PR #784, opened 2016-10-16
[WIP] Show emulations in html report

wardpeet:
Show emulation inside the html report.
It still needs some styling but what do you guys think?
**Note:  Started from cpu branch**

image

Issue #568, opened 2016-08-09
Report: no clear indication that results are for a 3g connection


ebidel:
It's unclear that the results are for a simulated 3g connection. 

Description text at the top of the report would be helpful that indicates the report is for:
- the first run of the page
- on a 3g simulated connection, with the exact network conditions that were simulated
- anything else I'm forgetting.

hsablonniere:
Should we also specify the device viewport that was "simulated"?

ebidel:
Definitely. All the settings should be included in the description. When it
shows a report, there should be no questions on what it just did :)

ebidel:
I think we should prioritize this now that the new report is in. https://github.com/GoogleChrome/lighthouse/pull/784 adds emulation info to the left nav, but I don't think that UI is ideal. 

@chowse would you be able to help us with designs? Perhaps some small text in the new top nav or n the footer?

ebidel:
This info is now behind a dropdown at the top of the report:

![screen shot 2017-02-01 at 11 15 32 am](https://cloud.githubusercontent.com/assets/238208/22522249/c6ca013c-e86f-11e6-961a-1134854cf3f2.png)

PR #1485, opened 2017-01-17
Show runtime configuration on report

WeiweiAtGit:
Added a section at the end of the report page which shows the runtime configuration of this report (e.g. disable CUP throttling).
Screenshot [updated]:
current-config

Will add some test cases later.

### DIFF
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -217,7 +217,6 @@ body {
 }
 
 .report-body__aggregations-container {
-  padding-top: var(--report-header-height);
   will-change: transform;
 }
 
@@ -327,7 +326,7 @@ body {
 .report-body__metadata {
   flex: 1 1 0;
   white-space: nowrap;
-  padding: 0 10px;
+  padding-right: 10px;
 }
 
 .report-body__buttons {
@@ -364,28 +363,22 @@ body {
   border: none;
 }
 
-.report-body__fixed-header-container {
+.report-body__header-container {
   background: #FAFAFA;
-  margin-left: var(--report-menu-width);
-  position: fixed;
-  will-change: transform;
-  z-index: 1;
-  max-width: calc( var(--report-width) - var(--report-menu-width));
-  width: calc(100vw - var(--report-menu-width));
 }
 
 .report-body__header {
   height: var(--report-header-height);
-  border-bottom: 1px solid #EBEBEB;
   display: flex;
   flex-direction: row;
   align-items: center;
   justify-content: flex-start;
   padding: 0 var(--heading-line-height) 0 0;
+  border-bottom: 1px solid #EBEBEB;
 }
 
 .report-body__header-toggle {
-  background-image: url('data:image/svg+xml;utf8,<svg fill="black" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg"><path d="M16.59 8.59L12 13.17 7.41 8.59 6 10l6 6 6-6z"/><path d="M0 0h24v24H0z" fill="none"/></svg>');
+  background-image: url('data:image/svg+xml;utf8,<svg fill="black" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg"><path d="M8.59 16.34l4.58-4.59-4.58-4.59L10 5.75l6 6-6 6z"/><path d="M0-.25h24v24H0z" fill="none"/></svg>');
   width: 24px;
   height: 24px;
   border: none;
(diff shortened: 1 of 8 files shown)


lighthouse-core/report/styles/report.css
```diff
@@ -834,6 +834,43 @@ body {
...
+
+.runtime-config__icon {
+  width: 18px;
+  height: 18px;
+  background-size: contain;
+  transform: translate(0, 1px);
+  background-position: center center;
+  background-repeat: no-repeat;
+  background-size: contain;
+  background-color: transparent;
```
samuelli: Move background-* into its on selector? `.report-body__icon, .runtime-config__icon`.

Perhaps consider using the shorthand for background too. https://developer.mozilla.org/en/docs/Web/CSS/background

brendankenny:
can you give a screenshot of this in context of the whole report?

WeiweiAtGit:
> can you give a screenshot of this in context of the whole report?

Sure : )
current-config-entire-page

WeiweiAtGit:
> Actually, the ticks/crosses & the negation "disabled" are super confusing; can we change that please?
> 
> Maybe use bullet points and change the text depending on the configuration:

I wanted it to be consistent with cli-flags (e.g. `--disable-cpu-throttling`). But yeah, we can remove the "disabled".
But I still prefer tick/cross over bullet points + "enabled/disabled", because tick/cross allows users to tell which option is enabled at a glance.

lighthouse-core/report/styles/report.css
WeiweiAtGit: Deleted `.report-body__icon {...}` because it's redundant.

ebidel: I think I saw @brendankenny mention this somewhere else, and I've been thinking the same...

Can we start a new perfx.css stylesheet and dump all the perfx stuff in that? The template can inject it dynamically based the context. We want to limit stylesheet bloat for CLI/CRX/Devtools users.

ebidel:
My initial feedback from looking at the screenshot is that the "Runtime configuration" section looks very similar to report results. Users might get confused and think these X's and checked items are part of their perf results (as opposed to the way the run was setup).

We may need to think about how to best [surface the emulation settings](https://github.com/GoogleChrome/lighthouse/issues/568) in the report. https://github.com/GoogleChrome/lighthouse/pull/784 was another approach of adding emulation settings to the report.

lighthouse-core/report/styles/report.css
WeiweiAtGit: > Can we start a new perfx.css stylesheet and dump all the perfx stuff in that?

Sure. I will separate all the perf-x related CSS code to another file. But for the particular case, I think runtime config section belongs to the main report not just perf-x.

WeiweiAtGit:
If this part needs to be under perf-x folder, ignore this PR for now. I will move it to perf-x folder and inject this to main report as a partial (when in `--interactive` mode) after #1455 is landed.

paulirish:
I think this should be in the core code, and not living in perf-x.

But I also have a different idea on UI which we can first discuss offline.

paulirish:
We discussed using a `<details><summary>Run configuration</summary> .... </details>` block for the config and putting in the top grey header of the report. Also "enabled/disabled"

lighthouse-core/runner.js
```diff
@@ -133,14 +133,24 @@ class Runner {
             a => Aggregate.aggregate(a, runResults.auditResults));
         }
 
+        const runtimeConfig = {
+          environment: [
+            {name: 'CPU Throttling', value: !opts.flags.disableCpuThrottling},
```
paulirish: let's put CPU throttling last of the three and add the extra detail hardcoded in here.

> Network throttling (150ms RTT, 1.6Mbps down, 750Kbps up): enabled
> Device emulation (Nexus 5X): enabled
> CPU throttling (5x slowdown): disabled

lighthouse-core/report/styles/report.css
```diff
@@ -363,23 +364,76 @@ body {
...
+  margin-left: 8px;
+}
+
+.report-body__fixed-header-container.expanded .report-body__header-toggle {
+  background-image: url('data:image/svg+xml;utf8,<svg fill="black" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg"><path d="M12 8l-6 6 1.41 1.41L12 10.83l4.59 4.58L18 14z"/><path d="M0 0h24v24H0z" fill="none"/></svg>');
+}
+
+.report-body__header-dropdown {
```
paulirish: I think it'd be better if the header just grew in side when you expand this.

So we could use a ▶ and it goes to ▼ when you click on it. `cursor: pointer` on the icon so its clearly actionable.

lighthouse-core/report/templates/report-template.html
```diff
@@ -31,27 +31,56 @@
...
+          <span class="export-section">
+            <button class="export-button js-export" title="Export report in different formats">Export...</button>
+            <ul class="export-dropdown">
+              <a href="#" data-action="print">Print...</a>
+              <a href="#" data-action="copy">Copy JSON...</a>
+              <a href="#" data-action="save-html">Save as HTML...</a>
+              <a href="#" data-action="save-json">Save as JSON...</a>
+            </ul>
+          </span>
+          <button class="report-body__icon share js-share" title="Share report on Github"></button>
```
paulirish: "GitHub"

ebidel:
Like that dropdown UI. Should we incorporate some sort of setting icon too?

https://material.io/icons/#ic_settings

lighthouse-core/report/styles/report.css
WeiweiAtGit: > I think it'd be better if the header just grew in side when you expand this.

Sorry, I don't have a clear idea what you mean here. Do you mean grow the header itself when we expand it (so it looks like the current UI but without the line between header bar and drop down)? If not, would you mind explaining in a little bit detail?

WeiweiAtGit:
> Should we incorporate some sort of setting icon too?

Hmm. But the top bar does not allow users to configure anything. I'm not sure if a setting icon fits here.

WeiweiAtGit:
Updated screenshots
When there are no blocked URL patterns:
header-dropdown

When there are some blocked URL patterns:
header-dropdown-with-url

No sure where to add runtime config in print view yet. Considered using the first version, but it feels wired if we have runtime config as an aggregation in print view but not in normal view.

WeiweiAtGit:
Changed the header bar from a fixed header to part of report content and improved UI.
current-config

Scroll down + Url blocking patterns
current-config-scrolled-down

lighthouse-core/report/styles/report.css
```diff
@@ -790,6 +839,7 @@ body {
 
 .export-section {
   position: relative;
+  display: none;
```
WeiweiAtGit: instead of hiding the export-button, hide the entire export-section when not in viewer context. So the dropdown won't cause the report page to be wider than it should be, which could make the report page scroll-able on x direction.

lighthouse-core/report/templates/report-template.html
```diff
@@ -69,10 +46,65 @@ <h1 class="menu__header-title">Lighthouse</h1>
...
+        </div>
+        <div class="report-body__header-content">
+          <section class="config-section">
+            <h2 class="config-section__title">Runtime Environment</h2>
+            <ul class="config-section__items">
+              {{#each runtimeConfig.environment }}
+              <li class="config-section__item">
+                <p class="config-section__desc">
+                  {{ this.name }} ({{ this.description }})
+                  <strong>: {{#if this.value }}Enabled{{else}}Disabled{{/if}}</strong>
```
samuelli: Don't bold `:` and remove the extra space before the `:` in the UI.

WeiweiAtGit: fixed. thx.

lighthouse-core/runner.js
```diff
@@ -260,6 +262,34 @@ class Runner {
...
+   * Get runtime configuration specified by the flags
+   * @param {!Object} flags
+   * @return {!Object} runtime config
+   */
+  static getRuntimeConfig(flags) {
+    const emulationDesc = emulation.getEmulationDesc();
+    const environment = [
+      {
+        name: 'CPU Throttling',
+        value: !flags.disableCpuThrottling,
```
paulirish: rename `value` to `enabled` ?

WeiweiAtGit: fixed

PR #1487, opened 2017-01-18
CRX: persistent background page -> event page

ebidel:
R: all

I'm not sure why the crx uses a persistent background page, but it doesn't need it. They are generally discouraged in favor of on-demand [event pages](https://developer.chrome.com/extensions/event_pages), which consume less memory in Chrome and are better for perf.

The changes here are mostly:
- Moving to an event page. This requires using `chrome.runtime.getBackgroundPage` which is async. I've created a promise and wrapped code that needs the background page in it.
- Moving code around (aka doing less work in `DOMContentLoaded`). We were defining a bunch of functions and delaying all work until `DOMContentLoaded`.

### DIFF
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -214,6 +214,7 @@ function initPopup(background) {
 
     siteURL = new URL(tabs[0].url);
 
+    // Show the user what URL is going to be tested.
     document.querySelector('header h2').textContent = siteURL.origin;
   });
 }
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -111,100 +111,108 @@ function createOptionItem(text, isChecked) {
   return listItem;
 }
 
-function onGenerateReportButtonClick() {
+/**
+ * Click event handler for Generate Report button.
+ * @param {!Window} background Reference to the extension's background page.
+ * @param {!Object<boolean>} selectedAggregations
+ */
+function onGenerateReportButtonClick(background, selectedAggregations) {
   showRunningSubpage();
 
   const feedbackEl = document.querySelector('.feedback');
   feedbackEl.textContent = '';
 
-  getBackgroundPage.then(background => {
-    background.loadSelectedAggregations().then(selectedAggregations => {
-      return background.runLighthouseInExtension({
-        flags: {
-          disableCpuThrottling: true
-        },
-        restoreCleanState: true
-      }, selectedAggregations);
-    }).catch(err => {
-      let message = err.message;
-      let includeReportLink = true;
+  background.runLighthouseInExtension({
+    flags: {
+      disableCpuThrottling: true
+    },
+    restoreCleanState: true
+  }, selectedAggregations).catch(err => {
+    let message = err.message;
+    let includeReportLink = true;
 
-      // Check for errors in how the user ran Lighthouse and replace with a more
-      // helpful message (and remove 'Report Error' link).
-      for (const [test, replacement] of Object.entries(NON_BUG_ERROR_MESSAGES)) {
-        if (message.includes(test)) {
-          message = replacement;
-          includeReportLink = false;
-          break;
-        }
+    // Check for errors in how the user ran Lighthouse and replace with a more
+    // helpful message (and remove 'Report Error' link).
+    for (const [test, replacement] of Object.entries(NON_BUG_ERROR_MESSAGES)) {
+      if (message.includes(test)) {
+        message = replacement;
+        includeReportLink = false;
+        break;
       }
+    }
 
-      feedbackEl.textContent = message;
+    feedbackEl.textContent = message;
 
-      if (includeReportLink) {
-        feedbackEl.className = 'feedback-error';
-        feedbackEl.appendChild(buildReportErrorLink(err));
-      }
+    if (includeReportLink) {
+      feedbackEl.className = 'feedback-error';
+      feedbackEl.appendChild(buildReportErrorLink(err));
+    }
 
-      hideRunningSubpage();
-      background.console.error(err);
-    });
+    hideRunningSubpage();
+    background.console.error(err);
   });
 }
 
 /**
  * Generates a document fragment containing a list of checkboxes and labels
  * for the aggregation categories.
- * @param {!Object<boolean>} selectedAggregations
  * @param {!Window} background Reference to the extension's background page.
- * @return {!DocumentFragment}
+ * @param {!Object<boolean>} selectedAggregations
  */
-function generateOptionsList(list, selectedAggregations, background) {
+function generateOptionsList(background, selectedAggregations) {
   const frag = document.createDocumentFragment();
 
-  const defaultAggregations = background.getDefaultAggregations();
-  defaultAggregations.forEach(aggregation => {
+  background.getDefaultAggregations().forEach(aggregation => {
     const isChecked = selectedAggregations[aggregation.name];
     frag.appendChild(createOptionItem(aggregation.name, isChecked));
   });
 
-  return frag;
+  const optionsList = document.querySelector('.options__list');
+  optionsList.appendChild(frag);
 }
 
 /**
  * Initializes the popup's state and UI elements.
  * @param {!Window} background Reference to the extension's background page.
  */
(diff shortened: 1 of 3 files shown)


brendankenny:
> I'm not sure why the crx uses a persistent background page

It might just be because of the generator we originally used to get an extension up and running.

How does this affect the extension debugging story?

brendankenny:
> How does this affect the extension debugging story?

looks like the answer is that the extension is persisted while being debugged. Not sure if that's good or bad for accurately debugging :)

ebidel:
> How does this affect the extension debugging story?

You mean dev'ing the extension? I only saw it go inactive a few times during testing but it fired right back up when I pressed the browser action button.

Apart from that, the background page should remain live as long as LH is doing its thing during a run.

brendankenny:
also @paulirish had a big refactor involving this file to make aggregation categories play well with devtools, so this will be fun to come home to :)

lighthouse-extension/app/src/popup.js
```diff
@@ -16,152 +16,117 @@
...
-    return /Chrome\/([0-9.]+)/.exec(navigator.userAgent)[1];
-  }
+function getChromeVersion() {
+  return /Chrome\/([0-9.]+)/.exec(navigator.userAgent)[1];
+}
 
-  function buildReportErrorLink(err) {
-    const reportErrorEl = document.createElement('a');
-    reportErrorEl.className = 'button button--report-error';
+function startSpinner() {
```
ebidel: done

lighthouse-extension/app/src/popup.js
```diff
@@ -16,152 +16,117 @@
...
+  statusEl.classList.add(subpageVisibleClass);
```
ebidel: done

lighthouse-extension/app/src/popup.js
```diff
@@ -16,152 +16,117 @@
...
+}
 
-    if (siteURL) {
-      qsBody += '**URL**: ' + siteURL + '\n';
-    }
-
-    qsBody += '**Error Message**: ' + err.message + '\n';
-    qsBody += '**Stack Trace**:\n ```' + err.stack + '```';
+function buildReportErrorLink(err) {
+  const MAX_ISSUE_ERROR_LENGTH = 60;
```
ebidel: done

lighthouse-extension/app/src/popup.js
```diff
@@ -186,26 +151,69 @@ document.addEventListener('DOMContentLoaded', _ => {
...
+  getBackgroundPage.then(background => {
+    statusEl = document.querySelector('.status');
+
+    /**
+     * Generates a document fragment containing a list of checkboxes and labels
+     * for the aggregation categories.
+     * @param {!Object<boolean>} selectedAggregations
+     * @return {!DocumentFragment}
+     */
+    function generateOptionsList(list, selectedAggregations) {
```
ebidel: done. but needed to add a background page param to make that work.

lighthouse-extension/app/src/popup.js
```diff
@@ -186,26 +151,69 @@ document.addEventListener('DOMContentLoaded', _ => {
...
+
+      return frag;
+    }
 
-  okButton.addEventListener('click', () => {
-    // Save selected aggregation categories on options page close.
-    const checkedAggregations = Array.from(optionsEl.querySelectorAll(':checked'))
-        .map(input => input.value);
-    background.saveSelectedAggregations(checkedAggregations);
+    if (background.isRunning()) {
```
ebidel: done

lighthouse-extension/app/src/popup.js
```diff
@@ -186,26 +151,69 @@ document.addEventListener('DOMContentLoaded', _ => {
...
-    }
+    background.listenForStatus(logstatus);
+    background.loadSelectedAggregations().then(aggregations => {
+      const frag = generateOptionsList(optionsList, aggregations);
+      optionsList.appendChild(frag);
+    });
 
-    siteURL = new URL(tabs[0].url);
-    siteNameEl.textContent = siteURL.origin;
+    const generateReportEl = document.getElementById('generate-report');
```
ebidel: done

lighthouse-extension/app/src/popup.js
```diff
@@ -186,26 +151,69 @@ document.addEventListener('DOMContentLoaded', _ => {
...
+    });
+
+    chrome.tabs.query({active: true, lastFocusedWindow: true}, function(tabs) {
+      if (tabs.length === 0) {
+        return;
+      }
+
+      siteURL = new URL(tabs[0].url);
+
+      document.querySelector('header h2').textContent = siteURL.origin;
```
ebidel: done

ebidel:
PTAL

lighthouse-extension/app/src/popup.js
```diff
@@ -182,15 +146,58 @@ document.addEventListener('DOMContentLoaded', _ => {
...
+  const frag = document.createDocumentFragment();
+
+  const defaultAggregations = background.getDefaultAggregations();
+  defaultAggregations.forEach(aggregation => {
+    const isChecked = selectedAggregations[aggregation.name];
+    frag.appendChild(createOptionItem(aggregation.name, isChecked));
+  });
+
+  return frag;
+}
```
brendankenny: this doesn't use the first `list` argument in here.

I was thinking you could also just have it fetch the background itself (though it doesn't really matter), and since `selectedAggregations` is only used in here (not in the parent), you could even absorb that too:

```js
/**
 * Generates a document fragment containing a list of checkboxes and labels
 * for the aggregation categories.
 * @return {!Promise<!DocumentFragment>}
 */
function generateOptionsList() {
  const frag = document.createDocumentFragment();

  return getBackgroundPage.then(background => {
    return background.loadSelectedAggregations().then(selectedAggregations => {
      const defaultAggregations = background.getDefaultAggregations();
      defaultAggregations.forEach(aggregation => {
        const isChecked = selectedAggregations[aggregation.name];
        frag.appendChild(createOptionItem(aggregation.name, isChecked));
      });
      return frag;
}
```

then below it would be
`generateOptionsList().then(frag => optionsList.appendChild(frag));`

brendankenny: (could still just pass in `background`, of course, since the caller already has a reference to it)

brendankenny:
more important than bikeshedding that :) did you see my earlier comment:

> The one difference I see is that when running lighthouse, clicking outside of the popup to close it, then reopening the popup it now does the very brief opacity transition to loading the status subpage every time, while before it appeared to be visible instantly. Fixable?

ebidel:
I decided to pass the `background` and the `selectedAggregations` promise. Felt more DRY and less wasteful then having extra ticks inside those functions:

```
background.loadSelectedAggregations().then(selectedAggregations => {
      generateOptionsList(background, selectedAggregations);

      const generateReportButton = document.getElementById('generate-report');
      generateReportButton.addEventListener('click', () => {
        onGenerateReportButtonClick(background, selectedAggregations);
      });
    });
```

I think the changes also fix the visual jank you were seeing. Because the background page is async, it's a bit harder to control the transitions when the popup first starts and you're already running LH. I think we always had that jank, but it was unnoticeable until now.

lighthouse-extension/app/src/popup.js
ebidel: redid this function a bit.

brendankenny:
ah, good call with `selectedAggregations`. One weird thing is I'm really not sure why `loadSelectedAggregations` is in the background page. Seems like it should be in the popup since that's where it's generated and used. But that can wait for another day

brendankenny:
> I think the changes also fix the visual jank you were seeing. Because the background page is async, it's a bit harder to control the transitions when the popup first starts and you're already running LH. I think we always had that jank, but it was unnoticeable until now.

Yeah, this traded it for a fade in at first launch, but it's pretty unnoticeable either way. I like this better, though others can argue :) I guess we could put the transition in a class we add after load, but I wonder if a flash of the wrong subpage is better than a fast fadeout of the wrong subpage.

paulirish:
> also @paulirish had a big refactor involving this file to make aggregation categories play well with devtools, so this will be fun to come home to :)

INDEED. 😁 

but overall a worthwhile change, so it's all good.

PR #1489, opened 2017-01-18
Report-in-Firefox fixes

### LINKED ISSUE #1124: Report: FF rendering issues
Lots of cleanup to do in FF.

- The Beta banner is gone
- the check/x's are gone in FF. That's probably because we're using some `-webkit-mask-image` action. They should really just be background images with `background-size: contain`.
- The help toggles are checkboxes instead of images

screen shot 2016-12-08 at 10 37 46 am

XhmikosR:
Opening this since I'm not sure when I'll have more time to fix the remaining issues.

| Before  | After |
| ------------- | ------------- |
| before | after |

If merged, fixes #1124.

/CC @ebidel @wardpeet: feel free to push to my branch. I also kept the original patch which URL encoded the SVGs and worked everywhere but it's a little messy solution.

Also, there's a discrepancy between Chrome and Firefox: in Firefox the badges don't look properly vertically aligned but in Chrome they look fine. Not sure how to fix that only for Firefox.


XhmikosR:
**Before:**
before

---

**After:**
after

wardpeet:
will have a look ^^ if @ebidel doesn't beat me to it 😃

XhmikosR:
Also note that this patch doesn't fix the checkbox issue; I thought I'd leave this for later.

XhmikosR:
I'm almost done with the checkbox, but I'm having an issue with the position of the help text :/

test


Should I push that too?

lighthouse-core/formatters/partials/critical-request-chains.html
```diff
@@ -1,62 +1,3 @@
-<style>
```
ebidel: Can we restore these files and keep the related styles within the partials? That doesn't seem necessary for this PR?

ebidel:
Let's do checkboxes in another PR.

lighthouse-core/formatters/partials/critical-request-chains.html
XhmikosR: Done. I just feel that having random styles in many places isn't good. But, yeah, it was unrelated change to this PR.

Issue #941, opened 2016-11-14
Proposed refactor: Check for required artifacts before calling audit()


paulirish:
The very top of every audit file's audit() function has a block [like this](https://github.com/GoogleChrome/lighthouse/blob/0ac3fe9f683d207270ccf6b23f067f1d91889037/lighthouse-core/audits/dobetterweb/uses-single-use-event-listeners.js#L49-L57):

```js
    if (typeof artifacts.EventListeners === 'undefined' ||
        artifacts.EventListeners === -1) {
      return SingleUseEventsAudit.generateAuditResult({
        rawValue: -1,
        debugString: 'EventListeners gatherer did not run'
      });
    } else if (artifacts.EventListeners.rawValue === -1) {
      return SingleUseEventsAudit.generateAuditResult(artifacts.EventListeners);
    }
```

It's extremely rote and decreases the signal:noise ratio of the source. I think we could get rid of it.

We could alternatively handle this more automatically. Something like:

1.  For each audit, `runner` grabs its `requiredArtifacts` property.
2. `runner` checks that each requiredArtifact exists on `artifacts` and if it doesn't, `generateAuditResult` with our -1 error
3. `runner` also checks if the artifact value is -1, and if so, `generateAuditResult` with the curried `debugString` on the artifact

This would add some lines to runner.js and drop 5-10 lines from every audit we have.  I bet it would uncover a bug or two as well. AFAIK, currently we don't heavily depend on `requiredArtifacts` being correct or anything, but we could and should, IMO.

Thoughts?

brendankenny:
sounds GREAT. I was going to make a separate bug about standardizing on an error/`debugString` format from gatherers and what audits should do with that. This would take care of all of that and enforce it.

The only thing left to add would be a note in the readme about what gatherers should return :)

brendankenny:
> This would add some lines to runner.js and drop 5-10 lines from every audit we have

and add a few tests to `runner-test.js` but drop one or two tests from every audit :) win/win/win

paulirish:
:+1:

Anyone wanna take this? It's a good first bug. And a good second bug. :)

jillesme:
Did anyone say good second bug? I'll take it!

jillesme:
I'm working on this. 

Currently I'm thinking something like (runner.js):

```js
...
const status = `Evaluating: ${audit.meta.description}`;
const requiredArtifacts = audit.meta.requiredArtifacts;
[...]

PR #1492, opened 2017-01-18
CRX: disable extensions before running lighthouse

### LINKED ISSUE #1483: Extension: disabled other extensions when running Lighthouse
To help [perf differences](https://github.com/GoogleChrome/lighthouse/issues/1449) between the crx and cli and reduce bugs like https://github.com/GoogleChrome/lighthouse/issues/1282, it would be useful to disable some/all the user's chrome extensions that can effect page load.

From the CRX docs, we can use the "management" permission to get a list of installed extensions and disable all, or only those that could muck with the page (e.g. use the`content_script` permission).

ref: https://developer.chrome.com/extensions/management

```
function enableAllOtherExtensions(enable) {
  installedExtensions.forEach(info => {
    chrome.management.setEnabled(info.id, enable);
  });
}

let installedExtensions = [];

// Run when background page fires up.
chrome.management.getAll(installs => {
  chrome.management.getSelf(lighthouseCrxInfo => {
    installedExtensions = installs.filter(info => {
      return info.id !== lighthouseCrxInfo.id && info.type === 'extension';
    });
  });
});

// Run when user initiates LH run:
enableAllOtherExtensions(true);

// Run when LH finishes run:
enableAllOtherExtensions(false);

```

ebidel:
R: all

Fixes #1483.

### DIFF
--- a/lighthouse-extension/app/src/lighthouse-background.js
+++ b/lighthouse-extension/app/src/lighthouse-background.js
@@ -41,6 +41,9 @@ const _flatten = arr => [].concat(...arr);
  * @param {!Promise}
  */
 function enableOtherChromeExtensions(enable) {
+  const str = enable ? 'enabling' : 'disabling';
+  log.warn('Chrome', `${str} ${installedExtensions.length} extensions.`);
+
   return new Promise(resolve => {
     let remaining = installedExtensions.length;
     installedExtensions.forEach(info => {
@@ -147,28 +150,16 @@ window.runLighthouseForConnection = function(connection, url, options, requested
   lighthouseIsRunning = true;
   updateBadgeUI(url);
 
-  return enableOtherChromeExtensions(false)
-      .then(_ => Runner.run(connection, runOptions)) // Run Lighthouse.
+  return Runner.run(connection, runOptions) // Run Lighthouse.
       .then(result => {
         lighthouseIsRunning = false;
         updateBadgeUI();
         filterOutArtifacts(result);
-
-        // Deliver results even if enabling extensions fails.
-        enableOtherChromeExtensions(true).catch(err => {
-          log.warn('Chrome', `Could not disable some extensions. ${err.message}`);
-        });
-
         return result;
       })
       .catch(err => {
         lighthouseIsRunning = false;
         updateBadgeUI();
-
-        enableOtherChromeExtensions(true).catch(err => {
-          log.warn('Chrome', `Could not enable/disable some extensions. ${err.message}`);
-        });
-
         throw err;
       });
 };
@@ -182,11 +173,22 @@ window.runLighthouseInExtension = function(options, requestedAggregations) {
   // Default to 'info' logging level.
   log.setLevel('info');
   const connection = new ExtensionProtocol();
-  return connection.getCurrentTabURL()
+  return enableOtherChromeExtensions(false)
+    .then(_ => connection.getCurrentTabURL())
     .then(url => window.runLighthouseForConnection(connection, url, options, requestedAggregations))
     .then(results => {
+      // Deliver results even if enabling extensions fails.
+      enableOtherChromeExtensions(true).catch(err => {
+        log.warn('Chrome', `Could not enable some extensions. ${err.message}`);
+      });
+
       const blobURL = window.createReportPageAsBlob(results, 'extension');
       chrome.tabs.create({url: blobURL});
+    }).catch(err => {
+      enableOtherChromeExtensions(true).catch(err => {
+        log.warn('Chrome', `Could not enable/disable some extensions. ${err.message}`);
+      });
+      throw err;
     });
 };
 
@@ -311,18 +313,19 @@ window.isRunning = function() {
   return lighthouseIsRunning;
 };
 
-// Get list of installed extensions that are enabled and can be disabled.
-// Extensions are not allowed to be disabled if they are under an admin policy.
-chrome.management.getAll(installs => {
-  chrome.management.getSelf(lighthouseCrxInfo => {
-    installedExtensions = installs.filter(info => {
-      return info.id !== lighthouseCrxInfo.id && info.type === 'extension' &&
-             info.enabled && info.mayDisable;
+// Run when in extension context, but not in devtools.
+if (window.chrome && chrome.runtime) {
+  // Get list of installed extensions that are enabled and can be disabled.
+  // Extensions are not allowed to be disabled if they are under an admin policy.
+  chrome.management.getAll(installs => {
+    chrome.management.getSelf(lighthouseCrxInfo => {
+      installedExtensions = installs.filter(info => {
+        return info.id !== lighthouseCrxInfo.id && info.type === 'extension' &&
+               info.enabled && info.mayDisable;
+      });
     });
   });
-});
 
-if (window.chrome && chrome.runtime) {
   chrome.runtime.onInstalled.addListener(details => {
     if (details.previousVersion) {
       // eslint-disable-next-line no-console
--- a/lighthouse-extension/app/src/lighthouse-background.js
+++ b/lighthouse-extension/app/src/lighthouse-background.js
@@ -42,11 +42,16 @@ const _flatten = arr => [].concat(...arr);
  */
 function enableOtherChromeExtensions(enable) {
   const str = enable ? 'enabling' : 'disabling';
-  log.warn('Chrome', `${str} ${installedExtensions.length} extensions.`);
+  log.log('Chrome', `${str} ${installedExtensions.length} extensions.`);
 
   return Promise.all(installedExtensions.map(info => {
-    return new Promise(resolve => {
-      chrome.management.setEnabled(info.id, enable, resolve);
+    return new Promise((resolve, reject) => {
+      chrome.management.setEnabled(info.id, enable, _ => {
+        if (chrome.runtime.lastError) {
+          reject(chrome.runtime.lastError);
+        }
+        resolve();
+      });
     });
   }));
 }
@@ -145,17 +150,17 @@ window.runLighthouseForConnection = function(connection, url, options, requested
   updateBadgeUI(url);
 
   return Runner.run(connection, runOptions) // Run Lighthouse.
-      .then(result => {
-        lighthouseIsRunning = false;
-        updateBadgeUI();
-        filterOutArtifacts(result);
-        return result;
-      })
-      .catch(err => {
-        lighthouseIsRunning = false;
-        updateBadgeUI();
-        throw err;
-      });
+    .then(result => {
+      lighthouseIsRunning = false;
+      updateBadgeUI();
+      filterOutArtifacts(result);
+      return result;
+    })
+    .catch(err => {
+      lighthouseIsRunning = false;
+      updateBadgeUI();
+      throw err;
+    });
 };
 
 /**
@@ -171,17 +176,11 @@ window.runLighthouseInExtension = function(options, requestedAggregations) {
     .then(_ => connection.getCurrentTabURL())
     .then(url => window.runLighthouseForConnection(connection, url, options, requestedAggregations))
     .then(results => {
-      // Deliver results even if enabling extensions fails.
-      enableOtherChromeExtensions(true).catch(err => {
-        log.warn('Chrome', `Could not enable some extensions. ${err.message}`);
-      });
-
+      enableOtherChromeExtensions(true);
       const blobURL = window.createReportPageAsBlob(results, 'extension');
       chrome.tabs.create({url: blobURL});
     }).catch(err => {
-      enableOtherChromeExtensions(true).catch(err => {
-        log.warn('Chrome', `Could not enable/disable some extensions. ${err.message}`);
-      });
+      enableOtherChromeExtensions(true);
       throw err;
     });
 };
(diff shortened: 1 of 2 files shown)


lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -26,11 +26,34 @@ const log = require('../../../lighthouse-core/lib/log');
...
+ * of the user's extension is created when the background page is started.
+ * @param {!boolean} enable If true, enables all other installed extensions.
+ *     False disables them.
+ * @param {!Promise}
+ */
+function enableOtherChromeExtensions(enable) {
+  return new Promise(resolve => {
+    let remaining = installedExtensions.length;
+    installedExtensions.forEach(info => {
+      chrome.management.setEnabled(info.id, enable, _ => {
```
brendankenny: since the extension is running this after a button click, I assume this gets around the docs warning about "In most cases this function must be called in the context of a user gesture"?

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -124,19 +147,30 @@ window.runLighthouseForConnection = function(connection, url, options, requested
...
-      updateBadgeUI();
-      filterOutArtifacts(result);
-      return result;
-    })
-    .catch(err => {
-      lighthouseIsRunning = false;
-      updateBadgeUI();
-      throw err;
-    });
+  return enableOtherChromeExtensions(false)
```
brendankenny: needs to be in `runLighthouseInExtension` as devtools runs `runLighthouseForConnection` (through `runLighthouseInWorker`) as well

(`updateBadgeUI` probably should be too. It gates itself internally by returning if no `chrome.runtime`, but that could quickly get confusing if we had a bunch of noop functions in the devtools path)

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -124,19 +147,30 @@ window.runLighthouseForConnection = function(connection, url, options, requested
...
+      .then(_ => Runner.run(connection, runOptions)) // Run Lighthouse.
+      .then(result => {
+        lighthouseIsRunning = false;
+        updateBadgeUI();
+        filterOutArtifacts(result);
+
+        // Deliver results even if enabling extensions fails.
+        enableOtherChromeExtensions(true).catch(err => {
+          log.warn('Chrome', `Could not disable some extensions. ${err.message}`);
```
brendankenny: `disable` -> `enable`? This seems like a big problem if it ever happens...worth bubbling up to the "Report Error" level?

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -124,19 +147,30 @@ window.runLighthouseForConnection = function(connection, url, options, requested
...
+        });
+
+        return result;
+      })
+      .catch(err => {
+        lighthouseIsRunning = false;
+        updateBadgeUI();
+
+        enableOtherChromeExtensions(true).catch(err => {
+          log.warn('Chrome', `Could not enable/disable some extensions. ${err.message}`);
```
brendankenny: feels like we should track what state we weren't able to achieve (mostly in case we can't reenable extensions)? Would probably need a separate variable from `lighthouseIsRunning` to track that though

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -277,6 +311,17 @@ window.isRunning = function() {
   return lighthouseIsRunning;
 };
 
+// Get list of installed extensions that are enabled and can be disabled.
+// Extensions are not allowed to be disabled if they are under an admin policy.
+chrome.management.getAll(installs => {
```
brendankenny: also needs to not run in devtools. Can just nest in the if statement below in this case.

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -26,11 +26,34 @@ const log = require('../../../lighthouse-core/lib/log');
...
+ * Enables or disables all other installed chrome extensions. The initial list
+ * of the user's extension is created when the background page is started.
+ * @param {!boolean} enable If true, enables all other installed extensions.
+ *     False disables them.
+ * @param {!Promise}
+ */
+function enableOtherChromeExtensions(enable) {
+  return new Promise(resolve => {
+    let remaining = installedExtensions.length;
+    installedExtensions.forEach(info => {
```
brendankenny: could also do 
```js
return Promise.all(installedExtensions.map(info => {
  return new Promise(resolve => {
    chrome.management.setEnabled(info.id, enable, resolve);
  });
})
```

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -26,11 +26,34 @@ const log = require('../../../lighthouse-core/lib/log');
...
+
+/**
+ * Enables or disables all other installed chrome extensions. The initial list
+ * of the user's extension is created when the background page is started.
+ * @param {!boolean} enable If true, enables all other installed extensions.
+ *     False disables them.
+ * @param {!Promise}
+ */
+function enableOtherChromeExtensions(enable) {
+  return new Promise(resolve => {
```
brendankenny: will you add a `log.log` or `log.verbose` about enabling or disabling extensions? Worth including `installedExtensions.length` in the log message?

brendankenny:
how does the permission prompt look for `management`?

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

lighthouse-extension/app/src/lighthouse-background.js
ebidel: What are you thinking? And how would it surface to the user?

lighthouse-extension/app/src/lighthouse-background.js
ebidel: Done. For the other, I don't think it's a good UX if LH fails because an extension fails to enable. That was my intention at least. I'm not sure how enabling _could_ fail, but users can still get results if they do.

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

lighthouse-extension/app/src/lighthouse-background.js
ebidel: Yea. Still falls within that click.

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

ebidel:
The permission prompt looks like this:

screen shot 2017-01-20 at 3 21 51 pm

lighthouse-extension/app/src/lighthouse-background.js
brendankenny: good question. I don't actually know a great way, but I'm concerned there will be some corner case of permissions or something where we end up silently disabling some people's extensions and we won't know about it until someone notices (or we get error reporting integrated)

ebidel:
PTAL

lighthouse-extension/app/src/lighthouse-background.js
ebidel: It would be nice to know if/when the enabling/disabling can fail, but it may be risky to make that fatal. I really wish we had analytics :\

As part of your error refactoring, do you want to think about ways to surface top-level errors like this (e.g. non-gatherer errors) to the report so people can file bugs on them?

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

lighthouse-extension/app/src/lighthouse-background.js
brendankenny: #941 is really only about errors in the gatherer->artifact->audit->report pipeline. @patrickhulce will have to figure this out to get error reporting working, though.

We could make a master bug, though

ebidel: https://github.com/GoogleChrome/lighthouse/issues/1512

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -124,19 +144,18 @@ window.runLighthouseForConnection = function(connection, url, options, requested
...
-      updateBadgeUI();
-      filterOutArtifacts(result);
-      return result;
-    })
-    .catch(err => {
-      lighthouseIsRunning = false;
-      updateBadgeUI();
-      throw err;
-    });
+  return Runner.run(connection, runOptions) // Run Lighthouse.
```
brendankenny: can revert this block of changes?

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -148,11 +167,22 @@ window.runLighthouseInExtension = function(options, requestedAggregations) {
...
   log.setLevel('info');
   const connection = new ExtensionProtocol();
-  return connection.getCurrentTabURL()
+  return enableOtherChromeExtensions(false)
+    .then(_ => connection.getCurrentTabURL())
     .then(url => window.runLighthouseForConnection(connection, url, options, requestedAggregations))
     .then(results => {
+      // Deliver results even if enabling extensions fails.
+      enableOtherChromeExtensions(true).catch(err => {
+        log.warn('Chrome', `Could not enable some extensions. ${err.message}`);
```
brendankenny: should upgrade to `log.error`, but will `enableOtherChromeExtensions` ever actually reject? [The docs](https://developer.chrome.com/extensions/management#method-setEnabled) mention that it *could* fail but don't mention *how* it will fail. Presumably it won't throw but will do the callback with an error or set `lastError` or something? We should hook that up to the `reject` callback

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -148,11 +167,22 @@ window.runLighthouseInExtension = function(options, requestedAggregations) {
...
+      });
+
       const blobURL = window.createReportPageAsBlob(results, 'extension');
       chrome.tabs.create({url: blobURL});
+    }).catch(err => {
+      enableOtherChromeExtensions(true).catch(err => {
+        log.warn('Chrome', `Could not enable/disable some extensions. ${err.message}`);
```
brendankenny: `log.error`

lighthouse-extension/app/src/lighthouse-background.js
ebidel: It's just a proper 4-space indent. https://github.com/GoogleChrome/lighthouse/pull/1492/files?w=1

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

lighthouse-extension/app/src/lighthouse-background.js
ebidel: I think I remember it throwing, but will check. If not, we can monitor `lastError`.

lighthouse-extension/app/src/lighthouse-background.js
brendankenny: only two spaces for chaining though, just to keep everyone on their toes :)

brendankenny: (I think this is one place we diverge from Google style, but it's so common in open source JS that we've pretty much unquestioningly done it throughout LH)

lighthouse-extension/app/src/lighthouse-background.js
brendankenny: looks like it does throw, but the error is from not checking `lastError` :) "`Unchecked runtime.lastError while running management.setEnabled`"

ebidel: done. this is now in the promise flow.

lighthouse-extension/app/src/lighthouse-background.js
ebidel: [We're not consistent](https://github.com/GoogleChrome/lighthouse/search?utf8=%E2%9C%93&q=++.then%28&type=Code), but I'll revert to be consistent with the file. 

If we want to enforce something other than Goog style, there really should be an eslint rule.

ebidel:
I've changed is so if an extension cannot be disabled, the chain rejects and we'll know about it via the extension bug report button. We should monitor this pretty closely on the next release. If there are errors coming in,  we'll want to adjust.

ebidel:
PTAL

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -148,11 +172,16 @@ window.runLighthouseInExtension = function(options, requestedAggregations) {
   // Default to 'info' logging level.
   log.setLevel('info');
   const connection = new ExtensionProtocol();
-  return connection.getCurrentTabURL()
+  return enableOtherChromeExtensions(false)
+    .then(_ => connection.getCurrentTabURL())
     .then(url => window.runLighthouseForConnection(connection, url, options, requestedAggregations))
     .then(results => {
+      enableOtherChromeExtensions(true);
```
brendankenny: these have to be returned for the rejection to get into the promise chain. 
```js
.then(results => {
  return enableOtherChromeExtensions(true).then(_ => {
    const blobURL = window.createReportPageAsBlob(results, 'extension');
    chrome.tabs.create({url: blobURL});
  });
}).catch(err => {
  return enableOtherChromeExtensions(true).then(_ => {
    throw err;
  });
});
```
maybe?

ebidel: done

ebidel:
PTAL

Issue #1456, opened 2017-01-12
Audit: CSS used selectors improvements


ebidel:
Seeing this in the DBW tester:

<img width="620" alt="screen shot 2017-01-11 at 8 02 02 pm" src="https://cloud.githubusercontent.com/assets/238208/21876567/d8ad060c-d838-11e6-8577-87692d5a48b2.png">

What about listing the rules that are not used instead of file snippets? Also, should URLs that have "0% unused rules" show up in the list?

patrickhulce:
@ebidel showing the unused rules was the original behavior but the consensus was that it was too overwhelming and so instead it just shows you the contents of the file (also helps in cases where there are many inline scripts in sites built with webpack). 0% rules used should show up but only if there's content and it has rules IMO :) will fix

ebidel:
Yea, just trying to think what is most useful for developers to take action from.

What I'm finding confusing now is that we're always showing general code snippets from the CSS file but that snippet won't contain the unused rule(s). That's a departure from other places in the report we only show violations.

Guess I'm wondering how valuable the snippets are if they're not violations? For example, below I'm using all these rules (chromestatus.com):

<img width="636" alt="screen shot 2017-01-13 at 10 41 25 am" src="https://cloud.githubusercontent.com/assets/238208/21941284/db425c24-d97c-11e6-9334-294ce86ddd70.png">

For the "100% usage entry"
- it makes it seem like I should remove the `[hidden]` rule.
- I'm already 100% awesome on that file. Maybe we shouldn't show a snippet at all?

patrickhulce:
I think the snippets sole value is for identification of the stylesheet in cases where the URL is not sufficient and in all of these cases you've mentioned the URL is relatively self-explanatory and thus the snippet is just noise. I'm somewhat in favor of bringing back listing the unused selectors but truncating the output at some threshold (maybe just first 50 or something?), or maybe just showing the stylesheet snippet for inline stylesheets where there's no useful URL information?

ebidel:
> maybe just showing the stylesheet snippet for inline stylesheets where there's no useful URL information?

+1. Good balance of noise / usefulness ratio. I have the same problem with the deprecations audit. Some deprecations don't return url/line combos. I used "Unable to determine URL":
https://github.com/GoogleChrome/lighthouse/pull/1470#issue-200767440, but we can figure out something better if you want.

brendankenny:
fixed by #1496

PR #1496, opened 2017-01-19
fix(unused-css-rules): remove stylesheet preview for sheets with URL

patrickhulce:
R: @ebidel 

addresses #1456

### DIFF
--- a/lighthouse-core/audits/unused-css-rules.js
+++ b/lighthouse-core/audits/unused-css-rules.js
@@ -33,7 +33,7 @@ class UnusedCSSRules extends Audit {
       description: 'Site does not have more than 10% unused CSS',
       helpText: 'Remove unused rules from stylesheets to reduce unnecessary ' +
           'bytes consumed by network activity. [Learn more](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery)',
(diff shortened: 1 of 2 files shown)


patrickhulce:
Before:
image


After:
image

lighthouse-core/audits/unused-css-rules.js
```diff
@@ -110,11 +111,16 @@ class UnusedCSSRules extends Audit {
...
-    return {
-      url: stylesheetInfo.header.sourceURL || 'inline',
-      label: `${percentUsed}% rules used`,
-      code: contentPreview.trim(),
-    };
+    let code;
+    let url = stylesheetInfo.header.sourceURL;
+    const label = `${percentUsed}% rules used`;
+
+    if (!url || url === pageUrl) {
```
ebidel: One thing I was curious about for later. When is url blank vs. the same as the page? Can you add a  comment?

patrickhulce: Oh whoops Brendan merged before I could :) it seems to be if it was in the original document or added later with script, which maybe a more sophisticated version later on could identify the URL of the script that added it too? I'll add the TODO in my upcoming rules -> bytes conversion PR for CSS.

PR #1122, opened 2016-12-08
refactor: get artifacts directly from running gatherer methods

brendankenny:
R: @paulirish @patrickhulce 

next part of #941

- Removes `this.artifact` from gatherers and instead have gatherers return the artifact directly in whatever value is returned last from `beforePass()`, `pass()`, or `afterPass()`.
- When gatherers reject on an error, we have a new opt-in property `recoverable` on the error that gather-runner checks. If set to `true`, it will turn that into an "error artifact", equivalent to the many versions of -1 artifacts we have today.
- If `recoverable` isn't present, it throws the error and lighthouse exits on the error. By making it explicit when a gatherer expects an error, this should help catch more errors than our current system (which absorbs all errors in the gatherer stage)
- gatherer's methods can return (or throw) synchronously or asynchronously

This PR doesn't actually take advantage of recoverable errors yet, as it was already getting really big. Merely implements support for them, adds unit tests for them, and gets rid of `gatherer.artifact` on all the current artifacts. So most work here is in `gather-runner.js` and `gather-runner-test.js`.

Feel free to bikeshed/propose other control flows.

Next PR will delete all the -1 nonsense from gatherers and replace with this system when we're certain this is the direction we want to go.

PR #1498, opened 2017-01-20
Start using errors instead of -1 for error artifacts

brendankenny:
@ebidel @patrickhulce 

This is the next (and sort of last) major step of #941, actually using the recoverable Error infrastructure set up in #1122. Most easily reviewed by commit:

The first adds the last check needed in runner: if an artifact required for an audit is an Error it no longer runs the audit at all, instead directly generating an error result on behalf of the audit. This lets us leave out the -1 checks from all future audits.

This is opt-in for now while we feel this out, but also included are four gatherer conversions to just throwing an error when there's an error and the simplifications that allows in the audits that depend on them. The first two (meta viewport and theme-color) are simple, the last commit is giving that treatment to the slightly more complicated `Styles` and `CSSUsage` gatherers and the two audits that need them.

I think this is looking good so far. Thoughts?

### DIFF
--- a/lighthouse-core/gather/gatherers/css-usage.js
+++ b/lighthouse-core/gather/gatherers/css-usage.js
@@ -29,8 +29,9 @@ class CSSUsage extends Gatherer {
       .catch(err => {
         // TODO(phulce): Remove this once CSS usage hits stable
         if (/startRuleUsageTracking/.test(err.message)) {
-          this.failure = 'CSS Usage tracking requires Chrome \u2265 56';
-          return;
+          const usageError = new Error('CSS Usage tracking requires Chrome \u2265 56');
+          usageError.recoverable = true;
+          throw usageError;
         }
 
         throw err;
@@ -44,11 +45,6 @@ class CSSUsage extends Gatherer {
       return driver.sendCommand('CSS.disable')
         .then(_ => driver.sendCommand('DOM.disable'))
         .then(_ => results.ruleUsage);
-    }).catch(err => {
-      return {
-        rawValue: -1,
-        debugString: this.failure || err,
-      };
     });
   }
 }
--- a/lighthouse-core/test/audits/viewport-test.js
+++ b/lighthouse-core/test/audits/viewport-test.js
@@ -18,20 +18,12 @@
 const Audit = require('../../audits/viewport.js');
 const assert = require('assert');
 
-/* global describe, it*/
+/* eslint-env mocha */
 
 describe('Mobile-friendly: viewport audit', () => {
-  it('fails when no input present', () => {
-    const audit = Audit.audit({
-      Viewport: -1
-    });
-    assert.equal(audit.rawValue, -1);
-    assert.equal(audit.debugString, 'Error in determining viewport');
-  });
-
   it('fails when HTML does not contain a viewport meta tag', () => {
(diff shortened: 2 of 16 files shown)


lighthouse-core/test/runner-test.js
```diff
@@ -127,39 +127,68 @@ describe('Runner', () => {
...
+    it('outputs an error audit result when required artifact was an Error', () => {
+      const url = 'https://example.com';
+
+      const config = new Config({
+        audits: [
+          'is-on-https'
+        ],
+
+        artifacts: {
+          HTTPS: {
```
patrickhulce: can you just change this to `null` and maybe move/add the comment about it being replaced here. was pretty confusing to skim :)

lighthouse-core/test/audits/viewport-test.js
```diff
@@ -18,20 +18,12 @@
...
-      Viewport: -1
-    });
-    assert.equal(audit.rawValue, -1);
-    assert.equal(audit.debugString, 'Error in determining viewport');
-  });
-
   it('fails when HTML does not contain a viewport meta tag', () => {
     return assert.equal(Audit.audit({
-      Viewport: ''
+      Viewport: null
```
patrickhulce: is this just the actual output when it doesn't contain one?

lighthouse-core/gather/gatherers/css-usage.js
```diff
@@ -29,8 +29,9 @@ class CSSUsage extends Gatherer {
       .catch(err => {
         // TODO(phulce): Remove this once CSS usage hits stable
         if (/startRuleUsageTracking/.test(err.message)) {
-          this.failure = 'CSS Usage tracking requires Chrome \u2265 56';
-          return;
+          const usageError = new Error('CSS Usage tracking requires Chrome \u2265 56');
+          usageError.recoverable = true;
+          throw usageError;
```
patrickhulce: nice!

lighthouse-core/gather/gatherers/css-usage.js
```diff
@@ -44,11 +45,6 @@ class CSSUsage extends Gatherer {
       return driver.sendCommand('CSS.disable')
         .then(_ => driver.sendCommand('DOM.disable'))
         .then(_ => results.ruleUsage);
-    }).catch(err => {
```
patrickhulce: The `afterPass` won't be called when the recoverable error is thrown right? Otherwise we'll fail on stable Chrome again

brendankenny: yeah, I thought of this immediately after opening the PR and it will fail :) need to add a nicer fix

lighthouse-core/test/audits/viewport-test.js
brendankenny: > is this just the actual output when it doesn't contain one?

Yeah, this would be a great place for a type checker :) `driver.querySelector()` and `element.getAttribute()` both return null if the query fails, so it really is a string or `null`

ebidel:
@brendankenny do you still want a review of this or are you thinking things will change if we invert (to opt-in fatal errors)?

brendankenny:
it will be very similar but yeah, probably worth waiting

PR #1505, opened 2017-01-21
Add table list formatter for HTML audits

ebidel:
R: all

This adds a new `<table>` formatter for HTML reports which increases clarity, fixes some overflow UI bugs we had with text, and will allow us to more easily add additional extended info in the future (e.g. extra properties from the gatherers, RUM data column, etc.)

An audit uses the formatter like so:

    const createTable = Formatter.getByName('table').createTable;
    ...
    extendedInfo: {
        formatter: Formatter.SUPPORTED_FORMATS.TABLE,
        value: {
          table: createTable({url: 'URL', protocol: 'Protocol'}, resources),
          results: resources
        }
    }

Notes:
- All the cells are `overflow: auto` so you can scroll them to reveal more of the url/code snippets
- the original `results` are also being returned in `extendedInfo` so unit testing is easier and so we have the full objects as returned by the gatherer. Might want to change this in the future b/c it does mean the JSON results will be larger than they need to be.
- This PR was getting big, so I only did 3 audits. When we're happy with this, we can move over other URLIST audits to use the TABLE formatter.

After: 

screen shot 2017-01-20 at 2 01 54 pm

screen shot 2017-01-20 at 2 01 58 pm

screen shot 2017-01-20 at 2 02 01 pm

Before:

screen shot 2017-01-20 at 4 04 18 pm
screen shot 2017-01-20 at 4 04 24 pm

### DIFF
--- a/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
+++ b/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
@@ -103,12 +103,10 @@ class UsesOptimizedImages extends Audit {
 
       totalWastedBytes += webpSavings.bytes;
       results.push({
-        cols: [
-          url,
-          `${originalKb} KB`,
-          `${webpSavings ? webpSavings.percent + '%' : ''}`,
-          `${jpegSavings ? jpegSavings.percent + '%' : ''}`
-        ]
+        url,
+        total: `${originalKb} KB`,
+        webpSavings: `${webpSavings ? webpSavings.percent + '%' : ''}`,
+        jpegSavings: `${jpegSavings ? jpegSavings.percent + '%' : ''}`
       });
       return results;
     }, []);
@@ -124,6 +122,16 @@ class UsesOptimizedImages extends Audit {
       debugString = `Lighthouse was unable to decode some of your images: ${urls.join(', ')}`;
     }
 
+    const createTable = Formatter.getByName(
+        Formatter.SUPPORTED_FORMATS.TABLE).createTable;
+
+    const table = createTable({
+      url: 'URL',
+      total: 'Original (KB)',
+      webpSavings: 'WebP savings',
+      jpegSavings: 'JPEG savings'
+    }, results);
+
     return UsesOptimizedImages.generateAuditResult({
       displayValue,
       debugString,
@@ -131,8 +139,8 @@ class UsesOptimizedImages extends Audit {
       extendedInfo: {
         formatter: Formatter.SUPPORTED_FORMATS.TABLE,
         value: {
-          headings: ['URL', 'Original (KB)', 'WebP savings', 'JPEG savings'],
-          rows: results
+          results,
+          table
         }
       }
     });
--- a/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
+++ b/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
@@ -123,16 +123,6 @@ class UsesOptimizedImages extends Audit {
       debugString = `Lighthouse was unable to decode some of your images: ${urls.join(', ')}`;
     }
 
-    const createTable = Formatter.getByName(
-        Formatter.SUPPORTED_FORMATS.TABLE).createTable;
-
-    const table = createTable({
-      url: 'URL',
-      total: 'Original (KB)',
-      webpSavings: 'WebP savings',
-      jpegSavings: 'JPEG savings'
-    }, results);
-
     return UsesOptimizedImages.generateAuditResult({
       displayValue,
       debugString,
@@ -141,7 +131,12 @@ class UsesOptimizedImages extends Audit {
         formatter: Formatter.SUPPORTED_FORMATS.TABLE,
         value: {
           results,
-          table
+          tableHeadings: {
+            url: 'URL',
+            total: 'Original (KB)',
+            webpSavings: 'WebP savings',
+            jpegSavings: 'JPEG savings'
+          }
         }
       }
     });
--- a/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
+++ b/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
@@ -89,21 +89,27 @@ class UsesOptimizedImages extends Audit {
       const url = URL.getDisplayName(image.url);
       const webpSavings = UsesOptimizedImages.computeSavings(image, 'webp');
 
-      let label = `${originalKb} KB total, webp savings: ${webpSavings.percent}%`;
       if (webpSavings.bytes > WEBP_ALREADY_OPTIMIZED_THRESHOLD_IN_BYTES) {
         hasAllEfficientImages = false;
       }
 
+      let jpegSavings;
       if (/(jpeg|bmp)/.test(image.mimeType)) {
-        const jpegSavings = UsesOptimizedImages.computeSavings(image, 'jpeg');
+        jpegSavings = UsesOptimizedImages.computeSavings(image, 'jpeg');
         if (jpegSavings.bytes > 0) {
           hasAllEfficientImages = false;
-          label += `, jpeg savings: ${jpegSavings.percent}%`;
         }
       }
 
       totalWastedBytes += webpSavings.bytes;
-      results.push({url, label});
+      results.push({
+        cols: [
+          url,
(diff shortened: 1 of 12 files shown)


patrickhulce:
Niiiiiiiiiiiice, I was needing something like this!

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
```diff
@@ -89,21 +89,25 @@ class UsesOptimizedImages extends Audit {
...
-          label += `, jpeg savings: ${jpegSavings.percent}%`;
         }
       }
 
       totalWastedBytes += webpSavings.bytes;
-      results.push({url, label});
+      results.push({
+        url,
+        total: `${originalKb} KB`,
+        webpSavings: `${webpSavings ? webpSavings.percent + '%' : ''}`,
```
patrickhulce: don't need the check here for `webpSavings`

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
```diff
@@ -89,21 +89,25 @@ class UsesOptimizedImages extends Audit {
...
+        jpegSavings: `${jpegSavings ? jpegSavings.percent + '%' : ''}`
```
patrickhulce: either add a check for it to be positive here or move `jpegSavings = ...` inside the if above?

patrickhulce: probably also move these out of `` strings :)

lighthouse-core/audits/dobetterweb/uses-passive-event-listeners.js
```diff
@@ -80,11 +80,21 @@ class PassiveEventsAudit extends Audit {
...
+    }, groupedResults);
+
     return PassiveEventsAudit.generateAuditResult({
       rawValue: groupedResults.length === 0,
       extendedInfo: {
-        formatter: Formatter.SUPPORTED_FORMATS.URLLIST,
-        value: groupedResults
+        formatter: Formatter.SUPPORTED_FORMATS.TABLE,
+        value: {
+          results: groupedResults,
```
patrickhulce: no changes needed to the mapping?

lighthouse-core/formatters/partials/table.html
```diff
@@ -0,0 +1,68 @@
+<style>
+.table_list {
+  margin-top: 8px;
+  border: 1px solid #EBEBEB;
```
patrickhulce: do we care about alphabetizing our rules?

lighthouse-core/formatters/table.js
```diff
@@ -0,0 +1,95 @@
...
+        throw new Error('Unknown formatter type');
+    }
+  }
+
+  /**
+   * Preps a formatted table (headings/col vals) for output.
+   * @param {!Object<string, string>} headings for the table. The order of this
+   *     object's key/value will be the order of the table's headings.
+   * @param {!Array<*>} results Audit results.
+   * @return {!{headings: string, rows: [{cols: [*]}]}} headings
```
patrickhulce: nit: no `!` needed, headings is `Array<string>` isn't it?

lighthouse-core/audits/dobetterweb/uses-passive-event-listeners.js
ebidel: nope. The table formatter will pull from the`groupedResults` object as needed, based on the keys used in the first arg to `createTable({url: 'URL', lineCol: 'Line/Col', type: 'Type', code: 'Snippet'})`.

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
ebidel: done

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
ebidel: done

lighthouse-core/formatters/partials/table.html
ebidel: There's a PR that alphabetizing selectors was brought up (can't find it now), but we deemed it unrealistic without tooling.

lighthouse-core/formatters/table.js
ebidel: thanks. done.

ebidel:
PTAL

lighthouse-core/formatters/partials/table.html
patrickhulce: a-ok by me!

lighthouse-core/audits/dobetterweb/uses-passive-event-listeners.js
patrickhulce: Ah so the extra info was already in there and was being ignored by URLLIST?

ebidel: Yea.

brendankenny:
To keep formatting out of the audits (and avoiding the `Formatter.getByName(Formatter.SUPPORTED_FORMATS.TABLE)` monster), rather than have each audit access the `createTable` function, you could do something like add a

```js
static getHelpers() {
  createTable
}
```
to the formatter, then have the `extendedInfo` be something like 
```js
extendedInfo: {
  formatter: Formatter.SUPPORTED_FORMATS.TABLE,
  value: {
    results,
    tableHeaders
  }
}
```
and then have `createTable` be applied to `results` in the partial to create the table at that point (and `pretty` could just call `createTable` inside itself)

WDYT?

brendankenny:
this is a huge improvement

lighthouse-core/formatters/table.js
```diff
@@ -0,0 +1,95 @@
...
+        return html;
+
+      default:
+        throw new Error('Unknown formatter type');
+    }
+  }
+
+  /**
+   * Preps a formatted table (headings/col vals) for output.
+   * @param {!Object<string, string>} headings for the table. The order of this
```
brendankenny: nit: can just do `!Object<string>` (since all object keys are strings)

lighthouse-core/formatters/table.js
```diff
@@ -0,0 +1,95 @@
...
+   * @return {{headings: Array<string>, rows: [{cols: [*]}]}} headings
+   */
+  static createTable(headings, results) {
+    const headingKeys = Object.keys(headings);
+
+    const rows = results.map(result => {
+      const cols = headingKeys.map(key => {
+        switch (key) {
+          case 'code':
+            // Wrap code snippets in markdown ticks.
```
brendankenny: worth documenting these special keys in jsdoc?

lighthouse-core/formatters/table.js
```diff
@@ -0,0 +1,95 @@
...
+      default:
+        throw new Error('Unknown formatter type');
+    }
+  }
+
+  /**
+   * Preps a formatted table (headings/col vals) for output.
+   * @param {!Object<string, string>} headings for the table. The order of this
+   *     object's key/value will be the order of the table's headings.
+   * @param {!Array<*>} results Audit results.
```
brendankenny: looks like you maybe expect this to be `!Array<!Object>`?

lighthouse-core/formatters/table.js
```diff
@@ -0,0 +1,95 @@
...
+   * @return {{headings: Array<string>, rows: [{cols: [*]}]}} headings
```
brendankenny: maybe `@return {{headings: !Array<string>, rows: !Array<{cols: !Array<*>}>}}`?

lighthouse-core/formatters/table.js
ebidel: done. added to the function description.

lighthouse-core/formatters/table.js
ebidel: done

lighthouse-core/formatters/table.js
ebidel: done

lighthouse-core/formatters/table.js
ebidel: done

ebidel:
PTAL. `getHelpers` refactor and new tests.

PR #1511, opened 2017-01-22
Move `--harmony` addition to smokehouse/offline-local/run-tests.sh.

### LINKED ISSUE #1508: Is `$__node_harmony` needed in extension?
It's causing issues when building on Windows so I wonder if it's needed for the extension.

Also, would be nice if the package.json `engines` values were consistent across all package.json files.

XhmikosR:


XhmikosR:
So apparently the flag is still needed; reverted that patch.

XhmikosR:
I don't seem to be able to get this right; I'm missing some parts in my understanding regarding tests.

So, if someone could take care of this, that would be better. We need to get rid of `$__node_harmony` from package.json so that it works on Windows too.

brendankenny:
smokehouse was node v4 compliant, but looks like the issue is that we brought `log` into smokehouse to use the cross platform unicode/console stuff, but `log` is *not* node v4 (without `--harmony`) compliant :)

XhmikosR:
Hmm, so I guess this patch is moot...

I was just trying to get rid of the environment variable from package.json
to no avail. :/

On Jan 24, 2017 22:12, "Brendan Kenny" <notifications@github.com> wrote:

smokehouse was node v4 compliant, but looks like the issue is that we
brought log into smokehouse to use the cross platform unicode/console
stuff, but log is *not* node v4 (without --harmony) compliant :)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleChrome/lighthouse/pull/1511#issuecomment-274924424>,
or mute the thread

brendankenny:
Sorry this has been such a pain. Closing as this will be taken care of by #1519 now.

PR #1519, opened 2017-01-24
remove support for node before v6

### LINKED ISSUE #449: move to requiring Node 6 when LTS
With the release of 6.2.2, I think it may be time.
- Node 6 will be the [next LTS version](https://github.com/nodejs/LTS) (skipping 5)
- it will give us a [few more es6 features](http://node.green/) like
  - [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) (avoiding the `|| defaultValue` dance and especially eliminating the verbose cases of checking a value that may be falsy since default parameters always check against `undefined`)
  - [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) (allowing us to avoid noise like [this](https://github.com/GoogleChrome/lighthouse/blob/230cd56dc732ad951dabbcc872dc7a291cb091c3/lighthouse-core/driver/drivers/extension.js#L91-L94) and [this](https://github.com/GoogleChrome/lighthouse/pull/430/files#r67022456))
  - [`Array.prototype.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)
  - [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) (which I'm a bit wary of because every person I've seen use it quickly starts abusing readability for the sake of cool destructuring, but there's no way up but through :)
- there are [no notable language changes in the pipeline for Node 7](http://node.green/) (sorry `**`), which means lighthouse devs can use node 6+ and any language feature used will be compatible with a node 6 baseline (modulo performance improvements). This has been a problem in a few PRs in the past, where someone running node 6 has to rewrite when their change breaks in testing on node 5. Only caveat here is `async`/`await`, which has enough demand that we may need to upgrade when that's turned on.
- I assume the new DevTools debugging? Is this going to be in 6 or skip to 7?

brendankenny:
fixes #449

just basic removal of version checks and logic specifically for getting `--harmony` piped into the CI system. New language features (mostly default parameters, rest args, and `[].includes()`) can come later.

@XhmikosR if I'm missing anything :)


XhmikosR:
Awesome! I'll have a close look at little later just in case you missed
something.

lighthouse-extension/package.json
```diff
@@ -2,11 +2,11 @@
   "name": "lighthouse-extension",
   "private": true,
   "engines": {
-    "node": ">=0.8.0"
+    "node": ">=6"
```
ebidel: should this match travis's 6.9.1? or vice versa?

lighthouse-viewer/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "lighthouse-viewer",
   "private": true,
   "engines": {
-    "node": ">=5"
+    "node": ">=6"
```
ebidel: same

package.json
```diff
@@ -8,7 +8,7 @@
     "chrome-debug": "./lighthouse-cli/manual-chrome-launcher.js"
   },
   "engines": {
-    "node": ">=5"
+    "node": ">=6"
```
ebidel: same

lighthouse-extension/package.json
brendankenny: yeah, I don't know how much we care about the precision on this. We're going to only be testing 6.9.1, so it makes sense that we shouldn't say we work in case important bugs were fixed before that...but on the other hand we're not testing with the latest v6, so it's also possible that some bug was fixed *after* 6.9.1 that materially affects execution...

XhmikosR:
But do you need 6.9.1 specifically? If not just setting this to 6 will
prevent any confusion since CI will always have the latest 6 version anyway.

brendankenny:
> But do you need 6.9.1 specifically?

Similar to how we were specifically testing 4.3.2 before, a key user is stuck with 6.9.1 for at least the immediate future.

It's probably best to test that exact version rather than test with latest 6 and keep an ear out for anything that might affect us (e.g. while there's been only a few changes between 6.9.1 and latest 6.9.4, there was [one destructuring case that no longer throws](https://github.com/nodejs/node/pull/10386), and it's *possible* we could use it in LH in the future and cause issues for anyone stuck with the slightly older version).

ebidel:
Ready after the resolve

PR #1520, opened 2017-01-24
.travis.yml: remove global installation of typescript.

XhmikosR:
It's listed in CLI's dependencies so this is redundant.


paulirish:
lgtm but tagging eric for signoff.

ebidel:
Should be fine. Viewer also brings it in local to the package. https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-viewer/package.json#L28

XhmikosR:
On a side note, I believe we should remove `npm run build-cli` too since that is in prepublish.

Issue #1516, opened 2017-01-24
npm prepublish deprecation


XhmikosR:
So, with node.js 7 which has npm v4 we get:

```
As of npm@5, `prepublish` scripts will run only for `npm publish`.
(In npm@4 and previous versions, it also runs for `npm install`.)
See the deprecation note in `npm help scripts` for more information.
```

So this should be changed to something else eventually.

ebidel:
Ack. Thanks. We're aware of the deprecation but decided to add the `prepublish` step for now. Discussed in https://github.com/GoogleChrome/lighthouse/pull/1439.

XhmikosR:
Better think before than be late :)

What's the alternative here? `install` / `postinstall`? I'm not aware of their pros and cons.

ebidel:
Really just a manual `yarn build-cli` when users check out the code. And we'll need to make sure it's documented before we publish the extension (to prevent the missing typescript build step we've experienced before).

`install`/`postinstall` are not good options because all users that install LH via `npm install` will have to run that unnecessary build step.

XhmikosR:
Isn't `postinstall` being run automatically like `prepublish`?

PR #1521, opened 2017-01-24
.travis.yml: remove `npm run build-cli`.

XhmikosR:
CLI is being built on install with a prepublish script. So now Travis behavior matches real installations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ before_script:
   - export LIGHTHOUSE_CHROMIUM_PATH="$(pwd)/chrome-linux/chrome"
   - sh -e /etc/init.d/xvfb start
   - ./lighthouse-core/scripts/download-chrome.sh
-  - npm run build-cli
   - npm run build-extension
   - npm run build-viewer
 script:


brendankenny:
I actually think we should do the opposite of this and remove the `prepublish`... 😬 

As discussed, `prepublish` on `install` is deprecated and node v8 (and npm 5) is coming out quite soon. We have a [release checklist](https://github.com/GoogleChrome/lighthouse/blob/master/CONTRIBUTING.md#release-guide) now.

If removing `prepublish` can't be done now I'd personally rather just leave in the extra build step until it can be.

ebidel:
Nothing stopping us from removing `prepublish` now.

XhmikosR:
This is just so that Travis is doing the same thing as real life installations. But, yeah, removing prepublish would be good too.

brendankenny:
> This is just so that Travis is doing the same thing as real life installations

If we remove the `prepublish` step (or when npm5 arrives and it's no longer run on install), Travis and real users cloning the git repo will need to run `npm run build-cli`

XhmikosR:
I'm just saying, that what we have now is bad in the sense that it can break without noticing. :)

ebidel:
@brendankenny do you want to keep the travis build and remove the `prepublish` script? Or should we take this for now?

brendankenny:
yeah, I'd personally rather leave as-is until we have a solution for #1516

XhmikosR:
All right, closing since until that happens it will be a while so it's not urgent.

My rationale was that if something breaks, CI won't fail as it is currently.

PR #1522, opened 2017-01-24
Switch back to plain `npm install`.

### LINKED ISSUE #1507: Move back to npm --prefix instead of using -C to install components
See https://ci.appveyor.com/project/XhmikosR/lighthouse/build/59/job/uis3t57c4558yf18 for an example

What does the `-C` npm flag do exactly?

XhmikosR:
`-C` didn't work on Windows.

Fixes #1507.


paulirish:
instead of 
`npm install -C ./lighthouse-cli`

how about just
`npm install --prefix ./lighthouse-cli`
?

is that good enough, or does it need the destination path explicit as well?

XhmikosR:
Unfortunately, it does for me on Windows.

Please check my comment in https://github.com/GoogleChrome/lighthouse/issues/1507#issuecomment-274967215

ebidel:
@paulirish we can do `cd ./lighthouse-cli && npm install` if you think it's too tricky. Anyone reading that link would instantly know what it does and wouldn't have to read up on `--prefix` (me). Also no need for pushd/popd, yea?

Either way, it would be nice to be consistent about the method in our main package.

XhmikosR:
Yeah, `cd` / `pushd` /`popd` would be better IMO.

XhmikosR:
I added the solution I personally prefer. Let me know which one you want.

paulirish:
sure. wfm!

XhmikosR:
You want to squash and rename the commit? I can do it tomorrow if you can't.

On a side note, the build* scripts work with `-C` but they create an empty
etc folder in the subfolders. This is probably a bug in npm. We could
change those to do the same with cd for consistency.

package.json
```diff
@@ -12,9 +12,9 @@
   },
   "scripts": {
     "install-all": "npm run install-cli && npm run install-extension && npm run install-viewer",
-    "install-cli": "npm install -C ./lighthouse-cli",
-    "install-extension": "npm install -C ./lighthouse-extension",
-    "install-viewer": "npm install -C ./lighthouse-viewer",
+    "install-cli": "cd ./lighthouse-cli && npm install && cd ..",
```
ebidel: you shouldnt need the `&& cd .`.

XhmikosR:
I'll test this tomorrow and change it if it works.

ebidel:
SGTM. Agreed we should do the `build` scripts too.

XhmikosR:
@ebidel: done. Feel free to rename the PR or change the commit message to something better.

PR #1524, opened 2017-01-24
use rest parameters instead of arguments

brendankenny:
enabled now that we have node v6 as the baseline.

sorry @ebidel, got to it first :)

### DIFF
--- a/lighthouse-core/lib/console-quieter.js
+++ b/lighthouse-core/lib/console-quieter.js
@@ -26,14 +26,14 @@ class ConsoleQuieter {
   static mute(opts) {
     ConsoleQuieter._logs = ConsoleQuieter._logs || [];
 
-    console.log = function() {
-      ConsoleQuieter._logs.push({type: 'log', args: arguments, prefix: opts.prefix});
+    console.log = function(...args) {
+      ConsoleQuieter._logs.push({type: 'log', args, prefix: opts.prefix});
     };
-    console.warn = function() {
-      ConsoleQuieter._logs.push({type: 'warn', args: arguments, prefix: opts.prefix});
+    console.warn = function(...args) {
+      ConsoleQuieter._logs.push({type: 'warn', args, prefix: opts.prefix});
     };
-    console.error = function() {
-      ConsoleQuieter._logs.push({type: 'error', args: arguments, prefix: opts.prefix});
+    console.error = function(...args) {
+      ConsoleQuieter._logs.push({type: 'error', args, prefix: opts.prefix});
     };
   }
 
--- a/lighthouse-core/lib/log.js
+++ b/lighthouse-core/lib/log.js
@@ -40,21 +40,22 @@ class Emitter extends EventEmitter {
    * Fires off all status updates. Listen with
    * `require('lib/log').events.addListener('status', callback)`
    * @param {string} title
-   * @param {...*} args
+   * @param {!Array<*>} argsArray
    */
-  issueStatus(title, ...args) {
+  issueStatus(title, argsArray) {
     if (title === 'status' || title === 'statusEnd') {
-      this.emit(title, args);
+      this.emit(title, [title, ...argsArray]);
     }
   }
 
   /**
    * Fires off all warnings. Listen with
    * `require('lib/log').events.addListener('warning', callback)`
-   * @param {...*} args
+   * @param {string} title
+   * @param {!Array<*>} argsArray
    */
-  issueWarning(...args) {
-    this.emit('warning', args);
+  issueWarning(title, argsArray) {
+    this.emit('warning', [title, ...argsArray]);
   }
 }
 
@@ -64,9 +65,8 @@ const loggingBufferColumns = 25;
 class Log {
 
   static _logToStdErr(title, argsArray) {
-    const args = [...argsArray];
     const log = Log.loggerfn(title);
-    log(...args);
+    log(...argsArray);
   }
 
   static loggerfn(title) {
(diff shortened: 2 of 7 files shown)


lighthouse-core/lib/log.js
```diff
@@ -40,8 +40,9 @@ class Emitter extends EventEmitter {
    * Fires off all status updates. Listen with
    * `require('lib/log').events.addListener('status', callback)`
    * @param {string} title
+   * @param {...*} args
    */
-  issueStatus(title, args) {
+  issueStatus(title, ...args) {
     if (title === 'status' || title === 'statusEnd') {
       this.emit(title, args);
```
patrickhulce: wouldn't this stick all current args in an array? will that impact anything outside of changed below?

ebidel:
yes
Yes
YES

lighthouse-core/lib/log.js
```diff
@@ -113,23 +115,23 @@ class Log {
     Log._logToStdErr(`${prefix}:${level || ''}`, [data.method, snippet]);
   }
 
-  static log(title) {
-    Log.events.issueStatus(title, arguments);
-    return Log._logToStdErr(title, Array.from(arguments).slice(1));
+  static log(title, ...args) {
+    Log.events.issueStatus(title, ...args);
+    return Log._logToStdErr(title, args);
```
patrickhulce: `_logToStdErr` doesn't seem to have been changed at all did it already support this?

lighthouse-core/lib/console-quieter.js
```diff
@@ -26,14 +26,14 @@ class ConsoleQuieter {
   static mute(opts) {
     ConsoleQuieter._logs = ConsoleQuieter._logs || [];
 
-    console.log = function() {
-      ConsoleQuieter._logs.push({type: 'log', args: arguments, prefix: opts.prefix});
+    console.log = function(...args) {
+      ConsoleQuieter._logs.push({type: 'log', args: args, prefix: opts.prefix});
```
patrickhulce: nit: can just do `{type: 'log', args,`

lighthouse-core/lib/log.js
brendankenny: yes. Kind of confusingly `args` was previously an array, and event consumers are expecting a single array from the event, but that made it awkward to call when switching everything else to rest params, so it was easiest to just have varargs and group them up into an array here.

lighthouse-core/lib/log.js
brendankenny: > `_logToStdErr` doesn't seem to have been changed at all did it already support this?

Maybe I'm misunderstanding you, but isn't this right? `args` is an array of all but the first argument, equivalent to the old `Array.from(arguments).slice(1))`?

lighthouse-core/lib/console-quieter.js
brendankenny: done

lighthouse-core/lib/log.js
patrickhulce: Oh, no I just misread the diff thinking it didn't have a second argument at all, disregard :)

lighthouse-core/lib/log.js
patrickhulce: ah ok

lighthouse-core/lib/log.js
```diff
@@ -113,23 +115,23 @@ class Log {
     Log._logToStdErr(`${prefix}:${level || ''}`, [data.method, snippet]);
   }
 
-  static log(title) {
-    Log.events.issueStatus(title, arguments);
-    return Log._logToStdErr(title, Array.from(arguments).slice(1));
+  static log(title, ...args) {
+    Log.events.issueStatus(title, ...args);
```
wardpeet: There is one difference with the previous code block.
...args doesn't contain title whereas arguments did.
to fix you should do `issueStatus(title, [title, ...args])`

lighthouse-core/lib/log.js
```diff
@@ -113,23 +115,23 @@ class Log {
...
 
-  static error(title) {
-    return Log._logToStdErr(`${title}:error`, Array.from(arguments).slice(1));
+  static error(title, ...args) {
+    return Log._logToStdErr(`${title}:error`, args);
   }
 
-  static verbose(title) {
+  static verbose(title, ...args) {
     Log.events.issueStatus(title);
```
wardpeet: See comment above about arguments, ...args inside the function will be an empty []

lighthouse-core/lib/log.js
brendankenny: ugh, I looked at this, thought it was terrible but had to be maintained, then promptly forgot about it, apparently. Thanks :)

lighthouse-core/lib/log.js
brendankenny: > See comment above about arguments, ...args inside the function will be an empty []

I don't think that applies here since it's only doing `Log.events.issueStatus(title)`, so it gets an empty array on master already.

Honestly I have no idea why it's doing `issueStatus` here with no args anyways. It doesn't make much sense.

brendankenny: looks like it was a bug introduced in https://github.com/GoogleChrome/lighthouse/commit/71195e4ac66f096c8505ca78321b523d9722270c#diff-e5b0c339f7ea6da4ec03ff079a28288cL100 I'll fix.

brendankenny:
updated (see second commit). Also reduced some of the ridiculous repetitions of rest/spread/rest/spread :)

ebidel:
😴  + ⛳️ + 🐏 s

Issue #1517, opened 2017-01-24
Audits: Detect byte and time savings from bundled resources


patrickhulce:
Across the byte efficiency audits time saved from data URI resources, stylesheets added from script is not accurately reflected. Ideally we'd be able to attribute these resources to the actual network record they come in on (the JavaScript or HTML file that included them), but at a minimum we can estimate the network throughput and time saved on average.

- [ ] Attribute inline CSS stylesheets to the appropriate HTML or JS file - https://github.com/GoogleChrome/lighthouse/issues/1544
- [X] ~Provide a byte estimate for inline stylesheets #1536~
- [X] ~Provide a time estimate for inline stylesheets #1536~
- [X] ~Attribute data URI images to the appropriate HTML/JS/CSS file~ no longer needed
- [X] ~Provide a time estimate for data URI images #1536~

ebidel:
I've been thinking it would be really nice to work towards very visual infographic-type callouts in the report. High level aggregations, but also breakdowns as you drill into each of these audits. Something like:
https://pageweight.imgix.com/jobs/99166054-29d7-413a-a0c4-1d7c8e30a8e2.html?utm_source=share-link

I'm visual. Users are visual. The world loves visuals!

brendankenny:
see also [May 2016 UX mocks](https://docs.google.com/document/d/1ZEPbE79DmMXLyUQ3k8pxQVQaiWxcg4wCDfLlwy3BZaE/edit). There are various Irish/Lewis mocks spread out in issues as well (cc @paulirish)

(no pie charts, tho :P)

patrickhulce:
#1500 is related to first bullet on attribution of inline stylesheets

patrickhulce:
Marking this as closed since the last remaining enhancement is now orthogonal to the original intent of the issue and is tracked separately in #1544.

Issue #1469, opened 2017-01-14
Image Optimization Audit: add metric mapping bytes saved <-> time saved


ebidel:
This has been mentioned a couple of times, but capturing in a bug....

The image optimization audit's "potential savings" for bytes is super useful:

<img width="718" alt="screen shot 2017-01-13 at 4 05 57 pm" src="https://cloud.githubusercontent.com/assets/238208/21949847/524f2efa-d9aa-11e6-8ff7-601406c804b9.png">

It would be great to also include how much time this would save the user.  Something like:

"Optimizing images would save X bytes and load x ms faster"

paulirish:
Thanks for filing. :) What a pro. 

One note: calculating time saved on fmp/TTI/etc requires critical path analysis which is tough. A good start is calculating the possibld change in "milliseconds spent downloading resources" These times will overlap but I think that's ok. Cc @wardpeet

ebidel:
That would also be jive well with the scripts/links blocking first render audit. Those return download times as the metric.

PR #1536, opened 2017-01-25
feat: byte efficiency estimates based on average throughput

patrickhulce:
addresses #1469 and several tasks of #1517

Switches the approach used previously from adding up the potential ms of each individual resource and instead estimating the time savings based on the average throughput across all requests and the total bytes saved. The previous approach had similar shortcomings to our current links-blocking-first-paint audit where we would report savings longer than the load of the page (because in reality these assets are loaded in parallel, not serially). This should give a more robust and realistic estimate of how many total milliseconds can be shaved off of the load time.

**Before:**
image

**After:**
image


brendankenny:
This is very cool.

Instead of living on `NetworkRecords`, does this make sense as a computed gatherer?

Would it also make sense to have an average time per domain? That makes certain questions slightly harder to answer (was the connection to a domain in some interval server limited or was it because there were a bunch of connections open to some other domain), but it clears up other things, like recognizing the use of a fast CDN for static assets while ads on the same page are loaded from a much slower server.

patrickhulce:
> Instead of living on NetworkRecords, does computeAverageThroughput make sense as a computed gatherer?

Good point I totally forgot about those :) yes, I'll convert it

> Would it also make sense to have an average time per domain?

I actually started with this, but ended up just going back to all requests for a couple reasons:
* Until we have network request attribution for *all* data URI images and stylesheets we'll still need to compute the total average throughput, so start there to keep it simple.
* If we accept my wild assumption that the client is typically the throughput bottleneck, the differences between origins wouldn't matter much.
* The way this metric gets used in the byte efficiency audits for the time being completely ignores response time and the fact that throughput changes over the duration of the request so the benefit you get from a CDN having low latency and quickly saturating the network wouldn't really be surfaced anyway. (Although that's a good idea for another audit, warning about just long server response times!)

Ideally, we'd have some fancy performance experimentation where we could cutoff downloading after X bytes have been received and really tell you what the savings would be

lighthouse-core/formatters/partials/table.html
```diff
@@ -26,6 +26,12 @@
 .table_list tr:hover {
   background-color: #fafafa;
 }
+.table_list td * {
```
ebidel: Perf isnt a huge deal on the report, but `*` is always nasty. Is there something more specific we can use? Right now it's just br, code blocks as children elements. RIght?

lighthouse-core/formatters/table.js
```diff
@@ -78,13 +78,18 @@ class Table extends Formatter {
 
     const rows = results.map(result => {
       const cols = headingKeys.map(key => {
+        let value = result[key];
+        if (typeof value === 'undefined') {
```
ebidel: nit: `!(key in value)`

lighthouse-core/formatters/table.js
```diff
@@ -78,13 +78,18 @@ class Table extends Formatter {
...
+
         switch (key) {
           case 'code':
-            return '`' + result[key].trim() + '`';
+            return '`' + value.trim() + '`';
           case 'lineCol':
             return `${result.line}:${result.col}`;
           default:
-            return result[key];
+            return String(value);
```
ebidel: doesnt handlebars do this for us?

lighthouse-core/report/report-generator.js
```diff
@@ -132,6 +132,7 @@ class ReportGenerator {
       // XSS, define a renderer that only transforms links and code snippets.
       // All other markdown ad HTML is ignored.
       const renderer = new marked.Renderer();
+      renderer.br = () => '<br>';
```
ebidel: are you using `<br>` anywhere?

lighthouse-core/formatters/table.js
patrickhulce: doesn't `let foo; 'foo' in {foo} === true`?

lighthouse-core/formatters/table.js
patrickhulce: No the results get directly passed to the pretty formatter above which is where `startsWith` handler above fails with numbers

lighthouse-core/formatters/partials/table.html
patrickhulce: Interesting what makes it so expensive? Reflows become a lot slower?

lighthouse-core/report/report-generator.js
patrickhulce: yes `  \n` in between the url and the content preview for stylesheets

lighthouse-core/formatters/table.js
ebidel: not sure how many undefined vals we get, but will a bunch of '--' cells be too noisy?

lighthouse-core/formatters/partials/table.html
ebidel: Yea. Selectors match right -> left. So `*` first matches every DOM node, then filters on `td`s, etc.

lighthouse-core/formatters/table.js
ebidel: ack

lighthouse-core/report/report-generator.js
ebidel: Ah ok. I have a PR coming that introduces `<pre>` to the markdown renderer. We could also use that for block level code snippets, removing the need for `<br>` and the css you have. WDYT?

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
```diff
@@ -114,7 +117,9 @@ class UsesOptimizedImages extends Audit {
 
     let displayValue = '';
     if (totalWastedBytes > 1000) {
-      displayValue = `${Math.round(totalWastedBytes / KB_IN_BYTES)}KB potential savings`;
+      const totalWastedKb = Math.round(totalWastedBytes / KB_IN_BYTES);
+      const totalWastedMs = Math.round(totalWastedBytes / networkThroughput * 1000);
+      displayValue = `${totalWastedKb}KB (~${totalWastedMs}ms) potential savings`;
```
paulirish: Because this number is super hand-wavy let's reduce the precision some more. Shall we round to the nearest ten?  e.g. 497 => 500 and  432 => 430.

lighthouse-core/formatters/table.js
patrickhulce: I found it super helpful for the CLI to differentiate between missing cells and (totally subjectively) more pleasing in the HTML, but open to other ideas

ebidel: Sounds like we might need some shared helps for the code and empty cell formatting. Should we do that here or consolidate later?

lighthouse-core/report/report-generator.js
patrickhulce: Yeah just not sure how to appropriately target the `pre` tag then for margin only when its following other text. Also I can't get the `pre` to generate with triple ticks, it still does `code` for me after cherry-picking :/

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
patrickhulce: sure sg

lighthouse-core/formatters/partials/table.html
patrickhulce: TIL :)

lighthouse-core/formatters/table.js
patrickhulce: Not sure what you mean, which code needs to be shared?

ebidel: https://github.com/GoogleChrome/lighthouse/pull/1536/files#diff-508b9167df0e380cb39ba29ab7deafa2R98 for example. I could use that over in the flexbox audit to fix the whitespace formatting of the code snippet.

lighthouse-core/report/report-generator.js
patrickhulce: Alright I kinda like this, hacky but with a look of intention :)
image

lighthouse-core/formatters/table.js
patrickhulce: Oh the fix for that is most likely the `~` added to the report HTML, that was some more aggressive edge case hunting, but yes we totally should. I'm in favor of getting this consistency PR through between the byte efficiency and then doing a bit of refactor cleanup of the stuff that's starting to be repeated everywhere, if that makes sense to you too?

ebidel: Makes sense. I couldn't get `~` to work but I might be doing Handlebad.

patrickhulce:
I'm not getting off this easy am I? PTAL :)

lighthouse-core/audits/unused-css-rules.js
```diff
@@ -33,19 +35,27 @@ class UnusedCSSRules extends Audit {
...
 
   /**
    * @param {!Array.<{header: {styleSheetId: string}}>} styles The output of the Styles gatherer.
+   * @param {!Array<NetworkRecord>} networkRecords
    * @return {!Object} A map of styleSheetId to stylesheet information.
    */
-  static indexStylesheetsById(styles) {
+  static indexStylesheetsById(styles, networkRecords) {
+    const indexedNetworkRecords = networkRecords
+        .filter(record => /css/.test(record.mimeType))
```
paulirish: rather than mimeType we might want to use `record.resourceType()`: 
https://github.com/ChromeDevTools/devtools-frontend/blob/b84fc38d5618d27b6a5b50d3948cb31a38488bfa/front_end/common/ResourceType.js#L183

lighthouse-core/audits/unused-css-rules.js
```diff
@@ -79,6 +89,42 @@ class UnusedCSSRules extends Audit {
...
+        .slice(0, PREVIEW_LENGTH * 5)
+        .replace(/( {2,}|\t)+/g, '  ') // remove leading indentation if present
+        .replace(/\n\s+}/g, '\n}') // completely remove indentation of closing braces
+        .trim(); // trim the leading whitespace
+
+    if (preview.length > PREVIEW_LENGTH) {
+      const firstRuleStart = preview.indexOf('{');
+      const firstRuleEnd = preview.indexOf('}');
+
+      if (firstRuleStart === -1 || firstRuleEnd === -1
```
paulirish: Is there anything from https://github.com/ChromeDevTools/devtools-frontend/blob/b84fc38d5618d27b6a5b50d3948cb31a38488bfa/front_end/gonzales/SCSSParser.js we can reuse

not sure, just curious if we can leave some of the parsing/traversal to another project.

lighthouse-core/gather/computed/network-throughput.js
```diff
@@ -0,0 +1,72 @@
...
+   * @return {number}
+   */
+  compute_(networkRecords) {
+    if (!networkRecords || !networkRecords.length) {
+      return 0;
+    }
+
+    let totalBytes = 0;
+    const timeBoundaries = networkRecords.reduce((boundaries, record) => {
+      if (/^data:/.test(record.url) || record.failed || !record.finished ||
```
paulirish: could use `record.parsedURL().scheme` here if you want.

lighthouse-core/audits/unused-css-rules.js
patrickhulce: Yeah I tried to use these but because everything is expressed in terms of lines and columns translating to absolute length ended up being even more complicated than this (and didn't really handle leading comments such as those used in every lighthouse file :) ) It also focuses a lot more on rules than rule-sets which made for tricky and generally confusing logic to show the first rule-set with its selector.

patrickhulce: Also I would love to just totally kill this at some point in the future by attributing where the inline sheet came from and/or maybe showing a few of the selectors

lighthouse-core/gather/computed/network-throughput.js
patrickhulce: yessss, this is what I've been looking for.

lighthouse-core/audits/unused-css-rules.js
patrickhulce: done

lighthouse-core/audits/unused-css-rules.js
paulirish: k sgtm

Issue #1537, opened 2017-01-25
CLI Error: Cannot read property 'startsWith' of undefined


patrickhulce:
I've already fixed this in #1536 but I'll split out a fix since that review will likely take some time.

patrickhulce:
Full stack trace:
```
Runtime error encountered: TypeError: Cannot read property 'startsWith' of undefined
    at row.cols.forEach.col (/Users/phulce/Code/Chrome/lighthouse/lighthouse-core/formatters/table.js:42:22)
    at Array.forEach (native)
    at table.rows.forEach.row (/Users/phulce/Code/Chrome/lighthouse/lighthouse-core/formatters/table.js:40:22)
    at Array.forEach (native)
    at result (/Users/phulce/Code/Chrome/lighthouse/lighthouse-core/formatters/table.js:38:22)
    at item.subItems.forEach.subitem (/Users/phulce/Code/Chrome/lighthouse/lighthouse-cli/printer.js:109:34)
    at Array.forEach (native)
    at aggregation.score.forEach.item (/Users/phulce/Code/Chrome/lighthouse/lighthouse-cli/printer.js:89:27)
    at Array.forEach (native)
    at results.aggregations.forEach.aggregation (/Users/phulce/Code/Chrome/lighthouse/lighthouse-cli/printer.js:84:27)
```

PR #1538, opened 2017-01-25
fix(table-formatter): account for undefined values

patrickhulce:
addresses #1537 
a more robust fix with tests that addresses the underlying issue can be found in #1536


brendankenny:
> a more robust fix with tests that addresses the underlying issue can be found in #1536

any reason you can't grab `table.js` and `table-formatter-test.js` from the PR and put into this one?

patrickhulce:
just that it has a slightly different direction that could reasonably be discussed and didn't want a broken master for longer than necessary

ebidel:
This is good protection we should have already been doing.

PR #1543, opened 2017-01-26
CLI: close tab when we're done with it.

### LINKED ISSUE #1419: Close tab after we're done with it
In the CLI case we spawn a new tab, but then we keep it open we we disconnect.
I keep a `chrome-debug` open while I'm developing or perf testing, but I end up with LOTS of old tabs that I manually clear. 

I think that if we did spawn a new tab, then we should definitely close it when we teardown.
ya?

#goodsecondbug

paulirish:
Fixes #1419

### DIFF
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -79,9 +79,11 @@ class CriConnection extends Connection {
               resolve(JSON.parse(data));
               return;
             } catch (e) {
+              // In the case of 'close' Chromium returns a string rather than JSON: goo.gl/7v27xD
               if (data === 'Target is closing') {
                 return resolve({message: data});
               }
+              return reject(e);
             }
           }
           reject(new Error(`Protocol JSON API error (${command}), status: ${response.statusCode}`));
@@ -119,6 +121,7 @@ class CriConnection extends Connection {
       this._ws.removeAllListeners();
       this._ws.close();
       this._ws = null;
+      this._pageId = null;
     });
   }
 
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -43,6 +43,7 @@ class CriConnection extends Connection {
   connect() {
     return this._runJsonCommand('new').then(response => {
       const url = response.webSocketDebuggerUrl;
+      this._pageId = response.id;
 
       return new Promise((resolve, reject) => {
         const ws = new WebSocket(url);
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/connections/cri.js
```diff
@@ -74,10 +75,16 @@ class CriConnection extends Connection {
         });
         response.on('end', _ => {
           if (response.statusCode === 200) {
-            resolve(JSON.parse(data));
-            return;
+            try {
+              resolve(JSON.parse(data));
+              return;
+            } catch (e) {
+              if (data === 'Target is closing') {
```
paulirish: The [chromium code](https://cs.chromium.org/chromium/src/content/browser/devtools/devtools_http_handler.cc?q=%22Target+is+closing%22&sq=package:chromium&l=593&dr=C) that handles this says

```c++
SendJson(connection_id, net::HTTP_OK, NULL, "Target is closing");
```

But unfortunately, it doesn't send JSON. It sends this raw string. 
Thus, the try/catch on this guy.

lighthouse-core/gather/connections/cri.js
```diff
@@ -108,10 +115,11 @@ class CriConnection extends Connection {
       log.warn('CriConnection', 'disconnect() was called without an established connection.');
       return Promise.resolve();
     }
-    this._ws.removeAllListeners();
-    this._ws.close();
-    this._ws = null;
-    return Promise.resolve();
+    return this._runJsonCommand(`close/${this._pageId}`).then(_ => {
```
patrickhulce: Will this close the browser too if it's the last tab left?

paulirish: It will not. It'll close the "window" but not the application.

lighthouse-core/gather/connections/cri.js
```diff
@@ -108,10 +115,11 @@ class CriConnection extends Connection {
...
+      this._ws.removeAllListeners();
+      this._ws.close();
+      this._ws = null;
```
brendankenny: I don't know when we'd ever reuse a cri connection instance, but since we null out `_ws`, null out `_pageId` as well?

lighthouse-core/gather/connections/cri.js
brendankenny: Will you add some variant of this as a comment in the code? At least for the string being checked against

lighthouse-core/gather/connections/cri.js
```diff
@@ -74,10 +75,16 @@ class CriConnection extends Connection {
...
           if (response.statusCode === 200) {
-            resolve(JSON.parse(data));
-            return;
+            try {
+              resolve(JSON.parse(data));
+              return;
+            } catch (e) {
+              if (data === 'Target is closing') {
+                return resolve({message: data});
+              }
```
brendankenny: Since you have the try/catch now, want to add a `return reject(e)` here, even though it wasn't caught in the promise chain before?

paulirish: sg

lighthouse-core/gather/connections/cri.js
paulirish: yup!

lighthouse-core/gather/connections/cri.js
paulirish: fair. done.

paulirish:
PTAL

brendankenny:
looks like some of the gather-runner tests failed because `fake-driver` [doesn't `return Promise.resolve();` in `disconnect`](https://github.com/GoogleChrome/lighthouse/blob/518b87b5602064291f3f263f1bfa998b409f1e5b/lighthouse-core/test/gather/fake-driver.js#L25).

PR #1547, opened 2017-01-26
Move date.now and console.time audits to table formatter

ebidel:
R: all

This also ads the table heads to the pretty print output. The alignment is not great, but otherwise it's hard to know what the columns mean. Ideally, we'd bring in a CLI table formatter, but ...g3 :(

screen shot 2017-01-26 at 12 51 48 pm

### DIFF
--- a/lighthouse-core/formatters/table.js
+++ b/lighthouse-core/formatters/table.js
@@ -34,6 +34,11 @@ class Table extends Formatter {
           }
 
           const table = Table.createTable(result.tableHeadings, result.results);
+          const headings = Object.keys(result.tableHeadings).map(key => {
+            return result.tableHeadings[key].toUpperCase();
+          });
+
+          output += `      ${headings.join(' ')}\n`;
 
           table.rows.forEach(row => {
             output += '      ';
@@ -66,6 +71,7 @@ class Table extends Formatter {
    *       code: wraps the value in ticks as a markdown code snippet.
    *       lineCol: combines the values for the line and col keys into a single
    *                value "line/col".
+   *       isEval: returns "yes" if the script was eval'd.
    *       All other values are passed through as is.
    * @param {!Array<!Object>} results Audit results.
    * @return {{headings: !Array<string>, rows: !Array<{cols: !Array<*>}>}}
@@ -88,6 +94,8 @@ class Table extends Formatter {
             return '`' + value.trim() + '`';
           case 'lineCol':
             return `${result.line}:${result.col}`;
+          case 'isEval':
+            return value ? 'yes' : '';
           default:
             return String(value);
         }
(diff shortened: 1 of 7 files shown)


brendankenny:
>  Ideally, we'd bring in a CLI table formatter, but ...g3

what about something simple like [text-table](https://www.npmjs.com/package/text-table) with no deps and tiny in size.

lighthouse-core/formatters/table.js
```diff
@@ -78,13 +83,16 @@ class Table extends Formatter {
...
     const rows = results.map(result => {
       const cols = headingKeys.map(key => {
+        const val = result[key];
         switch (key) {
           case 'code':
-            return '`' + result[key].trim() + '`';
+            return '`' + val.trim() + '`';
           case 'lineCol':
             return `${result.line}:${result.col}`;
+          case 'isEval':
```
patrickhulce: Not a huge fan of accumulating audit-specific display logic here, do we anticipate `isEval` cropping up in several other places too?

patrickhulce:
thoughts on using `URL.getDisplayName` on those urls if they're not eval? would be nice to standardize there :)

paulirish:
> what about something simple like text-table with no deps and tiny in size.

pretty good. here's their left-right align:

image

ebidel:
I'd like to prettifer the pretty print in another PR, but don't we need a module that's checked in to g3?

ebidel:
There's https://github.com/75lb/table-layout

lighthouse-core/formatters/table.js
ebidel: We have a few JS audits that this is relevant for.

ebidel:
@patrickhulce should we `URL.getDisplayName` in the table formatter so everyone gets it and audits don't have to call themselves?

brendankenny:
> I'd like to prettify the pretty print in another PR, but don't we need a module that's checked in to g3?

New modules can be checked in, it's just a big responsibility..."we" have to own compat and maintenance. Since it has no dependencies it makes that a lot easier since it'll only be the addition of a single module.

brendankenny:
> There's https://github.com/75lb/table-layout

yeah, there are a few with a lot more features, but e.g. table-layout brings in seven other modules in addition to itself. I guess it just depends on how much we need fancy table layout.

[cli-table](https://www.npmjs.com/package/cli-table) is another option, and only brings in [colors](https://www.npmjs.com/package/colors), but the compatibility there is harder as it's fixed on 1.03, we already have `~0.6.3` from speedline, and the g3 version is 1.1.2. So someone will have to figure out what changes were made in that range and what incompatibilities exist as a result :)

patrickhulce:
> should we URL.getDisplayName in the table formatter so everyone gets it and audits don't have to call themselves?

I like it in theory, but there are some complexities with inline/previews we have to consider and I was going to suggest accepting a parameter for the base URL so that cross-origin hosts would be shown, since those weren't so obvious for me on a few runs which would complicate the formatter a bit. Either way let's save for another PR?

ebidel:
SG

brendankenny:
table formatter test bugs on travis

ebidel:
fixed tests!

PR #1551, opened 2017-01-27
Report: fix rendering issues in FF. Fixes #1124

### LINKED ISSUE #1124: Report: FF rendering issues

ebidel:
R: all

This is an alternative to https://github.com/GoogleChrome/lighthouse/pull/1489.

The benefit is that it keeps the SVG unencoded (easier to maintain later) and also fixes the checkboxes in firefox. Didn't touch the Beta banner for FF. Probably not worth it.

screen shot 2017-01-26 at 7 30 29 pm

### DIFF
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -267,7 +267,7 @@ body {
   position: absolute;
   top: 0;
   right: 0;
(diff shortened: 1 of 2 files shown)


Janpot:
the beta banner seems still missing in your screenshot.

XhmikosR:
What @Janpot said plus the toggle help button looks bad using FF 52b1.

1498alt

Other than those 2, much cleaner solution, agreed 👍

ebidel:
I should have mentioned, I didn't touch the Beta banner for FF. Didn't think it's worth the trouble.

My screenshot is from FF54. It looks like they've fixed the behavior of `-moz-appearance` for checkboxes in later versions of FF.

XhmikosR:
All right, then I guess it makes sense to merge this but it will land on beta in like 2 months at best.

I'd still URL encode the banner though unless you plan to remove it soon :).

ebidel:
We've talked about that :)

XhmikosR:
All right, closing my PR then, and thanks for this!

ebidel:
Update: 

This is also rocking blue "?" icons now and black headings, as per @paulirish's css changes for devtools.

screen shot 2017-01-27 at 3 40 58 pm

PR #1553, opened 2017-01-27
Update readme.md

XhmikosR:
Use the `https` URL for viewer.


ebidel:
Thanks. As part of this, I've also switched on SSL for our gh pages.

PR #1555, opened 2017-01-27
fix: report proper first paint delay for blocking tags audits

patrickhulce:
Switches the blocking tag audits to report the maximum delay of first paint rather than a sum of the individual delays. Moves us more in alignment with the actual impact these violations have on the UX.
Also removes duplicate URLs from showing up twice to be inline with our use of "resources" and switches to table formatter.

**Before**:
image

**After**:
image

### DIFF
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -22,6 +22,7 @@
 'use strict';
 
 const Audit = require('../audit');
+const URL = require('../../lib/url-shim');
 const Formatter = require('../../formatters/formatter');
 
 class LinkBlockingFirstPaintAudit extends Audit {
@@ -65,7 +66,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
       endTime = Math.max(item.endTime, endTime);
 
       return {
-        url: item.tag.url,
+        url: URL.getDisplayName(item.tag.url),
         totalKb: `${Math.round(item.transferSize / 1024)} KB`,
         totalMs: `${Math.round((item.endTime - startTime) * 1000)}ms`
       };
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -58,9 +58,8 @@ class LinkBlockingFirstPaintAudit extends Audit {
 
     const filtered = artifact.filter(item => item.tag.tagName === tagFilter);
 
-    let startTime = Number.MAX_VALUE;
+    const startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), Number.MAX_VALUE);
     let endTime = 0;
-    startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), startTime);
 
     const results = filtered.map(item => {
       endTime = Math.max(item.endTime, endTime);
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -55,31 +55,43 @@ class LinkBlockingFirstPaintAudit extends Audit {
       };
     }
 
-    let totalSpendTime = 0;
-    const results = artifact.items
-      .filter(item => item.tag.tagName === tagFilter)
-      .map(item => {
-        totalSpendTime += item.spendTime;
+    const filtered = artifact.filter(item => item.tag.tagName === tagFilter);
 
-        return {
-          url: item.tag.url,
-          label: `delayed first paint by ${item.spendTime}ms`
-        };
-      });
+    let startTime = Number.MAX_VALUE;
+    let endTime = 0;
+    startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), startTime);
 
+    const results = filtered.map(item => {
+      endTime = Math.max(item.endTime, endTime);
+
+      return {
+        url: item.tag.url,
+        totalKb: `${Math.round(item.transferSize / 1024)} KB`,
+        totalMs: `${Math.round((item.endTime - startTime) * 1000)}ms`
+      };
+    });
+
+    const delayTime = Math.round((endTime - startTime) * 1000);
     let displayValue = '';
     if (results.length > 1) {
-      displayValue = `${results.length} resources delayed first paint by ${totalSpendTime}ms`;
+      displayValue = `${results.length} resources delayed first paint by ${delayTime}ms`;
     } else if (results.length === 1) {
-      displayValue = `${results.length} resource delayed first paint by ${totalSpendTime}ms`;
+      displayValue = `${results.length} resource delayed first paint by ${delayTime}ms`;
     }
 
     return {
       displayValue,
       rawValue: results.length === 0,
       extendedInfo: {
-        formatter: Formatter.SUPPORTED_FORMATS.URLLIST,
-        value: results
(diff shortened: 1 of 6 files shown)


lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -55,31 +56,43 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
+    let startTime = Number.MAX_VALUE;
+    let endTime = 0;
+    startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), startTime);
 
+    const results = filtered.map(item => {
+      endTime = Math.max(item.endTime, endTime);
+
+      return {
+        url: URL.getDisplayName(item.tag.url),
+        totalKb: `${Math.round(item.transferSize / 1024)} KB`,
```
ebidel: should we include the unit alongside the table value (what you have) or just in the header?

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -55,31 +56,43 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
+    const filtered = artifact.filter(item => item.tag.tagName === tagFilter);
 
-        return {
-          url: item.tag.url,
-          label: `delayed first paint by ${item.spendTime}ms`
-        };
-      });
+    let startTime = Number.MAX_VALUE;
+    let endTime = 0;
+    startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), startTime);
```
ebidel: would this work:

`let startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), Number.MAX_VALUE);`

patrickhulce: sure will, just looked weird above `endTime` but this looks weird too, I'll switch back

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
patrickhulce: we've been including it in each cell and header so far and I think it's fine to be a bit redundant for the sake of clarity

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -55,31 +56,42 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
-      .map(item => {
-        totalSpendTime += item.spendTime;
+    const filtered = artifact.filter(item => item.tag.tagName === tagFilter);
 
-        return {
-          url: item.tag.url,
-          label: `delayed first paint by ${item.spendTime}ms`
-        };
-      });
+    const startTime = filtered.reduce((t, item) => Math.min(t, item.startTime), Number.MAX_VALUE);
```
brendankenny: are these strictly overlapping or is it possible there are breaks between them?

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -55,31 +56,42 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
+    let endTime = 0;
 
+    const results = filtered.map(item => {
+      endTime = Math.max(item.endTime, endTime);
+
+      return {
+        url: URL.getDisplayName(item.tag.url),
+        totalKb: `${Math.round(item.transferSize / 1024)} KB`,
+        totalMs: `${Math.round((item.endTime - startTime) * 1000)}ms`
```
brendankenny: should this be `item.startTime` for just this item?

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
patrickhulce: it's possible there are breaks, what does it impact if there are breaks?

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
patrickhulce: overall startTime IMO (really it should be just when the original request finished but first possible request seems good enough). Using the overall start then highlights the biggest offenders the right way, since these block first paint, we want developers to target the ones that pushed it out the farthest regardless of when the request was started by Chrome.

patrickhulce:
@ebidel PTAL :)

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
brendankenny: > it's possible there are breaks, what does it impact if there are breaks?

If there are link elements blocking first paint, then the browser is off doing something else, then it gets back to parsing and finds more link elements blocking first paint, don't we only want to report the time due to the link elements?

brendankenny: in other words, is it possible for there to be a large difference between the union of time spent with these blocking first paint vs the first `startTime` subtracted from the last `endTime`

patrickhulce: It's possible for that difference to be large (though in practice I suspect it to be small unless people love putting huge script in the head or using document.write all over the place in which case there are other audits for them and I'm happy to punish them a little more than necessary here :) ) given our numbers because these are all purely network times not parse or anything else that would have generated these large gaps, but the goal here is for there to be 0 items blocking first paint and as far as I'm concerned it makes more sense to report that the violations pushed out first paint to at least X ms than to report the share of network time they occupied.

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
brendankenny: > Using the overall start then highlights the biggest offenders the right way, since these block first paint, we want developers to target the ones that pushed it out the farthest regardless of when the request was started by Chrome

Ah, I see now. So the idea is that the parser has seen these by the time the first request is sent, so even if it doesn't get around to requesting this item right away, it makes sense to punish for that full time? That makes sense for many cases it's of course not just that item's contribution to blocking time...e.g. if the browser is throttled on the number of connections to make and so a request gets pushed further out, eliminating that particular request won't speed up page by `item.endTime - startTime`.

OTOH, maybe with the change you're making to how the total blocked time is calculated, it's clear that while this individual request blocked first paint for this long, eliminating this request won't speed up first paint by that much (i.e. the individual delays are overlapping).

I might be out in the weeds here, though, so I'll defer to you all :) @paulirish for his opinion

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
brendankenny: yeah, `document.write` is definitely one that throws a wrench in the works. Unfortunately we don't have any audits that try to quantify the downsides of it, just true/false is it used.

brendankenny: FWIW, I wrote some quick code to calculate the blocking length from the union of these requests and I could find no page where the two numbers differed (no site had a break between overlapping requests). It's possible I didn't test enough bad sites, though :)

patrickhulce: phew! I can achieve it using a hypothetical combo of stylesheets returning a 500 (which Chrome will re-request!?) and HTML imports injected later with a delayed script just as the first blocking stylesheet comes in :) but glad that situation isn't common in practice haha

patrickhulce:
@brendankenny any more objections to approval or just waiting on @paulirish 's opinion?

PR #1558, opened 2017-01-27
Restore "first visual change" to output.

paulirish:
For a bit, these metrics have reported blanks: 
image

This was following the refactor of the speedindex extendedInfo. This PR fixes the output of CLI/HTML to correctly emit these timings.


ebidel:
Nice catch. I noticed this too at one point and forget to file a 🐛  :(

PR #1560, opened 2017-01-27
Start using errors instead of -1 for error artifacts

brendankenny:
retrying #1498 (as next step of #941), now with errors from gatherers being considered recoverable by default. If you want to stop a run, throw an error with a `fatal` property set to `true`.

Biggest benefit is that if something throws inside a gatherer that isn't something the gatherer cares about or is specifically watching for, it can just let the exception go. That means
- no more `catch` block in every gatherer
- no more `-1` artifacts...now they're just `Error` objects
- no more checking for `-1` artifacts in every audit. If one of an audit's `requiredArtifacts` is an error, the audit never runs and LH emits an error result on behalf of that audit.

First commits changes old `recoverable` infra to support `fatal`.
Other commits try switching over some of our gatherers to using this format.  The first two (`meta viewport` and `theme-color`) are simple, the last commit is giving that treatment to the slightly more complicated `Styles` and `CSSUsage` gatherers and the two audits that need them.

Thoughts?

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -110,14 +110,14 @@ class GatherRunner {
   }
 
   /**
-   * Catches any `recoverable` errors from the supplied promise, rejecting on
-   * the rest.
+   * Test any error output from the promise, absorbing non-fatal errors and
+   * throwing on fatal ones so that run is stopped.
    * @param {!Promise<*>} promise
    * @return {!Promise<*>}
    */
   static recoverOrThrow(promise) {
     return promise.catch(err => {
-      if (!err.recoverable) {
+      if (err.fatal) {
         throw err;
       }
     });
@@ -237,8 +237,8 @@ class GatherRunner {
   /**
    * Takes the results of each gatherer phase for each gatherer and uses the
    * last produced value (that's not undefined) as the artifact for that
-   * gatherer. If a recoverable error was rejected from a gatherer phase,
-   * uses that error as the artifact instead.
+   * gatherer. If a non-fatal error was rejected from a gatherer phase,
+   * uses that error object as the artifact instead.
    * @param {!Object<!Array<!Promise<*>>} gathererResults
    * @return {!Promise<!Artifacts>}
    */
@@ -257,7 +257,7 @@ class GatherRunner {
           }
           artifacts[gathererName] = artifact;
         }, err => {
-          // To reach this point, all errors are recoverable, so return err to
+          // To reach this point, all errors are non-fatal, so return err to
           // runner to handle turning it into an error audit.
           artifacts[gathererName] = err;
         });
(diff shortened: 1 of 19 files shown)


brendankenny:
screen shot 2017-01-27 at 15 48 14
:D

lighthouse-core/gather/gatherers/css-usage.js
```diff
@@ -29,8 +29,7 @@ class CSSUsage extends Gatherer {
       .catch(err => {
         // TODO(phulce): Remove this once CSS usage hits stable
         if (/startRuleUsageTracking/.test(err.message)) {
-          this.failure = 'CSS Usage tracking requires Chrome \u2265 56';
-          return;
+          throw new Error('CSS Usage tracking requires Chrome \u2265 56');
```
brendankenny: @patrickhulce now that errors are non-fatal by default, `afterPass` still throws but this error wins

paulirish:
Looks nice so far!

Do we have a location where we'd quit on a fatal error? Would love to see that in action.

brendankenny:
> Do we have a location where we'd quit on a fatal error

What do you mean? Like a test run really quits? We don't have any gatherer that throws a fatal error yet :) (and I'm not entirely sure what would warrant doing that at this point)

The `"rejects if a gatherer returns a fatal error"` test in `gather-runner-test` checks that gather-runner throws if a gatherer returns a fatal error, though.

brendankenny:
yah? nah?

lighthouse-core/gather/gatherers/css-usage.js
patrickhulce: cool beans

lighthouse-core/runner.js
```diff
@@ -177,6 +177,18 @@ class Runner {
             debugString: `Required ${artifactName} gatherer did not run.`
           });
         }
+
+        // If artifact was an error, it must be non-fatal (or gatherRunner would
+        // have thrown). Output error result on behalf of audit.
+        if (artifacts[artifactName] instanceof Error) {
```
patrickhulce: Any of the same machinery for audits? Or are we still treating all errors there as fatal?

lighthouse-core/gather/gather-runner.js
```diff
@@ -110,14 +110,14 @@ class GatherRunner {
...
-   * the rest.
+   * Test any error output from the promise, absorbing non-fatal errors and
+   * throwing on fatal ones so that run is stopped.
    * @param {!Promise<*>} promise
    * @return {!Promise<*>}
    */
   static recoverOrThrow(promise) {
     return promise.catch(err => {
-      if (!err.recoverable) {
+      if (err.fatal) {
```
patrickhulce: Any of the same machinery for audits? Or are we still treating all errors there as fatal?

patrickhulce:
Nice I like this direction a lot!! Can't wait to get that tracking threaded through...

lighthouse-core/runner.js
brendankenny: > Any of the same machinery for audits? Or are we still treating all errors there as fatal?

we can. There's no reason we can't also get rid of `generateAuditResult({rawValue: -1, debugString: 'whatever'})` and just let audits throw errors and teach the reporter(s) to handle output for those cases

brendankenny:
> LGTM. This is actually fairly low-key

indeed :)

brendankenny:
LGTM? I was planning on merging and doing these in chunks so I don't have to keep rebasing :)

brendankenny:
👍

PR #1561, opened 2017-01-27
Report: do not expand help text by default when audit is violated

ebidel:
R: all

This was discussed at the meeting today and I wanted to see the difference. It's small change but I think it goes a lonnnnnng way to reduce the visual density of the report.

Before:

screen shot 2017-01-27 at 3 46 20 pm

After:

screen shot 2017-01-27 at 3 45 48 pm


paulirish:
yup looks great

PR #1554, opened 2017-01-27
Viewer: add PWA images.

XhmikosR:
Fixes #1506

PR #1571, opened 2017-01-28
Viewer: add sw for ATHS

ebidel:
R: all

Follow up on https://github.com/GoogleChrome/lighthouse/pull/1554. Adds a shell sw so ATHS works.


ebidel:
cc @XhmikosR

XhmikosR:
Pardon my ignorance but what ATHS stands for?

Apart from that the changes look good. I was in the verge when I added manifest.json if it should be in root or not, but agreed this is simpler.

ebidel:
Ah. Short for "Add to Homescreen"

XhmikosR:
You guys make acronyms out of everything :P

lighthouse-viewer/app/src/main.js
```diff
@@ -43,3 +43,10 @@ Promise.all(loadPolyfillPromises).then(_ => {
   // eslint-disable-next-line no-new
   new LighthouseViewerReport();
 });
+
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('/sw.js', {scope: './'}).catch(e => {
```
paulirish: isn't that scope implied? (iirc, the default scope matches the path of the SW resource?)

lighthouse-viewer/app/index.html
```diff
@@ -21,7 +21,7 @@
...
-  <link rel="manifest" href="images/manifest.json">
+  <link rel="manifest" href="manifest.json">
```
paulirish: can you throw the manifest link above the favicon reference?
IMO makes sense before, based on priority.

lighthouse-viewer/app/src/main.js
ebidel: I've been told to ALWAYS use scope. @jeffposnick says so!

jeffposnick: I disavow ever saying that 😄  This specific use case should be fine without providing a value for`scope`, so you might as well leave it out.

My recommendations, for the record: http://stackoverflow.com/a/33881341/385997

lighthouse-viewer/app/index.html
ebidel: done

lighthouse-viewer/app/src/main.js
ebidel: done

ebidel: done

lighthouse-viewer/app/src/main.js
```diff
@@ -43,3 +43,10 @@ Promise.all(loadPolyfillPromises).then(_ => {
   // eslint-disable-next-line no-new
   new LighthouseViewerReport();
 });
+
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('/sw.js').catch(e => {
+    // eslint-disable-next-line no-console
+    console.error('Error during service worker registration:', e);
```
ebidel: actually, we probably want to register this at `${location.pathname}sw.js` so it is restricted to `googlechrome.github.io/lighthouse/viewer/` and also works on localhost.

ebidel: `navigator.serviceWorker.register('sw.js')` for that matter.

ebidel:
PTAL for sanity

lighthouse-viewer/app/sw.js
```diff
@@ -0,0 +1 @@
+// Intentionally empty for now.
```
brendankenny: needs license :)

lighthouse-viewer/app/src/main.js
```diff
@@ -43,3 +43,10 @@ Promise.all(loadPolyfillPromises).then(_ => {
   // eslint-disable-next-line no-new
   new LighthouseViewerReport();
 });
+
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('sw.js').catch(e => {
+    // eslint-disable-next-line no-console
```
brendankenny: since this file is only for running in the browser, want to just do `/* eslint-disable no-console */` at the top of the file? Or is it better to keep it around to cut down on `console` calls left over from development?

(the rule is a bit annoying for debugging in viewer and in the extension since both have the `gulp lint` task as a required prerequisite for `gulp build`, but it's not the end of the world)

lighthouse-viewer/app/sw.js
ebidel: done

lighthouse-viewer/app/src/main.js
ebidel: Just removed the `.catch` for now. We don't care if this shell sw fails to register.

PR #1280, opened 2016-12-23
Add AppVeyor CI support for Windows testing.

XhmikosR:
Fixes #1079.

**Preview**: <https://ci.appveyor.com/project/XhmikosR/lighthouse/history>

**Before merging**: one of the project maintainers needs to add AppVeyor to the repository and paste the image link so that I add it in readme.md and rebase.

PR #1572, opened 2017-01-29
Windows: Look into `LIGHTHOUSE_CHROMIUM_PATH` variable too.

### LINKED ISSUE #1472: Windows: Use `LIGHTHOUSE_CHROMIUM_PATH` if set
Currently, Chrome is always being loaded from the Program Files folder regardless of `LIGHTHOUSE_CHROMIUM_PATH` value.

Noticed this while working on #1280

XhmikosR:
Fixes #1472.

### DIFF
--- a/lighthouse-cli/chrome-finder.ts
+++ b/lighthouse-cli/chrome-finder.ts
@@ -152,6 +152,11 @@ export function win32() {
     process.env.PROGRAMFILES,
     process.env['PROGRAMFILES(X86)']
   ];
+
+  if (canAccess(process.env.LIGHTHOUSE_CHROMIUM_PATH)) {
+    installations.push(process.env.LIGHTHOUSE_CHROMIUM_PATH);
+  }
+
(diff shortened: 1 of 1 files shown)


paulirish:
sg. shall we add it for Mac too just so it's straightforward to reason about?

if this is set, then we should definitely use it rather than any sniffed location.

XhmikosR:
I guess it makes sense to take into consideration the variable. Although it should be documented somewhere.

I need this for #1280 and noticed that only on Linux the env var us taken into consideration.

brendankenny:
+1 for mac support and having it override any other choice (unless you use `--select-chrome` in addition to setting the variable).

win32 will need the `priorities` treatment that the other platforms get, and mac will need to set `process.env.LIGHTHOUSE_CHROMIUM_PATH` as highest priority in the same way linux does

XhmikosR:
@brendankenny: what does that mean? That this won't be merged currently?

PR #1575, opened 2017-01-29
Report: Fix HTML validation errors.

XhmikosR:
BTW, this fixes a 404 link along the way (the link with the double quotes).

### DIFF
--- a/lighthouse-core/formatters/partials/table.html
+++ b/lighthouse-core/formatters/partials/table.html
@@ -2,6 +2,7 @@
 .table_list {
   margin-top: 8px;
   border: 1px solid #EBEBEB;
+  border-spacing: 0;
   max-width: 100%;
 }
 .table_list th,
@@ -56,8 +57,7 @@
 {{#if rows}}
 <details class="subitem__details">
   <summary class="subitem__detail">More information</summary>
-  <table class="table_list {{#if_not_eq headings.length 2}}multicolumn{{/if_not_eq}}"
-         cellspacing="0">
+  <table class="table_list {{#if_not_eq headings.length 2}}multicolumn{{/if_not_eq}}">
     <thead>
       <tr>
         {{#each headings}}<th>{{this}}</th>{{/each}}
--- a/lighthouse-core/report/report-generator.js
+++ b/lighthouse-core/report/report-generator.js
@@ -56,7 +56,7 @@ class ReportGenerator {
 
     // Converts a name to a link.
     Handlebars.registerHelper('nameToLink', name => {
-      return name.toLowerCase().replace(/\s/, '-');
+      return name.toLowerCase().replace(/\s/g, '-');
     });
 
     // Figures out the total score for an aggregation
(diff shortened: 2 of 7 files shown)


XhmikosR:
BTW, this fixes a 404 link along the way (the link with the double quotes).

lighthouse-core/report/report-generator.js
```diff
@@ -56,7 +56,7 @@ class ReportGenerator {
 
     // Converts a name to a link.
     Handlebars.registerHelper('nameToLink', name => {
-      return name.toLowerCase().replace(/\s/, '-');
+      return name.toLowerCase().replace(/\s/g, '-');
```
XhmikosR: This is self-explanatory for me but just in case it isn't, say the `name` is `Progressive Web app`. Previously it became `progressive-web app` which is invalid.

XhmikosR:
Only 2 things left are:

1. #1578 
2. I'd specify the HTML `lang="en"` but being that is being used twice in tests I left it out. I could apply it if you guys agree, though.

lighthouse-core/formatters/partials/table.html
```diff
@@ -2,6 +2,7 @@
 .table_list {
   margin-top: 8px;
   border: 1px solid #EBEBEB;
+  border-spacing: 0;
```
ebidel: 👍

lighthouse-core/report/report-generator.js
ebidel: derp. mind attending a test for this?

XhmikosR: Not sure where and how; I haven't looked into tests yet.

XhmikosR: Would this be enough in /lighthouse-core/test/report/report-test.js

```js
  it('`nameToLink` works properly', () => {
    const reportGenerator = new ReportGenerator();
    const html = reportGenerator.generateHTML(sampleResults);
    assert.ok(/href="#progressive-web-app"/gm.test(html));
  });
```

ebidel: Yea, that's probably best for now. We really don't have a great way to test the registered handlebar helpers atm. The few tests we have, look at the generated output :\ 

Maybe make the asset more specific so there is a smaller chance of false positives in the future. We want make sure "progressive-web-app" is from the `nameToLink` output and not something else. There's also [another place to check](https://github.com/GoogleChrome/lighthouse/blob/5fd1cfeaee82b6484e824a61d0a9f36bd06074c7/lighthouse-core/report/templates/report-template.html#L78):

    assert.ok(/class="menu__link" href="#progressive-web-app"/gm.test(html));
    assert.ok(/id="progressive-web-app"/gm.test(html));

brendankenny: could also make the helpers static (and private) functions so they're testable on their own. It would make the tests a little easier to write and run a little faster.

ebidel: You beat me to it :) https://github.com/GoogleChrome/lighthouse/issues/1581

XhmikosR:
The Travis error seems unrelated to my changes.

ebidel:
Rebased for ya :)

XhmikosR:
Rebased, should be OK now. Only the `Element style not allowed as child of element li in this context.` errors are left.

XhmikosR:
I wouldn't have caught the 2 issues, one with the 404 link and the second with the spaces in the anchors if it weren't for the validator.

IMO it's a good safety measure to have it run and be HTML valid as close as possible :)

PR #1577, opened 2017-01-30
Add 2 test cases for URL blocking

WeiweiAtGit:
Ensure adding blocked URL patterns works correctly.

### DIFF
--- a/lighthouse-core/test/gather/gather-runner-test.js
+++ b/lighthouse-core/test/gather/gather-runner-test.js
@@ -46,7 +46,7 @@ class TestGathererNoArtifact extends Gatherer {
 
 const fakeDriver = require('./fake-driver');
 
-function getMockedEmulationDriver(emulationFn, netThrottleFn, cpuThrottleFn) {
+function getMockedEmulationDriver(emulationFn, netThrottleFn, cpuThrottleFn, blockUrlFn) {
   const Driver = require('../../gather/driver');
   const Connection = require('../../gather/connections/connection');
   const EmulationDriver = class extends Driver {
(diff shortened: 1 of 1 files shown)


lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -222,6 +225,28 @@ describe('GatherRunner', function() {
     });
   });
 
+  it('tells the driver to block given URL patterns when blockedUrlPatterns is given', () => {
+    const receivedUrlPatterns = [];
+    const urlPatterns = ['http://*.evil.com', '.jpg', '.woff2'];
```
paulirish: ".jpg" is equivalent to "*.jpg*" ?? if so, i didn't know that. :)

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -222,6 +225,28 @@ describe('GatherRunner', function() {
...
+    const driver = getMockedEmulationDriver(null, null, null, params => {
+      receivedUrlPatterns.push(params.url);
+    });
+
+    return GatherRunner.setupDriver(driver, {flags: {blockedUrlPatterns: urlPatterns.slice()}})
+      .then(() => assert.equal(receivedUrlPatterns.sort().join(), urlPatterns.sort().join()));
```
paulirish: not sure if `assert.equal` means much on these arrays. do we need `deepEqual` or `deepStrictEqual` ? /me shrugs.

lighthouse-core/test/gather/gather-runner-test.js
WeiweiAtGit: Yes. They are the same. I was also surprised.

lighthouse-core/test/gather/gather-runner-test.js
WeiweiAtGit: I'm not comparing arrays here. I sorted them and join the elements as a string respectively, then compare the strings. But yeah, `deepStrictEqual` seems better. Thx for your advice.

PR #1580, opened 2017-01-30
Cleanup: array.indexOf -> includes!

ebidel:
R: all

Now that we're on Node 6+, we can use awesome new features.


brendankenny:
woo!

ebidel:
done

PR #1584, opened 2017-01-31
feat: save each screenshot asset individually

patrickhulce:
Came across this while working towards WPT integration. Not sure what the use case was for the screenshots file, so I might be missing context of the original decision.

**New Output:**
image

### DIFF
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -118,9 +118,11 @@ function prepareAssets(artifacts, results) {
       .then(screenshots => {
         const traceData = Object.assign({}, trace);
         const html = screenshotDump(screenshots, results);
+        const imageData = screenshots.map(shot => shot.datauri.replace(/^.*?base64,/, ''));
 
         assets.push({
           traceData,
+          imageData,
           html
         });
       });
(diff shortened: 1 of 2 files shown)


paulirish:
We don't have a strong reason to keep around the `screenshots.html` file, tbh.  :)

But the downside of this is that we could end up making > 100 files, whereas before we are capped at 3 files max. 

Do we already have a need for these individual screenshot files?

patrickhulce:
> But the downside of this is that we could end up making > 100 files, whereas before we are capped at 3 files max.

Yeah, that's less than ideal. Making it an option sounds like it leads down the road to esoteric CLI flags and a matrix of flag compatibility. I definitely have a need to be able to consume individual JPEG files, but I could save them just as JSON directly? That seemed unusable for every other use case though, unless we output the HTML page *and* the JSON it uses.

brendankenny:
doesn't WPT use video recording? Are there limitations to using screenshots instead? (I can only think of that the debugger protocol stops sending over screenshots after a certain number)

patrickhulce:
> doesn't WPT use video recording?
Yes, but it collects screenshots for the report too since video isn't always enabled.

paulirish:
I would prefer this is under some other flag (probably that's not exposed to CLI right now) for the time being. 
Other than that, it works for me.

patrickhulce:
WPT doesn't need anymore so I'm going to punt on this.

PR #1585, opened 2017-01-31
Update mocha to v3.2.0.

XhmikosR:
Fixes various `npm i` warnings.

Fixes #1539.

PR #1586, opened 2017-01-31
Update yarn.lock with the latest dependencies.

XhmikosR:
Not sure what the policy is on this, I just noticed yarn.lock was pretty outdated when making #1585.


XhmikosR:
I'm gonna close this since by the time it will be merged it will be already obsolete.

PR #1587, opened 2017-01-31
Travis: Add `npm prune` before caching.

XhmikosR:
This will make sure cache is kept at the minimal state and clean.


XhmikosR:
I'm trying to see if this is actually needed. Not sure what exactly Travis caching is doing or if `npm i` takes cares of this. So this might not even be needed.

XhmikosR:
After looking into this, I believe it's not needed currently.

PR #1588, opened 2017-01-31
refactor: use public accessors for network records

patrickhulce:
Brought over from [discussion](https://github.com/GoogleChrome/lighthouse/pull/1513#discussion_r98522727)

The two biggest holdouts are `_resourceType` and `_initiator` which are functions and not getter properties.


brendankenny:
nice, though this does move the problem mentioned in https://github.com/GoogleChrome/lighthouse/pull/1513#discussion_r98522727 to the tests, so if the implementation of the getters change the test fixtures will presumably have to be updated.

Not sure how much of a danger that is, though, and I'd say simulating the getters in the test fixtures is definitely better than accessing the internal properties in the actual gatherers/audits...

brendankenny:
after discussion today, sounds like the solution is this PR + storing performance logs (or whatever format) for network activity for tests. Tests would use something like [`NetworkRecorder.recordsFromLogs()`](https://github.com/GoogleChrome/lighthouse/blob/81b891d491e95135f1f5daf4278e1e27c93ac510/lighthouse-core/lib/network-recorder.js#L147) to convert them into the `networkRecords` format gatherers and audits expect, and we can drop all use of the internal variable names for tests.

Main things we'd need to establish would be an easy format to create for creating test files and (if we don't end up using the performance logs we currently support) an easy function for conversion to `networkRecords`

paulirish:
honestly we're not in a big rush here. :)
let's drop this for now.

Issue #1512, opened 2017-01-23
Way to surface non-critical errors in the report


ebidel:
From https://github.com/GoogleChrome/lighthouse/pull/1492#pullrequestreview-17592494

For individual gatherers/audit errors, we surface `debugString`.

Right now, fatal errors stop the extension from running and we show a "report error" button. However, for non-fatal errors that happen outside of the gatherer -> audit -> report pipeline, we don't have a way to surface those errors to extension users.

Without analytics, we don't know about non-fatal errors. The HTML report could have a section where users see overall LH non-fatal errors and file bugs against us.

brendankenny:
It seems this could be broken down into
- a general system for logging or capturing errors in Lighthouse
- having those errors available for output in some way: surfacing in the report, error reporting if that's enabled, etc

brendankenny:
after #1591, we probably want to add the errors in the gatherer -> audit -> report pipeline to this as well. Some things might not make sense, but e.g. if we couldn't find FCP and didn't calculate a bunch of perf metrics it should probably be called out at the top as well (this was the spirit of @addyosmani's original #812)

We may want a better icon to mark them in the report as well :)

brendankenny:
one more thought for the CLI: as @patrickhulce notes in https://github.com/GoogleChrome/lighthouse/pull/1605#issue-204981590, if something gets logged at the beginning of the LH run, it probably won't get noticed. If we figure out a nice way to pipe errors to the report (or whatever), we could also drop them in the log at the end of the CLI run so they're very visible there too.

brendankenny:
yet another warning in the report issue. dupe of #1825 (or the other way around)

brendankenny:
List of things to warn on:
- Run on `localhost`, so `is-on-https` warning (amongst others?) might not be representative of when site moves to production/real server - (formerly #3228)
- redirected from HTTP to HTTPS (should only happen in the CLI), so `is-on-https` passes (after #3417) but risk of downgrade attack and performance impact of redirect (details in #3417)

patrickhulce:
addressed by #3692

Issue #1175, opened 2016-12-18
Skip HTTPS tests when running a 127.0.0.1 or localhost or .local URL


gauntface:
The following tests could be skipped for local URLs:

- Site is on HTTPS 
- Site redirects HTTP traffic to HTTPS
- Security: Site is on HTTPS

I'm also not sure if the HTTP/2 test makes sense for localhost, can you test H/2 on local origin?

wardpeet:
Might be ok to do this because we do this as well for localhost.
```
if (!parsedURL.protocol.includes('https') && parsedURL.hostname !== 'localhost') {
      log.warn('Lighthouse', 'The URL provided should be on HTTPS');
      log.warn('Lighthouse', 'Performance stats will be skewed redirecting from HTTP to HTTPS.');
}
```

XhmikosR:
I'd be interested in this. One note though: this should skip the tests if `127.0.0.1` is being used too.

patrickhulce:
I'd be interested in hearing more about the workflow of using lighthouse on internal/local servers. Is it generally production builds on a staging or development server for the developer to feel out performance iteratively? Is it for CI where a failure can't just be ignored and should fail the build?

Interested to hear what options would typically be set as I'd prefer to make config changes more visible/easier instead of one-off `if` checks for various hostnames :)

gauntface:
The reason I raised this issue initially was that gut checking changes on Lighthouse via the extension is somewhat skewed by the fact that there are results I can't pass on localhost / aren't worth making work on localhost.

I have been using it for CI of a "production version" of my site (Write up here: https://medium.com/@gauntface/the-old-rebuild-chestnut-4210e389647#.xpj3zm3za) The tests actually filter out some audit results because the scores fail / have invalid scores (i.e. a tti of -1). So for CI. this is easy enough to side-step the troubling audits.

patrickhulce:
Great writeup thanks @gauntface! Do you happen to have the lighthouseWrapper you described available anywhere? Maybe we'd be able to address some of the issues you're facing.

And interesting, so basically looking for automatically unchecking the "Network connections are secure" option in the extension?

gauntface:
Sure, the wrapper is here: https://github.com/gauntface/gf-site/blob/v3/test/utils/lighthouse-wrapper.js

For the most part it's just wrapping launching, using and killing chrome.

ebidel:
[...]

PR #1591, opened 2017-02-01
add explicit result error property for erroneous audit results

### LINKED ISSUE #1148: Getting Error: Expected expectedValue of type boolean, got number
```Runtime error encountered: Error: Expected expectedValue of type boolean, got number: Unable to get document body innerText: result was not a string```

Linux 16
node v7.0.0

### PR DESCRIPTION
**Pull request by brendankenny**
Related to #941 
Fixes #1148
Makes #1567 non-fatal again

This PR does two things
- audits can now throw non-fatal errors (like gatherers after #1560) and the `Error` object will be converted into an error result . Any audits that require artifacts from gatherers that have already been converted to using errors will get this behavior today.
- these error results are explicitly handled by `aggregate.js` instead of throwing.

I was going hold off on doing this, but it turns out to be necessary for converting gatherers to using proper `Error` objects for errors. When Runner got an error artifact, [it was outputting `-1` results from audits that depend on them](https://github.com/GoogleChrome/lighthouse/blob/4442f53cb6da4b005733674f0ae996f4c2e8a70c/lighthouse-core/runner.js#L189-L191), but quickly ran into an issue with audits that have expected boolean values, since [aggregate checks that the audit result and the expected result have the same type](https://github.com/GoogleChrome/lighthouse/blob/4442f53cb6da4b005733674f0ae996f4c2e8a70c/lighthouse-core/aggregator/aggregate.js#L90), otherwise it throws (this actually happens on master, such as in #1148, since e.g. the [viewport audit used to use `-1` as its error score](https://github.com/GoogleChrome/lighthouse/blob/539241e78e4549921967f5f0ae78b4f3afc59846/lighthouse-core/audits/viewport.js#L43-L46) but expects `true`/`false` in the config).

One benefit is it takes the currently fatal error from the missing FCP event (#1567) and automatically turns it into a error audit result now. Going forward we can convert all the audits to this system, which will simplify audits and audit tests much like what's been going on with gatherers.

### DIFF
--- a/lighthouse-cli/printer.ts
+++ b/lighthouse-cli/printer.ts
@@ -110,7 +110,9 @@ function createOutput(results: Results, outputMode: OutputMode): string {
         if (auditResult.comingSoon === true)
           return;
 
-        let lineItem = ` ${log.doubleLightHorizontal} ${formatAggregationResultItem(auditResult.score)} ${auditResult.description}`;
+        const formattedScore = auditResult.error ? `${log.redify('‽')}` :
+            `${formatAggregationResultItem(auditResult.score)}`;
+        let lineItem = ` ${log.doubleLightHorizontal} ${formattedScore} ${auditResult.description}`;
         if (auditResult.displayValue) {
           lineItem += ` (${log.bold}${auditResult.displayValue}${log.reset})`;
         }
--- a/lighthouse-core/aggregator/aggregate.js
+++ b/lighthouse-core/aggregator/aggregate.js
@@ -77,6 +77,12 @@ class Aggregate {
       throw new Error(msg);
     }
 
+    // Audit resulted in an error, so doesn't contribute to score.
+    // TODO: could do NaN instead, as score becomes somewhat meaningless.
+    if (result.error) {
+      return 0;
+    }
+
     if (typeof result === 'undefined' ||
         typeof result.score === 'undefined') {
       let msg =
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -34,6 +34,18 @@ class Audit {
   }
 
   /**
+   * @param {string} debugString
+   * @return {!AuditResult}
+   */
+  static generateErrorAuditResult(debugString) {
+    return this.generateAuditResult({
+      rawValue: null,
+      error: true,
+      debugString
+    });
+  }
+
+  /**
    * @param {!AuditResultInput} result
    * @return {!AuditResult}
    */
@@ -57,6 +69,7 @@ class Audit {
       score,
       displayValue: `${displayValue}`,
       rawValue: result.rawValue,
+      error: result.error,
       debugString: result.debugString,
       optimalValue: result.optimalValue,
       extendedInfo: result.extendedInfo,
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -33,6 +33,7 @@ span, div, p, section, header, h1, h2, li, ul {
   --good-color: #1ac123;
   --average-color: #ffae00;
   --unknown-color: #b3b3b3;
+  --warning-color: #f6be00;
   --gutter-gap: 12px;
   --gutter-width: 40px;
   --body-font-size: 14px;
@@ -895,6 +896,9 @@ body {
 }
 .score-poor-bg {
   background-color: var(--poor-color);
+}
+.score-warning-bg {
+  background-color: var(--warning-color);
 }
 .score-unknown-bg {
   background-color: var(--unknown-color);
--- a/lighthouse-core/report/templates/report-template.html
+++ b/lighthouse-core/report/templates/report-template.html
@@ -189,6 +189,8 @@ limitations under the License.
                   <div class="subitem-result">
                     {{#if subItem.comingSoon}}
                           <span class="subitem-result__unknown score-unknown-bg">N/A</span>
+                    {{else if subItem.error}}
+                      <span class="subitem-result__unknown score-warning-bg">N/A</span>
                     {{else}}
                       {{#if (is-bool subItem.score)}}
                         {{#if subItem.score}}
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -174,10 +174,7 @@ class Runner {
         if (noArtifact || noTrace) {
           log.warn('Runner',
               `${artifactName} gatherer, required by audit ${audit.meta.name}, did not run.`);
-          return audit.generateAuditResult({
-            rawValue: -1,
-            debugString: `Required ${artifactName} gatherer did not run.`
-          });
+          throw new Error(`Required ${artifactName} gatherer did not run.`);
         }
 
         // If artifact was an error, it must be non-fatal (or gatherRunner would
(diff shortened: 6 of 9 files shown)

### CODE REVIEW

File: lighthouse-core/aggregator/aggregate.js [hunk 1]
```diff
@@ -77,6 +77,12 @@ class Aggregate {
       throw new Error(msg);
     }
 
+    // Audit resulted in an error, so doesn't contribute to score.
```
brendankenny: For scoring: we're doing kind of a dumb thing on master currently, adding `-1/100` as the score for audits currently returning an error result of -1.

Originally I was going to return `NaN` here for error audit scores, but that seemed unfair to do until all audits had been converted to this system (e.g. an error in user-timing would become a NaN, but an error in first-meaningful-paint would simply be `-.001`), so I manually set those to contributing 0 to match current master behavior as much as possible.

We can reconsider this (in tandem with #1512) when all audits have been converted.

lighthouse-core/runner.js
```diff
@@ -186,14 +183,20 @@ class Runner {
...
       }
 
       return audit.audit(artifacts);
+    }).catch(err => {
+      if (err.fatal) {
+        throw err;
+      }
+
+      // Non-fatal errors become the result of the audit. Decorate with name.
+      err.name = audit.meta.name;
```
patrickhulce: hmm this will throw away `TypeError` and soforth right? Do we want to preserve any of that?

What does it look like to the user, similar to debugString? I don't see any logic added for it was it already happening :)

brendankenny: > hmm this will throw away TypeError and soforth right? Do we want to preserve any of that?

yeah, I wasn't wild about this, but the assumption that the audit result will have the audit name on `name` is built deep in several places in the runner and if you provide `auditResults` in the config (which affects a bunch of tests), so this seemed easier. Not sure if we really care about the name.

> What does it look like to the user, similar to debugString?

somehow I skipped this pretty important step :)

lighthouse-core/aggregator/aggregate.js
ebidel: @brendankenny and I talked this out. I think a "less complete" score that factors in fewer audits makes more sense than seeing `NaN` for the entire category. Basically, users should never see `NaN`. That looks like a bug.

brendankenny: > Basically, users should never see `NaN`. That looks like a bug

one question I'm immediately running into is what *should* it display in the report for these error results. In theory I could continue doing -1 for the numeric ones, but that's not great, and what about the boolean ones?

Could do red X from failing audits, greyed out icon from the coming soon audits, some kind of question mark icon, ..? Anyone have a good idea? Doesn't have to be our symbol for life :)

ebidel: Hmm, yea. 

I like the greyed out "-" icon for the failing ones. We should ditch the "coming soon" audits. They've been coming soon for.....ever. We could also use that icon for tests that we know are not passable for the user's setup (https://github.com/GoogleChrome/lighthouse/issues/1175#issuecomment-275476565) as a way to indicate something is up.

patrickhulce: I like the combo of displaying "-" in red and showing the message in a similar manner to debugString 👍

brendankenny:
ok, take 2 :)

For a couple of reasons `Error` objects don't work for audit results: they don't make it through JSON serialization (affecting the report generator and saving/loading audit results to JSON), they don't have a bunch of properties we expect, etc. This PR now takes a very simple approach and adds an `error`property to audit results, indicating whether or not there was a problem. `debugString` is still used as the error message, and the score becomes `null` (which does make it through JSON, unlike `NaN`). The `Audit` base class now has a `generateErrorAuditResult()` function to make this easy.

brendankenny:
Updated PR description so it's correct for the new version.

lighthouse-cli/printer.ts
```diff
@@ -110,7 +110,13 @@ function createOutput(results: Results, outputMode: OutputMode): string {
         if (auditResult.comingSoon === true)
           return;
 
-        let lineItem = ` ${log.doubleLightHorizontal} ${formatAggregationResultItem(auditResult.score)} ${auditResult.description}`;
+        let lineItem = ` ${log.doubleLightHorizontal}`;
+        if (auditResult.error) {
+          lineItem += ` ${log.redify('-')}`;
```
brendankenny: screen shot 2017-02-01 at 15 28 35

this is kind of weak, so feel free to suggest a different character :)

lighthouse-core/report/templates/report-template.html
```diff
@@ -189,6 +189,8 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
                   <div class="subitem-result">
                     {{#if subItem.comingSoon}}
                           <span class="subitem-result__unknown score-unknown-bg">N/A</span>
+                    {{else if subItem.error}}
+                      <span class="subitem-result__unknown score-error-bg">N/A</span>
```
brendankenny: this ends up looking like

screen shot 2017-02-01 at 15 31 48

lighthouse-cli/printer.ts
ebidel: "???". Not sure if that's much better.

lighthouse-core/report/templates/report-template.html
ebidel: Red is good b/c it signifies an error, but the icon is different (LH error vs. audit failure) so hopefully people will see the difference. "Audit error" in the debug string should help a lot.

lighthouse-core/audits/audit.js
```diff
@@ -34,6 +34,18 @@ class Audit {
...
 
   /**
+   * @param {string} debugString
+   * @return {!AuditResult}
+   */
+  static generateErrorAuditResult(debugString) {
+    return this.generateAuditResult({
+      rawValue: null,
+      error: true,
+      debugString: debugString
```
ebidel: nit: `debugString: debugString` -> `debugString`

brendankenny: 💥

lighthouse-cli/printer.ts
paulirish: > ‽

brendankenny: done :)

lighthouse-cli/printer.ts
```diff
@@ -110,7 +110,13 @@ function createOutput(results: Results, outputMode: OutputMode): string {
         if (auditResult.comingSoon === true)
           return;
 
-        let lineItem = ` ${log.doubleLightHorizontal} ${formatAggregationResultItem(auditResult.score)} ${auditResult.description}`;
+        let lineItem = ` ${log.doubleLightHorizontal}`;
+        if (auditResult.error) {
```
patrickhulce: nit: I think keeping the initialization inline and having a 
`const formattedScore = auditResult.error ? log.redify('') : formatAgg...` might be slightly more understandable then the concatenation but not up in arms over in

brendankenny: I like it

lighthouse-core/report/styles/report.css
```diff
@@ -894,6 +894,9 @@ body {
 .score-poor-bg {
   background-color: var(--poor-color);
 }
+.score-error-bg {
+  background-color: var(--poor-color);
```
brendankenny: I chose like...orangish. Based on devtools' alert icon when you do an invalid css property. Happy to change more :)

PR #1593, opened 2017-02-01
Fix some cross-browser compatibility issues

WeiweiAtGit:
In interactive mode, the report page will be opened in the default browser. However, the interactive mode report page only works in Chrome currently.
This PR adds support for Safari and Firefox.

### DIFF
--- a/lighthouse-cli/performance-experiment/report/scripts/perf-x.js
+++ b/lighthouse-cli/performance-experiment/report/scripts/perf-x.js
@@ -31,7 +31,8 @@ class ConfigPanel {
     this._messageField = this._configPanel.querySelector('.js-message');
     this._urlBlockingList = this._configPanel.querySelector('.js-url-blocking-patterns');
     this._urlBlockingStatus = {};
-    this._reportId = new URL(window.location).searchParams.get('id');
+    const match = location.search.match(/[?&]id=([^&]*)/);
+    this._reportId = match ? match[1]: '';
 
     const bodyToggle = this._configPanel.querySelector('.js-panel-toggle');
     bodyToggle.addEventListener('click', () => this._toggleBody());
@@ -83,15 +84,6 @@ class ConfigPanel {
         }
       });
     });
-
-    // get and recover blocked URL patterns of current run
-    fetch(`/flags?id=${this._reportId}`).then(response => {
-      return response.json();
-    }).then(flags => {
-      const blockedUrlPatterns = flags.blockedUrlPatterns || [];
-      blockedUrlPatterns.forEach(urlPattern => this.addBlockedUrlPattern(urlPattern));
-      this.log('');
-    });
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lighthouse-cli/performance-experiment/report/scripts/perf-x.js
```diff
@@ -31,7 +31,8 @@ class ConfigPanel {
     this._messageField = this._configPanel.querySelector('.js-message');
     this._urlBlockingList = this._configPanel.querySelector('.js-url-blocking-patterns');
     this._urlBlockingStatus = {};
-    this._reportId = new URL(window.location).searchParams.get('id');
+    const match = location.search.match(/[?&]id=([^&]*)/);
```
ebidel: I use a polyfill in Viewer for URLSearchParams: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-viewer/package.json#L33

Could you use that here?

lighthouse-cli/performance-experiment/report/scripts/perf-x.js
```diff
@@ -83,15 +84,6 @@ class ConfigPanel {
         }
       });
     });
-
-    // get and recover blocked URL patterns of current run
-    fetch(`/flags?id=${this._reportId}`).then(response => {
```
ebidel: you could also use a polyfill for this: 
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-viewer/package.json#L34

ebidel: But Safari 10.1 also has `fetch()`. A polyfill might not be worth it.

lighthouse-cli/performance-experiment/report/scripts/perf-x.js
```diff
@@ -100,15 +92,24 @@ class ConfigPanel {
...
       blockedUrlPatterns: this.getBlockedUrlPatterns()
     };
 
-    return fetch(`/rerun?id=${this._reportId}`, {method: 'POST', body: JSON.stringify(options)})
-      .then(response => response.text())
-      .then(newReportId => location.assign(`?id=${newReportId}`))
-      .catch(err => this.log(`Lighthouse Runtime Error: ${err}`));
+    const xhr = new XMLHttpRequest();
+    xhr.open('POST', `/rerun?id=${this._reportId}`, true);
+    xhr.onreadystatechange = () => {
```
ebidel: just use `xhr.onload`, then you don't need to both with `readyState` checking.

lighthouse-cli/performance-experiment/report/scripts/perf-x.js
WeiweiAtGit: thanks for your advice. Shall I move the dependencies from viewer to core? Or just add some dependencies to cli?

ebidel: We need to keep the deps for the viewer since it's a separate app. 

Since you've already done the work of moving over to the fallbacks, there's not a lot of benefit to adding polyfills for two small APIs. I'm fine with what you've got.

WeiweiAtGit: :)

lighthouse-cli/performance-experiment/report/scripts/perf-x.js
paulirish: lets not polyfill fetch()

lighthouse-cli/performance-experiment/report/scripts/perf-x.js
paulirish: can you add a comment here?  `// Compat: URL.searchParams isn't yet supported by _______ `
so later when that browser updates we'll know when we can remove it

PR #1169, opened 2016-12-16
WIP: Move copy button to main report

ebidel:
R:  all

This is the first part of #1166. It moves the copy button and butter bar functionality to the core report. While I was in report.css, I also tweaked the print stylesheet a bit.

This PR also introduces a package.json and gulpfile to lighthouse-core to browserify and build the report.js. Alternatively, we could move this up into the root and do build stuff there. It also introduces a `postinstall` script to mitigate folks checking out LH and not having the built report js.

The benefit of browserifying the report code is that we can author in es6 and have the report work in older browsers!

Feel free to bikeshed the build changes here.

PR #1594, opened 2017-02-01
Consolidate viewer functionality into main report

ebidel:
R: all

This is the finale of https://github.com/GoogleChrome/lighthouse/pull/1169. But unlike that PR, this does not introduce a build step to lighthouse-core!

Now, all generated reports (cli/extension/viewer) have the full export button goodness:

screen shot 2017-01-31 at 5 53 51 pm

The share button is still specific to viewer:

screen shot 2017-01-31 at 5 54 13 pm


lighthouse-core/report/scripts/lighthouse-report.js
```diff
@@ -26,27 +26,147 @@ class LighthouseReport {
...
+      dropdown.addEventListener('click', this.onExport);
+
+      document.addEventListener('copy', this.onCopy);
+    }
+  }
+
+  /**
+   * Handler copy events.
+   */
+  onCopy(e) {
```
ebidel: mostly moving code from viewer -> lighthouse-report.js

lighthouse-core/report/scripts/lighthouse-report.js
```diff
@@ -99,10 +219,37 @@ class LighthouseReport {
...
+    a.click();
+
+    setTimeout(_ => URL.revokeObjectURL(a.href), 500); // cleanup.
+  }
 }
 
-// If in Node, allow others to require us.  By default, browser code can juse
-// use the LighthouseReport class.
-if (typeof module !== 'undefined') {
+// Allow Node code to use require(). Browser code can use the LighthouseReport
```
ebidel: Open to suggestions on this. There might be a Node-ier want to do this but I think it's simple enough.

Viewer still browserifies the code, so we keep requires for that and omit them for the browser context.

lighthouse-core/report/scripts/logger.js
```diff
@@ -67,4 +67,6 @@ class Logger {
   }
 }
 
-module.exports = new Logger('#log'); // singleton
```
ebidel: got rid of the nasty dom singleton. Viewer and the core report each defines their own logger now.

lighthouse-core/report/templates/report-template.html
```diff
@@ -245,8 +247,41 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
 {{/each}}
 
 <script>
-window.addEventListener('DOMContentLoaded', _ => {
-  new LighthouseReport(window.lhresults);
+(function(exports) {
+'use strict';
+
+/**
+ * Note: getFilenamePrefix is duplicated from asset-saver.js
```
ebidel: I'm up for suggestions on this too. We could factor out `getFilenamePrefix` from asset-saver.js and make it work in both Node and the browser.

lighthouse-viewer/test/test-helpers.js
```diff
@@ -26,11 +26,14 @@ const PAGE = fs.readFileSync(path.join(__dirname, '../app/index.html'), 'utf8');
 function setupJsDomGlobals() {
   global.document = jsdom.jsdom(PAGE);
   global.window = global.document.defaultView;
+  global.logger = console;
```
ebidel: stub

ebidel:
rebased

ebidel:
rebased

lighthouse-core/test/report/report-test.js
```diff
@@ -78,15 +81,11 @@ describe('Report', () => {
     assert.ok(html.includes('&quot;lighthouseVersion'), 'lhresults were escaped');
     assert.ok(/Version: x\.x\.x/g.test(html), 'Version doesn\'t appear in report');
     assert.ok(html.includes('export-button'), 'page includes export button');
-
-    assert.ok(html.includes('printButton = document.querySelector'),
-                            'print button functionality attached');
-    assert.ok(html.includes('openButton = document.querySelector'),
-                            'open button functionality attached');
     assert.ok(html.includes('share js-share'), 'has share button');
```
paulirish: i guess it'd be a pain to assert that the share icon has display:none in this case, yeah?

ah well.

ebidel:
PTAL. Moved out `getFilenamePrefix` into a separate file so it could be require'd or used directly in the browser.

brendankenny:
this might need a fix now that `getFilenamePrefix` is used in a few other places due to #1601, sorry :(

lighthouse-core/lib/file-namer.js
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Generate a filenamePrefix of hostname_YYYY-MM-DD_HH-MM-SS
+ * Date/time uses the local timezone, however Node has unreliable ICU
+ * support, so we must construct a YYYY-MM-DD date format manually. :/
+ * @param {!Object} results
+ * @returns string
+ */
+function getFilenamePrefix(results) {
+  // eslint-disable-next-line no-undef
+  const hostname = new (URLConstructor || URL)(results.url).hostname;
```
brendankenny: `url-shim` already guards against this. Any way we could rely on that and avoid this (and the `require()` below)? Since `file-namer` isn't included via script tag it seems like it's doable.

lighthouse-core/report/scripts/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  "extends": "../../../.eslintrc.js",
+  "env": {
+    "browser": true
```
brendankenny: 👍

brendankenny:
> Part of the issue is probably `LighthouseViewerReport` extending `LighthouseReport` but also creating a `LighthouseReport` to display it, just not including any of the script parts of it. I wonder if there's more we can do to tease this apart in the future.

Thinking about this more, I realize part of this setup is precisely because we're trying to avoid a build step in lighthouse-core :)

lighthouse-core/lib/file-namer.js
ebidel: I couldn't see a way around it. 

For node code, we need to `require('url-shim')` to get `URL`. But if you do this instead:

```
let URL;
  URL = require('./url-shim');
  ...
}
```

You'll end up with `URL === undefined` in the browser. filer-namer.js gets `<script>` included in the main report here: https://github.com/GoogleChrome/lighthouse/pull/1594/files#diff-6a4779f17aa030f3223e4de92085eda9R227

ebidel:
> this might need a fix now that getFilenamePrefix is used in a few other places due to #1601 

We should be covered. `asset-saver.js` imports `require('file-namer.js').getFilenamePrefix` and still exports it for backwards compatibility.

lighthouse-core/report/templates/report-template.html
```diff
@@ -247,8 +249,14 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
 {{/each}}
 
 <script>
-window.addEventListener('DOMContentLoaded', _ => {
-  new LighthouseReport(window.lhresults);
+(function(exports) {
```
brendankenny: what do you think about moving this into its own script (report-startup or whatever) and included in `ReportGenerator.getReportJS` if `reportContext` is cli or extension? That seems like it may be clearer it's just for that context, but I'm not sure

brendankenny:
> We should be covered. `asset-saver.js` imports `require('file-namer.js').getFilenamePrefix` and still exports it for backwards compatibility.

Ah, I see. It seems like that should be dropped and the reference and tests moved from `asset-saver-test` to `file-namer-test`, though

lighthouse-core/report/templates/report-template.html
ebidel: We could, but does it buy anything?

Viewer has [similar init code](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-viewer/app/src/main.js#L44) and uses a main.js. But the only reason I put that a separate script was because the code uses `require()`.

brendankenny: My thinking was that it may make the execution flow clearer since `report-template.html` is used for the report and viewer but this script is only for report, not viewer

brendankenny: (come to think of it, none of the scripts injected by `getReportJS` are used by the viewer, right?)

ebidel: Right. Nothing [here](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/templates/report-template.html#L228-L247) ends up in Viewer. Because of that, we should avoid adding things to `getReport*` that are specific to the main report and never used elsewhere. This is a good example, where the init code is specific to the page. Otherwise the browserified bundle gets larger :\

ebidel:
> Ah, I see. It seems like that should be dropped and the reference and tests moved from asset-saver-test to file-namer-test, though

Done. PTAL

lighthouse-core/report/templates/report-template.html
brendankenny: > Because of that, we should avoid adding things to `getReport*` that are specific to the main report and never used elsewhere

arg, see this is why we need a diagram :S

If none of the `getReportJS` scripts end up used in viewer, how does that make the browserified bundle bigger?

brendankenny: oh, is it because we statically include them and `brfs` inlines them? `brfs` doesn't have an explicit `ignore` option, but we could probably add that to cut down on the viewer bundle

ebidel: "Ends up in the viewer" in the sense that viewer doesn't run [any of this stuff](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/templates/report-template.html#L228-L247) itself.

What does end up in viewer is anything we put in report-generator.js. For example, any html/css/js file we read as a string ends up in the browseried bundle. If you look in https://googlechrome.github.io/lighthouse/viewer/src/main.js, you can see html snippets, js file contents, css, data urls. It ends up being lots of stuff.

ebidel: > statically include them and brfs inlines them

Yep. But we can't ignore them. Remember, we still need those scripts for the "export as html" option. When users click that, we create a full version of the report as if it was generated by the CLI/extension. So it needs all the same juice. It's the [setup work](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/templates/report-template.html#L228-L247) that we leave out.

One day I will draw you a 🖼 ...on 🗒 .

brendankenny: > Yep. But we can't ignore them. Remember, we still need those scripts for the "export as html" option. When users click that, we create a full version of the report as if it was generated by the CLI/extension. So it needs all the same juice. It's the setup work that we leave out.

lol, but then it doesn't matter size-wise where the script lives, it's either inlined into the inlined `getReportTemplate()` html or in the inlined `getReportJS()` scripts

> One day I will draw you a 🖼 ...on 🗒 

yes please

ebidel: our own conservation of mass!

lighthouse-core/report/templates/report-template.html
```diff
@@ -241,8 +243,14 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
...
+
+exports.addEventListener('DOMContentLoaded', _ => {
+  exports.logger = new Logger('#log');
+  new LighthouseReport(exports.lhresults);
 });
+
+})(window);
 </script>
-{{/if}}
+{{/if}}
```
brendankenny: new line

lighthouse-core/report/styles/report.css
```diff
@@ -1040,7 +1079,11 @@ body {
...
-:root[data-report-context="viewer"] .share,
-:root[data-report-context="viewer"] .export-section {
+:root[data-report-context="viewer"] .share {
   display: initial;
 }
+
+/* app z-indexes */
+.log-wrapper {
+  z-index: 3;
+}
```
brendankenny: new line

lighthouse-core/report/report-generator.js
```diff
@@ -222,7 +222,11 @@ class ReportGenerator {
     if (reportContext === 'devtools') {
       return [];
     } else {
-      return [fs.readFileSync(path.join(__dirname, './scripts/lighthouse-report.js'), 'utf8')];
+      return [
+        fs.readFileSync(path.join(__dirname, './scripts/logger.js'), 'utf8'),
```
brendankenny: nit: can drop the `path.join` and just `fs.readFileSync(__dirname + '/scripts/logger.js'), 'utf8')`

lighthouse-core/report/templates/report-template.html
paulirish: okay on the one hand we have 
image
but then here we have
image

so yeah im pretty confused as well.

paulirish: okay here's how report-generator is consumed:
image

`viewer/app/src/main` and `lighthouse-background` are browserified and used in the viewer and extension/devtools, respectively.

question:

* why is report-generator including the source of `scripts/lighthouse-report` into the report when that file was already part of the bundle that shipped with the viewer

also, can we rename some of these files, because they all sound like the same thing right now. :)

ebidel: Is the confusion mostly about the inline `<script>` at the bottom of report-template.html (@paulirish's 2nd screenshot)? I can create a separate .js file with the `DOMContentLoaded` setup  and load it in the `getReportJS()` array. Would that help?

> why is report-generator including the source of scripts/lighthouse-report into the report... 

report-generator.js has always included the ligthhouse-report.js source into the report. I didn't change that setup. The only difference in this PR is that we're inlining more JS into the report (logger.js + filer-namer.js + lighthouse-report.js).

> ....when that file was already part of the bundle that shipped with the viewer

Viewer bundles report-generator.js **as-is** because it needs all the mechanics that generate the HTML report. Viewer needs all of report-generator.js because it dynamically generates an HTML report  when the user drops a json file, pastes a link, or deep links from github. The idea was that we have the same codepath (`ReportGenerator.generateHTML()`) generating HTML for all three: crx/cli/viewer.

> also, can we rename some of these files, because they all sound like the same thing right now. :)

Partially why I'm advocating we don't create more micro files :) lighthouse-report.js and the viewer's subclass (lighthouse-report-viewer.js) make sense to me, but I'm open to renaming.

lighthouse-core/report/templates/report-template.html
ebidel: done

lighthouse-core/report/styles/report.css
ebidel: done

lighthouse-core/report/report-generator.js
ebidel: Yea, just stuck with what we had in `getReportCSS` and `getExceptionTemplate`

ebidel:
Ready freddie.

lighthouse-core/report/templates/report-template.html
ebidel: Thinking about this some more, I'm not stoked on moving this:

```
window.addEventListener('DOMContentLoaded', _ => {
  window.logger = new Logger('#log');
  new LighthouseReport(window.lhresults);
});
```

into a separate script. There's little value in doing that. Plus, it'll need a license header and a bunch of eslint ignore boilerplate. All of that will add to our browserified bundles.

Next step is to kill some of that extra stuff in https://github.com/GoogleChrome/lighthouse/issues/1670

ebidel:
💥

paulirish:
🎐 🚶 🥇

Issue #1463, opened 2017-01-13
Aggregation refactor 


paulirish:
I'm currently working on a change so you can easily select what sort of audits you want to run.
In reality this means need to build a full config that includes all these things:

> passes, gatherers per pass, audit list, aggregations (both parent and child aggregations), and the audit list within each of those aggregations

There's a lot of relationships here to juggle and generally it works, but one particular sore spot is "aggregations." What are aggregations anyways? :)

This all led to an idea about a possible refactor..

-------

I have a proposal for adjusting our `config/default.json`, which will have implications for the pipeline from auditResults => report.

Basically, I'm interested in replacing our [big aggregations blob](https://github.com/GoogleChrome/lighthouse/blob/ce7927307d1d4a15d81eea6b4deda57cb38d6c25/lighthouse-core/config/default.json#L98-L120) with something like the following:

* **`reportCategories` array**. These are basically our parent aggregations like "PWA", "Fancier Stuff" etc. It's very presentational, and how a visual report will be generated with some semblance of order. 
* **`auditGroups` array**. The meat of our aggregations, but it's a flat list with no nesting. so `"App can load on offline/flaky connections"` is a sibling to `"Using modern protocols"` and `"Page load performance is fast"`. 
* **`auditGroupTags` object**. Describes the few tags that are applied to each `auditGroup`. These will be used for users configuring what they want to evaluate on a given run. 

Each `auditGroup` (née aggregation) gains an `id` property, a `reportCategory` property, and a `groupTags` one. They don't have any more `items` property which has 1 or more children, which a lot of code has special handling for. yay. :)

Here's an excerpt of a revised json around the [aggregations part](https://github.com/GoogleChrome/lighthouse/blob/ce7927307d1d4a15d81eea6b4deda57cb38d6c25/lighthouse-core/config/default.json#L98-L120):

```js
  ], // end of audits array

  "reportCategories": [
    {
      "name": "Progressive Web App",
      "description": "These audits validate the aspects of a Progressive Web App.",
      "id": "pwa_category",
      "scored": true
    }, {
      "name": "Fancier stuff",
      "description": "A list of newer features that you could be using in your app. These audits do not affect your score and are just suggestions.",
      "id": "fancy_bp_category",
      "scored": false
    }, {
      "name": "Performance Metrics",
      "description": "These encapsulate your app's performance.",
      "id": "perf_diagnostics_category",
      "scored": false
    } // and "Best Practices"..
  ],

[...]

PR #1598, opened 2017-02-01
Report: lingo cleanup, visual cleanup, remove redundant messaging

ebidel:
R: all

Users get confused when an audit description says "Site does not use x API..." and there's a red "X" for the audit result. "X" means the audit failed, but the double negative is confusing. I've heard the complaint a few times in the last couple of days and decided to fix it.

This PR changes the lingo we use in the report to be consistent throughout:

- Replace "Site does not use X" -> "Avoids X"
- Replace "Site uses..." -> "Page uses..." (we're testing a page and not the entire site)
- Remove redundant debugStrings like this:

screen shot 2017-02-01 at 9 40 06 am

- Removes categories from the audit names. Reduces visual clutter.

Before:

screen shot 2017-02-01 at 11 16 27 am

Now:

screen shot 2017-02-01 at 11 08 02 am

screen shot 2017-02-01 at 11 07 53 am

### DIFF
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -33,7 +33,7 @@ class Deprecations extends Audit {
     return {
       category: 'Deprecations',
       name: 'deprecations',
-      description: 'Site does not use deprecated APIs',
+      description: 'Page avoids deprecated APIs',
       helpText: 'We found some uses of deprecated APIs. Please consider migrating ' +
           'to a newer option. [Learn more](https://www.chromestatus.com/features#deprecated).',
       requiredArtifacts: ['ChromeConsoleMessages']
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -33,7 +33,7 @@ class Deprecations extends Audit {
     return {
       category: 'Deprecations',
       name: 'deprecations',
-      description: 'Page avoids deprecated APIs',
+      description: 'Avoids deprecated APIs',
       helpText: 'We found some uses of deprecated APIs. Please consider migrating ' +
           'to a newer option. [Learn more](https://www.chromestatus.com/features#deprecated).',
       requiredArtifacts: ['ChromeConsoleMessages']
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -34,7 +34,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
     return {
       category: 'Performance',
       name: 'link-blocking-first-paint',
-      description: 'Avoids using `<link>` that delay first paint',
+      description: 'Avoids `<link>` that delay first paint',
       helpText: 'Link elements are blocking the first paint of your page. Consider ' +
           'inlining critical links and deferring non-critical ones. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -34,7 +34,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
     return {
       category: 'Performance',
       name: 'link-blocking-first-paint',
-      description: 'Site does not use <link> that delay first paint',
+      description: 'Page avoids using `<link>` that delay first paint',
       helpText: 'Link elements are blocking the first paint of your page. Consider ' +
           'inlining critical links and deferring non-critical ones. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
--- a/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
@@ -34,7 +34,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
     return {
       category: 'Performance',
       name: 'link-blocking-first-paint',
-      description: 'Page avoids using `<link>` that delay first paint',
+      description: 'Avoids using `<link>` that delay first paint',
       helpText: 'Link elements are blocking the first paint of your page. Consider ' +
           'inlining critical links and deferring non-critical ones. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
--- a/lighthouse-core/audits/dobetterweb/script-blocking-first-paint.js
+++ b/lighthouse-core/audits/dobetterweb/script-blocking-first-paint.js
@@ -33,7 +33,7 @@ class ScriptBlockingFirstPaint extends Audit {
     return {
       category: 'Performance',
       name: 'script-blocking-first-paint',
-      description: 'Avoids using `<script>` in head that delay first paint',
+      description: 'Avoids `<script>` in head that delay first paint',
       helpText: 'Script elements are blocking the first paint of your page. Consider inlining ' +
           'critical scripts and deferring non-critical ones. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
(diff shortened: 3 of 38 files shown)


lighthouse-core/report/templates/report-template.html
```diff
@@ -153,9 +153,9 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
                 <li class="subitem {{#if subItem.comingSoon}}--coming-soon{{/if}} {{#if (shouldShowHelpText subItem.score)}}--show-help{{/if}}">
 
                   <p class="subitem__desc">
-                    {{#unless ../../scored }}
+                    <!--{{#unless ../../scored }}
```
ebidel: I left this in for now in case we want to bring it back some day. Also left the category in each audit.

ebidel:
cc @kaycebasques

lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -45,10 +45,6 @@ class Manifest extends Gatherer {
           let errorString;
           if (response.url) {
             errorString = `Unable to retrieve manifest at ${response.url}`;
-          } else {
-            // The driver will return an empty string for url and the data if
```
brendankenny: maybe retain comment above `if (response.url) {`?

lighthouse-core/report/templates/report-template.html
brendankenny: leaving category in is fine, I'm sure #1463 will have something to say about it anyways, but I'd say just delete this. We definitely don't want.

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -34,7 +34,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
     return {
       category: 'Performance',
       name: 'link-blocking-first-paint',
-      description: 'Site does not use <link> that delay first paint',
+      description: 'Avoids `<link>` that delay first paint',
```
paulirish: does this description also go through `marked` now?

lighthouse-core/audits/manifest-display.js
```diff
@@ -57,7 +57,7 @@ class ManifestDisplay extends Audit {
       rawValue: hasRecommendedValue,
       displayValue
     };
-    if (!hasRecommendedValue) {
+    if (manifest && !hasRecommendedValue) {
```
ebidel: This new check will only show the "Manifest display property should be set" debug string if there is a manifest and the display property is missing.  Here's what it looks like when a manifest is missing:

screen shot 2017-02-01 at 12 17 31 pm

lighthouse-core/audits/manifest-icons-min-144.js
```diff
@@ -42,8 +42,7 @@ class ManifestIconsMin144 extends Audit {
 
     if (icons.doExist(manifest) === false) {
       return ManifestIconsMin144.generateAuditResult({
-        rawValue: false,
-        debugString: 'WARNING: No icons found in the manifest'
+        rawValue: false
```
ebidel: done

lighthouse-core/audits/manifest-icons-min-192.js
```diff
@@ -45,8 +45,7 @@ class ManifestIconsMin192 extends Audit {
 
     if (icons.doExist(manifest) === false) {
       return ManifestIconsMin192.generateAuditResult({
-        rawValue: false,
-        debugString: 'WARNING: No icons found in the manifest'
+        rawValue: false
```
ebidel: done

lighthouse-core/audits/unused-css-rules.js
```diff
@@ -32,9 +32,10 @@ class UnusedCSSRules extends Audit {
     return {
       category: 'CSS',
       name: 'unused-css-rules',
-      description: 'Site does not have more than 10% unused CSS',
+      description: 'Page uses 90% of its CSS rules',
       helpText: 'Remove unused rules from stylesheets to reduce unnecessary ' +
-          'bytes consumed by network activity. [Learn more](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery)',
+          'bytes consumed by network activity. ' +
```
ebidel: it was over 100 chars

lighthouse-core/audits/viewport.js
```diff
@@ -27,7 +27,7 @@ class Viewport extends Audit {
     return {
       category: 'Mobile Friendly',
       name: 'viewport',
-      description: 'HTML has a `<meta name="viewport">` tag containing `width` or `initial-scale`',
+      description: 'Page contains a `<meta name="viewport">` tag with `width` or `initial-scale`',
```
ebidel: done

lighthouse-core/gather/gatherers/manifest.js
ebidel: done

kaycebasques:
Well, while we're on the topic, since you removed "site", why not also remove "page"? LH always and only audits the current page, no? So using "page" implies some distinction that does not exist.

lighthouse-core/audits/deprecations.js
```diff
@@ -33,7 +33,7 @@ class Deprecations extends Audit {
     return {
       category: 'Deprecations',
       name: 'deprecations',
-      description: 'Site does not use deprecated APIs',
+      description: 'Avoids deprecated APIs',
```
kaycebasques: I'm all for nixing `Site`, but `Avoids` strikes me as vaguer than `Does not use`. I suggest s/Avoid/Does not use/ throughout.

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
ebidel: you got it.

kaycebasques:
Dropping `page` also cleans up a lot of unnecessary text, so that should be right up errone's alleys

lighthouse-core/audits/deprecations.js
ebidel: The complaint we've been getting (and which this PR tries to address) is that "not" in  `"X" Does not use....` is confusing. The "X" icon + "not" appear to folks as a double negative.

I also think there could be something better than "avoids" . Any suggestions?

ebidel:
K. Dropped "page" from the audit's that use positive assertions in their descriptions

lighthouse-core/audits/deprecations.js
kaycebasques: Ah, OK. Hm... Well, we could drop the verb... E.g. `Deprecated APIs`. Else it might require inversing the titles... E.g. `Uses Up-To-Date APIs`

lighthouse-core/audits/manifest-display.js
brendankenny: oh, my brain is broken. Somehow I thought the diff was the reverse.

lighthouse-core/audits/deprecations.js
kaycebasques: Yeah... given the constraints of the current UI, I don't see a better alternative to `avoid`! Although something doesn't sit well with me about it.

*Rides off grumbling into the sunset*

lighthouse-core/audits/dobetterweb/script-blocking-first-paint.js
```diff
@@ -33,7 +33,7 @@ class ScriptBlockingFirstPaint extends Audit {
     return {
       category: 'Performance',
       name: 'script-blocking-first-paint',
-      description: 'Site does not use <script> in head that delays first paint',
+      description: 'Avoids `<script>` in head that delay first paint',
```
patrickhulce: Why delay? `<script>` is singular or referential to the idea of script, no?

"Avoids script in head that delays first paint" or
"Avoids scripts in head that delay first paint"

brendankenny:
maybe clean up the [accessibility audits `debugString` generator](https://github.com/GoogleChrome/lighthouse/blob/7117029a811efbb9c12971fbed9681bb77615a14/lighthouse-core/audits/accessibility/axe-audit.js#L58)? Could just get rid of the `rule.help` and the parentheses. Example:

screen shot 2017-02-01 at 16 04 04

kaycebasques:
The `learn more` after `1 element fails this test` can also be removed now that the a11y audits have links to docs.

brendankenny:
> Could just get rid of the rule.help and the parentheses 

actually maybe the whole `debugString` since the rule help is contained in the audit description and the "Failed on 1 element` is contained in the extended info header below :)

lighthouse-core/audits/dobetterweb/script-blocking-first-paint.js
ebidel: I read "Avoids `<script> in head...` as "Avoids scripts in head....", but I'll change it back to "Avoids `<script>` in head that delays first paint"

ebidel:
I'll do the a11y cleanup in another pr

PR #1601, opened 2017-02-02
fix: save artifacts and assets in output path

patrickhulce:
More changes needed I came across during WPT integration investigation.

* Default report when output is `pretty` will use same name and location as output path.
* Artifacts will use same name and location as output path or will use `getFilenamePrefix` and be suffixed with `.artifacts.log`
* Assets will use same name and location as output path.

**Examples**

`--output-path=~/mydir/foo.out --save-assets` generates
* `~/mydir/foo.out`
* `~/mydir/foo.report.html`
* `~/mydir/foo-0.trace.json`
* `~/mydir/foo-0.screenshots.html`

`--output-path=./report.json --output json --save-artifacts` generates 
* `./report.json`
* `./report.artifacts.log`

`--save-artifacts` generates
* `./<HOST>_<DATE>.report.html`
* `./<HOST>_<DATE>.artifacts.log`


lighthouse-core/lib/asset-saver.js
```diff
@@ -131,19 +131,20 @@ function prepareAssets(artifacts, results) {
...
-function saveAssets(artifacts, results) {
-  return prepareAssets(artifacts, results).then(assets => {
+function saveAssets(artifacts, options = {}) {
+  return prepareAssets(artifacts, options.title).then(assets => {
     assets.forEach((data, index) => {
-      const filenamePrefix = getFilenamePrefix(results);
       const traceData = data.traceData;
-      fs.writeFileSync(`${filenamePrefix}-${index}.trace.json`, JSON.stringify(traceData, null, 2));
-      log.log('trace file saved to disk', filenamePrefix);
+      const traceFilename = `${options.path}-${index}.trace.json`;
```
patrickhulce: I'm not super thrilled about calling this mixture of path and prefix `path` but passing both options everywhere seemed like overkill. My arm could be twisted to revert though... :)

lighthouse-cli/bin.ts
```diff
@@ -260,25 +260,31 @@ function runLighthouse(url: string,
     .then(chrome => chromeLauncher = chrome)
     .then(() => lighthouse(url, flags, config))
     .then((results: Results) => {
+      let promise = Promise.resolve(results);
```
brendankenny: maybe move all this printing stuff to its own function? Definitely getting hard to follow :)

lighthouse-cli/bin.ts
```diff
@@ -260,25 +260,31 @@ function runLighthouse(url: string,
...
+          flags.outputPath.replace(/\.\w{2,4}$/, '');
+      const resolvedPath = path.resolve(process.cwd(), baseFilename);
+
       // delete artifacts from result so reports won't include artifacts.
       const artifacts = results.artifacts;
       results.artifacts = undefined;
 
       if (flags.saveArtifacts) {
-        assetSaver.saveArtifacts(artifacts);
+        promise = promise.then(_ => assetSaver.saveArtifacts(artifacts, resolvedPath));
```
brendankenny: `saveArtifacts` is sync, so does wrapping it matter?

lighthouse-core/lib/asset-saver.js
```diff
@@ -131,19 +131,20 @@ function prepareAssets(artifacts, results) {
 /**
  * Writes trace(s) and associated screenshot(s) to disk.
  * @param {!Artifacts} artifacts
- * @param {!Results} results
+ * @param {{path: string, title: string}} options
  * @return {!Promise}
  */
-function saveAssets(artifacts, results) {
-  return prepareAssets(artifacts, results).then(assets => {
+function saveAssets(artifacts, options = {}) {
```
brendankenny: this doesn't appear to be treated as an optional parameter, so any need for the default?

lighthouse-core/lib/asset-saver.js
brendankenny: isn't it just passed to this one function? agreed that calling it `path` is kind of confusing

lighthouse-cli/bin.ts
```diff
@@ -260,25 +260,31 @@ function runLighthouse(url: string,
...
       }
       if (flags.saveAssets) {
-        return assetSaver.saveAssets(artifacts, results).then(() => results);
+        promise = promise.then(_ => assetSaver.saveAssets(artifacts, {
+          title: path.basename(resolvedPath),
```
brendankenny: not really clear what `title` is here, or why it's called that. file prefix?

lighthouse-core/lib/asset-saver.js
patrickhulce: it's this and saveArtifacts and used back in `bin.ts`

lighthouse-cli/bin.ts
patrickhulce: I suppose not, it just seemed like the odd duckling before

lighthouse-cli/bin.ts
patrickhulce: it becomes the title of the screenshots HTML page, which was essentially this basename before but it needed all the results passed in to compute it

lighthouse-core/lib/asset-saver.js
patrickhulce: yeah ended up not being optional anymore, I'll remove

lighthouse-cli/bin.ts
patrickhulce: will do

patrickhulce:
alright @brendankenny moved some things around and just got rid of title completely for now, hopefully it's clearer now

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,36 @@ function handleError(err: LighthouseError) {
   }
 }
 
+function saveResults(results: Results,
+                     flags: {output: any, outputPath: string, saveArtifacts: boolean, saveAssets: boolean}) {
+    let promise = Promise.resolve(results);
+    const cwd = process.cwd();
+    const configuredPath = !flags.outputPath || flags.outputPath === 'stdout' ?
+        assetSaver.getFilenamePrefix(results) :
+        flags.outputPath.replace(/\.\w{2,4}$/, '');
```
paulirish: instead of this edit we could do...

```js
const parsed = path.parse(flags.outputPath); 
parsed.base = parsed.base.replace(parsed.ext, '');
path.format(parsed);
```

but i don't know if that's really worth it. :)

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,36 @@ function handleError(err: LighthouseError) {
...
+    const resolvedPath = path.resolve(cwd, configuredPath);
+    const pathWithBasename = resolvedPath.includes(cwd) ?
+        resolvedPath.slice(cwd.length + 1) : resolvedPath;
+
+    // delete artifacts from result so reports won't include artifacts.
+    const artifacts = results.artifacts;
+    results.artifacts = undefined;
+
+    if (flags.saveArtifacts) {
+      assetSaver.saveArtifacts(artifacts, pathWithBasename);
```
paulirish: as mentioned offline, we can remove --save-artifacts flag and the functions if we want. AFAIK, they havent been used for ages and are only for debugging.

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,36 @@ function handleError(err: LighthouseError) {
...
+function saveResults(results: Results,
+                     flags: {output: any, outputPath: string, saveArtifacts: boolean, saveAssets: boolean}) {
+    let promise = Promise.resolve(results);
+    const cwd = process.cwd();
+    const configuredPath = !flags.outputPath || flags.outputPath === 'stdout' ?
+        assetSaver.getFilenamePrefix(results) :
+        flags.outputPath.replace(/\.\w{2,4}$/, '');
+    const resolvedPath = path.resolve(cwd, configuredPath);
+    const pathWithBasename = resolvedPath.includes(cwd) ?
+        resolvedPath.slice(cwd.length + 1) : resolvedPath;
```
paulirish: my read is that resolvedPath always includes cwd because we just did path.resolve on it.
and so we'll always slice,
taking the absolute file path back to a relative one.

This is 1) the same as `path.relative(cwd, resolvedPath)` AFAICT. but 2) i don't know why we nuke it, since its a more absolute location.

patrickhulce: Just nuked because with the default options the logging seemed a bit verbose to restate your `cwd`, don't have particularly strong feelings about it though if we're fine with the absolute path being printed.

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,36 @@ function handleError(err: LighthouseError) {
...
+
+    // delete artifacts from result so reports won't include artifacts.
+    const artifacts = results.artifacts;
```
patrickhulce: I think it feels weird have this part of the saving logic outside `saveResults` haha, compromise? `results = Object.assign({}, results, {artifacts: undefined});`

lighthouse-core/lib/asset-saver.js
```diff
@@ -90,24 +89,23 @@ img {
 /**
  * Save entire artifacts object to a single stringified file
```
patrickhulce: done

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,36 @@ function handleError(err: LighthouseError) {
   }
 }
 
+function saveResults(results: Results,
+                     flags: {output: any, outputPath: string, saveArtifacts: boolean, saveAssets: boolean}) {
+    let promise = Promise.resolve(results);
+    const cwd = process.cwd();
+    const configuredPath = !flags.outputPath || flags.outputPath === 'stdout' ?
```

lighthouse-cli/bin.ts
patrickhulce: I like the explicitness of it but not a huge fan of the triple lines, haha :)

readme.md
```diff
@@ -164,6 +164,23 @@ Options:
                      installations are found                                               [boolean]
 ```
 
+### Output Path Examples
+
+`--output-path=~/mydir/foo.out --save-assets` generates
```
brendankenny: I don't believe this actually works, though, because we don't do `~` expansion

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,39 @@ function handleError(err: LighthouseError) {
...
+        assetSaver.getFilenamePrefix(results) :
+        flags.outputPath.replace(/\.\w{2,4}$/, '');
+    const resolvedPath = path.resolve(cwd, configuredPath);
+    const pathWithBasename = resolvedPath.includes(cwd) ?
+        path.relative(cwd, resolvedPath) : resolvedPath;
+
+    const audits = results.audits;
+    const artifacts = results.artifacts;
+    // remove artifacts from result so reports won't include artifacts.
+    results = Object.assign({}, results, {artifacts: undefined});
```
brendankenny: this is fine (it makes some sense, we are dropping `artifacts` since we don't want the them in the saved/printed results), but since this resolves on the changed `results` I don't see the difference functionally?

So it still seems weird, but now it's that perfX relies on that side effect of `saveResults` to not get 100MB reports :) Not sure of the optimal configuration here, though

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,39 @@ function handleError(err: LighthouseError) {
...
+    let promise = Promise.resolve(results);
+    const cwd = process.cwd();
+    // Use the output path as the prefix for all generated files.
+    // If no output path is set, generate a file prefix using the URL and date.
+    const configuredPath = !flags.outputPath || flags.outputPath === 'stdout' ?
+        assetSaver.getFilenamePrefix(results) :
+        flags.outputPath.replace(/\.\w{2,4}$/, '');
+    const resolvedPath = path.resolve(cwd, configuredPath);
+    const pathWithBasename = resolvedPath.includes(cwd) ?
+        path.relative(cwd, resolvedPath) : resolvedPath;
```
brendankenny: I'm personally +1 for dropping this line since it definitely adds mental overhead when reading. If it's annoying in the log we could fix it where it's logged instead

lighthouse-cli/bin.ts
```diff
@@ -249,6 +249,39 @@ function handleError(err: LighthouseError) {
...
+
+    if (flags.saveAssets) {
+      promise = promise.then(_ => assetSaver.saveAssets(artifacts, audits, pathWithBasename));
+    }
+
+    if (flags.output === Printer.OutputMode[Printer.OutputMode.pretty]) {
+      promise = promise.then(_ => Printer.write(results, 'html', `${pathWithBasename}.report.html`));
+    }
+
+    return promise.then(_ => Printer.write(results, flags.output, flags.outputPath));
```
brendankenny: you're inheriting this, but it seems kind of magical that we rely on `Printer.write` to return `results`. How do you feel about doing a resolve on `results` here?

readme.md
patrickhulce: depends on your shell I suppose, should I change to an absolute path?

lighthouse-cli/bin.ts
patrickhulce: will do

lighthouse-cli/bin.ts
patrickhulce: only changes in the sense that the original results object is preserved, if we never intend artifacts on being consumed along with the rest of the results I'd advocate splitting them into distinct sets in the lighthouse return object. Since that'd probably break others at this point, I'll restructure a bit more instead.

lighthouse-cli/bin.ts
patrickhulce: removed the need for it to return anything at all

PR #1604, opened 2017-02-02
CRX: make disabling extension during run, optional

ebidel:
R: @brendankenny, all

This makes the enabling/disabling of extension before a LH run a user setting:

screen shot 2017-02-02 at 11 24 53 am

My worry was that some users would find the current default behavior (disable all extension before the run) unexpected and frustrating. Any extensions that don't save state when they're killed off will lose data. I suspect there are many that don't follow best practices. Instead, we're making this a setting for people to opt into.

@brendankenny I tried to move some of the bg methods into popup.html like we chatted about but it was getting trickier than it was worth. We allow the user to dismiss the popup while LH is running. It was getting gnarly to make a transient popup communicate state back to the bg page for persistence, then repopulate the UI when the user reopens the popup. I was also finding myself piping more `background` page variables through popup.js methods...which defeated the original purpose of the code refactor (separation of concerns between bg page and popup.js).

### DIFF
--- a/lighthouse-extension/app/src/lighthouse-background.js
+++ b/lighthouse-extension/app/src/lighthouse-background.js
@@ -26,8 +26,10 @@ const log = require('../../../lighthouse-core/lib/log');
 const ReportGenerator = require('../../../lighthouse-core/report/report-generator');
 
 const STORAGE_KEY = 'lighthouse_audits';
+const SETTINGS_KEY = 'lighthouse_settings';
 
 let installedExtensions = [];
+let DISABLE_EXTENSIONS_DURING_RUN = false;
 let lighthouseIsRunning = false;
 let latestStatusLog = [];
 
@@ -41,6 +43,10 @@ const _flatten = arr => [].concat(...arr);
  * @param {!Promise}
  */
 function enableOtherChromeExtensions(enable) {
+  if (!DISABLE_EXTENSIONS_DURING_RUN) {
+    return Promise.resolve();
+  }
+
   const str = enable ? 'enabling' : 'disabling';
   log.log('Chrome', `${str} ${installedExtensions.length} extensions.`);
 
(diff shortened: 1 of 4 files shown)


ebidel:
Tests are shoppy shop. Shop is the worst.

lighthouse-extension/app/src/popup.js
```diff
@@ -207,9 +209,11 @@ function initPopup() {
     const okButton = document.getElementById('ok');
     okButton.addEventListener('click', () => {
       // Save selected aggregation categories on options page close.
```
brendankenny: "Save settings on options page close"?

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -26,8 +26,10 @@ const log = require('../../../lighthouse-core/lib/log');
 const ReportGenerator = require('../../../lighthouse-core/report/report-generator');
 
 const STORAGE_KEY = 'lighthouse_audits';
+const SETTINGS_KEY = 'lighthouse_settings';
 
 let installedExtensions = [];
+let DISABLE_EXTENSIONS_DURING_RUN = false;
```
brendankenny: not a constant so shouldn't be all caps

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -251,41 +257,57 @@ window.getDefaultAggregations = function() {
 
 /**
  * Save currently selected set of aggregation categories to local storage.
- * @param {!Array<{name: string, audits: !Array<string>}>} selectedAggregations
+ * @param {!{selectedAggregations: Array<{name: string, audits: !Array<string>}>,
+ *           disableExtensions: boolean}} settings
+ * @param {!boolean} disableExtensions True if extensions should be disabled for the run.
```
brendankenny: not separate params, just one object

brendankenny:
> I tried to move some of the bg methods into popup.html like we chatted about but it was getting trickier than it was worth.

maybe (in a future PR) we can move it into a separate background script, then, so other clients won't get the extension-specific stuff

lighthouse-extension/app/src/lighthouse-background.js
ebidel: leftover from tinkering, done.

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

ebidel:
> maybe (in a future PR) we can move it into a separate background script, then, so other clients won't get the extension-specific stuff

Or we flip it and have something special for devtools. It's funky the devtools using the built background file from the extension. Those should really be separate targets.

ebidel:
PTAL

ebidel:
Tests are green

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -251,41 +257,56 @@ window.getDefaultAggregations = function() {
 
 /**
  * Save currently selected set of aggregation categories to local storage.
- * @param {!Array<{name: string, audits: !Array<string>}>} selectedAggregations
+ * @param {!{selectedAggregations: Array<{name: string, audits: !Array<string>}>,
```
brendankenny: somewhere along the line this became wrong. I believe its `@param {{selectedAggregations: !Array<string>, disableExtensions: boolean}} settings`

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -251,41 +257,56 @@ window.getDefaultAggregations = function() {
...
+  storage[SETTINGS_KEY].disableExtensions = disableExtensionsDuringRun;
+
+  // Save object to chrome local storage.
   chrome.storage.local.set(storage);
 };
 
 /**
  * Load selected aggregation categories from local storage.
- * @return {!Promise<!Object<boolean>>}
+ * @return {!Promise<{selectedAggregations: Array<{name: string, audits: !Array<string>}>,
```
brendankenny: this one was right, so `@return {!Promise<{selectedAggregations: !Object<boolean>, disableExtensions: boolean}>}`

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

lighthouse-extension/app/src/lighthouse-background.js
ebidel: done

ebidel:
Woot!

PR #1605, opened 2017-02-02
fix: warn about redirected urls

### LINKED ISSUE #715: "URL responds with a 200 when offline": initial URL or final URL?
This is an interesting case where a developer using Lighthouse might be looking for a strict or loose interpretation of their question:
- if you've visited a URL before, does loading that exact URL when offline load content? Or:
- if you've visited a URL before, does reloading that page (or running from the homescreen) while offline load content?

The difference is in the second case the URL being visited while offline may not be the initial URL entered if a redirect occurred. Two cases I have in smokehouse are
- `https://www.airhorner.com`, which redirects to `https://airhorner.com`, which _does_ work offline. We report it as not working offline because we reload the original URL and the redirect doesn't happen.
- `https://www.chromestatus.com/` which redirects to `https://www.chromestatus.com/features`, which _does_ work offline, but we again report the initial URL as not working offline.

When interpreting the question literally (does `https://www.airhorner.com` work while offline?) the answer is correct right now, but, again, this may be a surprising result if what you're really asking is if the page was open and you go offline, or it's refreshed while offline, does the page still work.

Two possible solutions:
- multiple offline tests
- warn when the initial URL changes due to redirects. "You tested https://example.com which redirected you to https://example.com/awesomePWA. You may want to test the second URL directly for complete results" or whatever

patrickhulce:
fixes #715 

How it looks. 
image

I stuck it after everything since putting at the top likely means no one will know.

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -281,6 +281,14 @@ function runLighthouse(url: string,
       return results;
     })
     .then((results: Results) => {
+      if (url !== results.url) {
+        const msg = `Redirect to ${results.url} detected. For best results, test this URL directly.`;
+        log.warn('Lighthouse CLI', msg);
+      }
+
+      return results;
+    })
+    .then((results: Results) => {
       if (flags.interactive) {
         return performanceXServer.hostExperiment({url, flags, config}, results);
       }


lighthouse-cli/bin.ts
```diff
@@ -281,6 +281,14 @@ function runLighthouse(url: string,
       return results;
     })
     .then((results: Results) => {
+      if (url !== results.url) {
```
paulirish: should we move this patch to runner.js so our other distributions get the log as well?

patrickhulce: How are logs surfaced there? I really think it's unlikely to have any impact unless it's prominent. Come to think of it, thoughts on injecting the warning into the report itself?

brendankenny: > Come to think of it, thoughts on injecting the warning into the report itself?

#1512 kind of captures the need for this. For stuff that's logged anyway, one easy fix is to capture any warning or error log messages and somehow display them in the report.

patrickhulce:
Just bumping this, @brendankenny do you think we should hold off until we fully resolve #1512?

lighthouse-cli/bin.ts
brendankenny: after #1677 I now realize you should do the comparison with `URL.equalWithExcludedFragments` so it doesn't warn on dropped fragments

patrickhulce: good call

brendankenny:
> Just bumping this, @brendankenny do you think we should hold off until we fully resolve #1512?

I feel like this is less effective while we figure out #1512 but still helpful, but agree with @paulirish that this makes a lot more sense inside lighthouse instead of in the CLI.

lighthouse-cli/bin.ts
brendankenny: actually, now I wonder if that's right :) Excluding the fragment makes sense for comparing against network records, but for sites doing some state based stuff with the fragment maybe we should warn that you ended up in a different state than you intended (at least according to the URL)

wdyt?

patrickhulce:
I'm going to close this in favor of starting work towards consistent error/warning reporting and surfacing.

PR #1620, opened 2017-02-02
fix: critical request chains error on empty

### LINKED ISSUE #1427: Extension Error: Cannot read property 'children' of undefined
**Lighthouse Version**: 1.4.1
**Chrome Version**: 55.0.2883.87
**URL**: http://10.1.10.7:8011/h_home.apw
**Error Message**: Cannot read property 'children' of undefined
**Stack Trace**:
 ```TypeError: Cannot read property 'children' of undefined
    at artifacts.requestCriticalRequestChains.then.chains (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:560:62)```

patrickhulce:
fixes #1427 

This seems to be another one of those cases that seems impossible to happen, but happens. I couldn't repro the error with the URL in the issue.

### DIFF
--- a/lighthouse-core/audits/critical-request-chains.js
+++ b/lighthouse-core/audits/critical-request-chains.js
@@ -64,8 +64,8 @@ class CriticalRequestChains extends Audit {
       }
 
       // Account for initial navigation
-      const initialNavigationKey = Object.keys(chains)[0];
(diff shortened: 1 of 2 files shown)


ebidel:
Thanks for taking care of this one. Glad to see it finally fixed.

PR #1623, opened 2017-02-03
Update more gatherers to use gather error instead of -1 artifacts

brendankenny:
followup to #1560, continuing work on #941. This removes catches, etc from gatherers, letting any exceptions handle themselves as gatherer errors. Separated by commit for easier review, but most of these are fairly straightforward.

This leaves only the Manifest gatherer and the deobetterweb gatherers to go; thought I'd break here before it got too unwieldy.


lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -32,7 +32,7 @@ class AxeAudit extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    const violations = artifacts.Accessibility.violations || [];
+    const violations = artifacts.Accessibility.violations;
```
brendankenny: now asserted as an array in the gatherer since [axe says](https://github.com/dequelabs/axe-core/blob/master/doc/API.md#results-object) any valid result will include the `violations` array.

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -63,20 +59,14 @@ class Accessibility extends Gatherer {
     })()`;
 
     return driver
-        .evaluateAsync(expression)
-        .then(returnedValue => {
-          if (!returnedValue) {
-            return Accessibility._errorAccessibility('Unable to parse axe results');
-          }
-
-          if (returnedValue.error) {
```
brendankenny: This appears to be from working around issues in the old `evaluateAsync` before @patrickhulce fixed up the error handling to properly reject. So, removed.

lighthouse-core/gather/gatherers/html.js
```diff
@@ -1,42 +0,0 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-'use strict';
-
-const Gatherer = require('./gatherer');
-
-class HTML extends Gatherer {
```
brendankenny: super old and unused

lighthouse-core/test/audits/service-worker-test.js
```diff
@@ -21,19 +21,6 @@ const assert = require('assert');
 /* eslint-env mocha */
 
 describe('Offline: Service Worker audit', () => {
-  it('reports driver error when given no Service Worker versions', () => {
```
ebidel: so all these tests are going through a single codepath now? https://github.com/GoogleChrome/lighthouse/pull/1591/files#diff-9476be2df7d812bc50da0dd73d73f2f8R205?

brendankenny: More or less, if I understand what you're asking. Audits never receive an error artifact; [runner produces an error audit result](https://github.com/GoogleChrome/lighthouse/blob/81b891d491e95135f1f5daf4278e1e27c93ac510/lighthouse-core/runner.js#L182) for them if any required artifacts are in error. That's tested by the test you linked.

ebidel: Yea ok. Just checking we're not losing test coverage.

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -63,20 +59,14 @@ class Accessibility extends Gatherer {
...
-            return Accessibility._errorAccessibility(returnedValue.error);
-          }
+      .evaluateAsync(expression)
+      .then(returnedValue => {
+        if (!returnedValue || !Array.isArray(returnedValue.violations)) {
```
paulirish: since we're nuking all the remaining checks on the violations array, lets just add an assertion in `accessibility-test.js` that we throw in this case.

brendankenny: good call

PR #823, opened 2016-10-24
stop interpreting driver manifest errors as failure to fetch manifest

brendankenny:
fixes #819 

currently any entries in `response.errors` are treated as a failure to fetch the manifest, even though they entires in `response.errors` are parsing errors. There are some caveats here: for instance, the debugger protocol will try to `JSON.parse()` a 404 message from a missing manifest and so will report a syntax error instead of the actual 404 error.

This PR works around this by ignoring the `errors` array and looking directly at the raw `data` and `url` returned.

This isn't perfect, however. Malformed JSON, for instance, will show up as `'Unable to retrieve manifest at ${response.url}`, but there's not really any signal we can use to get around that. The debugger protocol simply won't return anything on `response.data` in that case, and so is indistinguishable from e.g. a 404 page.

PR #1624, opened 2017-02-03
Update Manifest gatherer to use gather error instead of -1 artifact

brendankenny:
Part of #941. There a lot of audits that depend on the manifest gatherer, so separated this one from #1623.

The most impactful change I made was to make the manifest gatherer return `null` if the page has no manifest as it seems appropriate to have an explicit value for this (and obviously not having a manifest is not an error).

This sticks to the minimal redundancy of `debugString`s that #1598 started: if the page has no manifest, all the manifest audits will fail with no message. If the manifest was there but invalid JSON, they will all fail but only `manifest-exists` will report the `debugString`.

There's a good bit of redundancy here. It may make sense to split these into their own `audits/` subdirectory a la `accessibility/` and do some kind of base class thing (and unify testing...most of these tests are duplicates with only the tested property changed)

### DIFF
--- a/lighthouse-core/audits/cache-start-url.js
+++ b/lighthouse-core/audits/cache-start-url.js
@@ -37,21 +37,18 @@ class CacheStartUrl extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    let cacheHasStartUrl = false;
-    const manifest = artifacts.Manifest.value;
-    const cacheContents = artifacts.CacheContents;
-
-    if (!(manifest && manifest.start_url && manifest.start_url.value)) {
+    if (!artifacts.Manifest || !artifacts.Manifest.value) {
+      // Page has no manifest or was invalid JSON.
       return CacheStartUrl.generateAuditResult({
         rawValue: false,
-        debugString: 'start_url not present in Manifest'
       });
     }
 
-    if (!Array.isArray(cacheContents)) {
+    const manifest = artifacts.Manifest.value;
+    if (!(manifest.start_url && manifest.start_url.value)) {
       return CacheStartUrl.generateAuditResult({
         rawValue: false,
-        debugString: cacheContents.debugString || 'No cache detected'
+        debugString: 'start_url not present in Manifest'
       });
     }
 
@@ -67,7 +64,8 @@ class CacheStartUrl extends Audit {
     // here would be to resolve this more completely by asking the Service Worker about the start
     // URL. However that would also necessitate the cache contents gatherer relying on the manifest
     // gather rather than being independent of it.
-    cacheHasStartUrl = cacheContents.find(req => {
+    const cacheContents = artifacts.CacheContents;
+    const cacheHasStartUrl = cacheContents.find(req => {
       return (startURL === req || altStartURL === req);
     });
 
--- a/lighthouse-core/audits/manifest-exists.js
+++ b/lighthouse-core/audits/manifest-exists.js
@@ -40,6 +40,13 @@ class ManifestExists extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    if (!artifacts.Manifest) {
+      // Page has no manifest.
+      return ManifestExists.generateAuditResult({
+        rawValue: false
+      });
+    }
+
     return ManifestExists.generateAuditResult({
       rawValue: typeof artifacts.Manifest.value !== 'undefined',
       debugString: artifacts.Manifest.debugString
--- a/lighthouse-core/gather/gatherers/manifest.js
+++ b/lighthouse-core/gather/gatherers/manifest.js
@@ -26,15 +26,11 @@ const manifestParser = require('../../lib/manifest-parser');
  * the manifest parser.
  */
 class Manifest extends Gatherer {
-
-  static _errorManifest(errorString) {
-    return {
-      raw: undefined,
-      value: undefined,
-      debugString: errorString
-    };
-  }
-
+  /**
+   * Returns the parsed manifest or null, if the page had no manifest.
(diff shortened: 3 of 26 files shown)


lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -26,35 +26,28 @@ const manifestParser = require('../../lib/manifest-parser');
...
-          let errorString;
-          // The driver returns an empty string for url and the data if
-          // the page has no manifest.
           if (response.url) {
-            errorString = `Unable to retrieve manifest at ${response.url}`;
+            throw new Error(`Unable to retrieve manifest at ${response.url}`);
           }
 
-          return Manifest._errorManifest(errorString);
+          // The driver will return an empty string for url and the data if
```
paulirish: i'd keep this comment above the `if (response.url)` conditional

lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -26,35 +26,28 @@ const manifestParser = require('../../lib/manifest-parser');
...
-
+  /**
+   * Returns the parsed manifest or null, if the page had no manifest.
+   * @param {!Object} options
+   * @return {!Promise<?Manifest>}
+   */
   afterPass(options) {
     return options.driver.sendCommand('Page.getAppManifest')
       .then(response => {
         // We're not reading `response.errors` however it may contain critical and noncritical
```
paulirish: wdyt about handling these? it might help us differentiate between bad JSON, missing, or other issues.

lighthouse-core/audits/cache-start-url.js
```diff
@@ -37,21 +37,18 @@ class CacheStartUrl extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    let cacheHasStartUrl = false;
-    const manifest = artifacts.Manifest.value;
-    const cacheContents = artifacts.CacheContents;
-
-    if (!(manifest && manifest.start_url && manifest.start_url.value)) {
+    if (!artifacts.Manifest || !artifacts.Manifest.value) {
```
paulirish: Since `artifacts.Manifest` is now `null` in the non-exist case, should we just change all these to this?

```js
if (artifacts.Manifest === null) {
```

lighthouse-core/audits/manifest-exists.js
```diff
@@ -40,6 +40,13 @@ class ManifestExists extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    if (!artifacts.Manifest) {
```
paulirish: compare against null?

lighthouse-core/gather/gatherers/manifest.js
brendankenny: > wdyt about handling these?

do we have a bug tracking it? I assume you added that comment :)

> it might help us differentiate between bad JSON, missing, or other issues.

the first two should be handled, but yeah, other issues would be good to catch

lighthouse-core/gather/gatherers/manifest.js
brendankenny: hopefully clarified the comment

lighthouse-core/audits/cache-start-url.js
brendankenny: this is checking two error states.
1) no manifest, so the `artifacts.Manifest` is `null`. This could do the explicit check, but not sure if it's worth it as the gatherer will only return `null` or an object, so only null will be falsy.

2) invalid JSON, so `artifacts.Manifest.value` will be `undefined`. Invalid JSON is the only way that `artifacts.Manifest.value`, but we have to check for it.

   One solution I considered to get rid of the extra check was to throw if the JSON was invalid, but it felt a bit weird to declare a gatherer error when it really was an error in the user's data. It would also mean that every manifest audit would have a `debugString` of `Audit error: Manifest gatherer encountered an error: invalid JSON` (or whatever) . That's a lot of visual noise when it could just be on the `manifest-exists` audit as it is here. OTOH, invalid JSON in your manifest is a pretty big and obvious error and one the site owner should fix quickly, so maybe that's fine.

lighthouse-core/audits/manifest-exists.js
brendankenny: > compare against null?

as above, we can compare against null if you feel strongly about it, but null is the only falsy value you could get here (and if we added the check against null specifically and let through the other falsy values, `artifacts.Manifest.value` below would throw anyways)

lighthouse-core/audits/cache-start-url.js
paulirish: well i guess i'm saying we should handle the invalid JSON case via `response.errors` in the gatherer, so we can be more clear here.  otherwise these are the only audits that have to differentiate between two error states (plus they can't say which they're in)

brendankenny: Hmm, maybe I don't understand what you're saying.

Just to be clear, by invalid JSON I mean just unparseable JSON; if the manifest is parseable by `JSON.parse` but is invalid according to the manifest spec, that *will* show up in the individual audits.

This code *can* separate these two states, but I put them in one conditional on purpose because we want the audits to just fail and not output a `debugString` (otherwise we'd have 12 `debugString`s or whatever in the report saying the same thing).

lighthouse-core/gather/gatherers/manifest.js
paulirish: > do we have a bug tracking it? I assume you added that comment :)

https://github.com/GoogleChrome/lighthouse/pull/823

OHHHHHHHHHHHHHHHHHH WHAT NOW

PR #1626, opened 2017-02-03
Fix viewer's "Export -> Save as..." with Firefox.

### LINKED ISSUE #1574: Viewer: "Save as" doesn't work with Firefox
Using Firefox 52b1, Export -> Save as html or json does nothing and no error is logged in console.

Using Chrome 56, it works fine.

XhmikosR:
For some weird reason, Firefox needs `document.body.appendChild(a)` too.

Fixes #1574.

### DIFF
--- a/lighthouse-viewer/app/src/lighthouse-report-viewer.js
+++ b/lighthouse-viewer/app/src/lighthouse-report-viewer.js
@@ -361,9 +361,12 @@ class LighthouseViewerReport extends LighthouseReport {
     const a = document.createElement('a');
     a.download = `${filename}${ext}`;
     a.href = URL.createObjectURL(blob);
+    document.body.appendChild(a); // Firefox requires anchor to be in the DOM.
     a.click();
 
-    setTimeout(_ => URL.revokeObjectURL(a.href), 500); // cleanup.
+    // cleanup.
+    document.body.removeChild(a);
+    setTimeout(_ => URL.revokeObjectURL(a.href), 500);
   }
 
   /**


XhmikosR:
Only thing I wonder is if we should call `removeChild` too or not. Any feedback is appreciated.

XhmikosR:
I added a new pat6ch which also removes the node from DOM and changes the `setTimeout`'s delay to 0 which works the same for FF and Chrome.

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -361,9 +361,12 @@ class LighthouseViewerReport extends LighthouseReport {
     const a = document.createElement('a');
     a.download = `${filename}${ext}`;
     a.href = URL.createObjectURL(blob);
+    document.body.appendChild(a);
```
ebidel: `document.body.appendChild(a); // Firefox requires anchor be in the DOM.`

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -361,9 +361,12 @@ class LighthouseViewerReport extends LighthouseReport {
...
     a.click();
 
-    setTimeout(_ => URL.revokeObjectURL(a.href), 500); // cleanup.
+    // cleanup.
+    document.body.removeChild(a);
+    setTimeout(_ => URL.revokeObjectURL(a.href), 0);
```
ebidel: We may want to keep the 500ms delay just to be on the safe side.

XhmikosR:
@ebidel: done. I kept the interval to 500, even though it seems unneeded from my tests.

PR #1627, opened 2017-02-03
Minor CSS consistency changes:

XhmikosR:
* use lowercase hex colors
* use numeric font-weight values
* remove units from zero


ebidel:
Thanks for the change. IMO, it's generally hard to be consistent (and enforce) formatting without a proper CSS linter. But LGTM.

XhmikosR:
Agreed. Might be worth looking into [stylelint](https://github.com/stylelint/stylelint).

PR #1633, opened 2017-02-03
Use default args now that node 6 is required.

XhmikosR:
There's also `polyfillNode4Support` which I can't quite look into now. If the name is wrong, the function should be renamed otherwise cleaned up.

### DIFF
--- a/lighthouse-cli/printer.ts
+++ b/lighthouse-cli/printer.ts
@@ -46,10 +46,7 @@ function checkOutputPath(path: string): string {
   return path;
 }
 
-function formatAggregationResultItem(score: boolean | number | string, suffix?: string) {
-  // Until we only support node 6 we can not use default args.
-  suffix = suffix || '';
-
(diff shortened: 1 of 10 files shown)


brendankenny:
maybe we should do one big pass for places where we can drop the || dance and use default parameters? A quick search didn't turn up that many, but enough that a PR for each is going to get pretty noisy :)

XhmikosR:
Agreed, but I just looked for node comments and saw this one and the `polyfillNode4Support` function.

XhmikosR:
@brendankenny: please check now. I enabled ESLint's `no-param-reassign` rule locally to quickly see the errors.

brendankenny:
@XhmikosR I fixed the merge conflict in the github UI and immediately regretted it :) Do you want to back that out, fix the merge conflict and I can land?

XhmikosR:
@brendankenny: rebased, check now.

brendankenny:
thanks again!

PR #1634, opened 2017-02-04
Support traces missing FCP

### LINKED ISSUE #1567: Support traces where a FCP isn't found
e.g. airhorner, our favorite contentless webapp.

Issue forked from https://github.com/GoogleChrome/lighthouse/pull/1549#discussion_r98312812

Will need some tweaks in `trace-of-tab.js`

paulirish:
Thanks airhorner for introducing this edgecase of a website without content. :)

Fixes #1567

I was able to remove the FMP tolerance bit we had before. 

I've also verified that the pwmetrics-events bit works perfectly now with a trace missing fCP. :)

image
image


ebidel:
FYI. needs a rebase if you're not already aware.

paulirish:
Thx. Rebased!

tdresser:
"Use last fMPCandidate if no FMP is marked."

This is a bit scary. If we take a 30 second trace, and FMP is 28 seconds, and we regress it by 3 seconds, we'll see FMP decrease, correct?

paulirish:
>> Use last fMPCandidate if no FMP is marked.

> This is a bit scary. If we take a 30 second trace, and FMP is 28 seconds, and we regress it by 3 seconds, we'll see FMP decrease, correct?

Yes. :) I don't know how avoidable this is, since FMP is willing to consider new paints as FMP until an indeterminate time.

On the other hand, this means we will report the 28s FMP as our FMP even if the detector is waiting around for other paints that end up with lower significance. (I'd wager this is far more likely, out here at 30 seconds in)

In both cases, these are terrible numbers, so that's the key thing to communicate. So IMO if we score a 28s FMP as 0.4/10 and a 31s FMP as 0.3/10, then it seems generally okay.  
I think this overall 30s window-of-consideration is comfortable.. if it were only 10s, then this would be a much bigger problem. :)

(I just checked and right now any FMP after 13 seconds gets a score of <1 out of 100.)

tdresser:
Yeah, that's probably reasonable.

@irori

PR #1635, opened 2017-02-04
refactor: DRY byte efficiency audits

### LINKED ISSUE #1612: Sort resources by potential savings, do not report negligible results
In the following case, I'd like to sort by the potential savings

image_savings

And in this one, 0 is irrelevant.

css_savings

Consider making report more dense and sharp to improve actionability.

patrickhulce:
R: @pavelfeldman all

* Moves the byte efficiency audits to their own folder
* Creates a base audit class for byte efficiency audits to use
* Sorts unused byte rows by their total savings in descending order (biggest bang for buck first)
* Relaxes the thresholds for failure 
  * **CSS Usage** will now only fail if at least 10ms would have been saved by eliminating the unused CSS
  * **Optimized Images** will now fail at 1MB total waste (up from 100k) or a JPEG wastes 25k (up from 1 byte) or an image would have saved 100k from conversion to WebP (up from 50k)
  * **Responsive Images** will only show images that are larger than their displayed size with DPR accounted for and will only fail if one of those images is 25k larger than necessary (changed from 10% of the pixels)

## How it looks now
image
(CSS Usage is green because this was run without throttling and it wouldn't have saved at least 10ms)

## Examples of Verge de-noising

**Before**
image
**After**
image

fixes #1612 

The smokehouse tests will fail thanks to us ignoring negligible results now. Not sure if we can find an elegant way of lowering the threshold or we have to just use full-size examples in smokehouse.


ebidel:
@patrickhulce this will need a rebase. Rebasing against the latest master will also pick up https://github.com/GoogleChrome/lighthouse/pull/1654.

patrickhulce:
PTAL, I'll update smoke tests once there's general agreement on the more lax thresholds.

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+const Audit = require('../audit');
+const Formatter = require('../../formatters/formatter');
+
+const KB_IN_BYTES = 1024;
+
+class UnusedBytes extends Audit {
```
patrickhulce: will do!

patrickhulce:
Thanks for braving the scary diff @brendankenny! :)

> what happens when we get a byte efficiency audit that doesn't completely fit this mold

I think we can refactor the shared pieces between that audit and these to a helper, but there's so much shared between generating the audit results here it felt like a waste to do anything less.

> For smokehouse, it would be a shame to lose it for testing these audits. We probably do want to add some assets big enough to hit the thresholds, but replacing everything might be excessive.

Agreed, what about moving these to a new folder and start matching our smokehouse expectations files to our folder structure? That way there's less noise too. We'd only really introduce a minimum of ~100k total to exercise failure in each of these.

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -32,7 +29,7 @@ class UnusedCSSRules extends Audit {
     return {
       category: 'CSS',
       name: 'unused-css-rules',
-      description: 'Uses 90% of its CSS rules',
+      description: 'Avoids unnecessary CSS',
```
ebidel: "unnecessary CSS" might not be descriptive enough.
What do you think about `Avoids including unnecessary CSS` or `Avoids loading unnecessary CSS` to correlate with the theme of reducing bytes on page load?

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
```diff
@@ -41,7 +40,7 @@ class UsesOptimizedImages extends Audit {
     return {
       category: 'Images',
       name: 'uses-optimized-images',
-      description: 'Has optimized images',
+      description: 'Avoids unoptimized images',
```
ebidel: Affirmative descriptions still feel better to me (where we can use them). How about aligning with PSI? They use  a singular "Optimize images".  WDYT about "Optimizes images" for us?

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
ebidel: "unnecessary CSS" might not be descriptive enough.
What do you think about Avoids including unnecessary CSS or Avoids loading unnecessary CSS to correlate with the theme of reducing bytes on page load?

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
```diff
@@ -93,65 +80,56 @@ class UsesOptimizedImages extends Audit {
...
-            webpSavings: 'WebP Savings (%)',
-            jpegSavings: 'JPEG Savings (%)',
-          }
-        }
+      results,
+      tableHeadings: {
+        preview: '',
+        url: 'URL',
+        totalKb: 'Original (KB)',
+        wastedKb: 'Savings (KB)',
```
ebidel: just a thought: should we put total savings last...as a way to indicate it's representative of the webp/jpeg column that provides the most impact.

patrickhulce:
Yeah sounds good, I like 'loading unnecessary CSS'

lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
```diff
@@ -39,7 +38,7 @@ class UsesResponsiveImages extends Audit {
     return {
       category: 'Images',
       name: 'uses-responsive-images',
-      description: 'Has appropriately sized images',
+      description: 'Avoids oversized images',
```
ebidel: "Uses appropriately sized images" or "Loads appropriately sized images"?

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
patrickhulce: +1 I like that idea, also makes the math clear ORIG x PERCENT = SAVINGS

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
patrickhulce: I switched them because I sympathized with the idea that telling users they did something when that thing doesn't apply to them doesn't make a lot of sense. I like the affirmative descriptions in isolation, but I do think consistency between audits within a category that's all about *not* doing something (shipping bytes) feels a lot smoother.

I definitely prefer "Optimizes images" to the double negative here, but could use guidance on the CSS affirmative, `Uses most of its CSS` is pretty lame

ebidel: >  telling users they did something when that thing doesn't apply to them 

Sort of why I think "Uses optimized images" is better than PSI's "Optimizes images". 

The first says: "All the images LH found were optimized. Good job!"
The second says: "You're (actively) optimizing images" ...which as you say, may not be the case.

patrickhulce:
updated smokehouse to have a new byte efficiency tester since there were no objections, tests should pass now PTAL :)

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
patrickhulce: Any thoughts on CSS wording changes to also be affirmative? I'll give up soon if you don't have any ideas either and just live with the mismatch to push this through, haha :P

ebidel: Yea......hmm. We can stick with "avoids *" for now.

patrickhulce:
with @paulirish's improvements:

image

paulirish:
love it.

PR #1636, opened 2017-02-04
Add inline image preview to table formatter

ebidel:
R: @patrickhulce all

This adds image previews to the formatter. Clicking the previews opens the full size version.

screen shot 2017-02-03 at 5 21 30 pm

### DIFF
--- a/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
+++ b/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
@@ -144,8 +144,8 @@ class UsesOptimizedImages extends Audit {
         value: {
           results,
           tableHeadings: {
-            url: 'URL',
             previewUrl: '',
+            url: 'URL',
             total: 'Original (KB)',
             webpSavings: 'WebP Savings (%)',
             jpegSavings: 'JPEG Savings (%)',
--- a/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
+++ b/lighthouse-core/audits/dobetterweb/uses-optimized-images.js
@@ -94,7 +94,6 @@ class UsesOptimizedImages extends Audit {
       }
 
       const originalKb = Math.round(image.originalSize / KB_IN_BYTES);
-      const url = URL.getDisplayName(image.url);
       const webpSavings = UsesOptimizedImages.computeSavings(image, 'webp');
 
       if (webpSavings.bytes > WEBP_ALREADY_OPTIMIZED_THRESHOLD_IN_BYTES) {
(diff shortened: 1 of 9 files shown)


paulirish:
Looks very nice. :)

I'd put them in the left column, before the URL. The network panel has icons like that.

ebidel:
> I'd put them in the left column, before the URL. The network panel has icons like that.

Sure, we can do that. It'll require some reworking of the CSS. Right now we make the URL 40% of the content width so you see most of it.

ebidel:
@paulirish PTAL. Moved the image preview to the left of the URL.

screen shot 2017-02-06 at 12 12 22 pm

patrickhulce:
niiiice this looks great

lighthouse-core/formatters/table.js
```diff
@@ -91,6 +94,11 @@ class Table extends Formatter {
         }
 
         switch (key) {
+          case 'previewUrl':
+            if (/^image/.test(result.mimeType)) {
```
patrickhulce: This feels like we're really doubling down on magic happening in the table formatter/assumptions about other keys that don't seem to have a relationship to the caller. What if `preview` were an object that contained`url` and an optional `mimeType`? Do we have other use cases for non-image previews yet?

lighthouse-core/audits/dobetterweb/uses-optimized-images.js
```diff
@@ -116,7 +115,9 @@ class UsesOptimizedImages extends Audit {
 
       totalWastedBytes += webpSavings.bytes;
       results.push({
-        url,
+        url: URL.getDisplayName(image.url),
+        previewUrl: image.url,
+        mimeType: image.mimeType,
```
paulirish: instead of mimeType you _could_ use resourceType. The difference is that it'll report Image even if a jpg's mimetype was plaintext.

```js
record.resourceType() == Common.resourceTypes.Image;
```

but looking closer...
 this would have to be done back in the gatherer. and we're already using mimetype to determine png/jpg anyway... So i guess it's moot in this case. Something to remember for the next one. :)

cc @patrickhulce

patrickhulce: Yeah I think a comment like this was made on my 4th network record PR, after this was already done :) Though in this case we specifically needed to check mimetype in the gatherer to avoid GIF/SVG/etc

lighthouse-core/formatters/table.js
ebidel: Good idea. It is a bit magical. 

Done.

ebidel:
PTAL

PR #1640, opened 2017-02-04
Update --perf config to latest

paulirish:
once we have tags in the default config (#1463), we can swap to that and won't have it fall behind again.

but until then..

### DIFF
--- a/lighthouse-core/config/perf.json
+++ b/lighthouse-core/config/perf.json
@@ -2,8 +2,22 @@
   "passes": [{
     "recordNetwork": true,
     "recordTrace": true,
-    "gatherers": []
-  }],
+    "gatherers": [
+      "url",
+      "image-usage",
+      "content-width"
+    ]
+  },
+  {
+    "passName": "css-image-usage",
+    "recordNetwork": true,
+    "gatherers": [
+      "styles",
+      "css-usage",
+      "dobetterweb/optimized-images"
+    ]
+  }
+  ],
 
   "audits": [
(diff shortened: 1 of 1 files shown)


lighthouse-core/config/perf.json
```diff
@@ -12,7 +26,10 @@
     "time-to-interactive",
     "user-timings",
     "screenshots",
-    "critical-request-chains"
+    "critical-request-chains",
+    "unused-css-rules",
+    "dobetterweb/uses-optimized-images",
+    "dobetterweb/uses-responsive-images"
```
ebidel: should we add these audits as well:

links-blocking-first-paint
scripts-blocking-first-paint

paulirish: yes thanks!

paulirish: added

paulirish:
PTAL

Reports look like this:

image
image

lighthouse-core/config/perf.json
```diff
@@ -2,8 +2,23 @@
   "passes": [{
     "recordNetwork": true,
     "recordTrace": true,
-    "gatherers": []
-  }],
+    "gatherers": [
+      "url",
+      "image-usage",
```
brendankenny: doesn't look like it will matter, but any reason to include these in the first pass instead of the second?

paulirish: nope.

PR #1641, opened 2017-02-04
convert DoBetterWeb gatherers to new error system

brendankenny:
with #1624 this completes the transition of the gatherers to using actual errors instead of -1 artifacts :)

These are all pretty straightforward so I grouped them more coarsely by similar gatherers per commit.

The biggest change was I changed the artifact returned by the call site gatherers (`Date.now`, `document.write`, etc) so instead of returning `{usage: [/* locations of call sites */]}` they just return `[/* locations of call sites */]`. If that's bad I can revert, but the indirection seems unnecessary now that we've lived with these for a while now.

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
@@ -147,15 +147,14 @@ class TagsBlockingFirstPaint extends Gatherer {
     return options.driver.evaluateScriptOnLoad(scriptSrc);
   }
 
+  /**
+   * @param {!Object} options
+   * @param {{networkRecords: !Array<!NetworkRecord>}} tracingData
+   * @return {!Array<{tag: string, transferSize: number, startTime: number, endTime: number}>}
+   */
   afterPass(options, tracingData) {
     return TagsBlockingFirstPaint
-      .findBlockingTags(options.driver, tracingData.networkRecords)
-      .catch(err => {
-        return {
-          value: -1,
-          debugString: err.message
-        };
-      });
+      .findBlockingTags(options.driver, tracingData.networkRecords);
   }
 }
 
--- a/lighthouse-core/gather/gatherers/dobetterweb/websql.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/websql.js
@@ -43,21 +43,15 @@ class WebSQL extends Gatherer {
     });
   }
 
+  /**
+   * Returns WebSQL database information or null if none was found.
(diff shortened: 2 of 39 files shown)


lighthouse-core/gather/gatherers/dobetterweb/websql.js
```diff
@@ -43,21 +43,15 @@ class WebSQL extends Gatherer {
     });
   }
 
+  /**
+   * Returns WebSQL database information or null if none was found.
+   * @param {!Object} options
+   * @return {?{id: string, domain: string, name: string, version: string}}
```
brendankenny: this one also changed a bit. Before this was returning the object returned by the protocol (which has all information on a property named `database`) on the artifact on a property also named `database`, so to get e.g. the name you'd have to do `artifacts.WebSQL.database.database.name`. Now it's `artifacts.WebSQL.name`, and `artifacts.WebSQL` is null if no database was found.

ebidel:
> The biggest change was I changed the artifact returned by the call site gatherers (Date.now, document.write, etc) so instead of returning {usage: [/* locations of call sites */]} they just return [/* locations of call sites */]. If that's bad I can revert, but the indirection seems unnecessary now that we've lived with these for a while now.

I think part of this was a symptom of the -1 error system. It was easy to check `artifacts.AnchorsWithNoRelNoopener === -1` to know something went wrong with the gatherer. Coercing multiple values into the artifact result always felt weird to me. So we stuck actual results on a `.usages` property. With your changes, all that weirdness goes bye bye :)

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
```diff
@@ -147,15 +147,14 @@ class TagsBlockingFirstPaint extends Gatherer {
     return options.driver.evaluateScriptOnLoad(scriptSrc);
   }
 
+  /**
+   * @param {!Object} options
+   * @param {{networkRecords: !Array<!NetworkRecord>}} tracingData
```
ebidel: `{!{networkRecords: !Array<!NetworkRecord>}}`?

brendankenny:
> I think part of this was a symptom of the -1 error system

Yeah, good point.

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
brendankenny: record types like this are non-nullable, so the `!` is technically correct but unnecessary. It would have been really nice if everything had been non-nullable by default from the start :(

PR #1650, opened 2017-02-07
More concise unit test running

paulirish:
_Update: only dot now.. no more landing strip :(_

Using the landing strip reporter for us.  It looks like this:
untitled

Pretty nice as it gives some idea of progress through the suite.

Travis logs every screen change, so I swapped it to the `dot` reporter, which looks like this:
image

Errors in the suite are still reported the exact same way. It's just you see the progress differently

-----

a few other things:

* in chrome-launcher-test I silenced all the logging output (all about "waiting for browser.."
  * This required changing how we tell debug() to shut up, but I like this way better
* similarly, the `pipe` change in index-test was to avoid seeing "Please provide a url" logged to terminal
* And I also nuked some unused CSS in a driveby

PTAL!

### DIFF
--- a/lighthouse-cli/test/cli/chrome-launcher-test.js
+++ b/lighthouse-cli/test/cli/chrome-launcher-test.js
@@ -19,15 +19,18 @@
 require('../../compiled-check.js')('chrome-launcher.js');
 
 const ChromeLauncher = require('../../chrome-launcher.js').ChromeLauncher;
+const log = require('../../../lighthouse-core/lib/log');
 const assert = require('assert');
 
 /* eslint-env mocha */
 
 describe('ChromeLauncher', () => {
   it('doesn\'t fail when killed twice', () => {
+    log.setLevel('error');
     const chromeInstance = new ChromeLauncher();
     return chromeInstance.run()
       .then(() => {
+        log.setLevel();
         return Promise.all([
           chromeInstance.kill(),
           chromeInstance.kill()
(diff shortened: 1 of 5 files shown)


ebidel:
Nice. I always been fond of the dot b/c it gives you progress and extra information. Should we just stick with one, or do you like the runway? ✈️ are pretty cool

patrickhulce:
Nice, landing is pretty dope. I do miss that we won't have visibility into which tests take forever (although we haven't around to using that information much yet I guess) but the output was insane before.

brendankenny:
> Should we just stick with one, or do you like the runway? I mean, ✈️ are pretty cool.

ha, I probably would prefer dot locally as well. Something about the progress bar not actually being 1:1 with the tests feels unsatisfying to me

brendankenny:
why did you have to manually suppress the logging in `chrome-launcher-test.js` but not in e.g. `gather-runner-test.js`, which has gotten a bit annoying with it's logged errors but isn't logging here anymore?

paulirish:
> why did you have to manually suppress the logging in `chrome-launcher-test.js` but not in e.g. `gather-runner-test.js`, which has gotten a bit annoying with it's logged errors but isn't logging here anymore?

I wasn't seeing it logging, but if it is, i'm happy to apply the suppression there too. :)

paulirish:
`dot` FTW. changed it to always be that.

Also check out the height of this travis log: https://travis-ci.org/GoogleChrome/lighthouse/jobs/199341666  oh yeah. :)

LGTMs?

brendankenny:
> I wasn't seeing it logging, but if it is, i'm happy to apply the suppression there too. :)

ha, same, I was just wondering how that happened. Did a few lighthouse runs and diffed the logs and everything looks the same, though

Issue #1552, opened 2017-01-27
Viewer: minify style.css for production.


XhmikosR:

XhmikosR:
On a side note, since this is a one-page app, how about inlining CSS and JS after minification?

brendankenny:
Closing in favor of #2591 tracking issue

PR #1652, opened 2017-02-07
Report: include partial's css only once

### LINKED ISSUE #1578: Report: move all CSS to `head`
Apart from it not being valid HTML, it's hard to debug it as the style tags now appear through the whole `body`.

If you don't want to move everything to `report.css`, maybe a build step could take care of this instead.

ebidel:
R: all

Fixes #1578.

Each html partial defines its own `<style>`. While this is convenient, it ends up injecting a bunch of duplicated CSS into the final report:

screen shot 2017-02-06 at 9 41 38 pm

This PR creates .css files for each partial and injects the styles once, at the top of the page.


XhmikosR:
@ebidel: shouldn't this fix #1578?

ebidel:
@XhmikosR it should take care of that too.

XhmikosR:
Cool, nice! Now if you added CSS minifying it would fix #1552, but perhaps you want to tackle that another time.

ebidel:
Been playing around with https://github.com/kangax/html-minifier but it's a huge dependency.

XhmikosR:
Yeah, unfortunately that grew up a lot.

You could use a css minifier only for starters. It's just that with html-minifier everything would be taken care by one root dependency.

XhmikosR:
I mean, it's using clean-css, uglify-js for CSS and JS respectively. The other deps are for unicode characters and some other stuff.

Personally, I've been using html-minifier for years in all my projects.

lighthouse-core/formatters/partials/speedline.css
```diff
@@ -0,0 +1,3 @@
+.speedline-measures {
```
ebidel: @paulirish I think you got rid of this in another PR?

lighthouse-core/report/templates/report-template.html
```diff
@@ -25,19 +24,20 @@
...
+  <link rel="shortcut icon" href="https://raw.githubusercontent.com/GoogleChrome/lighthouse/master/assets/lh_logo_icon.png">
```
ebidel: This shaves some bytes but also allows us to keep the icon up to date.

lighthouse-core/report/templates/report-template.html
```diff
@@ -25,19 +24,20 @@
...
   <title>Lighthouse report: {{ url }}</title>
-  {{#each stylesheets }}
-    <style>{{{ this }}}</style>
```
ebidel: it was silly to create separate `<style>` for each each sheet.

lighthouse-core/report/templates/report-template.html
XhmikosR: Personally, I don't like that. You are loading a 1303x1303 image for what? For 48x48 max.

ebidel: In theory, the 20kb will be cached and more shareable by the browser across different instances of the report.

XhmikosR: Or... you can just link to viewer's 48x48 icon.

lighthouse-core/formatters/partials/speedline.css
brendankenny: #1650

brendankenny: looks like you can just no include a css file for it (or leave an empty one so there's a place for future css?)

ebidel: Removed the file. There are other partials that don't have a .css file.

ebidel: Done

lighthouse-core/report/templates/report-template.html
brendankenny: it's just the favicon, so not sure if it matters, but this means the report won't be self contained

lighthouse-core/report/report-generator.js
```diff
@@ -214,7 +214,20 @@ class ReportGenerator {
    * @return {!Array<string>} an array of CSS
    */
   getReportCSS() {
-    return [fs.readFileSync(path.join(__dirname, './styles/report.css'), 'utf8')];
+    // Cannot DRY this up and dynamically create paths because fs.readdirSync
+    // doesn't browserify well with a variable path. See https://github.com/substack/brfs/issues/36.
+    const partialStyles = [
+      fs.readFileSync(path.join(__dirname, '../formatters/partials',
```
brendankenny: nit: can just do `__dirname + '../formatters/partials/critical-request-chains.css'` (for all of these)

lighthouse-core/report/templates/report-template.html
```diff
@@ -25,19 +24,20 @@
...
-  {{/each}}
+  <style>
+    {{#each stylesheets }}
+      {{{ this }}}
```
paulirish: can you add this for better debugging and editing w/ devtools?

```css
 /*# sourceURL=inline-styles-{{@index}}.css */
```

lighthouse-core/report/report-generator.js
ebidel: done

lighthouse-core/report/templates/report-template.html
ebidel: added for the single `<style>`

paulirish:
Yup lgtm 
image

PR #1657, opened 2017-02-07
Report: use the same favicon as the viewer.

XhmikosR:
Saves 0.3KB.

Also switch to the `link rel=icon` which is spec-compliant.


brendankenny:
is this just reverting part of #1652?

And FWIW, the spec [says no prob to `shortcut`](https://html.spec.whatwg.org/multipage/semantics.html#rel-icon:rel-icon-5), even if it is increasingly archaic

XhmikosR:
Well, I had created this patch before the other one.

As for the spec yes shortcut icon became valid because of ancient IE and
only for that. Plain icon is the proper syntax.

It's your call what to do with this. I have expressed my personal opinion
on the other PR that loading a huge icon is overkill and second the report
isn't self contained anymore.

A middle solution would be to commit a 48x48 icon and link to that.

paulirish:
+1 to not loading a 1000px image for the favicon. TBH lighthouse should have something to say about that. :)

PR #1660, opened 2017-02-07
Report: refactor handelbar helpers into static methods

### LINKED ISSUE #1581: Report: refactor handelbar helpers into static methods
This will make them easier to test.

sendilkumarn:
fixes #1581


lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,233 @@
...
+    return {
+      nameToLink(name) {
+        return HandlebarHelpers.nameToLink(name);
+      },
+
+      getItemRating(value) {
+        return HandlebarHelpers.getItemRating(value);
+      },
+
+      'getTotalScore': HandlebarHelpers.getTotalScore,
```
patrickhulce: We can probably structure all of these helpers like this without having to create new functions.
Also unnecessary `'` :)

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,233 @@
...
+      'is-bool': (value) => {
+        return HandlebarHelpers.isBool(value);
+      },
+
+      not(value) {
+        return HandlebarHelpers.not(value);
+      },
+
+      if_eq(lhs, rhs, options) {
+        return HandlebarHelpers.ifEq(lhs, rhs, options);
```
patrickhulce: the `this` binding will break here and on `if_not_eq` currently, fixing above comment should help

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,233 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All rights reserved.
```
ebidel: 2017

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,233 @@
...
+    };
+    renderer.codespan = function(str) {
+      return `<code>${str}</code>`;
+    };
+    // eslint-disable-next-line no-unused-vars
+    renderer.code = function(code, language) {
+      return `<pre>${code}</pre>`;
+    };
+    // Nuke wrapper <p> tag that gets generated.
+    renderer.paragraph = function(str) {
```
ebidel: This is missing `renderer.image`: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/report-generator.js#L147.

Are you refactoring off the latest master? That one recently landed.

ebidel:
Ah, I see the rebase errors now :)

lighthouse-core/report/handlebar-helpers.js
brendankenny: > We can probably structure all of these helpers like this without having to create new functions

maybe we should just avoid defining this as a `class` and have it be an object literal? That way in `report-generator.js` it could just be

```js
const handlebarHelpers = require('./handlebar-helpers');
// ...
Handlebars.registerHelper(handlebarHelpers);
```

paulirish:
Nice. Thanks for the awesome contribution, @sendilkumarn :)

lighthouse-core/report/handlebar-helpers.js
sendilkumarn: Initially I had this as an object literal, and it works. Since everywhere we defined it as a class, wanted to align with others

sendilkumarn:
@ebidel rebased, I missed that merge by minutes :)

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,182 @@
...
+const getItemRating = value => {
+  if (typeof value === 'boolean') {
+    return value ? RATINGS.GOOD.label : RATINGS.POOR.label;
+  }
+  return calculateRating(value);
+};
+
+const handlebarHelpers = {
+  // Converts a name to a link.
+  'nameToLink': name => {
```
ebidel: you can ditch the quotes around these keys.

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,182 @@
...
+
+  // Converts the total score to a rating that can be used for styling.
+  'getTotalScoreRating': aggregation => {
+    const totalScore = getTotalScore(aggregation);
+    return calculateRating(totalScore);
+  },
+
+  // Converts a value to a rating string, which can be used inside the report
+  // for color styling.
+  'getItemRating': getItemRating,
```
ebidel: es6 shorthand is good here: `getItemRating,`

lighthouse-core/test/report/handlebar-helpers-test.js
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+const handlebarHelpers = require('../../report/handlebar-helpers.js');
+const assert = require('assert');
+
+/* eslint-env mocha */
+
+// Most of the functionality is tested via the Printer class, but in this
```
ebidel: also mostly tested through the various formatter tests.

lighthouse-core/report/handlebar-helpers.js
sendilkumarn: the problem is that for `is-bool` we need to have the quotes

we will have conflict with `quote-props` in eslint

lighthouse-core/report/handlebar-helpers.js
sendilkumarn: we can change `getTotalScore` too

lighthouse-core/report/handlebar-helpers.js
brendankenny: how does everyone feel about renaming the helper methods so they're all camelCase?

I don't feel strongly, but it would at least bring some consistency and avoid the quote thing (they aren't even consistent currently, obviously. some care camelCase, some are snake_case, and at least `is-bool` is kebab-case (which I just learned has a name))

The remaining question would be is e.g. `isNotEq` still readable in the html template, I guess.

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,182 @@
...
+  },
+
+  // value is boolean?
+  'is-bool': value => (typeof value === 'boolean'),
+
+  // !value
+  'not': value => !value,
+
+  // value === value2?
+  'if_eq': (lhs, rhs, options) => {
```
patrickhulce: I believe we can't actually use es6 shorthand here since it will change `this` to become the global instead of the applied context of this which is what we want, same goes for `if_not_eq`

lighthouse-core/report/handlebar-helpers.js
patrickhulce: +1 to all camelCase, though I do love me some 🍖 -case

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -0,0 +1,182 @@
...
+  let rating = RATINGS.POOR.label;
+  if (value >= RATINGS.GOOD.minScore) {
+    rating = RATINGS.GOOD.label;
+  } else if (value >= RATINGS.AVERAGE.minScore) {
+    rating = RATINGS.AVERAGE.label;
+  }
+  return rating;
+}
+
+const getTotalScore = aggregation => {
```
patrickhulce: can we use `function getTotalScore(aggregation)` instead? basically same length but gets us better stack traces

same for below

lighthouse-core/test/report/handlebar-helpers-test.js
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+const handlebarHelpers = require('../../report/handlebar-helpers.js');
+const assert = require('assert');
+
+/* eslint-env mocha */
+
+// Most of the functionality is tested via the Formatter tests, but in this
```
patrickhulce: awesome thanks!

lighthouse-core/report/handlebar-helpers.js
sendilkumarn: excellent then, I will convert them to 🐫 Case 👍

sendilkumarn:
@brendankenny 👍

patrickhulce:
thanks @sendilkumarn this is great!

PR #1665, opened 2017-02-07
feat: save devtools log

### LINKED ISSUE #1662: WPT compat: save DevTools message log with trace
WPT needs the log of messages received over the protocol which LH does not include with the trace.

patrickhulce:
fixes #1662

### DIFF
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -18,6 +18,7 @@
 
 const EventEmitter = require('events').EventEmitter;
 const log = require('../../lib/log.js');
+const MessageLog = require('./message-log');
 
 class Connection {
 
@@ -26,27 +27,7 @@ class Connection {
     /** @type {!Map<number, {resolve: function(*), reject: function(*), method: string}>}*/
     this._callbacks = new Map();
     this._eventEmitter = new EventEmitter();
-    this._messageLog = [];
-    this._recordMessages = false;
-  }
-
-  /**
-   * @return {!Array<Object>}
-   */
-  get messageLog() {
-    return this._messageLog;
-  }
-
-  resetMessageLog() {
-    this._messageLog = [];
-  }
-
-  beginRecording() {
-    this._recordMessages = true;
-  }
-
-  endRecording() {
-    this._recordMessages = false;
+    this.log = new MessageLog();
   }
 
   /**
@@ -130,10 +111,7 @@ class Connection {
       }));
     }
 
-    if (this._recordMessages) {
-      this._messageLog.push(object);
-    }
-
+    this.log.record(object);
     log.formatProtocol('<= event',
         {method: object.method, params: object.params}, 'verbose');
     this.emitNotification(object.method, object.params);
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -26,6 +26,27 @@ class Connection {
     /** @type {!Map<number, {resolve: function(*), reject: function(*), method: string}>}*/
     this._callbacks = new Map();
     this._eventEmitter = new EventEmitter();
+    this._messageLog = [];
+    this._recordMessages = false;
+  }
+
+  /**
+   * @return {!Array<Object>}
+   */
+  get messageLog() {
+    return this._messageLog;
+  }
+
+  resetMessageLog() {
+    this._messageLog = [];
+  }
+
+  beginRecording() {
+    this._recordMessages = true;
+  }
+
+  endRecording() {
+    this._recordMessages = false;
   }
 
   /**
@@ -108,6 +129,11 @@ class Connection {
         return object.result;
       }));
     }
+
+    if (this._recordMessages) {
+      this._messageLog.push(object);
+    }
+
     log.formatProtocol('<= event',
         {method: object.method, params: object.params}, 'verbose');
     this.emitNotification(object.method, object.params);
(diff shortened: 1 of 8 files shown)


lighthouse-core/lib/asset-saver.js
```diff
@@ -143,11 +145,14 @@ function prepareAssets(artifacts, audits) {
 function saveAssets(artifacts, audits, pathWithBasename) {
   return prepareAssets(artifacts, audits).then(assets => {
     assets.forEach((data, index) => {
-      const traceData = data.traceData;
       const traceFilename = `${pathWithBasename}-${index}.trace.json`;
-      fs.writeFileSync(traceFilename, JSON.stringify(traceData, null, 2));
```
paulirish: you mentioned the WPT trace parser doesn't do pretty JSON.
Should we start saving it in the right format? (or another PR i suppose)

lighthouse-core/gather/connections/connection.js
```diff
@@ -108,6 +129,11 @@ class Connection {
         return object.result;
       }));
     }
+
+    if (this._recordMessages) {
```
brendankenny: is this only for events, not command responses?

lighthouse-core/gather/connections/connection.js
```diff
@@ -26,6 +26,27 @@ class Connection {
...
+   */
+  get messageLog() {
+    return this._messageLog;
+  }
+
+  resetMessageLog() {
+    this._messageLog = [];
+  }
+
+  beginRecording() {
```
paulirish: i think we should have a whitelist filter to only record given domains.  that way we could only do Page/Network which is all that WPT needs (and same for Lighthouse in its artifact-driven run).

honestly the API doesn't really need to be configurable.. maybe just set the filter as a property on this class?

lighthouse-core/gather/connections/connection.js
```diff
@@ -26,6 +26,27 @@ class Connection {
...
     this._callbacks = new Map();
     this._eventEmitter = new EventEmitter();
+    this._messageLog = [];
+    this._recordMessages = false;
+  }
+
+  /**
+   * @return {!Array<Object>}
+   */
+  get messageLog() {
```
paulirish: maybe bump this stuff into it's own class and instantiate it in the connection constructor?

lighthouse-core/test/lib/asset-saver-test.js
```diff
@@ -78,6 +82,13 @@ describe('asset-saver helper', () => {
       fs.unlinkSync(traceFilename);
     });
 
+    it('devtools log file saved to disk with data', () => {
+      const filename = 'the_file-0.devtools.json';
+      const fileContents = fs.readFileSync(filename, 'utf8');
+      assert.ok(fileContents.includes('"message": "first"'));
```
paulirish: are there any constraints around the format for this file?  (aka not pretty print?)

lighthouse-core/gather/connections/connection.js
patrickhulce: yes, figured that would be distracting and not make very much sense without the request payload

lighthouse-core/lib/asset-saver.js
```diff
@@ -143,11 +145,14 @@ function prepareAssets(artifacts, audits) {
...
+      fs.writeFileSync(traceFilename, JSON.stringify(data.traceData, null, 2));
       log.log('trace file saved to disk', traceFilename);
 
+      const devtoolsLogFilename = `${pathWithBasename}-${index}.devtools.json`;
```
paulirish: can we use "*devtoolslog.json" rather?

patrickhulce:
WPT uses the message log to generate most of what you see in a report, network waterfalls, request breakdowns, timing information, etc. For that it needs the page and network events, and from the samples I've done the trace is *far* larger (~5-30x) than the devtools log.

lighthouse-core/test/lib/asset-saver-test.js
patrickhulce: no constraints on this one

lighthouse-core/lib/asset-saver.js
patrickhulce: I'm planning on handling the pretty printed file in WPT since I don't think its an excuse to be human-hostile here

lighthouse-core/lib/asset-saver.js
patrickhulce: will do

brendankenny:
> WPT uses the message log to generate most of what you see in a report, network waterfalls, request breakdowns, timing information, etc.

Got it. I do like @paulirish's idea of filtering messages saved.

> For that it needs the page and network events, and from the samples I've done the trace is far larger (~5-30x) than the devtools log.

Right, but we don't stream in the trace until after we're done tracing :) If it's just events and just for the two domains it should probably be fine

patrickhulce:
Filtering to just `Page` and `Network` cuts down about 33% of the file size.

lighthouse-core/gather/connections/connection.js
paulirish: yup, we dont need command reqs/responses.

lighthouse-core/gather/connections/connection.js
patrickhulce: will do

lighthouse-core/gather/connections/connection.js
patrickhulce: will do

brendankenny:
I didn't get to this fast enough, but some drive by thoughts when future changes touch the code:

- it may be better for the log to live totally in driver. Could easily happen in the `'notification'` event handler currently added in the driver constructor. That would keep it from being partially in connection and partially in driver reaching into connection, and it would mean connection wouldn't have to know about a class that knows about contents of the messages its handling.

- Should `MessageLog` be `DevtoolsLog` to match the asset being saved?

- Really need some comments :) No one reading is going to know what `MessageLog` is doing and why it's only saving those particular events

PR #1667, opened 2017-02-08
perf: gather all event listeners in parallel

brendankenny:
this was some low hanging performance fruit discussed last week, getting all the event listeners for all nodes in the page in parallel rather than waiting on each command to finish. We aren't enabling/disabling domains or listening for events, so this should be safe.

Profile for the `EventListeners` gatherer doesn't look all that different but it does save e.g. 700ms (1300 -> 600) when running on www.chromestatus.com.

before:
screen shot 2017-02-07 at 17 09 05

after:
screen shot 2017-02-07 at 17 09 20

(purple block under 11750ms is all the commands now being sent at once. Remaining scatter of blocks is now just returning messages, not interleaved returning messages and sending messages)

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
@@ -112,13 +112,10 @@ class EventListeners extends Gatherer {
    *     listeners found across the elements.
    */
   collectListeners(nodes) {
-    return nodes.reduce((chain, node) => {
-      return chain.then(prevArr => {
-        // Call getEventListeners once for each node in the list.
-        return this.getEventListeners(node.element ? node.element.nodeId : node)
-            .then(result => prevArr.concat(result));
-      });
-    }, Promise.resolve([]));
+    // Gather event listeners from each node in parallel.
+    return Promise.all(nodes.map(node => {
+      return this.getEventListeners(node.element ? node.element.nodeId : node);
+    })).then(nestedListeners => [].concat(...nestedListeners));
   }
 
   beforePass(options) {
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
```diff
@@ -112,13 +112,10 @@ class EventListeners extends Gatherer {
...
-      return chain.then(prevArr => {
-        // Call getEventListeners once for each node in the list.
-        return this.getEventListeners(node.element ? node.element.nodeId : node)
-            .then(result => prevArr.concat(result));
-      });
-    }, Promise.resolve([]));
+    // Gather event listeners from each node in parallel.
+    return Promise.all(nodes.map(node => {
+      return this.getEventListeners(node.element ? node.element.nodeId : node);
+    })).then(nestedListeners => [].concat(...nestedListeners));
```
ebidel: do you still need the concat?

```
return Promise.all(nodes.map(node => {
  return this.getEventListeners(node.element ? node.element.nodeId : node);
}))
```

brendankenny:
sneaky post-approval commit: @paulirish reminded me we had also discussed disabling websocket compression since our payloads are so tiny and the vast majority of the time in `sendCommand` and its callbacks is spent compressing and decompressing messages. So I've disabled it.

*However*, the win is basically imperceptible. The time spent per command has dropped, but the execution time of all commands was so small the reduction had little effect on overall execution time.

However however, the `ws` readme recommends:

> The [permessage-deflate] extension is enabled by default but adds a significant overhead in terms of performance and memory consumption. We suggest to use WebSocket compression only if it is really needed.

and there may come a time where we need a gatherer that runs many commands in a row that can't be run in parallel, so leaving compression off as the new default makes sense.

ebidel:
🚫 🏋️

paulirish:
LGTM! so good

PR #1668, opened 2017-02-08
perf: tracingModel to computed artifact

paulirish:
Right now we're generating a traceviewer tracingModel twice.

It's fairly expensive, so by moving it to a computed artifact, we get caching for free. Woo!

### DIFF
--- a/lighthouse-core/audits/time-to-interactive.js
+++ b/lighthouse-core/audits/time-to-interactive.js
@@ -63,20 +63,15 @@ class TTIMetric extends Audit {
    */
   static audit(artifacts) {
     const trace = artifacts.traces[Audit.DEFAULT_PASS];
-    const pendingSpeedline = artifacts.requestSpeedline(trace);
-    const pendingFMP = FMPMetric.audit(artifacts);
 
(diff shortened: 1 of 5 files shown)


brendankenny:
before:
screen shot 2017-02-07 at 19 09 00

after:
screen shot 2017-02-07 at 19 08 09

savings of 2 seconds (± whatever rounding `debug` uses)!

lighthouse-core/audits/time-to-interactive.js
```diff
@@ -63,20 +63,15 @@ class TTIMetric extends Audit {
...
-      const speedline = results[0];
-      const fmpResult = results[1];
-
-      // Process the trace
-      const tracingProcessor = new TracingProcessor();
-      const trace = artifacts.traces[Audit.DEFAULT_PASS];
-      const model = tracingProcessor.init(trace);
+    const pending = [
+      artifacts.requestSpeedline(trace),
+      FMPMetric.audit(artifacts),
```
brendankenny: I guess this could have moved to `traceOfTab` to avoid calling into another audit. Next time :)

Issue #1464, opened 2017-01-13
Problems with mesuring slow sites


denar90:
Hi. We got some [issue](https://github.com/paulirish/pwmetrics/issues/28) in `pwmetrics` project, where users and me personally faced some problems using `ligthouse` API for getting site metrics. So I used @paulirish [suggestion](https://github.com/paulirish/pwmetrics/issues/28#issuecomment-272351713) and run trace stuff (cloned repo and run `node lighthouse-cli --save-assets --save-artifacts https://www.cnet.com/special-reports/jony-ive-talks-about-putting-the-apple-touch-on-the-macbook-pro/ --output=json > results.json`). I'm attaching archive with results for kinda `slow` [site](https://www.cnet.com/special-reports/jony-ive-talks-about-putting-the-apple-touch-on-the-macbook-pro/). 
Hope it will help you guys.

P.S. Don't hesitate to ask me for additional info or smth.

[results.zip](https://github.com/GoogleChrome/lighthouse/files/704285/results.zip)

Thanks.

denar90:
One more thing. After debugging I found out that Lighthouse throws error: [`No usable firstMeaningfulPaint event found in trace`](https://github.com/GoogleChrome/lighthouse/blob/a7648e7eeef7dce5641147f2bf75725a084e8937/lighthouse-core/audits/first-meaningful-paint.js#L148) twice.

wardpeet:
looks like the trace is getting too big and you get a memory exception

tmf:
Hi @denar90 @wardpeet, 
First: a big thank you for this awesome project!

I also looked into this issue and found that there could be several reasons for this (not finding a firstMeaningfulPaint event in the trace):
- fMP (firstMeaningfulPaint) is recorded before the fCP (firstContentfulPaint) - [FCPFMP_TOLERANCE](https://github.com/GoogleChrome/lighthouse/blob/a7648e7eeef7dce5641147f2bf75725a084e8937/lighthouse-core/audits/first-meaningful-paint.js#L144): in this case, I think, the metric cannot be determined...
[...]

PR #1672, opened 2017-02-09
feat: configurable page timeout

### LINKED ISSUE #1671: WPT compat: configurable page timeout
with deep network throttling many pages fail to load in 25s which yields an incomplete trace/message log that isn't very useful

patrickhulce:
fixes #1671

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -141,16 +141,6 @@ if (cliFlags.listTraceCategories) {
 
 const url = cliFlags._[0];
 
-// Work around camelCase bug for default value in yargs 3.30.
-// see: https://github.com/yargs/yargs/issues/341
-if (!cliFlags.outputPath && cliFlags['output-path']) {
-  cliFlags.outputPath = cliFlags['output-path'];
-}
-
-if (!cliFlags.maxWaitForLoad && cliFlags['max-wait-for-load']) {
-  cliFlags.maxWaitForLoad = cliFlags['max-wait-for-load'];
-}
-
 let config: Object | null = null;
 if (cliFlags.configPath) {
   // Resolve the config file path relative to where cli was called.
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -64,7 +64,7 @@ const cliFlags = yargs
     'config-path',
     'perf',
     'port',
-    'timeout'
+    'max-wait-for-load'
   ], 'Configuration:')
   .describe({
     'disable-device-emulation': 'Disable Nexus 5X emulation',
@@ -77,7 +77,7 @@ const cliFlags = yargs
     'config-path': 'The path to the config JSON.',
     'perf': 'Use a performance-test-only configuration',
     'port': 'The port to use for the debugging protocol. Use 0 for a random port',
-    'timeout': 'The timeout in seconds to wait for the page to load before continuing',
+    'max-wait-for-load': 'The timeout in milliseconds to wait for the page to load before continuing. WARNING: Very high values can lead to large traces and instability',
     'skip-autolaunch': 'Skip autolaunch of Chrome when already running instance is not found',
     'select-chrome': 'Interactively choose version of Chrome to use when multiple installations are found',
     'interactive': 'Open Lighthouse in interactive mode'
@@ -117,7 +117,7 @@ Example: --output-path=./lighthouse-results.html`
   .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.pretty])
   .default('output-path', 'stdout')
   .default('port', 9222)
-  .default('timeout', Driver.MAX_WAIT_FOR_FULLY_LOADED / 1000)
+  .default('max-wait-for-load', Driver.MAX_WAIT_FOR_FULLY_LOADED)
   .check((argv: {listAllAudits?: boolean, listTraceCategories?: boolean, _: Array<any>}) => {
     // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
     // or --list-trace-categories. If not, stop the program and ask for a url
@@ -145,6 +145,10 @@ const url = cliFlags._[0];
 // see: https://github.com/yargs/yargs/issues/341
 if (!cliFlags.outputPath && cliFlags['output-path']) {
   cliFlags.outputPath = cliFlags['output-path'];
+}
+
+if (!cliFlags.maxWaitForLoad && cliFlags['max-wait-for-load']) {
+  cliFlags.maxWaitForLoad = cliFlags['max-wait-for-load'];
 }
 
 let config: Object | null = null;
@@ -283,7 +287,7 @@ function saveResults(results: Results,
 function runLighthouse(url: string,
                        flags: {port: number, skipAutolaunch: boolean, selectChrome: boolean, output: any,
                          outputPath: string, interactive: boolean, saveArtifacts: boolean, saveAssets: boolean
-                         timeout: number},
+                         maxWaitForLoad: number},
                        config: Object): Promise<undefined> {
 
   let chromeLauncher: ChromeLauncher;
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -77,7 +77,7 @@ const cliFlags = yargs
     'config-path': 'The path to the config JSON.',
     'perf': 'Use a performance-test-only configuration',
     'port': 'The port to use for the debugging protocol. Use 0 for a random port',
-    'max-wait-for-load': 'The timeout in milliseconds to wait for the page to load before continuing. WARNING: Very high values can lead to large traces and instability',
+    'max-wait-for-load': 'The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING: Very high values can lead to large traces and instability',
     'skip-autolaunch': 'Skip autolaunch of Chrome when already running instance is not found',
     'select-chrome': 'Interactively choose version of Chrome to use when multiple installations are found',
     'interactive': 'Open Lighthouse in interactive mode'
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -62,7 +62,8 @@ const cliFlags = yargs
     'list-trace-categories',
     'config-path',
     'perf',
-    'port'
+    'port',
+    'timeout'
   ], 'Configuration:')
   .describe({
     'disable-device-emulation': 'Disable Nexus 5X emulation',
@@ -75,6 +76,7 @@ const cliFlags = yargs
     'config-path': 'The path to the config JSON.',
     'perf': 'Use a performance-test-only configuration',
     'port': 'The port to use for the debugging protocol. Use 0 for a random port',
+    'timeout': 'The timeout in seconds to wait for the page to load before continuing',
     'skip-autolaunch': 'Skip autolaunch of Chrome when already running instance is not found',
     'select-chrome': 'Interactively choose version of Chrome to use when multiple installations are found',
     'interactive': 'Open Lighthouse in interactive mode'
@@ -114,6 +116,7 @@ Example: --output-path=./lighthouse-results.html`
   .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.pretty])
   .default('output-path', 'stdout')
   .default('port', 9222)
+  .default('timeout', 25)
   .check((argv: {listAllAudits?: boolean, listTraceCategories?: boolean, _: Array<any>}) => {
     // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
     // or --list-trace-categories. If not, stop the program and ask for a url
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -420,10 +420,11 @@ class Driver {
    * - MAX_WAIT_FOR_FULLY_LOADED milliseconds have passed.
    * See https://github.com/GoogleChrome/lighthouse/issues/627 for more.
    * @param {number} pauseAfterLoadMs
+   * @param {number} maxTimeoutMs
    * @return {!Promise}
    * @private
    */
-  _waitForFullyLoaded(pauseAfterLoadMs) {
+  _waitForFullyLoaded(pauseAfterLoadMs, maxTimeoutMs) {
```
paulirish: naming: i prefer the clarity of the old name.. how about `maxWaitForLoadedMs` ?

lighthouse-core/gather/driver.js
```diff
@@ -477,11 +478,15 @@ class Driver {
     const waitForLoad = _options.waitForLoad || false;
     const disableJS = _options.disableJavaScript || false;
     const pauseAfterLoadMs = (_options.flags && _options.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
+    let maxWaitMs = MAX_WAIT_FOR_FULLY_LOADED;
+    if (_options.flags && _options.flags.timeout) {
+      maxWaitMs = 1000 * _options.flags.timeout;
```
paulirish: `x  = y * 1000`

lighthouse-cli/bin.ts
```diff
@@ -114,6 +116,7 @@ Example: --output-path=./lighthouse-results.html`
   .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.pretty])
   .default('output-path', 'stdout')
   .default('port', 9222)
+  .default('timeout', 25)
```
paulirish: since the `MAX_WAIT_FOR_FULLY_LOADED` is still in driver, seems like we should let this value fall through as `undefined`? wdyt

lighthouse-core/gather/driver.js
patrickhulce: will do, anything magic I'm not aware of or just style? I know in this case it doesn't matter but I developed a habit of putting the type to coerce to first when dealing with potentially unknown types to avoid `{} + ''` vs `'' + {}` scenarios :)

lighthouse-core/gather/driver.js
patrickhulce: will do

lighthouse-cli/bin.ts
```diff
@@ -62,7 +62,8 @@ const cliFlags = yargs
     'list-trace-categories',
     'config-path',
     'perf',
-    'port'
+    'port',
+    'timeout'
```
brendankenny: `timeout` makes me think of a timeout for the whole run or something. Maybe some non-awkward version of `page-load-timeout`?

lighthouse-core/gather/driver.js
paulirish: ah i see. totally a style thing.. I'm just used to seeing the integers on the right side.

but good call on doing this for protection with +.

nbd either way

lighthouse-cli/bin.ts
patrickhulce: we double default disableCpuThrottling as well and the extra yargs hints to users won't hurt IMO

lighthouse-cli/bin.ts
patrickhulce: yeah I started with that but it felt so clunky and didn't make a ton of sense to have any other timeout, what about adopting the `max-wait-for-load` name from within if we change?

patrickhulce:
One more thing to note, this can get dangerous if you always set it very high. Testing the verge leads to ugly OOM crashes if you actually try and wait for it when a particularly nasty ad loads...

lighthouse-cli/bin.ts
brendankenny: main benefit of having it here is the `--help` printing, but if we want a default it has to happen internally as well for anyone using lighthouse as a module.

What about exposing on Driver as a constant, so it would still be single source of truth `.default('timeout', Driver.MAX_WAIT_FOR_FULLY_LOADED)`?

brendankenny:
what does WPT do in those cases? Is it possible to cleanly wrap up the trace before load has finished?

lighthouse-cli/bin.ts
patrickhulce: A bit overkill to bring in driver just for the constant to me. I don't think it's even that crazy that programmatic access and CLI access could have different sensible defaults at some point either, but I'm fine with removing if we like.

brendankenny: I think in general (at least from a documentation perspective) it's better to make sure they're the same, but in this case my main problem would be development in driver, where you'd have to remember why changes to `MAX_WAIT_FOR_FULLY_LOADED` aren't doing anything when the CLI is run.

patrickhulce:
> what does WPT do in those cases? Is it possible to cleanly wrap up the trace before load has finished?

AFAICT there's nothing preventing it from also suffering from the OOM case, it just has a timeout that's a little more reasonable that's more likely to contain the events we care about and doesn't process the traces in the same process (does it incrementally in a separate python process)

lighthouse-cli/bin.ts
patrickhulce: k, done

patrickhulce:
PTAL :)

lighthouse-cli/bin.ts
brendankenny: Thinking about this now, I wouldn't think `timeout` was the whole run, and there are better tools to accomplish that if you do want to limit that on the command line.

The only other thing I would guess as a user of this would be that `timeout` would apply to the entirety of each pass, as in not just page load but (∑`beforePass()`es + page load + ∑`afterPass()`es), to make sure a particular pass doesn't go on forever. As long as our gatherers are relatively bounded, though, these are roughly the same thing.

(I guess custom gatherers could go crazy, but that's on them)

I think I'm good with going with `--timeout`, but since it's such prime flag real estate (and not necessarily self-descriptive) maybe @paulirish and @ebidel want to weigh in.

lighthouse-cli/bin.ts
```diff
@@ -75,6 +77,7 @@ const cliFlags = yargs
     'config-path': 'The path to the config JSON.',
     'perf': 'Use a performance-test-only configuration',
     'port': 'The port to use for the debugging protocol. Use 0 for a random port',
+    'timeout': 'The timeout in seconds to wait for the page to load before continuing',
```
brendankenny: should we add a warning about large traces and OOM? If you're frustrated with slow page loads (#1464), a large number might be the first thing you try. A full warning may be more appropriate for the readme but more people will see something brief written here.

lighthouse-cli/bin.ts
```diff
@@ -114,6 +117,7 @@ Example: --output-path=./lighthouse-results.html`
   .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.pretty])
   .default('output-path', 'stdout')
   .default('port', 9222)
+  .default('timeout', Driver.MAX_WAIT_FOR_FULLY_LOADED / 1000)
```
brendankenny: are seconds the usual thing for command line timeouts? (obvs most of the js world tends towards ms)

lighthouse-core/gather/driver.js
```diff
@@ -477,11 +480,15 @@ class Driver {
     const waitForLoad = _options.waitForLoad || false;
     const disableJS = _options.disableJavaScript || false;
```
brendankenny: while you're in here, do you want to get rid of `_options` and make `options` a default `{}`

lighthouse-cli/bin.ts
```diff
@@ -278,7 +282,8 @@ function saveResults(results: Results,
 
 function runLighthouse(url: string,
                        flags: {port: number, skipAutolaunch: boolean, selectChrome: boolean, output: any,
-                         outputPath: string, interactive: boolean, saveArtifacts: boolean, saveAssets: boolean},
+                         outputPath: string, interactive: boolean, saveArtifacts: boolean, saveAssets: boolean
+                         timeout: number},
```
brendankenny: is this just for documentation? I wouldn't think TS would care since we don't use it here

lighthouse-cli/bin.ts
ebidel: I think we should be more specific. `--timeout` feels too generic and it's clear from this discussion, that it's ambiguous what it _could_ mean :) `page-load-timeout` or `max-wait-for-load` would be cool.

lighthouse-cli/bin.ts
patrickhulce: When it's a CLI interface for something whose order of magnitude is nearly always measured in seconds I've seen seconds more often than ms, but with a name change to more reflect that it's an internal code setting I think a move to ms is more reasonable.

lighthouse-cli/bin.ts
patrickhulce: Yeah just for documentation.

lighthouse-core/gather/driver.js
patrickhulce: will do

lighthouse-cli/bin.ts
patrickhulce: alright sounds good, I'll go with `max-wait-for-load` for consistency

lighthouse-cli/bin.ts
patrickhulce: Done

lighthouse-cli/bin.ts
```diff
@@ -143,6 +147,10 @@ if (!cliFlags.outputPath && cliFlags['output-path']) {
   cliFlags.outputPath = cliFlags['output-path'];
 }
 
+if (!cliFlags.maxWaitForLoad && cliFlags['max-wait-for-load']) {
```
paulirish: how would this be possible? AFAIK, yargs creates the former and makes sure they match at this point.

patrickhulce: The bug as mentioned in the above identical check makes this happen when the default is used.

lighthouse-cli/bin.ts
```diff
@@ -75,6 +77,7 @@ const cliFlags = yargs
     'config-path': 'The path to the config JSON.',
     'perf': 'Use a performance-test-only configuration',
     'port': 'The port to use for the debugging protocol. Use 0 for a random port',
+    'max-wait-for-load': 'The timeout in milliseconds to wait for the page to load before continuing. WARNING: Very high values can lead to large traces and instability',
```
ebidel: nit: The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING:...`

lighthouse-cli/bin.ts
ebidel: How do we feel about updating yargs? Current is 6.6.0 :(

patrickhulce: What was the google3 dependency on yargs that kept us here, are they game to update?

lighthouse-cli/bin.ts

lighthouse-cli/bin.ts
ebidel: suspect that's it. Just complaining out loud that that restriction continues to hold us back.

lighthouse-core/gather/driver.js
```diff
@@ -472,16 +475,17 @@ class Driver {
...
-  gotoURL(url, options) {
-    const _options = options || {};
-    const waitForLoad = _options.waitForLoad || false;
-    const disableJS = _options.disableJavaScript || false;
-    const pauseAfterLoadMs = (_options.flags && _options.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
+  gotoURL(url, options = {}) {
+    const waitForLoad = options.waitForLoad || false;
+    const disableJS = options.disableJavaScript || false;
+    const pauseAfterLoadMs = (options.flags && options.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
+    const maxWaitMs = (options.flags && options.flags.maxWaitForLoad) ||
```
ebidel: should we support `--max-wait-for-load = 0`?

readme.md
```diff
@@ -149,6 +149,11 @@ Configuration:
   --list-trace-categories       Prints a list of all required trace categories and exits   [boolean]
   --config-path                 The path to the config JSON.
   --perf                        Use a performance-test-only configuration                  [boolean]
+  --port                        The port to use for the debugging protocol. Use 0 for a
+                                random port.                                         [default: 9222]
+  --max-wait-for-load           The timeout in milliseconds to wait for the page to
```
ebidel: you'll want to update this description too to match the CLI help

lighthouse-core/gather/driver.js
patrickhulce: I would say that would lead to a useless lighthouse run that wouldn't return any results so filling in default is fine with me :)

lighthouse-cli/bin.ts
paulirish: Let's bump yargs from our 3.30.0 to 3.32.0.
It has the fix for these: https://github.com/yargs/yargs/commit/51c0063183486193c92117d2bdcc61a67387f3d4

patrickhulce: done, from the looks of it google3 is still on lighthouse ~1.1 so we likely will have quite a few other issues to iron out when they want to upgrade anyway...

denar90:
Thanks guys 🎉

PR #1673, opened 2017-02-09
Audit: DOM stats (total nodes, depth, width)

### LINKED ISSUE #1096: DBW: Page uses too many DOM elements
Profiling Google Apps, they use < 1000 DOM nodes. Too many can hurt performance. It's especially a problem for users of web components where many custom elements are deep within shadow roots.

Let's get a baseline of how many is too many and create an audit for it.

from @esprehn

ebidel:
R: all

Fixes #1096.

- The audit is now scored according to the distribution linked to in the source.
- Adds a "cards formatter" to present hero stat info. If we like the direction, there are some other stat candidates in the report we can move over. I played with hiding the cards behind a `<details>`, but it didn't feel right. It's kinda nice to see the cards to break up our walls of text. More grey, but they don't have to be...?!
- Smoke tests, formatter tests, and audit tests. The whole shebang.

Here's what she looks like now:

screen shot 2017-02-15 at 5 12 31 pm

As seen in the screenshot, there's a `title` attribute for the last two cards that gives users a selector snippet on mouse over. It  give context of the user's DOM tree.


ebidel:
> swaggy HTTP archive stats here 

I'm selfishly adding this audit to have more places to show httparchive data :)

lighthouse-core/gather/gatherers/dobetterweb/domnodes.js
```diff
@@ -0,0 +1,36 @@
...
+
+const Gatherer = require('../gatherer');
+
+class AllDOMNodes extends Gatherer {
+  /**
+   * @param {!Object} options
+   * @return {!Promise<!Array<!Object>>}
+   */
+  afterPass(options) {
+    return options.driver.querySelectorAll('body, body /deep/ *');
```
ebidel: It was in a comment I re-posted from an email thread:
https://github.com/GoogleChrome/lighthouse/issues/1096#issuecomment-264347889

Would really love more hard data on the topic, but I do trust whatever Elliot says. I benchmarked a few big apps in the past and they tend to hover around ~1500-2000 nodes (GMail, maps, inbox, github).

lighthouse-core/audits/dobetterweb/num-dom-nodes.js
```diff
@@ -0,0 +1,66 @@
...
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'num-dom-nodes',
+      description: 'Uses a small number of DOM nodes',
+      helpText: 'Browser engineers recommend pages contain fewer than ' +
+          `~${NumDOMNodes.MAX_DOM_NODES} DOM nodes. The sweet spot is around 60 elements wide x ` +
+          '32 elements deep. A large DOM can increase memory, cause longer ' +
+          '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
+          'and produce costly [layout reflows](https://developers.google.com/speed/articles/reflow). [Learn more](hhttps://developers.google.com/web/fundamentals/performance/rendering/).',
```
ebidel: done

lighthouse-core/gather/gatherers/dobetterweb/domnodes.js
esprehn: Gmail is 4000-5000 elements (depending on the view, sometimes it's 15k on long email threads). This github review page is ~1500. My facebook timeline is 2300 when loaded. I got my numbers sampling a bunch of popular websites.

btw the /deep/ trick won't work with Shadow DOM v1 because I don't think deep will match in there. You'll need to write your own traversal that crawls the entire document. It's unfortunate we don't have any tooling for querySelectorAll with Shadow DOM v1 yet. @hayatoito

ebidel: Here's the crbug: https://bugs.chromium.org/p/chromium/issues/detail?id=633007#c16

Looks like `>>>` in `qSA` is behind the experimental web platform flag.  We can probably move to something like https://gist.github.com/ebidel/fc1302d5fa1ef7c6d42fe8189acc3820 this cover all base.

DevTools also has https://chromedevtools.github.io/debugger-protocol-viewer/tot/DOM/#method-getFlattenedDocument, which might be even better.

ebidel: @patrickhulce we also have the same query here:
https://github.com/GoogleChrome/lighthouse/blob/c6aeb33855456e160474421d544c2477719184ae/lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js#L133

We could get trickier and create a shared driver method that returns a promise and caches the result after the first call. Thoughts?

lighthouse-core/audits/dobetterweb/num-dom-nodes.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * @param {!Artifacts} artifacts
+   * @return {!AuditResult}
+   */
+  static audit(artifacts) {
+    const nodes = artifacts.AllDOMNodes;
+
+    const rawValue = nodes.length <= NumDOMNodes.MAX_DOM_NODES;
```
brendankenny: it would be great to score these like we do the [FMP/TTI/etc audits](https://github.com/GoogleChrome/lighthouse/blob/3043cc3732bd6b9cfbc02d96579a6f21c86de08b/lighthouse-core/audits/first-meaningful-paint.js#L110-L117), so it isn't just "you passed this line and are now failing" and more like, "you are way out of the mainstream/what browsers can handle well, so you score a 25" or whatever.

It would also be nice to be able to quantify the harm done by the amount of DOM content, so it's less a specific number and more about how much harm you're doing yourself, but I don't know if that's possible...

lighthouse-core/audits/dobetterweb/num-dom-nodes.js
```diff
@@ -0,0 +1,66 @@
...
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'num-dom-nodes',
+      description: 'Uses a small number of DOM nodes',
+      helpText: 'Browser engineers recommend pages contain fewer than ' +
+          `~${NumDOMNodes.MAX_DOM_NODES} DOM nodes. The sweet spot is around 60 elements wide x ` +
+          '32 elements deep. A large DOM can increase memory, cause longer ' +
```
brendankenny: should we be looking at length of the longest paths, or distribution of path lengths in the DOM?

lighthouse-core/gather/gatherers/dobetterweb/domnodes.js
brendankenny: agree with @patrickhulce that this may be a really big thing to bring back over and keep around just for the length. It doesn't look like the array contents would be particularly helpful for others audits, just an array of node IDs that would then each be wrapped by [Elements](https://github.com/GoogleChrome/lighthouse/blob/3043cc3732bd6b9cfbc02d96579a6f21c86de08b/lighthouse-core/lib/element.js), but we wouldn't be able to tell them apart in any way without more protocol requests for each one...

Particularly if you're going to have to get tricky with a selector for Shadow DOM v1, maybe switch to `evaluateAsync` and ship back only the number of nodes?

brendankenny: > We could get trickier and create a shared driver method that returns a promise and caches the result after the first call. Thoughts?

We probably should measure, I guess :) Maybe it's not so bad.

Beyond perf, though, for this gatherer there doesn't seem to be any benefit to keeping the array around

lighthouse-core/audits/dobetterweb/num-dom-nodes.js
ebidel: Agreed. Can you point me to that graphing site?

brendankenny: yeah, sorry, we put the parameters for the curve at the top of each of the audits that use them, e.g. https://github.com/GoogleChrome/lighthouse/blob/3043cc3732bd6b9cfbc02d96579a6f21c86de08b/lighthouse-core/audits/first-meaningful-paint.js#L24-L27

you can edit the graph and then save it, which will give a new permalink for that particular graph to add as a comment. You have to create an account to save, but it can login via github :)

ebidel:
Giving this a big time beef up.

lighthouse-core/gather/gatherers/dobetterweb/domnodes.js
patrickhulce: Isn't Pavel adding a method specifically to stop us from doing that? :) Haha, if its just a few thousand node ids I'm not that concerned.

ebidel: `DOM.getFlattenedDocument`? I did a quick dirty benchmark between the two and found `querySelectorAll` to be faster:

Runs: 1337ms, 1033ms
```
  console.time('took');
  return options.driver.sendCommand('DOM.enable')
    .then(_ => options.driver.sendCommand('DOM.getFlattenedDocument', {
      depth: -1, pierce: true
    }))
    .then(result => {
      const nodes = result.nodes.filter(node => node.nodeType === 1);
      console.timeEnd('took');
      return nodes;
    });
```
 
Runs: 812ms, 425ms

```
    console.time('took');
    return options.driver.querySelectorAll('html, html /deep/ *')
      .then(nodes => {
        console.timeEnd('took');
        return nodes;
      });
```

IMO, `DOM.getFlattenedDocument` is a bit clunky to use. You have to filter for element nodes, but the biggest drawback is that `depth: -1, pierce: true` also traverses into iframes. We definitely don't want that.

patrickhulce: Yeah I'm fine with it here, just making a comment on why sharing the code may not be a great reason to justify fetching the nodes instead of length (if the other path that's using it will start using one that doesn't need to go and fetch information for each of the items individually *after* this call finishes :) )

ebidel:
PTAL. The audit UI has changed quite a bit, including a new formatter, and now being a scored audit. Take a look at the [description](https://github.com/GoogleChrome/lighthouse/pull/1673#issue-206386633).

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -0,0 +1,136 @@
...
+function elementPathInDOM(element) {
+  const path = [createSelectorsLabel(element)];
+  let node = element;
+  while (node) {
+    node = node.parentNode && node.parentNode.host ? node.parentNode.host : node.parentElement;
+    if (node) {
+      path.push(createSelectorsLabel(node));
+    }
+  }
+  return path.reverse();
```
patrickhulce: I don't particularly mind which one, but any gotchas with `unshift` instead?

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -0,0 +1,136 @@
...
+  };
+}
+
+class DOMStats extends Gatherer {
+  /**
+   * @param {!Object} options
+   * @return {!Promise<!Array<!Object>>}
+   */
+  afterPass(options) {
+    return options.driver.querySelectorAll('body, body /deep/ *')
```
patrickhulce: do we still need to do this if we're already calling evaluateAsync? could probably just return in `getDOMStats` without shipping all the nodes.

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -149,6 +149,36 @@ module.exports = [
       }
     }
   }, {
+    initialUrl: 'http://localhost:10200/dobetterweb/domtester.html?smallDOM',
```
patrickhulce: yay for moving out into some new testers! 🎉

lighthouse-core/formatters/partials/cards.html
```diff
@@ -0,0 +1,8 @@
+<ul class="subitem__details cards__container">
+  {{#each this}}
+    <div class="subitem__detail scorecard" {{#if snippet}}title="{{snippet}}"{{/if}}>
```
ebidel: I reserved `title` for the card title. Also wanted to make it obvious this was a code `snippet`.

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -0,0 +1,136 @@
...
+/* istanbul ignore next */
+function createSelectorsLabel(element) {
+  let name = element.localName;
+  const idAttr = element.getAttribute && element.getAttribute('id');
+  if (idAttr) {
+    name += `#${idAttr}`;
+  }
+  const className = element.className;
+  if (className) {
+    name += `.${className.split(' ').join('.')}`;
```
ebidel: done

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
ebidel: that works. done.

lighthouse-core/formatters/partials/cards.html
patrickhulce: oh whoops, alright

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
ebidel: good call. done.

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -0,0 +1,112 @@
...
+        SCORING_MEDIAN, SCORING_POINT_OF_DIMINISHING_RETURNS);
+    let score = 100 * distribution.computeComplementaryPercentile(stats.totalDOMNodes);
+
+    // Clamp the score to 0 <= x <= 100.
+    score = Math.max(0, Math.min(100, score));
+
+    const cards = [
+      {title: 'Total DOM Nodes', value: stats.totalDOMNodes.toLocaleString()},
+      {title: 'Max DOM Depth', value: stats.depth.max.toLocaleString(), snippet: depthSnippet},
+      {title: 'Max Children / Node', value: stats.width.max.toLocaleString(), snippet: widthSnippet}
```
ebidel: "Maximum Children". Done

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -0,0 +1,112 @@
...
+    const distribution = TracingProcessor.getLogNormalDistribution(
+        SCORING_MEDIAN, SCORING_POINT_OF_DIMINISHING_RETURNS);
+    let score = 100 * distribution.computeComplementaryPercentile(stats.totalDOMNodes);
+
+    // Clamp the score to 0 <= x <= 100.
+    score = Math.max(0, Math.min(100, score));
+
+    const cards = [
+      {title: 'Total DOM Nodes', value: stats.totalDOMNodes.toLocaleString()},
+      {title: 'Max DOM Depth', value: stats.depth.max.toLocaleString(), snippet: depthSnippet},
```
ebidel: done

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -0,0 +1,132 @@
...
+const Gatherer = require('../gatherer');
+
+/**
+ * Constructs a pretty label from element's selectors. For example, given
+ * <div id="myid" class="myclass">, returns 'div#myid.myclass'.
+ * @param {!HTMLElement} element
+ * @return {!string}
+ */
+/* istanbul ignore next */
+function createSelectorsLabel(element) {
```
ebidel: It does. Here's what it looks like:

screen shot 2017-02-16 at 6 30 06 pm

ebidel:
PTAL

PR #1675, opened 2017-02-09
feat: support skipping cache and storage clear

### LINKED ISSUE #1664: WPT compat: support cache warm run
WPT performs a cache cold run and a cache warm run while LH always clears all storage for the origin and disables the network cache.

patrickhulce:
fixes #1664

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -65,6 +65,7 @@ const cliFlags = yargs
     'port'
   ], 'Configuration:')
   .describe({
+    'disable-cache-clearing': 'Disable clearing of cache and storage before the run',
     'disable-device-emulation': 'Disable Nexus 5X emulation',
     'disable-cpu-throttling': 'Disable CPU throttling',
     'disable-network-throttling': 'Disable network throttling',
@@ -92,6 +93,7 @@ Example: --output-path=./lighthouse-results.html`
 
   // boolean values
   .boolean([
+    'disable-cache-clearing',
     'disable-device-emulation',
     'disable-cpu-throttling',
     'disable-network-throttling',
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -65,7 +65,7 @@ const cliFlags = yargs
     'port'
   ], 'Configuration:')
   .describe({
-    'disable-cache-clearing': 'Disable clearing of cache and storage before the run',
+    'disable-storage-reset': 'Disable clearing of cache and storage before the run',
     'disable-device-emulation': 'Disable Nexus 5X emulation',
     'disable-cpu-throttling': 'Disable CPU throttling',
     'disable-network-throttling': 'Disable network throttling',
@@ -93,7 +93,7 @@ Example: --output-path=./lighthouse-results.html`
 
   // boolean values
   .boolean([
-    'disable-cache-clearing',
+    'disable-storage-reset',
     'disable-device-emulation',
     'disable-cpu-throttling',
     'disable-network-throttling',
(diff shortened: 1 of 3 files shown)


lighthouse-cli/bin.ts
```diff
@@ -65,6 +65,7 @@ const cliFlags = yargs
     'port'
   ], 'Configuration:')
   .describe({
+    'disable-cache-clearing': 'Disable clearing of cache and storage before the run',
```
patrickhulce: other name ideas welcome, cache is more specific than what actually happens but is probably more familiar/understandable to users than a generic `storage`

ebidel: `disable-storage-reset` comes to mind. I actually think being more generic than "cache" is good for the reasons you mention. It applies to all forms of storage on the origin.

patrickhulce: will do

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -225,6 +225,60 @@ describe('GatherRunner', function() {
...
+      blockUrlPatterns: asyncFunc,
+    };
+
+    return GatherRunner.setupDriver(driver, {flags: {}}).then(_ => {
+      assert.equal(tests.calledDisableNetworkCache, true);
+      assert.equal(tests.calledClearStorage, true);
+    });
+  });
+
+  it('does not clear the network cache and origin storage when flag is set', () => {
```
ebidel: ...when disable-storage-reset flag is set

lighthouse-cli/bin.ts
```diff
@@ -65,6 +65,7 @@ const cliFlags = yargs
     'port'
   ], 'Configuration:')
   .describe({
+    'disable-storage-reset': 'Disable clearing of cache and storage before the run',
```
ebidel: nit: "Disable clearing the browser cache and other storage APIs before a run"

lighthouse-core/gather/gather-runner.js
```diff
@@ -91,13 +91,14 @@ class GatherRunner {
 
   static setupDriver(driver, options) {
     log.log('status', 'Initializing…');
+    const resetStorage = !(options.flags && options.flags.disableStorageReset);
```
brendankenny: is `options.flags` ever not defined here? If so, the `options.flags.blockedUrlPatterns` line below is in trouble

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -225,6 +225,60 @@ describe('GatherRunner', function() {
     });
   });
 
+  it('clears the network cache and origin storage', () => {
+    const asyncFunc = () => Promise.resolve();
+    const tests = {
+      calledDisableNetworkCache: false,
+      calledClearStorage: false,
+    };
+    const createCheck = variable => () => {
```
brendankenny: it would be nice to replace fake-driver with a nice interface around something like this :)

patrickhulce: or we could use `sinon` :)

lighthouse-core/gather/gather-runner.js
patrickhulce: oh yeah it probably always is, just was jumping around to other places where that wasn't true. i'll update

lighthouse-cli/bin.ts

lighthouse-core/test/gather/gather-runner-test.js
patrickhulce: done

lighthouse-core/test/gather/gather-runner-test.js
brendankenny: yeah, I'd be fine with that too.

Honestly I hate these tests involving the driver. Rely too much on knowledge of internal implementation instead of externally observable behavior. If we're going to assume the driver calls made and the order they're made in, I'd rather we just setup a debugger protocol record/replay system :)

PR #1677, opened 2017-02-10
ignore URL fragment when asserting page loaded

brendankenny:
came up in testing 1.5.0 for release. Runs fail testing e.g. `https://example.com/#/whatever`

Testing a URL with hash content will always fail the `GatherRunner.assertPageLoaded()` check so this uses `URL.equalWithExcludedFragments` instead of strict equality testing.

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -19,6 +19,7 @@
 const log = require('../lib/log.js');
 const Audit = require('../audits/audit');
 const path = require('path');
+const URL = require('../lib/url-shim');
 
 /**
  * Class that drives browser to load the page and runs gatherer lifecycle hooks.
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -135,7 +136,9 @@ class GatherRunner {
    * @param {!Array<WebInspector.NetworkRequest>} networkRecords
    */
   static assertPageLoaded(url, driver, networkRecords) {
-    const mainRecord = networkRecords.find(record => record.url === url);
+    const mainRecord = networkRecords.find(record => {
```
ebidel: maybe add a comment here that the network record never sees the hash so we need to test the URL without it. Obvious>

brendankenny: done

PR #1678, opened 2017-02-10
fix: table formatter image preview URL encode

patrickhulce:
R: @ebidel all

Image previews were broken for URLs with `)` since markdown can't handle them

Before
image

After
image

### DIFF
--- a/lighthouse-core/formatters/table.js
+++ b/lighthouse-core/formatters/table.js
@@ -94,7 +94,9 @@ class Table extends Formatter {
         switch (key) {
           case 'preview':
             if (/^image/.test(value.mimeType)) {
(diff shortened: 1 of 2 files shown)


brendankenny:
upstream issue: https://github.com/chjj/marked/issues/619 (looks like it won't be fixed)

patrickhulce:
FWIW in defense of marked I think it's actually an artifact of the markdown spec itself. There's no guarantee in an arbitrary link that an opening paren will be followed by a close paren.

Also relevant http://meta.stackexchange.com/questions/13501/links-to-urls-containing-parentheses

ebidel:
🚫 🖼

brendankenny:
> the markdown spec itself

"spec" :)

PR #1679, opened 2017-02-10
Geolocation audit: warn if page is non-secure origin

### LINKED ISSUE #1676: false denied state on GeolocationOnStart gatherer on some sites
Haven't looked into this yet, so it may be something obvious. Discovered when testing `http://angularexpo.com/`, in the CLI or extension. I get:

`Required GeolocationOnStart gatherer encountered an error: Unable to determine if this permission was requested on page load because it had already been set. Try resetting the permission and running Lighthouse again.`

Checking the state, it's seeing the permission as `denied` even though this is a new blank Chrome profile.

Interestingly the `NotificationOnStart` gatherer has no issue with this page, even though it's querying the permission state in the same way.

ebidel:
R: all

Fixes #1676

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/geolocation-on-start.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/geolocation-on-start.js
@@ -37,12 +37,20 @@ class GeolocationOnStart extends Gatherer {
    * @return {!Promise<!Array<{url: string, line: number, col: number}>>}
    */
   afterPass(options) {
-    return options.driver.queryPermissionState('geolocation')
+    return options.driver.evaluateAsync('(function(){return window.isSecureContext;})()')
+      .then(isSecureContext => {
+        if (!isSecureContext) {
+          throw new Error('Unable to determine if the Geolocation permission requested on page ' +
+              'load because the page is not hosted on a secure origin. The Geolocation API ' +
+              'requires an https URL.');
+        }
+      })
+      .then(_ => options.driver.queryPermissionState('geolocation'))
       .then(state => {
         if (state === 'granted' || state === 'denied') {
-          throw new Error('Unable to determine if this permission was requested on page load ' +
-              'because it had already been set. Try resetting the permission and running ' +
-              'Lighthouse again.');
+          throw new Error('Unable to determine if the Geolocation permission was ' +
+              `requested on page load because it was already ${state}. ` +
+              'Try resetting the permission and running Lighthouse again.');
         }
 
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/gatherers/dobetterweb/geolocation-on-start.js
```diff
@@ -37,12 +37,20 @@ class GeolocationOnStart extends Gatherer {
...
+          throw new Error('Unable to determine if the Geolocation permission requested on page ' +
+              'load because the page is not hosted on a secure origin. The Geolocation API ' +
+              'requires an https URL.');
+        }
+      })
+      .then(_ => options.driver.queryPermissionState('geolocation'))
+      .then(permissionState => {
+        if (permissionState === 'granted' || permissionState === 'denied') {
+          throw new Error('Unable to determine if the Geolocation permission was ' +
+              'requested on page load because it was already granted/denied. ' +
```
brendankenny: since we have the `permissionState` and we know it was either "granted" or "denied", should we just inject it into the string instead of saying "granted/denied"?

lighthouse-core/gather/gatherers/dobetterweb/geolocation-on-start.js
```diff
@@ -37,12 +37,20 @@ class GeolocationOnStart extends Gatherer {
    * @return {!Promise<!Array<{url: string, line: number, col: number}>>}
    */
   afterPass(options) {
-    return options.driver.queryPermissionState('geolocation')
-      .then(state => {
-        if (state === 'granted' || state === 'denied') {
-          throw new Error('Unable to determine if this permission was requested on page load ' +
-              'because it had already been set. Try resetting the permission and running ' +
-              'Lighthouse again.');
+    return options.driver.evaluateAsync('(function(){return window.isSecureContext;})()')
```
ebidel: Note: `isSecureContext` returns true for `https://` and `http://localhost`. The latter is a secure context for testing purposes. We may want to consider moving the HTTPs gatherer to this or supplementing it with this check for https://github.com/GoogleChrome/lighthouse/issues/1175.

lighthouse-core/gather/gatherers/dobetterweb/geolocation-on-start.js
ebidel: +1

ebidel: done

PR #1681, opened 2017-02-10
ignore core and viewer tests from npm module

brendankenny:
leaving in `lighthouse-cli/test` for now as it's relatively small (200KB, fixtures and all) and smokehouse is really convenient for testing that the npm module works before publishing

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -30,4 +30,5 @@ lighthouse-cli/types/*.map
 assets
 
 lighthouse-core/test
-lighthouse-viewer/test
+lighthouse-viewer
+lighthouse-extension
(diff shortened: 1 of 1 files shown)


ebidel:
Just leave out all of viewer? How about all of the extension too?

brendankenny:
> Just leave out all of viewer? How about all of the extension too?

I can test that. Is there any reason someone installing lighthouse globally might want to use viewer? I can't really think of one

ebidel:
Naw don't think so. They can always check out the source of they want those.

brendankenny:
done

PR #1686, opened 2017-02-10
report: Tweak colors so that we are WCAG2AA valid.

XhmikosR:
**before**

1-before

**after**

2-after

Now the report should be WCAG2AA valid.

### DIFF
--- a/lighthouse-core/formatters/partials/critical-request-chains.css
+++ b/lighthouse-core/formatters/partials/critical-request-chains.css
@@ -51,11 +51,11 @@
 }
 
 .cnc-node__tree-hostname {
-  color: #999;
+  color: #767676;
 }
 
 .initial-nav {
-  color: #a9a9a9;
+  color: #767676;
   margin: 10px 0 0;
   font-style: italic;
 }
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -29,10 +29,10 @@ span, div, p, section, header, h1, h2, li, ul {
   --text-color: #222;
   --secondary-text-color: #606060;
   --accent-color: #3879d9;
-  --poor-color: #eb211e;
-  --good-color: #1ac123;
-  --average-color: #ffae00;
-  --warning-color: #f6be00;
+  --poor-color: #e81e1b;
+  --good-color: #008a07;
+  --average-color: #9f6d00;
+  --warning-color: #947200;
   --gutter-gap: 12px;
   --gutter-width: 40px;
   --body-font-size: 14px;
@@ -273,8 +273,7 @@ body {
 }
 
 .menu__header-version {
-  opacity: 0.4;
-  color: #fff;
+  color: #aab3ed;
   font-family: var(--text-font-family);
   font-size: 14px;
   line-height: 1.5;
@@ -326,7 +325,7 @@ body {
 .menu__link {
   padding: 0 20px;
   text-decoration: none;
-  color: #777;
+  color: #767676;
   display: flex;
 }
 
@@ -613,7 +612,7 @@ body {
   text-align: center;
   font-size: 12px;
   border-top: 1px solid #ebebeb;
-  color: #999;
+  color: #767676;
 }
 
 .devtabs {


XhmikosR:
The red was fine, I'll check the rest when I'm back.

XhmikosR:
@ebidel: the other colors had issues too with WCAG2AA. Please verify because I only tested this in theory wince I don;t have URLs that use `--accent-color`, `--warning-color` and `--average-color`.

XhmikosR:
Especially warning and average; not sure their difference is clear enough or not.

XhmikosR:
Fixed one more issue with poor color. Now the patch should be final.

paulirish:
@XhmikosR this is a bit of a larger question, but... how about HSL? 
personally i find it massively easier to interpret at a glance. plus adjustments of lighter/darker are obviously more straightforward.

not sure how others feel tho. @ebidel

XhmikosR:
Personally I haven't had the need to use HSL yet, but from a quick look it does seem easier to interpret indeed.

This PR just makes sure the report is WCAG2AA valid, nothing more or less :)

paulirish:
@XhmikosR gotcha, sounds good. in that case, let me defer to @robdodson on this stuff

brendankenny:
Is this for color contrast? Sounds like a joke but isn't: maybe in CI we should run lighthouse then do a smokehouse run against the generated report. That would let us assert color contrast, etc

XhmikosR:
Yeah this is for color contrast in order to comply with WCAG2AA. I didn't go for WGAG2AAA because I thought "one step at a time".

XhmikosR:
IMO we should also run HTML validation on CI because in the past it has caught important things, see #1575.

robdodson:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

robdodson:
small nit: the version number under the word "Lighthouse" is still low contrast. I think if you set it to `#aab3ed` and remove the 0.4 `opacity` then it would meet WCAG2AA

XhmikosR:
It is valid in this branch since I've already changed it. How did you check
it?

On Feb 13, 2017 19:09, "Rob Dodson" <notifications@github.com> wrote:

small nit: the version number under the word "Lighthouse" is still low
contrast. I think if you set it to #aab3ed and remove the 0.4 opacity then
it would meet WCAG2AA

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleChrome/lighthouse/pull/1686#issuecomment-279456131>,
or mute the thread

XhmikosR:
Can someone report this guy? Doesn't seem right that one can randomly spam
PRs like that :/

ebidel:
@XhmikosR we've taken care of it.

XhmikosR:
Doesn't seem so :p

On Feb 13, 2017 19:40, "Eric Bidelman" <notifications@github.com> wrote:

@XhmikosR <https://github.com/XhmikosR> we've taken care of it.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleChrome/lighthouse/pull/1686#issuecomment-279464366>,
or mute the thread

paulirish:
just kidding but really _now_ it should be taken care of.

ebidel:
I blame the black header.

XhmikosR:
Hehe.

I reported him. Alternatively I guess you could lock the PRs but I won't be able to reply (I think).

XhmikosR:
Now about @robdodson's comment above, https://github.com/GoogleChrome/lighthouse/pull/1686/files#diff-7c4405f5b4874ecc329c5331597912e1R616

There's the change to the footer's color.

robdodson:
Sorry, I was referring to this version number in the top left

screen shot 2017-02-14 at 8 42 56 am

XhmikosR:
@robdodson: I see now. But still, I don't get any errors for that chunk at all.

I can change it to `#aab3ed` and remove `opacity` if you prefer.

robdodson:
I think I was testing in aXe and the chrome devtools extension. What're u
using to test?

XhmikosR:
I'm using http://squizlabs.github.io/HTML_CodeSniffer/.

No big deal, your change makes also sense. Just give me the final OK and I'll include it in this branch,

XhmikosR:
In fact your solution won't pass WCAG2AAA if there are plans to do so in the future.

XhmikosR:
I can make changes in another PR so that we are WCAG2AAA valid. The current changes are for WCAG2AA.

robdodson:
WCAG2AA should be fine. If you want to make the change I can sign off on the PR

XhmikosR:
I'd say better merge this for now and we can revisit it later. There are a
few improvements we could make later.

XhmikosR:
On a side note @paulirish we hit this failure on Windows quite often. Not sure what's causing but it seems the process can't be terminated for some reason.

Otherwise this should be merged so that we can move forward with more changes.

robdodson:
The version number being used is pretty important information. I'd like to still change it as mentioned in my prior comment before merging this PR

XhmikosR:
All right I'll change it. Just keep in mind it won't pass 3a and we'll need
to change it again then. Also I personally don't see it failing.

robdodson:
I used the color picker and Lea Verou's contrast tool to confirm it's failing
https://leaverou.github.io/contrast-ratio/#%237E87D1-on-%232935B3

We don't need to shoot for WCAG2AAA, my understanding is WCAG2AA is the generally accepted standard and AAA is for advanced cases when you know your audience might contain a larger percentage of folks who need very high contrast for example.

XhmikosR:
Rebased and also fixed one new issue with `.initial-nav`.

PR #1687, opened 2017-02-10
Hide disable extension setting

ebidel:
R: @brendankenny 

Didn't remove code or html for now so we can test if this setting is impactful.

### DIFF
--- a/lighthouse-extension/app/popup.html
+++ b/lighthouse-extension/app/popup.html
@@ -57,10 +57,10 @@ limitations under the License.
   </aside>
 
   <aside class="options subpage">
-    <h2 class="options__title">Settings</h2>
-    <div>
+    <h2 class="options__title" hidden>Settings</h2>
+    <div hidden>
       <label>
-        <input type="checkbox" class="setting-disable-extensions"
+        <input type="checkbox" class="setting-disable-extensions" disabled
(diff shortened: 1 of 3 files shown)


lighthouse-extension/app/popup.html
```diff
@@ -57,10 +57,10 @@ <h2 class="header-titles__url">...</h2>
   </aside>
 
   <aside class="options subpage">
-    <h2 class="options__title">Settings</h2>
-    <div>
+    <h2 class="options__title" hidden>Settings</h2>
+    <div hidden>
       <label>
-        <input type="checkbox" class="setting-disable-extensions"
+        <input type="checkbox" class="setting-disable-extensions" disabled
```
ebidel: for good measure

brendankenny:
does more need to be disabled in `lighthouse-background.js`? e.g. the part that gets all the current extensions run with or without the checkbox checked. 

maybe comment out the part that reads/writes `disableExtensionsDuringRun` from storage? mostly for people that may have already enabled, but also to be double extra sure :)

brendankenny:
LGTM and runs well. Do you want to file an issue to track testing of this?

ebidel:
https://github.com/GoogleChrome/lighthouse/issues/1689

PR #1692, opened 2017-02-11
Add trailing slash to test URLs.

XhmikosR:


brendankenny:
diversity in trailing slashes was actually intentional for the tests, but I guess that's not self-evident :)

XhmikosR:
I thought it was a mistake because the server does respond with the trailing slashes.

No big deal, it's mostly cosmetic, I just noticed it.

brendankenny:
yeah, the idea was that people aren't great at including it or not, so we should make sure we handle it.

Good in an integration test because while all (or most) of the places we compare URLs should get stuff like this unit tested, the interaction between where the test URL is canonicalized, where redirects are tracked, where we match network records, etc etc might have some weird interaction (also why the `finalUrl` is asserted too, even though it means we're subject to our test sites restructuring even slightly someday)

Issue #1158, opened 2016-12-14
RFC: V8 Performance Insights


addyosmani:
Lighthouse currently exposes a Performance Metrics section that includes items like Time-to-interactive. This is great, but what if we helped you understand _why_ you weren't doing great at some of these metrics? With attribution that ties back to your source.

A poor TTI may be due to scripts keeping the main thread busy. Some factors that contribute to this are script specific:

**JS bundle sizes**: Large, expensive JS bundles can proportionately take longer to compile and run. They'll almost certainly peg the main thread longer depending on the work being done. Profiling a few thousand sites, I've noticed the average React or Angular app ships more than 700KB+ of script while those generally hitting a pleasing TTI are shipping < 300KB of script. Similar to how we have a notion of [critical request chains](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/computed/critical-request-chains.js), we could highlight total JS size as a potential bottleneck 🔆.

Depending on the UX for this, we might also be able to hint at using techniques like code-splitting to break up those larger scripts and only ship down what is useful for a route. 

Another iteration on this could break down large scripts based on origin, so it's clear to see when large bundle sizes are an origin level concern vs something a third-party is pulling into your page. I've seen a few otherwise "okay" sites having a party on the main thread 🎉 because an ad was loading in all of Angular.js. It'd be good to highlight where the issue is your app/framework code vs things slightly out of your control.

Size alone is just perspective on this problem however.

**Parse/compile costs**: There's a need for more devs to understand parse/compile can be a real bottleneck. Many of the traces I look at have folks spending 4-5s in just this phase, often pushing out how soon a page is sufficiently booted up to be useful. Sites that are interactive within our current public budgets (<5s) typically spend < 3500ms in parse/compile (before trying to progressively load in anything else). Most of the Chrome loading team's mobile URL sets spend < 2.5s in parse/compile so I feel like starting with a 2.5-3.5s budget for this and tweaking over time might make sense.

Practically, we could introduce a notion of _long scripts_, highlighting the top 5-10 scripts on your page that are parse/compile heavy so you know where time is being spent ⌚️ 

We know that V8 spends a lot of time in the parse phase and there may be something interesting we can do here with runtime callstats to expose causes of slowness.

![screen shot 2016-12-14 at 8 11 19 am](https://cloud.githubusercontent.com/assets/110953/21189903/f1ddec24-c1d4-11e6-9995-10335b3e2895.jpg)

[...]

PR #1695, opened 2017-02-11
feat: Page execution timings (don't merge)

addyosmani:
This initial PR adds support for page execution timings to the report. The ultimate goal of the feature is to inform developers where time is being spent during load and could be augmented to just show this before they are interactive.

Preview (extension):
screen shot 2017-02-10 at 7 12 42 pm

CLI:
screen shot 2017-02-10 at 6 56 34 pm

**A few directions we could take this:**

- Improve approach to grouping. @paulirish shared this [nice](https://docs.google.com/spreadsheets/d/1jqbSfdkB5RMtuifJwQsGb70WUUhWvzir7hq41wYHXsU/edit#gid=0) categorization doc for Timeline events that could be used.
- Add support for V8 Runtime Call Stats (to address #1158). This could come later. 
- Eventually, I'd like to introduce page execution 'budgets' (e.g your Parse and Compile times are wayyyy too long. Read this doc on how to trim em' down) 
- Eventually distinguish between lifetime of page load and work done between FMP and TTI

### DIFF
--- a/lighthouse-core/audits/page-execution-timings.js
+++ b/lighthouse-core/audits/page-execution-timings.js
@@ -1,0 +1,120 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const Audit = require('./audit');
+const Formatter = require('../formatters/formatter');
+const TimelineModel = require('../lib/traces/devtools-timeline-model');
+
+function dumpTree(tree, timeValue) {
+  const result = new Map();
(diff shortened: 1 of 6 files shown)


tdresser:
We should probably be reusing code from: https://github.com/catapult-project/catapult/blob/master/tracing/tracing/extras/chrome/chrome_user_friendly_category_driver.html to do the categorization.

I'm not sure that we should do this in lighthouse, but I'm hoping that long term we'll have some UI surface which lets us expose this information to developers, split by whether the work was pre-TTI or post-TTI, and maybe even more precisely than that. I can imagine someone asking the question "why is my FMP so far after FCP?", and then we could give them a CPU time breakdown between those two points.

Does this use wall time or cpu time?

tdresser:
Ping re code sharing with https://github.com/catapult-project/catapult/blob/master/tracing/tracing/extras/chrome/chrome_user_friendly_category_driver.html.

addyosmani:
I was pretty sure I had commented on this a while back, but must have forgot to hit submit :)

I'm fully in support of reusing the code from https://github.com/catapult-project/catapult/blob/master/tracing/tracing/extras/chrome/chrome_user_friendly_category_driver.html. to do the categorization, @tdresser. I'll pick this work back up in the next week or two.

wardpeet:
would it be beneficial to show compile & execution time per script or do we just want this to manually trigger the user to go in devtools

benschwarz:
> would it be beneficial to show compile & execution time per script or do we just want this to manually trigger the user to go in devtools

It'd be great to have these stats on a per-script basis, as well as what @addyosmani has prepared so far. 

I'm going to push this around today and see where its at. IMO, we should get what we have shipped then start to look at enhancements thereafter.

paulirish:
As requested, not merging this. ;)

Closing in favor of ward's new PR: #3520

PR #1698, opened 2017-02-11
CSS consistency changes.

XhmikosR:
* use numeric values for `font-weight`
* stop using shorthands when one property is meant to change (`background`, `list-style`)
* use hex color values instead of rgb or color names when they are shorter
* use `outline: none` instead of zero
* remove duplicate property `border-radius`
* fix typo in comment
* minor HTML consistency changes

### DIFF
--- a/lighthouse-cli/performance-experiment/report/partials/config-panel.html
+++ b/lighthouse-cli/performance-experiment/report/partials/config-panel.html
@@ -1,7 +1,7 @@
 <style>
 .config-panel {
   padding: 0;
-  background: #fafafa;
+  background-color: #fafafa;
 }
 
 .config-panel__header {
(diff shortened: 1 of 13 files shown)


paulirish:
Do you have tooling that can help maintain this? I'm afraid it's going to be easy to regress these. :/

XhmikosR:
Unfortunately all the CSS tweaks I've done are made manually.

We could look into stylelint or something. But these changes are mostly cosmetic apart from not using shorthands.

XhmikosR:
Fixed one more issue, a duplicate `border-radius` property.

XhmikosR:
So any new thoughts about this? Things I should revert or add?

ebidel:
Agreed with Paul on the long-term maintenance of this stuff with out tools. That sad, this is lowkey. Can you rebase?

XhmikosR:
Done. I still don't get those random AppVeyor failures to kill the running instance but they seem to happen a lot.

PR #1701, opened 2017-02-12
Make ESLint rules stricter

XhmikosR:
I left the ones that need more changes as disabled so someone could toggle them easier in the future.

I only added the rules that aren't in eslint:recommended nor google config.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,23 +8,61 @@ module.exports = {
   },
   "rules": {
     // 2 == error, 1 == warning, 0 == off
+    "array-callback-return": 2,
+    "arrow-parens": [2, "as-needed"],
+    "block-scoped-var": 2,
+    "block-spacing": 2,
+    "dot-notation": 2,
+    "eqeqeq": 2,
     "indent": [2, 2, {
       "SwitchCase": 1,
       "VariableDeclarator": 2
+    }],
+    "keyword-spacing": [2, {
+      "before": true,
+      "after": true
     }],
     "max-len": [2, 100, {
       "ignoreComments": true,
       "ignoreUrls": true,
       "tabWidth": 2
     }],
+    "no-alert": 2,
+    "no-catch-shadow": 2,
+    "no-cond-assign": 2,
+    "no-confusing-arrow": [2, {
+      "allowParens": true
+    }],
+    "no-div-regex": 2,
+    "no-duplicate-imports": 2,
     "no-empty": [2, {
       "allowEmptyCatch": true
     }],
+    "no-eq-null": 2,
+    "no-eval": 2,
+    "no-extra-label": 2,
+    "no-floating-decimal": 2,
     "no-implicit-coercion": [2, {
       "boolean": false,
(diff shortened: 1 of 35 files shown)


brendankenny:
I'm personally a fan of stricter linting, but you may have an uphill battle on the necessity of most of these :)

I think some of these are good (`eqeqeq`, `dot-notation`, `no-useless-escape`), some we seem to already have (`array-bracket-spacing`, some of the disabled rules), and I'm definitely not a fan of a few -- `no-lonely-if`, `no-negated-condition` (sometimes more readable the other way), `no-path-concat` (unnecessary in cases like `fs.readFile`) stand out, and then it appears you edited code as if `no-undefined` was set to error, but didn't end up enabling it?

patrickhulce:
I'm a fan or indifferent on most of these, though I wouldn't mind seeing comma dangle become enforced for cleaner diffs :)

Why no on `no-lonely-if` @brendankenny not an `else if` fan?

brendankenny:
> Why no on no-lonely-if @brendankenny not an else if fan?

ha, no, often `else if` is better, but like with `no-negated-condition` sometimes it makes more sense to go the other way and having a rule for it is annoying whenever you run into it and seems like overkill. It's a stylistic thing that rarely or never would lead to an accidental bug.

The only real benefit (IMO, of course) would be as a Clippy reminder when refactoring so you don't miss where you could restructure your conditionals now that some part is removed or whatever.

paulirish:
I've personally been attracted to https://github.com/jlongster/prettier and interested in using that for style concerns. It's an automatic formatter that's extremely comprehensive. It's the only thing that handles style around long lines, optional wrapping, etc itself.

Going in this direction, I'd also adopt https://github.com/kentcdodds/prettier-eslint which passes prettify output on to eslint --fix to tackle any remaining things.

I'd prefer using prettier (and getting it run on file save) vs more eslint rules that I need to look up docs for.

XhmikosR:
@brendankenny: `array-bracket-spacing` fixed. It's quite possible I missed some since I based these changes on my own config and then removed the ones Google has. For the rest, it's your call, guys.

As you know, one can also disable the rule per block so, you know :P

XhmikosR:
@brendankenny: about `undefined`, yes I had it enabled but since you seem to do `foo.bar = undefined` in many places I had to disable it. Personally I'd like that rule enabled but I didn't want to change so many stuff.

XhmikosR:
The failure that happens in this branch doesn't seem to be caused by my changes since even if I revert the ESLint fixes, I still get the failures. No idea where that error comes from yet.

XhmikosR:
Opened #1817 for the test failure.

wardpeet:
@brendankenny @paulirish @patrickhulce 
any update what to do with this PR?

I'll give prettier a shot

paulirish:
Nothing seriously wrong with this patch but I think we don't have the bandwidth to debate some of the style details here.

Going to close. I have adopted prettier recently and find it useful for a lot of this stuff.. :)

PR #1704, opened 2017-02-12
Report: fix table overflow with Firefox.

XhmikosR:
**before**

**after**

### DIFF
--- a/lighthouse-core/formatters/partials/table.css
+++ b/lighthouse-core/formatters/partials/table.css
@@ -33,6 +33,7 @@
   font-family: monospace;
   display: block;
   margin: 0;
+  overflow-x: auto;
 }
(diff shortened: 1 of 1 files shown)


lighthouse-core/formatters/partials/table.css
```diff
@@ -33,6 +33,7 @@
   font-family: monospace;
   display: block;
   margin: 0;
+  overflow-y: hidden;
```
ebidel: `overflow-x: auto`. Tells the `pre` what to do (instead of telling it what not to do).
Would make me less nervous with tables, pre wrapping, and different browser compat in the mix. Clearly the latter is already an issue for FF.

XhmikosR: You are right. Changed it.

PR #1708, opened 2017-02-13
Create .codeclimate.yml

abbyarmada:
Abby from Code Climate here! There's more info in the email I've sent.

- Adds `eslint-3` channel to fix eslint errors.

Thanks!

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,30 @@
+engines:
+  csslint:
+    enabled: true
+  duplication:
+    enabled: true
+    config:
+      languages:
+      - ruby
+      - javascript
+      - python
+      - php
+  eslint:
+    enabled: true
+    channel: "eslint-3"
+  fixme:
+    enabled: true
+  rubocop:
+    enabled: true
+ratings:
+  paths:
+  - "**.css"
+  - "**.inc"
+  - "**.js"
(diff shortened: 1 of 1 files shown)


abbyarmada:
I signed it. ;)

paulirish:
**others**: For context, I was investigating if CodeClimate would give us much value and tried to run it against Lighthouse. It errored with a problem in eslint. I submitted a support email to report the bug. This PR is part of the response. :)

----

@abbyarmada I'm reading the docs at https://docs.codeclimate.com/docs/eslint and now see what this eslint-channel thing is all about. 
This opens up a larger conversation, but IMHO it's **rather unfortunate to have to define the version number of the linter I want to use**. In every project that uses eslint, the version of eslint they need is [right here](https://github.com/GoogleChrome/lighthouse/blob/6d5d4feb583c5648eed825f507089c728f1a9d7c/package.json#L44) in the package.json's devDependencies. So I would consider this to be CodeClimate's bug. 
Sure, any project definitely relying on CC on a daily basis will probably have the yml for configuration, but this bug adds friction for people curious about the service and who want to see how awesome it is.

paulirish:
Merging just so I can see what the results look like. 
Almost certain I will revert the commit afterwards. :)

PR #1717, opened 2017-02-13
js style: allow single line if statements

paulirish:
Update: See [comment below](https://github.com/GoogleChrome/lighthouse/pull/1717#issuecomment-279567658), as the PR changed 100%. The details at the top here are no longer valid.

-----

This is a proposal for slightly cleaner brace style. 

This style is the same as chrome devtools frontend style.
We are currently inheriting default eslint rules here. Over in `eslint-config-google` it does [say](https://github.com/google/eslint-config-google/blob/master/index.js#L81-L82) next to the `curly` rule :  

> TODO(philipwalton): add an option to enforce braces with
>                   // the exception of simple, single-line if statements

so what do ya'll think?

-----

Before:
```js
if (err) {
  $.util.log(err);
}
```

After:
```js
if (err)
  $.util.log(err);
```


patrickhulce:
I'm not really a huge fan of it because of the diff noise when needing to add a 2nd line.
OTOH I love me some one-line guard clauses, and it is more consistent with devtools, so I'm game!

ebidel:
Also 💘  the braces, all the time.

brendankenny:
I'm still not a fan of these :) Yes, it can be visually noisy at times, but I've definitely been burned by bugs around this, and I'm not a fan of the diff noise either.

> TODO(philipwalton): add an option to enforce braces with
> // the exception of simple, single-line if statements

FWIW, I believe this is referring to the case of

```js
if (err) $.util.log(err);
```
see https://google.github.io/styleguide/jsguide.html#formatting-braces-all

paulirish:
PTAL: I've reverted the previous changes and provided some new.

The new changes don't violate our existing eslint rules, so no config change necessary. ;)

I've done these changes manually (roughly)¹. I didn't apply the change in many cases, when it would make the lines more difficult to read. Personally, I prefer the new look in all cases included in this PR, but I'm open to reverting any in particular. (like the two assignments in here)

¹ _Say hello to `^\s+if (\(.{0,80}\))\s{\n(?:\s+)(.{0,60};)\n(?:\s+}$)`_

XhmikosR:
For what is worth, I believe 1) being consistent pays off and 2) having braces everywhere can prevent some issues. Just my 2 cents. :)

wardpeet:
I think I have the same opinion as the rest. I do love the curly brackets but if it's google styleguide I can live with the one liner but we should enforce it so we don't get mixes between one liners.

//cc @paulirish

brendankenny:
I'm in favor of dropping this PR rather than rebasing and then just naturally using the single-line if statements as they make sense (and aid code clarity) as code is added/changed.

paulirish:
sg. it's dead.

PR #1726, opened 2017-02-15
fix: clean up table styling

patrickhulce:
R: @paulirish all

**Before**
image

**After**
image

### DIFF
--- a/lighthouse-core/formatters/partials/table.css
+++ b/lighthouse-core/formatters/partials/table.css
@@ -41,7 +41,7 @@
 .table_list .table-column {
   text-align: right;
 }
-.table_list .table-column.table-column-pre {
+.table_list .table-column.table-column-pre, .table_list .table-column.table-column-code {
   text-align: left;
 }
 .table_list .table-column.table-column-url {
--- a/lighthouse-core/formatters/partials/table.css
+++ b/lighthouse-core/formatters/partials/table.css
@@ -41,6 +41,9 @@
 .table_list .table-column {
   text-align: right;
 }
+.table_list .table-column.table-column-pre {
+  text-align: left;
+}
 .table_list .table-column.table-column-url {
   text-align: left;
   width: 250px;
--- a/lighthouse-core/formatters/partials/table.css
+++ b/lighthouse-core/formatters/partials/table.css
@@ -28,11 +28,6 @@
 .table_list tr:hover {
   background-color: #fafafa;
 }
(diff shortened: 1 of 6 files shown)


lighthouse-core/report/report-generator.js
```diff
@@ -111,6 +111,20 @@ class ReportGenerator {
       }
     });
 
+    Handlebars.registerHelper('kebabCase', function(str) {
```
patrickhulce: would love to rebase against  #1660 and add tests here

lighthouse-core/formatters/partials/table.html
```diff
@@ -7,15 +7,15 @@
...
+          <th class="table-column table-column-{{ kebabCase @key }}">{{this}}</th>
         {{/each}}
       </tr>
     </thead>
     <tbody>
       {{#each rows}}
         <tr>
           {{#each cols}}
-            <td>{{ sanitize this}}</td>
+            <td class="table-column table-column-{{ kebabCase (lookup ../../headingKeys @index) }}">{{ sanitize this}}</td>
```
patrickhulce: don't like this, but not sure it's worth restructuring the format of rows just for class names here

lighthouse-core/formatters/partials/table.css
```diff
@@ -43,11 +38,23 @@
 .table_list em + code, .table_list em + pre {
   margin-top: 10px;
 }
-.table_list td img {
+.table_list .table-column {
+  text-align: right;
+}
+.table_list .table-column.table-column-url {
+  text-align: left;
+  width: 250px;
```
patrickhulce: could add more here, but on smaller screen sizes the optimized images table starts getting pretty squished, min-width doesn't seem to have an impact on fixed table width

lighthouse-core/formatters/partials/table.html
ebidel: Nice. It's useful to be able to target these. It's hard with a generic table, but specific data that might nee to look different here/there..

ebidel:
This LGTM. One thing I notice is the `text-align: right` makes the code snippets funky:

screen shot 2017-02-15 at 9 46 08 am

vs before:

screen shot 2017-02-15 at 9 46 12 am

patrickhulce:
oh thanks good catch @ebidel, fixed

ebidel:
LGTM

PR #1731, opened 2017-02-16
feat: allow extension of default config

patrickhulce:
would address part of #1730, it's rather out there so comments and thoughts welcome on other possible approaches

usage

```
{
  extends: true,
  passes: [
    {passName: "defaultPass", gatherers: [MyCustomGatherer]
  ],
  audits: [MyCustomAudit]
}
```

Issue #1730, opened 2017-02-16
WPT compat: play nicely with video recording


patrickhulce:
WPT already records the video of Lighthouse working its magic but a couple of things need to be done on our side to make processing possible since we're mostly a blackbox.

- [x] ~Support custom blank pages per pass through config~ (#1732)
- [x] ~Support custom blank page hold durations per pass through config~ (#1732)
- [x] ~Support config extensions~ (#1731)

PR #1732, opened 2017-02-16
feat: configurable blank page options

patrickhulce:
addresses parts of #1730 
would be used in conjunction with #1731

usage
```
{
  passes: [
    {passName: "defaultPass", blankPage: "http://google.com", blankDuration: 10000}
    ...
  ]
  ...
}
```

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -63,11 +63,12 @@ class GatherRunner {
    * reload. We do not `waitForLoad` on about:blank since a page load event is
    * never fired on it.
    * @param {!Driver} driver
+   * @param {url=} url
+   * @param {number=} duration
    * @return {!Promise}
    */
-  static loadBlank(driver) {
-    return driver.gotoURL('about:blank')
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -158,7 +159,9 @@ class GatherRunner {
    * @return {!Promise}
    */
   static beforePass(options, gathererResults) {
-    const pass = GatherRunner.loadBlank(options.driver);
+    const blankPage = options.config.blankPage;
```
ebidel: blankPage -> startPage?

patrickhulce: Since blank is how we referred to this concept before and WPT also uses the 'blank' nomenclature, I'd prefer to stick with it, but open to other ideas to make it clearer what's going on here. Any thoughts?

ebidel: Yea, not sure. `blankStartPageUrl`? haha. We should stick with the existing nomenclature though. I wasn't aware of that part.

PR #1735, opened 2017-02-16
fix: do not fail if chrome could not be killed

patrickhulce:
should deflake the appveyor tests that have been failing ~50% of the time because of this

### DIFF
--- a/lighthouse-cli/chrome-launcher.ts
+++ b/lighthouse-cli/chrome-launcher.ts
@@ -237,11 +237,7 @@ class ChromeLauncher {
             process.kill(-this.chrome.pid);
           }
         } catch (err) {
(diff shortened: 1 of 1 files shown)


XhmikosR:
But why do they fail in the first place? Is this normal?

patrickhulce:
@XhmikosR if I had to guess it's related to tab closing behavior and that there's a race between closing the tab which we don't wait for and killing Chrome. On Mac closing the last window does not usually quit the application, but the paradigm in Windows is usually to kill the process once the last window has been closed. In general, I think it's a bad idea to throw a fatal error for cleanup that has already been done.

XhmikosR:
I agree, I'm just wondering why this happens only on Windows and if it hides something else.

If the reason is what you suspect I guess it's OK.

PR #1740, opened 2017-02-17
Disable throttling for non-performance passes

paulirish:
This is the approach that I ended up with after playing around a bit.

* `useThrottling` prop in the config for each pass
* if that is true and the disable CLI flags aren't set, then we'll turn it on.
* if the prop is false for a pass we'll definitely make sure throttling is off
* when `goOnline` comes back it'll restore whatever throttling should be for that pass.
* `setThrottling` method in driver does the work. the old `beginEmulation` now defers to it for some parts.

wdyt?

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -677,25 +677,29 @@ class Driver {
     return this.sendCommand('Runtime.enable');
   }
 
-  beginEmulation(flags) {
+  beginEmulation(options) {
     return Promise.resolve().then(_ => {
-      if (!flags.disableDeviceEmulation) {
+      if (!options.flags.disableDeviceEmulation) {
         return emulation.enableNexus5X(this);
       }
+    }).then(_ => {
+      return this.setThrottling(options.flags, {useThrottling: true});
     });
   }
 
   setThrottling(cliFlags, passConfig) {
-    const emulations = [
-      emulation.disableNetworkThrottling(this),
-      emulation.disableCPUThrottling(this)
-    ];
+    const emulations = [];
 
-    if (!cliFlags.disableNetworkThrottling && passConfig.useThrottling) {
-      emulations.push(emulation.enableNetworkThrottling(this));
+    if (!passConfig.useThrottling) {
+      emulations.push(emulation.disableNetworkThrottling(this));
+      emulations.push(emulation.disableCPUThrottling(this));
+      return Promise.all(emulations);
     }
 
-    if (!cliFlags.disableCpuThrottling && passConfig.useThrottling) {
+    if (!cliFlags.disableNetworkThrottling) {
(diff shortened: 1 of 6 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -673,21 +673,31 @@ class Driver {
     return this.sendCommand('Runtime.enable');
   }
 
-  beginEmulation(flags) {
+  beginEmulation(options) {
```
patrickhulce: any reason to switch from flags to options? looks like this just uses flags

lighthouse-core/gather/driver.js
```diff
@@ -673,21 +673,31 @@ class Driver {
...
+      return this.setThrottling(options.flags, {useThrottling: true});
+    });
+  }
+
+  setThrottling(cliFlags, passConfig) {
     const emulations = [];
 
-    if (!flags.disableDeviceEmulation) {
-      emulations.push(emulation.enableNexus5X(this));
+    if (!passConfig.useThrottling) {
```
patrickhulce: this is one case I think having it nested would make it more readable, took me a bit to spot the early return (though that may be more diff noise and it's easier without the diff)

thoughts on the following?
```
if (useThrottling) {
  if (!disableNetwork) emulations.push()
  if (!disableCpu) emulations.push
} else {
  emulations.push(...)
}

return Promise.all()
```

lighthouse-core/gather/driver.js
```diff
@@ -696,11 +706,11 @@ class Driver {
...
   goOffline() {
-    return this.sendCommand('Network.enable').then(_ => {
-      return emulation.goOffline(this);
-    }).then(_ => {
-      this.online = false;
-    });
+    return this.sendCommand('Network.enable')
+      .then(_ => emulation.goOffline(this))
+      .then(_ => {
+        this.online = false;
```
patrickhulce: nit: could probably one-line this too

lighthouse-core/gather/driver.js
```diff
@@ -673,21 +673,31 @@ class Driver {
...
+    return Promise.resolve().then(_ => {
+      if (!options.flags.disableDeviceEmulation) {
+        return emulation.enableNexus5X(this);
+      }
+    }).then(_ => {
+      return this.setThrottling(options.flags, {useThrottling: true});
+    });
+  }
+
+  setThrottling(cliFlags, passConfig) {
```
ebidel: should passConfig just be a boolean, `useThrottling`?

ebidel: Ah, see you're passing in the `options.config` below. Guess it's fine.

lighthouse-core/gather/driver.js
paulirish: so much better. done.

PR #1741, opened 2017-02-17
DOMStat audit: fix for traversing shadow root anchor parents and svg .className

ebidel:
R: all

Couple of subtle edge cases found on https://www.chromestatus.com/features, which uses shadow dom and inline svg.

The first is traversing up through parent elements looking for shadow hosts / parentElements. Anchors have a `.host` property. So looking at `node.host` does not necessarily give you a shadow root :\ First time I've ran into this when working with web components!

Now we construct the depth properly:

screen shot 2017-02-16 at 9 55 10 pm

The second issue fixed is checking `svgNode.className`. Turns out, svg nodes return a `SVGAnimatedString = {animVal: string, baseVal: string}` instead of a string. Ugh! Switched to using the `class` attr instead.

Added chromestatus.com to the smoketests.

### DIFF
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -194,6 +194,20 @@ module.exports = [
       }
     }
   }, {
+    initialUrl: 'https://www.chromestatus.com/features',
+    url: 'https://www.chromestatus.com/features',
+    audits: {
+      'dom-size': {
+        score: 100,
+        extendedInfo: {
+          value: {
+            1: {value: '20'},
+            2: {snippet: 'Element with most children:\nul#versionlist.canaryisdev'}
+          }
+        }
+      }
+    }
+  }, {
     initialUrl: 'http://localhost:10200/online-only.html',
     url: 'http://localhost:10200/online-only.html',
     audits: {
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -194,20 +194,6 @@ module.exports = [
       }
     }
   }, {
-    initialUrl: 'https://www.chromestatus.com/features',
-    url: 'https://www.chromestatus.com/features',
-    audits: {
-      'dom-size': {
-        score: 100,
-        extendedInfo: {
-          value: {
-            1: {value: '20'},
-            2: {snippet: 'Element with most children:\nul#versionlist.canaryisdev'}
-          }
-        }
-      }
-    }
-  }, {
     initialUrl: 'http://localhost:10200/online-only.html',
     url: 'http://localhost:10200/online-only.html',
     audits: {
--- a/lighthouse-core/gather/gatherers/dobetterweb/domstats.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/domstats.js
@@ -39,9 +39,11 @@ function createSelectorsLabel(element) {
   if (idAttr) {
     name += `#${idAttr}`;
   }
-  const className = element.className;
+  // svg elements return SVGAnimatedString for .className, which is an object.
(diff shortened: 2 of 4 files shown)


lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -39,9 +39,11 @@ function createSelectorsLabel(element) {
   if (idAttr) {
     name += `#${idAttr}`;
   }
-  const className = element.className;
+  // svg elements return SVGAnimatedString for .className, which is an object.
+  // Use the class attr instead.
+  const className = element.getAttribute && element.getAttribute('class');
```
paulirish: I think elem.classList toString()s to the same values.  Can't double-check this ATM tho.

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -55,7 +57,11 @@ function elementPathInDOM(element) {
   const path = [createSelectorsLabel(element)];
   let node = element;
   while (node) {
-    node = node.parentNode && node.parentNode.host ? node.parentNode.host : node.parentElement;
+    // Anchor elements have a .host property. Be sure we've found a shadow root
+    // host and not an anchor.
+    const shadowHost = node.parentNode && node.parentNode.host &&
```
paulirish: Is there a more standard way of detecting it's a shadow host? This seems fairly oblique.

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
ebidel: Yep that works. Done.

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
ebidel: This is it :( https://developers.google.com/web/fundamentals/getting-started/primers/shadowdom?hl=en#create

Typically, you don't traverse upwards through the entire deep DOM tree like this. So it's a big of a special case for what we're doing.

paulirish:
Adding chromestatus as a smoke test will mean the total test time will grow a bit. But generally the coverage seems worth it. 
Should we add some more expectations to that run since we're doing it anyway?

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -194,6 +194,20 @@ module.exports = [
       }
     }
   }, {
+    initialUrl: 'https://www.chromestatus.com/features',
```
patrickhulce: Up until now we've entirely controlled the expectations here which can be run locally, can we not move this to belong with other external sites?

patrickhulce: Oh wait we're already hitting this in smokehouse, can we just add the assertion there?

ebidel: Removed this entry and added the 'dom-size' audit to the pwa_expectations file. Boom.

ebidel:
PTAL

ebidel:
Not sure what's up with travis. The smoketests pass locally for me. Maybe a caching issue?

patrickhulce:
hmm @ebidel it's failing for me locally too, it looks like we only run a subset of the config on those expectations only for PWA-ness :( I'm still in favor of adding the audit there and doing more comprehensive checking of the live sites over time.

ebidel:
@patrickhulce you mean move chromestatus back to dbw_expectations.js?

patrickhulce:
@ebidel oh no, I meant keeping it where it is, just adding the dom-size audit to the config for the external sites.

ebidel:
Crap, I didn't even know that file exists. Thanks for pointing it out.

ebidel:
Yay. Tests are passing.

patrickhulce:
Yay! Thanks :)

Issue #1613, opened 2017-02-02
Invert the report form for the discouraged API use


pavelfeldman:
Context:

![use_offline](https://cloud.githubusercontent.com/assets/883973/22571148/61e973d0-e953-11e6-9a4e-c86eeb6e55bc.png)

Providing a sharp actionable report would please the user more than repetitive praise for something they did not do. Consider inverting the report form here and listing / penalizing for the deprecated API use.

pavelfeldman:
I see more of these:
- Avoids Mutation Events in its own scripts
- Avoids document.write()
- Avoids requesting the geolocation permission on page load

.. and many more.

but I did not avoid any of these...

ebidel:
"Avoids" was introduced in https://github.com/GoogleChrome/lighthouse/pull/1598. "Does not use S..." was what we had before, but that was deemed confusing to see next to a green check. Discussion is here: https://github.com/GoogleChrome/lighthouse/pull/1598#pullrequestreview-19646677.

Either way, I'm not sure we can make everyone happy.

pavelfeldman:
I have no problem with "Avoid" as a term when I did avoid. I just don't like it reported for about:blank. As the project scales, you'll have hundreds of violations to track and reporting them all as passed for about:blank would not be reasonable. I'm suggesting to only list violations in the dedicated violations group.

ebidel:
How about something like this:

![screen shot 2017-02-02 at 4 09 42 pm](https://cloud.githubusercontent.com/assets/238208/22574161/169e85a0-e962-11e6-84d7-ac52f391b263.png)

For every section, passing all of the audits will give you an overall "pass" and hide the children. If you fail 1 or more, it'll open up and show the list.

pavelfeldman:
That looks better, but still seems a bit silly to say "Uses modern offline features" for the site that does not support offline at all. Just because it did not use WebSQL...

ebidel:
The subcategory names could probably use a rethinking, but I'm not sure the right answer without moving back to "Does not use...." headings. But I think the screenshot alleviates some of the original issue. Our new theme should be "less is more".

paulirish:
> ## Things we don't want you to do.. and you didn't. \o/

> ...

> ## Things we don't want you to do, but you DID! >:| 

> ...

ebidel:
[...]

PR #1742, opened 2017-02-17
Report: collapse aggregations when all audits pass

ebidel:
R: @patrickhulce @paulirish 

In the act of de-cluttering, this PR addresses some of of https://github.com/GoogleChrome/lighthouse/issues/1613 and implements the proposed solution in https://github.com/GoogleChrome/lighthouse/issues/1613#issuecomment-277126590.

Now, if all audits pass, the entire aggregation is collapsed. If one or more audits fail, the aggregation is opened up.

Before:

screen shot 2017-02-16 at 10 58 53 pm

After:

screen shot 2017-02-16 at 10 40 05 pm

when an audit fail: 

screen shot 2017-02-16 at 10 40 24 pm

BTW, the aggregation icons are slightly bigger than the subaudits. I thought the size difference was good but I'm happy to make them all the homogeneous. 

- Aggregations now always create an overall score. Please look over those changes as I may have missed something.
- Adds css transitions to the details dropdown.
- @patrickhulce turns out the `expectedValues`/`weights` are important and we'll need to use them after all :) aggregate.js uses them to type check and calculate an overall score based on the weights.

### DIFF
--- a/lighthouse-core/config/default.json
+++ b/lighthouse-core/config/default.json
@@ -258,73 +258,98 @@
       "name": "Using modern offline features",
       "audits": {
         "appcache-manifest": {
-          "expectedValue": false
+          "expectedValue": true,
+          "weight": 1
         },
         "no-websql": {
-          "expectedValue": false
+          "expectedValue": true,
+          "weight": 1
         }
       }
     }, {
       "name": "Using modern protocols",
       "audits": {
         "is-on-https": {
-          "expectedValue": false
+          "expectedValue": true,
+          "weight": 1
         },
         "uses-http2": {
-          "expectedValue": false,
-          "description": "Resources made by this application should be severed over HTTP/2 for improved performance."
+          "expectedValue": true,
+          "description": "Resources made by this application should be severed over HTTP/2 for improved performance.",
+          "weight": 1
         }
       }
     }, {
       "name": "Using bytes efficiently",
       "audits": {
-        "unused-css-rules": {},
-        "uses-optimized-images": {},
-        "uses-responsive-images": {}
+        "unused-css-rules": {
+          "expectedValue": true,
+          "weight": 1
+        },
+        "uses-optimized-images": {
+          "expectedValue": true,
+          "weight": 1
+        },
+        "uses-responsive-images": {
+          "expectedValue": true,
+          "weight": 1
+        }
       }
     }, {
       "name": "Using modern CSS features",
       "audits": {
         "no-old-flexbox": {
-          "expectedValue": false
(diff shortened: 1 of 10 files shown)


patrickhulce:
> aggregate.js uses them to type check and calculate an overall score based on the weights

sure it uses them *now* ;) this is awesome though!

lighthouse-core/config/default.json
```diff
@@ -387,11 +412,11 @@
           "weight": 1
         },
         "critical-request-chains": {
-          "expectedValue": 0,
```
patrickhulce: I thought this one actually was numeric, is it not?

paulirish:
Totally on board with the collapsing. Nice.

Regarding scoring, are you saying we'll now have scores for "Best practices" and "fancier stuff"?

ebidel:
> Regarding scoring, are you saying we'll now have scores for "Best practices" and "fancier stuff"?

Sorry. There's still only one score in the  report (PWA). The difference with this PR is that it calculates an overall score for each subitem. That's not surfaced in the report though.

Before (all `overall: 0` for unscored audits):

screen shot 2017-02-17 at 10 02 24 am

Now:

screen shot 2017-02-17 at 10 00 29 am

patrickhulce:
After playing with it for a bit, I really like it, but the UX of the cards/boxes could use some work since I don't think it plays well with the collapsed report. Everything is rather succinct and cleanly laid out and the three large uncollapsible boxes really jump out. Can we have those collapse too?

image

patrickhulce:
Yeah that's a fair point. Right now they just really stick out. If we used them more throughout I'd probably feel differently, but probably need to a strike a balance to prevent wall of text becoming wall of boxes :)

paulirish:
> Personally I like seeing them b/c it's something other than text.

Agree with that, but if those numbers are in good shape then it's not adding much value to always show them.
(Also maybe the target for each value should be in the card, as I don't know what numbers are good for children and depth)

ebidel:
Added target vals support to the cards and they're now hidden by default. PTAL.

screen shot 2017-02-17 at 12 40 26 pm

lighthouse-core/config/default.json
ebidel: The `rawValue` that we display is numeric, yes. But we treat the audit as a pass/fail, which means `expectedValue` needs to be a boolean. AFAICT, `expectedValue` should really be `expectedType`.

ebidel:
@patrickhulce you happy too?

patrickhulce:
Oh sorry yep!

Issue #1748, opened 2017-02-17
Responsive images: don't warn about duplicate images


patrickhulce:
Currently, you'll be warned about a wasted image if you use an image more than once even if one of the uses is legitimate.

ebidel:
Example: https://twitter.com/jescalan/status/832667692188450816

PR #1749, opened 2017-02-17
fix: responsive image audits for vectors and duplicates

### LINKED ISSUE #1747: Responsive images: don't warn about vectors
Vector sizing can be tricky to determine what's actually "waste", we should just ignore for now.

patrickhulce:
fixes #1747 and #1748

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
@@ -89,6 +89,7 @@ class UsesResponsiveImages extends Audit {
     let debugString;
     const DPR = contentWidth.devicePixelRatio;
     const resultsMap = images.reduce((results, image) => {
+      // TODO: give SVG a free pass until a detail per pixel metric is available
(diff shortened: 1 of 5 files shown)


lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
```diff
@@ -90,29 +87,32 @@ class UsesResponsiveImages extends Audit {
...
     const DPR = contentWidth.devicePixelRatio;
-    const results = images.reduce((results, image) => {
-      if (!image.networkRecord) {
+    const resultsMap = images.reduce((results, image) => {
+      if (!image.networkRecord || image.networkRecord.mimeType === 'image/svg+xml') {
         return results;
       }
 
       const processed = UsesResponsiveImages.computeWaste(image, DPR);
-      if (!processed) {
```
ebidel: we don't need the `null` check anymore?

patrickhulce: correct, it will either throw or return an object now

Issue #1448, opened 2017-01-10
Adopt handlebars precompiled templates


paulirish:
Currently we browserify all of handlebars for our extension and report-viewer builds. 
It would make more sense to adopt precompiled templates, which are much smaller and simpler to run: http://handlebarsjs.com/precompilation.html

https://metabroadcast.com/blog/precompilation-of-templates-with-handlebars

In that case, we'd require in the precomplied template + runtime rather than the original template file and all of `handlebars`. 
https://www.npmjs.com/package/gulp-handlebars

This will require some experimenting. Open for anyone to investigate.  ([For example](https://github.com/lazd/gulp-handlebars/pull/53)).. big plus if you've worked with precompiled handlebars before. ;)

brendankenny:
this was closed in #1752

PR #1752, opened 2017-02-18
Adopt handlebars precompiled templates

sendilkumarn:
#1448 

@paulirish this will need us to run `watch` even when we are testing locally on `cli`, in order to compile the templated. Otherwise we have to split the generateReport into 2 different methods one handles with html files and other with precompiled template file.

Note: This is my first attempt with precompile ;)


paulirish:
Hmm! So it looks like after this PR, a few of the compiled files actually get larger.

* viewer/main.js goes from 320kb to 344kb
* lighthouse-background.js goes from 2.5mb to 2.6mb

so it's not a win in filesize, at least. It might still be worth it for either runtime or complexity but i'm not sure about that yet.

regardless, i really appreciate you giving this one a shot, @sendilkumarn!  super interesting.

sendilkumarn:
@paulirish 
Sorry.
Apparently, I did not include partials. So after including partials and then removing html references. This is what I ended up with. ( completely dereferencing handlebar -> handlebar/runtime )

- viewer/main.js goes from 320kb to `283kb`
- lighthouse-background.js goes from 2,541,212 kb to `2,452,002kb`

lighthouse-extension/gulpfile.js
```diff
@@ -178,6 +207,14 @@ gulp.task('watch', ['lint', 'browserify', 'html'], () => {
     'app/src/**/*.js',
     '../lighthouse-core/**/*.js'
   ], ['browserify']);
+
+  gulp.watch([
+    '../lighthouse-core/**/*.html'
```
paulirish: this and the partials wildcard should overlap, right? does the order of these compile tasks matter?

lighthouse-core/report/templates/report-templates.js
```diff
@@ -0,0 +1,283 @@
+this["report"] = this["report"] || {};
```
paulirish: wondering if we should put this in gitignore instead.

lighthouse-extension/gulpfile.js
sendilkumarn: Yes. If we want to change anything, straightway in that precompiled file, this will help. 

Order, yeah we have to run compile first and then browserify.

lighthouse-core/report/templates/report-templates.js
sendilkumarn: That is debatable. If we add in gitignore,  we need to make sure that it will be generated on the fly for people who just clones the repo and runs node-cli command.

ebidel:
I like anything that makes our build smaller :)

But I'm not sure we should have the compile steps as part of the extension/viewer. For one, it's duplicated code. It would be nice to have shared tasks in a `gulp_scripts` dir in the top level repo. Speaking of which, is it time for a main gulpfile.js in `lighthouse/`?

sendilkumarn:
Yeah that would be awesome. It will also solves the above comment
`wondering if we should put this in gitignore instead.`

lighthouse-core/report/report-generator.js
```diff
@@ -164,7 +150,11 @@ class ReportGenerator {
         Handlebars.registerHelper(helpers);
       }
 
-      Handlebars.registerPartial(audit.name, formatter.getFormatter('html'));
+      const partials =reportPartials.report.partials;
```
patrickhulce: nit: space after `=`

lighthouse-core/report/report-generator.js
```diff
@@ -133,7 +119,7 @@ class ReportGenerator {
    * @return {string} HTML of the exception page.
    */
   renderException(err, results) {
-    const template = Handlebars.compile(this.getExceptionTemplate());
+    const template = reportTemplate.report.template['exception'];
```
patrickhulce: nit: `.exception`

lighthouse-core/report/report-generator.js
```diff
@@ -164,7 +150,11 @@ class ReportGenerator {
         Handlebars.registerHelper(helpers);
       }
 
-      Handlebars.registerPartial(audit.name, formatter.getFormatter('html'));
+      const partials =reportPartials.report.partials;
+      Handlebars.registerPartial(audit.name,
+        Handlebars.template(
+          partials[audit.name] ? partials[audit.name] : partials['empty-formatter']
```
patrickhulce: nit: `partials[audit.name] || partials['empty-formatter']`

sendilkumarn:
Thanks @patrickhulce  👍

lighthouse-core/report/report-generator.js
sendilkumarn: wondering why `eslint` did not complain ...

sendilkumarn:
Added a gulp folder to write the common tasks and reference from there.

gulp/compile-partials.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
```
ebidel: Can you drop our license header in these new files?

lighthouse-core/test/formatter/formatters-test.js
```diff
@@ -43,6 +42,7 @@ const walkTree = new Promise((resolve, reject) => {
 });
 
 describe('Formatters', () => {
+  /* TODO : Remove these tests and add appropriate tests based on handlebar pre-compiled templates
   it('has no formatters failing when getFormatter("html") is called', () => {
```
ebidel: Let's use `.skip` instead: https://github.com/GoogleChrome/lighthouse/blob/6943b99955bf9faf9ba47c2845fa547b6df72317/lighthouse-core/test/audits/screenshots-test.js#L32

lighthouse-extension/gulpfile.js
ebidel: So they can't be combined? Also can we utilize `config.report` and `config.partials` for the globs?

lighthouse-viewer/gulpfile.js
```diff
@@ -116,28 +123,39 @@ gulp.task('clean', () => {
...
+      runSequence('html');
+    });
 
-  gulp.watch([
-    'app/src/**/*.js',
-    '../lighthouse-core/**/*.js'
-  ], ['browserify']);
-});
+    gulp.watch([
+      '../lighthouse-core/**/*.html'
```
ebidel: same comment here.

gulp/compile-report.js
```diff
@@ -0,0 +1,35 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
```
ebidel: what do you think about combining these two files into one and exporting both functions? They're very similar.

sendilkumarn:
@ebidel `null-formatter` is used for `partials` with no `partial-template` right ? am I wrong ?

sendilkumarn:
👍 for travis I think that will be enough.

ebidel:
Right. But it has 0 LOC changed:

screen shot 2017-03-01 at 6 43 04 pm


Never seen that before. Not sure why that file is being touched.

sendilkumarn:
Well it is just an empty file.

I created it. Before that null-formatter js will just return an empty string. Since we cannot use that for pre-compiling templates. Added that file

ebidel:
Looks good @sendilkumarn! Did you see my comments on .travis.yml and package.json updates? https://github.com/GoogleChrome/lighthouse/pull/1752#pullrequestreview-24552317

sendilkumarn:
Yeah absolutely fine with that. 
[Check here](https://github.com/GoogleChrome/lighthouse/pull/1752#issuecomment-283527411)

I will add and push.

package.json
```diff
@@ -15,7 +15,7 @@
     "install-cli": "cd ./lighthouse-cli && npm install",
     "install-extension": "cd ./lighthouse-extension && npm install",
     "install-viewer": "cd ./lighthouse-viewer && npm install",
-    "build-all": "npm run build-cli && npm run build-extension && npm run build-viewer",
+    "build-all": "npm run build-cli && npm run build-extension && npm run build-viewer && gulp",
```
ebidel: Shouldn't `gulp` be first so the templates are compiled by the time other places need them?

sendilkumarn:
@ebidel yeah exactly. Thanks 👍

package.json
```diff
@@ -24,7 +24,7 @@
     "coverage": "node $(npm bin)/istanbul cover -x \"**/third_party/**\" _mocha -- $(find */test -name '*-test.js') --timeout 10000 --reporter progress",
     "coveralls": "npm run coverage && cat ./coverage/lcov.info | coveralls",
     "start": "node ./lighthouse-cli/index.js",
```
patrickhulce: do we want to add gulp to start too since it won't be able to run without it now?

patrickhulce: I'm leaning towards not incurring that cost every start, but a note in the README or something would be nice :)

sendilkumarn: do we need to have pre-compiled templates for `global installs` ? currently we are ignoring those.
 Do you think, it makes sense to add them under `develop` section ?

paulirish:
Huge. Thanks @sendilkumarn for taking this on. This was a yeoman's work and we're really jazzed about it.

sendilkumarn:
Thank you 👍

PR #680, opened 2016-09-21
Improved currentTab impl

paulirish:
Our current `chrome.windows.WINDOW_ID_CURRENT` is pretty flaky with devtools open. 

Also, `currentWindow: true` isn't ideal:

> The current window is the window that contains the code that is currently executing. It's important to realize that this can be different from the topmost or focused window.

The `lastFocusedWindow` property is what we want, plus `windowType` lets us make sure we're not selecting a chrome app/extension. https://developer.chrome.com/extensions/tabs#method-query

R=@wardpeet

Issue #1745, opened 2017-02-17
Unable to generate report for working site


mblayman:
Hi, I was excited to try Lighthouse on the production site that I work on. Unfortunately, the tool is erroring 100% of the time for me when I try to run it. Every time I run it against https://storybird.com/, I get the following message:

![image](https://cloud.githubusercontent.com/assets/636865/23073960/45fce478-f505-11e6-9eba-3491172d37fc.png)

I've tried other pages on Storybird as well as authenticated and non-authenticated. Nothing I've tried was successful.

I installed the extension today and am running Chrome 56.0.2924.87 (64-bit) on OS X 10.11.6. As a sanity check, I also attempted to restart the browser after installing the extension. As an additional check, I ran the tool against https://www.google.com/ and that worked so I have confidence that the installation of the extension is correct.

I don't know if there are any logs to inspect what went wrong so I'm not sure how to fix whatever the problem is. I'd love to use Lighthouse so any insight would be appreciated. Thanks!

mblayman:
As a bit of extra data, I tried to disable all but 1 option and ran three separate tests for three separate options to see if the problem was related to a single test. That failed.

Then I tried to run it with *all* of the options unchecked. That failed as well.

patrickhulce:
So sorry you're having such trouble @mblayman ! That's strange as both the CLI and extension worked perfectly for me on that site first try (also Chrome 56 on OS X). The logs of the CLI tool are a bit more verbose for this category of error if you'd like to give that a try and see if you get the same results. You can also inspect the extension itself to see more logs (during a Lighthouse run, go to `chrome://inspect#extensions` and click inspect under the Lighthouse option).

Would you be able to try your site using the npm module and see if you get the same results?

paulirish:
Does your site redirect to a subdomain or other domain? I was saying a similar thing yesterday on the CLI with "about:version" not working but it's post redirect "chrome://version" being fine. 
Either way.. it's our bug.

mblayman:
@patrickhulce thanks, I was unaware of `chrome://inspect#extensions`. When I ran it again, I got this in the logs:

<img width="1380" alt="image" src="https://cloud.githubusercontent.com/assets/636865/23075640/404b5b8a-f50b-11e6-87d7-9ed06682a042.png">

When I watch Lighthouse run, it seems to die every time just after the "Loading page & waiting for onload: HTTPRedirect, HTTPWithoutJavaScript."

[...]

PR #1753, opened 2017-02-18
Speculative fix for getCurrentTabURL

### LINKED ISSUE #1214: Extension Error: You must provide a url to the driver
**Lighthouse Version**: 1.3.0
**Chrome Version**: 55.0.2883.95
**Error Message**: You must provide a url to the driver
**Stack Trace**:
 ```Error: You must provide a url to the driver
    at Function.run (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:11221:29)
    at window.runLighthouseForConnection (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:12188:17)
    at connection.getCurrentTabURL.then.url (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:12201:25)```

paulirish:
I am hoping this will fix #1214, but since the repro is particularly tricky, we won't know until later.

I believe is ultimately a chrome bug that's behavior is here: [546696 - chrome.windows.getLastFocused returns the last opened window, not the last focused, when an extension popup is visible - chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=546696#c4) but essentially its interplay between extensions with `onFocusChanged` `getLastFocused` and here, `chrome.tabs.query({lastFocusedWindow: true})`..

Of note, previous to the currentTab change in #680 I don't believe we had this bug, but we did have trouble when debugging the extension background page with a devtools window. I'm OK with regressing our developer ergonomics to get this user bug squashed.

http://stackoverflow.com/a/28794338 (by Rob W) describes the best solution to be:
```js
chrome.tabs.query({
    active: true,
    currentWindow: true    
}
```

The best practices around getting current tab have changed throughout the past 7 years, and there appear to be some still open bugs, so this is what we're running up against.

This PR adds extra error logging around all this so we'll have more information if the problem persists.

### DIFF
--- a/lighthouse-core/gather/connections/extension.js
+++ b/lighthouse-core/gather/connections/extension.js
@@ -136,8 +136,7 @@ class ExtensionConnection extends Connection {
     return new Promise((resolve, reject) => {
       const queryOpts = {
         active: true,
-        lastFocusedWindow: true,
-        windowType: 'normal'
+        lastFocusedWindow: true
       };
 
       chrome.tabs.query(queryOpts, (tabs => {
@@ -159,7 +158,7 @@ class ExtensionConnection extends Connection {
   getCurrentTabURL() {
     return this._queryCurrentTab().then(tab => {
       if (!tab.url) {
-        log.error('ExtensionConnection', 'getCurrentTabURL returned empty string');
+        log.error('ExtensionConnection', 'getCurrentTabURL returned empty string', tab);
       }
       return tab.url;
     });
--- a/lighthouse-core/gather/connections/extension.js
+++ b/lighthouse-core/gather/connections/extension.js
@@ -136,7 +136,7 @@ class ExtensionConnection extends Connection {
     return new Promise((resolve, reject) => {
       const queryOpts = {
         active: true,
-        lastFocusedWindow: true
+        currentWindow: true
       };
 
       chrome.tabs.query(queryOpts, (tabs => {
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/connections/extension.js
```diff
@@ -157,7 +159,12 @@ class ExtensionConnection extends Connection {
    * Used by lighthouse-background to kick off the run on the current page
    */
   getCurrentTabURL() {
-    return this._queryCurrentTab().then(tab => tab.url);
+    return this._queryCurrentTab().then(tab => {
+      if (!tab.url) {
+        log.error('ExtensionConnection', 'getCurrentTabURL returned empty string', tab);
```
ebidel: Is logging the tab obj going to be useful?

paulirish: Indeterminable

paulirish:
@mblayman re #1745. would you be willing to see if this PR fixes things for you?

I've uploaded a build of the extension and attached to this comment. You can load it in developer mode (and disable the other lighthouse) to give it a try.

➡️  [lighthouse-1.5.99999.zip](https://github.com/GoogleChrome/lighthouse/files/791857/lighthouse-1.5.99999.zip)

mblayman:
@paulirish I just tried the extension you supplied and I still have the same behavior for `https://storybird.com`. It hits the same timeout that I reported in #1745. 🤔

paulirish:
okay thanks for giving it a go!

Issue #1690, opened 2017-02-11
Automation to track our build sizes


paulirish:
For a number of our build assets, we'd like to keep track of how big they are over time.

* `lighthouse-background.js`: Used in both extension and devtools. 
* `viewer/main.js`: Used in the online viewer
* the tarball that's published to NPM

Relatedly.. we're also interested in tracking the **duration of doing a default lighthouse run**. Potentially against both a simple site (example.com) and a more complex one (chromestatus.com).

This is pretty wide open for implementation.

Some of this could be done based on our travis/smokehouse runs.
No preferences on where the data gets beaconed to, but something that's easy for the team and community to view.

wardpeet:
We might beacon it to firebase database? As we already have the viewer on there.

paulirish:
FWIW the [pwmetrics project](https://github.com/paulirish/pwmetrics/) has all the code necessary to publish results to google sheets. That'd be nice as we can make the data public and fairly hackable from there.

wtgtybhertgeghgtwtg:
Something that might shave off a few megabytes would be replacing `handlebars`.  The package alone is about 2.5 MB; factor in its dependencies on `uglify-js`, `source-map`, and `async`, and the fact that the version of `yargs` in its dependency tree doesn't dedupe with the version `lighthouse` uses, and it adds up pretty quickly.

paulirish:
@wtgtybhertgeghgtwtg 

Here's a treemap of our current build: https://cdn.rawgit.com/paulirish/d86e390c53e8de23511bf880fdffe1cd/raw/ac5e4188bf8cf4c835b3d9232b2ab000404d1782/lighthouse-build.html

![image](https://cloud.githubusercontent.com/assets/39191/23302618/39c6665e-fa44-11e6-966a-dbdc4f84fdc8.png)

handlebars isn't small, but it does only top out at 146kb of the overall build.

wtgtybhertgeghgtwtg:
Sorry, I was looking at the version in `node_modules`.

wardpeet:
@paulirish I could look at this if you like.

wardpeet:
I have a working branch of this https://github.com/wardpeet/lighthouse/tree/feature/beacon

@paulirish should we close this one until we know what to do with this one?

brendankenny:
https://github.com/siddharthkp/bundlesize could be helpful. Threshold limit and the status message mentions the delta of the PR (doesn't track over time, though)

[...]

PR #1756, opened 2017-02-21
Reduce build size by 33%

ebidel:
R: all

In this spirit of #1690, I've managed to get the build down to 1.6MB. 33% savings!

screen shot 2017-02-19 at 4 09 33 pm

Other low hanging fruit is stripping out the html/js license comments in the inlined files (`readFileSync`).

This PR introduces uglify-harmony into the extension gulpfile. `gulp build:production` will build lighthouse-extension.js and also compress it. `gulp build` continues to only broswerify. I felt that we may still want to keep that around for speedlier development when making tweaks to the crx. Happy to consolidate though.


lighthouse-extension/gulpfile.js
```diff
@@ -155,6 +162,33 @@ gulp.task('browserify', cb => {
   runSequence('browserify-lighthouse', 'browserify-other', cb);
 });
 
+gulp.task('compilejs', () => {
+  const opts = {
+    mangle: false, // Preserve  array notation property access. Needed for artifact name lookups.
```
patrickhulce: Interesting how does this disrupt array notation specifically? I had thought mangle was fine unless `Function.name`/`with`/`eval` though I'm not familiar with uglify-harmony gotchas

ebidel: It's pretty tricky to track down where things break down, but without `mangle: false`, the audit names get rewritten:

screen shot 2017-02-21 at 8 55 57 am
screen shot 2017-02-21 at 8 56 25 am

Hard to say if it also creates other errors.

paulirish:
In `build-traceviewer-module.js` we use babel to do some lightweight minification: 

https://github.com/GoogleChrome/lighthouse/blob/7a5448e46fb365decf5f9f6d38a28fbcaad564f8/lighthouse-core/scripts/build-traceviewer-module.js#L134-L138

are you open to using that? this way we wouldn't have to deal with disabling a bunch of options like mangle and sideEffects.

ebidel:
For the life of me, I couldn't get babel, typescript, or closure to play nicely. https://github.com/babel/babili looked the most promising (to avoid transpilation) but it choked after a certain amount of code and stopped minifying it.

In terms of filesize reduction, I didn't see disabling a few uglify compression settings do much to adversely effect file size.

lighthouse-extension/gulpfile.js
```diff
@@ -155,6 +162,33 @@ gulp.task('browserify', cb => {
...
+    warnings: true,
+    compress: { // turn off a bunch stuff so we don't create unexpected errors.
+      unused: false,
+      properties: false,
+      dead_code: false,
```
XhmikosR: Isn't dead code supposed to be caught by ESLint?

lighthouse-extension/gulpfile.js
patrickhulce: Oh right because we do use `Function.name` there with the gatherer class name, maybe update that comment then? The accessors seems like the secondary issue to the primary problem of the gatherer being renamed :)

paulirish:
I put up https://github.com/GoogleChrome/lighthouse/pull/1765 which replaces uglify with babel. It has one benefit of not adding any new dependencies at all. :)

lighthouse-extension/gulpfile.js
ebidel: Even if it does, we're not running eslint over the browserified bundle.

PR #1758, opened 2017-02-21
Report: use accessible material colors

ebidel:
R: all

The report colors were recently adjusted to meet accessibility contrast for WCAG 2 AA, but the new average color is mustard and the red/green are quite deep. This PR dials them back a bit, returns the average color to an orange, but still keeps WCAG 2 AA at 18pt. I used [this](https://snook.ca/technical/colour_contrast/colour.htm) tool to verify the colors are still compliant.
The colors are from the MD color palette. 

cc @robdodson 

Before:

screen shot 2017-02-21 at 1 04 08 pm

After:

screen shot 2017-02-21 at 1 01 43 pm

### DIFF
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -29,10 +29,10 @@ span, div, p, section, header, h1, h2, li, ul {
   --text-color: #222;
   --secondary-text-color: #606060;
   --accent-color: #3879d9;
-  --poor-color: #e81e1b;
-  --good-color: #008a07;
-  --average-color: #9f6d00;
-  --warning-color: #947200;
+  --poor-color: #e53935; /* md red 600 */
+  --good-color: #43a047; /* md green 600 */
+  --average-color: #ef6c00; /* md orange 800 */
+  --warning-color: #757575; /* md grey 600 */
   --gutter-gap: 12px;
   --gutter-width: 40px;
   --body-font-size: 14px;


lighthouse-core/report/styles/report.css
```diff
@@ -29,10 +29,10 @@ span, div, p, section, header, h1, h2, li, ul {
...
   --secondary-text-color: #606060;
   --accent-color: #3879d9;
-  --poor-color: #e81e1b;
-  --good-color: #008a07;
-  --average-color: #9f6d00;
-  --warning-color: #947200;
+  --poor-color: #e53935; /* md red 600 */
+  --good-color: #43a047; /* md green 600 */
+  --average-color: #ef6c00; /* md orange 800 */
+  --warning-color: #757575; /* md grey 600 */
```
paulirish: In isolation this looks a bit odd, that a warning icon is done in gray.
I'm presuming you're okay with that because any warning debugString is displayed in red?

Personally I prefer a yellow/orange/mustard color for these error states, but I can live with this just fine. :)

ebidel: Right. My thinking was that colors should signify level of passing. Since we moved the warning icon to orange, it's felt like mixed messaging. e.g. orange for an audit error and orange for a 50/100 audit is kinda confusing. If anything, it might make more sense to make the failures red so users know they contribute 0 to the score.

paulirish: okay. i'm feeling like we'll soon graduate to sorting all results into three buckets:  exceptions/failures/passing. 
Once we do that, we can be very clear about how things bubble up to a score.
Also it generally seems weird to provide a score if there were exceptions in that section.

paulirish: In the meantime.. I'm ambivalent. :)

ebidel: > seems weird to provide a score if there were exceptions in that section.

Definitely need to shake that out. @brendankenny had similar feelings when putting in the error stuff.

robdodson:
image

XhmikosR:
Personally I see this doesn't pass WCAG2AA for at least `--good-color` which I can test. My recommendation would be `#2b882f`.

PR #1759, opened 2017-02-22
feat: add total byte weight audit

### LINKED ISSUE #584: Audit: total byte weight
From the I/O and PWA summit talks I watched, I saw that total transfered size has a cost and is a big problem in Africa and other regions where data is expensive.

> https://whatdoesmysitecost.com/

I also remember that when [Google+ did its redisign](https://developers.google.com/web/showcase/2015/googleplus), they talked a lot about their budget target : 

> We hit our goal of never downloading more than 60k of HTML, 60k of JavaScript and 60k of CSS at any one time!
1. Would this kind of audit be interesting and relevant?
2. What kind of scoring could we propose?
3. Should this be part of the **Performance Metrics** or is this something new which could group stuffs related to caching strategies (HTTP cache, SW cache...)?

This is maybe where configuration will help since the kind of business and target of the PWA have influence over the budget target.

patrickhulce:
fixes #584 

I'm not the biggest fan of this implementation (primarily because it feels like a rather arbitrary cutoff and not as actionable as the other audits) but thought it was a good place to start to gather feedback and ideas.

image


lighthouse-core/audits/byte-efficiency/total-byte-weight.js
```diff
@@ -0,0 +1,82 @@
...
+ /**
+  * @fileoverview
+  */
+'use strict';
+
+const Audit = require('./byte-efficiency-audit');
+const URL = require('../../lib/url-shim');
+
+// ~95th percentile http://httparchive.org/interesting.php?a=All&l=Feb%201%202017&s=All#bytesTotal
+const FAILURE_THRESHOLD_IN_BYTES = 5 * 1024 * 1024; // 5MB
```
patrickhulce: In practice this also seemed to pass on image heavy gallery sites assuming they made reasonable attempts to optimize the images, which gives me a little more confidence that it's not overly noisy.

ebidel: Haven't looked at what it is, but should we use the mobile stat instead of the desktop one?

ebidel:
WDYT about making this a [scored audit](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/dobetterweb/dom-size.js#L81-L90)? We could create a curve based on the httparchive numbers.

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
patrickhulce: It doesn't change the result very much.
5MB is ~94th on desktop ~96th on mobile and the roundedness of it is a nice bonus.

patrickhulce:
> WDYT about making this a scored audit?

I'm game, but I am a little hesitant because graphic heavy sites have legitimate reasons that they're in the higher percentiles. Curving between say 90 and 95 doesn't seem that useful either. What about start dropping off around 75th percentile, median at 90th, hitting 0 at 99th?

ebidel:
> have legitimate reasons that they're in the higher percentiles

Certainly. But isn't the goal of this audit to raise awareness to those byte-heavy sites; mention they _could_ be doing better? I don't care who you are, a 5MB on page load is too much :) Crazy that that's the median!

You'll definitely want to tweak the curve to make a decent score obtainable.

patrickhulce:
OK I added scoring and based the 100 target based on the maximum bytes on what one could conceivably download with 3G throttling and the 10s TTI goal. (1.5Mbps * 10s ~= 1830 KB)
[see chart](https://www.desmos.com/calculator/gpmjeykbwr) Also added a link to [whatdoesmysitecost](https://whatdoesmysitecost.com/)

1600 KB = 100
4000 KB = 50
5000 KB = 24

image

patrickhulce:
PTAL :)

Issue #970, opened 2016-11-18
Error: Unable to fetch webSocketDebuggerUrl with headless_shell


justinribeiro:
I realize this is on the edge, but currently when running `lighthouse-cli` or using lighthouse-core in custom scripts, `cri.js` will throw an error and exit when running against the `headless_shell` built from the tip (as of Nov 17):

```
status Disconnecting from browser... +0ms
Runtime error encountered: Error: Unable to fetch webSocketDebuggerUrl, status: 500
    at IncomingMessage.response.on._ (/usr/local/lib/node_modules/lighthouse/lighthouse-core/gather/connections/cri.js:79:18)
    at emitNone (events.js:91:20)
    at IncomingMessage.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

This is caused by `CriConnection connect()` method calling `/json/new`. In this case, `headless_shell` will return "Cannot create new page" as the `/json/new` API does not exist in `headless_shell` as per https://bugs.chromium.org/p/chromium/issues/detail?id=626847.

My question: leave the existing behavior and wait for 626847 to add the the backwards compatibility for `/json/new` or handle this in the new `Browser.createTarget` DevTools command?

Note: if anyone is reading this and you still want to run against `headless_shell` now, you can temporarily peg to 1.1.6 before the `Connection` refactor occurred.

brendankenny:
@pavelfeldman if you have any advice.

@justinribeiro do you know how [chrome-remote-interface](https://github.com/cyrus-and/chrome-remote-interface) does this that worked before? When I looked into it when Pavel changed to `Connection` in #800 it seemed like cri was doing the same http request to `/json/new`. I was just skimming the source, though, not stepping through it, so maybe I missed the actual execution path.

brendankenny:
maybe I'm just not understanding the `Browser.createTarget` command though (debugger protocol noob here :). How can we give a protocol command if we don't have a websocket to talk over yet?

justinribeiro:
@brendankenny `CriDriver` in 1.1.6 used ` chromeRemoteInterface.New()` and only warned on the inability to crete a new tab and re-used instead, which worked fine with headless_shell.

[...]

PR #1760, opened 2017-02-22
Reuse existing tab if creating new tab fails

paulirish:
Only happens in the CLI case, so far, as thats the only place we spawn a new tab.

Fixes https://github.com/GoogleChrome/lighthouse/issues/970

In this PR:

* We move most of the impl of `CriConnection.connect` into `_connectToSocket`
* We introduce a fallback mode if the `this._runJsonCommand('new')` rejects, where we reuse the first tab available. 
* We will make sure our chosen tab is in the foreground, as well.


paulirish:
This fixes the only major issue with headless chromium. \o/
@patrickhulce can you take a look tmw?

Issue #1443, opened 2017-01-09
Add --chrome-flags to CLI


paulirish:
Usecase: 

```sh
lighthouse --chrome-flags="--ignore-certificate-errors" https://localbuild.dev/site/
```

(Broken out from #1416)

brendankenny:
and (from other discussions), preferably we'd have some way of also passing in the sub-sub-flags to v8 via `--js-flags="--trace_deopt --etc"`.

Not sure if it should be a separate `--js-flags` flag or if it should be nested inside `--chrome-flags` so the value could just be passed to chrome directly (nested quotes might get weird, though).

paulirish:
sgtm!  I'd prefer we try the nested quotes route first, otherwise lighthouse will have to eat some complexity.

paulirish:
The current alternative to this is running the `chrome-debug` binary, which lighthouse adds when you do a global install. It can take additional flags..

```sh
chrome-debug --show-paint-rects
```

then run lighthouse, however you like.

Lighthouse will automatically use the chrome-debug instance. as it's got port 9222 open to connect to.

paulirish:
fixed by #1761

PR #1761, opened 2017-02-22
Add --chrome-flags option to CLI. (#1443)

tommycli:


paulirish:
@tommycli for whatever reason I'm seeing some failures in our integration tests after this PR:

https://travis-ci.org/GoogleChrome/lighthouse/jobs/204045562
https://travis-ci.org/GoogleChrome/lighthouse/jobs/204045561

The error is coming from https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/test/smokehouse/smokehouse.js#L92 with invalid JSON. So seemingly the smokehouse test is returning something that's not JSON on stdout despite the --json flag being set.

It's not immediately clear why your PR is causing this, but we'll need to sort that out.

paulirish:
@tommycli ah. it's the `flags.chromeFlags.split(' ')` 
if it was not specified, `flags.chromeFlags` will be undefined and this'll throw.

a `default('chrome-flags', '')` might make sense, but you could probably handle this other ways.

paulirish:
@tommycli do you want to finish this up?

paulirish:
ayup. lgtm! thanks

PR #1763, opened 2017-02-22
Add explicit strict null checks in.

samccone:


lighthouse-cli/bin.ts
```diff
@@ -276,11 +276,11 @@ function saveResults(results: Results,
...
 
-function runLighthouse(url: string,
+export function runLighthouse(url: string,
                        flags: {port: number, skipAutolaunch: boolean, selectChrome: boolean, output: any,
                          outputPath: string, interactive: boolean, saveArtifacts: boolean, saveAssets: boolean
                          maxWaitForLoad: number},
-                       config: Object): Promise<undefined> {
+                       config: Object | null): Promise<undefined> {
 
   let chromeLauncher: ChromeLauncher;
```
patrickhulce: why is this declaration allowed by TS when we know it starts off as undefined :/

lighthouse-cli/chrome-launcher.ts
```diff
@@ -31,17 +31,17 @@ const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
 const isWindows = process.platform === 'win32';
 
-class ChromeLauncher {
+export class ChromeLauncher {
```
patrickhulce: just learning here, what difference does this make to TS as opposed to `export {ChromeLauncher}`

lighthouse-cli/chrome-launcher.ts
```diff
@@ -221,7 +221,7 @@ class ChromeLauncher {
     });
   }
 
-  kill(): Promise<undefined> {
+  kill(): Promise<{}> {
```
patrickhulce: Interesting that these are not `Promise<void>`

lighthouse-cli/chrome-launcher.ts
samccone: none

patrickhulce: oh so just style?

lighthouse-cli/bin.ts
samccone: if we know it starts out  as undefined we should set it to config?: Object

It is optionally null since when we kick of
`return runLighthouse(url, cliFlags, config)`
config can be (is?) null
so we are assigning it to null.

/shrug

lighthouse-cli/chrome-launcher.ts
samccone: yeah just rolling it up a tiny bit

lighthouse-cli/bin.ts
patrickhulce: Oh I'm referring to the bug in typescript allowing me to declare `let chromeLauncher: ChromeLauncher;` when TS knows perfectly well that `undefined`, it's initial value, is not a valid member of that type. I want it to force me to declare `ChromeLauncher | undefined`

paulirish:
I'm interested in these improvements. however i will point out this PR was precipitated by the fact that we couldn't catch the error in #1761 (we dont check for `chromelauncher` before calling `chromelauncher.kill()`)

it appears that TS can't handle the promise flow here. a quick look and changing it to async/await does allow TS to spot the bug.  i believe @samccone is trying that out.

lighthouse-cli/bin.ts
```diff
@@ -276,38 +276,36 @@ function saveResults(results: Results,
...
+                       config: Object | null): Promise<{}|void> {
 
-      return promise;
-    })
-    .then(() => chromeLauncher.kill())
-    .catch(err => {
-      if (typeof chromeLauncher !== 'undefined') {
-        return chromeLauncher.kill().then(() => handleError(err), handleError);
-      }
+  let chromeLauncher: ChromeLauncher | undefined = undefined;
```
paulirish: lol.

lighthouse-cli/bin.ts
```diff
@@ -276,38 +276,36 @@ function saveResults(results: Results,
...
+
+    await saveResults(results, artifacts!, flags);
+    if (flags.interactive) {
+      await performanceXServer.hostExperiment({url, flags, config}, results);
+    }
+
+    return await chromeLauncher.kill();
+  } catch (err) {
+    if (typeof chromeLauncher !== 'undefined') {
+      await chromeLauncher!.kill();
```
paulirish: can you explain the `!`? new to me.

samccone: Non-null assertion operator
---

A new ! post-fix expression operator may be used to assert that its operand is non-null and non-undefined in contexts where the type checker is unable to conclude that fact. Specifically, the operation x! produces a value of the type of x with null and undefined excluded. Similar to type assertions of the forms <T>x and x as T, the ! non-null assertion operator is simply removed in the emitted JavaScript code.

lighthouse-cli/bin.ts
samccone: it is the truth

lighthouse-cli/bin.ts
samccone: and we need these due to a bug in typescript which I will file.

patrickhulce: https://github.com/Microsoft/TypeScript/issues/13884

Issue #1739, opened 2017-02-17
Is it time to deprecate the CLI report output?


paulirish:
Personally I won't try to read the results of the CLI output and I always reach for the HTML report.

Maybe that should be the default setup?

Thoughts?

ebidel:
+💯 

Personally, I never use the prettyprint CLI and find it cumbersome to author and maintain audits.

Another positive is that it would remove a BUNCH of code and tests.

patrickhulce:
+1
Also is a giant pain for trying to read the log output, maybe bring back `--view` as a way to immediately open the report without bringing up a local server?

ebidel:
Has anyone started this? I can take some/all of the audits.

paulirish:
I guess i technically already deprecated output=pretty. (Did so by making it non-default.)

One can still use it.. I'm thinking the CI scenario is the most likely for it to be valuable.

I'm not 100% on removing _all_ support for pretty at this point, though I do think it'd be fine to stop investing in our pretty formatters (table, crit path, etc). Those aren't worth it for us to maintain (or keep).

Issue #1714, opened 2017-02-13
WPT compat: Support multiple output formats


patrickhulce:
Need both HTML and JSON outputs, support for comma-delimited outputs would be nice.

PR #1764, opened 2017-02-23
Add --view flag. Don't default CLI report to stdout.

paulirish:
First step in closing #1739 (Is it time to deprecate the CLI report output?)

1. Adds a `--view` flag which will open the HTML report that the CLI generated in your default browser.
2. Default `output` changes from `pretty` to `none`, which means you dont get the full report flooding your terminal at the end. You _can_ re-enable it with `--output=pretty`, however. 
3. Log two little tips to stdout telling the CLI users about the change.

image

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -35,6 +35,7 @@ import * as Printer from './printer';
 import * as randomPort from './random-port';
 import {Results} from './types/types';
 const yargs = require('yargs');
+const opn = require('opn');
 
 interface LighthouseError extends Error {
   code?: string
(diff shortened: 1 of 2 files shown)


lighthouse-cli/bin.ts
```diff
@@ -269,8 +273,15 @@ function saveResults(results: Results,
       promise = promise.then(_ => assetSaver.saveAssets(artifacts, results.audits, resolvedPath));
     }
 
-    if (flags.output === Printer.OutputMode[Printer.OutputMode.pretty]) {
-      promise = promise.then(_ => Printer.write(results, 'html', `${resolvedPath}.report.html`));
+    if (flags.output === Printer.OutputMode[Printer.OutputMode.none]) {
```
patrickhulce: Unfortunate that one cannot get both pretty output and HTML output now, but we're killing it entirely and #1714 captures multiple output desires anyway

paulirish: true.

XhmikosR:
@paulirish: is it expected that now if `output=pretty` is specified, the report.html isn't being saved anymore?

patrickhulce:
@XhmikosR yes that was a consequence of this change as noted in https://github.com/GoogleChrome/lighthouse/pull/1764#discussion_r102771988

XhmikosR:
Ah, I see thanks. Not a big deal, but took me a while to see why it stopped to work. :P

PR #1765, opened 2017-02-23
Reduce build size (via babel instead)

paulirish:
Since I [keep going on about this](https://github.com/GoogleChrome/lighthouse/pull/1756#issuecomment-281544651), I figured I'd give it a try. This is a PR against @ebidel's open PR (#1756)

Very conservative settings for the minification. And luckily, no problems with the build at all, so far. 

I also retain the lines, which makes for happier stack traces. This costs about 100K, but I think it's worth it. Kinda funny, but here's what it looks like: [screenshot](https://cloud.githubusercontent.com/assets/39191/23244855/ff8204ce-f93c-11e6-88a9-7948424f2205.png)

The filesize breakdown:

| | size | 
|-|-|
| original | 2,537,217 bytes |
| uglify (1756) | 1,725,484 bytes (32% smaller) |
| babel (this pr) | 1,828,032 bytes  (28% smaller)  |
| babel w/o lines | 1,775,509 bytes  (30% smaller)  |

so @ebidel wdyt?

### DIFF
--- a/lighthouse-extension/gulpfile.js
+++ b/lighthouse-extension/gulpfile.js
@@ -12,8 +12,7 @@ const chromeManifest = require('gulp-chrome-manifest');
 const debug = require('gulp-debug');
 const eslint = require('gulp-eslint');
 const livereload = require('gulp-livereload');
-const uglifyJsHarmony = require('uglify-js-harmony');
-const uglify = require('gulp-uglify/minifier');
+const babel = require('babel-core');
 const tap = require('gulp-tap');
 const zip = require('gulp-zip');
 const LighthouseRunner = require('../lighthouse-core/runner');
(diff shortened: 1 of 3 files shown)


ebidel:
Yea looks good for no errors. Does this avoid transpilation? Couldn't find much info using babel just as a minifier.

paulirish:
> Does this avoid transpilation? Couldn't find much info using babel just as a minifier.

more or less. there are no plugins or presets so it doesn't work the AST anymore than it has to.

It will go source -> AST -> (walking AST to remove comments) -> codegen (without whitespace) -> binary

Technically I suppose that would be a transpilation, but since babel is opt-in for every transformation, nothing else is happening in the source beyond what we say.

ebidel:
Nice. I didn't realize you could give it no transforms and just run options. I'm cool with using this instead. Looks like we can squeeze more out too and eventually remove the inline comments.

PR #1771, opened 2017-02-24
Compact license header

paulirish:
This is motivated by wanting fewer than 16 lines of boilerplate at the top of every source file. :)

I have reviewed all internal licensing requirements (go/releasing/preparing) and this is the best we can do. The Golang project (and many projects in Go) use a very compact header. ([example](https://github.com/GoogleCloudPlatform/golang-samples/blob/e91e4141e3d689498a9989d2b88308d6ed2f027b/kms/crypter/crypter.go#L1-L3)) however while this shortform boilerplate is permitted for BSD and MIT, Apache requires all the text.

So essentially, we nuke all the newlines. :) 
And we drop from 16 total lines down to 5. 📉  sgtm!

As a start, this PR just changs a single source file header. If it's sgtm, then I will update _all_ files.

### DIFF
--- a/lighthouse-core/index.js
+++ b/lighthouse-core/index.js
@@ -1,18 +1,7 @@
 /**
- * @license
- * Copyright 2016 Google Inc. All rights reserved.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * @license Copyright 2016 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


patrickhulce:
> Apache requires all the text.

:( I don't have strong feelings. I don't like many lines, but I've started to ignore it and I also don't like really long lines, haha.

ebidel:
Is the goal to make the build smaller? Personally, I'd prefer keeping the full boilerplate in source and removing/re-adding at build time. We can use the gulp-license with `tiny: false` to re-add it to the bundle.

It's also worth noting the Polymer teams gets away with a minimal header and links to the full license ([example](https://github.com/Polymer/polymer/blob/master/src/standard/utils.html#L1-L9)). WDYT of that?

@tjsavage how was that done?

paulirish:
> Is the goal to make the build smaller?

nah that's not what i was targeting. (sorry for not being clear on that!)

its to **improve the developer story.** less scrolling. higher signal:noise ratio of the sourcebase.  since our implementations will be increasingly reviewed by folks (who aren't contributing), it'd be nice to improve the visual accessibility of the code.

> gulp-license 

Also from my research on this topic it looks like we probably shouldn't be using `tiny:true` in the built version. :/  Though TBH I didn't find any specific conversations of deployed JS code, as the constraints there are slightly different.

>  Polymer teams gets away with a minimal header

Yeah projects licensed under BSD and MIT (polymer is bsd) can get away with the shortform: https://goto.google.com/ooiik  (internal link)
And someone asked specifically for a more shortform for Apache but the guidance was to go long: https://goto.google.com/gxiub (internal). (That form would have def been ideal for me, but oh well. :/  )

brendankenny:
I breeze right over the license header as well, so it doesn't bother me. I've never really heard of anyone complaining about it before, actually :)

I will say that often when I look at other people's source code just the *shape* of the license header is a good first check of what license they're using, so it is a shame to lose that Apache 2.0 shape.

wardpeet:
I don't have any issues with the large license either. More scrolling is extra exercise 💪

PR #1777, opened 2017-02-25
Rewrite chrome URLs to compare them.

### LINKED ISSUE #1776: GatherRunner:error timeout reached for chrome://settings
When testing Chrome tip of tree with --chrome-flags="--disable-features=MaterialDesignSettings", or testing Stable without the above flag (it's disabled by default on Stable), chrome://settings fails to run.

This is with lighthouse also on master/1.5.1.

Full command:

```sh
lighthouse --chrome-flags="--disable-features=MaterialDesignSettings" chrome://settings
```

paulirish:
fixes #1776

### DIFF
--- a/lighthouse-core/lib/url-shim.js
+++ b/lighthouse-core/lib/url-shim.js
@@ -95,12 +95,11 @@ URL.getDisplayName = function getDisplayName(url) {
 };
 
 // There is fancy URL rewriting logic for the chrome://settings page that we need to work around.
-// Why? It was to be able to swap from chrome://settings to chrome://extensions with a JavaScript 
-// pushState call without forcing a page reload. :)
-// As a result, the final document URL doesn't match the URL that network layer surfaces.
-function rewriteChromeInternalUrl(url){
+// Why? Special handling was added by Chrome team to allow a pushState transition between chrome:// pages.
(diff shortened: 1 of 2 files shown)


paulirish:
@tommycli can you review?

tommycli:
Btw, I don't think I'm a valid reviewer.

lighthouse-core/lib/url-shim.js
```diff
@@ -94,13 +94,24 @@ URL.getDisplayName = function getDisplayName(url) {
   return name;
 };
 
+// There is fancy URL rewriting logic for the chrome://settings page that we need to work around.
+// Why? It was to be able to swap from chrome://settings to chrome://extensions with a JavaScript
+// pushState call without forcing a page reload. :)
+// As a result, the final document URL doesn't match the URL that network layer surfaces.
+function rewriteChromeInternalUrl(url) {
+  if (!url.startsWith('chrome://')) return url;
+  return url.replace(/^chrome:\/\/chrome\//, 'chrome://');
```
ebidel: it't not clear from your description that it uses `chrome:chrome` urls. Might want to update the summary.

paulirish: good call. you got it.

Issue #820, opened 2016-10-24
Calibration for cpu throttling


wardpeet:
When running lighthouse in emulation mode we have static values for cpu throttling and network throttling.
CPU throttling has the biggest bottleneck because it is dependend on system hardware. (5x slower)

It would be nice if we could add a calibration service that we can compare our results too.

Related issues:
[Issue #33](https://github.com/GoogleChrome/lighthouse/issues/33)
[PR #747](https://github.com/GoogleChrome/lighthouse/pull/747)

pedro93:
Hello,

Just wanted to say something regarding calibration runs. I have been doing performance audits at work and noticed a large variability on performance-related metrics between "simulated runs" and real devices. 

As a suggestion/idea perhaps it would be best to have something like 10 mobile simulated lighthouse runs and compare the average with 10 real mobile devices runs using the remote debugging protocol (if a user has one) and adjust the throttling rates (both CPU and network) to each user's hardware?

At least at my work, our mid level device is not a nexus 5X and it would be great if we could adjust the throttling to our case (without needing the device after the calibration) and even pass those values to our CD servers for continuous performance analysis.

paulirish:
@pedro93 just curious.. is it possible to share your data? privately would be fine.

we're also building out some tools of our own to measure the variability but it's not operational just yet.

Your idea is right on point, by the way. :)

paulirish:
research by @wwwillchen  indicated that this isn't necessary.  http://go/devtools-cpu-throttling

paulirish:
Somewhat related: speedcurve is using some estimates for their CPU throttling presets. The data is based off [Geekbench](https://www.geekbench.com/) values, which may or may not represent browser performance. :) http://support.speedcurve.com/synthetic-settings/cpu-throttle

paulirish:
We now think it's worthwhile to do this for lantern runs.
Breaking this out from #4521

patrickhulce:
This was deeply investigated and concluded with #9085. See that issue for more discussion.

PR #747, opened 2016-10-05
Add cpu throttle

wardpeet:

PR #1778, opened 2017-02-25
Enable CPU throttling (4.5x)

addyosmani:
This change enables CPU throttling by default and switches the slow-down from 5x -> 3.5x (as discussed with @paulirish in person). The 3.5x aligns closer with the CPU differences observed between a MBP and a Moto G4 (our current recommendation for 'average' mobile hardware).

screen shot 2017-02-25 at 12 13 46 am

If there are any further changes required to enable this, please feel free to 👋


addyosmani:
cc @wardpeet in case I'm missing anything.

ebidel:
@addyosmani how much closer does this take us to aligning with WPT results? Curious if you had time to run the same site against WPT and this PR?

lighthouse-core/gather/gather-runner.js
```diff
@@ -313,9 +313,8 @@ class GatherRunner {
       return Promise.reject(new Error('You must provide a config'));
     }
 
-    // CPU throttling is temporarily off by default
     if (typeof options.flags.disableCpuThrottling === 'undefined') {
```
paulirish: Afaik you can nuke this conditional

lighthouse-extension/app/src/popup.js
```diff
@@ -124,7 +124,7 @@ function onGenerateReportButtonClick(background, selectedAggregations) {
 
   background.runLighthouseInExtension({
     flags: {
```
paulirish: You can drop this flags object

paulirish:
Oh I think there's a note about mobile testing in the readme that can also use a tweak.

addyosmani:
@paulirish Woo. Thanks for the review! I'll get those three bits dropped shortly.

@ebidel Good questions :) The 3.5x number came from locally comparing traces seen on a Moto G4 over remote debugging vs LH on desktop.  Talking to Pat this week, he seemed supportive of the change getting us 90% of the way there in the short term.

I'm going to run some more experiments as a sanity check and report back here.

addyosmani:
@paulirish relevant to the multiplier: do we have any UMA data on the type of desktop/hardware the average DevTools user is on? Just want a sanity check that the MBP is a fair baseline to assume.

wardpeet:
Looks nice 👍 

I'm adding referring this #820 and #747 where we had a previous discussion.

addyosmani:
After trying to look at calibration for a while, I'm updating my recommendation to CPU throttling at 4.5x and tweaking that over time. @paulirish could you let me know if you'd be down with that and I'll update my PR?

Here are a few comparisons between WPT runs over a Moto G4 with 3G and DevTools using a 4.3-4.5x CPU throttling + custom network throttling closer to what WPT is doing:

Polymer Shop - closer with 4.5x CPU slowdown

image

ReactHN - closer with a 4.5x CPU throttling slowdown

image

Fashion United - close..ish with 4.5x CPU slowdown
image

Voice Memos with 4.5x

image

Code.gov with 4.3x CPU slowdown

image

Twitter.com (outlier) at 4x CPU (~1000ms slower)

image

addyosmani:
@paulirish I believe I've addressed your feedback and rebased against master. PTAL whenever you get a chance.

lighthouse-cli/bin.ts
```diff
@@ -124,8 +124,7 @@ Example: --output-path=./lighthouse-results.html`,
   // default values
   .default('chrome-flags', '')
   .default('disable-cpu-throttling', false)
-  .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.none])
-  .default('output-path', 'stdout')
```
paulirish: i don't think you meant to change this, right?

lighthouse-cli/bin.ts
```diff
@@ -124,8 +124,7 @@ Example: --output-path=./lighthouse-results.html`,
   // default values
   .default('chrome-flags', '')
   .default('disable-cpu-throttling', false)
```
paulirish: false => true ?

PR #1785, opened 2017-02-27
fix: deflake TTI audit by extending trace

### LINKED ISSUE #1784: TTI Audit fails on sites that do an amazing job
Because we look for a 500ms window after fMP where the main thread isn't too busy, sites that do [an incredible job](https://preact-pwa.appspot-preview.com/) of loading quickly can fail the TTI audit (5/20 times on my machine) because the trace was closed too quickly. The effect will be even more extreme for those running on localhost without throttling.

Naive solution: set a 500ms timeout before closing the trace to ensure there's enough time at the end for the window to exist.

Example trace (there's 488ms after first paint): https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=https://dl.dropboxusercontent.com/u/54768288/preact-pwa.appspot-preview.com_2017-02-27_10-19-53-0.trace.json

patrickhulce:
R: @paulirish all

fixes #1784

### DIFF
--- a/lighthouse-core/config/default.json
+++ b/lighthouse-core/config/default.json
@@ -3,7 +3,7 @@
     "passName": "defaultPass",
     "recordNetwork": true,
     "recordTrace": true,
-    "pauseAfterTraceMs": 500,
+    "pauseBeforeTraceEndMs": 500,
     "useThrottling": true,
     "gatherers": [
       "url",
--- a/lighthouse-core/config/default.json
+++ b/lighthouse-core/config/default.json
@@ -3,6 +3,7 @@
     "passName": "defaultPass",
     "recordNetwork": true,
     "recordTrace": true,
+    "pauseAfterTraceMs": 500,
     "useThrottling": true,
     "gatherers": [
       "url",
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -599,7 +599,7 @@ class Driver {
   /**
    * @param {number=} pauseAfterTraceMs Wait this many milliseconds before ending the trace
    */
-  endTrace(pauseAfterTraceMs = 500) {
+  endTrace(pauseBeforeTraceEndMs = 0) {
     return new Promise((resolve, reject) => {
       // When the tracing has ended this will fire with a stream handle.
       this.once('Tracing.tracingComplete', streamHandle => {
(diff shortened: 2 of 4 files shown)


lighthouse-core/config/default.json
```diff
@@ -3,6 +3,7 @@
     "passName": "defaultPass",
     "recordNetwork": true,
     "recordTrace": true,
+    "pauseAfterTraceMs": 500,
```
paulirish: the pause is previous to ending the trace, so i'd think we'd call it something like`recordTraceEndDelay`. wdyt?

lighthouse-core/gather/driver.js
```diff
@@ -596,7 +596,10 @@ class Driver {
       .then(_ => this.sendCommand('Tracing.start', tracingOpts));
   }
 
-  endTrace() {
+  /**
+   * @param {number=} pauseAfterTraceMs Wait this many milliseconds before ending the trace
+   */
+  endTrace(pauseAfterTraceMs = 500) {
```
paulirish: dont you want a default of 0, since you're setting the config specifically for this pass?

lighthouse-core/gather/driver.js
```diff
@@ -606,7 +609,7 @@ class Driver {
       });
 
       // Issue the command to stop tracing.
-      this.sendCommand('Tracing.end').catch(reject);
+      setTimeout(() => this.sendCommand('Tracing.end').catch(reject), pauseAfterTraceMs);
```
paulirish: let's just remind ourselves why this was neccessary. :) something like..

> Stop tracing, after an optional delay. Audits like TTI may require having a slightly longer trace

lighthouse-core/config/default.json
patrickhulce: Yeah true, just was very similar in spirit to the pauseAfterLoadMs so wanted to be consistent. How about `pauseBeforeTraceEnd`? Nix the `Ms` either way?

lighthouse-core/gather/driver.js
patrickhulce: I guess depends on if we want previous custom configs to be opt-in or opt-out of the fix. The myriad of ways to run lighthouse gets thorny with all these options :) I'm fine with 0 too since I'm guessing custom config usage is quite, quite low

lighthouse-core/gather/driver.js
patrickhulce: Will do 👍

lighthouse-core/config/default.json
paulirish: yah. i actually wrote that first. :) and then revised so it looks more similar to `recordTrace` since they're buddies

but yeah let's go with pauseBeforeTraceEnd since we both went there.

lighthouse-core/gather/driver.js
paulirish: yah. let's do 0. 

but since you mentioned it. can you add pauseBeforeTraceEnd into the perf.json as well? :)

outside of that one, custom configs are pretty rare yeah. so far we have no evidence anyone other than us are using them.

patrickhulce:
> would like the addition to perf.json

done

PR #1796, opened 2017-02-28
fixing the css bug for misaligned score badges on Firefox

### LINKED ISSUE #1762: Report: misaligned score badges on Firefox
I'm using Firefox 52.0 beta. See the attached screenshot:

screen shot 2017-02-22 at 10 32 40-fullpage

The issue seems to come from `.aggregation__header :-moz-any(.aggregation__header__score)` rule.

narendrashetty:
This PR fixes #1762


XhmikosR:
Why don't just use `margin-top: 2px`? Seems simpler to me.

narendrashetty:
@XhmikosR It's just the way I write. Whenever it is `absolute` positioned I prefer to use `top` and `left` instead of `margin`'s. Moreover `top` affects only the icon and `margin` can affect other items around it.

If you think I should change, let me know. both of it does the same.

PR #1804, opened 2017-03-02
fix: loosen blocking tags audit

### LINKED ISSUE #1770: Don't warn about blocking links with very short delays
Brought over from discussion in https://github.com/GoogleChrome/lighthouse/issues/1317#issuecomment-282144792. We're unable to differentiate async stylesheets from blocking stylesheets, so don't warn if the time spent was below our threshold.

patrickhulce:
fixes #1770


lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -50,7 +52,10 @@ class LinkBlockingFirstPaintAudit extends Audit {
   static computeAuditResultForTags(artifacts, tagFilter) {
     const artifact = artifacts.TagsBlockingFirstPaint;
 
-    const filtered = artifact.filter(item => item.tag.tagName === tagFilter);
+    const filtered = artifact.filter(item => {
+      return item.tag.tagName === tagFilter &&
+        (item.endTime - item.startTime) * 1000 >= LOAD_THRESHOLD_IN_MS;
```
patrickhulce: there might be varying opinions on this, but I'm in favor of not failing for <100ms since we're still notifying them of their existence in critical path

open to only loosening for blocking stylesheets (or a different threshold) and keeping all scripts or something though

ebidel: 100ms does feel a little long, but that's just my gut. I know we're calling them out in the critical chain, but it would still be nice to catch these faster links here in this audit. Otherwise, users might get confused why it shows up in one place and not the other.

patrickhulce:
@brendankenny @paulirish any thoughts here?

brendankenny:
I don't really have any expert knowledge to add here. Who could we rope in?

paulirish:
I haven't finished looking at this but this reply from patrick was useful in helping me understand what this was about

image


regardless, we should probably add some comments to the code so someone reading it will know what this is about. :)

patrickhulce:
> but this reply from patrick was useful in helping me understand what this was about

lol, message received I'll add some explanations :)

patrickhulce:
Any other objections to this? Didn't expect it to be so untouchable :)

paulirish:
I can't really connect these two things. :/ I'm probably confused.

A) Us flagging link `rel=preload` resources are blocking.  Oops.
B) Excluding resources that take <1s to load (from start to finish) from the report.

Fixing A seems good.
Doing B seems good.

But I don't see how doing B fixes A, as `loadUnder1seconds` isn't indicative of preloaded and/or nonblocking.

patrickhulce:
> But I don't see how doing B fixes A, as loadUnder1seconds isn't indicative of preloaded and/or nonblocking.

@paulirish we do have code to detect A properly, but it's on a setInterval that has to be triggered *while the link is loading*. B fixes A because we just can't detect the async ones if they loaded too quickly and our interval doesn't load.

patrickhulce:
@paulirish lowered to 50ms, 👍 

for posterity: MutationObserver was somewhat of a chicken and egg problem with being able to find the link elements to observe (since they are added with parsing not mutations), finding the link elements is the only reason to use the setInterval in the first place!

patrickhulce:
soooo this works much better in the wild but still doesn't tell the full story, and actually is fairly flaky in smoke testing :/ because everything is delayed, our injected script also gets delayed meaning it's back to square one in our tests with a race condition of the async switching back to rel=stylesheet too quickly

Issue #909, opened 2016-11-05
Audits Meta: Feature parity with PSI, MFT, DevTools


ebidel:
_updated by @paulirish jan 2018_ 

### For parity with [Pagespeed Insights](https://developers.google.com/speed/pagespeed/insights/)

- [x] Eliminate render-blocking JavaScript and CSS in above-the-fold content ([docs](aaaaaaaa))
- ~Prioritize visible content~ ([docs](https://developers.google.com/speed/docs/insights/PrioritizeVisibleContent))
  - not pursuing the exact definition of the original audit
  - [ ] reduce size of above-the-fold content
  - [x] large off-screen images #1807
- [x] Leverage browser caching #3460 ([docs](https://developers.google.com/speed/docs/insights/LeverageBrowserCaching))
- [X] Enable compression ([docs](https://developers.google.com/speed/docs/insights/EnableCompression))  #1513
- [x] Minify CSS, Minify JavaScript #3459 ([docs](https://developers.google.com/speed/docs/insights/MinifyResources))
- ~Minify HTML~
    - concluded the value is too low to pursue
- [x] Reduce server response time ([docs](https://developers.google.com/speed/docs/insights/Server))
- [x] Avoid landing page redirects ([docs](https://developers.google.com/speed/docs/insights/AvoidRedirects))
- [ ] Content is sized correctly for viewport ([docs](https://developers.google.com/speed/docs/insights/SizeContentToViewport)) (#1836)

### For parity with [Mobile-Friendly Test](https://search.google.com/search-console/mobile-friendly):

- [ ] Tap targets ([docs](https://developers.google.com/speed/docs/insights/SizeTapTargetsAppropriately)) (#4358)
   - basic tap target size detection that PSI does
- [x] Legible font sizes ([docs](https://developers.google.com/speed/docs/insights/UseLegibleFontSizes))
   - [docs](https://developers.google.com/speed/docs/insights/UseLegibleFontSizes)
- [x] Optimize Images ([docs](https://developers.google.com/speed/docs/insights/OptimizeImages))
   - we want to calculate the latency caused by unoptimized images.
- [x] Avoid plugins ([docs](https://developers.google.com/speed/docs/insights/AvoidPlugins)))
  - issue: #101 
- [ ] avoid interstitials ([docs](https://developers.google.com/webmasters/mobile-sites/mobile-seo/common-mistakes#avoid-interstitials))
   - cleanroom implementation needed 

[...]

PR #821, opened 2016-10-24
[Proposal] Report downloaded but offscreen images

mixed:
## Motivation

When I have analyzed load performance using devtools Almost performance bottleneck was network(images...). So, I have checked invisible image and applied lazy load image. I think that This practice is common sense.
## How to make it?

I have filtered image that checked size, position. So, I think that LH(LightHouse) have to support options size and position of image. If I ran LH that I think LH have to suggest perf enhance point. So. LH support 2 features.
- support option (size, position)
- perf enhance point

I have made prototype. What do you think about this feature?

image

reference)
- [AMP Prioritize resource loading](https://www.ampproject.org/learn/how-amp-works/#prioritize-resource-loading)
- [Lazy loading image library in github](https://github.com/search?l=JavaScript&q=lazy+image+loading&ref=searchresults&type=Repositories&utf8=%E2%9C%93)
- [Lazy loading image article in google](https://www.google.co.kr/search?q=Lazy+loading+image&oq=Lazy+loading+image&aqs=chrome..69i57j0l5.1281j0j4&sourceid=chrome&ie=UTF-8)

PR #1807, opened 2017-03-03
feat: add OffscreenImages Audit

patrickhulce:
Continues the byte efficiency audits by finding unused images (images that are mostly or entirely outside the viewport). Broken into 2 commits, 1 that renames ContentWidth to ViewportDimensions to accommodate the addition of height and the other that adds the audit with tests.

Partially addresses above the fold content issue for PSI #909


lighthouse-core/audits/byte-efficiency/unused-images.js
```diff
@@ -0,0 +1,135 @@
...
+class UnusedImages extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Images',
+      name: 'unused-images',
+      description: 'Unused images',
+      helpText: 'Images that are not above the fold should be lazy loaded on interaction',
```
ebidel: learn more link to https://developers.google.com/web/updates/2016/04/intersectionobserver?

lighthouse-core/audits/byte-efficiency/unused-images.js
```diff
@@ -0,0 +1,135 @@
...
+
+const Audit = require('./byte-efficiency-audit');
+const URL = require('../../lib/url-shim');
+
+const ALLOWABLE_OFFSCREEN_X = 100;
+const ALLOWABLE_OFFSCREEN_Y = 200;
+
+const IGNORE_THRESHOLD_IN_BYTES = 2048;
+
+class UnusedImages extends Audit {
```
ebidel: would "OffscreenImages" make more sense? Or "AboveTheFoldImages"

lighthouse-core/audits/byte-efficiency/unused-images.js
```diff
@@ -0,0 +1,135 @@
...
+      category: 'Images',
+      name: 'unused-images',
+      description: 'Unused images',
+      helpText: 'Images that are not above the fold should be lazy loaded on interaction',
+      requiredArtifacts: ['ImageUsage', 'ViewportDimensions', 'networkRecords']
+    };
+  }
+
+  /**
+   * @param {{top: number, right: number, bottom: number, left: number}} imageRect
```
ebidel: nit: `{ClientRect}`

lighthouse-core/audits/byte-efficiency/unused-images.js
```diff
@@ -0,0 +1,135 @@
...
+   * @param {{scrollWidth: number, scrollHeight: number}} viewportDimensions
+   * @return {number}
+   */
+  static computeVisiblePixels(imageRect, viewportDimensions) {
+    const scrollWidth = viewportDimensions.scrollWidth;
```
ebidel: Not clientWidth/Height of the viewport? scrollWidth/Height includes off screen content.

lighthouse-core/audits/byte-efficiency/unused-images.js
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * @param {!Object} image
+   * @param {{scrollWidth: number, scrollHeight: number}} viewportDimensions
+   * @return {?Object}
+   */
+  static computeWaste(image, viewportDimensions) {
+    const url = URL.getDisplayName(image.src, {preserveQuery: true});
+    const totalPixels = image.clientWidth * image.clientHeight;
+    const visiblePixels = this.computeVisiblePixels(image.clientRect, viewportDimensions);
+    const wastedRatio = 1 - visiblePixels / totalPixels;
```
ebidel: One thought about this audit is that it would be really nice to use `IntersectionObserver` for finding the visible ratio of the image to the viewport. That API was designed for this use case. One problem with that approach is that the observer only fires if the image clientrect intersects with the viewports. So, for offscreen images, you'd never get a callback.

paulirish:
A different approach would be to use the trace and see what images we decoded. If we didn't decode an image yet, we probably didn't need to download it.

This would be slightly more resilient in case images move location during load.

---

Also, did you look at #821 ? How does this approach differ?

lighthouse-core/audits/byte-efficiency/unused-images.js
patrickhulce: yeah `OffscreenImages` sounds good

lighthouse-core/audits/byte-efficiency/unused-images.js
patrickhulce: The existing naming of `scroll` and `viewport` for inner and outer somewhat confused me so I'm up for renaming to just inner and outer, but I think I do want `inner*` (which is scroll here), right? Everything else gives me the size of the screen not what content is being displayed.

lighthouse-core/audits/byte-efficiency/unused-images.js
patrickhulce: True, I open to one that would do set subtraction but that seemed too specific to this audit to put in image usage and felt weird breaking out into a new one. It also meant the settings for failure (margin and so forth) were located in the gatherer rather than with the audit.

We have a spectrum of this throughout LH though, so I'm curious what other opinions are. I'm generally more in favor of keeping gatherers purely informational/more logic-less and having failure conditions measured in audits.

patrickhulce:
> Also, did you look at #821 ? How does this approach differ?

Similar overall approach, mostly just updated to leverage the byte efficiency framework and image gatherer that's been added since then. Also adjustments to use innerHeight/Width instead.

lighthouse-core/audits/byte-efficiency/unused-images.js
patrickhulce: will do

lighthouse-core/audits/byte-efficiency/unused-images.js
patrickhulce: done

lighthouse-core/gather/gatherers/viewport-dimensions.js
```diff
@@ -21,18 +21,21 @@ const Gatherer = require('./gatherer');
...
 /* istanbul ignore next */
-function getContentWidth() {
+function getViewportDimensions() {
   // window.innerWidth to get the scrollable size of the window (irrespective of zoom)
   // window.outerWidth to get the size of the visible area
   // window.devicePixelRatio to get ratio of logical pixels to physical pixels
   return Promise.resolve({
+    scrollPosition: {x: window.scrollX, y: window.scrollY},
     scrollWidth: window.innerWidth,
+    scrollHeight: window.innerHeight,
```
ebidel: Ah I see these now. Yea, I would switch these to use the same name as the property you're stashing.

I typically use `document.documentElement.clientWidth` to get the viewport width, but inner* works great. That's what you want.

lighthouse-core/audits/content-width.js
```diff
@@ -55,8 +55,8 @@ class ContentWidth extends Audit {
       return '';
     }
 
-    return 'The content scroll size is ' + artifact.scrollWidth + 'px, ' +
-        'whereas the viewport size is ' + artifact.viewportWidth + 'px.';
+    return 'The content scroll size is ' + artifact.innerWidth + 'px, ' +
+        'whereas the viewport size is ' + artifact.outerWidth + 'px.';
```
ebidel: Think this sentence needs updating. The first var is the viewport width and the second is
the size of the browser window. Right?  Not sure you want `outerWidth/Height`. That is inclusive of the window chrome. https://developer.mozilla.org/en-US/docs/Web/API/Window/outerWidth

For "content scroll size", don't you want `scrollWidth` of the content?

ebidel: I'm on a plane so sorry for the short replies.

patrickhulce: No worries, and I was just going with the "if it ain't broke don't fix it" approach, but I agree `documentElement.scrollWidth !== documentElement.clientWidth` makes the intention a lot more clear. In practice, outerWidth seems to reflect the non-scaled true width of the viewport and innerWidth seems to reflect the width of the content

ebidel: Sorry. I'm still caught up in lingo and intended behavior. Does http://jsbin.com/murusonore/edit?html,console,output capture what you're after? 

`document.documentElement.scrollWidth == content scroll size`
`window.innerWidth == visible viewport size`

(In the bin example, both of these metrics are < `window.outerWidth` because it's running inside the iframe...which is smaller than the viewport.)

----
Other refs:

> `innerWidth`: Width (in pixels) of the browser window viewport including, if rendered, the vertical scrollbar.

https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth

> `outerWidth` gets the width of the outside of the browser window. It represents the width of the whole browser window including sidebar (if expanded), window chrome and window resizing borders/handles. 

https://developer.mozilla.org/en-US/docs/Web/API/Window/outerWidth

patrickhulce: I'm not after anything other than to make the audit have more accurate names than it had before the rename :) I'm honestly not sure what this audit is getting after since we already have the `viewport` audit checking for the most obvious condition under which this one will fail.

Proposal: rename `scrollWidth` variable to `viewportWidth`, rename `viewportWidth` to `windowWidth` and update the description accordingly. Do not change comparison of innerWidth to outerWidth.

Aside, my previous example of `scrollWidth != clientWidth` is actually non-sensical I mistakenly assumed they meant something different on documentElement than regular elements.

ebidel: SGTM on the renames.

patrickhulce: Ok talked to Paul and found the motivation of this audit and it's just actually measuring something totally different. I'll clean that up in a future PR :) Will rename for now.

patrickhulce:
> Who knows what the audit does, but at least the output text is matching what the code is doing

lol, I filed #1836 and linked it in your PSI meta issue

lighthouse-core/gather/gatherers/viewport-dimensions.js
```diff
@@ -21,38 +21,44 @@ const Gatherer = require('./gatherer');
...
   afterPass(options) {
     const driver = options.driver;
 
-    return driver.evaluateAsync(`(${getContentWidth.toString()}())`)
+    return driver.evaluateAsync(`(${getViewportDimensions.toString()}())`)
 
     .then(returnedValue => {
-      if (!Number.isFinite(returnedValue.scrollWidth) ||
-          !Number.isFinite(returnedValue.viewportWidth) ||
+      if (!Number.isFinite(returnedValue.innerWidth) ||
```
brendankenny: time for something more succinct here? No `Object.values` yet, but could do like
```js
const allNumeric = Object.keys(returnedValue).every(key => Number.isFinite(returnedValue[key]));
if (!allNumeric) {
  // ...
}
```
?

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -0,0 +1,136 @@
...
+        return results;
+      }
+
+      const processed = OffscreenImages.computeWaste(image, viewportDimensions);
+      if (processed instanceof Error) {
+        debugString = processed.message;
+        return results;
+      }
+
+      // Don't warn about an image that was also used appropriately
```
brendankenny: what about something like, "if an image was used more than once, warn about the least wasteful usage of it"

patrickhulce:
> does it make sense to punish for images that are partly on screen? Slicing them would not be a maintainable solution for developers, so I'm not sure how actionable the advice for those images will be

That's mostly what the extra margin of error is for, but I'm down for only notifying only if a minimum percent offscreen is reached, maybe 75+?

> "Images that are not above the fold should be lazy loaded on interaction" - should they? What's the downside vs e.g. loading below the fold images after TTI?

Will reword, and we should probably  check to see if that's already the case...

lighthouse-core/audits/byte-efficiency/offscreen-images.js
patrickhulce: done

lighthouse-core/gather/gatherers/viewport-dimensions.js
patrickhulce: done

ebidel:
@patrickhulce this needs a rebase.

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -0,0 +1,146 @@
...
+      const existing = results.get(processed.preview.url);
+      if (!existing || existing.wastedBytes > processed.wastedBytes) {
+        results.set(processed.preview.url, processed);
+      }
+
+      return results;
+    }, new Map());
+
+    return TTIAudit.audit(artifacts).then(ttiResult => {
+      const tti = ttiResult.extendedInfo.value.timestamps.timeToInteractive / 1000000;
```
brendankenny: `ttiInSeconds` or something like that? Any reason you aren't just using `ttiResult.rawValue`?

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -0,0 +1,146 @@
...
+      const results = Array.from(resultsMap.values()).filter(item => {
+        const isWasteful = item.wastedBytes > IGNORE_THRESHOLD_IN_BYTES &&
+            item.wastedPercent > IGNORE_THRESHOLD_IN_PERCENT;
+        const loadedEarly = item.requestStartTime < tti;
```
brendankenny: should add an explicit comment here (or in description or on devsite? we'll have to doc it somewhere) that offscreen images loaded after TTI are considered fine

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -0,0 +1,143 @@
...
+  };
+  const image = {src, networkRecord, clientRect};
+  Object.assign(image, size);
+  return image;
+}
+
+describe('OffscreenImages audit', () => {
+  const _ttiAuditOriginal = TTIAudit.audit;
+  before(() => {
+    const timeToInteractive = 2000000;
```
brendankenny: `timeToInteractiveInμs`? :) took me a second and some looking at the TTI audit itself to realize this wasn't just chosen to be a really high TTI so it would be longer than any trace we normally take

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -0,0 +1,143 @@
...
+const UnusedImages =
+    require('../../../audits/byte-efficiency/offscreen-images.js');
+const TTIAudit = require('../../../audits/time-to-interactive.js');
+const assert = require('assert');
+
+/* eslint-env mocha */
+function generateRecord(resourceSizeInKb, startTime = 0, mimeType = 'image/png') {
+  return {
+    mimeType,
+    startTime,
```
brendankenny: mention this is in seconds?

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -0,0 +1,143 @@
...
+        generateImage(generateSize(1000, 1000), [1000, 1000], generateRecord(50)),
+        generateImage(generateSize(50, 50), [0, 1500], generateRecord(200), urlB),
+        generateImage(generateSize(400, 400), [0, 1500], generateRecord(90), urlB),
+      ],
+    }).then(auditResult => {
+      assert.equal(auditResult.results.length, 1);
+    });
+  });
+
+  it('disregards lazily loaded images', () => {
```
brendankenny: maybe explicit `'disregards images loaded after TTI'`?

brendankenny:
Hmm, do we need to bring in something to slow down the byte efficiency tester page? Or induce more screenshots with a spinner? Now that TTI is needed for it (through the offscreen images audit) it looks like that page is flakey now when no screenshots are captured.

```sh
✘ difference at offscreen-images.score: found null, expected false
full found result: {
  "score": null,
  "rawValue": null,
  "error": true,
  "debugString": "Audit error: No screenshots found in trace",
  "name": "offscreen-images",
  "description": "Offscreen images"
}
```

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -0,0 +1,146 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ /**
+  * @fileoverview Checks to see if images are displayed only outside of the viewport.
+  */
+'use strict';
+
+const Audit = require('./byte-efficiency-audit');
+const TTIAudit = require('../time-to-interactive');
```
patrickhulce: i'm with ya, @paulirish not a fan of getting these values into computed artifact though 😩

lighthouse-core/audits/byte-efficiency/offscreen-images.js
patrickhulce: raw value isn't what I want, I need the actual devtools timestamp value

lighthouse-core/audits/byte-efficiency/offscreen-images.js
patrickhulce: done

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
patrickhulce: done

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
patrickhulce: done

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
patrickhulce: done

patrickhulce:
> Hmm, do we need to bring in something to slow down the byte efficiency tester page? Or induce more screenshots with a spinner?

It's already being slowed to 5s as part of this PR and not sure why it's flaking on travis, would love to get a reproducible failure case to have more ammo against that bug though. 6/6 smoke test successes locally as I wrote this

brendankenny:
ok, so travis is happy on a few reruns, so lets ship it. Not sure how to restart the appveyor builds

PR #1809, opened 2017-03-03
tweak table styles

paulirish:
GIF showing before and after

tablescrunch

3 changes

* wider URL column. I think we can afford this width now.
* The overflow auto on table cells seemed really awkward, so i went for ellipsis instead
  * to be honest, the ellipsis probably isn't better. so i could revert that.
* no more (inactive) vertical scrollbar on the URL

### DIFF
--- a/lighthouse-core/formatters/partials/table.css
+++ b/lighthouse-core/formatters/partials/table.css
@@ -10,7 +10,8 @@
 }
 .table_list th,
 .table_list td {
-  overflow: auto;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 .table_list th {
   background-color: #eee;
@@ -34,6 +35,7 @@
   display: block;
   margin: 0;
   overflow-x: auto;
+  overflow-y: hidden;
 }
 .table_list em + code, .table_list em + pre {
   margin-top: 10px;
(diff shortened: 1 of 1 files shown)


ebidel:
For a reasonable viewport width (I'm currently at 1110px), there's a lot more information loss in some of the audits with the increased width and ellipsis `overflow:hidden`.

Now:

screen shot 2017-03-02 at 7 24 11 pm

screen shot 2017-03-02 at 7 24 38 pm

Compared to before:

screen shot 2017-03-02 at 7 25 02 pm

The important thing is that we give users a way to discover the whole picture. 

What we really need is a better way to customize the styling of each table. One size fits all isn't working out.

lighthouse-core/formatters/partials/table.css
```diff
@@ -51,6 +53,7 @@
   text-align: left;
   width: 250px;
   white-space: nowrap;
+  width: 440px;
```
patrickhulce: can we stick the increased width of a couple ranges in a few media queries? this is probably too big as eric points out when the screen is smaller

wardpeet:
why don't we enable responsive tables? so a horizontal scrollbar is present?

paulirish:
k. closing this for now. 
report redesign will give more width to these tables and we'll sort out more details then.

PR #1813, opened 2017-03-03
feat: support multiple output modes

### LINKED ISSUE #1714: WPT compat: Support multiple output formats

patrickhulce:
fixes #1714 

not sure what to do exactly about the path, the meaning of the `--output-path` extension starts to fall apart a bit when you have multiple files so figured I'd stick with the current paradigm.

USAGE:

`lighthouse --output html --output json --output-path foobar.html` generates
* foobar.report.html
* foobar.report.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ last-run-results.html
 *.devtoolslog.json
 *.screenshots.html
 *.report.html
+*.report.json
+*.report.pretty
 *.artifacts.log
 
 closure-error.log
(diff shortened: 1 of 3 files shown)


patrickhulce:
> CI didn't like this PR, it seems. :(

Yeah tsc not part of test or smokehouse and apparently it still outputs even when compilation fails...? fixed

Issue #1802, opened 2017-03-02
Extension Error: Cannot read property 'fatal' of null


bbaronSVK:
**Lighthouse Version**: 1.5.2
**Chrome Version**: 56.0.2924.87
**URL**: http://www.auto.cz/
**Error Message**: Cannot read property 'fatal' of null
**Stack Trace**:
 ```TypeError: Cannot read property 'fatal' of null
    at promise.catch.err (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:9971:14)```

brendankenny:
should be fixed by #1818

PR #1818, opened 2017-03-05
Fixes no rel noopener when properties of anchor element cannot be resolved anymore

wardpeet:
Fixes issue #1802. The problem is that [getObjectProperty](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L511) rejects a null value which bubbles down into the [recoverOrThrow](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/gather-runner.js#L128) function where `err = null`.

I fixed the bug by using the getAttribute function because href is also an attribute and is still available. I also caught the reject(null) from the `getObjectProperty` inside `getProperty` and resolve it with null so in the future other gatherers won't crash.

It's a pretty weird bug. Objects get collected but whenever you want to resolve the properties by `Runtime.getProperties` the object only has a `__proto__` property. My guess is that the object gets garbage collected or something and v8 can not resolve it's props anymore. Of course this is just a guess.

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
@@ -29,7 +29,7 @@ class AnchorsWithNoRelNoopener extends Gatherer {
       .then(failingNodeList => {
         const failingNodes = failingNodeList.map(node => {
           return Promise.all([
-            node.getAttributes('href'),
+            node.getAttribute('href'),
             node.getAttribute('rel'),
             node.getAttribute('target')
           ]);
--- a/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
@@ -29,7 +29,7 @@ class AnchorsWithNoRelNoopener extends Gatherer {
       .then(failingNodeList => {
         const failingNodes = failingNodeList.map(node => {
           return Promise.all([
-            node.getAttribute('href'),
+            node.getProperty('href'),
             node.getAttribute('rel'),
             node.getAttribute('target')
           ]);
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
```diff
@@ -29,7 +29,7 @@ class AnchorsWithNoRelNoopener extends Gatherer {
       .then(failingNodeList => {
         const failingNodes = failingNodeList.map(node => {
           return Promise.all([
-            node.getProperty('href'),
+            node.getAttributes('href'),
```
patrickhulce: `getAttribute`?

wardpeet:
mmh weird probably a cmd + s typo :)

lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
wardpeet: mmh weird probably a cmd + s typo :)

wardpeet:
Looks like getAttribute fails on smokehouse when href is not set, moving back to getProperty which works with this PR too but isn't always possible to fetch the href.

wardpeet:
looks like chromestatus is broken on chrome canary 59..

brendankenny:
hmm, what do you think about instead of fixing this in `element.getProperty()` we instead make `driver.getObjectProperty()` *resolve* with `null` instead of [rejecting with it](https://github.com/GoogleChrome/lighthouse/blob/d0f1f73999ba97339e0d0c78dc8be74f16bb3b60/lighthouse-core/gather/driver.js#L511) when a property isn't found?

It makes some sense...an element not having a property isn't an error condition, so it shouldn't really reject, and it makes it so any driver error encountered in `Element.getProperty()` isn't swallowed

paulirish:
Looking again, I don't see why we are using `.getProperty()` here. AFAIK both `href` and `rel` will reflect changes to their property in their attribute.

That said, I agree with getProperty resolving with null instead of rejecting. 

so IMO we should do that plus switching this gatherer to [use getAttribute entirely](https://github.com/GoogleChrome/lighthouse/blob/d0f1f73999ba97339e0d0c78dc8be74f16bb3b60/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js#L32-L34).

wardpeet:
I kept the driver api as is because I thought I could be a requirement ^^ but that's even better!

paulirish:
thx for grabbing this!

PR #1821, opened 2017-03-06
fix(ReportGenerator): use formatter for audit partial

### LINKED ISSUE #1820: Report: template is missing table/card/urllist formatters
after running:

    npm run install-all
    npm run build-all

and running a report, we're missing the extended info `<details>` in audits. 

screen shot 2017-03-06 at 9 26 18 am

@paulirish @patrickhulce can one of you guys take a look? I'm OOO until Wed.

patrickhulce:
R: @paulirish 

fixes #1820

### DIFF
--- a/lighthouse-core/report/report-generator.js
+++ b/lighthouse-core/report/report-generator.js
@@ -151,8 +151,10 @@ class ReportGenerator {
       }
 
       const partials = reportPartials.report.partials;
-      Handlebars.registerPartial(audit.name,
-        Handlebars.template(partials[audit.name] || partials['null-formatter']));
+      const partial = partials[audit.name] ||
+          partials[audit.extendedInfo.formatter] ||
+          partials['null-formatter'];
+      Handlebars.registerPartial(audit.name, Handlebars.template(partial));
     });
   }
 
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/report-generator.js
```diff
@@ -151,8 +151,10 @@ class ReportGenerator {
       }
 
       const partials = reportPartials.report.partials;
-      Handlebars.registerPartial(audit.name,
-        Handlebars.template(partials[audit.name] || partials['null-formatter']));
+      const partial = partials[audit.name] ||
+          partials[audit.extendedInfo.formatter] ||
```
paulirish: should we try looking at `partials[audit.name]` at all or just keep it to `partials[audit.extendedInfo.formatter] || null`?
If we can, I'd prefer that, as it's more declarative which formatter will be used.

sendilkumarn: Am sorry for this. 😞  Wonder how `crc` was working 
@paulirish null will not work here. you have to pass it as a template object.

sendilkumarn:
Thanks @patrickhulce 👍

lighthouse-core/report/report-generator.js
patrickhulce: Yeah we have a mismatch of kebab-case and camelCase here, so better to address that then keep the extra `||`.

@sendilkumarn no worries, `critical-request-chains` worked because the name of the audit is the same as the formatter :)

paulirish: Cool. Kinda crazy that we are matching filenames against kebab case whereas formatter js files are hooked up directly. 
But w/e lets ship it.

Issue #1817, opened 2017-03-05
Tests failure with latest master


XhmikosR:
https://ci.appveyor.com/project/paulirish/lighthouse/build/364

Noticed it when I rebased a branch of mine so I restarted the master branch build et voilà.

patrickhulce:
@XhmikosR that looks like a link for a PR not master ;)

looks like the [real latest master failure](https://ci.appveyor.com/project/paulirish/lighthouse/build/364/job/7vmqjofpine4qrcf) was a flake on chromestatus removing a class from their biggest element, maybe we shouldn't be too reliant on the smoke test URLs changing 🤔

XhmikosR:
https://ci.appveyor.com/project/paulirish/lighthouse/build/364

paulirish:
fixed by #1824

PR #1824, opened 2017-03-06
test: Fix chromestatus smokehouse flake.

paulirish:
addressing the smokehouse flake that's hitting us on travis

image


paulirish:
fixes #1817

PR #1827, opened 2017-03-07
RTT = Latency * 2

johnboxall:
I _believe_ Chrome's throttling settings introduce delay as latency, or the time it takes for a signal to pass from one point to the other.

Round trip time is the time there and back again. 

So... if we're displaying Round Trip Time (RTT) it should be latency times two.

image

💕

### DIFF
--- a/lighthouse-core/lib/emulation.js
+++ b/lighthouse-core/lib/emulation.js
@@ -134,7 +134,7 @@ function getEmulationDesc() {
   return {
     'deviceEmulation': 'Nexus 5X',
     'cpuThrottling': `${CPU_THROTTLE_METRICS.rate}x slowdown`,
-    'networkThrottling': `${latency}ms RTT, ${byteToMbit(downloadThroughput)}Mbps down, ` +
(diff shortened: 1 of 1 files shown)


johnboxall:
I signed it

johnboxall:
I signed it!

paulirish:
Soooo.. our throttling does have some issues, but luckily this is being represented well.

in devtools throttling case, "latency" is actually RTT time.

what we do:

1. start measuring time from when the http request is sent 
2. when the response starts coming back, we decide wether to delay the response some more
1. if the actual RTT is still lower than the throttled setting, then we'll wait for that many ms to release the response into the browser

this implies that if a site normally has RTT of like 30-ish on all its requests, when you set 150ms latency in devtools... you'll see light green waiting time of 150ms on all of them.

Trying this out just now, this seems to be a little inconsistent. But regardless, our latency is applied just once, subtracts actual RTT, and is essentially reflective of the RTT.

---------

Another way to verify this.. using --save-assets you'll save the trace.. open in timeline and you should see the light shaded color in network requests (waiting time) as 150ms ish.  
image

paulirish:
given that, i'll close this PR. 

our throttling behavior does require some more research however. there are some improvements we could make.

PR #1829, opened 2017-03-07
Complete fix for #1820 - missing formatter output

ebidel:
R: @patrickhulce, all

https://github.com/GoogleChrome/lighthouse/pull/1821/files addressed #1820, but it did not completely address the missing formatter output. The card and urllist formatters were still being left out of the html report.

I've also added a test to make sure these special formatters that have obvious HTML hooks (things we can look for in a a test) are being caught.

### DIFF
--- a/lighthouse-core/formatters/formatter.js
+++ b/lighthouse-core/formatters/formatter.js
@@ -40,9 +40,9 @@ class Formatter {
   static _getFormatters() {
     this._formatters = {
       accessibility: require('./accessibility'),
-      card: require('./cards'),
+      cards: require('./cards'),
       criticalRequestChains: require('./critical-request-chains'),
-      urllist: require('./url-list'),
+      urlList: require('./url-list'),
       null: require('./null-formatter'),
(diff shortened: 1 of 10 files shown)


lighthouse-core/test/report/report-generator-test.js
```diff
@@ -93,6 +93,17 @@ describe('Report', () => {
 
   it('does not include script for devtools', () => {
```
patrickhulce: nit: change test name

lighthouse-core/formatters/formatter.js
```diff
@@ -40,9 +40,9 @@ class Formatter {
   static _getFormatters() {
     this._formatters = {
       accessibility: require('./accessibility'),
-      card: require('./cards'),
```
patrickhulce: doh!

lighthouse-core/test/report/report-generator-test.js
ebidel: ha yep. done.

PR #1830, opened 2017-03-08
Configure lighthouse run by whitelisting aggregations

paulirish:
This PR is brings back the unmerged patch applied to enable DevTools Audits 2.0 panel integration, and then cleans things up to be less terrible.

Screenshots of what this enables:

image
image

The big change is allowing the user to select which top-level aggregations they want, rather than which 2nd-level aggregations they want. :)

Note that the config is expecting some large refactors in the near future. This PR allows the config to remain as is, but the new `Config.generateConfigOfAggregations` will be useful for the new `config.settings.only` and `config.settings.skip` options.

**Reviewers:**

* Note that `getGatherersNeededByAudits` just moved places
* `auditPathToName` is mostly moved from `lighthouse-background` to config. ish.
* The other green in config.js is indeed new, (but the majority of it has been in the devtools distribution for the past month ;)
* config-test.js got a whitespace adjustment, so ?ws=1 to neutralize half of it.
* added some basic integration tests for the `Config.generateConfigOfAggregations` run.
* The devtools-side matching CL is here: https://codereview.chromium.org/2736153002. I'm filtering out "Fancier stuff" from that UI manually. ;)

window.runLighthouseInWorker and window.runLighthouseInExtension just can get an array like `["Performance", "Best Practices"]` now. Earlier they took a more complex object with booleans. While those booleans are necessary for saving/restoring settings, they shouldn't cross over.

Lastly, the modifications in popup's `generateReportButton` is done as there is _currently_ a bug where changing the extension settings doesn't apply those changes immediately. It's using stale settings. So I fixed that bug while I was in here. :)

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -320,11 +320,10 @@ class Config {
   }
 
  /**
-  * Filter out any unrequested aggregations from the config. If any audits are
-  * no longer needed by any remaining aggregations, filter out those as well.
+  * Filter out any unrequested items from the config, based on requested top-level aggregations.
   * @param {!Object} oldConfig Lighthouse config object.
   * @param {!Array<string>} aggregationNames Name values of aggregations to include.
-  * @return {undefined} config is mutated, instead of returning
+  * @return {Object} new config
   */
   static generateNewConfigOfAggregations(oldConfig, aggregationNames) {
     // 0. Clone config to avoid mutating it
@@ -343,13 +342,13 @@ class Config {
     const requiredGatherers = Config.getGatherersNeededByAudits(auditObjectsSelected);
 
     // 4. Filter to only the neccessary passes
-    config.passes = Config.getPassesNeededByGatherers(config.passes, requiredGatherers);
+    config.passes = Config.generatePassesNeededByGatherers(config.passes, requiredGatherers);
     return config;
   }
 
  /**
   * Return names of top-level aggregations from a config. Used by tools driving lighthouse-core
-  * @param {!Object} config Lighthouse config object.
+  * @param {{aggregations: !Array<{name: string}>}} Lighthouse config object.
   * @return {!Array<string>}  Name values of aggregations within
   */
   static getAggregationNames(config) {
@@ -358,7 +357,7 @@ class Config {
 
   /**
    * Find audits required for remaining aggregations.
-   * @param  {!Object} aggregations
+   * @param {!Object} aggregations
    * @return {!Set<string>}
    */
   static getAuditsNeededByAggregations(aggregations) {
@@ -373,9 +372,6 @@ class Config {
    * @return {Map}
    */
   static getMapOfAuditPathToName(config) {
-    // The `audits` property in the config is a list of paths of audits to run.
-    // `requestedAuditNames` is a list of audit *names*. Map paths to names, then
-    // filter out any paths of audits with names that weren't requested.
     const auditObjectsAll = Config.requireAudits(config.audits);
     const auditPathToName = new Map(auditObjectsAll.map((AuditClass, index) => {
       const auditPath = config.audits[index];
@@ -387,7 +383,7 @@ class Config {
 
   /**
    * From some requested audits, return names of all required artifacts
-   * @param  {!Object} audits
+   * @param {!Object} audits
    * @return {!Set<string>}
    */
   static getGatherersNeededByAudits(audits) {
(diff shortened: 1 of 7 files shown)


lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
   }
 
   /**
+  * Filter out any unrequested aggregations from the config. If any audits are
+  * no longer needed by any remaining aggregations, filter out those as well.
+  * @param {!Object} config Lighthouse config object.
+  * @param {!Array<string>} aggregationNames Name values of aggregations to include.
+  * @return {undefined} config is mutated, instead of returning
+  */
+  static generateConfigOfAggregations(config, aggregationNames) {
```
ebidel: "generate" implies something new is being created. Might want to use something like `removeUnrequestedAggregations`?

lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
...
+    // 3. Resolve which gatherers will need to run
+    const auditObjectsSelected = Config.requireAudits(config.audits);
+    const requiredGatherers = Config.getGatherersNeededByAudits(auditObjectsSelected);
+
+    // 4. Filter to only the neccessary passes
+    config.passes = Config.selectPassesNeededByGatherers(config.passes, requiredGatherers);
+  }
+
+  /**
+  * Filter out any unrequested aggregations from the config. If any audits are
```
ebidel: update the comment text (same as `generateConfigOfAggregations`). Indentation too.

lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
...
+  * no longer needed by any remaining aggregations, filter out those as well.
+  * @param {!Object} config Lighthouse config object.
+  * @return {!Array<string>}  Name values of aggregations within
+  */
+  static getAggregationNames(config) {
```
ebidel: do you use this method anywhere?

lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
...
+
+  // Find audits required for remaining aggregations.
+  static getAuditsNeededByAggregations(aggregations) {
+    const requestedItems = _flatten(aggregations.map(aggregation => aggregation.items));
+    const requestedAudits = _flatten(requestedItems.map(item => Object.keys(item.audits)));
+    return new Set(requestedAudits);
+  }
+
+  /**
+   * @return {map} map from audit path (seein in config.audits) to audit.name used in config.aggregations
```
ebidel: map -> Map

ebidel: seein?

ebidel: add docs for `config`

lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
...
+    const auditPathToName = new Map(auditObjectsAll.map((AuditClass, index) => {
+      const auditPath = config.audits[index];
+      const auditName = AuditClass.meta.name;
+      return [auditPath, auditName];
+    }));
+    return auditPathToName;
+  }
+
+  static getGatherersNeededByAudits(audits) {
+    // It's possible we didn't get given any audits (but existing audit results), in which case
```
ebidel: we didn't get -> we weren't

lighthouse-extension/gulpfile.js
```diff
@@ -183,7 +183,7 @@ gulp.task('clean', () => {
 });
 
 
-gulp.task('watch', ['lint', 'compileReport', 'compilePartials', 'browserify', 'html'], () => {
```
ebidel: yes!

lighthouse-core/config/config.js
paulirish: Yes. Agree the name isn't right. 
Alternatively I could do the json parse/stringify here rather than mutating the passed in config.

will noodle on that.

lighthouse-core/config/config.js
paulirish: good catch! :) right now i don't, but `window.getDefaultAggregations` should. And I was thinking that it'd make sense for tools embedding lighthouse to get access to their customization options easier.

I'll adapt window.getDefaultAggregations and add a test.

lighthouse-core/config/config.js
```diff
@@ -331,6 +320,101 @@ class Config {
...
+      return new Set();
+    }
+
+    return audits.reduce((list, audit) => {
+      audit.meta.requiredArtifacts.forEach(artifact => list.add(artifact));
+      return list;
+    }, new Set());
+  }
+
+  static selectPassesNeededByGatherers(passes, requiredGatherers) {
```
paulirish: yeah kinda of weird that some of these mutate, others return new. i'll try to normalize them

paulirish:
Okay updated.  ptal

Now no methods mutate existing data, they all return new objects. Easier to reason about, I hope!

lighthouse-core/config/config.js
```diff
@@ -330,6 +319,121 @@ class Config {
...
+  }
+
+  /**
+   * Creates mapping from audit path (used in config.audits) to audit.name (used in config.aggregations)
+   * @param {!Object} config Lighthouse config object.
+   * @return {Map}
+   */
+  static getMapOfAuditPathToName(config) {
+    // The `audits` property in the config is a list of paths of audits to run.
+    // `requestedAuditNames` is a list of audit *names*. Map paths to names, then
```
brendankenny: `requestedAuditNames` isn't a thing anymore, so need to update comment. Maybe just `config.audits` instead?

lighthouse-core/config/config.js
```diff
@@ -330,6 +319,121 @@ class Config {
...
+    // filter out any paths of audits with names that weren't requested.
```
brendankenny: don't filter here either

lighthouse-core/config/config.js
```diff
@@ -330,6 +319,121 @@ class Config {
...
+  }
+
+  /**
+   * Filters to only required passes and gatherers, returning a new passes object
+   * @param  {!Object} passes
+   * @param  {!Set<string>} requiredGatherers
+   * @return {!Object} fresh passes object
+   */
+  static getPassesNeededByGatherers(passes, requiredGatherers) {
+    const passesClone = JSON.parse(JSON.stringify(passes));
```
brendankenny: what about sticking with what you do elsewhere and call the parameter `oldPasses`?

lighthouse-core/config/config.js
```diff
@@ -330,6 +319,121 @@ class Config {
...
+    const filteredPasses = passesClone.map(pass => {
+      // remove any unncessary gatherers
+      pass.gatherers = pass.gatherers.filter(gathererName => {
+        gathererName = GatherRunner.getGathererClass(gathererName).name;
+        return requiredGatherers.has(gathererName);
+      });
+      return pass;
+    // remove any passes lacking concrete gatherers
+    }).filter(pass => pass.gatherers.length > 0);
```
brendankenny: won't this remove a pass that's just for a trace or networkRecord as long as there are other traces with gatherers remaining? (below seems to be for only when there are no gatherers at all).

It seems like (until we get finer grained information about *which* trace or network records an audit is interested in) if any audit requires `traces` or `networkRecords` we can't eliminate any passes that have `recordTrace` or `recordNetwork` set to true

lighthouse-core/config/config.js
```diff
@@ -330,6 +319,121 @@ class Config {
...
+
+    // handle the perf-only case (no specific gatherers, just trace & network)
+    if (filteredPasses.length === 0) {
```
brendankenny: generating a pass is a little odd here (though less so if you don't go with my suggestion of `filterPassesByGatherers` :)

Would it make sense just to not filter this out in the first place (re: above comment)? That way other options like the `passName` and `useThrottling` wouldn't be dropped

lighthouse-core/config/config.js
paulirish: we can eliminate recordNetwork ones but not recordTrace ones. this will require some better handling, but i'm not addressing it now.

logic updated to keep recordTrace ones, otherwise nuke pass if no gatherers.

paulirish:
PTAL.

PR #1832, opened 2017-03-08
Fail if PWA score is lower than `--fail-under`.

johnboxall:
> Adds `--fail-under=INT` option CLI which causes it to exit unsuccessfully if the PWA aggregate score is below the specified value.
> ```bash
> lighthouse --fail-under=38 https://www.example.com

We're using Lighthouse on CI to verify PWAs are hitting above a score and that over time they don't regress. *Sorta like coverage regression testing for performance!*

To detect whether scores have regressed, we use a small script that runs after `lighthouse` completes:

```js
// Return the score from a Lighthouse Report.
// Usage: node get-lighthouse-score.js $PATH_TO_LIGHTHOUSE_REPORT
const fs = require('fs')

// Dig the PWA score out of the output file.
const path = process.argv[2]
const report = fs.readFileSync(path, 'utf8')
const results = report.match(/<span class="section-result__points">(.*)<\/span>/)
const score = parseInt(results[1])

// Fail if we were under the threshold.
if (process.env.LIGHTHOUSE_FAIL_UNDER > score) {
   process.exit(2)
}
```

Having to include this helper script on everything is a pain, I'd rather we could just pass it as an option to Lighthouse.

Happy to add tests / do a second pass through if this is a direction the project is thinking of heading.

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -326,15 +326,18 @@ export async function runLighthouse(url: string,
       await performanceXServer.hostExperiment({url, flags, config}, results);
     }
 
-    await chromeLauncher.kill();
+    let promise = await chromeLauncher.kill();
 
-    const total = Math.ceil(results.aggregations[0].total * 100);
-    const shouldFail = flags.failUnder && flags.failUnder > total;
-    if (shouldFail) {
-        throw new Error(`Got ${total} needed ${flags.failUnder}`);
+    // FIXME: Cannot provide `--fail-under` unless config includes aggregations.
+    if (results.aggregations.length) {
+        const total = Math.ceil(results.aggregations[0].total * 100);
+        const shouldFail = flags.failUnder && flags.failUnder > total;
+        if (shouldFail) {
+            throw new Error(`Got ${total} needed ${flags.failUnder}`);
+        }
     }
 
-    return Promise.resolve(true)
+    return promise
   } catch (err) {
     if (typeof chromeLauncher !== 'undefined') {
       await chromeLauncher!.kill();


patrickhulce:
Thanks for the contribution! I'm in favor of the functionality, but we're considering some proposed changes to scoring that might make this slightly more complicated we'd have to think through.

I'd also prefer there to be a distinct exit code for this sort of failure as opposed to an ordinary runtime error exit code. Would definitely simplify the CI story though I like it!

ebidel:
I like the idea of this too. Adding to what @patrickhulce mentioned, we're planning to have several scored sections very soon. So we'd have to figure out how adding a single flag to core would work.

In practice, this feels like a perfect use case for using LH programmatically! @johnboxall, we have a sample here: https://github.com/GoogleChrome/lighthouse#using-programmatically. It also has a snippet of how to calculate the overall score across all sections. That would be future proof with our upcoming scoring changes.

BTW, instead of grepping the report HTML output, you can use  `--output=json` (or `{output: 'json'}` in the example above) to return the report as JSON. That might be less brittle for ya if we update the HTML structure in the future.

johnboxall:
Thanks for the feedback @patrickhulce & @ebidel.

Couple of thoughts:

1) We're aware of the other output types, but the HTML output is a nice build artifact for quickly spotting why things failed. (an alternative is we could dump the JSON in the viewer, but thats one more step)
2) WRT to multiple scoring sections – in Lighthouse's current implementation, I think it is a safe assumption that you want to fail on the score of the first aggregate in your configuration. In some future town, maybe this is coupled with a `--fail-on=$AGGREGATE_KEY` that allows you to specify some-non default aggregate to fail on.

Would love to see something like this make its way into a release in the future though. 🚀 

*PS. Thanks for the awesome tool!*

brendankenny:
you can do multiple output modes (like html and json) after #1813, but they will both be written to disk. We could fix `--output-path` to be supported for multiple output modes, so you could do e.g. HTML to disk and JSON to stdout, though that might end up confusing.

One issue with any overall score, of course, is that improvements in one area can mask regressions in others, so I'll agree with the others that we'd really want the option to fail on subscores. It feels kind of like we want smokehouse light here, but I'm not sure of a great way to do it without requiring a flag that would get complicated quickly.

addyosmani:
Fwiw, when I tried implementing something similar I found creating a separate custom module that failed on a particular error code (I used the JSON report) was relatively low-effort. Smokehouse proved a super helpful reference point. 

That said I agree with @johnboxall that not having to write a custom helper for each project trying to achieve this might encourage more folks to use Lighthouse in CI.

brendankenny:
we could add something to the new [`recipes/` directory](https://github.com/GoogleChrome/lighthouse/tree/master/docs/recipes) at the very least. Adding a utility that does this to package.json's `scripts` and/or `bin` could also make it easy to use without adding something super specific to the main binary (or adding to it to make it sufficiently flexible until it has too many options)

paulirish:
I'm in favor of landing this in our CLI.  I just want our config changes in
place first.

paulirish:
So I wanted to implement this against tip of tree since it's been a while.. here's what the implementation of a standalone CLI looks like:

```js
// usage:
//     node failunder.js http://paulirish.com --failing=50

const runLighthouse = require('lighthouse/lighthouse-cli/run.js').runLighthouse;

// CLI args
const args = process.argv.slice(2);
let startingUrl = args.find(flag => !flag.startsWith('--'));
let failingScore;
const failingFlag = args.find(flag => flag.startsWith('--failing='));
if (failingFlag) {
  failingScore = failingFlag.replace('--failing=', '');
}

failingScore = defaults(failingScore, 40);
startingUrl= defaults(startingUrl, 'https://example.com');

// these 3 flag values are required
const flags = { 
  port: 0,
  chromeFlags: '',
  output: 'json'
};
const pwaConfig = {
  'extends': 'lighthouse:default',
  'settings': {
    'onlyCategories': ['pwa']
  }
};

runLighthouse(startingUrl, flags, pwaConfig).then(results => {
  const pwaCat = results.reportCategories.find(cat => cat.id === 'pwa');

  process.stdout.write(`Lighthouse PWA Score: ${pwaCat.score} \n`);
  if (pwaCat.score <= failingScore) {
    process.exit(2);
  }
});

function defaults(val, def) {
  return typeof val === 'undefined' ? def : val;
}

```

Us adding this into CLI requires some careful work because we have 4 categories and we'd want to assert details about the score of just one of them.

Alternatively, this sort of thing could move out into its own module. Interestingly, both smokehouse (within lighthouse) and the `pwmetrics` module have [a configuration to assert against the LH result object](https://github.com/paulirish/pwmetrics/#all-available-configuration-options) and fail the build if certain thresholds were not met. 

I can see something like a `lighthouse-assert` module existing that takes a configuration and returns errors/warnings. At that point, with the configuration spec'd, then i think adding a `--expectations` flag to lighthouse-cli makes total sense.

cc @pedro93 @denar90

denar90:
Currently, everyone wanna use LH on CI, so I'm 👍 with `lighthouse-assert` to avoid different implementation across projects. 
Of course, it makes LH more complicated but for users, it's simpler. 
I think for some people it's blocker to integrate LH and CI for their project because they have to build their own implementations.
E.g. I have some `--expectations` implementation, after landing next LH version I have to make changes to my code fixing CI. If we are talking about OSS it will be awesome decrease time spending for this kind of things.
Also talking about OSS, having `--expectations` under the hood can make LH part of 90% simple sites, owners of which care about perf. 
Either for enterprise or for people with special needs, it will stay as it's, they can use their own `--expectations` implementations and of course support them.
Other +1, `smokehouse` also can reuse `lighthouse-assert` logic for its needs.

@ebidel @brendankenny wdyt?

Issue #1102, opened 2016-12-02
Add lighthouse ecosystem projects to readme


paulirish:
We could use a section on our readme that links to:

* https://github.com/addyosmani/webpack-lighthouse-plugin 
* https://github.com/justinribeiro/lighthouse-mocha-example
* https://github.com/paulirish/pwmetrics/

ebidel:
Done

PR #1835, opened 2017-03-08
Add related projects section

chrisdwheatley:
This covers #1102 by adding the suggested links to the readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -206,6 +206,12 @@ right corner and signing in to GitHub.
 
 > **Note**: shared reports are stashed as a secret Gist in GitHub, under your account.
 
+## Related Projects
+
+* [webpack-lighthouse-plugin](https://github.com/addyosmani/webpack-lighthouse-plugin)
+* [lighthouse-mocha-example](https://github.com/justinribeiro/lighthouse-mocha-example)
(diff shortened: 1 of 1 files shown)


chrisdwheatley:
I signed it!

ebidel:
Thanks for the contribution @chrisdwheatley! 

Wonder what you/others think about enabling the wiki and we can have folks add their projects there? We did something similar for https://github.com/Polymer/polymer/wiki/Who's-using-Polymer%3F.

paulirish:
IMO readme gets like 100x more eyes than wiki. I think we should only move this data to the wiki once we have like > 20 items to put here.

i <3 having this stuff in the readme like this.


@ebidel sgty?

paulirish:
@chrisdwheatley thx for the ez pz contribution! really appreciate it.

looking at https://www.npmjs.com/browse/depended/lighthouse i wonder if there's anything else worth adding. maybe lighthouse-batch?

ebidel:
SG. Can you also add https://github.com/ebidel/lighthouse-hue :)

paulirish:
did a followup commit: https://github.com/GoogleChrome/lighthouse/commit/342398c716ce98fd497a7e7d3dfa64a751f4eb1f

image

PR #1838, opened 2017-03-08
Simplify report generation: no pretty print, no formatter classes

### LINKED ISSUE #1739: Is it time to deprecate the CLI report output?

brendankenny:
This ended up being something of a monster since formatters have grown to touch so many parts of the codebase. This removes formatters completely, setting the stage for much simpler report refactors in the future.

The only remaining "formatting" is done with handlebar partials, and all that functionality now lives entirely with `lighthouse-core/report/`. The only thing any code outside that directory needs to know about formatting is the string specified by audits in their `extendedInfo.formatter` results (this hasn't changed at all from before, still on the `Formatter.SUPPORTED_FORMATS` object).

The individual commits stand on their own, but the last moves enough around that I'm not sure if going through them individually really helps with review. As mentioned in the commit messages, the PR:
- removes pretty printing on the command line (fixes #1739)
- simplifies the `--output` and `--output-path` options a bit since we got rid of pretty (see bin.ts/readme changes for details)
- moves all handlebar helpers from `getHelpers()` in individual formatter files to the central `handlebar-helpers.js`. This is somewhat questionable organizationally, but they all live in a global namespace when they're registered anyways, and having them in one place will aid with future refactoring.
- move partial html and css files from `formatters/partials/` to `report/partials/`
- eliminates the Formatter subclasses for each formatter since after #1752 the partial sources are retrieved from the precompiled partials anyway and with all the changes from above, they no longer have a use
- keep formatter unit tests but instead have them compile and test the html produced by the partials directly

In addition, `Formatter.SUPPORTED_FORMATS` is now auto-generated from the available partial files and `report-generator.js` now throws on unknown formatters, so it should cut down on issues like #1820

Tested in CLI and extension, no visible changes


lighthouse-core/audits/critical-request-chains.js
```diff
@@ -40,6 +40,58 @@ class CriticalRequestChains extends Audit {
...
+    }
+
+    walk(tree, 0);
+  }
+
+  /**
+   * Get stats about the longest initiator chain (as determined by time duration)
+   * @return {{duration: number, length: number, transferSize: number}}
+   */
+  static _getLongestChain(tree) {
```
brendankenny: move from the formatter to the audit. I think it was only in the formatter because we didn't have `extendedInfo` when this was first written

lighthouse-core/audits/critical-request-chains.js
```diff
@@ -77,7 +129,10 @@ class CriticalRequestChains extends Audit {
         optimalValue: this.meta.optimalValue,
         extendedInfo: {
           formatter: Formatter.SUPPORTED_FORMATS.CRITICAL_REQUEST_CHAINS,
-          value: chains
+          value: {
+            chains,
+            longestChain: CriticalRequestChains._getLongestChain(chains)
```
brendankenny: the change to the CriticalRequestChains `extendedInfo` is the only real externally-noticeable change. Added `longestChain` since it's used several times in the report, so no reason not to just precompute it

lighthouse-core/audits/critical-request-chains.js
```diff
@@ -40,6 +40,58 @@ class CriticalRequestChains extends Audit {
     };
   }
 
+  static _traverse(tree, cb) {
```
patrickhulce: did this used to live in the formatter?

lighthouse-core/report/formatter.js
```diff
@@ -0,0 +1,45 @@
...
+'use strict';
+
+const fs = require('fs');
+
+// Auto convert partial filenames to lookup for audits to use.
+const formatters = {};
+fs.readdirSync(__dirname + '/partials')
+  .filter(filename => filename.endsWith('.html'))
+  .forEach(filename => {
+    const baseName = filename.replace(/\.html$/, '');
```
patrickhulce: all of this is still lines up with eric's fix?

lighthouse-core/report/report-generator.js
```diff
@@ -132,30 +127,23 @@ class ReportGenerator {
...
-      }
-      const formatter = Formatter.getByName(audit.extendedInfo.formatter);
-      const helpers = formatter.getHelpers();
-      if (helpers) {
-        Handlebars.registerHelper(helpers);
+
+      const partialName = audit.extendedInfo.formatter;
+      const partial = reportPartials.report.partials[partialName];
+      if (!partial) {
+        throw new Error(`${auditName} requested unknown partial for formatting`);
```
patrickhulce: nice :)

lighthouse-core/audits/critical-request-chains.js
patrickhulce: already answered, ignore

lighthouse-core/report/formatter.js
brendankenny: not entirely, it changes somewhat, but it is all working.

I moved us from three possible cases (caps for the `SUPPORTED_FORMATS` name, camel for the actual value of that variable, and kebab for the partial name, so we had to convert from camel to kebab to map from formatter to partial), to just caps and kebab (so the value of the formatter property is the same as the name of the partial)

brendankenny: > so the value of the formatter property is the same as the name of the partial

(and that's now checked by [a test](https://github.com/GoogleChrome/lighthouse/pull/1838/files#diff-dd8576d3d840580d95a4a90f37ab89a2R31))

readme.md
```diff
@@ -34,7 +34,7 @@ Kick off a run by passing `lighthouse` the URL to audit:
 lighthouse https://airhorner.com/
 ```
 
-Lighthouse will prettyprint a report to CLI. You can control the output format by passing flags.
+Lighthouse will write an HTML report to file. You can control the output format by passing flags.
```
ebidel: "write the report to an HTML file"?

readme.md
```diff
@@ -66,8 +66,13 @@ Configuration:
 
 Output:
   --output       Reporter for the results, supports multiple values
-                         [choices: "none", "pretty", "json", "html"]               [default: "none"]
-  --output-path  The file path to output the results
+                                    [choices: "json", "html"]                    [default: "none"]
```
ebidel: `[default: "html"]` right?

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -43,13 +47,13 @@ function calculateRating(value) {
 }
 
 /**
- * Figures out the total score for an aggregation
- * @param {{total: number}} aggregation
- * @return {number}
+ * Format number.
```
ebidel: `formatNumber` formaters a number...

keanu2

lighthouse-core/report/formatter.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Formatter {
+  /**
+   * Getter that returns a list of supported formatters, mapping an all-caps
+   * name to the basename of the available formatter partials
+   * (e.g. `CRITICAL_REQUEST_CHAINS` to `critical-request-chains).
+   * @return {!Object<string, string>}
+   */
+  static get SUPPORTED_FORMATS() {
```
brendankenny: > does this need to be in caps?

no, I just left it how it was before so that audits don't have to change the `formatter: Formatter.SUPPORTED_FORMATS.USER_TIMINGS` (or whatever) line in `extendedInfo`.

Selecting the partial to use is a little awkward in general. It could also have been a string (like e.g `requiredArtifacts` is) since that's all that variable really is, but there is the advantage of if your editor does support code completion you get suggestions there. The `SUPPORTED_FORMATS` could have lived somewhere else, too (like on `ReportGenerator`).

Basically I didn't see an option that was legitimately better, so I just left it the same until we do come up with something :)

lighthouse-core/report/handlebar-helpers.js
```diff
@@ -161,7 +195,7 @@ const handlebarHelpers = {
    * @param {!Object} opts
    * @return {string}
    */
-  createTreeRenderContext(tree, opts) {
+  createTreeRenderContext(tree = {}, opts) {
```
ebidel: doc the default val?

lighthouse-core/report/formatter.js
ebidel: Tested the PR and everything looks good.

lighthouse-core/report/report-generator.js
```diff
@@ -21,14 +21,10 @@
 const Handlebars = require('handlebars/runtime');
 const handlebarHelpers = require('./handlebar-helpers');
 const reportTemplate = require('./templates/report-templates');
-const reportPartials = require('../formatters/partials/templates/report-partials');
+const reportPartials = require('../report/partials/templates/report-partials');
 const fs = require('fs');
 const path = require('path');
 
-function toKebabCase(string) {
```
ebidel: 🚫 🍖

readme.md
brendankenny: > "write the report to an HTML file"?

I mean, the report is the html, it's not writing it to an HTML file...I don't know.

readme.md
brendankenny: > `[default: "html"]` right?

whoops, yes. done.

lighthouse-core/report/handlebar-helpers.js
brendankenny: > formatNumber formaters a number

I got less creative as I went along and just stole it from the also excellent [`format time`](https://github.com/GoogleChrome/lighthouse/blob/fe88c4ab830eff5182ac6754a30be353f89503d6/lighthouse-core/report/handlebar-helpers.js#L188)

readme.md
ebidel: wording was a bit funny to me. "write your report to a file..."?

brendankenny: I think the point of the original statement is to point out what you get by default and call attention to the fact that there are flags to change that.

Agree it is a bit awkward. "Lighthouse will write an HTML report to file by default." to call attention to the fact that it's talking about the default behavior?

Also we could delete the line :)

lighthouse-core/report/handlebar-helpers.js
brendankenny: this was to work around a wonky unit test in development, but no longer necessary \o/ removed.

readme.md
ebidel: "write an HTML report to file" sounds incomplete. "to a file", "to an HTML file"....

What about " By default, Lighthouse writes the report to an HTML file. You can control the output format by passing flags."

brendankenny: 💥  done

brendankenny:
merged! looking at the open PRs, I only see #1847 and #1840 needing to be updated after this, and that should just require `const Formatter = require('../formatters/formatter');` -> `const Formatter = require('../report/formatter');`

paulirish:
@brendankenny thanks for checking! i've updated those PRs

PR #1841, opened 2017-03-09
Report: add legend to decipher iconography left to us by the ancient ones

ebidel:
R: all

Time for a legend kids! As we've added more icons, it's become unclear what everything means.

screen shot 2017-03-08 at 5 23 48 pm


lighthouse-core/report/styles/report.css
```diff
@@ -120,7 +126,6 @@ summary {
   margin: 0 auto;
   max-width: var(--report-width);
   background-color: #fff;
-  box-shadow: 0 0 6px 0 rgba(0,0,0,0.26);
```
ebidel: this is no longer visible, so removing.

paulirish:
My my this is very attractive. Nice stuff, man.

lighthouse-core/report/templates/report-template.html
```diff
@@ -226,7 +226,80 @@ <h2 class="config-section__title">Blocked URL Patterns</h2>
...
+                  <span class="aggregation__header__score score-good-bg subitem-result__good">Good</span>
+                  <span class="footer__key-description">Overall category score is 75-100</span>
+                </li>
+                <li>
+                  <span class="aggregation__header__score score-average-bg subitem-result__good">Average</span>
+                  <span class="footer__key-description">Overall category score is 45-74</span>
+                </li>
+                <li>
+                  <span class="aggregation__header__score score-poor-bg subitem-result__poor">Poor</span>
+                  <span class="footer__key-description">Overall category score is < 73</span>
```
sendilkumarn: is it < 45 ?

ebidel: Good catch. Done

ebidel:
> I'd normally argue the fact that we need a legend for report scoring is a UX smell and an argument that we need to fix the report itself, but that's in progress so hopefully this isn't a permanent solution 

100% agree. It's gotten out of control. I wanted to get this in before the report redo. No telling how long or what form that would take.

PR #1849, opened 2017-03-10
.npmignore more dev-only things

XhmikosR:
Also, ignores more files and folders like:

* gulp/
* lighthouse-cli/test/
* gulpfile.js
* .editorconfig
* .eslintignore
* .eslintrc.js
* latest.artifacts.log

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,38 +1,40 @@
-node_modules
-.DS_Store
-npm-debug.log
-.vscode
+# folders
+.vscode/
+assets/
+coverage/
+gulp/
 
-lighthouse-extension/dist
-lighthouse-core/third_party/src
-**/pages/scripts/lighthouse-report.js
-coverage
-
-lighthouse-cli/results
-results
-results.html
-last-run-results.html
-
-*.trace.json
-*.screenshots.html
-*.report.html
-
-closure-error.log
+lighthouse-core/scripts/
+lighthouse-core/test/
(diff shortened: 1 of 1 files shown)


XhmikosR:
Forgot to mention the file changes:

```
before: 3.080 Files, 465 Folders - 67,9 MB (71.260.044 bytes)
after:  3.033 Files, 452 Folders - 67,7 MB (70.995.752 bytes)

**Comment by brendankenny** [comment]
Traveling so I haven't been able to take a close look at the rest, but `lighthouse-cli/test/` is left in to allow [local testing of the module before publishing](https://github.com/GoogleChrome/lighthouse/blob/7c03886ff0b49377dea71b9718e2c188ed8f9545/CONTRIBUTING.md#release-guide)

**Comment by XhmikosR** [comment]
IMO, you shouldn't do that but it's your call.

If you stick with that, I guess this can be closed since most changes are develop related; for example `npm run lint` won't work after this.

**Comment by brendankenny** [comment]
We don't lint the module code, so that's no issue. The key is automated testing of the module (not only that it runs, but that it's running correctly) before publishing, and until there is some other system for that, `npm explore lighthouse -- npm run smokehouse` it is :)

**Comment by wardpeet** [comment]
adding should keep smokehouse working
```
lighthouse-cli/test/cli/
lighthouse-cli/test/fixtures/
```

PR #1851, opened 2017-03-10
Tweak report colors so that we are WCAG2AA valid.

XhmikosR:
| Before | After |
| ------------- |-------------| 
| 1before | 2after  |

Feel free to tweak this if you want to :)

### DIFF
--- a/lighthouse-core/report/partials/cards.css
+++ b/lighthouse-core/report/partials/cards.css
@@ -30,6 +30,7 @@
   justify-content: center;
   align-items: center;
   border-bottom: 1px solid #ebebeb;
+  color: #4f4f4f;
 }
 .scorecard-value {
   font-size: 28px;
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -27,11 +27,11 @@ span, div, p, section, header, h1, h2, li, ul {
 :root {
   --text-font-family: "Roboto", -apple-system, BlinkMacSystemFont,  "Segoe UI", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
   --text-color: #222;
-  --secondary-text-color: #606060;
+  --secondary-text-color: #565656;
   --accent-color: #3879d9;
-  --poor-color: #e53935; /* md red 600 */
-  --good-color: #43a047; /* md green 600 */
-  --average-color: #ef6c00; /* md orange 800 */
+  --poor-color: #df332f;
+  --good-color: #2b882f;
+  --average-color: #c15700;
   --warning-color: #757575; /* md grey 600 */
   --gutter-gap: 12px;
   --gutter-width: 40px;
@@ -90,7 +90,7 @@ html, body {
 }
 
 a {
-  color: #15c;
+  color: #0c50c7;
 }
 
 body {
@@ -338,7 +338,7 @@ summary {
 .menu__link {
   padding: 0 20px;
   text-decoration: none;
-  color: #767676;
+  color: #595959;
   display: flex;
 }
 
@@ -629,7 +629,7 @@ summary {
   margin-left: var(--report-menu-width);
   font-size: 12px;
   border-top: 1px solid var(--report-border-color);
-  color: #767676;
+  color: #565656;
   background-color: var(--report-header-bg-color);
   border-top: 1px solid var(--report-border-color);
 }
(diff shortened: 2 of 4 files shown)


patrickhulce:
I'm surprised this is more valid than the previous, I can barely distinguish warning/error/poor score after this change! This has seems like a tug of war with the brown color, and a trade-off between background contrast and contrast between items. IMO, we shouldn't change as moving from AA to AAA weakens the latter significantly.

XhmikosR:
That is why I say feel free to tweak this :)

It's clear though that the current colors do **not** pass WCAG 2AA on certain occasions.

For example the average bacjground color is `#ef6c00` on `#fff` which gives a 3.1 contrast ration for large text only. So I decided to go with WCAG2AAA, but I can revert some of the changes so that at least it's WCAG2AA valid.

patrickhulce:
Instead of weakening all of the colors further, how about we flip the score text to black? That yields good contrast ratios (5+) and still keeps inter-item comparisons scannable.

XhmikosR:
I thought about that too. I'll give it a go later unless one of you beats
me to it :)

patrickhulce:
Sweet, appreciate the vigilance on these and contributions @XhmikosR ! :)

XhmikosR:
@patrickhulce: now everything is passing for me for WCAG2AA and for WCAG2AAA there are only 12 issues with this branch; before it was like 70 or so.

ebidel:
Hmm. Yea the new colors were chosen to be AA compliant. IMO they look good
as is and we shouldn't continue to tweak them every couple of weeks.

XhmikosR:
True but they weren't
AA compliant.

ebidel:
Can you point out where it's failing? When they were updated before they
checked out.

XhmikosR:
See my comment above. They are not AS valid and IIRC I mentioned in the
pull request that changed the colors.

ebidel:
https://github.com/GoogleChrome/lighthouse/pull/1758 were good to go at 18pt. The good color you mentioned in the PR should have also been fine but I'll check it out when I'm not on mobile.

lighthouse-core/report/partials/cards.css
```diff
@@ -30,6 +30,7 @@
   justify-content: center;
   align-items: center;
   border-bottom: 1px solid #ebebeb;
+  color: #4f4f4f
```
ebidel: ;

lighthouse-core/report/styles/report.css
```diff
@@ -933,9 +933,9 @@ summary {
 
 .subitem-result__points {
   margin-top: calc((var(--subitem-line-height) - var(--subitem-font-size) - 4px) / 2);
-  background-color: #000;
+  background-color: fff;
```
ebidel: `#fff`

lighthouse-core/report/styles/report.css
```diff
@@ -933,9 +933,9 @@ summary {
...
   padding: 2px 4px;
-  color: #fff;
+  color: #000;
```
ebidel: IMO, this is pretty hard to read. Since the report is going through a bit of a redesign, let's keep it at `#fff` for now. That's more consistent with the rest of the report.

ebidel:
I double checked all the existing colors again on https://snook.ca/technical/colour_contrast/colour.html and they continue to pass "WCAG 2 AA Compliant (18pt+)" per the changes in #1758. So think we're good there. 

Update: I see the PR has changed since the screenshots were posted.

XhmikosR:
@ebidel: I reverted the black on white change.

Now, about the previous colors; they were valid only for 18pt or 14px and above.

I went back and and changed them for 2AA.

XhmikosR:
Screenshots updated with the current changes.

ebidel:
talked with @robdodson, and he suggested we could probably get away with keeping the existing colors if we made the icons slightly bigger (19px). Messing around with that a bit, it does do the trick. I can add those changes to this PR later today or start a new one.

robdodson:
19px == 14pt bold which is allowed in wcag2aa. I think the icons are chunky
enough to be considered bold. My hope is this solution presents a middle
ground that satisfies everyone.

XhmikosR:
Sounds good to me. Although the current state of this branch doesn't have  significant changes to the colors and aims for 2AA.

ebidel:
Closing in favor of https://github.com/GoogleChrome/lighthouse/pull/1856

XhmikosR:
Cool, give me some time to verify all is good with WCAG2AA in your branch.

PR #1855, opened 2017-03-13
feat(uses-optimized-images): add support for CORS images

### LINKED ISSUE #1467: Gatherer: Add support for cross-origin image optimization checks
Migrated from comment in #1452 regarding lack of support for cross-origin images. Canvas won't process them without opting in with `crossorigin='anonymous'` which all sites/images didn't support. Needs further investigation on how to proceed.

patrickhulce:
fixes #1467

### DIFF
--- a/lighthouse-cli/test/fixtures/byte-efficiency/tester.html
+++ b/lighthouse-cli/test/fixtures/byte-efficiency/tester.html
@@ -55,6 +55,10 @@ function generateInlineScriptWithSize(sizeInBytes, firstContent = '', used = fal
   <!-- PASS(responsive): image is used at full size -->
   <img src="lighthouse-unoptimized.jpg">
 
+  <!-- FAIL(optimized): image is not optimized -->
+  <!-- PASS(responsive): image is used at full size -->
+  <img src="https://github.com/GoogleChrome/lighthouse/raw/751a84547bfff97d1f6c4533d1be2266908dfdc9/assets/lighthouse-logo_no-light.png">
+
   <!-- PASSWARN(optimized): image is JPEG optimized but not WebP -->
   <!-- FAIL(responsive): image is 25% used at DPR 2 -->
   <img style="width: 256px; height: 170px;" src="lighthouse-1024x680.jpg">
--- a/lighthouse-cli/test/fixtures/byte-efficiency/tester.html
+++ b/lighthouse-cli/test/fixtures/byte-efficiency/tester.html
@@ -57,7 +57,7 @@ function generateInlineScriptWithSize(sizeInBytes, firstContent = '', used = fal
 
   <!-- FAIL(optimized): image is not optimized -->
   <!-- PASS(responsive): image is used at full size -->
-  <img src="https://github.com/GoogleChrome/lighthouse/raw/751a84547bfff97d1f6c4533d1be2266908dfdc9/assets/lighthouse-logo_no-light.png">
+  <img src="http://localhost:10503/byte-efficiency/lighthouse-unoptimized.jpg">
 
   <!-- PASSWARN(optimized): image is JPEG optimized but not WebP -->
   <!-- FAIL(responsive): image is 25% used at DPR 2 -->
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -132,6 +132,21 @@ class OptimizedImages extends Gatherer {
   }
 
   /**
+   * @param {!Object} driver
+   * @param {!Array<!Object>} imageRecords
+   * @return {!Promise<!Array<!Object>>}
+   */
+  computeOptimizedImages(driver, imageRecords) {
+    return imageRecords.reduce((promise, record) => {
+      return promise.then(results => {
+        return this.calculateImageStats(driver, record)
+          .catch(err => ({failed: true, err}))
+          .then(stats => results.concat(Object.assign(stats, record)));
+      });
+    }, Promise.resolve([]));
+  }
+
+  /**
    * @param {!Object} options
    * @param {{networkRecords: !Array<!NetworRecord>}} traceData
    * @return {!Promise<!Array<!Object>}
@@ -140,15 +155,7 @@ class OptimizedImages extends Gatherer {
     const networkRecords = traceData.networkRecords;
     const imageRecords = OptimizedImages.filterImageRequests(options.url, networkRecords);
 
-    return imageRecords.reduce((promise, record) => {
-      return promise.then(results => {
-        return this.calculateImageStats(options.driver, record).catch(err => {
-          return {failed: true, err};
-        }).then(stats => {
-          return results.concat(Object.assign(stats, record));
-        });
-      });
-    }, Promise.resolve([])).then(results => {
+    return this.computeOptimizedImages(options.driver, imageRecords).then(results => {
       const successfulResults = results.filter(result => !result.failed);
       if (results.length && !successfulResults.length) {
         throw new Error('All image optimizations failed');
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -79,7 +79,7 @@ class OptimizedImages extends Gatherer {
 
       seenUrls.add(record._url);
       const isOptimizableImage = /image\/(png|bmp|jpeg)/.test(record._mimeType);
-      const isSameOrigin = URL.hostsMatch(pageUrl, record._url);
+      const isSameOrigin = URL.originsMatch(pageUrl, record._url);
       const isBase64DataUri = /^data:.{2,40}base64\s*,/.test(record._url);
 
       if (isOptimizableImage) {
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -108,13 +108,13 @@ class OptimizedImages extends Gatherer {
     // Get the request body for cross-origin images to circumvent canvas CORS protections
     if (!networkRecord.isSameOrigin && !networkRecord.isBase64DataUri) {
       const requestId = networkRecord.requestId;
-      uriPromise = Promise.resolve()
-        .then(_ => driver.sendCommand('Network.enable'))
-        .then(_ => driver.sendCommand('Network.getResponseBody', {requestId}))
-        .then(resp => {
-          const body = resp.base64Encoded ? resp.body : '';
-          return `data:${networkRecord.mimeType};base64,${body}`;
-        });
+      uriPromise = driver.sendCommand('Network.getResponseBody', {requestId}).then(resp => {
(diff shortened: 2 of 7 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -123,13 +140,15 @@ class OptimizedImages extends Gatherer {
...
+    return imageRecords.reduce((promise, record) => {
+      return promise.then(results => {
+        return this.calculateImageStats(options.driver, record).catch(err => {
+          return {failed: true, err};
+        }).then(stats => {
+          return results.concat(Object.assign(stats, record));
+        });
       });
-    })).then(results => {
+    }, Promise.resolve([])).then(results => {
```
ebidel: nit: it might be nice to break this reduce up from the `.then` chain. There's already a lot going on. You could also remove some of the `{ return ...}` in the arrow functions and just do an implicit return.

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -101,16 +103,31 @@ class OptimizedImages extends Gatherer {
...
-        jpegSize: isBase64DataUri ? stats.jpeg.base64 : stats.jpeg.binary,
-        webpSize: isBase64DataUri ? stats.webp.base64 : stats.webp.binary,
-      };
+    let uriPromise = Promise.resolve(networkRecord.url);
+
+    // Get the request body for cross-origin images to circumvent canvas CORS protections
+    if (!networkRecord.isSameOrigin && !networkRecord.isBase64DataUri) {
+      const requestId = networkRecord.requestId;
+      uriPromise = Promise.resolve()
+        .then(_ => driver.sendCommand('Network.enable'))
```
patrickhulce: conceptually yes, but we enable network in a couple places without disabling, so I'm somewhat ambivalent. (in fact if we disable network properly elsewhere this might not work... :/)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: yeah I wasn't pleased about adding another layer, will do

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
brendankenny: > in fact if we disable network properly elsewhere this might not work

the enabled domains are tracked now, so if you enable and then disable, but someone had already disabled before you, it won't actually be disabled.

I've been thinking about putting a check at the end of every pass to yell about gatherers that didn't clean up after themselves by disabling everything they enable :)

lighthouse-cli/test/fixtures/byte-efficiency/tester.html
```diff
@@ -55,6 +55,10 @@
   <!-- PASS(responsive): image is used at full size -->
   <img src="lighthouse-unoptimized.jpg">
 
+  <!-- FAIL(optimized): image is not optimized -->
+  <!-- PASS(responsive): image is used at full size -->
+  <img src="https://github.com/GoogleChrome/lighthouse/raw/751a84547bfff97d1f6c4533d1be2266908dfdc9/assets/lighthouse-logo_no-light.png">
```
brendankenny: any way we can do this without a network connection? It's nice to be able to run `npm run smoke` on an airplane :)

patrickhulce: is that ability worth writing and maintaining a second static server that you need to remember to bring up? :)

brendankenny: > is that ability worth writing and maintaining a second static server that you need to remember to bring up? :)

actually, now that I think about it, `static-server.js` already [does start up two servers](https://github.com/GoogleChrome/lighthouse/blob/cd82bc7d6e16881861cdfc4fea2f66c5a3d6ad01/lighthouse-cli/test/fixtures/static-server.js#L87-L88) :) The byte efficiency test is testing against `http://localhost:10200/byte-efficiency/tester.html`, so I believe you'll be able to use `http://localhost:10503` as a second origin.

It may still be a total pain, but it's worth a quick try to see if it will work

brendankenny:
this is a really nice way of handling cross origin images, btw :)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -101,20 +103,50 @@ class OptimizedImages extends Gatherer {
...
+        .then(_ => driver.sendCommand('Network.getResponseBody', {requestId}))
+        .then(resp => {
+          const body = resp.base64Encoded ? resp.body : '';
```
brendankenny: how often does this fail? Worth throwing here (so the image stats is an error) instead of going through with an empty string?

patrickhulce: yep good call, better now than cryptic later :)

lighthouse-cli/test/fixtures/byte-efficiency/tester.html
patrickhulce: ha so we do. done :)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -101,20 +103,50 @@ class OptimizedImages extends Gatherer {
...
-        webpSize: isBase64DataUri ? stats.webp.base64 : stats.webp.binary,
-      };
+    let uriPromise = Promise.resolve(networkRecord.url);
+
+    // Get the request body for cross-origin images to circumvent canvas CORS protections
+    if (!networkRecord.isSameOrigin && !networkRecord.isBase64DataUri) {
+      const requestId = networkRecord.requestId;
+      uriPromise = driver.sendCommand('Network.getResponseBody', {requestId}).then(resp => {
+        if (!resp.base64Encoded) {
+          throw new Error('expected image to be base64 encoded');
```
brendankenny: if this ends up in the `debugString` won't this be a bit confusing for the user since the image wouldn't have been a data uri? What about just "unable to access cross-origin image body" or something since that's the important bit from their perspective.

brendankenny: though looking at the [`debugString` construction](https://github.com/GoogleChrome/lighthouse/blob/174766fc53b8315d5d2290f4864c04ebd30834fd/lighthouse-core/audits/byte-efficiency/uses-optimized-images.js#L120), the individual error message isn't surfaced, so maybe it's fine. Your call.

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -101,20 +103,50 @@ class OptimizedImages extends Gatherer {
...
-      const isBase64DataUri = networkRecord.isBase64DataUri;
-      const base64Length = networkRecord.url.length - networkRecord.url.indexOf(',') - 1;
-      return {
-        originalSize: isBase64DataUri ? base64Length : networkRecord.resourceSize,
-        jpegSize: isBase64DataUri ? stats.jpeg.base64 : stats.jpeg.binary,
-        webpSize: isBase64DataUri ? stats.webp.base64 : stats.webp.binary,
-      };
+    let uriPromise = Promise.resolve(networkRecord.url);
+
+    // Get the request body for cross-origin images to circumvent canvas CORS protections
```
brendankenny: *total* nit, but could start the comment with the conditional to make the context super clear: `// For cross-origin images, get the request body to circumvent canvas CORS protections`

lighthouse-cli/test/fixtures/byte-efficiency/tester.html
```diff
@@ -55,6 +55,10 @@
   <!-- PASS(responsive): image is used at full size -->
   <img src="lighthouse-unoptimized.jpg">
 
+  <!-- FAIL(optimized): image is not optimized -->
+  <!-- PASS(responsive): image is used at full size -->
+  <img src="http://localhost:10503/byte-efficiency/lighthouse-unoptimized.jpg">
```
brendankenny: and reused the image, nice :) If we had better smokehouse comparators would you also be able to assert that this goes through your new cross origin path?

brendankenny: > If we had better smokehouse comparators would you also be able to assert that this goes through your new cross origin path

looking at the `extendedInfo`, looks like the answer is no. Worth adding a `isCrossOrigin` flag on the `extendedInfo` object?

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -82,9 +82,11 @@ class OptimizedImages extends Gatherer {
       const isSameOrigin = URL.hostsMatch(pageUrl, record._url);
```
brendankenny: does this need to be looking at origins instead of just host so the protocol is included?

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: done

lighthouse-cli/test/fixtures/byte-efficiency/tester.html
patrickhulce: eh yeah sure :)

brendankenny: knowing you, you'd love expectations like
```js
'uses-optimized-images': {
  score: false,
  extendedInfo: {
    value: {
      results: {
        1: {
          isCrossOrigin: true
        },
        length: 4
      }
    }
  }
},
```
:)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: hm does this actually taint CORS/require headers? you won't be able to do https -> http and if you're doing http -> https we'll complain elsewhere anyway. I'm cool leaving it since nowhere else compares origins, but I guess we should go on a hunt to find all cases it might matter :)

brendankenny: I *believe* it does, but we'll need to check

brendankenny: just checked and confirmed an http page requesting an image from the same host but over https does indeed taint the canvas

brendankenny: > if you're doing http -> https we'll complain elsewhere anyway

this is a fair point, but I think it'll just be a matter of time until we get an issue filed on this :) Testing cross origin seems like the right thing to do to me in this case, but also agree the inconsistency between this and other gatherers/audits doing host comparison is less than ideal

lighthouse-cli/test/fixtures/byte-efficiency/tester.html
brendankenny: looks like this won't work because it depends on the order of the network records coming in (it's slot 0 on travis for whatever reason). This might need to wait until we have function-based comparisons in smokehouse

PR #1856, opened 2017-03-13
Report: increase icon size for a11y

ebidel:
R: all

Expands on https://github.com/GoogleChrome/lighthouse/pull/1851 by increasing the size of the icons to 19px to "cheat" at WCAG2AA. See https://github.com/GoogleChrome/lighthouse/pull/1851#issuecomment-285905938

Here's what everything looks like now:

screen shot 2017-03-13 at 3 16 37 pm

Note, the score labels are set to slightly smaller than they should be but ramping them up to 19px makes them far too big. I've set them to bold 16px, which gets us pretty close.

I don't think we should spend too much more time tweaking the report if things are changing.

### DIFF
--- a/lighthouse-core/report/partials/cards.css
+++ b/lighthouse-core/report/partials/cards.css
@@ -30,6 +30,7 @@
   justify-content: center;
   align-items: center;
   border-bottom: 1px solid #ebebeb;
+  color: #4f4f4f;
 }
 .scorecard-value {
   font-size: 28px;
--- a/lighthouse-core/report/partials/critical-request-chains.css
+++ b/lighthouse-core/report/partials/critical-request-chains.css
@@ -51,11 +51,11 @@
 }
 
 .cnc-node__tree-hostname {
-  color: #767676;
+  color: #595959;
 }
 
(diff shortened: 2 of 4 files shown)


wardpeet:
Probably a good reason why but why aren't you using more css vars for the colors. There are a few duplicates like `#595959`

paulirish:
> I don't think we should spend too much more time tweaking the report if things are changing.

agree. 

i'd actually prefer to close this PR now so that we can start focusing on the report redesign instead of these fixes.   ok with you?

ebidel:
I think we should still iterate. I'm still indifferent on a bunch of stuff
in the new report and it's unclear when it would see a first release. This
is good for existing users!

XhmikosR:
I'm still seeing an issue with `subitem-result__points.score-average-bg`: the contrast ratio is 3.08:1 in your branch while in mine it was passing.

paulirish:
### embiggened!

XhmikosR:
@paulirish: this isn't valid WCAG2AA as per my comment above.

PR #1847, opened 2017-03-10
Collapse the 9 manifest PWA audits into 3 

paulirish:
_updated everything here march 22_

Instead of relying on topic grouping, this changes the {add to homescreen, splash screen, omnibox theming} audits to be all or nothing. (Scoring wise, if you're missing manifest `short_name`, you shouldn't really be getting points for the other manifest items.)

Screenshot of the PWA audits now:
image

(I stripped out the topics for readability as the [double nesting](https://cloud.githubusercontent.com/assets/39191/24228517/f391a434-0f31-11e7-88b9-c5f17b04d7b6.png) is a little funny)

### Functional changes

* Checking for `display` in a2hs/install-banner is back! Based on the resolution in https://github.com/GoogleChrome/lighthouse/issues/495#issuecomment-287270541, we accept standalone/fullscreen/minimal-ui.
* The 144 and 192 icon checks graduated to 192 and 512 respectively. This fixes the longstanding #291. Both before and now we're requiring sizes that are larger than you technically _need_, but that seems OK.
  * The 512 icon will catch some sites out. `pwa.rocks` for example fails now.
* Manifest `name` isn't accepted as a fallback if `short_name` is missing any longer. Some sites like svgomg fail on this now.
### Technical changes

* Manifest checks have moved to a computed artifact (`manifestValues`).
* All but one of `audits/manifest-*.js` have been nuked. (short-name-length remains).
  * Their tests have moved to either the `manifestValues` tests or the PWA audits tests.
* the "add to home screen" thing is now corrected referred to as "webapp install banner" in all the places
* Flipped back on a skipped test that is good now :)

PTAL!

PR #937, opened 2016-11-13
Adds single-use event listener audit

ebidel:
R: @paulirish @brendankenny @GoogleChrome/lighthouse 

Fixes #872

screen shot 2016-11-13 at 12 04 37 pm

PR #1857, opened 2017-03-13
remove explicit required generateAuditResult() call

brendankenny:
basically forever we've required audits to call `AuditName.generateAuditResult()` on the result they're returning.

```js
static audit(artifacts) {
  return MyGreatAudit.generateAuditResult({
    rawValue: false,
  });
}
```

This just moves the call to the caller of the audit (`Runner._runAudit`) instead, making writing an audit require just a little less boilerplate :)

```js
static audit(artifacts) {
  return {
    rawValue: false,
  };
}
```


lighthouse-cli/types/types.ts
```diff
@@ -1,13 +1,25 @@
...
-  debugString: string;
-  score: number;
+  debugString?: string;
+  score: boolean|number;
   error?: boolean;
   description: string;
   name: string;
   category: string;
   helpText?: string;
-  requiredArtifacts?: Array<string>;
```
ebidel: wait, no longer needed?

brendankenny: this is for the `AuditResult`, not `Audit`. I'm not sure why `requiredArtifacts` was on this in the first place

lighthouse-core/audits/audit.js
```diff
@@ -34,22 +34,24 @@ class Audit {
...
       error: true,
       debugString
     });
   }
 
   /**
-   * @param {!AuditResultInput} result
-   * @return {!AuditResult}
+   * @param {!Audit} audit
+   * @param {!AuditResut} result
```
paulirish: AuditResut

lighthouse-core/test/audits/speed-index-metric-test.js
```diff
@@ -74,7 +74,6 @@ describe('Performance: speed-index-metric audit', () => {
     const artifacts = mockArtifactsWithSpeedlineResult(SpeedlineResult);
 
     return Audit.audit(artifacts).then(response => {
-      assert.equal(response.displayValue, '845');
```
paulirish: i saw when you were debugging this..

looks like you nuked it because `.displayValue` is never explictly set?

paulirish: If that's true, what is this bit doing? https://github.com/GoogleChrome/lighthouse/blob/3fefe78dd1b4a1b50496152422130e55eeee8cc4/lighthouse-core/audits/audit.js#L60-L63

lighthouse-core/audits/audit.js
brendankenny: nice catch :)

lighthouse-core/test/audits/speed-index-metric-test.js
brendankenny: > looks like you nuked it because `.displayValue` is never explictly set?

yes. That line in `audit.js` does the same thing as it did before (so the final result output by Lighthouse will have a `displayValue` of `'845'`), but since the unit test is calling `.audit()` directly, it's not getting the version run through `generateAuditResult` and so won't have `displayValue` on it

lighthouse-core/audits/audit.js
brendankenny: done

lighthouse-core/test/audits/speed-index-metric-test.js
paulirish: roger that.

brendankenny:
I believe technically all current PRs will still work (no danger in calling `Audit.generateAuditResult()` on the results of itself), but you can remove `Audit.generateAuditResult()` from your audit if you feel like it :)

#1840, #1847, #1804, #1695, and #937 looks like they can be updated if desired

addyosmani:
Thanks for the ping on the page execution breakdowns PR. This change looks pretty sane. LGTM3.

Issue #1568, opened 2017-01-28
Migrate TTI and EIL audits to using TraceOfTab


brendankenny:
and off their homegrown filter of trace events. Will require being able to identify top level slices ourselves.

paulirish:
Also `getNavigationStartEvt` in pwmetrics-events.js could migrate onto the computed artifact


fwiw i've started the migration off traceviewer in this branch: [rmtraceviewer](https://github.com/GoogleChrome/lighthouse/tree/rmtraceviewer). it's incomplete.

brendankenny:
fixed in #1858, #2347, etc etc

PR #1858, opened 2017-03-14
use FMP from TraceOfTab in TTI calc

brendankenny:
With FMP timing available in the `TraceOfTab` computed artifact after #1549, this moves TTI calculation to using that rather than calling into the FMP audit.

Helps with #1568

### DIFF
--- a/lighthouse-core/test/audits/time-to-interactive-test.js
+++ b/lighthouse-core/test/audits/time-to-interactive-test.js
@@ -35,7 +35,7 @@ describe('Performance: time-to-interactive audit', () => {
       assert.equal(output.rawValue, 1105.8, output.debugString);
       assert.equal(output.displayValue, '1105.8ms');
       assert.equal(output.extendedInfo.value.expectedLatencyAtTTI, 20.724);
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/audits/time-to-interactive-test.js
```diff
@@ -35,7 +35,7 @@ describe('Performance: time-to-interactive audit', () => {
       assert.equal(output.rawValue, 1105.8, output.debugString);
       assert.equal(output.displayValue, '1105.8ms');
       assert.equal(output.extendedInfo.value.expectedLatencyAtTTI, 20.724);
-      assert.equal(output.extendedInfo.value.timings.fMP, 1099.5);
+      assert.equal(output.extendedInfo.value.timings.fMP, 1099.523);
```
brendankenny: Test update because this was truncated to one digit after the decimal point by the `first-meaningful-paint` audit before, now matches the `fmpTiming.toFixed(3)` found in this audit

patrickhulce: aw, snap watch out report here comes precision!

Issue #1780, opened 2017-02-25
Print -> Save as PDF : Pages truncated


ashrithkulai:
Click on export and select Print, in the chrome dialog chose save as PDF option. 
It is showing only 6 pages. 
The content after **"Using bytes efficiently"** is truncated in the saved pdf.

wardpeet:
Thanks for reporting!

ebidel:
I have a sneaky suspicion[ expanding the `<details>` before print](https://github.com/GoogleChrome/lighthouse/blob/9ac57cb05efd5b0aba6684047382fff670860ca0/lighthouse-core/report/scripts/lighthouse-report.js#L204-L221) is not fully working as expected. Especially as the report grow. 

Chrome appears to be generating the PDF before everything is expanded.

PR #1860, opened 2017-03-14
Fix: Save as PDF: Pages truncated

ZZhaoTireless:
Hi,

I did some researches of the issue: Print -> [Save as PDF : Pages truncated #1780](https://github.com/GoogleChrome/lighthouse/issues/1780), and here is my solution.

The main problem I found is, once the `window.print()` is called. The printing process will take current page height immediately. In fact, the function [ _setupExpandDetailsWhenPrinting](https://github.com/GoogleChrome/lighthouse/blob/9ac57cb05efd5b0aba6684047382fff670860ca0/lighthouse-core/report/scripts/lighthouse-report.js#L204-L221) works as expected in my tests, i.e. all details are labeled `open=true` when triggering `print` event. 

So the only problem is, for example, the page height is 2112px before the expanding, and the printing take it as 3 pages pdf file; however, after the expanding, the page height is 9230px, and it requires 11 pages pdf file. Then, the last 8 pages are truncated in this case.

I found two possibilities. 

First, I could solve it by adding a `min-height: 10000px ` under the `@media print `. It is obviously a bad idea. However, it works, so it confirms my hypophysis above.

Second, since my searching work implies that there is no way to change the page height dynamically in CSS or after `window.print()` is called, I would like to solve the problem through calling expand before calling `window.print()`. I left the collapse listener for the event when printing complete.

Thanks,

Zixu Zhao

### DIFF
--- a/lighthouse-core/report/scripts/lighthouse-report.js
+++ b/lighthouse-core/report/scripts/lighthouse-report.js
@@ -203,7 +203,6 @@ class LighthouseReport {
    */
   expandDetailsWhenPrinting() {
     const details = Array.from(document.querySelectorAll('details'));
-
     details.map(detail => detail.open = true);
   }
 
--- a/lighthouse-core/report/scripts/lighthouse-report.js
+++ b/lighthouse-core/report/scripts/lighthouse-report.js
@@ -37,7 +37,7 @@ class LighthouseReport {
   }
 
   _addEventListeners() {
-    this._setupExpandDetailsWhenPrinting();
+    this._setUpCollaspeDetailsAfterPrinting();
 
     const headerContainer = document.querySelector('.js-header-container');
     if (headerContainer) {
@@ -127,6 +127,7 @@ class LighthouseReport {
         this.sendJSONReport();
         break;
       case 'print':
+        this.expandDetailsWhenPrinting();
         window.print();
         break;
       case 'save-json': {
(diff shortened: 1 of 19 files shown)


lighthouse-core/report/scripts/lighthouse-report.js
```diff
@@ -196,26 +197,31 @@ class LighthouseReport {
...
-        details.map(detail => detail.open = true);
-      });
       window.addEventListener('afterprint', _ => {
         details.map(detail => detail.open = false);
       });
     } else {
       // Note: while FF has media listeners, it doesn't fire when matching 'print'.
       window.matchMedia('print').addListener(mql => {
-        details.map(detail => detail.open = mql.matches);
+        if(!mql.matches) details.map(detail => detail.open = mql.matches);
```
ebidel: spacing: `if (!mql.matches)`. I'm also a fan of `{}`, even for one liners :)

lighthouse-core/report/scripts/lighthouse-report.js
```diff
@@ -196,26 +197,31 @@ class LighthouseReport {
...
+  /**
+   * Sets up listeners to collapse audit `<details>` when the user closes the
+   * print dialog, all `<details>` are collapsed.
+   */
+  _setUpCollaspeDetailsAfterPrinting() {
     const details = Array.from(document.querySelectorAll('details'));
 
     // FF and IE implement these old events.
     if ('onbeforeprint' in window) {
-      window.addEventListener('beforeprint', _ => {
```
ebidel: one thing we loose out on with this is `ctrl+p` (or printing from the browser chrome) no longer expands the dropdowns. Meaning users have to go through the "Export" dropdown for printing. That might be unavoidable though b/c of the bug this PR is trying to address.

PR #1861, opened 2017-03-14
test: deflake blocking links

patrickhulce:
To continue the nightmare that is the blocking stylesheets audit false positives, the last PR helped in the wild, but smokehouse added delays to *all* other blocking sheets that pushed out when our detection script could fire (thus making detection rather flaky). This is just an bad approach, and I still advocate for parsing the document we received at some point instead of all these hoops, but this will make travis more reliable for now.

Real fix should be made in protocol to add initiator information that we need.

### DIFF
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -21,12 +21,13 @@
 const http = require('http');
 const path = require('path');
 const fs = require('fs');
+const parseQueryString = require('querystring').parse;
 const parseURL = require('url').parse;
 
 function requestHandler(request, response) {
   const requestUrl = parseURL(request.url);
   const filePath = requestUrl.pathname;
-  const queryString = requestUrl.search;
+  const queryString = requestUrl.search && parseQueryString(requestUrl.search.slice(1));
   let absoluteFilePath = path.join(__dirname, filePath);
 
   if (filePath === '/zone.js') {
@@ -64,10 +65,9 @@ function requestHandler(request, response) {
     }
     response.writeHead(statusCode, headers);
 
-    if (queryString && queryString.includes('delay')) {
+    if (queryString && typeof queryString.delay !== 'undefined') {
       response.write('');
-      const matched = queryString.match(/delay=(\d+)/);
-      const delay = (matched && parseInt(matched[1])) || 2000;
+      const delay = parseInt(queryString.delay) || 2000;
       return setTimeout(finishResponse, delay, data);
     }
     finishResponse(data);
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -66,7 +66,9 @@ function requestHandler(request, response) {
 
     if (queryString && queryString.includes('delay')) {
       response.write('');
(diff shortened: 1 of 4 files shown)


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -59,7 +59,7 @@ module.exports = [
         extendedInfo: {
           value: {
             results: {
-              length: 3
+              length: 4
```
patrickhulce: we actually had 4 failures before, just 2 of them were de-duped with the same URL, now that I added different delays it should be 4.

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -71,7 +71,7 @@ class LinkBlockingFirstPaintAudit extends Audit {
       endTime = Math.max(item.endTime, endTime);
 
       return {
-        url: URL.getDisplayName(item.tag.url),
+        url: URL.getDisplayName(item.tag.url, {preserveQuery: true}),
```
patrickhulce: making it more sane to read the failure output for which one failed

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -21,12 +21,13 @@
 const http = require('http');
 const path = require('path');
 const fs = require('fs');
+const parseQueryString = require('querystring').parse;
```
brendankenny: ha, had no idea this was a node core API

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -64,9 +65,10 @@ function requestHandler(request, response) {
     }
     response.writeHead(statusCode, headers);
 
-    if (queryString && queryString.includes('delay')) {
+    if (queryString && typeof queryString.delay !== 'undefined') {
       response.write('');
-      return setTimeout(finishResponse, 2000, data);
+      const delay = parseInt(queryString.delay) || 2000;
```
brendankenny: `parseInt(queryString.delay, 10)`, just in case :)

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -64,9 +65,10 @@ function requestHandler(request, response) {
     }
     response.writeHead(statusCode, headers);
 
-    if (queryString && queryString.includes('delay')) {
+    if (queryString && typeof queryString.delay !== 'undefined') {
```
brendankenny: maybe add a comment here for future folks about how this will work? You can set the value for `delay` or any non-numeric value will default to 2000 ms?

patrickhulce: done

lighthouse-cli/test/fixtures/static-server.js
patrickhulce: done

patrickhulce:
>  just moving the script after the stylesheets and removing the delay?

Yeah that's the critical part which I originally discovered by varying the delays, but figured it was still better to have some robustness of different delays and clearer expression of how it works can't hurt :)

PR #1864, opened 2017-03-15
Fix service worker gatherer by waiting for active state

### LINKED ISSUE #1815: Flakiness with "Registers a service worker" audit
We have installed service worker, but lighthouse report saying service worker not installed. 
Site URL : https://fuzecaresystem.com/

Please check screenshot:
capture

wardpeet:
Fixes #1815 where we don't wait until the service worker is active. 
Happens mostly on slow connections

not sure if I should mock and create a test for this?

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -275,15 +275,15 @@ class Driver {
       const versionUpdatedListener = data => {
         // find a service worker with runningStatus that looks like active
         // on slow connections the serviceworker might still be installing
-        const nonRedundantServiceWorkers = data.versions.filter(sw => {
+        const activateCandidates = data.versions.filter(sw => {
           return sw.status !== 'redundant';
         });
-        const activeServiceWorker = nonRedundantServiceWorkers.find(sw => {
+        const activeServiceWorker = activateCandidates.find(sw => {
           return sw.status === 'activated';
         });
 
-        const hasActiveServiceWorker = nonRedundantServiceWorkers.length && activeServiceWorker;
-        if (!nonRedundantServiceWorkers.length || hasActiveServiceWorker) {
+        const hasActiveServiceWorker = activeServiceWorker;
+        if (!activateCandidates.length || hasActiveServiceWorker) {
           this.off('ServiceWorker.workerVersionUpdated', versionUpdatedListener);
           this.sendCommand('ServiceWorker.disable')
             .then(_ => resolve(data), reject);
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -278,11 +278,10 @@ class Driver {
         const activateCandidates = data.versions.filter(sw => {
           return sw.status !== 'redundant';
         });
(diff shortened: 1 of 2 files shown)


paulirish:
Do we have a fairly solid repro for the bug? 

I'm trying it with heavier throttling on both caltrainschedule and fuzecaresystem and i can't repro the bug.

wardpeet:
if you like I can add a failing test case to the mix. In my test case you can see the issue at end. First it's installing and the audit goes right through it, now it waits for activated.

I throttled my connection with charles and could reproduce it quite often. One time I didn't have a service worker, the other time I did.

wardpeet:
@paulirish here is a url that I could reproduce it with
https://fuzecaresystem.com/ see #1815  for more info

wardpeet:
Referencing #1815

stramel:
Can we get this merged in soon?

lighthouse-core/gather/driver.js
```diff
@@ -263,10 +263,25 @@ class Driver {
...
   getServiceWorkerVersions() {
     return new Promise((resolve, reject) => {
-      this.once('ServiceWorker.workerVersionUpdated', data => {
-        this.sendCommand('ServiceWorker.disable')
-          .then(_ => resolve(data), reject);
-      });
+      const versionUpdatedListener = data => {
+        // find a service worker with runningStatus that looks like active
+        // on slow connections the serviceworker might still be installing
+        const nonRedundantServiceWorkers = data.versions.filter(sw => {
```
ebidel: suggest `nonRedundantServiceWorkers` -> `activateCandidates`

lighthouse-core/gather/driver.js
```diff
@@ -263,10 +263,25 @@ class Driver {
...
+          return sw.status !== 'redundant';
+        });
+        const activeServiceWorker = nonRedundantServiceWorkers.find(sw => {
+          return sw.status === 'activated';
+        });
+
+        const hasActiveServiceWorker = nonRedundantServiceWorkers.length && activeServiceWorker;
```
ebidel: Do you need this boolean? Isn't `activeServiceWorker` enough?

wardpeet: when reading the code over again it seems like you are right

lighthouse-core/gather/driver.js
wardpeet: will update

lighthouse-core/gather/driver.js
```diff
@@ -272,10 +272,25 @@ class Driver {
...
+        // find a service worker with runningStatus that looks like active
+        // on slow connections the serviceworker might still be installing
+        const activateCandidates = data.versions.filter(sw => {
+          return sw.status !== 'redundant';
+        });
+        const activeServiceWorker = activateCandidates.find(sw => {
+          return sw.status === 'activated';
+        });
+
+        const hasActiveServiceWorker = activeServiceWorker;
```
ebidel: nite: let's just remove and use `activeServiceWorker`

stramel:
LGTM :+1:

PR #1866, opened 2017-03-16
feat(cli): add support for custom trace categories

### LINKED ISSUE #1663: WPT compat: include additional trace categories
LH is missing the following trace categories captured by WPT:
```
+      "v8.execute",
+      "devtools.timeline.frame",
+      "devtools.timeline.stack",
+      "disabled-by-default-blink.feature_usage",
```

patrickhulce:
fixes #1663


lighthouse-core/gather/driver.js
```diff
@@ -570,9 +570,14 @@ class Driver {
       });
   }
 
-  beginTrace() {
+  /**
+   * @param {{traceCategories: string=}=} flags
+   */
+  beginTrace(flags) {
+    const additionalCategories = flags && flags.traceCategories && flags.traceCategories.split(',');
+    const traceCategories = new Set(this._traceCategories.concat(additionalCategories || []));
```
paulirish: i kinda prefer moving this ` || []` to the above statement, so we can rely on it as an array.

paulirish: let's just drop in `const _uniq = arr => Array.from(new Set(arr));` and use it so we're not tossing this type from Set back to array a few lines later.

lighthouse-cli/bin.ts
```diff
@@ -78,6 +79,7 @@ const cliFlags = yargs
     'save-artifacts': 'Save all gathered artifacts to disk',
     'list-all-audits': 'Prints a list of all available audits and exits',
     'list-trace-categories': 'Prints a list of all required trace categories and exits',
+    'trace-categories': 'Additional categories to capture with the trace (comma-delimited).',
```
paulirish: how about `addl-trace-categories` ?  to clarify that this doesn't overwrite the defaults

lighthouse-core/gather/driver.js

lighthouse-cli/bin.ts
patrickhulce: `additional-` but done :)

lighthouse-core/gather/driver.js
```diff
@@ -570,9 +572,15 @@ class Driver {
       });
   }
 
-  beginTrace() {
+  /**
+   * @param {{additionalTraceCategories: string=}=} flags
+   */
+  beginTrace(flags) {
```
ebidel: nit: could use some fancy default params: `flags = {additionalTraceCategories: ''}` 
then: `const additionalCategories = flags.additionalTraceCategories.split(',').filter(String);`

patrickhulce: true, but would still need the additionalTraceCategories check so just gets rid of `flags &&` in favor of ` = {additionalTraceCategories: ''}` gotta save those bytes you know ;)

PR #1868, opened 2017-03-16
feat(image-usage): add support for CSS images

### LINKED ISSUE #978: DBW: site uses responsive images
## Tracked Features
- [X] ~Detect`img` elements that use assets that are too large for their displayed size (#1497)~
- [ ] Detect images used in CSS that are too large for their displayed size

## ORIGINAL
It would be nice to have an audit for testing if a site sends responsive images. We could test for different common viewports and DPR and see if the images served by the page vary in size.

Suggest techniques like ``, css `image-set()`, and `<picture>`.

Perhaps it could share a gatherer with https://github.com/GoogleChrome/lighthouse/issues/718

patrickhulce:
fixes #978

Supports an MVP of responsive CSS images. The area here is a lot more gray than with HTML image elements, so I've kept it very limited in scope, but I believe it should cover the most naive case that's likely to have bloated images.

After this lands, I'd like to do some cleanup so that we're not potentially fetching images multiple times in the optimized image gatherer and this one.


lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -36,9 +36,45 @@ function collectImageElementInfo() {
       clientHeight: element.clientHeight,
       naturalWidth: element.naturalWidth,
       naturalHeight: element.naturalHeight,
+      isCss: false,
       isPicture: element.parentElement.tagName === 'PICTURE',
     };
   });
+
+  // Only match basic background-image: url("http://host/image.jpeg") declarations
+  const ALLOWED_IMAGE = /^\s*url\("([^"]+)"\)\s*$/;
```
paulirish: lets do a `trim()` on the string before running this regex so we can drop the `\s*` on the ends

paulirish: quotes aren't required inside of a url().. And i personally rarely use them.
also some people use single quotes.

so let's allow these combos.  this guy (thanks SO) seems good:

```regex
/(?:\(['"]?)(.*?)(?:['"]?\))/
```

paulirish: and naming nit:  how about `css_url_regex` instead of `allowed_image`

patrickhulce: in getComputedStyle Chrome normalizes the URL to full URL in quotes :) in fact the spacing is probably totally useless too since I bet those are eliminated as well (just not as thoroughly tested by me)

paulirish:
Ah that's cool to know. :)

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -36,9 +36,46 @@ function collectImageElementInfo() {
...
       isPicture: element.parentElement.tagName === 'PICTURE',
     };
   });
+
+  // Chrome normalizes background image style from getComputedStyle to be an absolute URL in quotes.
+  // Only match basic background-image: url("http://host/image.jpeg") declarations
+  const CSS_URL_REGEX = /^url\("([^"]+)"\)$/;
+  // Only find images that aren't specifically scaled
+  const CSS_SIZE_REGEX = /(auto|contain|cover)/;
+  const cssImages = [...document.querySelectorAll('*')].reduce((images, element) => {
```
ebidel: we should probably make this shadow dom-aware: `document.querySelectorAll('html /deep/ *')`.

https://github.com/GoogleChrome/lighthouse/blob/b06e808a5eb6b0c077bd0d855b9ff469aad5b056/lighthouse-core/gather/gatherers/dobetterweb/domstats.js#L111

ebidel: Collecting all the dom nodes on the page + running `gCS` on them....wowza. How heavy is that?

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -36,9 +36,46 @@ function collectImageElementInfo() {
...
+    if (images.length - differentImages.length > 2) {
+      return differentImages;
+    }
+
+    images.push({
+      src: url,
+      clientWidth: element.clientWidth,
+      clientHeight: element.clientHeight,
+      naturalWidth: Number.MAX_VALUE,
+      naturalHeight: Number.MAX_VALUE,
```
ebidel: Is this b/c "css images have no natural size"? Might want to add that comment here too. I was expecting these to be `element.clientWidth/Height`.

lighthouse-core/gather/gatherers/image-usage.js
patrickhulce: heh yeah...you have any better ideas? i'm all ears :D

lighthouse-core/gather/gatherers/image-usage.js
patrickhulce: yup, will do

brendankenny:
Just to be clear, this isn't really a problem on many sites, even ones with a similar number of DOM nodes, so not sure exactly what's going on yet.

And of course it's next to impossible to profile the code we run in the context of the browser to find out...

patrickhulce:
@paulirish @brendankenny more feedback?

patrickhulce:
bump PTAL :)

paulirish:
> Still would be interesting to see where the time was going if we could easily profile the js code we run in the browser context.

we can if we just keep the trace on during this time.. just some gatherer juggling. :)

Issue #1876, opened 2017-03-17
Move perf metrics into performance section


paulirish:
once **Add TTI < 10s audit for PWA** #1840 is done, we can move perf metrics to their real home, out of the PWA section.

addyosmani:
Do we have a public note about why we're extending TTI's budget to < 10s? (just for historical context as folks look at how/why metric budgets have evolved).

paulirish:
@addyosmani there's two separate "audits" here

A) A boolean audit saying "is the TTI faster than 10s?". This is to mirror this item in the baseline checklist: 
![image](https://cloud.githubusercontent.com/assets/39191/24376450/c24637a6-12f0-11e7-8a5f-363349fef422.png)

B) sharing the value and scoring of TTI, along with the other key perf metrics. The scoring over here is separate and we can handle completely independently from the PWA requirement. FYI: With our current scoring curve (which honestly hasn't been validated), a 9900ms TTI gets a score of 11%. That seems a tad low, but I'm also okay with Lighthouse perf section having much higher standards for perf than the hard boolean requirements of PWA-ness.

patrickhulce:
This is already done in the categories landing of config v2. OK to close @paulirish or was there more to it (UI perhaps?) than just moving them out?

paulirish:
yah sgtm.

Issue #1875, opened 2017-03-17
Move accessibility into own top-level category


paulirish:
Rob this probably makes sense to do in your mega-PR adding all the new checks.

patrickhulce:
FWIW I don't think this really needs to be done prior to the report refactor. I've already done it in my config v2 branch with little effort.

paulirish:
Add patrick to this so he can include the category/topic thing as part of config v2.

ebidel:
Can we consider this taken care of then? v1 supports it now after Rob's PR and Patrick has it in v2.

paulirish:
yup. good call.

Issue #1874, opened 2017-03-17
Remove concept of sub-aggregations


paulirish:

patrickhulce:
Also don't necessarily see this as blocking though it'll be nice cleanup to do afterwards/to simplify :)

Issue #1826, opened 2017-03-07
[proposal] harmonize command line flags and config file(s)


brendankenny:
We need to simplify how options are specified and how you can figure out what options an individual gatherer or audit will be affected by to help users and also (hopefully) help keep our code clearer.

A few steps in that direction I've been thinking about, mostly leaning toward eslint-style configuration:
- have every flag on the command line also be able to be specified in the config
- use the same names for CLI flags and options in the config
- have some subset of these available per-pass where it makes sense (e.g. override throttling)
- have a clear hierarchy over what value wins when set at different levels (see e.g. [eslint's "complete configuration hierarchy"](http://eslint.org/docs/user-guide/configuring), which goes inline (per-pass option for us?) > CLI flag > config file)
- support some sort of [`--print-config`](http://eslint.org/docs/user-guide/command-line-interface#--print-config) which prints a very verbose output of exactly what will be run with the current options from the current location. This is the last safeguard so you can tell what really is going to happen when you run lighthouse both for these flags and for extendable configs added in #1731

The other side of the first bullet point is allowing every top-level config option to also be specified on the command line, which is something eslint allows. I'm not sure how helpful that really is in practice, but they do at least namespace the options behind e.g. `eslint --rule 'quotes: [2, double]'` so there isn't a total explosion of possible CLI flags

patrickhulce:
I was about to add "All options are overridable from CLI" to my features in the doc 😄

wardpeet:
would you go for the same name as in cli or make is camelCase what yargs does in code?
`disable-storage-reset` => `disableStorageReset`?

PR #1880, opened 2017-03-20
[WIP] feat: base config v2 functionality

patrickhulce:
Implements the base functionality for the move to config v2 upon which all other features for the new format can be implemented.

[design doc](https://docs.google.com/document/d/1cN_m591PAQF4tkEf5nJ5M93pI2tM_U64XolmwpOuiUU/edit?usp=sharing)
related #1874 #1875 #1876 #1463 #1826

Missing because WIP (will be fixed before merge):
- ~category description text~
- ~jsdocs~
- ~tests~
- consensus :)

Missing because MVP:
- settings object
- audit/gatherer options
- customizable scoring functions (both proposed are already implemented just hardcoded atm)
- handling of auditResults/artifacts/trace files, planned to be split from config entirely in v2

Current look:

image

### DIFF
--- a/lighthouse-core/config/v2/config.js
+++ b/lighthouse-core/config/v2/config.js
@@ -1,3 +1,21 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
 const path = require('path');
 const log = require('../../lib/log');
 
@@ -14,8 +32,7 @@ class ConfigV2 {
       configPath = path.resolve(__dirname, defaultConfigPath);
     }
 
-    const configDir = typeof configPath === 'string' ?
-        path.dirname(configPath) : process.cwd();
+    const configDir = typeof configPath === 'string' ? path.dirname(configPath) : process.cwd();
     if (!path.isAbsolute(configDir)) {
       throw new Error('Must pass an absolute path to config');
     }
@@ -144,7 +161,7 @@ class ConfigV2 {
         extendedConfigPath,
         path.resolve(configDir, extendedConfigPath),
         path.resolve(process.cwd(), extendedConfigPath),
-      ])
+      ]);
       extendedConfigJson = ConfigV2._require(extendedConfigPath);
     }
 
@@ -209,7 +226,7 @@ class ConfigV2 {
         implementation = ConfigV2._require(definition.path);
       }
 
-      return Object.assign({}, definition, {implementation})
+      return Object.assign({}, definition, {implementation});
     });
   }
 
@@ -225,7 +242,10 @@ class ConfigV2 {
     const gathererIds = new Set(gatherers.map(item => item.id));
     const usedGathererIds = new Set();
     const usedGathererNames = new Set(['traces', 'networkRecords']);
-    const requestedGathererNames = new Set(_flatten(audits.map(audit => audit.implementation.meta.requiredArtifacts)));
+
+    const allRequiredArtifacts = audits.map(audit => audit.implementation.meta.requiredArtifacts);
+    const requestedGathererNames = new Set(_flatten(allRequiredArtifacts));
+
     const passDefinitions = ConfigV2.objectToArray(passesObject);
     const passes = passDefinitions.map(definition => {
       const foundGatherers = definition.gatherers.map(id => {
@@ -247,14 +267,14 @@ class ConfigV2 {
       log.warn('config', `Gatherers are unused: ${unused.join(', ')}`);
     }
 
-    const usedButNotNeeded = _differenceAsArray(usedGathererNames, requestedGathererNames);
-    if (usedButNotNeeded.length) {
-      log.warn('config', `Gatherers were configured but not needed: ${usedButNotNeeded.join(', ')}`);
+    const usedNotNeeded = _differenceAsArray(usedGathererNames, requestedGathererNames);
+    if (usedNotNeeded.length) {
+      log.warn('config', `Gatherers were configured but not needed: ${usedNotNeeded.join(', ')}`);
     }
 
-    const neededButNotGathered = _differenceAsArray(requestedGathererNames, usedGathererNames);
-    if (neededButNotGathered.length) {
-      log.warn('config', `Gatherers were needed but not configured: ${neededButNotGathered.join(', ')}`);
+    const neededNotUsed = _differenceAsArray(requestedGathererNames, usedGathererNames);
+    if (neededNotUsed.length) {
+      log.warn('config', `Gatherers were needed but not configured: ${neededNotUsed.join(', ')}`);
     }
 
     return passes;
--- a/lighthouse-core/config/v2/config.js
+++ b/lighthouse-core/config/v2/config.js
@@ -14,35 +14,51 @@ class ConfigV2 {
       configPath = path.resolve(__dirname, defaultConfigPath);
     }
 
+    const configDir = typeof configPath === 'string' ?
+        path.dirname(configPath) : process.cwd();
+    if (!path.isAbsolute(configDir)) {
+      throw new Error('Must pass an absolute path to config');
+    }
+
     // Perform a shallow clone so we can adjust gatherers and audits
     configJson = Object.assign({}, configJson);
 
     // Resolve the paths of audits and gatherers
-    configJson.gatherers = ConfigV2.resolvePaths(configJson.gatherers, configPath, [
+    configJson.gatherers = ConfigV2.resolvePaths(configJson.gatherers, configDir, [
       path.join(__dirname, '../../gather/gatherers'),
     ]);
-    configJson.audits = ConfigV2.resolvePaths(configJson.audits, configPath, [
+    configJson.audits = ConfigV2.resolvePaths(configJson.audits, configDir, [
       path.join(__dirname, '../../audits'),
     ]);
 
     // Extend only after our paths have been resolved
-    configJson = ConfigV2.extendIfNecessary(configJson, configPath);
+    configJson = ConfigV2.extendIfNecessary(configJson, configDir);
 
     this._json = configJson;
     this._gatherers = ConfigV2.collectImplementations(configJson.gatherers);
     this._audits = ConfigV2.collectImplementations(configJson.audits);
-    this._passes = ConfigV2.computePasses(configJson, this._gatherers, this._audits);
+    this._passes = ConfigV2.computePasses(configJson.passes, this._gatherers, this._audits);
     this._report = configJson.report;
   }
 
+  /**
+   * Returns a deep clone of the final extended config JSON.
+   * @return {!Object}
+   */
   asJson() {
     return JSON.parse(JSON.stringify(this._json));
   }
 
+  /**
+   * @return {!Array<{id: string, implementation: !Audit}>}
+   */
   get audits() {
     return [...this._audits];
   }
 
+  /**
+   * @return {!Array<{id: string, gatherers: !Array<!Gatherer>}>}
+   */
   get passes() {
     return [...this._passes];
   }
@@ -51,6 +67,29 @@ class ConfigV2 {
     return this._report;
   }
 
+  /**
+   * Creates a new ConfigV2 object, used to stub for testing.
+   * @param {Object=} json
+   * @param {string=} path
+   */
+  static _createInstance(json, path) {
+    return new ConfigV2(json, path);
+  }
+
+  /**
+   * Requires the given file, used to stub for testing.
+   * @param {string} path
+   * @return {*}
+   */
+  static _require(path) {
+    return require(path);
+  }
+
+  /**
+   * Returns the first path that resolves via require.resolve or throws.
+   * @param {!Array<string>} paths
+   * @return {string}
+   */
   static _tryResolveUntilSuccess(paths) {
     for (const path of paths) {
       try {
@@ -61,6 +100,11 @@ class ConfigV2 {
     throw new Error(`Unable to locate ${paths[0]}`);
   }
 
+  /**
+   * Recursively merges all of the properties from extension into base.
+   * @param {*} base
+   * @param {*} extension
+   */
   static _mergeObjects(base, extension) {
     if (!base ||
         !extension ||
@@ -79,7 +123,13 @@ class ConfigV2 {
     return base;
   }
 
-  static extendIfNecessary(configJson, configPath) {
+  /**
+   * Computes the final extended JSON or returns the original if no extension was needed.
+   * @param {!Object} configJson
+   * @param {string} configDir
+   * @return {!Object}
+   */
+  static extendIfNecessary(configJson, configDir) {
     if (!configJson.extends) {
       return configJson;
     }
@@ -88,21 +138,29 @@ class ConfigV2 {
     let extendedConfigPath = configJson.extends;
     if (extendedConfigPath === 'lighthouse:default') {
       extendedConfigJson = defaultConfigJson;
-      extendedConfigPath = defaultConfigPath;
+      extendedConfigPath = path.resolve(__dirname, defaultConfigPath);
     } else {
       extendedConfigPath = ConfigV2._tryResolveUntilSuccess([
         extendedConfigPath,
-        path.resolve(configPath, extendedConfigPath),
+        path.resolve(configDir, extendedConfigPath),
         path.resolve(process.cwd(), extendedConfigPath),
       ])
-      extendedConfigJson = require(extendedConfigPath);
+      extendedConfigJson = ConfigV2._require(extendedConfigPath);
     }
 
-    const extendedConfig = new ConfigV2(extendedConfigJson, extendedConfigPath);
+    const extendedConfig = ConfigV2._createInstance(extendedConfigJson, extendedConfigPath);
     return ConfigV2._mergeObjects(extendedConfig.asJson(), configJson);
   }
 
-  static resolvePaths(object, configPath, searchPaths = []) {
+  /**
+   * Traverses all the keys in the object and sets the path property of the child object to the
+   * resolved path of the mentioned file. Throws if no file could be found for a child.
+   * @param {!Object} object
+   * @param {string} configDir
+   * @param {Array<string>=} searchPaths
+   * @return {!Object} A copy of the input object with path set for all children.
+   */
+  static resolvePaths(object, configDir, searchPaths = []) {
     object = Object.assign({}, object);
     Object.keys(object).forEach(key => {
       // We don't need to resolve objects that already have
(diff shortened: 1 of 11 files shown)


lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,182 @@
...
+
+class ConfigV2 {
+  constructor(configJson, configPath) {
+    if (!configJson) {
+      configJson = defaultConfigJson;
+      configPath = path.resolve(__dirname, defaultConfigPath);
+    }
+
+    // Perform a shallow clone so we can adjust gatherers and audits
+    configJson = Object.assign({}, configJson);
```
paulirish: I think you'll need to do parse/stringify otherwise the original configJson object can be mutated, right?

patrickhulce: I do shallow copies on any children I mutate and whenever this is exposed to outside for potential mutation it's parsed/stringified in `asJson`, this was mainly to preserve the ability to call from javascript with `audits: {custom: {implementation: MyAuditClass}}`

patrickhulce: or at least I try to do shallow copies everywhere, we'll see when I add tests :)

lighthouse-core/runner.js
```diff
@@ -134,14 +136,21 @@ class Runner {
...
         return {
           lighthouseVersion: require('../package').version,
           generatedTime: (new Date()).toJSON(),
           initialUrl: opts.initialUrl,
           url: opts.url,
-          audits: formattedAudits,
+          audits: resultsById,
           artifacts: runResults.artifacts,
           runtimeConfig: Runner.getRuntimeConfig(opts.flags),
+          categories,
```
paulirish: should we put this inside a `report` object on the result payload?  or title it `reportCategories` to communicate clearly that it's just presentational?

patrickhulce: Yeah I'm in favor of restructuring this object for v2, should probably add a section to the doc to discuss format

I was thinking something like the following?
```
{
  report: {
    meta: {version, time, url}, // stuff thats currently top-level
    config, // full config used (after extension, cli overrides, etc)
    score, // final overall score
    categories, // heirarchy of computed scores/results
    audits // individual results by audit id
  } // the stuff that gets written to report.json and whose contents can be used to render the HTML report without any additional information
  artifacts: {gatherers, trace, perfLog/devtoolsMessages} // used to quickly re-run analysis/perform later deep dives/debug/etc
}
```

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  static objectToList(object) {
+    return Object.keys(object).reduce((list, id) => {
+      list.push(Object.assign({id}, object[id]));
+      return list;
+    }, []).sort((itemA, itemB) => itemA.order - itemB.order);
+  }
+
+  static getGathererImplementations(gatherersObject) {
```
paulirish: this and `getAuditImplementations` appear to be the same. You leaving them WET for now expecting them to change in the future?

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,198 @@
...
+        usedGathererNames.add(gatherer.implementation.name);
+        return gatherer;
+      });
+
+      return Object.assign({}, definition, {gatherers: foundGatherers});
+    });
+
+    if (gathererIds.size !== usedGathererIds.size) {
+      const unused = _subtract(gathererIds, usedGathererIds);
+      log.warn('config', `Gatherers are unused: ${unused.join(', ')}`);
```
paulirish: is this a bad thing? or just noting they are being culled from the working config based on the selection.

I'm fine with `warn` but the text could be more neutral to indicate its okay. (I'm expecting that this will log a lot when you do `--perf`)

paulirish:
Super excited about this work.  :)

On the whole I'm super excited about the new declarative config. The v2/config is a big upgrade over the current config.

Let's get some tests in place. IMO Category text can wait for another PR. 

I spoke with @brendankenny  and we're thinking of him enabling the v2 flag which enables the DOM based report renderer. This will start as an empty page and we'll build it out from there.

lighthouse-core/config/v2/config.js
patrickhulce: Seemed like a good thing to warn about in case you misname or something, this function is likely to experience the most change when we add `only`/`skip` etc, so I'd say we can re-evaluate or just remove in the filtered case.

lighthouse-core/config/v2/config.js
patrickhulce: So they are, haha I'll DRY up. It was copypasta I expected to modify but didn't need to in the end

lighthouse-core/config/v2/config.js
brendankenny: > or at least I try to do shallow copies everywhere, we'll see when I add tests :)

test will be good :) `assert.notDeepStrictEqual` might be sufficient, but we could also do something with deep `Object.freeze` and see what throws (in strict mode)

Especially important for the browserified/lighthouse-as-a-module case since when loading the config json via `require()` any changes to the original object is persisted to the next use of it.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,60 @@
...
+      return Object.assign({}, category, {children, score});
+    });
+
+    const scoreByCategory = categories.reduce((scores, category) => {
+      scores[category.id] = category.score;
+      return scores;
+    }, {});
+
+    const score = this._geometricMean([
+      {score: scoreByCategory.pwa, weight: 5},
```
brendankenny: hard coding category names and weights here seems like a mistake. Were you waiting to add a `scoring` property? Waiting is good, but can you just add `weight` to the category objects and skip this? (I'll comment in the design doc)

I also think using `_geometricMean` here is premature until we've had a real discussion on scoring, but happy to have it here for now :P

lighthouse-core/runner.js
brendankenny: agree that top level `categories` here is a bit confusing. Can we call it `reportCategories` until we settle on a result object shape?

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,308 @@
...
+      "path": "dobetterweb/tags-blocking-first-paint"
+    },
+    "websql": {
+      "path": "dobetterweb/websql"
+    }
+  },
+
+  "passes": {
+    "defaultPass": {
+      "order": 10,
```
brendankenny: can we kick `order` from the MVP so we can discuss it in a separate PR? I'm still not really a fan and with no motivating code that needs it yet there's no rush

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,186 @@
...
+    // Resolve the paths of audits and gatherers
+    configJson.gatherers = ConfigV2.resolvePaths(configJson.gatherers, configPath, [
+      path.join(__dirname, '../../gather/gatherers'),
+    ]);
+    configJson.audits = ConfigV2.resolvePaths(configJson.audits, configPath, [
+      path.join(__dirname, '../../audits'),
+    ]);
+
+    // Extend only after our paths have been resolved
+    configJson = ConfigV2.extendIfNecessary(configJson, configPath);
```
brendankenny: I kind of want to punt extensions from the MVP as well, that's going to take some thought to get it right, but it looks like you're just matching current behavior here?

brendankenny:
I'd really like to clear up the audit/gatherer id/name/classname/key-in-various-objects distinctions a bit more.

- What is the relationship between the keys in the `gatherers` top level object and the `gatherers` object in each pass? Do the pass gatherer keys reference the keys in the top level object?
- Same thing for the `audits` top-level object keys vs the `id` values in each category's children array (I believe I asked this in a comment)
- Is there a way to use a gatherer twice with different input? This is hard since we use `artifacts[classname]` currently to identify the output of each gatherer, so we may want to table this for now.
- Is there a way to use an audit twice? Since we're letting audits take input with this new version this is going to be a great thing. Ideally I could do something like
```js
"time-to-interactive-cold": {
  "path": "time-to-interactive",
  "inputTrace": "firstPass"
},
"time-to-interactive-warm": {
  "path": "time-to-interactive",
  "inputTrace": "laterPass"
}
```
(modulo better naming/clearer handling of path or whatever, etc)
and then being able to refer to `time-to-interactive-cold` and `time-to-interactive-warm` as different things in my categories. This would require using the keys as ids though, and not something on the audit itself.

I haven't fully thought that through yet, so there may be issues with that :)

So, basically, it would be sweet to simplify the id/key/path thing a bit more and figure out for sure how we're going to allow single audits multiple times (bonus points for gatherers multiple times, though that seems a lot messier).

brendankenny:
kind of shitty idea, everyone feel free to disagree: to iterate on some of this a bit more but not block the new report, we could make even more of an MVP by making the config v2 exactly the same as v1 except have `report.categories` instead of `aggregations`. I think everyone is already in agreement on the `report.categories` stuff here.

`config.js` probably wouldn't need to change much from master and the report generation/scoring could stay as it is in this PR. It's mostly v2/config.js that wouldn't be needed yet.

Just to be clear, this is all really good and would just get rolled over to the next change, I'm just thinking about this because I don't want to slow us way down by making us talk about all these things.

...? :)

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,186 @@
+const path = require('path');
+const log = require('../../lib/log');
+
+const defaultConfigPath = './default.json';
+const defaultConfigJson = require(defaultConfigPath);
+
+const _flatten = arr => [].concat(...arr);
+const _subtract = (setA, setB) => Array.from(setA).filter(x => !setB.has(x));
```
patrickhulce: sure, and it did at first but it never gets used as a set, just as an array, I'll rename to `_differenceAsArray`

lighthouse-core/config/v2/config.js
patrickhulce: > it looks like you're just matching current behavior here?

The intent behind several of the config changes was specifically to enable clear and simple extension not needing logic or fancy behavior of any sort, so this is the final behavior as far as I'm concerned. That's what the structure of the new config enables :)

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,308 @@
+{
+  "version": 2,
```
patrickhulce: Agreed we should have verification of the config format but we explicitly are attempting to support two config versions at once to start as we iterate, so I think the version has value as a clue of which path to head toward. A try until it works approach with validation and error messages greatly complicates the MVP. How about `configVersion` until it becomes the default and we nuke the field entirely and rely on validation only?

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,308 @@
...
+    "content-width": {},
+    "service-worker": {},
+    "offline": {},
+    "http-redirect": {},
+    "html-without-javascript": {},
+    "chrome-console-messages": {},
+    "styles": {},
+    "css-usage": {},
+    "all-event-listeners": {
+      "path": "dobetterweb/all-event-listeners"
```
patrickhulce: Not sure what you're trying to say here. IMO, this is actually a step in the direction of decoupling our arbitrary repo folder structure from magic happening in the config. What leads you to believe our repo organization process is now inherent to the gatherer now?

lighthouse-core/config/v2/default.json
patrickhulce: sure

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,308 @@
...
+    "manifest-icons-min-192": {},
+    "manifest-icons-min-144": {},
+    "manifest-name": {},
+    "manifest-short-name": {},
+    "manifest-short-name-length": {},
+    "manifest-start-url": {},
+    "theme-color-meta": {},
+    "content-width": {},
+    "deprecations": {},
+    "aria-allowed-attr": {
```
patrickhulce: For audits, as with each of these toplevel objects, the keys are the `id` (currently `audit.meta.name`) of the audit (rather than have it redundantly specified inside the object as well). There's really no need for audits to have category or name in their implementation now either.

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,308 @@
...
+      "path": "dobetterweb/uses-passive-event-listeners"
+    }
+  },
+
+  "report": {
+    "categories": {
+      "pwa": {
+        "name": "Progressive Web App",
+        "description": "",
+        "children": [
```
patrickhulce: `children` was used forseeing the whole `subcategory` discussion occurring and the likely scenario that we will need to add categories as children of other categories. Keeping it generic with children, IMO, leaves open the possibility without a breaking change. In the proposal, each child has a `type: "audit"|"category"|"some-newfangled-thing"` but since all we use are audits today it seemed unnecessary.

lighthouse-core/gather/gather-runner.js
```diff
@@ -431,6 +431,11 @@ class GatherRunner {
   static instantiateGatherers(passes, rootPath) {
     return passes.map(pass => {
       pass.gatherers = pass.gatherers.map(gatherer => {
+        if (gatherer.implementation) {
```
patrickhulce: sure

lighthouse-core/report/v2/report-generator.js
patrickhulce: Yeah I was waiting to introduce scoring methods and since nothing scoring wise has changed despite lots of talk the past few months I thought I'd poke the bear and see what came of it :P

lighthouse-core/runner.js
patrickhulce: done

patrickhulce:
> audit/gatherer id/name/classname/key-in-various-objects

I seem to have failed in config clarity given that I think the biggest win of config v2 is getting rid all of most of this nonsense and moving everything to IDs (apart from artifact name being class name...still need to hash that one out). The keys for *all toplevel objects* (gatherers, audits, passes) are IDs of that object. This ID should how everything else in the config refers to that particular object. Examples below.

> What is the relationship between the keys in the gatherers top level object and the gatherers object in each pass? Do the pass gatherer keys reference the keys in the top level object?

The strings specified in a passes `gatherers` property are the IDs of gatherers. The long-term view was that these wouldn't need to be strings but potentially objects specifying a gatherer ID and potential options (for example collecting similar data across passes, or with slightly different settings).

> Same thing for the audits top-level object keys vs the id values in each category's children array (I believe I asked this in a comment)

Responded in comment too, but same story. These strings are the IDs of the audits.

> Is there a way to use a gatherer twice with different input? This is hard since we use artifacts[classname] currently to identify the output of each gatherer, so we may want to table this for now.

Correcto, as mentioned above still need to hash this out, but my big controversial proposal will be to start storing artifacts by gatherer ID, ignore class name completely, and move requiredArtifacts to an input property of an audit (with a defaults suggested by impl so config doesn't balloon with unnecessary wiring).....buuuuuuut one major change at a time :)

> Is there a way to use an audit twice? Since we're letting audits take input with this new version this is going to be a great thing. Ideally I could do something like

Bingo :)

> This would require using the keys as ids though, and not something on the audit itself.

yes! this is exactly what I'm trying to move to.

patrickhulce:
> making the config v2 exactly the same as v1 except have report.categories instead of aggregations

😢 😢 

Seriously though, IMO there's no need to have these set in stone until we actually flip the switch and kill config v1 making v2 the default. If we have buyer's remorse about passes as object/whatever, I don't see why we couldn't change it all around tomorrow (though I obviously hope major points could be ironed out in doc stage 😄 )

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,304 @@
...
+        "domstats",
+        "optimized-images",
+        "tags-blocking-first-paint",
+        "websql"
+      ]
+    }
+  },
+
+  "audits": {
+    "is-on-https": {},
```
ebidel: shall we alphabetize all of these listings?

lighthouse-core/config/v2/default.json
ebidel: I'm in favor of keeping "audits" for now. I read this block as "this category contains/runs these audits".  That naming is consistent with the list of "audits" above and is probably easier for contributors to make the linkage when adding new audits.

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,304 @@
...
+    },
+    "uses-http2": {
+      "path": "dobetterweb/uses-http2"
+    },
+    "uses-passive-event-listeners": {
+      "path": "dobetterweb/uses-passive-event-listeners"
+    }
+  },
+
+  "report": {
```
ebidel: I still thinks it's worth exploring modularity but am alright with adding a `report` section. 

IOW,`report` could be it's own file (`config/report.json`). In that way, we keep configuration separate from output generation. As as an example, I could see users wanting to use our settings.json with their own report.json to create a custom report. They wouldn't have to change a mega file. That said, maybe your extension model smooths over that nasty....?

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,304 @@
...
+          {"id": "manifest-icons-min-192", "weight": 1},
+          {"id": "manifest-exists", "weight": 1},
+          {"id": "theme-color-meta", "weight": 1},
+          {"id": "viewport", "weight": 1},
+          {"id": "content-width", "weight": 1}
+        ]
+      },
+      "performance": {
+        "name": "Performance",
+        "description": "",
```
ebidel: Might as well port over the descriptions from v1 while we're here.

lighthouse-core/report/v2/report-generator.js
ebidel: Anything related to scoring does feel like a future PR. There's enough to bit off in here with all the other changes :)

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,186 @@
...
+        path.resolve(configPath, rawPath), // for relative config usage
+        path.resolve(process.cwd(), rawPath), // for node module usage
+      ]);
+      const resolvedPath = ConfigV2._tryResolveUntilSuccess(possiblePaths);
+      object[key] = Object.assign({}, object[key], {path: resolvedPath});
+    });
+    return object;
+  }
+
+  static objectToList(object) {
```
patrickhulce: mainly for testing since it did do fancier things like ordering, but that's been nixed for a future PR and didn't seem worth moving

lighthouse-core/config/v2/config.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    const extendedConfig = new ConfigV2(extendedConfigJson, extendedConfigPath);
+    return ConfigV2._mergeObjects(extendedConfig.asJson(), configJson);
+  }
+
+  static resolvePaths(object, configPath, searchPaths = []) {
+    object = Object.assign({}, object);
+    Object.keys(object).forEach(key => {
+      // We don't need to resolve objects that already have
```
patrickhulce: fixed

lighthouse-core/config/v2/default.json
```diff
@@ -0,0 +1,304 @@
+{
+  "configVersion": 2,
+
+  "gatherers": {
```
patrickhulce: yep :) lists all gatherers with their ids and paths (if different from id) plus allowing options and multiple version of the same implementation later on down the road

lighthouse-core/config/v2/default.json
patrickhulce: I'm definitely not opposed to separating the presentation config from the runtime config at some point, easy extension eases the immediate pain IMO.

lighthouse-core/config/v2/default.json
patrickhulce: done

PR #1881, opened 2017-03-20
test(smokehouse): fix flakiness of smokehouse dom-size expectation

patrickhulce:
adds RegExp smokehouse to boot :)

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -111,9 +111,13 @@ function findDifference(path, actual, expected) {
     return null;
   }
 
+  if (typeof actual === 'string' && expected instanceof RegExp && expected.test(actual)) {
+    return null;
+  }
+
   // If they aren't both an object we can't recurse further, so this is the difference.
   if (actual === null || expected === null || typeof actual !== 'object' ||
-      typeof expected !== 'object') {
+      typeof expected !== 'object' || expected instanceof RegExp) {
     return {
       path,
       actual,
--- a/lighthouse-cli/test/smokehouse/pwa-expectations.js
+++ b/lighthouse-cli/test/smokehouse/pwa-expectations.js
@@ -83,7 +83,7 @@ module.exports = [
         extendedInfo: {
           value: {
             1: {value: '20'},
-            2: {snippet: 'Element with most children:\nul#versionlist.'}
+            2: {snippet: /ul.versionlist/}
           }
         }
       },


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -111,6 +111,10 @@ function findDifference(path, actual, expected) {
     return null;
   }
 
+  if (expected instanceof RegExp && expected.test(actual)) {
```
brendankenny: `expected.test()` will do `actual.toString()` internally, so this check should be good for any value of `actual`, but if it fails this check, `typeof expected` will still be `'object'`, so it could pass the conditional below and then try to recurse on the regexp object instead of returning the difference. This could lead to a false positive result since `Object.keys(/ul.versionlist/)` is `[]`, so any value of `actual` that has `typeof` `'object'` will pass. e.g. `findDifference('path', {}, /whatever/)` will find no difference.

Simplest thing would be to also add `|| expected instanceof RegExp` to the conditional below.

brendankenny: Although there is also the minuscule chance that `actual.toString()` could yield a string that somehow passes `expected.test()`, so you may also want to add a `typeof actual === 'string'` above and below (or rearrange somehow) if that's not a feature we want to support :)

brendankenny: surprised I don't see a `RegExp.isRegExp()` proposed anywhere

brendankenny:
how was this flaking, incidentally? What string was being generated instead?

brendankenny:
> how was this flaking, incidentally? What string was being generated instead?

from a recent run:

> difference at dom-size.extendedInfo.value[2].snippet: 
found 'Element with most children: ul#versionlist.canaryisdev`',
expected 'Element with most children: ul#versionlist'.

patrickhulce:
Never did I expect so many comments for this... haha

lighthouse-cli/test/smokehouse/smokehouse.js
patrickhulce: Really? :P

brendankenny: > Really? :P

well `expected instanceof RegExp` won't work if through some ridiculously circuitous process you're writing expectations values with a `RegExp` object from another realm, so the "real" way to check is like the old `isArray` check of `Object.prototype.toString.call(expected) === '[object RegExp]'`.

I would have thought someone, somewhere would have proposed `RegExp.isRegExp()` like `Array.isArray()` to TC39 at some point, but I couldn't find any record of such a thing :)

PR #1883, opened 2017-03-21
remove unused variables

denar90:
Should `AuditResult` be removed from types as well?

### DIFF
--- a/lighthouse-cli/ask.ts
+++ b/lighthouse-cli/ask.ts
@@ -20,7 +20,7 @@
 const readline = require('readline');
 
 function ask(question: string, options: string[]): Promise<string> {
-  return new Promise((resolve, reject) => {
+  return new Promise(resolve => {
     const iface = readline.createInterface(process.stdin, process.stdout);
     const optionsStr = options.map((o, i) => i + 1 + '. ' + o).join('\r\n');
 
--- a/lighthouse-cli/chrome-launcher.ts
+++ b/lighthouse-cli/chrome-launcher.ts
@@ -142,7 +142,7 @@ export class ChromeLauncher {
   }
 
   spawn(execPath: string): Promise<any[]> {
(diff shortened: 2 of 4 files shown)


ebidel:
Nice. Thanks for the cleanup.

Might be worth adding these to our tsconfig.json so the compile catches it in the future:

```
"noUnusedLocals": true,
"noUnusedParameters": true
```

Looks like there are a couple of other places that get flagged.

lighthouse-cli/ask.ts
```diff
@@ -20,7 +20,7 @@
 const readline = require('readline');
 
 function ask(question: string, options: string[]): Promise<string> {
-  return new Promise((resolve, reject) => {
+  return new Promise((resolve) => {
```
ebidel: to be consistent with the rest of the codebase, can you remove the brackets for the single param?

lighthouse-cli/chrome-launcher.ts
```diff
@@ -142,7 +142,7 @@ export class ChromeLauncher {
   }
 
   spawn(execPath: string): Promise<any[]> {
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
```
ebidel: remove `()`

lighthouse-cli/printer.ts
```diff
@@ -68,7 +68,7 @@ function createOutput(results: Results, outputMode: OutputMode): string {
  * Writes the output to stdout.
  */
 function writeToStdout(output: string): Promise<{}> {
-  return new Promise((resolve, reject) => {
+  return new Promise(resolve => {
```
ebidel: :)

lighthouse-cli/ask.ts
denar90: oh, sorry, missed these ...

PR #1889, opened 2017-03-22
CLI: remove npm prepublish

### LINKED ISSUE #1516: npm prepublish deprecation

ebidel:
Fixes #1516

R: all

Finally getting rid of this. It's deprecated in npm 5 and we have convenient build scripts now.

### DIFF
--- a/lighthouse-cli/compiled-check.js
+++ b/lighthouse-cli/compiled-check.js
@@ -6,7 +6,7 @@ const path = require('path');
 module.exports = function(filename) {
   if (!fs.existsSync(path.join(__dirname, filename))) {
     console.log('Oops! Looks like the CLI needs to be compiled. Please run:');
-    console.log('   npm run build-cli');
+    console.log('   npm run install-cli');
     console.log('More at: https://github.com/GoogleChrome/lighthouse#develop');
     process.exit(1);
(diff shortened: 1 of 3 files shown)


lighthouse-cli/compiled-check.js
```diff
@@ -6,7 +6,7 @@ const path = require('path');
 module.exports = function(filename) {
   if (!fs.existsSync(path.join(__dirname, filename))) {
     console.log('Oops! Looks like the CLI needs to be compiled. Please run:');
-    console.log('   npm run install-cli');
+    console.log('   npm run build-cli');
```
brendankenny: if `install-cli` has `build-cli` in it now, should this stay `install-cli` then?

ebidel: Suppose it could. Either works.

ebidel: done

PR #1892, opened 2017-03-23
handle driver.getObjectProperty results with no value property

### LINKED ISSUE #1891: lighthouse hangs on specific websites
Noticed the issue with 1.6.0 and 1.5.2, running lighthouse against some specific websites results in errors (lighthouse hangs until SIGINT). Specifically observed on thomson.co.uk with the following error:

status Retrieving: AnchorsWithNoRelNoopener +1s
(node:90614) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'value' of undefined
(node:90614) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

to reproduce:
lighthouse https://www.thomson.co.uk

brendankenny:
fixes #1891

`Runtime.getProperties` returns properties that, even if they're defined [might not have a value](https://chromedevtools.github.io/debugger-protocol-viewer/tot/Runtime/#type-PropertyDescriptor). This just returns null in that case. As mentioned in #1891, I'm not really sure why no value is being returned in this particular case, since no site I've seen has demonstrated this problem before. We may want to hold off on merging until we figure out the underlying cause so e.g. we could put a link like that into the smoke tests.

Also puts any thrown errors (or rejected promises) from the promise chain starting with `this.sendCommand` back into the main promise chain.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -496,7 +496,7 @@ class Driver {
   * @return {!Promise<string>} The property value, or null, if property not found
   */
   getObjectProperty(objectId, propName) {
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       this.sendCommand('Runtime.getProperties', {
         objectId,
         accessorPropertiesOnly: true,
@@ -507,12 +507,12 @@ class Driver {
         const propertyForName = properties.result
           .find(property => property.name === propName);
 
-        if (propertyForName) {
+        if (propertyForName && propertyForName.value) {
           resolve(propertyForName.value.value);
         } else {
           resolve(null);
         }
-      });
+      }).catch(reject);
     });
   }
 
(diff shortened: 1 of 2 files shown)


wardpeet:
I think I can reference this one https://github.com/GoogleChrome/lighthouse/pull/1818

This was an issue with getProperties also. My hunch was garbage collection or chrome letting go of the reference.

paulirish:
Here's how devtools handles these objects: https://github.com/ChromeDevTools/devtools-frontend/blob/56e9d0b2f6f83f4c0eb104b6f01cd14a706e4a61/front_end/sdk/RemoteObject.js#L667-L683

It handles a case where the remote object is expressed with a symbol instead, but since we don't have to handle every object type under the sun, I think we can safely ignore that.

PR #1896, opened 2017-03-23
feat(report): add v2 with categories

patrickhulce:
R: @brendankenny, all

ported from #1880 but...
- removes all config.js code and changes
- removes geometric mean and overall scoring
- toggles functionality via output mode (see below)

since we shot down a `--v2` flag this adds an additional output mode instead, I'm not a fan since it seems less generic and more opaque, so open to other suggestions of how to do parallel operation modes.


lighthouse-core/report/report-generator.js
```diff
@@ -149,13 +149,44 @@ class ReportGenerator {
...
+  /**
    * Generates the Lighthouse report HTML.
    * @param {!Object} results Lighthouse results.
    * @param {!string} reportContext What app is requesting the report (eg. devtools, extension)
    * @param {?Object} reportsCatalog Basic info about all the reports to include in left nav bar
+   * @param {boolean=} forceV2
    * @return {string} HTML of the report page.
    */
-  generateHTML(results, reportContext = 'extension', reportsCatalog) {
+  generateHTML(results, reportContext = 'extension', reportsCatalog, forceV2) {
```
paulirish: can we make a separate method like `generateDOMReport`? this will duplicate a bunch of this method but will make it easier to drop the old when we are ready to switch.
will kill boolean trap, too. yay :)

(also i kinda prefer using "dom" as the key word here. v2 can get overloaded and not sure which v2 talking about: config or report UI or report generator etc etc)

lighthouse-cli/printer.ts
```diff
@@ -22,8 +22,8 @@
  *   'json': JSON formatted results
  *   'html': An HTML report
  */
-enum OutputMode { json, html };
-type Mode = 'json' | 'html';
+enum OutputMode { json, html, htmlv2 };
+type Mode = 'json' | 'html' | 'htmlv2';
```
paulirish: htmlv2 -> domhtml

lighthouse-core/report/report-generator.js
brendankenny: +1 to duplicate in a new method. Easy to edit and not worry about existing report still working.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,67 @@
...
+    return (results.sum / results.weight) || 0;
+  }
+
+  /**
+   * Returns the report JSON object with computed scores.
+   * @param {{categories: !Object<{audits: !Array}>}} config
+   * @param {!Object<{score: number|boolean}>} resultsByAuditId
+   * @return {{categories: !Array<{audits: !Array<{score: number, result: !Object}>}>}}
+   */
+  generateReportJson(config, resultsByAuditId) {
```
paulirish: based on earlier discussion i think we concluded that lighthouse result object should be post-scoring. 
so its got scoring and categories and it's all done.

so maybe there's an `applyScoring` thing called in runner that's called before it returns everything

but i like this because people consuming LH over npm should get scores. 
that sound good?

patrickhulce: yeah config v2 version of `report-generator.js` had it returning score too but got removed for controversy over weights and geometric mean. i'll add it back for now

lighthouse-cli/printer.ts
patrickhulce: will do

lighthouse-core/report/report-generator.js
patrickhulce: did something kinda similar, but not completely...

patrickhulce:
@brendankenny @paulirish updated since comments with a slightly different approach (skipping old report entirely and just outputting the empty dom html file)

lighthouse-core/report/v2/report-template.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
brendankenny: need license

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,84 @@
...
+    const score = ReportGeneratorV2.arithmeticMean(categories);
+    return {score, categories};
+  }
+
+  /**
+   * @param {!Object} reportAsJson
+   */
+  generateReportHtml(reportAsJson) {
+    return REPORT_TEMPLATE
+      .replace(/%%LIGHTHOUSE_JSON%%/, JSON.stringify(reportAsJson))
```
paulirish: looks like encodeURIComponent would do the trick too.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,84 @@
...
+const path = require('path');
+
+const REPORT_TEMPLATE = fs.readFileSync(path.join(__dirname, './report-template.html'), 'utf8');
+// TODO: Setup a gulp pipeline to concat and minify the renderer files?
+const REPORT_JAVASCRIPT = fs.readFileSync(path.join(__dirname, './report-renderer.js'), 'utf8');
+
+class ReportGeneratorV2 {
+  /**
+   * Computes the weighted-average of the score of the list of items.
+   * @param {!Array<{score: number=, weight: number=}} items
```
brendankenny: have to do the unfortunate `(number|undefined)` here. `=` is just for optional params.

Do you really want optional weights or scores, though? `generateAuditResult` at least guarantees a score, and we use that for all our audits

brendankenny: also, do you want to handle audit errors explicitly here? In that case `score` will be null, so would be `Number(null) -> 0`, but may be worth handling explicitly

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Returns the report JSON object with computed scores.
+   * @param {{categories: !Object<{audits: !Array}>}} config
+   * @param {!Object<{score: number|boolean}>} resultsByAuditId
+   * @return {{categories: !Array<{audits: !Array<{score: number, result: !Object}>}>}}
+   */
+  generateReportJson(config, resultsByAuditId) {
+    const categories = Object.keys(config.categories).map(categoryId => {
+      const category = config.categories[categoryId];
+      category.id = categoryId;
```
brendankenny: any reason for the `.id`? (appears to only be used in the tests, and there only for testing that the assignment was correct)

lighthouse-core/report/v2/report-template.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
+  <title>Lighthouse Report</title>
+</head>
+<body>
+  <noscript>Lighthouse report requires JavaScript. Please enable.</noscript>
+  <script>window.__LIGHTHOUSE_JSON__ = %%LIGHTHOUSE_JSON%%</script>
```
patrickhulce: yup will do

lighthouse-core/report/v2/report-generator.js
patrickhulce: > (number|undefined)

will do

> Do you really want optional weights or scores, though?

Yeah, needing to explicitly set `weight: 0` on every unscored item seemed unnecessary and always having score doesn't necessarily make sense for explicitly non-scored audits so handling the unspecified case seems reasonable.

> do you want to handle audit errors explicitly here?

What do you envision explicit handling would do differently?

lighthouse-core/report/v2/report-generator.js
patrickhulce: it was used by the now defunct `getAggregations` and was the consistent format across config v2, but can remove if you think we won't use it again

lighthouse-core/report/v2/report-generator.js
patrickhulce: I generally prefer just unicode escaping `<` any objections?

patrickhulce: surprised they hide that option so much in the link actually :)

lighthouse-core/report/v2/report-template.html
patrickhulce: will do

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -0,0 +1,39 @@
...
+   * @param {!Element} element
+   */
+  render(element) {
+    const pre = document.createElement('pre');
+    pre.innerText = JSON.stringify(this._report, null, 2);
+    element.appendChild(pre);
+  }
+}
+
+const renderer = new ReportRenderer(window.__LIGHTHOUSE_JSON__);
```
ebidel: Not that we need it right now, but can we keep usage out of this file and initiate in the .html? Thinking about the the Viewer extension use case:

https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/templates/report-template.html#L336

lighthouse-core/report/v2/report-template.html
```diff
@@ -0,0 +1,13 @@
...
+  <script>%%LIGHTHOUSE_JAVASCRIPT%%</script>
```
ebidel: does devtools have any CSP restrictions? are we able to have inline `<script>`?

lighthouse-core/report/v2/report-generator.js
brendankenny: > I generally prefer just unicode escaping `<` any objections?

like `&#60;` or `&#x3c;`? My understanding is that `&lt;` is just shorthand for that, so seems fine, though the entity names have the advantage of being more human readable-ish.

(from a quick look it seems that source is a little out of date, too, or a little too conservative. Both `&apos;` (instead of `&#x27;'`) and `&sol;` (instead of `&#x2F;`) are [in the html spec](https://html.spec.whatwg.org/multipage/syntax.html#named-character-references) now)

lighthouse-core/report/v2/report-generator.js
brendankenny: > Yeah, needing to explicitly set weight: 0 on every unscored item seemed unnecessary and always having score doesn't necessarily make sense for explicitly non-scored audits so handling the unspecified case seems reasonable.

ah yes,  that's much better

> What do you envision explicit handling would do differently?

either a comment to make it clear that the code is purposefully handling that case (`// error audit result will have a null score which will be treated as 0` or whatever) or an explicit check of `item.error` and setting the score based on the result. Either way would make it easier for future code readers to figure out the error flow or for us to change error scoring to `NaN` or whatever.

patrickhulce: oh gotcha no change just explanation 👍  yeah for sure will do :)

lighthouse-core/report/v2/report-renderer.js
patrickhulce: will do

lighthouse-core/report/v2/report-template.html
patrickhulce: devtools probably won't use this at all actually, it should just include the renderer source files directly, this is pretty much exclusively for saving the report to an html file

lighthouse-core/report/v2/report-generator.js
patrickhulce: No I mean `\u003c`, the only place `</script>` can appear in valid JSON would be in strings which are replaced with JS escapes not HTML escapes

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -0,0 +1,36 @@
...
+  constructor(report) {
+    this._report = report;
+  }
+
+  /**
+   * @param {!Element} element
+   */
+  render(element) {
+    const pre = document.createElement('pre');
+    pre.innerText = JSON.stringify(this._report, null, 2);
```
ebidel: `.textContent`

patrickhulce:
PTAL, FYI appveyor is failing because of the issue identified in #1911

lighthouse-cli/bin.ts
```diff
@@ -293,9 +293,11 @@ function saveResults(results: Results,
           return innerPromise.then((_: Results) => Printer.write(results, outputType, outputPath));
         }, Promise.resolve(results));
       } else {
-        const outputPath = flags.outputPath || `${resolvedPath}.report.${flags.output}`;
+        let outputPath = flags.outputPath || `${resolvedPath}.report.${flags.output}`;
+        outputPath = outputPath.replace(/\.domhtml$/, '.html');
```
brendankenny: this is kind of a stretch, but since we support multiple output modes and you added support for letting lighthouse output both the html and the dom reports below, would it make sense to do `outputPath = outputPath.replace(/\.domhtml$/, '.dom.html');` or something so that they aren't written to the same file name?

To be fair, the only use case I can see for that is for development of the new report, but it might be helpful to be able to output both while we work on it

lighthouse-core/report/v2/report-generator.js
brendankenny: > but can remove if you think we won't use it again

I'd personally say remove and we can add back when we need them

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,87 @@
...
+
+      const audits = category.audits.map(audit => {
+        const result = resultsByAuditId[audit.id];
+        // Cast to number to catch `null` and undefined when audits error/do not run
```
brendankenny: `do not run` definitely seems like a case we should catch in config validation

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Returns the report HTML as a string with the report JSON and renderer JS inlined.
+   * @param {!Object} reportAsJson
+   * @return {string}
+   */
+  generateReportHtml(reportAsJson) {
+    return REPORT_TEMPLATE
+      .replace(/%%LIGHTHOUSE_JSON%%/, JSON.stringify(reportAsJson).replace(/</g, '\\u003c'))
```
brendankenny: this helps display but it doesn't prevent XSS. Escaping JS correctly is a pain, hence the html entity escaping and putting it in a div suggestion

lighthouse-core/report/v2/report-generator.js
```diff
@@ -0,0 +1,87 @@
...
+      .replace(/%%LIGHTHOUSE_JAVASCRIPT%%/, REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/'));
```
brendankenny: we control this, so no XSS concerns there

lighthouse-core/report/v2/report-generator.js
brendankenny: > this helps display but it doesn't prevent XSS

actually, I see what you mean now. If it's valid JSON you aren't breaking out of the object, so you really only have to worry about closing the script tag at that point. SGTM, though maybe save to a variable to make the sanitize step clear?

lighthouse-core/report/v2/report-generator.js
patrickhulce: not for XSS, same motivation as for why we currently escape `</script>` tags just not to strangely break

lighthouse-core/report/v2/report-generator.js
patrickhulce: will extract to variable 👍

lighthouse-core/report/v2/report-generator.js
patrickhulce: we do, but not when the required artifacts error and the audit isn't run :P

lighthouse-cli/bin.ts
patrickhulce: yeah sure

patrickhulce:
alright gatekeeper @brendankenny, do your worst

paulirish:
OH YEAHHHHHHHHH. 
Hello Future, here we come. 👓

PR #1906, opened 2017-03-24
Add doc for testing site with authentication

denar90:
Thx to @brendankenny and @patrickhulce

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -116,6 +116,12 @@ NOTE: specifying an output path with multiple formats ignores your specified ext
 * `./<HOST>_<DATE>.report.html`
 * `./<HOST>_<DATE>.artifacts.log`
 
+### Testing on a site with authentication
+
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -116,6 +116,20 @@ NOTE: specifying an output path with multiple formats ignores your specified ext
...
 
+### Testing on a site with authentication
+
+If you have installed Lighthouse globally 
+
+ - `chrome-debug`
+ - open and login to your site
+ - `lighthouse http://mysite.com`
+ 
+If you use repository
```
paulirish: lets get rid of this second part. 
also you can just `npm link .` to install the repo globally. :)

denar90: 👍
I was thinking about users who doesn't know about it, but you are right, if person using repo he/she should know about this kind of stuff.
nuking... 💥

PR #1911, opened 2017-03-27
fix: default(.json -> .js). Disable css usage audit

patrickhulce:
Breaking changes have landed in protocol (currently failing all AppVeyor builds) and we're punting on fixing the accuracy until after IO anyway, just disable for now.


paulirish:
this now becomes a disabled audit, though we're keeping it in the source tree for good reasons.

I wonder if theres somewhere else we can clearly indicate it's disabled, as unfortunately commenting out in the default.json isn't a possibility.

how about also changing the audit.meta.name from `"unused-css-rules"` to `"DISABLED__unused-css-rules"` or something?

brendankenny:
> as unfortunately commenting out in the default.json isn't a possibility

Time to `git mv default.json default.js`? :)

paulirish:
> Time to git mv default.json default.js? :)

YUP. patrick wants to, too. let's do it.

patrickhulce:
BOOM! 💥 
> paulirish added the RlsToStable label 16 minutes ago


wait, @paulirish changing default.js is not compatible with that plan...

patrickhulce:
i'll cut a separate PR for 1.6.0 branch?

patrickhulce:
PTAL for added json -> js :)

patrickhulce:
oh, oops, I literally just merged :) makes it easy to find files requires with grep/simple search though so i'm actually not as jazzed about extension-less requires

brendankenny:
fair (though I just saved a regex for that :). Main benefit is that it makes it transparent where the data is coming from (json or from executed code), but since this is the hardcoded default config it doesn't matter

patrickhulce:
> Main benefit is that it makes it transparent where the data is coming from

You mean opaque? ;) haha, that's exactly why I don't like it!

PR #1915, opened 2017-03-27
fix(url-shim): do not elide short names with preserveHost

patrickhulce:
R: all

### DIFF
--- a/lighthouse-cli/test/cli/printer-test.js
+++ b/lighthouse-cli/test/cli/printer-test.js
@@ -71,7 +71,7 @@ describe('Printer', () => {
   it('writes extended info', () => {
     const mode = Printer.OutputMode.html;
     const htmlOutput = Printer.createOutput(sampleResults, mode);
-    const outputCheck = new RegExp('\u2026dobetterweb/dbw_tester.css', 'i');
+    const outputCheck = new RegExp('dobetterweb/dbw_tester.css', 'i');
 
     assert.ok(outputCheck.test(htmlOutput));
   });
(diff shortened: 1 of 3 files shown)


patrickhulce:
> from …img/logo.jpg to /img/logo.jpg. Is that intentional?

Yep, I'll add that test to make it explicit.

> Also the double // in something like http://example.com//logo.jpg would no longer be preserved, though I don't know why someone would want it to be

Haha, fair :) I'll switch to another method

patrickhulce:
PTAL :)

brendankenny:
> I'll switch to another method

looks like you didn't? Does it matter?

brendankenny:
oh, looks like my example was wrong.

`URL.getDisplayName('http://example.com//logo.jpg', {preserveHost: true})` gives `example.com//logo.jpg`, but

`URL.getDisplayName('http://example.com/img/img//logo.jpg', {preserveHost: true})` drops the double // and gives `example.com/…img/logo.jpg`

patrickhulce:
Oh, yeah, realized it doesn't matter for the upfront case you mentioned
(captured by test), only in the middle of a URL, so stopped changing :) but
can continue if we think it'll come up

patrickhulce:
I'm sticking by it too FYI because 
A) it should be rare
B) the way we currently express path parts and ellipsis isn't great to begin with. `/a/b//foo.js` after today becomes `...b/foo.js` but would become just `.../foo.js` after a fix (since the second path part is empty)

bump :)

brendankenny:
> the way we currently express path parts and ellipsis isn't great to begin with

yeah, it's just a display URL, so, SGTM. But we're assigning you the bug when someone inevitably files an issue because `///` in their URLs is essential to their workflow :P

patrickhulce:
> But we're assigning you the bug when someone inevitably files an issue because /// in their URLs is essential to their workflow :P

hahaha deal!

PR #1918, opened 2017-03-27
fix: simpler https audit

### LINKED ISSUE #1175: Skip HTTPS tests when running a 127.0.0.1 or localhost or .local URL

### LINKED ISSUE #1904: "Uses HTTPS" result is incorrect on WPT
LH report:  https://www.webpagetest.org/lighthouse.php?test=170324_KB_5MBY&run=1
WPT details: https://www.webpagetest.org/result/170324_KB_5MBY/

On this report you can see "uses https" fails but the test URL is "https://...". It was captured on chrome stable on android. 

We've seen some issues with the Security domain in Headless and it's possible we're encountering similar problems here. 

Does someone have an android they can debug this with?

patrickhulce:
fixes #1904 

Moves the HTTPS audit to a simpler method just ensuring no requests were loaded over `http` more debuggable and informative to the user as well which URLs need to be https. Also considers localhost to be secure to fix #1175


lighthouse-core/audits/is-on-https.js
```diff
@@ -41,9 +43,25 @@ class HTTPS extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    const networkRecords = artifacts.networkRecords[Audit.DEFAULT_PASS];
+    const insecureRecords = networkRecords
+        .filter(record => record.scheme === 'http' && record.domain !== 'localhost')
```
paulirish: i'd prefer to whitelist what we consider secure, rather than building up a blacklist.  

check out https://www.chromium.org/Home/chromium-security/security-faq#TOC-Which-origins-are-secure-
i think we only want to do https/wss and localhost for now.

lighthouse-core/test/runner-test.js
```diff
@@ -46,7 +46,7 @@ describe('Runner', () => {
     const url = 'https://example.com';
     const config = new Config({
       audits: [
-        'is-on-https'
+        'theme-color-meta'
```
paulirish: im just about to nuke this guy. ;)  can you use another

brendankenny:
it is nice to match Chrome's/devtools' notions of security, though, and not have to update to match when it changes (e.g. `neutral` today for mixed content images may someday be more strict).

Would it make sense to still do `driver.getSecurityState`, keep the whole security state object, not just `state.schemeIsCryptographic` so we can get the reasoning, and then combine it with record counts in the audit (like it appears the Security panel does too now)?

We can definitely ditch all the timeout nonsense. That's from ancient times, and if we actually needed to worry about every driver call timing out, Lighthouse as a whole would be in a ton of trouble :)

paulirish:
> Would it make sense to still do driver.getSecurityState,

basically the `Security` domain isn't implemented on Android, and hasn't been for over a year. So, while I don't really _want_ to, i think we have to drop that approach and determine the basic security state on our own.

lighthouse-core/test/runner-test.js
patrickhulce: 😢

lighthouse-core/audits/is-on-https.js
patrickhulce: done

lighthouse-core/test/runner-test.js
```diff
@@ -392,18 +390,16 @@ describe('Runner', () => {
...
       audits: [
-        'is-on-https'
+        'content-width'
       ],
 
       artifacts: {
-        HTTPS: {
-          value: true
-        }
+        ViewportDimensions: '#ffffff'
```
paulirish: white dimensions? that's kinda fun.

lighthouse-core/audits/is-on-https.js
```diff
@@ -32,18 +37,42 @@ class HTTPS extends Audit {
...
   static audit(artifacts) {
+    const networkRecords = artifacts.networkRecords[Audit.DEFAULT_PASS];
+    const insecureRecords = networkRecords
+        .filter(record => !HTTPS.isSecureRecord(record))
+        .map(record => ({url: URL.getDisplayName(record.url, {preserveHost: true})}));
+
+    let displayValue = '';
+    if (insecureRecords.length > 1) {
+      displayValue = `${insecureRecords.length} insecure requests found`;
+    } else if (insecureRecords.length === 1) {
```
paulirish: looks like we're setting the same displayValue in both cases. 
Regardless, I'm +1 on ignoring the pluralization issue and keeping the impl simple.

lighthouse-core/test/runner-test.js
patrickhulce: heh, this turns out this value is meaningless to tests then :)

lighthouse-core/audits/is-on-https.js
ebidel: What about one string to rule them all:

```
if (insecureRecords.length) {
  displayValue = `Insecure requests: ${insecureRecords.length}`;
}
```

We should handle the pluralization though. #polish

lighthouse-core/test/runner-test.js
```diff
@@ -153,17 +151,17 @@ describe('Runner', () => {
       const config = new Config({
         audits: [
           // requires the HTTPS artifact
```
brendankenny: `ViewportDimensions`

lighthouse-core/test/runner-test.js
brendankenny: it's not testing that the `content-width` didn't throw an error, though, just that the artifacts make it through. Should still update the assertion value to something less confusing for the next reader

lighthouse-core/test/runner-test.js
patrickhulce: done

lighthouse-core/test/runner-test.js
patrickhulce: done

Issue #1917, opened 2017-03-27
Report: top-level warn if initialUrl and finalUrl are different


ebidel:
We need to figure out some way of notifying the user if the `finalUrl` ended up on a different origin than the `initialUrl`. Mainly for the CLI and WPT cases where users can put in a URL. For the CRX / DT, users typically end up on the final URL before testing.

Separating this out from https://github.com/GoogleChrome/lighthouse/issues/715. @brendankenny said:

> Sounds good. To close the loop: another issue with initial vs final URL is clearing storage, which is done per origin. If a redirect goes cross origin the cache, SW, etc are not cleared.

> In the CLI this is only a problem if not using the default clean Chrome. In the extension you're usually already on the final (possibly redirected) URL, except when sites redirect you to a different origin due to mobile emulation.

brendankenny:
HTTP -> HTTPS redirects should get this warning too (#2397)

connorjclark:
I'm not sure we need to have different warnings if just the origin changes. Let's do just a generic warning if requested !== final.

PR #1919, opened 2017-03-27
Offline audit: add warning if initial url != final url

### LINKED ISSUE #715: "URL responds with a 200 when offline": initial URL or final URL?

ebidel:
R: @brendankenny all

Fixes #715. I ran into this when debugging and was very confused.  Especially since the report header shows the `finalUrl` and we never tell the user there was a redirect.

https://github.com/GoogleChrome/lighthouse/issues/1917 is still open to message a top-level waring if initial url != final url.

Test with: 
```
# no warning
npm run fast -- --output=html --view https://www.chromestatus.com/features
# warning
npm run fast -- --output=html --view https://www.chromestatus.com
```

### DIFF
--- a/lighthouse-core/audits/works-offline.js
+++ b/lighthouse-core/audits/works-offline.js
@@ -30,7 +30,7 @@ class WorksOffline extends Audit {
       helpText: 'If you\'re building a Progressive Web App, consider using a service worker so ' +
           'that your app can work offline. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).',
-      requiredArtifacts: ['Offline']
+      requiredArtifacts: ['Offline', 'URL']
     };
   }
 
@@ -39,8 +39,16 @@ class WorksOffline extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    let debugString;
+    if (artifacts.URL.initialUrl !== artifacts.URL.finalUrl) {
+      debugString = 'WARNING: You may be failing this check because your test URL ' +
+          `(${artifacts.URL.initialUrl}) was redirected to ${artifacts.URL.finalUrl}. ` +
+          'Try testing the second URL directly.';
+    }
+
     return {
-      rawValue: artifacts.Offline === 200
+      rawValue: artifacts.Offline === 200,
+      debugString
     };
   }
 }
--- a/lighthouse-core/audits/works-offline.js
+++ b/lighthouse-core/audits/works-offline.js
@@ -16,6 +16,7 @@
  */
 'use strict';
 
+const URL = require('../lib/url-shim');
 const Audit = require('./audit');
 
 class WorksOffline extends Audit {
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/works-offline.js
```diff
@@ -39,8 +39,16 @@ class WorksOffline extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    let debugString;
+    if (artifacts.URL.initialUrl !== artifacts.URL.finalUrl) {
+      debugString = 'WARNING: You may be failing this check because your test URL ' +
+          `(${artifacts.URL.initialUrl}) was redirected to ${artifacts.URL.finalUrl}. ` +
```
brendankenny: quotes or something to delineate the second URL? could get messy

lighthouse-core/audits/works-offline.js
```diff
@@ -39,8 +39,16 @@ class WorksOffline extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    let debugString;
+    if (artifacts.URL.initialUrl !== artifacts.URL.finalUrl) {
```
brendankenny: do you want to use `URL.equalWithExcludedFragments` here? I don't see how you could get into a redirect situation where it would matter, but as far as network requests are concerned, with or without the fragment would be the same URL.

lighthouse-core/audits/works-offline.js
ebidel: done

lighthouse-core/audits/works-offline.js
ebidel: done

lighthouse-core/test/audits/works-offline-test.js
```diff
@@ -18,18 +18,38 @@
...
+    });
+
+    assert.equal(output.rawValue, true);
+    assert.ok(!output.debugString);
+  });
+
+  it('warns if initial url does not match final url', () => {
+    const output = Audit.audit({
+      Offline: 200,
+      URL: {initialUrl: URL, finalUrl: '${URL}/features'}
```
ebidel: whoops. this was passing just because they didn't match :)

PR #1923, opened 2017-03-28
Cherry picks for 1.6

ebidel:
R: all

Some cherry picks from master so they make it in 1.6. Of note, this includes:

- update-notifier - that'll be nice to have in place when we do the next major release
- report: legend at the bottom
- report: print functionality fixes
- report: adjustments to colors and icon sizes for a11y

All of these have already landed.


paulirish:
I think we should probably also get #1918 cherrypicked so that WPT wont be lying about HTTPS to err'one.

ebidel:
@paulirish that one doesnt merge cleanly because of the changes in https://github.com/GoogleChrome/lighthouse/commit/174766fc53b8315d5d2290f4864c04ebd30834fd.

The other thing we could do is just roll the current master as the next release. Hold up on merging https://github.com/GoogleChrome/lighthouse/pull/1896 until we do a release.

PR #1925, opened 2017-03-28
update perf.json

paulirish:
--perf was broken for a little bit after the gatherer rename of content-width

will be happy to switch to config v2 later on. :)

### DIFF
--- a/lighthouse-core/config/perf.json
+++ b/lighthouse-core/config/perf.json
@@ -7,7 +7,7 @@
     "gatherers": [
       "url",
       "image-usage",
-      "content-width",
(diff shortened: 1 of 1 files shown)


lighthouse-core/config/perf.json
```diff
@@ -16,8 +16,6 @@
     "recordNetwork": true,
     "useThrottling": false,
     "gatherers": [
-      "styles",
-      "css-usage",
       "dobetterweb/tags-blocking-first-paint",
```
patrickhulce: could probably totally nuke this pass now 💥

PR #1928, opened 2017-03-29
perf(report): Remove screenshots audit

paulirish:


brendankenny:
looks like [sample.json(https://github.com/GoogleChrome/lighthouse/blob/fdadb93bddd2de610addba08c9e0efd0d6b638a5/lighthouse-core/test/results/sample.json) and [dbw_tester-perf-results.json](https://github.com/GoogleChrome/lighthouse/blob/fdadb93bddd2de610addba08c9e0efd0d6b638a5/lighthouse-core/test/fixtures/dbw_tester-perf-results.json) need to be updated after this and the various other changes that have been made. We should probably script that...

paulirish:
yah. those have gotten stale real quick.

Issue #1887, opened 2017-03-21
[meta] V2 Report Burndown


paulirish:
_This list will be updated_

### Overview
* [ ] Add config v2 MVP #1880 [@patrickhulce]
* [ ] Add `--v2` flag. [@brendankenny]

### Report sprint
* [ ] Add report DOM renderer MVP [@dgozman]
  * [ ] Re-sort topics by failures [@robdodson]
  * [ ] Hide passing audits [@ebidel]
  * [ ] Render Markdown to DOM [@brendankenny]
  * [ ] Migrating various formatters to DOM [team]

### Audit IA refactor
* [ ] Perf metrics move [@patrickhulce]
* [ ] PWA topics collapse into audits [@paulirish]

### PWA definition - Align PWA items w/ Checklist  [@ebidel]
* [x] add TTI < 10s audit [@paulirish]
* [ ] Add placeholders for crossbrowser, etc
* [ ] Express PWA boolean along with score (decide how score and boolean relate)

wardpeet:
if it helps I can make the report cross browser ^^

ebidel:
@paulirish since we can test the additional PWA checklist items, do we need fake audits for them (crossbrowser, each page has an url, page transitions dont fee janky)? There's already a note at the top of the section that says our items are a subset of the checklist and links to it:

<img width="667" alt="screen shot 2017-03-22 at 4 51 21 pm" src="https://cloud.githubusercontent.com/assets/238208/24225838/ef04af5e-0f1f-11e7-86c5-c388a9313a48.png">

paulirish:
@ebidel not sure! we could do fake audits (a la comingSoon) or some paragraph text at the end of the section, or something else. The paragraph text doesn't sound so bad to me.

For expressing the PWA boolean I think we should plan on showing like a green checkmark if you're good.  So maybe something like

> ### Congrats, you're bonafide! 
> #### You've met the requirements for a [baseline progressive web app]()*.

> *Baseline PWAs require a few criteria that Lighthouse didn't test. 1) The site must work crossbrowser. 2) Each page must have a URL. 3) Transitions between pages dont feel janky. [Read more here]().

Just spitballing, so whatever. :)

paulirish:
closing in favor of the io burndown issue. #1937 

i filed the new #2059 to capture our convo here about the untested PWA reqs

PR #1933, opened 2017-03-29
Add basic rendering to report generator v2

dgozman:
Supports categories, audits, scores, lists, blocks and text.
Changed is-on-https audit to show list of urls as an example.
Clearly needs some design love.

Ref #1887


paulirish:
FYI y'all: it's very basic. 

Screenshot: 
image

lighthouse-core/report/v2/report-styles.css
```diff
@@ -0,0 +1,160 @@
...
+  box-sizing: border-box;
+}
+
+.hidden {
+  display: none !important;
+}
+
+/* Text */
+
+.lighthouse-text {
```
patrickhulce: We tried to follow BEM in the previous report, do we have any opinions about it in the new world?

I'm personally not a huge fan, but I like the idea of having some convention to standardize around.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+   * @param {ReportJSON} report
+   * @return {Element}
    */
-  render(element) {
-    const pre = document.createElement('pre');
-    pre.textContent = JSON.stringify(this._report, null, 2);
-    element.appendChild(pre);
+  renderReport(report) {
+    try {
+      var element = this._renderReport(report);
```
patrickhulce: might have to make a decision if we follow devtools style and conventions or lighthouse (or drift the rest of lighthouse to devtools?). use of `const`, `let` vs. `var`, object shorthand, etc

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+    return element;
+  }
+
+  /**
+   * @param {Error} e
+   * @return {Element}
+   */
+  _renderException(e) {
+    var element = this._createElement('div', 'lighthouse-exception');
+    element.textContent = '' + e;
```
patrickhulce: thoughts on `e.stack` instead for a little more info?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+      title += ': ' + audit.result.displayValue;
+    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
+    if (audit.result.details)
+      element.appendChild(this._renderDetails(audit.result.details));
+    return element;
   }
 };
 
+// /** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */
+// var DetailsJSON;
```
dgozman: Linter doesn't like var here, but using let will pollute the scope. Is there a known way for typedefs?

brendankenny: > Linter doesn't like var here, but using let will pollute the scope. Is there a known way for typedefs?

yeah, I was thinking about this when looking at enabling closure compilation of the project again. If we do more typedefs in files just meant for node/browserification, the nice thing is that those will be limited in scope to the file. For stuff in the browser it's harder. A few ideas:

1) we could add an .eslintrc to this directory to modify the rules, since in theory this js file will only ever run in a browser context and not in node (never say never, though :)
2) NTI doesn't support typedefs as namespaces, but would doing `ReportRenderer.DetailsJSON` work here since the namespace would be a real object?
3) we live with the pollution and in the future add something to `report-generator.js` that strips it out
4) We live with the pollution and pick names that will only ever be used for typedefs

paulirish: My vote is for 3.  The pollution seems very manageable, for the limited number of typedefs we have.

brendankenny: > My vote is for 4

SGTM2.

We were also just talking and realized it doesn't matter if this is `var` or `let` here anyways, since even if we use `var` here if somewhere else a global with the same name is declared with `let` or `const` it will throw an error. So I'm good with `let` and global pollution.

lighthouse-core/report/v2/report-styles.css
paulirish: Agree with having a convention. That said, I don't love any of the options. But I think we need to pick one regardless

Honestly BEM probably has the most tooling support of any CSS convention, so I favor that.

Should we migrate to it in this PR? I could go either way.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+    return element;
+  }
+
+  /**
+   * @param {Error} e
+   * @return {Element}
+   */
+  _renderException(e) {
+    const element = this._createElement('div', 'lighthouse-exception');
+    element.textContent = String(e);
```
paulirish:
```js
String(e.stack)
```
so we can get the message and callstack.

dgozman: Done

lighthouse-core/report/v2/report-styles.css
dgozman: Adopted BEM.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+      title += ': ' + audit.result.displayValue;
+    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
+    if (audit.result.details)
+      element.appendChild(this._renderDetails(audit.result.details));
+    return element;
   }
 };
 
+/** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */
+let DetailsJSON;
```
brendankenny: can do `let DetailsJSON; // eslint-disable-line no-unused-vars` for each of these to disable the lint failure on just these lines

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+   * @param {number} score
+   * @param {string} title
+   * @param {string} description
+   * @return {Element}
+   */
+  _renderScore(score, title, description) {
+    const element = this._createElement('div', 'lighthouse-score');
+    const value = this._createChild(element, 'div', 'lighthouse-score__value');
+    value.textContent = score.toLocaleString(undefined, {maximumFractionDigits: 1});
+    if (score <= 50)
```
ebidel: For scoring, you can use this from the existing report:

```
const RATINGS = {
  GOOD: {label: 'good', minScore: 75},
  AVERAGE: {label: 'average', minScore: 45},
  POOR: {label: 'poor'}
};
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+
+  /**
+   * @param {number} score
+   * @param {string} title
+   * @param {string} description
+   * @return {Element}
+   */
+  _renderScore(score, title, description) {
+    const element = this._createElement('div', 'lighthouse-score');
+    const value = this._createChild(element, 'div', 'lighthouse-score__value');
```
ebidel: could easily get away without having a `_createChild`:

`element.appendChild(this._createElement('div', 'lighthouse-score__value'));`

lighthouse-core/report/v2/report-template.html
```diff
@@ -21,14 +21,15 @@
...
 </head>
 <body>
   <noscript>Lighthouse report requires JavaScript. Please enable.</noscript>
   <script>window.__LIGHTHOUSE_JSON__ = %%LIGHTHOUSE_JSON%%</script>
   <script>%%LIGHTHOUSE_JAVASCRIPT%%</script>
   <script>
-    const renderer = new ReportRenderer(window.__LIGHTHOUSE_JSON__);
-    renderer.render(document.body);
+    const renderer = new ReportRenderer(document);
+    document.body.appendChild(renderer.renderReport(window.__LIGHTHOUSE_JSON__));
```
ebidel: Hopefully in the near future, we can move some of the UI into static HTML that gets populated. Is there a plan for that? I'm still very concerned about the maintainability and scalability of a pure dom-based approach.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+      return this._renderException(e);
+    }
+  }
+
+  /**
+   * @param {string} name
+   * @param {string=} className
+   * @returns {Element}
+   */
+  _createElement(name, className) {
```
dgozman: Let's do it when we need it?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+    parent.appendChild(child);
+    return child;
+  }
+
+  /**
+   * @param {Element} parent
+   * @param {string} text
+   * @return {Node}
+   */
+  _createTextChild(parent, text) {
```
dgozman: Nope. I'll remove it. Thanks!

lighthouse-core/report/v2/report-renderer.js
dgozman: _createChild returns an element for further manipulations.

lighthouse-core/report/v2/report-template.html
dgozman: This is against the idea of report being embeddable and template-free.
Regarding maintainability: we plan to implement some basic primitives like list/table/url/etc, so that on average nobody has to touch report generator at all.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -79,9 +80,11 @@ class ReportGeneratorV2 {
   generateReportHtml(reportAsJson) {
     const sanitizedJson = JSON.stringify(reportAsJson).replace(/</g, '\\u003c');
     const sanitizedJavascript = REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/');
+    const sanitizedCss = REPORT_CSS.replace(/<\//g, '\\u003c/');
```
patrickhulce: @ebidel I responded to that. This wasn't for XSS, this was for future-proof against us somewhere using `</script>` in a string and scratching heads for why the whole report is broken. We do it [currently](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/report-generator.js#L40) too.

patrickhulce: Can rename to `escaped` or something if we want to make that clearer

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -19,19 +19,190 @@
...
+   * @param {number} score
+   * @param {string} title
+   * @param {string} description
+   * @return {Element}
+   */
+  _renderScore(score, title, description) {
+    const element = this._createElement('div', 'lighthouse-score');
+    const value = this._createChild(element, 'div', 'lighthouse-score-value');
+    value.textContent = score.toLocaleString(undefined, {maximumFractionDigits: 1});
+    if (score <= 50)
```
ebidel: Sorry for the double comments. Github was being weird today.

lighthouse-core/report/v2/report-generator.js
ebidel: We'd know right away if the whole report was broken :) The current `_escapeScriptTags` does the same as your `sanitizedJson`, which sanitizes user input in the json results. But our own stuff should be safe (`REPORT_JAVASCRIPT` and `REPORT_CSS`)

lighthouse-core/report/v2/report-template.html
```diff
@@ -21,14 +21,15 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
   <title>Lighthouse Report</title>
+  <style>%%LIGHTHOUSE_CSS%%</style>
```
ebidel: This one causes vscode's syntax highlighting to blow up. WDYT about making this replace str a css comment: <style>/*%%LIGHTHOUSE_CSS%%*/</style>?

lighthouse-core/report/v2/report-template.html
ebidel: I don't think those goals are at odds with a declarative approach. For example, we could declare `<template>` partials in the page:

```
<template id="template-audit">
  <div class="lighthouse-audit">
    <div class="lighthouse-score">
      <div class="lighthouse-score__value"><!--score--></div>
      <div class="lighthouse-score__text">
        <div class="lighthouse-score__title"><!--title--></div>
        <div class="lighthouse-score__description"><!--detail--></div>
      </div>
    </div>
  </div>
</template>
```
http://jsbin.com/bojinipecu/edit?html,output

And various embedders could instantiate slightly different `<template>`s based on their needs. Alternatively, override a `render` method and add their extra functionality. It's still 100% DOM, but it would be easier to reason about app structure and make modifications to the UI.

This would also lend itself nicely to creating reusable chunks of markup+functionality in the form of web components. Each component would be responsible for its own css/html/js. DevTools could extend a given component if it needs something different.

```
const categories = __LIGHTHOUSE_JSON__.reportCategories;

const container = document.querySelector('lighthouse-category');
categories.forEach(audit => {
  const auditEl = document.createElement('lighthouse-audit');
  auditEl.audit = audit;
  container.appendChild(auditEl);
});
```

might produce something like:

```
<lighthouse-category>
  <lighthouse-audit id="critical-request-chains">
    #shadow-root
      <div class="lighthouse-score">...</div>
      ...
      <card-fortmatter></card-fortmatter>
  </lighthouse-audit>
  <lighthouse-audit>...</lighthouse-audit>
</lighthouse-category>
```

paulirish:
There's not too much left here. 
@ebidel do you want to finish off this PR? if not, i'm happy to.

ebidel:
Closing down for https://github.com/GoogleChrome/lighthouse/pull/1951

PR #1938, opened 2017-03-30
Improve types for cli

denar90:
Some small improvements for cli types.
Also we wanna reuse types form LH in pwmetrics and right now we have to extends it -https://github.com/paulirish/pwmetrics/blob/master/types/types.ts#L5-L15,  but I hope there no objections to have it here :)

### DIFF
--- a/lighthouse-cli/types/types.ts
+++ b/lighthouse-cli/types/types.ts
@@ -10,6 +10,10 @@ interface AuditResult {
   };
 }
 
+interface AuditResults {
+  [metric: string]: AuditResult
+}
+
 interface AuditFullResult {
   rawValue: boolean|number;
   displayValue: string;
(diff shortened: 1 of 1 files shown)


brendankenny:
> Also we wanna reuse types form LH in pwmetrics and right now we have to extends it -https://github.com/paulirish/pwmetrics/blob/master/types/types.ts#L5-L15, but I hope there no objections to have it here :)

we're taking another look at the closure compiler to check the jsdocs we do (often) use. *If* we do go down that path, it might be good to set up [`clutz`](https://github.com/angular/clutz) to automatically generate ts declaration files for your use case (there are a few issues with it still, but it could probably still get us at least 90% of the way there)

denar90:
@brendankenny thx :)

PR #1944, opened 2017-03-31
gentlejack

gentlejack:
Declarative report with using polymer 2


paulirish:
We're still not clear on this history of this PR. (btw both me and Github staff looked into it and didn't see anything fishy with its origin)  ¯\_(ツ)_/¯

But regardless I think we can close it now.

For search purposes, let's make sure it'll show up with any of these keywords: polymer polymer2 custom elements, web components, templates

<3z

PR #1947, opened 2017-03-31
feat(tti): increase robustness of perf metrics

patrickhulce:
First round of perf metric investigation, to be judged using #1936 :) Follow-up work includes adding in network reverse search methods for comparison.

- Adds two additional TTI candidates that don't rely on visual completeness.
	- First 500ms window after fMP where 90th percentile EIL is <50ms
	- First 5000ms window after fMP where 90th percentile EIL is <50ms (if at the end of trace, uses a minimum window of as low as 1000ms)
- Falls back to fMP when speedline fails to find screenshots in trace.
- Adds timing data to trace-of-tab computed artifact.

### DIFF
--- a/lighthouse-core/audits/estimated-input-latency.js
+++ b/lighthouse-core/audits/estimated-input-latency.js
@@ -44,9 +44,9 @@ class EstimatedInputLatency extends Audit {
     };
   }
 
-  static calculate(speedline, model, trace) {
+  static calculate(tabTrace, model, trace) {
     // Use speedline's first paint as start of range for input latency check.
-    const startTime = speedline.first;
+    const startTime = tabTrace.timings.firstMeaningfulPaint;
 
     const latencyPercentiles = TracingProcessor.getRiskToResponsiveness(model, trace, startTime);
     const ninetieth = latencyPercentiles.find(result => result.percentile === 0.9);
@@ -84,11 +84,12 @@ class EstimatedInputLatency extends Audit {
     const trace = artifacts.traces[this.DEFAULT_PASS];
 
     const pending = [
-      artifacts.requestSpeedline(trace),
+      artifacts.requestTraceOfTab(trace),
       artifacts.requestTracingModel(trace)
     ];
-    return Promise.all(pending).then(([speedline, model]) =>
-        EstimatedInputLatency.calculate(speedline, model, trace));
+    return Promise.all(pending).then(([tabTrace, model]) => {
+      return EstimatedInputLatency.calculate(tabTrace, model, trace);
+    });
   }
 }
 
--- a/lighthouse-core/audits/estimated-input-latency.js
+++ b/lighthouse-core/audits/estimated-input-latency.js
@@ -45,7 +45,6 @@ class EstimatedInputLatency extends Audit {
   }
 
   static calculate(tabTrace, model, trace) {
-    // Use speedline's first paint as start of range for input latency check.
     const startTime = tabTrace.timings.firstMeaningfulPaint;
 
     const latencyPercentiles = TracingProcessor.getRiskToResponsiveness(model, trace, startTime);
--- a/lighthouse-core/audits/time-to-interactive.js
+++ b/lighthouse-core/audits/time-to-interactive.js
@@ -42,10 +42,9 @@ class TTIMetric extends Audit {
    * @param {number} maxTime
    * @param {{model: !Object, trace: !Object}} data
    * @param {number=} windowSize
-   * @param {number=} minimumWindowSize
    * @return {{timeInMs: number|undefined, currentLatency: number, foundLatencies: !Array}}
    */
-  static _forwardWindowTTI(minTime, maxTime, data, windowSize = 500, minimumWindowSize = 500) {
+  static _forwardWindowTTI(minTime, maxTime, data, windowSize = 500) {
     // Find first window where Est Input Latency is <50ms at the 90% percentile.
     let startTime = minTime - 50;
     let endTime;
@@ -60,7 +59,7 @@ class TTIMetric extends Audit {
       startTime += 50;
       endTime = startTime + windowSize;
       // If there's no more room in the trace to look, we're done.
-      if (maxTime - startTime < minimumWindowSize) {
+      if (endTime > maxTime) {
         return {currentLatency, foundLatencies};
       }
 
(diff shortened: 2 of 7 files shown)


lighthouse-core/audits/time-to-interactive.js
```diff
@@ -90,45 +187,23 @@ class TTIMetric extends Audit {
...
-        }
-        // Get our expected latency for the time window
-        const latencies = TracingProcessor.getRiskToResponsiveness(
-          model, trace, startTime, endTime, percentiles);
-        const estLatency = latencies[0].time;
-        foundLatencies.push({
-          estLatency: estLatency,
-          startTime: startTime.toFixed(1)
-        });
+      const times = {fmpTiming, visuallyReadyTiming, endOfTraceTime};
```
paulirish: can you put a `lastTraceEvtTs` into trace-of-tab?

```js
const traceEndTsInMS = trace.traceEvents.reduce(
          (acc, traceEvent) => Math.max(acc, traceEvent.ts), 0) / 1000;
```
it's from will's https://github.com/GoogleChrome/lighthouse/pull/1936/files#diff-fd99fc8068c6555d1fb3d1c70df7c29cR82

wec can then reuse here (and maybe drop the traceviewer dep) and also in will's #1936

lighthouse-core/test/audits/estimated-input-latency-test.js
```diff
@@ -37,8 +37,8 @@ describe('Performance: estimated-input-latency audit', () => {
     const artifacts = generateArtifactsWithTrace({traceEvents: pwaTrace});
     return Audit.audit(artifacts).then(output => {
       assert.equal(output.debugString, undefined);
-      assert.equal(output.rawValue, 17.4);
-      assert.equal(output.displayValue, '17.4ms');
+      assert.equal(output.rawValue, 16.7);
```
paulirish: to be clear, these changes are only due to the shift from speedline.first to FMP. yes?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -92,7 +92,16 @@ class TraceOfTab extends ComputedArtifact {
       .filter(e => e.pid === startedInPageEvt.pid)
       .sort((event0, event1) => event0.ts - event1.ts);
 
+    const metrics = {navigationStart, firstPaint, firstContentfulPaint, firstMeaningfulPaint};
+    const timings = {};
+
+    Object.keys(metrics).forEach(metric => {
+      timings[metric + 'Ts'] = metrics[metric] && metrics[metric].ts / 1000;
```
paulirish: the structure of the extInfo objects in TTI/FMP includes both a timings object and a timestamps object. should we do the same here?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -92,7 +92,16 @@ class TraceOfTab extends ComputedArtifact {
...
+      timings[metric] = timings[metric + 'Ts'] - timings.navigationStartTs;
```
paulirish: adding this guy in here means that if we pass along the data in extInfo then tools like pwmetrics could skip this work: https://github.com/paulirish/pwmetrics/blob/master/lib/metrics.ts#L54-L61

lighthouse-core/audits/time-to-interactive.js
```diff
@@ -37,6 +37,99 @@ class TTIMetric extends Audit {
...
+  }
+
+  /**
+   * @param {{fmpTiming: number, visuallyReadyTiming: number, endOfTraceTime: number}} times
+   * @param {{model: !Object, trace: !Object}} data
+   * @return {{timeInMs: number|undefined, currentLatency: number, foundLatencies: !Array}}
+   */
+  static findTTIAlpha(times, data) {
+    return TTIMetric._forwardWindowTTI(
+      Math.max(times.fmpTiming, times.visuallyReadyTiming),
```
paulirish: can you add a comment here to clarify you are using fmp as fallback?

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -44,9 +44,9 @@ class EstimatedInputLatency extends Audit {
     };
   }
 
-  static calculate(speedline, model, trace) {
+  static calculate(tabTrace, model, trace) {
     // Use speedline's first paint as start of range for input latency check.
```
patrickhulce: done

lighthouse-core/test/audits/estimated-input-latency-test.js
patrickhulce: yup

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: sgtm

lighthouse-core/audits/time-to-interactive.js
patrickhulce: done, but naming `traceEnd` :) to match with event names :)

lighthouse-core/audits/time-to-interactive.js
patrickhulce: done

lighthouse-core/audits/time-to-interactive.js
```diff
@@ -63,9 +63,6 @@ class TTIMetric extends Audit {
         return {currentLatency, foundLatencies};
       }
 
-      // Clip to the trace end
-      endTime = Math.min(endTime, maxTime);
```
paulirish: hmm! why'd you nuke this?
 
fwiw: https://docs.google.com/document/d/1b9slyaB9yho91YTOkAQfpCdULFkZM9LqsipcX3t7He8/edit#heading=h.ooalxriljq1i has details on the clipping approach for EIL

cc @brendankenny

patrickhulce: oh I added this when I did the 5000ms - 1000ms if we reached the end of trace for WPT TTI, but after our convo on Friday about how fake that is I nuked the variable window and just kept the 5 seconds so it wasn't necessary anymore

paulirish: gotcha. ok

patrickhulce:
> want to update pwmetrics-events so we can get these new TTIs plotted in the trace?

done

image

PR #1951, opened 2017-04-01
Report generator v2 basics (more stuff)

ebidel:
R: all

This takes the feedback from https://github.com/GoogleChrome/lighthouse/pull/1933 and brings over more of our styling.

I suggest **reviewing by commit**. Just mine should be fine.

>  Supports categories, audits, scores, lists, blocks and text.
Changed is-on-https audit to show list of urls as an example.
Clearly needs some design love.
> Ref #1887 

Updated look from from https://github.com/GoogleChrome/lighthouse/pull/1933:

screen shot 2017-04-01 at 2 13 08 pm


lighthouse-core/report/v2/report-styles.css
```diff
@@ -0,0 +1,176 @@
...
+}
+
+.--hidden {
+  display: none !important;
+}
+
+/* Text */
+
+/*.lighthouse-text {
+  white-space: nowrap;
```
ebidel: These aren't needed yet so I commented them out. There was also a lot of unnecessary CSS flexbox that was removed. Divs are already block level elements 👍

paulirish:
What a pro. Looking good, dude.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,184 @@
...
+    this._document = document;
+  }
+
+  /**
+   * @param {ReportJSON} report
+   * @return {Element}
+   */
+  renderReport(report) {
+    try {
+      const element = this._renderReport(report);
```
patrickhulce: nit: just nuke the variable?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,184 @@
...
+   * @param {string=} className
+   * @param {Array<Object>=} attrs Attribute key/val pairs.
+   * @returns {Element}
+   */
+  _createElement(name, className, attrs = []) {
+    const element = this._document.createElement(name);
+    if (className) {
+      element.className = className;
+    }
+    for (const [key, val] of Object.entries(attrs)) {
```
patrickhulce: huh? This is really hard to parse, and I can't find usage of it to help :) What's wrong with `attrs.forEach(attr => el.setAttribute(attr.key, attr.value))`?

patrickhulce: oooooh is the doctype/default wrong and you mean `attrs` to be an object?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,184 @@
...
+    if (audit.result.details) {
+      element.appendChild(this._renderDetails(audit.result.details));
+    }
+    return element;
   }
 };
+})();
+
+/** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */
+let DetailsJSON; // eslint-disable-line no-unused-vars
```
patrickhulce: what will break if we stick these in the iife? needed for closure in other files?

lighthouse-core/report/v2/report-styles.css
```diff
@@ -0,0 +1,176 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ :root {
+  --text-font-family: '.SFNSDisplay-Regular', 'Helvetica Neue', 'Lucida Grande', sans-serif;
+  --body-font-size: 13px;
+
+  --secondary-text-color: #565656;
```
patrickhulce: now that Lighthouse reports are publicly exposed over on WPT it'd be nice to settle exactly what browsers we want to support and test the report on as we make decisions about what platform features to use, how hard they are to polyfill, etc

lighthouse-core/report/v2/report-renderer.js
ebidel: Note sure. My closure forward declarations foo is 100% non-existent. cc @brendankenny

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js
ebidel: Yea sorry, docs are wrong. Done.

lighthouse-core/report/v2/report-styles.css
ebidel: +1. I've been operating under the assumption that our baseline is the latest version of each modern browser.

Good news is [everyone has CSS custom props](http://caniuse.com/#feat=css-variables) except for Edge. And they'll get them in their next release (Edge 15).

Regardless, Edge will be an investment that we haven't made before and it might not be worth it. They're catching up on features quickly. Any work we put in now might be moot in the near future. FWIW, < 0.60% Viewer users are using Edge...so a fraction of the already small browser share :) I'd be curious to see the analytics for WPT though.

One thing that would also be interesting is offloading polyfills and broader browser support to Viewer. The more code we to ship in the binary, the larger the report output becomes.

ebidel:
PTAL

lighthouse-core/report/v2/report-styles.css
patrickhulce: > One thing that would also be interesting is offloading polyfills and broader browser support to Viewer. The more code we to ship in the binary, the larger the report output becomes.

Definitely, I see a world in which a hosted polyfill'd/browserify'd/whatever'd renderer bundle being included just via a single `<script src="jsdelivr.net/lighthouse/v2.x>` in WPT-style reports via CLI flag or something and DevTools/opt-in CLI users being the only ones to consume the source directly

> I'd be curious to see the analytics for WPT though.

I think this is the more important information to track when considering if we need to support Edge

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
-    element.appendChild(pre);
+  _renderText(text) {
+    const element = this._createElement('div', 'lighthouse-text');
+    element.textContent = text.text;
+    return element;
+  }
+
+  /**
+   * @param {DetailsJSON} block
+   * @return {Element}
```
brendankenny: `{!DetailsJSON}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    const element = this._createElement('div', 'lighthouse-block');
+    for (const item of block.items) {
+      element.appendChild(this._renderDetails(item));
+    }
+    return element;
+  }
+
+  /**
+   * @param {DetailsJSON} list
+   * @return {Element}
```
brendankenny: `{!DetailsJSON}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    for (const item of list.items) {
+      items.appendChild(this._renderDetails(item));
+    }
+    element.appendChild(items);
+    return element;
+  }
+
+  /**
+   * @param {Error} e
+   * @return {Element}
```
brendankenny: `{!Error}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+   */
+  _renderException(e) {
+    const element = this._createElement('div', 'lighthouse-exception');
+    element.textContent = String(e.stack);
+    return element;
+  }
+
+  /**
+   * @param {ReportJSON} report
+   * @return {Element}
```
brendankenny: `{!ReportJSON}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    const element = this._createElement('div', 'lighthouse-report');
+    for (const category of report.reportCategories) {
+      element.appendChild(this._renderCategory(category));
+    }
+    return element;
+  }
+
+  /**
+   * @param {CategoryJSON} category
+   * @return {Element}
```
brendankenny: `{!CategoryJSON}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    element.appendChild(this._renderScore(category.score, category.name, category.description));
+    for (const audit of category.audits) {
+      element.appendChild(this._renderAudit(audit));
+    }
+    return element;
+  }
+
+  /**
+   * @param {AuditJSON} audit
+   * @return {Element}
```
brendankenny: `{!AuditJSON}`, `{!Element}`

lighthouse-core/report/v2/report-renderer.js
brendankenny: Closure will follow scoping rules for these. In this case it doesn't matter at this point since they're only ever used in this file

lighthouse-core/report/v2/report-styles.css
brendankenny: > These aren't needed yet so I commented them out

just delete then?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -16,6 +16,38 @@
 'use strict';
 
 /* eslint-env browser */
+/* eslint indent: ["error", 2, { "outerIIFEBody": 0 }] */
+
+(function() {
```
brendankenny: why the IIFE?

lighthouse-core/audits/is-on-https.js
```diff
@@ -72,6 +72,11 @@ class HTTPS extends Audit {
       extendedInfo: {
         formatter: Formatter.SUPPORTED_FORMATS.URL_LIST,
         value: insecureRecords
+      },
+      details: {
```
brendankenny: two thoughts on details:

1) I get the point that the `type: 'text'` on every node is basically JSON HTML, but it seems like it's valuing simplicity in the report renderer at the expense of more verbose and less readable JSON results. A simple fix in this case would be to let the header type determine the type of rendering used for items (and if we ever encounter the need for mixed-type items we can deal with it then)

2) This can wait for a followup PR: we shouldn't have an `extendedInfo` and a `details` as those are basically synonyms. Either we should pick clearer names for the two (one is for report rendering and the other is going to be for information someone somewhere might need maybe), or just have `details` and have `extendedInfo` be a property on it

lighthouse-core/closure/typedefs/AuditResult.js
```diff
@@ -51,6 +51,9 @@ AuditResult.prototype.optimalValue;
 /** @type {(AuditExtendedInfo|undefined|null)} */
 AuditResult.prototype.extendedInfo;
 
+/** @type {(Object|undefined)} */
```
brendankenny: if `details` is so regular, it would be nice if these were actually typed :)

lighthouse-core/audits/is-on-https.js
brendankenny: > I get the point that the type: 'text' on every node is basically JSON HTML, but it seems like it's valuing simplicity in the report renderer at the expense of more verbose and less readable JSON results

Now that I look at the typedef, I'm even more curious about the thinking behind the `details` object. `/** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */` seems like an overgeneralization. It's cool you could render a list in the header of a list, but...why? Seems better to make a new `type` at that point, otherwise a reader will just wonder what on earth is going on.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+
+  /**
+   * @param {DetailsJSON} details
+   * @return {Element}
    */
-  constructor(report) {
-    this._report = report;
+  _renderDetails(details) {
+    switch (details.type) {
+    case 'text':
```
paulirish: current eslint config wants it flat. (fwiw prettify expects another level). 

i dont really care but if we're bikeshedding then i'll  add i much preferred when the `case` was on the same line as the resulting statement. :)

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+   * @param {string=} className
+   * @param {Object=} attrs Attribute key/val pairs.
+   * @returns {Element}
+   */
+  _createElement(name, className, attrs = {}) {
+    const element = this._document.createElement(name);
+    if (className) {
+      element.className = className;
+    }
+    for (const [key, val] of Object.entries(attrs)) {
```
paulirish: `/* eslint-env browser */` already present. im comfy with assuming this is only run clientside

lighthouse-core/report/v2/report-renderer.js
brendankenny: > current eslint config wants it flat

is it because of the eslint indent directive added to the file? When I try the same thing in a random other file in the project it wants it indented

lighthouse-core/report/v2/report-renderer.js
ebidel: to prevent leakage on window

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
-   * @param {!Object} report
+   * @param {Document} document
+   */
+  constructor(document) {
+    this._document = document;
+  }
+
+  /**
+   * @param {ReportJSON} report
+   * @return {Element}
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    } catch (e) {
+      return this._renderException(e);
+    }
+  }
+
+  /**
+   * @param {string} name
+   * @param {string=} className
+   * @param {Object=} attrs Attribute key/val pairs.
+   * @returns {Element}
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+      return this._renderReport(report);
+    } catch (e) {
+      return this._renderException(e);
+    }
+  }
+
+  /**
+   * @param {string} name
+   * @param {string=} className
+   * @param {Object=} attrs Attribute key/val pairs.
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+      element.setAttribute(key, val);
+    }
+    return element;
+  }
+
+  /**
+   * @param {number} score
+   * @param {string} title
+   * @param {string} description
+   * @return {Element}
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+    column.appendChild(this._createElement('div', 'lighthouse-score__title')).textContent = title;
+    column.appendChild(
+      this._createElement('div', 'lighthouse-score__description')).textContent = description;
+
+    return element;
+  }
+
+  /**
+   * @param {DetailsJSON} details
+   * @return {Element}
```

lighthouse-core/report/v2/report-renderer.js
brendankenny: > to prevent leakage on window
     
But this is the only script on the page :)

lighthouse-core/report/v2/report-renderer.js
ebidel: yea i prefer indented by our eslint prevents it.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -25,19 +57,183 @@
...
+      return this._renderList(details);
+    default:
+      throw new Error(`Unknown type: ${details.type}`);
+    }
   }
 
   /**
-   * @param {!Element} element
+   * @param {DetailsJSON} text
+   * @return {Element}
```

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js
brendankenny: looks like it's because the `/* eslint indent: ["error", 2, { "outerIIFEBody": 0 }] */` line in the file is [overriding our default](https://github.com/GoogleChrome/lighthouse/blob/bf6a0a4ca8875df3ccfb973f71e6a1eae79451d4/.eslintrc.js#L11-L14), which calls out the switch thing explicitly. /* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */ should work in that case

lighthouse-core/audits/is-on-https.js
ebidel: I'd be in favor of renaming `extendedInfo` to `details`. We don't need both and `extendedInfo` is really just a details object with a specified formatter.

When I was looking over this code, I also found the generalization of `_renderDetails` to be premature. Not sure we need recursive rendering of details just yet (`_renderDetails` calls `_renderBlock|List|Text` which in turn, call `_renderDetails`). That's nice if we ever have nests of `<details>` but to date, haven't need that.

ebidel: But just to be clear, let's do all of that in another PR.

paulirish: I'm in favor of keeping extendedInfo and details as separate things. (The names could change though).

I really like that extendedInfo can hold all the metadata so downstream tools can consume and analyze the extra information we have on the audit. 

For example, pwmetrics, lighthouse-plots are consuming data that's in extendedInfo. And `perf-battle` is, as well: https://github.com/shakyShane/perf-battle/blob/31943652eb08c7bb1047805032b99d0aff82ad11/src/output-size.ts#L17

I think its wise for this AuditResult object to cleanly separate the data we need to render the report (in a reliable structure) vs the extra data we can offer (with no structure guarantees).


-------

> Not sure we need recursive rendering of details just yet 

how about datagrid where the first column is a URL? URL will likely need its own `render` as it'll get forked for devtools revealing.

paulirish:
btw can we try to preserve authorship when we squash land this? I'll volunteer to flatten to two commits.

lighthouse-core/audits/is-on-https.js
patrickhulce: > how about datagrid where the first column is a URL? URL will likely need its own render as it'll get forked for devtools revealing.

or when URL is only sometimes a URL like in stylesheets that are inline and have previews. I think this is a very clean way to handle these situations and is a great step forward from the hackiness of old formatters. 

Also +1 to keeping `extendedInfo` and `details` separate for Paul's reasons and an aspiration to keep the presentation and raw data as separate as possible. All for renaming to make the intention of each object clearer though :)

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-styles.css
ebidel: done

lighthouse-core/audits/is-on-https.js
ebidel: On interesting, granularity is good. I was thinking devtools would replace entire methods (`_renderTable`) with its own implementation or call us and annotate after the fact. Still waiting on answers to how the latter works :)

brendankenny: > For example, pwmetrics, lighthouse-plots are consuming data that's in extendedInfo

> with no structure guarantees

sounds like a good way to break on every release...

> how about datagrid where the first column is a URL? URL will likely need its own render as it'll get forked for devtools revealing.

wouldn't a column of type URL work in that case? That's how the table formatter works now.

> or when URL is only sometimes a URL like in stylesheets that are inline and have previews

this is more of an issue, but still not clear why it can't be handled in the report, e.g. the element could default to the column type unless it specifies its own. Or whatever we come up with when we get to it. Anything to avoid `{type: 'text', text: 'my text'}` everywhere to enable an arbitrary nesting feature we'll rarely use anywhere.

(I also doubt that longterm our CSS will allow arbitrary nesting of components, so it does seem like the DOM generation code should mirror that)

brendankenny: > I was thinking devtools would replace entire methods (`_renderTable`) with its own implementation

I think that could still happen, unless we bring the devtools table renderer into the report

ebidel:
Travis is super happy right now. We feel good about merging this for now? Details/extendedInfo to be worked out....

paulirish:
I'm happy. I think Patrick's happy. @brendankenny u ok if Eric merges?

ebidel:
I am the Squashinator! 💪

paulirish:
yes

PR #1956, opened 2017-04-03
Docs: update gulp recipe to return promise chain. Fixes #1955

### LINKED ISSUE #1955: Process.exit inside gulp example
What is the reason process.exit() is required at the end of the run? 
https://github.com/GoogleChrome/lighthouse/blob/master/docs/recipes/gulp/gulpfile.js#L75
I'm trying to integrate pwmetrics as part of my CI-chain, but currently the only solution is to spawn it in a different process. I've asked the question within the pwmetrics repo, but they redirected me to lighthouse: https://github.com/paulirish/pwmetrics/issues/88

ebidel:


docs/recipes/gulp/gulpfile.js
```diff
@@ -70,21 +70,23 @@ const runLighthouse = function() {
...
 };
 
 /**
  * Handle error
  */
-const handleError = function() {
+const handleError = function(e) {
   stopServer();
+  console.error(e);
   process.exit(1);
```
paulirish: shouldn't we just rethrow the error here instead of exiting?

ebidel: Either works.

paulirish: aiight. @Bubbit does exit on error matter to you?

Bubbit: I would do both, one commented out with a short explanation (for example in my CI pipeline I would let it through, as I'm running multiple test tools), but I guess the gulp standard is returning the promise?

ebidel: updated to throw

ebidel:
PTAL

PR #1959, opened 2017-04-03
fix: add more helpful error message when fMP is missing

patrickhulce:
filed #1960 to track the issue itself


addyosmani:
👍 Better than what we had here before. 

I do wonder with failures like this as a whole whether we should be more proactive in telling users to re-run Lighthouse vs considering it a fail state they can't recover from.

brendankenny:
> I do wonder with failures like this as a whole whether we should be more proactive in telling users to re-run Lighthouse vs considering it a fail state they can't recover from.

yeah, we already see this in plots, where we should add plotting of the error rate as one of our metrics, but should also probably discard those runs to get full samples

PR #1963, opened 2017-04-04
Report v2: check/x boolean audits, expand details failing audits, optimal val

ebidel:
R: all

Before: 

screen shot 2017-04-04 at 12 51 49 pm

After: 

screen shot 2017-04-04 at 12 49 07 pm

- Expands failing audits + expander arrows
- Support for boolean audits with X/check
- Renders `optimalValue` text on audit title. Final UI we'll need to work out.

### DIFF
--- a/docs/hacking-tips.md
+++ b/docs/hacking-tips.md
@@ -16,11 +16,13 @@ node --trace-warnings lighthouse-cli http://example.com
 ## Updating fixture dumps
 
 `lighthouse-core/test/results/samples_v2.json` is generated from running LH against
-dbw_tester.html. To update this file:
+dbw_tester.html. To update this file, run:
 
 ```sh
 npm run start -- --output=json --output-path=lighthouse-core/test/results/sample_v2.json http://localhost:8080/dobetterweb/dbw_tester.html
 ```
+
+After updating, consider deleting any irrelevant changes from the diff (exact timings, timestamps, etc). Be sure to run the tests.
 
 ## Iterating on the v2 report
 
--- a/docs/hacking-tips.md
+++ b/docs/hacking-tips.md
@@ -13,9 +13,18 @@ Use [`--trace-warnings`](https://medium.com/@jasnell/introducing-process-warning
 node --trace-warnings lighthouse-cli http://example.com
 ```
 
+## Updating fixture dumps
+
+`lighthouse-core/test/results/samples_v2.json` is generated from running LH against
+dbw_tester.html. To update this file:
+
+```sh
+npm run start -- --output=json --output-path=lighthouse-core/test/results/sample_v2.json http://localhost:8080/dobetterweb/dbw_tester.html
+```
+
 ## Iterating on the v2 report
 
-This'll generate new reports from the same results json.
+This will generate new reports from the same results json.
 
 ```sh
 # capture some results first:
--- a/lighthouse-core/report/v2/report-generator.js
+++ b/lighthouse-core/report/v2/report-generator.js
@@ -97,14 +97,12 @@ class ReportGeneratorV2 {
   generateReportHtml(reportAsJson) {
     const sanitizedJson = JSON.stringify(reportAsJson).replace(/</g, '\\u003c');
     const sanitizedJavascript = REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/');
-    // Remove script in templates files just to be safe.
-    const sanitizedTemplates = REPORT_TEMPLATES.replace(/<script>([\s\S]*?)<\/script>/g, '');
 
     return ReportGeneratorV2.replaceStrings(REPORT_TEMPLATE, [
       {search: '%%LIGHTHOUSE_JSON%%', replacement: sanitizedJson},
       {search: '%%LIGHTHOUSE_JAVASCRIPT%%', replacement: sanitizedJavascript},
       {search: '/*%%LIGHTHOUSE_CSS%%*/', replacement: REPORT_CSS},
-      {search: '%%LIGHTHOUSE_TEMPLATES%%', replacement: sanitizedTemplates},
+      {search: '%%LIGHTHOUSE_TEMPLATES%%', replacement: REPORT_TEMPLATES},
     ]);
   }
 }
--- a/lighthouse-core/report/v2/report-generator.js
+++ b/lighthouse-core/report/v2/report-generator.js
@@ -17,13 +17,12 @@
 'use strict';
 
 const fs = require('fs');
-const path = require('path');
 
-const REPORT_TEMPLATE = fs.readFileSync(path.join(__dirname, './report-template.html'), 'utf8');
+const REPORT_TEMPLATE = fs.readFileSync(__dirname + '/report-template.html', 'utf8');
 // TODO: Setup a gulp pipeline to concat and minify the renderer files?
-const REPORT_JAVASCRIPT = fs.readFileSync(path.join(__dirname, './report-renderer.js'), 'utf8');
-const REPORT_CSS = fs.readFileSync(path.join(__dirname, './report-styles.css'), 'utf8');
-const REPORT_TEMPLATES = fs.readFileSync(path.join(__dirname, './templates.html'), 'utf8');
+const REPORT_JAVASCRIPT = fs.readFileSync(__dirname + '/report-renderer.js', 'utf8');
+const REPORT_CSS = fs.readFileSync(__dirname + '/report-styles.css', 'utf8');
+const REPORT_TEMPLATES = fs.readFileSync(__dirname + '/templates.html', 'utf8');
 
 class ReportGeneratorV2 {
   /**
--- a/lighthouse-core/report/v2/report-generator.js
+++ b/lighthouse-core/report/v2/report-generator.js
@@ -23,6 +23,7 @@ const REPORT_TEMPLATE = fs.readFileSync(path.join(__dirname, './report-template.
 // TODO: Setup a gulp pipeline to concat and minify the renderer files?
 const REPORT_JAVASCRIPT = fs.readFileSync(path.join(__dirname, './report-renderer.js'), 'utf8');
 const REPORT_CSS = fs.readFileSync(path.join(__dirname, './report-styles.css'), 'utf8');
+const REPORT_TEMPLATES = fs.readFileSync(path.join(__dirname, './templates.html'), 'utf8');
 
 class ReportGeneratorV2 {
   /**
@@ -97,11 +98,14 @@ class ReportGeneratorV2 {
   generateReportHtml(reportAsJson) {
     const sanitizedJson = JSON.stringify(reportAsJson).replace(/</g, '\\u003c');
     const sanitizedJavascript = REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/');
+    // Remove script in templates files just to be safe.
+    const sanitizedTemplates = REPORT_TEMPLATES.replace(/<script>([\s\S]*?)<\/script>/g, '');
 
     return ReportGeneratorV2.replaceStrings(REPORT_TEMPLATE, [
       {search: '%%LIGHTHOUSE_JSON%%', replacement: sanitizedJson},
       {search: '%%LIGHTHOUSE_JAVASCRIPT%%', replacement: sanitizedJavascript},
       {search: '/*%%LIGHTHOUSE_CSS%%*/', replacement: REPORT_CSS},
+      {search: '%%LIGHTHOUSE_TEMPLATES%%', replacement: sanitizedTemplates},
     ]);
   }
 }
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -105,7 +105,7 @@ class ReportRenderer {
   }
 
   /**
-   * @param {!DocumentFragment|Element} element DOM node to populate with values.
+   * @param {!DocumentFragment|!Element} element DOM node to populate with values.
    * @param {number} score
    * @param {string} scoringMode
    * @param {string} title
@@ -141,6 +141,13 @@ class ReportRenderer {
     }
     if (audit.result.optimalValue) {
       title += ` (target: ${audit.result.optimalValue})`;
+    }
+
+    // Append audit details to header section so the entire audit is within a <details>.
+    const header = tmpl.querySelector('.lighthouse-score__header');
+    header.open = audit.score < 100; // expand failed audits
+    if (audit.result.details) {
+      header.appendChild(this._renderDetails(audit.result.details));
     }
 
     return this._populateScore(tmpl, audit.score, scoringMode, title, description);
@@ -257,14 +264,6 @@ class ReportRenderer {
   _renderAudit(audit) {
     const element = this._createElement('div', 'lighthouse-audit');
     element.appendChild(this._renderAuditScore(audit));
-
-    // Append audit details to to header section so the entire audit is within one <details>.
-    const header = element.querySelector('.lighthouse-score__header');
-    header.open = audit.score < 100; // expand failed audits
-    if (audit.result.details) {
-      header.appendChild(this._renderDetails(audit.result.details));
-    }
-
     return element;
   }
 }
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -23,9 +23,7 @@
  */
 
 /* eslint-env browser */
-/* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */
 
-(function(self) {
 const RATINGS = {
   PASS: {label: 'pass', minScore: 75},
   AVERAGE: {label: 'average', minScore: 45},
@@ -144,7 +142,7 @@ class ReportRenderer {
   }
 
   /**
-   * @param {!Element} element DOM node to populate with values.
+   * @param {!DocumentFragment|Element} element DOM node to populate with values.
    * @param {number} score
    * @param {string} scoringMode
    * @param {string} title
@@ -311,9 +309,8 @@ class ReportRenderer {
 }
 
 // Exports
-self.ReportRenderer = ReportRenderer;
-self.DOM = DOM;
-})(self);
+window.ReportRenderer = ReportRenderer;
+window.DOM = DOM;
 
 /** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */
 let DetailsJSON; // eslint-disable-line no-unused-vars
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -25,7 +25,7 @@
 /* eslint-env browser */
 /* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */
 
-(function() {
+(function(self) {
 const RATINGS = {
   PASS: {label: 'pass', minScore: 75},
   AVERAGE: {label: 'average', minScore: 45},
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -15,56 +15,128 @@
  */
 'use strict';
 
-/* eslint-env browser */
-/* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */
-
-(function() {
-const RATINGS = {
-  GOOD: {label: 'good', minScore: 75},
-  AVERAGE: {label: 'average', minScore: 45},
-  POOR: {label: 'poor'}
-};
-
-/**
- * Convert a score to a rating label.
- * @param {number} value
- * @return {string}
- */
-function calculateRating(value) {
-  if (typeof value === 'boolean') {
-    return value ? RATINGS.GOOD.label : RATINGS.POOR.label;
-  }
-
-  let rating = RATINGS.POOR.label;
-  if (value >= RATINGS.GOOD.minScore) {
-    rating = RATINGS.GOOD.label;
-  } else if (value >= RATINGS.AVERAGE.minScore) {
-    rating = RATINGS.AVERAGE.label;
-  }
-  return rating;
-}
-
-/**
- * Format number.
- * @param {number} number
- * @return {string}
- */
-function formatNumber(number) {
-  return number.toLocaleString(undefined, {maximumFractionDigits: 1});
-}
-
 /**
  * @fileoverview The entry point for rendering the Lighthouse report based on the JSON output.
  *    This file is injected into the report HTML along with the JSON report.
  *
  * Dummy text for ensuring report robustness: </script> pre$`post %%LIGHTHOUSE_JSON%%
  */
-window.ReportRenderer = class ReportRenderer {
+
+/* eslint-env browser */
+/* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */
+
+(function() {
+const RATINGS = {
+  PASS: {label: 'pass', minScore: 75},
+  AVERAGE: {label: 'average', minScore: 45},
+  FAIL: {label: 'fail'}
+};
+
+/**
+ * Convert a score to a rating label.
+ * @param {!number} score
+ * @param {!string} precision One of 'binary', 'numeric', 'grade'.
+ * @return {!string}
+ */
+function calculateRating(score, precision) {
+  if (precision === 'numeric') {
+    let rating = RATINGS.FAIL.label;
+    if (score >= RATINGS.PASS.minScore) {
+      rating = RATINGS.PASS.label;
+    } else if (score >= RATINGS.AVERAGE.minScore) {
+      rating = RATINGS.AVERAGE.label;
+    }
+    return rating;
+  } else if (precision === 'grade') {
+    // Not implemented yet.
+  }
+
+  // Treat as binary by default.
+  return score === 100 ? RATINGS.PASS.label : RATINGS.FAIL.label;
+}
+
+/**
+ * Format number.
+ * @param {!number} number
+ * @return {!string}
+ */
+function formatNumber(number) {
+  return number.toLocaleString(undefined, {maximumFractionDigits: 1});
+}
+
+class DOM {
   /**
    * @param {!Document} document
    */
   constructor(document) {
     this._document = document;
+  }
+
+  /**
+   * Sets the text content of an element, safely. If the element does not exist,
+   * results in a noop.
+   * @param {Element} element
+   * @param {!string} text
+   * @return {Element}
+   */
+  static setText(element, text) {
+    if (element) {
+      element.textContent = text;
+    }
+    return element;
+  }
+
+  /**
+   * Adds a class to an element, safely. If the element does not exist, results
+   * in a noop.
+   * @param {Element} element
+   * @param {...!string} classes
+   * @return {Element}
+   */
+  static addClass(element, ...classes) {
+    if (element) {
+      element.classList.add(...classes);
+    }
+    return element;
+  }
+
+  /**
+   * @param {string} name
+   * @param {string=} className
+   * @param {!Object<string, string>=} attrs Attribute key/val pairs.
+   * @return {!Element}
+   */
+  createElement(name, className, attrs = {}) {
+    const element = this._document.createElement(name);
+    if (className) {
+      element.className = className;
+    }
+    for (const [key, val] of Object.entries(attrs)) {
+      element.setAttribute(key, val);
+    }
+    return element;
+  }
+
+  /**
+   * @param {!string} selector
+   * @return {!DocumentFragment} A clone of the template content.
+   * @throws {Error}
+   */
+  cloneTemplate(selector) {
+    const template = this._document.querySelector(selector);
+    if (!template) {
+      throw new Error(`Template not found: template${selector}`);
+    }
+    return this._document.importNode(template.content, true);
+  }
+}
+
+class ReportRenderer {
+  /**
+   * @param {!Document} document
+   */
+  constructor(document) {
+    this._dom = new DOM(document);
   }
 
   /**
@@ -80,64 +152,47 @@ window.ReportRenderer = class ReportRenderer {
   }
 
   /**
-   * @param {string} name
-   * @param {string=} className
-   * @param {!Object<string, string>=} attrs Attribute key/val pairs.
+   * @param {!Element} tmpl Parent node to populate with values.
+   * @param {!{value: string, precision: string}} score
+   * @param {!string} title
+   * @param {!string} description
    * @return {!Element}
    */
-  _createElement(name, className, attrs = {}) {
-    const element = this._document.createElement(name);
-    if (className) {
-      element.className = className;
-    }
-    for (const [key, val] of Object.entries(attrs)) {
-      element.setAttribute(key, val);
-    }
-    return element;
+  _populateScore(tmpl, score, title, description) {
+    const rating = calculateRating(score.value, score.precision);
+
+    // Fill in the blanks.
+    const value = tmpl.querySelector('.lighthouse-score__value');
+    DOM.setText(value, formatNumber(score.value));
+    DOM.addClass(value, `lighthouse-score__value--${rating}`,
+                        `lighthouse-score__value--${score.precision}`);
+
+    DOM.setText(tmpl.querySelector('.lighthouse-score__title'), title);
+    DOM.setText(tmpl.querySelector('.lighthouse-score__description'), description);
+
+    return tmpl;
   }
 
   /**
-   * @param {number} score
-   * @param {string} title
-   * @param {string} description
+   * @param {!{value: string, precision: string}} score
+   * @param {!string} title
+   * @param {!string} description
    * @return {!Element}
    */
-  _renderScore(score, title, description, auditOrCategory) {
-    const isAudit = 'result' in auditOrCategory;
-    const tagName = isAudit ? 'details' : 'div';
+  _renderAuditScore(score, title, description) {
+    const tmpl = this._dom.cloneTemplate('#tmpl-lighthouse-audit-score');
+    return this._populateScore(tmpl, score, title, description);
+  }
 
-    // Score template. Audits get a details dropdown. Categories don't.
-    const tmpContainer = this._createElement('div');
-    tmpContainer.innerHTML = `<div class="lighthouse-score">
-      <div class="lighthouse-score__value"><!-- filled --></div>
-      <${tagName} class="lighthouse-score__header">
-        <summary class="lighthouse-score__snippet">
-          <span class="lighthouse-score__title"><!-- filled --></span>
-          <div class="lighthouse-score__arrow" title="See audits"></div>
-        </summary>
-        <div class="lighthouse-score__description"><!-- filled --></div>
-      </${tagName}>
-    </div>`;
-
-    // Fill in the blanks.
-    const value = tmpContainer.querySelector('.lighthouse-score__value');
-    value.textContent = formatNumber(score);
-    value.classList.add(`lighthouse-score__value--${calculateRating(score)}`);
-
-    if (typeof score === 'boolean') {
-      value.classList.add('lighthouse-score__value--boolean');
-    }
-
-    tmpContainer.querySelector('.lighthouse-score__title').textContent = title;
-    tmpContainer.querySelector('.lighthouse-score__description').textContent = description;
-
-    const header = tmpContainer.querySelector('.lighthouse-score__header');
-    header.open = score < 1;
-    if (isAudit && auditOrCategory.result.details) {
-      header.appendChild(this._renderDetails(auditOrCategory.result.details));
-    }
-
-    return tmpContainer.firstChild;
+  /**
+   * @param {!{value: string, precision: string}} score
+   * @param {!string} title
+   * @param {!string} description
+   * @return {!Element}
+   */
+  _renderCategoryScore(score, title, description) {
+    const tmpl = this._dom.cloneTemplate('#tmpl-lighthouse-category-score');
+    return this._populateScore(tmpl, score, title, description);
   }
 
   /**
@@ -162,7 +217,7 @@ window.ReportRenderer = class ReportRenderer {
    * @return {!Element}
    */
   _renderText(text) {
-    const element = this._createElement('div', 'lighthouse-text');
+    const element = this._dom.createElement('div', 'lighthouse-text');
     element.textContent = text.text;
     return element;
   }
@@ -172,7 +227,7 @@ window.ReportRenderer = class ReportRenderer {
    * @return {!Element}
    */
   _renderBlock(block) {
-    const element = this._createElement('div', 'lighthouse-block');
+    const element = this._dom.createElement('div', 'lighthouse-block');
     for (const item of block.items) {
       element.appendChild(this._renderDetails(item));
     }
@@ -184,14 +239,14 @@ window.ReportRenderer = class ReportRenderer {
    * @return {!Element}
    */
   _renderList(list) {
-    const element = this._createElement('details', 'lighthouse-list');
+    const element = this._dom.createElement('details', 'lighthouse-list');
     if (list.header) {
-      const summary = this._createElement('summary', 'lighthouse-list__header');
+      const summary = this._dom.createElement('summary', 'lighthouse-list__header');
       summary.textContent = list.header.text;
       element.appendChild(summary);
     }
 
-    const items = this._createElement('div', 'lighthouse-list__items');
+    const items = this._dom.createElement('div', 'lighthouse-list__items');
     for (const item of list.items) {
       items.appendChild(this._renderDetails(item));
     }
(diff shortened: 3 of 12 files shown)


lighthouse-core/report/v2/report-renderer.js
```diff
@@ -98,19 +102,42 @@ window.ReportRenderer = class ReportRenderer {
    * @param {string} description
    * @return {!Element}
    */
-  _renderScore(score, title, description) {
-    const element = this._createElement('div', 'lighthouse-score');
+  _renderScore(score, title, description, auditOrCategory) {
+    const isAudit = 'result' in auditOrCategory;
+    const tagName = isAudit ? 'details' : 'div';
+
+    // Score template. Audits get a details dropdown. Categories don't.
```
ebidel: if we wanted to, we could make the whole category collapsable. For now, I decide to keep it as v1.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -215,12 +243,15 @@ window.ReportRenderer = class ReportRenderer {
...
     }
-    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
-    if (audit.result.details) {
-      element.appendChild(this._renderDetails(audit.result.details));
+    if (audit.result.optimalValue) {
+      title += ` (target: ${audit.result.optimalValue})`;
     }
+
+    element.appendChild(
+        this._renderScore(audit.result.score, title, audit.result.helpText, audit));
```
patrickhulce: why can't we stick to `audit.score`?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -31,6 +31,10 @@ const RATINGS = {
  * @return {string}
  */
 function calculateRating(value) {
+  if (typeof value === 'boolean') {
```
patrickhulce: :( can we just keep that logic in the report generator?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -98,19 +102,42 @@ window.ReportRenderer = class ReportRenderer {
...
    */
-  _renderScore(score, title, description) {
-    const element = this._createElement('div', 'lighthouse-score');
+  _renderScore(score, title, description, auditOrCategory) {
+    const isAudit = 'result' in auditOrCategory;
+    const tagName = isAudit ? 'details' : 'div';
+
+    // Score template. Audits get a details dropdown. Categories don't.
+    const tmpContainer = this._createElement('div');
+    tmpContainer.innerHTML = `<div class="lighthouse-score">
```
patrickhulce: 😱

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -98,19 +102,42 @@ window.ReportRenderer = class ReportRenderer {
    * @param {string} description
    * @return {!Element}
    */
-  _renderScore(score, title, description) {
-    const element = this._createElement('div', 'lighthouse-score');
+  _renderScore(score, title, description, auditOrCategory) {
```
patrickhulce: the doctype lies! since we're shifting away from displaying high precision scores is this still necessary? what about a `precision: binary|grade|numeric` property or something for each audit and we can say goodbye to `typeof`

lighthouse-core/report/v2/report-renderer.js
ebidel: That doesn't have boolean information. Booleans come back `score: 0`

lighthouse-core/report/v2/report-renderer.js
ebidel: Don't be scared. Completely safe since we control all inputs for the initial set.

lighthouse-core/report/v2/report-renderer.js
patrickhulce: Yeah moved discussion to https://github.com/GoogleChrome/lighthouse/pull/1963#discussion_r109763756 :)

lighthouse-core/report/v2/report-renderer.js
ebidel: Score could also be an object with value/precision props. Or maybe`type` for the latter. Examples:

```
category.score = {value: 100, precision: 'numeric'}
audit.score = {value: true, precision: 'binary'}
audit.score = {value: 'A', precision: 'grade'}
```

brendankenny: can someone sum up what was decided? My thoughts:

- audits are going to score in some fashion, whether that's 0-100, bucketed into grades, or pass/fail. Even if we require all audits to report a score as a number, it's still going to be an integer in that interval, buckets of every 10ish numbers, or 0/100. It makes some sense then for the audit to declare how it scored (perhaps on audit.meta), even if the reporter changes how the score is presented.

- If on the other hand (or in addition to the above), we allow specifying the precision type in the config (on one of the audits in `categories.categoryName.audits`) it seems like we'd need some way of converting the score reported by the audit into the new format. `binary` -> `numeric` is easy, but (as discussed), what's the true/false cutoff for `numeric` -> `binary` and where do you specify it?

patrickhulce: Here's my take on what was discussed

Immediate actions: 
* add a `scorePrecision` or similarly named property to the performance audits `meta` object specifying a `numeric` precision
* add default `scorePrecision` of `binary` to the result
* update ReportRenderer to use the `audit.result.scorePrecision` combined with `audit.score` to decide how to display (no more `typeof score === 'boolean'` outside of ReportGenerator)

Potential future course of action that need further discussion: We remove `score` from the audit result object and instead have them return values that are compared to the declared target/point of diminishing returns/median values and a score is generated for them, boolean audits return whether they pass or fail as they do today and a score of 100/0 is generated for them. We explore the possibility of introducing letter grades and changing the score curves.

brendankenny: > add a `scorePrecision` or similarly named property to the performance audits meta object specifying a numeric precision

should be noted that is basically just giving `expectedValue` a new name :)

It will be nice to always have a numeric score, though.

patrickhulce: if you mean renaming a property to align with it's meaning and changing the types of values it can hold then yes it's basically that :)

brendankenny: an enum is an enum :P

ebidel: @patrickhulce do you want to craft those changes in a PR? I am working on some other stuff with `<template>` and can rebase when that lands.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -93,24 +116,73 @@ window.ReportRenderer = class ReportRenderer {
...
+  /**
+   * @param {!number} score
+   * @param {!string} title
+   * @param {!string} description
+   * @param {!(AuditJSON|CategoryJSON)} auditOrCategory
+   * @return {!Element}
+   */
+  _renderScore(score, title, description, auditOrCategory) {
+    const isAudit = 'result' in auditOrCategory;
+    const precision = isAudit ? 'binary' : 'category'; // TODO(ericbidelman): placeholder
```
ebidel: fyi @patrickhulce

lighthouse-core/report/v2/report-renderer.js
patrickhulce: @ebidel sure thing

patrickhulce: #1967

ebidel:
Removed `.innerHTML`! 

It's been replaced by templates.html which includes declared html partials (using `<template>`). It will also be the future home for ported handlebars formatters and other UI scaffolding. Templates get added to `report-template.html` using our injection technique. It's unclear how devtools will bring in templates.html.

This PR also introduces a `DOM` class to organize DOM-related code. It includes wrapper methods for setting `.textContent` and `.classList` so we never run into errors with querySelector'd returning `null`. Plan to add some renderer tests.

The `typeof === 'boolean'` stuff is also removed. I've preemptively hardcoded `precision` placeholders for audit and cateogry scores. For now, `_renderAuditScore` and `_renderCategoryScore` construct an object: `score = {value: audit.score, precision: 'numeric'}`. The expectation is that `report-generator.js` will setup this object when it creates results and we can remove those lines.

PTAL.

ebidel:
Now with `scoringMode`. PTAL

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -93,24 +116,81 @@ window.ReportRenderer = class ReportRenderer {
...
+   */
+  _populateScore(tmpl, score, title, description) {
+    const rating = calculateRating(score.value, score.scoringMode);
 
-    const column = element.appendChild(this._createElement('div', 'lighthouse-score__text'));
-    column.appendChild(this._createElement('div', 'lighthouse-score__title')).textContent = title;
-    column.appendChild(
-      this._createElement('div', 'lighthouse-score__description')).textContent = description;
+    // Fill in the blanks.
+    const value = tmpl.querySelector('.lighthouse-score__value');
```
paulirish: s/value/elem/

ebidel: when for `valueEl` so disambiguate.

ebidel:
Added tests. I'm done.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -64,15 +71,31 @@ window.ReportRenderer = class ReportRenderer {
...
+   * @param {!string} text
+   * @return {Element}
    */
-  renderReport(report) {
-    try {
-      return this._renderReport(report);
-    } catch (e) {
-      return this._renderException(e);
+  static setText(element, text) {
+    if (element) {
```
brendankenny: if we're worried about these, should they just throw? Seems like it's definitely an error since we're creating the templates *and* the selectors looking within them

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -199,8 +279,10 @@ window.ReportRenderer = class ReportRenderer {
    * @return {!Element}
    */
   _renderCategory(category) {
-    const element = this._createElement('div', 'lighthouse-category');
-    element.appendChild(this._renderScore(category.score, category.name, category.description));
+    const element = this._dom.createElement('div', 'lighthouse-category');
+    const score = {value: Math.round(category.score), scoringMode: 'numeric'};
+    element.appendChild(
```
brendankenny: `_renderCategoryScore` only takes three arguments

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -212,19 +294,35 @@ window.ReportRenderer = class ReportRenderer {
...
+    }
+    if (audit.result.optimalValue) {
+      title += ` (target: ${audit.result.optimalValue})`;
     }
-    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
-    if (audit.result.details) {
-      element.appendChild(this._renderDetails(audit.result.details));
+
+    const score = {value: audit.score, scoringMode: audit.result.scoringMode};
+    element.appendChild(this._renderAuditScore(score, title, audit.result.helpText, audit));
```
brendankenny: also doesn't need the fourth argument

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,36 +104,93 @@ window.ReportRenderer = class ReportRenderer {
...
-    value.textContent = formatNumber(score);
-    value.classList.add(`lighthouse-score__value--${calculateRating(score)}`);
+  /**
+   * @param {!Element} tmpl Parent node to populate with values.
+   * @param {!{value: string, scoringMode: string}} score
+   * @param {!string} title
+   * @param {!string} description
+   * @return {!Element}
+   */
+  _populateScore(tmpl, score, title, description) {
```
brendankenny: should this (and the render functions below) be called something like `_populateResult` instead of `_populateScore` since they include the score, title, and description of the audit/category?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -199,8 +279,10 @@ window.ReportRenderer = class ReportRenderer {
    * @return {!Element}
    */
   _renderCategory(category) {
-    const element = this._createElement('div', 'lighthouse-category');
-    element.appendChild(this._renderScore(category.score, category.name, category.description));
+    const element = this._dom.createElement('div', 'lighthouse-category');
+    const score = {value: Math.round(category.score), scoringMode: 'numeric'};
```
brendankenny: any reason to have this as an object? Seems fine as separate arguments to `renderCategoryScore`/`renderAuditScore` (and you may not even need `scoringMode` in there, at least not yet)

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/* eslint-env mocha */
+/* global document self */
```
brendankenny: `/* eslint-env mocha, browser */`?

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+
+const assert = require('assert');
+const fs = require('fs');
+const path = require('path');
+const jsdom = require('jsdom');
+const sampleResults = require('../../results/sample_v2.json');
+
+const TEMPLATES_FILE = fs.readFileSync(
+    path.join(__dirname, '../../../report/v2/templates.html'), 'utf8');
```
brendankenny: don't need to joinnnn

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+
+function setupJsDomGlobals() {
+  global.document = jsdom.jsdom(TEMPLATES_FILE);
```
brendankenny: shouldn't this test in the full context of the `report-template.html` page (with assets inlined)?

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+  global.document = undefined;
+  global.window = undefined;
+  global.self = undefined;
+}
+
+setupJsDomGlobals(); // Run jsdom setup code before report renderer is executed.
+require('../../../report/v2/report-renderer.js');
+
+describe('DOM', () => {
+  const window = self;
```
brendankenny: `window` appears unused?

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+  describe('setText', () => {
+    it('sets text of element', () => {
+      const el = document.createElement('div');
+      const result = DOM.setText(el, 'hello world');
+      assert.deepStrictEqual(result, el);
+      assert.equal(result.textContent, 'hello world');
+    });
+
+    it('does not fail when given null element argument', () => {
+      assert.doesNotThrow(() => {
```
brendankenny: can drop the surrounding `assert.doesNotThrow` and just execute it. If it throws the test will fail

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+  describe('addClass', () => {
+    it('adds classes to element', () => {
+      const el = document.createElement('div');
+      const result = DOM.addClass(el, 'class1', 'class2');
+      assert.ok(result.classList.contains, 'class1');
+      assert.ok(result.classList.contains, 'class2');
+    });
+
+    it('does not fail when given null element argument', () => {
+      assert.doesNotThrow(() => {
```
brendankenny: don't need `doesNotThrow` wrapper

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+    it('should clone a template', () => {
+      const dom = new DOM(document);
+      const clone = dom.cloneTemplate('#tmpl-lighthouse-audit-score');
+      assert.ok(clone.querySelector('.lighthouse-score'));
+    });
+
+    it('fails when template cannot be found', () => {
+      const dom = new DOM(document);
+      assert.throws(() => {
+        const clone = dom.cloneTemplate('#unknown-selector'); // eslint-disable-line no-unused-vars
```
brendankenny: can you just not assign to a var to lose the eslint directive?

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+      });
+    });
+  });
+});
+
+describe('ReportRenderer V2', () => {
+  const window = self;
```
brendankenny: isn't the global `window` already set to `self`? Or...something? The globals are confusing here (and only used to access `ReportRenderer`)

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -0,0 +1,177 @@
...
+
+    it('should render a report', () => {
+      const renderer = new ReportRenderer(document);
+      const output = renderer.renderReport(sampleResults);
+      assert.ok(output.classList.contains('lighthouse-report'));
+    });
+
+    it('should render an exception for invalid input', () => {
+      const renderer = new ReportRenderer(document);
+      const output = renderer.renderReport({});
```
brendankenny: Since we're not really validating anything right now this relies on implementation details. What about something explicit like
```js
const output = renderer.renderReport({
  get reportCategories() { throw new Error() }
});
```

brendankenny:
> there's stuff i could bikeshed.. like the DOM library API and how the templates are added to the page. But on the whole I'm happy here. Let's keep it moving.

If it really is just bikeshedding stuff that's fine, but this is the kind of thing that gets written once and we're stuck with from then on unless it bothers you so much in the future you file a bug/change it.

Now's the time to bring it up, even if we get to it in a followup :)

lighthouse-core/report/v2/report-renderer.js
brendankenny: just kidding, you do need `scoringMode` for the class name, but still seems like they should be separate parameters :)

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -15,47 +15,54 @@
...
 }
 
-/**
- * @fileoverview The entry point for rendering the Lighthouse report based on the JSON output.
- *    This file is injected into the report HTML along with the JSON report.
- *
- * Dummy text for ensuring report robustness: </script> pre$`post %%LIGHTHOUSE_JSON%%
- */
-window.ReportRenderer = class ReportRenderer {
+class DOM {
```
brendankenny: feels like something I know ahead of time you're going to disagree with me on :) but I don't personally see the need for a separate `DOM` class (yet, at least), especially because the split with `ReportRenderer` doesn't seem completely clear. Anything wrong with these living there?

ebidel: I knew you were going to bring this up! I wasn't convinced adding a bunch of generic methods to ReportRenderer made sense either. They could live as standalone helper functions, but devtools may need to extend them later.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -97,11 +98,14 @@ class ReportGeneratorV2 {
   generateReportHtml(reportAsJson) {
     const sanitizedJson = JSON.stringify(reportAsJson).replace(/</g, '\\u003c');
     const sanitizedJavascript = REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/');
+    // Remove script in templates files just to be safe.
+    const sanitizedTemplates = REPORT_TEMPLATES.replace(/<script>([\s\S]*?)<\/script>/g, '');
```
ebidel: I agree with you. I recall that `el.innerHTML = '<div>just a string</div>';` was deemed unsafe and that's our own markup :) The argument from _someonnnnne_ was that we may add things later and not realize the mistakes.

lighthouse-core/report/v2/report-renderer.js
ebidel: @paulirish asked for them so that the fails gracefully if a node cant be found. My pref would be to remove them altogether. Tests should really be catching missing nodes in templates.html (markup we control). It would also get rid of the `DOM` and simplify render methods a bit if we just set things directly (`.classList`, `.textContent`).

I'm also ok keeping.

lighthouse-core/report/v2/report-generator.js
```diff
@@ -23,6 +23,7 @@ const REPORT_TEMPLATE = fs.readFileSync(path.join(__dirname, './report-template.
 // TODO: Setup a gulp pipeline to concat and minify the renderer files?
 const REPORT_JAVASCRIPT = fs.readFileSync(path.join(__dirname, './report-renderer.js'), 'utf8');
 const REPORT_CSS = fs.readFileSync(path.join(__dirname, './report-styles.css'), 'utf8');
+const REPORT_TEMPLATES = fs.readFileSync(path.join(__dirname, './templates.html'), 'utf8');
```
patrickhulce: does that work just fine on windows?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,36 +104,93 @@ window.ReportRenderer = class ReportRenderer {
...
+    });
     return element;
   }
 
   /**
-   * @param {number} score
-   * @param {string} title
-   * @param {string} description
+   * @param {!string} selector
+   * @return {!DocumentFragment} A clone of the template content.
```
ebidel: Technically a doc frag, which is not a full element. We use it like an `!Element` though.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -15,47 +15,54 @@
...
   AVERAGE: {label: 'average', minScore: 45},
-  POOR: {label: 'poor'}
+  FAIL: {label: 'fail'}
 };
 
 /**
  * Convert a score to a rating label.
- * @param {number} value
- * @return {string}
+ * @param {!number} score
```
patrickhulce: idk though @ebidel the great `Object<string, string>` debacle haunts us throughout our codebase now after listenin' to this guy 😉

docs/hacking-tips.md
```diff
@@ -13,9 +13,18 @@ Use [`--trace-warnings`](https://medium.com/@jasnell/introducing-process-warning
...
 ```
 
+## Updating fixture dumps
+
+`lighthouse-core/test/results/samples_v2.json` is generated from running LH against
+dbw_tester.html. To update this file:
+
+```sh
+npm run start -- --output=json --output-path=lighthouse-core/test/results/sample_v2.json http://localhost:8080/dobetterweb/dbw_tester.html
+```
```
ebidel: done

lighthouse-core/report/v2/report-generator.js
ebidel: done

lighthouse-core/report/v2/report-renderer.js
ebidel: good to know. removed.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -15,47 +15,54 @@
...
+    let rating = RATINGS.FAIL.label;
+    if (score >= RATINGS.PASS.minScore) {
+      rating = RATINGS.PASS.label;
+    } else if (score >= RATINGS.AVERAGE.minScore) {
+      rating = RATINGS.AVERAGE.label;
+    }
+    return rating;
+  }
+
+  // Treat as binary by default.
```
ebidel: Right, good call. Left over from before. Removed.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -64,15 +71,31 @@ window.ReportRenderer = class ReportRenderer {
   }
 
   /**
-   * @param {!ReportJSON} report
-   * @return {!Element}
+   * Sets the text content of an element, safely. If the element does not exist,
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -64,15 +71,31 @@ window.ReportRenderer = class ReportRenderer {
...
+      element.textContent = text;
+    }
+    return element;
+  }
+
+  /**
+   * Adds a class to an element, safely. If the element does not exist, results
+   * in a noop.
+   * @param {Element} element
+   * @param {...!string} classes
```

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,36 +104,93 @@ window.ReportRenderer = class ReportRenderer {
...
-  _createElement(name, className, attrs = {}) {
+  createElement(name, className, attrs = {}) {
     const element = this._document.createElement(name);
     if (className) {
       element.className = className;
     }
-    for (const [key, val] of Object.entries(attrs)) {
-      element.setAttribute(key, val);
-    }
+    Object.keys(attrs).forEach(key => {
```
ebidel: +1

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,36 +104,93 @@ window.ReportRenderer = class ReportRenderer {
...
 
-    const column = element.appendChild(this._createElement('div', 'lighthouse-score__text'));
-    column.appendChild(this._createElement('div', 'lighthouse-score__title')).textContent = title;
-    column.appendChild(
-      this._createElement('div', 'lighthouse-score__description')).textContent = description;
+    // Fill in the blanks.
+    const valueEl = tmpl.querySelector('.lighthouse-score__value');
+    DOM.setText(valueEl, formatNumber(score.value));
+    DOM.addClass(valueEl, `lighthouse-score__value--${rating}`,
+                          `lighthouse-score__value--${score.scoringMode}`);
```

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: done

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: done

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: I kind of like being explicit.

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: sloppy...but done!

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: done

lighthouse-core/report/v2/report-generator.js
brendankenny: > does that work just fine on windows?

Yeah, Windows understands / and the `fs` methods will take care of the rest. The only real worry (that I know of) is if you want to specify an absolute path from the root, which is different cross platform, but we're using `__dirname` so it's not an issue here

lighthouse-core/report/v2/report-renderer.js
brendankenny: But it isn't just grabbing the `content`, it also calls `document.importNode` which appears to return a Node.

lighthouse-core/report/v2/report-generator.js
brendankenny: I think this is different enough, though, because it's not markup assembled in (potentially) multiple places, it's all in one spot. So I think it's better on that front. I also don't like calling it sanitized unless it is actually sanitized :)

lighthouse-core/report/v2/report-renderer.js
brendankenny: well, TECHNICALLY, `Object<string>` still works, Closure just won't type check it :)

lighthouse-core/report/v2/report-renderer.js
brendankenny: commit pushed yet?

brendankenny: jk you haven't pushed yet :)

lighthouse-core/report/v2/report-renderer.js
brendankenny: > @paulirish asked for them so that the report fails gracefully if a node cant be found.

I think @paulirish was asking before we had completely settled on the approach used here, so it may be less of an issue. Queries failing fast (as soon as we try to set `textContent` or add to the `classList`) is a good thing, I think. I also like the idea of moving some of the query verification to tests, but if setting `element.textContent` throws a TypeError, just testing e.g. `_renderAuditScore` will verify the relevant elements exist if it doesn't throw :)

lighthouse-core/report/v2/report-renderer.js
ebidel: TBH, it's a terrible API and name. `importNode()` returns the node you import. Since we're importing a `DocumentFragment`, we get back a `Node` of that type. 

screen shot 2017-04-06 at 12 35 37 pm

lighthouse-core/report/v2/report-renderer.js
ebidel: Just pushed the first set of changes. Updating tests to come.

lighthouse-core/report/v2/report-renderer.js
brendankenny: > `importNode()` returns the node you import

ah, I see. Should `_populateResult` take a `!DocumentFragment`, then?

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: Right. I was just testing template output and the render methods in this file, not how they fit into the entire app. Using the whole page would need `ReportGeneratorV2`, which seems more end to end.

What I've done: added are more tests to report-generator-test.js so we know all the templates get injected.

ebidel:
PTAL. I think I got everything.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,39 +96,104 @@ window.ReportRenderer = class ReportRenderer {
...
   }
 
   /**
+   * @param {!AuditJSON} audit
+   * @return {!Element}
+   */
+  _renderAuditScore(audit) {
+    const tmpl = this._dom.cloneTemplate('#tmpl-lighthouse-audit-score');
+
+    const scoringMode = audit.result.scoringMode;
```
brendankenny: How we've gotten here has made sense, but it is weird that `scoringMode` is on `audit.result` but the score is on `audit` directly

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,39 +96,104 @@ window.ReportRenderer = class ReportRenderer {
...
+  renderReport(report) {
+    try {
+      return this._renderReport(report);
+    } catch (e) {
+      return this._renderException(e);
+    }
+  }
+
+  /**
+   * @param {!DocumentFragment|Element} element DOM node to populate with values.
```
brendankenny: this is only ever called with a `DocumentFragment`, though

lighthouse-core/report/v2/report-generator.js
brendankenny: really think we should remove this, or call it "templatesWithScriptTagsRemoves" or something. Current version is easily worked around with an attribute on the script tag...I don't think this is an arms race we want to even start.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -15,26 +15,31 @@
...
 /* eslint-env browser */
-/* eslint indent: [2, 2, { "SwitchCase": 1, "outerIIFEBody": 0 }] */
 
-(function() {
 const RATINGS = {
-  GOOD: {label: 'good', minScore: 75},
+  PASS: {label: 'pass', minScore: 75},
   AVERAGE: {label: 'average', minScore: 45},
-  POOR: {label: 'poor'}
+  FAIL: {label: 'fail'}
```
brendankenny: does PASS/FAIL work here? Seems weird to have a three state pass/fail system :)

lighthouse-core/report/v2/report-renderer.js
brendankenny: still think we should remove. Earlier (and non-silent) failures are better failures.

ebidel: @paulirish says, "I am low stress on this. Do whateva :)"

Note to future me: give PI 24hrs and he will be low stress.

lighthouse-core/report/v2/report-generator.js
ebidel: done

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js
ebidel: removed

lighthouse-core/report/v2/report-renderer.js
ebidel: Not really a pass/fail system. Just for labeling/coloring right now. We'll probably need to expand in the future if we support things like grades. Seems fine for now.

lighthouse-core/report/v2/report-renderer.js
ebidel: yep, but it support both

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -81,39 +96,104 @@ window.ReportRenderer = class ReportRenderer {
...
+    }
+
+    return this._populateResult(tmpl, audit.score, scoringMode, title, description);
+  }
+
+  /**
+   * @param {!CategoryJSON} category
+   * @return {!Element}
+   */
+  _renderCategoryScore(category) {
```
ebidel: I'm actually going to change "_populateResult" back to `_populateScore`. It's populating the score dom with values.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -212,19 +292,25 @@ window.ReportRenderer = class ReportRenderer {
...
-    }
-    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
-    if (audit.result.details) {
-      element.appendChild(this._renderDetails(audit.result.details));
+    const element = this._dom.createElement('div', 'lighthouse-audit');
+    element.appendChild(this._renderAuditScore(audit));
+
+    // Append audit details to to header section so the entire audit is within one <details>.
+    const header = element.querySelector('.lighthouse-score__header');
+    if (header) {
```
ebidel: it was meant to protect against `null` querySelector results like methods. But removed since we don't care about that anymore.

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -212,19 +292,25 @@ window.ReportRenderer = class ReportRenderer {
...
+
     return element;
   }
-};
-})();
+}
+
+// Exports
+window.ReportRenderer = ReportRenderer;
+window.DOM = DOM;
```
ebidel: Mocha needs it, but I've switched to proper module exports for Node env.

lighthouse-core/report/v2/report-renderer.js
brendankenny: > Not really a pass/fail system. Just for labeling/coloring right now.

yeah, I was mostly just commenting on the names since PASS/FAIL is usually a binary state. Fine for now

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -86,34 +85,78 @@ window.ReportRenderer = class ReportRenderer {
...
+  _cloneTemplate(selector) {
+    const template = this._document.querySelector(selector);
+    if (!template) {
+      throw new Error(`Template not found: template${selector}`);
+    }
+    return this._document.importNode(template.content, true);
+  }
+
+  /**
+   * @param {!DocumentFragment|Element} element DOM node to populate with values.
```
brendankenny: `{!DocumentFragment|!Element}` then

lighthouse-core/report/v2/report-renderer.js
brendankenny: > I'm actually going to change `_populateResult` back to `_populateScore`. It's populating the score dom with values

my issue was that if someone was going into the codebase to see where audits' titles and descriptions are set, they probably wouldn't think it would be in `_populateScore`. If the issue is the template's id, that can change too :P

This is all going to change a lot in the near future, so if you feel strongly we can defer

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -213,18 +256,22 @@ window.ReportRenderer = class ReportRenderer {
...
     const element = this._createElement('div', 'lighthouse-audit');
-    let title = audit.result.description;
-    if (audit.result.displayValue) {
-      title += ': ' + audit.result.displayValue;
-    }
-    element.appendChild(this._renderScore(audit.score, title, audit.result.helpText));
+    element.appendChild(this._renderAuditScore(audit));
+
+    // Append audit details to to header section so the entire audit is within one <details>.
+    const header = element.querySelector('.lighthouse-score__header');
```
brendankenny: since this depends on markup introduced by `_renderAuditScore`, move header/details stuff in there?

lighthouse-core/report/v2/templates.html
```diff
@@ -0,0 +1,26 @@
+<!-- Lighthouse category score -->
+<template id="tmpl-lighthouse-category-score">
+  <div class="lighthouse-score">
+    <div class="lighthouse-score__value"><!-- fill me --></div>
+    <div class="lighthouse-score__header">
+      <div class="lighthouse-score__snippet">
+        <span class="lighthouse-score__title"><!-- fill me --></span>
+      </div>
+      <div class="lighthouse-score__description"><!-- fill me --></div>
```
brendankenny: worth putting in a comment(s) to show where audits (and details for the below template) will go?

lighthouse-core/test/report/v2/report-generator-test.js
```diff
@@ -143,6 +146,20 @@ describe('ReportGeneratorV2', () => {
       assert.ok(result.includes('LIGHTHOUSE_JAVASCRIPT'), 'cannot be tricked');
     });
 
+    it('should inject the report templates', () => {
+      const page = jsdom.jsdom(new ReportGeneratorV2().generateReportHtml({}));
+      const templates = jsdom.jsdom(TEMPLATES_FILE);
+
+      assert.deepEqual(page.querySelectorAll('template[id^="tmpl-"]'),
```
brendankenny: `deepStrictEqual`? Deep equality checking also gets weird with e.g. empty objects and properties on prototypes ([it doesn't check them](https://nodejs.org/api/assert.html#assert_assert_deepequal_actual_expected_message)), so it may be worth doing an explicit comparison with the expected elements

lighthouse-core/report/v2/templates.html
ebidel: I like the idea, but right now it's not worth it. Things are going to move around as we iterate. Don't want comments to get out of whack.

lighthouse-core/test/report/v2/report-generator-test.js
ebidel: done

lighthouse-core/report/v2/report-renderer.js

lighthouse-core/report/v2/report-renderer.js

ebidel:
PTAL

lighthouse-core/report/v2/templates.html
brendankenny: > I like the idea, but right now it's not worth it. Things are going to move around as we iterate. Don't want comments to get out of whack.

you say that now, but we'll have to see if anyone remembers later on :)

lighthouse-core/test/report/v2/report-generator-test.js
```diff
@@ -143,6 +146,20 @@ describe('ReportGeneratorV2', () => {
       assert.ok(result.includes('LIGHTHOUSE_JAVASCRIPT'), 'cannot be tricked');
     });
 
+    it('should inject the report templates', () => {
+      const page = jsdom.jsdom(new ReportGeneratorV2().generateReportHtml({}));
+      const templates = jsdom.jsdom(TEMPLATES_FILE);
+
+      assert.deepStrictEqual(page.querySelectorAll('template[id^="tmpl-"]'),
```
brendankenny: just checked, `assert.deepStrictEqual` also doesn't work here (besides checking that there's the same number of templates) because `Object.keys()` on the two `HTMLTemplateElement`s returns `[]`, so any two `HTMLTemplateElement`s are considered deep equal. You'll probably need to manually check the innerHTMLs are equal or something

brendankenny: could do something like
```js
const page = jsdom.jsdom(new ReportGeneratorV2().generateReportHtml({}));
const templates = jsdom.jsdom(TEMPLATES_FILE);

const pageTemplates = page.querySelectorAll('template[id^="tmpl-"]');
const rawTemplates = templates.querySelectorAll('template[id^="tmpl-"]');
rawTemplates.forEach((expectedTemplate, index) => {
  assert.strictEqual(pageTemplates[index].innerHTML, expectedTemplate.innerHTML);
});
```

ebidel: switched to using lengths.

brendankenny: > switched to using lengths

lulz, well the `assert.deepEqual` was checking that, but I guess at least now it's explicit

Issue #1927, opened 2017-03-28
Add browserVersion to lighthouse results


wwwillchen:
Similar to `lighthouseVersion`, we should annotate the results with the `browserVersion`. We can use the information from http://localhost:9222/json/version  (e.g. "Chrome/59.0.3051.0")

paulirish:
> We can use the information from http://localhost:9222/json/version

In the devtools and extension cases, we dont have this endpoint.

We _could_ fork based on the connection type... we could also evaluateScript to ask the page itself.. Maybe another option.. ?

brendankenny:
> We could fork based on the connection type

yeah, could have a `driver.getChromeVersion()` or whatever, have it ask the connection and have the implementation in the subclasses.

The extension can just check `navigator.userAgent`, is there an equivalent for devtools that also supports remote debugging a device?

paulirish:
remote debugging we have the /json/version endpoint so we're good there.

in devtools case we have the protocol, but not the /json/version endpoint.
AFAIK, we'll have to evaluateScript in this case.

PR #1968, opened 2017-04-05
Add user agent to lighthouse report

wwwillchen:
Capturing userAgent string which includes browser version & OS info.

Fixes https://github.com/GoogleChrome/lighthouse/issues/1927

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -91,7 +91,7 @@ class GatherRunner {
       }));
   }
 
-  static setupDriver(driver, options) {
+  static setupDriver(driver, gathererResults, options) {
     log.log('status', 'Initializing…');
     const resetStorage = !options.flags.disableStorageReset;
     // Enable emulation based on flags
@@ -101,7 +101,8 @@ class GatherRunner {
       .then(_ => driver.cacheNatives())
       .then(_ => resetStorage && driver.cleanAndDisableBrowserCaches())
       .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
-      .then(_ => driver.blockUrlPatterns(options.flags.blockedUrlPatterns || []));
+      .then(_ => driver.blockUrlPatterns(options.flags.blockedUrlPatterns || []))
+      .then(_ => gathererResults.userAgent = [driver.getUserAgent()]);
   }
 
   static disposeDriver(driver) {
@@ -326,7 +327,7 @@ class GatherRunner {
 
     return driver.connect()
       .then(_ => GatherRunner.loadBlank(driver))
-      .then(_ => GatherRunner.setupDriver(driver, options))
+      .then(_ => GatherRunner.setupDriver(driver, gathererResults, options))
 
       // Run each pass
       .then(_ => {
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -91,6 +91,12 @@ class GatherRunner {
       }));
   }
 
+  /**
+   * @param {!Driver} driver
+   * @param {!Object} gathererResults
+   * @param {!Object} options
+   * @return {!Promise}
+   */
   static setupDriver(driver, gathererResults, options) {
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/connections/connection.js
```diff
@@ -19,13 +19,31 @@
...
+ * @interface
+ */
+class ConnectionDelegate {
+  /**
+   * Evaluate an expression in the context of the current page.
+   * Returns a promise that resolves on the expression's value.
+   * @param {string} expression
+   * @return {!Promise<*>}
+   */
+  evaluateAsync(expression) { }
```
patrickhulce: this function already exists in `driver`, rather than implement this in each connection type thoughts on just using the userAgent method everywhere for now?

lighthouse-core/runner.js
```diff
@@ -144,7 +144,8 @@ class Runner {
           score = report.score;
         }
 
-        return {
+        return opts.driver.getChromeVersion().then(version => ({
```
wwwillchen: make it a gatherer runner and store it in artifacts.

brendankenny: > collect it in gatherer runner and store it in artifacts.

I like doing it earlier, it makes it kind of a prominent step having it here

brendankenny: actually, we do need it in gather-runner, as runner doesn't necessarily have a driver or a browser to talk to. I'd be ok with a `ChromeVersion` artifact on there.

Is there value in saving the whole UA string? e.g operating system might be helpful.

lighthouse-core/gather/connections/connection.js
brendankenny: > this function already exists in driver, rather than implement this in each connection type thoughts on just using the userAgent method everywhere for now?

yeah, I like this. Not sure why I skimmed over @paulirish suggesting that in #1927 :)

`driver.evaluateAsync('/Chrome\\/([0-9.]+)/.exec(navigator.userAgent)[1]')` would work (have to double escape the `/`, arguably a bug in our current `evaluateAsync`), or just `driver.evaluateAsync('navigator.userAgent')` if you want to do the regexing on the node side

lighthouse-core/runner.js
wwwillchen: I don't think it hurts to save the whole UA string. It might be valuable to later add chromeVersion and  platform (i.e.. operating system) as additional metadata properties.

lighthouse-core/gather/gather-runner.js
```diff
@@ -91,7 +91,13 @@ class GatherRunner {
       }));
   }
 
-  static setupDriver(driver, options) {
+  /**
+   * @param {!Driver} driver
+   * @param {!Object} gathererResults
```
brendankenny: `@param {!Object<string, !Array<!Promise<*>>} gathererResults` (should probably make a typedef at some point :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -101,7 +107,8 @@ class GatherRunner {
       .then(_ => driver.cacheNatives())
       .then(_ => resetStorage && driver.cleanAndDisableBrowserCaches())
       .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
-      .then(_ => driver.blockUrlPatterns(options.flags.blockedUrlPatterns || []));
+      .then(_ => driver.blockUrlPatterns(options.flags.blockedUrlPatterns || []))
+      .then(_ => gathererResults.userAgent = [driver.getUserAgent()]);
```
brendankenny: as much as the capitalization makes me cringe, this should be `gathererResults.UserAgent` to match the other names coming from gatherers

lighthouse-core/gather/gather-runner.js
wwwillchen: Made it a type-def :) I noticed all of them were missing a ">"

brendankenny: > I noticed all of them were missing a ">"

d'oh, nice catch

PR #1973, opened 2017-04-06
fix: do not throw when checking URL equality

### LINKED ISSUE #1972: Fatal error when running Lighthouse on certain YouTube pages
Upstream bug is https://github.com/jsdom/whatwg-url/issues/50 and underlying issue is that some network records have invalid URLs according to whatwg-url but not Chrome (if there's a hyphen at indices 3 and 4 as in `https://r15---sn-o097znl7.googlevideo.com/generate_204?conn2`)

Work around in progress

patrickhulce:
fixes #1972

### DIFF
--- a/lighthouse-core/test/lib/url-shim-test.js
+++ b/lighthouse-core/test/lib/url-shim-test.js
@@ -29,6 +29,12 @@ describe('URL Shim', () => {
     assert.doesNotThrow(_ => new URL(url));
   });
 
+  it.skip('handles URLs with multiple dashes', () => {
+    // from https://github.com/GoogleChrome/lighthouse/issues/1972
+    const url = 'https://r15---sn-o097znl7.googlevideo.com/generate_204?conn2';
+    assert.doesNotThrow(_ => new URL(url));
+  });
+
   it('safely identifies valid URLs', () => {
     assert.ok(URL.isValid('https://5321212.fls.net/page?query=string#hash'));
     assert.ok(URL.isValid('https://localhost:8080/page?query=string#hash'));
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/lib/url-shim-test.js
```diff
@@ -29,6 +29,12 @@ describe('URL Shim', () => {
     assert.doesNotThrow(_ => new URL(url));
   });
 
+  it.skip('handles URLs with multiple dashes', () => {
+    // from https://github.com/GoogleChrome/lighthouse/issues/1972
+    const url = 'https://r15---sn-o097znl7.googlevideo.com/generate_204?conn2';
+    assert.doesNotThrow(_ => new URL(url));
```
brendankenny: can drop the `doesNotThrow` and just construct the URL

patrickhulce: I noticed you've been on a kick lately to banish `doesNotThrow` from our tests. Do you not think it enhances the clarity that the statement is being executed exclusively to check for side effects? There's a reason the function exists :)

brendankenny: ha, fair. I guess it depends on what the purpose of the test is. In this case where you're not testing the output I guess it makes some sense to show what you are testing...but there seems to be a fine line between that and the fact that every test that isn't running `assert.throws()` is implicitly asserting that it isn't going to throw.

I guess in this case the regression is that the constructor was throwing, not that it was parsing incorrectly, so I guess it's not the worst thing to be explicit about that with the assertion statement.

patrickhulce: > not the worst thing

I'll take that as your resounding approval 😆

brendankenny: 🎺

PR #1974, opened 2017-04-06
feat: add npm run clean

patrickhulce:
We have a lot of different file extensions now :)


package.json
```diff
@@ -19,6 +19,7 @@
     "build-cli": "cd ./lighthouse-cli && npm run build",
     "build-extension": "cd ./lighthouse-extension && npm run build",
     "build-viewer": "cd ./lighthouse-viewer && npm run build",
+    "clean": "rm *.report.html *.report.dom.html *.report.json *.screenshots.html *.devtoolslog.json *.trace.json || true",
```
brendankenny: `rimraf` for easy cross platform support?

patrickhulce: ah yeah good call

PR #787, opened 2016-10-17
TTI: Update scoring PODR

addyosmani:
For #780. This will probably require some discussion. Alex's latest post on PWAs sets the bar for TTI to <5s. The current point of diminishing returns we use is 1700. I propose we raise this closer to 5000ms (here I use 4500) which has the following impact on the curve:

screen shot 2016-10-16 at 6 47 39 pm

For a site like housing.com, this takes the TTI score from:

**66** Time To Interactive (alpha) (3940.8ms)

to

**100** Time To Interactive (alpha) (3564.1ms)

cc @paulirish and @brendankenny as I believe they wrote a lot of the original code for the TTI implem.

PR #1977, opened 2017-04-06
fix: loosen TTI scoring

### LINKED ISSUE #1976: appveyor build broken on --perf smoketest
example run: https://ci.appveyor.com/project/paulirish/lighthouse/build/928/job/j34fb8vx4rwjh6rp

@patrickhulce it looks like you might have been trying to fix this [here](https://github.com/GoogleChrome/lighthouse/pull/1958/files#diff-b8e76f1137b1f350d675e8c5cc1ed3b6L5) by extending the trace long enough to find a quiet window, but no dice. Any insight?

patrickhulce:
mostly fixes #1976 

AppVeyor seems to use *really* slow machines, the online-only page takes 2 seconds to load...

### DIFF
--- a/lighthouse-core/audits/time-to-interactive.js
+++ b/lighthouse-core/audits/time-to-interactive.js
@@ -13,9 +13,9 @@ const TracingProcessor = require('../lib/traces/tracing-processor');
 const Formatter = require('../report/formatter');
 
 // Parameters (in ms) for log-normal CDF scoring. To see the curve:
-//   https://www.desmos.com/calculator/jlrx14q4w8
-const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
-const SCORING_MEDIAN = 5000;
+//   https://www.desmos.com/calculator/klsen8gqbf
+const SCORING_POINT_OF_DIMINISHING_RETURNS = 4000;
+const SCORING_MEDIAN = 10000;
 // This aligns with the external TTI targets in https://goo.gl/yXqxpL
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/time-to-interactive.js
```diff
@@ -13,9 +13,9 @@ const TracingProcessor = require('../lib/traces/tracing-processor');
 const Formatter = require('../report/formatter');
 
 // Parameters (in ms) for log-normal CDF scoring. To see the curve:
-//   https://www.desmos.com/calculator/jlrx14q4w8
-const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
-const SCORING_MEDIAN = 5000;
+//   https://www.desmos.com/calculator/klsen8gqbf
+const SCORING_POINT_OF_DIMINISHING_RETURNS = 4000;
```
patrickhulce: I won't merge this, but I propose it since we're not releasing for awhile anyway and we should totally redo our scoring before we do. It's driven me nuts, if we say target = 5000, you shouldn't get a 50/100 for hitting our target. This also differentiates more between 10s (which would qualify as PWA) and say 20s.

brendankenny: yeah, definitely don't include this :) There are much better forums for such changes

brendankenny:
You might wan to start this off with an email to lighthouse-core

brendankenny:
(cf. #787)

PR #1979, opened 2017-04-07
Generate DOM for markdown links in help text

ebidel:
R: all

Just support in help text / descriptions.

screen shot 2017-04-07 at 1 57 59 pm

Will add a test once https://github.com/GoogleChrome/lighthouse/pull/1963 lands.

### DIFF
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -183,18 +183,17 @@ class ReportRenderer {
 
   /**
    * @param {string} text
-   * @return {!Element}
+   * @return {!HTMLSpanElement}
    */
   _convertMarkdownLinksToElement(text) {
     const element = this._createElement('span');
 
     // Split on markdown links (e.g. [some link](https://...)).
-    const parts = text.split(/(\[(.*?)\]\((https?:\/\/.*?)\))/g);
+    const parts = text.split(/\[(.*?)\]\((https?:\/\/.*?)\)/g);
 
     while (parts.length) {
-      // Remove the same number of elements as there are capture groups.
-      // eslint-disable-next-line no-unused-vars
-      const [preambleText, fullMarkdownLink, linkText, linkHref] = parts.splice(0, 4);
+      // Pop off the same number of elements as there are capture groups.
+      const [preambleText, linkText, linkHref] = parts.splice(0, 3);
       element.appendChild(this._document.createTextNode(preambleText));
 
       // Append link if there are any.
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -195,7 +195,7 @@ class ReportRenderer {
       // Remove the same number of elements as there are capture groups.
       // eslint-disable-next-line no-unused-vars
       const [preambleText, fullMarkdownLink, linkText, linkHref] = parts.splice(0, 4);
-      element.appendChild(document.createTextNode(preambleText));
+      element.appendChild(this._document.createTextNode(preambleText));
 
       // Append link if there are any.
       if (linkText && linkHref) {
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -202,12 +202,8 @@ class ReportRenderer {
         a.rel = 'noopener';
         a.target = '_blank';
         a.textContent = linkText;
-        try {
-          a.href = (new URL(linkHref)).href;
-          element.appendChild(a);
-        } catch (e) {
-          element.appendChild(this._document.createTextNode(linkText));
-        }
+        a.href = (new URL(linkHref)).href;
+        element.appendChild(a);
       }
     }
 
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -201,9 +201,13 @@ class ReportRenderer {
         const a = this._createElement('a');
         a.rel = 'noopener';
         a.target = '_blank';
-        a.href = linkHref;
         a.textContent = linkText;
-        element.appendChild(a);
+        try {
+          a.href = (new URL(linkHref)).href;
+          element.appendChild(a);
+        } catch (e) {
+          element.appendChild(this._document.createTextNode(linkText));
+        }
       }
     }
 
--- a/lighthouse-core/report/v2/report-renderer.js
+++ b/lighthouse-core/report/v2/report-renderer.js
@@ -120,7 +120,8 @@ class ReportRenderer {
                           `lighthouse-score__value--${scoringMode}`);
 
     element.querySelector('.lighthouse-score__title').textContent = title;
-    element.querySelector('.lighthouse-score__description').textContent = description;
+    element.querySelector('.lighthouse-score__description')
+        .appendChild(this._convertMarkdownLinksToElement(description));
 
     return element;
   }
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,36 @@ class ReportRenderer {
   }
 
   /**
+   * @param {string} text
+   * @return {!Element}
```
brendankenny: `@return {!HTMLSpanElement}`. A function description here would be good too :)

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,36 @@ class ReportRenderer {
...
+   */
+  _convertMarkdownLinksToElement(text) {
+    const element = this._createElement('span');
+
+    // Split on markdown links (e.g. [some link](https://...)).
+    const parts = text.split(/(\[(.*?)\]\((https?:\/\/.*?)\))/g);
```
brendankenny: any reason to capture `fullMarkdownLink` if it's going to be unused?

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,36 @@ class ReportRenderer {
...
+
+    while (parts.length) {
+      // Remove the same number of elements as there are capture groups.
+      // eslint-disable-next-line no-unused-vars
+      const [preambleText, fullMarkdownLink, linkText, linkHref] = parts.splice(0, 4);
```
brendankenny: if you want to keep the larger capture group, you can also rename `fullMarkdownLink` to `_` to indicate that it's unused and drop the `// eslint-disable-next-line no-unused-vars`

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -26,10 +26,55 @@ const sampleResults = require('../../results/sample_v2.json');
...
 
 describe('ReportRenderer V2', () => {
-  describe('renderReport', () => {
-    const document = jsdom.jsdom(TEMPLATE_FILE);
-    const renderer = new ReportRenderer(document);
+  const document = jsdom.jsdom(TEMPLATE_FILE);
+  const renderer = new ReportRenderer(document);
+
+  describe('createElement', () => {
+    it('creates a simple element using default values', () => {
```
brendankenny: what's new and what's not is very confusing when you move things around :P

lighthouse-core/test/report/v2/report-renderer-test.js
```diff
@@ -26,10 +26,55 @@ const sampleResults = require('../../results/sample_v2.json');
...
+      assert.ok(clone.querySelector('.lighthouse-score'));
+    });
+
+    it('fails when template cannot be found', () => {
+      assert.throws(() => renderer._cloneTemplate('#unknown-selector'));
+    });
+  });
+
+  describe('_convertMarkdownLinksToElement', () => {
+    it('converts links', () => {
```
brendankenny: A few more corner cases would be nice. With/without preamble text and text afterwards, etc

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,36 @@ class ReportRenderer {
...
+      element.appendChild(this._document.createTextNode(preambleText));
+
+      // Append link if there are any.
+      if (linkText && linkHref) {
+        const a = this._createElement('a');
+        a.rel = 'noopener';
+        a.target = '_blank';
+        a.href = linkHref;
```
brendankenny: do we want to verify the url? Maybe someday it would be good if we start using this in more places, but today it would start checking our help text URLs, which would be nice

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,36 @@ class ReportRenderer {
...
+   * @return {!Element}
+   */
+  _convertMarkdownLinksToElement(text) {
+    const element = this._createElement('span');
+
+    // Split on markdown links (e.g. [some link](https://...)).
+    const parts = text.split(/(\[(.*?)\]\((https?:\/\/.*?)\))/g);
+
+    while (parts.length) {
+      // Remove the same number of elements as there are capture groups.
```
brendankenny: I always forget that `splice` affects both creates a new array *and* modifies the original, so maybe change the comment to something like `// Pop off the same number of elements as there are capture groups.` to make it clear?

lighthouse-core/report/v2/report-renderer.js
brendankenny: ooh, can also do `const [preambleText, , linkText, linkHref]`, though don't know how our eslint would react

lighthouse-core/report/v2/report-renderer.js
brendankenny: you could also avoid the `splice` by doing something like
```js
let parts = text.split(/(\[(.*?)\]\((https?:\/\/.*?)\))/g);
while (parts.length) {
  const [preambleText, fullMarkdownLink, linkText, linkHref, ...rest] = parts;
  // ...
  parts = rest;
}
```
more explicit, though the extra assignment is kind of annoying

lighthouse-core/report/v2/report-renderer.js
ebidel: Is the type of element important We're not specifying subclass type for the other methods.

lighthouse-core/report/v2/report-renderer.js
ebidel: what do you mean by verify?

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: i like to keep you on your toes.

lighthouse-core/report/v2/report-renderer.js
brendankenny: > Is the type of element important We're not specifying subclass type for the other methods.

not really, though it doesn't hurt. I'm more thinking for documentation purposes it's nice to have the information that your linkified text is coming back in a `<span>`.

lighthouse-core/test/report/v2/report-renderer-test.js
ebidel: done

lighthouse-core/report/v2/report-renderer.js
ebidel: i just removed the extra group. no longer needed.

lighthouse-core/report/v2/report-renderer.js
ebidel: nope done.

lighthouse-core/report/v2/report-renderer.js
brendankenny: > what do you mean by verify?

sorry, I meant like check that it's a valid url. `a.href = new URL(linkHref).href` or whatever

lighthouse-core/report/v2/report-renderer.js
ebidel: updated the comment.

lighthouse-core/report/v2/report-renderer.js

ebidel:
PTAL

lighthouse-core/report/v2/report-renderer.js
```diff
@@ -181,6 +182,39 @@ class ReportRenderer {
...
+      if (linkText && linkHref) {
+        const a = this._createElement('a');
+        a.rel = 'noopener';
+        a.target = '_blank';
+        a.textContent = linkText;
+        try {
+          a.href = (new URL(linkHref)).href;
+          element.appendChild(a);
+        } catch (e) {
+          element.appendChild(this._document.createTextNode(linkText));
```
brendankenny: should we fail silently here? Seems like a big enough mistake that it should complain loudly and end up in `_renderException`

ebidel: I think it's better to fail silently here. If a link is busted for some reason, it shouldn't block the whole report from being generated.

brendankenny: but why is the link busted? Right now, at least, they're all coming from helpText. An error in one of those should be fixed immediately

ebidel: done. 

FWIW, b/c of the regex, the only time we'd get an invalid URL is if someone wrote `https://`. or used something crazy in help text like `[boom](🍣🍺)`.

ebidel:
This ready. Let's get it landed so other PRs can proceed.

patrickhulce:
let's merge this so we can rebase #1987 and #1992 💥

PR #1987, opened 2017-04-10
refactor: split report-renderer into multiple files

patrickhulce:
In preparation for introducing tags/grouped audit renderers the report-renderer.js started getting unwieldy. This splits it into 3 files DOM, DetailsRenderer, and ReportRenderer.

I know @brendankenny had Eric kill DOM but c'mon having these duplicated everywhere is bad and the class allows for easy test injection. I'll split up the test files too, but currently it conflicts with @ebidel's PR and figured I'd wait before I get too far.


paulirish:
works for me

ebidel:
Works for me too. You know I love that `DOM`.

ebidel:
@patrickhulce you good on this? I don't want to get too deep and have rebase nightmares.

patrickhulce:
yep I'm ready if you are!

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -0,0 +1,91 @@
...
+    element.appendChild(items);
+    return element;
+  }
+}
+
+if (typeof module !== 'undefined' && module.exports) {
+  module.exports = DetailsRenderer;
+}
+
+/** @typedef {{type: string, text: string|undefined, header: DetailsJSON|undefined, items: Array<DetailsJSON>|undefined}} */
```
brendankenny: this won't work (it's scoped to the module like regular variables). I believe you can put it on `DetailsRenderer`, but then it'll have to be referred to as `DetailsRenderer.DetailsJSON` in report-renderer.js

brendankenny: although you can wait on that, because these are broken anyways (no non-null and I believe union types need parens when nested like this)

patrickhulce: done

Issue #1937, opened 2017-03-30
[meta] I/O Burndown


paulirish:
_Feel free to comment or edit. Check items once PRs are up for review, rather than when merged._

---------

### Report and Config refactors
* [x] ReportinDOM: Render the Lighthouse report without innerHTML [team]
  * [x] Refactor from Handlebars -> DOM (#1933) _#blocksDevTools_  [@dgozman]
  * [x] No more markdown links [@ebidel]
  * [ ] Remove markdown from descriptions. [?]
  * [ ] Migrate formatters to DOM [@ebidel, others]
    * [x] cards https://github.com/GoogleChrome/lighthouse/pull/1992
    * [x] table [@paulirish]
    * [ ] critical request chain [@ebidel]
    * [ ] nodes incl revealing (a11y) [@paulirish]
    * [ ] user timings [?]
    * [ ] url list [@patrickhulce]
  * [x] show [passing tests UI](https://docs.google.com/document/u/1/d/1jFoI5Anwm033n7SDabqUbjiWMzY2X31n5CqP254RH3A/edit#) [@paulirish]
  * [x] [performance UI](https://docs.google.com/document/u/1/d/1jFoI5Anwm033n7SDabqUbjiWMzY2X31n5CqP254RH3A/edit) for filmstrip & metrics ([crbug.com/717082](https://bugs.chromium.org/p/chromium/issues/detail?id=717082)) [?]
  * [x] perf diagnostics: sparklines of estimated cost (#2112) [@patrickhulce]
  * [ ] ~~Report UI: view system (#2111, [crbug.com/717079](https://bugs.chromium.org/p/chromium/issues/detail?id=717079))~~ punt
  * [x] Report UI: gauge scores
  * [ ] Report UI: other misc report needs
* [x] DevTools UI: Sidebar ([crbug.com/717076](https://bugs.chromium.org/p/chromium/issues/detail?id=717076)) [@pavelfeldman]
* [x] DevTools UI: modal for starting new audit. ([crbug.com/717072](https://bugs.chromium.org/p/chromium/issues/detail?id=717072)) [@pavelfeldman]
* [ ] DevTools UI: share/context menu. print?  --- _may be punted_
* [x] Configv2: Aggregations -> Categories (#1935) [@patrickhulce]
* [x] Configv2: Subitems -> Tags (topics) [@patrickhulce]
* [x] Configv2: category/audit whitelisting via settings (#1988)  [@patrickhulce]

[...]

PR #1988, opened 2017-04-10
feat: support category and audit whitelisting

patrickhulce:
addresses burndown items for I/O #1937 

usage to run just TTI, PSI, and the PWA audits
```js
{
  settings: {
     onlyAudits: ["time-to-interactive", "speed-index-metric"],
     onlyCategories: ["pwa"]
  }
}
```


lighthouse-core/config/config.js
```diff
@@ -344,14 +356,21 @@ class Config {
    * Filter out any unrequested categories from the categories object.
    * @param {!Object<string, {audits: !Array<{id: string}>}>} categories
    * @param {Array<string>=} categoryIds
+   * @param {Array<string>=} auditIds
```
brendankenny: looks like we missed this before and these should be `{!Array<string>=}`

lighthouse-core/config/config.js
```diff
@@ -344,14 +356,21 @@ class Config {
    * Filter out any unrequested categories from the categories object.
```
brendankenny: need to update

lighthouse-core/config/perf.json
```diff
@@ -1,114 +1,6 @@
...
-    "pauseBeforeTraceEndMs": 5000,
-    "useThrottling": true,
-    "gatherers": [
-      "image-usage",
-      "viewport-dimensions",
-      "dobetterweb/domstats",
-      "dobetterweb/tags-blocking-first-paint",
-      "dobetterweb/optimized-images"
-    ]
+  "extends": "lighthouse:default",
```
brendankenny: this is a lie depending on the current check just being truthiness of `extends`, right?

patrickhulce: yup

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/config/config.js
patrickhulce: done

patrickhulce:
> Does it make more sense to have onlyCategories and onlyAudits properties instead of the less explicit only prop?

I'm fine with this too, but with the goal of having all config settings map to CLI flags that seemed slightly polluting. WFM if others agree to split though.

> We need to be clear about the hierarchy here (but where?)

Agreed, this keeps it simple by being 100% additive so if it's listed in `only` it's guaranteed to run, added warnings too.

paulirish:
> Does it make more sense to have onlyCategories and onlyAudits properties instead of the less explicit only prop?

I'm fine with both and consider them equivalent from an ergonomics POV.

patrickhulce:
this waiting on anything else? PTAL :)

lighthouse-core/config/config.js
```diff
@@ -341,17 +351,47 @@ class Config {
...
+   * @param {!Array<string>=} categoryIds
+   * @param {!Array<string>=} auditIds
    * @return {!Object<string, {audits: !Array<{id: string}>}>}
    */
-  static filterCategories(oldCategories, categoryIds = []) {
+  static filterCategoriesAndAudits(oldCategories, categoryIds = [], auditIds = []) {
     const categories = {};
 
+    // warn if the category is not found
+    categoryIds.forEach(category => {
```
brendankenny: `category` -> `categoryId`

lighthouse-core/config/config.js
```diff
@@ -341,17 +351,47 @@ class Config {
...
+      if (!oldCategories[category]) {
+        log.warn('config', `unrecognized category: ${category}`);
+      }
+    });
+
+    // warn if the audit is not found in a category
+    auditIds.forEach(audit => {
```
brendankenny: `audit` -> `auditId`

lighthouse-core/config/config.js
```diff
@@ -341,17 +351,47 @@ class Config {
...
    * @return {!Object<string, {audits: !Array<{id: string}>}>}
    */
-  static filterCategories(oldCategories, categoryIds = []) {
+  static filterCategoriesAndAudits(oldCategories, categoryIds = [], auditIds = []) {
     const categories = {};
 
+    // warn if the category is not found
+    categoryIds.forEach(category => {
+      if (!oldCategories[category]) {
+        log.warn('config', `unrecognized category: ${category}`);
```
brendankenny: need a clearer way of saying this, since users will see it without context. Not sure what would be good though... maybe just ``unrecognized category in `onlyCategories`: ${category}``?

lighthouse-core/config/config.js
```diff
@@ -341,17 +351,47 @@ class Config {
...
+
+    // warn if the audit is not found in a category
+    auditIds.forEach(audit => {
+      const foundCategory = Object.keys(oldCategories).find(category => {
+        const audits = oldCategories[category].audits;
+        return audits.find(candidate => candidate.id === audit);
+      });
+
+      if (!foundCategory) {
+        log.warn('config', `unrecognized audit: ${audit}`);
```
brendankenny: same for this

lighthouse-core/config/config.js
```diff
@@ -341,17 +351,47 @@ class Config {
...
+      }
+
+      if (categoryIds.includes(foundCategory)) {
+        log.warn('config', `${foundCategory} category already includes "${audit}"`);
```
brendankenny: I'm not sure how to word this one either. ``${audit} in `onlyAudits` already included by `onlyCategories` entry ${foundCategory}``? Or something less terrible :)

lighthouse-core/test/config/config-test.js
```diff
@@ -265,6 +265,64 @@ describe('Config', () => {
...
+    const config = new Config({
+      extends: true,
+      settings: {
+        onlyCategories: ['performance'],
+        onlyAudits: ['is-on-https'],
+      },
+    });
+
+    assert.ok(config.audits.length, 'inherited audits by extension');
+    assert.ok(config.audits.length < origConfig.audits.length, 'filtered out audits');
```
brendankenny: could you do `origConfig.categories.performance.audits.length + 1`? Seems future compatible

lighthouse-core/test/config/config-test.js
```diff
@@ -384,39 +442,61 @@ describe('Config', () => {
...
+      assert.equal(config.audits.length, auditCount, 'audit filtering failed');
+    });
+
+    it('should support redundant filtering', () => {
+      const config = Config.generateNewFilteredConfig(origConfig, ['pwa'], ['is-on-https']);
+      const selectedCategory = origConfig.categories.pwa;
+      const auditCount = Object.keys(selectedCategory.audits).length;
+      assert.equal(config.passes.length, 3, 'incorrect # of passes');
+      assert.equal(config.audits.length, auditCount, 'audit filtering failed');
+    });
```
brendankenny: tests for the new warnings would be good, too. `log.events.addListener('warning', callback)` can be an easy way to do that rather than going to the trouble of overriding `log`, like is done [elsewhere](https://github.com/GoogleChrome/lighthouse/blob/605a4579c5cd0eeba9ea71a3dcf38d7817d1b86a/lighthouse-core/test/config/config-test.js#L88-L95) in that file

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -99,7 +99,7 @@ function filterOutArtifacts(result) {
  * @return {!Promise}
  */
 window.runLighthouseForConnection = function(connection, url, options, categoryIDs) {
-  const newConfig = Config.generateNewConfigOfCategories(defaultConfig, categoryIDs);
+  const newConfig = Config.generateNewFilteredConfig(defaultConfig, categoryIDs);
```
brendankenny: should we switch this to just
```js
const newConfig = {
  "extends": "lighthouse:default",
  "settings": {
    "onlyCategories": categoryIDs
  }
};
```

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/config/config.js
patrickhulce: done

lighthouse-core/test/config/config-test.js
patrickhulce: done

lighthouse-core/test/config/config-test.js
patrickhulce: done

lighthouse-extension/app/src/lighthouse-background.js
patrickhulce: done

PR #1992, opened 2017-04-11
Report v2: port over cards formatter

ebidel:
R: all

screen shot 2017-04-10 at 10 10 26 pm

This should wait until https://github.com/GoogleChrome/lighthouse/pull/1979 and https://github.com/GoogleChrome/lighthouse/pull/1987 land and will probably need rebasing, but review can start now. The changes are standalone from those.

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -24,7 +24,7 @@ class DetailsRenderer {
   }
 
   /**
-   * @param (!DetailsRenderer.DetailsJSON|!DetailsRenderer.CardsDetailsJSON)} details
+   * @param {(!DetailsRenderer.DetailsJSON|!DetailsRenderer.CardsDetailsJSON)} details
    * @return {!Element}
    */
   render(details) {
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -24,7 +24,7 @@ class DetailsRenderer {
   }
 
   /**
-   * @param {!DetailsJSON} details
+   * @param (!DetailsRenderer.DetailsJSON|!DetailsRenderer.CardsDetailsJSON)} details
    * @return {!Element}
    */
   render(details) {
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -33,6 +33,8 @@ class DetailsRenderer {
         return this._renderText(details);
       case 'block':
         return this._renderBlock(details);
+      case 'cards':
+        return this._renderCards(details);
       case 'list':
         return this._renderList(details);
       default:
@@ -78,6 +80,36 @@ class DetailsRenderer {
     for (const item of list.items) {
       items.appendChild(this.render(item));
     }
+    element.appendChild(items);
+    return element;
+  }
+
+  /**
+   * @param {!DetailsJSON} list
+   * @return {!Element}
+   */
+  _renderCards(list) {
+    const element = this._dom.createElement('details', 'lighthouse-details');
+    if (list.header) {
+      element.appendChild(this._dom.createElement('summary')).textContent = list.header.text;
+    }
+
+    const items = this._dom.createElement('div', 'lighthouse-scorecards');
+    for (const item of list.items) {
+      const card = items.appendChild(
+          this._dom.createElement('div', 'lighthouse-scorecard', {title: item.snippet}));
+      const titleEl = this._dom.createElement('div', 'lighthouse-scorecard__title');
+      const valueEl = this._dom.createElement('div', 'lighthouse-scorecard__value');
+      const targetEl = this._dom.createElement('div', 'lighthouse-scorecard__target');
+
+      card.appendChild(titleEl).textContent = item.title;
+      card.appendChild(valueEl).textContent = item.value;
+
+      if (item.target) {
+        card.appendChild(targetEl).textContent = `target: ${item.target}`;
+      }
+    }
+
     element.appendChild(items);
     return element;
   }
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -85,18 +85,18 @@ class DetailsRenderer {
   }
 
   /**
-   * @param {!DetailsJSON} list
+   * @param {!CardsDetailsJSON} details
    * @return {!Element}
    */
-  _renderCards(list) {
+  _renderCards(details) {
     const element = this._dom.createElement('details', 'lighthouse-details');
-    if (list.header) {
-      element.appendChild(this._dom.createElement('summary')).textContent = list.header.text;
+    if (details.header) {
+      element.appendChild(this._dom.createElement('summary')).textContent = details.header.text;
     }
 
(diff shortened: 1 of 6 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -81,6 +83,36 @@ class DetailsRenderer {
     element.appendChild(items);
     return element;
   }
+
+  /**
+   * @param {!DetailsJSON} list
+   * @return {!Element}
+   */
+  _renderCards(list) {
```
brendankenny: not a list :)

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -81,6 +83,36 @@ class DetailsRenderer {
...
+    const element = this._dom.createElement('details', 'lighthouse-details');
+    if (list.header) {
+      element.appendChild(this._dom.createElement('summary')).textContent = list.header.text;
```
brendankenny: why are we keeping the verbose `header: {type: 'text', text: 'View details'}` format if we don't do anything with it (here or in the list formatter)?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -81,6 +83,36 @@ class DetailsRenderer {
...
+    }
+
+    const items = this._dom.createElement('div', 'lighthouse-scorecards');
```
brendankenny: nit: need to differentiate types of things here. maybe call this `parentElement` or something?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -81,6 +83,36 @@ class DetailsRenderer {
     element.appendChild(items);
     return element;
   }
+
+  /**
+   * @param {!DetailsJSON} list
```
brendankenny: this isn't a `DetailsJSON`, though (specifically `list.items` isn't a `Array<DetailsJSON>|undefined`). Maybe you want to create a new `CardsDetailsJSON` or whatever? I'm also not clear where this leaves us on the whole "new `details` property has a standard structure" thing

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -37,7 +37,9 @@ class DOM {
       element.className = className;
     }
     Object.keys(attrs).forEach(key => {
-      element.setAttribute(key, attrs[key]);
+      if (attrs[key] !== undefined) {
```
brendankenny: need to update the `attrs` type signature if adding this, but it would probably be better to keep it strict and make the caller correct it if they don't want an attribute

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: You mean just use a string: `header: 'View details'`?

lighthouse-core/report/v2/renderer/dom.js
ebidel: `{!Object<?string, string>=}`?

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done

lighthouse-core/report/v2/renderer/dom.js
brendankenny: if it shows up in `Object.keys()` but is `undefined` then it would be `{!Object<string, (string|undefined)>=}` but it's still silly to be passing in undefined attributes. If it's just in that one spot, why not make a `const attributes = item.snippet ? {title: item.snippet} : undefined` or whatever there

lighthouse-core/report/v2/renderer/details-renderer.js
brendankenny: > You mean just use a string: `header: 'View details'`?

or whatever. I'm just wondering why it's a full `DetailsJSON` if we don't use it

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -81,6 +83,36 @@ class DetailsRenderer {
     element.appendChild(items);
     return element;
   }
+
+  /**
+   * @param {!CardsDetailsJSON} details
```
brendankenny: `render()` needs to take a `(DetailsRenderer.DetailsJSON|DetailsRenderer.CardsDetailsJSON)` now...this is why we need to turn type checking back on :)

lighthouse-core/report/v2/renderer/dom.js
ebidel: Every caller will have to do that. I'd rather just prevent `attr="undefined"` in one place. Kinda what templating languages do for you ;)

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done. closure is the worst.

lighthouse-core/report/v2/renderer/dom.js
ebidel: done

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -24,7 +24,7 @@ class DetailsRenderer {
   }
 
   /**
-   * @param {!DetailsJSON} details
+   * @param (!DetailsRenderer.DetailsJSON|!DetailsRenderer.CardsDetailsJSON)} details
```
brendankenny: missing a `{`

lighthouse-core/report/v2/renderer/details-renderer.js
brendankenny: closure is the worst but not for this reason. The parameter can be one of two different things, so... :P

lighthouse-core/report/v2/renderer/dom.js
brendankenny: > Every caller will have to do that. I'd rather just prevent attr="undefined" in one place. Kinda what templating languages do for you ;)

I'd imagine most templating languages would give you `title="undefined"` in this case (which I assume is why the [current partial](https://github.com/GoogleChrome/lighthouse/blob/605a4579c5cd0eeba9ea71a3dcf38d7817d1b86a/lighthouse-core/report/partials/cards.html#L5) has `{{#if snippet}}title="{{snippet}}"{{/if}}` in it). I'd generally consider accidentally passing in attributes with `undefined` values a bug and it would be nice to catch that

ebidel: Exactly. It's replacing that conditional. The good templating systems do this : http://jsbin.com/damayuhiwa/edit?html,output.

brendankenny: I guess. You should document that behavior in that case so no one is surprised by it

brendankenny:
need a rebase now

lighthouse-core/report/v2/renderer/dom.js
ebidel: done

ebidel:
rebasing all my life.

#K-Ci & JoJo

PR #1991, opened 2017-04-11
ignore proto expectation errors from traceviewer

paulirish:
(really, all of the console.error() calls are muted by this)

I am replacing with `console.assert(true, ...)` just to not overcomplicate the replacement.

PR #1998, opened 2017-04-11
Remove babel transpile of traceviewerjs

paulirish:
node 4 didn't support `new.target` or destructuring. We don't need to handle these cases any longer.

(from #1991)


.eslintignore
```diff
@@ -23,4 +23,4 @@ lighthouse-core/report/templates/report-templates.js
 lighthouse-core/report/partials/templates/report-partials.js
 
 # Generated files
-plots/out/**
+plots/out*
```
brendankenny: whats the distinction you're making here?

lighthouse-core/scripts/build-traceviewer-module.js
```diff
@@ -126,18 +124,11 @@ function convertImport(src) {
...
       // adjust and transpile for usage in node 4+
-      function polyfillNode4SupportAndMinify(dest, scriptsContent) {
-        let transformed = scriptsContent;
-        if (dest.endsWith('/slice.js')) {
-          // no babel plugin that can transpile `new.target`, so this is done manually
-          transformed = transformed.replace('if (new.target)', 'if (!(this instanceof Slice))');
-        }
-        transformed = babel.transform(transformed, {
-          plugins: ['transform-es2015-destructuring'],
+      function minify(dest, scriptsContent) {
```
brendankenny: update node 4+ comment above

.eslintignore
paulirish: Personally, i've been taking old `out` folders and renaming them to `out-old-thevergeruns` etc. cc @wwwillchen in case he has a preference for handling measure runs differently.

lighthouse-core/scripts/build-traceviewer-module.js
paulirish: good call. done

PR #2000, opened 2017-04-12
Report: header, footer, leftnav, share, oh my

ebidel:
R: all

- Adds the header back to the report, redesigned to match the devtools mocks. The "Runtime settings" dropdown wasn't mocked.
- Centers the report and gives it a max width for easier reading. I like it, but we can keep the justified left.
- support for multiple calls to `renderReport`(e.g. future where we have re-run audit)
- responsive tweaks
- Logger is now a toast at the bottom of the page
- also adds test for some of our helpers.

Most of the code is porting. I'll try to annotate what's what.

screen shot 2017-04-12 at 3 36 10 pm
screen shot 2017-04-12 at 3 36 19 pm


lighthouse-core/report/v2/logger.js
```diff
@@ -0,0 +1,73 @@
...
+
+'use strict';
+
+/* eslint-env browser */
+
+/**
+ * Logs messages via a UI butter.
+ * @class
+ */
+class Logger {
```
ebidel: this is a straight port from v1.

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,273 @@
...
+
+/* globals Logger */
+/* eslint-env browser */
+
+class ReportFeatures {
+
+  /**
+   * @param {!Document} document
+   */
+  constructor(document) {
```
ebidel: this is a port from v1. Only difference is that the constructor does not automatically attache event listeners. You have to call `attach()` now.

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,273 @@
...
+
+    this._copyAttempt = false;
+  }
+
+  /**
+   * Copies the report JSON to the clipboard (if supported by the browser).
+   */
+  onCopyButtonClick() {
+    if (window.ga) {
+      window.ga('send', 'event', 'report', 'copy');
```
ebidel: Would love to keep GA for viewer. Should be fine since this is a noop for contexts that don't load ga.

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,273 @@
...
+        // Since Viewer generates its page HTML dynamically from report JSON,
+        // run the ReportGenerator. For everything else, the page's HTML is
+        // already the final product.
+        // if (e.target.dataset.context !== 'viewer') {
+        //   htmlStr = this._document.documentElement.outerHTML;
+        // } else {
+        //   const reportGenerator = new ReportGeneratorV2();
+        //   htmlStr = reportGenerator.generateReportHTML(this.json);
+        // }
+        // TODO: fix viewer.
```
ebidel: Need to make this one work again.

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -54,24 +55,53 @@ function formatNumber(number) {
   return number.toLocaleString(undefined, {maximumFractionDigits: 1});
 }
 
+/**
+ * Format time.
+ * @param {string} date
+ * @return {string}
+ */
+function formatDateTime(date) {
```
ebidel: port from v1

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -54,24 +55,53 @@ function formatNumber(number) {
...
   }
 
   /**
    * @param {!ReportJSON} report
-   * @return {!Element}
+   * @param {!Element} container Parent element to render the report into.
    */
-  renderReport(report) {
+  renderReport(report, container) {
+    container.innerHTML = ''; // Remove previous report.
```
ebidel: I'm hoping this one isn't contentious :)

lighthouse-core/report/v2/report-template.html
```diff
@@ -27,11 +27,16 @@
...
+  <div id="lighthouse-content"></div>
+  <div id="lighthouse-log"></div>
   <script>%%LIGHTHOUSE_JAVASCRIPT%%</script>
   <script>window.__LIGHTHOUSE_JSON__ = %%LIGHTHOUSE_JSON%%;</script>
   <script>
-    const renderer = new ReportRenderer(document);
-    document.body.appendChild(renderer.renderReport(window.__LIGHTHOUSE_JSON__));
+    window.addEventListener('DOMContentLoaded', _ => {
+      window.renderer = new ReportRenderer(document);
+      const container = document.querySelector('#lighthouse-content');
```
ebidel: made sense to render into a container rather than replace the whole world for every call to render.

lighthouse-core/report/v2/templates.html
```diff
@@ -18,9 +18,231 @@
...
+        <div class="lighthouse-toggle-arrow" title="See audits"></div>
       </summary>
       <div class="lighthouse-score__description"><!-- fill me --></div>
     </details>
   </div>
 </template>
+
+<!-- Lighthouse header -->
+<template id="tmpl-lighthouse-heading">
+  <style>
```
ebidel: For these  one-time-use partials....instead of creating separate .css file for each (or adding to the main stylesheet), I chose to keep related styles together with the markup that requires them. One nice thing about this is that these styles don't get added to the page unless a renderer uses the template. So if devtools renders its own header, we're not penalized.

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -44,11 +44,12 @@ class DOM {
 
   /**
    * @param {string} selector
+   * @param {Element=} context Parent node to query for templates within.
    * @return {!DocumentFragment} A clone of the template content.
    * @throws {Error}
    */
-  cloneTemplate(selector) {
-    const template = this._document.querySelector(selector);
+  cloneTemplate(selector, context = this._document) {
```
patrickhulce: nice

lighthouse-core/report/v2/renderer/logger.js
```diff
@@ -0,0 +1,73 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+/* eslint-env browser */
```
patrickhulce: I've been avoiding setting this and preferring explicit globals instead, so when we test in node it's really explicit which globals it depends on, thoughts?

lighthouse-core/report/v2/renderer/report-features.js
patrickhulce: when we do, instead of generating it on the fly I'd be interested to publish the bundled JS with our releases instead that the viewer can just reference. This way the viewer can actually start supporting all versions going forward too

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -54,25 +55,58 @@ function formatNumber(number) {
...
+ * Format time.
+ * @param {string} date
+ * @return {string}
+ */
+function formatDateTime(date) {
+  const options = {
+    month: 'numeric', day: 'numeric', year: 'numeric',
+    hour: 'numeric', minute: 'numeric', timeZoneName: 'short'
+  };
+  let formatter = new Intl.DateTimeFormat('en-US', options);
```
patrickhulce: I believe there's an issue outstanding for our permanent use of `en-US`, since we're starting fresh, thoughts on just using `.toISOString()` for now and not bringing our baggage with us yet? :)

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,273 @@
...
+        }
+      });
+    }
+  }
+  /**
+   * Downloads a file (blob) using a[download].
+   * @param {Blob|File} blob The file to save.
+   */
+  _saveFile(blob) {
+    const filename = window.getFilenamePrefix({
```
patrickhulce: we should really get around to fixing `getFilenamePrefix` haha

lighthouse-core/report/v2/renderer/logger.js
ebidel: Sure, we ca do explicit globals.

lighthouse-core/report/v2/renderer/report-renderer.js
ebidel: The devtools mocks make this even nicer:

screen shot 2017-04-13 at 1 43 47 pm

What if we go with that?

lighthouse-core/report/v2/renderer/logger.js
ebidel: done

lighthouse-core/report/v2/renderer/report-renderer.js
patrickhulce: sgtm!

ebidel:
Working on some tests...

ebidel:
Added tests for the non-interactive stuff (newer header/footer renders). Also added tests for some of the format helpers.

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -196,7 +274,12 @@ class ReportRenderer {
 }
 
 if (typeof module !== 'undefined' && module.exports) {
-  module.exports = ReportRenderer;
+  module.exports = {
```
brendankenny: if these need to be exported just make them static members?

OTOH, if it's just for testing (I think it is right now?), I'd rather just test the effects on the final generated report so anyone doing future refactors isn't spending their time moving stuff around just for tests...the score is just correctly formatted or it's not

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -54,25 +54,58 @@ function formatNumber(number) {
...
   }
 
   /**
    * @param {!ReportJSON} report
+   * @param {!Element} container Parent element to render the report into.
    * @return {!Element}
    */
-  renderReport(report) {
+  renderReport(report, container) {
+    container.innerHTML = ''; // Remove previous report.
```
brendankenny: can't you still do `container.textContent = '';`

lighthouse-core/report/v2/renderer/logger.js
```diff
@@ -0,0 +1,71 @@
...
+    this.el.textContent = msg;
+    this.el.classList.add('show');
+    if (autoHide) {
+      this._id = setTimeout(_ => {
+        this.el.classList.remove('show');
+      }, 7000);
+    }
+  }
+
+  warn(msg) {
```
brendankenny: need jsdoc on these

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -147,13 +180,58 @@ class ReportRenderer {
...
+    return footer;
+  }
+
+  /**
+   * @param {!ReportJSON} report
    * @return {!Element}
    */
   _renderReport(report) {
-    const element = this._dom.createElement('div', 'lh-report');
+    const container = this._dom.createElement('div', 'lh-content');
```
brendankenny: there's a `.lh-content` div inside a `#lh-content` div? Or is this a old code? Appears to be discarded since `element` is returned and attached elsewhere in the DOM

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  closeExportDropdown() {
+    this.exportButton.classList.remove('active');
+  }
+
+  /**
+   * Click handler for export button.
+   */
+  onExportButtonClick(e) {
```
brendankenny: need jsdoc on these `e`s

lighthouse-core/report/v2/renderer/dom.js
paulirish: i have a conflict with this in flight: https://github.com/GoogleChrome/lighthouse/pull/2002/files#diff-67023326fb4e20403076afc5d18d67f2R51

I can adopt yours pretty easily though.

lighthouse-core/report/v2/renderer/report-features.js
brendankenny: can we just remove for now since it's non-functional code?

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+
+    if (!e.target.dataset.action) {
+      return;
+    }
+
+    switch (e.target.dataset.action) {
+      case 'copy':
+        this.onCopyButtonClick();
+        break;
+      case 'open-viewer':
```
brendankenny: remove?

lighthouse-core/report/v2/renderer/dom.js
paulirish: are you comfortable with using `this.templatesContext` instead? 

```js
cloneTemplate(selector) {
  const template = this.templatesContext.querySelector(selector);
```

this would be ideal for me. I'd set it right after instantiating the instance. 
it would default to `this._document`

paulirish: i have this implemented in https://github.com/GoogleChrome/lighthouse/pull/2002. 
wdyt?

ebidel: https://github.com/GoogleChrome/lighthouse/pull/2002#discussion_r111594934

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,272 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/* globals document window URL Blob Logger */
+
+class ReportFeatures {
```
ebidel: `ReportUIFeatures` sgtm.

lighthouse-core/report/v2/renderer/report-features.js
ebidel: It's always been functional in viewer (only context that loads GA atm).

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -44,11 +44,12 @@ class DOM {
 
   /**
    * @param {string} selector
+   * @param {Element=} context Parent node to query for templates within.
```
ebidel: done

lighthouse-core/report/v2/renderer/logger.js
ebidel: done

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,272 @@
...
+      dropdown.addEventListener('click', this.onExport);
+
+      this._document.addEventListener('copy', this.onCopy);
+    }
+    this._document.addEventListener('keydown', this.printShortCutDetect);
+  }
+
+  /**
+   * Adds export button and print functionality to the report.
+   * @param {ReportJSON=} report
```
ebidel: Not anymore without viewer. Done.

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+   */
+  attach(report) {
```
ebidel: done

lighthouse-core/report/v2/renderer/report-features.js
ebidel: done

lighthouse-core/report/v2/renderer/report-features.js
ebidel: Disabled this in the UI for now. Left in the code.

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+        const jsonStr = JSON.stringify(this.json, null, 2);
+        this._saveFile(new Blob([jsonStr], {type: 'application/json'}));
+        break;
+      }
+      case 'save-html': {
+        let htmlStr = '';
+
+        // Since Viewer generates its page HTML dynamically from report JSON,
+        // run the ReportGenerator. For everything else, the page's HTML is
+        // already the final product.
```
ebidel: done

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+    if (e.keyCode === 27) { // ESC
+      this.closeExportDropdown();
+    }
+  }
+
+  /**
+   * Opens a new tab to the online viewer and sends the local page's JSON results
+   * to the online viewer using postMessage.
+   */
+  sendJSONReport() {
```
ebidel: left in

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+      this.expandDetailsWhenPrinting();
+    }
+  }
+
+  /**
+   * Expands audit `<details>` when the user prints the page.
+   * Ideally, a print stylesheet could take care of this, but CSS has no way to
+   * open a `<details>` element.
+   */
+  expandDetailsWhenPrinting() {
```
ebidel: Renamed and updated the selectors. 

The listeners need to be setup before printing happens. To further complicate things, there are two ways users can print a report: using "ctrl+p", and when click Export -> Print. The printing stuff is weird due to the browser compat. `window.addEventListener('beforeprint', this.expandAllDetails);` doesnt even work correctly in FF :(

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,272 @@
...
+    const reportContainer = this._document.querySelector('.lighthouse-categories');
+    const details = Array.from(reportContainer.querySelectorAll('details'));
+    details.map(detail => detail.open = true);
+  }
+
+  /**
+   * Sets up listeners to collapse audit `<details>` when the user closes the
+   * print dialog, all `<details>` are collapsed.
+   */
+  _setUpCollaspeDetailsAfterPrinting() {
```
ebidel: done

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,276 @@
...
+      window.matchMedia('print').addListener(mql => {
+        if (!mql.matches) {
+          details.map(detail => detail.open = mql.matches);
+        }
+      });
+    }
+  }
+  /**
+   * Downloads a file (blob) using a[download].
+   * @param {Blob|File} blob The file to save.
```
ebidel: done

lighthouse-core/report/v2/renderer/report-renderer.js
ebidel: Ha. Done.

lighthouse-core/report/v2/renderer/report-renderer.js
ebidel: Yea old code. Removed.

lighthouse-core/report/v2/renderer/report-renderer.js
ebidel: All of our exports in these browser files are just for testing.

lighthouse-core/report/v2/renderer/logger.js
```diff
@@ -0,0 +1,71 @@
+/**
```
ebidel: IDK. I'm just porting :)

Knowing the separation of LH proper and integrations of LH is still unclear. Right now the operating assumption is that we're keeping everything we currently have b/c https://github.com/GoogleChrome/lighthouse/issues/1937 doesn't capture these types of details.

For me, the dream is that there will be all sorts of existing tools that adopt a LH report. It's impossible to foresee all those integration touch points. Rather than designing our system around a particular _integration_ , it might be easier to flip it. IOW, the integration (e.g. devtools) decides what parts of LH it wants to throw away or replace. Removing handlebars, adding `templateContext` are a step towards that.

For the features stuff, I've updated it to take the approach in https://github.com/GoogleChrome/lighthouse/commit/b12f2ea70a4f786138c24730d6d569f77d26dbfc. Devtools can decide not to load the feature (including logger). 

---

Some open thoughts/questions for everyone. "**Integration X**" can be substituted for  "devtools".

- The core report should contain the features we want to show up in the most places. For example, if you run LH in the CLI/extension/WPT/Gulliver/LHaaS, you'll get the UI we vend (logger, export, etc.).
- What happens if you generate an html report in **Integration X** and open it later? Do you see LH's base UI or their version?
- What happens if you open that same report in Viewer?
- CSS is the easiest way to get rid of base UI: `#lh-log { display: none}`. We're already [doing that today](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/styles/report.css#L63).
   - Alternatively, **Integration X** loads its own stylesheets and effectively changes the entire report.
- If you also want to nuke code, don't inject `ReportUIFeature` ([here](https://github.com/GoogleChrome/lighthouse/pull/2000/files#diff-8423c2d032b8588d5009d6660287d634R40)), or stub files: `class Logger { log() {} }`. We could even provide those files for people if it makes sense.

ebidel:
**Update**- as discussed in today's meeting, this now includes our left nav with scores for each category. The left nav is not functional, but it takes us towards the devtools mocks. 

screen shot 2017-04-14 at 10 48 49 pm

patrickhulce:
rebased for ya @ebidel, biggest change was extracting the format helpers to a new file

paulirish:
Alright, brendan and I walked through the features and components in this PR and identified how we'll handle for the devtools case.

### ✅  Will include and use as-is in DevTools

* print report (w/ details expanding behavior)
* download HTML
* download JSON
* copy JSON to clipboard
* header
* header's runtime setting
* footer

### ✅  Will include, but subclass in DevTools

* **`_saveFile`** - DT has its own way to save files to disk, so we'll override for that.
* t**he Share menu component** - devtools has a native `UI.ContextMenu` component that we can use. If you expose this sort of API, we can maximize overlap:
```js
contextMenu.appendItem(menuItemTitleStr, handlerFn);
// ... later on ...
contextMenu.show();
```
I think this also means we don't have declarative menu items over in `.lh-export .export__dropdown` in the template. They should be added at runtime.

### ❌  Will not include in devtools right now

* **Share to viewer** - I'd like to defer adding this to devtools, as I don't want to walk through the security concerns right now. We could explore in 2 months. Can we separate out this into a new file? (Maybe then reportFeatures exposes a `registerFeature` method that will end up calling `contextMenu.appendItem` etc).
* **Logger toasts component** - Proposal: emit events from reportFeatures.. like dispatchEvent of [`CustomEvent`](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent) on the `.lh-report` dom element. The standalone logger component will addEventListener for `"lh-error"` events.. In DevTools, we'll also listen but just send to console. Hows that sound?
* **Google Analytics** - (the `ga.send` calls) - Same idea as above, but probably an event called `"lh-telemetry"` or whatever. [Looks like](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events) objects can be attached to the `detail` property of the event.

----------

#### actual action item summary (because so many words (sorry)):

* move all share-to-viewer stuff to separate file
* extract all logger mentions from report-features and adopt event-based API for logger
* remove `ga.send` mentions from report-features and dispatch events for those.
* refactor the `.lh-export` dropdown to add items at runtime.

aside from these bits, i'm happy with the rest and looking forward to some nice looks on the report! :)

paulirish:
> then no problem as static methods? It has the benefit of making the global and the export parallel, too. But again, testing implementation details isn't all that useful. Asserting that `calculateRating(10)` returns `'fail'` means that a refactor of how scoring is calculated has to update tests, even if the refactor maintains unchanged scoring thresholds in the actual report. The tests are already using `sampleResults`; would it be feasible to pull out the scores of a report generated from those and test for how they're marked?

I'm also supportive of putting this on as static methods.
I'd also prefer this vs adding the new format-helpers.js file. 🐈

brendankenny:
turning on closure checking produced a number of errors in here, some legitimate, some from just things like `document.execCommand`'s return type (apparently) being implementation specific, so the default externs don't have a type for them. I can help sort them out on Monday.

ebidel:
@brendankenny I need your help with the remaining closure issues. Casting didn't take care of `document.execCommand` or the `MessageEvent` stuff.

@paulirish 

> move all share-to-viewer stuff to separate file

I've added a few other methods that can be overridden. See if that works. See example. below.

Instead of worrying about potential (and future) security concerns, **Lighthouse should really just go through a full security review** before we launch DT. There are more questionable areas of the codebase than opening a tab and sending JSON to it :)

I think it's also good to keep features that have already landed in the project.  If you really want to omit features from DT, can the DT integration just remove them?

> extract all logger mentions from report-features and adopt event-based API for logger

Easy to do already. See example below.

> remove ga.send mentions from report-features and dispatch events for those.

Done! Although, kinda silly that something trivial has turned into something more complex. But hopefully we can use `_fireEventOn` for more stuff?

```
if (self.ga) {
  self.ga('send', 'event', 'report', 'copy');
}
```

```
/**
 * Fires a custom DOM event on target.
 * @param {string} name Name of the event.
 * @param {!Document|!Element} target DOM node to fire the event on.
 * @param {Object<{detail: Object<string, *>}>=} detail Custom data to include.
 */
_fireEventOn(name, target = this._document, detail) {
  const event = new CustomEvent(name, detail ? {detail} : null);
  this._document.dispatchEvent(event);
}

// Elsewhere in the app.
document.addEventListener('lh-analytics', e => {
  if (self.ga) {
    self.ga(e.detail.cmd, e.detail.fields);
  }
});
```

> refactor the .lh-export dropdown to add items at runtime.

In the example. Although,  the `UI.ContextMenu()` in devtools looks like it renders a system UI menu. Not sure if that's what you want or not.

screen shot 2017-04-25 at 8 47 21 pm

---

Ok, so here was my thinking for the feature stuff. Choosing what features go in and out of DT is going to be really adhoc. The code separation is only going to get more confusing. What I did with `ReportUIFeatures` was to design everything to be overridable. DT can override methods as it needs, or even swap in an entirely new implementation. Either way, everything gets DI'd into `ReportRenderer` and it JustWorks.

Proof is in the pudding. All of the tweaks suggested above can be done like so:

```
class ReportUIFeaturesDevTools extends ReportUIFeatures {
  /**
   * Routes Logger to console instead of the UI toast.
   * @override
   */
  _setupLogger() {
    this.logger = console;
    this.logger.hide = function() {};
  }

  // @override
  initFeatures(...args) {
    super.initFeatures(...args);
    // ... devtools adds more stuff at runtime ...
  }

  // @override
  _setupExportButton() {
    super._setupExportButton();

    // ... devtools uses its API to build the context menu.
    // const dropdown = this._dom.find('.lh-export__dropdown', this._document);
    // dropdown.textContent = '';
    // const menu = UI.ContextMenu();
    // menu.appendItem('Save HTML', this.);
    // menu.appendItem('Item 2', function handler {});
    // menu.appendItem('Item 3', function handler {});
  }

  // @override
  _saveFile(blob) {
    // ... devtools calls its own API ...
  }

  // @override
  sendJSONReport() {
    // super.sendJSONReport = null; // nuke it if you're paranoid.
    console.warn('DT: not implemented yet');
  }

  // @override
  onExport(e) {
    // ... devtools does its own thing ...
  }
}
```

patrickhulce:
hahaha these commits bring me so much joy 😂 
image

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+    this._resetUIState();
+    this._document.addEventListener('keydown', this.printShortCutDetect);
+    this._document.addEventListener('copy', this.onCopy);
+  }
+
+  /**
+   * Fires a custom DOM event on target.
+   * @param {string} name Name of the event.
+   * @param {!Document|!Element} target DOM node to fire the event on.
+   * @param {Object<{detail: Object<string, *>}>=} detail Custom data to include.
```
brendankenny: Not exactly sure where this is going, but from the usage in `onCopyButtonClick`, at least, this could be `@param {{cmd: string, fields: !Object<string, *>}=} detail`, but not sure if that's the intention

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -127,7 +127,7 @@ class DOM {
   /**
    * Helper for context.querySelectorAll. Returns an Array instead of a NodeList.
    * @param {string} query
-   * @param {!DocumentFragment|!Element} context
+   * @param {!Document|!DocumentFragment|!Element} context
    * @return {!Array<Element>}
```
brendankenny: should return `!Array<!Element>`

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+      e.clipboardData.setData('text/plain', JSON.stringify(this.json, null, 2));
+      this.logger.log('Report JSON copied to clipboard');
+    }
+
+    this._copyAttempt = false;
+  }
+
+  /**
+   * Copies the report JSON to the clipboard (if supported by the browser).
+   */
```
brendankenny: I don't think there's any hope of getting `this._document.queryCommandSupported` and `this._document.execCommand` to work, so you can add `@suppress {reportUnknownTypes}` to the jsdocs of this function to get it to shut up. The only downside is any other errors in the same category in this method will be silenced (none right now)

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+    e.preventDefault();
+    const el = /** @type {!Element} */ (e.target);
+    el.classList.toggle('active');
+    this._document.addEventListener('keydown', this.onKeyDown);
+  }
+
+  /**
+   * Resets the state of page before capturing the page for export.
+   * When the user opens the exported HTML page, certain UI elements should
+   * be in their open state (not opened) and the templates should be refreshly
```
brendankenny: > be in their open state (not opened)

yuh?

> refreshly stamped

freshly stamped?

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+   * stamped.
+   */
+  _resetUIState() {
+    this.logger.hide();
+    this.closeExportDropdown();
+    this._dom.findAll('template[data-stamped]', this._document).forEach(t => {
```
brendankenny: should this be in this class? It feels like it's a higher-level detail (it's hard to connect to other side of things in `dom.js`)

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+        break;
+      case 'save-json': {
+        const jsonStr = JSON.stringify(this.json, null, 2);
+        this._saveFile(new Blob([jsonStr], {type: 'application/json'}));
+        break;
+      }
+      case 'save-html': {
+        this._resetUIState();
+
+        const htmlStr = this._document.documentElement.outerHTML;
```
brendankenny: don't we want this to generate the report so it's just a copy of this report rather than a copy of its DOM?

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+  sendJSONReport() {
+    const VIEWER_ORIGIN = 'https://googlechrome.github.io';
+    const VIEWER_URL = `${VIEWER_ORIGIN}/lighthouse/viewer/`;
+
+    // Chrome doesn't allow us to immediately postMessage to a popup right
+    // after it's created. Normally, we could also listen for the popup window's
+    // load event, however it is cross-domain and won't fire. Instead, listen
+    // for a message from the target app saying "I'm open".
+    self.addEventListener('message', function msgHandler(/** @type {!Event} */ evt) {
+      const e = /** @type {!MessageEvent} */ (evt);
```
brendankenny: `MessageEvent` is a [templated type](https://github.com/google/closure-compiler/blob/0a2b4d12fa3f30cfe4d9a184f3c0cc3e349313d4/externs/browser/html5.js#L2168-L2174), so you can type the `data` property on it in the type declaration. e.g. `/** @type {!MessageEvent<{opened: boolean}>} */`

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+
+      if (e.origin !== VIEWER_ORIGIN) {
+        return;
+      }
+
+      if (e.data['opened']) {
```
brendankenny: surprised eslint isn't complaining here about not doing `e.data.opened`

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+  _saveFile(blob) {
+    const filename = self.getFilenamePrefix({
+      url: this.json.url,
+      generatedTime: this.json.generatedTime
+    });
+
+    const ext = blob.type.match('json') ? '.json' : '.html';
+    const href = URL.createObjectURL(blob);
+
+    const a = this._dom.createElement('a');
```
brendankenny: might want to do `const a = /** @type {!HTMLAnchorElement} */ (this._dom.createElement('a'));` so closure knows what this is up to

lighthouse-core/report/v2/renderer/report-features.js
```diff
@@ -0,0 +1,326 @@
...
+  }
+}
+
+if (typeof module !== 'undefined' && module.exports) {
+  module.exports = ReportUIFeatures;
+} else {
+  self.ReportUIFeatures = ReportUIFeatures;
+
+  /** @type {function({url: string, generatedTime: string}): string} */
+  self.getFilenamePrefix; // eslint-disable-line no-unused-expressions
```
brendankenny: Isn't this already declared in the global scope?

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -22,30 +22,48 @@
...
+      if (this._uiFeatures) {
+        this._uiFeatures.initFeatures(report, container);
+      }
     } catch (/** @type {!Error} */ e) {
-      return this._renderException(e);
+      container.textContent = '';
+      element = container.appendChild(this._renderException(e));
     }
+
+    return /** @type {!Element} **/ (element);
```
brendankenny: can also split into `element = this._renderReport(report); container.appendChild(element);` lines if you don't want to cast

brendankenny: actually, does this even need to return anything anymore? Return value isn't being used in `report-template.html` anymroe, and if adding to the container now, not even needed for testing

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -69,17 +88,91 @@ class ReportRenderer {
...
+  _renderReportHeader(report) {
+    const header = this._dom.cloneTemplate('#tmpl-lh-heading', this._templateContext);
+    this._dom.find('.lh-config__timestamp', header).textContent =
+        formatDateTime(report.generatedTime);
+    const url = this._dom.find('.lh-metadata__url', header);
+    url.href = report.url;
+    url.textContent = report.url;
+
+    const env = this._dom.find('.lh-env__items', header);
+    report.runtimeConfig.environment.forEach(runtime => {
```
brendankenny: s/runtime/env, maybe? (would need to rename `env` above to `envContainer` or whatever)

lighthouse-core/report/v2/renderer/report-features.js
brendankenny: oh, I see the issue, `file-namer.js` needs to be added to the closure compiled files list (and should use this one's `{url: string, generatedTime: string}` instead of `Object` like in that file)

lighthouse-core/report/v2/report-template.html
```diff
@@ -27,15 +27,28 @@
...
+      const categoryRenderer = new CategoryRenderer(dom, detailsRenderer);
+      const features = new ReportUIFeatures(dom);
+      const renderer = new ReportRenderer(dom, categoryRenderer, features);
+
+      const container = document.querySelector('main');
+      renderer.renderReport(window.__LIGHTHOUSE_JSON__, container);
+    });
+
+    document.addEventListener('lh-analytics', e => {
+      if (self.ga) {
```
brendankenny: no need for `self`, I don't think

lighthouse-core/closure/third_party/google_universal_analytics_api.js
```diff
@@ -0,0 +1,80 @@
...
+ * @externs
+ */
+
+
+/**
+ * @param {string} methodName
+ * @param {...?} var_args
+ * @return {?}
+ */
+function ga(methodName, var_args) {}
```
brendankenny: I don't think putting functions on this function will actually work. Is the rest of this file needed right now?

Definitely should avoid unknown type and do `...*` for `var_args` and no return type if it's not needed. Right now the second argument is being used as `!Object<string, *>` and not any more arguments after that. Is that better?

brendankenny:
we definitely need some sanity around dependency injection vs overriding. There's basically no help right now on how to use stuff or how stuff works in these report files except reading all of them. Things that are injected need stuff that's overridden and vice versa, so some clarity with very clear decoupling boundaries would be amazing

brendankenny:
(that would be through a group effort, not expecting you to solve this eric :)

brendankenny:
> > move all share-to-viewer stuff to separate file

> I've added a few other methods that can be overridden. See if that works. See example. below.

> Instead of worrying about potential (and future) security concerns, Lighthouse should really just go through a full security review before we launch DT. There are more questionable areas of the codebase than opening a tab and sending JSON to it :)

even easier would be https://github.com/GoogleChrome/lighthouse/pull/2000#discussion_r111712249 :) viewer code doesn't work, it's self contained, and can easily be added once we figure out viewer and get it working

lighthouse-core/closure/third_party/google_universal_analytics_api.js
ebidel: I took this entire thing from closure's common externs.

lighthouse-core/report/v2/renderer/report-features.js
ebidel: Unfortunately, we need the whole page (outer shell, templates, styles, etc.) when exporting to html. [Same as before.](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/scripts/lighthouse-report.js#L147).

brendankenny: can't we just include [`generateReportHtml`](https://github.com/GoogleChrome/lighthouse/blob/e46244d64030c335e84f7b73f8e8a883d60328bd/lighthouse-core/report/v2/report-generator.js#L130-L140) (with `REPORT_JAVASCRIPT`, `REPORT_CSS`, and `REPORT_TEMPLATES` inlined) and use that to generate the thing we save? We just need to make sure we refer to things by reference at one point so we don't infinitely recurse :)

lighthouse-core/report/v2/renderer/format-helpers.js
```diff
@@ -0,0 +1,76 @@
...
+  const tz = formatter.resolvedOptions().timeZone;
+  if (!tz || tz.toLowerCase() === 'etc/unknown') {
+    options.timeZone = 'UTC';
+    formatter = new Intl.DateTimeFormat('en-US', options);
+  }
+  return formatter.format(new Date(date));
+}
+
+if (typeof module !== 'undefined' && module.exports) {
+  module.exports = {
```
ebidel: Hmm, even when I uncomment that line the compilation succeeds. Although I don't seem to get a closure-output.js file

brendankenny: > Hmm, even when I uncomment that line the compilation succeeds

er, you're right. I guess ignore this :) not sure what changed

> Although I don't seem to get a closure-output.js file

you have to set `PRINT_CODE` to true to get the output. Normally it's just checking, not outputting the compiled code

ebidel:
Changes are in and we're 💚 . 

Dropping the mic on this one 🎤 folks. Any other change, ya'll do in followup prs.

paulirish:
merging to unblock the merge of other stuff. 
thanks for the 2w slog on this monster, everyone. 
💟

Issue #1981, opened 2017-04-07
Improve lighthouse results `extendedInfo` structure guarantee


wwwillchen:
Lighthouse results `extendedInfo` property tree is sometimes not defined. I came across this issue when working on plots, and I created a `safeGet` function as a workaround: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/traces/pwmetrics-events.js.

As @brendankenny is working on re-enabling closure compiler, we'll want to have a less hacky way of dealing this issue and still be type-checking friendly.

paulirish:
dupe of #2165 basicallly

Issue #489, opened 2016-06-29
Properly account for non-top-level `toplevel` tasks for estimating input latency


brendankenny:
@skyostil says:

"The Blink scheduler runs several tasks within a single message loop iteration to reduce the overhead of going in and out of the loop. Here's an example:

![1d4d6c54-405a-4432-81f1-1acef5f731fb](https://cloud.githubusercontent.com/assets/316891/16472071/30cba0e0-3e16-11e6-9464-4bf47a177664.png)

There's just one `MessageLoop::RunTask` at the top, but three `TaskQueueManager::ProcessTaskFromWorkQueue`, i.e., three actual tasks that ran.

To avoid hardcoding too much I suggest you look for the innermost slices with the "toplevel" category. This should land you at `ProcessTaskFromWorkQueue` on the main thread and should also work with Lucky Luke on the browser side....

Input can sneak in at any time between `ProcessTaskFromWorkQueue`-- no need for it to go via the message loop."

patrickhulce:
should be addressed by #2013

PR #2013, opened 2017-04-14
feat: add firstInteractive

patrickhulce:
Adds firstInteractive as defined [here](https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#). Also changes all references from TTI to first-interactive. Removing TTI and replacing it with visually ready is in a different pending PR (currently a blocking bug in speedline I'm investigating)


lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -88,7 +88,10 @@ class TraceOfTab extends ComputedArtifact {
       firstMeaningfulPaint = lastCandidate;
     }
 
-    const onLoad = keyEvents.find(e => e.name === 'MarkLoad' && e.ts > navigationStart.ts);
+    const onLoad = frameEvents.find(e => e.name === 'loadEventStart' && e.ts > navigationStart.ts);
```
patrickhulce: @paulirish `MarkLoad` was not onload in most cases I saw during my investigations, what's supposed to be the difference between it and loadEventStart?

paulirish: markload should actually line up with loadEventEnd. if its not, then it must be selecting the wrong frame.

loadEventEnd is the dfn we want to use, so let's adjust to that instead.

lighthouse-core/lib/traces/pwmetrics-events.js
```diff
@@ -131,39 +131,15 @@ class Metrics {
         }
       },
       {
-        name: 'Time to Interactive (vAlpha)',
-        id: 'tti',
+        name: 'First Interactive (vAlpha)',
```
paulirish: vBeta 😎

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -38,18 +38,21 @@ class ComputedArtifact {
 
   /**
    * Request a computed artifact, caching the result on the input artifact.
-   * @param {!OBject} artifact
+   * @param {!Object} artifact
+   * @param {!Object=} artifacts
    * @return {!Promise}
    */
-  request(artifact) {
+  request(artifact, artifacts) {
```
paulirish: pretty interesting, but I like it. leads to more clean implementations.

can you rename to `allArtifacts` to clarify, since we have 3 diff sorts in here.

lighthouse-core/audits/first-interactive.js
```diff
@@ -0,0 +1,76 @@
...
+
+'use strict';
+
+const Audit = require('./audit');
+const TracingProcessor = require('../lib/traces/tracing-processor');
+const Formatter = require('../report/formatter');
+
+// Parameters (in ms) for log-normal CDF scoring. To see the curve:
+//   https://www.desmos.com/calculator/uti67afozh
+const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
```
paulirish: presuming these are spitballed?

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -63,20 +62,21 @@ class LoadFastEnough4Pwa extends Audit {
...
-      const isFast = timeToInteractive < MAXIMUM_TTI;
+    const trace = artifacts.traces[Audit.DEFAULT_PASS];
+    return artifacts.requestFirstInteractive(trace, artifacts).then(firstInteractive => {
+      const ttfiValue = firstInteractive.timeInMs;
+      const isFast = ttfiValue < MAXIMUM_TTI;
 
       const extendedInfo = {
         formatter: Formatter.SUPPORTED_FORMATS.NULL,
-        value: {areLatenciesAll3G, allRequestLatencies, isFast, timeToInteractive}
+        value: {areLatenciesAll3G, allRequestLatencies, isFast, ttfiValue}
```
paulirish: let's call it the full `timeToFirstInteractive`

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,163 @@
...
+const TracingProcessor = require('../../lib/traces/tracing-processor');
+
+const LONELY_TASK_FMP_DISTANCE = 5000;
+const LONELY_TASK_ENVELOPE_SIZE = 250;
+const LONELY_TASK_NEIGHBOR_DISTANCE = 1000;
+
+const MAX_QUIET_WINDOW_SIZE = 5000;
+const TRACE_BUSY_MSG = 'trace was busy the entire time';
+
+class FirstInteractive extends ComputedArtifact {
```
paulirish: lets add a fileoverview or other description that links to this doc and briefly describes wtf this does
https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -88,7 +88,10 @@ class TraceOfTab extends ComputedArtifact {
...
+    const domContentLoaded = frameEvents.find(
+      e => e.name === 'domContentLoadedEventStart' && e.ts > navigationStart.ts
```
paulirish: for both this and `loadEvent`, usually the `End` is used, but in this case, it actually makes more sense for us to focus on the browser's POV of these events... so 👍  on using `Start`

lighthouse-core/gather/computed/computed-artifact.js
brendankenny: the problem is that it breaks the cache and `requiredArtifacts` dependency checks, since the computed artifact could theoretically access any artifact it wanted at this point. In this case it's only accessing other computed artifacts, so it's fine, but we can't guarantee it.

I have a proposal I've been thinking about for a while (it also makes typing nicer for the artifacts object). I'll throw it together quickly and put it in a PR

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: oh so it is fired per frame afterall then 👍

lighthouse-core/gather/computed/computed-artifact.js
patrickhulce: Does this proposal allow for passing more than one artifact? I wanted to make consistently interactive a computed artifact too, but it needs trace + networkRecords

lighthouse-core/audits/first-interactive.js
patrickhulce: yeah I did it based on my hypothetical grading rubric from A+ -> F, each letter gets 20 point range with the median (C) being "passing" but not good. By this scale, A+ is basically the point of diminishing returns and A- extends to around 4 seconds.

paulirish:
FYi I just finished a branch that moves networkRecords into a computed
artifact. Shouldn't matter much for this though.

lighthouse-core/gather/computed/computed-artifact.js
brendankenny: > Does this proposal allow for passing more than one artifact?

No, but we definitely need that too. I can add it in a followup PR (or you can if you want). Basically computedArtifact's `_cache` needs to support a tuple as a key instead of just objects directly. There's no really efficient way to do that in JS today, but luckily these caches all have like one item in them :)

brendankenny:
> FYi I just finished a branch that moves networkRecords into a computed artifact. Shouldn't matter much for this though.

I'm not sure how this would work, so interested in seeing it :)

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: 👍  moved it up from compute_ and added some detail

lighthouse-core/lib/traces/pwmetrics-events.js
patrickhulce: done

patrickhulce:
rebased PTAL :)

lighthouse-core/gather/computed/trace-of-tab.js
paulirish: Actually......

The doc uses dclEnd rather than dclStart.
Since it's used as `fi = Math.max(DCL, FI)`... the End value seems much more accurate. (Handlers are setup during DCL, and the main thread is unavailable during this time)

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+    const FMP = traceOfTab.timings.firstMeaningfulPaint;
+    const DCL = traceOfTab.timings.domContentLoaded;
+    const traceEnd = traceOfTab.timings.traceEnd;
+
+    if (traceEnd - FMP < MAX_QUIET_WINDOW_SIZE) {
+      throw new Error('trace not at least 5 seconds longer than FMP');
+    }
+
+    const longTasks = TracingProcessor.getMainThreadTopLevelEvents(traceModel, trace, FMP)
+        .filter(evt => evt.end - evt.start >= 50 && evt.end > FMP);
```
paulirish: extract 50 to LONG_TASKS_THRESHOLD const

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+    const navStart = traceOfTab.timestamps.navigationStart;
+    const FMP = traceOfTab.timings.firstMeaningfulPaint;
+    const DCL = traceOfTab.timings.domContentLoaded;
+    const traceEnd = traceOfTab.timings.traceEnd;
+
+    if (traceEnd - FMP < MAX_QUIET_WINDOW_SIZE) {
+      throw new Error('trace not at least 5 seconds longer than FMP');
+    }
+
+    const longTasks = TracingProcessor.getMainThreadTopLevelEvents(traceModel, trace, FMP)
```
paulirish: rename to longTasksAfterFMP

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -246,6 +235,28 @@ class TraceProcessor {
...
+   * @param {number=} startTime Optional start time (in ms) of range of interest. Defaults to trace start.
+   * @param {number=} endTime Optional end time (in ms) of range of interest. Defaults to trace end.
+   * @return {!Array<{start: number, end: number}>}
+   */
+  static getMainThreadTopLevelEvents(model, trace, startTime = 0, endTime = Infinity) {
+    // Find the main thread via the first TracingStartedInPage event in the trace
+    const startEvent = trace.traceEvents.find(event => event.name === 'TracingStartedInPage');
+    const mainThread = TraceProcessor._findMainThreadFromIds(model, startEvent.pid, startEvent.tid);
+
+    // TODO(bckenny): filter for top level slices ourselves?
```
paulirish: can we fix our longstanding #489 while you're here?

```js
// See https://github.com/GoogleChrome/lighthouse/issues/489
const schedulableTaskTitle = 'TaskQueueManager::ProcessTaskFromWorkQueue';

const events = mainThread.sliceGroup.slices.filter(slice => {
  return slice.title === schedulableTaskTitle && 
      slice.end > startTime && 
      slice.start < endTime;
});
```

Just tested this guy out and he's good. This also matches what we see in https://codereview.chromium.org/2610183003/patch/140001/150003

lighthouse-core/test/lib/traces/pwmetrics-events-test.js
```diff
@@ -26,7 +26,7 @@ const dbwResults = require('../../fixtures/dbw_tester-perf-results.json');
 /* eslint-env mocha */
 describe('metrics events class', () => {
   it('exposes metric definitions', () => {
-    assert.equal(Metrics.metricsDefinitions.length, 12, '12 metrics not exposed');
+    assert.equal(Metrics.metricsDefinitions.length, 10, '10 metrics not exposed');
```
paulirish: feel free to change this test. looks like it's getting annoying for you. :)

lighthouse-core/lib/traces/pwmetrics-events.js
```diff
@@ -131,39 +131,15 @@ class Metrics {
         }
       },
       {
-        name: 'Time to Interactive (vAlpha)',
```
paulirish: let's keep TTI in here for now. 
i'd like them to live side-by-side for a while for folks to get comfortable.
then we can drop.

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+
+const MAX_QUIET_WINDOW_SIZE = 5000;
+const TRACE_BUSY_MSG = 'trace was busy the entire time';
+
+/**
+ * @fileoverview This artifact identifies the time the page is "first interactive" as defined below
+ * @see https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#
+ *
+ * First Interactive marks the first moment when a website is minimally interactive:
+ *    > Enough (but maybe not all) UI components shown on the screen are interactive
```
brendankenny: is including this without a disclaimer misleading since this is the goal of the heuristic but is not actually tested here? If I were reading this I would be scanning the code for where event listeners are instrumented.

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * @param {!Array<{start: number, end: number>}} longTasks
+   * @param {number} i
+   * @param {number} FMP
+   * @param {number} traceEnd
+   * @return {number} The last index of the lonely task envelope,
+   *    -1 if not a lonely task
+   */
+  static getLastLonelyTaskIndex(longTasks, i, FMP, traceEnd) {
```
paulirish: Over in #2023  you're using a candidate + while loop pattern that I find considerably easier to follow the logic of.

Is it possible to use the same pattern here?

take sorted list of tasks, lastLonelyCandidate is the first, `while (lastLonelyCandidate.start < windowEnd)`... kind of a thing..

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+ * First Interactive marks the first moment when a website is minimally interactive:
+ *    > Enough (but maybe not all) UI components shown on the screen are interactive
+ *    > The page responds to user input in a reasonable time on average, but it’s ok if this
+ *      response is not always immediate.
+ *
+ * First Interactive is defined as the first period after FMP of N-seconds where there are no
+ * long tasks that are not "lonely".
+ *
+ *    > t = time in seconds since FMP
+ *    > N = 4 * e^(-0.045 * t) + 1
```
brendankenny: they're in the doc, but example values (and/or desmos curve) would be good here

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: yeah originally I was using DCL as the start point for search when using start made some sense, but now that it's just the actual firstInteractive point, using end is much better

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/lib/traces/pwmetrics-events.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+ *    > a "lonely" task is an envelope of 250ms that contains a set of long tasks that have at least
+ *      1 second of padding before and after the envelope that contain no long tasks.
```
brendankenny: should include the FMP + 5s def?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+      if (isQuiet) {
+        return windowStart;
+      }
+    }
+
+    throw new Error(TRACE_BUSY_MSG);
+  }
+
+  /**
+   * @param {!Object} trace
```
brendankenny: should type these. `Object` essentially disables typing since it's the base class for all classes

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -246,6 +235,28 @@ class TraceProcessor {
   }
 
   /**
+   * Provides the top level events on the main thread.
+   * @param {!traceviewer.Model} model
+   * @param {{traceEvents: !Array<!Object>}} trace
+   * @param {number=} startTime Optional start time (in ms) of range of interest. Defaults to trace start.
+   * @param {number=} endTime Optional end time (in ms) of range of interest. Defaults to trace end.
+   * @return {!Array<{start: number, end: number}>}
```
brendankenny: `slice.duration` is used in `getMainThreadTopLevelEventDurations`, not sure if there's anything else needed on these objects

lighthouse-core/test/lib/traces/pwmetrics-events-test.js
patrickhulce: done, yes it was :)

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -205,23 +205,12 @@ class TraceProcessor {
...
     const durations = [];
     let clippedLength = 0;
-    mainThread.sliceGroup.topLevelSlices.forEach(slice => {
-      // Discard slices outside range.
-
-      if (slice.end <= startTime || slice.start >= endTime) {
-        return;
-      }
-
+    slices.forEach(slice => {
```
brendankenny: you're harshing EQT perf here :S

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -246,6 +235,28 @@ class TraceProcessor {
...
+    // Find the main thread via the first TracingStartedInPage event in the trace
+    const startEvent = trace.traceEvents.find(event => event.name === 'TracingStartedInPage');
+    const mainThread = TraceProcessor._findMainThreadFromIds(model, startEvent.pid, startEvent.tid);
+
+    // TODO(bckenny): filter for top level slices ourselves?
+    const events = mainThread.sliceGroup.topLevelSlices.filter(slice => {
+      return slice.end > startTime && slice.start < endTime;
+    });
+
+    events.sort((a, b) => a.start - b.start);
```
brendankenny: maybe at least sort in caller so `getMainThreadTopLevelEventDurations` doesn't sort twice?

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: for just `getLastLonelyTaskIndex` or the whole thing? The trickier piece here that makes everything annoying is the complicated base case differentiation. I planned to fix that by treating the entire span before FMP as one giant long task which will make the base case the same as considering the first long task, but not sure if I should do in this PR. If it's too hard to grok right now I can do that switch now instead of a follow-up, wdyt?

FWIW, I totally agree it's not nearly as easy to follow, but at least a tiny bit that's because lonely tasks are way harder to explain that simple quiet windows too :)

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -246,6 +235,28 @@ class TraceProcessor {
...
 
   /**
+   * Provides the top level events on the main thread.
+   * @param {!traceviewer.Model} model
+   * @param {{traceEvents: !Array<!Object>}} trace
+   * @param {number=} startTime Optional start time (in ms) of range of interest. Defaults to trace start.
+   * @param {number=} endTime Optional end time (in ms) of range of interest. Defaults to trace end.
+   * @return {!Array<{start: number, end: number}>}
+   */
+  static getMainThreadTopLevelEvents(model, trace, startTime = 0, endTime = Infinity) {
```
brendankenny: can also use `model.bounds.min` and `model.bounds.max` for defaults here for slightly more self-evident values

lighthouse-core/gather/computed/first-interactive.js
brendankenny: this is difficult to follow right now. "getLastLonelyTaskIndex" is kind of inscrutable...is it looking for the last one in the same envelope as the one at index `i`?

> I planned to fix that by treating the entire span before FMP as one giant long task which will make the base case the same as considering the first long task

from this description, at least, this sounds like a generalization that will make the code harder to follow, but the proof is in the implementation :) If there are three(?) different zones (pre-FMP, FMP+5s, post-FMP+5s), what about explicitly handling them individually?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,185 @@
...
+ *      response is not always immediate.
+ *
+ * First Interactive is defined as the first period after FMP of N-seconds where there are no
+ * long tasks that are not "lonely".
+ *
+ *    > t = time in seconds since FMP
+ *    > N = 4 * e^(-0.045 * t) + 1
+ *    > a "lonely" task is an envelope of 250ms that contains a set of long tasks that have at least
+ *      1 second of padding before and after the envelope that contain no long tasks.
+ */
```
brendankenny: also add `DOMContentLoaded` bound?

lighthouse-core/audits/first-interactive.js
```diff
@@ -0,0 +1,76 @@
...
+const Audit = require('./audit');
+const TracingProcessor = require('../lib/traces/tracing-processor');
+const Formatter = require('../report/formatter');
+
+// Parameters (in ms) for log-normal CDF scoring. To see the curve:
+//   https://www.desmos.com/calculator/rjp0lbit8y
+const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
+const SCORING_MEDIAN = 10000;
+// This aligns with the external TTI targets in https://goo.gl/yXqxpL
+const SCORING_TARGET = 5000;
```
patrickhulce: +1 I'll remove

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: oh puh-lease if we cared about perf here it wouldn't have taken months to realize the trace model was being constructed *multiple* times let alone that TTI runs this forEach, filter, and sort about 100 times more than necessary :P

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: hm, that actually feels a little less self-evident to me as someone who doesn't know the shape and meaning of the model properties well, we also wouldn't be able to specify them in the function signature anymore

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: added duration, or I could switch it to `!Object` if you prefer :)

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: image

😆  sorry I had to, seriously though it'd be nice to have a conversation about what we actually care about typing and what we don't, these are all computed artifacts that do not make sense writing one-off types for in every place that uses them but also would give no value to the developer as documentation if they were typed somewhere else and the type was just `!Trace` or something equally useless that just matched the name of the parameter.

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: added disclaimer

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -63,20 +62,21 @@ class LoadFastEnough4Pwa extends Audit {
     const areLatenciesAll3G = allRequestLatencies.every(val =>
         val === undefined || val > latency3gMin);
 
-    return TTIMetric.audit(artifacts).then(ttiResult => {
-      const timeToInteractive = ttiResult.extendedInfo.value.timings.timeToInteractive;
-      const isFast = timeToInteractive < MAXIMUM_TTI;
+    const trace = artifacts.traces[Audit.DEFAULT_PASS];
+    return artifacts.requestFirstInteractive(trace).then(firstInteractive => {
+      const timeToFirstInteractive = firstInteractive.timeInMs;
+      const isFast = timeToFirstInteractive < MAXIMUM_TTI;
```
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: yeah I'll work on rewriting it

lighthouse-core/gather/computed/first-interactive.js
brendankenny: the plan is still to get it going everywhere, so effort here pays off later. I personally think as documentation it still has value (certainly more value than a bunch of `Object`s, might as well not include them at that point) and besides, for 

> these are all computed artifacts that do not make sense writing one-off types for in every place that uses them

someone already has you [covered](https://github.com/GoogleChrome/lighthouse/blob/c153f625bbe855d6795103ea54a38699048e8e7a/lighthouse-core/closure/typedefs/ComputedArtifacts.js) :)

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: > oh puh-lease if we cared about perf here it wouldn't have taken months to realize the trace model was being constructed multiple times let alone that TTI runs this forEach, filter, and sort about 100 times more than necessary :P

I didn't say anything about TTI :)

patrickhulce: touché

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: > hm, that actually feels a little less self-evident to me as someone who doesn't know the shape and meaning of the model properties well

`-Infinity` for `startTime`, then?

> we also wouldn't be able to specify them in the function signature anymore

that would actually work. You can refer to earlier params in default parameters

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: alright, alright you got me, done

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: > -Infinity for startTime, then?

👍 

> that would actually work. 

TIL 
```
function noway({x, y} = {x: 3, y: 3}, woah = x) {
  console.log('boom', woah)
}
> noway()
boom 3
```

patrickhulce:
PTAL :)

patrickhulce:
PTAL :)

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+   * @param {number} windowEnd
+   * @return {!Array<{start: number, end: number, duration: number}>}
+   */
+  static getTaskGroupsInWindow(tasks, startIndex, windowEnd) {
+    const groups = [];
+
+    let lastEndTime = -Infinity;
+    let currentGroup = null;
+
+    // consider all tasks that could possibly be part of a group starting before windowEnd
```
paulirish: is the group starting before windowEnd or the tasks?
also here it's called `considerationWindowEnd`

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+    const considerationWindowEnd = windowEnd + MIN_TASK_GROUP_PADDING;
+    const couldTaskBeInGroup = task => task && task.start < considerationWindowEnd;
```
paulirish: `couldTaskBelongToGroup`?  unless tasks can be in multiple groups

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+    return (4 * exponentiationComponent + 1) * 1000;
+  }
+
+  /**
+   * @param {!Array<{start: number, end: number}>} tasks
+   * @param {number} startIndex
+   * @param {number} windowEnd
+   * @return {!Array<{start: number, end: number, duration: number}>}
+   */
+  static getTaskGroupsInWindow(tasks, startIndex, windowEnd) {
```
paulirish: another naming bikeshed. wdyt about `Cluster`/`TaskCluster` rather than group? Implies the proximity that's part of this..

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+    let lastEndTime = -Infinity;
+    let currentGroup = null;
+
+    // consider all tasks that could possibly be part of a group starting before windowEnd
+    const considerationWindowEnd = windowEnd + MIN_TASK_GROUP_PADDING;
+    const couldTaskBeInGroup = task => task && task.start < considerationWindowEnd;
+    for (let i = startIndex; couldTaskBeInGroup(tasks[i]); i++) {
+      const task = tasks[i];
+
+      // check if enough time has passed to start a new group
```
paulirish: > if enough time has elapsed, we'll create a new group

or cluster

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+      const windowEnd = windowStart + windowSize;
+
+      // Check that we have a long enough trace
+      if (windowEnd > traceEnd) {
+        throw new Error(TRACE_BUSY_MSG);
+      }
+
+      const isTooCloseToFMP = group => group.start < FMP + MIN_TASK_GROUP_FMP_DISTANCE;
+      const isTooLong = group => group.duration > MAX_TASK_GROUP_DURATION;
+      const isBadTaskGroup = group => isTooCloseToFMP(group) || isTooLong(group);
```
paulirish: you could de-extract this guy. seems better to inline him in the .find() below.

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * @param {!Array<{start: number, end: number}>} tasks
+   * @param {number} startIndex
+   * @param {number} windowEnd
+   * @return {!Array<{start: number, end: number, duration: number}>}
+   */
+  static getTaskGroupsInWindow(tasks, startIndex, windowEnd) {
+    const groups = [];
+
+    let lastEndTime = -Infinity;
```
paulirish: previousTaskEndTime

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: yeah I like cluster more, 👍

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: we need to consider all tasks before `considerationWindowEnd` but we only want clusters that start before `windowEnd`

consider the case where window end is 15s, there's a 100ms task from 14.9-15s and a 500ms task from 15.5-16s, we need that later task to be clustered with the first so we can properly identify that main thread isn't quiet, but if that 100ms task weren't there we want to ignore it. Make sense? Not sure how to condense that explanation into a better comment though suggestions welcome :)

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: I agree it's a bit much and I started with inlining when my names were shorter, but inlining now creates a linebreak and seems marginally better to keep the two conditions on the same line and keep each item a one-liner. still prefer a two-line inliner over this?

lighthouse-core/gather/computed/first-interactive.js
paulirish: would you consider `considerationWindowEnd` to be a `clusteringWindowEnd` ?

honestly i think you should just use the two sentences you just said ^ as the comment. :)

lighthouse-core/gather/computed/first-interactive.js
paulirish: eh. not a huge deal to me.

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: haha alright sg

patrickhulce:
alright @brendankenny all you :)

patrickhulce:
@brendankenny 🐜 ⛰ 🐒 🐈 🐩

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+      if (windowEnd > traceEnd) {
+        throw new Error(TRACE_BUSY_MSG);
+      }
+
+      const isTooCloseToFMP = cluster => cluster.start < FMP + MIN_TASK_CLUSTER_FMP_DISTANCE;
+      const isTooLong = cluster => cluster.duration > MAX_TASK_CLUSTER_DURATION;
+      const isBadCluster = cluster => isTooCloseToFMP(cluster) || isTooLong(cluster);
+
+      const taskClusters = FirstInteractive.getTaskClustersInWindow(longTasks, i + 1, windowEnd);
+      const hasBadTaskClusters = taskClusters.find(isBadCluster);
```
brendankenny: nit: `taskClusters.some(isBadCluster)` is probably more idiomatic

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+ * First Interactive is defined as the first period after FMP of N-seconds that has no bad task
+ * clusters.
+ *
+ *    > t = time in seconds since FMP
+ *    > N = f(t) = 4 * e^(-0.045 * t) + 1
+ *      5 = f(0) = 4 + 1
+ *      3 ~= f(15) ~= 2 + 1
+ *      1 ~= f(∞) ~= 0 + 1
+ *    > a "bad task cluster" is a cluster of 1 or more long tasks with no more than 1s of idle time
+ *      between each task that does one of the following
```
brendankenny: `with no more than 1s of idle time between each task` is kind of confusing... probably even `with less than 1s of idle time` would be better (depending on how loose you want to play with the openness of the interval :)

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+   * @return {number}
+   */
+  static getRequiredWindowSizeInMs(t) {
+    const tInSeconds = t / 1000;
+    const exponentiationComponent = Math.exp(-.045 * tInSeconds);
+    return (4 * exponentiationComponent + 1) * 1000;
+  }
+
+  /**
+   * @param {!Array<{start: number, end: number}>} tasks
```
brendankenny: jsdoc method description would be 💯

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    return 'FirstInteractive';
+  }
+
+  /**
+   * @param {number} t The time passed since FMP in miliseconds.
+   * @return {number}
+   */
+  static getRequiredWindowSizeInMs(t) {
+    const tInSeconds = t / 1000;
+    const exponentiationComponent = Math.exp(-.045 * tInSeconds);
```
brendankenny: we should probably use the canonical `-Math.log(2) / 15` instead of `-.045` since we're going to want to compare against other implementations. The difference is small but will definitely be noticeable when e.g. comparing analysis of the same saved trace

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+
+    let previousTaskEndTime = -Infinity;
+    let currentCluster = null;
+
+    // Examine all tasks that could possibly be part of a cluster starting before windowEnd.
+    // Consider the case where window end is 15s, there's a 100ms task from 14.9-15s and a 500ms
+    // task from 15.5-16s, we need that later task to be clustered with the first so we can properly
+    // identify that main thread isn't quiet.
+    const clusteringWindowEnd = windowEnd + MIN_TASK_CLUSTER_PADDING;
+    const couldTaskBelongToCluster = task => task && task.start < clusteringWindowEnd;
```
brendankenny: nit: maybe `isInClusteringWindow`? "could" is a low bar for descriptiveness :)

brendankenny: I've done zero performance testing, so I don't have a good basis for worrying, but *if* calculating FirstInteractive is slow for some sites, doing out-of-bounds array accessing really isn't going to help things. What about something like an explicit regular `i < tasks.length` here and then do the `if (!couldTaskBelongToCluster) break;` inside the loop?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    for (let i = startIndex; couldTaskBelongToCluster(tasks[i]); i++) {
+      const task = tasks[i];
+
+      // if enough time has elapsed, we'll create a new cluster
+      if (task.start - previousTaskEndTime > MIN_TASK_CLUSTER_PADDING) {
+        if (currentCluster) {
```
brendankenny: can you instead push upon creation of the cluster array or is that going to miss one the other way?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    return clusters
+      // add some useful information about the cluster
+      .map(tasks => {
+        const start = tasks[0].start;
+        const end = tasks[tasks.length - 1].end;
+        const duration = end - start;
+        return {start, end, duration, tasks};
+      })
+      // filter out clusters that started after the window because of our clusteringWindowEnd
+      .filter(cluster => cluster.start < windowEnd);
```
brendankenny: ah, nice

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    // Consider the case where window end is 15s, there's a 100ms task from 14.9-15s and a 500ms
+    // task from 15.5-16s, we need that later task to be clustered with the first so we can properly
+    // identify that main thread isn't quiet.
+    const clusteringWindowEnd = windowEnd + MIN_TASK_CLUSTER_PADDING;
+    const couldTaskBelongToCluster = task => task && task.start < clusteringWindowEnd;
+    for (let i = startIndex; couldTaskBelongToCluster(tasks[i]); i++) {
+      const task = tasks[i];
+
+      // if enough time has elapsed, we'll create a new cluster
+      if (task.start - previousTaskEndTime > MIN_TASK_CLUSTER_PADDING) {
```
brendankenny: I'm very curious why you went this way instead of clustering by 250ms and then doing a `isNextClusterTooClose` or whatever with 1000ms. Was there something easier this way or just felt more natural to you or...?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+      clusters.push(currentCluster);
+    }
+
+    return clusters
+      // add some useful information about the cluster
+      .map(tasks => {
+        const start = tasks[0].start;
+        const end = tasks[tasks.length - 1].end;
+        const duration = end - start;
+        return {start, end, duration, tasks};
```
brendankenny: any reason to keep `tasks` around? (not on the return type and looks like not used, so OK to drop?)

brendankenny:
I think there is one case this is missing: once a task is selected as the start of a window, `getTaskClustersInWindow` doesn't consider if the next tasks examined should have been clustered with that first task.

So, e.g. (in test form)
```js
mainThreadEvents = [
  {start: 5500, end: 5600},
  {start: 6000, end: 6100},
];

const result = firstInteractive.computeWithArtifacts({}, {}, {
  timings: {
    firstMeaningfulPaint: 1000,
    domContentLoaded: 800,
    traceEnd: 24000,
  },
  timestamps: {
    navigationStart: 0,
  },
});
```
says that FirstInteractive is at 5600, even though that's less than 1000ms away from the next long task, so it should be 6100 (correct me if the example is wrong; it's late :)

I believe the fix could be as simple as adding a
```js
if (i + 1 < longTasks.length &&
    longTasks[i + 1].start - windowStart <= MIN_TASK_CLUSTER_PADDING) {
  continue;
}
```
after the `windowEnd > traceEnd` check in `findQuietWindow`, but there might be a better/prettier way of doing it

lighthouse-core/audits/first-interactive.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc. All rights reserved.
```
patrickhulce: it's whatever it was copied from :)

lighthouse-core/audits/first-interactive.js
```diff
@@ -0,0 +1,73 @@
...
+        score = Math.min(100, score);
+        score = Math.max(0, score);
+        score = Math.round(score);
+
+        const displayValue = Math.round(firstInteractive.timeInMs / 10) * 10;
+        return {
+          score,
+          rawValue: firstInteractive.timeInMs,
+          displayValue: `${displayValue.toLocaleString()}ms`,
+          optimalValue: this.meta.optimalValue,
```
patrickhulce: done

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -20,18 +20,17 @@
...
-const TTIMetric = require('./time-to-interactive');
 const Emulation = require('../lib/emulation');
 
 const Formatter = require('../report/formatter');
 
-// Maximum TTI to be considered "fast" for PWA baseline checklist
+// Maximum TTFI to be considered "fast" for PWA baseline checklist
 //   https://developers.google.com/web/progressive-web-apps/checklist
-const MAXIMUM_TTI = 10 * 1000;
+const MAXIMUM_TTFI = 10 * 1000;
```
patrickhulce: I'll revert

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+const MAX_QUIET_WINDOW_SIZE = 5000;
+const TRACE_BUSY_MSG = 'trace was busy the entire time';
+
+/**
+ * @fileoverview This artifact identifies the time the page is "first interactive" as defined below
+ * @see https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/edit#
+ *
+ * First Interactive marks the first moment when a website is minimally interactive:
+ *    > Enough (but maybe not all) UI components shown on the screen are interactive
+ *      DISCLAIMER: This is assumed by virtue of the fact that the CPU is idle, actual event
```
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: ha done

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+ *      1 ~= f(∞) ~= 0 + 1
+ *    > a "bad task cluster" is a cluster of 1 or more long tasks with no more than 1s of idle time
+ *      between each task that does one of the following
+ *        > Starts within the first 5s after FMP.
+ *        > Spans more than 250ms from the start of the earliest task in the cluster to the end of the
+ *          latest task in the cluster.
+ *
+ * If this timestamp is earlier than DOMContentLoaded, use DOMContentLoaded as firstInteractive.
+ */
+class FirstInteractive extends ComputedArtifact {
```
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: Not sure what you mean by clustering by 250ms since in this new way of viewing things that's more a condition on a task cluster rather than how a cluster is defined, what you're saying sounds mostly to me like why did I change it away from what I had before :)

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: was just for debugging, can drop

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+        const end = tasks[tasks.length - 1].end;
+        const duration = end - start;
+        return {start, end, duration, tasks};
+      })
+      // filter out clusters that started after the window because of our clusteringWindowEnd
+      .filter(cluster => cluster.start < windowEnd);
+  }
+
+  /**
+   * @param {number} FMP
```
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+   * @param {number} traceEnd
+   * @param {!Array<{start: number, end: number>}} longTasks
+   * @return {number}
+   */
+  static findQuietWindow(FMP, traceEnd, longTasks) {
+    if (longTasks.length === 0 ||
+        longTasks[0].start > FMP + MAX_QUIET_WINDOW_SIZE) {
```
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+      return FMP;
+    }
+
+    // FirstInteractive must start at the end of a long task, consider each long task and
+    // examine the window that follows it.
+    for (let i = 0; i < longTasks.length; i++) {
+      const event = longTasks[i];
```
patrickhulce: it wasn't even really used, killed

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+      const windowStart = event.end;
+      const windowSize = FirstInteractive.getRequiredWindowSizeInMs(windowStart - FMP);
+      const windowEnd = windowStart + windowSize;
+
+      // Check that we have a long enough trace
+      if (windowEnd > traceEnd) {
+        throw new Error(TRACE_BUSY_MSG);
+      }
+
+      const isTooCloseToFMP = cluster => cluster.start < FMP + MIN_TASK_CLUSTER_FMP_DISTANCE;
```
patrickhulce: moved outside the loop, they're separate for readability and clarity of each condition that must be true

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * @param {!Trace} trace
+   * @param {!tr.Model} traceModel
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {{timeInMs: number, timestamp: number}}
+   */
+  computeWithArtifacts(trace, traceModel, traceOfTab) {
+    const navStart = traceOfTab.timestamps.navigationStart;
+    const FMP = traceOfTab.timings.firstMeaningfulPaint;
```
patrickhulce: yeah I'll throw

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    const DCL = traceOfTab.timings.domContentLoaded;
+    const traceEnd = traceOfTab.timings.traceEnd;
+
+    if (traceEnd - FMP < MAX_QUIET_WINDOW_SIZE) {
+      throw new Error('trace not at least 5 seconds longer than FMP');
+    }
+
+    const longTasksAfterFMP = TracingProcessor.getMainThreadTopLevelEvents(traceModel, trace, FMP)
+        .filter(evt => evt.end - evt.start >= LONG_TASK_THRESHOLD && evt.end > FMP);
```
patrickhulce: yup and yup

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,206 @@
...
+    return {
+      timeInMs: valueInMs,
+      timestamp: (valueInMs + navStart) * 1000,
+    };
+  }
+
+  /**
+   * @param {!Object} trace
+   * @param {!Object} artifacts
+   * @return {!Object}
```
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -246,6 +236,29 @@ class TraceProcessor {
...
+    const startEvent = trace.traceEvents.find(event => event.name === 'TracingStartedInPage');
+    const mainThread = TraceProcessor._findMainThreadFromIds(model, startEvent.pid, startEvent.tid);
+
+    const events = mainThread.sliceGroup.slices.filter(slice => {
+      return slice.title === SCHEDULABLE_TASK_TITLE &&
+          slice.end > startTime &&
+          slice.start < endTime;
+    });
+
+    events.sort((a, b) => a.start - b.start);
```
patrickhulce: not a fan, but I can live

lighthouse-core/test/gather/computed/first-interactive-test.js
```diff
@@ -0,0 +1,215 @@
...
+
+const FirstInteractive = require('../../../gather/computed/first-interactive');
+const firstInteractive = new FirstInteractive();
+const TracingProcessor = require('../../../lib/traces/tracing-processor.js');
+
+const assert = require('assert');
+
+/* eslint-env mocha */
+describe('FirstInteractive computed artifact:', () => {
+  let mainThreadEvents;
```
patrickhulce: nope its for the `TracingProcessor.getMainThreadTopLevelEventsStub`

lighthouse-core/test/gather/computed/first-interactive-test.js
```diff
@@ -0,0 +1,215 @@
...
+        },
+        timestamps: {
+          navigationStart: 0,
+        },
+      });
+
+      assert.equal(result.timeInMs, 3400);
+    });
+
+    it('should return DCL when DCL is after interactive', () => {
```
patrickhulce: done

lighthouse-core/test/gather/computed/first-interactive-test.js
```diff
@@ -0,0 +1,215 @@
...
+        {start: 2200, end: 10000},
+        {start: 11000, end: 11500},
+
+        // first lonely task group
+        {start: 12750, end: 12825},
+        {start: 12850, end: 12930},
+        {start: 12935, end: 12990},
+
+        {start: 14000, end: 17000},
+      ];
```
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
paulirish: 1. i would prefer to do the sort as early as possible as we dont want audits to have to know things are unsorted
1. i'm decently confident that the sort is done by tracingmodel already, so TBH this line can probably just be nuked.

patrickhulce:
> says that FirstInteractive is at 5600, even though that's less than 1000ms away from the next long task, so it should be 6100 (correct me if the example is wrong; it's late :)

yup good call, didn't have a test for this before and got missed in refactor

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: I'm adding tests with actual traces today now that it's in good shape so I'll find out and nuke the sort in caller if necessary

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: nope I'm keeping again :)

patrickhulce:
alright I believe all feedback addressed, added a couple tests for called out cases and the ones @brendankenny pointed out, added a new trace with 5s quiet on pwa.rocks to assert TTFI is correct with actual trace

lighthouse-core/gather/computed/first-interactive.js
brendankenny: > Not sure what you mean by clustering by 250ms since in this new way of viewing things that's more a condition on a task cluster rather than how a cluster is defined, what you're saying sounds mostly to me like why did I change it away from what I had before :)

yeah, I was curious about your reason for switching

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: > I'm adding tests with actual traces today now that it's in good shape so I'll find out and nuke the sort in caller if necessary

was this 👍 ?

lighthouse-core/audits/load-fast-enough-for-pwa.js
brendankenny: > nope I'm keeping again :)

I think @paulirish took the AI to get the message out about this for you :)

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: yep pre-sorted by model!

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: oh this just seemed more straightforward after talking it out with paul, easier to explain exactly what's happening without looking at code

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,222 @@
...
+    const FMP = traceOfTab.timings.firstMeaningfulPaint;
+    const DCL = traceOfTab.timings.domContentLoaded;
+    const traceEnd = traceOfTab.timings.traceEnd;
+
+    if (traceEnd - FMP < MAX_QUIET_WINDOW_SIZE) {
+      throw new Error('trace not at least 5 seconds longer than FMP');
+    }
+
+    if (!FMP || !DCL) {
+      throw new Error(`No ${FMP ? 'domContentLoaded' : 'firstMeaningfulPaint'} event in trace`);
```
brendankenny: should we throw in `traceOfTab` just to ruin all dependent audits' days?

lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -0,0 +1,222 @@
...
+
+    let previousTaskEndTime = -Infinity;
+    let currentCluster = null;
+
+    // Examine all tasks that could possibly be part of a cluster starting before windowEnd.
+    // Consider the case where window end is 15s, there's a 100ms task from 14.9-15s and a 500ms
+    // task from 15.5-16s, we need that later task to be clustered with the first so we can properly
+    // identify that main thread isn't quiet.
+    const clusteringWindowEnd = windowEnd + MIN_TASK_CLUSTER_PADDING;
+    const isInClusteringWindow = task => task && task.start < clusteringWindowEnd;
```
brendankenny: nit: `task` will always be defined here now

lighthouse-core/test/gather/computed/first-interactive-test.js
```diff
@@ -0,0 +1,290 @@
...
+/* eslint-env mocha */
+describe('FirstInteractive computed artifact:', () => {
+  let computedArtifacts;
+
+  beforeEach(() => {
+    computedArtifacts = Runner.instantiateComputedArtifacts();
+  });
+
+  it('throws on short traces', () => {
+    return firstInteractive.compute_({traceEvents: tooShortTrace}, computedArtifacts).then(() => {
```
brendankenny: nit: for these, can this call `computedArtifacts(requestFirstInteractive({traceEvents: tooShortTrace})` so it doesn't have to call the underlying `compute_`?

brendankenny:
oh, sorry for the conflict :S

patrickhulce:
> oh, sorry for the conflict :S

geez even after approval you're givin' me a hard time ;)

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: I'm down

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: I only moved FMP since DCL is specific to TTFI but everything needs FMP basically

lighthouse-core/test/gather/computed/first-interactive-test.js
patrickhulce: done

lighthouse-core/gather/computed/first-interactive.js
patrickhulce: well, nevermind we explicitly test that trace-of-tab can handle those traces without paints...reverting for now and we can do in a followup PR if necessary

brendankenny: > reverting for now and we can do in a followup PR if necessary

I guess having #1981 open will have to suffice for now :(

brendankenny:
⏱🖌🏁⏱

benschwarz:
👏 Great work @patrickhulce && @brendankenny

PR #2017, opened 2017-04-15
bump master to v1.7.0-alpha

paulirish:


paulirish:
:two: :zero: :one: :seven:   v:one:.:seven:.:zero:alpha

patrickhulce:
we have a crap ton of breaking changes, what about v2-alpha? :)

paulirish:
Ya, sgtm

PR #2018, opened 2017-04-15
allow computed artifacts to request other computed artifacts

brendankenny:
There's no reason that a computed artifact shouldn't be able to access other computed artifacts using the same artifact that it received. Since it can't access any new artifacts, just things based on the artifact it's already received, the cache will still work and computation of audits->requiredArtifacts->required gatherers still works.

should fulfill the needs of #2013

I've made a `ComputedArtifacts` type for this object, and had `Artifacts` extend it. Originally I was going to do that for the actual artifact object as well (make the computedArtifact object the prototype of the artifact object), but realized it wasn't necessary.

### DIFF
--- a/lighthouse-core/gather/computed/computed-artifact.js
+++ b/lighthouse-core/gather/computed/computed-artifact.js
@@ -18,8 +18,15 @@
 'use strict';
 
 class ComputedArtifact {
-  constructor() {
-    this.cache = new Map();
+  /**
+   * @param {!ComputedArtifacts} allComputedArtifacts
+   */
+  constructor(allComputedArtifacts) {
+    /** @private {!Map} */
+    this._cache = new Map();
+
+    /** @private {!ComputedArtifacts} */
+    this._allComputedArtifacts = allComputedArtifacts;
   }
 
   /* eslint-disable no-unused-vars */
@@ -27,10 +34,12 @@ class ComputedArtifact {
   /**
    * Override to implement a computed artifact. Can return a Promise or the
    * computed artifact itself.
-   * @param {!Object} artifact Input to computation.
+   * @param {*} artifact Input to computation.
+   * @param {!ComputedArtifacts} allComputedArtifacts Access to all computed artifacts.
+   * @return {*}
    * @throws {Error}
    */
-  compute_(artifact) {
+  compute_(artifact, allComputedArtifacts) {
     throw new Error('compute_() not implemented for computed artifact ' + this.name);
   }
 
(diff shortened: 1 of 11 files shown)


brendankenny:
> Originally I was going to do that for the actual artifact object as well (make the computedArtifact object the prototype of the artifact object), but realized it wasn't necessary.

on the other hand, `Object.assign()` not altering computedArtifacts depends on the order in which its called, so maybe this should switch to the more explicit `Object.create()`...

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -18,36 +18,45 @@
...
+   * @return {!Promise<*>}
    */
   request(artifact) {
-    if (this.cache.has(artifact)) {
-      return Promise.resolve(this.cache.get(artifact));
+    if (this._cache.has(artifact)) {
+      return Promise.resolve(this._cache.get(artifact));
     }
 
     return Promise.resolve().then(_ => this.compute_(artifact)).then(computedArtifact => {
```
patrickhulce: don't you need to pass it in here?

brendankenny: > don't you need to pass it in here?

er, yeah, probably kind of important :)

brendankenny: done

brendankenny:
PTAL

lighthouse-core/gather/gather-runner.js
```diff
@@ -430,13 +430,16 @@ class GatherRunner {
...
+   * @return {!ComputedArtifacts}
+   */
   static instantiateComputedArtifacts() {
     const computedArtifacts = {};
     require('fs').readdirSync(path.join(__dirname, 'computed')).forEach(function(file) {
       // Drop `.js` suffix to keep browserify import happy.
       file = file.replace(/\.js$/, '');
       const ArtifactClass = require('./computed/' + file);
-      const artifact = new ArtifactClass();
+      const artifact = new ArtifactClass(computedArtifacts);
```
patrickhulce: really not a fan of this intertwined computedArtifacts, artifact object, but I guess it works for now

brendankenny: > really not a fan of this intertwined computedArtifacts, artifact object, but I guess it works for now

sure, it uses inheritance for delegation, but on the plus side it guarantees invariants on our artifact dependency graph and it's nicely isolated, so it's ready for an easy refactor at any time :)

PR #2019, opened 2017-04-15
Report v2: Basic table formatter

paulirish:
It's not very pretty but the basics are there.

* adds thumbnail renderer
* adds URL renderer (just displays as text for now)
* adds table renderer (obviously)
* introduces `AuditResult.details.itemHeaders` for the `thead` cells

screenshot, lol:
image

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -70,7 +70,7 @@ class Audit {
    * Table cells will use the type specified in headings[x].itemType. However a custom type
    * can be provided: results[x].propName = {type: 'code', text: '...'}
    * @param {!Audit.Headings} headings
-   * @param {!Array<!Object<string, string>>} results
+   * @param {!Array<!Object<string, *>>} results
    * @return {!Array<!DetailsRenderer.DetailsJSON>}
    */
   static makeV2TableRows(headings, results) {
@@ -157,13 +157,20 @@ class Audit {
 
 module.exports = Audit;
 
+/** @typedef {
+ * !Array<{
+ *   key: string,
+ *   itemType: string,
+ *   text: string,
+ * }>}
+ */
+Audit.Headings; // eslint-disable-line no-unused-expressions
+
 /** @typedef {{
- *     key: string,
- *     itemType: string,
- *     text: string,
+ *   results: !Array<!Object<string, string>>,
+ *   headings: !Audit.Headings,
+ *   passes: boolean,
+ *   debugString: (string|undefined)
  * }}
  */
-Audit.Heading; // eslint-disable-line no-unused-expressions
-
-/** @typedef {!Array<!Audit.Heading>} */
-Audit.Headings; // eslint-disable-line no-unused-expressions
+Audit.HeadingsResult; // eslint-disable-line no-unused-expressions
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -81,7 +81,7 @@ class Audit {
 
         return {
           type: heading.itemType,
-          text: value[heading.key]
+          text: value
         };
       });
     });
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -57,6 +57,61 @@ class Audit {
   }
 
   /**
+   * @param  {!Audit.Headings} headings
+   * @return {*}
+   */
+  static makeV1TableHeadings(headings) {
+    const tableHeadings = {};
+    headings.forEach(heading => tableHeadings[heading.key] = heading.text);
+    return tableHeadings;
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
+   * @return {!Array<!DetailsRenderer.DetailsJSON>}
+   */
+  static makeV2TableRows(headings, results) {
+    const tableRows = results.map(item => {
+      return headings.map(heading => {
+        return {
+          type: heading.itemType,
+          text: item[heading.key]
+        };
+      });
+    });
+    return tableRows;
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @return {!Array<!DetailsRenderer.DetailsJSON>}
+   */
+  static makeV2TableHeaders(headings) {
+    return headings.map(heading => ({
+      type: 'text',
+      text: heading.text
+    }));
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
+   * @return {!DetailsRenderer.DetailsJSON}
+   */
+  static makeV2TableDetails(headings, results) {
+    const v2TableHeaders = Audit.makeV2TableHeaders(headings);
+    const v2TableRows = Audit.makeV2TableRows(headings, results);
+    return {
+      type: 'table',
+      header: 'View Details',
+      itemHeaders: v2TableHeaders,
+      items: v2TableRows
+    };
+  }
+
+
+  /**
    * @param {!Audit} audit
    * @param {!AuditResult} result
    * @return {!AuditFullResult}
@@ -97,3 +152,16 @@ class Audit {
 }
 
 module.exports = Audit;
+
+
+/** @typedef {{
+ *     key: string,
+ *     itemType: string,
+ *     text: string,
+ * }}
+ */
+Audit.Heading; // eslint-disable-line no-unused-expressions
+
+
+/** @typedef {!Array<!Audit.Heading>} */
+Audit.Headings; // eslint-disable-line no-unused-expressions
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -110,7 +110,6 @@ class Audit {
     };
   }
 
-
   /**
    * @param {!Audit} audit
    * @param {!AuditResult} result
@@ -153,7 +152,6 @@ class Audit {
 
 module.exports = Audit;
 
-
 /** @typedef {{
  *     key: string,
  *     itemType: string,
@@ -161,7 +159,6 @@ module.exports = Audit;
  * }}
  */
 Audit.Heading; // eslint-disable-line no-unused-expressions
-
 
 /** @typedef {!Array<!Audit.Heading>} */
 Audit.Headings; // eslint-disable-line no-unused-expressions
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -57,8 +57,8 @@ class Audit {
   }
 
   /**
-   * @param  {!Audit.Headings} headings
-   * @return {*}
+   * @param {!Audit.Headings} headings
+   * @return {!Array<string>}
    */
   static makeV1TableHeadings(headings) {
     const tableHeadings = {};
@@ -67,8 +67,8 @@ class Audit {
   }
 
   /**
-   * @param  {!Audit.Headings} headings
-   * @param  {!Object} results
+   * @param {!Audit.Headings} headings
+   * @param {!Array<!Object<string, string>>} results
    * @return {!Array<!DetailsRenderer.DetailsJSON>}
    */
   static makeV2TableRows(headings, results) {
@@ -84,7 +84,7 @@ class Audit {
   }
 
   /**
-   * @param  {!Audit.Headings} headings
+   * @param {!Audit.Headings} headings
    * @return {!Array<!DetailsRenderer.DetailsJSON>}
    */
   static makeV2TableHeaders(headings) {
@@ -95,8 +95,8 @@ class Audit {
   }
 
   /**
-   * @param  {!Audit.Headings} headings
-   * @param  {!Object} results
+   * @param {!Audit.Headings} headings
+   * @param {!Array<!Object<string, string>>} results
    * @return {!DetailsRenderer.DetailsJSON}
    */
   static makeV2TableDetails(headings, results) {
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -67,6 +67,8 @@ class Audit {
   }
 
   /**
+   * Table cells will use the type specified in headings[x].itemType. However a custom type
+   * can be provided: results[x].propName = {type: 'code', text: '...'}
    * @param {!Audit.Headings} headings
    * @param {!Array<!Object<string, string>>} results
    * @return {!Array<!DetailsRenderer.DetailsJSON>}
@@ -74,9 +76,12 @@ class Audit {
   static makeV2TableRows(headings, results) {
     const tableRows = results.map(item => {
       return headings.map(heading => {
+        const value = item[heading.key];
+        if (typeof value === 'object' && value.type) return value;
+
         return {
           type: heading.itemType,
-          text: item[heading.key]
+          text: value[heading.key]
         };
       });
     });
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -95,8 +95,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
 
   /**
    * @param {!Artifacts} artifacts
-   * @return {{results: !Array<Object>, tableHeadings: Object,
-   *     passes: boolean=, debugString: string=}}
+   * @return {{results: !Array<Object>, headings: !Audit.Headings, debugString: string=}}
    */
   static audit_(artifacts) {
     const images = artifacts.ImageUsage;
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -132,15 +132,18 @@ class OffscreenImages extends ByteEfficiencyAudit {
         const loadedEarly = item.requestStartTime < ttiTimestamp;
         return isWasteful && loadedEarly;
       });
+
+      const headings = [
+        {key: 'preview', itemType: 'thumbnail', text: ''},
+        {key: 'url', itemType: 'url', text: 'URL'},
+        {key: 'totalKb', itemType: 'text', text: 'Original'},
+        {key: 'potentialSavings', itemType: 'text', text: 'Potential Savings'},
+      ];
+
       return {
         debugString,
         results,
-        tableHeadings: {
-          preview: '',
-          url: 'URL',
-          totalKb: 'Original',
-          potentialSavings: 'Potential Savings',
-        }
+        headings,
       };
     });
   }
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -96,7 +96,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
 
   /**
    * @param {!Artifacts} artifacts
-   * @return {{results: !Array<Object>, headings: !Audit.Headings, debugString: string=}}
+   * @return {!Audit.HeadingsResult}
    */
   static audit_(artifacts) {
     const images = artifacts.ImageUsage;
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -83,6 +83,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
     return {
       url,
       preview: {
+        type: 'thumbnail',
         url: image.networkRecord.url,
         mimeType: image.networkRecord.mimeType
       },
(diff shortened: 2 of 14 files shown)


lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -179,14 +179,17 @@ class UnusedCSSRules extends Audit {
       return UnusedCSSRules.mapSheetToResult(indexedSheets[sheetId], pageUrl);
     }).filter(sheet => sheet && sheet.wastedBytes > 1024);
 
+
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
```
patrickhulce: we can override this per row correct? it'd be nice to finally have a nice solution for displaying the inline code preview in the url column

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
...
+    return headings.map(heading => ({
+      type: 'text',
+      text: heading.text
+    }));
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
+   * @return {!DetailsRenderer.DetailsJSON}
```
ebidel: the return obj is slightly different than `DetailsJSON`. Should it be a new type? `TableDetailsJSON` or whatever.

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -132,15 +132,18 @@ class OffscreenImages extends Audit {
...
       return {
         debugString,
         results,
-        tableHeadings: {
-          preview: '',
-          url: 'URL',
-          totalKb: 'Original',
-          potentialSavings: 'Potential Savings',
-        }
+        headings,
```
ebidel: The `@return` docstring of this method needs to be updated to match.

lighthouse-core/audits/audit.js
ebidel: Oh, you have it below :) `DetailsRenderer.TableDetailsJSON`

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -88,6 +116,33 @@ class DetailsRenderer {
     return element;
   }
 
+
+  /**
+   * @param {!DetailsRenderer.TableDetailsJSON} details
```
ebidel: is `TableDetailsJSON` typedef'd anywhere? not seeing it.

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -55,6 +61,28 @@ class DetailsRenderer {
...
+
+  /**
+   * @param {!DetailsRenderer.DetailsJSON} obj
+   * @return {!Element}
+   */
+  _renderThumbnail(obj) {
+    const element = this._dom.createElement('img', 'lh-thumbnail');
+    element.src = obj.text.url;
+    // ignore obj.text.mimeType
+    element.alt = 'Image preview';
```
ebidel: Should we also add `element.title = obj.text.url;` so there's a full url when users hover over the image?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -88,6 +116,33 @@ class DetailsRenderer {
     return element;
   }
 
+
+  /**
+   * @param {!DetailsRenderer.TableDetailsJSON} details
+   * @return {!Element}
+   */
+  _renderTable(details) {
+    const element = this._dom.createElement('details', 'lh-details', {open: true});
```
ebidel: should we open by default? There could be a  lot of stuff in the table.

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -88,6 +116,33 @@ class DetailsRenderer {
...
+    if (details.header) {
+      element.appendChild(this._dom.createElement('summary')).textContent = details.header;
+    }
+
+    const tableElem = element.createChild('table', 'lh-table lh-table__multicolumn');
```
ebidel: Do you need `lh-table__multicolumn`? That was an impl detail of v1's handlebars table.

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -23,6 +23,30 @@ class DOM {
...
+    const instance = this;
+
+    /**
+     * From devtools' DOMExtension.js (minus the 3rd customElementType param)
+     * @param {string} elementName
+     * @param {string=} className
+     * @return {!Element}
+     */
+    self.Element.prototype.createChild = function(elementName, className) {
+      const element = instance.createElement(elementName, className);
```
ebidel: pipe through the `attrs` param too?

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -23,6 +23,30 @@ class DOM {
...
+      this.appendChild(element);
+      return element;
```
ebidel: nit: `return this.appendChild(element);`

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -23,6 +23,30 @@ class DOM {
...
+    if (self.Element.prototype.createChild) return;
+    const instance = this;
+
+    /**
+     * From devtools' DOMExtension.js (minus the 3rd customElementType param)
+     * @param {string} elementName
+     * @param {string=} className
+     * @return {!Element}
+     */
+    self.Element.prototype.createChild = function(elementName, className) {
```
ebidel: If we're comfortable with modifying built in prototypes, then I have a bunch of other helpers to add :)

lighthouse-core/report/v2/report-styles.css
```diff
@@ -301,4 +301,15 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
   padding: 3px;
 }
 
+
+.lh-table {
+  --image-preview: 24px;
+}
+
+img.lh-thumbnail {
```
ebidel: just `.lh-thumbnail`? Seems ok to be tag-agnostic.

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
paulirish: to have an individual row change the itemType? that seems a little weird. whyd we want diff itemtypes in the column?

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: sg

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: nah. was only for development. will remove. :)

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: nope. will nuke.

lighthouse-core/report/v2/renderer/dom.js
paulirish: check out https://github.com/ChromeDevTools/devtools-frontend/blob/master/front_end/dom_extension/DOMExtension.js .. we can polyfill whatever from there we want. (or easier: just load that file in the report)

lighthouse-core/report/v2/renderer/dom.js
paulirish: i dont think we can because in devtools this method will already exist and it doesnt do attrs

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
   }
 
   /**
+   * @param  {!Audit.Headings} headings
+   * @return {*}
```
brendankenny: `!Array<string>`?

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
...
+   */
+  static makeV1TableHeadings(headings) {
+    const tableHeadings = {};
+    headings.forEach(heading => tableHeadings[heading.key] = heading.text);
+    return tableHeadings;
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
```
brendankenny: not an object. `!Array<!Object<string, string>>`? It's hard to follow the old `createTable` helper

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
   }
 
   /**
+   * @param  {!Audit.Headings} headings
```
brendankenny: just one space (here and elsewhere). Is this a vscode setting that needs to be changed?

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
...
+          text: item[heading.key]
+        };
+      });
+    });
+    return tableRows;
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @return {!Array<!DetailsRenderer.DetailsJSON>}
```
brendankenny: this is fine for now, but it won't actually work if we were to turn on type checking on this file. Closure won't be able to see the typedef since it's in another module. We'll have to think of how we want to do that (move report typedefs to their own file?)

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
...
+  static makeV2TableHeaders(headings) {
+    return headings.map(heading => ({
+      type: 'text',
+      text: heading.text
+    }));
+  }
+
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
```
brendankenny: same type as above `results`

lighthouse-core/audits/audit.js
```diff
@@ -97,3 +151,14 @@ class Audit {
...
+
+/** @typedef {{
+ *     key: string,
+ *     itemType: string,
+ *     text: string,
+ * }}
+ */
+Audit.Heading; // eslint-disable-line no-unused-expressions
+
+/** @typedef {!Array<!Audit.Heading>} */
```
brendankenny: is it worth making a typedef for this?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -55,6 +61,28 @@ class DetailsRenderer {
...
+    return element;
+  }
+
+  /**
+   * @param {!DetailsRenderer.DetailsJSON} obj
+   * @return {!Element}
+   */
+  _renderThumbnail(obj) {
+    const element = this._dom.createElement('img', 'lh-thumbnail');
+    element.src = obj.text.url;
```
brendankenny: this isn't a thing on `DetailsJSON`

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,60 @@ class Audit {
...
+  /**
+   * @param  {!Audit.Headings} headings
+   * @param  {!Object} results
+   * @return {!Array<!DetailsRenderer.DetailsJSON>}
+   */
+  static makeV2TableRows(headings, results) {
+    const tableRows = results.map(item => {
+      return headings.map(heading => {
+        return {
+          type: heading.itemType,
```
brendankenny: why not do this in `_renderTable` rather than in the JSON report?

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: in all the css-based ones the "url column" doesn't always have a url and it's a code snippet preview instead since it's an inline style. it'd be nice to have the details renderer support that too so the code snippets can say `{type: "code", content: ".inline-style {...}"}` or something

ebidel:
@paulirish how's this coming? We've got a lot of audits that need this guy.

paulirish:
> @paulirish how's this coming? We've got a lot of audits that need this guy.

roger that. on it.

lighthouse-core/audits/audit.js
paulirish: shrugz

I'm okay with either.
patrick is fine with either but thinks its slightly better to have all things coming into renderer be our details objects
dgozman also feels the same way, and thinks that if we wanted to conserve the size of the LH object we should shorten `type` to `t` etc. ;) ;) 

so I'm comfortable expanding the objects over here before we deliver the LHR.

brendankenny: > so I'm comfortable expanding the objects over here before we deliver the LHR.

I think with our less rigid details taxonomy now I'm actually more OK with this extra verboseness :)

We can keep an eye on report file size and how much is taken up by `"type": "text"` and figure out some defaults (default to text or default to header itemType) if/when needed

lighthouse-core/audits/audit.js
paulirish: Actually yeah. Basically we pass around this typedef and never use the Audit.Heading one directly. Not entirely worth it to nuke this guy, i tried.

paulirish:
PTAL. updated.

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,65 @@ class Audit {
...
+   * can be provided: results[x].propName = {type: 'code', text: '...'}
+   * @param {!Audit.Headings} headings
+   * @param {!Array<!Object<string, string>>} results
+   * @return {!Array<!DetailsRenderer.DetailsJSON>}
+   */
+  static makeV2TableRows(headings, results) {
+    const tableRows = results.map(item => {
+      return headings.map(heading => {
+        const value = item[heading.key];
+        if (typeof value === 'object' && value.type) return value;
```
patrickhulce: 🎉

lighthouse-core/report/v2/report-styles.css
```diff
@@ -424,4 +424,14 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
   }
 }
 
+.lh-table {
+  --image-preview: 24px;
```
patrickhulce: nit: `--image-preview-size` or something?

ebidel:
@brendankenny to you

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -134,3 +197,21 @@ DetailsRenderer.ListDetailsJSON; // eslint-disable-line no-unused-expressions
  * }}
  */
 DetailsRenderer.CardsDetailsJSON; // eslint-disable-line no-unused-expressions
+
+/** @typedef {{
+ *     type: string,
+ *     header: ({text: string}|undefined),
+ *     items: !Array<!Array<!DetailsRenderer.DetailsJSON>>,
+ *     itemHeaders: !Array<!DetailsRenderer.DetailsJSON>
```
brendankenny: by letting these be `DetailsJSON` we're basically descending into recursive madness again since in theory when you call `this.render` on them they could embed a whole other table (but of course, our css would quickly fall over itself if that sort of thing occurred). This leaves these basically untyped (except our assurance to the compiler that we're correct when we cast in `this.render`)

I'm not sure of the right solution...maybe we could separate top level rendering (anything currently in `this.render`) from more primitive-level rendering (just text, url, and thumbnail as of this PR) in a future PR or something, but we would need to make whatever we did non-annoying.

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -54,6 +60,32 @@ class DetailsRenderer {
...
+    element.classList.add('lh-text__url');
+    return element;
+  }
+
+  /**
+   * @param {!DetailsRenderer.ThumbnailDetails} value
+   * @return {!Element}
+   */
+  _renderThumbnail(value) {
+    if (/^image/.test(value.mimeType) === false) {
```
brendankenny: make a note of this in a jsdoc description?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -54,6 +60,32 @@ class DetailsRenderer {
...
+      return this._dom.createElement('span');
+    }
+
+    const element = this._dom.createElement('img', 'lh-thumbnail');
+    element.src = value.url;
+    element.alt = 'Image preview';
```
brendankenny: best we can do?

lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
```diff
@@ -80,8 +81,7 @@ class UsesResponsiveImages extends ByteEfficiencyAudit {
 
   /**
    * @param {!Artifacts} artifacts
-   * @return {{results: !Array<Object>, tableHeadings: Object,
-   *     passes: boolean=, debugString: string=}}
+   * @return {{results: !Array<Object>, headings: !Audit.Headings, passes: boolean=, debugString: string=}}
```
brendankenny: `boolean=` would be `(boolean|undefined)`, but doesn't look like it can be undefined?
`string=` would be `(string|undefined)`

(`=` only works for `@params`...technically it means optional param, which just so happens to manifest as being `undefined`)

lighthouse-core/audits/byte-efficiency/uses-request-compression.js
```diff
@@ -46,7 +46,7 @@ class ResponsesAreCompressed extends ByteEfficiencyAudit {
   /**
    * @param {!Artifacts} artifacts
    * @param {number} networkThroughput
-   * @return {!AuditResult}
+   * @return {{results: !Array<Object>, passes: boolean=, headings: !Audit.Headings, debugString: string=}}
```
brendankenny: see below

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
```diff
@@ -63,8 +63,7 @@ class UsesOptimizedImages extends ByteEfficiencyAudit {
 
   /**
    * @param {!Artifacts} artifacts
-   * @return {{results: !Array<Object>, tableHeadings: Object,
-   *     passes: boolean=, debugString: string=}}
+   * @return {{results: !Array<Object>, headings: !Audit.Headings, passes: boolean=, debugString: string=}}
```
brendankenny: samsies

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -95,8 +96,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
 
   /**
    * @param {!Artifacts} artifacts
-   * @return {{results: !Array<Object>, tableHeadings: Object,
-   *     passes: boolean=, debugString: string=}}
+   * @return {{results: !Array<Object>, headings: !Audit.Headings, debugString: string=}}
```
brendankenny: see notes below

lighthouse-core/audits/audit.js
brendankenny: if that's the case, what about 
```js
/** @typedef {
 * !Array<{
 *   key: string,
 *   itemType: string,
 *   text: string,
 * }>}
 */
Audit.Headings;
```
and skip `Audit.Heading`?

lighthouse-core/audits/audit.js
brendankenny: results `value`s might not just be a `string`, then?

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: done

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: was matching our v1. :)  i changed it to empty string, which seems better, actually.

lighthouse-core/audits/audit.js
paulirish: fixed

lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
paulirish: thanks. made a new typdef for this guy since he was everywhere.

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: > we're basically descending into recursive madness again

are we though?


`detailsjson` only has type and text.. it can't contain other types.
really it and `thumbnail` should include the word "item" as they are ALWAYS leaf nodes.

and the `render()` method should accept all of these leaf & group types. the real bug seems that render's param is typed to a leaf node when we know its not.

brendankenny: `detailsjson` is the base (structural) type that is the superset of all of them. It only has fields all of them have: `type` and optional `text` (really text is there from when they all had `text` and should probably be removed). `render()` takes that and casts according to its `type` to the appropriate type, and we'll take responsibility to make sure that if `type` is set to x, then that object really is of `type` x. It's a hack, but simple (we only have to have humans in the loop checking the top level `details` object), and can easily be reworked once we know the full outlines of the types we need.

Allowing arbitrary nesting means that it's much less simple since we have to guarantee that all the nested items and itemHeaders are really whatever their `type` says. It also means we're implicitly breaking the no-typedef-cycles, just getting around it by casting, which is what I meant by recursive madness.

If you're saying we should have a different type explicitly for leaf nodes to make this better, I think that's great

paulirish: ya makes sense. let's sort this out in voice as i have a few ideas.

PR #2024, opened 2017-04-17
feat: add category validation to config

patrickhulce:


lighthouse-core/config/config.js
```diff
@@ -145,6 +145,30 @@ function validatePasses(passes, audits, rootPath) {
   });
 }
 
+function validateCategories(categories, audits) {
+  if (!categories || !audits) {
```
brendankenny: what about a `categories` with audits but no `audits` array? Wouldn't that fall under the `could not find audit` error?

lighthouse-core/config/config.js
```diff
@@ -145,6 +145,30 @@ function validatePasses(passes, audits, rootPath) {
...
+    return;
+  }
+
+  const auditIds = new Set(audits.map(audit => audit.meta.name));
+  Object.keys(categories).forEach(categoryId => {
+    const category = categories[categoryId];
+    if (!category.audits) {
+      throw new Error(`${categoryId} category has no audits`);
```
brendankenny: do we care about this case? Seems more like a `log.warn` situation, maybe? Also this throws an error for no `audits` array, but not for an empty array (which would be covered by "has no audits")

lighthouse-core/config/config.js
```diff
@@ -145,6 +145,30 @@ function validatePasses(passes, audits, rootPath) {
...
+    }
+
+    category.audits.forEach((audit, index) => {
+      if (!audit.id) {
+        throw new Error(`${categoryId} category is missing an audit id at index ${index}`);
+      }
+
+      if (!auditIds.has(audit.id)) {
+        throw new Error(`could not find ${audit.id} audit`);
```
brendankenny: maybe add a ` in category ${categoryId}`?

lighthouse-core/config/config.js
brendankenny: if we do care, needs a test :)

patrickhulce: I'll remove

lighthouse-core/config/config.js
patrickhulce: I had that, but there was`auditResults` biting us here. I'll just add for both.

lighthouse-core/config/config.js
patrickhulce: that reads like we were looking for the audit in the category, I'll use `for category ${categoryId}`?

brendankenny: > that reads like we were looking for the audit in the category

ah, good point. SG

lighthouse-core/config/config.js
```diff
@@ -145,6 +145,31 @@ function validatePasses(passes, audits, rootPath) {
   });
 }
 
+function validateCategories(categories, audits, auditResults) {
+  if (!categories) {
+    return;
+  }
+
+  if (!audits && !auditResults) {
+    throw new Error('could not find audits or auditResults');
```
brendankenny: this feels like total nitpicking, because this is indeed an error state, but this doesn't really feel like validating categories any more. Maybe `validateCategories` should just use an empty array if both don't exist and the fact that both exist should be checked in some overall validator

(currently this error state is caught in runner (and we [test it there](https://github.com/GoogleChrome/lighthouse/blob/9349c894a6a9d4f9b1f0a74ce4555433591d98aa/lighthouse-core/test/runner-test.js#L274)) since we didn't have `Config` back when it was written. It probably should be checked in Config so it never gets that far)

Feel free to disagree :)

patrickhulce: I'll get rid of it

lighthouse-core/config/config.js
```diff
@@ -145,6 +145,27 @@ function validatePasses(passes, audits, rootPath) {
   });
 }
 
+function validateCategories(categories, audits, auditResults = []) {
```
brendankenny: it shouldn't be possible for `auditResults` to ever be undefined. At worst it'll be [`null`](https://github.com/GoogleChrome/lighthouse/blob/9349c894a6a9d4f9b1f0a74ce4555433591d98aa/lighthouse-core/config/config.js#L288), so it won't ever get the default value

brendankenny: (lol sorry)

patrickhulce: 😝

Issue #2022, opened 2017-04-17
Audit: Prefixing Flexbox is currently considered an error


zouhir:
My CSS currently has Flexbox and I did lose points because of Prefixing.

**The error I got was:**

> **Using modern CSS features**
> **Avoids old CSS flexbox** 
> `The 2009 spec of Flexbox is deprecated and is 2.3x slower than the latest spec`

The feedback details in lighthouse found the error in this selector:
```css
.row {
    display: -ms-flexbox;
}
```
white in my CSS, it actually has prefixed and non-prefixed version of `display` property as the following:
```css
.row {
    display: -ms-flexbox;
    display: flex;
    -ms-flex-align: start;
    align-items: flex-start;
    -ms-flex-wrap: wrap;
    flex-wrap: wrap;
}
```
I think having that prefix will make my website more progressively enhanced for browsers that only support older slow specs and I believe modern ones should be using the non prefixed, modern faster version.

I'll raise it as a bug for now, keen to learn more about that issue and don't mind looking into it once I get the right knowledge on how LH should deal with Flexbox and maybe other modern CSS (I did signup for Google's OSS agreement btw)

*screenshot*
![screen shot 2017-04-17 at 16 36 34](https://cloud.githubusercontent.com/assets/5052316/25081772/b4dce240-238f-11e7-837f-df2f7710f91f.png)

hexalys:
Was just going to file this bug. So I'll comment here. I was [made aware](https://twitter.com/nekrtemplar/status/853780845139767297) prefixed flexbox was discouraged.

This shouldn't be the case for reasons stated at https://github.com/ai/browserslist/issues/98 and https://github.com/autoprefixer/autoprefixer.github.io/issues/9

More broadly. I don't think it's wise for tools or search engine audits to suggest that any prefixes for CSS layout should be removed until usage reached 0.0x%. `border-radius` and `box-shadow` can be an exception, since those are cosmetic only (and now in the near completely obsolete category). However, adding layout related prefixes can simply insure that sites aren't completely unusable in older browsers still being used at large, even if usage is in low digits.

brendankenny:
see #1710 for some recent discussion

(in the meantime, FWIW, while flagging this with an x when old flexbox is used is debatable, you aren't losing points to it. Nothing in the "Best Practices" section affects your score)

ebidel:
Just to be clear, Lighthouse is not a tool for "search engine audits".

[...]

PR #2025, opened 2017-04-17
Add fancier stuff to default config

ebidel:
R: @patrickhulce all

Not sure if this was intentional. A couple of audits are no longer being run after the switch to the new config.

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -687,6 +687,14 @@ module.exports = {
         {"id": "deprecations", "weight": 1},
         {"id": "manifest-short-name-length", "weight": 1},
       ]
+    },
+    "fancy": {
+      "name": "Fancier stuff",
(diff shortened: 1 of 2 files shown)


patrickhulce:
intentionally left out since they weren't expressed as being carried over to the new report in all the mocks of top level categories, unintentionally forgotten to highlight for discussion :)

I'm personally not the biggest fan of these right now as I believe they're noisy and not drop-in replacements, but I tend to sympathize with the likes of #2022 generally. We should get some proper 'warning'/'info'/'did you know?' styling that makes them not show up as errors 😃

ebidel:
Hmm, I thought we were going to filter. We have UI treatment for `informative: true` audits that we can leverage.

brendankenny:
where did this end up? @paulirish

paulirish:
I'm fine with adding no-consoletime to Best Practices and marking as informative

i think no-datenow should be removed, but we should discuss it in voice.

paulirish:
#2155 is tracking the return of these two

PR #2029, opened 2017-04-18
[WIP] Add audits renderer support + GroupByTag renderer

patrickhulce:
R: @ebidel @dgozman all

I wanted to throw this out there as my first *very* rough take on what different renderers would look like code-wise and get feedback before going any further. Please ignore UI :)

* Audits (optionally) specify a `renderer` property in their `meta` object similar to `scoringMode`
* Audits (optionally) specify a `tag` property in the config under their category entry
* `AuditsRenderer` groups audits by renderer and hands off the rendering to the specified class, this can be extended for a `PerformanceMetricRenderer` for the metrics and `PerformanceHintRenderer` for the blocking tags/byte efficiency audits.

If this rough angle looks good I'll close this PR and we can make the changes in stages:

1. I'll extract the pass/fail grouping to GroupByStatusRenderer and add the base logic for handling the `renderer` property.
2. I'll add the tags object to the config and report JSON with a shell of a GroupByTagRenderer.
3. Hand off to @robdodson to modify GroupByTagRenderer and apply tags to the a11y audits in whatever way makes the most sense there.

image

### DIFF
--- a/lighthouse-core/audits/accessibility/document-title.js
+++ b/lighthouse-core/audits/accessibility/document-title.js
@@ -33,6 +33,7 @@ class DocumentTitle extends AxeAudit {
       category: 'Accessibility',
       name: 'document-title',
       description: 'Document has a `<title>` element.',
+      renderer: AxeAudit.RENDERERS.GROUP_BY_TAG,
       helpText: 'Screen reader users use page titles to get an overview of the contents of ' +
           'the page.',
       requiredArtifacts: ['Accessibility']
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -37,6 +37,16 @@ class Audit {
   }
 
   /**
+   * @return {{GROUP_BY_STATUS: string, GROUP_BY_TAG: string}}
+   */
+  static get RENDERERS() {
+    return {
+      GROUP_BY_STATUS: 'GroupByStatus',
+      GROUP_BY_TAG: 'GroupByTag',
+    };
+  }
+
+  /**
    * @throws {Error}
    */
   static get meta() {
@@ -85,6 +95,7 @@ class Audit {
       debugString: result.debugString,
       optimalValue: result.optimalValue,
       extendedInfo: result.extendedInfo,
+      renderer: audit.meta.category === 'Accessibility' ? Audit.RENDERERS.GROUP_BY_TAG : Audit.RENDERERS.GROUP_BY_STATUS,
       scoringMode: audit.meta.scoringMode || Audit.SCORING_MODES.BINARY,
       informative: audit.meta.informative,
       name: audit.meta.name,
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -316,6 +316,7 @@ class Config {
     this._audits = Config.requireAudits(configJSON.audits, this._configDir);
     this._artifacts = expandArtifacts(configJSON.artifacts);
     this._categories = configJSON.categories;
+    this._tags = configJSON.tags;
 
     // validatePasses must follow after audits are required
     validatePasses(configJSON.passes, this._audits, this._configDir);
@@ -567,6 +568,11 @@ class Config {
   /** @type {Object<{audits: !Array<{id: string, weight: number}>}>} */
   get categories() {
     return this._categories;
+  }
+
+  /** @type {Object<string, {title: string, description: string}>} */
+  get tags() {
+    return this._tags;
   }
 }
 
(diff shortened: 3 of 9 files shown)


lighthouse-core/audits/accessibility/document-title.js
```diff
@@ -33,6 +33,7 @@ class DocumentTitle extends AxeAudit {
       category: 'Accessibility',
       name: 'document-title',
       description: 'Document has a `<title>` element.',
+      renderer: AxeAudit.RENDERERS.GROUP_BY_TAG,
```
patrickhulce: example of what it looks like on the audit side

lighthouse-core/audits/audit.js
```diff
@@ -85,6 +95,7 @@ class Audit {
       debugString: result.debugString,
       optimalValue: result.optimalValue,
       extendedInfo: result.extendedInfo,
+      renderer: audit.meta.category === 'Accessibility' ? Audit.RENDERERS.GROUP_BY_TAG : Audit.RENDERERS.GROUP_BY_STATUS,
```
patrickhulce: this line is a total hack to make a quick demo, would be more like the line below `audit.meta.renderer | Audit.RENDERERS.GROUP_BY_STATUS`

brendankenny:
I've only skimmed this so far, but do we need both audits and the config to know about how the results will be grouped? Ideally this would be config-side only and the audit wouldn't have to care about it.

brendankenny:
OTOH, I think I see what your larger thinking is here, and it feels like it will make a lot of sense for e.g. `PerformanceMetricRenderer`: take all the audits that can be rendered like that and pass them off to that audit renderer.

Should grouping be a more fundamental capability of the report renderer, though? As noted, it seems like it can be expressed without `Audit.meta` participation, and separating grouping would allow e.g. multiple groups of `PerformanceMetricRenderer` within a category without adding that in the future through a recursive call back to `AuditsRenderer.render`.

patrickhulce:
>  Ideally this would be config-side only and the audit wouldn't have to care about it.

I started with config-only with the renderer specified in the tag, but it occurred to me that the accessibility audits are really the only ones that have this special tag relationship. I'm relatively ambivalent about putting the renderer in config on each audit vs. in the meta, but I see it following the path of scoring modes where the renderer that can be used is tied to shape of the results object and therefore isn't really configuration at all but inherent to the audit.

> Should grouping be a more fundamental capability of the report renderer, though, rather than just one renderer in a set of them?

How do you envision this being expressed?

brendankenny:
> I'm relatively ambivalent about putting the renderer in config on each audit vs. in the meta, but I see it following the path of scoring modes where the renderer that can be used is tied to shape of the results object and therefore isn't really configuration at all but inherent to the audit.

yeah, I agree the renderer shouldn't be in the config. That's partially why I was suggesting grouping should be something different.

> How do you envision this being expressed?

It would depend on what styling we want exactly, but basically make grouping part of the loop over audits in `_renderCategory`. Have tags in the config, then `_renderCategory` collect groups of audits with tags and single audits without them, then have it pass each of those off to `AuditsRenderer`, where it can deal with the way that audit/group wants to be rendered. Whatever `AuditsRenderer` passes back can either be appended top level in the category element or nested in a tag-group element (which has the group's title and description added).

brendankenny:
There are some downsides (e.g. the same `tag` property couldn't be context dependent based on renderer, we'd need yet another tag thing), but it allows nesting of any type of audit in a group (e.g. a group of performance metrics for first load and a group for second load), and it allows grouping without the audit having to opt into it via its `meta` settings.

patrickhulce:
hm that's interesting, but wouldn't that force tag application into grouping by tag as opposed to allowing the current group by passing? I had seen a world in which audits could have tags that were basically useful hover or descriptive information. 

This is problematic to discuss because we have no specific outlined use cases for tags other than a11y :) haha, if having a tag always means that they're grouped for display I'd vote for expressing that as a category that could be referenced by a parent instead since that's our chosen mechanism for display grouping atm.

robdodson:
Just as an aside, we should update the text for the accessibility audit so it doesn't say "These apps validate that your apps work for all users."

We should probably phrase it as, "These tests check for common accessibility issues."

patrickhulce:
After a talk with Dmitry and doing some more thinking, I think it might be better in the immediate term to instead just create meta audits for these particular use cases, leverage the details rendering (and avoid creating two distinct rendering systems), and leave audits as individually renderable items for now. Taking stabs at potential future use cases and increasing the complexity now seems like a bad idea when we can easily solve our problems pre-IO with an already used solution. Basically, how about we do what we originally planned to do for the a11y audits instead of concocting a new `aggregation` object we're stuck with for a long time :)

Thoughts @brendankenny ?

patrickhulce:
Closing in favor of a CategoryRenderer that is aware of individual audits that is forth-coming.

Issue #1969, opened 2017-04-05
waitForNetworkIdle gets hosed by main thread activity


patrickhulce:
DevTools does not send `RequestStarted` events while the main thread is busy. This means that for a site that downloads a hefty script as its last resource, Lighthouse will stop recording if the parse and execution time exceeds our PAUSE_AFTER_LOAD value which is currently 500ms.

I have a host of example traces of `https://travis-ci.org/GoogleChrome/lighthouse/branches` where we normally stop recording ~8 seconds before the page is actually finished loading due to a 3s script execution that kicks off all of the actual page resources.

wardpeet:
Is this a bug we will fix or rather wait for devtools/chrome?

patrickhulce:
@wardpeet we're going to have to work around it, apparently from the DevTools side it's working as intended to keep overhead of instrumentation low. We still get the accurate timestamps of the network requests so it's really only for detecting how long to trace.

patrickhulce:
not that big of a deal anymore

PR #2031, opened 2017-04-18
fix: increase pause after load to 5s

patrickhulce:
testing a stop gap for #1969 until we have a Tracing domain event to listen for, I'll rebuild this PR a dozen times to see if we get the same failure rate on the tests too :)


patrickhulce:
well that's not a good sign already...
image

PR #2032, opened 2017-04-18
Introduces isPWA boolean

ebidel:
Part of https://github.com/GoogleChrome/lighthouse/issues/1937.

Took at stab at this. Right now, it's based off of the pwa category score being 100. Nothing more.

Also added a small UI check if you're "certified". We don't have to do that but I figured people might want to see that in a report.

screen shot 2017-04-18 at 12 56 48 pm

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -672,7 +672,7 @@ class Config {
     return this._groups;
   }
 
-  /** @type {Object<string, {title: string, description: string}>|undefined} */
+  /** @type {Object<string, {name: string, url: string, description: string, audits: !Array<{id: string, minScore: number}>}>|undefined} */
   get certifications() {
     return this._certifications;
   }
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -149,7 +149,7 @@ function validatePasses(passes, audits, rootPath) {
   });
 }
 
-function validateCategories(categories, audits, auditResults, groups) {
+function validateCategories(categories, audits, auditResults, groups, certifications) {
   if (!categories) {
     return;
   }
@@ -158,7 +158,9 @@ function validateCategories(categories, audits, auditResults, groups) {
       audits.map(audit => audit.meta.name) :
       auditResults.map(audit => audit.name);
   Object.keys(categories).forEach(categoryId => {
-    categories[categoryId].audits.forEach((audit, index) => {
+    const category = categories[categoryId];
+
+    category.audits.forEach((audit, index) => {
       if (!audit.id) {
         throw new Error(`missing an audit id at ${categoryId}[${index}]`);
       }
@@ -175,6 +177,31 @@ function validateCategories(categories, audits, auditResults, groups) {
         throw new Error(`${audit.id} references unknown group ${audit.group}`);
       }
     });
+
+    // Validate certification (if any).
+    if (category.certification) {
+      const certificationId = category.certification;
+      const certification = certifications[certificationId];
+      if (!certification) {
+        throw new Error(`${categoryId} references unknown certification ${certificationId}`);
+      }
+
+      certification.audits.forEach((certAudit, index) => {
+        if (!certAudit.id) {
+          throw new Error(`missing an audit id at certification ${certificationId}[${index}]`);
+        }
+
+        if (!certAudit.minScore) {
+          throw new Error(
+              `missing a minScore for '${certAudit.id}' in certification ${certificationId}`);
+        }
+
+        if (!category.audits.find(audit => audit.id === certAudit.id)) {
+          throw new Error(`certification ${certificationId} requires audit '${certAudit.id}', ` +
+              `but no such audit exists in ${categoryId}`);
+        }
+      });
+    }
   });
 }
 
@@ -325,10 +352,12 @@ class Config {
     this._artifacts = expandArtifacts(configJSON.artifacts);
     this._categories = configJSON.categories;
     this._groups = configJSON.groups;
+    this._certifications = configJSON.certifications;
 
     // validatePasses must follow after audits are required
     validatePasses(configJSON.passes, this._audits, this._configDir);
-    validateCategories(configJSON.categories, this._audits, this._auditResults, this._groups);
+    validateCategories(configJSON.categories, this._audits, this._auditResults, this._groups,
+        this._certifications);
   }
 
   /**
@@ -590,6 +619,11 @@ class Config {
   /** @type {Object<string, {title: string, description: string}>|undefined} */
   get groups() {
     return this._groups;
+  }
+
+  /** @type {Object<string, {title: string, description: string}>|undefined} */
+  get certifications() {
+    return this._certifications;
   }
 }
 
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -429,7 +429,7 @@ class Config {
     // 5. Filter out certifications that can no longer apply.
     const oldCertifications = config.certifications || {};
     config.certifications = Config.filterCertifications(oldCertifications, config.categories,
-        config.audits);
+        requestedAuditNames);
 
     return config;
   }
@@ -487,14 +487,14 @@ class Config {
    * Filter out any certifications with required audits not in auditIds.
    * @param {!Object<string, {audits: !Array<{id: string}>}>} oldCertifications
    * @param {!Object<string, {audits: !Array<{id: string}>}>} categories
-   * @param {!Array<string>} auditIds
+   * @param {!Set<string>} auditIds
    * @return {!Object<string, {audits: !Array<{id: string}>}>}
    */
   static filterCertifications(oldCertifications, categories, auditIds) {
     const certifications = {};
     Object.keys(oldCertifications).filter(certificationId => {
       const certification = oldCertifications[certificationId];
-      return certification.audits.every(audit => auditIds.includes(audit.id));
+      return certification.audits.every(audit => auditIds.has(audit.id));
     }).forEach(certificationId => {
       certifications[certificationId] = oldCertifications[certificationId];
     });
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -178,30 +178,47 @@ function validateCategories(categories, audits, auditResults, groups, certificat
       }
     });
 
-    // Validate certification (if any).
-    if (category.certification) {
-      const certificationId = category.certification;
-      const certification = certifications[certificationId];
-      if (!certification) {
-        throw new Error(`${categoryId} references unknown certification ${certificationId}`);
+    if (category.certification && !certifications[category.certification]) {
+      throw new Error(`${categoryId} references unknown certification '${category.certification}'`);
+    }
+  });
+}
+
+function validateCertifications(certifications, audits, auditResults) {
+  if (!certifications) {
+    return;
+  }
+
+  const auditIds = audits ?
+      audits.map(audit => audit.meta.name) :
+      auditResults.map(audit => audit.name);
+
+  Object.keys(certifications).forEach(certificationId => {
+    const certification = certifications[certificationId];
+
+    if (!certification.url) {
+      throw new Error(`certification ${certificationId} must have a url`);
+    }
+
+    if (!certification.description) {
+      throw new Error(`certification ${certificationId} must have a description`);
+    }
+
+    certification.audits.forEach((certAudit, index) => {
+      if (!certAudit.id) {
+        throw new Error(`missing an audit id at certification ${certificationId}[${index}]`);
       }
 
-      certification.audits.forEach((certAudit, index) => {
-        if (!certAudit.id) {
-          throw new Error(`missing an audit id at certification ${certificationId}[${index}]`);
-        }
+      if (!auditIds.includes(certAudit.id)) {
+        throw new Error(
+            `could not find ${certAudit.id} audit for certification ${certificationId}`);
+      }
 
-        if (!certAudit.minScore) {
-          throw new Error(
-              `missing a minScore for '${certAudit.id}' in certification ${certificationId}`);
-        }
-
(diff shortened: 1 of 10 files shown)


patrickhulce:
short and sweet I like it, any thoughts on exposing this at the toplevel object vs. in the category? I kinda like it contained within the category but that does mean every consumer has to do `results.reportCategories.find(cat => cat.id === 'pwa').isPwa`

ebidel:
Yea, it might be easy for consumers to check a top level prop rather than a buried obj.

Either is fine with me.

ebidel:
`isPWA` is also top level now. Kept it part of the category too so the renderer UI can access it easily instead of poking back at the report. Not thrilled about the duplication though.

paulirish:
I want to discuss how we represent the PWA score + passing threshold.  My take is we want to reward only the best best apps with a score of 100, to incentivize performance improvements. So we should set the bar for isPWA passing a bit lower.

### _A wild strawman proposal appears!_

#### Passing isPWA ✓
* **PWA score of 70, isPWA  ✓**: All PWA audits as passing. (FastEnough audit is passing). The TTI is scoring a 70/100.✜. 
* **PWA score of 85, isPWA ✓**: Same, but TTI is scoring 85/100.
* **PWA score of 100, isPWA ✓**: Same, but TTI is scoring 100/100.

#### Failing isPWA ✘
* **PWA score of 60, isPWA ✘**: All PWA audits as passing, except for FastEnough audit, which fails as TTI is scoring < 70.
* **PWA scores of 63.8 - 69.9**: Impossible. Once you're failing one of the PWA booleans, you're losing full points for that audit. (Currently we're at 11 audits, so each fail is worth ~6.3 pts, assuming even weighting.)

#### TTI Scoring
✜ TTI scoring has come up a lot, but one solution is: 70/100 is the score for whatever the "fast enough" definition of the baseline checklist says. (Currently its 10s, discussions around it moving to 5s). For example, that scoring curve could look like this: https://www.desmos.com/calculator/whus3whb2i (median at 15.7s, podr at 2s). 70% score at 10s. 80% score at 7.5s. 90% at 5.1s. (**All** the disclaimers about TTI changing apply here)

patrickhulce:
Lots to parse in there @paulirish :) but I like the idea of basically saying your final PWA score is `min(pwa.score, timeToInteractive.score)` but have the boolean still be based on `pwa.score === 100`

paulirish:
@brendankenny is gonna iterate on this. we discussed a solution and with some slight changes this will be goooooooood.

ebidel:
@brendankenny are you going add to this PR? Not sure what was discussed :)

brendankenny:
yeah, I'm going to adapt it. Plan is basically just a new thing in config to make the check more detailed but also separate from category score. Should hit most of what we want and give us flexibility if we decide to use a different approach later

ebidel:
@brendankenny do you think we'll land this for IO?

brendankenny:
added "certifications" to the config, allowing scoring of a category to be independent of the certification of it. e.g. a site can be ✔PWA but we still give them a bad score cause they're really slow or whatever.

Uses the same styling as before (just a green check).

Wanted to get implementation feedback before adding tests.

lighthouse-core/config/config.js
```diff
@@ -175,6 +177,31 @@ function validateCategories(categories, audits, auditResults, groups) {
         throw new Error(`${audit.id} references unknown group ${audit.group}`);
       }
     });
+
+    // Validate certification (if any).
+    if (category.certification) {
```
paulirish: do we want to validate any certification in the config, even if a category doesn't call it out?

lighthouse-core/report/v2/report-generator.js
```diff
@@ -118,21 +118,25 @@ class ReportGeneratorV2 {
...
 
-    // Add isPWA signal for PWA category.
-    let isPWA = false;
-    const pwaCategory = categories.find(cat => cat.id === 'pwa');
-    if (pwaCategory) {
-      isPWA = pwaCategory.score === 100;
-      pwaCategory.isPWA = isPWA;
-    }
+      // Check certification (if any).
+      let certified;
```
paulirish: isCertified

lighthouse-core/config/config.js
brendankenny: > do we want to validate any certification in the config

like just check that the audits a certification is requesting will be run? We could do that, but I'm already running into issues with filtering the config with `onlyAudits`. For this version of validation, I was going to drop the `certification` property if `onlyAudits` filtered out a required audit rather than show an error. If we want to validate any certification in the config, I'm not sure what to do in that case...strip out the certification altogether?

lighthouse-core/report/v2/report-generator.js
brendankenny: done

lighthouse-core/config/config.js
```diff
@@ -591,6 +620,11 @@ class Config {
   get groups() {
     return this._groups;
   }
+
+  /** @type {Object<string, {title: string, description: string}>|undefined} */
```
patrickhulce: this type doesn't look right?

lighthouse-core/config/default.js
```diff
@@ -646,11 +646,32 @@ module.exports = {
...
+        {"id": "content-width", "minScore": 100},
+      ]
+    }
+  },
   "categories": {
     "pwa": {
       "name": "Progressive Web App",
       "weight": 1,
       "description": "These audits validate the aspects of a Progressive Web App, as specified by the baseline [PWA Checklist](https://developers.google.com/web/progressive-web-apps/checklist).",
+      "certification": "is-pwa",
```
patrickhulce: why is it a property of a category? there seems to be no relationship at all to the category shouldn't certifications just be evaluated and returned as their own thing?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -97,17 +97,30 @@ class CategoryRenderer {
...
 
     const gaugeContainerEl = this._dom.find('.lh-score__gauge', tmpl);
     const gaugeEl = this.renderScoreGauge(category);
     gaugeContainerEl.appendChild(gaugeEl);
 
-    return this._populateScore(tmpl, score, 'numeric', category.name, category.description);
+    const element = this._populateScore(
+        tmpl, score, 'numeric', category.name, category.description);
+
+    if (certDefinitions && category.certification && category.isCertified) {
```
patrickhulce: someone once told me we should trust our closure types and not be defensive beyond them... ;)

lighthouse-core/report/v2/report-generator.js
```diff
@@ -118,7 +118,19 @@ class ReportGeneratorV2 {
...
 
       const categoryScore = ReportGeneratorV2.arithmeticMean(audits);
-      return Object.assign({}, category, {audits, score: categoryScore});
+
+      // Check certification (if any).
+      let isCertified;
+      if (category.certification) {
+        // Certification and required audits existence checked by Config.
+        const certification = config.certifications[category.certification];
+        isCertified = certification.audits.reduce((result, certAudit) => {
```
patrickhulce: nit: I think Paul is not a reduce fan, could just use `.every` instead

paulirish:
I had envisioned a slight tweak on this, but lmk wyt:

* the LHR has a certifications object
* within that object is a `isCertified` boolean
* and the reportCategory references that certification by ID.

lighthouse-core/config/config.js
paulirish: i see.

okay then.  this works for me.

lighthouse-core/report/v2/report-generator.js
paulirish: i'm okay with it here. (as it is reducing down to a single value, rather than pushing items onto an array).

but now that you mention it, every() seems better. :)

lighthouse-core/report/v2/renderer/category-renderer.js
paulirish: but, backwards compat? _shrug_

patrickhulce: haha I agree @paulirish! just wanted to point it out to those who were down on it to give me some slack :P

brendankenny: > backwards compat

hehe, no :P

In this case `certDefinitions` is an optional argument (though arguably perf and a11y categories should get it too), a category might not have a certification, and we only add a check mark if the category is certified...there's no red x for failing certification

brendankenny: > perf and a11y categories should get it too

er, perf and a11y category *renderers* should get it too

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -97,17 +97,30 @@ class CategoryRenderer {
 
   /**
    * @param {!ReportRenderer.CategoryJSON} category
+   * @param {!Object<string, !ReportRenderer.CertificationJSON>=} certDefinitions
```
patrickhulce: > In this case certDefinitions is an optional argument

typedef needs fixin'

lighthouse-core/report/v2/report-generator.js
brendankenny: every is nice

lighthouse-core/config/config.js
brendankenny: no, that's a c/p mistake

brendankenny:
ok, so I think I like the idea of structural certifications, where certifications exist independent of categories, and they pass if the audits in question meet a threshold.

Questions:
- how to deal with blacklisting things via `onlyAudits` or `onlyCategories`? Audits will be removed, and suddenly certifications won't be able to pass?
- do we care about failing certifications? One way to fix the above is a LHR either gets a certification or it does't, and "doesn't" may include not meeting minimum scores (failing) *or* the audit(s) in question never ran
- How do we deal with config error checking? Hard fail upon referencing audits that don't exist is ideal, but if we go with the above, we're going to have to allow references to audits that don't exist in the config, which isn't ideal. Easy to miss a misspelled audit name and that a certification just never passes

lighthouse-core/config/default.js
brendankenny: > why is it a property of a category?

does Paul's idea sound good to you? Top level evaluated certifications, and a category can say, "hey, put a checkmark next to me based on if the certification with this name passed"?

patrickhulce:
> how to deal with blacklisting things via onlyAudits or onlyCategories? Audits will be removed, and suddenly certifications won't be able to pass?

> How do we deal with config error checking? Hard fail upon referencing audits that don't exist is ideal, but if we go with the above, we're going to have to allow references to audits that don't exist in the config, which isn't ideal. Easy to miss a misspelled audit name and that a certification just never passes

I'd say we just filter out certifications that can't be met by the result of `onlyAudits`/`onlyCategories` when we're doing the filtering stage, that way we still hard fail for misspellings and such but only run certifications when we actually can.

> do we care about failing certifications? One way to fix the above is a LHR either gets a certification or it does't, and "doesn't" may include not meeting minimum scores (failing) or the audit(s) in question never ran

IMO we should just provide them in the results JSON for now and special case the `is-pwa` and future `is-exemplary-pwa` for display as checkmarks in particular places.

Later on certifications could each appear as nav items to click and visit an overview page for that certification that shows the failures/what comprises the audit/whatever we want

lighthouse-core/config/default.js
patrickhulce: sgtm for now ya 👍

brendankenny:
updated to top level certifications. Like 15 extending corner cases to still test, tho

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -97,17 +97,33 @@ class CategoryRenderer {
...
+    const element = this._populateScore(
+        tmpl, score, 'numeric', category.name, category.description);
+
+    // If category wants a certification and the certification has passed, add a checkmark.
+    if (certifications && category.certification) {
+      const certification = certifications[category.certification];
+      if (certification.isCertified) {
+        const titleEl = element.querySelector('.lh-score__title');
+        const link = /** @type {!Element} */ (titleEl.appendChild(this._dom.createSpanFromMarkdown(
+            `[✔](${certification.url})`)));
```
patrickhulce: should we put an x if it didn't meet the requirements?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -97,17 +97,33 @@ class CategoryRenderer {
...
+        link.classList.add('lh-score__isCertified');
```
patrickhulce: nit: `__certification` or maybe `--certified` on oh `lh-score`? at the very least `__is-certified` :)

lighthouse-core/report/v2/renderer/category-renderer.js
paulirish: nah.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -320,28 +336,30 @@ class CategoryRenderer {
   /**
    * @param {!ReportRenderer.CategoryJSON} category
    * @param {!Object<string, !ReportRenderer.GroupJSON>} groups
+   * @param {!Object<string, !ReportRenderer.CertificationJSON>} certifications
    * @return {!Element}
    */
-  render(category, groups) {
+  render(category, groups, certifications) {
```
ebidel: I've been thinking that we should be passing the `report` object around instead of adding additional params to all of the render methods + piping it throughout. Each method can pull the properties it cares about. WDYT?

paulirish:
Once this boolean (via whitelisting) is in place, then we can move manifest-name-short-enough from Best Practices into PWA

paulirish:
No more demand for this fella it seems. 

As of May, a 100 PWA score === this checkmark. We don't see that changing in the short term.

brendankenny:
🎉 🎂

PR #2033, opened 2017-04-18
refactor: move instantiateComputedArtifacts to Runner

brendankenny:
very small refactor to fix something that's bothered me for a while: computed artifacts aren't part of gathering (we use them even if artifacts are provided by the config and Chrome is never run), so creating them should live in runner. Main benefit is we don't have both `runner` *and* `gather-runner` calling `instantiateComputedArtifacts`.

Also fixes a minor bug where the `computedArtifact` base class was being used as a computed artifact. Since it doesn't have a `name` property, it was showing up as `artifacts.requestundefined()` :)

originally part of #2018...but I forgot to push

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -18,7 +18,6 @@
 
 const log = require('../lib/log.js');
 const Audit = require('../audits/audit');
-const path = require('path');
 const URL = require('../lib/url-shim');
 
 /**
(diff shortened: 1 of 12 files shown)


patrickhulce:
Hm what's a small refactor when very small is 146 lines? ;) I have no complaints

brendankenny:
> Hm what's a small refactor when very small is 146 lines? ;)

ha, but a good quarter of that is just s/GatherRunner/Runner, tests are basically just switching files, etc :P

paulirish:
thanks for this cleanup. paves the way for https://github.com/GoogleChrome/lighthouse/compare/dumpartifacts ;)

PR #2034, opened 2017-04-18
Plots: Visualize metrics per site per run

### LINKED ISSUE #2011: plots: add graph-per-site page
@wwwillchen 

It would be great to have a second graph page that looks at the transpose of the current data: rather than a graph of all sites per metric, a graph of all metrics per site.

That would allow easier comparison of the values and spread of variations on metrics, for instance.

wwwillchen:
Example:
plot-per-site

Closes #2011

### DIFF
--- a/plots/README.md
+++ b/plots/README.md
@@ -23,5 +23,11 @@ $ npm run measure
 $ npm run analyze
 
 # View visualization
+
+# Sites by metric
 # Open index.html in browser
+
+# Metrics by site
+# Open plot-per-site.html in browser
 ```
+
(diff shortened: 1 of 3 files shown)


brendankenny:
hmm, I was imagining a box-and-whisker plot per metric to be able to eyeball the distributions. Plus otherwise I end up with

screen shot 2017-04-18 at 15 38 21

:)

patrickhulce:
> I was imagining a box-and-whisker

same here :) too many lines makes my eyes hurt haha

wwwillchen:
Thanks for the feedback. I'll open up a separate PR which will show box plots (metrics per site). I think this plot can be helpful for analyzing a few particular problematic runs but I'll leave it up to the team if you want to merge this or close this.

PR #2037, opened 2017-04-19
Report v2: category score gauges

ebidel:
R: all

First go at the v2 mock score badges. I'm calling these gauges. Needs to be combined with https://github.com/GoogleChrome/lighthouse/pull/2000.

Example usage: `<div class="lh-gauge" data-progress="25">`.  #wouldbeasweetwebcomponent

screen shot 2017-04-18 at 9 14 53 pm

### DIFF
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -60,16 +60,17 @@ class DOM {
       throw new Error(`Template not found: template${selector}`);
     }
 
-    const clone = this._document.importNode(template.content, true);
+    const clone = /** @type {!DocumentFragment} */ (
+        this._document.importNode(template.content, true));
 
     // Prevent duplicate styles in the DOM. After a template has been stamped
-    // for the first time, remove its styles so they're not re-added.
-    if (template.used) {
-      Array.from(clone.querySelectorAll('style')).forEach(style => style.remove());
+    // for the first time, remove the clone's styles so they're not re-added.
+    if (template.hasAttribute('data-stamped')) {
+      this.findAll('style', clone).forEach(style => style.remove());
     }
-    template.used = true;
+    template.setAttribute('data-stamped', true);
 
-    return /** @type {!DocumentFragment} */ (clone);
+    return clone;
   }
 
   /**
@@ -121,6 +122,16 @@ class DOM {
       throw new Error(`query ${query} not found`);
     }
     return result;
+  }
+
+  /**
+   * Helper for context.querySelectorAll. Returns an Array instead of a NodeList.
+   * @param {string} query
+   * @param {!DocumentFragment|!Element} context
+   * @return {!Array<Element>}
+   */
+  findAll(query, context) {
+    return Array.from(context.querySelectorAll(query));
   }
 }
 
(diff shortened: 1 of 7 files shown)


lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -60,7 +60,17 @@ class DOM {
...
       throw new Error(`Template not found: template${selector}`);
     }
-    return /** @type {!DocumentFragment} */ (this._document.importNode(template.content, true));
+
+    const clone = this._document.importNode(template.content, true);
+
+    // Prevent duplicate styles in the DOM. After a template has been stamped
+    // for the first time, remove its styles so they're not re-added.
+    if (template.used) {
+      Array.from(clone.querySelectorAll('style')).forEach(style => style.remove());
```
paulirish: qSA results are iterable, so you don't need the array.from. 
https://www.chromestatus.com/feature/5691608351637504

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -148,6 +148,33 @@ class ReportRenderer {
   }
 
   /**
+   * @param {!ReportRenderer.CategoryJSON} category
+   * @return {!DocumentFragment}
+   */
+  _renderScoreGauge(category) {
```
patrickhulce: I'm not really in love with a bunch of isolated components filling `report-renderer.js`, but there's not really an obvious place for them to go now that details seems to not be what I thought it was haha

lighthouse-core/report/v2/renderer/dom.js
wardpeet: this will fail though if we have a report-viewer online and people watch it in edge.

My point it doesn't harm to have it here?

ebidel: Right. Sadly, I think we need to hold out for Edge support.

lighthouse-core/report/v2/renderer/report-renderer.js
ebidel: I moved it into category-renderer.js now that we have it.

lighthouse-core/report/v2/renderer/dom.js
paulirish: oh didnt know we included Edge in our supported browsers for the report. I'd be happy to exclude, but no strong feelings.

paulirish: ah. `findAll` method added. sgtm!

paulirish:
so so good. report is on its way to looking sharp. 🔪

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -112,6 +123,16 @@ class DOM {
     }
     return result;
   }
+
+  /**
+   * Helper for context.querySelectorAll. Returns an Array instead of a NodeList.
+   * @param {string} query
+   * @param {!DocumentFragment|!Element} context
+   * @return {!Array<Element>}
```
brendankenny: should be `{!Array<!Element>}`. Technically fine here, since `Element` is part of `null|Element`, but closure should be complaining where this is being called and null isn't checked.

Looks like the typing is being lost in the arrow functions passed to `forEach`, but the old type inference isn't complaining and the NTI has too many bugs elsewhere to turn on yet :(

PR #2023, opened 2017-04-17
feat: add consistently interactive audit

patrickhulce:
https://docs.google.com/document/d/1GGiI9-7KeY3TPqS3YT271upUVimo-XiL5mwWorDUD4c/preview

PR #2040, opened 2017-04-19
Add audit to check if start_url is cached by SW

wardpeet:
I couldn't reuse the offline gatherer as it checks the current page from being cached by SW or not for the works-offline audit.

I added a start-url gatherer which lives in a pass other than offline as I need the manifest to work which fails if the current requested page is offline.

### DIFF
--- a/lighthouse-cli/test/smokehouse/pwa-config.js
+++ b/lighthouse-cli/test/smokehouse/pwa-config.js
@@ -36,14 +36,14 @@ module.exports = {
     recordNetwork: true,
     gatherers: [
       'service-worker',
-      'offline'
+      'offline',
+      'start-url'
     ]
   },
   {
     gatherers: [
       'dobetterweb/domstats',
-      'http-redirect',
-      'start-url',
+      'http-redirect'
     ]
   }],
 
--- a/lighthouse-cli/test/smokehouse/pwa-config.js
+++ b/lighthouse-cli/test/smokehouse/pwa-config.js
@@ -42,7 +42,8 @@ module.exports = {
   {
     gatherers: [
       'dobetterweb/domstats',
-      'http-redirect'
+      'http-redirect',
+      'start-url',
     ]
   }],
 
--- a/lighthouse-cli/test/smokehouse/pwa-expectations.js
+++ b/lighthouse-cli/test/smokehouse/pwa-expectations.js
@@ -93,46 +93,6 @@ module.exports = [
   },
 
   {
-    initialUrl: 'https://www.chromestatus.com/features',
-    url: 'https://www.chromestatus.com/features',
-    audits: {
-      'dom-size': {
-        score: 100,
-        extendedInfo: {
-          value: {
-            1: {value: '20'},
-            2: {snippet: /ul.versionlist/}
-          }
-        }
-      },
-      'is-on-https': {
-        score: true
-      },
-      'redirects-http': {
-        score: true
-      },
-      'service-worker': {
-        score: true
-      },
-      'works-offline': {
-        score: true
-      },
-      'webapp-install-banner': {
-        score: true
-      },
-      'splash-screen': {
-        score: true
-      },
-      'themed-omnibox': {
-        score: true
-      },
-      // 'cache-start-url': {
-      //   score: true
-      // }
-    }
-  },
-
-  {
     initialUrl: 'https://jakearchibald.github.io/svgomg/',
     url: 'https://jakearchibald.github.io/svgomg/',
     audits: {
--- a/lighthouse-cli/test/smokehouse/pwa-expectations.js
+++ b/lighthouse-cli/test/smokehouse/pwa-expectations.js
@@ -78,6 +78,46 @@ module.exports = [
         score: false
       },
       'webapp-install-banner': {
+        score: false
+      },
+      'splash-screen': {
+        score: true
+      },
+      'themed-omnibox': {
+        score: true
+      },
+      // 'cache-start-url': {
+      //   score: true
+      // }
+    }
+  },
+
+  {
+    initialUrl: 'https://www.chromestatus.com/features',
+    url: 'https://www.chromestatus.com/features',
+    audits: {
+      'dom-size': {
+        score: 100,
+        extendedInfo: {
+          value: {
+            1: {value: '20'},
+            2: {snippet: /ul.versionlist/}
+          }
+        }
+      },
+      'is-on-https': {
+        score: true
+      },
+      'redirects-http': {
+        score: true
+      },
+      'service-worker': {
+        score: true
+      },
+      'works-offline': {
+        score: true
+      },
+      'webapp-install-banner': {
         score: true
       },
       'splash-screen': {
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -24,6 +24,7 @@ module.exports = {
     "gatherers": [
       "service-worker",
       "offline",
+      "start-url",
     ]
   },
   {
@@ -32,7 +33,6 @@ module.exports = {
     "gatherers": [
       "http-redirect",
       "html-without-javascript",
-      "start-url",
     ]
   }, {
     "passName": "dbw",
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -23,7 +23,7 @@ module.exports = {
     "useThrottling": false,
     "gatherers": [
       "service-worker",
-      "offline"
+      "offline",
     ]
   },
   {
@@ -31,7 +31,8 @@ module.exports = {
     "useThrottling": false,
     "gatherers": [
       "http-redirect",
-      "html-without-javascript"
+      "html-without-javascript",
+      "start-url",
     ]
   }, {
     "passName": "dbw",
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -711,20 +711,26 @@ class Driver {
     });
   }
 
-  endNetworkCollect() {
+  endNetworkCollect(opts = {}) {
+    const pauseAfterLoadMs = (opts.flags && opts.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
     return new Promise((resolve, reject) => {
-      this.off('Network.requestWillBeSent', this._networkRecorder.onRequestWillBeSent);
-      this.off('Network.requestServedFromCache', this._networkRecorder.onRequestServedFromCache);
-      this.off('Network.responseReceived', this._networkRecorder.onResponseReceived);
-      this.off('Network.dataReceived', this._networkRecorder.onDataReceived);
-      this.off('Network.loadingFinished', this._networkRecorder.onLoadingFinished);
-      this.off('Network.loadingFailed', this._networkRecorder.onLoadingFailed);
-      this.off('Network.resourceChangedPriority', this._networkRecorder.onResourceChangedPriority);
+      const waitForNetworkIdle = this._waitForNetworkIdle(pauseAfterLoadMs).promise;
 
-      resolve(this._networkRecords);
+      waitForNetworkIdle.then(() => {
+        const networkRecorder = this._networkRecorder;
+        this.off('Network.requestWillBeSent', networkRecorder.onRequestWillBeSent);
+        this.off('Network.requestServedFromCache', networkRecorder.onRequestServedFromCache);
+        this.off('Network.responseReceived', networkRecorder.onResponseReceived);
+        this.off('Network.dataReceived', networkRecorder.onDataReceived);
+        this.off('Network.loadingFinished', networkRecorder.onLoadingFinished);
+        this.off('Network.loadingFailed', networkRecorder.onLoadingFailed);
+        this.off('Network.resourceChangedPriority', networkRecorder.onResourceChangedPriority);
 
-      this._networkRecorder = null;
-      this._networkRecords = [];
+        resolve(this._networkRecords);
+
+        this.networkRecorder = null;
+        this._networkRecords = [];
+      }).catch(err => reject(err));
     });
   }
 
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -711,26 +711,20 @@ class Driver {
     });
   }
 
-  endNetworkCollect(opts = {}) {
-    const pauseAfterLoadMs = (opts.flags && opts.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
+  endNetworkCollect() {
     return new Promise((resolve, reject) => {
-      const waitForNetworkIdle = this._waitForNetworkIdle(pauseAfterLoadMs).promise;
+      this.off('Network.requestWillBeSent', this._networkRecorder.onRequestWillBeSent);
+      this.off('Network.requestServedFromCache', this._networkRecorder.onRequestServedFromCache);
+      this.off('Network.responseReceived', this._networkRecorder.onResponseReceived);
+      this.off('Network.dataReceived', this._networkRecorder.onDataReceived);
+      this.off('Network.loadingFinished', this._networkRecorder.onLoadingFinished);
+      this.off('Network.loadingFailed', this._networkRecorder.onLoadingFailed);
+      this.off('Network.resourceChangedPriority', this._networkRecorder.onResourceChangedPriority);
 
-      waitForNetworkIdle.then(() => {
-        const networkRecorder = this._networkRecorder;
-        this.off('Network.requestWillBeSent', networkRecorder.onRequestWillBeSent);
-        this.off('Network.requestServedFromCache', networkRecorder.onRequestServedFromCache);
-        this.off('Network.responseReceived', networkRecorder.onResponseReceived);
-        this.off('Network.dataReceived', networkRecorder.onDataReceived);
-        this.off('Network.loadingFinished', networkRecorder.onLoadingFinished);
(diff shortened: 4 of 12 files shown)


lighthouse-core/gather/gatherers/offline.js
```diff
@@ -30,9 +30,8 @@ class Offline extends Gatherer {
         record._fetchedViaServiceWorker;
     }).pop(); // Take the last record that matches.
 
-    return options.driver.goOnline(options).then(_ => {
-      return navigationRecord ? navigationRecord.statusCode : -1;
-    });
+    return options.driver.goOnline(options)
+      .then(_ => navigationRecord ? navigationRecord.statusCode : -1);
```
patrickhulce: 👏 👏 👏

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All rights reserved.
```
patrickhulce: nit: 2017 :)

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
...
+const ManifestGatherer = require('./manifest');
+
+class StartUrl extends ManifestGatherer {
+  afterPass(options) {
+    let startUrl = options.url;
+
+    return super.afterPass(options)
+      .then(manifest => {
+        if (!manifest) {
+          return Promise.reject(`Unable to retrieve manifest at ${options.url}`);
```
patrickhulce: Oh it doesn't reject if it couldn't find it?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        startUrl = manifest.value.start_url.raw;
+      })
+      .then(_ => options.driver.goOffline())
+      .then(_ => options.driver.evaluateAsync(
+          `fetch('${startUrl}').then(response => response.status)
```
patrickhulce: did we decide we're ok with this failing for CORS and https redirect issues?

lighthouse-core/test/gather/gatherers/start-url-test.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright 2016 Google Inc. All rights reserved.
```
patrickhulce: nit: 2017

lighthouse-core/config/default.js
```diff
@@ -22,15 +22,16 @@ module.exports = {
...
     ]
   },
   {
     "passName": "redirectPass",
     "useThrottling": false,
     "gatherers": [
       "http-redirect",
-      "html-without-javascript"
+      "html-without-javascript",
+      "start-url",
```
brendankenny: putting it in this pass seems like it could be problematic if this site doesn't do http->https redirect since the base url would then be the http version

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > Oh it doesn't reject if it couldn't find it?

I believe this handles the case when the page has no manifest and so `Manifest.afterPass` returns `null`. Do we want to say that more explicitly? `No web app manifest found on page ${options.url}` or something?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+'use strict';
+
+const ManifestGatherer = require('./manifest');
+
+class StartUrl extends ManifestGatherer {
+  afterPass(options) {
+    let startUrl = options.url;
+
+    return super.afterPass(options)
```
brendankenny: this is kind of horrifying :) What do you think about moving most of the manifest gatherer's `afterPass` method to `driver.getParsedManifest` or something? The only awkward part is maybe passing in `options.url`, but that's not the worst thing

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
...
+      .then(manifest => {
+        if (!manifest) {
+          return Promise.reject(`Unable to retrieve manifest at ${options.url}`);
+        }
+
+        startUrl = manifest.value.start_url.raw;
```
brendankenny: I believe we want `manifest.value.start_url.value` here and then need to reject if there isn't a `value`. 

We may also want to surface the `manifest.value.start_url.debugString` (e.g. if it's an invalid URL or cross-origin) and warn that Lighthouse is testing the default URL (`options.url` in this case) instead.

I wonder if we should show which `start_url` we tested regardless, actually, since that could be really annoying if you have an error in your `start_url` and so we use the default instead, or we're calculating the url relative to the document URL differently than the developer expects

lighthouse-core/gather/gatherers/start-url.js
brendankenny: also need to check `manifest.value` to make sure there wasn't a parse error (and reject on `manifest.debugString` in that case, maybe

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > did we decide we're ok with this failing for CORS and https redirect issues?

for now, at least, [`start_url`](https://w3c.github.io/manifest/#start_url-member) is required to be same-origin, so CORS shouldn't be an issue but https redirect could definitely be a problem

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -0,0 +1,45 @@
...
+      })
+      .then(_ => options.driver.goOffline())
+      .then(_ => options.driver.evaluateAsync(
+          `fetch('${startUrl}').then(response => response.status)
+            .catch(err => -1)`
```
brendankenny: I hated keeping the 200/-1 thing from the offline test. Is there more information we could surface here, especially in the case of failure? (definitely ok to punt this to a followup, though)

lighthouse-cli/test/smokehouse/pwa-config.js
```diff
@@ -42,7 +42,8 @@ module.exports = {
   {
     gatherers: [
       'dobetterweb/domstats',
-      'http-redirect'
+      'http-redirect',
+      'start-url',
```
brendankenny: definitely need some expectations for the smoke tests as this is going to be a tricky one :)

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > for now, at least, start_url is required to be same-origin, so CORS shouldn't be an issue but https redirect could definitely be a problem

actually, I guess whether it was cross domain or http while the `start_url` was https that would still be cross origin, so `manifest.value.start_url` would fall back to `options.url` and have a `debugString` of [`'ERROR: start_url must be same-origin as document'`](https://github.com/GoogleChrome/lighthouse/blob/2b7a3607cfa282e64eb327f82f64dc0ee5fd66e3/lighthouse-core/lib/manifest-parser.js#L138-L144)

lighthouse-core/gather/gatherers/start-url.js
wardpeet: should I change the manifest gatherer to reject instead?

Code below how it is now
```
if (!response.data) {
          if (response.url) {
            throw new Error(`Unable to retrieve manifest at ${response.url}`);
          }

          return null;
        }
```

lighthouse-core/gather/gatherers/start-url.js
wardpeet: I can always check if both urls (start_url & lighthouse url) are the same origin. If they are not I can throw an error as start_url should be same origin

lighthouse-core/config/default.js
wardpeet: so should I move it to the dbw pass or create a new one?

lighthouse-core/gather/gatherers/start-url.js
wardpeet: I don't mind moving it, I only need the url if I want to parse the manifest. So an option is to get the raw manifest inside driver and let the gatherer parse it with the lib. Which isn't a big problem as the logic is inside another file.

lighthouse-core/gather/gatherers/start-url.js
wardpeet: we shouldn't test the the original url as it's an other audit that takes care of this (offline.js). I should just fail with extra information.

lighthouse-core/gather/gatherers/start-url.js
wardpeet: if we could keep the network gatherer alive I could get a network record. If that's the case I can also move this audit back to offline pass. (i'll take this discussing offline with you)

lighthouse-cli/test/smokehouse/pwa-config.js
wardpeet: will do :)

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: yeah good point I was thinking of the LH test URL redirecting to a different domain but we're getting this from the manifest.

aside: should we stringify here in case there are non-unicode escaped `'`s? or should we throw if they have a wonky start_url in the manifest anyway?

wardpeet: we should fail if it's a wonky start_url. The manifest audit should fail as well as I guess it's not a valid start_url?

lighthouse-core/gather/gatherers/start-url.js
brendankenny: yeah, this makes sense to me

lighthouse-core/gather/gatherers/start-url.js
brendankenny: no, I think it should stay as it is there, because not having a manifest for other audits isn't an error, there just isn't one. Not having a manifest here is also arguably not an error, but I'm fine with treating it as one for now

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > we shouldn't test the the original url as it's an other audit that takes care of this (offline.js). I should just fail with extra information.

I think we actually need to use it.

For all error cases (missing `start_url` or invalid URL, etc), the web app manifest spec has the value of `start_url` fall back to the URL of the document instead, so I think that's what we need to do. As long as the manifest exists and is JSON, there will always be a `start_url` on the parsed result.

ManifestParser takes care of all that ([here](https://github.com/GoogleChrome/lighthouse/blob/2faa6867d03130b8331062dea8c0a7dc6938246d/lighthouse-core/lib/manifest-parser.js#L111-L144)), so this code doesn't need to handle any of this, it just might want to surface `manifest.value.start_url.debugString` as a warning if there was one

brendankenny: > then need to reject if there isn't a `value`

I was actually wrong here. As above, as long as there is *some* manifest made up of *some* kind of native JSON, there will be a `manifest.value.start_url.value` with a valid URL on it

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > I can always check if both urls (start_url & lighthouse url) are the same origin. If they are not I can throw an error as start_url should be same origin as @brendankenny said but don't let the fetch fail I can just throw earlier.

just to be clear, I don't think this gatherer/audit should check this, since `ManifestParser` already took care of all that. I would say test the `start_url` and we should pipe through the `debugString` if there is one notifying if manifestParser noticed anything weird about their manifest

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > if we could keep the network gatherer alive I could get a network record. 

as discussed, let's file an issue on this and do it in the future. I think it'll be really useful beyond this gatherer, too

lighthouse-core/config/default.js
paulirish: we think it actually could live in the offlinePass. order of all the afterPasses should work!

plus nicer for people that only do PWA stuff. just 2 passes instead of 3.

wardpeet: Sadly it can not :( As I need the start url from the manifest and I can't retrieve the manifest unless the page loaded so if I go offline and the page can not be fetched I cannot get the manifest anymore.

Another option is to fail with a debugstring as I could not get the manifest?

lighthouse-cli/test/smokehouse/pwa-expectations.js
```diff
@@ -78,6 +78,46 @@ module.exports = [
...
+        score: true
+      },
+      // 'cache-start-url': {
+      //   score: true
+      // }
+    }
+  },
+
+  {
+    initialUrl: 'https://www.chromestatus.com/features',
```
patrickhulce: I'm a little confused, why is this being added twice? Maybe the GitHub diff is just misleading

lighthouse-core/gather/driver.js
```diff
@@ -675,20 +697,26 @@ class Driver {
     });
   }
 
-  endNetworkCollect() {
+  endNetworkCollect(opts = {}) {
+    const pauseAfterLoadMs = (opts.flags && opts.flags.pauseAfterLoad) || PAUSE_AFTER_LOAD;
```
patrickhulce: don't love that we have to have this logic twice :/ also seems a little strange that we have to wait for idle again since we could have explicitly timed out trying to wait for network idle the first time, I'm guessing you want to be sure you're not giving up while it's being requested?

patrickhulce: also as an aside, the implementation of consistently interactive (#2023) increases the pauseafterload to 5s which would be kinda painful to wait an extra 5s here too

lighthouse-cli/test/smokehouse/pwa-expectations.js
wardpeet: I did it twice, the only difference is `webapp-install-banner` XD
Or maybe a a smoketest is enough?

paulirish:
We need to run this new PWA checks including this across a bunch of PWAs to be confident we're happy with this.

@wardpeet wanna run that if we give you like 50 URLs? :)

wardpeet:
@paulirish sure send me a list! :)

paulirish:
@wardpeet just sent you a list

merging this so that we can get more eyes on it (and its results).

fchristant:
I hope this is the right place to offer some feedback on this new check. I'm specifically triggered because yesterday I scored 100% on the PWA check, yet this morning it fails due to this new check. I don't care too much about the score, but the bigger issue is that my PWA no longer classifies as installable.

My SW strategy is very modest and basic. I may extend upon it in the future, but right now my strategy is to simply shows users a dedicated "offline" page when they are disconnected. So naturally, I cache that offline page's content during the install step. My SW simply passes through every network request to the network stack, and only if it fails and it was a HTML request, do I redirect to the cached 'offline' page.

To further satisfy PWA requirements, I'm setting a start_url, to the only value that makes sense in my case: '/'. It does not make sense in my strategy to cache '/', since my SW will never show a cached version of '/'. But I'm caching it anyway, just to see if that passes this check. It doesn't. 

Here's my SW code:

https://gist.github.com/fchristant/440ea5077db60fff2a14108ce152ca0b

It seems to me that the error text is misleading in this setup. I do cache the start URL, but the check does not seem to detect it. Does this mean my PWA really does not qualify as installable? Or does it pass and is this a false positive?

wardpeet:
@fchristant thanks for reaching out! Do you have a live url we can test it on? If not I can get your sw running but a live url is easier to start debugging

fchristant:
Thanks for the quick reply :) I do have a live URL but it running on a Debian VM from my Windows desktop. The URL will only be accessible as long as my machine is on. Here it is:

https://www.eon.earth/index.php/template/empty

As you can see, I'm coding a minimum template here for a new web app. This page pretty much is about testing all the stuff and setup in the HTML header. 

I'll be away for the afternoon and return in the evening, yet keeping the machine on. I hope this is enough opportunity for you to have a look. Thanks a lot in advance for any help!

fchristant:
I just saw something weird. Not sure if it makes a difference for this issue, but reporting it anyway. This is the valid test URL:

https://www.eon.earth/template/empty

If I run Lighthouse on that URL (from the extension), by the time it is done the URL has become:

https://www.eon.earth/index.php/template/empty

I wonder how that happened? Or whether it matters?

The back-end of this app is PHP CodeIgniter. All URLs go through a front controller which is index.php. However, I have specific URL rewriters in place at Apache to hide this front controller so that the URLs become prettier. I'm just surprised/curious what triggers this strange redirect.

wardpeet:
It does matter as the url isn't the same anymore and lighthouse loses track of what url to look at. I'll see why the redirect happens

fchristant:
Just checking in to see if there's any news?

fchristant:
Reporting on a few experiments I did:

Setting my start URL in the manifest to the full absolute path of the root domain (instead of using '/') and matching the exact same path in my cache URL makes no difference. This test is to exclude relative/absolute paths being a factor. 

Temporarily setting my 'offline' page to be the root page and responding with that when being offline, makes no difference either. 

Being more explicit about named caches makes no difference.

It seems that Lighthouse in this scenario doesn't detect what is in cache at all, no matter what I put in it. Which leaves me wondering...is this strategy fundamentally incompatible with this check, or is the check itself faulty (a false positive)? 

It's a highly common recipe that is repeated all over the web in cookbooks, MDN docs, etc.

PR #2041, opened 2017-04-19
Plots: Visualize metrics per site with box plots

### LINKED ISSUE #2011: plots: add graph-per-site page

wwwillchen:
Based on the feedback from https://github.com/GoogleChrome/lighthouse/pull/2034, this is another take on visualizing perf metrics per site using box plots. It does a better job of showing many runs for a given site, but doesn't readily show how a particular run may affect multiple metrics (couldn't find a plotly option to do this).

**Example:**
box-plots-metrics-per-site

Closes #2011

### DIFF
--- a/plots/metrics-per-site.html
+++ b/plots/metrics-per-site.html
@@ -1,0 +1,29 @@
+<!--
+Copyright 2017 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


patrickhulce:
b-e-a-utiful

PR #2002, opened 2017-04-13
Adjust report-renderer init for devtools

paulirish:
Gettin the new report going in DevTools and wanted some changes like these.

1. Since I'm subclassing, I need to pass in those subclasses.
1. Globals weren't being provided automatically. (I think because devtools uses eval?)

e.g. my subclass currently looks like (UPDATED):
image

Screenshot btw: 
image

PR #2043, opened 2017-04-20
add Closure type checking for report v2

brendankenny:
@paulirish already took this most of the way there in #2002, this is just cleaning up the rest and hitting some pretty strict rules, which is nice. It has to be manually run right now (`npm run closure`)...I wasn't sure if I should enable it or not yet in CI.

Two things that might be controversial:
1) `DetailsJSON` has been flattened into basically a base class (`DetailsJSON` with a `type` and optional `text` field) and derived classes `ListDetailsJSON`, and `CardDetailsJSON` (names are terrible :). Inheritance isn't actually used, however.

   Precipitating reason is that you can't do circular typedefs, so these weren't being checked at all. DevTools compilation didn't notice because it's not using the NewTypeInference engine (to be fair, we aren't either by default because of some lingering bugs). We *could* do an actual `@interface`, which handles circular references, but we aren't actually using the recursive nature of `DetailsJSON` anywhere and it really just makes it harder to figure out what type any particular place is dealing with. If every property on `DetailsJSON` was going to be optional except for `type`, and `type` determines which properties are defined on a particular instance and what function deals with rendering it...that sounds  a whole lot like a separate type, so lets just call it that.

   When dispatched from `_render` to their particular `_render*` method, they're cast to the specific type needed. Whole thing works because these are structural types, so as long as they fit the structure they're good to go.

   For lists and cards I tried to narrow the property types to how they're actually used. `_renderBlock` is never actually used, so I have no information there, so I just deleted rather than deal with it. Next PR that needs blocks can add it back :)

2) As mentioned offline to some, I've replaced instances of `context.querySelector(query)` (which can return `null` if nothing is found) with `DOM.find(context, query)` which never returns null so we don't have to constantly assure closure that it's not null. We use it for finding stuff in templates, so if the target isn't found we either screwed up the template or the query, and `DOM.find` helpfully throws an exception in that case.

### DIFF
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -55,7 +55,7 @@ class DOM {
    * @throws {Error}
    */
   cloneTemplate(selector, context) {
-    const template = /** @type {HTMLTemplateElement} */ (context.querySelector(selector));
+    const template = /** @type {?HTMLTemplateElement} */ (context.querySelector(selector));
     if (!template) {
       throw new Error(`Template not found: template${selector}`);
     }
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -101,11 +101,11 @@ class DOM {
   /**
    * Guaranteed context.querySelector. Always returns an element or throws if
    * nothing matches query.
-   * @param {!DocumentFragment|!Element} context
    * @param {string} query
+   * @param {!DocumentFragment|!Element} context
    * @return {!Element}
    */
-  static find(context, query) {
+  find(context, query) {
     const result = context.querySelector(query);
     if (result === null) {
       throw new Error(`query ${query} not found`);
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -22,6 +22,7 @@ class DOM {
    * @param {!Document} document
    */
   constructor(document) {
+    /** @private {!Document} */
     this._document = document;
   }
 
@@ -33,9 +34,7 @@ class DOM {
    *     set the attribute on the node.
    * @return {!Element}
    */
-  createElement(name, className, attrs) {
-    // TODO(all): adopt `attrs` default arg when https://codereview.chromium.org/2821773002/ lands
-    attrs = attrs || {};
+  createElement(name, className, attrs = {}) {
     const element = this._document.createElement(name);
     if (className) {
       element.className = className;
@@ -56,7 +55,7 @@ class DOM {
    * @throws {Error}
    */
   cloneTemplate(selector, context) {
-    const template = context.querySelector(selector);
+    const template = /** @type {HTMLTemplateElement} */ (context.querySelector(selector));
     if (!template) {
       throw new Error(`Template not found: template${selector}`);
     }
@@ -80,7 +79,7 @@ class DOM {
 
       // Append link if there are any.
       if (linkText && linkHref) {
-        const a = this.createElement('a');
+        const a = /** @type {!HTMLAnchorElement} */ (this.createElement('a'));
         a.rel = 'noopener';
         a.target = '_blank';
         a.textContent = linkText;
@@ -97,6 +96,21 @@ class DOM {
    */
   document() {
     return this._document;
+  }
+
+  /**
+   * Guaranteed context.querySelector. Always returns an element or throws if
+   * nothing matches query.
+   * @param {!DocumentFragment|!Element} context
+   * @param {string} query
+   * @return {!Element}
+   */
+  static find(context, query) {
+    const result = context.querySelector(query);
+    if (result === null) {
+      throw new Error(`query ${query} not found`);
+    }
+    return result;
   }
 }
 
--- a/lighthouse-core/report/v2/renderer/report-renderer.js
+++ b/lighthouse-core/report/v2/renderer/report-renderer.js
@@ -22,7 +22,7 @@
  * Dummy text for ensuring report robustness: </script> pre$`post %%LIGHTHOUSE_JSON%%
  */
 
-/* globals self */
+/* globals self, DOM */
 
 const RATINGS = {
   PASS: {label: 'pass', minScore: 75},
@@ -60,9 +60,11 @@ class ReportRenderer {
    * @param {!DetailsRenderer} detailsRenderer
    */
   constructor(dom, detailsRenderer) {
+    /** @private {!DOM} */
     this._dom = dom;
+    /** @private {!DetailsRenderer} */
     this._detailsRenderer = detailsRenderer;
-
+    /** @private {!Document|!Element} */
     this._templateContext = this._dom.document();
   }
 
@@ -73,7 +75,7 @@ class ReportRenderer {
   renderReport(report) {
     try {
       return this._renderReport(report);
-    } catch (e) {
+    } catch (/** @type {!Error} */ e) {
       return this._renderException(e);
     }
   }
@@ -88,13 +90,13 @@ class ReportRenderer {
    */
   _populateScore(element, score, scoringMode, title, description) {
     // Fill in the blanks.
-    const valueEl = element.querySelector('.lh-score__value');
+    const valueEl = DOM.find(element, '.lh-score__value');
     valueEl.textContent = formatNumber(score);
     valueEl.classList.add(`lh-score__value--${calculateRating(score)}`,
-                          `lh-score__value--${scoringMode}`);
+        `lh-score__value--${scoringMode}`);
 
-    element.querySelector('.lh-score__title').textContent = title;
-    element.querySelector('.lh-score__description')
+    DOM.find(element, '.lh-score__title').textContent = title;
+    DOM.find(element, '.lh-score__description')
         .appendChild(this._dom.createSpanFromMarkdown(description));
 
     return /** @type {!Element} **/ (element);
@@ -128,7 +130,7 @@ class ReportRenderer {
     }
 
     // Append audit details to header section so the entire audit is within a <details>.
-    const header = tmpl.querySelector('.lh-score__header');
+    const header = /** @type {!HTMLDetailsElement} */ (DOM.find(tmpl, '.lh-score__header'));
     header.open = audit.score < 100; // expand failed audits
     if (audit.result.details) {
       header.appendChild(this._detailsRenderer.render(audit.result.details));
@@ -218,15 +220,17 @@ if (typeof module !== 'undefined' && module.exports) {
 
 /**
  * @typedef {{
- *     id: string, weight:
- *     number, score: number,
+ *     id: string,
+ *     weight: number,
+ *     score: number,
  *     result: {
  *       description: string,
  *       displayValue: string,
  *       helpText: string,
  *       score: (number|boolean),
  *       scoringMode: string,
- *       details: (!DetailsRenderer.DetailsJSON|!DetailsRenderer.CardsDetailsJSON|undefined)
+ *       optimalValue: number,
+ *       details: (!DetailsRenderer.DetailsJSON|undefined)
  *     }
  * }}
  */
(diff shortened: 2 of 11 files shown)


brendankenny:
> DevTools compilation didn't notice because it's not using the NewTypeInference engine (to be fair, we aren't either by default because of some lingering bugs). 

`new_type_inf` is commented out here. NTI is way better at many things: inferring types, noticing where code is tightening a type, supporting more ES6 stuff, etc, but there are some things it doesn't support yet, and since it's stricter with "there's a mistake here" -> "that's an error" (as opposed to "well, you probably don't want me to be a bother"), we can't really use it yet.

I have used it to set up the typing in this PR, but disabled it for now. Uncomment that line to see its output. Some examples:
- it doesn't support `@implicitCast` yet, which is needed to not see an error on setting `textContent`.
- it gets confused by the default parameter on `attrs = {}` meaning the param can't be undefined anymore
- it loses type information of what's in an array in the callback sent to `Array.prototype.filter` and when destructuring

Notably the old type inference *also* gets confused by the last two, it just doesn't complain, which is arguably worse.

The Closure Compiler team is doing a large push on NTI right now, most especially finishing the ES6->ES6 compiler so type checking can be done on the actual code instead of transpiled ES5 code, so I'm hoping that our particular bugs will be squashed in the next month or two.

<details>
<summary>Full NTI output:</summary>

```
> lighthouse@1.7.0-alpha.1 closure
> cd lighthouse-core && node closure/closure-type-checking.js

gulp-google-closure-compiler: /report/v2/renderer/details-renderer.js:52: WARNING - The right side in the assignment is not a subtype of the left side.
Expected : string
Found    : string|undefined
More details:
The found type is a union that includes an unexpected type: undefined
    element.textContent = text.text;
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/details-renderer.js:83: WARNING - Attempt to use nullable type Node|null.
      element.appendChild(this._dom.createElement('summary')).textContent = details.header.text;
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/details-renderer.js:94: WARNING - Attempt to use nullable type Node|null.
      card.appendChild(titleEl).textContent = item.title;
      ^^^^

/report/v2/renderer/details-renderer.js:95: WARNING - Attempt to use nullable type Node|null.
      card.appendChild(valueEl).textContent = item.value;
      ^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/details-renderer.js:98: WARNING - Attempt to use nullable type Node|null.
        card.appendChild(targetEl).textContent = `target: ${item.target}`;
        ^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/dom.js:40: WARNING - This GETPROP expression has the unknown type.
      ^^^^^^^^^^^^^^^^^

/report/v2/renderer/dom.js:43: WARNING - Attempt to use nullable type IObject<string,string|undefined>|undefined.
      const value = attrs[key];
                    ^^^^^

/report/v2/renderer/dom.js:77: WARNING - This GETPROP expression has the unknown type.
      const [preambleText, linkText, linkHref] = parts.splice(0, 3);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/dom.js:77: WARNING - Violation: References to object props that are typed as `unknown` are discouraged. See https://github.com/google/closure-compiler/wiki/JS-Conformance-Framework#banunknowntypedclasspropsreferences
The property "value" on type "IIterableResult<?>"
      const [preambleText, linkText, linkHref] = parts.splice(0, 3);
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/dom.js:81: WARNING - This AND expression has the unknown type.
      if (linkText && linkHref) {
          ^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:182: WARNING - Property score never defined on any type in the program
    const passedAudits = category.audits.filter(audit => audit.score === 100);
                                                         ^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:182: WARNING - This GETPROP expression has the unknown type.
    const passedAudits = category.audits.filter(audit => audit.score === 100);
                                                         ^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:183: WARNING - This NAME expression has the unknown type.
    const nonPassedAudits = category.audits.filter(audit => !passedAudits.includes(audit));
                                                                                   ^^^^^

/report/v2/renderer/report-renderer.js:185: WARNING - This GETPROP expression has the unknown type.
    for (const audit of nonPassedAudits) {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:185: WARNING - Violation: References to object props that are typed as `unknown` are discouraged. See https://github.com/google/closure-compiler/wiki/JS-Conformance-Framework#banunknowntypedclasspropsreferences
The property "value" on type "IIterableResult<?>"
    for (const audit of nonPassedAudits) {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:197: WARNING - This GETPROP expression has the unknown type.
    for (const audit of passedAudits) {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/report/v2/renderer/report-renderer.js:197: WARNING - Violation: References to object props that are typed as `unknown` are discouraged. See https://github.com/google/closure-compiler/wiki/JS-Conformance-Framework#banunknowntypedclasspropsreferences
The property "value" on type "IIterableResult<?>"
    for (const audit of passedAudits) {
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

0 error(s), 17 warning(s), 96.4% typed

Closure compilation successful.
```
</details>

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -73,7 +75,7 @@ class ReportRenderer {
   renderReport(report) {
     try {
       return this._renderReport(report);
-    } catch (e) {
+    } catch (/** @type {!Error} */ e) {
```
brendankenny: this is an example of stuff not needed with NTI

lighthouse-core/test/report/v2/renderer/details-renderer-test.js
```diff
@@ -95,31 +81,6 @@ describe('DetailsRenderer', () => {
       assert.equal(items.children.length, 3, 'did not render children');
     });
 
-    it('renders nested structures', () => {
```
patrickhulce: why? even if we don't care about blocks we certainly care about text within lists

brendankenny: > why? even if we don't care about blocks we certainly care about text within lists

oh, I didn't actually keep reading and see the list part :)

brendankenny:
> I'm not liking the idea of already nuking recursive rendering as that was kinda the point of this structure in the first place. Table will definitely need this and list should already need it for text.

This PR was aiming to satisfy two constraints: do actual type checking and don't change how things are currently being used. And just to be clear, a DAG of typedefs works fine (e.g. lists have items that are text or urls or images or etc...just not items that are lists or things that contain lists).

Currently the only thing actually using the recursive structure was the unused `_renderBlocks`. No render method (including `_renderBlocks`) was treating the `header` as an actual `DetailsJSON` (nothing was even checking the `type` of header to make sure it was a `text` node, let alone pass it back to `_render`).

Only some types had `items` defined, and then you had to check at runtime if it was type `cards` or type `list` to know what was actually in those items. That's not (statically) checking types anymore if we call all those things by the same name. If every `_render*` method is going to be bespoke and only call back into `_render` for a limited number of other types, that's easily represented by a DAG.

And if and when we need them, we can switch over to actual record types/interfaces instead of typedefs to support circular references.

lighthouse-core/test/report/v2/renderer/details-renderer-test.js
brendankenny: actually, without the block part it's just a duplicate of the `renders lists with headers` test above, so it's probably fine to remove and get more detailed list tests in #2019

patrickhulce:
> No render method (including _renderBlocks) was treating the header as an actual DetailsJSON (nothing was even checking the type of header to make sure it was a text node, let alone pass it back to _render).

This is a bad thing, not something we should cement with type-checking that won't support the proper method.

> If every _render* method is going to be bespoke and only call back into _render for a limited number of other types, that's easily represented by a DAG.

If this is really going to be the case then I misunderstood what direction details was headed and definitely do not like it 😆  I'll dismiss my review because I think I'm out of sync with how the new world of LH front-end is taking shape...

brendankenny:
> If this is really going to be the case then I misunderstood what direction details was headed and definitely do not like it 😆 I'll dismiss my review because I think I'm out of sync with how the new world of LH front-end is taking shape...

yeah, there's probably a meta issue here: "wtf is `details`" :)

The current queue of report PRs just send us further down this path, so we should probably figure that out first if we don't want to go that way

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -56,7 +55,7 @@ class DOM {
    * @throws {Error}
    */
   cloneTemplate(selector, context) {
-    const template = context.querySelector(selector);
+    const template = /** @type {HTMLTemplateElement} */ (context.querySelector(selector));
```
paulirish: compiling this PR in devtools i get this error:

```
/Users/paulirish/chromium/src/third_party/WebKit/Source/devtools/front_end/audits2/lighthouse/renderer/dom.js:58: ERROR - Type "HTMLTemplateElement" nullability not marked explicitly with "?" (nullable) or "!" (non-nullable)
    const template = /** @type {HTMLTemplateElement} */ (context.querySelector(selector));
```

that's the only error though!

brendankenny: interesting! The devtools compiler must have an extra check for explicit nullability?

brendankenny: fixed! made it explicitly nullable so the following if statement makes sense :)

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -88,13 +90,13 @@ class ReportRenderer {
    */
   _populateScore(element, score, scoringMode, title, description) {
     // Fill in the blanks.
-    const valueEl = element.querySelector('.lh-score__value');
+    const valueEl = DOM.find(element, '.lh-score__value');
```
paulirish: need to use the instance because we can't rely on this global being present.
so i guess let's drop the static and use it as `this._dom.find(...`

also, suggestion on alternative method signature above...

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -98,6 +97,21 @@ class DOM {
...
   }
+
+  /**
+   * Guaranteed context.querySelector. Always returns an element or throws if
+   * nothing matches query.
+   * @param {!DocumentFragment|!Element} context
+   * @param {string} query
+   * @return {!Element}
+   */
+  static find(context, query) {
```
paulirish: to be more like https://www.w3.org/TR/selectors-api2/ it would make more sense to do `find(query, context)`.
i'd prefer it, as well.

lighthouse-core/report/v2/renderer/dom.js
paulirish: thx. 
image

brendankenny:
ok, `find(query, context)` now has good param order and is no longer static. Also added closure compiler check into travis and `npm test`

brendankenny:
> Also added closure compiler check into travis

currently, at least, this is adding 5-6 seconds on travis, so currently tied with `lint` for our fastest CI test :)

paulirish:
Yeah closure compilation is back! and on travis!

⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️ ⛵️

PR #2045, opened 2017-04-20
browserify axe-core as a string, not a Buffer

brendankenny:
@paulirish 

removes the slightly shady looking inlining of axe-core as a `Buffer` from the browserified output:

`const axe=Buffer("LyohIGFYZSB2Mi4xLjcKICogQ29weXJpZ2h0IChjKSAyMDE2IERlcXVlIFN5c3...`

and uses a regular string instead:

`const axe="/*! aXe v2.1.7\n * Copyright (c) 2016 Deque Systems, Inc.\n *\n * You...`

much more clear about what's going on there and actually saves 37KB in `lighthouse-background.js`!

may be a little faster too since it doesn't have to `.toString()` the polyfilled `Buffer`, but I don't see any measurable difference in the face of the axe testing that runs immediately afterwards :)

### DIFF
--- a/lighthouse-core/gather/gatherers/accessibility.js
+++ b/lighthouse-core/gather/gatherers/accessibility.js
@@ -16,18 +16,18 @@
  */
 'use strict';
 
-/* global document */
+/* global window, document */
 
 const Gatherer = require('./gatherer');
 const fs = require('fs');
-const axe = fs.readFileSync(require.resolve('axe-core/axe.min.js'), 'utf8');
+const axeLibSource = fs.readFileSync(require.resolve('axe-core/axe.min.js'), 'utf8');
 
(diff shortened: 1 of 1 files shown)


brendankenny:
> lol that's hilarious, nice find!

@paulirish spotted it :)

paulirish:
The two `axe`'s here tripped up both me and @patrickhulce so i renamed the source and made the in-page call explicitly using `window.axe`. sgty?

paulirish:
Nice fix!

brendankenny:
😠 
screen shot 2017-04-20 at 16 45 47

👼 
screen shot 2017-04-20 at 16 46 11

brendankenny:
especially cause of
screen shot 2017-04-20 at 16 51 19
:P

PR #2052, opened 2017-04-22
move key closure warnings to errors

brendankenny:
tests won't fail if all the errors are actually just warnings :)

Moved some key warnings to errors. We can move more in or some back as we see fit


lighthouse-core/closure/closure-type-checking.js
```diff
@@ -48,25 +48,24 @@ gulp.task('compile-report', () => {
     warning_level: process.env.CI ? 'QUIET' : 'VERBOSE',
     jscomp_error: [
       'checkTypes',
+      'missingProperties',
+      'accessControls',
+      'const',
+      'visibility',
+      'reportUnknownTypes',
```
paulirish: add a comment here saying we can move reportUnknownTypes back to warning if necessary?

PR #2057, opened 2017-04-24
feat: add groups to config, accessibility renderer

patrickhulce:
addresses topics item in #1937 

I'd prefer to leave UI bikeshedding for when we implement the real version that's going in DevTools, but here's a screenshot
image


lighthouse-core/config/default.js
```diff
@@ -632,41 +666,41 @@ module.exports = {
...
+        {"id": "aria-required-parent", "weight": 1, "tags": ["accessibility-aria"]},
+        {"id": "aria-roles", "weight": 1, "tags": ["accessibility-aria"]},
+        {"id": "aria-valid-attr-value", "weight": 1, "tags": ["accessibility-aria"]},
+        {"id": "aria-valid-attr", "weight": 1, "tags": ["accessibility-aria"]},
+        {"id": "audio-caption", "weight": 1, "tags": ["accessibility-correct-attributes"]},
+        {"id": "button-name", "weight": 1, "tags": ["accessibility-element-names"]},
+        {"id": "bypass", "weight": 1, "tags": ["accessibility-describe-contents"]},
+        {"id": "color-contrast", "weight": 1, "tags": ["accessibility-color-contrast"]},
+        {"id": "definition-list", "weight": 1, "tags": ["accessibility-well-structured"]},
+        {"id": "dlitem", "weight": 1, "tags": ["accessibility-well-structured"]},
```
ebidel: nit: we could shorten these to `a11y-*`. Doesn't all this stuff end up in our json dump?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -142,7 +142,16 @@ class CategoryRenderer {
    * @param {!ReportRenderer.CategoryJSON} category
    * @return {!Element}
    */
-  render(category) {
+  render(category, tags) {
```
ebidel: tags needs a doc string

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -142,7 +142,16 @@ class CategoryRenderer {
...
+    switch (category.id) {
+      case 'accessibility':
+        return this._renderAccessibilityCategory(category, tags);
+      default:
+        return this._renderDefaultCategory(category);
+    }
+  }
+
+  _renderDefaultCategory(category) {
```
ebidel: docs

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -167,6 +176,69 @@ class CategoryRenderer {
     element.appendChild(passedElem);
     return element;
   }
+
+  _renderAuditGroup(audits, tag) {
```
ebidel: docs

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -167,6 +176,69 @@ class CategoryRenderer {
...
+      const auditGroupElem = this._dom.createElement('details', 'lh-audit-group');
+      const auditGroupHeader = this._dom.createElement('div', 'lh-audit-group__header');
+      auditGroupHeader.textContent = tag.title;
+
+      const auditGroupDescription = this._dom.createElement('div', 'lh-audit-group__description');
```
ebidel: you can style the details arrow like the rest by adding a selector like https://github.com/GoogleChrome/lighthouse/blob/2ae69be650e4a8e303ac847b1fe4b076132ac567/lighthouse-core/report/v2/report-styles.css#L228-L234 and generating markup like https://github.com/GoogleChrome/lighthouse/blob/2ae69be650e4a8e303ac847b1fe4b076132ac567/lighthouse-core/report/v2/templates.html#L19-L22. `lh-score__arrow` is the gem.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -167,6 +176,69 @@ class CategoryRenderer {
...
+      const auditGroupSummary = this._dom.createElement('summary', 'lh-audit-group__summary');
+      auditGroupSummary.appendChild(auditGroupHeader);
+
+      auditGroupElem.appendChild(auditGroupSummary);
+      auditGroupElem.appendChild(auditGroupDescription);
+      audits.forEach(audit => auditGroupElem.appendChild(this._renderAudit(audit)));
+      return auditGroupElem;
+  }
+
+  _renderAccessibilityCategory(category, tags) {
```
ebidel: docs

lighthouse-core/report/v2/report-styles.css
```diff
@@ -245,6 +245,29 @@ body {
...
+}
+
+.lh-audit-group__description {
+  font-size: 16px;
+  color: var(--secondary-text-color);
+  margin-top: calc(var(--default-padding) / 2);
+}
+
+.lh-audit-group__summary::-webkit-details-marker {
+  float: right;
```
ebidel: i dont think you need these styles if using `lh-score__arrow`

lighthouse-core/config/config.js
```diff
@@ -568,6 +569,11 @@ class Config {
   get categories() {
     return this._categories;
   }
+
+  /** @type {Object<{title: string, description: string}>} */
```
patrickhulce: it's an object where the values are `{title: string, description: string}` but IIRC we actually need `Object<string,{title: string, description: string}>` now, will update

lighthouse-core/config/default.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: I saw that but wasn't convinced we wanted to reuse those, `lh-score` gives me the idea that it's a scored thing specific to audit or category when these are just display groupings. If we want to reuse the same CSS, maybe we can extract it to a different distinct reusable component without tying the category/audit dependency?

lighthouse-core/config/config.js
ebidel: Ah! That makes more sense :)

lighthouse-core/report/v2/renderer/category-renderer.js
ebidel: Yea, we can rename the CSS if you need to.

"distinct reusable component"....interesting idea :)

patrickhulce: > "distinct reusable component"....interesting idea. Tell me more

hahaha, I've never had anything against organization of code into components :)

patrickhulce:
rebased against the mega-branch lookin' fly now 😎 
image

ebidel:
Should we have the dropdown arrows on the categories? Otherwise, it's not obvious you can expand them.

screen shot 2017-04-28 at 3 18 33 pm

patrickhulce:
> Should we have the dropdown arrows on the categories? Otherwise, it's not obvious you can expand them.

Yeah we should, fixed

ebidel:
Also now noticing the arrow don't turn. They're always in the down state:

screen shot 2017-04-28 at 3 30 25 pm

lighthouse-core/config/config.js
```diff
@@ -568,6 +569,11 @@ class Config {
   get categories() {
     return this._categories;
   }
+
+  /** @type {Object<string, {title: string, description: string}>} */
```
brendankenny: I don't believe this would be nullable. Is it `|undefined`?

lighthouse-core/config/default.js
```diff
@@ -632,41 +666,41 @@ module.exports = {
...
-        {"id": "meta-refresh", "weight": 1},
-        {"id": "meta-viewport", "weight": 1},
-        {"id": "object-alt", "weight": 1},
-        {"id": "tabindex", "weight": 1},
-        {"id": "td-headers-attr", "weight": 1},
-        {"id": "th-has-data-cells", "weight": 1},
-        {"id": "valid-lang", "weight": 1},
-        {"id": "video-caption", "weight": 1},
-        {"id": "video-description", "weight": 1},
+        {"id": "accesskeys", "weight": 1, "tags": ["a11y-correct-attributes"]},
```
brendankenny: - since this is scoped, should we just call this `group` or something?
- since we're only supporting the first entry, should we just drop the array?
- since we're requiring a tag for accessibility audits, should there be a check with a clear warning?

ebidel:
Ah interesting. Looks like we need to be more specific with our selectors for that. Now that there are sub sub arrows :)

lighthouse-core/config/config.js
patrickhulce: if we're bein' real it's `*` for most of these defs since the config can stick whatever it wants here ;) but yeah could be undefined I'll add

lighthouse-core/config/default.js
patrickhulce: I'm against doing a one off config thing per category when we can re-use the same concept, but are you proposing just renaming `tags` -> `groups` ?

> since we're requiring a tag for accessibility audits, should there be a check with a clear warning?

sure

patrickhulce: > since we're only supporting the first entry, should we just drop the array?

accessibility only uses one, but forcing one already seems bad when we just planned last week to achieve the PWA goals with multiple tags on a one

patrickhulce:
> Changing to this works:
.lh-expandable-details[open] > .lh-expandable-details__summary > .lh-toggle-arrow

yup that's what I got

patrickhulce:
> how do you feel about tests for this

like we should definitely have some 👍

patrickhulce:
PTAL this is blocking #2111 :)

lighthouse-core/config/default.js
paulirish: brendan is thinking `group` making it 1:1 instead of 1:many. So each a11y audit only belongs to a single group. Same with perf.

Doing this means we wouldn't have the ability for an audit to live in two "groups", but that seems decent for now. I can't think of where we actually NEED the extensibility of 1audit:many tags.
So perhaps it's better to keep it simple for now.  

@patrickhulce is there a 1:many usecase you're keeping in mind?

brendankenny:
> accessibility only uses one, but forcing one already seems bad when we just planned last week to achieve the PWA goals with multiple tags on a one

understood, I'm just thinking if we only have like three uses of this hardcoded into each categoryRenderer, having a generic `tags` section seems like overkill (general solution for a small number of situations) and less descriptive than it could be (if we just called them what they each were). If we start approaching more uses of these than that, then a generic solution does start looking better.

lighthouse-core/config/default.js
patrickhulce: I had thought we decided to use tags in the `certifications`-style rendition of PWA checklist, but if that's a totally distinct "tag" so to speak then I'm down with 1:1 `group`

patrickhulce: talked to paul, I was hearing/thinking random things. certifications are to be totally separate from tags (phew!) and group sgtm 👍

paulirish:
lgtm on groups refactor

patrickhulce:
@brendankenny you appeased? :)

paulirish:
ELL GEE TEE EMM

PR #2060, opened 2017-04-25
Small status logging improvements

paulirish:
1. Add two lifecycle logs
   - 'Analyzing and running audits...'
   - 'Generating results...'
2. Exceptions caught in audits will log an warning to the console
   - Exceptions in gatherers are already emitted, so we're good there.
1. `Loading page & waiting for onload` => `Loading page & waiting for idle`. 
   - And instead of just listing gatherer classnames we add the handy text of "Collecting data for "...

screenshot: 
image

----

This was mostly motivated by this exception from the screenshot:
> consistently-interactive:warn Caught exception: TracingProcessor.getMainThreadTopLevelEvents is not a function

I was getting that exception, but was only surfaced in the report. On the CLI i had no idea it was happening.

### DIFF
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -93,6 +93,12 @@ class Runner {
         return artifacts;
       });
 
+
+      run = run.then(artifacts => {
+        log.log('status', 'Analyzing and running audits...');
+        return artifacts;
+      });
+
       // Run each audit sequentially, the auditResults array has all our fine work
       const auditResults = [];
       for (const audit of config.audits) {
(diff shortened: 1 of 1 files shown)


lighthouse-core/runner.js
```diff
@@ -202,6 +210,7 @@ class Runner {
     // Fill remaining audit result fields.
     }).then(auditResult => Audit.generateAuditResult(audit, auditResult))
     .catch(err => {
+      log.warn(audit.meta.name, `Caught exception: ${err.message}`);
```
brendankenny: this is going to log gatherer errors twice, though. Not sure if we should care though (since gatherer errors are audit errors too, sort of).

paulirish: ya true. i saw that earlier. seems okay though. better chance we'll see it since the we've emitted 80 lines of "Evaluating..." since the gatherers logged out. :)

PR #2074, opened 2017-04-26
Smokehouse: log out the command we spawn with

paulirish:
image

Handy so you can retry the exact command to debug.


brendankenny:
is that going to be useful since most rely on the test server running?

paulirish:
well, for this netrecordsComputed branch i was only failing in smokehouse, and no where else.. so i needed to repro the exact situation in order to debug it.

so yes? :)

paulirish:
If you're interested.. also logging the static-server command now: 

image

wdyt?

brendankenny:
I think that's cool, though it's pretty noisy. You couldn't just run `yarn smokehouse` locally?

Issue #2075, opened 2017-04-27
report-styles.css are global


paulirish:
A devtools integration item here. :)

These styles at the top of [report-styles.css](https://github.com/GoogleChrome/lighthouse/blob/8b47006efc2e4e27753d3eed5928f55652964447/lighthouse-core/report/v2/report-styles.css#L17-L53) end up applying to all of chrome devtools. :)
The `body` rule is the particular conflict.

@ebidel we could..

1. scope these to within the existing `.lh-report` element.
2. add a new `.lh-root` element to be a little more explicit.
1. something else? (like use [`unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/unset) perhaps..)

ebidel:
4. Shadow DOM!!!!!!!!!!!!!!!!!!!!

I like `.lh-root`. Here ya go: https://github.com/GoogleChrome/lighthouse/pull/2076

PR #2076, opened 2017-04-27
Scope styles to prevent collisions with Devtools UI

ebidel:
R: @paulirish 

Fixes https://github.com/GoogleChrome/lighthouse/issues/2075

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -40,7 +40,7 @@
   box-sizing: border-box;
 }
 
-body.lh-root {
+.lh-root {
   font-family: var(--text-font-family);
   font-size: var(--body-font-size);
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/v2/report-styles.css
```diff
@@ -36,19 +36,19 @@
   --lh-category-score-width: 50px;
 }
 
-* {
+.lh-root * {
   box-sizing: border-box;
 }
 
-body {
+body.lh-root {
```
paulirish: in devtools case this won't be a body tag, so i'd favor doing just tagless `.lh-root`

ebidel: done

paulirish:
🤕 
👖   

bod-less to the max

Issue #1615, opened 2017-02-02
A11y audits: Wall of text / false positives for contrast ratio violations


pavelfeldman:
Consider delivering a screenshot with questionable areas outlined. If that is not feasible due to the false positives, consider dropping the audit. I am not aware of a good CSS-based implementation of this due to bg images. A more reliable implementation would render a screenshot with and without the text and then would compare those.

![contrast_ratio](https://cloud.githubusercontent.com/assets/883973/22571660/6bedc49c-e955-11e6-908c-e1e1ee90fc12.png)

rossPatton:
i'm also having issues with this. this check does not appear reliable or consistent

1 ) my current site passes WAVE and google accessibility audit tools here but not when using lighthouse and

2 ) elements are passing with less contrast than some elements that are failing with higher contrast. see attached

it basically feels like it's impossible to pass this test without having a black and white page. it should be removed if it's broken. or preferably fixed to be consistent with other tools

bigger text with darker background color, fails test:
![screen shot 2017-03-08 at 6 36 09 pm](https://cloud.githubusercontent.com/assets/2379901/23729104/1ef00378-042e-11e7-9301-2b70f95be6da.png)

smaller text, with lighter background color, passes test:
![screen shot 2017-03-08 at 6 36 18 pm](https://cloud.githubusercontent.com/assets/2379901/23729106/23a1a520-042e-11e7-9f1e-18e0129831c1.png)

and for comparisons sake, what the above look like when running chrome in high contrast mode

again, this passes
![screen shot 2017-03-08 at 6 39 28 pm](https://cloud.githubusercontent.com/assets/2379901/23729223/9fd7e06e-042e-11e7-9597-794b5af639a9.png)

but this doesn't
![screen shot 2017-03-08 at 6 39 34 pm](https://cloud.githubusercontent.com/assets/2379901/23729227/a913d1f6-042e-11e7-9f2a-02dd8635c2a6.png)

ebidel:
cc @robdodson

robdodson:
I'm happy to look into this but it may take a bit cuz I'm traveling to Sydney later today.

@rossPatton are you able to share a URL that I can test against?

alice:
There are two issues here:

1. axe-core returns a query selector for each node violating each rule. I have work in progress to allow it to return a live element for the top frame, which should help with this in general.
2. Potential false positives for the color contrast rule. As @robdodson said, live examples would help us and the axe-core developers dig into this issue.

[...]

PR #2090, opened 2017-04-27
Bump axe-core

robdodson:
Should fix false positives reported in #1615

I verified that the site mentioned in https://github.com/GoogleChrome/lighthouse/issues/1615#issuecomment-286830244 is no longer failing contrast checks.


patrickhulce:
that might be the most beautiful yarn.lock diff I've ever seen 😆

PR #2093, opened 2017-04-28
simplify readme launcher example

brendankenny:
this removes a few excess lines in the readme programmatic example. Maybe arguable, but I think `skipAutolaunch` is a distraction that can be saved for those who want to go deep

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -147,27 +147,21 @@ The example below shows how to setup and run Lighthouse programmatically as a No
 assumes you've installed Lighthouse as a dependency (`yarn add --dev lighthouse`).
 
 ```javascript
-const Lighthouse = require('lighthouse');
+const lighthouse = require('lighthouse');
 const ChromeLauncher = require('lighthouse/lighthouse-cli/chrome-launcher.js').ChromeLauncher;
 const Printer = require('lighthouse/lighthouse-cli/printer');
 
 function launchChromeAndRunLighthouse(url, flags, config) {
   const launcher = new ChromeLauncher({port: 9222, autoSelectChrome: true});
 
-  return launcher.isDebuggerReady()
-    .catch(() => {
-      if (flags.skipAutolaunch) {
-        return;
-      }
-      return launcher.run(); // Launch Chrome.
-    })
-    .then(() => Lighthouse(url, flags, config)) // Run Lighthouse.
+  return launcher.run() // Launch Chrome.
+    .then(() => lighthouse(url, flags, config)) // Run Lighthouse.
     .then(results => launcher.kill().then(() => results)) // Kill Chrome and return results.
     .catch(err => {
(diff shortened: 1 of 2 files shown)


brendankenny:
and if we want to code golf, also came up with
```js
function launchChromeAndRunLighthouse(url, flags, config) {
  const launcher = new ChromeLauncher({port: 9222, autoSelectChrome: true});

  const lighthousePromise = launcher.run().then(() => lighthouse(url, flags, config));
  const cleanUp = () => launcher.kill().then(() => lighthousePromise);

  return lighthousePromise.then(cleanUp, cleanUp);
}
```
which is simpler but may not seem that way :) Curious if we can make this even simpler to express

brendankenny:
(I've verified that it actually does work normally *and* actually does kill Chrome and print the error if there was an error thrown in Lighthouse)

ebidel:
Curious why you don't use `isDebuggerReady`? I've gotten into bad places without it and needed to safeguard. The connection to the debugger protocol sometimes cannot be established. Using just `launcher.run()`, sometimes it works, sometimes it doesn't.

brendankenny:
In the master version (and bin.ts) the initial `isDebuggerReady` is only checking if there's a Chrome already open and ready to talk, then the catch statement after is for if there's no Chrome open and so `isDebuggerReady` throws, at which point `launcher.run()` can be called. If it doesn't throw, nothing is launched.

To me it makes sense to remove that...bin.ts only has it because we want the CLI to support both pre-launched Chrome and launching Chrome. In theory, if you're writing your own simple Lighthouse script, you know whether or not you'll have Chrome launched already (and if you really need to support both cases, you're an advanced user who can dig in more/read our super awesome docs for our new ChromeLauncher module :)

brendankenny:
(importantly `launcher.run()` calls `isDebuggerReady` (via `waitUntilReady`) up to 10 times internally before returning)

readme.md
```diff
@@ -147,27 +147,21 @@ The example below shows how to setup and run Lighthouse programmatically as a No
 assumes you've installed Lighthouse as a dependency (`yarn add --dev lighthouse`).
 
 ```javascript
-const Lighthouse = require('lighthouse');
+const lighthouse = require('lighthouse');
 const ChromeLauncher = require('lighthouse/lighthouse-cli/chrome-launcher.js').ChromeLauncher;
```
ebidel: `const {ChromeLauncher} = require('lighthouse/lighthouse-cli/chrome-launcher');`

ebidel:
Can you update https://github.com/GoogleChrome/lighthouse/blob/master/docs/recipes/gulp/gulpfile.js#L51 too?

brendankenny:
> Can you update

done

paulirish:
@ebidel we good?

PR #2098, opened 2017-04-28
feat: new audit styles

patrickhulce:
since the devtools folks will be picking up next week figured I'd throw up what I had locally before they start. it wasn't a ton of work lost if we just close this

## Mock
image

## Impl so far

image

### DIFF
--- a/lighthouse-core/report/v2/renderer/category-renderer.js
+++ b/lighthouse-core/report/v2/renderer/category-renderer.js
@@ -56,7 +56,6 @@ class CategoryRenderer {
 
     // Append audit details to header section so the entire audit is within a <details>.
     const header = /** @type {!HTMLDetailsElement} */ (this._dom.find('.lh-score__header', tmpl));
-    header.open = audit.score < 100; // expand failed audits
     if (audit.result.details) {
       header.appendChild(this._detailsRenderer.render(audit.result.details));
     }
@@ -98,6 +97,11 @@ class CategoryRenderer {
   _renderCategoryScore(category) {
     const tmpl = this._dom.cloneTemplate('#tmpl-lh-category-score', this._templateContext);
     const score = Math.round(category.score);
+
+    const gaugeContainerEl = this._dom.find('.lh-score__gauge', tmpl);
+    const gaugeEl = this.renderScoreGauge(category);
+    gaugeContainerEl.appendChild(gaugeEl);
+
     return this._populateScore(tmpl, score, 'numeric', category.name, category.description);
   }
 
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -36,7 +36,7 @@
   --report-content-width: calc(var(--report-width) + var(--report-menu-width));
 
   --lh-score-highlight-bg: #fafafa;
-  --lh-score-icon-background-size: 17px;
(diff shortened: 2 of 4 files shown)


ebidel:
I chatted with @paulirish about this offline, but we both agreed that some of the new DevTools UI mocks are kinda lack luster. The list in particular is really plain. Not very inspiring to motivate a developer.

If devtools does end up going with a more generic UI, I wonder if we can diverge a bit with a different stylesheet that DT loads? Small things like this should be pretty easy to tweak.

patrickhulce:
Yeah fair enough, it's easy for devtools to drop in styles like these anyway :)

paulirish:
wait what happened here. this PR was pretty awesome. 

let's bring it back?

paulirish:
@ebidel re your comment. I agreed with you that X's on the left are superior. :)
That was mostly it. This PR didnt change that and generally improved things. I think we should apply this, even if we'll do another pass on visuals.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -56,7 +56,6 @@ class CategoryRenderer {
 
     // Append audit details to header section so the entire audit is within a <details>.
     const header = /** @type {!HTMLDetailsElement} */ (this._dom.find('.lh-score__header', tmpl));
-    header.open = audit.score < 100; // expand failed audits
```
paulirish: looks like debugText is printed even if the header isn't `open`?

this header only reveals helpText and any `details` stuff?

lighthouse-core/report/v2/report-styles.css
```diff
@@ -182,53 +181,54 @@ summary {
...
   content: none;
 }
 
 .lh-score__value--pass {
-  background: var(--pass-color);
+  color: var(--pass-color);
 }
 
 .lh-score__value--pass::after {
-  background: url('data:image/svg+xml;utf8,<svg width="12" height="12" viewBox="0 0 12 12" xmlns="http://www.w3.org/2000/svg"><title>pass</title><path d="M9.17 2.33L4.5 7 2.83 5.33 1.5 6.66l3 3 6-6z" fill="white" fill-rule="evenodd"/></svg>') no-repeat 50% 50%;
```
ebidel: Can we keep the svg icons? They look cleaner than the ascii. You can change the size and `fill` color as needed.

ebidel:
@paulirish can we at least remove some of the lines? There are so. many. lines. and a lot of white. Not much to break things up anymore.

What about using a smaller guage for the sore?

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: yup, that was unchanged do we want to stick debugText in details?

lighthouse-core/report/v2/report-styles.css
patrickhulce: done, looks much sharper 👌

patrickhulce:
mini-gauges lookin' ✈️ 
image

ebidel:
It's the little things. This looks muuuuuch better!

ebidel:
@patrickhulce looks like tests need adjusting.

patrickhulce:
adjusted

PR #2099, opened 2017-04-28
Add script to compile renderer against TOT devtools frontend

### LINKED ISSUE #2050: Compile renderer against DevTools frontend
Now with brendan's new compilation in place, we should also be able to check that we compile against the devtools frontend, which has slightly different compilation settings.

Basically i'm thinking we'll use latest devtools frontend, copy our renderer sources in there, and run its closure.

-------------

We have a few prerequisites:
1. paul lands the ReportRenderer CL in devtools
1. will adds a hook to compile_frontend.py to provide a custom path for protocol_externs
1. paul adds a hook to our roll-to-devtools.sh to provide a different `$frontend_dir` location.

-----------

Then I think the flow looks something like this

Add latest devtools frontend to devDeps:
```
 "devtools-frontend-tot": "git://github.com/ChromeDevTools/devtools-frontend.git",
```
Probably add [ChromeDevTools/devtools-protocol) to devDeps as well.

Run roll-to-devtools.sh with `./node_modules/devtools-frontend-tot/front_end`, which copies our renderer files over.

Reach into the frontend and run closure:
```sh
 npm explore devtools-frontend-tot -- npm run closure
```

paulirish:
fixes #2050 

output:

image

### DIFF
--- a/lighthouse-core/scripts/compile-against-devtools.sh
+++ b/lighthouse-core/scripts/compile-against-devtools.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+local_script_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+lhroot_path="$local_script_path/../../"
+
+frontend_path="$lhroot_path/node_modules/temp-devtoolsfrontend"
+protocol_path="$lhroot_path/node_modules/temp-devtoolsprotocol"
+
+# clone if they're not there
+if [ ! -d "$frontend_path" ]; then
+  git clone --depth=1 git://github.com/ChromeDevTools/devtools-frontend.git "$frontend_path"
+fi 
+if [ ! -d "$protocol_path" ]; then
+  git clone --depth=1 git://github.com/ChromeDevTools/devtools-protocol.git "$protocol_path"
+fi
+
+# update to latest
+cd "$frontend_path" && git reset --hard && git fetch origin master && git checkout --force origin/master
+cd "$protocol_path" && git reset --hard && git fetch origin master && git checkout --force origin/master
+
+cd "$lhroot_path" || exit 1
+
+# copy renderer and lh backgrond into this devtools checkout
+yarn devtools -- "$frontend_path/front_end/"
+
+# compile
+python "$frontend_path/scripts/compile_frontend.py" --protocol-externs-file "$protocol_path/externs/protocol_externs.js"
+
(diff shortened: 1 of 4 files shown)


lighthouse-core/scripts/compile-against-devtools.sh
```diff
@@ -0,0 +1,60 @@
...
+if [ ! -d "$protocol_path" ]; then
+  git clone --depth=1 git://github.com/ChromeDevTools/devtools-protocol.git "$protocol_path"
+fi
+
+# update to latest
+cd "$frontend_path" && git reset --hard && git fetch origin master && git checkout --force origin/master
+cd "$protocol_path" && git reset --hard && git fetch origin master && git checkout --force origin/master
+
+
+cd "$lhroot_path" || exit 1
```
patrickhulce: does this fail sometimes?

package.json
```diff
@@ -32,6 +32,7 @@
     "unit": "bash lighthouse-core/scripts/run-mocha.sh --default",
     "closure": "cd lighthouse-core && node closure/closure-type-checking.js",
     "devtools": "bash lighthouse-core/scripts/roll-to-devtools.sh",
+    "compdevtools": "bash lighthouse-core/scripts/compile-against-devtools.sh",
```
patrickhulce: I know you love 🖐 but I didn't know what this meant haha, can we do `compile-devtools`? it's only 4 more characters :)

paulirish: sure sure. 
i found and installed `https://github.com/mklabs/yarn-completions` which does completions for our scripts on `yarn run` so i'm happy now. :)

lighthouse-core/scripts/compile-against-devtools.sh
paulirish: no but https://www.shellcheck.net/ recommends it so i do it. :)

paulirish:
@brendankenny merge when ready

paulirish:
> should we add this to CONTIBUTING?

sure. 

> should we add this to testing?

the plan is to flip it on once we get it passing on master. (last i checked master wasn't yet passing)

> After a perusal it appears this is non-destructive on my Lighthouse repo, so I can run multiple times/run in an in-progress change.

totally. run as much as you like. doesn't affect your LH files.

brendankenny:
> last i checked master wasn't yet passing

yeah, that's what I just ran into.  LGTM!

paulirish:
added a note to readme:
image

PR #2100, opened 2017-04-28
followup for issues in #2000

brendankenny:
also update closure for mid-month release

@paulirish this will hopefully make #2099 a bit smoother


lighthouse-core/closure/closure-type-checking.js
```diff
@@ -32,14 +32,20 @@ gulp.task('compile-report', () => {
...
+
+    'lib/file-namer.js',
     'report/v2/renderer/*.js',
   ])
 
   // Ignore `module.exports` and `self.ClassName = ClassName` statements.
   .pipe(replace(/^\s\smodule\.exports = \w+;$/gm, ';'))
   .pipe(replace(/^\s\sself\.(\w+) = \1;$/gm, ';'))
 
+  // Remove node-specific code from file-namer so it can be included in report.
```
brendankenny: kind of dumb to do such a custom solution for just this file, but since this is the only file that's completely intended to live in both node and the browser (all the other renderer files are only used in node for testing), I went with it. If/when we have more files that need to live in both worlds we can add a more general solution.

lighthouse-core/report/v2/renderer/report-ui-features.js
```diff
@@ -64,7 +64,7 @@ class ReportUIFeatures {
    * Fires a custom DOM event on target.
    * @param {string} name Name of the event.
    * @param {!Node=} target DOM node to fire the event on.
-   * @param {Object<{detail: Object<string, *>}>=} detail Custom data to include.
+   * @param {*=} detail Custom data to include.
```
patrickhulce: 😱

lighthouse-core/closure/closure-type-checking.js
```diff
@@ -32,14 +32,20 @@ gulp.task('compile-report', () => {
...
+  .pipe(replace(/^\s\smodule\.exports = {\w+};$/gm, ';'))
```
paulirish: happens to not match the `module.exports` line in the renderer files:
```js
module.exports = ReportRenderer;
```
but yah this seems to be the right solution.

lighthouse-core/report/v2/renderer/report-ui-features.js
brendankenny: hehe, that's what it really is! https://dom.spec.whatwg.org/#dictdef-customeventinit
```idl
dictionary CustomEventInit : EventInit {
  any detail = null;
};
```

lighthouse-core/closure/closure-type-checking.js
brendankenny: that should be caught by the existing`.pipe(replace(/^\s\smodule\.exports = \w+;$/gm, ';'))` above.

I also turned `module.exports` to write-only, so anything that ends up tricking the regexes will fail the compile :)

brendankenny: > write-only

by which I obviously mean read-only

PR #2062, opened 2017-04-25
Move networkRecords to a computed artifact

paulirish:
This PR moves networkRecords into a computed artifact. It's a little different than other ones though:

* A number of gatherers rely on `tracingData.networkRecords`. As such, we continue to generate these at runtime, so they're available in `afterPass`.  However after the gathering phase is done... we trash this `tracingData.networkRecords` data.  Afterwards when auditing, the networkRecords are recomputed from the `artifacts.devtoolsLogs` which these days is collected and stored.

**The primary motivation** here is we want enable splitting up Gathering from Auditing. (So you could  Gather on one machine and, later, Audit on another.) However, `networkRecords` contains circular references, so currently `artifacts` is not serializable. After this PR is merged, all our artifacts are serializable. \o/

----

This PR also has a **breaking change** regarding our ingesting of `artifacts.performanceLog`. It's now `artifacts.devtoolsLogs` and that must be an object (typically with a first key of `defaultPass`).

---

Reviewers: [The w=1 url ](https://github.com/GoogleChrome/lighthouse/compare/netrecordsComputed?w=1)will help, as there's some decent whitespace changes in here.

There's also more changes we *could* consider in scope for this PR. (Reducing the double `request...` calls we have in a few audits now.) However, based on the # of lines this'll touch, I'd recommend I finish that up in a followup.

PR #2102, opened 2017-04-29
Remove recordNetwork

paulirish:
Now, we always actually record it. (And it's separated from `recordTrace`)

This flag was differentiating on if we expose it to gatherers and artifacts objects.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -129,17 +129,17 @@ function validatePasses(passes, audits, rootPath) {
     });
   });
 
-  // Log if multiple passes require trace or network recording and could overwrite one another.
+  // Log if multiple passes require trace could overwrite one another.
   const usedNames = new Set();
   passes.forEach((pass, index) => {
-    if (!pass.recordNetwork && !pass.recordTrace) {
+    if (!pass.recordTrace) {
       return;
     }
 
     const passName = pass.passName || Audit.DEFAULT_PASS;
     if (usedNames.has(passName)) {
-      log.warn('config', `passes[${index}] may overwrite trace or network ` +
-          `data of earlier pass without a unique passName (repeated name: ${passName}.`);
+      log.warn('config', `passes[${index}] may overwrite trace ` +
+          ` of earlier pass without a unique passName (repeated name: ${passName}.`);
     }
     usedNames.add(passName);
   });
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -129,13 +129,9 @@ function validatePasses(passes, audits, rootPath) {
     });
   });
 
-  // Log if multiple passes require trace could overwrite one another.
+  // Passes should have unique passName's. Warn otherwise.
   const usedNames = new Set();
   passes.forEach((pass, index) => {
-    if (!pass.recordTrace) {
-      return;
-    }
-
     const passName = pass.passName || Audit.DEFAULT_PASS;
     if (usedNames.has(passName)) {
       log.warn('config', `passes[${index}] may overwrite trace ` +
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -129,13 +129,12 @@ function validatePasses(passes, audits, rootPath) {
     });
   });
 
-  // Passes should have unique passName's. Warn otherwise.
+  // Passes must have unique `passName`s. Throw otherwise.
   const usedNames = new Set();
-  passes.forEach((pass, index) => {
+  passes.forEach(pass => {
     const passName = pass.passName || Audit.DEFAULT_PASS;
     if (usedNames.has(passName)) {
-      log.warn('config', `passes[${index}] may overwrite trace ` +
-          ` of earlier pass without a unique passName (repeated name: ${passName}.`);
+      throw new Error(`Passes must have unique names (repeated passName: ${passName}.`);
     }
     usedNames.add(passName);
   });
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -131,8 +131,18 @@ function validatePasses(passes, audits, rootPath) {
 
   // Passes must have unique `passName`s. Throw otherwise.
   const usedNames = new Set();
-  passes.forEach(pass => {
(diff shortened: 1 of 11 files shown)


lighthouse-core/config/config.js
```diff
@@ -129,17 +129,13 @@ function validatePasses(passes, audits, rootPath) {
...
   passes.forEach((pass, index) => {
-    if (!pass.recordNetwork && !pass.recordTrace) {
-      return;
-    }
-
     const passName = pass.passName || Audit.DEFAULT_PASS;
     if (usedNames.has(passName)) {
-      log.warn('config', `passes[${index}] may overwrite trace or network ` +
-          `data of earlier pass without a unique passName (repeated name: ${passName}.`);
+      log.warn('config', `passes[${index}] may overwrite trace ` +
```
brendankenny: `may overwrite data` since it still could overwrite network records?

We should probably throw here instead of warning as there's no "may" about it anymore for network records. Can just be `Passes must have unique names (repeated name: ${passName}.` or something?

lighthouse-core/config/config.js
```diff
@@ -129,17 +129,13 @@ function validatePasses(passes, audits, rootPath) {
     });
   });
 
-  // Log if multiple passes require trace or network recording and could overwrite one another.
+  // Passes should have unique passName's. Warn otherwise.
```
brendankenny: drop apostrophe

lighthouse-core/gather/gather-runner.js
```diff
@@ -249,9 +249,9 @@ class GatherRunner {
       return driver.endNetworkCollect();
     }).then(networkRecords => {
       GatherRunner.assertPageLoaded(options.url, driver, networkRecords);
-
-      // Network records only given to gatherers if requested by config.
-      config.recordNetwork && (passData.networkRecords = networkRecords);
+      // expose devtoolsLog & networkRecords to gatherers
```
brendankenny: capital E

brendankenny: there's also a `// Network is always recorded for internal use, even if not saved as artifact.` comment that needs updating

lighthouse-core/gather/gather-runner.js
```diff
@@ -249,9 +249,9 @@ class GatherRunner {
...
+      passData.devtoolsLog = driver.devtoolsLog;
```
brendankenny: this is already handled above in `config.recordTrace` block

lighthouse-core/gather/gather-runner.js
```diff
@@ -356,10 +356,10 @@ class GatherRunner {
               const passName = config.passName || Audit.DEFAULT_PASS;
               if (config.recordTrace) {
                 tracingData.traces[passName] = passData.trace;
-                tracingData.devtoolsLogs[passName] = passData.devtoolsLog;
               }
-              config.recordNetwork &&
-                  (tracingData.networkRecords[passName] = passData.networkRecords);
+
+              tracingData.devtoolsLogs[passName] = passData.devtoolsLog;
```
brendankenny: this has the same problem as #2062 in that there's no `passData.devtoolLog` if there was no trace. Leave inside the conditional until that's fixed?

lighthouse-core/gather/gather-runner.js
```diff
@@ -356,10 +356,10 @@ class GatherRunner {
...
+              tracingData.networkRecords[passName] = passData.networkRecords;
```
brendankenny: `// If requested by config` comment above needs update

lighthouse-core/config/config.js
brendankenny: I guess that's hard to differentiate in a comment :)
``// Passes should have unique `passName`s. Warn otherwise.``? 
``// Passes should all have a unique passName. Warn otherwise.``?

brendankenny:
test errors are from not all passes having names in smokehouse's [pwa-config.js](https://github.com/GoogleChrome/lighthouse/blob/9ccfd00d7b1d4f4ca264c0a055528fc293754e4f/lighthouse-cli/test/smokehouse/pwa-config.js) so network records are overwriting each other. So, probably 👍  to throw in config.js to catch that early

lighthouse-core/gather/gather-runner.js
paulirish: good catch. devtoolsLog should never be within a `recordTrace` conditional.

lighthouse-core/gather/gather-runner.js
paulirish: fixing that in this PR now. :)

paulirish:
PTAL

lighthouse-core/gather/gather-runner.js
```diff
@@ -85,8 +85,9 @@ class GatherRunner {
     return Promise.resolve()
       // Begin tracing only if requested by config.
       .then(_ => options.config.recordTrace && driver.beginTrace(options.flags))
-      // Network is always recorded for internal use, even if not saved as artifact.
+      // Network is always recorded for gatherer use
       .then(_ => driver.beginNetworkCollect(options))
+      .then(_ => driver.beginDevtoolsLog())
```
patrickhulce: nit: should we do this the other way around? feasibly there could be some network request recorded before the devtools log start I guess, idk

patrickhulce:
love it

paulirish:
@brendankenny this work for you?

brendankenny:
> I'll make a PR on top of this :)

actually, no I won't :) I really need #2062 to do it cleanly.

Speaking of which, what do you think about reverting the last commit? It should really be in #2062 anyways...this PR can get by pretty easily without touching devtoolsLog and #2062 is already touching it, so it makes a lot more sense over there

brendankenny:
(or just merge them together or whatever if this is getting crazy since they're both mostly reviewed)

brendankenny:
OK, removed `devtoolsLog` stuff and added a special warning for multiple unnamed passes (so you wouldn't get a warning about repeated passName 'defaultPass' even though you never named any of your passes 'defaultPass). I didn't remove defaultPass though because that's a lot more work in a lot more places and not worth it right now :)

paulirish:
yup. this WFM.

PR #2104, opened 2017-04-29
artficact renames

paulirish:
A few renames I've been eager to make 

This PR is on top of #2062 btw. 

If approved, let's merge 2062 to master first.

wdyt?


brendankenny:
hmm, I thought the logic was that since the artifacts were arrays of traces and arrays of devtools logs, then `traces` and `devtoolsLogs` made sense. Most (all?) current audits only require *a* trace or *a* devtools log, but until we get per-pass assignability for those, we're still passing in the arrays and they're selecting which one to look at.

I guess maybe explain why they bother you enough to write this PR? :)

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
```diff
@@ -55,8 +55,8 @@ class TotalByteWeight extends ByteEfficiencyAudit {
    * @return {!Promise<!AuditResult>}
    */
   static audit(artifacts) {
-    const devtoolsLogs = artifacts.devtoolsLogs[ByteEfficiencyAudit.DEFAULT_PASS];
-    return artifacts.requestNetworkRecords(devtoolsLogs).then(networkRecords => {
+    const devtoolsLog = artifacts.devtoolsLogs[ByteEfficiencyAudit.DEFAULT_PASS];
```
patrickhulce: change `artifacts.devtoolsLog` too

lighthouse-core/test/gather/fake-driver.js
```diff
@@ -65,11 +65,11 @@ module.exports = {
   beginNetworkCollect() {},
   endNetworkCollect() {
     return Promise.resolve(
-      recordsFromLogs(require('../fixtures/perflog.json'))
+      recordsFromLogs(require('../fixtures/devtoolslog.json'))
```
patrickhulce: should this name change to `recordsFromLog`?

paulirish:
> I guess maybe explain why they bother you enough to write this PR? :)

The primary motivator: "devtoolsLogs" is a mouthful and its part of our external API so i'm eager to smooth it out. Our other artifacts don't have pluralized names, but admittedly that's as they're not pass-specific. 

`networkRecords` has had the same name at the pass level and multi-pass level already, so i don't consider this all that odd.

Slightly less preferable for me, but I'd also be open to `devtoolsLog` -> `protocolLog`. And there I could stomach `protocolLogs` easier. But, still, this is a less favorable solution.

paulirish:
closing because aint nobody got time for renames

PR #2105, opened 2017-04-29
handleJavascriptDialogs, closes #1939

### LINKED ISSUE #1939: Catch and dimiss javascript dialogs/alerts/prompts
lighthouse http://jetcost.com

jetcost seems to call window.confirm during a lighthouse test, and so long as the window.confirm is open, lighthouse continues to wait. tested with 1.6.3

cedricbellet:
Please let me know if something is missing.


cedricbellet:
Hm. Looks like I've got a few changes to make (including changing my email). BRB.

benschwarz:
What if the prompt type isn't just an alert / confirm? (Could be a prompt too)

stramel:
@benschwarz See #2106 Which is the follow up to this

PR #2106, opened 2017-04-29
handleJavascriptDialogs (#1939)

### LINKED ISSUE #1939: Catch and dimiss javascript dialogs/alerts/prompts

cedricbellet:
fixes #1939


lighthouse-core/gather/driver.js
```diff
@@ -833,6 +833,16 @@ class Driver {
     const promiseArr = urlPatterns.map(url => this.sendCommand('Network.addBlockedURL', {url}));
     return Promise.all(promiseArr);
   }
+
+  handleJavaScriptDialogs() {
```
ebidel: You'll need to put the enable in the promise chain:

```javascript
handleJavaScriptDialogs() {
  this.sendCommand('Page.enable').then(_ => {
    this.on('Page.javascriptDialogOpening', _ => {
      this.sendCommand('Page.handleJavaScriptDialog', {
        accept: true,
        promptText: 'Lighthouse prompt response',
      });
    });
  });
}
```

Can you also add a doc string to this method?

ebidel: My first reaction was that maybe we should use `this.once` instead of `this.on`. IOW, dismiss the first prompt we see and ignore the rest. Thinking towards the future when we have auditing after the page load (SPAs), this will dismiss all prompts the page throws up. That might not be something we want to do.

That said, it's a ways out and this `on` is probably safer for now.

ebidel: maybe`dismissJavaScriptDialogs`

cedricbellet: sounds good, will make the changes.

cedricbellet:
I still need to add content to driver-test.js and am not sure why Travis is unhappy about the method. Am I missing something?

brendankenny:
> am not sure why Travis is unhappy about the method. 

yeah, that's a little hard to figure out because the stack trace is so truncated.

You'll need to add a mocked version of `dismissJavaScriptDialogs` to [`fake-driver.js`](https://github.com/GoogleChrome/lighthouse/blob/9ccfd00d7b1d4f4ca264c0a055528fc293754e4f/lighthouse-core/test/gather/fake-driver.js) so that unit tests can call into GatherRunner without having a real browser to talk to. Looks like a `return Promise.resolve();` like most of the others in there will be sufficient since the method has nothing to do since it only makes changes via the debugger protocol.

paulirish:
@googlebot shh bby.

I added the method to the mock driver just to see if that resolves tests completely.

lighthouse-core/gather/driver.js
```diff
@@ -833,6 +833,22 @@ class Driver {
...
+
+  /**
+   * Dismiss JavaScript dialogs (alert, confirm, prompt), providing a
+   * generic promptText in case the dialog is a prompt.
+   * @return {!Promise}
+   */
+  dismissJavaScriptDialogs() {
+    return this.sendCommand('Page.enable').then(_ => {
+      this.once('Page.javascriptDialogOpening', _ => {
+        this.sendCommand('Page.handleJavaScriptDialog', {
```
brendankenny: if this rejects, this is intentionally unhandled (since we need to resolve as soon as `'Page.enable'` is sent). Should we add a note that that's it's intended for future code readers?

paulirish:
@cedricbellet thanks!

PR #2110, opened 2017-05-01
[ChromeLauncher] Catch createConnection error

benschwarz:
### What happens
When there is no Chrome instance ready to accept a connection on the CRI port, node's `net.createConnection` will throw an exception. 

### What was changed
This change wraps the call to `createConnection` in a try/catch.

### Demo of `createConnection` in action 💥 
error

### DIFF
--- a/lighthouse-cli/chrome-launcher.ts
+++ b/lighthouse-cli/chrome-launcher.ts
@@ -178,15 +178,22 @@ export class ChromeLauncher {
   // resolves if ready, rejects otherwise
   isDebuggerReady(): Promise<{}> {
     return new Promise((resolve, reject) => {
-      const client = net.createConnection(this.port);
-      client.once('error', err => {
-        this.cleanup(client);
+      let client
+
+      try {
+        client = net.createConnection(this.port);
+        client.once('error', err => {
+          this.cleanup(client);
+          reject(err);
+        });
+        client.once('connect', () => {
+          this.cleanup(client);
+          resolve();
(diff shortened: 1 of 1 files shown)


patrickhulce:
@benschwarz thanks for investigating! are you seeing this in the wild though? the console example you have is consistent with our existing error handling

```js
const net = require('net')

try {
 const conn = net.createConnection(1234)
 conn.on('error', () => console.log('async error!!'))
} catch (err) {
 console.log('caught!!!!!', err)
}
```

```sh
> node foo.js
async error!!
```

benschwarz:
@patrickhulce I was referring to https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/chrome-launcher.ts#L181, where there is no try/catch for the call to `createConnection`. 

It happens on slower VPS instances (where, I guess chrome takes a little longer to start CRI). At the moment, this is only happening on my test builds.

patrickhulce:
Right I totally don't doubt the port being slow and the error happening, just that the `catch` path will ever be triggered since it seems to be calling the `.on('error', handler)` just fine in that scenario

benschwarz:
You're totally right. I've no idea why I'm seeing this throwing an error 🤔 
Closing until see it again. Sorry for the disturbance @patrickhulce!

patrickhulce:
no worries, node REPL trying to be helpful and printing unhandled errors has caused many superfluous bug investigations myself :)

PR #2115, opened 2017-05-01
Remove renderException

paulirish:
This was mostly introduced because in the extension, we were generating the report in the background page and not the report. So errors were absorbed and harder to track down.

Now that exceptions during rendering happen right on the page itself (CLI, ext, and DT), we can remove this complexity.


ebidel:
Testing need looking at.

PR #2119, opened 2017-05-01
Rename *-unit scripts to unit-*

paulirish:
Should always start with the word that's the same. "unit" "install" build"  followed by "viewer" 
"cli" etc..

Also adding back aliases because it's hard to remember which to use.


patrickhulce:
thoughts on moving to some colon namespacing?

```sh
yarn test # the whole shebang
yarn test:lint # lints all
yarn test:unit # units all
yarn test:unit:core # units core

yarn test:core # runs 'yarn test:lint:core && yarn test:unit:core && ...'
```

paulirish:
we'll kill the aliases in a week or so once muscle memory is adjusted. :)

PR #2120, opened 2017-05-01
handle multiple javascript modal prompts

brendankenny:
followup to #2106. Solution there only closes the first dialog popup, but we want to always close them, including over a page reload per pass (the test site in #1939 must persist your selection somehow because it now works).

Discovered while adding a smoketest that we handle dismissing modal prompts, so PR also includes that :)


brendankenny:
@ebidel suggested one in https://github.com/GoogleChrome/lighthouse/issues/1939#issuecomment-290574898. Even worse than geolocation or notification prompts, so yeah, we really should :)

paulirish:
while you're here can you add a `log.warn` to log that we're dismissing a modal.

brendankenny:
> while you're here can you add a log.warn to log that we're dismissing a modal.

done

PR #2123, opened 2017-05-02
improve instructions for running behind a login

jimthedev:
As a user it isn't immediately clear that debug-chrome needs to be run and remain open when running lighthouse cli to access an area of a site that requires a login.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -120,7 +120,7 @@ NOTE: specifying an output path with multiple formats ignores your specified ext
 
  - `chrome-debug`
  - open and login to your site
- - `lighthouse http://mysite.com`
+ - in a separate terminal tab `lighthouse http://mysite.com`
 
 ## Testing on a mobile device
 


paulirish:
Fair. Good call.

jimthedev:
<3

hrvojegolcic:
So if I understand correctly:
- I have 2 tabs opened in the browser
- In each tab the same page and I'm logged-in in both tabs
- I have devtools opened for each tab
- In one of the tabs I run audit for accessibility

But anways it constantly clears the session and logs me out before the audit is started. Then only audit is done for the login page. What shall I do?

Thanks a lot!

patrickhulce:
@hrvojegolcic if you're not storing auth state in a cookie, then there's nothing that can be done from within DevTools at the moment

These instructions are for the CLI where you can also add the `--disable-storage-reset` flag to avoid clearing out localStorage and other caches.

esoyke:
Does the `--disable-storage-reset` flag work? My app was using local storage to maintain the session and this flag wasn't preventing a reset. I refactored my site to use cookies instead to store the authentication, but Lighthouse is clearing out the cookies as well and login still fails.

PR #2126, opened 2017-05-02
Add time to first byte audit

wardpeet:
Succes:
image

Error:
image


lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+const Audit = require('./audit');
+const Formatter = require('../report/formatter');
+const URL = require('../lib/url-shim');
+
+const TTFB_THRESHOLD = 200;
```
patrickhulce: we should probably use a variable threshold since throttling could easily make every request look like a "failure", maybe we determine a threshold based on the minimum TTFB?

either way we should probably just mark as `informative: true` in the `meta` and give the top X requests sorted by slowness

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * @param {!Artifacts} artifacts
+   * @return {!AuditResult}
+   */
+  static audit(artifacts) {
+    const networkRecords = artifacts.networkRecords[Audit.DEFAULT_PASS];
+    const results = [];
+    const walk = (node) => {
```
patrickhulce: unfortunate that all consumers have to do the walk, I wonder if we should offer a flattened list as a computed artifact too?

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'time-to-firstbyte',
+      description: 'Time To FirstByte',
+      helpText: 'Time To First Byte identifies the time at which your server sends a response.',
+      requiredArtifacts: ['URL', 'networkRecords']
```
patrickhulce: I don't think we actually use the URL artifact do we? Just the `url-shim`?

lighthouse-core/config/default.js
```diff
@@ -615,6 +620,7 @@ module.exports = {
         {"id": "first-meaningful-paint", "weight": 5},
         {"id": "speed-index-metric", "weight": 1},
         {"id": "estimated-input-latency", "weight": 1},
+        {"id": "time-to-firstbyte", "weight": 1},
```
patrickhulce: going forward we're only going to be scoring the actual metrics in performance, so we can set this weight to 0

nit: could you move this down below with the other 'informative' audits so we don't split the metrics

lighthouse-core/test/audits/time-to-firstbyte-test.js
```diff
@@ -0,0 +1,89 @@
...
+      {_url: 'https://google.com/', _requestId: '0', _timing: {receiveHeadersEnd: 500, sendEnd: 200}},
+      {_url: 'https://google.com/styles.css', _requestId: '1', _timing: {receiveHeadersEnd: 414, sendEnd: 200}},
+      {_url: 'https://google.com/image.jpg', _requestId: '2', _timing: {receiveHeadersEnd: 600, sendEnd: 400}},
+    ];
+    const artifacts = {networkRecords: {defaultPass: networkRecords}};
+
+    artifacts.requestCriticalRequestChains = () => {
+      return Promise.resolve(
+        {
+          '0': {
```
patrickhulce: offering a flattened version would make the tests prettier too :)

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+class TTFBMetric extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'time-to-firstbyte',
+      description: 'Time To FirstByte',
+      helpText: 'Time To First Byte identifies the time at which your server sends a response.',
```
ebidel: Maybe a Leare more link to https://developers.google.com/web/tools/chrome-devtools/network-performance/issues

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+
+class TTFBMetric extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'time-to-firstbyte',
+      description: 'Time To FirstByte',
```
ebidel: 'Time To First Byte'

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: Do you mean to go through all requests? Or just sort the critical ones?
🤔 throttling shouldn't mess up the ttfb as it's a server side metric? So what variable are we thinking of? 10%? 

TIL: `informative: true` , didn't know we had these options 😛

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: no sorry forgot to remove it, at first I used this to calculate ttfb of the main document but moved to critical assets later on

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: thanks for this! Was looking for one but didn't find one immediately

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: would be easier though as we only need the tree for the critical request chain

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+      helpText: 'Time To First Byte identifies the time at which your server sends a response.',
+      requiredArtifacts: ['URL', 'networkRecords']
+    };
+  }
+
+  static caclulateTTFB(record) {
+    const timing = record._timing;
+
+    return timing.receiveHeadersEnd - timing.sendEnd;
```
ebidel: Curious...do we want sendEnd - sendStart? Seeing Souders and other [articles](https://developers.googleblog.com/2013/12/measuring-network-performance-with.html) calculate TTFB that way.

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+    const networkRecords = artifacts.networkRecords[Audit.DEFAULT_PASS];
+    const results = [];
+    const walk = (node) => {
+      const children = Object.keys(node);
+
+      children.forEach(id => {
+        const child = node[id];
+
+        const networkRecord = networkRecords.find(record => {
+          return record._requestId === id;
```
ebidel: nit: use one line.

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+        });
+
+        if (networkRecord) {
+          const ttfb = TTFBMetric.caclulateTTFB(networkRecord);
+          results.push({
+            url: URL.getDisplayName(networkRecord._url),
+            ttfb: `${Math.round(ttfb)}ms`,
```
ebidel: `${Math.round(ttfb)} ms`

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+          walk(child.children);
+        }
+      });
+    };
+
+    return artifacts.requestCriticalRequestChains(networkRecords).then(tree => {
+      walk(tree);
+
+      const recordsOverBudget = results.filter(row =>
+        row.rawTTFB > TTFB_THRESHOLD + TTFB_THRESHOLD_BUFFER);
```
ebidel: 4 space indent

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+      let displayValue;
+
+      if (recordsOverBudget.length) {
+        displayValue = recordsOverBudget.length +
+        ` critical request(s) went over the ${TTFB_THRESHOLD}ms threshold`;
```
ebidel: 4 space indent

ebidel: space before "ms": ` ${TTFB_THRESHOLD} ms`

lighthouse-core/audits/time-to-firstbyte.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      return {
+        rawValue: recordsOverBudget.length === 0,
+        displayValue,
+        extendedInfo: {
+          formatter: Formatter.SUPPORTED_FORMATS.TABLE,
+          value: {
+            results,
+            tableHeadings: {
```
ebidel: For the v2 report, you'll want to put this in a `details` table. @paulirish is working on that in https://github.com/GoogleChrome/lighthouse/pull/2019

lighthouse-core/test/audits/time-to-firstbyte-test.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * Copyright 2016 Google Inc. All rights reserved.
```
ebidel: 2017

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: I've copied these headers from devtools. Not sure which to follow

patrickhulce: yeah this is measuring more like response time than TTFB `responseStart - startTime` is true TTFB but that doesn't seem to be the primary thing we want to be auditing, perhaps we should rename the audit?

lighthouse-core/audits/time-to-firstbyte.js
patrickhulce: nit: and maybe throw in `.toLocaleString()` for good measure :)

lighthouse-core/audits/time-to-firstbyte.js
patrickhulce: I was thinking just expose as a `flattened` property of the existing computed artifact so we don't bother with the computation twice although if you're feeling fancy we could refactor the existing one to call `requestCriticalRequests` and then build the tree :)

lighthouse-core/audits/time-to-firstbyte.js
patrickhulce: well kinda `receiveHeadersEnd - sendEnd` will be at least `RTT/2 + server response` time, right? if we're emulating RTT to 400ms for example everything would start to be flagged

lighthouse-core/audits/time-to-firstbyte.js
ebidel: It would be sweet if DT had the same names as resource timing. Alas. Do we have access to `responseStart` or is that the same as `sendEnd`? https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-ResourceTiming

wardpeet: I believe we don't have true TTFB in devtools.

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: seems right indeed, I believe we're not able to calculate real TTFB. I could do an estimate of a roundtrip as you calculate download time

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: sounds good to me

lighthouse-core/audits/time-to-firstbyte.js
wardpeet: oh i'll have a look

wardpeet:
I had a few changes ready to go, was fixing the details export of audit and cleaning up the criticalRequest change

paulirish:
@wardpeet sounds good. let's do that in a new one.

PR #2118, opened 2017-05-01
feat: add styling for informative audits

patrickhulce:
paving the way for #2112, #2044, and splitting up the byte efficiency audits into the sparkline versions and the more aggressive "did you know" versions

basically informative audits will always look like a blue information icon and will not look like pass/fails, the score now simply controls visibility and auto-openness

image

PR #2128, opened 2017-05-03
Travis: assert renderer can compile against devtools

paulirish:
Output:

* https://travis-ci.org/GoogleChrome/lighthouse/jobs/227773873#L606
* https://travis-ci.org/GoogleChrome/lighthouse/jobs/227773872#L607

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ script:
   - yarn closure
   - yarn smoke
   - yarn smokehouse
+  - yarn compile-devtools
 after_success:
   - yarn coveralls
 after_failure:


brendankenny:
Nice!

I'm curious if it would have caught https://github.com/GoogleChrome/lighthouse/pull/2118/commits/0d6991707691072cabcdad8516c79103c02f0fd1 (I'm disappointed that our current closure check didn't alert that it wasn't on the type and I'm not sure why it didn't)

paulirish:
@brendankenny ill try real quick

brendankenny:
I guess the frontend stuff can't be cached on travis by design? Our travis run times have been edging upwards recently

paulirish:
> I guess the frontend stuff can't be cached on travis by design? 

well a git fetch of latest from those repos would take less time than a brand new clone.
so actually a cache would work out pretty okay for those two folders

paulirish:
@brendankenny i checked out f8a45f97b268b49dfec862d7ec2826721738732f which was the commit right before the closure fix

the task **didn't** catch it:
image

brendankenny:
> the task didn't catch it

weird. I'll investigate (though prob post launch)

PR #2132, opened 2017-05-03
Users should strive for < target value. Not hitting it

ebidel:
R: all

This has been confusing people for a while. Finally got around to address it.  Our messaging makes it seem like users should be hitting these numbers, when in fact, we want them to do better than the targets!

screen shot 2017-05-02 at 7 15 38 pm


paulirish:
sg. we'll probably change this target stuff again soon, but this is an improvement nonetheless.

re change, i've been thinking of providing a hint of the scoring curve. maybe tooltip on the score which says "A score of 80: 3220ms. A score of 90: 2610ms"

ebidel:
Maybe the sparklines could should some of those notches?

PR #2135, opened 2017-05-03
Don't filter artifacts out of LHR

### LINKED ISSUE #2124: expose API in lighthouse-background to access artifacts
Currently we do a `filterOutArtifacts` for both extension and devtools to exclude the artifacts from the lighthouseResult object we hand over.

There's uses for DevTools to access artifacts such as trace and devtoolsLog. 
We should make it easy to request those artifacts from bg script.

Strawman API:

```js
window.getArtifacts(lighthouseResult);
```

we cache the artifacts using  `lighthouseResult` as a key. So this works for multiple runs.  

Seems like a decent start.

paulirish:
Fixes #2124

this implements an alternate approach than what was proposed in issue.
now: artifacts are always included on the object and receiver chooses to ditch them or not.

* CLI: ditched in `bin.ts` right after `lighthouse()` completes (already implemented)
* extension: ditched in `runLighthouseInExtension` right after `runLighthouseInExtension` completes
* devtools: will be ditched before results sent into `renderReport`

### DIFF
--- a/lighthouse-extension/app/src/lighthouse-background.js
+++ b/lighthouse-extension/app/src/lighthouse-background.js
@@ -170,7 +170,6 @@ window.runLighthouseInWorker = function(port, url, options, categoryIDs) {
  */
 window.createReportPageAsBlob = function(results, reportContext) {
   performance.mark('report-start');
-  filterOutArtifacts(results);
   const reportGenerator = new ReportGenerator();
   let html;
   try {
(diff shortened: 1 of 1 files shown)


lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -170,7 +170,7 @@ window.runLighthouseInWorker = function(port, url, options, categoryIDs) {
  */
 window.createReportPageAsBlob = function(results, reportContext) {
   performance.mark('report-start');
-
+  filterOutArtifacts(results);
```
brendankenny: did you mean to include this second one? `runLighthouseInExtension` would run this twice if it's here too

paulirish: Nuked

PR #2136, opened 2017-05-03
Closure: handle rejection from stream

paulirish:
we were getting unhandled rejection exceptions in closure-type-checking:
image

This fixes it.


brendankenny:
No unhandled promise rejection, but still no errors like we see locally.

I think we need to remove the `warning_level: process.env.CI ? 'QUIET' : 'VERBOSE',` line altogether, as setting it to `QUIET` locally causes no error exit code. Now why `QUIET` would silence even errors, I don't know...

PR #2137, opened 2017-05-03
make closure test errors fatal; fix errors on master

brendankenny:


lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -233,21 +233,23 @@ class CategoryRenderer {
     element.id = category.id;
     element.appendChild(this._renderCategoryScore(category));
 
-    const auditsGroupedByGroup = category.audits.reduce((indexed, audit) => {
+    const auditsGroupedByGroup = /** @type {!Object<string,
+        {passed: !Array<!ReportRenderer.AuditJSON>,
+        failed: !Array<!ReportRenderer.AuditJSON>}>} */ ({});
```
brendankenny: open to less awkward type cast here, but not sure how. One option is to split `auditsGroupedByGroup` into two `!Object<string, !Array<!ReportRenderer.AuditJSON>>`, one for passes and one for failed, but wanted to minimize change to @patrickhulce's code

brendankenny:
hold off for a second, intentionally reintroducing error to check that it catches it :)

brendankenny:
confirmed it fails the travis run now, ready for review :)

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: eh I'm cool with splitting into passed and failed arrays instead if it makes it cleaner here

brendankenny:
> I still get 46 errors complaining

I think you might need to `yarn`

lighthouse-core/report/v2/renderer/category-renderer.js
paulirish: this typecast lgtm. its quite clear, i'd say.

lighthouse-core/closure/closure-type-checking.js
```diff
@@ -51,7 +51,7 @@ gulp.task('compile-report', () => {
     // new_type_inf: true,
     language_in: 'ECMASCRIPT6_STRICT',
     language_out: 'ECMASCRIPT5_STRICT',
-    warning_level: process.env.CI ? 'QUIET' : 'VERBOSE',
+    warning_level: 'VERBOSE',
```
paulirish: :) good call

PR #2138, opened 2017-05-03
Report css variable tweaks

paulirish:
two things:

1. We want to reuse some of the constants in the CSS variables outside of the report.. for example: 
image

So I've split `lh-root` into `lh-root` and `lh-vars`, where lh-vars can be applied to a subtree anywhere you want those variables.

2. The new category gauge was getting crowded on more narrow widths:

before:
image

after:
image

### DIFF
--- a/lighthouse-core/report/v2/report-template.html
+++ b/lighthouse-core/report/v2/report-template.html
@@ -24,7 +24,7 @@ limitations under the License.
   <title>Lighthouse Report</title>
   <style>/*%%LIGHTHOUSE_CSS%%*/</style>
 </head>
-<body class="lh-root">
+<body class="lh-root lh-vars">
   <noscript>Lighthouse report requires JavaScript. Please enable.</noscript>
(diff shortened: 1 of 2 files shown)


patrickhulce:
> The new category gauge was getting crowded on more narrow widths:

nice catch 👍

lighthouse-core/report/v2/report-template.html
```diff
@@ -24,7 +24,7 @@
   <title>Lighthouse Report</title>
   <style>/*%%LIGHTHOUSE_CSS%%*/</style>
 </head>
-<body class="lh-root">
+<body class="lh-root lh-vars">
```
patrickhulce: do we use root for anything now?

patrickhulce:
> @patrickhulce is merging like crazy!!!

ask and ye shall receive 😆

lighthouse-core/report/v2/report-template.html
paulirish: we do yes.

PR #2140, opened 2017-05-03
feat: performance category style++ (sparklines)

patrickhulce:
Changes:
* Adds group properties to performance audits
* Moves performance category closer to mocks
* Removes failure states from the byte-efficiency audits now roughly grouped as "performance hints"

~~Feedback needed to remove WIP label:~~
* ~~What to do with unoptimized resources audits that have no results? I actually like not showing them at all, but can see the confusion of where'd it go, maybe an overall passed section at the bottom?~~ 
Grouped in a passed section at bottom
* ~~We cool with the sparkline renderer living in category renderer? (I vote yes for now, but long-term likely splitting out each category to a new class since there's not much else one-off we plan to add for I/O outside performance)~~ 
This is basically regular review feedback at this point, but I'm going to hold off writing tests against it until I get a cursory review just in case :)

Screenshot:
image

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -20,13 +20,31 @@ const Audit = require('../audit');
 const Formatter = require('../../report/formatter');
 
 const KB_IN_BYTES = 1024;
-const WASTEFUL_THRESHOLD_IN_BYTES = 20 * KB_IN_BYTES;
+
+const WASTED_MS_FOR_AVERAGE = 150;
+const WASTED_MS_FOR_POOR = 750;
 
 /**
  * @overview Used as the base for all byte efficiency audits. Computes total bytes
  *    and estimated time saved. Subclass and override `audit_` to return results.
  */
 class UnusedBytes extends Audit {
+  /**
+   * @param {number} wastedMs
+   * @return {number}
+   */
+  static scoreForWastedMs(wastedMs) {
+    if (wastedMs === 0) {
+      return 100;
+    } else if (wastedMs < WASTED_MS_FOR_AVERAGE) {
+      return 90;
+    } else if (wastedMs < WASTED_MS_FOR_POOR) {
+      return 65;
+    } else {
+      return 0;
+    }
+  }
+
   /**
    * @param {number} bytes
    * @return {string}
@@ -88,6 +106,8 @@ class UnusedBytes extends Audit {
         .sort((itemA, itemB) => itemB.wastedBytes - itemA.wastedBytes);
 
     const wastedBytes = results.reduce((sum, item) => sum + item.wastedBytes, 0);
+    const wastedKb = Math.round(wastedBytes / KB_IN_BYTES);
+    const wastedMs = Math.round(wastedBytes / networkThroughput * 100) * 10;
 
     let displayValue = result.displayValue || '';
     if (typeof result.displayValue === 'undefined' && wastedBytes) {
@@ -99,12 +119,16 @@ class UnusedBytes extends Audit {
     return {
       debugString,
       displayValue,
-      rawValue: typeof result.passes === 'undefined' ?
-          wastedBytes < WASTEFUL_THRESHOLD_IN_BYTES :
-          !!result.passes,
+      rawValue: wastedMs,
+      score: UnusedBytes.scoreForWastedMs(wastedMs),
       extendedInfo: {
         formatter: Formatter.SUPPORTED_FORMATS.TABLE,
-        value: {results, tableHeadings: result.tableHeadings}
+        value: {
+          wastedMs,
+          wastedKb,
+          results,
+          tableHeadings: result.tableHeadings,
+        },
       }
     };
   }
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -34,15 +34,14 @@ class UnusedBytes extends Audit {
    * @return {number}
    */
   static scoreForWastedMs(wastedMs) {
-    if (wastedMs === 0) {
+    if (wastedMs === 0)
       return 100;
-    } else if (wastedMs < WASTED_MS_FOR_AVERAGE) {
+    else if (wastedMs < WASTED_MS_FOR_AVERAGE)
       return 90;
-    } else if (wastedMs < WASTED_MS_FOR_POOR) {
+    else if (wastedMs < WASTED_MS_FOR_POOR)
       return 65;
-    } else {
+    else
       return 0;
-    }
   }
 
   /**
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -21,7 +21,7 @@ const Formatter = require('../../report/formatter');
 
 const KB_IN_BYTES = 1024;
 
-const WASTED_MS_FOR_AVERAGE = 150;
+const WASTED_MS_FOR_AVERAGE = 300;
 const WASTED_MS_FOR_POOR = 750;
 
 /**
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -34,14 +34,10 @@ class UnusedBytes extends Audit {
    * @return {number}
    */
   static scoreForWastedMs(wastedMs) {
-    if (wastedMs === 0)
-      return 100;
-    else if (wastedMs < WASTED_MS_FOR_AVERAGE)
-      return 90;
-    else if (wastedMs < WASTED_MS_FOR_POOR)
-      return 65;
-    else
-      return 0;
+    if (wastedMs === 0) return 100;
+    else if (wastedMs < WASTED_MS_FOR_AVERAGE) return 90;
+    else if (wastedMs < WASTED_MS_FOR_POOR) return 65;
+    else return 0;
   }
 
   /**
--- a/lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
@@ -28,9 +28,6 @@ const ByteEfficiencyAudit = require('./byte-efficiency-audit');
 const URL = require('../../lib/url-shim');
 
 const IGNORE_THRESHOLD_IN_BYTES = 2048;
-const TOTAL_WASTED_BYTES_THRESHOLD = 1000 * 1024;
-const JPEG_ALREADY_OPTIMIZED_THRESHOLD_IN_BYTES = 25 * 1024;
-const WEBP_ALREADY_OPTIMIZED_THRESHOLD_IN_BYTES = 100 * 1024;
 
 class UsesOptimizedImages extends ByteEfficiencyAudit {
   /**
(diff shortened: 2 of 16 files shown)


lighthouse-core/config/default.js
```diff
@@ -588,6 +588,18 @@ module.exports = {
...
+    "perf-metric": {
+      "title": "Metrics",
+      "description": "These metrics encapsulate your app's performance across a number of dimensions."
+    },
+    "perf-hint": {
+      "title": "Unoptimized Resources",
+      "description": "These are opportunities to speed up your application by optimizing the following resources."
+    },
+    "perf-info": {
+      "title": "Diagnostics",
```
patrickhulce: could use a better name "Other" wasn't very catchy

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -116,6 +116,94 @@ class CategoryRenderer {
...
+    }
+    return element;
+  }
+
+  /**
+   * @param {!Array<!ReportRenderer.AuditJSON>} audits
+   * @param {!ReportRenderer.GroupJSON} group
+   * @return {!Element}
+   */
+  _renderAuditGroup(audits, group, renderAudit) {
```
patrickhulce: just moved, not new

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -20,14 +20,32 @@ const Audit = require('../audit');
...
 /**
  * @overview Used as the base for all byte efficiency audits. Computes total bytes
  *    and estimated time saved. Subclass and override `audit_` to return results.
  */
 class UnusedBytes extends Audit {
   /**
+   * @param {number} wastedMs
+   * @return {number}
+   */
+  static scoreForWastedMs(wastedMs) {
```
patrickhulce: felt cumbersome to use the standard curve here since the points at which we'll style as failure and average are what we really care about, not the median/diminishing points

ebidel:
> What to do with unoptimized resources audits that have no results? I actually like not showing them at all, but can see the confusion of where'd it go, maybe an overall passed section at the bottom?

For ones that have no results (the audit is fully passing), could we show it in the passing section. Instead of a sparkline, show a green check or something?


Look at this categorization, it feels like our naming is off. Especially now that we're showing passing and fails audits in different sections. Example: it's confusing to see "Unoptimized images" in a passing section. One thought...

What if we change titles to action we want the users to taking: So this:

"Unoptimized Resources"
    "Unoptimized images"
    "Oversized images"
    "Render-blocking scripts"
    "Compression enabled for server responses"

becomes a task list:

"Optimize resources" or simply "Resources" (single word like the others)
    "Optimize images"
    "Properly size images"
    "Reduce render-blocking scripts"
    "Enable compression for server responses"

^ those titles also align well with PSI 👍

patrickhulce:
> For ones that have no results (the audit is fully passing), could we show it in the passing section. Instead of a sparkline, show a green check or something?

It felt really weird having a passing section in the middle, thoughts on having a single passing section at the bottom and only have information appear in "Resources" and "Diagnostics" if they're <100?

> becomes a task list:
"Optimize resources" or simply "Resources" (single word like the others)
"Optimize images"
"Properly size images"
"Reduce render-blocking scripts"
"Enable compression for server responses"

I like it 👍

ebidel:
> thoughts on having a single passing section at the bottom

That's what I was imagining too :) Sorry, haven't looked closely at this PR yet.

patrickhulce:
rebased and updated, removing WIP label so come at me!

lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
```diff
@@ -40,7 +40,7 @@ class UsesOptimizedImages extends ByteEfficiencyAudit {
     return {
       category: 'Images',
       name: 'uses-optimized-images',
-      description: 'Unoptimized images',
+      description: 'Optimize images',
```
paulirish: i had mocked this section out with titles more in line with "Your problems" rather than the "Suggested solutions" you do here.

but since the big communication here is YOUR SAVINGS, it seems good to sit those numbers next to the action you must take.

sgtm

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
paulirish: sg. wanna kill the braces ;)?

lighthouse-core/config/default.js
```diff
@@ -588,6 +588,18 @@ module.exports = {
     }]
   }],
   "groups": {
+    "perf-metric": {
+      "title": "Metrics",
+      "description": "These metrics encapsulate your app's performance across a number of dimensions."
+    },
+    "perf-hint": {
+      "title": "Resources",
```
paulirish: s/Resources/Opportunities/  ??

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -116,6 +116,109 @@ class CategoryRenderer {
...
+
+    const titleEl = this._dom.createElement('div', 'lh-perf-hint__title');
+    titleEl.textContent = audit.result.description;
+    summary.appendChild(titleEl);
+
+    const sparklineContainerEl = this._dom.createElement('div', 'lh-perf-hint__sparkline');
+    const sparklineEl = this._dom.createElement('div', 'lh-sparkline');
+    const sparklineBarEl = this._dom.createElement('div', 'lh-sparkline__bar');
+    sparklineBarEl.style.width = extendedInfo.value.wastedMs / scale * 100 + '%';
+    sparklineEl.appendChild(sparklineBarEl);
```
paulirish: fwiw you do have `this._dom.createChildOf` if you want to nuke these append lines.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -186,45 +291,58 @@ class CategoryRenderer {
...
+    const hintAudits = category.audits
+        .filter(audit => audit.group === 'perf-hint' && audit.score < 100)
+        .sort((auditA, auditB) => auditB.result.rawValue - auditA.result.rawValue);
+    if (hintAudits.length) {
+      let maxWaste = 0;
+      hintAudits.forEach(audit => maxWaste = Math.max(maxWaste, audit.result.rawValue));
+
+      const scale = Math.ceil(maxWaste / 1000) * 1000;
+      const hintAuditsEl = this._renderAuditGroup(hintAudits, groups['perf-hint'],
+          audit => this._renderPerfHintAudit(audit, scale));
```
paulirish: should this go through our typical `render()` method?  if so, we'd need to smush `scale` into the audit object and set up the `type` etc..

that seems pretty reasonable, but lemme know if you're thinkin otherwise.
i'd say if we don't then it probably shouldn't be prefixed with `_render` as its done differently than the other `_render**()`'s

paulirish: nvm. categoryRenderer doesn't do it like this. :)

lighthouse-core/config/default.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: 👍  done

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
patrickhulce: yup

paulirish: oh sorry i meant:

```js
      if (wastedMs === 0) return 100;
      else if (wastedMs < WASTED_MS_FOR_AVERAGE) return 90;
      else if (wastedMs < WASTED_MS_FOR_POOR) return 65;
      else return 0;
```

this is a super nit either way. w/e

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -116,6 +116,99 @@ class CategoryRenderer {
...
+   * @param {!ReportRenderer.GroupJSON} group
+   * @param {function(!ReportRenderer.AuditJSON): !Element=} renderAudit
+   * @return {!Element}
+   */
+  _renderAuditGroup(audits, group, renderAudit) {
+    if (!renderAudit) {
+      renderAudit = this._renderAudit.bind(this);
+    }
+
+    const auditGroupElem = this._dom.createElement('details',
```
paulirish: i kinda think we dont need another `details` nesting here.
how about just a header for the title and the items following.. not collapsible.

patrickhulce: this is re-used for general audit groups (in accessibility for example), worth forking the logic to explicitly not nest in a details?

paulirish: oh ok. let's leave it alone. 

turns out details element fires a `toggle` event when it gets touched. but you can't preventDefault it. anyway, sure w're good.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -116,6 +116,99 @@ class CategoryRenderer {
...
+    return element;
+  }
+
+  /**
+   * @param {!Array<!ReportRenderer.AuditJSON>} audits
+   * @param {!ReportRenderer.GroupJSON} group
+   * @param {function(!ReportRenderer.AuditJSON): !Element=} renderAudit
+   * @return {!Element}
+   */
+  _renderAuditGroup(audits, group, renderAudit) {
```
brendankenny: this is kind of an oddball, injecting a renderAudit function when everything so far doesn't do this. Rather than invert control, could this function just return the `auditGroupElem` that the caller will `appendChild` its audits to?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -116,6 +116,99 @@ class CategoryRenderer {
   }
 
   /**
+   * @param {!ReportRenderer.AuditJSON} audit
+   * @param {number} scale
+   * @return {!Element}
+   */
+  _renderPerfHintAudit(audit, scale) {
+    const extendedInfo = /** @type {!CategoryRenderer.PerfHintExtendedInfo}
```
brendankenny: could this be in `details`? It would be nice to not split report rendering info across `details` and `extendedInfo` if we don't have to

brendankenny:
sorry for the slow comments :)

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: discussed a lot with paul on that topic, basically by porting over so much of the way things were before and deciding against a generic solution for rendering categories/groups for now (and basically harding switch statements instead) it doesn't really fit into details cleanly since details is exclusively for expanded audit information not the audit itself, it'd be nice to retackle this when we migrate audit groups/audits themselves to details

brendankenny: yeah, my concern is breaking the seal on `extendedInfo` :/ These things tend to grow with time (see: `extendedInfo` itself). It also breaks the `details`->rendering info, `extendedInfo` -> debug/for other tools info split. Really it would be better even if it was exactly the same but called basically anything else :)

patrickhulce: ha, yeah +1 to keeping an object that is purely informational with no rendering awareness, which extendedInfo should be once we rip out formatters and whatnot. I see a difference between "renderer-aware" and "used by renderer" though. The former is bad and the latter seems fine in the short-term.

PR #2145, opened 2017-05-04
Add line/code renderer to deprecations audit

ebidel:
screen shot 2017-05-03 at 9 16 57 pm

### DIFF
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -24,6 +24,7 @@
 
 const Audit = require('./audit');
 const Formatter = require('../report/formatter');
+const URL = require('../lib/url-shim');
 
 class Deprecations extends Audit {
   /**
@@ -65,12 +66,13 @@ class Deprecations extends Audit {
     function makeDepractionsV2(entries) {
       return entries.filter(log => log.entry.source === 'deprecation').map(log => {
         // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
-        const label = log.entry.lineNumber ? `line: ${log.entry.lineNumber}` : 'line: ???';
-        const url = log.entry.url || 'Unable to determine URL';
+        const url = URL.isValid(log.entry.url) ? URL.getDisplayName(log.entry.url) : '';
         return {
           type: 'code',
           text: log.entry.text,
-          header: {type: 'text', text: `${url} (${label})`}
+          url,
+          source: log.entry.source,
+          lineNumber: log.entry.lineNumber
         };
       });
     }
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -61,9 +61,10 @@ class Deprecations extends Audit {
         });
 
     /**
+     * @param {!Array<!DetailsRenderer.CodeDetailsJSON>} entries
      * @return {!DetailsRenderer.CodeDetailsJSON} details
      */
-    function makeDepractionsV2(entries) {
+    function makeDeprecationsV2(entries) {
       return entries.filter(log => log.entry.source === 'deprecation').map(log => {
         // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
         const url = URL.isValid(log.entry.url) ? URL.getDisplayName(log.entry.url) : '';
@@ -94,7 +95,7 @@ class Deprecations extends Audit {
       details: {
         type: 'list',
         header: {type: 'text', text: 'Deprecations & warnings'},
-        items: makeDepractionsV2(entries)
+        items: makeDeprecationsV2(entries)
       }
     };
   }
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -59,6 +59,19 @@ class Deprecations extends Audit {
           }, log.entry);
         });
 
+    function makeDepractionsV2(entries) {
+      return entries.filter(log => log.entry.source === 'deprecation').map(log => {
+        // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
+        const label = log.entry.lineNumber ? `line: ${log.entry.lineNumber}` : 'line: ???';
+        const url = log.entry.url || 'Unable to determine URL';
+        return {
+          type: 'code',
+          text: log.entry.text,
+          header: {type: 'text', text: `${url} (${label})`}
+        };
+      });
+    }
+
     let displayValue = '';
     if (deprecations.length > 1) {
       displayValue = `${deprecations.length} warnings found`;
@@ -72,6 +85,11 @@ class Deprecations extends Audit {
       extendedInfo: {
         formatter: Formatter.SUPPORTED_FORMATS.URL_LIST,
         value: deprecations
+      },
+      details: {
+        type: 'list',
+        header: {type: 'text', text: 'Deprecations & warnings'},
+        items: makeDepractionsV2(entries)
       }
     };
   }
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -59,6 +59,9 @@ class Deprecations extends Audit {
           }, log.entry);
         });
 
+    /**
+     * @return {!DetailsRenderer.CodeDetailsJSON} details
+     */
     function makeDepractionsV2(entries) {
       return entries.filter(log => log.entry.source === 'deprecation').map(log => {
         // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -60,23 +60,24 @@ class Deprecations extends Audit {
           }, log.entry);
         });
 
-    /**
-     * @param {!Array<!DetailsRenderer.CodeDetailsJSON>} entries
-     * @return {!DetailsRenderer.CodeDetailsJSON} details
-     */
-    function makeDeprecationsV2(entries) {
-      return entries.filter(log => log.entry.source === 'deprecation').map(log => {
-        // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
-        const url = URL.isValid(log.entry.url) ? URL.getDisplayName(log.entry.url) : '';
-        return {
-          type: 'code',
-          text: log.entry.text,
-          url,
-          source: log.entry.source,
-          lineNumber: log.entry.lineNumber
-        };
-      });
-    }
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -134,3 +155,12 @@ DetailsRenderer.ListDetailsJSON; // eslint-disable-line no-unused-expressions
  * }}
  */
 DetailsRenderer.CardsDetailsJSON; // eslint-disable-line no-unused-expressions
+
+/**
+ * @typedef {{
+ *     type: string,
+ *     text: string,
+ *     header: (!DetailsRenderer.DetailsJSON|undefined),
```
brendankenny: to avoid circular typedefs, can do `header: {type: string, text: (string|undefined)}` like `ListDetailsJSON` does with its `items`

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -102,6 +104,25 @@ class DetailsRenderer {
...
+   * @param {!DetailsRenderer.CodeDetailsJSON} details
+   * @return {!Element}
+   */
+  _renderCode(details) {
+    const element = this._dom.createElement('div', 'lh-details');
+
+    const pre = this._dom.createElement('pre', 'lh-code');
+    pre.textContent = details.text;
+
+    if (details.header) {
```
brendankenny: `header` seems like it's overloading the term since it's per-item here, maybe there's something better to call it (based on the screenshot, at least?). I don't have any good suggestions, though

lighthouse-core/audits/deprecations.js
```diff
@@ -59,6 +59,19 @@ class Deprecations extends Audit {
           }, log.entry);
         });
 
+    function makeDepractionsV2(entries) {
```
brendankenny: jsdoc would be nice here just to not require looking through the impl to see the return type

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: I was using `label` but decided to keep it consistent with the rest of the details renderers. And it's more or less a "header" for the code snippet.

lighthouse-core/audits/deprecations.js
ebidel: done. note this does go away when we ditch extendedInfo

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done

ebidel:
PTAL

paulirish:
do we want to group by the entry text? in case there are a bunch of /deep/ errors (or whatever) from various locations?

Also how about visualizing similar to how they're shown in the console, where the location is on the right side:

image

paulirish:
Combing the two might look something like this:

image

ebidel:
My understanding is that the deprecation system only warns about first use, so we won't see a bunch of `/deep/`. Not sure about violations. What pages are you testing against in those screenshots?

ebidel:
Looks like the console now. I also added a warning icon for audit errors. 

screen shot 2017-05-04 at 11 09 23 am

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -47,6 +49,16 @@ class DetailsRenderer {
    * @param {!DetailsRenderer.DetailsJSON} text
    * @return {!Element}
    */
+  _renderURL(text) {
```
ebidel: stole this from https://github.com/GoogleChrome/lighthouse/pull/2019

patrickhulce:
nice! can we have the warning icon not mess with where the text starts and shift it to the left? maybe just beneath the X basically?

lighthouse-core/audits/deprecations.js
```diff
@@ -59,6 +60,23 @@ class Deprecations extends Audit {
           }, log.entry);
         });
 
+    /**
+     * @return {!DetailsRenderer.CodeDetailsJSON} details
```
brendankenny: `{!Array<!DetailsRenderer.CodeDetailsJSON>}`?

also needs `entries`

lighthouse-core/audits/deprecations.js
```diff
@@ -59,6 +60,23 @@ class Deprecations extends Audit {
...
+     */
+    function makeDepractionsV2(entries) {
```
brendankenny: any reason this declaration is nested in `audit()`? Move it out?

ebidel:
@patrickhulce the debug text was aligned with "Open" and this just drops an icon in front of that text. We could align it with the audit X/check but that might be confusing?

I'm ok with ditching it too. Maybe the red it enough .

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -36,6 +36,8 @@ class DetailsRenderer {
         return this._renderText(details);
       case 'cards':
         return this._renderCards(/** @type {!DetailsRenderer.CardsDetailsJSON} */ (details));
+      case 'code':
```
brendankenny: how do you feel about calling this something like `codeSnippet`? It feels like `code` will be a more primitive renderer more similar to `url` or `text`...just a `<pre>` and maybe (someday) some syntax highlighting or something

patrickhulce:
yeah just feels a bit weird, maybe I only actually want the text on the second line to be aligned with the first line instead :)

paulirish:
>  So we won't see a bunch of /deep/. Not sure about violations. What pages are you testing against in those screenshots?

I was hand-editing those screenshots but it started from the console of https://chromedevtools.github.io/devtools-protocol/

paulirish:
new look generally lgtm. plenty to bikeshed but nothing big.

ebidel:
Here's left aligned. Not sure that it's better.

screen shot 2017-05-04 at 1 36 03 pm

ebidel:
Oh snap. Didn't realize the viewer was Polymer.

patrickhulce:
yeah no the original indent feels a bit better, if we just have the icon stand on its own with some margin before the text sgtm

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -134,3 +164,14 @@ DetailsRenderer.ListDetailsJSON; // eslint-disable-line no-unused-expressions
  * }}
  */
 DetailsRenderer.CardsDetailsJSON; // eslint-disable-line no-unused-expressions
+
+/**
+ * @typedef {{
+ *     type: string,
+ *     text: string,
```
brendankenny: what is `text` in this case? like a description of the code snippet that follows?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -134,3 +164,14 @@ DetailsRenderer.ListDetailsJSON; // eslint-disable-line no-unused-expressions
...
+ *     url: (string|undefined),
+ *     source: string,
```
brendankenny: what about `code` instead of `source` (to differentiate `source` meaning like the URL or wherever the code is from)

ebidel:
Also: 

screen shot 2017-05-04 at 1 41 37 pm

ebidel:
Non red also looks decent:

screen shot 2017-05-04 at 1 47 05 pm

ebidel:
LOL. probably should land and tweak later. This stuff is easy to change.

patrickhulce:
ooooh I like this one :) haha yeah it's all good
image

lighthouse-core/audits/deprecations.js
ebidel: done

lighthouse-core/audits/deprecations.js
ebidel: Didn't think it was worth elevating to the class. This func is going away when we ditch extended info.

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: We should land the table renderer so I can use it in this PR :) Then we can make a generic `code` renderer.

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: It's the text for the code snippet. For this audit, they're the console message text. We're just that text in `<pre>` so it looks like DT's log output.

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: `url` is where the code came from :) I stuck to `source` b/c it's what the protocol uses: `logEntry.source`.

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: Lolz. This happened while I was responding. BRB!

ebidel:
This is using the table renderer now. Simplifies things a big and `renderCode` became a standalone renderer for `<pre>`.

The table render needs some UI love, but we can do that later when the rest of the audits are in that need it. 

screen shot 2017-05-04 at 3 17 09 pm

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -53,19 +55,19 @@ class DetailsRenderer {
    * @param {!DetailsRenderer.DetailsJSON} text
    * @return {!Element}
    */
-  _renderText(text) {
-    const element = this._dom.createElement('div', 'lh-text');
-    element.textContent = text.text;
+  _renderURL(text) {
```
patrickhulce: ha, any reason to flip these? just adds noise

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -217,3 +229,14 @@ DetailsRenderer.TableDetailsJSON; // eslint-disable-line no-unused-expressions
...
+/**
+ * @typedef {{
+ *     type: string,
+ *     text: string,
+ *     url: (string|undefined),
+ *     source: string,
+ *     lineNumber: (string|undefined)
+ * }}
+ */
+DetailsRenderer.CodeDetailsJSON; // eslint-disable-line no-unused-expressions
```
brendankenny: delete now? `renderCode` can just take a `DetailsRenderer.DetailsJSON` now, I believe

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: lost track of which was which :)

lighthouse-core/report/v2/renderer/details-renderer.js
ebidel: done

paulirish:
let's do it.

PR #2150, opened 2017-05-04
Allow opn & update-notifier CLI deps to be optional.

paulirish:


lighthouse-cli/shim-modules.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2016 Google Inc. All rights reserved.
```
ebidel: 2017

lighthouse-cli/shim-modules.ts
```diff
@@ -0,0 +1,50 @@
...
+/** @fileoverview Some users may be unable to install the full dependency tree,
+ * especially for the CLI. `opn` and `update-notifier` in particular have some
+ * uncommon transitive dependencies, so these shims will let the modules no-op
+ * if the real dependency is not installed.
+ */
+
+export let opn = () => {
+	function shim(str: String, obj: Object) {
+		console.error('module `opn` not installed. Not opening browser.');
+		return [str, obj]
```
ebidel: ;

lighthouse-cli/shim-modules.ts
```diff
@@ -0,0 +1,50 @@
...
+	} catch (e) {
+		return shim;
+ }
+};
+
+
+export let updateNotifier = () => {
+	function shim(obj: Object) {
+		console.error('module `update-notifier` not installed. Not checking for new version.');
+		return { notify: () => obj };
```
ebidel: `{notify: () => obj}`

lighthouse-cli/bin.ts
```diff
@@ -35,10 +35,12 @@ import * as Printer from './printer';
 import * as randomPort from './random-port';
 import {Results} from './types/types';
 const yargs = require('yargs');
-const opn = require('opn');
-const updateNotifier = require('update-notifier');
 const pkg = require('../package.json');
 
+// accept noop modules for these, so the real dependency is optional.
+const opn = require('./shim-modules').opn();
```
brendankenny: should we do this or just try/catch their assignment and guard their usage with an  `if (opn)` and `if (updateNotifier)`? They're used so rarely it seems easy to do the second option

ebidel: I was thinking the same, but we may have more of these in the future. It might be nice to keep our codebase clean rather than littering if with a bunch of forks.

brendankenny: it's just that for these two it's literally just
```js
if (updateNotifier) {
  updateNotifier({pkg}).notify();
}
```
and
```ts
if (flags.view && opn) {
  opn(outputPath, {wait: false});
}
```
the new thing has logged warnings, which is nice, but it is a lot more code to get that same effect :)

paulirish: the try/catch in the same file is really nasty and i really preferred moving it out into shim-modules.

so i think it's worth keeping that.

However, i'd be fine with guarding the calls so we don't need to mock their API.

patrickhulce: oooooh not literal `optionalDependencies` just handle them optionally :) I gotcha now

is there any benefit to having these be individual functions vs just inlining the exports? feels a little weird as functions

paulirish: done

paulirish: >  i'd be fine with guarding the calls so we don't need to mock their API.

done

paulirish: >>  i'd be fine with guarding the calls so we don't need to mock their API.

>done

undone. i think its worth the overhead to make sure silent failures aren't silent.

lighthouse-cli/bin.ts
```diff
@@ -35,10 +35,12 @@ import * as Printer from './printer';
 import * as randomPort from './random-port';
 import {Results} from './types/types';
 const yargs = require('yargs');
-const opn = require('opn');
-const updateNotifier = require('update-notifier');
 const pkg = require('../package.json');
 
+// accept noop modules for these, so the real dependency is optional.
+const opn = require('./shim-modules').opn;
```
patrickhulce: nit: really up our es imports cred and do `import {opn, updateNotifier} from './shim-modules'`?

lighthouse-cli/shim-modules.ts
```diff
@@ -0,0 +1,36 @@
...
+try {
+	updateNotifier = require('update-notifier');
+} catch (e) {
+	updateNotifier = function shimUpdateNotifier() {
+		console.error('module `update-notifier` not installed. Not checking for new version.');
+		return {notify: () => false};
+	};
+}
+
+export {
```
patrickhulce: 👌  beautiful

Issue #2146, opened 2017-05-04
Operation Yaquina Bay


paulirish:
Named after the Yaquina Bay lighthouse which, _like a fast webpage that you can see it load quickly from start to finish_, is short enough that it doesn't take long to see the whole thing:
![image](https://cloud.githubusercontent.com/assets/39191/25690902/b8770a26-304a-11e7-9c4e-9a5c6fb20ec4.png)

----

OYB is about making Lighthouse itself fast. I mean.. a performance tool should probably be performant. 😉 

### Goals: 

1. Make the default 4 category run of lighthouse faster
2. Improve the run duration for users who want to run a single category

The full start-to-finish LH run of cnn.com takes 3 minutes. Oh my. Waiting for load of the first pass takes 25 seconds, so that's the absolute fastest we can make the entire LH run take right now.  

While cutting the time to 1/6th is a stretch **we should at least cut it in half**.

-------

While profiling LH runtimes, we uncovered a few issues:

1. The disabled cache for the entire run (#2089) slows down passes 3 and 4. 
1. The Styles gatherer is generally taking >10s on heavy publisher sites
1. Speedline is also going >10s on long pageloads 
1. Retrieving trace can take like 10s easily.
1. TTI(beta) / LoadFastEnough audits are costly (~5s) on larger traces.

I think we can do a few selective fixes here to drive our total run duration down pretty easily.

patrickhulce:
Update on "TTI(beta) / LoadFastEnough audits are costly (~5s) on larger traces":

tl;dr remove old TTI (alpha) code before I/O will get us a few seconds, nothing can be done about TTFI in the short term since it's only ~20ms of computation after getting a trace model

For firstInteractive, getting the trace model accounts for roughly 99.5% of the time spent (3110ms spent in getting trace model, 17ms spent filtering to long tasks and computing the result) so if we do away with that and compute durations slices ourselves then we can likely get a pretty big win. However, we'd really just shift the cost to EIL since it also uses trace model. This seems an unlikely dependency to remove before I/O and likely to introduce bugs if we maintain ourselves, but I might investigate further if time permits.

For old TTI, getting the trace model/running speedline only accounts for ~66% of the time spent (4738ms to get trace model/speedline since trace model should be cached by then, 2422ms computing) so there's likely more room for improvement but killing that entirely should save us that time anyway

patrickhulce:
> Retrieving trace can take like 10s easily.

[...]

PR #2151, opened 2017-05-04
perf(driver): only disable cache on trace passes

### LINKED ISSUE #2089: Disk cache is disabled for the whole run
We disable network cache at the beginning of the run (`setCacheDisabled` in driver) and never turn it back on.

Subsequent passes don't neccessarily need cache to be disabled, so we're just slowing those down.

1. why do we disable if we just cleared the whole thing? (i am not sure) image
1. do we ever need it disabled? Probably for the service-worker offline, checks, but that's probably it.

patrickhulce:
fixes #2089 (though it doesn't really seem to have any tangible impact on #2146 😢  maybe this isn't actually working? from what I can tell, master seemed to be using the cache on subsequent passes already)

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -750,18 +750,12 @@ class Driver {
         .then(_ => this.online = true);
   }
 
-  cleanAndDisableBrowserCaches() {
-    return Promise.all([
-      this.clearBrowserCache(),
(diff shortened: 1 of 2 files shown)


paulirish:
we'll pick this up in the https://github.com/GoogleChrome/lighthouse/compare/dontdisablecache branch


😀

PR #2153, opened 2017-05-04
perf(config): disable styles gatherer

patrickhulce:
in the name of Operation Yaquina Bay (#2146)

shaves off ~40s on heavy sites like www.cnn.com


brendankenny:
need to remove smokehouse expectations too

patrickhulce:
> At some point we're going to have to be honest with ourselves about these disabled gatherers/audits, as well. What's the likelihood that they'll come back? If they aren't on anyone's plate, they probably aren't, and they're just bitrotting in the meantime. It may be better just to delete them completely and have a link in the disabled-audit issue to the last commit where they existed for a hypothetical future person who'll take them up and make them fast so they can run again

I agree, but it's at least on my plate after I/O to pick up CSS usage and it's been nice to have the code maintained and thought about during refactors instead of having a larger rewrite later. Not sure if that applies *as much* to this audit, but I'm in favor of keeping the gatherer at least around in the code.

brendankenny:
Oh, sorry, I don't mind at all if it has someone who really is meaning to get back to it with a timeframe (though commenting out in the default config for the project does look a bit janky :), but how much does anyone love old-flexbox? The styles gatherer is going to be difficult to get to run quickly. And then there's `CacheStartUrl` and friends...

patrickhulce:
> but how much does anyone love old-flexbox

heh, yeah... :)

paulirish:
lgtm on disabling this gatherer and old-flexbox audit.

++ on patrick picking up css/js coverage audits which will bring back this gatherer. i am sure there are perf wins to be had.

i prefer commenting out as these cases fall into two options: 1) we have explicit plans to bring it back in the short term or 2) the discussion will take some time to sort out

PR #2156, opened 2017-05-05
perf(gather-runner): Clear cache selectively per-pass

### LINKED ISSUE #2089: Disk cache is disabled for the whole run

paulirish:
Instead of once at the beginning, we now clear on a pass-level, but in reality this is the same as it was the first pass that we need it for.

Additionally, don't leave "disable cache" on. Just clear it and start from empty when necessary.

Cache is cleared if we're on a "perf pass" as defined by `recordTrace && useThrottling`.

_In the name of Operation Yaquina Bay #2146_. fixes #2089

------------------

From a perf perspective the wins are not as noticeable except on high-end hardware. That's because disk cache isn't really significantly faster than the network. lol. Usually pass #4 drops ~10% in duration (for google news, it was from 900ms total to 800ms total)

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -38,7 +38,7 @@ let GathererResults; // eslint-disable-line no-unused-vars
  *     ii. beginEmulation
  *     iii. enableRuntimeEvents
  *     iv. evaluateScriptOnLoad rescue native Promise from potential polyfill
- *     v. cleanAndDisableBrowserCaches
+ *     v. cleanBrowserCaches
  *     vi. clearDataForOrigin
  *
  * 2. For each pass in the config:
@@ -110,7 +110,7 @@ class GatherRunner {
       .then(_ => driver.enableRuntimeEvents())
       .then(_ => driver.cacheNatives())
       .then(_ => driver.dismissJavaScriptDialogs())
-      .then(_ => resetStorage && driver.cleanAndDisableBrowserCaches())
+      .then(_ => resetStorage && driver.cleanBrowserCaches())
       .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
       .then(_ => gathererResults.UserAgent = [driver.getUserAgent()]);
   }
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -118,6 +118,10 @@ class GatherRunner {
       .then(_ => driver.enableRuntimeEvents())
       .then(_ => driver.cacheNatives())
       .then(_ => driver.dismissJavaScriptDialogs())
+<<<<<<< HEAD
+=======
+      .then(_ => resetStorage && driver.cleanBrowserCaches())
+>>>>>>> Don't disable disk cache, just clear it once for perf.
       .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
       .then(_ => gathererResults.UserAgent = [driver.getUserAgent()]);
   }
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -49,7 +49,8 @@ let GathererResults; // eslint-disable-line no-unused-vars
  *   B. GatherRunner.pass()
  *     i. beginTrace (if requested) & beginDevtoolsLog
  *     ii. GatherRunner.loadPage()
- *       a. navigate to options.url (and wait for onload)
+ *       a. cleanBrowserCaches
+ *       b. navigate to options.url (and wait for onload)
  *     iii. all gatherers' pass()
  *   C. GatherRunner.afterPass()
  *     i. endTrace (if requested) & endDevtoolsLog & endThrottling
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -86,13 +86,20 @@ class GatherRunner {
...
-      flags: options.flags,
-    }).then(finalUrl => {
-      options.url = finalUrl;
-    });
+    const resetStorage = !options.flags.disableStorageReset;
+    const recordTrace = options.config.recordTrace;
+    const useThrottling = options.config.useThrottling;
+    return Promise.resolve()
+      // Clear disk & memory cache if it's a perf run
+      .then(_ => resetStorage && recordTrace && useThrottling && driver.cleanBrowserCaches())
```
brendankenny: this heuristic seems a little magical, but I can't think of a better one. At some point we'll want an escape hatch (for if you want a perf pass but not a trace...e.g. running your own gatherers or in-page metrics), but in theory finishing #1826 would mean you'd be able to set a `disableStorageReset` per-pass in the config

lighthouse-core/gather/gather-runner.js
```diff
@@ -86,13 +86,20 @@ class GatherRunner {
...
    */
   static loadPage(driver, options) {
-    return driver.gotoURL(options.url, {
-      waitForLoad: true,
-      disableJavaScript: !!options.disableJavaScript,
-      flags: options.flags,
-    }).then(finalUrl => {
-      options.url = finalUrl;
-    });
+    const resetStorage = !options.flags.disableStorageReset;
```
brendankenny: we usually let `options.flags` optionally exist, but I don't care much :)

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -242,10 +242,10 @@ describe('GatherRunner', function() {
     });
   });
 
-  it('clears the network cache and origin storage', () => {
+  it('clears origin storage', () => {
```
brendankenny: do you want to check that the cache gets turned off and back on again?

lighthouse-core/gather/gather-runner.js
```diff
@@ -86,13 +86,20 @@ class GatherRunner {
...
+    const recordTrace = options.config.recordTrace;
+    const useThrottling = options.config.useThrottling;
+    return Promise.resolve()
```
brendankenny: what do you think about moving this to `pass`? It seems more like setup for navigating (like `driver.beginDevtoolsLog` and `driver.beginTracing`) rather than just the act of navigating (putting it here also means going to `about:blank` also goes through these steps, FWIW)

lighthouse-core/test/gather/gather-runner-test.js
paulirish: it doesnt go off and on, actually.  in the ended didnt go with what the ticket explained. :)

just a one time "Clear everything" at the start of the pass.

added test for that.

lighthouse-core/gather/gather-runner.js
paulirish: ya

lighthouse-core/gather/gather-runner.js
paulirish: yah seems good.

moved beginDevtoolsLog() to be in the promise chain too.

paulirish:
ptal

lighthouse-core/test/gather/gather-runner-test.js
brendankenny: > it doesnt go off and on, actually

well I meant whatever this is doing here, then :P

```js
// Toggle 'Disable Cache' to evict the memory cache
.then(_ => this.sendCommand('Network.setCacheDisabled', {cacheDisabled: true}))
.then(_ => this.sendCommand('Network.setCacheDisabled', {cacheDisabled: false}));
```

brendankenny:
> it'd be nice if we could get more smoke-style tests when we do these protocol-only changes, but I can't think of an obvious, quick way that wouldn't be flaky

Agreed. We've talked before about adding a smokehouse pass that uses completely custom gatherers doing things to verify gatherRunner/driver behavior...we could maybe write one that loads a site, then loads it again in a "perf pass", and checks the network records to make sure it didn't come from the cache...but that might still be flaky. Similar technique could be used to check #2168

lighthouse-core/test/gather/gather-runner-test.js
paulirish: not really. :) 

the only thing to really assert is that we want for the first sendCommand's promise to resolve before sending the second. (but technically that's not even necessary)

lighthouse-core/gather/gather-runner.js
```diff
@@ -38,17 +38,18 @@ let GathererResults; // eslint-disable-line no-unused-vars
...
  *   A. GatherRunner.beforePass()
  *     i. navigate to about:blank
  *     ii. all gatherers' beforePass()
  *   B. GatherRunner.pass()
- *     i. beginTrace (if requested) & beginDevtoolsLog
- *     ii. GatherRunner.loadPage()
+ *     i. cleanBrowserCaches() (if it's a perf run)
+ *     ii. beginDevtoolsLog()
+ *     iii. beginTrace (if requested)
+ *     iiv. GatherRunner.loadPage()
```
brendankenny: `iv` (and `v` below)

Issue #1999, opened 2017-04-11
Reduce size of JSON results: remove duplication


ebidel:
We currently duplicate a lot of content in `audits` and `reportCategories[x].audits.result`. `extendedInfo` is going away, but `details`, meta data, and other audit info is duplicated in both. sections. For a complex site, this can make our JSON results huge (e.g. many event listeners, urls, etc.). If the audit is used in multiple categories, the problem only gets worse :(

Example:

```
"audits": {
    "is-on-https": {
      "score": false,
      "displayValue": "1 insecure request found",
      "rawValue": false,
      "extendedInfo": {
        "formatter": "url-list",
        "value": [
          {
            "url": "ajax.googleapis.com/…3.2.1/jquery.min.js"
          }
        ]
      },
      "scoringMode": "binary",
      "name": "is-on-https",
      "category": "Security",
      "description": "Uses HTTPS",
      "helpText": "All sites should be protected with HTTPS, even ones that don't handle sensitive data. HTTPS prevents intruders from tampering with or passively listening in on the communications between your app and your users, and is a prerequisite for HTTP/2 and many new web platform APIs. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/https).",
      "details": {
        "type": "list",
        "header": {
          "type": "text",
          "text": "Insecure URLs:"
        },
        "items": [
          {
            "type": "text",
            "text": "ajax.googleapis.com/…3.2.1/jquery.min.js"
          }
        ]
      }
    },
...
"reportCategories": [
    {
      "name": "Progressive Web App",
      "weight": 1,
      "description": "These audits validate the aspects of a Progressive Web App. They are a subset of the [PWA Checklist](https://developers.google.com/web/progressive-web-apps/checklist).",
      "audits": [
        {
          "id": "is-on-https",
          "weight": 1,
          "result": {
            "score": false,
            "displayValue": "1 insecure request found",
            "rawValue": false,
            "extendedInfo": {
              "formatter": "url-list",
              "value": [
                {
                  "url": "ajax.googleapis.com/…3.2.1/jquery.min.js"
                }
              ]
            },
            "scoringMode": "binary",
            "name": "is-on-https",
            "category": "Security",
            "description": "Uses HTTPS",
[...]

PR #2159, opened 2017-05-05
Add details to "is fast on 3g" audit

ebidel:
R: all

This is really handy for debugging...knowing what requests failed the check.
screen shot 2017-05-05 at 10 23 23 am

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -25,7 +25,6 @@
 
 const Audit = require('./audit');
 const Emulation = require('../lib/emulation');
-
 const Formatter = require('../report/formatter');
 
 // Maximum TTFI to be considered "fast" for PWA baseline checklist
@@ -41,7 +40,7 @@ class LoadFastEnough4Pwa extends Audit {
       category: 'PWA',
       name: 'load-fast-enough-for-pwa',
       description: 'Page load is fast enough on 3G',
-      helpText: 'Satisfied if the _Time To Interactive_ duration is shorter than _10 seconds_, as defined by the [PWA Baseline Checklist](https://developers.google.com/web/progressive-web-apps/checklist). Network throttling is required (specifically: RTT latencies >= 150 RTT are expected).',
+      helpText: 'Satisfied if the Time To Interactive duration is shorter than 10 seconds, as defined by the [PWA Baseline Checklist](https://developers.google.com/web/progressive-web-apps/checklist). Network throttling is required (specifically: RTT latencies >= 150 RTT are expected).',
       requiredArtifacts: ['traces', 'devtoolsLogs']
     };
   }
@@ -57,16 +56,21 @@ class LoadFastEnough4Pwa extends Audit {
         // Ignore requests that don't have timing data or resources that have
         // previously been requested and are coming from the cache.
         if (!record._timing || record._fromDiskCache || record._fromMemoryCache) {
-          return undefined;
+          return {url: record._url};
         }
 
         // Use DevTools' definition of Waiting latency: https://github.com/ChromeDevTools/devtools-frontend/blob/66595b8a73a9c873ea7714205b828866630e9e82/front_end/network/RequestTimingView.js#L164
-        return record._timing.receiveHeadersEnd - record._timing.sendEnd;
+        const latency = record._timing.receiveHeadersEnd - record._timing.sendEnd;
+
+        return {
+          url: record._url,
+          latency: latency.toLocaleString(undefined, {maximumFractionDigits: 2})
+        };
       });
 
       const latency3gMin = Emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.latency - 10;
       const areLatenciesAll3G = allRequestLatencies.every(val =>
-          val === undefined || val > latency3gMin);
+          val.latency === undefined || val.latency > latency3gMin);
 
       const trace = artifacts.traces[Audit.DEFAULT_PASS];
       return artifacts.requestFirstInteractive(trace).then(firstInteractive => {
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -41,7 +40,7 @@ class LoadFastEnough4Pwa extends Audit {
       category: 'PWA',
       name: 'load-fast-enough-for-pwa',
       description: 'Page load is fast enough on 3G',
-      helpText: 'Satisfied if the _Time To Interactive_ duration is shorter than _10 seconds_, as defined by the [PWA Baseline Checklist](https://developers.google.com/web/progressive-web-apps/checklist). Network throttling is required (specifically: RTT latencies >= 150 RTT are expected).',
```
paulirish: we'll need to decide if we want to essentially rename "FI" to "TTI" and bury the old one as version 0.1.

if we dont then we'll have to change this text.

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -57,16 +56,21 @@ class LoadFastEnough4Pwa extends Audit {
         // Ignore requests that don't have timing data or resources that have
         // previously been requested and are coming from the cache.
         if (!record._timing || record._fromDiskCache || record._fromMemoryCache) {
-          return undefined;
+          return {url: record._url};
```
paulirish: since we're now using an object we can be more explicit

```js
const ret = {url: record._url};

if (!record._timing)) ret.isMissingTiming = true;
if (record._fromDiskCache || record._fromMemoryCache) ret.isFromCache = true;

if (ret.isFromCache || ret.isMissingTiming) return ret;

ret.latency = record._timing.receiveHeadersEnd - record._timing.sendEnd;
ret.latencyStr = ret.latency.toLocaleString(undefined, {maximumFractionDigits: 2});

return ret;
```

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -57,16 +56,21 @@ class LoadFastEnough4Pwa extends Audit {
...
+        return {
+          url: record._url,
+          latency: latency.toLocaleString(undefined, {maximumFractionDigits: 2})
+        };
       });
 
       const latency3gMin = Emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.latency - 10;
       const areLatenciesAll3G = allRequestLatencies.every(val =>
-          val === undefined || val > latency3gMin);
+          val.latency === undefined || val.latency > latency3gMin);
```
paulirish:
```js
const excludedLatencies = allRequestLatencies.filter(val => val.isFromCache || val.isMissingTiming);
const includedLatencies = allRequestLatencies.filter(val => !excludedLatencies.includes(val));

// assert includedLatencies.length > 1

const areLatenciesAll3G = includedLatencies.every(val => val.latency > latency3gMin);
```

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -78,12 +82,18 @@ class LoadFastEnough4Pwa extends Audit {
           value: {areLatenciesAll3G, allRequestLatencies, isFast, timeToFirstInteractive}
         };
 
+        const details = Audit.makeV2TableDetails([
+          {key: 'url', itemType: 'url', text: 'URL'},
+          {key: 'latency', itemType: 'text', text: 'Latency (ms)'},
```
paulirish: key: 'latencyStr'

and you could add another column for "Cached?" in here..

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -78,12 +82,18 @@ class LoadFastEnough4Pwa extends Audit {
...
+        ], allRequestLatencies.filter(record => record.latency !== undefined));
```
paulirish: and replace this guy with `excludedLatencies`

brendankenny:
This is helpful for debugging, but isn't it irrelevant for the audit result? The audit is for TTFI, the 3g check is just to make sure we can do the check. If a user wants to know why their TTIFI was too slow this table doesn't help.

ebidel:
> This is helpful for debugging, but isn't it irrelevant for the audit result?

We say "Network throttling is required (specifically: RTT latencies >= 150 RTT are expected)." Without seeing which requests didn't meet that requirement, it's hard to know why you failed the audit.

lighthouse-core/audits/load-fast-enough-for-pwa.js
ebidel: I can use more vars in the code if it'll make that part more clear, but I'm thinking we should start being diligent about [sliming down audit results](https://github.com/GoogleChrome/lighthouse/issues/1999). Extra bits are nice, but if we're not going to use them in a renderer, they're extra weight.

lighthouse-core/audits/load-fast-enough-for-pwa.js
ebidel: Is a cached column worth showing? In theory, this table doesn't contain any requests that come from the cache.

brendankenny:
But it's non-actionable information that looks like it's something they're doing, but unless they disabled throttling it's a lighthouse bug -- we didn't throttle everything correctly (or it's something like the caching bug). They can't slow down server responses in order to pass the test :)

We could only include the details when `areAllLatencies3G` fails, but this feels more like a `debugString` situation

ebidel:
> We could only include the details when areAllLatencies3G fails, but this feels more like a debugString situation

Yea, we should do that. `debugString` might be good, but we're not setup for that.

ebidel:
Changed so we only show the details when `areLatenciesAll3G !== true`.

lighthouse-core/audits/load-fast-enough-for-pwa.js
brendankenny: > if we dont then we'll have to change this text.

I'd like to keep it a little while longer, at least. I want to get a few plots/ runs in with it

probably makes sense to change the test regardless to reduce confusion with Consistently Interactive

PR #2160, opened 2017-05-05
perf: consolidate DBW pass into defaultPass

### LINKED ISSUE #1737: Migrate discouraged API use audits to protocol-based violation report
A [patch](https://chromium.googlesource.com/chromium/src/+/dbd9d48e4e782cd4830d971f273468599d1a2125 ) has landed in Chromium that should make it easier to collect information on:
- scroll-blocking listeners
- document.write
- Notification.requestPermission calls outside of user gesture
- navigator.geolocation calls outside of user gesture

Those are reported as `verbose` console messages with the `violation` source.

patrickhulce:
in the name of Operation Yaquina Bay (#2146), closes #1737

shaves off ~40s on heavy sites like cnn.com (latest run when combined with other changes for me now completes in <1 minute 🎉 )

Changes:
* Converts event-listeners, document.write, and permission requests on load to use the violation entries coming from the protocol instead of bespoke in-page execution
* Deletes the corresponding gatherers associated with them
* Moves all remaining audits into the performance pass since they all have `afterPass` entries only
* Deletes `no-console-time` and `no-datenow` since they monkeypatch functions with capture callsites
* Convert `tags-blocking-first-paint`'s `beforePass` false positive detection into other checks using `isLinkPreload` flag and ensuring that the request actually finished before FCP.

Caveats:
* Violations don't seem to be coming through for mutation event listeners yet, will follow up with @pavelfeldman 
* Fancier stuff that got removed could *potentially* be enabled during the perf pass depending on how heavy the callsite capturing is, but further testing is needed
* smoke tests already pass, but unit tests will need to be updated if this gets cursory 👍  :)


patrickhulce:
bueller...bueller...

lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
```diff
@@ -118,18 +118,16 @@ class EventListeners extends Gatherer {
     })).then(nestedListeners => [].concat(...nestedListeners));
   }
 
-  beforePass(options) {
```
paulirish: this change fixes the 2nd half of https://github.com/GoogleChrome/lighthouse/issues/2163

and doing it this way means we don't need pavel's CL 

\o/

patrick, you should followup with pavel separately if you'd rather migrate to violations vs this approach. but this can be post-IO

lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
```diff
@@ -118,18 +118,16 @@ class EventListeners extends Gatherer {
...
    * @return {!Promise<!Array<!Object>>}
    */
   afterPass(options) {
-    return this.unlistenForScriptParsedEvents()
+    this.driver = options.driver;
+    this._parsedScripts = new Map();
+    return Promise.resolve()
+      .then(() => this.listenForScriptParsedEvents())
+      .then(() => this.unlistenForScriptParsedEvents())
       .then(_ => options.driver.querySelectorAll('body, body /deep/ *')) // drill into shadow trees
```
paulirish: if you want to fix the 1st half of https://github.com/GoogleChrome/lighthouse/issues/2163
we're awfully close to it right now.

i'm also fine with it coming in separately.

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
```diff
@@ -104,6 +86,7 @@ function filteredAndIndexedByUrl(networkRecords) {
     // Include 404 scripts/links generated by the parser because they are likely blocking.
     if (isHtml || isParserScriptOrStyle || (isFailedRequest && isParserGenerated)) {
       prev[record._url] = {
+        isLinkPreload: record.isLinkPreload,
```
paulirish: this is a little new.

CL: https://codereview.chromium.org/2751023005
cr rev: 460094

it's not yet in stable chrome (58), but its in 59.

how do you want to handle stable?

lighthouse-core/config/default.js
```diff
@@ -121,8 +109,6 @@ module.exports = {
     "dobetterweb/external-anchors-use-rel-noopener",
     "dobetterweb/geolocation-on-start",
     "dobetterweb/link-blocking-first-paint",
-    "dobetterweb/no-console-time",
-    "dobetterweb/no-datenow",
     "dobetterweb/no-document-write",
     "dobetterweb/no-mutation-events",
```
paulirish: im seeing this: 
image

related to your change?

lighthouse-core/config/default.js
```diff
@@ -31,27 +40,6 @@ module.exports = {
       "http-redirect",
       "html-without-javascript"
     ]
-  }, {
-    "passName": "dbw",
-    "useThrottling": false,
-    "gatherers": [
-      "chrome-console-messages",
-      // "styles",
-      // "css-usage",
```
paulirish: lets keep the commented out css-usage? :)

lighthouse-core/config/default.js
```diff
@@ -12,8 +12,17 @@ module.exports = {
...
       "manifest",
+      "chrome-console-messages",
       "image-usage",
-      "accessibility"
+      "accessibility",
+      "dobetterweb/all-event-listeners",
+      "dobetterweb/anchors-with-no-rel-noopener",
+      "dobetterweb/appcache",
+      "dobetterweb/domstats",
+      "dobetterweb/optimized-images",
```
paulirish: in this gatherer, we flip `Network` domain on and off. 

i'm not sure why though.  I suppose its for the `'Network.getResponseBody'` we do. there's a chance this method will work without network being enabled. lets doublecheck that?  looking at response-compression.js, it appears to get response bodies while network is disabled:

image


(also `driver.getRequestContent` exists now which it should probably use?)

it _could_ be a problem because we may allow more network events into the devtoolsLog than what we intended. but yeah... if we dont need to enable/disable in there then let's remove.

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -92,16 +72,16 @@ module.exports = [
...
-        score: false,
-        extendedInfo: {
-          value: {
-            results: {
-              length: 6
-            }
-          }
-        }
-      },
+      // 'no-mutation-events': {
```
paulirish: add a comment about why its excluded right now?

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -45,23 +45,26 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
+      requiredArtifacts: ['TagsBlockingFirstPaint', 'traces']
     };
   }
 
   /**
    * @param {!Artifacts} artifacts
    * @param {string} tagFilter The tagName to filter on
+   * @param {number=} endedBefore The trace milisecond timestamp that offending tags must have ended
+   *    before (typically first contentful paint).
    * @param {number=} loadThreshold Filter to resources that took at least this
```
paulirish: `loadDurationThreshold`

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
```diff
@@ -45,23 +45,26 @@ class LinkBlockingFirstPaintAudit extends Audit {
...
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
-      requiredArtifacts: ['TagsBlockingFirstPaint']
+      requiredArtifacts: ['TagsBlockingFirstPaint', 'traces']
     };
   }
 
   /**
    * @param {!Artifacts} artifacts
    * @param {string} tagFilter The tagName to filter on
+   * @param {number=} endedBefore The trace milisecond timestamp that offending tags must have ended
```
paulirish: `endTimeMax`

lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
patrickhulce: +1 to separate there's already a lot going on here

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
patrickhulce: IMO it's not the end of the world to have it fire false positives for a while since there's now a mitigation that won't fire if it didn't actually occur before the paint

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
patrickhulce: doesn't need to be anymore I removed

lighthouse-core/audits/audit.js
```diff
@@ -57,6 +57,18 @@ class Audit {
   }
 
   /**
+   * @param {!Artifacts} artifacts
+   * @param {!RegExp} pattern
+   * @return {!Array}
+   */
+  static getViolationResults(artifacts, pattern) {
```
patrickhulce: sure

lighthouse-core/config/default.js
patrickhulce: we don't need network here anymore after #2154

lighthouse-core/config/default.js
patrickhulce: ha, alright, it's more dangerous now though since uncommenting could really mess up everything in the perf pass

lighthouse-core/config/default.js
patrickhulce: probably, we won't be getting the same level of detail now that we aren't listening the entire time. strange that it shows up when there are no entries though...

lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
patrickhulce: I have no strong opinions, are the mutation listeners bad enough that we should put them in everyone's console or just check for in LH? basically, is there a legitimate reason to still use them?

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
paulirish: k. sgtm

lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js
paulirish: >  basically, is there a legitimate reason to still use them?

no. you can use mutation observers. but the events are from a dark time.
only really shitty legacy code should be using them. i hope.

patrickhulce: alright cool then violation sgtm

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
patrickhulce: done

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
patrickhulce: done

lighthouse-core/audits/dobetterweb/link-blocking-first-paint.js
paulirish: need a push?

lighthouse-core/config/default.js
paulirish: sg. do  you want to remove the network.enable/disable stuff in that gatherer now, then?

lighthouse-core/audits/dobetterweb/no-console-time.js
```diff
@@ -1,82 +0,0 @@
-/**
- * @license
- * Copyright 2016 Google Inc. All rights reserved.
- *
```
ebidel: why were these audits removed!?

ebidel: @patrickhulce

patrickhulce: @ebidel 

They were the only audits left that had perf impact and would need their own pass. Somewhere there was also a thread with @brendankenny on the dangers of letting these run during the perf pass which is also why I recommended not enabling them until we proved they had minimal impact. 

From the PR description:

> Deletes no-console-time and no-datenow since they monkeypatch functions with capture callsites
> Fancier stuff that got removed could potentially be enabled during the perf pass depending on how heavy the callsite capturing is, but further testing is needed

ebidel:
Right but can't we keep the code around for people to run themselves?

patrickhulce:
> Right but can't we keep the code around for people to run themselves?

Oh sure I have no qualms with that 👍  I'll defer to @brendankenny who expressed disgruntlement at the time when I wanted to keep code around for unused audits :)

PR #2164, opened 2017-05-05
Report: convert rest of URL list audits

patrickhulce:
didn't actually implement a URL list renderer since tables seem to be fine

image

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -77,7 +77,7 @@ class Audit {
     const tableRows = results.map(item => {
       return headings.map(heading => {
         const value = item[heading.key];
-        if (typeof value === 'object' && value.type) return value;
+        if (typeof value === 'object' && value && value.type) return value;
 
         return {
           type: heading.itemType,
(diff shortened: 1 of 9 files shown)


lighthouse-core/audits/audit.js
```diff
@@ -77,7 +77,7 @@ class Audit {
     const tableRows = results.map(item => {
       return headings.map(heading => {
         const value = item[heading.key];
-        if (typeof value === 'object' && value.type) return value;
+        if (typeof value === 'object' && value && value.type) return value;
```
paulirish: the 2nd condition? how would it pass the typeof if it was already falsey?

patrickhulce: `null` 💔

paulirish: ah my bad. ;)

ebidel:
Do you want to add a few assert in these tests that make sure `details` is there?

patrickhulce:
> Do you want to add asserts in the tests to make sure details is there?

👍 done

PR #2167, opened 2017-05-05
feat: backcompat config in report

### LINKED ISSUE #2165: Don't break back-compat of the LHR
We'll have people saving lhr.json files more often now. And if they drag one from folder into devtools audits, they'll expect to see a report.

My proposal to make this easier:

* the LHR gets a `.config` property which is the actual config.
* do the [`generateReportJson()`/etc work](https://github.com/GoogleChrome/lighthouse/blob/581eec37cb08cd9d13be7da4ad41465c70bdd64f/lighthouse-core/runner.js#L136-L150)  inside of the `generateHTML` method
* optional: we could even expose a more static `scoreReport` method in case other modules wanted scores/certifications. (Similar public API-ness makes sense for the pw-metrics-events metric gathering_

But to paraphrase pavel, the sooner you start guaranteeing the backcompat of this API the better.

patrickhulce:
fixes #2165 
This changes both report generation functions in report generator to accept a lighthouse results object with a `.config` property OR their original inputs (backcompat ftw)

### DIFF
--- a/lighthouse-core/report/v2/report-generator.js
+++ b/lighthouse-core/report/v2/report-generator.js
@@ -96,11 +96,17 @@ class ReportGeneratorV2 {
 
   /**
    * Returns the report JSON object with computed scores.
-   * @param {{categories: !Object<{audits: !Array}>}} config
-   * @param {!Object<{score: ?number|boolean|undefined}>} resultsByAuditId
+   * @param {{categories: !Object<{audits: !Array}>, config: ?Object, audits: ?Object}} config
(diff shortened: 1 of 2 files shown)


brendankenny:
doesn't this get back to the larger question of if there's a difference between the json report and lhr?

also, one issue is that config isn't always serializable. We'd need to drop live audits/gatherer classes if we want them to be

patrickhulce:
yeah actually this isn't going to work :/

Issue #1480, opened 2017-01-17
Consider adding Chrome stable to Travis matrix


patrickhulce:
Context: Recently CSS usage broke Lighthouse master on Chrome stable since the protocol method wasn't added until v56 (#1473). 

It would be great to have Travis automatically check against these types of errors. Not sure how we would want to handle the failures/tests though since we would still want to ensure the smoke tests for canary were run.

brendankenny:
Simplest solution: smoke tests take a combined ~1.5 minutes on travis right now. We could easily download Chrome latest *and* Chrome stable and just double that up. Obviously that's a lot of duplication of work, though, and only going to get more painful as the smoke tests grow.

patrickhulce:
Oh @brendankenny I took that as a given, I meant more how do we handle the failures that result because stable lacks the feature support. Example: the CSS usage gatherer will fail on stable as expected and set a debug string which won't pass the expectations file. Should we maintain separate expectations, mark as OK to fail, etc?

wardpeet:
what do you guys think about 2 travis checks? I for code linting & 1 for smokehouse?

brendankenny:
for these differences between stable and canary, I think we're just going to have to go back to being stricter about not landing new audits or driver commands until supported by Chrome stable, rather than having a different expectation for each version.

This will still be a bit hard with audit changes due to Chrome changes, e.g. new deprecation warnings coming in Canary (one of the errors in the log in https://github.com/GoogleChrome/lighthouse/issues/1573#issuecomment-276129392)

patrickhulce:
fixed by #2547 ! 🎉  moving forward we'll probably want to stick to tests that only pass in stable

XhmikosR:
Cool, I'll add it in my long to-do list to update AppVeyor soon-ish

On Jun 20, 2017 22:13, "Patrick Hulce" <notifications@github.com> wrote:

fixed by #2547 <https://github.com/GoogleChrome/lighthouse/pull/2547> ! 🎉
moving forward we'll probably want to stick to tests that only pass in
stable

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/GoogleChrome/lighthouse/issues/1480#issuecomment-309859078>,
or mute the thread

PR #2168, opened 2017-05-05
perf: request block stylesheets and images on redirect pass

patrickhulce:
shaves off ~10s on certain sites like theverge.com

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -28,17 +28,7 @@ module.exports = {
     "passName": "redirectPass",
     "useThrottling": false,
     // Speed up the redirect pass by blocking stylesheets, fonts, and images
-    "blockedUrlPatterns": [
-      "*.css",
-      "*.jpg",
-      "*.jpeg",
-      "*.png",
-      "*.gif",
-      "*.svg",
-      "*.ttf",
-      "*.woff",
-      "*.woff2",
-    ],
+    "blockedUrlPatterns": ["*.css", "*.jpg", "*.jpeg", "*.png", "*.gif", "*.svg", "*.ttf", "*.woff", "*.woff2"],
     "gatherers": [
       "http-redirect",
       "html-without-javascript"
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -27,6 +27,18 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
+    // Speed up the redirect pass by blocking stylesheets, fonts, and images
+    "blockedUrlPatterns": [
+      "*.css",
+      "*.jpg",
+      "*.jpeg",
+      "*.png",
+      "*.gif",
+      "*.svg",
+      "*.ttf",
+      "*.woff",
+      "*.woff2",
+    ],
     "gatherers": [
       "http-redirect",
       "html-without-javascript"
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -853,9 +853,7 @@ class Driver {
   }
 
   /**
-   * This method sets the URL patterns to be blocked and is called at the beginning of each pass.
-   * No "clearing" is done at the end of the pass since the next pass will set to [] if necessary.
-   * @param {!Array<string>} urls
+   * @param {!Array<string>} urls URL patterns to block. Wildcards ('*') are allowed.
    * @return {!Promise}
    */
   blockUrlPatterns(urls) {
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -859,7 +859,13 @@ class Driver {
    * @return {!Promise}
    */
   blockUrlPatterns(urls) {
-    return this.sendCommand('Network.setBlockedURLs', {urls});
+    return this.sendCommand('Network.setBlockedURLs', {urls})
+      .catch(err => {
+        // TODO: remove this handler once m59 hits stable
+        if (!/wasn't found/.test(err.message)) {
+          throw err;
+        }
+      });
   }
 
   /**
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -852,6 +852,12 @@ class Driver {
     return collectUsage;
   }
 
+  /**
+   * This method sets the URL patterns to be blocked and is called at the beginning of each pass.
+   * No "clearing" is done at the end of the pass since the next pass will set to [] if necessary.
+   * @param {!Array<string>} urls
+   * @return {!Promise}
+   */
   blockUrlPatterns(urls) {
     return this.sendCommand('Network.setBlockedURLs', {urls});
   }
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -852,9 +852,8 @@ class Driver {
     return collectUsage;
   }
 
-  blockUrlPatterns(urlPatterns) {
-    const promiseArr = urlPatterns.map(url => this.sendCommand('Network.addBlockedURL', {url}));
-    return Promise.all(promiseArr);
+  blockUrlPatterns(urls) {
+    return this.sendCommand('Network.setBlockedURLs', {urls});
   }
 
   /**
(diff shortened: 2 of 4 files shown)


lighthouse-core/config/default.js
```diff
@@ -27,6 +27,18 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
+    // Speed up the redirect pass by blocking stylesheets, fonts, and images
+    "blockedUrlPatterns": [
+      "*.css",
```
paulirish: formatting nit:  this is pretty low signal info, compared to the rest of config. 
can these all share a line?

lighthouse-core/gather/gather-runner.js
```diff
@@ -45,7 +45,8 @@ let GathererResults; // eslint-disable-line no-unused-vars
  * 2. For each pass in the config:
  *   A. GatherRunner.beforePass()
  *     i. navigate to about:blank
- *     ii. all gatherers' beforePass()
+ *     ii. Enable network request blocking for specified patterns
```
paulirish: > set patterns for network request blocking

and you can remove "vii. blockUrlPatterns" from above.

lighthouse-core/gather/driver.js
```diff
@@ -852,9 +852,8 @@ class Driver {
     return collectUsage;
   }
 
-  blockUrlPatterns(urlPatterns) {
-    const promiseArr = urlPatterns.map(url => this.sendCommand('Network.addBlockedURL', {url}));
-    return Promise.all(promiseArr);
+  blockUrlPatterns(urls) {
```
paulirish: jsdoc?

`{!Array<string>}` yeah?

lighthouse-core/config/default.js
patrickhulce: yup, done

lighthouse-core/gather/gather-runner.js
patrickhulce: done

lighthouse-core/gather/driver.js
patrickhulce: yup, done

patrickhulce:
> Can you add a comment on the blockUrlPatterns method that mentions this trick?

Yeah good call done, blocking will cease when we disconnect, right?

lighthouse-core/gather/driver.js
```diff
@@ -852,9 +852,14 @@ class Driver {
...
-    const promiseArr = urlPatterns.map(url => this.sendCommand('Network.addBlockedURL', {url}));
-    return Promise.all(promiseArr);
+  /**
+   * This method sets the URL patterns to be blocked and is called at the beginning of each pass.
+   * No "clearing" is done at the end of the pass since the next pass will set to [] if necessary.
+   * @param {!Array<string>} urls
+   * @return {!Promise}
+   */
+  blockUrlPatterns(urls) {
+    return this.sendCommand('Network.setBlockedURLs', {urls});
```
paulirish: unfortunately this command was renamed/added recently..

commit: https://chromium.googlesource.com/chromium/src/+/01c043f5f744262981becd3d45a79981ad3f6b67

m58 (base rev 454471 ) which shipped to stable 2 weeks ago doesn't have this method. It has the add/remove pair instead.

IMO this is optional, and our other use of requestblocking is unused so i think we can ignore the failure. let's catch and exclude errors?

doublecheck this works with a stable m58 build

patrickhulce: what's going to be our policy going forward about handling stable chrome and protocol changes? it seems like it's mostly taken a backseat and we emphasize canary, should we add  a stable config to travis if we want to support it (#1480)?

confirmed working now (with an error log) in 58

lighthouse-core/gather/driver.js
```diff
@@ -852,9 +852,20 @@ class Driver {
     return collectUsage;
   }
 
-  blockUrlPatterns(urlPatterns) {
-    const promiseArr = urlPatterns.map(url => this.sendCommand('Network.addBlockedURL', {url}));
-    return Promise.all(promiseArr);
+  /**
+   * This method sets the URL patterns to be blocked and is called at the beginning of each pass.
+   * No "clearing" is done at the end of the pass since the next pass will set to [] if necessary.
```
brendankenny: since some future gatherers or other consumers of lighthouse (e.g. performanceExperiment) may want to use this method, probably makes more sense to move the bulk of this comment into `beforePass` where it's used like this

lighthouse-core/gather/driver.js
```diff
@@ -852,9 +852,20 @@ class Driver {
...
+   * @param {!Array<string>} urls
```
brendankenny: maybe worth adding `param` doc similar to what the protocol docs has? ("URL patterns to block. Wildcards ('*') are allowed.")

lighthouse-core/gather/gather-runner.js
```diff
@@ -171,9 +170,12 @@ class GatherRunner {
    * @return {!Promise}
    */
   static beforePass(options, gathererResults) {
+    const blockedUrls = (options.config.blockedUrlPatterns || [])
+      .concat(options.flags.blockedUrlPatterns || []);
```
brendankenny: how do unions of url patterns work? Last one wins? (just curious)

brendankenny: aside: we really need to get the overloaded naming of `options` under control here :)

lighthouse-core/config/default.js
```diff
@@ -27,6 +27,8 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
+    // Speed up the redirect pass by blocking stylesheets, fonts, and images
```
brendankenny: love having this in the default config as an example for custom configs

lighthouse-core/gather/gather-runner.js
brendankenny: > how do unions of url patterns work? Last one wins? 

oh, is `!` not supported? I guess in that case it doesn't matter

paulirish: ! isnt supported, correct.

lighthouse-core/config/default.js
```diff
@@ -27,6 +27,8 @@ module.exports = {
...
+    "blockedUrlPatterns": ["*.css", "*.jpg", "*.jpeg", "*.png", "*.gif", "*.svg", "*.ttf", "*.woff", "*.woff2"],
```
brendankenny: not sure about this one, but does it make sense to have these apply across `beforePass`, `pass`, and `afterPass`? Maybe it would make more sense/be less surprising to have them blocked only during `pass`, analogous to `recordTrace`, `useThrottling`, etc cc/ @paulirish

lighthouse-core/gather/driver.js

lighthouse-core/gather/driver.js

lighthouse-core/config/default.js
patrickhulce: yeah few random thoughts here
1) we don't have any gatherers that load *new* content that wasn't already requested by the page
2) we should probably discourage that behavior as something outside of Lighthouse realm anyway, and suggest requesting on the node side if a third party integration wants to do that rather than an `evaluateAsync(fetch('my-url'))`
3) since afterPass doesn't freeze the page or anything and we're not guaranteed to be done at the 25s timeout it felt weird to unblock these urls while the page could feasibly still be requesting or retrying them

brendankenny: > we don't have any gatherers that load new content that wasn't already requested by the page

that's actually why I brought it up, so it wouldn't surprise us when someone does write a gatherer that tries it :)

> we should probably discourage that behavior as something outside of Lighthouse realm anyway, and suggest requesting on the node side if a third party integration wants to do that rather than an evaluateAsync(fetch('my-url'))

but that's definitely something *we* rely on when we need to get around authentication or whatever. As an easy workaround to some of those issues, fetching in the context of the page can't be beat.

> since afterPass doesn't freeze the page or anything and we're not guaranteed to be done at the 25s timeout it felt weird to unblock these urls while the page could feasibly still be requesting or retrying them

yeah, good point, and this is what really separates it from `useThrottling`. While matching their behavior makes some sense from a user-surprise perspective, disabling `useThrottling` in `afterPass` means the run goes faster due to potential work going faster, while disabling url blocking potentially means the run going slower due to (potentially) *making* work.

My real hesitation is magic while having no data about actual downsides. This is why we need users making custom gatherers :)

In the meantime, I suspect we'll have to revisit scheduling of all of this for #1826 anyways, so I'm good with this.

PR #2171, opened 2017-05-06
report: add tooltips to sparklines

### LINKED ISSUE #2169: Sparkline improvements
The numbers of the right should have `title` attributes so that when you hover over them, they hint at the meaning of the numbers. Maybe one for the sparkline too.

screen shot 2017-05-05 at 4 34 05 pm

For pages that do well, one "failing" audit with a small sparkline looks like a bug. Not sure how to handle these.

screen shot 2017-05-05 at 4 28 34 pm

patrickhulce:
fixes #2169 

image

we kind of want to be intentionally vague here since we don't guarantee anything in particular improves by that amount. I'll also be moving through here in a bit to address the other component of that issue by splitting these into "definitely do" and moving the more aggresive/opinionated ones into "Diagnostics"


paulirish:
@patrickhulce lint error
image

patrickhulce:
@paulirish fixed

Issue #1702, opened 2017-02-12
Invalid JSON format on toepoke.co.uk manifest.json


toepoke:
Hi,

When running the report I get the following error:

"ERROR: file isn't valid JSON: SyntaxError: Unexpected token ﻿ in JSON at position 0"

It's been driving me mental for weeks!  Turns out it's because the manifest.json was saved with a byte-order-mark (I'm on Windows).  Saved without the BOM and all works well.

May I suggest adding a warning to the documentation so others don't have a similar battle?

Many thanks,
Franz.

XhmikosR:
As far as I can tell LH is using the standard `JSON.parse` method.

Might be worth looking into https://github.com/sindresorhus/strip-bom/blob/master/index.js

brendankenny:
Not sure what the correct thing to do here is. A BOM *is* an invalid character in JSON, but in some cases clients will drop the BOM before attempting to parse.

At the very least we could warn that the first character is a BOM to give a better error message, though.

If I test a manifest saved with a BOM, the application panel in devtools lists no information for the manifest. Since IIUC devtools is exposing what Chrome has actually parsed from the manifest, this implies Chrome doesn't understand the manifest either (devtools does not report a JSON SyntaxError, though, or any error at all).

If the browser doesn't understand it, we should definitely mark it as an error, just with a better error message.

OTOH, (one of?) the native Chrome JSON parser(s) [explicitly skips BOMs when encountered](https://cs.chromium.org/chromium/src/base/json/json_parser.cc?type=cs&l=88). Not sure how that interacts with the manifest parser.

toepoke:
HI @brendankenny,

To confirm when I was saving with BOM I didn't see the manifest parameters in DevTools.  They magically start appearing when saved without BOM - which is in line with what you're seeing.

If it's not feasible to raise an error, perhaps a warning that _always_ appears in the report would be sufficient?  I'm just conscious of others falling into the same issue.

If I can be of any assistance, please give me a shout.

XhmikosR:
IMO, we shouldn't do something special. It's an error in the parser so Chrome will not load it as I understand it.

paulirish:
@toepoke when you look at the manifest panel in devtools, do you see warnings or errors?
![image](https://cloud.githubusercontent.com/assets/39191/23564521/1da6a2a2-ffff-11e6-89ac-42c28e374f44.png)

also same question for @zigmandel, as it seems like a related issue.

[...]

PR #2175, opened 2017-05-07
Throw error when manifest is BOM encoding

wardpeet:
Fixes issue #1702


paulirish:
I didn't read the full history of #1702 but i have some recent news:

AFAIK the chrome manifest parser had no problem with BOM but devtools' application panel did. (is that confirmed?)
This was fixed just a few weeks back though: https://codereview.chromium.org/2837943003

so afaik, there won't be a problem with BOM parsing in chrome anymore.

With that status, should we still throw in this situation?

wardpeet:
Let me give it a spin if we get our manifest back from devtools.
If we do get it without any problems this PR is useless 😄

Before the devtools fix it gave a null manifest

I'll close it myself when i've tested it

paulirish:
hmmmm!

we might still have a problem then.

the devtools fix was only in the frontend, so yeah looks like we'll need this patch afterall.

(real chrome doesn't use the protocol method. that's the difference here)

lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -48,6 +49,12 @@ class Manifest extends Gatherer {
           return null;
         }
 
+        const isBomEncoded = Buffer.from(response.data).slice(1, 4)
+          .equals(new Buffer(BOM_ENCODING));
+        if (isBomEncoded) {
+          throw new Error('Manifest is encoded with BOM. Please remove the BOM encoding');
```
paulirish: rather than throw error let's just remove the BOM and let it proceed.

wardpeet:
@paulirish fixed! ;)

paulirish:
Nice I also tested with the fixture in https://bugs.chromium.org/p/chromium/issues/detail?id=706926 which is what we validated the devtools fix against.
lg.

lighthouse-core/test/gather/gatherers/manifest-test.js
```diff
@@ -47,6 +47,46 @@ describe('Manifest gatherer', () => {
...
+              url: EXAMPLE_MANIFEST_URL
+            });
+          }
+        },
+        url: EXAMPLE_DOC_URL
+      };
+    };
+
+    const promises = [];
+    promises.push(manifestGather.afterPass(getDriver(manifestWithBOM))
```
brendankenny: I don't think this will throw even if still bad...won't it show with an "isn't valid JSON" `debugString`?

It would be good to check the result from `afterPass`...`result.raw` should have the raw string passed in, so it should have length equal to `manifestWithBOM` minus the BOM. And it would be good to check some of the parsed values to make sure it's being parsed

lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -36,6 +37,12 @@ class Manifest extends Gatherer {
   afterPass(options) {
     return options.driver.getAppManifest()
       .then(response => {
+        const isBomEncoded = Buffer.from(response.data).slice(0, BOM_ENCODING.length)
+          .equals(new Buffer(BOM_ENCODING));
```
brendankenny: what's the compatibility like if we skip the Buffer slicing for the check (so files without a BOM don't have to go through that) and do `const isBomEncoded = response.data.charCodeAt(0) === 65279;`. Does that work cross platform?

lighthouse-core/test/gather/gatherers/manifest-test.js
wardpeet: damn forgot to change the test text as we throwed before it should say pass when manifest has BOM

lighthouse-core/gather/gatherers/manifest.js
wardpeet: I can test on windows, linux & mac just to be sure

wardpeet:
@brendankenny tested your change on Windows 10, mac and linux and seems to work for me.

paulirish:
Awesome. Thanks for testing that out. We'll let this PR and TTFB chill until we cut 2.0, and then we'll incorporate. Sorry for making you wait!

wardpeet:
@paulirish you never have to say sorry 😉  

It's no urgent bug so it can wait

wardpeet:
@brendankenny everything should be ready to go

Issue #2092, opened 2017-04-28
Make the ChromeLauncher API into a real thing.


paulirish:
<details>
<summary>
Here's a few incarnations I'm seeing of people using the ChromeLauncher API:
</summary>

![image](https://cloud.githubusercontent.com/assets/39191/25513433/e90fec4a-2b89-11e7-85b8-eaff641fd9b5.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513435/ed717542-2b89-11e7-840d-a062a7b14b3e.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513445/fe9a72ec-2b89-11e7-8b0f-64fbd9bd816f.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513458/1c54209e-2b8a-11e7-8d21-33a5fd9bd501.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513492/563ca20e-2b8a-11e7-9abd-9e4a0617969b.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513503/612bfab6-2b8a-11e7-9ff5-a84890cf5ec2.png)
![image](https://cloud.githubusercontent.com/assets/39191/25513512/70a358e0-2b8a-11e7-80f0-cf4d31350bcd.png)

and this dude who's just like "wtf is autoselect?": https://github.com/siddharthkp/lighthouse-ci/blob/c3938a2523434f26ea7963b0546fb241d4f21d36/index.js

</details>

_(click above)_

and not to mention the example in lighthouse's readme [which is like whaaaaa.](https://github.com/GoogleChrome/lighthouse#using-programmatically).

There's a lot of variety and nothing I'm seeing in those snippets actually _looks correct._

**Feels like we could use chrome-launcher module that has a real first-class API.**

Something with a very clear API surface to handle launch, getting the port number (if set with 0), ability to kill.. 

It'd also be nice to have sigint [kill the process automatically as well](https://github.com/paulirish/pwmetrics/blob/4070b62ccc0e84c25fd81ce6e3e32c284f8e639b/lib/index.ts#L61-L71). (our poor manual-chrome-launcher doesn't have this currently).

wdyt?

+ @samccone @ebidel

ebidel:
In!

brendankenny:
> getting the port number (if set with 0), ability to kill

[...]

PR #2176, opened 2017-05-07
Warn users if they have not yet generated the file.

samccone:
Prior to this fix users would be confused why their unit tests were
failing. The error was a missing file during a require, which made it seem like
something was messed up with their local clone. Now we can throw a
useful error message.

ref #2092


paulirish:
@googlebot shh bby.

PR #2177, opened 2017-05-07
Cli testing

samccone:
Prior to this fix, the tests *had* to be run from the root lighthouse
dir, this fix makes this test more portable, so for instance you can
run the tests from within the cli folder if you wanted

ref #2092

### DIFF
--- a/lighthouse-cli/package.json
+++ b/lighthouse-cli/package.json
@@ -4,9 +4,11 @@
   "main": "index.js",
   "scripts": {
     "build": "tsc",
-    "dev": "tsc -w"
+    "dev": "tsc -w",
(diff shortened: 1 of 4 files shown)


lighthouse-cli/package.json
```diff
@@ -4,9 +4,11 @@
   "main": "index.js",
   "scripts": {
     "build": "tsc",
-    "dev": "tsc -w"
+    "dev": "tsc -w",
+    "test": "mocha --reporter dot test/**/*-test.js"
   },
   "devDependencies": {
+    "mocha": "^3.3.0",
```
paulirish: can you match the `"^3.2.0",` that we do in root?

samccone: good call

PR #2180, opened 2017-05-08
CLI: extract bin flag def from index

### LINKED ISSUE #2092: Make the ChromeLauncher API into a real thing.

samccone:
* Extract Flags interface instead of defining inline in the function defs.
* Extract bin flags into a standalone file since this is more about config and less about logic.

ref #2092

depends on 
https://github.com/GoogleChrome/lighthouse/pull/2173

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -25,16 +25,15 @@ const assetSaver = require('../lighthouse-core/lib/asset-saver.js');
 const getFilenamePrefix = require('../lighthouse-core/lib/file-namer.js').getFilenamePrefix;
 import {ChromeLauncher} from './chrome-launcher';
 import * as Commands from './commands/commands';
+import {getFlags, Flags} from './cli-flags';
 const lighthouse = require('../lighthouse-core');
 const log = require('../lighthouse-core/lib/log');
-const Driver = require('../lighthouse-core/gather/driver.js');
 import * as path from 'path';
 const perfOnlyConfig = require('../lighthouse-core/config/perf.json');
 const performanceXServer = require('./performance-experiment/server');
 import * as Printer from './printer';
 import * as randomPort from './random-port';
 import {Results} from './types/types';
-const yargs = require('yargs');
 const pkg = require('../package.json');
 
 // accept noop modules for these, so the real dependency is optional.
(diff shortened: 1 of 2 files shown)


lighthouse-cli/bin.ts
```diff
@@ -27,107 +27,37 @@ import {ChromeLauncher} from './chrome-launcher';
...
-        .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.html])
-        .default('port', 9222)
-        .default('max-wait-for-load', Driver.MAX_WAIT_FOR_FULLY_LOADED)
-        .check((argv: {listAllAudits?: boolean, listTraceCategories?: boolean, _: Array<any>}) => {
-          // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
-          // or --list-trace-categories. If not, stop the program and ask for a url
-          if (!argv.listAllAudits && !argv.listTraceCategories && argv._.length === 0) {
-            throw new Error('Please provide a url');
-          }
+interface Flags {
```
patrickhulce: shouldn't this interface be defined and exported in `cli-flags.ts`?

paulirish: separately, thanks for extracting this. the # of times this type was inlined was killing me.

samccone:
All updated PTAL.

lighthouse-cli/bin.ts
```diff
@@ -322,39 +216,35 @@ function saveResults(results: Results,
...
+      await chromeLauncher!.kill();
     }
 
-    // This will be reenabled once flags are extracted to standalone interface.
-    // clang-format off
+    return handleError(err);
+  }
+}
+
+// This will be reenabled once flags are extracted to standalone interface.
```
paulirish: is that now?

lighthouse-cli/cli-flags.ts
```diff
@@ -0,0 +1,115 @@
+const yargs = require('yargs');
```
paulirish: add license header

use strict?

lighthouse-cli/bin.ts
```diff
@@ -186,29 +84,27 @@ if (cliFlags.output === Printer.OutputMode[Printer.OutputMode.json] && !cliFlags
...
-  return randomPort.getRandomPort().then(portNumber => {
-    flags.port = portNumber;
-    log.verbose('Lighthouse CLI', `Using generated port ${flags.port}.`);
+    log.verbose('Lighthouse CLI', 'Generating random port.');
+    return randomPort.getRandomPort().then(portNumber => {
+      flags.port = portNumber;
+      log.verbose('Lighthouse CLI', `Using generated port ${flags.port}.`);
+    });
   });
-  })
```
paulirish: welcome back, semicolon.

lighthouse-cli/cli-flags.ts
samccone: done

lighthouse-cli/bin.ts
samccone: done

lighthouse-cli/bin.ts
samccone: 👋

PR #2181, opened 2017-05-08
Report: Perf metrics filmstrip (WIP)

paulirish:
pavel wrote this perf filmstrip as a devtools patch: https://codereview.chromium.org/2861053003/
so here i am upstreaming it.

There's some rough edges and I wanted to get a few thoughts before proceeding.

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -333,7 +333,7 @@ export async function runLighthouse(url: string,
     // Filter out artifacts except for screenshots in traces to minimize report size.
     var traces = results.artifacts.traces;
     for (var pass in traces)
-      traces[pass]['traceEvents'] = traces[pass]['traceEvents'].filter(e => e['cat'] === 'disabled-by-default-devtools.screenshot');
+      traces[pass]['traceEvents'] = traces[pass]['traceEvents'].filter((e: {cat: string}) => e['cat'] === 'disabled-by-default-devtools.screenshot');
     results.artifacts = { traces: traces };
 
     const artifacts = results.artifacts;
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -330,8 +330,14 @@ export async function runLighthouse(url: string,
     const chromeLauncher = await getDebuggableChrome(flags)
     const results = await lighthouse(url, flags, config);
 
+    // Filter out artifacts except for screenshots in traces to minimize report size.
+    var traces = results.artifacts.traces;
+    for (var pass in traces)
+      traces[pass]['traceEvents'] = traces[pass]['traceEvents'].filter(e => e['cat'] === 'disabled-by-default-devtools.screenshot');
+    results.artifacts = { traces: traces };
+
     const artifacts = results.artifacts;
-    delete results.artifacts;
+    // delete results.artifacts;
 
     await saveResults(results, artifacts!, flags);
     if (flags.interactive) {
--- a/lighthouse-core/report/v2/renderer/category-renderer.js
+++ b/lighthouse-core/report/v2/renderer/category-renderer.js
@@ -209,6 +209,13 @@ class CategoryRenderer {
   }
 
   /**
+   * @param {*} artifacts
+   */
+  setArtifacts(artifacts) {
+    this._artifacts = artifacts;
+  }
+
+  /**
    * @param {!Document|!Element} context
    */
   setTemplateContext(context) {
@@ -302,7 +309,11 @@ class CategoryRenderer {
     metricAuditsEl.open = true;
     element.appendChild(metricAuditsEl);
 
-
+    const artifacts = this._artifacts;
+    const timelineElement = this._renderPerfTimeline(metricAudits, artifacts);
+    const metricsDescEl = metricAuditsEl.querySelector(':scope > .lh-audit-group__description');
+    if (metricsDescEl) console.log('wtffff');
+    metricsDescEl.parentElement.insertBefore(timelineElement, metricsDescEl.nextSibling);
 
     const hintAudits = category.audits
         .filter(audit => audit.group === 'perf-hint' && audit.score < 100)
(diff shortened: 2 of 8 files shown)


lighthouse-cli/bin.ts
```diff
@@ -330,8 +330,14 @@ export async function runLighthouse(url: string,
     const chromeLauncher = await getDebuggableChrome(flags)
     const results = await lighthouse(url, flags, config);
 
+    // Filter out artifacts except for screenshots in traces to minimize report size.
+    var traces = results.artifacts.traces;
+    for (var pass in traces)
+      traces[pass]['traceEvents'] = traces[pass]['traceEvents'].filter(e => e['cat'] === 'disabled-by-default-devtools.screenshot');
```
paulirish: we need screenshots, and right now they're only in the trace.
we could:

1. keep this monstrosity
2. bring back the screenshots audit
3. throw screenshots into extInfo,etc  of the psi audit

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -336,6 +350,86 @@ class CategoryRenderer {
...
+    const fmp = metricAudits.find(a => a.id === 'first-meaningful-paint').result;
+    if (!fmp || !fmp['extendedInfo'])
+      return;
+
+    const tti = metricAudits.find(a => a.id === 'time-to-interactive').result;
+    if (!tti || !tti['extendedInfo'])
+      return;
+
+    const navStart = fmp['extendedInfo']['value']['timestamps']['navStart'];
+    const markers = [
```
paulirish: it'd be lovely to use [`metricsDefinitions()` in pwmetrics-events.js](https://github.com/GoogleChrome/lighthouse/blob/5fa684ef993f13f9028a306fdfb5e7566c27f7fc/lighthouse-core/lib/traces/pwmetrics-events.js#L47) for this section here, however we're currently in the standalone report.

lighthouse-core/report/v2/report-styles.css
```diff
@@ -574,6 +574,64 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
...
+    margin: 4px 0 6px;
+    width: calc(68px * 7);
+}
+
+.lh-timeline-subtitle {
+    color: #01579B;
+}
+
+.lh-timeline-bar {
+    border-top: 4px solid #03A9F4;
```
paulirish: i'd like to color these according to audit score, instead.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -336,6 +350,86 @@ class CategoryRenderer {
...
+      {
+        title: 'First contentful paint',
```
paulirish: question:  should we include the audit entry for these metrics if we have them here?
probably not
but that means we'll have to include their helpText in their visualization.

image


if we do that, then we'll only show bars for the metrics audits we have, with the exception of EIL, which is our non pageload metric. 
and this also means if we want to show "first contentful paint" or "first visual change" etc.. then they should be real top-level audits.

that sound reasonable?

lighthouse-cli/bin.ts
```diff
@@ -330,8 +330,14 @@ export async function runLighthouse(url: string,
     const chromeLauncher = await getDebuggableChrome(flags)
     const results = await lighthouse(url, flags, config);
 
+    // Filter out artifacts except for screenshots in traces to minimize report size.
+    var traces = results.artifacts.traces;
+    for (var pass in traces)
```
ebidel: braces would be good here

lighthouse-cli/bin.ts
```diff
@@ -330,8 +330,14 @@ export async function runLighthouse(url: string,
     const chromeLauncher = await getDebuggableChrome(flags)
     const results = await lighthouse(url, flags, config);
 
+    // Filter out artifacts except for screenshots in traces to minimize report size.
+    var traces = results.artifacts.traces;
```
ebidel: const

lighthouse-cli/bin.ts
patrickhulce: I'm in favor of bringing back a screenshots audit but only in thumbnail form and perhaps using SSIM to throttle the exported frames to major changes and not show every loading spinner animation, etc

I can throw together a PR for this if we're interested.

lighthouse-cli/bin.ts
ebidel: Are we not linting .ts files?

lighthouse-cli/bin.ts
```diff
@@ -330,8 +330,14 @@ export async function runLighthouse(url: string,
...
+    for (var pass in traces)
+      traces[pass]['traceEvents'] = traces[pass]['traceEvents'].filter((e: {cat: string}) => e['cat'] === 'disabled-by-default-devtools.screenshot');
+    results.artifacts = { traces: traces };
+
     const artifacts = results.artifacts;
-    delete results.artifacts;
+    // delete results.artifacts;
```
ebidel: remove

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: there's been a need to consume these in a handy place and not have DCL, onLoad, traceEnd, etc scattered across TTI and random other audits.

wdyt about having a meta audit for the `metricsDefinitions` that can reference the others/pull unrelated ones out of the trace, and pwmetrics-events just injects everything that's there?

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: sgtm, though honestly it may make more sense to keep speed index out of the timeline too since people might be looking for what happened at that point not knowing it's an average

lighthouse-cli/bin.ts
paulirish: +1 on fixing speedline perf, but let's not block the filmstrip here on it. 

*nod* on a dedicated screenshots audit. seems good to have it top-level than hide it in ext-info.

lighthouse-core/report/v2/renderer/category-renderer.js
ebidel: I like the idea of having top level audits for all of the ones that have bars. But that can also wait until after IO.

Should we consolidate these two sections to remove duplicated info? We could put the score next to the bar and make the bar section expandable with help text. Move others without bars (e.g. EIL) to the bottom so the bars stay together.

lighthouse-core/report/v2/renderer/category-renderer.js
paulirish: hmm! maybe? can you explain more what this would look like?

lighthouse-core/report/v2/renderer/util.js
```diff
@@ -42,10 +42,11 @@ class Util {
   /**
    * Format number.
    * @param {number} number
+   * @param {number} decimalPlaces
    * @return {string}
    */
-  static formatNumber(number) {
-    return number.toLocaleString(undefined, {maximumFractionDigits: 1});
+  static formatNumber(number, decimalPlaces = 1) {
```
ebidel: nice

lighthouse-core/report/v2/renderer/util.js
```diff
@@ -42,10 +42,11 @@ class Util {
   /**
    * Format number.
    * @param {number} number
+   * @param {number} decimalPlaces
```
ebidel: `{number=} decimalPlaces Number of decimal places to include. Defaults to 1.`

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -336,6 +350,86 @@ class CategoryRenderer {
...
+        const img = this._dom.createChildOf(thumbnail, 'img');
+        img.src = 'data:image/jpg;base64,' + frameForTime;
+      }
+    }
+
+    for (const marker of markers) {
+      const markerElement = this._dom.createChildOf(timelineElement, 'div', 'lh-timeline-marker');
+      this._dom.createChildOf(markerElement, 'div', 'lh-timeline-bar').style.width =
+          (100 * (marker.value / timeSpan) | 0) + '%';
+      this._dom.createChildOf(markerElement, 'span').textContent = `${marker.title}: `;
```
patrickhulce: do you think it's worth trying to share any of this sort of logic with the sparkline elements? it's not a lot and they are rendered differently though, so I'm probably not for it, but throwing it out there

lighthouse-core/report/v2/report-styles.css
```diff
@@ -574,6 +574,64 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
...
+
+.lh-filmstrip .frame {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    padding: 4px;
+    flex: none;
+}
+
+.lh-filmstrip .frame .thumbnail {
```
patrickhulce: do we want to do BEM for these?

lighthouse-core/report/v2/report-generator.js
```diff
@@ -18,19 +18,24 @@
...
-  fs.readFileSync(__dirname + '/renderer/details-renderer.js', 'utf8'),
-  fs.readFileSync(__dirname + '/../../lib/file-namer.js', 'utf8'),
-  fs.readFileSync(__dirname + '/renderer/logger.js', 'utf8'),
-  fs.readFileSync(__dirname + '/renderer/report-ui-features.js', 'utf8'),
-  fs.readFileSync(__dirname + '/renderer/category-renderer.js', 'utf8'),
-  fs.readFileSync(__dirname + '/renderer/report-renderer.js', 'utf8'),
-].join(';\n');
-const REPORT_CSS = fs.readFileSync(__dirname + '/report-styles.css', 'utf8');
-const REPORT_TEMPLATES = fs.readFileSync(__dirname + '/templates.html', 'utf8');
+  readScript('/renderer/util.js'),
```
ebidel: might be nice to make these relative to it's obvious: `'./renderer/util.js'`

lighthouse-core/report/v2/report-styles.css
```diff
@@ -574,6 +574,64 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
...
+    display: flex;
+    width: 54px;
+    height: 100%;
+    flex-direction: row;
+    align-items: center;
+    margin: 4px 0 2px;
+    border: 2px solid transparent;
+    box-shadow: 0 0 3px #bbb, 2px 1px 1px -1px #7c80a0;
```
ebidel: I think you've got 4 space indents

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: Basically I'm proposing to move the logic that collects the metrics from pwmetrics-events.js into an audit instead.

audits/metrics.js
```js
const TTFI = require('./first-interactive')

class Metrics extends Audit {
  static audit(artifacts) {
    const traceOfTab = await artifacts.requestTraceOfTab()
    const fmp = await FMP.audit(artifacts)
    const firstInteractive = await TTFI.audit(artifacts)

    return {
      extendedInfo: {
        value: [
           {name: 'first-interactive', timeInMs: firstInteractive.extendedInfo.value.timings.firstInteractive, timestamp: ...},
           {name: 'on-load', timeInMs: traceOfTab.timings.onLoad, timestamp: ...},
           {name: 'trace-end', timeInMs: traceOfTab.timings.traceEnd, timestamp: ...},
           ...
        ]
      }
    }
  }
}
```

pwmetrics-events.js
```js
class Metrics {
  ...
  metricsDefinitions(auditResults) {
    return auditResults['metrics'].extendedInfo.value;
  }
}
```

lighthouse-cli/bin.ts
patrickhulce: I don't think we have tslint in our deps anywhere :)

lighthouse-core/report/v2/report-generator.js
patrickhulce: unrelated, this is kosher with browserify? I think we have some other comments and verbose lines to remove :)

lighthouse-cli/bin.ts
patrickhulce: oh this is actually orthogonal to speedline perf IMO, we should definitely do that too, but even if we didn't we can still leverage the SSIM to only return thumbnails that are substantively different

lighthouse-core/report/v2/report-generator.js
ebidel: Ignore this comment. It won't work.

paulirish:
superceded by #2239 and #2256

PR #2182, opened 2017-05-08
README: update CLI help with latest. Give common examples

ebidel:
- Provide usage examples in CLI `--help`
- Update README with latest output from `--help`
- Also changed `additionalFlags` -> `chromeFlags` so we're consistent. It's confusing that we use `--chrome-flags` in some places (LH CLI) and `additionalFlags` in other places (`chrome-debug`, `ChromeLauncher`). `--chrome-flags` better signals that these flags are Chrome specific and not Lighthouse flags.

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -51,13 +51,13 @@ const cliFlags = yargs
   .version(() => pkg.version)
   .showHelpOnFail(false, 'Specify --help for available options')
 
-  .usage('$0 url')
-  .example('$0 url --view', 'Opens the HTML report in a browser after the run completes')
-  .example('$0 url --config-path=./myconfig.js', 'Runs Lighthouse with your own configuration: custom audits, report generation, etc.')
-  .example('$0 url --output=json --output-path=./report.json --save-assets', 'Save trace, screenshots, and named JSON report.')
-  .example('$0 url --disable-device-emulation --disable-network-throttling', 'Disable device emulation')
-  .example('$0 url --chrome-flags="--window-size=412,732"', 'Launch Chrome with a specific window size')
-  .example('$0 url --quiet --chrome-flags="--headless"', 'Launch Headless Chrome, turn off logging')
+  .usage('lighthouse <url>')
+  .example('lighthouse <url> --view', 'Opens the HTML report in a browser after the run completes')
+  .example('lighthouse <url> --config-path=./myconfig.js', 'Runs Lighthouse with your own configuration: custom audits, report generation, etc.')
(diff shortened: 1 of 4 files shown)


paulirish:
it's super tall:
image


can you add this guy into the `yargs` chain?
```js
.wrap(yargs.terminalWidth())
```
image

lighthouse-cli/bin.ts
```diff
@@ -52,6 +52,12 @@ const cliFlags = yargs
   .showHelpOnFail(false, 'Specify --help for available options')
 
   .usage('$0 url')
+  .example('$0 url --view', 'Opens the HTML report in a browser after the run completes')
```
paulirish: IMO we should just use "lighthouse" rather than whatever $0 resolves to.

how about with brackets on the url too: `lighthouse <url> --view`

lighthouse-cli/bin.ts
```diff
@@ -209,7 +216,7 @@ function getDebuggableChrome(flags: {skipAutolaunch: boolean, port: number, sele
                                chromeFlags: string}): Promise<ChromeLauncher> {
   const chromeLauncher = new ChromeLauncher({
     port: flags.port,
-    additionalFlags: flags.chromeFlags.split(' '),
+    chromeFlags: flags.chromeFlags.split(' '),
```
paulirish: this is a breaking change and some other folks rely on the API, but @samccone's work here will break them too and hopefully have some docs so its easy to do the right thing.

all that said, +1 to this change.

readme.md
```diff
@@ -44,43 +44,78 @@ $ lighthouse --help
...
+  --disable-cpu-throttling      Disable CPU throttling[boolean] [default: false]
+  --disable-network-throttling  Disable network throttling             [boolean]
+  --skip-autolaunch             Skip autolaunch of Chrome when already running
+                                instance is not found                  [boolean]
+  --select-chrome               Interactively choose version of Chrome to use
+                                when multiple installations are found  [boolean]
+  --interactive                 Open Lighthouse in interactive mode    [boolean]
+
+Examples:
+  lighthouse-cli/index.js url --view        Opens the HTML report in a browser
```
paulirish: i'd repeat this paste with the new terminalWidth change, for less wrapping.

lighthouse-cli/bin.ts
ebidel: Yea, I figure we're covered under the 2.0 veil

readme.md
ebidel: done

lighthouse-cli/bin.ts
ebidel: done

ebidel:
PTAL

PR #2183, opened 2017-05-08
Plots: make measure script more flexible (CLI args)

wwwillchen:
Add flags, help docs, and making it better all around.


lighthouse-core/lib/emulation.js
```diff
@@ -41,10 +41,13 @@ const NEXUS5X_USERAGENT = {
     '(KHTML, like Gecko) Chrome/59.0.3033.0 Mobile Safari/537.36'
 };
 
+const LATENCY_FACTOR = 3.75;
+const THROUGHPUT_FACTOR = 1.1;
```
patrickhulce: nit: should we model both of these as multipliers so it's just multiply for both?

plots/measure.js
```diff
@@ -29,9 +30,47 @@ const ChromeLauncher = require('../lighthouse-cli/chrome-launcher.js').ChromeLau
...
+
+const FLAGS = {
+  output: 'json',
+  disableCpuThrottling: DISABLE_CPU_THROTTLING,
+  disableNetworkThrottling: DISABLE_NETWORK_THROTTLING,
+  disableDeviceEmulation: DISABLE_DEVICE_EMULATION,
+};
+console.log('Running lighthouse with flag\n disableNetworkThrottling: ', FLAGS.disableNetworkThrottling);
+
+const SUBSET = [
```
brendankenny: running plots/ this weekend, it was very handy to split this to loading from file so I could quickly switch between a couple of different sets of test sites. If you're adding `yargs` anyways, it would be great to support loading a list of URLs from a path passed in as a CLI option :)

plots/measure.js
```diff
@@ -150,20 +178,30 @@ main();
  * Returns a promise chain that analyzes all the sites n times.
  * @return {!Promise}
  */
-function runAnalysis() {
+function runAnalysisWithNewChromeInstances() {
```
brendankenny: 👍

lighthouse-core/lib/emulation.js
paulirish: +1.

it's equal to `throughput * 90/99` but you might as well simplify it to `throughput * 0.9`

brendankenny:
is this ready to review? And will need a rebase with recent changes to chromeLauncher

brendankenny:
> is this ready to review?

was that a yes? :)

wwwillchen:
ping @brendankenny for a review :)

brendankenny:
will need a rebase :)

plots/measure.js
```diff
@@ -119,46 +167,112 @@ const URLS = [
...
     return;
   }
 
-  const launcher = ChromeLauncher.launch();
-
-  return launcher.then(() => runAnalysis())
-  .catch(err => console.error(err))
-  .then(() => launcher.kill());
+  if (REUSE_CHROME) {
+    ChromeLauncher.launch({port: 9222}).then(launcher => {
```
paulirish: I've been running multiple plots at a time recently.. Launcher is getting a unique port by default, but you have to get it off `launcher.port` and pass into `lighthouse()`.. Here's what I did today: 
image


Can you support this functionality?  You could just pass it along like this or create measure instances that hold this state.. wdyt?

plots/measure.js
```diff
@@ -31,9 +59,29 @@ const ChromeLauncher = require('../chrome-launcher/chrome-launcher.js');
...
+const SITES_PATH = args['sites-path'];
+const SUBSET = args['subset'];
+const SITE = args['site'];
+const CUSTOM_NUMBER_OF_RUNS = args['n'];
 
-const URLS = [
+// Running it n + 1 times if the first run is deliberately ignored
+// because it has different perf characteristics from subsequent runs
+// (e.g. DNS cache which can't be easily reset between runs)
+const NUMBER_OF_RUNS = (CUSTOM_NUMBER_OF_RUNS || 20) + (KEEP_FIRST_RUN ? 0 : 1);
```
paulirish: doesn't seem like an uppercase const as its super conditional.

plots/measure.js
```diff
@@ -31,9 +59,29 @@ const ChromeLauncher = require('../chrome-launcher/chrome-launcher.js');
 const Printer = require('../lighthouse-cli/printer');
 const assetSaver = require('../lighthouse-core/lib/asset-saver.js');
 
-const NUMBER_OF_RUNS = 20;
+const DISABLE_DEVICE_EMULATION = args['disable-device-emulation'];
```
paulirish: tbh seems kinda odd to establish these all as uppercase consts as they are dependent on CLI input.

plots/measure.js
```diff
@@ -119,46 +167,112 @@ const URLS = [
...
+  if (SITES_PATH) {
+    return require(path.resolve(__dirname, SITES_PATH));
+  }
+
+  if (SITE) {
+    return [SITE];
+  }
+
+  if (SUBSET) {
+    return [
```
paulirish: how about we just slice the top of the full sites list instead of repeating here?

plots/measure.js
paulirish: it's unexpected that number_of_runs == 1 will provide two runs for each site. Can we make sure the number here will match # of actual runs done?

plots/utils.js
```diff
@@ -52,8 +53,34 @@ function shuffle(array) {
   }
 }
 
+/**
+ * @param {string} filePath
+ */
+function removeRecursive(filePath) {
```
paulirish: unused?

plots/measure.js
```diff
@@ -119,47 +147,117 @@ const URLS = [
...
 ];
 
-/**
- * Launches Chrome once at the beginning, runs all the analysis,
- * and then kills Chrome.
- * TODO(chenwilliam): measure the overhead of starting chrome, if it's minimal
- * then open a fresh Chrome instance for each run.
- */
+function getUrls() {
+  if (args['sites-path']) {
```
paulirish: since we already have this, why not toss the default SITES into a file and have sites-path default to that one.

plots/measure.js
```diff
@@ -119,47 +147,117 @@ const URLS = [
...
+    return require(path.resolve(__dirname, args['sites-path']));
+  }
+
+  if (args['site']) {
+    return [args['site']];
+  }
+
+  if (args['subset']) {
+    return [
```
paulirish: and it seems like we could just extract this into a diff sites_subset.js file?

plots/measure.js
```diff
@@ -182,20 +281,27 @@ function singleRunAnalysis(url, id, {ignoreRun}) {
...
  * @param {string} url
  * @param {string} outputPath
  * @param {string} assetsPath
  * @param {{ignoreRun: boolean}} options
  * @return {!Promise}
  */
-function analyzeWithLighthouse(url, outputPath, assetsPath, {ignoreRun}) {
-  const flags = {output: 'json'};
+function analyzeWithLighthouse(launcher, url, outputPath, assetsPath, {ignoreRun}) {
+  const flags = {
```
paulirish: if ignoreRun is true then there's no reason to have throttling on. I'd force disable cpu & network throttling in these cases.

wwwillchen: done

plots/sites.js
```diff
@@ -0,0 +1,103 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/* eslint-disable no-unused-expressions */
+[
```
paulirish: `module.exports = [` ...

plots/sites_subset.js
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+/* eslint-disable no-unused-expressions */
+[
```
paulirish: `module.exports = [`  ...

plots/measure.js
```diff
@@ -31,135 +59,107 @@ const ChromeLauncher = require('../chrome-launcher/chrome-launcher.js');
...
-  'https://fc2.com',
-  'https://www.pixnet.net/',
-  'https://dropbox.com',
-  'https://zhihu.com',
-  'https://whatsapp.com',
-  'https://alibaba.com',
-  'https://ask.com',
-  'https://bbc.com'
-];
+const keepFirstRun = args['keep-first-run'] || !args['reuse-chrome'];
```
paulirish: you can use camelcase instead of this square bracket dance.  

`args.keepFirstRun` etc. yargs does it for you.

plots/measure.js
```diff
@@ -31,135 +59,107 @@ const ChromeLauncher = require('../chrome-launcher/chrome-launcher.js');
...
+    } else {
+      throw new Error('Must pass a js or json file to --sites-path');
+    }
+  }
+
+  if (args['subset']) {
+    return eval(fs.readFileSync(path.resolve(__dirname, 'sites_subset.js'), 'utf-8'));
+  }
+
+  return eval(fs.readFileSync(path.resolve(__dirname, 'sites.js'), 'utf-8'));
```
paulirish: instead. you can set the default on yargs:

```js
.default('sites-path', 'sites.js')
```

plots/measure.js
```diff
@@ -31,135 +59,107 @@ const ChromeLauncher = require('../chrome-launcher/chrome-launcher.js');
...
+const keepFirstRun = args['keep-first-run'] || !args['reuse-chrome'];
+const numberOfRuns = args['n'] || 3;
+
+function getUrls() {
+  if (args['site']) {
+    return [args['site']];
+  }
+
+  if (args['sites-path']) {
+    const sitesPath = path.resolve(__dirname, args['sites-path']);
```
paulirish:
```js
if (args.sitesPath) {
  return require(path.resolve(__dirname, args.sitesPath));
}
```

plots/sites.js
wwwillchen: done

plots/measure.js
wwwillchen: Yeah - this is nice. I did it the more verbose way because the downside is require() will hold onto these files indefinitely in node's module cache. It's definitely OK now since the sites list are needed for the entire lifetime of the script but if we start loading a lot of different sites files, we'll need to watch out for heap usage.

wwwillchen:
@paulirish - please take another look.

plots/plot-per-site.html
```diff
@@ -0,0 +1,29 @@
...
+-->
+
+<!doctype html>
+<html>
+
+<head>
+  <script src="https://cdn.plot.ly/plotly-1.25.2.min.js"></script>
+</head>
+
+<body>
```
paulirish: add this page to the `<nav>` of the other two html files?

plots/plot-per-site.html
```diff
@@ -0,0 +1,29 @@
+<!--
```
paulirish: rename this to bars-per-site ?

plots/measure.js
```diff
@@ -17,11 +17,40 @@
 'use strict';
 
 /* eslint-disable no-console */
-
+const fs = require('fs');
 const path = require('path');
 const parseURL = require('url').parse;
 
 const mkdirp = require('mkdirp');
+const args = require('yargs')
```
paulirish: with the new style of invoking, can you update the readme?

to me, it makes more sense to use the `node script.js` invocation rather than the `npm run` one now that we have a CLI. but no big deal.

plots/plot-per-site.html
wwwillchen: done

plots/README.md
```diff
@@ -14,17 +14,13 @@ You need to build lighthouse first.
 ### Generating & viewing charts
 
 ```
-# View all commands
-$ cd plots
-$ yarn run
-
 # Run lighthouse to collect metrics data
-$ yarn measure
+$ node measure.js
```
paulirish: you got a whole CLI for measure.js now!. at the very least you should paste the `--help` output into the readme

PR #2184, opened 2017-05-08
Lighthouse Canary

paulirish:
We've shipped a few releases from master that have the latest and greatest.

* extension: https://chrome.google.com/webstore/detail/lighthouse-canary/gcladjiblfghhokpimgadfdmblkocoae
* node/cli:  `npm install -g lighthouse@canary`  which will install `2.0.0-alpha.3`

The changes here were made before shipping.

Which of these do we want to bring to master?

my POV: we should have all these changes except the extension changing its name & icon.


lighthouse-cli/bin.ts
```diff
@@ -134,7 +134,7 @@ Example: --output-path=./lighthouse-results.html`,
   // default values
   .default('chrome-flags', '')
   .default('disable-cpu-throttling', false)
-  .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.html])
+  .default('output', Printer.GetValidOutputOptions()[Printer.OutputMode.domhtml])
```
patrickhulce: 👍  when are we ready to knock off html entirely?

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -168,14 +169,13 @@ window.runLighthouseInWorker = function(port, url, options, categoryIDs) {
...
+window.createReportPageAsBlob = function(results) {
   performance.mark('report-start');
-  const reportGenerator = new ReportGenerator();
   let html;
   try {
-    html = reportGenerator.generateHTML(results, reportContext);
+    html = new ReportGeneratorV2().generateReportHtml(results);
   } catch (err) {
-    html = reportGenerator.renderException(err, results);
+    html = new ReportGenerator().renderException(err, results);
```
brendankenny: can probably ditch this too (like in the CLI) since an exception in v2 generation is extremely unlikely

lighthouse-extension/app/manifest.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "__MSG_appName__",
-  "version": "1.6.0",
+  "version": "2.0.0.1",
```
brendankenny: probably best to ditch this change as it might make releasing somewhat weird (unless we don't start with 2.0.0?)

lighthouse-extension/app/_locales/en/messages.json
```diff
@@ -1,10 +1,10 @@
 {
   "appName": {
-    "message": "Lighthouse",
-    "description": "Helping you avoid Progressive Web App woes."
+    "message": "Lighthouse Canary",
```
ebidel: yea, keep existing name

lighthouse-extension/app/manifest.json
```diff
@@ -1,12 +1,12 @@
...
   "minimum_chrome_version": "56",
   "manifest_version": 2,
   "description": "__MSG_appDescription__",
   "icons": {
-    "16": "images/lh_logo_bg.png",
-    "128": "images/lh_logo_bg.png"
+    "16": "images/lh_logo_canary_bg.png",
```
ebidel: keep existing icons

lighthouse-extension/app/manifest.json
paulirish: CWS doesnt allow text in the version names, so i can't do "alpha". I was hesitant to number it as 1.x, though.

what do you suggest? 

i can bring it back to 1.99.x but i'll have to get a new extension ID. (CWS wont let you go backwards)

lighthouse-extension/app/_locales/en/messages.json
paulirish: done. added separate file for these

lighthouse-extension/app/manifest.json
paulirish: done. added separate file for these

lighthouse-extension/app/src/lighthouse-background.js
paulirish: k done.

lighthouse-cli/bin.ts
paulirish: in late may let's do a two-day sprint of just deletions. :)

paulirish:
updated.

instead of modifying the extension name/icons i created a duplicate file with those edits. that way we don't have to hop back and forth between branches to do this. :)

lighthouse-extension/app/manifest.json
ebidel: Yea CWS is annoying like this. I've done date releases in the past: `2019-05-09`.

The crx version doesn't have to correspond to the LH version. We have the latter in the report.

ebidel: Anyway, this version doesn't really matter in the grand scheme of things. It's just for auto updating.

brendankenny: > CWS doesnt allow text in the version names, so i can't do "alpha". I was hesitant to number it as 1.x, though.
>
> what do you suggest?

I'm good with it now that it's in the canary manifest. I was just wanting to avoid having to skip 2.0.0 and ship 2.0.1 to LH stable

paulirish: sg guys.

i like dates on the releases ill try that out.

PR #2085, opened 2017-04-27
Remove images from critical request chains.

abacon:
Fixes https://github.com/GoogleChrome/lighthouse/issues/1550. Hope this is still useful!

PR #2191, opened 2017-05-09
Update: Use array of non-critical resource types

stramel:
@paulirish Follow up PR for #2085

Figured I would add it since I suggested and had time now

### DIFF
--- a/lighthouse-core/gather/computed/critical-request-chains.js
+++ b/lighthouse-core/gather/computed/critical-request-chains.js
@@ -43,7 +43,7 @@ class CriticalRequestChains extends ComputedArtifact {
     if (nonCriticalResourceTypes.includes(resourceTypeCategory) {
       return false;
     }
- 
+
     // Treat favicons as non-critical resources
     if (request.mimeType === 'image/x-icon' ||
         (request.parsedURL && request.parsedURL.lastPathComponent === 'favicon.ico')) {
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/computed/critical-request-chains.js
```diff
@@ -34,12 +34,13 @@ class CriticalRequestChains extends ComputedArtifact {
    */
   isCritical(request) {
     // XHRs are fetched at High priority, but we exclude them, as they are unlikely to be critical
```
paulirish: the comments relate to the `nonCriticalResourceTypes` so they shoudl probably be right on top of it.

lighthouse-core/gather/computed/critical-request-chains.js
```diff
@@ -34,12 +34,13 @@ class CriticalRequestChains extends ComputedArtifact {
...
-      return false;
-    }
     // Images are also non-critical.
-    if (resourceTypeCategory === WebInspector.resourceTypes.Image._category) {
+    const resourceTypeCategory = request._resourceType && request._resourceType._category;
+    const nonCriticalResourceTypes = [
+      WebInspector.resourceTypes.Image._category,
+      WebInspector.resourceTypes.XHR._category
+    ];
+    if (nonCriticalResourceTypes.includes(resourceTypeCategory)) {
```
paulirish: the `request._resourceType && ` conflicts with this. 

one solution: you can do an early `return false` if `_resourceType` is falsy.

paulirish: ah nevermind. at worst it'll be `[string,string].includes(undefined)`  which will just fail, but not throw.

no change necessary for this.

paulirish:
appreciate the assist on that one!

Issue #2192, opened 2017-05-09
TTFB Audit followup


paulirish:
Forking from #2126 

@wardpeet added:

> I had a few changes ready to go, was fixing the details export of audit and cleaning up the criticalRequestChains but I can open an extra PR just for that 😄 

i am also seeing this in the new report
![image](https://cloud.githubusercontent.com/assets/39191/25870050/89584f40-34b7-11e7-8a2f-72f916857691.png)

and this in the old one:
![image](https://cloud.githubusercontent.com/assets/39191/25870064/9a76b406-34b7-11e7-937c-0c569b670ba7.png)

ebidel:
Is "details export" adding a details rendering to the audit? FYI, I'm working on a criticalRequestChains details renderer.

paulirish:
i _think_ so, but cc @wardpeet to be sure.

it'll just be a basic `table` renderer.

> i'm working on a criticalRequestChains details renderer.

awesome.

paulirish:
Additional for @wardpeet: rename `time-to-firstbyte` to `time-to-first-byte`

wardpeet:
I was looking into the gatherer as @patrickhulce mentioned:
> I was thinking just expose as a flattened property of the existing computed artifact so we don't bother with the computation twice although if you're feeling fancy we could refactor the existing one to call requestCriticalRequests and then build the tree :)

wardpeet:
Just need a rebase and #2231 is ready.

Some records don't have _timing so I ignore them

brendankenny:
closed by #2231

PR #2194, opened 2017-05-09
Disable TTFB audit temporarily for stability.

paulirish:
As it says on the tin
cc #2192

---

also including a driveby fix for this:
image


paulirish:
> there wasn't a smokehouse expectation to undo too right?

there may have been...

paulirish:
>> there wasn't a smokehouse expectation to undo too right?

> there may have been...

nope we're good.

PR #2197, opened 2017-05-09
refactor: makes non-finished network records available

patrickhulce:
split from https://github.com/GoogleChrome/lighthouse/pull/2023#discussion_r114909191

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/total-byte-weight.js
+++ b/lighthouse-core/audits/byte-efficiency/total-byte-weight.js
@@ -61,7 +61,7 @@ class TotalByteWeight extends ByteEfficiencyAudit {
         let results = [];
         networkRecords.forEach(record => {
           // exclude data URIs since their size is reflected in other resources
-          if (record.scheme === 'data') return;
+          if (record.scheme === 'data' || !record.finished) return;
 
           const result = {
             url: URL.getDisplayName(record.url),
--- a/lighthouse-core/audits/byte-efficiency/total-byte-weight.js
+++ b/lighthouse-core/audits/byte-efficiency/total-byte-weight.js
@@ -61,6 +61,7 @@ class TotalByteWeight extends ByteEfficiencyAudit {
         let results = [];
         networkRecords.forEach(record => {
           // exclude data URIs since their size is reflected in other resources
+          // exclude unfinished requests since they won't have transfer size information
           if (record.scheme === 'data' || !record.finished) return;
 
           const result = {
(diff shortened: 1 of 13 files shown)


brendankenny:
> available internally

Can you clarify "internally"? I don't see the significance

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
```diff
@@ -61,7 +61,7 @@ class TotalByteWeight extends ByteEfficiencyAudit {
         let results = [];
         networkRecords.forEach(record => {
           // exclude data URIs since their size is reflected in other resources
-          if (record.scheme === 'data') return;
+          if (record.scheme === 'data' || !record.finished) return;
```
brendankenny: update comment

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -56,7 +56,7 @@ class LoadFastEnough4Pwa extends Audit {
         // Ignore requests that don't have timing data or resources that have
         // previously been requested and are coming from the cache.
         const fromCache = record._fromDiskCache || record._fromMemoryCache;
-        if (!record._timing || fromCache) {
+        if (!record._timing || fromCache || !record.finished) {
```
brendankenny: update comment (and it seems like it may make more sense to put `!record.finished` first)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -73,7 +73,7 @@ class OptimizedImages extends Gatherer {
   static filterImageRequests(pageUrl, networkRecords) {
     const seenUrls = new Set();
     return networkRecords.reduce((prev, record) => {
-      if (seenUrls.has(record._url)) {
+      if (seenUrls.has(record._url) || !record.finished) {
```
brendankenny: nit: does it make more sense to reverse conditions?

brendankenny:
travis is super slow today

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
patrickhulce: done

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: done

patrickhulce:
comments updated

> it probably makes sense to move the record.finished check first as it's the more
(most?) fundamental differentiator for "this audit isn't for this kind of request"

I sort of disagree, the fact that this is the most obvious and requires the least thought makes it ok to put it in the afterthought position and devote the most attention/focus to the domain-specific exclusions which come first

patrickhulce:
> Can you clarify "internally"? I don't see the significance

it was a minor point signifying that we still filter out unfinished records almost everywhere so it's mostly an internal change

brendankenny:
> the fact that this is the most obvious and requires the least thought 

Well that's what I mean. Really they could be filtered out before entering into these loops (a la critical request chains)  or removed with their own conditional check, but people around here don't seem to like multiple single-line conditionals in a row :)

Anyway, it's good as is.

Issue #2201, opened 2017-05-10
Flaky test: link-blocking-first-paint results === 4


paulirish:
[This assertion of only 4 blocking stylesheets](https://github.com/GoogleChrome/lighthouse/blob/d0e0d35d7d8bceaedcb5925da5f9345e5324cb37/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js#L62) on dbw_tester is failing sometimes.

Travis likes to think there are 5.

It includes [this guy](https://github.com/GoogleChrome/lighthouse/blob/d0e0d35d7d8bceaedcb5925da5f9345e5324cb37/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html#L40-L41):

```html
<link rel="preload" href="./dbw_tester.css?delay=2000&async=true" as="style" onload="this.rel = 'stylesheet'">
```

probably the new link preload logic?

example travis runs:  
* https://travis-ci.org/GoogleChrome/lighthouse/jobs/230589047#L537
* https://travis-ci.org/GoogleChrome/lighthouse/jobs/230580380#L545

patrickhulce:
well that really sucks, I can't repro this locally to investigate either :/

patrickhulce:
wait, is travis using an old cache has anyone checked? it might have an old canary that didn't support isLinkPreload

patrickhulce:
clearing the caches has gotten #2204 to pass 4 of 4 rebuilds (8 of 8 jobs) so I think we're good here 👍

PR #2202, opened 2017-05-10
tests: exclude score checking and flaky link preload assertion.

paulirish:
make master green again

ref #2201


brendankenny:
why is this flaky when it wasn't before?

paulirish:
@brendankenny the link new `isLinkPreload ` stuff in https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js is _probably_ at the root here. 

my guess is its  racy and fails 20% of the time. and the previous travis builds for it were not  running enough.

paulirish:
and fwiw the `&isdisabled` bit is just so its easier to match up which requests are what on the smokehouse output side. :)

PR #2203, opened 2017-05-10
Launcher type improvements.

samccone:
3 fixes in 3 commits, how about that.

Switch to default argument assignment.
Since we now only support node 6+ we can drop this ugly fallback.

--

Remove lossey type cast from chrome launcher.

--

Add type safety to defaults.  …
Since we now have generics in typescript, we can leverage them in our
default function to have type safety.

ref #2092

### DIFF
--- a/lighthouse-cli/chrome-launcher.ts
+++ b/lighthouse-cli/chrome-launcher.ts
@@ -32,9 +32,9 @@ const execSync = childProcess.execSync;
 const isWindows = process.platform === 'win32';
 
 export class ChromeLauncher {
-  prepared: Boolean = false;
+  prepared = false;
   pollInterval: number = 500;
-  autoSelectChrome: Boolean;
+  autoSelectChrome: boolean;
   TMP_PROFILE_DIR: string;
   outFile?: number;
   errFile?: number;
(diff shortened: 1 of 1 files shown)


lighthouse-cli/chrome-launcher.ts
```diff
@@ -44,20 +44,17 @@ export class ChromeLauncher {
...
-  }) {
-    opts = opts || {};
-
+  } = {}) {
     // choose the first one (default)
-    this.autoSelectChrome = defaults(opts.autoSelectChrome, true);
-    this.startingUrl = defaults(opts.startingUrl, 'about:blank');
-    this.chromeFlags = defaults(opts.chromeFlags, []);
-    this.port = defaults(opts.port, 9222);
+    this.autoSelectChrome = defaults<Boolean>(opts.autoSelectChrome, true);
```
paulirish: why is this casting necessary if these types are defined in the method signature?

is it possible to avoid providing the type twice?

lighthouse-cli/chrome-launcher.ts
```diff
@@ -141,7 +138,7 @@ export class ChromeLauncher {
         .then(execPath => this.spawn(execPath));
   }
 
-  spawn(execPath: string): Promise<any[]> {
+  spawn(execPath: string) {
```
paulirish: noob question: why can we get rid of the return type?

samccone: the return type can be 100% inferred 👍

lighthouse-cli/chrome-launcher.ts
paulirish: ya much better.

PR #2206, opened 2017-05-10
Add DomNode renderer

paulirish:
Adding a new node/element renderer for the report. Primarily assisting all the a11y audits right now.

~I've also added a `--a11y` CLI flag because 1) it was super easy and 2) it was really handy. But I can revert, no big deal.~

This also enables node revealing/hovering in devtools.  The WIP DevTools-side patch:
https://codereview.chromium.org/2849463002/diff/1/third_party/WebKit/Source/devtools/front_end/audits2/Audits2Panel.js

It looks like this in the standalone report:
image

### DIFF
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -30,6 +30,7 @@ const lighthouse = require('../lighthouse-core');
 const log = require('../lighthouse-core/lib/log');
 import * as path from 'path';
 const perfOnlyConfig = require('../lighthouse-core/config/perf.json');
+const a11yOnlyConfig = require('../lighthouse-core/config/a11y.json');
 const performanceXServer = require('./performance-experiment/server');
 import * as Printer from './printer';
 import * as randomPort from './random-port';
@@ -66,6 +67,8 @@ if (cliFlags.configPath) {
   config = require(cliFlags.configPath);
 } else if (cliFlags.perf) {
   config = perfOnlyConfig;
+} else if (cliFlags.a11y) {
+  config = a11yOnlyConfig;
 }
 
 // set logging preferences
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -30,7 +30,6 @@ const lighthouse = require('../lighthouse-core');
 const log = require('../lighthouse-core/lib/log');
 import * as path from 'path';
 const perfOnlyConfig = require('../lighthouse-core/config/perf.json');
-const a11yOnlyConfig = require('../lighthouse-core/config/a11y.json');
 const performanceXServer = require('./performance-experiment/server');
 import * as Printer from './printer';
 import * as randomPort from './random-port';
@@ -67,8 +66,6 @@ if (cliFlags.configPath) {
   config = require(cliFlags.configPath);
 } else if (cliFlags.perf) {
   config = perfOnlyConfig;
-} else if (cliFlags.a11y) {
-  config = a11yOnlyConfig;
 }
 
 // set logging preferences
--- a/lighthouse-cli/cli-flags.ts
+++ b/lighthouse-cli/cli-flags.ts
@@ -82,6 +82,7 @@ export function getFlags() {
         'chrome-flags':
             'Custom flags to pass to Chrome (space-delimited). For a full list of flags, see http://peter.sh/experiments/chromium-command-line-switches/.',
         'perf': 'Use a performance-test-only configuration',
+        'a11y': 'Use an accessibility-test-only configuration',
         'port': 'The port to use for the debugging protocol. Use 0 for a random port',
         'max-wait-for-load':
             'The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING: Very high values can lead to large traces and instability',
@@ -106,8 +107,8 @@ Example: --output-path=./lighthouse-results.html`,
       .boolean([
         'disable-storage-reset', 'disable-device-emulation', 'disable-cpu-throttling',
         'disable-network-throttling', 'save-assets', 'save-artifacts', 'list-all-audits',
-        'list-trace-categories', 'perf', 'view', 'skip-autolaunch', 'select-chrome', 'verbose',
-        'quiet', 'help', 'interactive'
+        'list-trace-categories', 'perf', 'a11y', 'view', 'skip-autolaunch', 'select-chrome',
+        'verbose', 'quiet', 'help', 'interactive'
       ])
       .choices('output', GetValidOutputOptions())
 
--- a/lighthouse-cli/cli-flags.ts
+++ b/lighthouse-cli/cli-flags.ts
@@ -82,7 +82,6 @@ export function getFlags() {
         'chrome-flags':
             'Custom flags to pass to Chrome (space-delimited). For a full list of flags, see http://peter.sh/experiments/chromium-command-line-switches/.',
         'perf': 'Use a performance-test-only configuration',
-        'a11y': 'Use an accessibility-test-only configuration',
         'port': 'The port to use for the debugging protocol. Use 0 for a random port',
         'max-wait-for-load':
             'The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING: Very high values can lead to large traces and instability',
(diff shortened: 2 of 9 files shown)


lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+            node.textContent.trim().length === 0) continue;
+        index++;
+      }
+      return index;
+    }
+
+    const path = [];
+    while (node && node.parentNode) {
+      const index = getNodeIndex(node);
+      path.push([index, node.nodeName]);
```
patrickhulce: is this pavel's special node selector methodology? what would be the downside to augmenting a regular CSS selector with `nth-child`?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -141,6 +145,21 @@ class DetailsRenderer {
   }
 
   /**
+   * @param {!DetailsRenderer.NodeDetailsJSON} item
+   * @return {!Element}
+   */
+  _renderNode(item) {
```
patrickhulce: node is slightly ambiguous at this level, can we make it DOMNode or something? when I just looked at the title of the PR I thought it was a renderer running in "node.js" 😆

lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -49,6 +49,14 @@ class DOM {
   }
 
  /**
+   * @param {string} text
+   * @return {!Node}
+   */
+  createText(text) {
```
patrickhulce: this doesn't seem used?

lighthouse-cli/bin.ts
```diff
@@ -66,6 +67,8 @@ if (cliFlags.configPath) {
   config = require(cliFlags.configPath);
 } else if (cliFlags.perf) {
   config = perfOnlyConfig;
+} else if (cliFlags.a11y) {
```
patrickhulce: only downside here is that people might think you could do `lighthouse --perf --a11y` and get both, what if we supported that by nixing the config files and just appending each category that's enabled to `onlyCategories` of a default extending config

I can pickup that separately though

lighthouse-cli/cli-flags.ts
```diff
@@ -82,6 +82,7 @@ export function getFlags() {
         'chrome-flags':
             'Custom flags to pass to Chrome (space-delimited). For a full list of flags, see http://peter.sh/experiments/chromium-command-line-switches/.',
         'perf': 'Use a performance-test-only configuration',
+        'a11y': 'Use an accessibility-test-only configuration',
```
ebidel: if you keep the flag, please update the readme.

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+    // Augment the node objects with outerHTML snippet & custom path string
+    axeResult.violations.forEach(v => v.nodes.forEach(node => {
+      node.path = getNodePath(node.element);
+      node.snippet = getOuterHTMLSnippet(node.element);
+    }));
+
+    return axeResult;
   });
+
+  // Adapated from DevTools' SDK.DOMNode.prototype.path
```
ebidel: Adapted

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+  //   https://github.com/ChromeDevTools/devtools-frontend/blob/7a2e162ddefd/front_end/sdk/DOMModel.js#L530-L552
+  // TODO: Doesn't handle frames or shadow roots...
```
ebidel: Can you file a bug on this? We have some other "handle shadow root" updates to make in the codebase. We can address them all after io.

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+  function getNodePath(node) {
+    /* global Node */
```
ebidel: prefer this at the top.

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+    function getNodeIndex(node) {
+      let index = 0;
+      while (node = node.previousSibling) {
+        // skip empty text nodes
+        if (node.nodeType === Node.TEXT_NODE &&
+            node.textContent.trim().length === 0) continue;
```
ebidel: 2 space indent

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -43,7 +44,47 @@ function runA11yChecks() {
...
+    while (node && node.parentNode) {
+      const index = getNodeIndex(node);
+      path.push([index, node.nodeName]);
+      node = node.parentNode;
+    }
+    path.reverse();
+    return path.join(',');
+  }
+
+  function getOuterHTMLSnippet(node) {
```
ebidel: docs

lighthouse-core/gather/gatherers/accessibility.js
paulirish: yeah [in the protocol docs](https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-pushNodeByPathToFrontend) it says "proprietary format". :)

impl's: [backend side](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/inspector/InspectorDOMAgent.cpp?type=cs&q=pushNodeByPathToFrontend&sq=package:chromium&l=2063),[frontend side](https://github.com/ChromeDevTools/devtools-frontend/blob/7a2e162ddefd/front_end/sdk/DOMModel.js#L530-L552)

downsides of using selectors:

1. can't reference text nodes, document fragments, or the doctype node
1. can't pierce shadow dom

lighthouse-core/report/v2/renderer/dom.js
paulirish: true.

lighthouse-cli/bin.ts
paulirish: heh true. ill probably kick this out of the PR. :)

lighthouse-cli/cli-flags.ts
paulirish: ill probably nuke, but thx!

lighthouse-core/gather/gatherers/accessibility.js
paulirish: will move to the top top.

look like `global`  and `env` are file-wide so we couldnt have it be scoped to a function anyway. bummer.

paulirish:
I'm getting a ": Object reference chain is too long" error on some pages:
image

It's due to an object being > 1000 objects deep.

I imagine it's something that happens to the Axe result object on big pages (e.g. https://developers.google.com/web/tools/chrome-devtools/ )  but it's not clear. 

cc @robdodson

lighthouse-core/gather/gatherers/accessibility.js
robdodson: I think there's also work ongoing in axe to support shadow dom. @alice may know more.

robdodson:
Hm I'm not able to reproduce this error. Do you have some specific steps I could follow? I tried running axe extension, axe cli, and audits against the link you provided and they all work fine for me.

lighthouse-core/gather/gatherers/accessibility.js
alice: I don't have any updates here, sorry. I know Marcy was thinking of taking a look at it but she might have gotten swamped with other things.

paulirish:
> Hm I'm not able to reproduce this error. Do you have some specific steps I could follow?

I'd say try it with this patch? it's possible the new `elemRef` option is what's triggering it.

dylanb:
@paulirish I was also not able to reproduce this. I tried using an extension and I tried by simply going to the page, using the console to execute the axe-core 2.2.0 Javascript and then running `axe.run` - is there some particular environment I need to reproduce this?

paulirish:
> I'm getting a ": Object reference chain is too long" error on some pages:

I'm pretty sure I managed to squash all these.
I pretty much neuter the return object from axe, in order to verify there are no circular references within it. https://github.com/GoogleChrome/lighthouse/pull/2206/commits/4b00b65a08f72b0f636ef769c861da11404b4015

thank you guys for testing this and trying to give it a shot.

paulirish:
@dylanb i don't think it'll be a problem for us anymore, but ultimately the challenge was that we wanted `JSON.stringify(axeResultObject)` to succeed and it was running into circular JSON problems. It doesn't necessarily have to be a requirement that the axeResultObject is serializable, but it might help other consumers of the data.
Cheers!

lighthouse-core/gather/gatherers/accessibility.js
patrickhulce: filed #2243

paulirish:
@patrickhulce this is ready for another look right?

patrickhulce:
yep tests passing and ready for review

dylanb:
@paulirish It was in fact a bug, it has been fixed. Upgrade to 2.2.1 to get the fix https://github.com/dequelabs/axe-core/pull/334

PR #2209, opened 2017-05-10
fix: remove unconverted markdown in byte-efficiency details

patrickhulce:
The only remaining instance of unrendered markdown I saw that wasn't ticks was the italic items in byte efficiency audits table, did someone already complete this?

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -49,11 +49,14 @@ class UnusedBytes extends Audit {
   }
 
   /**
+   * @param {number} bytes
    * @param {number} percent
    * @return {string}
    */
-  static toSavingsString(percent = 0) {
-    return Math.round(percent).toLocaleString() + '%';
+  static toSavingsString(bytes = 0, percent = 0) {
+    const kbDisplay = this.bytesToKbString(bytes);
+    const percentDisplay = Math.round(percent).toLocaleString() + '%';
+    return `${kbDisplay} (${percentDisplay})`;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -49,14 +49,11 @@ class UnusedBytes extends Audit {
...
-   * @param {number} bytes
    * @param {number} percent
    * @return {string}
    */
-  static toSavingsString(bytes = 0, percent = 0) {
-    const kbDisplay = this.bytesToKbString(bytes);
-    const percentDisplay = Math.round(percent).toLocaleString() + '%';
-    return `${kbDisplay} _${percentDisplay}_`;
+  static toSavingsString(percent = 0) {
+    return Math.round(percent).toLocaleString() + '%';;
```
ebidel: ;;

ebidel:
For optimize images, it combines the absolute and % savings for webp and jpeg. Do we not want to show wasted bytes anymore?

screen shot 2017-05-10 at 10 03 39 am

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
patrickhulce: done

patrickhulce:
> For optimize images, it combines the absolute and % savings for webp and jpeg. Do we not want to show wasted bytes anymore?

That specific audit is about to be split in a PR today, but the table gets quite crowded when displaying as separate columns, I'm in favor of just communicating one of the values until we have more robust recursive rendering in table cells for a `block` item or something

ebidel:
You could just keep the absolute + percentage concat for now. I kinda like to see the absolute value of what the file would be. If it's not precise, then we should remove.

patrickhulce:
alright then just changed `_` to parens

PR #2210, opened 2017-05-10
feat: table styles

patrickhulce:
Before
image

After
image

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -95,6 +95,7 @@ class Audit {
   static makeV2TableHeaders(headings) {
     return headings.map(heading => ({
       type: 'text',
+      itemType: heading.itemType,
       text: heading.text
     }));
   }
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -81,7 +81,7 @@ class Audit {
 
         return {
           type: heading.itemType,
-          styleAs: heading.styleAs,
+          styleClass: heading.styleClass,
           text: value
         };
       });
@@ -96,7 +96,7 @@ class Audit {
   static makeV2TableHeaders(headings) {
     return headings.map(heading => ({
       type: 'text',
-      styleAs: heading.styleAs,
+      styleClass: heading.styleClass,
       text: heading.text
     }));
   }
@@ -164,6 +164,7 @@ module.exports = Audit;
  *   key: string,
  *   itemType: string,
  *   text: string,
+ *   styleClass: (string|undefined),
  * }>}
  */
 Audit.Headings; // eslint-disable-line no-unused-expressions
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -81,7 +81,6 @@ class Audit {
 
         return {
           type: heading.itemType,
-          styleAs: heading.styleAs,
           text: value
         };
       });
@@ -96,7 +95,6 @@ class Audit {
   static makeV2TableHeaders(headings) {
     return headings.map(heading => ({
       type: 'text',
-      styleAs: heading.styleAs,
       text: heading.text
     }));
   }
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -81,6 +81,7 @@ class Audit {
 
         return {
           type: heading.itemType,
+          styleAs: heading.styleAs,
           text: value
         };
       });
@@ -95,6 +96,7 @@ class Audit {
   static makeV2TableHeaders(headings) {
     return headings.map(heading => ({
       type: 'text',
+      styleAs: heading.styleAs,
       text: heading.text
     }));
   }
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -81,7 +81,7 @@ class Audit {
 
         return {
           type: heading.itemType,
-          styleClass: heading.styleClass,
+          styleAs: heading.styleAs,
           text: value
         };
       });
@@ -96,7 +96,7 @@ class Audit {
   static makeV2TableHeaders(headings) {
     return headings.map(heading => ({
       type: 'text',
-      styleClass: heading.styleClass,
+      styleAs: heading.styleAs,
       text: heading.text
     }));
   }
@@ -164,7 +164,6 @@ module.exports = Audit;
  *   key: string,
  *   itemType: string,
  *   text: string,
- *   styleClass: (string|undefined),
  * }>}
  */
 Audit.Headings; // eslint-disable-line no-unused-expressions
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -134,8 +134,8 @@ class OffscreenImages extends ByteEfficiencyAudit {
       });
 
       const headings = [
-        {key: 'preview', itemType: 'thumbnail', text: '', styleAs: 'preview'},
-        {key: 'url', itemType: 'url', text: 'URL', styleAs: 'url'},
+        {key: 'preview', itemType: 'thumbnail', text: ''},
+        {key: 'url', itemType: 'url', text: 'URL'},
         {key: 'totalKb', itemType: 'text', text: 'Original'},
         {key: 'potentialSavings', itemType: 'text', text: 'Potential Savings'},
       ];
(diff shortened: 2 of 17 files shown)


lighthouse-core/audits/dobetterweb/notification-on-start.js
```diff
@@ -49,7 +49,7 @@ class NotificationOnStart extends ViolationAudit {
     const results = ViolationAudit.getViolationResults(artifacts, /notification permission/);
 
     const headings = [
-      {key: 'url', itemType: 'url', text: 'URL'},
+      {key: 'url', itemType: 'url', text: 'URL', styleAs: 'url'},
```
ebidel: Another approach I was exploring is to allow headers to define their own styling: `style: "width:50px"`, then use that in the renderer. It does pollute the audits a bit with rendering information, but it may be more flexible in the long-term. Also cool with sticking with classes for now.

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -88,7 +88,7 @@ class LoadFastEnough4Pwa extends Audit {
         const recordsWithLatencies = allRequestLatencies.filter(val => val !== undefined);
 
         const details = Audit.makeV2TableDetails([
-          {key: 'url', itemType: 'url', text: 'URL'},
+          {key: 'url', itemType: 'url', text: 'URL', styleAs: 'url'},
```
ebidel: nit: maybe`styleClass`?

lighthouse-core/report/v2/report-styles.css
```diff
@@ -602,6 +602,38 @@ summary.lh-passed-audits-summary::-webkit-details-marker {
...
 .lh-table {
   --image-preview-size: 24px;
+  border: 1px solid var(--report-secondary-border-color);
+  border-collapse: collapse;
+  table-layout: fixed;
+  width: 100%;
+}
+
+.lh-table thead {
+  background: #ccc;
```
ebidel: `--report-secondary-border-color` has less weight. If you don't want to overload `--report-secondary-border-color`, maybe define a new var.

ebidel:
Test!

patrickhulce:
> Test!

yeah we don't have any tests for tables at all so it took slightly longer :P

lighthouse-core/report/v2/report-styles.css
patrickhulce: new var it is

lighthouse-core/audits/audit.js
```diff
@@ -81,6 +81,7 @@ class Audit {
 
         return {
           type: heading.itemType,
+          styleAs: heading.styleAs,
```
brendankenny: looks like you'll have to update the `Audit.Headings` typedef?

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -134,8 +134,8 @@ class OffscreenImages extends ByteEfficiencyAudit {
       });
 
       const headings = [
-        {key: 'preview', itemType: 'thumbnail', text: ''},
-        {key: 'url', itemType: 'url', text: 'URL'},
+        {key: 'preview', itemType: 'thumbnail', text: '', styleAs: 'preview'},
+        {key: 'url', itemType: 'url', text: 'URL', styleAs: 'url'},
```
brendankenny: what's the difference between styling as URL and type URL? Can we make them more orthogonal so it's clearer how/when to use one or the other or both?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -190,6 +194,7 @@ if (typeof module !== 'undefined' && module.exports) {
 /**
  * @typedef {{
  *     type: string,
+ *     styleAs: (string|undefined),
```
brendankenny: since this is so table specific, it probably makes sense to make a new `TableItemJSON` or something

lighthouse-core/audits/audit.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/offscreen-images.js
patrickhulce: changed to `styleClass` per eric's suggestion and difference is structure of dom (which renderer to use) vs. styles to apply (which css to apply). they will typically move together seemed bad to force them to always be the same

patrickhulce: possible example in the near future if we keep KB and % in byte efficiency

```
{key: 'wastedKb', itemType: 'text', styleClass: 'bold'},
{key: 'wastedPercent', itemType: 'text', styleClass: 'italicized'},
```

lighthouse-core/report/v2/renderer/details-renderer.js
patrickhulce: I didn't see it as table-only we just didn't have a need for text styling yet, should I add back in the class to the text element too?

lighthouse-core/audits/byte-efficiency/offscreen-images.js
brendankenny: ah, `styleClass` does make it clearer. Is there a more general name we can use over `url` to avoid confusion there too? (`preview` is also pretty non-descriptive, but I don't know if we can do better)

lighthouse-core/report/v2/renderer/details-renderer.js
brendankenny: > I didn't see it as table-only we just didn't have a need for text styling yet

only the table renderer looks for it and only table css exists to support it, so I think it makes sense to make it clear it's limited there until someone comes along and adds broader support.

brendankenny: > should I add back in the class to the text element too?

can you explain what you mean :)

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -134,8 +134,8 @@ class OffscreenImages extends ByteEfficiencyAudit {
       });
 
       const headings = [
-        {key: 'preview', itemType: 'thumbnail', text: ''},
-        {key: 'url', itemType: 'url', text: 'URL'},
+        {key: 'preview', itemType: 'thumbnail', text: '', styleClass: 'preview'},
+        {key: 'url', itemType: 'url', text: 'URL', styleClass: 'url'},
```
ebidel: @brendankenny convinced me that this could feel repetitive and unclear for contributors.

@patrickhulce wdyt, about `renderTable` having a switch statement in it that decides which column class to apply based on the `details.type`. Right now we don't have complicated styles with hybrid outputs. I think that would be ok for now.

patrickhulce: alright talked with paul too, for now we'll just add `lh-table-column--${item.type}` and get rid of this `styleClass` business

lighthouse-core/audits/byte-efficiency/offscreen-images.js
patrickhulce: no longer needed :)

lighthouse-core/report/v2/renderer/details-renderer.js
patrickhulce: I had originally implemented styleClass support for arbitrary text elements but it wasn't needed to do just the basic table styling, so I punted to do separately, now we're getting rid of it so won't be added back anyhow

patrickhulce:
@brendankenny changes still requested?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -126,15 +126,18 @@ class DetailsRenderer {
     const theadElem = this._dom.createChildOf(tableElem, 'thead');
     const theadTrElem = this._dom.createChildOf(theadElem, 'tr');
 
-    for (const heading of details.itemHeaders) {
-      this._dom.createChildOf(theadTrElem, 'th').appendChild(this.render(heading));
-    }
+    details.itemHeaders.forEach((heading, index) => {
+      const itemType = details.items[0][index].type;
```
brendankenny: this kind of goes counter to supporting different types per item in the column :) Do we have places where we use that yet? Could cause issues depending on which item is first.

FWIW the headers (as generated by `Audit.makeV2TableDetails`) have an `itemType` property, we just drop them from the typedef by the time it gets to the renderer. Could pipe them through to here if you wanted.

patrickhulce: yeah, hence why I wanted to separate this from the type :P haha I'll plumb itemType through for now

paulirish:
@patrickhulce merge at will

PR #2211, opened 2017-05-10
update 'unoptimized imgs' helpText and link to doc

kaycebasques:


lighthouse-core/audits/byte-efficiency/uses-optimized-images.js
```diff
@@ -42,10 +42,10 @@ class UsesOptimizedImages extends ByteEfficiencyAudit {
       name: 'uses-optimized-images',
       description: 'Optimize images',
       informative: true,
-      helpText: 'Images should be optimized to save network bytes. ' +
-        'The following images could have smaller file sizes when compressed with ' +
-        '[WebP](https://developers.google.com/speed/webp/) or JPEG at 80 quality. ' +
-        '[Learn more about image optimization](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization).',
+      helpText: 'Optimized images take less time to download and save cellular data. ' +
+        'Consider converting the following images to WebP, or lowering their JPEG quality ' +
```
ebidel: If the audit passes (user has optimized all images), "Consider converting the following images..." won't make sense. We should keep the helpText agnostic of a pass/fail state.

kaycebasques: How about we stick to Patrick's approach [here](https://github.com/GoogleChrome/lighthouse/issues/2213#issuecomment-300598841)

Describe why the audit is important, and then just a link to the doc to find out how to fix

kaycebasques:
@patrickhulce SGTM

brendankenny:
resolution on this now that #2216 has landed?

patrickhulce:
> resolution on this now that #2216 has landed? 

I think we can close, but defer to @kaycebasques in case he'd still want changes.

PR #2212, opened 2017-05-10
update "oversized imgs" helpText and link to doc

kaycebasques:


lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
```diff
@@ -41,9 +41,10 @@ class UsesResponsiveImages extends ByteEfficiencyAudit {
       description: 'Properly size images',
       informative: true,
       helpText:
-        'Image sizes served should be based on the device display size to save network bytes. ' +
-        'Learn more about [responsive images](https://developers.google.com/web/fundamentals/design-and-ui/media/images) ' +
-        'and [client hints](https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints).',
+        'Serve images that are smaller than the user\'s viewport to save cellular data ' +
+        'and improve load time. Consider using responsive images or client hints for the ' +
```
ebidel: should we link "client hints" to something?

kaycebasques: I actually think we should just stick to @patrickhulce's idea to restrict the `helpText` to just the "why" of the audit, and leave the "how" in the reference doc (when it's available)

When a reference doesn't yet exist, then we can link to "how" pages, such as client hints

Otherwise we're going to endlessly go back and forth on when we should include "how" content, and when we should omit

kaycebasques: So my proposal is to remove that sentence entirely

lighthouse-core/audits/byte-efficiency/uses-responsive-images.js
```diff
@@ -41,9 +41,9 @@ class UsesResponsiveImages extends ByteEfficiencyAudit {
       description: 'Properly size images',
       informative: true,
       helpText:
-        'Image sizes served should be based on the device display size to save network bytes. ' +
-        'Learn more about [responsive images](https://developers.google.com/web/fundamentals/design-and-ui/media/images) ' +
-        'and [client hints](https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints).',
+        'Serve images that are smaller than the user\'s viewport to save cellular data ' +
```
patrickhulce: this may be nitpicky but since we're improving the description here... :) it's not just the viewport size but anywhere that's bigger than its rendered size, what about just "Serve images that are appropriately sized to save cellular data..." and all the nitty details can be at the Learn More?

kaycebasques: Didn't know that.... SGTM

kaycebasques: P.S. that's not a nit, that's a significant change in behavior

Somebody might fail the audit, read the description I wrote, and be like "wtf why am I failing this, the image is not larger than the viewport"

patrickhulce: haha, well I said nitpicky because the linked doc describes it correctly with "Lighthouse compares the size of the rendered image against the size of the actual image." and the previous description wasn't exactly illuminating here and just said "based on the display size" which could be interpreted as the "screen size" instead of "rendered size"

but yeah I'm glad we're fixing it :)

kaycebasques:
I put in a bug on the doc b/c I actually wrote it with the wrong mental
model

Issue #2189, opened 2017-05-09
PWA fast enough: Update text about metric


paulirish:
The audit checks TTFI but talks about TTI.

Last discussion here: https://github.com/GoogleChrome/lighthouse/pull/2159/files#r115051696

PR #2215, opened 2017-05-10
Install banner audit: failures should say what the user didnt do

ebidel:
Matches what the "Configure for a splash screen audit" does. It was confusing that the failure help text was saying the "site registers a sw", when in fact, it did not.

screen shot 2017-05-10 at 2 06 37 pm

Also fixes https://github.com/GoogleChrome/lighthouse/issues/2189.

### DIFF
--- a/lighthouse-core/audits/time-to-interactive.js
+++ b/lighthouse-core/audits/time-to-interactive.js
@@ -27,8 +27,8 @@ class TTIMetric extends Audit {
     return {
       category: 'Performance',
       name: 'time-to-interactive',
-      description: 'Time To First Interactive (alpha)',
-      helpText: 'Time to First Interactive identifies the time at which your app appears to be ready ' +
+      description: 'Time To Interactive (alpha)',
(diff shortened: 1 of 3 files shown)


lighthouse-core/audits/time-to-interactive.js
```diff
@@ -27,8 +27,8 @@ class TTIMetric extends Audit {
     return {
       category: 'Performance',
       name: 'time-to-interactive',
-      description: 'Time To Interactive (alpha)',
```
paulirish: nope.  :)
 that's https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/first-interactive.js which is a different metric altogether.

brendankenny:
lint error and somehow this is causing a unit test error. Looks like it may be checking the error string?

lighthouse-core/audits/time-to-interactive.js
brendankenny: oh, right. Yeah, it's `load-fast-enough-for-pwa` that needs updating to First Interactive

PR #2216, opened 2017-05-10
refactor: split optimized images into WebP and optimize audit

### LINKED ISSUE #2044: opinionated "unoptimized images" audit is opinionated
Just a general first impression of the "unoptimized images" audit now that I've dived into it

The comments in the audit source say that the audit fails when:

* a JPEG / bitmap can be at least 10KB smaller when canvas-encoded
* any image can be at least 100KB smaller when WebP-encoded
* total savings can be greater than 1MB by encoding all images with WebP

So really, the audit is mainly "look at how much you can save with WebP". But it still doesn't have support Firefox, Edge, or Safari, so that seems like a moot point for a solid chunk of our audience

Is there any way to optimize imgs using the original format? I know it's kinda arbitrary, b/c you could optimize the heck out of a JPEG, but if it looks like garbage, then the optimized version isn't valid. But for me personally, I'd be way more interested if LH said "here's an optimized version of your img, and this is how much you could save". Maybe LH could do a side-by-side right there so that the user can see whether the optimized version is valid.

Maybe another way to approach this would be to just flag images larger than some arbitrary size? Or to warn when the total amount of images exceeds some arbitrary limit?

patrickhulce:
closes #2044 and does some drive by cleanup of byte efficiency audits

Splits the audit to track two concerns separately:
1. **Are you doing the most sensible optimizations for your existing image format?**
JPEG doesn't have an insane amount of metadata or is too high quality, can be expanded to do the equivalent of PNG crush too.
2. **Are you using the most efficient image format for your content?**
Currently just a WebP check but could be fancier for your PNG could actually be a JPEG, etc.

Also reduces the aggressiveness of both of these by increasing the quality to 85 and increasing the ignore threshold to 4KB for JPEG and 8KB for WebP.

Questions:
* ~~Should the WebP check be moved down into diagnostics?~~ No
* Should we rename "Optimize images" since it's just tracking the lowest hanging fruit optimizations, if so, to what?

screenshot: (added by paul)

image


ebidel:
This makes sense to me. I think we should keep the audits in "Opportunities" (not move them to Diagnostics) as they are potential improvements to be made.

One thing, can you group the 3 image audits together, probably in this order:
1. Properly size images
2. Optimize images
3. Convert images to WebP

screen shot 2017-05-14 at 3 37 35 pm

"Convert images to WebP" could also be "Serve images as WebP" to suggest that this would be in addition to serving other formats. But that would also address Kayce's concerns in the bug.

ebidel:
Needs a rebase.

patrickhulce:
also turns out we've been hiding `offscreen-images` from the perf section this whole time, so I added that back

patrickhulce:
@ebidel PTAL rebased and updated order :)

paulirish:
I'm really happy with this split. Good call.

I like the title of both of these audits now. seems good.

PR #2220, opened 2017-05-10
perf: make network quiet threshold configurable per pass

patrickhulce:
in the name of OYB, shaves off ~9 seconds from all runs with the default passes

renames pauseAfterLoad to the more appropriate networkQuietThreshold, previously the flag was for speeding up the tests and was not documented in the CLI

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -30,7 +30,7 @@ module.exports = {
   {
     "passName": "offlinePass",
     "useThrottling": false,
-    "networkQuietThresholdMs": 500,
+    "networkQuietThresholdMs": 100,
     "gatherers": [
       "service-worker",
       "offline",
@@ -40,7 +40,7 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
-    "networkQuietThresholdMs": 500,
+    "networkQuietThresholdMs": 100,
     // Speed up the redirect pass by blocking stylesheets, fonts, and images
     "blockedUrlPatterns": ["*.css", "*.jpg", "*.jpeg", "*.png", "*.gif", "*.svg", "*.ttf", "*.woff", "*.woff2"],
     "gatherers": [
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -4,6 +4,7 @@ module.exports = {
   "passes": [{
     "passName": "defaultPass",
     "recordTrace": true,
+    "networkQuietThresholdMs": 5000,
     "pauseBeforeTraceEndMs": 5000,
     "useThrottling": true,
     "gatherers": [
@@ -29,6 +30,7 @@ module.exports = {
   {
     "passName": "offlinePass",
     "useThrottling": false,
+    "networkQuietThresholdMs": 500,
     "gatherers": [
       "service-worker",
       "offline",
@@ -38,6 +40,7 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
+    "networkQuietThresholdMs": 500,
     // Speed up the redirect pass by blocking stylesheets, fonts, and images
     "blockedUrlPatterns": ["*.css", "*.jpg", "*.jpeg", "*.png", "*.gif", "*.svg", "*.ttf", "*.woff", "*.woff2"],
     "gatherers": [
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -4,8 +4,9 @@ module.exports = {
   "passes": [{
     "passName": "defaultPass",
     "recordTrace": true,
+    "pauseAfterLoadMs": 5000,
     "networkQuietThresholdMs": 5000,
-    "pauseBeforeTraceEndMs": 5000,
+    "pauseAfterNetworkQuietMs": 2500,
     "useThrottling": true,
     "gatherers": [
       "url",
@@ -30,7 +31,8 @@ module.exports = {
   {
     "passName": "offlinePass",
     "useThrottling": false,
-    "networkQuietThresholdMs": 100,
+    // Just wait for onload
+    "networkQuietThresholdMs": 0,
     "gatherers": [
       "service-worker",
       "offline",
@@ -40,7 +42,8 @@ module.exports = {
   {
     "passName": "redirectPass",
     "useThrottling": false,
-    "networkQuietThresholdMs": 100,
+    // Just wait for onload
+    "networkQuietThresholdMs": 0,
     // Speed up the redirect pass by blocking stylesheets, fonts, and images
     "blockedUrlPatterns": ["*.css", "*.jpg", "*.jpeg", "*.png", "*.gif", "*.svg", "*.ttf", "*.woff", "*.woff2"],
     "gatherers": [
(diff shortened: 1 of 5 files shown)


lighthouse-core/config/default.js
```diff
@@ -29,6 +30,7 @@ module.exports = {
   {
     "passName": "offlinePass",
     "useThrottling": false,
+    "networkQuietThresholdMs": 500,
```
paulirish: do we even need to wait 500 on these two passes?

brendankenny:
wait, this was only 5000 because of #2023 :P It was 500 before

lighthouse-core/config/default.js
patrickhulce: yeah probably not we really just need the main page, right?

brendankenny:
This explains the end of trace ending up 10 seconds after most of the other metrics had completed, since we have the double 5 second wait.

Can we eliminate one of `networkQuietThresholdMs` or `pauseBeforeTraceEndMs` while we're at it? They're almost exactly the same, the only difference is it's
1) network quiet or onload
2) wait `networkQuietThresholdMs`
3) run all gatherer.pass()es
4) wait `pauseBeforeTraceEndMs`
5) end trace

It doesn't seem like both are necessary. Arguably `networkQuietThresholdMs` could go in the position of `pauseBeforeTraceEndMs`, because if a gatherer is doing anything in its `pass` method in a pass with `recordTrace`, it wants its activities to be recorded by the trace and delay TTI or whatever is being run on that pass.

patrickhulce:
> Can we eliminate one of networkQuietThresholdMs or pauseBeforeTraceEndMs while we're at it? 

The answer is how much TTCI failure are we willing to tolerate? Most sites should be fine without pauseBeforeTraceEnd, but there are definite cases where it will fail. Maybe we make it optional?

brendankenny:
for the default `defaultPass`, at least, there aren't any gatherers with a `pass()` method, so the actual execution is
1) network quiet or onload
2) wait networkQuietThresholdMs
3) wait pauseBeforeTraceEndMs
4) end trace

so we can just get rid of one and wait the sum of the current two, if we need that long

patrickhulce:
> so we can just get rid of one and wait the sum of the current two, if we need that long

It's not *quite* that simple, we can consolidate for sure but they serve different purposes. Increasing networkQuietThreshold to 10s might cause many more timeouts than the current approach, but adding 5s to pauseBeforeTraceEnd would also fail to capture network quiet periods more frequently than the current approach.

`networkQuietThreshold` at 5s is just a requirement of TTCI and we'll have to suffer through if we want to reliably determine it (though we could split into a 5s 2-inflight requirement and 500ms 0-inflight requirement).

Given `networkQuietThreshold` fixed at 5s, `pauseBeforeTraceEnd` should really be interpreted as "what's the longest we might expect a reasonable page to have main thread activity after all network requests have finished" Example: if the last network request finishes and then there's a single 50ms long task, we don't want that to make TTCI undeterminable. 5 seconds is quite generous, but 2s of main thread isn't that unheard of so we can play around with the number here.

No matter what we can definitely skip pauseBeforeTraceEnd in the timeout case which is the current worst case scenario and shave off 5s.

brendankenny:
I think I see what you're saying. Basically we want five seconds of network quiet, but in theory the first few seconds of that window could contain a long task, so we need to wait longer?

If I understand you correctly, there's not a ton of difference in waiting for seven seconds of network quiet. A single network request could go out after five seconds cancelling the network quiet detection but before the CPU quieted down, but with the current scheme three network requests going out after `networkQuietThreshold` but less than five seconds after CPU quiet would have the same effect. Three new requests is less likely than one but it's not great.

Basically this is all hand wavey and a post hoc justification and maybe that's what smells about it :) The fact that we don't have a way to observe the main thread is the primary issue with finding a good solution. If `networkQuietThreshold` was really all about TTCI, it would be listening for dropping down to two, not zero, network requests for five seconds. What it was originally for was that was our best guess at "loaded" when we wrote it (better than onload + pause), and now it's functioning as a clumsy heuristic for CPU quiet. "5s 2-inflight requirement and 500ms 0-inflight requirement" suffers from the same issue.

For now, how do you feel about renaming `pauseBeforeTraceEnd` to `pauseAfterNetworkQuiet` or whatever (`theLimitOfOurPatienceWaitingForAPossibleLongTask`) and moving it out of `endTrace` and into `gather-runner` where it belongs (it doesn't make sense as an option to ending the trace and it delays everything afterwards, not just the ending the trace, so might as well make it prominent).

brendankenny:
We should also consider (in the future) upping `_waitForNetworkIdle` to wait for one request (and maybe even two) and see its effect on average runtime. We've suspected there are a number of sites that hit the 25s timeout that don't need to because they have a single request open (Does FMP still allow one(?) open request when declaring the page ready and the last candidate official?). Many sites might not budge, but average runtime could be decreased significantly.

and we should add time to run lighthouse to plots/ cause we don't know what the average runtime is :)

patrickhulce:
> For now, how do you feel about renaming pauseBeforeTraceEnd to pauseAfterNetworkQuiet or whatever (theLimitOfOurPatienceWaitingForAPossibleLongTask) and moving it out of endTrace

ya let's do it

patrickhulce:
OK lots of new changes :)

I've split the thresholds into 3 to accurately track what they're trying to do and moved it all into the page load

* network-2-idle and network-2-busy have been added to network recorder
* pauseAfterLoad has been split into pauseAfterLoad and pauseAfterNetworkQuiet to appropriately account for the fact that in networkQuiets case we've already had 5s of 2-quiet silence, so we probably don't need another full 5
* networkQuietThreshold is used as the debounce time for network-2-idle instead of abusing pauseAfterLoad
* There's no more pauseBeforeTraceEnd, it's intention is captured by pauseAfterLoad and pauseAfterNetworkQuiet

lighthouse-core/gather/driver.js
```diff
@@ -25,13 +25,15 @@ const URL = require('../lib/url-shim');
...
+const DEFAULT_PAUSE_AFTER_LOAD = 0;
+const DEFAULT_NETWORK_QUIET_THRESHOLD = 5000;
+const DEFAULT_PAUSE_AFTER_NETWORK_QUIET = 0;
 
 const _uniq = arr => Array.from(new Set(arr));
 
 class Driver {
   static get MAX_WAIT_FOR_FULLY_LOADED() {
-    return 25 * 1000;
+    return 30 * 1000;
```
patrickhulce: increased to account for the fact that the trace will end abruptly after this time instead of 5s after it, so we still aren't capturing any larger traces than 30s

lighthouse-core/gather/driver.js
```diff
@@ -25,13 +25,15 @@ const URL = require('../lib/url-shim');
 const log = require('../lib/log.js');
 const DevtoolsLog = require('./devtools-log');
 
-const PAUSE_AFTER_LOAD = 5000;
+const DEFAULT_PAUSE_AFTER_LOAD = 0;
```
paulirish: can you add comments above these to explain what they are for? otherwise folks will have to backtrack to this discussion to totally understand. :)

patrickhulce: yup

PR #2226, opened 2017-05-11
Report: add stub for NodeDetails renderer

paulirish:
yesterday, I rolled to devtools with some LH changes that aren't yet in master here (they're in #2206)

as a result, i broke `compile-devtools`. :)
https://travis-ci.org/GoogleChrome/lighthouse/jobs/231226842#L661

This adds a stub that restores compilation while 2206 gets sorted.

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -44,6 +44,8 @@ class DetailsRenderer {
         return this._renderTable(/** @type {!DetailsRenderer.TableDetailsJSON} */ (details));
       case 'code':
         return this._renderCode(details);
+      case 'node':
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -144,6 +146,15 @@ class DetailsRenderer {
   }
 
   /**
+   * @param {!DetailsRenderer.NodeDetailsJSON} item
+   * @return {!Element}
+   * @protected
+   */
+  renderNode(item) {
```
patrickhulce: still gonna change this name though right? ;)

paulirish: ya ill change to `renderDomNode` later on

paulirish:
travis is good: https://travis-ci.org/GoogleChrome/lighthouse/jobs/231235163#L612

PR #2228, opened 2017-05-11
refactor: prepare for TTI removal

patrickhulce:
this moves the metrics currently referenced by pwmetric-events into other audits so that TTI can be easily removed

there's getting to be a lot of churn on those test fixtures which will need to happen again when TTI is removed, @paulirish thoughts on how lenient you're willing to be in those tests to avoid this?


lighthouse-core/audits/speed-index-metric.js
```diff
@@ -62,6 +62,13 @@ class SpeedIndexMetric extends Audit {
         throw new Error('Error in Speedline calculating Speed Index (speedIndex of 0).');
       }
 
+      let visuallyReadyInMs = undefined;
+      speedline.frames.forEach(frame => {
+        if (frame.getPerceptualProgress() >= 85 && typeof visuallyReadyInMs === 'undefined') {
```
brendankenny: just use a for loop with a conditional instead of this? :)

lighthouse-core/audits/first-meaningful-paint.js
```diff
@@ -84,35 +80,40 @@ class FirstMeaningfulPaint extends Audit {
...
-    };
-
+  static calculateScore(traceOfTab) {
     // Expose the raw, unchanged monotonic timestamps from the trace, along with timing durations
     const extendedInfo = {
       timestamps: {
-        navStart: getTs(evts.navigationStart),
-        fCP: getTs(evts.firstContentfulPaint),
-        fMP: getTs(evts.firstMeaningfulPaint)
+        navStart: traceOfTab.timestamps.navigationStart * 1000,
```
brendankenny: lol timestamps and units

PR #2218, opened 2017-05-10
Adds Critical Request Chain details renderer

ebidel:
screen shot 2017-05-10 at 3 08 44 pm

This one was a real pain :) Handlebars -> real code. Ugh. FWIW, I did not attempt rewrite Paul's tree code, so please keep bikeshedding to a minimum. The helpers come from handelbars-helpers, but I've renamed them slightly.

Other things this PR does:
- Moves the url-shim's 'getDisplayName` into `Util.getDisplayName` so we can use it in report renderers.
- Pipes `_templateContext` into the details-renderer so we can use templates in that file.

@brendankenny is going to finish off the closure errors.

---
This change is [Reviewable](https://reviewable.io/reviews/googlechrome/lighthouse/2218)

PR #2233, opened 2017-05-12
Fix: Catch all images as non-critical

### LINKED ISSUE #2219: Critical req chains: ignore image/* mimeType
Sites like http://veerle.duoh.com/ issue requests for images. Especially with favicons the devtools side doesn't identify them as `CategoryType === Image`.

See the screenshot in https://github.com/GoogleChrome/lighthouse/pull/2218 as evidence that this is still a problem. :)

So [over in isCritical](https://github.com/GoogleChrome/lighthouse/blob/3f22558aca0285b8b05585d4e1fa2597ef318ae1/lighthouse-core/gather/computed/critical-request-chains.js#L29-L55), we should exclude all requests that were sent with contentType `image/*`

@stramel wanna take this?

stramel:
Fixes #2219 Sorry it took so long!

Hope this helps :grimacing:

### DIFF
--- a/lighthouse-core/gather/computed/critical-request-chains.js
+++ b/lighthouse-core/gather/computed/critical-request-chains.js
@@ -37,17 +37,13 @@ class CriticalRequestChains extends ComputedArtifact {
 
     // XHRs are fetched at High priority, but we exclude them, as they are unlikely to be critical
     // Images are also non-critical.
+    // Treat any images missed by category, primarily favicons, as non-critical resources
     const nonCriticalResourceTypes = [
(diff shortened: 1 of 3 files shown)


stramel:
For some reason, it keeps discarding my updates to my original description.

Ran a quick test against http://veerle.duoh.com/ with @ebidel's changes to visualize the request chain and you can compare it against #2218. All favicons have been removed.

image

paulirish:
Yup. This LGTM.

thx!

stramel:
You're welcome!

PR #2234, opened 2017-05-12
Switch to containerized Trusty

### LINKED ISSUE #2224: Use container-based infrastructure on Travis
Hey,

I'm just sharing my experience on the subject because I had a hard time making things work on Travis for Speed Racer. I basically took a lot of inspiration on what you do (thanks for that btw).

I'm just wondering why you would need `sudo`?

stramel:
Addresses #2224 

Total run time seems slower, but it is containerized /shrug Up to you guys

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ matrix:
   include:
     - node_js: "6.9.1"
     - node_js: "7"
-sudo: required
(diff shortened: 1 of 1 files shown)


paulirish:
Let's let this guy chill under after I/O. :)

PR #2236, opened 2017-05-12
Include commit hash and version to browser extension

### LINKED ISSUE #2190: Expose git hash to extension build - followup
see this comment for updates: https://github.com/GoogleChrome/lighthouse/issues/2190#issuecomment-322321863

-------

The lighthouse-background.js file is built from some snapshot of master
But we dont have an easy wasy of finding out which commit it was. We want to know both in the chrome extension and in devtools.

For devtools, it seems sufficient to have something like this at the top of the file:

```js
// lighthouse, browserified. v2.0.0-alpha.3 (5dc01be664e883db58566f6358e9c15f67a03d6e)

```

we could probably live with that solution for the extension as well.

@wardpeet wanna try this out?

wardpeet:
fixes #2190


wardpeet:
image

wardpeet:
Rebased and ready to go!
image

lighthouse-extension/app/src/popup.js
```diff
@@ -49,6 +49,10 @@ function getLighthouseVersion() {
   return chrome.runtime.getManifest().version;
 }
 
+function getLighthouseRevision() {
```
brendankenny: bikeshed: can we call this `getLighthouseCommitHash` or something like that (and update references to `REVISION` below)? I wasn't sure what "revision" was referring to until I got to where it was defined :)

`**Lighthouse Revision**` below could be `**Lighthouse Commit**`, for instance.

Just an idea, though. Open to other ways people refer to the exact commit

lighthouse-extension/gulpfile.js
```diff
@@ -149,6 +162,13 @@ gulp.task('browserify-other', () => {
     .pipe(tap(file => {
       let bundle = browserify(file.path); // , {debug: true}); // for sourcemaps
       bundle = applyBrowserifyTransforms(bundle);
+
+      // replace revision
+      bundle.transform('browserify-versionify', {
+        placeholder: '__REVISION__',
+        version: REVISION,
+      });
```
brendankenny: could we drop the `insertGlobalVars`/`browserify-versionify` parts and just do a `.pipe(gulpReplace('__REVISION__', REVISION))`?

lighthouse-extension/gulpfile.js
```diff
@@ -174,6 +194,7 @@ gulp.task('compilejs', () => {
       file.contents = new Buffer(minified);
       return file;
     }))
+    .pipe(banner(BANNER))
```
brendankenny: Since it's a build step (and so just devDependencies), I personally don't particularly mind using yet another gulp library instead of writing our own function. It seems like `gulp-header` would let you inject this comment pretty easily at the top.

package.json
```diff
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@types/node": "^6.0.45",
     "babel-core": "^6.16.0",
+    "browserify-versionify": "^1.0.6",
```
brendankenny: I think we want this (or whatever new deps we add) in `lighthouse-extension/package.json` so that it's scoped to only install for those who are working on/with the extension

lighthouse-extension/gulpfile.js
brendankenny: > `.pipe(gulpReplace('__REVISION__', REVISION))`

we could make the string to replace even more unique if necessary

lighthouse-extension/app/src/popup.js
wardpeet: I don't mind what it's called

lighthouse-extension/gulpfile.js
wardpeet: Sure, I can take a dependency, just didn't want to introduce another one but yeah dev dependency doesn't really matter

package.json
wardpeet: I believe we also wanted this inside the report, or is this really only extension?

wardpeet:
@brendankenny @paulirish just notifying you guys. Pretty sure it's ok now.

yarn.lock
```diff
@@ -1008,6 +1008,10 @@ find-index@^0.1.1:
   version "0.1.1"
   resolved "https://registry.yarnpkg.com/find-index/-/find-index-0.1.1.tgz#675d358b2ca3892d795a1ab47232f8b6e2e0dde4"
 
+find-root@^0.1.1:
```
brendankenny: is this still necessary? Seems like changes to this file should have been dropped since no changes to root package.json

lighthouse-extension/gulpfile.js
```diff
@@ -15,10 +15,20 @@ const livereload = require('gulp-livereload');
...
 const zip = require('gulp-zip');
+const gulpReplace = require('gulp-replace');
+const header = require('gulp-header');
 const LighthouseRunner = require('../lighthouse-core/runner');
+const pkg = require('../package.json');
 
 const distDir = 'dist';
 
+const VERSION = pkg.version;
+const COMMITHASH = require('child_process')
```
brendankenny: nit: maybe `COMMIT_HASH`?

yarn.lock
wardpeet: I cleaned the yarn.lock with executing `yarn`, should be good to go

brendankenny:
reverted part of a comment removed by a merge

PR #2238, opened 2017-05-12
Update network throttling to simulate more realistic network conditions

wwwillchen:
Based on the work in https://github.com/GoogleChrome/lighthouse/pull/2183 which I'll land as a follow-up.

### DIFF
--- a/lighthouse-core/lib/emulation.js
+++ b/lighthouse-core/lib/emulation.js
@@ -43,14 +43,15 @@ const NEXUS5X_USERAGENT = {
 
 /**
  * Adjustments needed for DevTools network throttling to simulate
- * more realistic network throttling conditions
+ * more realistic network conditions.
+ * See: crbug.com/721112
  */
 const LATENCY_FACTOR = 3.75;
 const THROUGHPUT_FACTOR = 0.9;
 
 const TARGET_LATENCY = 150; // 150ms
 const TARGET_DOWNLOAD_THROUGHPUT = Math.floor(1.6 * 1024 * 1024 / 8); // 1.6Mbps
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/emulation.js
```diff
@@ -41,10 +41,24 @@ const NEXUS5X_USERAGENT = {
...
+/**
+ * Adjustments needed for DevTools network throttling to simulate
+ * more realistic network throttling conditions
+ */
+const LATENCY_FACTOR = 3.75;
+const THROUGHPUT_FACTOR = 0.9;
+
+const TARGET_LATENCY = 150; // 150ms
+const TARGET_DOWNLOAD_THROUGHPUT = Math.floor(1.6 * 1024 * 1024 / 8); // 1.6Mbps
+const TARGET_UPLOAD_THROUGHPUT = Math.floor(750 * 1024 / 8 * THROUGHPUT_FACTOR); // 750Kbps
```
patrickhulce: why is the throughput factor included in the target here?

lighthouse-core/lib/emulation.js
```diff
@@ -41,10 +41,24 @@ const NEXUS5X_USERAGENT = {
     '(KHTML, like Gecko) Chrome/59.0.3033.0 Mobile Safari/537.36'
 };
 
+/**
+ * Adjustments needed for DevTools network throttling to simulate
+ * more realistic network throttling conditions
```
patrickhulce: should we link off to any explanations or references?

lighthouse-core/lib/emulation.js
wwwillchen: whoops - fixed.

lighthouse-core/lib/emulation.js
wwwillchen: Added a link to the crbug (not sure whether to make the analysis doc public yet)

PR #2239, opened 2017-05-12
feat: add filmstrip to report

patrickhulce:
#2181 was a basic port of the logic currently in devtools, this addresses the first part (filmstrip) of it in a sustainable way by putting the screenshot thumbnails directly into the LHR and rendering them in the performance category

part two will add the metrics and adjust the timeline to match the metrics

image


lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -0,0 +1,119 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
+const Audit = require('./audit');
+const jpeg = require('jpeg-js');
+
+const NUMBER_OF_THUMBNAILS = 10;
+const THUMBNAIL_HEIGHT = 100;
```
paulirish: it'd be better to constrain on width, so we can be sure that 10thumbnails has a reliable total width

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -0,0 +1,119 @@
...
+    return artifacts.requestSpeedline(trace).then(speedline => {
+      const thumbnails = [];
+      const analyzedFrames = speedline.frames.filter(frame => !frame.isProgressInterpolated());
+
+      for (let i = 1; i <= NUMBER_OF_THUMBNAILS; i++) {
+        const targetTimestamp = speedline.beginning + speedline.complete * i / NUMBER_OF_THUMBNAILS;
+
+        let targetFrame = null;
+        if (i === NUMBER_OF_THUMBNAILS) {
+          targetFrame = analyzedFrames[analyzedFrames.length - 1];
```
paulirish: s/targetFrame/frameForTime/

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -0,0 +1,119 @@
...
+  static audit(artifacts) {
+    const trace = artifacts.traces[Audit.DEFAULT_PASS];
+    const cachedThumbnails = new Map();
+
+    return artifacts.requestSpeedline(trace).then(speedline => {
+      const thumbnails = [];
+      const analyzedFrames = speedline.frames.filter(frame => !frame.isProgressInterpolated());
+
+      for (let i = 1; i <= NUMBER_OF_THUMBNAILS; i++) {
+        const targetTimestamp = speedline.beginning + speedline.complete * i / NUMBER_OF_THUMBNAILS;
```
paulirish: pavel had a slightly diff approach so that we'd favor timestamps that are multiples of 100ms. it did make for nicer numbers, but not a huge deal IMO

```js
      const numberOfFrames = 8;
      const roundToMs = 100;
      const timeStep = (Math.ceil(timeSpan / numberOfFrames / roundToMs)) * roundToMs;
```

brendankenny:
is this taking over from #2181?

patrickhulce:
> is this taking over from #2181?

yeah description added for better explanation :)

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: we didn't really have room over screenshots to have full ms on smaller widths, so seconds to to tenths are all that are shown on the UI, have a preference for pavel's method?

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: yeah sg

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: done

lighthouse-core/audits/screenshot-thumbnails.js
brendankenny: is `speedline.complete` the time to completion, not the time *of* completion?

lighthouse-core/test/audits/screenshot-thumbnails-test.js
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const assert = require('assert');
+
+const Runner = require('../../runner.js');
+const Audit = require('../../audits/screenshot-thumbnails');
```
brendankenny: s/`Audit`/`ScreenshotThumbnailsAudit`

lighthouse-core/test/audits/screenshot-thumbnails-test.js
```diff
@@ -0,0 +1,48 @@
...
+describe('Screenshot thumbnails', () => {
+  it('should extract thumbnails from a trace', () => {
+    const artifacts = Object.assign({
+      traces: {defaultPass: pwaTrace}
+    }, Runner.instantiateComputedArtifacts());
+
+    return Audit.audit(artifacts).then(results => {
+      results.rawValue.forEach((result, index) => {
+        const framePath = path.join(__dirname,
+            `../fixtures/traces/progressive-app-frame-${index}.jpg`);
```
brendankenny: can we put these in a directory of their own?

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -0,0 +1,119 @@
...
+
+        cachedThumbnails.set(frameForTimestamp, base64Data);
+        thumbnails.push({
+          timing: Math.round(targetTimestamp - speedline.beginning),
+          timestamp: targetTimestamp * 1000,
+          data: base64Data,
+        });
+      }
+
+      return {rawValue: thumbnails};
```
brendankenny: I think we assume that `rawValue` will be either a `boolean` or a `number` (which matters when creating the `score` from the `rawValue`). Not an issue here because the score is never used, but it would be nice to avoid having a special audit like this which only works because of the specific renderer path it hits. Ideally it would work if someone else wanted to mix it into their config without a perf section, for instance.

Can we mark it informative, have a `false` `rawValue` and stick the thumbnails in `details`?

brendankenny: (if we wanted to go crazy we could put the thumbnails in a list or table which `_renderPerformanceCategory` could ignore but others could use, but we should add tests and stuff to make sure it works when mixed in not in a perf category, so we can wait on that)

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -0,0 +1,119 @@
...
+   *
+   * @param {{width: number, height: number, data: !Array<number>}} imageData
+   * @return {{width: number, height: number, data: !Array<number>}}
+   */
+  static scaleImageToThumbnail(imageData) {
+    const scaledWidth = THUMBNAIL_WIDTH;
+    const scaleFactor = imageData.width / scaledWidth;
+    const scaledHeight = Math.floor(imageData.height / scaleFactor);
+
+    const outPixels = new Uint8Array(scaledWidth * scaledHeight * 4);
```
benschwarz: This is very clever @patrickhulce

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: right-o `end` is the timestamp of completion

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: I wasn't a fan of those options originally since `details` maps to the details renderer and, again, rendering info not related to how to display the audit itself just its collapsible extraneous information. In the generic view, "rawValue" as it's the raw screenshot data not formatted for a renderer, made the most sense to me for third parties consuming the JSON, and makes it clear that a generic renderer does not support this. I'll mark as informative and explicitly set the score so rawValue isn't used there for sure, but I'd want to avoid sticking into details as a list or table when it's distinctly something *not* handled or treated as either of those.

lighthouse-core/test/audits/screenshot-thumbnails-test.js
patrickhulce: done

lighthouse-core/test/audits/screenshot-thumbnails-test.js
patrickhulce: done

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: thanks :)

benschwarz:
I don't know that it should be a goal of lighthouse to do this, but I'll share this at any rate. 

I've already added a `screenshots` audit to Calibre's Lighthouse implementation, but I do not scale the images. If 'other' tools were to use this audit, it might be appropriate to return full size images. 

If this audit did return full size images, I'd use it for Calibre. Otherwise, I'll stick with the custom `screenshots` audit that I added. (It is essentially exactly the same as yours, without the resize)

patrickhulce:
@benschwarz we actually used to have a full-sized screenshots audit but removed it because it greatly increased the size of the JSON results and decided if you want the screenshots you could still save the trace, the main purpose of this particular audit is to provide a mechanism to create the HTML report based just on the JSON without also needing the full trace saved

lighthouse-core/audits/screenshot-thumbnails.js
paulirish: admittedly there isn't a great solution here. and i think we face the same question with @benschwarz's HAR audit.

rawValue should be either false or true. (I don't care much either way, but feels like true is good unless it failed to finish, right?)

brendankenny: `rawValue` is completely inaccurately named, but it shouldn't be punished for that :)

> I wasn't a fan of those options originally since `details` maps to the details renderer and, again, rendering info not related to how to display the audit itself just its collapsible extraneous information.

I've actually been thinking about this distinction since #2140, and I increasingly think it actually makes sense to use `details` in this way.

tl;dr: the non-default category renderers already render audits however they want. In the same way, there's no reason why `details` can't be collapsed extra information *or* top level display information depending on what the particular category renderer chooses to display.

The performance category renderer is already bastardizing what the default category renderer does...e.g. any audit without a `group` is ignored completely, and with `screenshot-thumbnails` it plucks it out specially and renders it specially. The perf category renderer is pretty much inescapably special due to what we need from it. But really the method dispatch difference between `_renderDefaultCategory` and `_renderPerformanceCategory` is a lot larger than the conceptual difference.

`this._renderFilmstrip()` *is* a details renderer (conceptually), it's just that `renderAudit()` -> `renderAuditScore()` -> `detailsRenderer.render()` has been flattened into the few lines that call `renderFilmstrip()`. It's a `renderAuditScore()` with the top audit information hidden and the `details` rendered not in a `<details>` element. This could easily be made explicit with a `renderPerfAudit()` method that dispatches based on `group` or whatever, but it's probably not worth it :)

And putting the info in `details` means that this and future audits like this are compatible with other category renderers...if someone wants to use the `ScreenshotThumbnails` in their own custom report, it would work without them having to make their own perf category with all the right setup.

patrickhulce: Updated to move into details

lighthouse-core/report/v2/report-styles.css
```diff
@@ -372,6 +372,29 @@ summary {
...
+  padding: var(--default-padding);
+}
+
+.lh-filmstrip__frame {
+  text-align: center;
+}
+
+.lh-filmstrip__timestamp {
+  margin-bottom: 5px;
+  font-weight: bold;
```
paulirish: nuke the bold

patrickhulce: you got it 👍

PR #2241, opened 2017-05-12
add lighthouse calculation time to json results

brendankenny:
happy to bikeshed on the name and where it should be taken, but basically it would be nice to be able to track how long a lighthouse run takes. I was going to do it in plots/, but @wwwillchen pointed out that it's actually a generally useful thing to have in the LHR.

We could put this inside `runner`, but it's actually kind of nice to track config and connection construction too. This does ignore work done in `bin.ts`, however.

### DIFF
--- a/lighthouse-core/index.js
+++ b/lighthouse-core/index.js
@@ -38,9 +38,10 @@ const Config = require('./config/config');
  */
 
 module.exports = function(url, flags = {}, configJSON) {
-  return new Promise((resolve, reject) => {
+  const startTime = Date.now();
+  return Promise.resolve().then(_ => {
     if (!url) {
-      return reject(new Error('Lighthouse requires a URL'));
+      throw new Error('Lighthouse requires a URL');
     }
 
     // set logging preferences, assume quiet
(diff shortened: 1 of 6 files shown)


lighthouse-core/index.js
```diff
@@ -38,9 +38,10 @@ const Config = require('./config/config');
  */
 
 module.exports = function(url, flags = {}, configJSON) {
-  return new Promise((resolve, reject) => {
+  const startTime = Date.now();
```
patrickhulce: Audit error: `Avoids Date.now() In Its Own Scripts` ;)

brendankenny:
no, that was my best effort at "timeDoingStuff" :) `totalRunTime` SGTM

brendankenny:
I guess the only confusing thing is `runtimeConfig`, `runtime` vs `runTime` :)

paulirish:
how about a timing object: 
```js
LHR.timing.total // 43752 (ms)
```
(i imagine we'll want breakdown by pass, gather audit later on)

patrickhulce:
> how about a timing object:

love `timing` no need to add `run` prefixes IMO 👍

brendankenny:
`timing` sounds great

brendankenny:
I'm adding this to plots/, just sec :)

brendankenny:
plots/ now graphs the total execution time for runs (and can expand as we add other timings)

PR #2245, opened 2017-05-13
Extract chrome-launcher to a standalone thing.

samccone:
🚨  This is a breaking change to any consumer of the launcher 🚨  
----

Pull chrome-launcher into a standalone folder and add a new standalone package.json.

This will enable us to publish chrome-launcher as a standalone module on npm for direct consumption.

PR action:

* Update the path to chrome launcher used by the plot consumer. 
* Add chrome-launcher build / test scripts to all of the existing global mechanisms to enable seemless testing and building
* Add chrome-launcher tests to travis to prevent regressions

----

impacted users: 23~ on github
https://github.com/search?l=JavaScript&q=chrome-launcher.js&type=Code&utf8=%E2%9C%93

----

ref #2092


paulirish:
Nothing too big in this PR, just renames and a move. 

generally lgtm. will fine-tooth it in a bit.

chrome-launcher/test/global-mocha-hooks.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const assert = require('assert');
+
+/**
+ * THIS FILE IS A DUPLICATE OF lighthouse-core/test/global-mocha-hooks.js
```
patrickhulce: do we really need to do this?

patrickhulce:
we could also temporarily add a shim file in chrome-launcher.js `module.exports = require('../chrome-launcher/chrome-launcher');`

chrome-launcher/test/global-mocha-hooks.js
samccone: good question. I defer to @paulirish

samccone:
> assuming the prep for it being standalone is coming later (removing requires into core, logging control, etc)?

That was my thought :)

chrome-launcher/test/global-mocha-hooks.js
paulirish: nah. don't carry it over.

we'll delete these in a week. they haven't provided value in a long while.

samccone: done

samccone:
> we could also temporarily add a shim file in chrome-launcher.js module.exports = require('../chrome-launcher/chrome-launcher');

Since we are changing the API that is going to require a code change by any consumer, not sure we want this old path to work anymore.

PR #2247, opened 2017-05-13
Enhance error wording around busy traces.

paulirish:
to make these more understandable: 

image


lighthouse-core/gather/computed/first-interactive.js
```diff
@@ -26,7 +26,8 @@ const MIN_TASK_CLUSTER_PADDING = 1000;
 const MIN_TASK_CLUSTER_FMP_DISTANCE = 5000;
 
 const MAX_QUIET_WINDOW_SIZE = 5000;
-const TRACE_BUSY_MSG = 'trace was busy the entire time';
+const TRACE_BUSY_MSG = 'The main thread was busy for the entire trace recording. ' +
+   'First Interactive requires the main thread to reach a more idle state.';
```
patrickhulce: nit: 'a more idle state' feels weird, can we say 'idle for several seconds' or something?

paulirish: sgtm. done.

fixed tests too

PR #2257, opened 2017-05-16
README: fix launched chrome example

ebidel:
Also removes calc'ing overall score based on old aggregations.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -160,7 +160,7 @@ assumes you've installed Lighthouse as a dependency (`yarn add --dev lighthouse`
 
 ```javascript
 const lighthouse = require('lighthouse');
-const ChromeLauncher = require('lighthouse/chrome-laucher/chrome-launcher.js');
+const ChromeLauncher = require('lighthouse/chrome-launcher/chrome-launcher.js');
 const Printer = require('lighthouse/lighthouse-cli/printer');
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -160,16 +160,15 @@ assumes you've installed Lighthouse as a dependency (`yarn add --dev lighthouse`
...
 const Printer = require('lighthouse/lighthouse-cli/printer');
 
 function launchChromeAndRunLighthouse(url, flags, config) {
   const launchedChrome = ChromeLauncher.launch();
 
   launchedChrome
     .then(() => lighthouse(url, flags, config)) // Run Lighthouse.
-    .then(results => launchedChrome.kill())
-    .then(() => results) // Kill Chrome and return results.
+    .then(results => launchedChrome.kill().then(() => results)) // Kill Chrome and return results.
```
stramel: Isn't this an anti-pattern? Wouldn't keeping the promise chain flat be better?

samccone:
oops

readme.md
ebidel: `results` isn't in scope. You could stash a var, but I actually think that complicates the example.

stramel: Ah, yeah you would have to return results from it to use it on the next iteration.. gotcha.

Issue #2258, opened 2017-05-16
ChromeLauncher errors


ebidel:
`yarn add lighthouse@^2.x`

Then run this code:

```
const Lighthouse = require('lighthouse');
const ChromeLauncher = require('lighthouse/chrome-launcher/chrome-launcher');

function launchChromeAndRunLighthouse(url, flags, config = null) {
  return ChromeLauncher.launch()                         // Launch Chrome.
    .then(() => Lighthouse(url, flags, config))          // Run Lighthouse.
    .then(results => chrome.kill().then(() => results)); // Kill Chrome, results.
}

const flags = {output: 'json'};
launchChromeAndRunLighthouse('https://example.com', flags).then(results => {
  console.log(results);
});
```

Produces a error for a missing compiled file:

<img width="517" alt="screen shot 2017-05-15 at 7 11 47 pm" src="https://cloud.githubusercontent.com/assets/238208/26088000/26b27b6c-39a9-11e7-8c35-147564699ed6.png">

When I build that file, I still get errors:

<img width="532" alt="screen shot 2017-05-15 at 7 18 08 pm" src="https://cloud.githubusercontent.com/assets/238208/26087961/ee61cfd8-39a8-11e7-9e41-1e5e91fa1184.png">

samccone:
investigating

samccone:
from the root of lighthouse:

```js
const Lighthouse = require('./lighthouse-core/index.js');
const ChromeLauncher = require('./chrome-launcher/chrome-launcher.js');

function launchChromeAndRunLighthouse(url, flags, config = null) {
  let chrome;

  return ChromeLauncher.launch().then(v => {
    chrome = v;
    flags.port = v.port;
    return Lighthouse(url, flags, config);
  }).then(results => {
    return chrome.kill().then(() => results);
  });
}

const flags = {
  output: 'json'
};
launchChromeAndRunLighthouse('https://example.com', flags).then(results => {
  console.log(results);
}).catch(e => console.log(e));
```

--

The other issue was just a publish issue it looks like

samccone:
another option if you want to maybe have port collisions 

![image](https://cloud.githubusercontent.com/assets/883126/26088912/aad61354-39ae-11e7-894a-36d6425f7231.png)

samccone:
over to @paulirish for the publish fix / issue

paulirish:
will fix the publish!

samccone:
[...]

PR #2260, opened 2017-05-16
Docs: Update usage example

paulirish:
picking up from #2258. partial fix for that


docs/recipes/basic-use-via-module/readme.md
```diff
@@ -0,0 +1,7 @@
+# Basic usage of Lighthouse as a node module
+
+Lighthouse's node API is rather basic. You'll likely want to require the chrome-launcher and lighthouse itself.
+
+The return object is rather large, but seek and you shall find. ;)
```
ebidel: remove 2nd "rather"

ebidel:
Kinda worried that these will get out of sync with each other. Do we need a recipe if there's a top level readme?

paulirish:
> Kinda worried that these will get out of sync with each other. Do we need a recipe if there's a top level readme?

no big for me. It was mostly motivated by wanting to make sure the script worked, so I had to actually run it. ;)

all good tho!

PR #2261, opened 2017-05-16
Report: don't display score gauge header if there's only one.

paulirish:
Got tired of looking at tops of my perf/a11y reports that look like this:
image

This PR doesn't show the overview gauges if there's only one.

### DIFF
--- a/lighthouse-core/report/v2/renderer/report-renderer.js
+++ b/lighthouse-core/report/v2/renderer/report-renderer.js
@@ -139,18 +139,17 @@ class ReportRenderer {
    */
   _renderReport(report) {
     const container = this._dom.createElement('div', 'lh-container');
-    const isSoloCategory = report.reportCategories.length === 1;
-    let scoreHeader;
-
     container.appendChild(this._renderReportHeader(report)); // sticky header goes at the top.
     container.appendChild(this._renderReportNav(report));
-
     const reportSection = container.appendChild(this._dom.createElement('div', 'lh-report'));
+
+    let scoreHeader;
+    const isSoloCategory = report.reportCategories.length === 1;
     if (!isSoloCategory) {
       scoreHeader = reportSection.appendChild(this._dom.createElement('div', 'lh-scores-header'));
     }
-    const categories = reportSection.appendChild(this._dom.createElement('div', 'lh-categories'));
 
+    const categories = reportSection.appendChild(this._dom.createElement('div', 'lh-categories'));
     for (const category of report.reportCategories) {
       if (scoreHeader) {
         scoreHeader.appendChild(this._categoryRenderer.renderScoreGauge(category));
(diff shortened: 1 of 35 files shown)


lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -139,18 +139,21 @@ class ReportRenderer {
...
-    const scoreHeader = reportSection.appendChild(
-        this._dom.createElement('div', 'lh-scores-header'));
+    const scoreHeaderClass = `lh-scores-header ${isSoloCategory ? 'lh-scores-header__solo' : ''}`;
+    const scoreHeader = reportSection.appendChild(this._dom.createElement('div', scoreHeaderClass));
 
     const categories = reportSection.appendChild(this._dom.createElement('div', 'lh-categories'));
+
     for (const category of report.reportCategories) {
-      scoreHeader.appendChild(this._categoryRenderer.renderScoreGauge(category));
+      if (!isSoloCategory)
```
brendankenny: I thought we didn't allow no braces multiline if statement

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -139,18 +139,21 @@ class ReportRenderer {
...
+    const isSoloCategory = report.reportCategories.length === 1;
 
     container.appendChild(this._renderReportHeader(report)); // sticky header goes at the top.
     container.appendChild(this._renderReportNav(report));
 
     const reportSection = container.appendChild(this._dom.createElement('div', 'lh-report'));
 
-    const scoreHeader = reportSection.appendChild(
-        this._dom.createElement('div', 'lh-scores-header'));
+    const scoreHeaderClass = `lh-scores-header ${isSoloCategory ? 'lh-scores-header__solo' : ''}`;
```
brendankenny: Can we just not have a score header if < 2 categories?

paulirish: well i was thinking that too, but our styling uses that header's margin to get out from under the pos:fixed header.

lighthouse-core/report/v2/renderer/report-renderer.js
paulirish: au contraire! 

(it was a surprise to me, too)

brendankenny: #2263 :P

lighthouse-core/report/v2/renderer/report-renderer.js
paulirish: ok so the better fix would be to put that margin on something else.
i guess that's the action necessary here.

paulirish:
@brendankenny ptal

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -139,18 +139,22 @@ class ReportRenderer {
    */
   _renderReport(report) {
     const container = this._dom.createElement('div', 'lh-container');
+    const isSoloCategory = report.reportCategories.length === 1;
+    let scoreHeader;
```
brendankenny: move these two lines down to above `if (!isSoloCategory) {`?

brendankenny:
style nit still applies but other than that, still looks good :P

PR #2266, opened 2017-05-16
fix: remove TTI (alpha) from performance category

patrickhulce:
This hides TTI from being displayed but still leaves it in the audits so it will appear in the trace if `--save-assets` and you run with the default or with `onlyAudits: ['time-to-interactive']`

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -675,7 +675,6 @@ module.exports = {
         {"id": "first-meaningful-paint", "weight": 5, "group": "perf-metric"},
         {"id": "speed-index-metric", "weight": 1, "group": "perf-metric"},
         {"id": "estimated-input-latency", "weight": 1, "group": "perf-metric"},
-        {"id": "time-to-interactive", "weight": 5, "group": "perf-metric"},
         {"id": "first-interactive", "weight": 5, "group": "perf-metric"},
         {"id": "consistently-interactive", "weight": 5, "group": "perf-metric"},
         {"id": "link-blocking-first-paint", "weight": 0, "group": "perf-hint"},
--- a/lighthouse-cli/test/smokehouse/perf/expectations.js
+++ b/lighthouse-cli/test/smokehouse/perf/expectations.js
@@ -34,14 +34,8 @@ module.exports = [
           }
         }
       },
-      'time-to-interactive': {
+      'first-interactive': {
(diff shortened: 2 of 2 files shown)


lighthouse-core/config/default.js
```diff
@@ -675,7 +675,6 @@ module.exports = {
         {"id": "first-meaningful-paint", "weight": 5, "group": "perf-metric"},
         {"id": "speed-index-metric", "weight": 1, "group": "perf-metric"},
         {"id": "estimated-input-latency", "weight": 1, "group": "perf-metric"},
-        {"id": "time-to-interactive", "weight": 5, "group": "perf-metric"},
```
paulirish: /home/travis/build/GoogleChrome/lighthouse/lighthouse-cli/test/smokehouse/smokehouse.js:174
      throw new Error(`Config did not trigger run of expected audit ${auditName}`);
      ^
Error: Config did not trigger run of expected audit time-to-interactive
    at auditNames.map.auditName (/home/travis/build/GoogleChrome/lighthouse/lighthouse-cli/test/smokehouse/smokehouse.js:174:13)

patrickhulce: fixed

denar90:
What about https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/traces/pwmetrics-events.js#L134-L144?
Should it be also cleaned up?

patrickhulce:
@denar90 we're not actually removing the logic or the audit until later (probably next week) this just hides it from the report and removes it from `--perf`, still getting the timestamp in the trace for a few more days is intentional :)

denar90:
I'm asking because TTI is not present in audits object which is causing some errors for pwmetrics. I'm trying to integrate the latest master from c908e81 commit.

patrickhulce:
@denar90 yeah that's probably because you're referencing the [perf.json](https://github.com/paulirish/pwmetrics/blob/4070b62ccc0e84c25fd81ce6e3e32c284f8e639b/lib/index.ts#L6) config, if you still want to surface the old TTI you'll have to explicitly include it in your config, otherwise you can ignore errors from missing events like we do https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/traces/pwmetrics-events.js#L212

denar90:
@patrickhulce thanks 👍

PR #2278, opened 2017-05-16
Add details to mutation events audit

ebidel:
Think this is the last one left.

screen shot 2017-05-16 at 1 40 52 pm

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -531,7 +531,6 @@ span.lh-node:hover {
 
 .lh-code {
   text-overflow: ellipsis;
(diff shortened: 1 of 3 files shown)


lighthouse-core/report/v2/report-styles.css
```diff
@@ -531,7 +531,6 @@ span.lh-node:hover {
 
 .lh-code {
   text-overflow: ellipsis;
-  overflow: hidden;
```
patrickhulce: what is it now, auto or visible?

ebidel: auto. 

> can you at least scroll to see the rest of it?

yep, they're scrollable.

PR #2279, opened 2017-05-16
fix: always ensure tracing is off before starting

### LINKED ISSUE #2272: Fix the "tracing already started" bug
Our primary issue for it: #1091 

Current bandaid proposal while we find a reliable repro...   call `Tracing.stop` early and consume any error. 

Caseq agrees this is harmless and worth a try.

patrickhulce:
fixes #2272

### DIFF
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -46,16 +46,16 @@ class Connection {
    * Call protocol methods
    * @param {!string} method
    * @param {!Object} params
-   * @param {boolean=} silent
+   * @param {Object=} options
    * @return {!Promise}
    */
-  sendCommand(method, params = {}, silent = false) {
+  sendCommand(method, params = {}, options) {
     log.formatProtocol('method => browser', {method, params}, 'verbose');
     const id = ++this._lastCommandId;
     const message = JSON.stringify({id, method, params});
     this.sendRawMessage(message);
     return new Promise((resolve, reject) => {
-      this._callbacks.set(id, {resolve, reject, method, silent});
+      this._callbacks.set(id, {resolve, reject, method, options});
     });
   }
 
@@ -99,7 +99,7 @@ class Connection {
 
       return callback.resolve(Promise.resolve().then(_ => {
         if (object.error) {
-          const logLevel = callback.silent ? 'verbose' : 'error';
+          const logLevel = callback.options && callback.options.silent ? 'verbose' : 'error';
           log.formatProtocol('method <= browser ERR', {method: callback.method}, logLevel);
           throw new Error(`Protocol error (${callback.method}): ${object.error.message}`);
         }
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -46,16 +46,16 @@ class Connection {
    * Call protocol methods
    * @param {!string} method
    * @param {!Object} params
-   * @param {Object=} options
+   * @param {{silent: boolean}=} cmdOpts
    * @return {!Promise}
    */
-  sendCommand(method, params = {}, options = {}) {
+  sendCommand(method, params = {}, cmdOpts = {}) {
     log.formatProtocol('method => browser', {method, params}, 'verbose');
     const id = ++this._lastCommandId;
     const message = JSON.stringify({id, method, params});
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/connections/connection.js
```diff
@@ -46,15 +46,16 @@ class Connection {
    * Call protocol methods
    * @param {!string} method
    * @param {!Object} params
+   * @param {boolean=} silent
```
patrickhulce: open to much better names, `silenceError` felt like it would supress the error itself, `silenceErrorLogging` too verbose?

lighthouse-core/gather/driver.js
```diff
@@ -710,9 +711,23 @@ class Driver {
...
+      // ensure tracing is stopped before we can start
+      // see https://github.com/GoogleChrome/lighthouse/issues/1091
+      .then(_ => this.endTraceIfStarted())
       .then(_ => this.sendCommand('Tracing.start', tracingOpts));
   }
 
+  endTraceIfStarted() {
+    return new Promise((resolve) => {
+      const traceCallback = () => resolve();
+      this.once('Tracing.tracingComplete', traceCallback);
```
paulirish: why listen here? what about

```js
this.sendCommand('Tracing.end').then(resolve, resolve);
```

lighthouse-core/gather/driver.js
```diff
@@ -187,9 +187,10 @@ class Driver {
    * Call protocol methods
    * @param {!string} method
    * @param {!Object} params
+   * @param {boolean=} silent
    * @return {!Promise}
    */
-  sendCommand(method, params) {
+  sendCommand(method, params, silent) {
```
paulirish: im not crazy about the boolean trap here.  do we need this at all if we follow my idea below?

lighthouse-core/gather/driver.js
patrickhulce: nah it moves on too quickly and still fails fatally with `tracing already started`, this seemed preferable to an arbitrary timeout

lighthouse-core/gather/driver.js
patrickhulce: yeah regardless of that change the connection is currently hard coded for all messages with an error property to log them even if you don't care about the promise resolution

brendankenny: the boolean trap is pretty gnarly. Before we hardcoded some error message to ignore it...we could do the same again. There could also be some kind of `log` trickery so that this call to `sendCommand` could silence an error it was expecting then turn the silencing off again. Pretty hacky ideas, though...

patrickhulce:
@paulirish @brendankenny switched away from bool PTAL :)

lighthouse-core/gather/connections/connection.js
```diff
@@ -46,15 +46,16 @@ class Connection {
    * Call protocol methods
    * @param {!string} method
    * @param {!Object} params
+   * @param {Object=} options
    * @return {!Promise}
    */
-  sendCommand(method, params = {}) {
+  sendCommand(method, params = {}, options = {}) {
```
paulirish: sg, but can rename to `cmdOpts` or something? so it's clear this is different from our beloved `options` friend?

brendankenny: Should also document it, not just it takes some options :) 

Wonder if it should be something like a regex so it's not just unconditionally ignoring errors. That would resemble the temporary fix we had for silencing domain enable/disable errors, which seemed like a good idea at the time

brendankenny:
wdyt about `ignoreErrorPattern` or something instead of `silent`?

patrickhulce:
> wdyt about ignoreErrorPattern or something instead of silent?

I think it's a slightly better bandaid for the situation at hand but introduces unnecessary complexity for something we want to get rid of in forthcoming catharsis :)

brendankenny:
Wait, what are we removing that would fix the underlying problem? Or are you saying this is a temporary fix for something we want to actually fix?

patrickhulce:
> Wait, what are we removing that would fix the underlying problem?

Oh that the connection class tries to do anything more than verbose logging to begin with! Isn't that what you were proposing anyway?

paulirish:
let's wrap this guy up. i'm OK with merging and giving it a go.. see what happens.

to you, @brendankenny

PR #2282, opened 2017-05-17
Convert code snippets in audit titles

ebidel:
This converts markdown ticks in audit titles.

Esp. makes the a11y audits look better.

screen shot 2017-05-16 at 8 32 43 pm

### DIFF
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -101,7 +101,7 @@ class DOM {
    * @param {string} text
    * @return {!Element}
    */
-  createSpanFromMarkdown(text) {
+  convertMarkdownLinkSnippets(text) {
     const element = this.createElement('span');
 
     // Split on markdown links (e.g. [some link](https://...)).
--- a/lighthouse-core/report/v2/renderer/dom.js
+++ b/lighthouse-core/report/v2/renderer/dom.js
@@ -127,6 +127,28 @@ class DOM {
   }
 
   /**
+   * @param {string} text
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/v2/renderer/dom.js
```diff
@@ -127,6 +127,28 @@ class DOM {
   }
 
   /**
+   * @param {string} text
+   * @return {!Element}
+   */
+  convertMarkdownCodeSnippets(text) {
```
patrickhulce: do you not want to fold this in to `createSpanFromMarkdown`? if not can we rename to `convertMarkdownLinkSnippets` or something?

patrickhulce:
looks sooooo much better ❤️

lighthouse-core/report/v2/renderer/dom.js
ebidel: Not sure that's easy to do. They're different regexs and construct different dom. We might be able to combine, but following the code would be really messy.

Renamed.

ebidel:
PTAL

stramel:
@patrickhulce @ebidel I just ran the report on 2.0.0-alpha.7 and it didn't covert them.

image

paulirish:
aye. this commit isn't in alpha.7.. the latest commit that had was 603c92031eaae868293d5a4300ca61703d4cc67f which was a few before this landed.

stramel:
@paulIrish, Thanks for clarifying! I was pretty sure it just missed the build but wanted to make sure. 

v2 is really starting to look great! :+1:

PR #2283, opened 2017-05-17
fix(loadfast4pwa): ignore requests with startTime < issueTime

### LINKED ISSUE #2281: LoadFast4Pwa is still flaky
There are still cases where load fast for pwa is still flaky, but seems to be only in extension/devtools. It seems to fail to throttle two requests on paulirish.com, one of which is QUIC and is to be expected but the other is just a regular H2 request. The common thread is that the start time seems to be messed up (start time is negative)

Example:
image

patrickhulce:
fixes #2281

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -63,6 +63,12 @@ class LoadFastEnough4Pwa extends Audit {
           return;
         }
 
+        // Disregard requests with an invalid start time, (H2 request start times are sometimes less
+        // than issue time and even negative which throws off timing)
+        if (record._startTime < record._issueTime) {
+          return;
+        }
+
         // Use DevTools' definition of Waiting latency: https://github.com/ChromeDevTools/devtools-frontend/blob/66595b8a73a9c873ea7714205b828866630e9e82/front_end/network/RequestTimingView.js#L164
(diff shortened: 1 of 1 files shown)


lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -63,6 +63,12 @@ class LoadFastEnough4Pwa extends Audit {
           return;
         }
 
+        // Disregard requests with an invalid start time, (H2 request start times are sometimes less
+        // than issue time and even negative which throws off timing)
+        if (record._startTime < record._issueTime) {
```
patrickhulce: checking this requires the checks from above to pass and seemed complicated enough to devote its own if check and comment

brendankenny:
Is this working around a devtools bug?

If there's just one request from one of these origins, we'll end up allowing it to do anything, which seems to go against the intention to warn when we can't tell if it wasn't throttled properly...

lighthouse-core/audits/load-fast-enough-for-pwa.js
paulirish: it wasn't quic was it? we can identify quic requests in here.

paulirish:
landing this now.
we'll also research and followup with dgozman to understand if his throttling change messed with this.

also: im convinced of now passing loadfast4pwa and outputting a debugstring. :)

PR #2246, opened 2017-05-13
ChromeLauncher v1.0

samccone:
📔  note:

> This CL consists of individual commits that build on each other, each commit is an atomic unit of work to make reviewing easier, as well as the commit history understandable.

#### Outline of the work done in this CL.

* Clean up use of any typing for chrome-finder by being smart with types. 🤔  🍎 
* Limit class interface for chrome launcher instance with private. 👮 
* Extract flags to a standalone module since it is more config than code. 🐴 
* Add missing type info for mkdirp. 🔍 
* Extract util functions to a standalone component since they are not coupled to the implementation of chrome launcher. :man_farmer: 
* make use of await / async. 😴 
* update plots to consume the new api. 📈 
* Add a new ability for the launcher to handle SIGINT 🔪 .
* launcher will now if a non-zero port is passed tried to connect to an existing connection there, if no connection is found it will launch chrome with the passed port exposed and carry on. :zero: 
* Made use of the port 0 pattern and moved random port selection into launcher :goal_net: 
* abstract mkdir pattern for multiple platforms 🔡 
* remove autoSelector noise from old api 💀 
* add the ability for a user to pass in a custom chrome path 🍮 

----

ref #2092

PR #2284, opened 2017-05-17
give port to plots for new chromelauncher

brendankenny:
#2246 left this out. #2183 already has this fix, but we may not review it until next week.


brendankenny:
yeah, I figure better to wait for #2183

samccone:
ick, sorry!

PR #2294, opened 2017-05-17
polish: show audits with debug string, don't fail loadfast4pwa on network latencies, works-offline change

patrickhulce:
* doesn't hide passed audits with a debug string in "passed audits" section
* only shows debug string on `works-offline` if the audit failed
* flips loadfast4pwa to pass (but still show debug string) when latencies seem unrealistic
* drive by text fix for loadfast4pwa TTI -> TTFI
image

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -63,6 +63,12 @@ class LoadFastEnough4Pwa extends Audit {
           return;
         }
 
+        // Disregard requests with an invalid start time, (H2 request start times are sometimes less
+        // than issue time and even negative which throws off timing)
+        if (record._startTime < record._issueTime) {
+          return;
+        }
+
         // Use DevTools' definition of Waiting latency: https://github.com/ChromeDevTools/devtools-frontend/blob/66595b8a73a9c873ea7714205b828866630e9e82/front_end/network/RequestTimingView.js#L164
         const latency = record._timing.receiveHeadersEnd - record._timing.sendEnd;
         const latencyInfo = {
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -41,7 +41,7 @@ class LoadFastEnough4Pwa extends Audit {
       category: 'PWA',
       name: 'load-fast-enough-for-pwa',
       description: 'Page load is fast enough on 3G',
-      helpText: 'Satisfied if the Time To Interactive duration is shorter than 10 seconds, as defined by the [PWA Baseline Checklist](https://developers.google.com/web/progressive-web-apps/checklist). Network throttling is required (specifically: RTT latencies >= 150 RTT are expected).',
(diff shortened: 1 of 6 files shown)


paulirish:
can you retitle PR to indicate more than 1 change. especially about the loadfast change.

patrickhulce:
> can you retitle PR to indicate more than 1 change. especially about the loadfast change.

ya, retitled

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -63,12 +63,6 @@ class LoadFastEnough4Pwa extends Audit {
           return;
         }
 
-        // Disregard requests with an invalid start time, (H2 request start times are sometimes less
-        // than issue time and even negative which throws off timing)
-        if (record._startTime < record._issueTime) {
-          return;
-        }
```
paulirish: lets keep this

lighthouse-core/audits/works-offline.js
```diff
@@ -41,14 +41,16 @@ class WorksOffline extends Audit {
    */
   static audit(artifacts) {
     let debugString;
+    const passed = artifacts.Offline === 200;
     if (!URL.equalWithExcludedFragments(artifacts.URL.initialUrl, artifacts.URL.finalUrl)) {
-      debugString = 'WARNING: You may be failing this check because your test URL ' +
-          `(${artifacts.URL.initialUrl}) was redirected to "${artifacts.URL.finalUrl}". ` +
-          'Try testing the second URL directly.';
+      const explanation = passed ? 'Your test URL' :
```
brendankenny: do we want to show this at all if they passed? If they passed, it means that the `initialUrl` successfully redirected to `finalUrl` and `finalUrl` successfully loaded, so it seems like everything is good in that case?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -343,7 +343,8 @@ class CategoryRenderer {
 
     const manualAudits = category.audits.filter(audit => audit.result.manual);
     const nonManualAudits = category.audits.filter(audit => !manualAudits.includes(audit));
-    const passedAudits = nonManualAudits.filter(audit => audit.score === 100);
+    const passedAudits = nonManualAudits.filter(audit => audit.score === 100 &&
+        !audit.result.debugString);
```
brendankenny: 👍

brendankenny: add this to the `renders the passed audits` test, maybe?

lighthouse-core/audits/works-offline.js
patrickhulce: had that originally actually but it makes some sense that we flag that we don't know for sure if the original URL they wanted can be loaded offline, but the message is still a little off, I'll adjust it

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: ya done

lighthouse-core/audits/works-offline.js
patrickhulce: eh you know what, I'll just hide it if it passed, we don't have any actionable information for them in that scenario anyway since LH won't test it

brendankenny:
Oh, there's one unit test issue. LGTM after that :)

PR #2297, opened 2017-05-18
Updated report styles

chowse:
Updated the Lighthouse report's styling to better match the DevTools UI (with options to restyle it using CSS variables).


paulirish:
Before and after

image

chowse:
I'm using a 12px baseline font for this design. For the standalone report, I'd prefer to use a 14px baseline (should only require tweaking a few CSS variables). Just let me know how we'll differentiate a DevTools- and CLI-generated report and I can make the adjustments.

paulirish:
>  Just let me know how we'll differentiate a DevTools- and CLI-generated report 

@chowse we have `.lh-devtools` on the lh-root container in devtools.  

Also the `.lh-narrow` i mentioned before? It's only applied in standalone report.

patrickhulce:
> Also the .lh-narrow i mentioned before? It's only applied in standalone report. So probably not as useful as I indicated before. :)

Adding it to devtools is just a few line change though if we need it

paulirish:
And fwiw here's this PR applied within devtools:

image

chowse:
The same report without the dense DevTools styling:

image

lighthouse-core/report/v2/report-styles.css
```diff
@@ -520,25 +611,33 @@ span.lh-node:hover {
...
 
 .lh-audit-group {
-  padding-top: var(--default-padding);
-  padding-bottom: var(--default-padding);
+  padding-top: var(--lh-audit-group-vpadding);
   border-top: 1px solid var(--report-secondary-border-color);
+  padding-left: var(--expandable-indent);
+}
+
+.lh-audit-group[open] {
```
patrickhulce: does this have any content?

lighthouse-core/report/v2/report-styles.css
```diff
@@ -639,24 +750,20 @@ span.lh-node:hover {
 }
 
 .lh-category .lh-audit {
-  margin-left: calc(var(--default-padding) * 2);
```
patrickhulce: are none of these rules needed anymore then?

lighthouse-core/report/v2/templates.html
```diff
@@ -347,11 +369,11 @@ <h1 class="leftnav__header__title">Lighthouse</h1>
...
     .lh-gauge {
-      --circle-size: 80px;
+      --circle-size: calc(2.5 * var(--header-font-size));
       --circle-size-half: calc(var(--circle-size) / 2);
-      --circle-background: #eee;
-      --circle-border-width: 8px;
-      --inset-size: calc(var(--circle-size) - var(--circle-border-width));
+      --circle-background: #ccc;
+      --circle-border-width: 2px;
+      --inset-size: calc(var(--circle-size) - 2* var(--circle-border-width));
```
patrickhulce: nit: `2 *`

paulirish:
one thing I lose from the previous design is the green encoding of the passing group


image


in this view i'd love to more clearly see that we do have passing items in the group there.. maybe two overlapping green checks on the right of "8 Passed Audits" ?

ev1stensberg:
I can submit a fix to #2385 once this gets merged

brendankenny:
@paulirish do you have a sense of remaining things to resolve? Can we make a list?

paulirish:
Yah. I have a branch pushed with some more work. I'll prep a list of remaining work tonight.

PR #2298, opened 2017-05-18
polish: sizing tweaks for extremely narrow viewports

patrickhulce:
fixes https://bugs.chromium.org/p/chromium/issues/detail?id=723362

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -624,9 +624,15 @@ span.lh-node:hover {
 
 .lh-scores-header {
   display: flex;
+  overflow-x: auto;
   margin: var(--report-header-height) 0 0 0;
   padding: calc(var(--default-padding) * 2) var(--default-padding);
   border-bottom: 1px solid var(--report-border-color);
+}
+
+.lh-categories {
+  width: 100%;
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/v2/report-styles.css
```diff
@@ -739,6 +745,7 @@ summary.lh-passed-audits-summary::-webkit-details-marker,
 @media screen and (max-width: 767px) {
   .lh-report {
     margin-left: 0;
+    width: 100%;
```
paulirish: i think we should one min-width on `lh-report` or similar container...

to avoid this sorta thing

narrow

patrickhulce: sure, it seems to be a grab bag across devtools which panes shrink forever like this and which don't shrink at all and scroll as soon as it's less narrower than content, at what size do we want to stop shrinking?

paulirish: something reasonable. 

i trust your judgement. :)

patrickhulce: added min-width 400px which is enough to cover the header gauges to prevent them from being their own separate scroll box, can probably lower it after #2297 :)

PR #2300, opened 2017-05-18
polish: handle traces with base64 encoded args

patrickhulce:
fixes the `atob is not defined` error on https://plustimer.firebaseapp.com/ that results from the traceviewer trying to parse base64 encoded args embeded in a trace event used by Zone 

actually surprised we haven't run into this yet, it only happens when calling from node

also includes a drive by fix to increase the wait after load to 5250 since I've seen a couple traces where the trace ended 4999 or 4990 ms after quiet when the site was `onLoad` bound


patrickhulce:
PTAL :)

PR #2310, opened 2017-05-19
Fix: Allow template literals in linting

stramel:
Not sure if it was left out for a reason or just missed. Google's default preset has this exact quotes setting so I think it could just be removed.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -35,7 +35,6 @@ module.exports = {
       "argsIgnorePattern": "(^reject$|^_$)",
       "varsIgnorePattern": "(^_$)"
     }],
-    "quotes": [2, "single", {"allowTemplateLiterals": true}],
     "strict": [2, "global"],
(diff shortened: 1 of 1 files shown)


paulirish:
hmm.. is our linting rejecting temp literals now? not sure what diff this is making

paulirish:
(didn't mean to close)

stramel:
@paulirish Hmmm, I misunderstood the point of this command however, this is still the recommended setup for quotes by eslint. Really the only thing it is allowing is the use of backtick without multi-line or variable injection. /shrug I can close or remove the quote property out of the eslint and let the google preset handle it.

wardpeet:
I rather be explicit though. If you don't need the `` don't use it so I'm glad not changing it but if google preset has it we prolly should move to it too.

paulirish:
falling back to the google preset works for me. Thanks.

PR #2312, opened 2017-05-19
Total byte audit reports full URL

johnboxall:
Currently the `total-byte-weight` audit only reports `getURLDisplayName` for URLs.

This can make it challenging to understand what resource the audit is talking about, as sites we work with include 3rd party scripts with mystery names.

Without the complete URL, it is impossible to create detailed CI tests that fail if certain resources are bigger than a threshold. In our case, we often do not control the size of all assets on the page, so we don't want to fail on the overall threshold... just thing we have control on.

This change displays the full URL.

| Before | After |
|---|---|
| image | image|

I used the test page https://lighthouse-url-example.surge.sh/ with a few examples of URLs that we may encounter in the wild.

_With the longer URLs, the styling could probably use some love (I could also imagine terrible ideas like having the full URL available on hover) – happy to pursue that if this is a direction you'd like to head in. Assuming this is a direction you're looking at going in, also happy to update the relevant tests._

❤️

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -72,8 +72,10 @@ class DetailsRenderer {
    * @return {!Element}
    */
   _renderURL(text) {
-    const element = this._renderText(text);
+    const parsedURL = Util.parseURL(text.text);
+    const element = this._renderText({text: parsedURL.file});
     element.classList.add('lh-text__url');
+    element.title = text.text;
     return element;
   }
 
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -77,9 +77,9 @@ class DetailsRenderer {
     let displayedURL;
     let title;
     try {
-      displayedURL = Util.parseURL(text.text).file;
+      displayedURL = Util.parseURL(url).file;
       title = url;
-    } catch (e) {
+    } catch (/** @type {!Error} */ e) {
       if (!(e instanceof TypeError)) {
         throw e;
       }
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -72,10 +72,30 @@ class DetailsRenderer {
    * @return {!Element}
    */
   _renderURL(text) {
-    const parsedURL = Util.parseURL(text.text);
-    const element = this._renderText({text: parsedURL.file});
+    const url = text.text || '';
+
+    let displayedURL;
+    let title;
+    try {
+      displayedURL = Util.parseURL(text.text).file;
+      title = url;
+    } catch (e) {
+      if (!(e instanceof TypeError)) {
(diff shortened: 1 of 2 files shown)


wardpeet:
As I do like the idea we changed it for a reason as we didn't want to show the full url as it is irrelevant in most cases.

Maybe better to work with a toggle button or tooltip.. 

Cc @ebidel @paulirish

paulirish:
@johnboxall thx for the patch! appreciate it.

We've intended to have the full URL there in a tooltip. But based on the patch, it looks like it's not there. :/

johnboxall:
@patrickhulce sure 👍

johnboxall:
👨‍🎨 

image

wardpeet:
LGTM! Just have to fix the linting errors where travis throws on.

johnboxall:
@wardpeet 💁

> Just have to fix the linting errors where travis throws on.

So about that 😛 

1️⃣ It's not clear to me how I'd resolve [this](https://travis-ci.org/GoogleChrome/lighthouse/jobs/236048350#L737) type warning and the related [error](https://travis-ci.org/GoogleChrome/lighthouse/jobs/236048350#L592-L611):

```
/tmp/tmpvGobNX/front_end/audits2/lighthouse/renderer/details-renderer.js:75: WARNING - actual > parameter 1 of Util.parseURL does not match formal parameter
 found   : (string|undefined)
 required: string
     const parsedURL = Util.parseURL(text.text);
                                     ^^^^^^^^^
```
Relaxing the type on `parseURL` seems like the wrong move – I could do something silly like make an intermediate method with a more forgiving type definition ... **Any pointers?**

2️⃣ It's looks like I'll need to [re-generate](https://github.com/GoogleChrome/lighthouse/blob/master/docs/hacking-tips.md#updating-fixture-dumps) `sample_v2.json` to resolve these [errors](https://travis-ci.org/GoogleChrome/lighthouse/jobs/236048350#L413-L589) with something like:

```sh
node lighthouse-cli/test/fixtures/static-server.js
npm run start -- --output=json --output-path=lighthouse-core/test/results/sample_v2.json http://localhost:8080/dobetterweb/dbw_tester.html
```

The diff is large. **Can you confirm this is expected?**

🙏

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -72,8 +72,10 @@ class DetailsRenderer {
    * @return {!Element}
    */
   _renderURL(text) {
-    const element = this._renderText(text);
+    const parsedURL = Util.parseURL(text.text);
```
patrickhulce: we're going to have to try to be backwards compatible here now that we've released

maybe we can make parseURL do this safely, accept `{string|undefined}` and return null if the URL was not parsable for some reason? then in here we can check if we were able to parse and either set the tooltip with the full text or fallback to the original behavior

on the plus side this should address both the linting and the test failures :)

wardpeet:
thanks for the effort @johnboxall.

@patrickhulce wdyt? 😄

paulirish:
Most good. Thanks @johnboxall !

PR #2330, opened 2017-05-22
fix: do not flag blob urls as insecure

### LINKED ISSUE #2323: blob url shouldn't be flagged as insecure
when using Blob and window object url to load it, it flags it as an insecure request
for example:
`var blob = new Blob(['SOME JS WEB WORKER CONTENT']);
var worker = new Worker(window.URL.createObjectURL(blob));
worker.postMessage({});`
This loads a blob url in the browser network requests and is flagged as insecure.

patrickhulce:
fixes #2323

### DIFF
--- a/lighthouse-core/audits/is-on-https.js
+++ b/lighthouse-core/audits/is-on-https.js
@@ -20,7 +20,7 @@ const Audit = require('./audit');
 const Formatter = require('../report/formatter');
 const URL = require('../lib/url-shim');
 
-const SECURE_SCHEMES = ['data', 'https', 'wss', 'chrome', 'chrome-extension'];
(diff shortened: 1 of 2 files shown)


patrickhulce:
> Looks good. Just to confirm, chrome treats than the same way?

yup

```
var debug = {hello: "world"};
var blob = new Blob([JSON.stringify(debug, null, 2)], {type : 'application/json'});
var url = URL.createObjectURL(blob);
fetch(url).then(resp => resp.json()).then(console.log);
> Object {hello: "world"}
fetch('http://same-domain').then(resp => resp.json()).then(console.log);
> Mixed Content: The page at 'https://developer.mozilla.org/en-US/docs/Web/API/Blob' was loaded over HTTPS, but requested an insecure resource 'http://developer.mozilla.org/en-US/docs/Web/API/Blob'. This request has been blocked; the content must be served over HTTPS.
```

PR #2343, opened 2017-05-23
extension: Fix formatting of bug reports

paulirish:
Reusing the refactor [that's on the devtools side](https://codereview.chromium.org/2888263002/diff/50001/third_party/WebKit/Source/devtools/front_end/audits2/Audits2Panel.js).

This fixes the incorrect triple-tick we currently have (with no newlines inbetween).

It also simplifies this bit.

### DIFF
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -60,24 +60,21 @@ function hideRunningSubpage() {
 }
 
 function buildReportErrorLink(err) {
-  let qsBody = '**Lighthouse Version**: ' + getLighthouseVersion() + '\n';
-  qsBody += '**Chrome Version**: ' + getChromeVersion() + '\n';
-
-  if (siteURL) {
-    qsBody += '**URL**: ' + siteURL + '\n';
-  }
-
-  qsBody += '**Error Message**: ' + err.message + '\n';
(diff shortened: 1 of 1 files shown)


lighthouse-extension/app/src/popup.js
```diff
@@ -60,24 +60,21 @@ function hideRunningSubpage() {
...
-  let titleError = err.message;
-
-  if (titleError.length > MAX_ISSUE_ERROR_LENGTH) {
-    titleError = `${titleError.substring(0, MAX_ISSUE_ERROR_LENGTH - 3)}...`;
-  }
-  const title = encodeURI('title=Extension Error: ' + titleError);
-  const body = '&body=' + encodeURI(qsBody);
+  const errorTitle = err.message.substring(0, MAX_ISSUE_ERROR_LENGTH);
+  const title = encodeURI('title=Extension Error: ' + errorTitle);
+  const body = '&body=' + encodeURI(issueBody.trim());
```
patrickhulce: this also needs `encodeURIComponent`

paulirish:
Redid it with URLSearchParams. Gets encoding for free.

Bonus: If any of these characters ("&", "+", and "=") were in error message or stack, they would have broken the issue body (though the report would partially make it). This is fixed now.

PR #2347, opened 2017-05-24
fix: normalize all times to navStart, remove traceviewer model

### LINKED ISSUE #2235: Remove traceviewer dependency
We don't use it for much anymore, and it's very slow at building a model.

WIP branch here: https://github.com/GoogleChrome/lighthouse/tree/rmtraceviewer

a few test failures but it's pretty close.

### LINKED ISSUE #2336: Sites with redirects have wonky stats
I was puzzling over some recent metrics on a plots run of http://www.vkontakte.ru/ which reported traceEnd as 5s shorter than the length of the trace. This is because we take our stats relative to navstart which removes the time of the initial HTTP request, it seems we have inconsistent handling somewhere though because TTCI failed even though there is 5s of quiet.

Example Screenshot
image

patrickhulce:
brings us closer to closing #2235 (still there for stats we can remove separately)

Would like to do some more testing of how this impacts metrics, but the current state is pretty bad with ~10% of sites (just did 100 from alexa top 1000) failing to compute TTCI when it should have been computed (there was a 5s quiet of both network and CPU but due to the inconsistency between traceviewer model and network record timestamps the quiet periods didn't appear to line up in the audit)

~~NOTE: moving to first navStart is optional and can easily be reverted, I wanted to test this out since it still meets the goal of the tests of capturing the correct FMP it just reports longer times which in some cases is definitely the right thing to do and the failure cases are unknown~~ (reverted)


patrickhulce:
did a plots run on this and it brought TTCI failures down from 20% to 13% so that seems like a win, of the remaining 13%, 8% are timeouts and 5% are all for react-hn cases where there's a random long task that is impossible to predict from other signals (it's on a setinterval/settimeout or something) until we have protocol support for CPU quiet

I'm comfortable opening this up for review :)

brendankenny:
what was the plots population size? Runs per site?

can we move the navstart change in a separate PR?

patrickhulce:
🏏 🏏

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -134,12 +129,16 @@ class TraceProcessor {
    * @return {!Array<{percentile: number, time: number}>}
    */
   static getRiskToResponsiveness(tabTrace, startTime, endTime, percentiles) {
-    const events = tabTrace.processEvents.filter(e => e.ts !== 0);
-    const lastEvent = events[events.length - 1];
+    const navStart = tabTrace.navigationStartEvt.ts;
+    const traceStart = tabTrace.processEvents
```
paulirish: should we move these two (traceStart/traceEnd) into tabTrace ?  i'm okay either way.

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -200,22 +190,44 @@ class TraceProcessor {
...
+      if (start > endTime || end < startTime) return events;
+
+      const currentEvt = {
+        start,
+        end,
+        duration: event.dur / 1000,
+        events: [event],
+      };
+
+      if (currentEvt.end < lastEvt.end) {
```
paulirish: can you add this kind of comment?

    // note: processEvents is already sorted by event start

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -200,22 +190,44 @@ class TraceProcessor {
...
+        // The current event is completely inside the previous event, add to previous
+        lastEvt.events.push(event);
+      } else if (currentEvt.start < lastEvt.end) {
```
paulirish: this is possible? 

we're in a single process, though admittedly not a single thread. since we're only looking at these scheduleable tasks, i dont think they should overlap..

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -200,22 +190,44 @@ class TraceProcessor {
...
+        // The current event overlaps with the previous event, extend the duration and add to previous
+        lastEvt.events.push(event);
```
paulirish: are these childEvents being used somewhere?

lighthouse-core/test/lib/traces/tracing-processor-test.js
```diff
@@ -47,25 +47,6 @@ describe('TracingProcessor lib', () => {
     });
   });
 
-  it('doesn\'t throw when user_timing events have a colon', () => {
```
paulirish: why remove?

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: yeah good call

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: yup, that's one of the reasons the event count was so far off from before

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: no but seemed like a bad idea to throw them away

lighthouse-core/test/lib/traces/tracing-processor-test.js
patrickhulce: because we don't ask tracing processor to do anything anymore, I can bring back though I guess it doesn't hurt anything

patrickhulce: wait not nevermind it does hurt something there's no init function at all anymore :)

brendankenny:
what's the expected difference between this and what trace viewer produces today? I'm seeing some pretty different numbers of events

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -241,23 +188,45 @@ class TraceProcessor {
...
+        lastEvt.end = currentEvt.end;
+        lastEvt.duration = currentEvt.end - currentEvt.start;
```
brendankenny: should this be `lastEvt.duration = currentEvt.end - lastEvt.start;` (or `currentEvt.end - currentEvt.start`) if extending duration by the end of `currentEvt` (but keeping the same `start`)?

patrickhulce:
> I'm seeing some pretty different numbers of events

how different is "pretty different"? many of the differences I saw originally in Paul's branch were eliminated by handling the overlapping events, there are likely more edge cases we're not handling but this fixes the ~7% of *definitely* wrong interpretations of the trace I was seeing. Hard to do more comparisons of what's "right" since no ground truth exists

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: heh made the same mistake as in your or :) yeah `lastEvt.end - lastEvt.start` or `currentEvt.end - lastEvt.start`

patrickhulce: added tests for this func :)

brendankenny:
> how different is "pretty different"?

I called `getMainThreadTopLevelEventDurations` on the same trace using master and the new implementation. Durations should cancel out timing changes due to the navStart shift.

In the new implementation I got +35% number of top level events and some pretty different timings, though some of that may be due to it being difficult to tell which events match up and which are in the new 35% due to their being so many events.

But there were also some missing from the new version, and they could make a significant difference in our metrics. e.g. here's all event durations over 100ms:

    master: 105.71, 116.33, 144.134, 191.014, 239.333, 391.718, 505.825, 599.512, 2563.194
    #2347:  105.71, 116.33, 144.134, 191.014, 239.333,          505.825,          2563.194

I can dig in more and look at the actual events behind these missing durations to see where they're going.

> Hard to do more comparisons of what's "right" since no ground truth exists

This was extracting all of the top level events in the trace (using `-Infinity` and `Infinity` as `startTime` and `endTime`). I would expect trace viewer to handle that correctly and we could treat that as ground truth, but are there issues with that I'm missing?

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: > heh made the same mistake as in your or

lol

lighthouse-core/lib/traces/tracing-processor.js
paulirish: you're saying you have traces where these do overlap? can we look at them?

lighthouse-core/lib/traces/tracing-processor.js
paulirish: kinda feel like we should throw them away since we have no use for them.

lighthouse-core/test/lib/traces/tracing-processor-test.js
paulirish: ah true! :) can move this to `user-timing-test.js` to make sure we can inject a trace with these colons and everything goes through fine?

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: yeah the test fixtures do `fixtures/traces/progressive-app.json`

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -98,7 +98,11 @@ class TraceOfTab extends ComputedArtifact {
       .filter(e => e.pid === startedInPageEvt.pid)
       .sort((event0, event1) => event0.ts - event1.ts);
 
-    const traceEnd = trace.traceEvents.reduce((max, evt) => {
+    const earliestTraceEvt = trace.traceEvents.reduce((min, evt) => {
```
patrickhulce: done

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -124,7 +128,9 @@ class TraceOfTab extends ComputedArtifact {
       timings,
       timestamps,
       processEvents,
```
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -241,23 +188,45 @@ class TraceProcessor {
...
-    return mainThread.sliceGroup.slices.filter(slice => {
-      return slice.title === SCHEDULABLE_TASK_TITLE &&
-          slice.end > startTime &&
-          slice.start < endTime;
-    });
+  static getMainThreadTopLevelEvents(tabTrace, startTime = -Infinity, endTime = Infinity) {
+    let lastEvt = {start: -Infinity, end: -Infinity};
+    const topLevelEvents = [];
+    // note: processEvents is already sorted by event start
+    for (const event of tabTrace.processEvents) {
```
patrickhulce: done

lighthouse-core/test/lib/traces/tracing-processor-test.js
patrickhulce: donezo

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -177,57 +121,60 @@ class TraceProcessor {
...
-  static getRiskToResponsiveness(model, trace, startTime, endTime, percentiles) {
+  static getRiskToResponsiveness(tabTrace, startTime, endTime, percentiles) {
+    const navStart = tabTrace.navigationStartEvt.ts;
+    const traceStart = tabTrace.earliestTraceEvt.ts;
+    const traceEnd = tabTrace.latestTraceEvt.ts;
+
     // Range of responsiveness we care about. Default to bounds of model.
-    startTime = startTime === undefined ? model.bounds.min : startTime;
-    endTime = endTime === undefined ? model.bounds.max : endTime;
+    startTime = startTime === undefined ? (traceStart - navStart) / 1000 : startTime;
```
brendankenny: the `- navStart` is somewhat surprising here. Maybe add a comment explaining that they need to be shifted too? Or could use `tabTrace.timings.traceEnd` and `tabTrace.timings.traceStart` (if it was added to `trace-of-tab`) to get the already computed bounds

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -241,23 +188,31 @@ class TraceProcessor {
...
+
+      const start = (event.ts - tabTrace.navigationStartEvt.ts) / 1000;
+      const end = (event.ts + event.dur - tabTrace.navigationStartEvt.ts) / 1000;
+      if (start > endTime || end < startTime) continue;
+
+      topLevelEvents.push({
+        start,
+        end,
+        duration: event.dur / 1000,
+        events: [event],
```
brendankenny: since it'll be just the one event now, drop the array and just have `event`? (or `rawEvent` or whatever...also may not really be needed)

lighthouse-core/test/audits/estimated-input-latency-test.js
```diff
@@ -37,7 +37,7 @@ describe('Performance: estimated-input-latency audit', () => {
     const artifacts = generateArtifactsWithTrace({traceEvents: pwaTrace});
     return Audit.audit(artifacts).then(output => {
       assert.equal(output.debugString, undefined);
-      assert.equal(output.rawValue, 16.4);
+      assert.equal(output.rawValue, 16.2);
```
brendankenny: this and the TTI change, are they just from the navStart shift?

lighthouse-core/test/gather/computed/first-interactive-test.js
```diff
@@ -43,7 +44,14 @@ describe('FirstInteractive computed artifact:', () => {
 
   it('should compute firstInteractive', () => {
     return computedArtifacts.requestFirstInteractive(acceptableTrace).then(output => {
-      assert.equal(Math.round(output.timeInMs), 1587);
+      assert.equal(Math.round(output.timeInMs), 1582);
```
brendankenny: also just from navStart shift?

lighthouse-core/test/lib/traces/tracing-processor-test.js
```diff
@@ -205,20 +188,63 @@ describe('TracingProcessor lib', () => {
...
+      const ret = TracingProcessor.getMainThreadTopLevelEventDurations(tabTrace);
       const durations = ret.durations;
 
       function getDurationFromIndex(index) {
         return Number(durations[index].toFixed(2));
       }
 
       assert.equal(durations.filter(dur => isNaN(dur)).length, 0, 'NaN found');
-      assert.equal(durations.length, 652);
+      assert.equal(durations.length, 651);
```
brendankenny: why the change? Seems like it shouldn't lose any events

lighthouse-core/audits/time-to-interactive.js
```diff
@@ -191,7 +190,7 @@ class TTIMetric extends Audit {
       }
 
       const times = {fmpTiming, visuallyReadyTiming, traceEndTiming};
-      const data = {tabTrace, model, trace};
+      const data = {tabTrace, trace};
```
patrickhulce: or we can just kill the audit entirely :)

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -87,12 +87,8 @@ class EstimatedInputLatency extends Audit {
...
     const trace = artifacts.traces[this.DEFAULT_PASS];
 
-    const pending = [
-      artifacts.requestTraceOfTab(trace),
-      artifacts.requestTracingModel(trace)
-    ];
-    return Promise.all(pending).then(([tabTrace, model]) => {
-      return EstimatedInputLatency.calculate(tabTrace, model, trace);
+    return artifacts.requestTraceOfTab(trace).then(tabTrace => {
+      return EstimatedInputLatency.calculate(tabTrace);
```
patrickhulce: done

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -177,57 +121,60 @@ class TraceProcessor {
   /**
    * Calculates the maximum queueing time (in ms) of high priority tasks for
    * selected percentiles within a window of the main thread.
-   * @see https://docs.google.com/document/d/1b9slyaB9yho91YTOkAQfpCdULFkZM9LqsipcX3t7He8/preview
-   * @param {!traceviewer.Model} model
-   * @param {{traceEvents: !Array<!Object>}} trace
+   * @see https://docs.google.com/document/d/18gvP-CBA2BiBpi3Rz1I1ISciKGhniTSZ9TY0XCnXS7E/preview
```
patrickhulce: yeah probably a bad rebase from the earlier branch, I think paul's been on the hunt for this for a while ;) fixed

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: better question, should we just default these bounds to navstart and traceEnd now instead? negative start time doesn't really make a whole lot of sense and we take it to be the beginning of time everywhere we consume this data

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -177,57 +121,60 @@ class TraceProcessor {
...
+   * Provides durations in ms of all main thread top-level events
+   * @param {!TraceOfTabArtifact} tabTrace
    * @param {number} startTime Optional start time (in ms) of range of interest. Defaults to trace start.
    * @param {number} endTime Optional end time (in ms) of range of interest. Defaults to trace end.
    * @return {{durations: !Array<number>, clippedLength: number}}
    */
-  static getMainThreadTopLevelEventDurations(model, trace, startTime, endTime) {
-    const slices = TraceProcessor.getMainThreadTopLevelEvents(model, trace, startTime, endTime);
+  static getMainThreadTopLevelEventDurations(tabTrace, startTime = -Infinity, endTime = Infinity) {
+    const topLevelTasks = TraceProcessor.getMainThreadTopLevelEvents(tabTrace, startTime, endTime);
```
patrickhulce: sgtm

lighthouse-core/lib/traces/tracing-processor.js
patrickhulce: just nuked it we don't use it I forgot to remove

lighthouse-core/audits/time-to-interactive.js
brendankenny: haha, good point. I'm fine with leaving as is and deleting it next :)

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -107,7 +114,7 @@ class TraceOfTab extends ComputedArtifact {
       firstPaint,
       firstContentfulPaint,
       firstMeaningfulPaint,
-      traceEnd,
+      traceEnd: latestTraceEvt,
```
patrickhulce: I had a very good explanation for why I named them this instead of traceEnd (that these were supposed to be just about the main thread, `latestMainThreadEvt` `earliestMainThreadEvt` and distinct from `traceEnd`) but the rest of the code betrays me. In fact, there are a number of things in here that I strongly feel like I accidentally dropped a stash...

patrickhulce: yeah what do you think about using navStart and traceEnd timings instead of all this other nonsense entirely?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -38,7 +38,7 @@ class TraceOfTab extends ComputedArtifact {
 
   /**
    * @param {{traceEvents: !Array}} trace
-   * @return {!{processEvents: !Array<TraceEvent>, startedInPageEvt: TraceEvent, navigationStartEvt: TraceEvent, firstContentfulPaintEvt: TraceEvent, firstMeaningfulPaintEvt: TraceEvent}}
+   * @return {!TraceOfTabArtifact}
```
patrickhulce: sure

lighthouse-core/test/audits/estimated-input-latency-test.js
patrickhulce: yep, can confirm. the trace for this one now skips the first 15ms and the TTI one skips the first 5ms, adding back this delta yields the old results

lighthouse-core/test/gather/computed/first-interactive-test.js
patrickhulce: yep

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: > better question, should we just default these bounds to navstart and traceEnd now instead? negative start time doesn't really make a whole lot of sense and we take it to be the beginning of time everywhere we consume this data

ha, sorry I was too slow to respond :)

I think it would be better to still have `getMainThreadTopLevelEvents` and `getMainThreadTopLevelEventDurations` return all events, even ones before navStart.

Our current uses don't need anything before then, but we'll hopefully be doing more sophisticated trace analysis in the future (e.g. over multiple navigations, though we'll have several other complications to figure out before then :) and I think having them return all of the main thread/top level events in the trace will be what you would expect from them.

Having a note that the times are relative to navStart (so negative times are just before that) doesn't seem too confusing.

brendankenny: - I guess on the other hand we could say something like "use a `startTime` of `-Infinity` to get all top level events", but that's pretty awkward
- On the other other hand, getting everything after navStart is our typical use case right now, so it's kind of awkward that you have to specify 0 for that.
- On the original hand, though, it still makes sense to me that these would return all of the top level events in the main thread of the trace when no timing bounds are specified, so idk :)

brendankenny: > getting everything after navStart is our typical use case right now

though that's not completely true. It looks like all of our calls have to deal with events earlier than they care about anyways, so it doesn't seem harmful to default to all events, even before navStart

patrickhulce: I gotta disagree that it makes sense to return everything. The entire rest of the LH world completely ignores time before navstart and having this one spot that doesn't do that is what led to total failure on 10% of sites. Having them return everything would require *every* caller to filter out the negative events or provide a non-default parameter at which point why bother having a default at all?

brendankenny: but they do all filter out negative events or provide their own `startTime`. 

> why bother having a default at all?

start of trace/end of trace defaults are no defaults in a way, since it's pretty much just letting the trace you passed in define the bounds.

I guess it comes down to what these methods are for, just TTI-like things or any sort of processing where you need to extract the top level events.

idk, @paulirish, as a user, what would you expect from these methods?

paulirish: i think getRiskToResponsiveness makes sense to be aware of navstart, whereas getMainThreadTopLevelEventDurations shouldnt. that seems to be the state of things here, so i'm pretty good.

I would appreciate a comment being left about the shifting for navstart. (it seems like the jsdoc comment is incorrect now)

brendankenny: OK, I'm good with what we have here. I'll file an issue after this PR with a proposal to separate concerns slightly but still make sure all our audits can easily do the right thing by default (don't pay attention to events before navStart. This works well for all our current audits, though.

lighthouse-core/test/lib/traces/tracing-processor-test.js
```diff
@@ -205,20 +188,63 @@ describe('TracingProcessor lib', () => {
...
+      const ret = TracingProcessor.getMainThreadTopLevelEventDurations(tabTrace);
       const durations = ret.durations;
 
       function getDurationFromIndex(index) {
         return Number(durations[index].toFixed(2));
       }
 
       assert.equal(durations.filter(dur => isNaN(dur)).length, 0, 'NaN found');
-      assert.equal(durations.length, 652);
+      assert.equal(durations.length, 645);
```
brendankenny: can you confirm this one as well? Just the new navStart default?

brendankenny:
need a more recent rebase? travis is still complaining about javascriptDialog

paulirish:
Ya I didn't rebase this one.

lighthouse-core/test/lib/traces/tracing-processor-test.js
patrickhulce: 6 of them definitely the 7th is eluding me, trying to track it down

patrickhulce: ah found the 7th it's a result of `!dur` they return an event that has no duration which we'll ignore

patrickhulce:
🎉 🎉

brendankenny:
with this change, the compiled lighthouse-background.js just dropped by 562KB, around 25%

PR #2349, opened 2017-05-24
Report app manifest errors

### LINKED ISSUE #2348: 'Unable to retrieve manifest' when manifest file is available but invalid
The manifest for my site was actually invalid, but the Lighthouse report looks as if the manifest file couldn't be retrieved:

image

paulirish:
fixes #2348

If the manifest has JSON errors, it currently fails with something like this:

But we do have more useful errors coming from protocol we could share... for example.

This PR implements that.
Also driveby to remove superfluous word "required"

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -267,9 +267,17 @@ class Driver {
   getAppManifest() {
     return this.sendCommand('Page.getAppManifest')
       .then(response => {
-        // We're not reading `response.errors` however it may contain critical and noncritical
-        // errors from Blink's manifest parser:
-        //   https://chromedevtools.github.io/debugger-protocol-viewer/tot/Page/#type-AppManifestError
+        // handle errors
+        if (response.errors) {
+          response.errors.forEach(error => {
+            log.warn('driver', 'Manifest error', error.message);
+          });
+          const criticalError = response.errors.find(error => error.critical);
+          if (criticalError) {
+            return Promise.reject(new Error(`Manifest error: ${criticalError.message}`));
+          }
+        }
+        // handle failed manifest request or 404
         if (!response.data) {
           // If the data is empty, the page had no manifest.
           return null;
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -267,19 +267,20 @@ class Driver {
   getAppManifest() {
     return this.sendCommand('Page.getAppManifest')
       .then(response => {
-        // handle errors
+        // Handle JSON syntax errors and manifest correctness warnings
         if (response.errors) {
+          // Possible issues: https://cs.chromium.org/search/?q=f:manifest_parser.cc+AddErrorInfo&type=cs
           response.errors.forEach(error => {
             log.warn('driver', 'Manifest error', error.message);
           });
+          // Possible critical issues: invalid JSON (location reported) or manifest root is an array
           const criticalError = response.errors.find(error => error.critical);
           if (criticalError) {
-            return Promise.reject(new Error(`Manifest error: ${criticalError.message}`));
+            throw new Error(`Manifest error: ${criticalError.message}`);
           }
         }
-        // handle failed manifest request or 404
+        // Handle failed manifest request or 404
         if (!response.data) {
-          // If the data is empty, the page had no manifest.
           return null;
         }
 
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -189,7 +189,7 @@ class Runner {
           log.warn('Runner', `${artifactName} gatherer, required by audit ${audit.meta.name},` +
             ` encountered an error: ${artifactError.message}`);
           throw new Error(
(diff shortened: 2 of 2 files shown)


brendankenny:
a dbw_tester bad manifest might not be a terrible idea

brendankenny:
> I don't know what the protocol considers critical errors though.

at [first glance](https://cs.chromium.org/chromium/src/content/renderer/manifest/manifest_parser.cc?l=447&gs=cpp%253Acontent%253A%253Aclass-ManifestParser%253A%253AAddErrorInfo(const%2Bstd%253A%253Abasic_string%253Cchar%252C%2Bstd%253A%253Achar_traits%253Cchar%253E%252C%2Bstd%253A%253Aallocator%253Cchar%253E%2B%253E%2B%2526%252C%2Bbool%252C%2Bint%252C%2Bint)%2540chromium%252F..%252F..%252Fcontent%252Frenderer%252Fmanifest%252Fmanifest_parser.cc%257Cdef&gsn=AddErrorInfo&ct=xref_usages) it appears to only be used for JSON errors, but maybe there are errors higher up the chain that are also labeled critical? If not, not sure how we were labelling JSON parse errors as "unable to retrieve"...

lighthouse-core/gather/driver.js
```diff
@@ -277,9 +277,17 @@ class Driver {
     return new Promise((resolve, reject) => {
       this.sendCommand('Page.getAppManifest')
         .then(response => {
-          // We're not reading `response.errors` however it may contain critical and noncritical
-          // errors from Blink's manifest parser:
-          //   https://chromedevtools.github.io/debugger-protocol-viewer/tot/Page/#type-AppManifestError
+          // handle errors
```
brendankenny: maybe make this explicit and say `// Handle JSON syntax errors` or something?

lighthouse-core/gather/driver.js
```diff
@@ -277,9 +277,17 @@ class Driver {
...
+          if (response.errors) {
+            response.errors.forEach(error => {
+              log.warn('driver', 'Manifest error', error.message);
```
brendankenny: not sure if we want to warn on all of these (if it's not a critical warning the manifest is still functional, and we don't warn on something like all uncaught exceptions), but don't feel strongly :)

lighthouse-core/gather/driver.js
```diff
@@ -277,9 +277,17 @@ class Driver {
...
+            });
+            const criticalError = response.errors.find(error => error.critical);
+            if (criticalError) {
+              return reject(new Error(`Manifest error: ${criticalError.message}`));
```
brendankenny: while you're in here, do you want to flatten this promise? Rather than wrapping, can just `return this.sendCommand()...` and these `return reject(new Error())`s can just be `throw new Error()`

paulirish: patrick did this mostly in #2549

lighthouse-core/gather/driver.js
paulirish: i prefer warning on all for now. can quiet it if its annoying. But right now the repeated debugStrings in the report are the more annoying thing.

lighthouse-core/gather/driver.js
paulirish: sg. also detailed what issues these deal with.

paulirish:
updated. @brendankenny ptal

lighthouse-core/gather/driver.js
```diff
@@ -267,11 +267,20 @@ class Driver {
...
+          response.errors.forEach(error => {
+            log.warn('driver', 'Manifest error', error.message);
+          });
+          // Possible critical issues: invalid JSON (location reported) or manifest root is an array
+          const criticalError = response.errors.find(error => error.critical);
+          if (criticalError) {
+            throw new Error(`Manifest error: ${criticalError.message}`);
+          }
+        }
+        // Handle failed manifest request or 404
```
brendankenny: this doesn't usually work (or even get this far), fyi. DevTools will happily try to parse a 404 response like "Cannot GET /no-manifest.json" and return a JSON.parse error instead of the fact that it 404ed

paulirish: ah! good call. thanks.

lighthouse-core/gather/driver.js
```diff
@@ -267,11 +267,20 @@ class Driver {
...
+        // Handle JSON syntax errors and manifest correctness warnings
+        if (response.errors) {
+          // Possible issues: https://cs.chromium.org/search/?q=f:manifest_parser.cc+AddErrorInfo&type=cs
+          response.errors.forEach(error => {
+            log.warn('driver', 'Manifest error', error.message);
+          });
+          // Possible critical issues: invalid JSON (location reported) or manifest root is an array
+          const criticalError = response.errors.find(error => error.critical);
+          if (criticalError) {
+            throw new Error(`Manifest error: ${criticalError.message}`);
```
patrickhulce: We really don't want errors like these reported to Sentry, but this does seem like a valid thing to display as a fatal error for an audit, I guess we should bring back `ExpectedError`?

PR #2351, opened 2017-05-24
fix(connection): gracefully handle missing method

### LINKED ISSUE #2321: Lighthouse 2.0.0 crash
Tested with multiple URLs, all crash. 

OS: macOS 10.12.4

```
$ node -v && npm -v
v6.10.0
3.10.10
$ lighthouse --version
2.0.0
$ lighthouse http://google.com
  ChromeLauncher Waiting for browser. +0ms
  ChromeLauncher Waiting for browser... +1ms
  ChromeLauncher Waiting for browser..... +514ms
  ChromeLauncher Waiting for browser.....✓ +3ms
  Lighthouse:warn The URL provided should be on HTTPS +422ms
  Lighthouse:warn Performance stats will be skewed redirecting from HTTP to HTTPS. +1ms
  status Initializing… +510ms
/usr/local/lib/node_modules/lighthouse/lighthouse-core/lib/log.js:111
    const maxLength = columns - data.method.length - prefix.length - loggingBufferColumns;
                                           ^

TypeError: Cannot read property 'length' of undefined
    at Function.formatProtocol (/usr/local/lib/node_modules/lighthouse/lighthouse-core/lib/log.js:111:44)
    at CriConnection.handleRawMessage (/usr/local/lib/node_modules/lighthouse/lighthouse-core/gather/connections/connection.js:111:9)
    at WebSocket.ws.on.data (/usr/local/lib/node_modules/lighthouse/lighthouse-core/gather/connections/cri.js:78:37)
    at emitTwo (events.js:106:13)
    at WebSocket.emit (events.js:191:7)
    at Receiver.ontext (/usr/local/lib/node_modules/lighthouse/node_modules/ws/lib/WebSocket.js:841:10)
    at /usr/local/lib/node_modules/lighthouse/node_modules/ws/lib/Receiver.js:536:18
    at Receiver.applyExtensions (/usr/local/lib/node_modules/lighthouse/node_modules/ws/lib/Receiver.js:371:5)
    at /usr/local/lib/node_modules/lighthouse/node_modules/ws/lib/Receiver.js:508:14
    at Receiver.flush (/usr/local/lib/node_modules/lighthouse/node_modules/ws/lib/Receiver.js:347:3)
```

patrickhulce:
fixes #2321 

also observed errors in devtools when we get command callbacks for commands LH didn't send

### DIFF
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -92,7 +92,7 @@ class Connection {
     const object = JSON.parse(message);
     // Remote debugging protocol is JSON RPC 2.0 compiant. In terms of that transport,
     // responses to the commands carry "id" property, while notifications do not.
-    if (object.id) {
+    if (this._callbacks.has(object.id)) {
       const callback = this._callbacks.get(object.id);
       this._callbacks.delete(object.id);
 
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -107,6 +107,8 @@ class Connection {
         return object.result;
       }));
     } else if (object.id) {
+      // In DevTools we receive responses to commands we did not send which we cannot act on, so we
+      // just log these occurrences.
       const error = object.error && object.error.message;
       log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
           {method: object.method, params: error || object.result}, 'verbose');
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -106,11 +106,15 @@ class Connection {
           {method: callback.method, params: object.result}, 'verbose');
         return object.result;
       }));
+    } else if (object.id) {
+      const error = object.error && object.error.message;
+      log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
+          {method: object.method, params: error || object.result}, 'verbose');
+    } else {
+      log.formatProtocol('<= event',
+          {method: object.method, params: object.params}, 'verbose');
+      this.emitNotification(object.method, object.params);
     }
-
-    log.formatProtocol('<= event',
-        {method: object.method, params: object.params}, 'verbose');
-    this.emitNotification(object.method, object.params);
   }
 
   /**
--- a/lighthouse-core/lib/log.js
+++ b/lighthouse-core/lib/log.js
@@ -108,11 +108,12 @@ class Log {
    */
   static formatProtocol(prefix, data, level) {
     const columns = (!process || process.browser) ? Infinity : process.stdout.columns;
-    const maxLength = columns - data.method.length - prefix.length - loggingBufferColumns;
+    const method = data.method || '?????';
+    const maxLength = columns - method.length - prefix.length - loggingBufferColumns;
     // IO.read blacklisted here to avoid logging megabytes of trace data
-    const snippet = (data.params && data.method !== 'IO.read') ?
+    const snippet = (data.params && method !== 'IO.read') ?
       JSON.stringify(data.params).substr(0, maxLength) : '';
-    Log._logToStdErr(`${prefix}:${level || ''}`, [data.method, snippet]);
+    Log._logToStdErr(`${prefix}:${level || ''}`, [method, snippet]);
   }
 
   static log(title, ...args) {


patrickhulce:
> Just a bit unsure about the ????? method name when it's unknown.

Yeah using `unknown` just blended in a bit too much with the rest of the logs and this makes it a little clearer something weird is going on, open to suggestions though :)

brendankenny:
did anyone have any more insight into the cause of this? We don't want to crash in the faces of this, but without error tracking this will start to get lost with this change. The devtools case is more understandable, but feedback commands we didn't send is something we should probably track down

patrickhulce:
> did anyone have any more insight into the cause of this?

Nope :) it seems impossible from the stack trace in the issue since if it fails in the resolve it should have definitely failed the first time when it was sending the message (because it actually gets the method when the callback was set)

> but without error tracking this will start to get lost with this change

Good thing we're approved for sentry now let's get it installed post-branch :D

lighthouse-core/gather/connections/connection.js
```diff
@@ -92,7 +92,7 @@ class Connection {
     const object = JSON.parse(message);
     // Remote debugging protocol is JSON RPC 2.0 compiant. In terms of that transport,
     // responses to the commands carry "id" property, while notifications do not.
-    if (object.id) {
+    if (this._callbacks.has(object.id)) {
```
brendankenny: what do you think about making this more explicit than treating it like an event if no callback (it's not really an event since the message does have an `id`)? Could do a third case with a call to `formatProtocol` with an explicit `?????` (and no emit).

Also with level `error` so it hits `log.error`? Not sure about that one if this is a regular occurrence in devtools land

brendankenny:
> Good thing we're approved for sentry now let's get it installed post-branch :D

file an issue in that case? We don't want to forget that we're silencing this

lighthouse-core/gather/connections/connection.js
patrickhulce: sure

patrickhulce:
@brendankenny filed #2364

lighthouse-core/gather/connections/connection.js
```diff
@@ -106,11 +106,15 @@ class Connection {
           {method: callback.method, params: object.result}, 'verbose');
         return object.result;
       }));
+    } else if (object.id) {
+      const error = object.error && object.error.message;
```
brendankenny: add a comment about why this might happen (pavel's old `"while notifications do not"` comment above probably should also be updated as it implies only two states)

lighthouse-core/gather/connections/connection.js
```diff
@@ -106,11 +106,15 @@ class Connection {
...
+      log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
+          {method: object.method, params: error || object.result}, 'verbose');
```
brendankenny: normalize `object.method` here rather than in `formatProtocol`?

brendankenny: if it's an error, `params` probably shouldn't be set to the message?

brendankenny: and should it still throw if there's an error?

lighthouse-core/gather/connections/connection.js
```diff
@@ -106,11 +106,15 @@ class Connection {
           {method: callback.method, params: object.result}, 'verbose');
         return object.result;
       }));
+    } else if (object.id) {
+      const error = object.error && object.error.message;
+      log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
```
brendankenny: `disowned` may not be a good word here (sounds like LH did send the command but doesn't care about it now :), but not sure of a good replacement.

lighthouse-core/gather/connections/connection.js
patrickhulce: > normalize object.method here rather than in formatProtocol?

I'd really rather not, that's exactly what got us into this mess of fatally erroring on unexpected input, unless I'm misunderstanding what you mean by "normalize" haha

> if it's an error, params probably shouldn't be set to the message?

where would you like to log the message then?

> and should it still throw if there's an error?

that would make these fatal in the future which seems bad. the point is that we don't care about these "disowned" (better name pending you 😄 ) events

lighthouse-core/gather/connections/connection.js
patrickhulce: done

lighthouse-core/gather/connections/connection.js
brendankenny: `nonrequested`?

lighthouse-core/gather/connections/connection.js
brendankenny: >> normalize object.method here rather than in formatProtocol?

> I'd really rather not, that's exactly what got us into this mess of fatally erroring on unexpected input, unless I'm misunderstanding what you mean by "normalize"

I mean `const method = object.method || '?????';` here instead of in `formatProtocol` *just* for the logging, not for how it's handled.

> where would you like to log the message then?

For better or worse, `formatProtocol` expects a `data.method` and `data.params`, so it's better to be strict in sending that than to just let it support whatever. `formatProtocol`'s existence now is kind of dumb (all it's really doing beyond `log.verbose` is clipping to the end of the line), but rather than eliminating, maybe we can just quickly revise the parameters rather than overload them?

> that would make these fatal in the future which seems bad

yeah, makes sense.

Issue #2217, opened 2017-05-10
Optimized images: compare quality/size of Chrome-encoded images to other encoders


patrickhulce:
The [current method](https://github.com/GoogleChrome/lighthouse/blob/3f22558aca0285b8b05585d4e1fa2597ef318ae1/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js#L41) used to determine if an image is optimized is to get the resulting data URI of running the image through a `canvas` in the browser at a particular quality setting. However, I've observed that this quality setting does not necessarily track that of other encoders (such as the one used in GIMP for example, their ~90 is our ~85).

It'd be nice to compare and contrast the results of canvas encoding with popular encoders to inform our quality decisions and make sure we're being reasonable. Automating this so we can track it over time for regressions would be great too.

brendankenny:
for jpg, from my reading we [pass off to `libjpeg-turbo` with a simple 0.8 -> 80% quality conversion](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/image-encoders/JPEGImageEncoder.cpp?gsn=Encode&l=192), so this might simplify to comparing `libjpeg-turbo`'s quality levels to popular image editors, `ImageOptim` defaults, etc

jpg encoding might be platform specific, but I believe this covers desktop, at least

patrickhulce:
A recent [skia roll](https://chromium.googlesource.com/chromium/src/+/86074971b9640dbf59869071b4a18bec51c61424) has changed this, it looks like there are [3 possibilities](https://cs.chromium.org/chromium/src/ui/gfx/codec/jpeg_codec.cc?type=cs&l=47) it will delegate to?

paulirish:
@patrickhulce is this all set with trace-event based details?

patrickhulce:
the comparison is still worth knowing IMO to image opt even if it's fairly low priority

PR #2354, opened 2017-05-24
fix: bump jpeg quality

patrickhulce:
related #2217 

at some point recently the JPEG library used by chrome has changed so that the quality or file size is now much lower for the same quality setting which is causing smoke tests to fail if they don't use the cached download of Chrome (see https://travis-ci.org/GoogleChrome/lighthouse/jobs/235694158)

this bumps it to give roughly the same output as it does under Chrome stable but we should investigate this more and try to bisect for the cause or track it better

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -51,7 +51,7 @@ function getOptimizedNumBytes(url) {
         canvas.width = img.width;
         context.drawImage(img, 0, 0);
 
-        const jpeg = getTypeStats('image/jpeg', 0.85);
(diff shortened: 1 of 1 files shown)


patrickhulce:
@ebidel #2217 is for looking into how it compares to other encoders and automating the tracking of it over time, do we want a separate issue to track down this specific change?

patrickhulce:
@brendankenny yeah libjpeg-turbo is one of 3 possibilities I'm not sure under what circumstances the others get triggered would have to do some more digging

https://cs.chromium.org/chromium/src/ui/gfx/codec/jpeg_codec.cc?type=cs&l=47

patrickhulce:
Culprit looks like https://chromium.googlesource.com/chromium/src/+/86074971b9640dbf59869071b4a18bec51c61424

PR #2353, opened 2017-05-24
Revert "Switch to containerized Trusty"

patrickhulce:
Been seeing a lot more travis flakey failures (12/12 recent failures of known green branches), want to give this branch a try to see if it improves.

Reverts GoogleChrome/lighthouse#2234

PR #2356, opened 2017-05-24
test(smokehouse): add numeric comparisons

patrickhulce:
replaces #2353

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -29,7 +29,7 @@ const DEFAULT_EXPECTATIONS_PATH = 'pwa-expectations';
 
 const PROTOCOL_TIMEOUT_EXIT_CODE = 67;
 const RETRIES = 3;
-const NUMERICAL_EXPECTATION_REGEXP = /(<|>)(\d+)/;
+const NUMERICAL_EXPECTATION_REGEXP = /^(<|>)(\d+)$/;
 
 
 /**
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -29,7 +29,7 @@ const DEFAULT_EXPECTATIONS_PATH = 'pwa-expectations';
 
 const PROTOCOL_TIMEOUT_EXIT_CODE = 67;
 const RETRIES = 3;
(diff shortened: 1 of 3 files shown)


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -29,6 +29,7 @@ const DEFAULT_EXPECTATIONS_PATH = 'pwa-expectations';
 
 const PROTOCOL_TIMEOUT_EXIT_CODE = 67;
 const RETRIES = 3;
+const NUMERICAL_EXPECTATION_REGEXP = /(<|>)(\d+)/;
```
samccone: `/^(<|>)(\d+)$/`

to tighten it up a bit.

patrickhulce: done

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -105,6 +106,30 @@ function runLighthouse(url, configPath, saveAssetsPath) {
 }
 
 /**
+ * Checks if the actual value matches the expectation. Does not recursively search. This supports
+ *    - Greater than/less than operators, e.g. "<100", ">90"
+ *    - Regular expressions
+ *    - Strict equality
+ *
+ * @param {*} actual
```
samccone: if i am reading your code correctly..

```
{string|number} actual
{RegExp|string} expected
```

I think these are the types

patrickhulce: nah it can be any, the Object.is catch-all does boolean and number expectations too, objects will also be passed in but will always be false

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -29,6 +29,7 @@ const DEFAULT_EXPECTATIONS_PATH = 'pwa-expectations';
 
 const PROTOCOL_TIMEOUT_EXIT_CODE = 67;
 const RETRIES = 3;
+const NUMERICAL_EXPECTATION_REGEXP = /^(<|>)(\d+)$/;
```
paulirish: seems like it should handle `<=` as people would expect it, yah?

patrickhulce: heh yeah I suppose you're right

PR #2357, opened 2017-05-25
Enable passing of a custom userDataDir to launcher

### LINKED ISSUE #2291: Allow custom TMP_PROFILE_DIR
Launcher kept failing when started with default user data dir on macOS Sierra.  Now this probably is an issue with chrome or directory permissions on my system.

However, I was easily able to fix this by passing a custom --user-data-dir in additionalFlags
This does double the --user-data-dir argument, and maybe we could allow for it to customizing TMP_PROFILE_DIR in configuration.

samccone:
* rename TMP_DIR to a more accurate name. Since this is a private
internal value, this is non-breaking.

Fixes #2291
Ref #2092

### DIFF
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -5,16 +5,19 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
-    "test": "mocha --reporter dot test/**/*-test.js",
+    "test": "mocha -r ts-node/register --reporter dot test/**/*-test.ts",
     "test-formatting": "test/check-formatting.sh",
-    "format": "clang-format -i -style=file *.ts"
+    "format": "clang-format -i -style=file **/*.ts *.ts"
   },
   "devDependencies": {
+    "@types/mocha": "^2.2.41",
     "clang-format": "^1.0.50",
     "mocha": "^3.2.0",
+    "ts-node": "^3.0.4",
     "typescript": "2.2.1"
   },
   "dependencies": {
+    "@types/core-js": "^0.9.41",
     "@types/mkdirp": "^0.3.29",
     "@types/node": "6.0.66"
   }
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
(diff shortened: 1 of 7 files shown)


samccone:
CI failure is just due to flake.

paulirish:
✅  Confirmed it works via adding a `userDataDir` option into the https://github.com/GoogleChrome/lighthouse/blob/master/docs/readme.md#using-programmatically example.

doesn't delete if passed a custom path. clears the profile if we generated.

Nice!

paulirish:
Tests? wdyt

wardpeet:
would be nice to test this on appveyor 😄

samccone:
Added some simple tests @paulirish 

to test the destroy tmp, I am going to have to do some more heavy handed refactors, let me know what you think so far before I keep going.

paulirish:
We discussed offline:

* Don't install the `.chrome` property on the launcher instance. if anything it should be in the return object of `launch()`
* move `if (this.opts.userDataDir === undefined) {` to inside of `destroyTmp` so you can test that without faking a `close` event and a `this.chrome` interface.
* verify tmpdir is deleted when it should be and not deleted when it shouldnt be

chrome-launcher/package.json
```diff
@@ -5,16 +5,19 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
-    "test": "mocha --reporter dot test/**/*-test.js",
+    "test": "mocha -r ts-node/register --reporter dot test/**/*-test.ts",
```
paulirish: expand to `--require`

chrome-launcher/package.json
```diff
@@ -5,16 +5,19 @@
   "scripts": {
     "build": "tsc",
     "dev": "tsc -w",
-    "test": "mocha --reporter dot test/**/*-test.js",
+    "tests": "mocha -r ts-node/register --reporter dot test/**/*-test.ts",
```
paulirish: why remove `yarn test` ? Would prefer to continue the convention..

paulirish: can you use `--require` instead of `-r` ?

paulirish: `tests` -> `test`.  here and in run_mocha

samccone:
updated

samccone:
Thanks everyone for the reviews

PR #2361, opened 2017-05-25
swap math.round with util.formatNumber

### LINKED ISSUE #2320: Lighthouse 2.0.0 FF issues
I'm using Firefox 54 beta on Windows 7 64-bit

1. Some of the new SVG icons don't show up here
2. The score on the sidebar shows NaN

screen shot 2017-05-20 at 20 01 33-fullpage

ev1stensberg:
Fixes #2320 
---
#### Previously:

`formatNumber` returns a string value from `Number.toLocaleString`, and `math.round` can't round the number.

#### Currently:

`math.round` receives an integer instead of a string, which makes the conversion complete, not returning `NaN`. Note that we first do `math.round` and then use `formatNumber` instead of the opposite, as prior.

--- 
Didn't change test for `Utils.formatNumber`, as I didn't touch it. Furthermore, I'd prefer a common conversion point for [this](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/v2/renderer/category-renderer.js#L105) in relation to the line I've changed, as we're basically doing the same thing. This ensures that scores are persistent and co-reliant.

---

## Before:
skjermbilde 2017-05-25 kl 11 06 01
skjermbilde 2017-05-25 kl 11 07 41

## After:
skjermbilde 2017-05-25 kl 10 59 27

skjermbilde 2017-05-25 kl 11 10 22

### DIFF
--- a/lighthouse-core/report/v2/renderer/report-renderer.js
+++ b/lighthouse-core/report/v2/renderer/report-renderer.js
@@ -127,7 +127,7 @@ class ReportRenderer {
       this._dom.find('.leftnav-item__category', navItem).textContent = category.name;
       const score = this._dom.find('.leftnav-item__score', navItem);
       score.classList.add(`lh-score__value--${Util.calculateRating(category.score)}`);
-      score.textContent = Util.formatNumber(Math.round(category.score));
+      score.textContent = Math.round(category.score);
       nav.appendChild(navItem);
     }
     return leftNav;
--- a/lighthouse-core/report/v2/renderer/report-renderer.js
+++ b/lighthouse-core/report/v2/renderer/report-renderer.js
@@ -127,7 +127,7 @@ class ReportRenderer {
       this._dom.find('.leftnav-item__category', navItem).textContent = category.name;
       const score = this._dom.find('.leftnav-item__score', navItem);
       score.classList.add(`lh-score__value--${Util.calculateRating(category.score)}`);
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -127,7 +127,7 @@ class ReportRenderer {
       this._dom.find('.leftnav-item__category', navItem).textContent = category.name;
       const score = this._dom.find('.leftnav-item__score', navItem);
       score.classList.add(`lh-score__value--${Util.calculateRating(category.score)}`);
-      score.textContent = Math.round(Util.formatNumber(category.score));
+      score.textContent = Util.formatNumber(Math.round(category.score));
```
ebidel: Agreed about a consistent calculation of the final category score.  We have a couple of places where you call `Math.round(score)`. 

That said, this nav score should be the same as the [others](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/v2/renderer/category-renderer.js#L105). Those just use `Math.round`.

`Math.round(category.score)`

ev1stensberg: Do you want me to change it?

ebidel: Yes please.

ev1stensberg: Done :)

PR #2367, opened 2017-05-25
fix: deprecate old HTML report

### LINKED ISSUE #2344: WebPageTest is using 2.0.0 w/ the old report
Example: https://www.webpagetest.org/lighthouse.php?test=170523_CK_1C1G&run=1

Doesn't look all that bad, but we should flip the `html` default to the new report, so everybody has the new stuff, eh?

patrickhulce:
closes #2344 and opens the door to fully remove the old report code, users who request --output html will receive the current report instead (including WPT)


paulirish:
image

patrickhulce:
ha, fixed

Issue #1913, opened 2017-03-27
Remove usage of /deep/


ebidel:
Intent to deprecate has gone out. Eventually `/deep/` will be aliased to `""`. This means our usage of `/deep/` in gathers (`document.querySelecotorAll('html /deep/ *')`) will become `document.querySelecotorAll('html *')` and we'll miss out on nodes within shadow trees. 

We'll need to switch over to DOM traversal.

email that went to blink-dev: https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/68qSZM5QMRQ/pT2YCqZSomAJ

ebidel:
`/deep/` is being removed in Chrome 60. I'll get going on this. 

https://www.chromestatus.com/feature/4964279606312960

ebidel:
Filed an upstream FR ^ (https://github.com/ChromeDevTools/devtools-protocol/issues/19) to add better support for this int he DT protocol. We can get all nodes with the substrees fairly easily, but making `DOM.querySelector[All]` shadow dom aware would be more flexible, long-term.

PR #2371, opened 2017-05-26
Remove /deep/ usage

ebidel:
Fixes https://github.com/GoogleChrome/lighthouse/issues/1913

Also enables shadow dom on the rel="noopener" audit 👊

### DIFF
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -203,7 +203,7 @@ module.exports = [
         },
         details: {
           items: {
-            0: {value: '6,025'},
+            0: {value: '6,037'},
             1: {value: '9'},
             2: {value: '6,003'}
           }
@@ -225,7 +225,7 @@ module.exports = [
         },
         details: {
           items: {
-            0: {value: '25'},
+            0: {value: '37'},
             1: {value: '9'},
             2: {value: '9'}
           }
@@ -247,7 +247,7 @@ module.exports = [
         },
         details: {
           items: {
-            0: {value: '24'},
+            0: {value: '33'},
             1: {value: '7'},
             2: {value: '9'}
           }
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -196,7 +196,7 @@ module.exports = [
         score: 0,
         extendedInfo: {
           value: {
-            0: {value: '6,025'},
+            0: {value: '6,037'},
             1: {value: '9'},
             2: {value: '6,003'}
           }
@@ -218,7 +218,7 @@ module.exports = [
         score: 100,
         extendedInfo: {
           value: {
-            0: {value: '25'},
+            0: {value: '37'},
             1: {value: '9'},
             2: {value: '9'}
           }
@@ -240,7 +240,7 @@ module.exports = [
         score: 100,
         extendedInfo: {
           value: {
-            0: {value: '24'},
+            0: {value: '33'},
             1: {value: '7'},
             2: {value: '9'}
           }
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -748,7 +748,7 @@ class Driver {
    *     True by default.
    * @return {!Promise<!Array<!Element>>} The found elements, or [], resolved in a promise
    */
-  getAllElementsInDocumentAndFrames(pierce = true) {
+  getElementsInDocument(pierce = true) {
     return this.sendCommand('DOM.getFlattenedDocument', {depth: -1, pierce})
       .then(result => {
         const elements = result.nodes.filter(node => node.nodeType === 1);
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -743,6 +743,20 @@ class Driver {
   }
 
   /**
+   * Returns the flattened list of all DOM nodes within the document.
+   * @param {boolean=} pierce Whether to pierce through shadow trees and iframes.
+   *     True by default.
+   * @return {!Promise<!Array<!Element>>} The found elements, or [], resolved in a promise
+   */
+  getAllNodesInDocument(pierce = true) {
+    return this.sendCommand('DOM.getFlattenedDocument', {depth: -1, pierce})
+      .then(result => {
+        const elements = result.nodes.filter(node => node.nodeType === 1);
+        return elements.map(node => new Element({nodeId: node.nodeId}, this));
+      });
+  }
+
+  /**
    * @param {{additionalTraceCategories: string=}=} flags
    */
   beginTrace(flags) {
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -748,7 +748,7 @@ class Driver {
    *     True by default.
    * @return {!Promise<!Array<!Element>>} The found elements, or [], resolved in a promise
    */
-  getAllNodesInDocument(pierce = true) {
+  getAllElementsInDocumentAndFrames(pierce = true) {
     return this.sendCommand('DOM.getFlattenedDocument', {depth: -1, pierce})
       .then(result => {
         const elements = result.nodes.filter(node => node.nodeType === 1);
--- a/lighthouse-core/gather/gatherers/image-usage.js
+++ b/lighthouse-core/gather/gatherers/image-usage.js
@@ -134,7 +134,7 @@ class ImageUsage extends Gatherer {
     }, {});
 
     const expression = `(function() {
-      ${DOMHelpers.getElementsInDocument.toString()};
+      ${DOMHelpers.getElementsInDocumentFn}; // define function on page
       return (${collectImageElementInfo.toString()})();
     })()`;
 
--- a/lighthouse-core/gather/gatherers/image-usage.js
+++ b/lighthouse-core/gather/gatherers/image-usage.js
@@ -11,8 +11,9 @@
 'use strict';
 
 const Gatherer = require('./gatherer');
+const DOMHelpers = require('../../lib/dom-helpers.js');
 
-/* global window, document, Image */
+/* global window, getElementsInDocument, Image */
 
 /* istanbul ignore next */
 function collectImageElementInfo() {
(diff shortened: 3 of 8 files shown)


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -152,7 +152,7 @@ module.exports = [
         score: 0,
         extendedInfo: {
           value: {
-            0: {value: '6,025'},
+            0: {value: '6,037'},
```
ebidel: TIL that `document.querySelector('html, html /deep/ *')` returns different results for shadow dom v0 vs shadow dom v1! I've used it in the past for v0 (e.g. chromestatus returns correct results), but the dbw tester uses shadow dom v1, so we weren't catching all the nodes that we should have been. No idea why v0 and v1 are different, but they deprecated `/deep/` before v1 came out. So may be it was buggy with the new API and they never fixed it.

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -38,7 +39,10 @@ function collectImageElementInfo() {
     };
   }
 
-  const htmlImages = [...document.querySelectorAll('img')].map(element => {
+  const allElements = getElementsInDocument();
```
ebidel: small perf one on the reduced qSA calls :)

lighthouse-core/gather/driver.js
```diff
@@ -688,6 +688,20 @@ class Driver {
   }
 
   /**
+   * Returns the flattened list of all DOM nodes within the document.
+   * @param {boolean=} pierce Whether to pierce through shadow trees and iframes.
+   *     True by default.
+   * @return {!Promise<!Array<!Element>>} The found elements, or [], resolved in a promise
+   */
+  getAllNodesInDocument(pierce = true) {
```
paulirish: you filter on nodetype and then return our element instance.
and this pierces iframes whereas the dom helper doesnt.

`getAllElementsInDocumentAndFrames` ?

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -139,7 +144,12 @@ class ImageUsage extends Gatherer {
       return map;
     }, {});
 
-    return driver.evaluateAsync(`(${collectImageElementInfo.toString()})()`)
+    const expression = `(function() {
+      ${DOMHelpers.getElementsInDocument.toString()};
```
paulirish: took me a moment to figure out what you're doing here.

we should use this library consistently. If this makes the most sense, then in `anchors-with-no-rel-noopener.js` we should use it similar rather than `${fn.toString()}(selector)`

lighthouse-core/lib/dom-helpers.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+    }
+  };
+  _findAllElements(document.querySelectorAll('*'));
+
+  return results;
+}
+
+module.exports = {
+  getElementsInDocument
```
paulirish: since this is only going to be consumed as a string, i think we should toString it at this point. We can append a suffix to help clarify that this is for clientside. perhaps `getElementsInDocumentFn`?

lighthouse-core/gather/driver.js
```diff
@@ -688,6 +688,20 @@ class Driver {
...
+    return this.sendCommand('DOM.getFlattenedDocument', {depth: -1, pierce})
+      .then(result => {
+        const elements = result.nodes.filter(node => node.nodeType === 1);
```
stramel: Should this use the `Node.ELEMENT_NODE` enum here?

paulirish: Sure, why not. :)

lighthouse-core/lib/dom-helpers.js
```diff
@@ -0,0 +1,52 @@
...
+function getElementsInDocument(selector) {
+  const results = [];
+
+  const _findAllElements = nodes => {
+    for (let i = 0, el; el = nodes[i]; ++i) {
+      if (!selector || el.matches(selector)) {
+        results.push(el);
+      }
+      // If the element has a shadow root, dig deeper.
+      if (el.shadowRoot) {
```
stramel: Hmmmm should this be using `root` here instead of `shadowRoot` or `shadyRoot`? Didn't really find any documentation on `root`.

lighthouse-core/gather/driver.js
stramel: @paulirish @ebidel this is what I had come up with for querying through shadow dom.

```js
  queryAllElements(query, includeUserAgentShadowDOM = true) {
    return this.sendCommand('DOM.performSearch', {query, includeUserAgentShadowDOM})
      .then(({searchId, resultCount}) => {
        if (resultCount === 0) {
          return [];
        }

        return this.sendCommand('DOM.getSearchResults', {
          searchId,
          fromIndex: 0,
          toIndex: resultCount
        });
      })
      .then(nodeList => {
        const elementList = [];
        nodeList.nodeIds.forEach(nodeId => {
          if (nodeId !== 0) {
            elementList.push(new Element({nodeId}, this));
          }
        });
        return elementList;
      });
  }
```

lighthouse-core/lib/dom-helpers.js
```diff
@@ -0,0 +1,52 @@
...
+        _findAllElements(el.shadowRoot.querySelectorAll('*'));
+      }
+    }
+  };
+  _findAllElements(document.querySelectorAll('*'));
```
stramel: Can you support passing in a starting node? Defaulting to document? Should be pretty easy

lighthouse-core/lib/dom-helpers.js
ebidel: done

lighthouse-core/gather/gatherers/image-usage.js
ebidel: done

lighthouse-core/gather/driver.js
ebidel: done

lighthouse-core/gather/driver.js
ebidel: This is driver code. No `Node.ELEMENT_NODE` :(

lighthouse-core/lib/dom-helpers.js
ebidel: `. shadowRoot` is the standard property. `root` and `shadyRoot` are Polymer APIs

lighthouse-core/lib/dom-helpers.js
ebidel: not sure we need that flexibility quite yet. The method name is `getElementsInDocument`, so it's obvious that it acts on the entire DOM.

ebidel:
PTAL

ebidel:
@paulirish wdyt?

ebidel:
I will be merging this today if there are no comments :)

brendankenny:
10 other PRs will be happy to hear about our new policy :P

lighthouse-core/lib/dom-helpers.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return results;
+}
+
+module.exports = {
+  getElementsInDocumentFn: getElementsInDocument.toString()
```
patrickhulce: would it be too much to call it `getElementsInDocumentFnString`?

ebidel: done

ebidel:
PTAL

lighthouse-core/gather/driver.js
```diff
@@ -743,6 +743,20 @@ class Driver {
   }
 
   /**
+   * Returns the flattened list of all DOM nodes within the document.
+   * @param {boolean=} pierce Whether to pierce through shadow trees and iframes.
+   *     True by default.
+   * @return {!Promise<!Array<!Element>>} The found elements, or [], resolved in a promise
+   */
+  getAllElementsInDocumentAndFrames(pierce = true) {
```
brendankenny: avoid the boolean trap for now by just not making `pierce` optional until it's needed?

lighthouse-core/gather/driver.js
```diff
@@ -743,6 +743,20 @@ class Driver {
   }
 
   /**
+   * Returns the flattened list of all DOM nodes within the document.
```
brendankenny: all DOM *element* nodes maybe, to be specific?

lighthouse-core/gather/gatherers/image-usage.js
brendankenny: wait, what's the perf improvement?

ebidel: Oh, just that we went from 2 qSA to 1. Before we had a sep query just for `document.querySelectorAll('img')`. But now, `allElements` is inclusive of all the img elements on the page. Filtering `allElements` is probably not much faster, but it's nice to combine the collection calls.

lighthouse-core/gather/driver.js
ebidel: I was modeling after the [DT protocol APIs](https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getFlattenedDocument), which make `pierce` optional. With the exception that they default to `false`. We want to default to `true`.

lighthouse-core/gather/driver.js
ebidel: Think the `@return` value covers us.

ebidel:
> it's unfortunate not to have a driver-answer to "how do I get all the elements matching a selector, including in shadow trees". Any way we can wrap that up in a usable driver method?

Agreed. Pavel wasn't a fan of supporting this at the protocol level:
https://github.com/ChromeDevTools/devtools-protocol/issues/19

lighthouse-core/gather/driver.js
brendankenny: I meant in regards to the "nodes" part...it's only returning nodes with `nodeType` 1. Could also just be "Returns the flattened list of all DOM elements within the document."

lighthouse-core/gather/driver.js
brendankenny: right, but `getAllElementsInDocumentAndFrames(false)` is confusing on its own. Don't get all elements?

We could change it somehow to avoid the boolean trap, but I was just suggesting that we don't need to because we don't actually have any uses for not piercing.

If you really think it should stay, though, we should rename somehow (if `pierce` is false it won't be looking in the "frames" part of `getAllElementsInDocumentAndFrames`), and maybe make the parameter `noPierce` so that `undefined` can just act as false as is usually expected?

brendankenny:
> Agreed. Pavel wasn't a fan of supporting this at the protocol level

So is there anything we can do on LH's side? I guess one issue would be is it possible for `evaluateAsync` to send back node IDs compatible with the ones the protocol returns

lighthouse-core/gather/driver.js
ebidel: In that case, I think we should go back to what I had originally: `getAllElementsInDocument(pierce = true)`. 

Closest to the DT protocol for both func and param naming. The docs already explain what `pierce` does ("pierces through iframes and shadow roots when true"). It doesn't need to be in the method name.

PR #2377, opened 2017-05-26
Remove outline from active clicks & show green bar on successful audits

ev1stensberg:
## Before 
skjermbilde 2017-05-26 kl 21 40 55
## After

skjermbilde 2017-05-27 kl 01 41 03

Optional merge. Fixing svg-icons issue right in mozilla ( might exist in IE too), seems to be related to `details` and `summary` not being correctly set in relation to the `::-webkit-details-marker` selector.

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -73,10 +73,6 @@ a {
   color: #0c50c7;
 }
 
-summary:active {
-  outline: none;
-}
-
 summary {
   cursor: pointer;
 }
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -73,6 +73,10 @@ a {
   color: #0c50c7;
 }
 
+summary:active {
+  outline: none;
+}
+
 summary {
   cursor: pointer;
 }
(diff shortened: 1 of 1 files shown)


ev1stensberg:
Please rerun check, 99 / 100 on LH :(

lighthouse-core/report/v2/report-styles.css
```diff
@@ -73,6 +73,10 @@ a {
   color: #0c50c7;
 }
 
+summary:active {
+  outline: none;
```
ebidel: The browser's default outlines are hideous,  but they're important for a11y :(

One thing we could try is using the green bar for `:active`, but nuking the outline. That way there's a visual indication but it's a bit nicer.

cc @robdodson

ev1stensberg: Want me to change that?

ev1stensberg:
@ebidel Actually looks pretty sweet with outline! 
skjermbilde 2017-05-27 kl 13 34 57

ev1stensberg:
Do you want me to do the same for the grey ones as well? (Manual Checks? )

ebidel:
What do others think? I wasn't super keen about the green bar at first, but it's growing on me.

wardpeet:
I don't like it that much to be honest. The active state disrupt the view a bit too much. Perhaps it's better to have the padding all the time, even on not active state so the only thing that changes is the background. But if the others like this change it's ok for me 👍 

Not sure what issue this fixes? The arrows should be fixed by using a `:before`.

patrickhulce:
we'll have to figure out what we want to do about #2297, these changes would be largely obsolete

I kinda like the new green bar if we end up keeping these styles for LH outside devtools :) but I'd prefer to avoid forking the styles too much

brendankenny:
Looks like we are fixing up #2297 to do a bunch of styling changes after all, so closing this PR in favor of it.

After that PR lands we can do another round of evaluating the usability of the report, making sure we're clearly marking passing/failing sections, etc.

PR #2378, opened 2017-05-27
Fix icons on firefox by decoding svg imgs

ev1stensberg:
## Before:
skjermbilde 2017-05-27 kl 18 11 41

## After:
skjermbilde 2017-05-27 kl 18 10 38

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -219,12 +219,12 @@ span.lh-node:hover {
 }
 
 .lh-score--informative .lh-score__value::after {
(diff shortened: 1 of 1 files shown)


paulirish:
It's nice having these in utf8 svg as they remain very editable.

I looked into this issue and found these articles which affirm my belief this shouldn't be a crossbrowser issue:

* https://css-tricks.com/probably-dont-base64-svg/
* https://codepen.io/tigt/post/optimizing-svgs-in-data-uris


then after debugging a bit it turns out the problem with the `.lh-debug::before` triangle icon is the `#` in the `fill` attribute:
image

Taylor's post mentions that #'s need to be escaped. :)  But you can alternatively just use `hsl` color instead of hex and sort that out.


So it seems like these ones can be fixed just by changing the fill colors from hex to hsl. :)

ebidel:
The `fills` are definitely it. We ran into that back in https://github.com/GoogleChrome/lighthouse/issues/1124#issuecomment-271991648.

ev1stensberg:
I thought it was the issue too, but people on the googs explained as if the problem wasn't just the hashes being invalidated. 

💯 Agree about utf 8 being most convenient , will change this ✌️

ev1stensberg:
Ups, reverted this as I did `git rebase -i HEAD` and dropped the latest commit, hold on ..

ev1stensberg:
This is ready

paulirish:
honestly, this little bug is probably worth a lighthouse audit in its own right: "SVG background-images are cross-browser" :)

paulirish:
fixes (partiallly) #2320

PR #2381, opened 2017-05-27
Run npm install/build tasks in parallel

paulirish:
This makes install-all and build-all about 2-3x faster

On travis this takes:

* `install-all` from 28s to 12s.
* `build-all` from 38s to 27s.


wardpeet:
you probably have to change the package.json to first install the concurrently package and than move forward.

I also noticed that concurrently has some old npm packages. I'll create a PR to update their chalk dependency.

package.json
```diff
@@ -11,12 +11,12 @@
     "node": ">=6"
   },
   "scripts": {
-    "install-all": "yarn && yarn install-launcher && yarn install-cli && yarn install-extension && yarn install-viewer",
+    "install-all": "yarn && concurrently --kill-others-on-fail \"yarn install-launcher\" \"yarn install-cli\" \"yarn install-extension\" \"yarn install-viewer\"",
```
brendankenny: this [may or may not work on linux](https://github.com/kimmobrunfeldt/concurrently/issues/4#issuecomment-217124689), apparently

wardpeet: I can test real quick on linux

brendankenny: to clarify, it's the `--kill-others-on-fail` that may not work on linux

brendankenny: at worst it'll just continue to run them all, even if one failed

wardpeet: works on linux! (ubuntu)
image

brendankenny:
appveyor seems unhappy but not sure why

paulirish:
Using this module now: https://github.com/madarche/npm-run-posix-or-windows/blob/master/bin/cli.js

and basically letting windows get the slow path since it can't be trusted with being fancy.

paulirish:
Looking at travis builds from today:

* `install-all` from 32s to 9.7s. _(3.3x faster)_
* `build-all` from 101s to 33s.    _(3.0x faster)_

🚀

PR #2382, opened 2017-05-27
Appveyor fixes: fresh yarn, only node6 build

paulirish:
We were seeing this error in the appveyor logs:
image

It seemed to point at this yarn bug: https://github.com/yarnpkg/yarn/issues/2591

Appveyor has an older build of Yarn on the bots by default (0.21.3), so i've installed a fresh version of yarn (>= 0.24.5) in the build script. It sorts it out.

cc @XhmikosR 

-------

As a driveby: enabled `fast_finish` which means "immediately finish build once one of the jobs fails." We wanted this for travis too, but they don't have it. odd.

------

_Update_

Two more drivebys:

1. Our OSS free plan only allows 1 concurrent job at a time. Since appveyor VMs are so slow, i dont think its worth it to always wait for 2 back-to-back builds with node 6 and node 7. So i've ditched the node7 build.
2. I enabled the ["rolling builds" option](http://help.appveyor.com/discussions/problems/586-cancel-queued-builds-for-old-commits-of-the-same-branch) in the appveyor settings UI. It cancels queued builds if there's a fresh commit on the branch, must like the recently added travis beta experiment.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,6 +7,10 @@ environment:
   matrix:
     - nodejs_version: "6"
       platform: x86
+    # node 7 is skipped, as appveyor only allows 1 concurrent job
+    # and we want appveyor finishing ASAP. see #2382
+    # - nodejs_version: "7"
+    #   platform: x86
 
 build: off
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,9 +5,7 @@ version: "{build}"
 environment:
   fast_finish: true
   matrix:
-    - nodejs_version: "6.9.1"
-      platform: x86
-    - nodejs_version: "7"
+    - nodejs_version: "6"
       platform: x86
 
 build: off
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -26,8 +26,10 @@ test_script:
   - node --version
   - npm --version
   - yarn --version
-  - yarn lint
-  - yarn unit
+  - which yarn
+  - npm install yarn -g
+  - which yarn
+  - yarn install-all
   - yarn smoke
   - yarn smokehouse
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,6 +3,7 @@ clone_depth: 10
 version: "{build}"
 
 environment:
+  fast_finish: true
   matrix:
     - nodejs_version: "6.9.1"
       platform: x86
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -26,6 +26,7 @@ test_script:
   - node --version
   - npm --version
   - yarn --version
+  - which yarn
```
XhmikosR: Windows has `where` which is a native tool.

.appveyor.yml
```diff
@@ -3,16 +3,16 @@ clone_depth: 10
 version: "{build}"
 
 environment:
+  fast_finish: true
   matrix:
-    - nodejs_version: "6.9.1"
-      platform: x86
-    - nodejs_version: "7"
+    - nodejs_version: "6"
```
XhmikosR: So, you don't want to test node.js 7 anymore?

.appveyor.yml
paulirish: `which` does work in this case. _shrug_

XhmikosR:
@paulirish: instead of us updating yarn, how about I [ping AppVeyor](https://github.com/appveyor/ci/issues/1576) to update to the latest yarn. :)

About the rest of your comments, Travis does have it.

```yml
matrix:
  fast_finish: true
```

So you better bring .travis.yml and .appveyor.yml on par especially for the node.js versions.

.appveyor.yml
XhmikosR: It works because we have MSYS in `PATH` :P

.appveyor.yml
paulirish: ideally i do, but i'm not sure the delay is worth it.

appveyor is significantly slower of a VM than our travis bots, and they don't have a reasonable timeout on inactivity. which means they're very regularly our last CI to finish.

so this evens the playing field a bit, and i don't think missing out on node7/windows is all that bad.

paulirish:
> @paulirish: instead of us updating yarn, how about I ping AppVeyor to update to the latest yarn. :)

awesome thank you!

.appveyor.yml
XhmikosR: Ah, OK, that's reasonable. Maybe have a comment for the future in this file?

I'm not affiliated with AppVeyor but I guess you could report any issues to them :)
On a side note, did you compare cold phase? Because caching isn't enabled if the build fails.

PR #2392, opened 2017-05-30
fixed incorrect comments

mixed:
When I build fail, I saw below message. I think that below comment is wrong.
```
yarn install; yarn build;
```


patrickhulce:
ah, the commands to run depend on which directory you're in @mixed :)
It's true the more common case will likely be running into this issue when developing lighthouse core rather than just chrome-launcher. We could also alias 'install-launcher' and 'build-launcher' in the launcher repo to run the regular commands, haha. thoughts @samccone?

mixed:
Thanks @patrickhulce 

I got it. :) I ran into the root folder and met the message. If I ran it in the `chrome-launcher` folder, it would not have been a problem. If so, I hope that [it will be modified as well.](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/compiled-check.js#L9)

PR #2393, opened 2017-05-30
Add remote-device flag to test mobile devices

### LINKED ISSUE #2368: Testing on mobile devices no longer works
I've upgraded to 2.0.0. and found that [testing on mobile devices](https://github.com/GoogleChrome/lighthouse/blob/master/docs/readme.md#testing-on-a-mobile-device) no longer work. Lighthouse launches a new Chrome tab on my Goobuntu machine.

I've verified that port forwarding works by navigating in Chrome to localhost:9222:

image

However, when I run `lighthouse --disable-device-emulation --disable-cpu-throttling https://google.com`, LH opens a new Chrome tab on my desktop, not on the Android device.

ev1stensberg:
Enables mobile emulation to be the default port `9222` instead of `0` on device emulations. This enables mobile emulations instead of defaulting to the browser. Users can also use the remote debug port to forward I assume, without this getting affected. Fixes #2368

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -57,7 +57,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling`, `--port=9222` and potentially `--disable-network-throttling`. If you want to test on a mobile device without dealing with those flags, you could use `--remote-device` instead, which enables these features by default.
+You'll likely want to use the `--remote-device` flag which disables CPU and network throttling, disables device emulation, and defaults the port to 9222.
 
 ```sh
 $ adb kill-server
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -57,7 +57,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling`, `--port=9222` and potentially `--disable-network-throttling`. If you want to test on a mobile device without dealing with those flags, you could use `lighthouse --remote-device` instead, which enables these features by default.
+You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling`, `--port=9222` and potentially `--disable-network-throttling`. If you want to test on a mobile device without dealing with those flags, you could use `--remote-device` instead, which enables these features by default.
 
 ```sh
 $ adb kill-server
@@ -69,7 +69,7 @@ $ adb devices -l
 
 $ adb forward tcp:9222 localabstract:chrome_devtools_remote
 
-$ lighthouse --disable-device-emulation --disable-cpu-throttling https://mysite.com
+$ lighthouse --remote-device https://mysite.com
 ```
 
 ## Lighthouse as trace processor
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -57,7 +57,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling` and potentially `--disable-network-throttling`.
+You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling`, `--port=9222` and potentially `--disable-network-throttling`. If you want to test on a mobile device without dealing with those flags, you could use `lighthouse --remote-device` instead, which enables these features by default.
 
 ```sh
 $ adb kill-server
--- a/lighthouse-cli/bin.ts
+++ b/lighthouse-cli/bin.ts
@@ -47,10 +47,12 @@ const cliFlags = getFlags();
 
 if (cliFlags.remoteDevice) {
   cliFlags.port = defaults(cliFlags.port, 9222);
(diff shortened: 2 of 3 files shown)


ev1stensberg:
Mind rerunning on node v6?

patrickhulce:
Chrome launcher was meant to be a generic utility that shouldn't inherit all of LH idiosyncrasies. Additionally, I'm not sure we actually want to always use 9222 if device mode is disabled which is a rather opaque behavior. The use case described in #2368 is not necessarily true when this flag is set either (I use this flag all the time to test desktop versions of sites where I still wish to preserve the random port functionality), wdyt? I'm somewhat in favor of keeping the existing functionality and if a specific port should be used, the user should specify `--port XXXX`

ev1stensberg:
Keeping it minimal in userland should be a concern too, we can add a flag for testing on remote instead for people to avoid confusion

patrickhulce:
> Keeping it minimal in userland should be a concern too, we can add a flag for testing on remote/emulated device instead for people to avoid confusion

This is a great idea 👍  it can be rather burdensome right now to have to say `--port XXXX --disable-network-throttling --disable-cpu-throttling --disable-device-emulation` saying `--remote-device` or something that flips all of these bits would be nice

ev1stensberg:
Gotcha, will revert and reimplement at this branch shortly :)

ev1stensberg:
There 👍

lighthouse-cli/bin.ts
```diff
@@ -44,6 +44,13 @@ interface LighthouseError extends Error {
 
 const cliFlags = getFlags();
 
+if (cliFlags.remoteDevice) {
+  cliFlags.port = 9222;
```
patrickhulce: let's only set this if they haven't manually specified one

lighthouse-cli/cli-flags.ts
```diff
@@ -72,6 +73,7 @@ export function getFlags() {
         'disable-device-emulation': 'Disable Nexus 5X emulation',
         'disable-cpu-throttling': 'Disable CPU throttling',
         'disable-network-throttling': 'Disable network throttling',
+        'remote-device': 'Enable remote device emulation',
```
patrickhulce: how about we describe exactly what it does "Disables CPU/Network throttling, emulation, and defaults the protocol port to 9222`?

docs/readme.md
```diff
@@ -57,7 +57,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling` and potentially `--disable-network-throttling`.
+You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling`, `--port=9222` and potentially `--disable-network-throttling`. If you want to test on a mobile device without dealing with those flags, you could use `--remote-device` instead, which enables these features by default.
```
patrickhulce: Let's promote this a little more :) "You'll likely want to use the `--remote-device` flag which disables CPU and network throttling, disables device emulation, and defaults the port to 9222"

lighthouse-cli/bin.ts
```diff
@@ -44,6 +44,13 @@ interface LighthouseError extends Error {
...
+  cliFlags.disableNetworkThrottling = true;
```
patrickhulce: I'll defer to @paulirish if we want to disable network throttling by default in this scenario or just the other three

ev1stensberg:
Review the readme a second time to make sure I revised it correctly @patrickhulce 👍

lighthouse-cli/bin.ts
ev1stensberg: This is what's left, sorry for the port mistake, replaced in the latest commit!

ev1stensberg:
@brendankenny 

> Defaulting to port 9222 when the port flag is 0 (the usual default) might interfere with that depending on if we want to support random ports on a mobile device, for instance.

If this is the case, we'd need a mechanism to forward to the right port with `adb` 

> bike shedding on --remote-device? "remote" is kind of confusing to me in this context, but it does mirror "remote debugging", a well established term.

Could go with `remote-emulation`,  `remote-testing` or `native-device`

> Like @patrickhulce's suggestion about the port, it feels like the precedence for these flags should go the other way, where the individual flags win over the --remote-device (e.g. you want to do it on a real device but you still want to throttle CPU for some reason)

Enabled to override params, sorry for jr mistake 😆 

> one thing that might change the need for this quite a bit is @samccone's ideas/plans on switching over to adb in js. This flag might actually be perfect for that mode, but maybe he can weigh in so we don't have to change it again not far down the line :)

Was actually thinking about that, and it's a great opportunity to avoid defaulting to port 9222, but we'd need a mechanism to set the port correct with `adb`. That is doable when Sam implements the `adb in js`

ev1stensberg:
Anything I can do at my end to get this ready for merge?

lighthouse-cli/bin.ts
```diff
@@ -44,6 +45,14 @@ interface LighthouseError extends Error {
 
 const cliFlags = getFlags();
 
+if (cliFlags.remoteDevice) {
+  cliFlags.port = defaults(cliFlags.port, 9222);
```
brendankenny: I don't think this will work, since we set defaults for `port` and `disable-cpu-throttling` at least (to `0` and `false`, respectively), and `defaults()` checks for `undefined`.

I believe we can remove the `disable-cpu-throttling` default, though, since we can just use `undefined` as a falsy value in the normal case.

lighthouse-cli/bin.ts
```diff
@@ -44,6 +45,14 @@ interface LighthouseError extends Error {
...
+  cliFlags.disableNetworkThrottling = defaults(cliFlags.disableNetworkThrottling, true);
+  console.log(cliFlags.disableNetworkThrottling)
```
brendankenny: leftover console.log from development?

ev1stensberg: Yep, sorry!

lighthouse-cli/bin.ts
ev1stensberg: Gotcha. Idea behind default was to assign those if supplied. If a user supplies `--remote-device` and `port=8080`, it won't get the override, rather stay at `9222`

ev1stensberg: Same goes for the other flags, if the user wants to deactivate, say `disableNetworkThrottling`, he/she can do that now.

paulirish:
I think we'll move forward with this.. maybe with something like `--preset=remote-device`. No real rush here but generally this seems handy.

ev1stensberg:
If you want me to work on this now, I'll get to it. Need feedback of what you decide to do first

brendankenny:
Going to close for now. It does seem like it would be a handy flag, but we can wait for more demand/a clear plan

Issue #2383, opened 2017-05-28
HTTP->HTTPS redirect via JS fails redirects-http audit


PaulKinlan:
I created a quick sample https://pwa.glitch.me/

I don't have control over the serving on the static site so I can't easily 30X redirect HTTP to HTTPS. Instead I have a little piece of javascript that changes the protocol from http to https.

According to the [docs] it should pass the test.

>How to pass the audit
>
> Use canonical links in the head of your HTML to help search engines figure out the best way to get to the page.
> 
> Configure your server to redirect HTTP traffic to HTTPS. See your server's documentation to figure out the best way to do this.
> ...
> Lighthouse changes the page's URL to http, loads the page, and then waits for the event from the Chrome Debugger that indicates that the page is secure. If Lighthouse does not receive the event within 10 seconds then the audit fails.

Not sure if it is the test that is broken, or the docs that need more clarification.

[docs]: https://developers.google.com/web/tools/lighthouse/audits/http-redirects-to-https

patrickhulce:
This situation should also hopefully be fixed by #2396

ev1stensberg:
Doable when we add the initial URL [here](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L535) and [here](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L66), later validating it the same way as we did at #2396

ev1stensberg:
⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️  ⭐️ ⭐️ 
![skjermbilde 2017-06-01 kl 17 08 30](https://cloud.githubusercontent.com/assets/16735925/26686397/11d918c6-46ed-11e7-9fa0-1c3eeeb17b69.png)
[...]

PR #2396, opened 2017-05-30
Manually check protocol for http -> https redirect

### LINKED ISSUE #2363: HTTP redirects to HTTPS audit doesn't work on android
When you run Lighthouse on WPT or on any Android device, the security domain does not work properly and thus fails the HTTP redirect check, [see this report as example](https://www.webpagetest.org/lighthouse.php?test=170525_5S_14TQ&run=1). 

We should do what we did for is on HTTPS and just check for the scheme.

ev1stensberg:
fixes #2363

I’m not sure if I did this right. Seems too simple, but may be my lack of knowledge of what `afterPass` is doing behind the scenes.

### DIFF
--- a/lighthouse-core/gather/gatherers/http-redirect.js
+++ b/lighthouse-core/gather/gatherers/http-redirect.js
@@ -17,6 +17,7 @@
 'use strict';
 
 const Gatherer = require('./gatherer');
+const URL = require('../../lib/url-shim');
 
 /**
  * This gatherer changes the options.url so that its pass loads the http page.
@@ -39,9 +40,13 @@ class HTTPRedirect extends Gatherer {
     let isRedirect = false;
 
     if(options.driver._httpsArr.length !== 0) {
-      const httpsExpectURL = options.url.replace(/^http/, 'https');
       options.driver._httpsArr.forEach( (httpsUrl) => {
-        if(httpsUrl.url === httpsExpectURL) {
+        if(new URL(options.url).protocol === 'http:') {
+          const RedirectURL = options.url.replace(/^http/, 'https');
+          if(RedirectURL === httpsUrl.url) {
+            isRedirect = true;
+          }
+        } else if(httpsUrl.url === options.url) {
           isRedirect = true;
         } else if(httpsUrl.path === '/') {
           isRedirect = true;
--- a/lighthouse-core/gather/gatherers/http-redirect.js
+++ b/lighthouse-core/gather/gatherers/http-redirect.js
@@ -17,10 +17,11 @@
 'use strict';
 
 const Gatherer = require('./gatherer');
+const URL = require('../../lib/url-shim');
 
 /**
  * This gatherer changes the options.url so that its pass loads the http page.
- * After load it detects if its on a crypographic scheme.
+ * After load it detects if its protocol is https
  * TODO: Instead of abusing a loadPage pass for this test, it could likely just do an XHR instead
  */
 class HTTPRedirect extends Gatherer {
@@ -36,39 +37,18 @@ class HTTPRedirect extends Gatherer {
   }
 
   afterPass(options) {
-    // Reset the options.
+    // explicitly use this._ to make test pass
+    this._url = options.url;
+    const checkURLAfterDelay = new Promise(resolve => {
+      setTimeout(resolve, 5000);
+    }).then(_ => {
+      this._url = options.url;
+      return {value: new URL(this._url).protocol === 'https:'};
+    }).catch(_ => {
+      throw new Error('Couldn\'t resolve redirect');
+    });
     options.url = this._preRedirectURL;
-
-    // Allow override for faster testing.
-    const timeout = options._testTimeout || 10000;
-
-    const securityPromise = options.driver.getSecurityState()
-      .then(state => {
-        return {
-          value: state.schemeIsCryptographic
-        };
-      });
-
-    let noSecurityChangesTimeout;
-    const timeoutPromise = new Promise((resolve, reject) => {
-      // Set up a timeout for ten seconds in case we don't get any
-      // security events at all. If that happens, bail.
-      noSecurityChangesTimeout = setTimeout(_ => {
-        reject(new Error('Timed out waiting for HTTP redirection.'));
-      }, timeout);
-    });
-
-    return Promise.race([
-      securityPromise,
-      timeoutPromise
-    ]).then(result => {
-      // Clear timeout. No effect if it won, no need to wait if it lost.
-      clearTimeout(noSecurityChangesTimeout);
-      return result;
-    }).catch(err => {
-      clearTimeout(noSecurityChangesTimeout);
-      throw err;
-    });
+    return checkURLAfterDelay.then(result => result);
   }
 }
 
--- a/lighthouse-core/gather/gatherers/http-redirect.js
+++ b/lighthouse-core/gather/gatherers/http-redirect.js
@@ -38,8 +38,15 @@ class HTTPRedirect extends Gatherer {
   afterPass(options) {
     let isRedirect = false;
 
-    if(options.driver._httpsArr.length !== 0 && options.driver._httpsArr[0].scheme === 'https') {
-      isRedirect = true;
+    if(options.driver._httpsArr.length !== 0) {
+      const httpsExpectURL = options.url.replace(/^http/, 'https');
+      options.driver._httpsArr.forEach( (httpsUrl) => {
+        if(httpsUrl.url === httpsExpectURL) {
+          isRedirect = true;
+        } else if(httpsUrl.path === '/') {
+          isRedirect = true;
+        }
+      });
     }
     options.url = this._preRedirectURL;
     return {value: isRedirect};
--- a/lighthouse-core/gather/gatherers/http-redirect.js
+++ b/lighthouse-core/gather/gatherers/http-redirect.js
@@ -17,7 +17,6 @@
 'use strict';
 
 const Gatherer = require('./gatherer');
-const URL = require('../../lib/url-shim');
 
 /**
  * This gatherer changes the options.url so that its pass loads the http page.
@@ -37,11 +36,20 @@ class HTTPRedirect extends Gatherer {
   }
 
   afterPass(options) {
-    const checkURLAfterDelay = new Promise(resolve => {
-      resolve({value: new URL(options.url).protocol === 'https:'});
-    });
+    let isRedirect;
+
+    if(options.driver._httpArr[0].url === options.url) {
+      isRedirect = false;
+    }
+    if(options.driver._httpsArr[0].scheme === 'https') {
+      isRedirect = true;
+    }
+    if(!isRedirect) {
+      isRedirect = false;
+    }
     options.url = this._preRedirectURL;
-    return checkURLAfterDelay.then(result => result);
+
+    return {value: isRedirect};
   }
 }
 
--- a/lighthouse-core/gather/gatherers/http-redirect.js
+++ b/lighthouse-core/gather/gatherers/http-redirect.js
@@ -36,19 +36,12 @@ class HTTPRedirect extends Gatherer {
   }
 
   afterPass(options) {
-    let isRedirect;
+    let isRedirect = false;
 
-    if(options.driver._httpArr[0].url === options.url) {
-      isRedirect = false;
-    }
     if(options.driver._httpsArr[0].scheme === 'https') {
       isRedirect = true;
     }
-    if(!isRedirect) {
-      isRedirect = false;
-    }
     options.url = this._preRedirectURL;
-
     return {value: isRedirect};
   }
 }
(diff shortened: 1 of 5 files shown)


ev1stensberg:
Should I remove the old tests to test against old -> new urls accordingly?

lighthouse-core/gather/gatherers/http-redirect.js
```diff
@@ -36,39 +36,9 @@ class HTTPRedirect extends Gatherer {
...
-    ]).then(result => {
-      // Clear timeout. No effect if it won, no need to wait if it lost.
-      clearTimeout(noSecurityChangesTimeout);
-      return result;
-    }).catch(err => {
-      clearTimeout(noSecurityChangesTimeout);
-      throw err;
-    });
+    return {
+      value: options.url.includes('https')
```
patrickhulce: maybe slightly more robust to bring in `lighthouse-core/libs/url-shim` and do `new URL(options.url).protocol === 'https:'`?

ev1stensberg: Actually tried to implement it as:

ev1stensberg:
Proof of concept, adding tests in a sec! 
skjermbilde 2017-05-31 kl 00 31 56
skjermbilde 2017-05-31 kl 00 30 26

lighthouse-core/gather/gatherers/http-redirect.js
```diff
@@ -27,48 +28,12 @@ class HTTPRedirect extends Gatherer {
 
   constructor() {
     super();
-    this._preRedirectURL = undefined;
-  }
-
-  beforePass(options) {
-    this._preRedirectURL = options.url;
-    options.url = this._preRedirectURL.replace(/^https/, 'http');
```
patrickhulce: we still need this part though, that's what lets us test the redirect when an https url is given :)

ev1stensberg: :((

ev1stensberg:
Done now, I'd squash this 🥑  on acceptance!

patrickhulce:
want to give `https://pwa.glitch.me/` a run too to make sure this fixes #2383 too? 🐦 🐦 💎

ev1stensberg:
Can make a dive for it tomorrow, it's 1AM here 🌜

lighthouse-core/gather/gatherers/http-redirect.js
```diff
@@ -36,39 +37,9 @@ class HTTPRedirect extends Gatherer {
...
-      timeoutPromise
-    ]).then(result => {
-      // Clear timeout. No effect if it won, no need to wait if it lost.
-      clearTimeout(noSecurityChangesTimeout);
-      return result;
-    }).catch(err => {
-      clearTimeout(noSecurityChangesTimeout);
-      throw err;
-    });
+    return {
```
brendankenny: can remove all the `this._preRedirectURL` stuff up above, as well

ev1stensberg: Did, but @patrickhulce told me to let it stay, I'll look into it tomorrow 👍

patrickhulce: @brendankenny how else will the URL to be tested be changed if they provide a URL already on `https`?

brendankenny: sorry, meant only `this._preRedirectURL` doesn't need to be a member variable anymore. Still need the URL change

dandv:
Would this also fix http://m.uber.com being flagged as not redirecting to HTTPS, when in fact it is, via HSTS?

image

image

On the other hand, http://blueseed.com doesn't fail the redirect check, even though the same HSTS mechanism is used.

ev1stensberg:
Only checks for `HTTPS` right now @dandv, should I integrate to check header for `HSTS` as well @patrickhulce @brendankenny ?

ev1stensberg:
Works here. Tested on Google Pixel XL
skjermbilde 2017-05-31 kl 15 26 59

lighthouse-core/gather/gatherers/http-redirect.js
ev1stensberg: Changed in new commit along with failing tests. I added some clauses to make dates work too

ev1stensberg:
Fixes #2383 , but I need to fix the promise-race from the gross implementation in `driver`. Any suggestions? Also need to remove the security module from the mocks & src

brendankenny:
Not sure what's going on in 3132c89. driver methods generally shouldn't navigate, that should be driven by gather-runner.

What was the implementation before the latest commit missing? JS redirects?

ev1stensberg:
Yep. One thing we could do, is to add another `this.something` [here](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L65) and assign the host url to that when the first network monitor starts, and later just fetch it with a util that has the same implementation as the one I just committed. This way, there's no promise race or anything, just the same logic we've used in the previous commits. 

Also, if we decide to make a method like the one I just implemented, it's easier to look for headers and other minor bugs that might occur in hindsight. 

I dislike the approach I made in this commit, as the promise-race makes the runtime 👎 . Good side is however, that we could do what I just mentioned above. 

Also experimented a bit with checking for SW'ers, we could do that, but that is a one-time bug that could be dropped if we go with what I initially described.

ev1stensberg:
`this.something` could be reassigned [here](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L535) to match the initial route

paulirish:
fwiw here is the diff before the latest commit: https://github.com/GoogleChrome/lighthouse/compare/3132c8915230f1131ccfc59279df8ba4f7ca6fe8~

brendankenny:
hmm, @ev1stensberg, maybe I don't understand what you're trying to do. From `getRedirectURL`, it appears that it's trying to navigate to the http page, and watches the completed network requests looking for the https url. Is that correct?

If that's the case, it's probably easier to look at the `networkRecords` for that pass, which should already have all this data. That's available in `afterPass` (see e.g. [`response-compression`](https://github.com/GoogleChrome/lighthouse/blob/400457b22483510badbd3394db7fc6e335237997/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js#L62) as an example of use there).

But I'm not sure if that'll fix all our issues:
1) I don't actually know if redirecting in js should pass the check. It's arguably as secure as an http redirect (or even HSTS if the first visit to the site is over http), I guess, so maybe it's fine.
2) Should the https redirect have to take the user back to the same exact url with the more secure protocol? That seems stricter than necessary.
3) we run the `http-redirect` gatherer [in the same pass as `html-without-javascript`](https://github.com/GoogleChrome/lighthouse/blob/400457b22483510badbd3394db7fc6e335237997/lighthouse-core/config/default.js#L50-L51), so I don't believe the js to do the redirect will run anyways.
4) Why aren't HSTS redirects being picked up by our [redirect monitor](https://github.com/GoogleChrome/lighthouse/blob/400457b22483510badbd3394db7fc6e335237997/lighthouse-core/gather/driver.js#L536-L542)?

These seem complicated enough that they're worth figuring out on their own. In the meantime, it would be great to get the fix in for #2363 before figuring all that out.

ev1stensberg:
Should I revert the latest commit & pick up on a separate PR?

brendankenny:
> Should I revert the latest commit & pick up on a separate PR?

SG, but looks like you already did :)

ev1stensberg:
Yea, common sense 😺

ev1stensberg:
This is ready for the first roud now I assume, will implement what we've discussed / the idea I had tomorrow or so!

ev1stensberg:
This is ready now, we had to update the URLs after waiting approx 5secs for full load. Implemented `this._url` so that tests also can pick up the context.

ev1stensberg:
So much stress for a simple change :(

ev1stensberg:
Hm, weird , Unit Test still failing, worked locally. Lemme get this finalized tomorrow. This module sure is fine-grained 🎭

ev1stensberg:
So I can repro the Unit test, as the `setTimeout` isn't valid in Mocha 👎  Could we leave the `afterPass` tests for the smoke tests? Would be able to remove the `this._url` too and use the prev implementation of `options.url`
skjermbilde 2017-06-01 kl 12 23 44

ev1stensberg:
So I tested http://jakearchibald.github.io/svgomg/ which should redirect after load. However, `httpsRedirect` doesn't pick this up. I'll reimplement this entire PR, as it's not working as expected, way to fragile. I'll be using `Network` to view page redirects to check against the URL we were given in `httpsRedirect`

ev1stensberg:
Should fix https redirect issues

ev1stensberg:
Signing this of @patrickhulce @brendankenny @paulirish 👍  Fixes @dandv 's issue, #2383 & #2363

ev1stensberg:
skjermbilde 2017-06-01 kl 22 02 55
Anything I can do to make sure it doesn't happen, or should we wait for #2402 ?

brendankenny:
> Anything I can do to make sure it doesn't happen, or should we wait for #2402 ?

yes, that will be fixed by #2402

ev1stensberg:
Gotcha, sorry for the super-embarrasing jrdev mistakes! This is ready on my end, I removed tests for `afterPass`, as the tests didn't pick up `options.driver._httpsArr`, should be bullet-proof through smoketests though!

lighthouse-core/gather/gatherers/http-redirect.js
```diff
@@ -36,39 +36,20 @@ class HTTPRedirect extends Gatherer {
...
-    const timeout = options._testTimeout || 10000;
-
-    const securityPromise = options.driver.getSecurityState()
-      .then(state => {
-        return {
-          value: state.schemeIsCryptographic
-        };
+    let isRedirect = false;
+
+    if(options.driver._httpsArr.length !== 0) {
```
patrickhulce: I'm not really understanding why this is necessary @ev1stensberg could you elaborate? As @brendankenny mentioned we're already saving every network request that is issued (which will have flags for redirect, status code, scheme, etc) that you could check in `options`. I'm also interested why jakearchibald's site did not work with the previous approach, the `afterPass` is already quite some time after load and the page has settled, maybe we should just fix the redirect monitor instead if we're not picking up those changes :)

UPDATE: oh wait, does he do a `location = ...` too? we probably only examine redirects right now

lighthouse-core/gather/gatherers/http-redirect.js
```diff
@@ -36,39 +36,20 @@ class HTTPRedirect extends Gatherer {
...
+      const httpsExpectURL = options.url.replace(/^http/, 'https');
+      options.driver._httpsArr.forEach( (httpsUrl) => {
+        if(httpsUrl.url === httpsExpectURL) {
```
patrickhulce: I'd say we don't want to enforce that they redirect to the exact entered URL just with a scheme change, for example when testing `http://theverge.com` it ultimately sends you to `https://www.theverge.com/` which seems fine. @brendankenny did you have a different scenario in mind that you wanted exact equality?

lighthouse-core/gather/gatherers/http-redirect.js
ev1stensberg: Smoketests are failing if it's not checked for, as it results to `undefined`. And yeah, he also is I think. Is there something I need to change here?

lighthouse-core/gather/gatherers/http-redirect.js
ev1stensberg: I formed this after `http://m.uber.com` fails to resolve to https on the prev approach. With `'/'`, we are assuming the root route is https if it's in the arr. Anything I need to change here?

ev1stensberg:
Left comments, for the removed tests, I can try to wrap a class to make them pass instead of removing. Is there anything I need to change that I've failed to implement? / need change?

ev1stensberg:
`options.url` didn't pick up the latest URL through `location` change. If we default to looking at network records, we're now checking if we are on a http protocol for `options.url`, if that is the case, we have to look for a `https` url that matches the one that we've got in `httpsArr`. If we're at https already, this should be picked up by network records anyway, so we just have to test the two urls against each other. It might be that we cannot use protocol to check, as pages could be `m.uber.com`, so we're checking if we're on a root website if this happens. 

Also, in the latest commit, I've added a protocol check before looking at network, so that we can fetch if `options.url` has already been updated. This way, we're covering a lot of cases.

paulirish:
@ev1stensberg are you pushing only to trigger tests?

you can do any of these on your own fwiw:

```sh
yarn unit-core
yarn test
yarn smoke
yarn smoketest
```

ev1stensberg:
The `noopsi` commit was to try to see if the tests were buggy at the CI, I usually run those before I commit. Sorry for the excessive commits.. #jrdev

paulirish:
Closing this as I'm going to take a stab at it based on the plan in https://github.com/GoogleChrome/lighthouse/issues/2363#issuecomment-313768112

Even, Thanks again for trying to make this work!

ev1stensberg:
No worries, good luck!

PR #2398, opened 2017-05-30
feat: tooltips for all formatted URLs

patrickhulce:
extends the goodness of #2312 to all audits formatting URLs

also adds an `elideDataURI` method to url shim to prevent enormous images from filling the LHR multiple times

### DIFF
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -60,7 +60,6 @@ class Deprecations extends Audit {
         });
 
     const deprecationsV2 = entries.filter(log => log.entry.source === 'deprecation').map(log => {
-      // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
       return {
         type: 'code',
         text: log.entry.text,
--- a/lighthouse-core/audits/deprecations.js
+++ b/lighthouse-core/audits/deprecations.js
@@ -24,7 +24,6 @@
 
 const Audit = require('./audit');
 const Formatter = require('../report/formatter');
-const URL = require('../lib/url-shim');
 
 class Deprecations extends Audit {
   /**
@@ -62,11 +61,10 @@ class Deprecations extends Audit {
 
     const deprecationsV2 = entries.filter(log => log.entry.source === 'deprecation').map(log => {
       // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
-      const url = URL.isValid(log.entry.url) ? URL.getURLDisplayName(log.entry.url) : '';
       return {
         type: 'code',
         text: log.entry.text,
-        url,
+        url: log.entry.url,
         source: log.entry.source,
         lineNumber: log.entry.lineNumber
       };
--- a/lighthouse-core/lib/url-shim.js
+++ b/lighthouse-core/lib/url-shim.js
@@ -97,6 +97,7 @@ URL.getURLDisplayName = function getURLDisplayName(url, options) {
 };
 
 /**
+ * Limits data URIs to 100 characters, returns all other strings untouched.
  * @param {string} url
  * @return {string}
  */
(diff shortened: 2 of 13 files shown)


lighthouse-core/audits/deprecations.js
```diff
@@ -62,11 +61,10 @@ class Deprecations extends Audit {
 
     const deprecationsV2 = entries.filter(log => log.entry.source === 'deprecation').map(log => {
       // CSS deprecations can have missing URLs and lineNumbers. See https://crbug.com/680832.
```
ebidel: Shall we remove the comment?

patrickhulce: done

lighthouse-core/lib/url-shim.js
```diff
@@ -96,6 +96,19 @@ URL.getURLDisplayName = function getURLDisplayName(url, options) {
   return Util.getURLDisplayName(new URL(url), options);
 };
 
+/**
+ * @param {string} url
```
brendankenny: add description

lighthouse-core/lib/url-shim.js
```diff
@@ -96,6 +96,19 @@ URL.getURLDisplayName = function getURLDisplayName(url, options) {
...
+ * @return {string}
+ */
+URL.elideDataURI = function elideDataURI(url) {
```
brendankenny: I guess preferring `elideDataUri` is already a losing battle :(

lighthouse-core/lib/url-shim.js
```diff
@@ -96,6 +96,19 @@ URL.getURLDisplayName = function getURLDisplayName(url, options) {
...
+  try {
+    const parsed = new URL(url);
+    return parsed.protocol === 'data:' ? url.slice(0, 100) : url;
```
brendankenny: no clue when it could matter, but what about `parsed.href.slice(0, 100)`

lighthouse-core/lib/url-shim.js
patrickhulce: ha, yeah I've mostly given in at this point

lighthouse-core/lib/url-shim.js
patrickhulce: done

lighthouse-core/lib/url-shim.js
patrickhulce: that seems slightly more confusing to me since I started second guessing if parsed.href is somehow different from `url`, is there a reason you prefer it?

patrickhulce:
> is there a general rule for this? When should an audit author be sure to shorten data URLs?

when there's a reasonable chance that the URL can be massive i.e. it can be a very large image, here I've done it to all audits where we might expect an image and left it off for those we don't

brendankenny:
> when there's a reasonable chance that the URL can be massive i.e. it can be a very large image, here I've done it to all audits where we might expect an image and left it off for those we don't

is it this way (vs just in `_renderURL` for all callers) just to save `details` size?

lighthouse-core/lib/url-shim.js
brendankenny: > that seems slightly more confusing to me since I started second guessing if parsed.href is somehow different from url, is there a reason you prefer it?

making a reader second guess is a downside, but it does mean that any canonicalization or whatever that the URL constructor does is included.

OTOH, I have no clue if it would actually do anything to the beginning of a data URL that parsed successfully :)

`parsed.toString().slice(0, 100)` could be less confusing but does the same thing

patrickhulce:
> just to save details size?

correct-o

lighthouse-core/lib/url-shim.js
patrickhulce: > it does mean that any canonicalization or whatever that the URL constructor does is included

I'm a little confused why we want this, it's going to be reprocessed by a `new URL` by the renderer in `_renderUrl` anyhow, so why would we want to mess with it more here?

brendankenny: yeah, I guess I'm thinking about `URL.elideDataURI` in isolation/other future uses. To me it makes sense to get a shortened version of the canonicalized URL out of calling it, but without a known change that could be made to it, up to you

PR #2399, opened 2017-05-30
rmtraceviewer branch: reduce -> for..of

paulirish:
@patrickhulce i figured i'd save you the annoyance. :)


patrickhulce:
so I set out to try and defend this on [jsperf](https://jsperf.com/for-of-vs-reduce) but `for..of` is 40% faster so I can't 😆

Issue #2364, opened 2017-05-25
Non-fatal events to track


patrickhulce:
This is a list of all the non-fatal error-like events we might want to track once we have Sentry setup.

* Protocol messages with no corresponding callback
* Failed audits
* Failed gatherers
* Failed image conversions in optimize-images
* Audit debug strings for byte-efficiency, loadfast4pwa

paulirish:
'blocked' by #2420

PR #2402, opened 2017-05-31
handle Page.handleJavaScriptDialog rejection

### LINKED ISSUE #2141: intermittent travis failures on "could not handle JavaScript dialog"
This is from the smoke test introduced in #2120. handleJavaScriptDialog

screen shot 2017-05-03 at 12 35 30

Behavior seems to be it fails, the promise rejection is unhandled ([intentionally](https://github.com/GoogleChrome/lighthouse/blob/134b9b0b2235ed34f80b40c75166d1652160cd8b/lighthouse-core/gather/driver.js#L847)), then the test sits there and spins until travis hits its 10 minute timeout.

I'm not sure how the protocol is failing to dismiss the dialog (and only on travis? or are we just not running `yarn smoke` enough locally?), but the only solution I know of right now is removing the `window.confirm` from `dbw_tester`, which would be unfortunate.

Any other ideas?

paulirish:
Here's how to repro this bug:

1. open dbw_tester and find the `window.confirm()` call.
2. repeat it like 7 times.
2. comment out the "--quiet" flag in smokehouse.js to see output
3. run `bash lighthouse-cli/test/smokehouse/dobetterweb/run-tests.sh`
4. when the page loads for the first time, you should see these confirm prompts.
5. hit space bar frequently to try to dismiss the prompt before the protocol dismisses them.

 
--------------------

AFAICT, this rejection was unhandled because it's unexpected and we wanted it to tank the run... is that right?

Anyway, I handle it here and log it as an error:

image

---------

fixes #2141
related issues: #2120, #1939, #2106

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -929,13 +929,8 @@ class Driver {
       // No need to wait for this event bind to complete, so no promise returned
       this.on('Page.javascriptDialogOpening', data => {
         log.warn('Driver', `${data.type} dialog opened by the page automatically suppressed.`);
-        // In some cases, handleJavaScriptDialog will reject, but it appears to only happen when the
-        // the dialog already dismissed.
-        // Since it's handled and no longer blocking the page, we're happy.
-        this.sendCommand('Page.handleJavaScriptDialog', {
-          accept: true,
-          promptText: 'Lighthouse prompt response',
-        }).catch(err => log.warn('Driver', err.message));
+        this.sendCommand('Page.handleJavaScriptDialog', {accept: false}, {silent: true})
+        .catch(_ => this.sendCommand('Page.handleJavaScriptDialog', {accept: true}, {silent: true}));
       });
     });
   }
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -938,9 +938,8 @@ class Driver {
    * @return {!Promise}
    */
   dismissJavaScriptDialogs() {
-    const waitForDismiss = 300;
+    const waitForDismiss = 2000;
     const silent = {silent: true};
-
 
     return this.sendCommand('Page.enable').then(_ => {
       // No need to wait for this event bind to complete, so no promise returned
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -926,13 +926,16 @@ class Driver {
    */
   dismissJavaScriptDialogs() {
     return this.sendCommand('Page.enable').then(_ => {
+      // No need to wait for this event bind to complete, so no promise returned
       this.on('Page.javascriptDialogOpening', data => {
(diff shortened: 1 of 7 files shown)


patrickhulce:
so this whole time the dialog actually was being handled by something else it's just being noisy that it couldn't handle it?

paulirish:
> so this whole time the dialog actually was being handled by something else it's just being noisy that it couldn't handle it?

yes, mostly.

the "could not handle javascript dialog" is the fallback error if the JavaScriptDialogManager fails to "handle" the dialog instance: https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/page_handler.cc?type=cs&q=%22could+not+handle+javascript+dialog%22&sq=package:chromium&l=451

I think it was left in to prevent silent failures, but it does seem unexpected in the travis scenario we have..

lighthouse-core/gather/driver.js
```diff
@@ -929,11 +929,10 @@ class Driver {
       this.on('Page.javascriptDialogOpening', data => {
         log.warn('Driver', `${data.type} dialog opened by the page automatically suppressed.`);
 
-        // rejection intentionally unhandled
         this.sendCommand('Page.handleJavaScriptDialog', {
           accept: true,
           promptText: 'Lighthouse prompt response',
-        });
+        }).catch(err => log.error('driver', err.message));
```
brendankenny: uppercase `Driver` to match other logs in the file

brendankenny:
after discussion offline, absorbing errors here seems OK. The whole point of #2106 was to make it so Lighthouse wasn't stalling out on modal popups. As long as they're being handled, that's fine from LH's perspective.

It is a shame that the errors are being lost completely, but the command seems so noisy that tracking the errors we get (#2364) seem like it would be next to useless. More important would be an error here ("could not handle JavaScript dialog") *and* the LH run still stalls out. Based on how often we see the error in Travis currently, however, if this was at all likely it should show up almost immediately there :)

brendankenny:
oh, wait

> More important would be an error here ("could not handle JavaScript dialog") and the LH run still stalls out.

from #2141:

> Behavior seems to be it fails, the promise rejection is unhandled (intentionally), then the test sits there and spins until travis hits its 10 minute timeout.

so that's already the behavior? This change will silence the unhandled promise rejection but will the LH run actually complete?

patrickhulce:
> so that's already the behavior? This change will silence the unhandled promise rejection but will the LH run actually complete?

ya, I don't expect any meaningful improvement out of this other than a change in log message :)

paulirish:
correct. chrome wil continue to not handle these dialogs pretty frequently. we'll now emit warning logs when it happens, but the run will actually complete. YAyyy

paulirish:
gah still have a problem: https://travis-ci.org/GoogleChrome/lighthouse/jobs/238133366

image

patrickhulce:
wanna add the advice in #2423 to this? :D

PR #2408, opened 2017-05-31
Update optimized-images overview jsdoc

ebidel:
@patrickhulce 

No longer the case for jpeg.


patrickhulce:
heh, yeah we need a conversion table tracked over time since it's ~85 quality of what used to be the JPEG encoder before 2 weeks ago 😆

PR #2409, opened 2017-05-31
Updated readme to add port number when testing a site with authentication

### LINKED ISSUE #2404: Update Readme for Instructions for running lighthouse cli for site with authentication.
The current instructions in the documentation for running lighthouse against a site where you need to login first are:

- Run chrome-debug
- navigate to and log in to your site
- in a separate terminal tab lighthouse http://mysite.com

However, from what I can see for this to work when you run the lighthouse command you would need to add --port and the port number the chrome-debug instance is running on otherwise it opens a new instance of Chrome.

Happy to open a PR if this is the best way but it might be there's a better way of doing this !

mikerhyssmith:
fixes #2404

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -51,7 +51,7 @@ instance with an open debugging port.
 
 - Run `chrome-debug`
 - navigate to and log in to your site
-- in a separate terminal tab `lighthouse http://mysite.com`
+- in a separate terminal tab `lighthouse http://mysite.com --port port-number` with the debugging port number your Chrome instance is running on
 
 ## Testing on a mobile device
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -51,7 +51,7 @@ instance with an open debugging port.
 
 - Run `chrome-debug`
 - navigate to and log in to your site
-- in a separate terminal tab `lighthouse http://mysite.com --port port-number` with the debugging port number your Chrome instance is running on
+- in a separate terminal tab `lighthouse http://mysite.com --port port-number` with the debugging port number your Chrome instance is running on, by default this will be port 9222
 
 ## Testing on a mobile device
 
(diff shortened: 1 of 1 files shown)


docs/readme.md
```diff
@@ -51,7 +51,7 @@ instance with an open debugging port.
 
 - Run `chrome-debug`
 - navigate to and log in to your site
-- in a separate terminal tab `lighthouse http://mysite.com`
+- in a separate terminal tab `lighthouse http://mysite.com --port port-number` with the debugging port number your Chrome instance is running on
```
patrickhulce: maybe add a note that it's 9222 by default or use it as the example port?

brendankenny:
wait, wasn't the whole point that `chrome-debug` now uses a random port and doesn't have a default?

patrickhulce:
> now uses a random port and doesn't have a default?

I thought Lighthouse was looking to launch one on a random port but the instructions told you to open one already at 9222?

patrickhulce:
Oh no it tells you to use chrome-debug nevermind ignore me :) why don't we default chrome-debug to 9222 then?

paulirish:
> why don't we default chrome-debug to 9222 then?

I made that call. Mostly because it's possible of a port conflict and I wanted to make sure that `chrome-debug` command always works even if you have a second one already launched.

The output now says what port it's on, so it seems kinda reasonable to ask the user to use that port.
image

wdyt

mikerhyssmith:
Is it worth adding a note in the readme that chrome-debug will log the port its running on to then use in the lighthouse command ?

patrickhulce:
aight carry on then @mikerhyssmith that sounds good to me :)

paulirish:
> Is it worth adding a note in the readme that chrome-debug will log the port its running on to then use in the lighthouse command ?

yup. let's do that :)

Issue #1980, opened 2017-04-07
[meta] Plots: improve infrastructure for analyzing perf metrics


wwwillchen:
Building off the work in https://github.com/GoogleChrome/lighthouse/pull/1936, I'll work on the following tasks:

### Improve quality
* [x] Run this on a compute instance in Asia to test the hypothesis that the high variance for international sites is driven by network latency
* [ ] Testing effect of launching (or not) a new Chrome for every test (#1982)

### Improve visualizations
* Create a way to view archived results with a URL (e.g. github pages)
* Enable comparisons over different versions of Lighthouse (allows you to easily see how a commit will change the metrics on real-world sites)

### Plots as CLI
* Create flags to configure which sites to analyze
* Create a `--out` flag to customize output path

### Other
* Create a suite of tests to ensure `/plots/` doesn't break
* Add scores to Metrics (https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/lib/traces/pwmetrics-events.js)
* Make plotly a devDependency within `/plots/` so it can be run locally (currently plotly is fetched from CDN)
* Integrate with Travis CI to tag a PR with a before-and-after analysis (optional?)

slohmes:
Hello, I'm excited to see this feature grow!

Following the link in the Lighthouse plots readme (https://googlechrome.github.io/lighthouse/plots/) gives me a 404. I don't see this reported in any open issues - is it new, or expected? I'm happy to either open an issue to remove the link from the documentation, if the page is not currently supported, or to open an issue to fix the plots page, if it should be working (or maybe that should be included in this issue).

Does anyone have clarity around the expected status of the plots page?

paulirish:
@slohmes fixed the 404: https://googlechrome.github.io/lighthouse/plots/

we really like plots too, but we haven't been improving it much recently. would be happy for you to explore it and see if it works for you.

brendankenny:
no more plots 😢

PR #2411, opened 2017-05-31
plots: smoke test for happy case

wwwillchen:
A simple smoke test for plots. Will add more over time.

Closes part of #1980


plots/test/smoke.sh
```diff
@@ -0,0 +1,33 @@
...
+if [ -d "$out_path" ]; then
+  mv $out_path $out_backup_path
+  echo "Moved existing plots/out folder to plots/$out_backup_folder"
+fi
+
+node $measure_script --site https://google.com/ -n 2 --disable-network-throttling --disable-cpu-throttling
+node $analyze_script --ci
+node $clean_script
+
+echo "Finished plots smoke test without errors"
```
patrickhulce: dontcha need some `set -e` goodness to make this fail?

wwwillchen: yup.

plots/analyze.js
```diff
@@ -45,6 +48,9 @@ function main() {
     GENERATED_RESULTS_PATH,
     `var generatedResults = ${JSON.stringify(generatedResults)}`
   );
+  if (args.ci) {
```
paulirish: you can do 

```js
if (process.env.CI)
```

and avoid adding a custom flag

plots/test/smoke.sh
```diff
@@ -0,0 +1,35 @@
...
+
+echo "Starting plots smoke test"
+
+if [ -d "$out_path" ]; then
+  mv $out_path $out_backup_path
+  echo "Moved existing plots/out folder to plots/$out_backup_folder"
+fi
+
+node $measure_script --site https://google.com/ -n 2 --disable-network-throttling --disable-cpu-throttling
+node $analyze_script --ci
```
paulirish: do you want to assert anything about the analyze results?

plots/analyze.js
wwwillchen: Done

plots/test/smoke.sh
wwwillchen: I'll do this is in a follow-up PR - I think many of the potential issues (e.g. lighthouse changes that break plots) would be caught without assertions. 

I'm thinking about processing the output and replace non-deterministic values (e.g. timing values) and then diff it with a golden file.

paulirish:
Only +30s to travis times. Not bad. :)

PR #2417, opened 2017-06-01
link to "offscreen images" reference

kaycebasques:


patrickhulce:
nice! not sure what our policy is going forward on using "TTI" now though, I'm assuming publicly we're trying to say `TTI = time to consistently interactive`? We changed the language of `loadfast4pwa` to say "time to first interactive", but that's rather LH specific at the moment.

kaycebasques:
How about I change it to `... improve page load speed and time to Consistently Interactive.`

patrickhulce:
I was mostly just hoping to remove the use of "(TTI)" as an acronym in the short help text since AFAIK we still talk generally about "time to interactive" being important. I'll defer to others for specific language guidance. I'm fine with whatever.

cc: @paulirish

kaycebasques:
Oh, OK. I'm fine with that.

kaycebasques:
I removed the acronym and changed "time to interactive" to lowercase to indicate that we're talking about a general concept, not a specific metric

Issue #73, opened 2016-03-24
Cannot read property '_destroySSL' of undefined at destroySSL


gauntface:
Just tried running a lighthouse trace and hit the following issues:
- launch-chrome.sh is OS X. Would it be worth considering moving this to an npm script and using [which](https://www.npmjs.com/package/which) to pick the version of chrome to use
- Node 5.9.1 is required. We could add a check for >= 5.9 and throw a warning otherwise
- On linux I get the following error:

    TypeError: Cannot read property '_destroySSL' of undefined
    at destroySSL (_tls_wrap.js:370:7)

Going to dig into that one and try and figure out the problem but wanted to raise an issue on it..

gauntface:
Looks like:

    TypeError: Cannot read property '_destroySSL' of undefined
    at destroySSL (_tls_wrap.js:370:7)

is coming from the chrome remote interface `close()` call. Not sure there is anything that can be done to silence the error

paulirish:
I tried to diagnose this exception but only got so far. I'm pretty sure it's a node core bug, as that area has gotten some fixes.

paulirish:
> TypeError: Cannot read property '_destroySSL' of undefined at destroySSL (_tls_wrap.js:370:7)

Took another stab last night. AFAIK, this is due to the `request` call for the manifest, and not cri's `close()`. :relieved:  So we'll be done with it soon.

paulirish:
Broke out the other two issues into #106 and #105. 

Retitled this to be just the destroySSL. This will be fixed when #83 is done. \o/

paulirish:
fixed by #83

paulirish:
Just ran into this problem again, while working on `pwmetrics`.. here's what I found out:

the _destroySSL stuff is happening inside _tls_wrap, which the `request` module uses:
![image](https://cloud.githubusercontent.com/assets/39191/21671005/64045bf4-d2cd-11e6-8a01-0e811d2537b3.png)

As you can see, the `this` is passed as a parameter to `setImmediate` but a heartbeat later, that param isn't there. And why not?

`chrome-devtools-frontend` has been polyfilling setImmediate, and unfortunately for us... incorrectly: 
![image](https://cloud.githubusercontent.com/assets/39191/21671031/8c57cc1c-d2cd-11e6-85d0-9cbae2461825.png)

:)

ebidel:
[...]

PR #2420, opened 2017-06-01
core(lib): add error reporting

### LINKED ISSUE #2364: Non-fatal events to track

patrickhulce:
fixes #2364


brendankenny:
an overview of your design would be helpful (and be useful as docs for CONTRIBUTING). e.g. difference between expected errors and non-fatal ones, when to capture message vs exception, when things will be automatically collected and when it should be explicitly called (like the stack trace comment)

lighthouse-cli/bin.ts
```diff
@@ -175,10 +176,22 @@ export async function runLighthouse(
     url: string, flags: Flags, config: Object|null): Promise<{}|void> {
   let launchedChrome: LaunchedChrome|undefined;
 
+  if (typeof flags.disableErrorReporting === 'undefined') {
```
brendankenny: should be `enableErrorReporting`?

lighthouse-cli/bin.ts
```diff
@@ -175,10 +176,22 @@ export async function runLighthouse(
...
+    flags.disableErrorReporting = !isErrorReportingEnabled;
+  }
+
   try {
     launchedChrome = await getDebuggableChrome(flags);
     flags.port = launchedChrome.port;
-    const results = await lighthouse(url, flags, config);
+    const results = await lighthouse(url, flags, config, {
+      environment: isDev() ? 'development' : 'production',
+      serverName: 'redacted',
```
brendankenny: ?

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -122,10 +123,17 @@ class LoadFastEnough4Pwa extends Audit {
         }
 
         if (!areLatenciesAll3G) {
+          // eslint-disable-next-line max-len
```
brendankenny: move not needed?

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -122,10 +123,17 @@ class LoadFastEnough4Pwa extends Audit {
...
+          const debugString = `First Interactive was found at ${timeToFirstInteractive.toLocaleString()}, however, the network request latencies were not sufficiently realistic, so the performance measurements cannot be trusted.`;
+
+          Sentry.captureMessage('Network request latencies were not realistic', {
```
brendankenny: what is the level of detail we want to log for this sort of thing? I guess magnitude of the issue is something, and then can cross reference with the site being tested?

lighthouse-core/gather/driver.js
```diff
@@ -287,7 +288,7 @@ class Driver {
             }
 
             // If both the data and the url are empty strings, the page had no manifest.
-            return reject('No web app manifest found.');
+            return reject(ExpectedError(new Error('No web app manifest found.')));
```
brendankenny: this seems more like impetus to add a third return possibility (e.g. `null`)

lighthouse-core/gather/gather-runner.js
```diff
@@ -132,12 +133,15 @@ class GatherRunner {
...
+   * @param {string=} gathererName
    * @return {!Promise<*>}
    */
-  static recoverOrThrow(promise) {
+  static recoverOrThrow(promise, gathererName) {
     return promise.catch(err => {
       if (err.fatal) {
         throw err;
+      } else {
+        Sentry.captureException(err, {tags: {gatherer: gathererName}, level: 'warning'});
```
brendankenny: if this is just for non-fatal errors, may make more sense in `collectArtifacts` (or even `runner`)

lighthouse-core/gather/driver.js
brendankenny: (and this whole method will hopefully get fixed in https://github.com/GoogleChrome/lighthouse/pull/2349#discussion_r118798623)

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -77,7 +79,7 @@ class StartUrl extends Gatherer {
 
   afterPass(options, tracingData) {
     if (!this.startUrl) {
-      return Promise.reject(new Error('No start_url found inside the manifest'));
+      return Promise.reject(ExpectedError(new Error('No start_url found inside the manifest')));
```
brendankenny: if we go with `ExpectedError`, it should definitely be `new ExpectedError` or `createExpectedError` and should settle on whether to pass a string or an exception in

lighthouse-core/lib/sentry.js
```diff
@@ -0,0 +1,54 @@
+const log = require('./log');
+
+// Fix the polyfill. See https://github.com/GoogleChrome/lighthouse/issues/73
+self.setImmediate = function (callback) {
```
brendankenny: prefer rest args, but what did happen to the fix for #73?

lighthouse-core/lib/sentry.js
```diff
@@ -0,0 +1,54 @@
...
+  const args = [...arguments].slice(1);
+  Promise.resolve().then(() => callback(...args));
+  return 0;
+};
+
+const noop = () => undefined;
+const sentryDelegate = module.exports = {
```
brendankenny: this delegate is very hard to follow but you probably already know that :)

lighthouse-core/runner.js
```diff
@@ -33,6 +34,8 @@ class Runner {
 
     const config = opts.config;
 
+    Sentry.init(opts.flags.disableErrorReporting === false, opts.environmentData || {});
```
brendankenny: definitely should be `opts.flags.enableErrorReporting === true`

brendankenny: is having ` || {}` going to be confusing in the data? runner may want to try to infer more to differentiate different environments (but you might already be planning that)

brendankenny:
and, just to be clear, the structure of things seems to be what we've been discussing for a while/what I was imaginging, so it's looking good to me :)

The biggest unfortunate thing is the expected/regular/fatal exception split, but we've also been anticipating having to go down that path for a while as well. Maybe we can revisit resolving on a sentinel value instead of rejecting on an "expected" exception, but avoid going back to the -1 world by still having them automatically managed by gather-runner/runner.

lighthouse-cli/bin.ts
patrickhulce: that makes opt-out (which is the only usable flag) become `--no-enable-error-reporting` which feels odd, admittedly it definitely feels odd to check `disableErrorReporting` in runner but I didn't think we had any other case where we renamed the CLI flag from CLI to core

lighthouse-cli/bin.ts
patrickhulce: they send the serverName by default which is potentially private information we definitely we don't need to collect, this doesn't actually fix it but was trying it out

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: ya i'll revert

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: yeah saving the diff to 150ms sgtm

lighthouse-core/lib/sentry.js
patrickhulce: its still just waiting for us to bump devtools :)

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: passing a string with `createExpectedError` has the problem of sentry showing you the line where the error was created and not the more useful line here, and I thought you didn't like things extending from `Error` for some reason but I can't remember why? haha

lighthouse-core/gather/gather-runner.js
patrickhulce: why there? the method calling out that errors are explicitly handled was the first place that came to my mind or where this logic might live if I was looking for it, so open to much better explanations than that haha :)

lighthouse-core/runner.js
patrickhulce: mentioned above too, but why "definitely"?

lighthouse-cli/bin.ts
```diff
@@ -151,8 +152,8 @@ function saveResults(results: Results, artifacts: Object, flags: Flags) {
         return innerPromise.then((_: Results) => Printer.write(results, outputType, outputPath));
       }, Promise.resolve(results));
     } else {
-      const outputPath =
-          flags.outputPath || `${resolvedPath}.report.${typeToExtension(flags.output)}`;
+      const outputPath = flags.outputPath ||
```
patrickhulce: this was actually totally breaking vscode-insiders syntax highlighting :) filed an issue about it

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,67 @@
...
+
+async function prompt() {
+  if (!process.stdout.isTTY || process.env.CI) {
+    // Default non-interactive sessions to false
+    return false;
+  }
+
+  let timeout: NodeJS.Timer;
+
+  const prompt = inquirer.prompt([
```
paulirish: getting an exception in the prompt() method:

image

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,67 @@
+import {existsSync} from 'fs';
+import {join as joinPath} from 'path';
+import {inquirer, Configstore} from './shim-modules';
+const log = require('../lighthouse-core/lib/log');
+
+const MAXIMUM_WAIT_TIME = 20 * 1000;
+
+const MESSAGE = [
+  'Lighthouse would like to report back any errors that might occur while auditing. \n  ',
+  'Information such as the URL you are auditing, its subresources, your operating system, Chrome, ',
```
paulirish: this message is pretty huge. (See screenshot) can we break it up?
fwiw the message string [can be colorized](https://github.com/yeoman/insight/blob/a0430e2cb100e45c3d073ca80c7940c165b21f65/lib/index.js#L117) before going into prompt()

> **Lighthouse requests permission to anonymously report back runtime exceptions**
> This can include data such as the test URL, its subresources, your OS, Chrome version, and Lighthouse version. 
> **May _Lighthouse_ report this data to aid in improving the tool?**

lighthouse-cli/sentry-prompt.ts
patrickhulce: hmm strange, I can't repro, it happens every time?

lighthouse-cli/sentry-prompt.ts
patrickhulce: sure thing

patrickhulce:
> can you add us to the sentry project so we can see the results?

@paulirish yeah the free plan only allows one member though, time to apply for OSS account or whip out the gcard?

paulirish:
OSS acct for sho

lighthouse-core/runner.js
```diff
@@ -47,6 +51,19 @@ class Runner {
       return Promise.reject(err);
     }
 
+    const sentryContext = {
+      url: opts.url,
+      deviceEmulation: !opts.flags.disableDeviceEmulation,
+      networkThrottling: !opts.flags.disableNetworkThrottling,
+      cpuThrottling: !opts.flags.disableCpuThrottling,
+    };
+    Sentry.mergeContext({extra: Object.assign({}, environmentData.extra, sentryContext)});
```
paulirish: wdyt about moving this context setting stuff into lib/sentry ?

patrickhulce: ya done

patrickhulce:
A not insignificant portion of this PR is now changing the behavior of the StartUrl gatherer to pass the debugString instead of an error, so might be worth breaking that out into a separate PR at this point

brendankenny:
> so might be worth breaking that out into a separate PR at this point

I'd say 👍

lighthouse-cli/cli-flags.ts
```diff
@@ -56,6 +56,7 @@ export function getFlags() {
           ],
           'Configuration:')
       .describe({
+        'enable-error-reporting': 'Enables error reporting (on by default)',
```
brendankenny: maybe we can come up with a different flag name, as it isn't really the case that it's on by default (instead it prompts by default), and a first glance may alarm some

lighthouse-core/gather/gather-runner.js
brendankenny: IMO it's *just* for pulling fatal errors out of the stream. Non-fatal errors are handled on down the line (and due to a continuing lack of `fatal` errors, we should start to consider getting rid of it altogether :)

`collectArtifacts` is explicitly for collating the results of gatherers, including any errors, and has a slot ready made for [the sentry call](https://github.com/GoogleChrome/lighthouse/blob/8fce1465f376eddb88197142b5a2d0c86dc8eb66/lighthouse-core/gather/gather-runner.js#L307)

CONTRIBUTING.md
```diff
@@ -52,6 +52,52 @@ Don't:
...
+} catch (err) {
+  Sentry.captureException(err, {
+    tags: {audit: 'audit-name'},
+    level: 'warning',
+  });
+  doFallbackThing();
+}
+```
+
+#### If you throw an error in an audit to communicate audit failure instead of using the score property, *use createExpectedError*.
```
brendankenny: > If you throw an error in an audit to communicate audit failure instead of using the score property, *use createExpectedError*

after the changes to manifest fetching/start_url, are expected errors still needed? It sounds more like we should just discourage throwing errors to indicate an expected negative result

patrickhulce: yes I've removed createExpectedError entirely now, will update

patrickhulce:
filed and rebased against #2549

paulirish:
Blocked on #2549

patrickhulce:
alright folks I think we are ready for a re-review! 🎉

patrickhulce:
bump ⛰

lighthouse-cli/sentry-prompt.ts
paulirish: `yarn install-all` fixes. :)

paulirish:
Just ran it anew. Gave it one accidental exception and a deliberate one and its reporting well. 👍

lighthouse-core/lib/sentry.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+'use strict';
+
+const log = require('lighthouse-logger');
+
+// eslint-disable-next-line max-len
+const SENTRY_URL = 'https://a6bb0da87ee048cc9ae2a345fc09ab2e:63a7029f46f74265981b7e005e0f69f8@sentry.io/174697';
+
+// Fix the polyfill. See https://github.com/GoogleChrome/lighthouse/issues/73
+global.setImmediate = function(callback, ...argsForCallback) {
```
patrickhulce: this is what actually needs the fix, so I'm in favor of leaving here until we roll to newer version

lighthouse-cli/cli-flags.ts
```diff
@@ -104,6 +106,7 @@ Example: --output-path=./lighthouse-results.html`,
 
       // default values
       .default('chrome-flags', '')
+      .default('enable-error-reporting', true)
```
patrickhulce: then we won't even prompt? at that point I'd just abandon sentry entirely 😆

lighthouse-core/runner.js
```diff
@@ -36,6 +37,13 @@ class Runner {
       return Promise.reject(err);
     }
 
+    const sentryContext = Sentry.init(opts);
+    Sentry.captureBreadcrumb({
+      message: 'Run started',
+      category: 'lifecycle',
+      data: sentryContext,
+    });
```
patrickhulce: the point of the breadcrumb was to get easy access to the context (URL, throttling settings, etc), I'm not sure what we'd want to track at the end since we didn't have anything fatal along the way :)

lighthouse-cli/cli-flags.ts
```diff
@@ -58,6 +58,8 @@ export function getFlags(manualArgv?: string) {
           ],
           'Configuration:')
       .describe({
+        'enable-error-reporting':
+            'Enables error reporting (prompts by default, setting this flag to false will force off error reporting).',
```
patrickhulce: done

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,66 @@
+import {existsSync} from 'fs';
```
patrickhulce: done

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,66 @@
...
+import {join as joinPath} from 'path';
+
+import {Configstore, inquirer} from './shim-modules';
+
+const log = require('lighthouse-logger');
+
+const MAXIMUM_WAIT_TIME = 20 * 1000;
+// clang-format off
+const MESSAGE = [
+  `${log.reset}Lighthouse is requesting permission to anonymoously report back runtime exceptions.\n  `,
```
patrickhulce: heh

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,66 @@
...
+  `${log.reset}This can include data such as the test URL, its subresources, your OS, Chrome version, and Lighthouse version.\n  `,
+  `May ${log.green}Lighthouse${log.reset} ${log.bold}report this data to aid in improving the tool?`,
+].join('');
```
patrickhulce: done

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (typeof isErrorReportingEnabled === 'boolean') {
+    return isErrorReportingEnabled;
+  }
+
+  isErrorReportingEnabled = await prompt();
+  configstore.set('isErrorReportingEnabled', isErrorReportingEnabled);
+  return isErrorReportingEnabled;
+}
+
+export function isDev() {
```
patrickhulce: done

lighthouse-cli/test/cli/run-test.js
```diff
@@ -25,6 +25,7 @@ describe('CLI run', function() {
     const url = 'chrome://version';
     const filename = path.join(process.cwd(), 'run.ts.results.json');
     const flags = getFlags(`--output=json --output-path=${filename} ${url}`);
+    flags.enableErrorReporting = false;
```
patrickhulce: otherwise the mock and expected errors from tests get reported to sentry, this test skips bin.ts where we do the prompting/setting to false for CI

lighthouse-cli/run.ts
```diff
@@ -118,13 +118,13 @@ export function saveResults(results: Results, artifacts: Object, flags: Flags) {
 }
 
 export async function runLighthouse(
-    url: string, flags: Flags, config: Object|null): Promise<{}|void> {
+    url: string, flags: Flags, config: Object|null, environment?: Object): Promise<{}|void> {
```
patrickhulce: environmentData it is

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -111,6 +112,20 @@ class LoadFastEnough4Pwa extends Audit {
...
 
         if (!areLatenciesAll3G) {
+          const sentryContext = Sentry.getContext();
+          const hadThrottlingEnabled = sentryContext && sentryContext.extra &&
+              sentryContext.extra.networkThrottling;
+
+          if (hadThrottlingEnabled) {
+            const violatingLatency = firstRequestLatencies
+                .find(item => Number(item.latency) < latency3gMin);
+            Sentry.captureMessage('Network request latencies were not realistic', {
```
patrickhulce: yeah

lighthouse-cli/cli-flags.ts
brendankenny: > then we won't even prompt? at that point I'd just abandon sentry entirely 😆

arg, I think this is the second time I've mixed this up :)

Taking my confusion as a sign that it's generally confusing (not necessarily true :), and since it's not only encountered reading the code, but also in `--help` as `[default: true]`, would leaving it without a default and prompting on `cliFlags.enableErrorReporting === undefined` in `bin.ts` work instead? Other ideas?

patrickhulce: sure that'll do

patrickhulce: done

lighthouse-core/lib/sentry.js
brendankenny: > until we roll to newer version

patrickhulce: heh hey it's only been ~10 months anything can happen!

patrickhulce:
bump ⛰

package.json
```diff
@@ -75,13 +75,16 @@
   "dependencies": {
     "axe-core": "2.2.1",
     "chrome-devtools-frontend": "1.0.422034",
+    "configstore": "^3.1.0",
```
brendankenny: just a note so we remember, we should explore `optionalDependencies` for these (and the other shimmed ones) in the future to at least signal that these aren't necessary for folks looking for a slimmer bundle of dependencies.

brendankenny:
PTAL

lighthouse-core/runner.js
```diff
@@ -36,6 +37,13 @@ class Runner {
       return Promise.reject(err);
     }
 
+    const sentryContext = Sentry.init(opts);
```
patrickhulce: Not clear what you're proposing to me since we need the options before we can start sending the errors. If you're talking about buffering the events and sending them after Sentry has been initialized that sounds like future work than initial introduction of error reporting to me :)

lighthouse-cli/cli-flags.ts
```diff
@@ -58,6 +58,8 @@ export function getFlags(manualArgv?: string) {
           ],
           'Configuration:')
       .describe({
+        'enable-error-reporting':
+            'Enables error reporting (prompts once by default, setting this flag will force error reporting to that state).',
```
patrickhulce: done

lighthouse-core/runner.js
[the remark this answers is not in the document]
brendankenny: I was talking about the first thing, not buffering earlier events or anything (maybe we could eventually do that for CLI errors or something but "doesn't ever start" is the kind of thing that people would actually report manually, so probably isn't a pressing issue)

> this is mostly the earliest point we have all the available information to initialize

Does it need more than the `environmentData` made in `bin.ts` and `flags`? It seems like the `Sentry.init({environmentData, flags: cliFlags})` call could just be made there (or `lighthouse-core/index.js` might be better), but maybe I'm missing a step that's happening in between

patrickhulce:
alright I've rebased and kicked the tires of

* timeout flow
* CI flow
* answer yes/no flow
* remembered yes/no flow
* explicit yes/no flow
* failure report flow
* extension

anything I'm missing?

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+import {Configstore, inquirer} from './shim-modules';
+
+const log = require('lighthouse-logger');
+
+const MAXIMUM_WAIT_TIME = 20 * 1000;
+// clang-format off
+const MESSAGE =
+  `${log.reset}Lighthouse is requesting permission to anonymously report back runtime exceptions.\n  ` +
```
paulirish: I'm reading this corp doc: https://goto.google.com/gpxxy on how to handle consent prompts..

Basically we need to clearly describe:

1. What this is about
2. What are the benefits to the user
3. What are the consequences to the user
4. What apps are affected
5. What is the name of this feature/thing and how can a user change their setting in the future
6. Where can the user learn more

There's some other UX suggestions which are pretty good, too. it's a nice read.


-------------

Let's create a `docs/error-reporting.md` which has some end-user-facing documentation on this. including:

  * how to change the setting (delete `~/.config/configstore/lighthouse.json` ?)  
  * the stuff you already put into the readme.
  * reported data includes test URL, its subresources, your OS, Chrome version, and Lighthouse version

And I'm thinking we tweak our consent opt-in message. (I got some inspiration from `yo`'s one)

Currently:

> Lighthouse is requesting permission to anonymously report back runtime exceptions.
> This can include data such as the test URL, its subresources, your OS, Chrome version, and Lighthouse version.
> May Lighthouse report this data to aid in improving the tool?

Proposed:

> We're constantly trying to improve Lighthouse and its reliability
> May we anonymously report runtime exceptions to improve the tool over time?
> Details: https://github.com/GoogleChrome/lighthouse/blob/master/docs/error-reporting.md

lighthouse-cli/sentry-prompt.ts
```diff
@@ -0,0 +1,63 @@
...
+  `${log.reset}This can include data such as the test URL, its subresources, your OS, Chrome version, and Lighthouse version.\n  ` +
+  `May ${log.green}Lighthouse${log.reset} ${log.bold}report this data to aid in improving the tool?`;
+// clang-format on
+
+async function prompt() {
+  if (!process.stdout.isTTY || process.env.CI) {
+    // Default non-interactive sessions to false
```
paulirish: wdyt about defaulting CI to true?

patrickhulce: I'm split, I'm definitely against in *our* CI so an opt-out env variable seems good.

It'd be nice to gather errors from folks running Lighthouse in CI, but silently opting them in seems bad. Can always use `--enable-error-reporting` to skip this part entirely and run with error reporting so I'd vote no to default CI true.

lighthouse-cli/sentry-prompt.ts
patrickhulce: done and done

lighthouse-cli/sentry-prompt.ts
patrickhulce: I do like the idea of enabling this for our smoke tests on master though  👍  followup work

paulirish: ok sg

lighthouse-core/runner.js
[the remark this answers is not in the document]
paulirish: talked to brendan about this. he doesn't like that we have to pass along this environment data from file to file, augmenting method signatures along the way. the environment data in question: https://github.com/GoogleChrome/lighthouse/blob/49c061c323fcf7767d7ef04f112fb60515f6f3ec/lighthouse-cli/bin.ts#L72-L78

i'm like "sure, fine". 🦊 

**proposal:**

* move the setup of `name`, `environment`, and `release` into runner.js. no reason why it shouldn't be defined then.
* kill `tags.channel` because hey we are only recording from the cli. deal with the channels problem later when we have channels.
* restore the method signatures of runLighthouse and runner.run etc

brendan read this and is like "cool". 🐼 

@patrickhulce wdyt?

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
...
+## What data gets reported?
+
+* The URL you tested
+* The runtime settings used (throttling enabled/disabled, emulation, etc)
+* The message, stack trace, and associated data of the error
+* Your Lighthouse version
+* Your Chrome version
+* Your operating system
+
+## How do I opt-in/opt-out?
```
brendankenny: maybe just have "How do I opt-in?" to be clear it's only opt-in?

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
...
+The first time you run the CLI you will be prompted with a message asking you if Lighthouse can anonymously report runtime exceptions. You can give a direct response of `yes` or `no` to save your response, and you will not be prompted again. If no response is given within 20 seconds, a `no` response will be assumed, and you will not be prompted again. Non-interactive terminal sessions and invocations with the `CI` environment variable set will automatically be opted out.
```
brendankenny: "You can give a direct response of `yes` or `no` to save your response, and you will not be prompted again." -> "You can give a direct response of `yes` or `no`, and you will not be prompted again.", maybe?

brendankenny: "will automatically be opted out." -> "will automatically not be opted in." maybe? (slightly more awkward but clearer there is no state that needs to be opted out of :)

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
...
+
+The CLI also has two flags to control error reporting that will override the saved preference. Running Lighthouse with `--enable-error-reporting` will report errors regardless of the saved preference, and running Lighthouse with `--no-enable-error-reporting` will *not* report errors regardless of the saved preferences.
+
+## How do I change my opt-in/opt-out preference?
```
brendankenny: `How do I change my opt-in preference?`?

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
+# Error Reporting Explained
+
+## What's going on?
+
+The Lighthouse team is constantly trying to improve the reliability of our tools. We've added functionality to anonymously report runtime exceptions, given your consent, using [Sentry](https://sentry.io/welcome/). We will use this information to detect new bugs and avoid regressions.
```
brendankenny: maybe
`"If you give your consent, we've added functionality to anonymously report runtime exceptions using [Sentry](https://sentry.io/welcome/)."`

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
...
+
+Only CLI users are currently impacted. Node module consumers are opted out by default without a prompt, and the DevTools and extension integrations have no automatic error reporting functionality.
```
brendankenny: maybe just include Node module consumers with Devtools and extension users for now since it won't be happening for them?

docs/error-reporting.md
```diff
@@ -0,0 +1,30 @@
...
+
+## What will happen if I opt-in?
+Runtime exceptions will be reported to the team along with information on your environment such as the URL you tested, your OS, and Chrome version. See [what data gets reported](#what-data-gets-reported).
+
+## What will happen if I opt-out?
```
brendankenny: `"What will happen if I don't opt-in?"`?

docs/error-reporting.md
paulirish: also "y" and "n" work. 
and just hitting enter defaults to "no". this should be mentioned as well.

docs/error-reporting.md
patrickhulce: heh, but the functionality is there if they give consent or not, the phrase is modifying "report" ;)

rephrased

docs/error-reporting.md
patrickhulce: yeah was trying to be explicit that the code isn't even there, but that was probably confusing

docs/error-reporting.md
patrickhulce: done

docs/error-reporting.md
patrickhulce: done

docs/error-reporting.md
patrickhulce: done

docs/error-reporting.md
patrickhulce: done

paulirish:
@patrickhulce change lgtm

heads up: this discussion is unresolved https://github.com/GoogleChrome/lighthouse/pull/2420/files#r145858868

lighthouse-core/runner.js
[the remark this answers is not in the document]
patrickhulce: kk, couple thoughts

the only signature that is modified is runLighthouse, runner.run is untouched
doing an fs.existsSync on .git directory directly in the node module and browser bundled code for a purely cli concern feels nastier than an extra optional argument, another consequence is that the package.json will also get pulled in as a dependency for the bundle when previously it was not 

totally cool with nuking tags.channel for now, but doesn't seem like it's hurting anything if we stick with passed in environmentData

if the signature disruption is a concern, how about we move Sentry.init to CLI land then (I believe this was brendan's original suggestion?)

paulirish:
@brendankenny lets do thisssssssssss. @patrickhulce now understood what you were going for in that init() thread. :)

lighthouse-cli/run.ts
```diff
@@ -138,13 +138,13 @@ export function saveResults(results: Results, artifacts: Object, flags: Flags) {
 }
 
 export async function runLighthouse(
-    url: string, flags: Flags, config: Object|null): Promise<{}|void> {
+    url: string, flags: Flags, config: Object|null, environmentData?: Object): Promise<{}|void> {
```
brendankenny: not needed anymore?

lighthouse-cli/bin.ts
```diff
@@ -59,6 +65,23 @@ if (cliFlags.output === Printer.OutputMode[Printer.OutputMode.json] && !cliFlags
...
+      name: 'redacted', // prevent sentry from using hostname
+      environment: isDev() ? 'development' : 'production',
+      release: pkg.version,
+      tags: {
+        channel: 'cli',
+      },
+    },
+  });
+
+  return await runLighthouse(url, cliFlags, config);
```
brendankenny: nit: since the function is async anyway, can drop the `await` here

patrickhulce: done

lighthouse-cli/run.ts
patrickhulce: done

paulirish:
epic PR. super excited about this. \o/

lighthouse-core/runner.js
zixuan75: ok right

Issue #2276, opened 2017-05-16
Operation Catharsis


paulirish:
Just a meta tracking issue for capturing the cleanup tasks we've been deliberately deferring.

![image](https://cloud.githubusercontent.com/assets/39191/26121422/fcf41d6e-3a28-11e7-8455-2f273ebad18d.png)

For example:

* [ ] move v2 report up a directory into `lighthouse-core/report/`
* [ ] slim down LHR https://github.com/GoogleChrome/lighthouse/issues/1999
* [x] remove v1 report, incl handlebars, gulp for core ~#2596~
* [x] remove TTI v1 impl ~#2452~
* [x] remove superfluous quotes in config/default.js ~#2445~
* [x] remove v1 aggregations from config/default.js and codebase ~#2426~
* [x] simultaneous build/install: (`yarn install-launcher &  yarn install-cli &  yarn install-extension & yarn install-viewer & wait`) ~#2381~

We'll dig in to this right after 2.0.

Feel free to comment with the other stuff you're looking forward to.

brendankenny:
> remove v1 report, incl handlebars, gulp for core

I started this in a branch ([`rmv1`](https://github.com/GoogleChrome/lighthouse/tree/rmv1)) but realized it would break all the viewer tests since they currently rely on v1. We could skip them for now, but would rather get #2338 underway first

PR #2426, opened 2017-06-03
remove aggregations

brendankenny:
part of #2276

it is time to say goodbye :)


brendankenny:
the diff on `runner-test.js` looks like more is changing than actually is.

[`?w=1`](https://github.com/GoogleChrome/lighthouse/pull/2426/files?w=1#diff-9476be2df7d812bc50da0dd73d73f2f8) makes it clearer that it's just one test being removed and one test switching from a config with an aggregation to one with a category

paulirish:
ya got some conflicts now :)

brendankenny:
all ✅ and travis says 💚 so farewell aggregations

brendankenny:
(note: aggregations died on the way back to their home planet)

paulirish:
best best.

stramel:
6k deletions 😂😂😂

PR #2427, opened 2017-06-03
add units to LoadFastEnough debug string

brendankenny:
tiny change. From

`First Interactive was found at 808.352, however, the network request latencies were not sufficiently realistic, so the performance measurements cannot be trusted.`

to

`First Interactive was found at 808.352ms, however, the network request latencies were not sufficiently realistic, so the performance measurements cannot be trusted.`

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -116,7 +116,7 @@ class LoadFastEnough4Pwa extends Audit {
           return {
             rawValue: false,
             // eslint-disable-next-line max-len
-            debugString: `First Interactive was at ${timeToFirstInteractive.toLocaleString()}. More details in the "Performance" section.`,
+            debugString: `First Interactive was at ${timeToFirstInteractive.toLocaleString()}ms. More details in the "Performance" section.`,
             extendedInfo
           };
         }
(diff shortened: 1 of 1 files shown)


lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -116,7 +116,7 @@ class LoadFastEnough4Pwa extends Audit {
           return {
             rawValue: false,
             // eslint-disable-next-line max-len
-            debugString: `First Interactive was at ${timeToFirstInteractive.toLocaleString()}. More details in the "Performance" section.`,
+            debugString: `First Interactive was at ${timeToFirstInteractive.toLocaleString()}ms. More details in the "Performance" section.`,
```
paulirish: we separate numbers and unit with `\xa0` these days.

but you might as well just use `Util.formatMilliseconds` which will round it to the tens place, fixing your other concern with these numbers. :)

brendankenny: good call, updated

paulirish:
ready to land. just waiting for travis

brendankenny:
rebased to get the JavaScriptDialog change. Now I understand why you were appeared on that commit @paulirish :)

PR #2432, opened 2017-06-04
Fix US-centric wording

khempenius:
Not everyone uses dollars as their currency.


khempenius:
@patrickhulce Oops, sorry about that. It should now be fixed.

PR #2443, opened 2017-06-05
update smokehouse PWA expectations

brendankenny:
updates the PWA smokehouse run so that it runs all the latest PWA checks (and will stay current with them going forward) and updates expectations to match

### DIFF
--- a/lighthouse-core/gather/computed/manifest-values.js
+++ b/lighthouse-core/gather/computed/manifest-values.js
@@ -67,7 +67,7 @@ class ManifestValues extends ComputedArtifact {
       {
         id: 'shortNameLength',
         failureText: 'Manifest `short_name` will be truncated when displayed on the homescreen',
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/computed/manifest-values.js
```diff
@@ -67,7 +67,7 @@ class ManifestValues extends ComputedArtifact {
       {
         id: 'shortNameLength',
         failureText: 'Manifest `short_name` will be truncated when displayed on the homescreen',
-        validate: manifest => manifest.value.short_name.value &&
+        validate: manifest => !!manifest.value.short_name.value &&
```
brendankenny: added the `!!` so that it's always a boolean. If the site had no `short_name`, this would return `undefined` which would then get `JSON.stringified` out of existence in the JSON report

ebidel: good catch

PR #2445, opened 2017-06-06
de-jsonify default.js

brendankenny:
`default.js` was [once](https://github.com/GoogleChrome/lighthouse/pull/1911) `default.json`, and still has the double quotes to prove it. This updates to remove those, add a license header, enable eslint, and sprinkle in some trailing commas to hopefully help with `git blame` in the future while this commit is ruining it in the present :)

Disabled `max-len` for this file because of the description strings and this is a config file, not a normal script file, but happy to turn back on if others want it.

part of #2276

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -1,303 +1,321 @@
-/* eslint-disable */
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License');
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+'use strict';
+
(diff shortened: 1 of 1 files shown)


lighthouse-core/config/default.js
```diff
@@ -1,303 +1,321 @@
...
-        {"id": "webapp-install-banner", "weight": 1},
-        {"id": "splash-screen", "weight": 1},
-        {"id": "themed-omnibox", "weight": 1},
-        {"id": "viewport", "weight": 1},
-        {"id": "content-width", "weight": 1},
-        {"id": "pwa-cross-browser", "weight": 0, "group": "manual-pwa-checks"},
-        {"id": "pwa-page-transitions", "weight": 0, "group": "manual-pwa-checks"},
-        {"id": "pwa-each-page-has-url", "weight": 0, "group": "manual-pwa-checks"}
+  categories: {
+    'pwa': {
```
paulirish: you can unquote these category titles

lighthouse-core/config/default.js
```diff
@@ -1,303 +1,321 @@
-/* eslint-disable */
+/**
```
patrickhulce: nit: small headers now :)

lighthouse-core/config/default.js
brendankenny: > you can unquote these category titles

eslint doesn't like because `best-practices` is on the same level and needs quotes

lighthouse-core/config/default.js
brendankenny: > nit: small headers now :)

lulz I was hoping I'd get merged first :)

PR #2451, opened 2017-06-07
fix: better debug message for missing network times

patrickhulce:
Improves the debug information when network throughput can't be computed and defaults the result to Infinity instead of 0 which makes more sense since the only use of networkThroughput is to be the divisor (the computed artifact is also used by total byte weight which makes sense to say 0 ms time and still report the byte savings)

This was affecting chrome URLs which don't have any timing information, results from before and after for `chrome://version` below (the optimized images error is for tainting canvas which apparently all chrome images do)

**Before**
image

**After**
image

### DIFF
--- a/lighthouse-core/gather/computed/network-throughput.js
+++ b/lighthouse-core/gather/computed/network-throughput.js
@@ -15,6 +15,7 @@ class NetworkThroughput extends ComputedArtifact {
   /**
    * Computes the average throughput for the given records in bytes/second.
    * Excludes data URI, failed or otherwise incomplete, and cached requests.
+   * Returns Infinity if there were no analyzable network records.
    *
    * @param {?Array<!WebInspector.NetworkRequest>} networkRecords
    * @return {number}
--- a/lighthouse-core/gather/computed/network-throughput.js
+++ b/lighthouse-core/gather/computed/network-throughput.js
@@ -20,9 +20,7 @@ class NetworkThroughput extends ComputedArtifact {
    * @return {number}
    */
   compute_(networkRecords) {
-    if (!networkRecords || !networkRecords.length) {
-      return 0;
-    }
+    networkRecords = networkRecords || [];
 
     let totalBytes = 0;
     const timeBoundaries = networkRecords.reduce((boundaries, record) => {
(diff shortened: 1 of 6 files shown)


lighthouse-core/gather/computed/network-throughput.js
```diff
@@ -38,6 +36,10 @@ class NetworkThroughput extends ComputedArtifact {
       return boundaries;
     }, []).sort((a, b) => a.time - b.time);
 
+    if (!timeBoundaries.length) {
```
brendankenny: can you document this in the jsdoc?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -193,6 +193,12 @@ class CategoryRenderer {
     const descriptionEl = this._dom.createChildOf(element, 'div', 'lh-perf-hint__description');
     descriptionEl.appendChild(this._dom.convertMarkdownLinkSnippets(audit.result.helpText));
 
+    if (audit.result.debugString) {
```
brendankenny: what is this adding that wasn't there before? Not clear from the screenshot

lighthouse-core/test/audits/byte-efficiency/byte-efficiency-audit-test.js
```diff
@@ -85,7 +85,7 @@ describe('Byte efficiency base audit', () => {
         {wastedBytes: 450, totalBytes: 1000, wastedPercent: 50},
         {wastedBytes: 400, totalBytes: 450, wastedPercent: 50},
       ],
-    });
+    }, 1000);
```
brendankenny: need new tests, too

lighthouse-core/gather/computed/network-throughput.js
patrickhulce: done

lighthouse-core/test/audits/byte-efficiency/byte-efficiency-audit-test.js
patrickhulce: done

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: it was, but not if it didn't completely fail (this was sort of drive by, there are other debugStrings that can be set)

patrickhulce: added a test now for it

brendankenny: > it was, but not if it didn't completely fail

it was...what? :P

patrickhulce:
> a new test in network-throughput-test.js to replicate what was coming in from chrome:// URLs would be good too, but otherwise LGTM

done

brendankenny:
travis is greeeeeen

PR #2452, opened 2017-06-07
refactor: remove old TTI code

patrickhulce:
catharsis #2276


patrickhulce:
travis is happy except for the compile-devtools step which appears to be a devtools problem broken on master

PR #2453, opened 2017-06-07
remove beta badge from extension

paulirish:
It's been gone from the report for a while.

But it's remained in the popup:
image

This nukes it.

### DIFF
--- a/lighthouse-core/report/styles/report.css
+++ b/lighthouse-core/report/styles/report.css
@@ -257,17 +257,6 @@ summary {
   background-blend-mode: luminosity;
 }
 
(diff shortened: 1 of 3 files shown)


paulirish:
While I mostly want to do this I'll add two items we can bikeshed:

1. it's actually a rather prominent link to github. maybe there's value in that. (but since there is no status-bar, it's a blind click and probably not all that useful)
2. we could hypothetically use the badge to indicate the version number instead.

ebidel:
We have links to GitHub and version numbers in the report. The extensions management page also had the version. We good without it :)

PR #2454, opened 2017-06-07
refactor(trace-of-tab): return timestamps in microseconds

### LINKED ISSUE #2116: .timestamps should always be in microseconds
We `x / 1000` in a number of places on these timestamp values.  It's a little odd. I asked patrick if he had a guess why and he shared this punctuation-free explanation:

> something (fmp maybe?) was exposed as ts / 1000 and as the metrics got collected they're all used that way but it's now this hideous monster of misunderstand why they're not the real timestamps

We should keep these integers in their original value.
We'll break pwmetrics with this change, but that's okay. :)

This is low-priority relative to everything else but we should include it in 2.0

patrickhulce:
fixes #2116

### DIFF
--- a/lighthouse-core/audits/consistently-interactive.js
+++ b/lighthouse-core/audits/consistently-interactive.js
@@ -54,7 +54,7 @@ class ConsistentlyInteractiveMetric extends Audit {
    * @return {!Array<!TimePeriod>}
    */
   static _findNetworkQuietPeriods(networkRecords, traceOfTab) {
-    const traceEnd = traceOfTab.timestamps.traceEnd;
+    const traceEndTsInMs = traceOfTab.timestamps.traceEnd / 1000;
 
     // First collect the timestamps of when requests start and end
     const timeBoundaries = [];
@@ -94,7 +94,7 @@ class ConsistentlyInteractiveMetric extends Audit {
 
     // Check if the trace ended in a quiet period
     if (numInflightRequests <= ALLOWED_CONCURRENT_REQUESTS) {
-      quietPeriods.push({start: quietPeriodStart, end: traceEnd});
+      quietPeriods.push({start: quietPeriodStart, end: traceEndTsInMs});
     }
 
     return quietPeriods;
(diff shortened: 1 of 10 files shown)


brendankenny:
meta comment: where should we document this? It's good to know (and important to know when you do anything with these :)

Probably trace-of-tab, but maybe also in https://github.com/GoogleChrome/lighthouse/blob/master/docs/architecture.md?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -107,8 +107,8 @@ class TraceOfTab extends ComputedArtifact {
     const timestamps = {};
 
     Object.keys(metrics).forEach(metric => {
-      timestamps[metric] = metrics[metric] && metrics[metric].ts / 1000;
-      timings[metric] = timestamps[metric] - timestamps.navigationStart;
+      timestamps[metric] = metrics[metric] && metrics[metric].ts;
+      timings[metric] = (timestamps[metric] - navigationStart.ts) / 1000;
```
brendankenny: I do wish we had a better name for these than `timings`, but may be too late :)

lighthouse-core/audits/consistently-interactive.js
```diff
@@ -54,7 +54,7 @@ class ConsistentlyInteractiveMetric extends Audit {
    * @return {!Array<!TimePeriod>}
    */
   static _findNetworkQuietPeriods(networkRecords, traceOfTab) {
-    const traceEnd = traceOfTab.timestamps.traceEnd;
+    const traceEndTsInMs = traceOfTab.timestamps.traceEnd / 1000;
```
brendankenny: nit: move this down to where it's used?

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: ha a battle for another day :)

lighthouse-core/audits/consistently-interactive.js
patrickhulce: aw I liked that each function in this file got this aliasing out of the way at the beginning :( 
did it make you look for its use too much right afterwards?

brendankenny: ha, yeah, exactly. I figured that's what this was doing, but then my next thought was, "wait, what is this even used for?"

patrickhulce:
got another PR comin' with some more groovy docs 📜  📝

PR #2456, opened 2017-06-07
local implementation of LogNormalDistribution

brendankenny:
creates a `lib/statistics.js` as a new home for a local implementation of `LogNormalDistribution`. I didn't change the interface at all, though it may be a little silly to keep the same interface when we literally only use the `computeComplementaryPercentile` method. Happy to take opinions on that :)

This is our last dependency on traceviewer, so also removes `third_party/traceviewer-js/`. Reduces `lighthouse-background.js` by about another 100KB.


brendankenny:
🔢 📈 📉

PR #1174, opened 2016-12-17
Reject when main request fails

Janpot:
Tests might need some reordering

PR #2461, opened 2017-06-07
fix(http-redirect): do not error if no server listens on http

### LINKED ISSUE #1978: Site Redirects HTTP Traffic to HTTPS shouldn't stop the execution if failling
Site Redirects HTTP Traffic to HTTPS shouldn't stop the execution if failling

Actually if you don't server http at all, the test is faillingbut it also stop the process and the report is not generated

See logs
[logLightHouse.txt](https://github.com/GoogleChrome/lighthouse/files/905585/logLightHouse.txt)

### LINKED ISSUE #2459: Unable to load the page: timeout reached - Server misconfigured
**Initial URL**: https://mysite.mwhtools.com/mySiteApp/
This is in the new audit tool and all the options are checked

patrickhulce:
fixes #2459 (and probably closes #1978)

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -51,6 +51,8 @@ module.exports = {
   {
     passName: 'redirectPass',
     useThrottling: false,
+    // Move on even if there's no HTTP server
+    assertPageLoaded: false,
     // Just wait for onload
     networkQuietThresholdMs: 0,
     // Speed up the redirect pass by blocking stylesheets, fonts, and images
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -252,7 +252,9 @@ class GatherRunner {
       log.log('status', status);
       const devtoolsLog = driver.endDevtoolsLog();
       const networkRecords = NetworkRecorder.recordsFromLogs(devtoolsLog);
-      GatherRunner.assertPageLoaded(options.url, driver, networkRecords);
+      if (config.assertPageLoaded !== false) {
+        GatherRunner.assertPageLoaded(options.url, driver, networkRecords);
+      }
       log.verbose('statusEnd', status);
 
       // Expose devtoolsLog and networkRecords to gatherers


patrickhulce:
> I wonder if we can flip around the assertion. We don't actually care (here) that the url is the same as what was initially requested (+ redirects) as assertPageLoaded is checking. The real purpose of #1174 was to fast-fail in cases that LH would end up checking Chrome's offline or error page because it can be easy to miss that was happening and you end up assuming the report was for the test site.

Don't really understand what you're proposing here. Only assert that the URL was the same for the offline pass? I think that would still fall prey to all your objections to this approach, so I must be misunderstanding.

> Would it be possible for us to instead get better at checking if we loaded a UA-provided error page and stopping the run only in that case?

It seems wise to make sure we got a response for the URL that they asked us to check (though we could definitely loosen it to accept HSTS pre-loaded urls). This is a relatively special case where we muck with the URL and try to visit a URL that we don't actually know is valid, so it seems reasonable to have special logic for URLs we come up with ourselves.

> We'll still have to get better at detecting 301 and HSTS and pre-pinned redirects and whatever else is tricking our redirect watcher in driver.

Do we have a reproducible case where this is actually messing with the gatherer? We can definitely make some improvements to the redirect watcher, but that's somewhat orthogonal to fixing http-redirect since at the very least we could simple `evaluateAsync(window.location.href)` and check its https without improving the redirect watcher at all.

brendankenny:
Yeah, I don't know if this is actually workable, I just wanted to make sure to explore if our check could be smarter rather than adding another config option, putting the decision (and the need to remember to make the decision) of whether to assert on us instead of users.

> This is a relatively special case where we muck with the URL

This is the biggest problem and the way this gatherer has to be special is the worst. It would be really nice if we could just curl for headers :)

My half baked idea:

1) Prominently warn if we're testing a different URL than was entered (#1917). On the one hand this may be noisy if the redirect was expected, so maybe we don't warn about (or warn less prominently) about some kinds of redirects. This is more of an issue with the CLI, since generally the extension and DevTools will be testing a tab already redirected to the final URL (the big exception is if the site redirects to a mobile URL when emulating). This is also where we need to get better at some of the remaining redirect-tracking failure cases so we'll stop getting tripped up by the [`!mainRecord` clause in `assertPageLoaded`](https://github.com/GoogleChrome/lighthouse/blob/d59fd6dcad28ed5099778d0456632693b8c8fb0b/lighthouse-core/gather/gather-runner.js#L145) and can only watch the actual failures.

2) Fail the run in some way when the site actually fails to load. With (1) implemented, this could become much narrower in scope, limited to cases where the request really failed. That still leaves this *expected* 404 case, when the redirect gatherer tries to load an HTTP URL that doesn't exist. One idea:

   The `assertPageLoaded` failure mode could be less fatal. Rather than exiting the whole LH run on a request failure, we could just throw errors on all the `afterPass` slots on the gatherers in that pass, which will get bundled up and reported as audit errors (and so a failure in the report). The `redirects-http` audit would either pass if redirect was successful, fail with a `debugString` if the http site doesn't redirect, or fail with an Error if there is no http site. Not perfect categorization, but, again, this is the worst of all gathererers :)

In the case that the URL was entered incorrectly (or a test site is accidentally not deployed yet to that URL or whatever) and all passes error, the report would be full of errors that the site wouldn't load. We'd likely want to be smarter here by surfacing a clearer warning...e.g. if there is no `finalUrl`, propagate a top of the page warning for that too.

brendankenny:
I can try out an implementation of this tomorrow if you'd rather look at other stuff :)

patrickhulce:
> I can try out an implementation of this tomorrow if you'd rather look at other stuff :)

aight you sound pretty passionate I won't try and stop you ;)

paulirish:
> I can try out an implementation of this tomorrow if you'd rather look at other stuff :)

@brendankenny i believe the next action on this PR is yours

patrickhulce:
@brendankenny your enthusiasm to do this yourself doesn't seem to have held up over the last month ;) wanna give this another look?

paulirish:
@brendankenny described his approach and @patrickhulce is happy with it.

wardpeet:
we need this :D

patrickhulce:
agreed, @brendankenny any progress in the six weeks since "tomorrow"? ;)

paulirish:
just testing.. please ignore:

Closed nothing.

paulirish:
Imma merge.

brendankenny:
Sorry, out of the office today, but I still have that fix so we can silence the bugs.

PR #2466, opened 2017-06-08
Theme Color wording fixes

khempenius:
Updated description
- Lighthouse has no idea what your brand colors are - it's really just checking to see whether the address bar has been themed in any way.

Corrected reference to manifest theme color ("theme_color" vs. "theme-color")

Changes for clarity:
- I think "apply the same theme site-wide" could be misinterpreted as altering the web site appearance (vs. the browser) in some way. It's also slightly ambiguous as to whether "applying a theme" is the same thing as "upgrading the address bar", or whether it represents a larger subset of functionalities.

### DIFF
--- a/lighthouse-core/audits/themed-omnibox.js
+++ b/lighthouse-core/audits/themed-omnibox.js
@@ -27,7 +27,7 @@ class ThemedOmnibox extends MultiCheckAudit {
     return {
       category: 'PWA',
       name: 'themed-omnibox',
-      description: 'Address bar matches brand colors',
+      description: 'Address bar has been themed',
       helpText: 'The browser address bar can be themed to match your site. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/address-bar).',
       requiredArtifacts: ['Manifest', 'ThemeColor']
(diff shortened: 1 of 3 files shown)


patrickhulce:
aside: revisiting those docs I'm not sure we should require the manifest value to be set at all since we require meta theme-color...

> If you page already has a theme-color meta tag — for example <meta name="theme-color" content="#2196F3"> — then the page level configuration will be used instead of the value in the manifest.

ebidel:
Some internal discussion on whether the address bar theming should be required: https://goo.gl/kXzeO0

FWIW, it's currently not part of the Baseline PWA checklist.

brendankenny:
@kaycebasques if you have the time to take a look

brendankenny:
> If you page already has a theme-color meta tag — for example — then the page level configuration will be used instead of the value in the manifest.

wasn't there still some difference between the two? I can never keep track. We should just match the checklist, though, which makes these decisions easier (i.e. somebody else's job to keep track :)

brendankenny:
hi @khempenius, sorry for the slow movement on this. Would you mind rebasing with current master? The main thing that changed since you opened the PR was the themed-omnibox `helpText` changed (in #2487).

I would resolve the merge conflict on your behalf, but I believe when I've done that before and then squashed, the commit author was switched to me, so that's no good :)

khempenius:
@brendankenny No worries. This should be good to go now.

brendankenny:
Thanks!

Issue #2293, opened 2017-05-17
[meta] List of sites that give Lighthouse trouble


patrickhulce:
## To Investigate
- [ ] https://h5.m.taobao.com/trip/home/index.html, from #3838, "no DCL in trace", "no screenshots in trace" 100% reproducible

## Axe failures
- [ ] http://visiondirect.com (from GTech) 🐌 🐌 🐌  invalid selector, tracked by https://github.com/dequelabs/axe-core/issues/563

## Service worker/offline issues
- [ ] https://www.lenskart.com/ tracked by #3724 
- [ ] https://next.weekplan.net
- [ ] https://www.flipkart.com from #3736

## Stopped tracing too early 
- [ ]  https://www.instagram.com/p/BNQ98s4BTEa/ ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302154750)) tracked by #3734
- [ ] https://weather.com ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302233951))
- [ ] https://www.davidemaggio.it/  ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-309488204))
- [ ] http://lampsplus.com (from GTech)
- [ ] http://valuecityfurniture.com (from GTech)

## Site misbehavior (aborted request, failed document, extremely slow server, etc)
- [ ]  https://app.jalantikus.com/ ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302154750))
- [ ] http://createspace.com (from GTech)
- [ ] http://gifts.com (from GTech)
- [ ] https://music.amazon.com/ (from CDS) tracked by #3735

## Working but needed more than 30s
- [x]  http://mcdcc.org/ 😢 
- [x]  http://pwa.voot.com/ ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302154750)) 💨 🐌 
- [x] http://m.aliexpress.com ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302155884)) needs ~60+ seconds!
- [x] https://nytimes.com ([from comment](https://github.com/GoogleChrome/lighthouse/issues/2293#issuecomment-302233951)) 💨 🐌 
- [x] http://hollisterco.com (from GTech) 🐌  
- [x] http://petfooddirect.com (from GTech) 🐌  needs ~60+ seconds!
[...]

PR #2473, opened 2017-06-09
fix(driver): wait for CPU idle

### LINKED ISSUE #2475: Persistent 'Audit error: The main thread was busy for the entire trace recording.' when running via webpagetest
As mentioned in #2434 I'm seeing an issue in Lighthouse when running via webpagetest that I'm hoping someone can help investigate.

Here's what I'm seeing https://www.webpagetest.org/result/170608_KH_4a8be412cf0ef89d56e3cb13a8eed676/ (this is running as Mobile - Faster 3G from https://www.webpagetest.org/easy)

Here's a similar test but using Mobile - Emerging Markets where the waterfall looks pretty similar but the Lighthouse results don't fail as much https://www.webpagetest.org/result/170607_PW_1d454612c736a65ddff058cf16b8842e/

The website is https://nextjs-hn.chrisdwheatley.com/, when running Lighthouse via the Chrome extension or in the devtools in Chrome Canary I'm unable to replicate the issue.

If you have any thoughts as to what's causing this issue it'd be much appreciated.

patrickhulce:
should fix #2475 (the issue raised in https://github.com/tastejs/hacker-news-pwas/pull/63), the large class of remaining sites in https://github.com/GoogleChrome/lighthouse/issues/2293, and the last 10% that were left after #2347 

🎉 🎉 🎉 🎉 🎉 

Best explained by the comment:
```
/**
 * Installs a PerformanceObserver in the page to monitor for longtasks and resolves when there have
 * been no long tasks for at least waitForCPUQuiet ms. The promise will not resolve until the
 * `markAsResolvable` function on the return object has been called. This is to prevent promise resolution
 * before some important point in time such as network quiet or document load.
 * @param {number} waitForCPUQuiet
 * @return {{promise: !Promise, cancel: function(), markAsResolvable: function()}}
 */
_waitForCPUIdle(waitForCPUQuiet) {
```

This change also makes the `pauseAfterNetworkQuiet` heuristic no longer necessary, so it's been removed as well to keep our number of config flags the same.


lighthouse-core/gather/driver.js
```diff
@@ -460,27 +520,27 @@ class Driver {
...
       maxWaitForLoadedMs) {
     let maxTimeoutHandle;
 
     // Listener for onload. Resolves pauseAfterLoadMs ms after load.
     const waitForLoadEvent = this._waitForLoadEvent(pauseAfterLoadMs);
-    // Network listener. Resolves pauseAfterNetworkQuietMs after when the network has been idle for
-    // networkQuietThresholdMs.
-    const waitForNetworkIdle = this._waitForNetworkIdle(networkQuietThresholdMs,
-        pauseAfterNetworkQuietMs);
+    // Network listener. Resolves when the network has been idle for networkQuietThresholdMs.
```
addyosmani: Might be worth expanding comment to Network and CPU listeners or adding a similar line above `...const waitForCPUIdle` for consistency.

lighthouse-core/config/default.js
```diff
@@ -13,8 +13,8 @@ module.exports = {
     passName: 'defaultPass',
     recordTrace: true,
     pauseAfterLoadMs: 5250,
-    networkQuietThresholdMs: 5000,
-    pauseAfterNetworkQuietMs: 2500,
+    networkQuietThresholdMs: 5250,
```
addyosmani: Is the idea here to realign `networkQuietThresholdMs` and the new `cpuQuietThresholdMs` with the `pauseAfterLoadMs` delay instead of the lower pause thresholds previously being used?

lighthouse-core/gather/driver.js
```diff
@@ -425,6 +425,66 @@ class Driver {
...
+        .then(timeSinceLongTask => {
+          if (typeof timeSinceLongTask === 'number' && timeSinceLongTask >= waitForCPUQuiet) {
+            log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
+            resolve();
+          } else {
+            log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
+            tryLater();
+          }
+        });
+    }
```
addyosmani: I'm not quite as close to this code as @brendankenny or @paulirish might be, but the general approach to using long-tasks here sgtm.

lighthouse-core/config/default.js
patrickhulce: the historical background to `pauseAfterNetworkQuiet` was that we needed a way to account for CPU work after scripts have finished downloading, so it became the number for "what's the longest after network quiet we might reasonably expect a long task to occur"

now that this role has been subsumed by a bona fide CPU idle check, we don't need it anymore, however, we still want a little bit of slack in our cutoff thus `5250` instead of straight `5000`

`pauseAfterLoad` is probably still worth keeping around for now as there are cases like the ember + fastboot where "quiet" can be reached too quickly without really reaching the load state, and it's good to see the full timeline as we try to improve the other heuristics

lighthouse-core/gather/driver.js
patrickhulce: oops, meant to do that thanks!

patrickhulce:
> what about just adding it once in gatherRunner.setupDriver in the style of driver.cacheNatives

Getting rid of markAsResolvable would be nice, but I'm not sure the trade-off of moving this specific functionality that's only for waitForCPUIdle into setupDriver is that much better?

patrickhulce:
bump :)

wardpeet:
does it make sense to cache the native peformanceObserver as we do with Promise and Error?

patrickhulce:
@wardpeet I think it's much less of a problem here since for Promise and Error we relied on them being the actual original classes for instanceof and protocol promise support. Here we just want the PerformanceObserver functionality without caring about the specific object it references. Much like we do with all the other browser APIs we use (Image, DOM APIs, setTimeout, etc)

wardpeet:
@patrickhulce 👍  sounds reasonable 😄 

Can't wait to have this landed!

patrickhulce:
🏏 🏏

patrickhulce:
🏏 🏏

lighthouse-core/gather/driver.js
```diff
@@ -460,27 +520,28 @@ class Driver {
    * See https://github.com/GoogleChrome/lighthouse/issues/627 for more.
```
paulirish: this method description need an update. Can you flesh it out a bit so we won't have to look back up the PR when trying to read the code?

lighthouse-cli/test/smokehouse/tricky-ttci/expectations.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Expected Lighthouse audit values for --perf tests
+ */
+module.exports = [
+  {
+    initialUrl: 'http://localhost:10200/tricky-ttci.html',
+    url: 'http://localhost:10200/tricky-ttci.html',
+    audits: {
+      'first-interactive': {
```
patrickhulce: love it, gotta check those power levels

Issue #2482, opened 2017-06-11
noopener audit should also allow `noreferrer`


ngyikp:
Failing HTML:
```html
<a href="https://example.com" target="_blank" rel="noreferrer">text</a>
```

According to https://mathiasbynens.github.io/rel-noopener/ , using `noreferrer` by itself fixes the vulnerability too.

If `noreferrer` is enough, then wouldn't putting `noopener` next to it would be redundant?

patrickhulce:
yep that's right good catch @ngyikp!

ripsonowy:
not only noreferrer but also other, for example: nofollow, external etc...

patrickhulce:
@ripsonowy I don't believe nofollow and external solve the performance and hijacking vulnerabilities that are targeted by this audit. Could you elaborate why you believe those also solve the issue?

PR #2485, opened 2017-06-12
Add rel=noreferrer to audit gather

ev1stensberg:
#2482 Could somebody test this though?

### DIFF
--- a/lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
+++ b/lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
@@ -17,7 +17,7 @@ class ExternalAnchorsUseRelNoopenerAudit extends Audit {
     return {
       category: 'Performance',
       name: 'external-anchors-use-rel-noopener',
-      description: 'Opens external anchors using `rel="noopener"`',
+      description: 'Opens external anchors using `rel="noopener"` or `rel="noreferrer"`',
       helpText: 'Open new tabs using `rel="noopener"` to improve performance and ' +
           'prevent security vulnerabilities. ' +
           '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/noopener).',
--- a/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js
@@ -14,7 +14,7 @@ class AnchorsWithNoRelNoopener extends Gatherer {
    */
   afterPass(options) {
     const driver = options.driver;
-    return driver.querySelectorAll('a[target="_blank"]:not([rel~="noopener"])')
+    return driver.querySelectorAll('a[target="_blank"]:not([rel~="noopener"]):not([rel~="noreferrer"])')
       .then(failingNodeList => {
         const failingNodes = failingNodeList.map(node => {
           return Promise.all([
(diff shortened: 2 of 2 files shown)


ebidel:
According to the spec:
> For historical reasons, the `noreferrer` keyword implies the behavior associated with the `noopener` keyword when present on a hyperlink that creates a new browsing context. That is, `<a href="..." rel="noreferrer" target="_blank">` has the same behavior as `<a href="..." rel="noreferrer noopener" target="_blank">`.

So, you're correct that browsers which implement `rel="noopener"` will treat `rel="noreferrer"` as `rel="noreferrer noopener"`. IOW, you'll get the same security and perf benefits. Verifiable here: [example](http://jsbin.com/xajatuveru/edit?html,output). However, I'm sure if if the reverse is true for browsers that don't support it. 

A big motivation of this audit is about the performance benefit it brings, so we should be careful about changing the description language if `rel="noreferrer"` (by itself) doesn't do anything for perf. Using `rel="noreferrer"` also has other implications for tools that take in the`HTTP_REFERER` header.

IMO, `rel="noopener"` is the more modern approach and [support is decent](http://caniuse.com/#search=noopener) across browsers. Alternatively, we could just add a note to https://developers.google.com/web/tools/lighthouse/audits/noopener that mentions adding `rel="noreferrer"` for older browsers.  WDYT? cc @kaycebasques

patrickhulce:
> if rel="noreferrer" (by itself) doesn't do anything for perf

@ebidel how would `noreferrer` not give you the benefits? I thought we just established from the spec it is a strict superset of the disassociation. 33% global exposure to a serious potential phishing issue is pretty high to not suggest adding `noreferrer` at all, the ROI on adding it is so so good. I don't think we should enforce it as a requirement since there are very valid reasons to keep the Referrer header, but it's worth highlighting in the description of helptext IMO

ebidel:
> I thought we just established from the spec it is a strict superset of the disassociation.

It would be nice for someone to verify this in a browser that doesn't support `rel="noopener"`. This page would work: http://jsbin.com/xajatuveru/edit?html,output

> worth highlighting in the description of helptext IMO

Any details not required to pass the audit should live in the [audit references](https://developers.google.com/web/tools/lighthouse/audits/noopener). https://github.com/GoogleChrome/lighthouse/blob/master/CONTRIBUTING.md#helptext-guidelines

patrickhulce:
> Any details not required to pass the audit should live in the audit references. 

Bitten by my own proposal 😆 maybe I do actually want to enforce it then, it seems important for a developer to be aware of this exposure and exclusively mentioning noopener here would give no indication of that (and thus no incentive to click Learn More since it can be deduced from the help text alone how to make it pass in that case), passing the audit would still leave a large vulnerability window.

> It would be nice for someone to verify this in a browser that doesn't support rel="noopener".

Ah gotcha, yes all of the above is conditioned on whether they actually behave as stated :)

paulirish:
if our ultimate recommendation is that `rel="noreferrer noopener"` is the most comprehensive way to set this (with all the browser support), then we should just audit for that.  otherwise it's unclear what we want and what they should do.

ev1stensberg:
I'll let you decide what 2 do with this

jimmywarting:
I actually prefer noreferrer instead of noopener in one of my websites.
It's more anonymas which is more important for my users. (leaving of the referrer header)

I think either or should be fine, let the user decide which he wants to use

paulirish:
#4920 took care of this.

PR #2486, opened 2017-06-12
Link to appcache doc

kaycebasques:


kaycebasques:
I've omitted the stuff about using a service worker instead, since that falls under the category of "how to fix the audit", which we agreed we wouldn't cover anymore in the `helpText` when a reference doc exists. See [helpText guidelines](https://github.com/GoogleChrome/lighthouse/blob/master/CONTRIBUTING.md#helptext-guidelines)

PR #2491, opened 2017-06-13
Fix typo

piperchester:


ebidel:
According to https://en.oxforddictionaries.com/usage/less-or-fewer, "less" is ok since we're talking about a number (12) with a measurement (characters).

patrickhulce:
[NY Times stylebook also agrees](https://afterdeadline.blogs.nytimes.com/2011/03/01/fewer-vs-less/?_r=0) on this point

paulirish:
cc @kaycebasques this sgty?

kaycebasques:
So long as we effectively transfer the knowledge to our readers, I generally don't worry about stuff like this

IMO either word conveys the essential info

piperchester:
Thanks everyone! 👍

PR #2499, opened 2017-06-14
feat(config): add audit blacklist support

patrickhulce:
the path to enabling coverage by default isn't looking good, so we're gonna need a mechanism for skipping audits but including a version with them in the config :)

happy to bikeshed on naming

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -293,10 +293,13 @@ class Config {
     // Generate a limited config if specified
     if (configJSON.settings &&
         (Array.isArray(configJSON.settings.onlyCategories) ||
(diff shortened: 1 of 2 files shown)


ev1stensberg:
Just curious, why is the smoketests throwing at CI now and then?

patrickhulce:
> Just curious, why is the smoketests throwing at CI now and then?

couple reasons, we have some obstacles that actually have flaky implementations (like dismiss dialog), then other times the containerized trusty doesn't get enough resources for a particular run and the page load where we assert reasonable performance no longer has reasonable performance, also some lingering bugs that we can't reliably repro locally for and only happens sometimes to travis builds (no screenshots found in trace)

PR #2518, opened 2017-06-16
fix(unused-css-rules): update to support new coverage format

patrickhulce:
~config changes are commented out until #2499 lands and I can split into a `full-config.js` and a `default-config.js`~

a couple caveats to note
* I had to remove `Number Unused Rules` because the new coverage only reports string offsets of *used* portions, I had a version of this PR that actually did the mapping of lines/columns to string offsets and counted the number of rules but it required the parsed stylesheets and because the parsing so frequently fails this meant that we rarely got them anyway (not to mentioned it greatly increased the complexity and readability of what was going on)
* The size numbers reported here are still not going to match what you see in DevTools since the point of this is to audit is to highlight the network savings not the uncompressed/expanded CSS bytes that the coverage tool reports. Example: on paulirish.com DevTools reports 240kb of unused CSS when the total downloaded CSS is only ~40kb
* The URL reported by `header.sourceURL` doesn't seem to be 100% accurate, I added a debugString and hopefully we'll be able to track some more cases with #2420 
* The perf implications of enabling coverage were somewhat inconclusive but the styles gatherer is what can really take significant amounts of time, so I stopped investigating and opted for #2499 instead since we'll have a place going forward to collect these "extras" and make them accessible with a simple flag or checkbox instead of forcing the slowdown on all users


lighthouse-core/config/default.js
```diff
@@ -37,6 +36,14 @@ module.exports = {
...
     ]
   },
+  // {
+  //   passName: 'extrasPass',
+  //   useThrottling: false,
+  //   gatherers: [
+  //     'styles',
+  //     'css-usage',
+  //   ],
+  // },
```
brendankenny: with #2499 landed, time to uncomment?

patrickhulce: yep rebased it all 👍

patrickhulce:
PTAL :)

patrickhulce:
🏏 🏏

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +63,44 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
+
+    for (const usedRule of stylesheetInfo.usedRules) {
+      usedUncompressedBytes += usedRule.endOffset - usedRule.startOffset;
+    }
+
+    // If we don't know for sure how many bytes this sheet used on the network,
+    // we can guess it was roughly the size of the content gzipped.
+    const totalBytes = stylesheetInfo.networkRecord ?
+        stylesheetInfo.networkRecord.transferSize :
+        Math.round(stylesheetInfo.content.length / 3);
```
brendankenny: since this is the full run, not the default run, instead of estimating like this can we just actually run gzip on it like `response-compression` does?

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +63,44 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
-        stylesheetInfo.unused.push(rule);
+        stylesheetInfo.usedRules.push(rule);
       }
     });
+  }
+
+  /**
+   * @param {!Object} stylesheetInfo
+   * @param {boolean} isInline
+   * @return {{debugString: string|undefined, wastedBytes: number, totalBytes: number}}
```
brendankenny: `wastedPercent`, too

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -163,21 +175,27 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
         {key: 'potentialSavings', itemType: 'text', text: 'Potential Savings'},
       ];
 
+      const debugString = results
+          .map(result => result.debugString)
+          .filter(Boolean)
+          .pop();
+
       return {
+        debugString,
```
brendankenny: add to return type

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +63,44 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
+    let usedUncompressedBytes = 0;
+    const totalUncompressedBytes = stylesheetInfo.content.length;
+
+    for (const usedRule of stylesheetInfo.usedRules) {
+      usedUncompressedBytes += usedRule.endOffset - usedRule.startOffset;
+    }
+
+    // If we don't know for sure how many bytes this sheet used on the network,
+    // we can guess it was roughly the size of the content gzipped.
+    const totalBytes = stylesheetInfo.networkRecord ?
```
brendankenny: will this be weird for inline stylesheets that came in a page that was gzipped?

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: Since we don't actually know it was gzipped anyway I'm inclined to not introduce that complexity here too, but I don't feel that strongly.

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: It will be weird for inline stylesheets that came in a page that *was not* gzipped since we'll underestimate the savings.

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -13,17 +13,16 @@ module.exports = [
...
-      //   extendedInfo: {
-      //     value: {
-      //       results: {
-      //         length: 2
-      //       }
-      //     }
-      //   }
-      // },
+      'unused-css-rules': {
+        score: '<100',
```
patrickhulce: sure, the score gets bucketed by the generic byte efficiency audit though so it's not that useful. I added an assertion on the "wastedKb" instead which is a little more explicit and useful

lighthouse-cli/test/smokehouse/byte-efficiency/run-tests.sh
```diff
@@ -4,7 +4,7 @@ node lighthouse-cli/test/fixtures/static-server.js &
 
 sleep 0.5s
 
-config="lighthouse-core/config/default.js"
+config="lighthouse-core/config/full-config.js"
```
patrickhulce: done

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
brendankenny: > It will be weird for inline stylesheets that came in a page that was not gzipped since we'll underestimate the savings.

ah, yeah, that was based on how it was (I believe) filtering out networkRecords for non-stylesheets before

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +61,41 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
+      const wasCompressed = !networkRecord ||
+          networkRecord._transferSize !== networkRecord._resourceSize;
+      totalBytes = wasCompressed ? Math.round(totalUncompressedBytes / 3) : totalUncompressedBytes;
+    }
+
+    const percentUnused = (totalUncompressedBytes - usedUncompressedBytes) / totalUncompressedBytes;
+    const wastedBytes = Math.round(percentUnused * totalBytes);
+
+    return {
+      wastedBytes,
```
brendankenny: not worth logging `debugString` anymore for network records not found (it'll show up in HTTP Archive data that way)? Or we could sentry it, maybe. Otherwise we won't have any feedback on how often we're failing at finding network records...or it might not matter :)

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +61,41 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
+
+    const networkRecord = stylesheetInfo.networkRecord;
+    let totalBytes = networkRecord && stylesheetInfo.networkRecord.transferSize;
+    if (!networkRecord || networkRecord._resourceType._name !== 'stylesheet') {
+      // If we don't know for sure how many bytes this sheet used on the network,
+      // we can guess it was roughly the size of the content length.
+      const wasCompressed = !networkRecord ||
+          networkRecord._transferSize !== networkRecord._resourceSize;
+      totalBytes = wasCompressed ? Math.round(totalUncompressedBytes / 3) : totalUncompressedBytes;
+    }
```
brendankenny: this is a little confusing to follow. What do you think of something like `totalTransferredBytes` for `totalBytes` and then maybe folding the first short circuit into the conditional like
```js
let totalTransferredBytes;
if (networkRecord && networkRecord._resourceType._name === 'stylesheet') {
  totalTransferredBytes = networkRecord.transferSize;
} else {
  // If we don't know for sure how many bytes this sheet used on the network,
  // we can guess it was roughly the size of the content length.
  const wasCompressed = !networkRecord ||
      networkRecord._transferSize !== networkRecord._resourceSize;
  totalTransferredBytes = wasCompressed ? Math.round(totalUncompressedBytes / 3) :
      totalUncompressedBytes;
}
```
the variable rename alone might be enough to clear things up, though

patrickhulce: yeah sg, done

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
patrickhulce: yeah didn't seem worth it anymore, the cases where it happens now are wholly expected and there's no sense being noisy

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -66,14 +61,47 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
+    let usedUncompressedBytes = 0;
+    const totalUncompressedBytes = stylesheetInfo.content.length;
+
+    for (const usedRule of stylesheetInfo.usedRules) {
+      usedUncompressedBytes += usedRule.endOffset - usedRule.startOffset;
+    }
+
+    const networkRecord = stylesheetInfo.networkRecord;
+    let totalTransferredBytes;
+    if (!networkRecord) {
```
patrickhulce: @paulirish how's this?

Issue #2413, opened 2017-05-31
plots: graph metrics over time (e.g. different lighthouse versions)


wwwillchen:
The Lighthouse team wants to see how performance metrics have evolved over time as lighthouse has changed.

### Phase 1 - Plots over time locally
* This allows you to select multiple out folders and creates a comparison chart

### Phase 2 - Nightly plots
* Does a nightly run on master and records the artifacts (in Google Cloud Storage or GitHub gh-pages?)
* Link to a site that shows plots over time

Part of https://github.com/GoogleChrome/lighthouse/issues/1980

patrickhulce:
dupe of #3461

PR #2520, opened 2017-06-16
plots: dashboard - identify variance over lighthouse versions

wwwillchen:
_Updated code: please review the code :)_

**Example:**
screenshot from 2017-06-29 14 40 29

If you need some dummy data, you can copy [this file](https://gist.github.com/wwwillchen/c9cfd0671244ce905d51f3f91219fa39) into `plots/out/dashboard-results.js` and then open `dashboard/index.html` or I can show you the demo in person.

addresses part of: #2413


patrickhulce:
image

wwwillchen:
@patrickhulce @paulirish - please take a look.

plots/dashboard/dashboard.js
```diff
@@ -0,0 +1,235 @@
...
+}
+
+/**
+ * Navigation Start is usually not a very informative metric.
+ */
+const metrics = Object.keys(dashboardResults).filter(m => m !== 'Navigation Start');
+let currentMetric = metrics[0];
+let numberOfBatchesToShow = 0;
+
+function main() {
```
patrickhulce: this file is kinda hard to follow and grok, wdyt about grouping some of these into classes/objects/sections by usage? i.e. event handlers, rendering, init, etc

plots/dashboard/index.html
```diff
@@ -0,0 +1,41 @@
+<!doctype html>
+<html>
+
+<head>
+  <script src="https://cdn.plot.ly/plotly-1.25.2.min.js"></script>
```
patrickhulce: blocking script in head! 😱  :)

plots/package.json
```diff
@@ -1,7 +0,0 @@
-{
-  "main": "index.js",
-  "scripts": {
-    "measure": "node measure.js",
```
patrickhulce: aw I liked `yarn measure`

plots/dashboard/dashboard.js
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Navigation Start is usually not a very informative metric.
+   */
+  const metrics = Object.keys(dashboardResults).filter(m => m !== 'Navigation Start');
+
+  const renderingScheduler = new RenderingScheduler();
+  const charts = new Charts(renderingScheduler);
+  const dashboard = new Dashboard(metrics, charts);
```
patrickhulce: niiiiice muuuuch easier to follow!

minor nit, personally I like to separate the rendering logic from a constructor to an explicit `render()` method so it's clear what attaches to the DOM and what does basic setup and whatnot (if there's no need to keep instance variables at all some could even be static and the classes just used for logical grouping but totally up to you)

wwwillchen: thanks, done. I'll keep the instance variables (otherwise I think I'd need to use singletons or global variables).

patrickhulce: oh yeah wasn't implying that you *should* make them static, just that if it shook out without the need for instance variables didn't have to keep em that way

plots/package.json
patrickhulce: yeah sgtm

plots/dashboard/dashboard.css
```diff
@@ -0,0 +1,119 @@
+* {
```
brendankenny: license

plots/generate-dashboard.js
```diff
@@ -0,0 +1,129 @@
...
+const childProcess = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+const opn = require('opn');
+const args = require('yargs')
+  .example('node $0 --input-root dir-with-subdirectories')
+  .example('node $0 --inputs out-1 out-2 out-3')
+  .describe({
+    'inputs': 'paths to out directory generated by measure.js',
```
paulirish: could we instead have a single `--input`? (And tbh, at that point, you might as well skip the arg).

It'd look like this:

```sh
node generate-dashboard out-a out-b out-x

node generate-dashboard outParentFolder
```

Take the directory or its children, depending on the contents

plots/generate-dashboard.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * @param {!Array<string>} inputPaths
+ */
+function analyzeInputPaths(inputPaths) {
+  for (const inputPath of inputPaths) {
+    childProcess.execSync(`node analyze.js ${inputPath}`, {
+      env: Object.assign({}, process.env, {
+        CI: '1'
```
paulirish: add a comment that this is to prevent opn() from running. 
and collapse this Obj.assign into one line?

brendankenny:
> null is treated as 0 in when calculating the percentiles.

can they be filtered out before calculating? Adding 0s to the sample will bring the median downward when usually (but not always) missing results come from what would be much higher values.

For the time being we could filter nulls out and then in the future add something like a warning that there were too many missing values in a `plots/` run and either it needs to be rerun or there's a serious regression we let slip in that needs to be fixed :)

wwwillchen:
@brendankenny I tried out your suggestion and ran into an issue when all the runs for a particular batch of results for a site have a null value. Essentially plotly gets confused about where to plot. I think we can go with treating null as 0 for now and then figuring out a better solution in a f/u PR.

### Before (treating null as 0)
before

### After (filtering null)
after

brendankenny:
hmm, if nulls are filtered out of the array before calling `calculatePercentile` on it and the array ends empty, shouldn't `calculatePercentile` already return 0 for the median, 20%, and 80%? I'm ok with having those runs drop to 0 on the graph (it'll be really clear that something was wrong), I just meant nulls should be removed from the array before calculating a percentile.

(unless that's what you did and I'm missing the step that's messing up the graph)

wwwillchen:
@brendankenny updated it so it retuns 0 if the array is empty (before if the array was empty it was returning undefined).

plots/dashboard/dashboard.js
```diff
@@ -0,0 +1,254 @@
...
+    const index = (sorted.length - 1) * percentile;
+    const lower = Math.floor(index);
+    const upper = lower + 1;
+    const weight = index % 1;
+
+    if (upper >= sorted.length) {
+      return sorted[lower] || 0;
+    }
+    return sorted[lower] * (1 - weight) + sorted[upper] * weight;
+  }
```
brendankenny: how do you feel about
```js
calculatePercentile(array, percentile) {
  const sorted = array.filter(x => x !== null).sort((a, b) => a - b);

  if (sorted.length === 0) {
    return 0;
  }
  if (sorted.length === 1 || percentile <= 0) {
    return sorted[0];
  }
  if (percentile >= 1) {
    return sorted[sorted.length - 1];
  }

  const index = (sorted.length - 1) * percentile;
  const lower = Math.floor(index);
  const upper = lower + 1;
  const weight = index % 1;

  return sorted[lower] * (1 - weight) + sorted[upper] * weight;
}
```

plots/dashboard/dashboard.js
```diff
@@ -0,0 +1,254 @@
...
+   * Calculate the value at a given percentile
+   * Based on: https://gist.github.com/IceCreamYou/6ffa1b18c4c8f6aeaad2
+   * @param {!Array<number>} array
+   * @param {number} percentile should be from 0 to 1
+   */
+  calculatePercentile(array, percentile) {
+    if (array.length === 0) {
+      return 0;
+    }
+    if (percentile <= 0) {
```
wwwillchen: yep - fixed.

plots/dashboard/dashboard.js
```diff
@@ -0,0 +1,254 @@
...
+      return array[0];
+    }
+    if (percentile >= 1) {
+      return array[array.length - 1];
+    }
+    const sorted = array.slice().sort((a, b) => a - b).filter(x => x !== null);
```
wwwillchen: done

plots/dashboard/dashboard.js
wwwillchen: looks good.

PR #2527, opened 2017-06-18
Throw better error message when url is not auditable

### LINKED ISSUE #2525: Extension Error: Cannot access contents of the page. Extension manifest must
**Lighthouse Version**: 2.1.0
**Chrome Version**: 61.0.3132.0
**Initial URL**: view-source:chrome-extension://akdgnmcogleenhbclghghlkkdndkjdjc/linkinfo.html?id=e8b97e2a876941dc590bce68bd81e921
**Error Message**: Cannot access contents of the page. Extension manifest must request permission to access the respective host.
**Stack Trace**:
```
Error: Cannot access contents of the page. Extension manifest must request permission to access the respective host.
    at Object.chrome.debugger.attach._ [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:9837:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

wardpeet:
should throw a message that don't support non auditable pages like view-source, ...
Fixes #2525

Not sure it's the best solution though

### DIFF
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -22,8 +22,8 @@ const NON_BUG_ERROR_MESSAGES = {
   // The user tries to review an error page or has network issues
   'Unable to load the page': 'Unable to load the page. Please verify the url you ' +
       'are trying to review.',
-  'Cannot access contents of the page': 'The Lighthouse extension can only audit ' +
-      ' http & https pages.',
+  'Cannot access contents of the page': 'Lighthouse can only audit URLs that start' +
+      ' with http:// or https://.',
 };
 
 const MAX_ISSUE_ERROR_LENGTH = 60;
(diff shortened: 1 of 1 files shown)


lighthouse-extension/app/src/popup.js
```diff
@@ -21,7 +21,9 @@ const NON_BUG_ERROR_MESSAGES = {
       'Chrome-specific urls. If necessary, use the Lighthouse CLI to do so.',
   // The user tries to review an error page or has network issues
   'Unable to load the page': 'Unable to load the page. Please verify the url you ' +
-      'are trying to review.'
+      'are trying to review.',
+  'Cannot access contents of the page': 'The Lighthouse extension can only audit ' +
+      ' http & https pages.',
```
ebidel: "Lighthouse can only audit URLs that start with http:// or https://."

brendankenny:
does this actually fix #2525? The error pattern (`Unable to load the page`) won't match the error in that issue and so the updated explanation won't be shown

paulirish:
I think you're looking at the wrong one?

the exception message is "Cannot access contents of the page. Extension manifest must request permission to access the respective host."

and this PR looks for "Cannot access contents of the page"

brendankenny:
> I think you're looking at the wrong one?

oh, ha, new line :) 👍 👍

PR #2528, opened 2017-06-18
extract logger from lighthouse

### LINKED ISSUE #2526: Release logger as its own module
As mentioned in https://github.com/GoogleChrome/lighthouse/pull/2358#issue-231228347, "I vote that we pull out logger into a standalone module to consume to in both lighthouse and in launcher."

Currently `chrome-launcher` is pulling in the entirety of `lighthouse` just to use the logger.

The dependency is [here](https://github.com/GoogleChrome/lighthouse/blob/c20b28500d692393b8c4c32542c4471921a0fe61/chrome-launcher/chrome-launcher.ts#L16).

samccone:
Fixes #2526


chrome-launcher/chrome-launcher.ts
```diff
@@ -13,7 +13,7 @@ import {DEFAULT_FLAGS} from './flags';
 import {makeTmpDir, defaults, delay} from './utils';
 import * as net from 'net';
 const rimraf = require('rimraf');
-const log = require('lighthouse/lighthouse-core/lib/log');
+const log = require('lighthouse-logger');
```
patrickhulce: oh boy haha, can we just grab the lighthouse org instead? 😃 
* `require('@lighthouse/logger')`
* `require('@lighthouse/chrome-launcher')`
* `require('@lighthouse/all-the-things')`

patrickhulce: https://www.npmjs.com/org/lighthouse looks like it's already taken, but no one is using it...

wardpeet: maybe we should use `@____lighthouse` as an org 😛

lighthouse-viewer/gulpfile.js
```diff
@@ -83,7 +83,6 @@ gulp.task('browserify', () => {
           removeComments: true
         })
         .transform('brfs')
-        .ignore('../lighthouse-core/lib/log.js')
```
ebidel: think viewer still wants to omit logger from the brfs output.

`.ignore('../lighthouse-logger/index.js')`?

lighthouse-extension/yarn.lock
```diff
@@ -152,124 +144,14 @@ astw@^2.0.0:
...
-babel-code-frame@^6.16.0, babel-code-frame@^6.22.0:
+babel-code-frame@^6.16.0:
   version "6.22.0"
   resolved "https://registry.yarnpkg.com/babel-code-frame/-/babel-code-frame-6.22.0.tgz#027620bee567a88c32561574e7fd0801d33118e4"
   dependencies:
     chalk "^1.1.0"
     esutils "^2.0.2"
     js-tokens "^3.0.0"
 
-babel-core@^6.0.2, babel-core@^6.23.0:
```
paulirish: yargs and babel-core removed from this lockfile. I guess it was just stale?

yarn.lock
```diff
@@ -632,6 +632,12 @@ debug@2.2.0, debug@^2.1.1, debug@^2.2.0:
   dependencies:
     ms "0.7.1"
 
+debug@^2.6.8:
```
paulirish: you're bumping to latest debug with this change, yes?

lighthouse-viewer/gulpfile.js
brendankenny: > think viewer still wants to omit logger from the brfs output.

FWIW, this is just a bundle-size optimization that won't matter after viewer switches to LH2 in #2521

lighthouse-extension/yarn.lock
samccone: indeed

yarn.lock
samccone: correct

lighthouse-viewer/gulpfile.js
samccone: updated per @ebidel's comment

samccone:
Thanks y'all

PR #2538, opened 2017-06-19
is-on-https: check record.protocol for blob urls

### LINKED ISSUE #2523: Audit bug: Blob urls being flagged as insecure
Bit of a difficult one to explain this one.. When a read from it shows as insecure in the audit, I don't seem to be able to properly explain this as the blob is secure.... 

https://i.imgur.com/0sP5s1H.png

ebidel:
Fixes #2523.

Adds an additional check for the protocol/scheme. FWIW, `record.scheme` isn't populated for `blob:` urls but `.protocol` is!

### DIFF
--- a/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
+++ b/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
@@ -316,6 +316,13 @@ function deprecationsTest() {
   window.webkitStorageInfo.PERSISTENT // FAIL
 }
 
+function isOnHttps() {
+  // Check blob URL.
+  const img = document.createElement('img');
+  img.src = URL.createObjectURL(blob);
+  document.body.appendChild(img); // PASS
+}
+
 // Figure out which tests to fun.
 const params = new URLSearchParams(location.search);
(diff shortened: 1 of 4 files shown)


lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -316,6 +316,13 @@
   window.webkitStorageInfo.PERSISTENT // FAIL
 }
 
+function isOnHttps() {
+  // Check blob URL.
+  const img = document.createElement('img');
+  img.src = URL.createObjectURL(blob);
```
brendankenny: @ebidel I don't think this is actually working? `Uncaught ReferenceError: blob is not defined`.

The expectation after this change should have `length: 2` (the current extendedInfo has `http://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js` in it, not the blob URL)

ebidel: Oops. Yea the smoketest wasn't testing anything. Bug prevented the image from being aded to the DOM.
Will fix.

PR #2539, opened 2017-06-19
feat(config): add support for audit blacklisting in isolation

paulirish:
I tried to test #2499 and have run into some problems.

@patrickhulce i think this test is sound, but it's failing. Can you review?


patrickhulce:
@paulirish oh uh current assumption was you'd use in conjunction with `only*` so by itself includes`[]` and then skips your audit (which won't be there because nothing was included), I can add support for that though :)

lighthouse-core/config/config.js
```diff
@@ -381,9 +381,13 @@ class Config {
...
    * @return {!Object<string, {audits: !Array<{id: string}>}>}
    */
-  static filterCategoriesAndAudits(oldCategories, categoryIds = [], auditIds = [],
-      skipAuditIds = []) {
+  static filterCategoriesAndAudits(oldCategories, categoryIds, auditIds, skipAuditIds) {
     const categories = {};
+    const filterByIncludedCategory = !!categoryIds;
+    const filterByIncludedAudit = !!auditIds;
+    categoryIds = categoryIds || [];
+    auditIds = auditIds || [];
```
brendankenny: Maybe we should drop the default `[]` for the whitelisted audits and categories, since there's a difference between `undefined` (whitelist everything) and `[]` (whitelist nothing). It'll require a few more conditionals below, but it may make things easier to read/reason about

patrickhulce: that was what `filterByIncludedCategory` and `filterByIncludedAudit` are attempting to capture which I thought were more clear than `if (categoryIds && ...` 

the first section is all just warnings which `filterByIncludedCategory` have no impact on and adding the conditionals there just seemed like noise

patrickhulce:
not sure whose responsibility this is now haha @brendankenny do you still not like the `[]`?

brendankenny:
> not sure whose responsibility this is now haha @brendankenny do you still not like the `[]`?

sorry, no, just a style question. It's good as is if you prefer it.

brendankenny:
- should the blacklist always win? That is, if something is whitelisted *and* blacklisted, should it always end up blacklisted?

What are users likely to expect in that case?

   I couldn't find any good examples out there of programs that support both a whitelist and a blacklist to see what they do. The closest is things like `.gitignore` files or some glob libraries, but they have the benefit of order to decide what wins (whichever is last, blacklisting or whitelisting, wins).

   One answer is to make it an error (instead of just warning in the log) if an audit is in `onlyAudits` and `skipAudits`, but that doesn't inform what to do if an audit is included by `onlyCategories` and in `skipAudits`. That seems like a legitimately useful case (do this whole category except this audit), so I don't think we would want to eliminate being able to do that.

- we may need to revisit this when adding full support for extending configs. It would be nice to have "last one wins" in that case so that you could e.g. whitelist an audit that was blacklisted in a config that you're extending without having to manually add back all the pass/gatherer/audit stuff needed for that audit

brendankenny:
eric brought up another possible option: don't allow both `skipAudits` and `onlyAudits` in the same config, but do allow `onlyCategories` with either of those. Kind of similar to just throwing an error when there's an audit in both lists, it does have the advantage of reducing the cases where Lighthouse has to guess what the config means and we have to document the corner case behavior. Instead it's just a fast fail and the user has to be more explicit.

patrickhulce:
> eric brought up another possible option: don't allow both skipAudits and onlyAudits in the same config, but do allow onlyCategories with either of those. 

good idea, i'll get on that

PR #2540, opened 2017-06-19
fix: add support for skipAudits used in isolation

patrickhulce:


lighthouse-core/config/config.js
```diff
@@ -381,9 +381,13 @@ class Config {
    * @param {!Array<string>=} skipAuditIds
    * @return {!Object<string, {audits: !Array<{id: string}>}>}
    */
-  static filterCategoriesAndAudits(oldCategories, categoryIds = [], auditIds = [],
-      skipAuditIds = []) {
+  static filterCategoriesAndAudits(oldCategories, categoryIds, auditIds, skipAuditIds) {
```
ebidel: doc string need updating now.

patrickhulce: why? they're still optional

patrickhulce:
@ebidel I'm merging into the branch that added the test that's currently failing :)

paulirish:
ok let's merge this and in that PR we can decide if we want the functionality. :)
thx

lighthouse-core/config/config.js
ebidel: True. Not sure what the MO is for closure docs. IIRC, if you don't specify a default arg in the arg list, the var needs to be prefixed with `opt_`. That might be outdated though. cc @brendankenny

brendankenny: we haven't required it and the compiler doesn't really care anymore. It is nice to document in lieu of having the explicit `opt_` in the param name, though

PR #2541, opened 2017-06-20
Add script to capture travis build durations

paulirish:
Here's the last 40 days of travis build durations:
image

Pretty cool data. Halfway down the X axis you'll see the variance kick off. AFAICT, this seems to be travis's fault. It starts right when the "fix appveyor" commits land on master. (around the #7972's of build numbers)  But these commits ONLY touch appveyor.yml, so they couldn't matter less to Travis. 
image

EDIT: i've amended the script to tag where the build is from so we can see just master branch builds too:
image

spreadsheet: https://docs.google.com/spreadsheets/d/197M4SRTdDbbpg4uRDCr-tI5l48L2qsAGkxIcIncKKew/edit#gid=0

cc @wardpeet

### DIFF
--- a/lighthouse-core/scripts/log-travis-durations.rb
+++ b/lighthouse-core/scripts/log-travis-durations.rb
@@ -1,0 +1,59 @@
+##
+# @license Copyright 2017 Google Inc. All Rights Reserved.
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+##
+
+# @fileoverview This script logs durations of travis builds, for plotting
+
+# step 1. generate github token and and use in .netrc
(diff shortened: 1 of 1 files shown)


stramel:
@paulirish What is the unit on Y-axis?

paulirish:
@stramel good question :).  Y axis is **time in seconds.**  
500s is a little bit over 8 minutes.

lighthouse-core/scripts/log-travis-durations.rb
```diff
@@ -0,0 +1,59 @@
...
+Travis.access_token = access_token
+
+$days_of_builds_to_consider = 5
+
+puts "build, branch, state, started_at, duration"
+
+def log_durations
+  repository = Travis::Repository.find('GoogleChrome/lighthouse')
+  repository.each_build do |build|
+    # next unless Integer(build.number) < 8283
```
brendankenny: what was this for?

lighthouse-core/scripts/log-travis-durations.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    next if build.started_at.nil?
```
brendankenny: maybe add a comment mentioning when this would occur

lighthouse-core/scripts/log-travis-durations.rb
paulirish: leaving it in in case people want to look for a specific range.

PR #2549, opened 2017-06-20
refactor(StartUrl): switch from error to debugString object

patrickhulce:
split out from #2420 to reduce the noise there


lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,29 @@ class StartUrl extends Gatherer {
...
       .then(manifest => {
-        if (!manifest.value.start_url || !manifest.value.start_url.raw) {
-          return Promise.reject(new Error(`No web app manifest found on page ${options.url}`));
+        if (!manifest || !manifest.value.start_url || !manifest.value.start_url.raw) {
+          throw new Error(`No web app manifest found on page ${options.url}`);
         }
 
         if (manifest.value.start_url.debugString) {
-          return Promise.reject(new Error(manifest.value.start_url.debugString));
+          throw new Error(manifest.value.start_url.debugString);
```
brendankenny: do we want this to be an error? It'll be things like `invalid start_url relative to ${manifestUrl}` or `start_url must be same-origin as document`, which is nice to put in front of the developer, but the parsed manifest will still have a valid `start_url` in those cases.

Not sure if we want to fall on the "this is broken" side or the side of how the browser actually treats it since we don't have a great logging spot for that type of message if no error here

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,29 @@ class StartUrl extends Gatherer {
...
         }
 
         this.startUrl = manifest.value.start_url.value;
-      }).then(_ => this.executeFetchRequest(options.driver, this.startUrl));
+      })
+      .then(_ => this.executeFetchRequest(options.driver, this.startUrl))
+      .catch(err => {
+        this.debugString = err.message;
```
brendankenny: probably better to set `this.debugString` above and `return` rather than throwing errors and catching everything here. Doing that would also let errors from `getAppManifest`, the manifest parser, and `executeFetchRequest` get through without all being turned into debugStrings

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,29 @@ class StartUrl extends Gatherer {
...
+      });
   }
 
   afterPass(options, tracingData) {
-    if (!this.startUrl) {
-      return Promise.reject(new Error('No start_url found inside the manifest'));
+    if (this.debugString || !this.startUrl) {
+      const debugString = this.debugString || 'No start_url found inside the manifest';
+      return Promise.resolve({debugString});
```
brendankenny: `statusCode` of `-1` in this case?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -76,7 +81,9 @@ class StartUrl extends Gatherer {
     }).pop(); // Take the last record that matches.
 
     return options.driver.goOnline(options)
-      .then(_ => navigationRecord ? navigationRecord.statusCode : -1);
+      .then(_ => {
+        return {statusCode: navigationRecord ? navigationRecord.statusCode : -1};
```
brendankenny: `debugString` in the `-1` case?

lighthouse-core/gather/driver.js
```diff
@@ -276,7 +276,7 @@ class Driver {
             }
 
             // If both the data and the url are empty strings, the page had no manifest.
-            return reject('No web app manifest found.');
+            return resolve(null);
```
patrickhulce: sure

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,29 @@ class StartUrl extends Gatherer {
   pass(options) {
     return options.driver.getAppManifest()
       .then(response => {
-        return manifestParser(response.data, response.url, options.url);
+        return response && manifestParser(response.data, response.url, options.url);
       })
       .then(manifest => {
-        if (!manifest.value.start_url || !manifest.value.start_url.raw) {
-          return Promise.reject(new Error(`No web app manifest found on page ${options.url}`));
+        if (!manifest || !manifest.value.start_url || !manifest.value.start_url.raw) {
```
patrickhulce: done

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: it wasn't it was being caught below

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: done

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: that was kinda the point though since all of this could reasonably fail not because of a bug but just because of the page not having a service worker, fine with switching for now though and going back to this if it's too noisy

lighthouse-core/gather/driver.js
```diff
@@ -264,25 +264,18 @@ class Driver {
...
-
-            // If both the data and the url are empty strings, the page had no manifest.
-            return reject('No web app manifest found.');
-          }
+    return this.sendCommand('Page.getAppManifest')
+      .then(response => {
+        // We're not reading `response.errors` however it may contain critical and noncritical
+        // errors from Blink's manifest parser:
+        //   https://chromedevtools.github.io/debugger-protocol-viewer/tot/Page/#type-AppManifestError
+        if (!response.data) {
```
brendankenny: why drop the `response.url` check? If the manifest 404s, for instance, `response.data` will be `undefined` but that should probably be considered an error state, not that there is no manifest

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,28 @@ class StartUrl extends Gatherer {
...
-        if (!manifest.value.start_url || !manifest.value.start_url.raw) {
-          return Promise.reject(new Error(`No web app manifest found on page ${options.url}`));
+        if (!manifest || !manifest.value) {
+          this.debugString = `No web app manifest found on page ${options.url}`;
+          return;
         }
 
         if (manifest.value.start_url.debugString) {
-          return Promise.reject(new Error(manifest.value.start_url.debugString));
+          this.debugString = manifest.value.start_url.debugString;
```
brendankenny: should this be a state with a `debugString` *and* try to still fetch the `start_url`?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,24 +49,28 @@ class StartUrl extends Gatherer {
...
         this.startUrl = manifest.value.start_url.value;
-      }).then(_ => this.executeFetchRequest(options.driver, this.startUrl));
+        return this.executeFetchRequest(options.driver, this.startUrl);
+      });
   }
 
   afterPass(options, tracingData) {
-    if (!this.startUrl) {
-      return Promise.reject(new Error('No start_url found inside the manifest'));
+    if (this.debugString || !this.startUrl) {
```
brendankenny: seems like either there's a `debugString` or a `startUrl` for all possible states?

lighthouse-core/gather/driver.js
patrickhulce: IMO, anything that's not a bug should not say 'GathererError: blah', if the manifest 404s that's not a bug so it shouldn't be an error, at best we can give them the slightly different debug strings, is that what you're pushing back against?

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: hm, why? there's nothing to fetch if the start_url is bad

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: not if executeFetchRequest fails

lighthouse-core/gather/driver.js
brendankenny: Saying "No web app manifest found on page" is definitely wrong in that case, so a better error string is needed. But I'm not sure I agree about Gatherer errors. I have no problem with a method fetching a manifest throwing on a manifest URL that doesn't exist.

But I assume the issue is us conflating Lighthouse bugs with exceptional states? We may need to get back into that conversation to get the functionality we need. We don't want to slide back into -1 land where every audit has to handle all possible gatherer error states, but agreed we probably need a better separation between the classes of errors.

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > hm, why? there's nothing to fetch if the start_url is bad

if there's a parseable JSON file there's always a good `start_url`, even if it has to fall back to just the page's URL.

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > not if manifestParser/getAppManifest fails, but I guess that's a different can of worms altogether

yeah, in that case (if `pass` rejects) `afterPass` should be ignored, but it will still run, which is weird. We should probably fix that...

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: oh I see now, faulty start_urls are always ignored

debugString seems rather useless then there will always be a usable start_url...

lighthouse-core/gather/driver.js
patrickhulce: > I have no problem with a method fetching a manifest throwing on a manifest URL that doesn't exist.

I've got no problem with this throwing either just the overall pass throwing instead of returning debugString. There should be some distinguishing from a user's perspective between a bug they have no control over and something they need to take action on.

patrickhulce: Also re: 

> Saying "No web app manifest found on page" is definitely wrong in that case, so a better error string is needed.

I'd don't actually agree that "Unable to retrieve manifest" is better than "No web app manifest found on page", as a user receiving the former I may believe that it's LH problem not being able to retrieve it and there's nothing for me to do whereas the latter gets me to check my manifest. There's probably room for improvement on both, but calling it "definitely wrong" seems overly critical. It's not wrong, there was no usable manifest found.

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > debugString seems rather useless then there will always be a usable start_url

it is useful for debugging to see if your manifest isn't acting like you think it should...e.g. you have one start_url in mind and it's defaulting to another. That might be really hard to track down (or even notice that it's happening) otherwise.

I guess that makes me fall on the side of the current code and just show the error :) Lighthouse can be stricter than the spec in that case

patrickhulce: I guess this gets into the division of labor between DevTools and LH. LH shouldn't be your debugger IMO outside of general messages to get you on the right track/where to start looking. If everything is working from the PWA perspective (and document url was the right start url), do you really care about being flagged that you don't have a valid start url?

Either way, I added back the debug string to still do the check so if it passes it should pass with the debugString

lighthouse-core/gather/driver.js
brendankenny: maybe we need to start laying out the failure states for an audit ahead of time so we can decide on them once (in theory) and then it's obvious when they're implemented or not :)

If there's no `<link rel="manifest">` in the page, it's obviously not an error, the audits simply just won't pass. However, to me, if there is a `<link rel="manifest">` and the url 404s, that is an error worth reporting to the user, as much as a manifest made up of invalid JSON would be. They tried to include one and it's failing for some reason.

Agreed "Unable to retrieve manifest" is a bad error string for that and we should be as specific as possible to help with those situations. Also totally on board with exploring non-exception error returns, expected errors, etc after this so we can separate the two (or more?) clear classes of errors we're dealing with here.

lighthouse-core/gather/gatherers/start-url.js
brendankenny: > Either way, I added back the debug string to still do the check so if it passes it should pass with the debugString

I think that's a good solution for now and the devtools/LH discussion is a good one to revisit now that Audits 2.0 is a real thing. Pavel and Paul also love drawing diagrams about that sort of division so I'm sure they'd be happy to get in on it :)

lighthouse-core/gather/driver.js
brendankenny: > "Unable to retrieve manifest" is a bad error string for that and we should be as specific as possible to help with those situations

One downside of the current `Page.getAppManifest` protocol call is that it gives no network status information (as far as I can tell). So e.g. if the server returns a 404 with a `Cannot GET /manifest.json` response string, the protocol will just return a `JSON.parse` error from trying to parse that string and we just have to infer that there was a 404. Maybe we could see about getting more detail there.

brendankenny: (the devtools Application panel suffers from the same problem, only showing "Errors and warnings: `Line: 1, column: 1, Unexpected token.`" and you have to notice the 404 in the console/network panel to see what happened)

patrickhulce: > One downside of the current Page.getAppManifest protocol call is that it gives no network status information (as far as I can tell).

Yeah feels like it should return something similar to an artifact `{value: null, debugString: ''}` or whatever

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: > Pavel and Paul also love drawing diagrams about that sort of division so I'm sure they'd be happy to get in on it :)

haha :)

paulirish:
@brendankenny will review and we'll get this sorted today!

lighthouse-core/gather/gatherers/manifest.js
```diff
@@ -27,19 +27,16 @@ class Manifest extends Gatherer {
   afterPass(options) {
     return options.driver.getAppManifest()
       .then(response => {
+        if (!response || !response.data) {
```
brendankenny: `!response.data` not needed here?

lighthouse-core/audits/webapp-install-banner.js
```diff
@@ -73,25 +73,30 @@ class WebappInstallBanner extends MultiCheckAudit {
...
     return artifacts.requestManifestValues(artifacts.Manifest).then(manifestValues => {
-      WebappInstallBanner.assessManifest(manifestValues, failures);
-      WebappInstallBanner.assessServiceWorker(artifacts, failures);
-      WebappInstallBanner.assessOfflineStartUrl(artifacts, failures);
-
-      return {
-        failures,
-        manifestValues
-      };
+      const result = {failures, manifestValues};
```
brendankenny: if `result` already has `failures` and `manifestValues` on it but all three are being passed into these methods, should they instead just have a single `result` parameter?

lighthouse-core/audits/webapp-install-banner.js
```diff
@@ -73,25 +73,30 @@ class WebappInstallBanner extends MultiCheckAudit {
...
+      WebappInstallBanner.assessManifest(manifestValues, failures, result);
+      WebappInstallBanner.assessServiceWorker(artifacts, failures, result);
+      WebappInstallBanner.assessOfflineStartUrl(artifacts, failures, result);
```
brendankenny: `MultiCheckAudit` is already somewhat confusing and this doesn't exactly simplify things :)

Short of a `MultiCheckAudit` refactor, what about something like a `warnings` array, which is used for `debugString`, just like `failures` but they won't fail the audit?

Just an idea, maybe we can come up with something simpler/more self-explanatory

lighthouse-core/lib/url-shim.js
```diff
@@ -103,7 +103,7 @@ URL.elideDataURI = function elideDataURI(url) {
 // Why? Special handling was added by Chrome team to allow a pushState transition between chrome:// pages.
 // As a result, the network URL (chrome://chrome/settings/) doesn't match the final document URL (chrome://settings/).
 function rewriteChromeInternalUrl(url) {
-  if (!url.startsWith('chrome://')) return url;
+  if (!url || !url.startsWith('chrome://')) return url;
```
brendankenny: what new path needs this?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,34 +49,40 @@ class StartUrl extends Gatherer {
...
     return options.driver.getAppManifest()
       .then(response => {
-        return manifestParser(response.data, response.url, options.url);
+        return response && manifestParser(response.data, response.url, options.url);
       })
       .then(manifest => {
-        if (!manifest.value.start_url || !manifest.value.start_url.raw) {
-          return Promise.reject(new Error(`No web app manifest found on page ${options.url}`));
+        if (!manifest || !manifest.value) {
+          this.debugString = `No usable web app manifest found on page ${options.url}`;
```
brendankenny: if there's a `manifest` but no `manifest.value`, there will be a `manifest.debugString` probably worth keeping

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,34 +49,40 @@ class StartUrl extends Gatherer {
...
+          return;
         }
 
         if (manifest.value.start_url.debugString) {
-          return Promise.reject(new Error(manifest.value.start_url.debugString));
+          this.debugString = manifest.value.start_url.debugString;
```
brendankenny: maybe a comment here as a reminder that even with a `debugString` there will be a fallback start_url

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -49,34 +49,40 @@ class StartUrl extends Gatherer {
...
     const networkRecords = tracingData.networkRecords;
     const navigationRecord = networkRecords.filter(record => {
       return URL.equalWithExcludedFragments(record._url, this.startUrl) &&
         record._fetchedViaServiceWorker;
     }).pop(); // Take the last record that matches.
 
     return options.driver.goOnline(options)
-      .then(_ => navigationRecord ? navigationRecord.statusCode : -1);
+      .then(_ => {
+        let debugString = this.debugString;
```
brendankenny: can we handle these cases separately for clarity? I think there's:
- `debugString` and no `startUrl`
- `debugString` and no matching `navigationRecord`
- matching `navigationRecord` (and maybe a `debugString` from `start_url` parsing and other future messages)

we could probably give more information (e.g. `start_url` was fetched but had some error, wasn't covered by the SW, etc), but that can wait for another day

lighthouse-core/gather/gatherers/manifest.js
patrickhulce: done

lighthouse-core/audits/webapp-install-banner.js
patrickhulce: done

lighthouse-core/lib/url-shim.js
patrickhulce: this was a drive-by fix of something I encountered on another branch, can't remember what site now though it was over a month ago :P

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: done

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: done

lighthouse-core/lib/url-shim.js
patrickhulce: but to answer your question the `equalWithExcludedFragments` can now be run on a non-existent startUrl which causes this to fail for every site that doesn't have the manifest

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: done

patrickhulce:
PTAL

lighthouse-core/lib/url-shim.js
brendankenny: > but to answer your question the `equalWithExcludedFragments` can now be run on a non-existent startUrl which causes this to fail for every site that doesn't have the manifest

ha, I had forgotten we had done that. I guess that's fine vs moving the `.map(rewriteChromeInternalUrl);` into the try/catch

lighthouse-core/audits/webapp-install-banner.js
```diff
@@ -66,32 +67,37 @@ class WebappInstallBanner extends MultiCheckAudit {
...
+
     if (!hasOfflineStartUrl) {
-      failures.push('Manifest start_url is not cached by a Service Worker');
+      result.failures.push('Manifest start_url is not cached by a Service Worker');
+
+      if (artifacts.StartUrl.debugString) {
+        result.failures.push(artifacts.StartUrl.debugString);
+      }
+    } else if (artifacts.StartUrl.debugString) {
+      result.debugString = artifacts.StartUrl.debugString;
```
brendankenny: weird, I left this comment before (and it was in the email), but it was eaten in the web interface:

> `MultiCheckAudit` is already somewhat confusing and this doesn't exactly simplify things :)

> Short of a `MultiCheckAudit` refactor, what about something like a `warnings` array, which is used for `debugString`, just like failures but they won't fail the audit?

> Just an idea, maybe we can come up with something simpler/more self-explanatory

so here (for this idea) it would `result.warnings.push(artifacts.StartUrl.debugString);` whether or not it `hasOfflineStartUrl`, and then `MultiCheckAudit` would handle appending `result.warnings.join(', ')` to the `debugString`

lighthouse-core/test/audits/webapp-install-banner-test.js
```diff
@@ -157,7 +157,7 @@ describe('PWA: webapp install banner audit', () => {
 
     return WebappInstallBannerAudit.audit(artifacts).then(result => {
       assert.strictEqual(result.rawValue, true);
-      assert.equal(result.debugString, 'Warning!');
+      assert.equal(result.debugString, 'Warnings: Warning!');
```
brendankenny: :)

lighthouse-core/audits/webapp-install-banner.js
patrickhulce: done

PR #2552, opened 2017-06-20
update yarn.lock

ebidel:
Updating from latest master.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -45,13 +45,6 @@ ajv@^4.7.0:
     co "^4.6.0"
     json-stable-stringify "^1.0.1"
 
-ajv@^4.9.1:
-  version "4.11.5"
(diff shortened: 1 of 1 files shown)


paulirish:
these appear to be dependencies of `codecov@2.2.0`
http://npm.anvaka.com/#/view/2d/codecov

brendankenny:
> `+ Hits         4745     4837      +92`
> `- Misses        855      923      +68`
> `- Partials       14       20       +6`

that seems implausible, codecov

ebidel:
So what's up? We don't want an updated yarn.lock?

paulirish:
@ebidel im basically saying that i think this is incorrect.
your deletions are implicit dependencies of `codecov` which was recently added.

try updating to the latest version of yarn? 

is anyone else seeing this diff in their lock?

paulirish:
eh. i applied it and everything seems to be fine. 
i have no idea why i wasn't seeing this diff.


sorry dude. let's do it.

PR #2557, opened 2017-06-20
feat(config): add extends lighthouse:full

patrickhulce:
last step to support #2518 and the forthcoming js coverage PR

split into several commits for easier viewing

1. squashed version of #2539 upon which this relies
2. moves the default config to a new file called 'full-config.js'
3. adds the necessary support to continue the ability to extend the default config in a non-breaking way (people who directly require the default.js file and filter themselves will be broken, but anyone using the LH filtering features will be fine)


brendankenny:
otherwise should be fine after updating for the config.js changes that landed

patrickhulce:
> I'm not sure if there's much to argue for that, though.

I do! That way would make the diff cleaner :) I'll give it a whirl

patrickhulce:
this feels a lot better, PTAL :)

paulirish:
Yeah that was a great call. I'm happy with this.

brendankenny:
if we start including a lot more audits in the repo that aren't included in the default config we may want to split `Runner.getAuditList` and `Runner.getGathererList` so things that won't be running in the extension/devtools don't get pulled into `lighthouse-background.js`

patrickhulce:
> if we start including a lot more audits in the repo that aren't included in the default config we may want to split Runner.getAuditList and Runner.getGathererList so things that won't be running in the extension/devtools don't get pulled into lighthouse-background.js

My intention is to add a checkbox in devtools to run these extra/slow audits so we'd still need them to be browserified, I actually was second guessing if not having them in the default config would get them in there at first haha :)

brendankenny:
oh, I was going to guess that at first, but isn't e.g. coverage kind of redundant in devtools when there's a better interface in other panels? I guess we could link between the report and those panels...

PR #2561, opened 2017-06-21
fix(domstats): prevent infinite loop

### LINKED ISSUE #2532: Lighthouse does not finish
* **Lighthouse Version**: 2.1.0
* **Chrome version**: 59.0.3071.104
* **url**: https://www.davidemaggio.it/
* **Error message**: none
* **Stack trace**: none

lighthouse is not triggering any error, but it does not finish.
using CLI + verbose it gets stuck at `Runtime.evaluate +1ms {"expression":"(function wrapInNativePromise`

patrickhulce:
fixes #2532

the url was https://www.davidemaggio.it/ the parentNode.host of the `html` tag is the `body` tag and vice versa

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/domstats.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/domstats.js
@@ -48,9 +48,12 @@ function createSelectorsLabel(element) {
  */
 /* istanbul ignore next */
 function elementPathInDOM(element) {
+  const visited = new Set();
   const path = [createSelectorsLabel(element)];
   let node = element;
   while (node) {
+    visited.add(node);
+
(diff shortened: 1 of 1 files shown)


brendankenny:
> the parentNode of the `html` tag is the `body` tag and vice versa

I wouldn't think this would be possible, but looks like it's some ShadowRoot issue. `html.parentNode` is `document`, so it should stop there, but `html.parentNode.host` is the `body` within `html`, so `elementPathInDOM`'s traversal goes back into itself.

brendankenny:
also, looks like you need to be in mobile emulation to trigger this

ebidel:
Wow. Very Very strange. Maybe something to do with AMP (doesn't affect polymer sites).

cc @pbakaus

Issue #2510, opened 2017-06-15
Move download-chrome.sh to chrome-launcher and make easily callable


patrickhulce:
At one point we published the `download-chrome.sh` script with lighthouse that was handy for downstreamers to use when running in CI. It might be a nice feature of chrome-launcher to provide a similar mechanism to download/install chrome cross platform as well.

busticated:
fwiw, it would be awesome to have something like https://github.com/Medium/phantomjs

using this, i simply add the `phantomjs-prebuilt` dependency to my package.json, `npm i` (w/ caching) in my CI env, and run tests. works great. would love to drop in a chrome-powered replacement.

paulirish:
@busticated yeah true. we have no short term plans to make a production-ready `chrome-stable-prebuilt` npm package but we definitely see the value in it. 

If someone wants to start working on this, we'd be happy to see that!

PR #2562, opened 2017-06-21
Docs: Using chrome-launcher on CI

denar90:


chrome-launcher/README.md
```diff
@@ -97,3 +97,22 @@ chromeLauncher.launch({
   console.log(`Chrome debugging port running on ${chrome.port}`);
 });
 ```
+
+### Continuous Integration
+
+Unlike development environment, on CI environment Chrome binary is not present.
```
ebidel: Rework of the paragraph: 

In a CI environment like Travis, Chrome may not be installed.  If you want to use `chrome-launcher`, you can install Chrome using Lighthouse's `download-chrome.sh` script:

    curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/master/lighthouse-core/scripts/download-chrome.sh | bash

Then in `.travis.yml`, use it like so:

patrickhulce: yikes really need #2510 right about now huh ;) maybe pin to a specific version instead of master since we change things?

denar90: @ebidel @patrickhulce good points 👍

PR #2569, opened 2017-06-22
[chrome-launcher]: Add missing mkdir and rimraf dependencies

### LINKED ISSUE #2568: chrome-launcher missing dependencies
It looks like installing `chrome-launcher` by itself doesn't work because it uses `mkdirp` and `rimraf`, but does not have them listed as dependencies in `chrome-launcher/package.json`.

I have a branch ready to go with a fix and will have a PR in shortly.

devrelm:
Fixes #2568 

There's a bunch of dependencies removed from `yarn.lock` that I'm not so sure of. They may be indicative of more missing dependencies, but all of the package scripts run fine, so they may have just been extraneous packages that had yet to have been removed from the yarn.lock file.

### DIFF
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -21,7 +21,9 @@
     "@types/core-js": "^0.9.41",
     "@types/mkdirp": "^0.3.29",
     "@types/node": "6.0.66",
-    "lighthouse-logger": "^1.0.0"
+    "lighthouse-logger": "^1.0.0",
+    "mkdirp": "0.5.1",
+    "rimraf": "2.2.8"
   },
   "version": "0.2.0",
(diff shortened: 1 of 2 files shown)


pedro93:
Shouldn't Travis CI run only chrome-launcher related tests? This PR seemed to fail while running lighthouse specific tasks: 
https://travis-ci.org/GoogleChrome/lighthouse/jobs/245789420#L704 (a task about lighthouse-cli that run successfullly)
https://travis-ci.org/GoogleChrome/lighthouse/jobs/245789420#L845 (why should yarn compile-devtools run in the context of chrome-launcher?)

paulirish:
shipped in chrome-launcher@0.2.1

PR #2571, opened 2017-06-22
Improve tests around chrome launcher and flags.

samccone:
Implementation / Test details

This PR introduces a dependency on sinon for testing, which enables us
to use spys and stubs without having to fake our own implementation
(yay)
The PR exposes the ability to mock out the spawn call for testability
(yay)
The PR DRYs up some duplicated test code through the use of beforeEach
and afterEach (yay)


chrome-launcher/test/chrome-launcher-test.ts
```diff
@@ -6,59 +6,76 @@
...
   openSync: () => {},
-  closeSync: () => {}
+  closeSync: () => {},
+  writeFileSync: () => {}
 };
 
 describe('Launcher', () => {
-  it('accepts and uses a custom path', async () => {
+  beforeEach(() => {
     log.setLevel('error');
```
paulirish: does this setLevel business change anything? i didn't see anything logging when these calls werent made.

chrome-launcher/test/chrome-launcher-test.ts
```diff
@@ -6,59 +6,76 @@
...
+
+    try {
+      await chromeInstance.launch();
+    } catch (err) {
+      return Promise.reject(err);
+    }
 
+    const chromeFlags = spawnStub.getCall(0).args[1] as string[];
+
+    assert.ok(chromeFlags.find(f => f.startsWith('--remote-debugging-port')))
```
paulirish: can we also assert one of the flags.ts flags is present?

chrome-launcher/test/chrome-launcher-test.ts
samccone: oh i commented on the old closed PR.

basically since log is stateful, if anyone were to set log (which places in the code do sometimes) this essentially resets the log level for each test run.

paulirish: Turns out logger wasn't logging anything because setLevel was never being called.

samccone:
ah thanks!

PR #2575, opened 2017-06-23
fixes viewer CSS specificity and event listener removal

brendankenny:
sorry, two last viewer flubs
1) This was an existing issue with the old viewer too: binding the `message` handler to `this` meant that a newly created function was being passed to `addEventListener`, so later attempts to remove the handler via a reference to the original, non-bound function was having no effect. Noticed when refreshing the viewer kept loading the same report as long as the triggering html report was still open in a tab.

2) save-as-gist is hidden by the CSS in the `<template>`s, which is after viewer.css in the page, so they win. This !importants it. Why does anyone let me write CSS

### DIFF
--- a/lighthouse-core/report/v2/renderer/report-ui-features.js
+++ b/lighthouse-core/report/v2/renderer/report-ui-features.js
@@ -235,6 +235,7 @@ class ReportUIFeatures {
     // after it's created. Normally, we could also listen for the popup window's
     // load event, however it is cross-domain and won't fire. Instead, listen
     // for a message from the target app saying "I'm open".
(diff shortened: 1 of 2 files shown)


paulirish:
> save-as-gist is hidden by the CSS in the <template>s, which is after viewer.css in the page, so they win.

should the templates be first in order? seems kinda more correct.

brendankenny:
Yeah, that makes sense, though would the CSS have to be after not only the templates but where the templates end up inserted into the DOM, too?

paulirish:
ah yes. this is correct. :)   well then.. !important ftw.

Issue #2556, opened 2017-06-20
Launcher ECONNREFUSED flake


paulirish:
This is a bit of carryover from #2544. It was failing 100% of the time and now its more like 5%.

Example build: https://travis-ci.org/GoogleChrome/lighthouse/jobs/245116083
![image](https://user-images.githubusercontent.com/39191/27357770-7058a968-55c8-11e7-8ffe-701d879fcd06.png)

Things to note:

* Last night, this error thrown for EVERY chrome launch, like all smokehouse runs. After ward's fix it appears to only affect the Launcher tests
* It's intermittent. On this build, only 1 of the 3 travis jobs had a problem.

There might be still a bug on the travis/chrome side, but seems like we have room to handle this from within launcher. (port sniffing & retry logic?)
At the very least, more usable error messages and extra logging.

wdyt @samccone 

_edit: he said.._

![image](https://user-images.githubusercontent.com/39191/27358753-e67e5684-55cc-11e7-9bd6-b2951ca2cee7.png)

wardpeet:
not sure it's relevant but in smokehouse we have
```
// Lighthouse sometimes times out waiting to for a connection to Chrome in CI.
  // Watch for this error and retry relaunching Chrome and running Lighthouse up
  // to RETRIES times. See https://github.com/GoogleChrome/lighthouse/issues/833
```

just a sidenote on smoke tests we add a delay of 0.5s which maybe gives travis some extra room to breath between runs?

paulirish:
pwmetrics has had a workaround for this issue for a long time now: https://github.com/paulirish/pwmetrics/blob/2eb1266f0997050dd56435a2dac5b34dd26e1b82/lib/index.ts#L129-L154 (basically just adding retry logic. ) 


I've put up a proposed fix in #2603

brendankenny:
hey now, smokehouse has had a workaround for even longer :P

Issue #2544, opened 2017-06-20
Broken build: connect ECONNREFUSED


paulirish:
I cleared the travis caches, which so far has revealed TWO bugs.

The first was fixed by changing our install command: #2543 
The second is this issue.

Any tests that rely on connecting to the browser appear to fail right now. Example build: https://travis-ci.org/GoogleChrome/lighthouse/jobs/244797088

Screenshotting the failures:
![image](https://user-images.githubusercontent.com/39191/27317350-9c644dfc-553a-11e7-8cc8-b688e97bbb16.png)
![image](https://user-images.githubusercontent.com/39191/27317355-a53cea24-553a-11e7-9c10-fa2e55a5500c.png)
![image](https://user-images.githubusercontent.com/39191/27317359-abf5381c-553a-11e7-8ee3-1d6a55c3ad44.png)

So far I'm in the blind on the fix. Ideas?

paulirish:
@wardpeet [has been debugging](https://user-images.githubusercontent.com/39191/27348431-e402285a-55a8-11e7-9332-0b5537a4159f.png) this basically since I reported the bug.  **what a hero.**

He's been using the  travis-ci docker image  locally to debug.

I'm posting some of his debugging notes.

-------------

The first error encountered was this:
```sh
travis@4b833d423e4b:~/builds/GoogleChrome/lighthouse$ $(pwd)/chrome-linux/chrome

/home/travis/builds/GoogleChrome/lighthouse/chrome-linux/chrome: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 
`GLIBCXX_3.4.18' not found (required by /home/travis/builds/GoogleChrome/lighthouse/chrome-linux/chrome)`

/home/travis/builds/GoogleChrome/lighthouse/chrome-linux/chrome: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 
`GLIBCXX_3.4.19' not found (required by /home/travis/builds/GoogleChrome/lighthouse/chrome-linux/chrome)`
```

Ward fixed that via `sudo apt-get install libstdc++-4.9-dev`.

Second error was this:
```sh
/home/travis/builds/GoogleChrome/lighthouse/chrome-linux/chrome: 
undefined symbol: gdk_screen_get_monitor_scale_factor
```

This is only reproducible on latest chromium. Beta and stable chromium do not encounter this error.
Internet doesn't have much to say about this error.

[...]

Issue #2462, opened 2017-06-07
ECONNREFUSED When launching multiple Chrome instances


LucianoGanga:
Hello! 
I don't know why, but when I try to execute multiple instances of Chrome in a Linux env, I get this error: 

```
[ERROR] connect ECONNREFUSED 127.0.0.1:39537
at Object.exports._errnoException (util.js:1050:11)
at exports._exceptionWithHostPort (util.js:1073:20)
at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1093:14)
```
I've been trying to find more information about this, but I don't know where to start from. 

Any idea about what may be the problem?

patrickhulce:
How are you launching and consuming chrome? Lighthouse as a node module? `chrome-launcher` and your own connection to Chrome?

If you're using the latest chrome-launcher module, it should be [waiting until the port is available](https://github.com/GoogleChrome/lighthouse/blob/master/chrome-launcher/chrome-launcher.ts#L239) but you'll have to make sure you're awaiting for the promise resolution before connecting.

LucianoGanga:
Thanks for your answer Patrick! 

I'm using the NPM version of lighthouse 2.0.0. 

Then, I'm launching chrome here: 
https://github.com/LucianoGanga/simple-headless-chrome/blob/beta/lib/browser.js#L149

Then I connect to that instance here: 
https://github.com/LucianoGanga/simple-headless-chrome/blob/beta/lib/browser.js#L183

(I use "Chrome Remote Interface" for that: https://github.com/cyrus-and/chrome-remote-interface)

ezesculli:
So far, It appears to be an specific Linux problem (because it works fine in our macs, but not on our servers):

We're basically reproducing this on Linux when:

```
const chromeInstance = require('lighthouse/chrome-launcher')
chromeOptions = {
    port: 0,
    autoSelectChrome: true, // False to manually select which Chrome install.
    additionalFlags: ['--disable-gpu', '--headless', '--enable-logging']
  }

chromeInstance.launch(chromeOptions)
// Instance 1 -> Chrome Launched succesfully

chromeInstance.launch(chromeOptions)
// Instance 2 -> Chrome Launched succesfully

chromeInstance.launch(chromeOptions)
// Instance 3 -> Error: connect ECONNREFUSED 127.0.0.1:60520

chromeInstance.launch(chromeOptions)
// Instance 4 -> Error: connect ECONNREFUSED 127.0.0.1:40814
[...]

PR #2603, opened 2017-06-27
Log Chrome's stderr in case of failure

paulirish:
To improve debugging of #2556, #2544, #2462 , this PR will log out the contents of Chrome's stderr.

It was only through this technique that I found that Chrome had a fatal crash when I was trying to run it in a docker container:

> `[0627/002609.343419:FATAL:zygote_host_impl_linux.cc(107)] No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.`

We'll see if this will uncover the travis flake (which I presume to be unrelated to `--no-sandbox`)

---------

Also in this PR: we add the `logLevel` option to chrome-launcher.

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -30,6 +30,7 @@ export interface Options {
   handleSIGINT?: boolean;
   chromePath?: string;
   userDataDir?: string;
+  logLevel?: string;
 }
 
 export interface LaunchedChrome {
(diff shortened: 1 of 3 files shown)


chrome-launcher/chrome-launcher.ts
```diff
@@ -237,11 +241,19 @@ export class Launcher {
             })
             .catch(err => {
               if (retries > 10) {
+                log.error('ChromeLauncher', err.message);
```
brendankenny: will this end up logging this twice?

brendankenny:
need to `yarn format`, apparently

paulirish:
Ya. But since it's the most important part of that wall of text it seems ok.

chrome-launcher/chrome-launcher.ts
paulirish: Yes. But since it's the most important part of that wall-of-text it seems ok. :)

PR #2611, opened 2017-06-28
fix(driver): move performance observer registration to setupDriver

patrickhulce:
yo dawg @brendankenny I heard you like things in `setupDriver`

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -931,7 +931,7 @@ class Driver {
    * @return {!Promise}
    */
   registerPerformanceObserver() {
-    return this.evaluteScriptOnNewDocument(`(${registerPerformanceObserver.toString()})()`);
+    return this.evaluteScriptOnNewDocument(`(${registerPerformanceObserverInPage.toString()})()`);
   }
 
   /**
@@ -1088,7 +1088,7 @@ function wrapRuntimeEvalErrorInBrowser(err) {
  * property on window to the end time of the last long task.
  * instanbul ignore next
  */
-function registerPerformanceObserver() {
+function registerPerformanceObserverInPage() {
   window.____lastLongTask = window.performance.now();
   const observer = new window.PerformanceObserver(entryList => {
     const entries = entryList.getEntries();
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -206,7 +206,7 @@ class Driver {
    * @param {string} scriptSource
    * @return {!Promise<string>} Identifier of the added script.
    */
-  evaluateScriptOnLoad(scriptSource) {
+  evaluteScriptOnNewDocument(scriptSource) {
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -206,7 +206,7 @@ class Driver {
    * @param {string} scriptSource
    * @return {!Promise<string>} Identifier of the added script.
    */
-  evaluateScriptOnLoad(scriptSource) {
+  evaluteScriptOnNewDocument(scriptSource) {
```
brendankenny: 👍

lighthouse-core/gather/driver.js
```diff
@@ -1081,7 +1088,7 @@ function wrapRuntimeEvalErrorInBrowser(err) {
  * property on window to the end time of the last long task.
  * instanbul ignore next
  */
-function installPerformanceObserver() {
+function registerPerformanceObserver() {
```
brendankenny: I think I prefer the driver method and this function to have different names...but I don't feel strongly if you like it better this way.

lighthouse-core/gather/driver.js
```diff
@@ -464,8 +464,6 @@ class Driver {
         });
     }
 
-    log.verbose('Driver', `Installing longtask listener for CPU idle.`);
-    this.evaluateScriptOnLoad(`(${installPerformanceObserver.toString()})()`);
```
brendankenny: doesn't this mean we can also get rid of `markAsResolvable` since registration and polling are separated?

patrickhulce: nope, that's separate. `markAsResolvable` is for ensuring that we wait for no long tasks *after* network idle and not fire early while we're still downloading massive scripts or something

lighthouse-core/gather/driver.js
patrickhulce: how about `registerPerformanceObserverInPage` or something to signify the different context?

lighthouse-core/gather/driver.js
brendankenny: Right, but now `waitForCpuIdle` can just be chained after `waitForNetworkIdle` rather than initialized at the same time and then marked as resolvable in the chain.

lighthouse-core/gather/driver.js
brendankenny: SGTM

lighthouse-core/gather/driver.js
patrickhulce: Oh I gotcha, yeah sure

brendankenny: We could also land this since it works fine so we can get it in the release, and then do a refactor of waitForLoad later.

patrickhulce: > We could also land this since it works fine so we can get it in the release, and then do a refactor of waitForLoad later.

sg :) I was blinded by my desire to not have a messy rebase of a local branch that was abusing markAsResolvable for the waitForCustomPerformanceMark 😆

PR #2613, opened 2017-06-28
refactor(critical-request-chains): split out critical request chains

patrickhulce:
preparing to do some predictive perf analysis and need access to the full request chains, this PR separates out the criticality filter from the wiring of 'initiator'

### DIFF
--- a/lighthouse-core/gather/computed/request-chains.js
+++ b/lighthouse-core/gather/computed/request-chains.js
@@ -29,6 +29,8 @@ class RequestChains extends ComputedArtifact {
 
   /**
    * Links together network requests (filtered by predicate) with their initiator, building a DAG.
+   * When the predicate function returns false for a network request, that request and all of its children
+   * will be excluded from the graph. If no predicate is provided, all requests are included.
    *
    * @param {!Array<!WebInspector.NetworkRequest>} networkRecords
    * @param {function(!WebInspector.NetworkRequest):boolean=} predicate
--- a/lighthouse-core/gather/computed/request-chains.js
+++ b/lighthouse-core/gather/computed/request-chains.js
@@ -14,8 +14,8 @@ class RequestChains extends ComputedArtifact {
   }
 
   /**
-   *
-   * @param {!} record
+   * @param {!WebInspector.NetworkRequest} record
+   * @return {{url: string, startTime: number, endTime: number, responseReceivedTime: number, transferSize: number}}
    */
   static flattenRecord(record) {
     return {
@@ -28,10 +28,11 @@ class RequestChains extends ComputedArtifact {
   }
 
   /**
+   * Links together network requests (filtered by predicate) with their initiator, building a DAG.
    *
    * @param {!Array<!WebInspector.NetworkRequest>} networkRecords
    * @param {function(!WebInspector.NetworkRequest):boolean=} predicate
-   * @return {!ComputedArtifacts}
+   * @return {{request: !Object, children: !Object}}
    */
   static extractChain(networkRecords, predicate) {
     predicate = predicate || (() => true);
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/computed/request-chains.js
```diff
@@ -0,0 +1,122 @@
...
+      url: record._url,
+      startTime: record.startTime,
+      endTime: record.endTime,
+      responseReceivedTime: record.responseReceivedTime,
+      transferSize: record.transferSize
+    };
+  }
+
+  /**
+   * Links together network requests (filtered by predicate) with their initiator, building a DAG.
```
paulirish: can you add a comment on how the predicate is used?
AFAICT if it returns false on a request, that request and any of its children will be excluded from the graph. Is that right?

lighthouse-core/gather/computed/request-chains.js
```diff
@@ -0,0 +1,122 @@
...
+    const requestChains = {};
+    for (const request of requestsToExamine) {
+      // Work back from this request up to the root. If by some weird quirk we are giving request D
+      // here, which has ancestors C, B and A (where A is the root), we will build array [C, B, A]
+      // during this phase.
+      const ancestors = [];
+      let ancestorRequest = request.initiatorRequest();
+      let node = requestChains;
+      while (ancestorRequest) {
+        // If the parent request isn't a high priority request it won't be in the
```
brendankenny: should update comment to be more generic

lighthouse-core/gather/computed/request-chains.js
patrickhulce: yep, that's how it was used over in critical. done

lighthouse-core/gather/computed/request-chains.js
```diff
@@ -0,0 +1,124 @@
...
+      };
+    }
+
+    return requestChains;
+  }
+
+  /**
+   * @param {!DevtoolsLog} devtoolsLog
+   * @param {!ComputedArtifacts} artifacts
+   * @return {!Promise<!Object>}
```
brendankenny: no one's taken this on yet but we probably should :)

brendankenny: oh, just kidding, we did it in [`crc-details-renderer`](https://github.com/GoogleChrome/lighthouse/blob/57bcf43a661199d55a7f39ecc008815a2da1fcc6/lighthouse-core/report/v2/renderer/crc-details-renderer.js#L203). Can use `CriticalRequestChainRenderer.CRCNode` but we can make it a generic (not just critical) RequestNode type?

patrickhulce: done

lighthouse-core/gather/computed/request-chains.js
patrickhulce: done

patrickhulce:
I don't foresee anyone needing to call `extractChain` with a predicate anytime soon, and if they do they can implement tuple arguments for computed gatherers :) this was mostly a convenience to maintain the existing contract of CriticalRequestChains without causing a bunch of unnecessary noise

paulirish:
Tests are busted in a few ways, it looks like: https://travis-ci.org/GoogleChrome/lighthouse/jobs/248562571

lighthouse-core/gather/computed/request-chains.js
```diff
@@ -0,0 +1,124 @@
...
+    return 'RequestChains';
+  }
+
+  /**
+   * @param {!WebInspector.NetworkRequest} record
+   * @return {!CriticalRequestChainRenderer.CRCRequest}
+   */
+  static flattenRecord(record) {
+    return {
+      requestId: record.requestId,
```
paulirish: let's add priority into this record. (you'll have to normalize between initialpriority and final, i think?

patrickhulce:
based on my digging around last night I'm not going to need this afterall since I'll need to reconstruct dependencies myself for predictive perf, sorry for the wasted cycles :)

Issue #2614, opened 2017-06-28
Headless-Chrome example does not launch headless chrome


asanchezgiraldo:
I am running your example provided in the documentation but I am still not getting Chrome to run headless.

My config is 
```json
    "chai": "^4.0.2",
    "htmlparser2": "^3.9.2",
    "lighthouse": "^2.1.0",
    "mocha": "^3.4.2",
    "supertest": "^3.0.0"
```
 Mac Sierra 10.12.5
 Chrome Version 59.0.3071.115 (Official Build) (64-bit)
 Node v6.1.0

the code is

```nodejs
const lighthouse = require('lighthouse');
const chromeLauncher = require('lighthouse/chrome-launcher');

function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
  return chromeLauncher.launch().then(chrome => {
    flags.port = chrome.port;
    return lighthouse(url, flags, config).then(results =>
      chrome.kill().then(() => results));
  });
}

const flags = {
  chromeFlags: ['--headless']
};

launchChromeAndRunLighthouse('https://github.com', flags).then(results => {
  // Use results!
});
```

ebidel:
You need to pass `flags` to the launcher: `chromeLauncher.launch(flags)`. We'll update that doc :)

asanchezgiraldo:
@ebidel now is working, thank you so much for your reply

PR #2615, opened 2017-06-28
docs: headless doc was missing flag

ebidel:
Fixes https://github.com/GoogleChrome/lighthouse/issues/2614

### DIFF
--- a/docs/headless-chrome.md
+++ b/docs/headless-chrome.md
@@ -75,7 +75,7 @@ const lighthouse = require('lighthouse');
 const chromeLauncher = require('lighthouse/chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
-  return chromeLauncher.launch().then(chrome => {
+  return chromeLauncher.launch(flags).then(chrome => {
     flags.port = chrome.port;
     return lighthouse(url, flags, config).then(results =>
       chrome.kill().then(() => results));


docs/headless-chrome.md
```diff
@@ -75,7 +75,7 @@ const lighthouse = require('lighthouse');
 const chromeLauncher = require('lighthouse/chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
-  return chromeLauncher.launch().then(chrome => {
+  return chromeLauncher.launch(flags).then(chrome => {
```
paulirish: hate to bikeshed a small commit like this but...

should we call this `options` instead of `flags`? 

also we're technically providing the same flags to both LH and CLauncher, even tho launcher has [unique options](https://github.com/GoogleChrome/lighthouse/tree/master/chrome-launcher#launchopts). it just so happens that `chromeFlags` and `port` are the same in both.
so a slightly more correct example would have different options objects for both launcher and LH. :/
wdyt?

ebidel: In the interest of keeping things simple, we should keep them simple :)

paulirish: aiight sgtm

docs/headless-chrome.md
```diff
@@ -75,7 +75,7 @@ const lighthouse = require('lighthouse');
 const chromeLauncher = require('lighthouse/chrome-launcher');
```
cmtasha143: D

PR #2616, opened 2017-06-28
feature(launcher): add launch retries via LaunchManager

### LINKED ISSUE #2556: Launcher ECONNREFUSED flake

paulirish:
Expected to fix #2556 

Introduce LaunchManager which handles retrying Chrome on connection failure as also SIGINT binding. (We've been rebinding it)

For retrying, if we hit `ECONNREFUSED`, we'll retry up to 3 times total, then reject.

image

Also a few superficial changes inside of the launcher class:

1.`cleanup` was specific to the `isDebuggerReady` work. I've nested it there
2. `isDebuggerReady` renamed to `doesDebuggingPortConnect` because "Ready" doesn't say much. (open to bikeshedding, tho)
3. Tweaked the logging around waiting for this connection, notably moving the polling logs into `verbose`:

image

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -18,10 +18,13 @@ const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
 const isWindows = process.platform === 'win32';
 const _SIGINT = 'SIGINT';
-const _SIGINT_EXIT_CODE = 130;
 const _SUPPORTED_PLATFORMS = new Set(['darwin', 'linux', 'win32']);
 
 type SupportedPlatforms = 'darwin'|'linux'|'win32';
+
+interface LighthouseError extends Error {
+  code?: string
+}
 
 export interface Options {
   startingUrl?: string;
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
   spawn?: typeof childProcess.spawn;
 }
 
-export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
-  opts.handleSIGINT = defaults(opts.handleSIGINT, true);
+/**
+ * LaunchManager handles launch retries and SIGINT binding
+ */
+export class LaunchManager {
+  private instance?: Launcher;
+  private opts: Options;
+  private isSigintBound: boolean;
+  private maxLaunchAttempts: number;
+  private attempts: number;
 
-  const instance = new Launcher(opts);
-
-  // Kill spawned Chrome process in case of ctrl-C.
-  if (opts.handleSIGINT) {
-    process.on(_SIGINT, async () => {
-      await instance.kill();
-      process.exit(_SIGINT_EXIT_CODE);
-    });
+  constructor() {
+    this.isSigintBound = false;
+    this.maxLaunchAttempts = 3;
+    this.attempts = 1;
   }
 
-  await instance.launch();
+  setOptions(opts: Options = {}) {
+    this.opts = opts;
+  }
 
-  return {pid: instance.pid!, port: instance.port!, kill: async () => instance.kill()};
+  async launchInstance(): Promise<LaunchedChrome> {
+    if (this.instance) {
+      Promise.reject(new Error('Already a launcher instance, can\'t create a second.'));
+    }
+
+    this.instance = new Launcher(this.opts);
+
+    this.handleSigint();
+    await this.instance.launch().catch(err => {
+      return this.retry(err);
+    });
+
+    return {
+      pid: this.instance.pid!,
+      port: this.instance.port!,
+      kill: async () => this.instance!.kill()
+    };
+  }
+
+  async retry(err: LighthouseError) {
+    if (err.code !== 'ECONNREFUSED') {
+      Promise.reject(err);
+    }
+    this.attempts++;
+    if (this.attempts <= this.maxLaunchAttempts) {
+      log.warn(
+          'ChromeLauncher',
+          `Connection refused. Retrying attempt #${this.attempts}/${this.maxLaunchAttempts}`);
+    } else {
+      log.error('ChromeLauncher', 'Reached maximum relaunch attempts. Quitting...');
+      return Promise.reject(err);
+    }
+
+    await this.instance!.kill();
+    this.instance = undefined;
+    return this.launchInstance();
+  }
+
+  async handleSigint() {
+    this.opts.handleSIGINT = defaults(this.opts.handleSIGINT, true);
+    // Kill spawned Chrome process in case of ctrl-C.
+    if (this.opts.handleSIGINT && !this.isSigintBound) {
+      this.isSigintBound = true;
+      process.on(_SIGINT, async () => {
+        await this.instance!.kill();
+      });
+    }
+  }
 }
 
+/**
+ * Launcher handles the lifecycle of a unique process of Chrome
+ */
 export class Launcher {
   private tmpDirandPidFileReady = false;
   private pollInterval: number = 500;
@@ -143,7 +201,7 @@ export class Launcher {
 
       // If an explict port is passed first look for an open connection...
       try {
-        return await this.isDebuggerReady();
+        return await this.doesDebuggingPortConnect();
       } catch (err) {
         log.log(
             'ChromeLauncher',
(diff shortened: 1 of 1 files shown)


chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
+  private instance?: Launcher;
+  private opts: Options;
+  private isSigintBound: boolean;
+  private maxLaunchAttempts: number;
+  private attempts: number;
+
+  constructor() {
+    this.isSigintBound = false;
+    this.maxLaunchAttempts = 3;
+    this.attempts = 1;
```
patrickhulce: I'm a little confused the fact it's already `1`, how about initializing to `0` and treating it as 'failedAttempts'?

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
 }
 
-export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
-  opts.handleSIGINT = defaults(opts.handleSIGINT, true);
+/**
+ * LaunchManager handles launch retries and SIGINT binding
+ */
+export class LaunchManager {
+  private instance?: Launcher;
+  private opts: Options;
```
patrickhulce: are we missing some ts settings, isn't this `opts?`

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
+    this.attempts = 1;
+  }
+
+  setOptions(opts: Options = {}) {
+    this.opts = opts;
+  }
+
+  async launchInstance(): Promise<LaunchedChrome> {
+    if (this.instance) {
+      Promise.reject(new Error('Already a launcher instance, can\'t create a second.'));
```
patrickhulce: need to return this?

unrelated I wonder if ts async is fancy enough to wrap the whole thing in a `promise.resolve().then()` so we could just throw it

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
+    await this.instance!.kill();
+    this.instance = undefined;
+    return this.launchInstance();
+  }
 
-  return {pid: instance.pid!, port: instance.port!, kill: async () => instance.kill()};
+  async handleSigint() {
+    this.opts.handleSIGINT = defaults(this.opts.handleSIGINT, true);
+    // Kill spawned Chrome process in case of ctrl-C.
+    if (this.opts.handleSIGINT && !this.isSigintBound) {
```
patrickhulce: nice

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
+      this.isSigintBound = true;
+      process.on(_SIGINT, async () => {
```
patrickhulce: snap this works for proc.exit!? that's awesome

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,79 @@ export interface ModuleOverrides {
...
+      pid: this.instance.pid!,
+      port: this.instance.port!,
+      kill: async () => this.instance!.kill()
+    };
   }
 
-  await instance.launch();
+  async retry(err: LighthouseError) {
+    if (err.code !== 'ECONNREFUSED') {
+      Promise.reject(err);
```
patrickhulce: need to return this?

chrome-launcher/chrome-launcher.ts
```diff
@@ -310,3 +367,9 @@ export class Launcher {
     });
   }
 };
+
+const manager = new LaunchManager();
```
brendankenny: We should avoid a singleton if possible. Won't this cause issues e.g. launching two instances of Chrome from a single node process and that `attempts` won't reset between runs?

chrome-launcher/chrome-launcher.ts
paulirish: you mean being async in handling these? Yeah tested it a few times, it works well.

Also didn't mention i got rid of `process.exit()` in launcher as it was technically unnecessary.

chrome-launcher/chrome-launcher.ts
paulirish: changed to a `console.assert` which just throws right away.

chrome-launcher/chrome-launcher.ts
paulirish: done

chrome-launcher/chrome-launcher.ts
paulirish: don't understand what you're asking here.. hmm?

chrome-launcher/chrome-launcher.ts
paulirish: done

paulirish:
You know it's also possible none of this actually fixed the root issue... 😭 😭 

https://travis-ci.org/GoogleChrome/lighthouse/jobs/248176453#L1871

image

chrome-launcher/chrome-launcher.ts
patrickhulce: oh I'm just confused how typescript can let this compile when it's not initialized in the constructor, it will definitely be undefined...

chrome-launcher/chrome-launcher.ts
patrickhulce: sweet, yeah didn't know it'd wait on a promise here and then exit, just thought if you handled it you had to exit yourself at that point

chrome-launcher/chrome-launcher.ts
patrickhulce: yeah good point, shouldn't be a problem to create one for each launch though should it?

paulirish:
> But is there any reason to separate the two?

I tried this at first. The `retry()` method is odd though. Let's say its on the instance:

```js
  retry() {
    this.kill();  // roughly ok
    log('about to retry');
    return launch(this.opts);  // instance.retry() is now creating a fresh instance
    // hopefully the state/return values all work out? 
  }
```

It didn't feel great. Plus the sigint handling means you have 1 listener bound but inside of it you're going to `kill()` whatever launcher instance is active. And you're limited in keeping track of your active instance. 

That's roughly how we got to here.

paulirish:
I'm restarting travis a few times to see if I can get it to flake. 
So far it looks like it's doing well.

chrome-launcher/chrome-launcher.ts
```diff
@@ -200,53 +255,53 @@ export class Launcher {
...
         reject(err);
       });
       client.once('connect', () => {
-        this.cleanup(client);
+        cleanup(client);
         resolve();
       });
     });
+
+    function cleanup(client?: net.Socket) {
```
samccone: any advantage to inlining this?

chrome-launcher/chrome-launcher.ts
```diff
@@ -200,53 +255,53 @@ export class Launcher {
...
+        client.unref();
+      }
+    }
   }
 
-  // resolves when debugger is ready, rejects after 10 polls
+  // resolves when debugging port is ready, polls 10 times every 500ms
   private waitUntilReady() {
-    const launcher = this;
+    const maxPortCheckRetries = 10;
```
samccone: class private const

chrome-launcher/chrome-launcher.ts
```diff
@@ -310,3 +364,9 @@ export class Launcher {
     });
   }
 };
+
+const manager = new LaunchManager();
+export function launch(opts: Options = {}): Promise<LaunchedChrome> {
```
samccone: compiler can infer the return type here

chrome-launcher/chrome-launcher.ts
```diff
@@ -45,24 +48,76 @@ export interface ModuleOverrides {
   spawn?: typeof childProcess.spawn;
 }
 
-export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
-  opts.handleSIGINT = defaults(opts.handleSIGINT, true);
+/**
+ * LaunchManager handles launch retries and SIGINT binding
+ */
+export class LaunchManager {
```
samccone: should we pull this into another file?

chrome-launcher/chrome-launcher.ts
patrickhulce: do we want it to?

samccone: it is redundant :)

paulirish:
We shouldn't need this any longer. This was the wrong fix. 

POSIX sometimes holds ports as used a little longer. A revised approach will use Chrome to find the port (a la puppeteer)

But in the meantime, https://github.com/GoogleChrome/chrome-launcher/pull/19 and https://github.com/GoogleChrome/chrome-launcher/pull/21 handle this.

PR #2623, opened 2017-06-29
update "page load fast enough on 3G" helpText

kaycebasques:


lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -31,7 +31,9 @@ class LoadFastEnough4Pwa extends Audit {
       category: 'PWA',
       name: 'load-fast-enough-for-pwa',
       description: 'Page load is fast enough on 3G',
-      helpText: 'Satisfied if First Interactive is less than 10 seconds, as defined by the [PWA Baseline Checklist](https://developers.google.com/web/progressive-web-apps/checklist). Network throttling is required (specifically: RTT latencies >= 150 RTT are expected).',
+      helpText: 'Ensuring that page load is fast enough on a 3G network ensures that ' +
```
brendankenny: maybe reword to avoid Ensuring/ensures?

kaycebasques: How about:

A fast page load over a 3G network ensures a good mobile user experience.

paulirish: that works for me. 

kayce can you add something like ` Network throttling is required (specifically: RTT latencies >= 150 RTT are expected)` to the doc page?  This is an important detail. :)

side note: we'll have to make sure the tone of these are consistent.. e.g. if we always go with a matter-of-fact sentence rather than describing the audit.

kaycebasques:
@paulirish Re: describing the audit, we decided that if a reference doc exists, then the help text just explains why the audit is important. Else, the help text can be a stopgap for describing the audit

see [helpText guidelines](https://github.com/GoogleChrome/lighthouse/blob/master/CONTRIBUTING.md#helptext-guidelines)

PR #2638, opened 2017-07-04
Speedup travis

wardpeet:
Smoke tests:
- Create separate configs to reduce gatherers and audits

Travis tasks:
- Make everything concurrent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ cache:
     - lighthouse-viewer/node_modules
     - /home/travis/.rvm/gems/
 install:
-  - gem install travis-artifacts
   - yarn
   # travis can't handle the parallel install (without caches)
   - yarn run install-all:task:windows
@@ -51,21 +50,27 @@ jobs:
       node_js: "6.9.1"
       env: UNIT=true
       install: skip
+      addons: skip
     - node_js: "6.9.1"
       env: SMOKEHOUSE=true
       install: skip
+      addons: skip
     - node_js: "7"
       env: UNIT=true
       install: skip
+      addons: skip
     - node_js: "7"
       env: SMOKEHOUSE=true
       install: skip
+      addons: skip
     - node_js: "8"
       env: UNIT=true
       install: skip
+      addons: skip
     - node_js: "8"
       env: SMOKEHOUSE=true
       install: skip
+      addons: skip
 before_cache:
   # the `yarn compile-devtools` task adds these to node_modules, which slows down caching
   - rm -rf ./node_modules/temp-devtoolsfrontend/
@@ -76,3 +81,5 @@ after_failure:
 #  - grep 'No screenshots' perf.json && travis-artifacts upload --path perf-0.trace.json
 addons:
   chrome: stable
+  artifacts:
+    s3_region: "us-west-2"
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ cache:
     - lighthouse-cli/node_modules
     - lighthouse-extension/node_modules
     - lighthouse-viewer/node_modules
-    - /home/travis/.rvm/gems/
 install:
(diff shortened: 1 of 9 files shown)


wardpeet:
@paulirish sorry for taking so long...

.travis.yml
```diff
@@ -32,14 +32,10 @@ before_script:
   - sleep 3 # wait for xvfb to boot
   - yarn build-all
 script:
-  - yarn test-cli-formatting
-  - yarn test-launcher-formatting
-  - yarn lint
-  - yarn unit
-  - yarn closure
+  - yarn test-cli-formatting & yarn test-launcher-formatting & yarn lint & yarn unit & yarn closure & yarn compile-devtools & wait
```
paulirish: will a non-zero exit code in these background jobs trigger a failure?

By default I don't think it does, which is partially why i avoided doing it like so.

.travis.yml
```diff
@@ -32,14 +32,10 @@ before_script:
...
   - yarn smoke
   - yarn smokehouse
-  - yarn compile-devtools
```
paulirish: why remove this one?

wardpeet: it's not removed it's also running together in the background jobs trigger

stramel:
@wardpeet Assuming build matrix didn't work? :/

lighthouse-cli/test/smokehouse/offline-config.js
```diff
@@ -0,0 +1,35 @@
...
+  extends: 'lighthouse:default',
+  settings: {
+    onlyAudits: [
+      'is-on-https',
+      'redirects-http',
+      'service-worker',
+      'works-offline',
+      'viewport',
+      'without-javascript',
+      'user-timings',
```
patrickhulce: this is kind of a lot that we assert for something named "offline-config" 😆  is there anything that seems redundant we can punt?

patrickhulce:
this is great thanks @wardpeet!

to @stramel 's build matrix point, it would be awesome to split out the smoke tests at least from the rest so we can 
* at a glance tell if it was a likely flake or real failure
* just rebuild those jobs instead of the whole shebang

.travis.yml
patrickhulce: how much of the time difference is a result of this change does it make reading the output a lot more difficult?

stramel:
Is matrix not autogenerating properly with `node`, and `env`? Also, it may be useful to set `fast_finish: true` and then `allow_failures` section for flaky test suites

wardpeet:
@stramel thanks forgot about build matrix. Yes, just experimenting for now.
It looks like https://docs.travis-ci.com/user/build-stages/ is more something I'm looking for

@brendankenny probably a good idea to do a full smoketest as well. Offline might indeed be the best one to keep a full config on.

stramel:
@wardpeet Oh, nice! I think a combination of build matrix with build stages is exactly what we need! fail_fast should help. Allow_failures would be cool to try out for those flaky test suites

brendankenny:
@wardpeet how are you feeling about this? Hard to follow the commits :)

If any individual optimization is causing trouble, we could do this in pieces instead, landing what's working and pushing the rest to the next PR

stramel:
I agree with @brendankenney, I didn't mean to cause this much headache lol

wardpeet:
@brendankenny might be better indeed, most of it is just exploring travis-ci XD. It boots up a new docker image which comes with a cost as well so looking for the best option here but I can definitely make a seperate PR for the smoke tests

paulirish:
* Run master on node 6,7,8
* Run PRs on just node 6. 3 concurrent.
* Can adopt cloud storage for cache. Get shared across 3 concurrent workers
  * Will can help with image cleanup
* A bunch of other things i missed. :D

wardpeet:
i'm going to close this one and create separate PRs

paulirish:
👍  yeoman's work right here! thanks ward!

PR #2641, opened 2017-07-05
upgrade rimraf to latest

### LINKED ISSUE #2618: Chrome launcher's use of rimraf causes jest tests to fail
Quoted from https://github.com/isaacs/rimraf/issues/155 -- I think the true issue comes from lighthouse however:

> When I run chrome-headless I get the error

```
assert.js:81
  throw new assert.AssertionError({
  ^
AssertionError: false == true
    at rmdir (/Users/fabio/repo/prerender-chrome-headless/node_modules/chrome-launcher/node_modules/rimraf/rimraf.js:159:5)
    at /Users/fabio/repo/prerender-chrome-headless/node_modules/chrome-launcher/node_modules/rimraf/rimraf.js:95:13
    at FSReqWrap.oncomplete (fs.js:114:15)
```

> The offending line is this [one](https://github.com/isaacs/rimraf/blob/master/rimraf.js#L217) and the assertion fails because originalEr appears not to be an instance of Error.

...

> You can reproduce it by cloning [this repo](https://github.com/en-japan-air/prerender-chrome-headless) and running npm test.

> I'm on Mac OSX 10.12.4
> Node 7.10.0

> It happens on node 6.11.0 and 8.1.2 as well.

For some reason, when lighthouse calls `rimraf`, it causes standard `fs` errors to not be `instanceof Error`. This is causing `rimraf`'s assertions to fail.

cc @piuccio

paulirish:
fixes #2618, https://github.com/isaacs/rimraf/issues/155, https://github.com/facebook/jest/issues/3942

I bisected and rimraf did fix this EPERM bug as of 2.3.0. it was one of these commits: https://github.com/isaacs/rimraf/compare/v2.2.8...v2.3.0

but i'm bumping all the way to 2.6.1 because why not

### DIFF
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -25,7 +25,7 @@
     "@types/node": "6.0.66",
     "lighthouse-logger": "^1.0.0",
     "mkdirp": "0.5.1",
-    "rimraf": "2.2.8"
+    "rimraf": "^2.6.1"
   },
   "version": "0.3.0",
   "description": "Launch latest Chrome with the devtools-protocol port open",
--- a/chrome-launcher/yarn.lock
+++ b/chrome-launcher/yarn.lock
@@ -1182,10 +1182,6 @@ right-align@^0.1.1:
   dependencies:
     align-text "^0.1.1"
 
-rimraf@2.2.8:
-  version "2.2.8"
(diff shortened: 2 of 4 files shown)


patrickhulce:
nice!

joaomilho:
hey people, when will this be published?

paulirish:
@joaomilho this was shipped in chrome launcher already: 0.3.1

we didn't ship a lighthouse with it just yet. that'll happen later this week.

joaomilho:
yesh, I saw that :D thanks!

PR #2645, opened 2017-07-06
log protocol error data field

paulirish:
We just found a case (outside of Lighthouse) where we were getting this error:

     Error: Protocol error (Network.continueInterceptedRequest): Invalid parameters

however there were extra details once `error.data` was logged as well:

    Error: Protocol error (Network.continueInterceptedRequest): Invalid parameters (interceptionId: string value expected)

cc @aslushnikov


paulirish:
We just found a case outside of Lighthouse where we were getting this error:

     Error: Protocol error (Network.continueInterceptedRequest): Invalid parameters

however there were extra details once `error.data` was logged as well:

    Error: Protocol error (Network.continueInterceptedRequest): Invalid parameters (interceptionId: string value expected)

PR #2654, opened 2017-07-08
Expose LHR to modules consuming cli/run.ts

paulirish:
Ran into this while writing https://github.com/GoogleChrome/lighthouse/pull/1832#issuecomment-313823375

If you just want the whole sequence and then want the LHR, it's really awkward. Adding this return at the end seems harmless. Sg?

----------------

Driveby fix to not log by default in launcher. Doesn't change LH usecases, just solo-launcher cases.

### DIFF
--- a/lighthouse-cli/test/cli/run-test.js
+++ b/lighthouse-cli/test/cli/run-test.js
@@ -25,11 +25,18 @@ describe('CLI run', function() {
     const url = 'chrome://version';
     const filename = path.join(process.cwd(), 'run.ts.results.json');
     const flags = getFlags(`--output=json --output-path=${filename} ${url}`);
-    return run.runLighthouse(url, flags, fastConfig).then(_ => {
+    return run.runLighthouse(url, flags, fastConfig).then(passedResults => {
       assert.ok(fs.existsSync(filename));
       const results = JSON.parse(fs.readFileSync(filename, 'utf-8'));
       assert.equal(results.audits.viewport.rawValue, false);
+
+      // passed results match saved results
+      assert.strictEqual(results.generatedTime, passedResults.generatedTime);
+      assert.strictEqual(results.url, passedResults.url);
+      assert.strictEqual(results.audits.viewport.rawValue, passedResults.audits.viewport.rawValue);
+      assert.deepStrictEqual(results.timing, passedResults.timing);
+
       fs.unlinkSync(filename);
     });
-  });
+  }).timeout(60000);
 });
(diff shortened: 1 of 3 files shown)


paulirish:
`: Promise<{}|void> ` ftw.

brendankenny:
> `: Promise<{}|void>` ftw.

> isn't particularly inspiring :P

lighthouse-cli/test/cli/run-test.js
```diff
@@ -25,11 +25,18 @@ describe('CLI run', function() {
...
+      // passed results match saved results
+      assert.strictEqual(results.generatedTime, passedResults.generatedTime);
+      assert.strictEqual(results.url, passedResults.url);
+      assert.strictEqual(results.audits.viewport.rawValue, passedResults.audits.viewport.rawValue);
+      assert.deepStrictEqual(results.timing, passedResults.timing);
+
       fs.unlinkSync(filename);
     });
-  });
+  }).timeout(60000);
```
patrickhulce: yikes :)

paulirish: based on a comment of yours ive been thinking we should add these so that we don't need to pass `--timeout` to the mocha CLI all the time.
it'd certainly be better for us to mark tests specifically that are expected to take over 2000ms.

patrickhulce: Yeah totally agree on marking individual tests 👍, just scary that it needed to go to 60 haha

paulirish: ah. well...... here was my thinking

60 is currently our timeout.
i can't imagine any mocha `it()` test taking longer than this one in particular.  
so i'm just matching that.  :)

on my machine this test takes 7s to complete, which means on appveyor it could take 45s?
shrug

lighthouse-cli/test/cli/run-test.js
```diff
@@ -25,11 +25,18 @@ describe('CLI run', function() {
...
+    return run.runLighthouse(url, flags, fastConfig).then(passedResults => {
       assert.ok(fs.existsSync(filename));
       const results = JSON.parse(fs.readFileSync(filename, 'utf-8'));
       assert.equal(results.audits.viewport.rawValue, false);
+
+      // passed results match saved results
+      assert.strictEqual(results.generatedTime, passedResults.generatedTime);
+      assert.strictEqual(results.url, passedResults.url);
+      assert.strictEqual(results.audits.viewport.rawValue, passedResults.audits.viewport.rawValue);
+      assert.deepStrictEqual(results.timing, passedResults.timing);
```
brendankenny: maybe audits/reportCategories length?

lighthouse-cli/test/cli/run-test.js
patrickhulce: 7 to 45 😆 sad but true #appveyorproblems

PR #2659, opened 2017-07-09
--select-chrome and --skip-autolaunch flags removed

### LINKED ISSUE #2617: --select-chrome and --skip-autolaunch should be removed from the CLI
They don't do anything any longer.

arturmiz:
Fixes #2617


chrome-launcher/chrome-finder.ts
```diff
@@ -46,25 +52,43 @@ export function darwin() {
...
+    {regex: new RegExp(process.env.CHROME_PATH), weight: 151}
   ];
   // clang-format on
 
   return sort(installations, priorities);
 }
 
+function resolveChromePath() {
+  if (canAccess(process.env.CHROME_PATH)) {
+    return process.env.CHROME_PATH as string;
```
brendankenny: is the `as string` necessary here if `canAccess()` doesn't complain in the line above?

paulirish: gotta run, but ya that looks safe:

image

paulirish: fixed in a6bbcaba44d7f44865896115338cc1a52635f54a

PR #2671, opened 2017-07-11
config rebuild: keep defaultPass

paulirish:
example config:

```js
module.exports = {
  "extends": "lighthouse:default",
  "settings": {
    "onlyAudits": [
      "is-on-https"
    ]
  }
}
```

This ends up filtering out the defaultPass as there are no audits consuming a trace from it.
There are other ways to fix this. 
This is but one.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -540,6 +540,8 @@ class Config {
     }).filter(pass => {
       // remove any passes lacking concrete gatherers, unless they are dependent on the trace
       if (pass.recordTrace) return true;
+      // Always keep defaultPass
+      if (pass.passName === 'defaultPass') return true;
       return pass.gatherers.length > 0;
     });
     return filteredPasses;


lighthouse-core/config/config.js
```diff
@@ -540,6 +540,8 @@ class Config {
     }).filter(pass => {
       // remove any passes lacking concrete gatherers, unless they are dependent on the trace
       if (pass.recordTrace) return true;
+      // Always keep defaultPass
+      if (pass.passName === 'defaultPass') return true;
```
patrickhulce: I'm still partial to checking if they requested the `devtoolsLog` artifact for keeping it, but that only applies to like 2 audits so I'm cool with this too

brendankenny: > I'm still partial to checking if they requested the `devtoolsLog` artifact for keeping it

yeah, that would have gotten us closer to where we need to be wrt to ditching `defaultPass`.

Also needs a test :P

PR #2676, opened 2017-07-11
add bundlesize to track our build size

### LINKED ISSUE #1690: Automation to track our build sizes

paulirish:
fixes #1690

This should be reporting via Commit Status API, too..

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,11 +50,13 @@
     "fast": "yarn start -- --disable-device-emulation --disable-cpu-throttling --disable-network-throttling",
     "smokehouse": "node lighthouse-cli/test/smokehouse/smokehouse.js",
     "deploy-viewer": "cd lighthouse-viewer && gulp deploy",
+    "bundlesize": "bundlesize",
     "plots-smoke": "bash plots/test/smoke.sh"
   },
   "devDependencies": {
     "@types/node": "^6.0.45",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -98,5 +100,11 @@
   "bugs": {
     "url": "https://github.com/GoogleChrome/lighthouse/issues"
   },
+  "bundlesize": [
+    {
+      "path": "./lighthouse-extension/dist/scripts/lighthouse-background.js",
+      "threshold": "400 Kb"
```
brendankenny: Isn't that too low?

paulirish: It's a gzipped size.

image

right now it doesn't support non-gzipped size reporting.

paulirish:
Now on this PR: 
image

PR #2682, opened 2017-07-12
fix(driver): prevent PerfObserver from being garbage collected

### LINKED ISSUE #2667: TTCI error: Client-side PerformanceObserver doesn't pick up all long tasks
The improvement from #2473 still isn't perfect. The problem seems to be that not everything Lighthouse considers a long task is considered a long task by the PerformanceObserver/Chrome.

An example from mobile.twitter.com where the performance observer did not report any long tasks during the highlighted portion even though there were 2.

image

The first seems like a bug because I thought layout should be included?
image

The second might be because the self-time is only 43ms with compile separated out and might be below the threshold (though this also seems like a bug)?
image

cc: @paulirish for help interpreting possible explanations here

patrickhulce:
fixes #2667 

well this was a fun 6-hour rabbit hole :)

~~I'm still going to talk to Shubie and file any necessary bugs in Chrome but gist is that our PerformanceObserver was not receiving longtask notifications when it should have been. The "fix" seems to be delayed registration of the PerformanceObserver and adding a meaningless setInterval (presumably to ensure our script doesn't get discarded?)~~ 
It turned out that it was a known limitation that the PerformanceObserver will still be garbage collected even if it's still meant to be receiving events. Attaching a reference to window for now, but Shubie's filed and will fix the underlying issue (https://bugs.chromium.org/p/chromium/issues/detail?id=742530).

I have a smoketest file I can commit, but wasn't sure if we wanted to add another *very* long smoke test (long by design since it stops delivering notifications 10+ seconds in to the load) for just this one issue.


lighthouse-core/gather/driver.js
```diff
@@ -1107,7 +1113,11 @@ function registerPerformanceObserverInPage() {
       }
     }
   });
+
   observer.observe({entryTypes: ['longtask']});
+  // HACK: A PerformanceObserver will be GC'd if there are no more references to it, so attach it to
+  // window to ensure we still receive longtask notifications
```
paulirish: add link to https://crbug.com/742530

patrickhulce:
yeah I'll add the HTML file but not add the script 👍

lighthouse-core/gather/driver.js

patrickhulce:
> what does cancelled do for this?

@brendankenny that was a drive-by fix for cases where CPU idle was never reached so you don't get an annoying "Unhandled promise rejection error" in certain cases

We call cancel and clear the timeout but if we were in the middle of waiting on the page to finish executing JS (waiting on the `evaluateAsync` promise) then we currently kick off a new timeout anyway 😆

brendankenny:
ah, got it. In that case, we can ditch keeping the reference to `lastTimeout` around?

brendankenny:
> can we add the smoketest but skip it by default? nice to have the repro available for manual checking.

I don't think we should keep it around at the very least if it's not easily runnable (script to run it, expectations, etc).

Even then it doesn't feel great without a two-tier testing setup, since we won't ever notice when it breaks or goes out of date. Tree falls in the forest, etc.

paulirish:
it's easily runnable. i'd argue it's worth it since this was a 6-hour bug that we wouldn't want to have to rediagnose and recreate a new testcase for.
But it'd be fine with putting the testcase in a gist and linking to that from a comment.

brendankenny:
> it's easily runnable

oh, as just `lighthouse http://localhost:8080/really-tricky-ttci.html` and inspect the report? I guess that's true. I was thinking as a smokehouse test, still. It's not exactly automated, though.

patrickhulce:
> can we ditch the reference to lastTimeout?

No we still want to clear the timeout we just *also* don't want to kick off a new one, unless I'm missing something?

patrickhulce:
> oh, as just lighthouse http://localhost:8080/really-tricky-ttci.html and inspect the report? I guess that's true. I was thinking as a smokehouse test, still. It's not exactly automated, though.

I actually think that's the way you'd want to run it when manually checking anyway, visually inspecting the logs was much quicker and easier when investigating this. If we don't have it set to run automatically then manual inspection is the only case we should optimize for. Also fine with the gist idea.

brendankenny:
It leaves the timeout in place, but when `checkForQuiet` is next called back, `cancelled` will be true and it would immediately quit.

Functionally the same, but keeps the pieces of state for scheduling (`isResolvable` and a cancelling mechanism) still at two instead of three.

brendankenny:
> I actually think that's the way you'd want to run it when manually checking anyway, visually inspecting the logs was much quicker and easier when investigating this. If we don't have it set to run automatically then manual inspection is the only case we should optimize for. Also fine with the gist idea.

Gist idea sounds good to me, too. Nice to have an already written test page.

patrickhulce:
> It leaves the timeout in place, but when checkForQuiet is next called back, cancelled will be true and it would immediately quit.

Yeah but why keep the user waiting for 5 extra seconds after the run just to keep the state variable count down to 2? I'll be sending a refactor PR to get rid of isResolvable soon anyhow.

Gist it is.

brendankenny:
good point, especially if we get to a point where Lighthouse can immediately quit after gathering (e.g. GAR-land). It feels like there's a simpler control flow hiding in there, but ditching `isResolvable` will help :)

PR #2687, opened 2017-07-13
Fix chrome finder on linux & osx

### LINKED ISSUE #2683: launcher: Chrome-finder is grabbing old homebrew chromes
I have a canary installed in this path: `/Users/paulirish/.homebrew/Caskroom/google-chrome-canary/latest/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary`

Unfortunately, this Canary is really old. And I have another Canary installed in `/Applications/` that is my primary and it's up to date.

Prioritization-wise, it'd make sense here to favor the `/Applications/` one above a homebrew install.  I do believe a previous @paulirish wanted the reverse which is why we're in this state. 😁

wardpeet:
We added priority regexes of `LIGHTHOUSE_CHROMIUM_PATH` and `CHROME_PATH` even when they weren't populated so we got the regex `{(:?)}` which matches everything so the first chrome in the list got hit.

We now do explicit checks to add the regex only if they env params are populated

Fixes #2683

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
 before_script:
   - gem install travis-artifacts
   - export DISPLAY=:99.0
-  - export LIGHTHOUSE_CHROMIUM_PATH="$(pwd)/chrome-linux/chrome"
+  - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
   - sh -e /etc/init.d/xvfb start
   - sleep 3 # wait for xvfb to boot
   - yarn build-all
--- a/chrome-launcher/chrome-finder.ts
+++ b/chrome-launcher/chrome-finder.ts
@@ -130,7 +130,8 @@ export function linux() {
   }
 
   const priorities: Priorities = [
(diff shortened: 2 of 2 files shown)


wardpeet:
Should I create a test file just like we have for chrome-launcher?
introduce the `export ModuleOverrides` pattern?

arturmiz:
Oops, my bad... thanks for fixing this as I was the one who introduced this bug 👍

wardpeet:
@arturmiz no worries 😉  It happens to the best of us!

paulirish:
> this looks good to me, though it looks like maybe `yarn test-launcher-formatting` isn't happy? May need to adjust a line break or something.

fixed

> Related to the code but not this fix, do you mind switching [`.travis.yml`](https://github.com/GoogleChrome/lighthouse/blob/bbe2191fdc98e958d0eef3d26e749fd531169106/.travis.yml#L30) to using `CHROME_PATH` instead of `LIGHTHOUSE_CHROMIUM_PATH`?

fixed

paulirish:
thx @wardpeet. saved my bacon!

wardpeet:
thanks @paulirish for the cleanup

Issue #2668, opened 2017-07-10
Decide on & document stable vs. experimental properties of the LHR


patrickhulce:
Currently audits need to maintain certain properties of their return object for backwards compatibility in DevTools/viewer (details, name, description, etc). However, it's not clear to external consumers what properties are safe to use within a major version and which are not (e.g. [we almost removed all of extendedInfo](https://github.com/GoogleChrome/lighthouse/pull/2596) in a patch version bump which people were likely using from previous versions).

We should decide and document which properties we're committed to keeping until v3 and make it easier to confidently consume LH results.

brendankenny:
this would go well with whatever docs generator we come up with for #2463 :)

patrickhulce:
Mostly done by #3495, future work might be reworking the extendedInfo to be more stable.

PR #2697, opened 2017-07-17
track number of completed requests in total-byte-weight extendedInfo

brendankenny:
AFAICT we don't record the raw number of network requests made by a page anywhere, since the total-byte-weight array of assets is limited to the top 10.

There are some interesting patterns with that number, apparently regardless of content/size of the requests (though details there may offer even more interesting stuff). The pattern doesn't end at 10, which is as far as the array length will reveal, so this keeps the raw number around in the `extendedInfo` (neatly sidestepping the central question of #2668 for now :) so the investigation can continue.

If we do record that number in the report, let me know and we can close instead.


brendankenny:
Yeah I was thinking total-byte-weight instead of critical-request-chains because it's not just priority requests, but I also don't have a strong preference, so if anyone does, speak up soon! :)

PR #2702, opened 2017-07-17
feat: add base fast mode config

patrickhulce:
in the name of FastMode (#2703)

starting point for the FastMode config that I'll be adding to/burning down on for fast mode work

### DIFF
--- a/lighthouse-core/config/fast-config.js
+++ b/lighthouse-core/config/fast-config.js
@@ -1,0 +1,36 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


patrickhulce:
done!

lighthouse-core/config/fast-config.js
```diff
@@ -0,0 +1,36 @@
...
+    ],
+  },
+  passes: [
+    {
+      passName: 'defaultPass',
+      useThrottling: false,
+      pauseAfterLoadMs: 0,
+      networkQuietThresholdMs: 500,
+      cpuQuietThresholdMs: 500,
+      gatherers: [],
```
paulirish: this is kinda confusing.

passName above makes sure this will "overwrite" the default `defaultPass` definition. And indeed the 29-32 properties overwrite.
But the [array is `concat`'d](https://github.com/GoogleChrome/lighthouse/blob/5d610563f55961e919fffb1f95a73aa5bda5deb0/lighthouse-core/config/config.js#L249) so the default gatherers are still included. 

Can we add a short comment here to explain what's going on? Or augment configuration.md and link to that.

Issue #2703, opened 2017-07-17
Implement FastMode MVP


patrickhulce:
The future is comprised of many things, one of which is FastMode. This issue is a placeholder and gathering point for all things FastMode. Excerpts from the yet-to-be-finalized doc are below.

## Goals
* Deliver Lighthouse's functional audits and mobile performance insights **as fast as the PageSpeed Insights** web frontend.
* Provide a coarse mobile performance score using a trace and network log that were taken **from a page load that was not throttled**.
* Identify and fix any missing and necessary **critical path information** in the tracing/network stack in Chrome.
* (Stretch) **Estimate the First Meaningful Paint and Time to Consistently Interactive** with high enough precision to disable throttling by default in the DevTools audits panel.

## Background
Lighthouse's performance evaluations today are heavily based on metrics that analyze a specific load of the page (first meaningful paint, time to consistently interactive, perceptual speed index, etc) under the assumption that the load being analyzed is representative of the load a user would experience. To receive an accurate picture of mobile performance, Lighthouse loads the page under mobile emulation with heavy network and CPU throttling. The metrics on the throttled run then provide a somewhat realistic view of end-user-perceived load performance. However, this throttled load of the page comes at a cost. Getting results is significantly slower, the size of trace files that need to be parsed is much larger, and the resources of the host machine are not fully utilized during the run. Lighthouse FastMode would need to be able to evaluate expected performance on mobile without relying on network and CPU throttling.

Lighthouse today also performs several checks that require significant evaluation time in the browser context such as the aXe accessibility library, re-compression of images, and walking the entire DOM tree for problematic event listeners. Lighthouse FastMode might need to move some of this logic into Chrome to see the speed improvements necessary for delivering results in a PageSpeed Insights timeframe.

## Overview
The performance of a webpage is determined by the execution path of its resources and tasks. If a system could perfectly capture the dependency tree and execution plan of every resource and task, estimating performance under different conditions would be as simple as manipulating the duration of each node in this execution plan DAG and selecting the completion time of the maximal node. Unfortunately, there are many challenges with capturing the dependency tree that complicate this method that Lighthouse FastMode will need to overcome. 

[...]

PR #2705, opened 2017-07-17
feat(computed-artifact): support arbitrarily many inputs

### LINKED ISSUE #2275: support tuples as computed artifact input
this would allow moving [TTCI](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/consistently-interactive.js) to a computed artifact, as it needs both a `devtoolsLog` and a `trace` as input

patrickhulce:
fixes #2275

opens the door for #2703 to be built on computed artifacts and re-used in other places

### DIFF
--- a/lighthouse-core/gather/computed/computed-artifact.js
+++ b/lighthouse-core/gather/computed/computed-artifact.js
@@ -120,7 +120,7 @@ class ComputedArtifact {
 
   /**
    * Request a computed artifact, caching the result on the input artifact.
-   * @param {*} artifact
+   * @param {...*} artifacts
    * @return {!Promise<*>}
    */
   request(...artifacts) {
--- a/lighthouse-core/gather/computed/computed-artifact.js
+++ b/lighthouse-core/gather/computed/computed-artifact.js
@@ -17,6 +17,10 @@ class ComputedArtifact {
     this._allComputedArtifacts = allComputedArtifacts;
   }
 
+  get requiredNumberOfArtifacts() {
+    return 1;
+  }
+
   /* eslint-disable no-unused-vars */
 
   /**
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/gather/computed/computed-artifact-test.js
```diff
@@ -44,4 +65,23 @@ describe('ComputedArtifact base class', () => {
...
+    const mockComputed = {computed: true};
+    const computedArtifact = new MultipleInputArtifact(mockComputed);
+
+    const obj0 = {value: 1};
+    const obj1 = {value: 2};
+    const obj2 = {value: 3};
+
+    return computedArtifact.request(obj0, obj1)
+      .then(result => assert.equal(result, 0))
+      .then(() => assert.deepEqual(computedArtifact.lastArguments, [obj0, obj1, mockComputed]))
```
brendankenny: does testing `lastArguments` add more information? It seems like testing `result` is sufficient to check that a cached result is being returned.

lighthouse-core/test/gather/computed/computed-artifact-test.js
```diff
@@ -12,22 +12,43 @@ const assert = require('assert');
...
+  }
+}
+
 describe('ComputedArtifact base class', () => {
+  it('tests correct number of inputs', () => {
+    const singleInputArtifact = new TestComputedArtifact();
+    const multiInputArtifact = new MultipleInputArtifact();
+
+    return Promise.resolve()
+      .then(() => singleInputArtifact.request(1))
```
brendankenny: the file is already using `_` for non-mocha ignorable parameters, so should probably match that

lighthouse-core/test/gather/computed/computed-artifact-test.js
patrickhulce: nothing has been testing that compute actually received the proper arguments, this is a pretty critical thing considering leaving the original line untouched (aside from the variable rename) would have totally borked every computed artifact

lighthouse-core/test/gather/computed/computed-artifact-test.js
patrickhulce: while we're on this topic, why is that style a lot of places haha? is it the extra character?

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -31,21 +35,103 @@ class ComputedArtifact {
...
+   */
+  _cacheSet(artifacts, result) {
+    return this._performCacheOperation(artifacts, (cache, key) => cache.set(key, result));
+  }
+
+  /**
+   * Asserts that the length of the array is the same as the number of inputs the class expects
+   * @param {!Array<*>} artifacts
+   */
+  _assertCorrectNumberOfArtifacts(artifacts) {
```
patrickhulce: optional arguments won't work because it relies on artifacts array being the same length for every call

consider
```js
artifacts.requestMyComputed(1)
artifacts.requestMyComputed(1, 2)
```

we get a `cannot read property has of undefined` error because the cache map was created expecting a nesting level of 1 but needed 2, it seems like a reasonable sanity check and keeps the internal structure simple. if a computed artifact wants to have something as optional it can support passing in `undefined` as the n-th argument instead, but it'll have to be explicitly passed for now

lighthouse-core/test/gather/computed/computed-artifact-test.js
brendankenny: > while we're on this topic, why is that style a lot of places haha? is it the extra character?

it predates the use of eslint on lighthouse #8 :)

I think the original conversation was that it's nice to indicate that an argument is passed in but is ignored (borrowed from several other languages) and got stretched to cover even when there are no parameters (presumably because it's one fewer characters).

Mocha does mess with that a bit, though.

lighthouse-core/gather/computed/computed-artifact.js
brendankenny: ah, yeah, we'd have to move to storing a value and a Map per node, which isn't so bad. IMO `requiredNumberOfArtifacts` is a code smell (this is a cache, not a type system), but since we don't have any audits that even want to use multiple arguments yet, I'm ok with waiting for a user if you don't want to implement now.

lighthouse-core/test/gather/computed/computed-artifact-test.js
brendankenny: > nothing has been testing that compute actually received the proper arguments

well, the computed artifact subclass tests that don't take shortcuts are :)

Anyway, that's fine.

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -31,21 +35,103 @@ class ComputedArtifact {
...
+      const className = this.constructor.name;
+      throw new Error(`${className} requires ${expected} artifacts but ${actual} were given`);
+    }
+  }
+
   /* eslint-enable no-unused-vars */
 
   /**
    * Request a computed artifact, caching the result on the input artifact.
    * @param {*} artifact
```
patrickhulce: done

patrickhulce:
@brendankenny PTAL

PR #2709, opened 2017-07-18
link remaining a11y audits to aXe docs

kaycebasques:
The current plan is to link all of the a11y audits to their respective sources in the aXe docs.

These audits currently link to references I wrote before the a11y category ballooned to 50+ audits.

Deprecating the a11y docs I wrote reduces my workload while also ensuring a consistent UX (if you click on "Learn more" in an a11y doc now, you can always expect to land on an aXe doc)


paulirish:
@kaycebasques i gave you a PR for that small change: https://github.com/kaycebasques/lighthouse/pull/1

looks like i can't update your fork directly so.. :)  wanna merge my PR?

kaycebasques:
@paulirish done

PR #2710, opened 2017-07-18
refactor: extract computeLogNormalScore method

patrickhulce:


paulirish:
Niice. Good call.

patrickhulce:
yeah I had a branch that copied/pasted this 2 more times and just thought enough is enough haha

lighthouse-core/audits/audit.js
```diff
@@ -33,6 +35,27 @@ class Audit {
...
   /**
+   * Computes and clamps the score between 0 and 100 based on the measured value, and the two control
+   * points on the log-normal distribution (point of diminishing returns and the median value).
+   *
+   * @param {number} measuredValue
+   * @param {number} diminishingReturnsValue
+   * @param {number} medianValue
+   * @return {number}
+   */
+  static computeScore(measuredValue, diminishingReturnsValue, medianValue) {
```
brendankenny: since not all audits compute their score this way, would it make more sense for this to live in `statistics.js` still? and/or a more specific method name

patrickhulce: > since not all audits compute their score this way

ah, but they should :) any more specific name ideas, `computeLogNormalScore`?

brendankenny: that's probably good if it stays in audit.js. It would be good to note in the jsdoc description something about that it returns the percentage of sites with values higher than `measuredValue` from the distribution with the given attributes. But englished good.

patrickhulce: done

PR #2715, opened 2017-07-19
Gather-runner: Get useragent before emulating

### LINKED ISSUE #2612: bug: report's userAgent is emulated userAgent
We added [`driver.getUserAgent`](https://github.com/GoogleChrome/lighthouse/blob/fadaef95137181aa41ed44fd78f17d3a8b85a2c3/lighthouse-core/gather/driver.js#L88-L90) so that we could capture the version of Chrome we're running in, however [we're calling it *after* we've set up device emulation](https://github.com/GoogleChrome/lighthouse/blob/fadaef95137181aa41ed44fd78f17d3a8b85a2c3/lighthouse-core/gather/gather-runner.js#L101-L106), so the report only includes the emulated userAgent (if emulating, obvs).

We should get the userAgent first thing (or however else to make sure it's independent of emulation).

As a bonus, we should probably also stick the emulated userAgent string in the report's [`runtimeConfig.environment`](https://github.com/GoogleChrome/lighthouse/blob/fadaef95137181aa41ed44fd78f17d3a8b85a2c3/lighthouse-core/runner.js#L324) since we do change that from time to time as new Chrome versions come out.

paulirish:
fixes #2612

~Pavel introduced [`Input.setIgnoreInputEvents`](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-setIgnoreInputEvents) so that the user can't click or interact with the page while it's being audited.~

~It's been [injected into the lighthouse connection](https://github.com/ChromeDevTools/devtools-frontend/blob/f55402717f7f73fdb75b3d33d7587d9e92ab6447/front_end/audits2/Audits2Panel.js#L543-L546) in the devtools case for 2 months now.  And obviously it makes more sense to not have it hacked in there like that. :)~

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -30,7 +30,6 @@ let GathererResults; // eslint-disable-line no-unused-vars
  *     v. register a performance observer
  *     vi. register dialog dismisser
  *     vii. clearDataForOrigin
- *     viii. ignore the user's input events
  *
  * 2. For each pass in the config:
  *   A. GatherRunner.beforePass()
@@ -106,8 +105,7 @@ class GatherRunner {
       .then(_ => driver.cacheNatives())
       .then(_ => driver.registerPerformanceObserver())
       .then(_ => driver.dismissJavaScriptDialogs())
-      .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
-      .then(_ => driver.sendCommand('Input.setIgnoreInputEvents', {ignore: true}));
+      .then(_ => resetStorage && driver.clearDataForOrigin(options.url));
   }
 
   static disposeDriver(driver) {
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -104,7 +107,8 @@ class GatherRunner {
       .then(_ => driver.registerPerformanceObserver())
       .then(_ => driver.dismissJavaScriptDialogs())
       .then(_ => resetStorage && driver.clearDataForOrigin(options.url))
-      .then(_ => gathererResults.UserAgent = [driver.getUserAgent()]);
+      .then(_ => gathererResults.UserAgent = [driver.getUserAgent()])
```
patrickhulce: heh, want to move this line above begin emulation while you're in here updating the docs :)

paulirish: sg done

paulirish:
> does this prevent artificially-generated input events from having an effect?

turns out... Yes.. yes it does. At least the ones you create over the protocol, with the rest of the `Input` domain.

As such i've decided to nuke it entirely and have left the ez bugfix in this PR. Retitled too.

cheers. 🥈

patrickhulce:
> turns out... Yes.. yes it does.

heh, oops

PR #2721, opened 2017-07-21
fix(script-blocking-first-paint): ignore latent resources

patrickhulce:
scripts that are injected into the head after first paint are falsely flagged as blocking currently


lighthouse-core/test/audits/dobetterweb/script-blocking-first-paint-test.js
```diff
@@ -29,6 +32,12 @@ describe('Script Block First Paint audit', () => {
         },
         {
           tag: scriptDetails,
+          transferSize: 100,
+          startTime: 15,
```
brendankenny: maybe add a comment indicating that this is after fCP and so expected to not show up? Not really obvious due to different units of time, unfortunately.

patrickhulce: done

Issue #2727, opened 2017-07-23
Add chrome host as parameter 


salvocanna:
I'm looking to use lighthouse (which is really well done) programmatically integrating it inside another node app and I'm having some issues as the node app and headless-chrome are running inside two different containers.

[This](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/connections/cri.js#L13-L26) assumes chrome is always located on `localhost` and allows you to change only the port.

Would it make sense to call the constructor with a config object like `{ host: DEFAULT_HOST, port: DEFAULT_PORT }` or I'm just doing something weird and there's another way to specify the host?

patrickhulce:
Not doing anything wrong, I believe you're just the first request we've had for this :)

Makes sense to have as an option though it could be plumbed through as `port` is now. Any chance you'd want to take a swing at a PR?

FrederickGeek8:
I believe the pull request I just created has fulfilled your request. You might want to check it out and make sure that everything is as you want it.

Additionally, you may want to note that the `hostname` is the second parameter of the constructor function. I made this choice as to not disrupt and current implementations of CriConnection.

salvocanna:
Sorry for such a late answer.
From my point of view the host should take precedence over the port, so I just moved the config in the constructor to an object with defaults.
I've also made a pull request (#2793) - whichever makes more sense

PR #2728, opened 2017-07-24
Add chrome hostname to CriConnection & eslint fix

FrederickGeek8:
This issue fulfills the request made in issue #2727. This change should allow for Chrome to be located on a hostname other than `localhost`, as outlined in the issue. It may be noted that I added the hostname parameter to be the second parameter in the constructor so that it does not break any previous implementations of CriConnection.

Additionally, I made a small change that allowed for the yarn tests to be passed. No functionality was changed in this (the second) commit.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -13,7 +13,6 @@ const chromeLauncher = require('lighthouse/chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
   return chromeLauncher.launch().then(chrome => {
-    flags.host = 'localhost';
     flags.port = chrome.port;
     return lighthouse(url, flags, config).then(results =>
       chrome.kill().then(() => results));
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -13,6 +13,7 @@ const chromeLauncher = require('lighthouse/chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
   return chromeLauncher.launch().then(chrome => {
+    flags.host = 'localhost';
     flags.port = chrome.port;
     return lighthouse(url, flags, config).then(results =>
       chrome.kill().then(() => results));
--- a/lighthouse-cli/cli-flags.ts
+++ b/lighthouse-cli/cli-flags.ts
@@ -12,9 +12,9 @@ const Driver = require('../lighthouse-core/gather/driver.js');
 import {GetValidOutputOptions, OutputMode} from './printer';
 
 export interface Flags {
-  port: number, chromeFlags: string, output: any, outputPath: string,
-      interactive: boolean, saveArtifacts: boolean, saveAssets: boolean, view: boolean,
-      maxWaitForLoad: number, logLevel: string, host: string
+  port: number, chromeFlags: string, output: any, outputPath: string, interactive: boolean,
+      saveArtifacts: boolean, saveAssets: boolean, view: boolean, maxWaitForLoad: number,
+      logLevel: string, host: string
 }
 
 export function getFlags(manualArgv?: string) {
--- a/lighthouse-cli/cli-flags.ts
+++ b/lighthouse-cli/cli-flags.ts
@@ -12,9 +12,9 @@ const Driver = require('../lighthouse-core/gather/driver.js');
 import {GetValidOutputOptions, OutputMode} from './printer';
 
 export interface Flags {
-  port: number, chromeFlags: string, output: any, outputPath: string, interactive: boolean,
-      saveArtifacts: boolean, saveAssets: boolean, view: boolean, maxWaitForLoad: number,
-      logLevel: string
+  port: number, host: string, chromeFlags: string, output: any, outputPath: string,
+      interactive: boolean, saveArtifacts: boolean, saveAssets: boolean, view: boolean,
+      maxWaitForLoad: number, logLevel: string
 }
 
 export function getFlags(manualArgv?: string) {
@@ -53,7 +53,7 @@ export function getFlags(manualArgv?: string) {
       .group(
           [
             'save-assets', 'save-artifacts', 'list-all-audits', 'list-trace-categories',
-            'additional-trace-categories', 'config-path', 'chrome-flags', 'perf', 'port',
+            'additional-trace-categories', 'config-path', 'chrome-flags', 'perf', 'port', 'host',
             'max-wait-for-load'
           ],
           'Configuration:')
@@ -77,6 +77,7 @@ export function getFlags(manualArgv?: string) {
             CHROME_PATH: Explicit path of intended Chrome binary. If set must point to an executable of a build of Chromium version 54.0 or later. By default, any detected Chrome Canary or Chrome (stable) will be launched.
             `,
         'perf': 'Use a performance-test-only configuration',
+        'host': 'The host to use for the debugging protocol.',
         'port': 'The port to use for the debugging protocol. Use 0 for a random port',
         'max-wait-for-load':
             'The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING: Very high values can lead to large traces and instability',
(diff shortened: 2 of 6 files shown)


wardpeet:
TBH I rather have the eslint fix out of this PR but not my call to make.

FrederickGeek8:
Tests weren't compiling without it so I saw it as a necessary fix. I can remove it if requested however...

lighthouse-core/gather/connections/cri.js
```diff
@@ -10,18 +10,20 @@ const WebSocket = require('ws');
...
 
-const hostname = 'localhost';
+const DEFAULT_HOSTNAME = 'localhost';
 const CONNECT_TIMEOUT = 10000;
 const DEFAULT_PORT = 9222;
 
 class CriConnection extends Connection {
   /**
    * @param {number=} port Optional port number. Defaults to 9222;
+   * @param {string} hostname Optional hostname. Defaults to localhost.
```
wardpeet: `@param {string}` should be `@param {string=}` as we use closure compiler for our minification. 

ref: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#optional-parameter-in-a-param-annotation

wardpeet:
@FrederickGeek8 nonetheless thanks for the PR! 🎉 

Lets wait for @paulirish @patrickhulce @brendankenny to pick this review up

FrederickGeek8:
Sorry about that, thought I put that in. Should be fixed now.

patrickhulce:
Awesome thanks so much for jumping on this @FrederickGeek8! 

What do you think about plumbing it through to https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/index.js#L38 (read from `flags.hostname`) as well so the programmatic use case will be covered?

FrederickGeek8:
Like so? Also, do you think that anything should be added to documentation?

patrickhulce:
perfect👌  and great question!

Adding this to the [CLI](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/cli-flags.ts) will give us the CLI documentation for free (which you can then use to update the copy of the output found in the README). 

We don't really have comprehensive documentation of the programmatic flags yet, but you could add a nod to setting `flags.host` over in the first example in [using programmatically docs](https://github.com/GoogleChrome/lighthouse/blob/master/docs/readme.md#using-programmatically) for now.

FrederickGeek8:
I added some documentation to the mentioned locations. I also fixed the PR conflict.

paulirish:
Presumably this does indeed work? I had thought the remote-debugging-port was only listening for localhost connections. Was there another CLI flag you needed to pass to relax that, or am i wrong?

patrickhulce:
> Presumably this does indeed work? I had thought the remote-debugging-port was only listening for localhost connections. Was there another CLI flag you needed to pass to relax that, or am i wrong?

Maybe there's a special way docker works its magic for networking containers together? I seem to remember it having something unique going on.

brendankenny:
Sounds plausible. @FrederickGeek8 or @salvocanna can you comment to clarify your use case/workflow?

salvocanna:
@paulirish indeeed, I'm using [yukinying/chrome-headless-browser-docker](https://github.com/yukinying/chrome-headless-browser-docker) and [this line](https://github.com/yukinying/chrome-headless-browser-docker/blob/master/chrome/Dockerfile#L33) forces the bind to 0.0.0.0 so allowing external connections.

I'm using one docker container with the full application in node using lighthouse too and another container to host the actual browser. Seems to work fine updating the hostname.

docs/readme.md
```diff
@@ -13,6 +13,7 @@ const chromeLauncher = require('lighthouse/chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
   return chromeLauncher.launch().then(chrome => {
+    flags.host = 'localhost';
```
brendankenny: I think we should actually remove this, sorry :) We don't document any of the other possible flags here, and we want to keep this example as simple as possible for copy/pasting to get started.

We should start a separate section to document the flags (or at least point to the CLI flags as the same as what this accepts), but that's a change for a different PR.

lighthouse-cli/cli-flags.ts
```diff
@@ -12,9 +12,9 @@ const Driver = require('../lighthouse-core/gather/driver.js');
 import {GetValidOutputOptions, OutputMode} from './printer';
 
 export interface Flags {
-  port: number, chromeFlags: string, output: any, outputPath: string, interactive: boolean,
-      saveArtifacts: boolean, saveAssets: boolean, view: boolean, maxWaitForLoad: number,
-      logLevel: string
+  port: number, host: string, chromeFlags: string, output: any, outputPath: string,
+      interactive: boolean, saveArtifacts: boolean, saveAssets: boolean, view: boolean,
+      maxWaitForLoad: number, logLevel: string
```
brendankenny: the diff would be a lot clearer (just adding `host: string`) if this had just been appended on the end here, but no biggie :)

lighthouse-viewer/app/src/github-api.js
```diff
@@ -116,7 +116,9 @@ class GithubApi {
             const filename = Object.keys(json.files)
                 .find(filename => filename.endsWith(GithubApi.LH_JSON_EXT));
             if (!filename) {
-              throw new Error(`Failed to find a Lighthouse report (*${GithubApi.LH_JSON_EXT}) in gist ${id}`);
+              throw new Error(
```
brendankenny: drop this change from the PR?

lighthouse-core/gather/connections/cri.js
```diff
@@ -10,18 +10,20 @@ const WebSocket = require('ws');
...
 const DEFAULT_PORT = 9222;
 
 class CriConnection extends Connection {
   /**
    * @param {number=} port Optional port number. Defaults to 9222;
+   * @param {string=} hostname Optional hostname. Defaults to localhost.
    * @constructor
    */
-  constructor(port) {
+  constructor(port, hostname) {
```
brendankenny: I think this is an older file, so it doesn't use them yet, but do you want to switch these two to use default parameters instead of `||`?

Slight difference in `port` since it won't turn 0 into 9222, but we don't support port 0 at this stage anyways (trying to find a debugger on a random port makes no sense)

lighthouse-cli/cli-flags.ts
```diff
@@ -77,6 +77,7 @@ export function getFlags(manualArgv?: string) {
             CHROME_PATH: Explicit path of intended Chrome binary. If set must point to an executable of a build of Chromium version 54.0 or later. By default, any detected Chrome Canary or Chrome (stable) will be launched.
             `,
         'perf': 'Use a performance-test-only configuration',
+        'host': 'The host to use for the debugging protocol.',
```
brendankenny: should copy this to https://github.com/GoogleChrome/lighthouse/blob/master/readme.md#cli-options, too (like #2757 just did for `CHROME_PATH`)

lighthouse-viewer/app/src/github-api.js
FrederickGeek8: This was to fix a conflict that was introduced from a different PR somewhere along the road.

lighthouse-core/gather/connections/cri.js
```diff
@@ -10,19 +10,18 @@ const WebSocket = require('ws');
...
 const DEFAULT_PORT = 9222;
 
 class CriConnection extends Connection {
   /**
    * @param {number=} port Optional port number. Defaults to 9222;
+   * @param {string=} hostname Optional hostname. Defaults to localhost.
    * @constructor
    */
-  constructor(port) {
+  constructor(port = DEFAULT_PORT, hostname = DEFAULT_HOSTNAME) {
```
brendankenny: these still need to be saved on `this.port` and `this.hostname` (cause of the test failures)

FrederickGeek8: My bad...

FrederickGeek8:
@brendankenny Just a heads up, I made the requested changes.

paulirish:
I renamed `host` to `hostname` everywhere, but this is now landed.

PR #2730, opened 2017-07-24
Fix eslint issue max chars in github-api.js

wardpeet:


FrederickGeek8:
Just FYI (not sure if it matters), I fixed this in #2728

paulirish:
thx

PR #2733, opened 2017-07-24
Ignoring other extension assets in request compression audit

### LINKED ISSUE #2696: Ignore assets from an extension in text compression audit
I have the [Google Dictionary](https://chrome.google.com/webstore/detail/google-dictionary-by-goog/mgijmajocgfcbeboacabfgobmjgjcoja?hl=en) installed and when I run Lighthouse I get a message for `/content.min.css` text compression is not enabled.

It would be cool if Lighthouse could know what assets are being loaded by chrome extensions and which are being loaded by the site.

Lighthouse 2.2.1

arturmiz:
Fixes #2696

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-request-compression.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-request-compression.js
@@ -14,7 +14,6 @@ const URL = require('../../lib/url-shim');
 
 const IGNORE_THRESHOLD_IN_BYTES = 1400;
 const IGNORE_THRESHOLD_IN_PERCENT = 0.1;
-const CHROME_EXTENSION_PROTOCOL = 'chrome-extension:';
 
 class ResponsesAreCompressed extends ByteEfficiencyAudit {
   /**
@@ -41,12 +40,8 @@ class ResponsesAreCompressed extends ByteEfficiencyAudit {
   static audit_(artifacts) {
     const uncompressedResponses = artifacts.ResponseCompression;
 
-    function skipUnrelatedResponses(record) {
-      return URL.getProtocol(record.url) !== CHROME_EXTENSION_PROTOCOL;
-    }
-
     const results = [];
-    uncompressedResponses.filter(skipUnrelatedResponses).forEach(record => {
+    uncompressedResponses.forEach(record => {
       const originalSize = record.resourceSize;
       const gzipSize = record.gzipSize;
       const gzipSavings = originalSize - gzipSize;
--- a/lighthouse-core/audits/byte-efficiency/uses-request-compression.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-request-compression.js
@@ -14,6 +14,7 @@ const URL = require('../../lib/url-shim');
 
 const IGNORE_THRESHOLD_IN_BYTES = 1400;
 const IGNORE_THRESHOLD_IN_PERCENT = 0.1;
+const CHROME_EXTENSION_PROTOCOL = 'chrome-extension:';
 
 class ResponsesAreCompressed extends ByteEfficiencyAudit {
   /**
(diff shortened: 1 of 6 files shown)


lighthouse-core/lib/url-shim.js
```diff
@@ -128,4 +128,13 @@ URL.equalWithExcludedFragments = function(url1, url2) {
   }
 };
 
+/**
+ * Returns protocol of the given URL
+ * @param {string} url
+ * @return {string}
+ */
+URL.getProtocol = function(url) {
+  return new URL(url).protocol;
```
brendankenny: generally we've only been adding these methods to this shim if it needs a built-in try/catch. Otherwise maybe just use `new URL(url).protocol` directly in `uses-request-compression.js`

lighthouse-core/audits/byte-efficiency/uses-request-compression.js
```diff
@@ -40,8 +41,12 @@ class ResponsesAreCompressed extends ByteEfficiencyAudit {
   static audit_(artifacts) {
     const uncompressedResponses = artifacts.ResponseCompression;
 
+    function skipUnrelatedResponses(record) {
+      return URL.getProtocol(record.url) !== CHROME_EXTENSION_PROTOCOL;
```
patrickhulce: we recently discovered that calling `new URL` in a loop on the network records can be crazy slow due to data URIs (see #2701 for discussion), how about checking `record.protocol`/`record.scheme` or doing a simple `url.startsWith` if there's still some strong opinions about avoiding the network records 

cc: @brendankenny

lighthouse-core/audits/byte-efficiency/uses-request-compression.js
```diff
@@ -40,8 +41,12 @@ class ResponsesAreCompressed extends ByteEfficiencyAudit {
...
+    }
+
     const results = [];
-    uncompressedResponses.forEach(record => {
+    uncompressedResponses.filter(skipUnrelatedResponses).forEach(record => {
```
patrickhulce: Now that I'm reading the audit I realize this actually just processes what's already been compressed over in the gatherer. 

Let's filter it [there](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js#L28) instead so we don't waste cycles re-compressing something we don't care about. Sorry for the bad tip in the issue!

lighthouse-core/audits/byte-efficiency/uses-request-compression.js
brendankenny: > we recently discovered that calling `new URL` in a loop on the network records can be crazy slow due to data URIs

gah, oh yeah. I wonder if we can automate testing for audits that will run into this.

We could at least generate a network records fixture (or devtoolsLog now, I guess), that would make it easy to test an audit with many data URIs (and add it to a "writing an audit protips" section in CONTRIBUTING). With @paulirish's desire to individually annotate tests with timeouts, it would be a red flag if an individual test was taking more than two seconds to parse the test fixture.

arturmiz: @patrickhulce That was my initial plan for that but then I discovered that url-shim and thought it may be better approach to put it in there :) As I recall there is only `record.url` no `record.protocol` so I'll go with `record.url.startsWith`

patrickhulce: Ah awesome :) you should have access to `protocol` if you end up moving it to the gatherer ;) https://github.com/GoogleChrome/lighthouse/pull/2733#discussion_r129152300

lighthouse-core/test/audits/byte-efficiency/uses-response-compression-test.js
```diff
@@ -10,9 +10,9 @@ const ResponsesAreCompressedAudit =
   require('../../../audits/byte-efficiency/uses-request-compression.js');
 const assert = require('assert');
 
-function generateResponse(filename, type, originalSize, gzipSize) {
+function generateResponse(filename, type, originalSize, gzipSize, schema = 'http:') {
```
patrickhulce: nit: can probably revert this change too now

arturmiz: Yep, I must have skipped that :)

patrickhulce:
🎉  nice job @arturmiz!

PR #2734, opened 2017-07-24
open extension's report in new window

### LINKED ISSUE #1177: Running the extension inside an incognito window does not open generated report
Running Lighthouse on a website in an incognito window, when the extension generated it opens a new tab with an URL such as `blob:chrome-extension://<EXTENSION-ID>/<REPORT-ID>`. However, incognito disallows accessing this output and displays a HTTP ERROR 404 page. Consequently copying the URL back into a normal Chrome window does show the correct report.

I run Lighthouse inside an incognito window because I have other extensions running that I do not want to interfere with my performance tests.

brendankenny:
fixes #1177

allows opening the report when Lighthouse was run on an incognito tab.

(if folks hate always opening in a new window, we could try opening in a new window only when run in incognito or something)


paulirish:
Blob urls and incognito. lol.

PR #2737, opened 2017-07-25
All audits must specify helpText and failureDescription

paulirish:
I just got a release blocking bug on the Chromium side because password pasting audit doesn't show any helptext: https://bugs.chromium.org/p/chromium/issues/detail?id=748381

Indeed it's weird to expand a details/summary in the report to see nothing.

I've added that and then decided to throw in this case and the previously implemented failureDescription case.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -193,12 +193,16 @@ function assertValidAudit(auditDefinition, auditPath) {
   if (typeof auditDefinition.meta.failureDescription !== 'string' &&
     auditDefinition.meta.informative !== true &&
     auditDefinition.meta.scoringMode !== Audit.SCORING_MODES.NUMERIC) {
-    log.warn('config', `${auditName} has no failureDescription and should.`);
+    throw new Error(`${auditName} has no failureDescription and should.`);
   }
 
   if (typeof auditDefinition.meta.helpText !== 'string') {
     throw new Error(
       `${auditName} has no meta.helpText property, or the property is not a string.`
+    );
+  } else if (auditDefinition.meta.helpText === '') {
+    throw new Error(
+      `${auditName} has an empty meta.helpText string. Please add a description for the UI.`
     );
   }
 
--- a/lighthouse-core/test/runner-test.js
+++ b/lighthouse-core/test/runner-test.js
@@ -187,6 +187,7 @@ describe('Runner', () => {
       category: 'ThrowThrow',
       name: 'throwy-audit',
       description: 'Always throws',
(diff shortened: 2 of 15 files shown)


patrickhulce:
> a release blocking bug

😆

lighthouse-core/test/runner-test.js
```diff
@@ -187,6 +187,7 @@ describe('Runner', () => {
       category: 'ThrowThrow',
       name: 'throwy-audit',
       description: 'Always throws',
+      failureDescription: 'Always throws is failing, natch',
```
patrickhulce: 😆

lighthouse-core/config/config.js
```diff
@@ -193,13 +193,17 @@ function assertValidAudit(auditDefinition, auditPath) {
...
-    log.warn('config', `${auditName} has no failureDescription and should.`);
+    throw new Error(`${auditName} has no failureDescription and should.`);
   }
 
   if (typeof auditDefinition.meta.helpText !== 'string') {
     throw new Error(
       `${auditName} has no meta.helpText property, or the property is not a string.`
     );
+  } else if (auditDefinition.meta.helpText === '') {
+    throw new Error(
```
patrickhulce: worth noting this might break some custom audits that had no need for helpText since they didn't generate an HTML report, which is unfortunate

brendankenny: > worth noting this might break some custom audits that had no need for helpText since they didn't generate an HTML report, which is unfortunate

yeah, I started the trend by checking that everything had a string `helpText` (though `''` was acceptable), but it does kind of suck that even if you're doing output for CI only you still need a `helpText`, a `description` and a `fallbackDescription`.

- We could switch back to the log warning @paulirish had added before for `fallbackDescription`, but it is handy to have a real error so that tests fail when adding a new audit
- or we could pipe in the report output type and do extra audit validation for the html report or move this validation to somewhere in the report generator
- or we could add some of these checks to our unit tests so that the things all audits need are still checked here with an error and everything else is a warning (so custom audits get a hint of how they should work if they want to work with an html report). We have something like this in [`runner-test.js`](https://github.com/GoogleChrome/lighthouse/blob/7c461bf0de546f6939ee961de5e215ef694ebf0c/lighthouse-core/test/runner-test.js#L359-L367) where we check that all core audits filename and ID match.

paulirish: ⬆️  YUP everything that he said.

i added 513c4bda to make sure our guidance on adding a custom audit also includes these guys. :)

I am gonna merge and we'll see if any custom audit authors balk. 😀

PR #2739, opened 2017-07-25
Update first-interactive.js helpText

kaycebasques:


lighthouse-core/audits/first-interactive.js
```diff
@@ -22,8 +22,9 @@ class FirstInteractiveMetric extends Audit {
       category: 'Performance',
       name: 'first-interactive',
       description: 'First Interactive (beta)',
-      helpText: 'The first point at which necessary scripts of the page have loaded ' +
-          'and the CPU is idle enough to handle most user input.',
+      helpText: 'First Interactive marks the time at which the ' +
+          'page is interactive enough to handle most user input. ' +
```
paulirish: i'd go with some of the text we have in the source comments

> marks the first moment when a website is minimally interactive.

And leave it there, allowing the docs to explain in more detail, as this metric has a lot going on under the hood.

kaycebasques: OK, but I'm gonna stick with `page` rather than website, so that we're not giving the impression that LH is somehow auditing other pages, as well

kaycebasques:
@paulirish updated per your feedback

PR #2740, opened 2017-07-25
Update consistently-interactive.js helpText

kaycebasques:


lighthouse-core/audits/consistently-interactive.js
```diff
@@ -33,8 +33,9 @@ class ConsistentlyInteractiveMetric extends Audit {
       category: 'Performance',
       name: 'consistently-interactive',
       description: 'Consistently Interactive (beta)',
-      helpText: 'The point at which most network resources have finished loading and the ' +
-          'CPU is idle for a prolonged period.',
+      helpText: 'Consistently Interactive marks the time at which the page is ' +
+          'interactive enough to handle all user input. ' +
+          '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/consistently-interactive).',
```
paulirish: CI also depends on network idle. (specifically, we've reached just 2 remaining open network requests).

I don't think this detail needs to be here in the helptext, but it should probably be in the docs.

Furthermore, the docs say "Everything shown on the screen is interactive." but we're not testing for this really. We assume it, since main thread is idle.

kaycebasques:
Hey @paulirish I updated the text to complement the first interactive, i.e. it now it just says

`Consistently Interactive marks the time at which the page is fully interactive.`

I think this is better since you mention that CI is about more than just user input

I'll put in a doc update now...

kaycebasques:
doc update: https://github.com/google/WebFundamentals/pull/4912

PR #2749, opened 2017-07-26
fix(anchors-with-no-rel): ignore same origin links

### LINKED ISSUE #2743: NoRelOpenerAudit: exclude same origin links (regression)
https://github.com/GoogleChrome/lighthouse/commit/91ad83a37c25e213a6ef8e9166e86bdc6fc5dbdd#diff-f49bd8cfa49aa46e97b212f038581c35 introduced a regression whereby same-origin links are being considered in the audit. Relative and same-origin links should be ignored by the audit.

Example from chromestatus. These relative links shouldn't show up in the report:

screen shot 2017-07-26 at 9 15 09 am

Since gatherer was switched to injecting code directly into the page (rather than using driver code), we can move back to `node.href` [here](https://github.com/GoogleChrome/lighthouse/blob/6d862b0ccc4d8e6c26ce28d9f9fbc14cd4c4dd64/lighthouse-core/gather/gatherers/dobetterweb/anchors-with-no-rel-noopener.js#L22). That will get us back to using fully resolved urls and the checks for same-originess will pass :)

patrickhulce:
fixes #2743

we actually already had the smoketest to catch this but it was updated without catching it in review last time, also fixes a rather humorous smokehouse bug where we said `difference at undefined, found undefined but expected undefined`

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -163,7 +163,7 @@ function findDifference(path, actual, expected) {
     const expectedValue = expected[key];
 
     if (!(key in actual)) {
-      return {keyPath, undefined, expectedValue};
+      return {path: keyPath, actual: undefined, expected: expectedValue};
     }
 
     const actualValue = actual[key];
--- a/lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
+++ b/lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
@@ -39,7 +39,7 @@ class ExternalAnchorsUseRelNoopenerAudit extends Audit {
     const failingAnchors = artifacts.AnchorsWithNoRelNoopener
       .filter(anchor => {
         try {
-          return anchor.href === '' || new URL(anchor.href).host !== pageHost;
+          return new URL(anchor.href).host !== pageHost;
         } catch (err) {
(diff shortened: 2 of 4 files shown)


lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
```diff
@@ -39,7 +39,7 @@ class ExternalAnchorsUseRelNoopenerAudit extends Audit {
     const failingAnchors = artifacts.AnchorsWithNoRelNoopener
       .filter(anchor => {
         try {
-          return anchor.href === '' || new URL(anchor.href).host !== pageHost;
+          return new URL(anchor.href).host !== pageHost;
```
patrickhulce: this check was redundant given the try catch handling

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -163,7 +163,7 @@ function findDifference(path, actual, expected) {
     const expectedValue = expected[key];
 
     if (!(key in actual)) {
-      return {keyPath, undefined, expectedValue};
+      return {path: keyPath, actual: undefined, expected: expectedValue};
```
brendankenny: whoops, well that won't work.

(`{undefined}` alone seems like a ridiculous thing that should be caught instead of giving `{undefined: undefined}` :)

Issue #2753, opened 2017-07-27
[Bug] --chrome-flags does not parse quoted flags with spaces correctly


jeremywiebe:
# Problem 

When using the `--chrome-flags` command-line option, it is impossible to pass embedded quoted arguments. A common use for us is to pass a customized user agent. For example:

```sh
lighthouse --chrome-flags="--user-agent=\"iPhone Test\"" www.google.com
```

This results in a user agent of `iPhone` being passed to Chrome, but then ` Test\"` being passed as a raw parameter. Chrome interprets the second parameter as a URL and you end up getting multiple tabs open. You can see an example in the screenshot below.

![image](https://user-images.githubusercontent.com/77138/28648800-57707de4-7225-11e7-917c-9604272196e5.png)

I've traced this down to [this](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/run.ts#L36) line in `run.ts` where it calls `this.chromeFlags.split(' ')`.  

# Proposed Solution 

[Edit] I've been thinking about this more. Lighthouse-cli already has a dependency on `yargs` so I'm going to try using that to parse the `chromeFlags` instead of doing it manually.

I propose that we implement proper argument parsing of the `chromeFlags` to take into account quoted strings. The simple approach (which I've got a working solution for) is to step through the `chromeFlags` string and accumulate characters. If we encounter an escaped string we ignore it and continue. Once we reach a space that isn't inside a quoted string we "split" at that point and add the characters to that point as a discrete argument.

jeremywiebe:
This is now resolved with the completion of #2754

PR #2754, opened 2017-07-27
Fix: Basic chrome-flags parsing for embedded quotes

jeremywiebe:
Related Issue: https://github.com/GoogleChrome/lighthouse/issues/2753

This PR fixes `--chrome-flags` parsing so that the flags can contain spaces if quoted properly.

## TODO
- [x] Handle other methods of embedding quotes (single quotes)
    - Arguments are now handled by `yargs`
- [x] Unit tests

### DIFF
--- a/lighthouse-cli/run.ts
+++ b/lighthouse-cli/run.ts
@@ -11,6 +11,8 @@ import * as Printer from './printer';
 import {Results} from './types/types';
 import {Flags} from './cli-flags';
 import {launch, LaunchedChrome} from '../chrome-launcher/chrome-launcher';
+
+const yargs = require('yargs');
 const lighthouse = require('../lighthouse-core');
 const log = require('lighthouse-logger');
 const getFilenamePrefix = require('../lighthouse-core/lib/file-namer.js').getFilenamePrefix;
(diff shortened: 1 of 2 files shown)


jeremywiebe:
Should have contributor stuff fixed up now.

patrickhulce:
@jeremywiebe thanks very much for this contribution!! 

Since we're already using yargs, I wonder if we could instead just parse this string with [yargs-parser](https://github.com/yargs/yargs-parser) instead of handrolling our own. This way we tackle the limitations you highlight without additional maintenance burden. Would you want to take a swing at that? :)

jeremywiebe:
@patrickhulce That's a great idea. I am in the process of reworking the solution to use `yargs` and will update this PR shortly.

jeremywiebe:
@patrickhulce I have used `yargs` directly as I couldn't see anything that had a direct dependency on `yargs-parser`. It also turns out that `yargs` seems to be able to parse the options without the current `yargs` configuration getting in the way.  

I did find that it includes some keys that we need to strip out and then when converting back to a list of strings I had to re-format them as command-line parameters before handing them off to Chrome. Let me know what you think.

lighthouse-cli/run.ts
```diff
@@ -27,13 +29,36 @@ interface LighthouseError extends Error {
   code?: string
 }
 
+function parseChromeFlags(flags: string) {
+  let args = yargs(flags).argv;
+  const allKeys = Object.keys(args);
+
+  // Remove any "mangled" arguments (args that contain dashes are duplicated as
+  // camelCased variants, but the dashed version remains)
+  let keysToFilter = allKeys.filter(k => k.indexOf('-') !== -1)
```
patrickhulce: maybe to make this easier to follow we could just do a basic filter and map in one go?

```js
return Object.keys(args)
  .filter(key => {
    return key !== '_' &&
      !key.startsWith('$') &&
      !/[A-Z]/.test(key) // no camelcase keys (can also just use what you have now :) )
  })
  .map(key => {
    return args[key] === true ? ...
  })
```

patrickhulce: could also export this function separately and throw a test on here if you want to make sure it keeps the same functionality

jeremywiebe: Yeah, that shortens it up nicely!!  Thanks. I've applied your suggestion.

brendankenny:
for the test, would be good to include something like the (currently failing) example in #2817 as well

jeremywiebe:
@patrickhulce @brendankenny Looking at that `cli/run-test.js` it feels like the wrong level to test passing chrome flags properly. Before I start into some refactoring, are you opposed to some of the helper functions being exported so they are testable? Realistically as long as `parseChromeFlags` does the right thing that should be enough verification.

patrickhulce:
[way ahead of ya](https://github.com/GoogleChrome/lighthouse/pull/2754#discussion_r131011023) ;)

yep that's exactly what I had in mind! no need to take on our cleanup work for this, throwing an export on parseChromeFlags is 👌 and we'll come around for some refactoring another time

lighthouse-cli/run.ts
jeremywiebe: Ah, I missed this second comment about exporting the function. Ok, "great minds think alike" I guess.  😏

jeremywiebe:
Ok, I've added unit tests and all tests and coverage are passing. Anything else needed to get this merged?

PR #2767, opened 2017-07-27
Add lighthouse query param to axe links

robdodson:


benschwarz:
@robdodson, If other tools slurp up Lighthouse results, then the source of the request won't exactly be lighthouse. Is that a concern?

robdodson:
speaking with Brendan, it doesn't sound like there is any non-awkward way to make this data configurable. And if a tool is consuming Lighthouse then, in a way, those links are still coming from lighthouse. So.. I don't think it's a big problem today. Maybe down the line if LH is more configurable we can readdress this

brendankenny:
yeah, I think in this case it's more just to give axe a friendly heads up on where the extra traffic is coming from as opposed to a something binding or required

PR #2775, opened 2017-07-28
Clarification of hasOfflineStartUrl rule

karolklp:
Error message is confusing for some users (https://stackoverflow.com/questions/45330087/manifest-start-url-is-not-cached-by-a-service-worker) and technically it's not true. There's check if the start_url is returned, and if not the message that it's not cached is returned. There's possibility for service worker to cache it but not return it and receive this error message.


mbj36:
This should be merged ASAP

karolklp:
Sure, that's great @paulirish.

PR #2793, opened 2017-07-29
Add hostname as parameter for GoogleChrome/lighthouse#2727

salvocanna:
Convert the parameter configs to object to make it more extensible and add the hostname parameter.
This is actually in conflict with https://github.com/GoogleChrome/lighthouse/pull/2728
Whichever makes more sense, as long as we can have a host param :)

### DIFF
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -10,19 +10,22 @@ const WebSocket = require('ws');
 const http = require('http');
 const log = require('lighthouse-logger');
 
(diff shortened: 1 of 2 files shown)


salvocanna:
CLA: I signed it!

paulirish:
Did you confirm this works? I thought I recall remote-debugging-port only listening to localhost by default. 😕

brendankenny:
Closing in favor of #2728 

@salvocanna please weigh in on that PR so we can make sure it meets everyone's needs 👍🎉

PR #2846, opened 2017-08-04
Add license to lighthouse-logger.

samccone:


mikesherov:
Thank you!

brendankenny:
is the LICENSE file necessary with one at the repo root? Or is this to make sure it's in the npm module? If so, should we add one to `chrome-launcher` as well?

samccone:
If someone consumes this module directly then yes for those users it is needed.

mikesherov:
Just as FYI, either LICENSE file or license field works for my use case, which is we use `license-to-fail` to autoscan our deps and fail builds on it and that tool looks in NPM license field or for LICENSE file

mikesherov:
Thanks again!

PR #2848, opened 2017-08-04
report: draw metric lines over the screenshots.

paulirish:
based on discussion that lighthouse team members didn't know these things shared the same timescale. :)

lines only draw on hover:

image

### DIFF
--- a/lighthouse-core/report/v2/report-styles.css
+++ b/lighthouse-core/report/v2/report-styles.css
@@ -384,10 +384,15 @@ span.lh-node:hover {
 
 /* Perf Timeline */
 
+.lh-timeline-container {
+  overflow: hidden;
+  border-top: 1px solid #eee;
+}
+
 .lh-timeline {
   padding: 0;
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/v2/report-styles.css
```diff
@@ -384,10 +385,15 @@ span.lh-node:hover {
...
 
+.lh-timeline-container {
+  overflow: hidden;
+  border-top: 1px solid var(--metric-timeline-rule-color);
+}
+
 .lh-timeline {
   padding: 0;
   padding-bottom: 0;
-  min-width: calc(var(--lh-filmstrip-thumbnail-width) * 10 + var(--default-padding) * 2);
```
patrickhulce: :/ not a major fan of this one without something to put there

paulirish: going to leave alone the empty space on the right for now. After I add in the button to view trace i'll revisit this.

Issue #2797, opened 2017-07-29
[chrome-launcher] Memory leak: cleanup SIGINT when instance is killed


andrejewski:
Every time a new Chrome instance is launched via `chrome-launcher`'s [`.launch()`](https://github.com/GoogleChrome/lighthouse/blob/07817f90e538dbbd2e8e430e12a1deab569d718c/chrome-launcher/chrome-launcher.ts#L56) a listener is placed on the SIGINT event which is never cleaned up. After 10 launches, Node warns about a potential memory leak:

```
(node:23024) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 SIGINT listeners added. Use emitter.setMaxListeners() to increase limit
```

The [kill function returned from `launch()`](https://github.com/GoogleChrome/lighthouse/blob/07817f90e538dbbd2e8e430e12a1deab569d718c/chrome-launcher/chrome-launcher.ts#L64) should clean this listener up.

mikecardwell:
I've provided an alternative fix which I believe is better. The one provided by JarrodCTaylor would still trigger the EventEmitter memory leak warnings for me, as I'm starting up lots of Chrome instances at the same time. My fix ensures that there is a maximum of one SIGINT listener at any one point in time.

paulirish:
fixed by #2959

PR #2850, opened 2017-08-04
[chrome-launcher] Memory leak: cleanup SIGINT when instance is killed

JarrodCTaylor:
Closes issue #2797

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -51,17 +51,24 @@ export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
 
   const instance = new Launcher(opts);
 
+  let sigintListener =
+      async () => {
+    await instance.kill(sigintListener);
+    process.exit(_SIGINT_EXIT_CODE);
+  }
+
   // Kill spawned Chrome process in case of ctrl-C.
   if (opts.handleSIGINT) {
-    process.on(_SIGINT, async () => {
-      await instance.kill();
-      process.exit(_SIGINT_EXIT_CODE);
-    });
+    process.on(_SIGINT, sigintListener);
   }
 
   await instance.launch();
 
-  return {pid: instance.pid!, port: instance.port!, kill: async () => instance.kill()};
+  return {
+    pid: instance.pid!,
+    port: instance.port!,
+    kill: async () => instance.kill(sigintListener)
+  };
 }
 
 export class Launcher {
@@ -272,7 +279,7 @@ export class Launcher {
     });
   }
 
-  kill() {
+  kill(sigintListener?: Function) {
     return new Promise(resolve => {
       if (this.chrome) {
         this.chrome.on('close', () => {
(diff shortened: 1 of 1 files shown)


JarrodCTaylor:
`yarn test` completes successfully for me locally, I'm not a TS aficionado though so possible I am missing something.

chrome-launcher/chrome-launcher.ts
```diff
@@ -51,17 +51,24 @@ export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
...
+    process.on(_SIGINT, sigintListener);
   }
 
   await instance.launch();
 
-  return {pid: instance.pid!, port: instance.port!, kill: async () => instance.kill()};
+  return {
+    pid: instance.pid!,
+    port: instance.port!,
+    kill: async () => instance.kill(sigintListener)
```
patrickhulce: instead of passing the sigint listener in here, how about we move the `if (opts.handleSIGINT)` to the Launcher instead?

that would keep this function cleaner and allow you to save the listener for removal to `this` instead

JarrodCTaylor:
@patrickhulce looks like this introduced a codecov issue. Any suggestions on how to best resolve?

image

chrome-launcher/chrome-launcher.ts
```diff
@@ -285,6 +285,7 @@ export class Launcher {
             execSync(`taskkill /pid ${this.chrome.pid} /T /F`);
           } else {
             process.kill(-this.chrome.pid);
+            process.removeListener(_SIGINT, this.sigintListener);
```
patrickhulce: this brings up a decent point, should we only bind to signint if it's not windows then?

chrome-launcher/chrome-launcher.ts
```diff
@@ -171,6 +166,11 @@ export class Launcher {
       this.chromePath = installations[0];
     }
 
+    // Kill spawned Chrome process in case of ctrl-C.
+    if (defaults(this.opts.handleSIGINT, true)) {
+      process.on(_SIGINT, this.sigintListener.bind(this));
```
patrickhulce: Is this actually going to be removed later? Bind will return a new object that you're not saving anywhere (you're just removing the class-wide `this.sigintListener`) Maybe we should just keep the anonymous inline arrow func and save to `this._sigintListener`

chrome-launcher/chrome-launcher.ts
wardpeet: as far as I know windows doesn't have a sigint listener on process

chrome-launcher/chrome-launcher.ts
JarrodCTaylor: Good catch about bind returning a new object. Not sure I follow what/where you are proposing with keeping and saving the arrow function?

patrickhulce: oh just something like

```
if (handleSIGINT) {
  this._signintListener = async () => {
    await this.kill();
    process.exit(...);
  }
  process.on(_SIGINT, this._sigintListener)
}

paulirish:
Thanks for working with us on this. We decided to go with #2959 to fix this... but i def appreciate your contribution here. :)

PR #2851, opened 2017-08-05
add flag to FMP to indicate if it fell back to FMP candidate

brendankenny:
there's no way to tell from the LH results that `trace-of-tab` fell back to a `firstMeaningfulPaintCandidate` event for the FMP event.  #2420 will generate a Sentry entry for that case, but it would be handy to have it in the report so that e.g. the HTTPArchive data can be queried to get frequency numbers on this as well.

Happy to bikeshed on this and/or change so that it will also work for logging when/if we turn on #2244

### DIFF
--- a/lighthouse-core/gather/computed/trace-of-tab.js
+++ b/lighthouse-core/gather/computed/trace-of-tab.js
@@ -69,6 +69,7 @@ class TraceOfTab extends ComputedArtifact {
     let firstMeaningfulPaint = frameEvents.find(
       e => e.name === 'firstMeaningfulPaint' && e.ts > navigationStart.ts
     );
+    let fmpIsFallback = false;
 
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -69,13 +69,15 @@ class TraceOfTab extends ComputedArtifact {
     let firstMeaningfulPaint = frameEvents.find(
       e => e.name === 'firstMeaningfulPaint' && e.ts > navigationStart.ts
     );
+    let fmpIsFallback = false;
```
paulirish: `fmpFellBack`

brendankenny: done

patrickhulce:
> when/if we turn on

😢

Issue #2474, opened 2017-06-09
Lighthouse treats "shrink-to-fit" property in meta-tag "viewport" as invalid


zhirzh:
In lighthouse audits,
this will pass:
`<meta name="viewport" content="width=device-width, initial-scale=1">`
this will fail:
`<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">`
with message `Invalid properties found: {"shrink-to-fit":"no"}.`

patrickhulce:
Thanks for the report @zhaoz! Looks like our dependency for metaviewport parsing [hasn't been updated in quite some time](https://github.com/dontcallmedom/metaviewport-parser).

zhirzh:
@patrickhulce that's not even my handle :joy:

patrickhulce:
oops 😆  fixed! haha sorry for the same other similarly named LH-contributor :)

ev1stensberg:
What should we do to chase this down?

patrickhulce:
@ev1stensberg since that package doesn't seem very well maintained, we should probably add a temporary workaround to ignore `shrink-to-fit` in the audit for now

In the long term we'll need to decide if we want to...

* work with the owner to get a fix in place and published
* bring it into `third_party` and fix
* drop the package and parse the values ourselves

ev1stensberg:
I can work on long term if you want to, better fix now than never :)

hmelenok:
Guys please review it dontcallmedom/metaviewport-parser/pull/4

hmelenok:
@patrickhulce i fix it for you

hmelenok:
@patrickhulce GoogleChrome/lighthouse/pull/2863 with updated version

paulirish:
fixed by #2863

PatNeedham:
I'm still getting that same problem on the chrome extension @jamesvclements is referring to as well, ran on the default build of create-react-app.

ev1stensberg:
Needs reinvestigation @patrickhulce

brendankenny:
I think this is due to the release of the extension lagging behind (it's still on 2.3.0). We're hoping to do a release of 2.5.0 on all platforms early next week.

Foorack:
Any update on this? I am still experiencing this in Chrome 62.

patrickhulce:
@Foorack this was fixed on august 7, use Chrome Canary, the CLI, or the extension for latest LH version with the fix

PR #2863, opened 2017-08-07
Lighthouse treats "shrink-to-fit" property in meta-tag "viewport" as invalid

hmelenok:
Fix for #2474 

screen shot 2017-08-07 at 12 33 34

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "jpeg-js": "0.1.2",
     "json-stringify-safe": "5.0.1",
     "lighthouse-logger": "^1.0.0",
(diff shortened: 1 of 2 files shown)


hmelenok:
I signed it!

paulirish:
The package.json change looks great, but the yarn.lock diff is much bigger than i'd expect.

want to try that part again?

hmelenok:
Hey @paulirish, thanks  to check it out. It's fixed

paulirish:
thanks very much for making this happen in both projects!

PR #2867, opened 2017-08-07
Update lighthouse-logger location in readme

mikerhyssmith:
A small Readme change to correct location of lighthouse-logger !

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -43,12 +43,12 @@ You can also craft your own config (e.g. [plots-config.js](https://github.com/Go
 
 ### Turn on logging
 
-If you want to see log output as Lighthouse runs, include the `log` module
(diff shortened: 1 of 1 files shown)


docs/readme.md
```diff
@@ -48,7 +48,7 @@ and set an appropriate logging level in your code. You'll also need to pass
 the `logLevel` flag when calling `lighthouse`.
 
 ```javascript
-const log = require('lighthouse/lighthouse-core/lib/log');
```
ebidel: can you change `log` -> `lighthouse-logger` on line 46

paulirish:
I updated it to use the straight module reference. It's a runtime dep of lighthouse so its' consumed like that. Kinda odd, but hey.

PR #2898, opened 2017-08-09
[chrome-launcher] Publish type definitions instead of source TypeScript files

### LINKED ISSUE #2896: TypeScript undefined-argument error in chrome-finder.ts
I'm getting the following errors when compiling my project:

```
59     priorities.push({regex: new RegExp(process.env.LIGHTHOUSE_CHROMIUM_PATH), weight: 150});
                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(59,40): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.


63     priorities.push({regex: new RegExp(process.env.CHROME_PATH), weight: 151});
                                          ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(63,40): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.


71   if (canAccess(process.env.CHROME_PATH)) {
                   ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(71,17): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.


75   if (canAccess(process.env.LIGHTHOUSE_CHROMIUM_PATH)) {
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(75,17): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.


139     priorities.push({regex: new RegExp(process.env.LIGHTHOUSE_CHROMIUM_PATH), weight: 100});
                                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(139,40): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.


143     priorities.push({regex: new RegExp(process.env.CHROME_PATH), weight: 101});
                                           ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/chrome-launcher/chrome-finder.ts(143,40): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

Here's my tsconfig:

```
{
  "compilerOptions": {
    "target": "ES2017",
    "module": "commonjs",
    "lib": [
      "es2017",
      "es2017.object"
    ],
    "sourceMap": true,
    "outDir": "./bin",
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "moduleResolution": "node",
    "baseUrl": "./",
    "paths": {
      "*": ["./src/@types/*", "*"]
    },
    "pretty": true
  },
  "include": [
    "./src/**/*.*"
  ]
}
```

I'm not 100% sure what's causing it, because it doesn't actually happen when I build chrome-launcher locally, but I do know that there are 2 fixes:

1) The `canAccess` method already checks whether its argument is truthy, so we can update its signature to take a `string | null | undefined`.
2) The calls to the `RegExp` constructor can have a `!` appended to the environment variable to override the `strictNullChecks` tsconfig option for that expression.

devrelm:
This _should_ fix #2896 and reduce errors caused by different ts versions, ts configs, and ts declaration file versions.

In general, this PR enables the option in chrome-launcher's tsconfig to produce declaration files and updates the .npmignore to ignore `.ts` files.

One additional change was the addition of the `@types/rimraf` package. This had to be done, as the compiler was complaining about exporting a private type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ lighthouse-cli/*.js
 !chrome-launcher/compiled-check.js
 chrome-launcher/*.map
 chrome-launcher/*.js
+chrome-launcher/*.d.ts
 
 lighthouse-cli/commands/*.map
 lighthouse-cli/commands/*.js
--- a/chrome-launcher/.npmignore
+++ b/chrome-launcher/.npmignore
@@ -9,3 +9,9 @@ test/
 .eslintrc.js
 .travis.yml
 gulpfile.js
+
+# exclude source TypeScript files
+*.ts
+
+# allow TypeScript Declaration Files
+!*.d.ts
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -7,12 +7,12 @@
 
 import * as childProcess from 'child_process';
 import * as fs from 'fs';
+import * as net from 'net';
+import * as rimraf from 'rimraf';
 import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
 import {DEFAULT_FLAGS} from './flags';
 import {makeTmpDir, defaults, delay} from './utils';
-import * as net from 'net';
-const rimraf = require('rimraf');
 const log = require('lighthouse-logger');
 const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -7,7 +7,7 @@
     "test": "mocha --require ts-node/register --reporter=dot test/**/*-test.ts --timeout=10000",
     "coverage": "nyc yarn test && nyc report --reporter=text-lcov > lcov.info",
     "test-formatting": "test/check-formatting.sh",
-    "format": "clang-format -i -style=file **/*.ts *.ts"
+    "format": "scripts/format.sh"
   },
   "devDependencies": {
     "@types/mocha": "^2.2.41",
--- a/chrome-launcher/package.json
+++ b/chrome-launcher/package.json
@@ -23,6 +23,7 @@
     "@types/core-js": "^0.9.41",
     "@types/mkdirp": "^0.3.29",
     "@types/node": "6.0.66",
+    "@types/rimraf": "^0.0.28",
     "lighthouse-logger": "^1.0.0",
     "mkdirp": "0.5.1",
     "rimraf": "^2.6.1"
--- a/chrome-launcher/scripts/format.sh
+++ b/chrome-launcher/scripts/format.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+FILES="`find . -type f \! -path '*node_modules*' \! -name '*.d.ts' -name '*.ts'`"
+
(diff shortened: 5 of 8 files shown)


patrickhulce:
I agree with publishing the `.d.ts` files instead 👍  consumers shouldn't have to deal with our typescript compilation quirks. Thanks for taking this on @devrelm!

samccone:
lgtm. thanks!

brendankenny:
looks like formatting needs to be done for this change. Can fix with `cd chrome-launcher && yarn format` (and can check locally that it worked with `yarn test-launcher-formatting`)

devrelm:
Yeah, I actually noticed that my change to .npmignore would ignore the `d.ts` files, so I've got a couple changes to make.

I wasn't seeing the formatting error locally, but I actually realized that it's because of a similar globbing issue: the current `*.ts` and `**/*.ts` globs used to pick up files for formatting were picking up d.ts files as well. I wasn't seeing the issue locally because I had removed them.

devrelm:
@brendankenny I just added a change to `./chrome-launcher/test/check-formatting.sh` to not check the formatting of *.d.ts files. Globs don't quite cut it for selecting files that match `*.ts`, but _not_ `*.d.ts`, so I switched it to use `find` instead. Let me know if you can think of a better way.

I've also updated my commit that touched `.npmignore` to (hopefully) make sure that d.ts files will be included as well.

devrelm:
So one final thought: I should probably also update the chrome-launcher `format` package.json script. It's similarly using `*.ts` and `**/*.ts` file expansion globs, so it'll be erroneously formatting *.d.ts files as well.

I see two options here:

1) make a similar change, calling `find`, in which case I'd probably just make a function that could be shared between the format and test scripts, so that we only have to make future updates in one place.

2) use the [`--glob`](/angular/clang-format#globbing-files) option for `clang-format`, which uses [node-glob](https://www.npmjs.com/package/node-glob) and so has more options to exclude files.

I'll go for the second option for now since it'll be quick, but I'll be happy to change it if you have a different preference.

devrelm:
Alright. I just did the first option, though I didn't make a shared function. Apparently importing scripts with a relative path is not a solved problem in bash, and I don't feel like going down that rabbit hole right now.

brendankenny:
@samccone can you take another look?

devrelm:
@samccone I ran `npm pack` on this locally on it appears to include the `d.ts` files and exclude the typescript files. I'm not sure if you want to sanity-check it yourself as well.

PR #2939, opened 2017-08-10
Fix errors inside spawnPromise being ignored

alexhorn:
If an error inside spawnPromise occurs it is being ignored as the async function is passed to new Promise(...) as a regular function.

This leads to the following output if you pass an empty string for chromePath:
```
(node:19352) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Bad argument
(node:19352) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -177,10 +177,10 @@ export class Launcher {
 
   private async spawnProcess(execPath: string) {
     // Typescript is losing track of the return type without the explict typing.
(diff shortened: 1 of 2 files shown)


alexhorn:
I have now added a test

alexhorn:
@patrickhulce Is this ok?

patrickhulce:
@alexhorn yep looks great thanks for the ping and the contribution! :)

Issue #2812, opened 2017-08-01
Documentation: How is the optimal and the median value calculated?


jhnns:
Hey all 👋 ,

thanks for this awesome tool. When I started to try it out, I was wondering what values are considered to be "good". They are obviously related to the RAIL model, but some of the `optimalValue`s look like arbitrary values from my point of view. 

After digging the source code, I found [this comment about the audit parameters](https://github.com/GoogleChrome/lighthouse/blob/fadaef95137181aa41ed44fd78f17d3a8b85a2c3/lighthouse-core/audits/first-meaningful-paint.js#L12-L15). After doing a bit research, I concluded that CDF is the [cumulative distribution function](https://en.wikipedia.org/wiki/Cumulative_distribution_function).

Could you elaborate on the reasoning behind this formula? I could imagine that more developers are interested in this topic.

addyosmani:
cc @kaycebasques 

A few folks have pinged me this week asking how the current medians are calculated and walking them through the source to our metric audits has been necessary. It would be 💯 if there was a note in the docs that clarify where LH uses the CDF and why _some_ of the numbers used are not quite so arbitrary.

patrickhulce:
will be fixed by #3436

vinamratasingal:
Given that we now have scoring docs, this question becomes obsolete... only question is, can we delete values to optimal/median values within the code itself?

vinamratasingal:
Patrick will delete reference to optimal/median values within the code.

Issue #2376, opened 2017-05-26
Remove references to target/optimalValue


gauntface:
For speed-index-metric I'm getting a rawValue which is lass than the optimal value but a score less than 100.

For example I can get the following info: `Value was 1222, the optimal value is < 1,250 but the score was 98`.

This get's more confusing when I can get values over 1250 and still have the same score of 98.

How is the scoring of this calculated and is the "optimal value" really worth even mentioning if hitting it doesn't mean 100.

patrickhulce:
We've been removing the optimal value in other metrics due to confusion and emphasis on the timeline instead, +1 to removing it for the remaining metrics too 👍

brendankenny:
death to `optimalValue`! The only thing going for it over just putting it in the text description if *really* needed is that it gets some nicer formatting. We could always expand the supported markdown if we wanted to keep that around.

brendankenny:
As discussed last week:
- could delete `optimalValue` (`target` in the report)
- could also better explain target value/what score really means

gauntface:
I'd say remove it.

vinamratasingal:
Closed by #3774

PR #2952, opened 2017-08-11
PSI audit: hitting ideal target should give you 100 score

ebidel:
@gauntface brought this to my attention and it's been confusing for a long time. TLDR is that users should get a score of 100/100 if you they hit the ideal target we suggest.

Example - Matt got 99/100 for having a PSI of 966. Lighthouse says the target is 1250:
"99 Perceptual Speed Index: 966 (target: < 1,250)"

So intuitively, he thought "why don't I have 100/100 for hitting the target!?!?"

### DIFF
--- a/lighthouse-core/audits/speed-index-metric.js
+++ b/lighthouse-core/audits/speed-index-metric.js
@@ -9,8 +9,9 @@ const Audit = require('./audit');
 const Util = require('../report/v2/renderer/util');
 
 // Parameters (in ms) for log-normal CDF scoring. To see the curve:
-// https://www.desmos.com/calculator/mdgjzchijg
-const SCORING_POINT_OF_DIMINISHING_RETURNS = 1250;
+// https://www.desmos.com/calculator/rva61mekla
+const SCORING_POINT_OF_DIMINISHING_RETURNS = 1840;
+const SCORING_TARGET = 1250; // Ideal. Users receive score of 100.
 const SCORING_MEDIAN = 5500;
 
 class SpeedIndexMetric extends Audit {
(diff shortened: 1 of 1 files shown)


gauntface:
❤️

brendankenny:
agreed it's confusing, but what about changing the description instead of the score curve? The curve isn't necessarily reflective of reality, but any change is going to be disruptive to anyone watching that number, so we probably shouldn't unless/until we have data to back it up.

"Target" is probably just a poor choice of word. Thinking about the curve we use, we chose the the point of diminishing returns as a control point because a site at that point is
a) pretty fast
b) could *possibly* still improve a little bit
c) it would continue to take more and more work to improve, and it's already pretty good, so the score improvement starts to decrease per effort to improve (hence "point of diminishing returns") to encourage focusing on other metrics at that point

So for Lighthouse developers, "target" makes sense because once you reach that point you're doing great and don't need to worry about this audit now! And a ~98% score is great! And technically the target says "< 1,250" :)

But for Lighthouse users, it is probably confusing/annoying that hitting the target doesn't give you 100%.

Other possible options:
- keep the same curve but make the audit description give a "target" of 900-ish, which would round up to 100%
- drop the "target" from the description and give more guidance on the curve in the docs for the audit (#2812)

patrickhulce:
Other possible option that should be familiar coming from me: don't show the 99! The scoring curve and targets are aligned in that anything that is target or better is an A+ 😛

brendankenny:
> Other possible option that should be familiar coming from me: don't show the 99!

If humans can't handle getting a 99, this may be our only option :)

[vqv751i0cyecjwcfzcu2mpxy5yk](https://www.youtube.com/watch?v=6h4GD2feZ8Q)

ebidel:
+1 on grading. Exact scoring is the root of so many of our issues!

brendankenny:
Incidentally, we're probably being pretty harsh on PSI scoring. Here's PSI scores across the entire HTTPArchive data set (blue) vs our score curve (red)

PSI curve

If a site has a PSI of around 9000, it's just above the 50th percentile in the HTTPArchive population, but you get almost 30 points off of that and score a 24% in Lighthouse

(tails are barely affected by this correction, though, so @gauntface's score doesn't really budge when judged against reality :)

brendankenny:
should we close and open an issue? Or alter the PR to remove target?

brendankenny:
Closing for tracking bug in #2376 :)

PR #2959, opened 2017-08-12
Better SIGINT handling in chrome-launcher

mikecardwell:
This is a fix for https://github.com/GoogleChrome/lighthouse/issues/2797

Currently we add a SIGINT handler for each launched Chrome instance to kill that instance on exit. My change makes it so that we track the list of running Chrome instances. When that list goes from 0 to 1, a SIGINT handler is added which kills all tracked Chrome instances. When that list goes from 1 to 0, the handler is removed.

So there is only ever a maximum of one SIGINT listener, which prevents the EventEmitter memory leak warnings.

I believe that my fix is better than the other one provided in the issue as it allows for you to have more than 10 simultaneously running Chrome instances without an EventEmitter memory leak warning.

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -23,6 +23,8 @@ const _SUPPORTED_PLATFORMS = new Set(['darwin', 'linux', 'win32']);
 
 type SupportedPlatforms = 'darwin'|'linux'|'win32';
 
+const instances = new Set();
+
 export interface Options {
   startingUrl?: string;
   chromeFlags?: Array<string>;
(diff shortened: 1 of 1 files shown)


mikecardwell:
I signed it!

mikecardwell:
I believe that my fix is better than the one provided in the issue as it allows for you to have more than 10 simultaneously running Chrome instances without an EventEmitter memory leak warning.

patrickhulce:
Thanks for the PR @mikecardwell! This lines up well with the approach @paulirish takes in #2616 of moving toward a manager for multiple instances.

chrome-launcher/chrome-launcher.ts
```diff
@@ -46,22 +48,35 @@ export interface ModuleOverrides {
...
   const instance = new Launcher(opts);
 
   // Kill spawned Chrome process in case of ctrl-C.
-  if (opts.handleSIGINT) {
-    process.on(_SIGINT, async () => {
-      await instance.kill();
-      process.exit(_SIGINT_EXIT_CODE);
-    });
+  if (opts.handleSIGINT && instances.size === 0) {
+    process.on(_SIGINT, sigintListener);
```
aslushnikov: @paulirish @JoelEinbinder this gonna yield "max listeners" error in case of 11 simultaneously launched browsers.

JoelEinbinder: instances is a global variable, so it only adds one listener per project.

aslushnikov: ah i see. So the "handleSIGINT" option is not per-browser but per-process.

PR #3028, opened 2017-08-17
launcher: mute any audio

paulirish:
via https://github.com/GoogleChrome/puppeteer/issues/350


JoelEinbinder:
Is there a way to enable audio again via flags. I'd want headful mode to have audio.

logiclabs:
I'm attempting to use chrome launcher to record a screengrab of a website and would like to include audio.  As this flag is in the defaults, there appears to be no way to override the behaviour (short of forking the entire code).  I guess the above comments suggest that it should never be used for audio..

Would it be possible to make this a non-default flag as an option that defaults to true?

PR #3084, opened 2017-08-23
Image aspect ratio

### LINKED ISSUE #2579: Image audit: check for correct aspect ratio
The aspect ratio of an image element should match the aspect ratio of the image displayed, even if the dimensions are different.

Test by comparing `img.naturalWidth`/`img.naturalHeight` with `img` element `clientWidth`/`clientHeight`.

Need to ensure image element `load` event has fired before testing.

ThisIzKp:
Fixes #2579 - Added Image aspect-ratio audit under best practices category. Checks for image natural width/height & client width/height and warns the user if the difference is more than 0.5.


ThisIzKp:
I'm sure this is not really good enough to merge, still, a lot of work might be left. Just wanted someone to do the review so that I can improve the code and see what am I missing out here and continue working.

ThisIzKp:
I signed it!

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+
+class ImageAspectRatio extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Images',
+      name: 'image-aspect-ratio',
+      description: 'Uses Images with appropriate aspect ratio',
```
patrickhulce: @kaycebasques any thoughts or guidance on descriptions and helpText here? I had thought we decided to move away from the "uses" since it can be confusing if there are no images at all, but I can't remember what we settled on.

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * @param {!Object} image
+   * @return {?Object}
+   */
+  static computeAspectRatios(image) {
+    const url = URL.elideDataURI(image.src);
+    const actualAspectRatio = (image.naturalWidth / image.naturalHeight).toFixed(2);
+    const usedAspectRatio = (image.clientWidth / image.clientHeight).toFixed(2);
```
patrickhulce: nit: let's move the `toFixed` conversion here to the end where its stringified so we don't have these extra `parseFloat` calls

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+    const doesRatiosMatch = Math.abs(
```
patrickhulce: s/does/do/

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+
+    let debugString;
+    const resultsMap = new Map();
+    images.filter(image => {
+      // filter out images that don't have a proper url and width/height
+      return image.networkRecord && image.clientWidth && image.clientHeight;
+    }).forEach(image => {
+      const processed = ImageAspectRatio.computeAspectRatios(image);
+
+      if(processed instanceof Error) {
```
patrickhulce: nit: space after if

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+        debugString = processed.message;
+        return;
+      }
+      resultsMap.set(processed.preview.url, processed);
```
patrickhulce: Using the map was useful where we needed to de-dupe the URLs but here we probably want to flag every occurrence of wrong aspect ratio? Can just directly push onto the results array in that case and skip if the ratios don't match rather than filter it below

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    const results = Array.from(resultsMap.values())
+      .filter(item => !item.doesRatiosMatch);
+
+    const headings = [
+      {key: 'preview', itemType: 'thumbnail', text: ''},
+      {key: 'url', itemType: 'url', text: 'URL'},
+      {key: 'usedAspectRatio', itemType: 'text', text: 'Aspect Ratio (Original)'},
```
patrickhulce: "Original" is a bit confusing here, how about "Displayed"?

lighthouse-core/test/audits/image-aspect-ratio-test.js
```diff
@@ -0,0 +1,86 @@
...
+            generateRecord()
+          )
+        ]
+      });
+
+      assert.equal(result.rawValue, data.listed ? 1 : 0);
+    });
+  }
+
+  testImage('is very large than natural aspect ratio', {
```
patrickhulce: s/very large/much larger/

lighthouse-core/test/audits/image-aspect-ratio-test.js
```diff
@@ -0,0 +1,86 @@
...
+    listed: true,
+    clientSize: [800, 500],
+    naturalSize: [200, 200]
+  });
+
+  testImage('is large than natural aspect ratio', {
```
patrickhulce: s/large/larger/

lighthouse-core/test/audits/image-aspect-ratio-test.js
```diff
@@ -0,0 +1,86 @@
...
+    listed: false,
+    clientSize: [400, 300],
+    naturalSize: [200, 200]
+  });
+
+  testImage('is very small than natural aspect ratio', {
```
patrickhulce: s/very small/much smaller/

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param {!Object} image
+   * @return {?Object}
+   */
+  static computeAspectRatios(image) {
+    const url = URL.elideDataURI(image.src);
+    const actualAspectRatio = (image.naturalWidth / image.naturalHeight).toFixed(2);
+    const usedAspectRatio = (image.clientWidth / image.clientHeight).toFixed(2);
+    const doesRatiosMatch = Math.abs(
+      parseFloat(actualAspectRatio) - parseFloat(usedAspectRatio)
+    ) < 0.5;
```
patrickhulce: 0.5 is a bit large I meant .05 if you got this from my suggestion :) otherwise a 500x1000 would pass even if the original were 1000x1000, let's also keep this allowable slack consistent by checking the inverse ratio if they're both larger than 1

i.e. `if (usedAspectRatio > 1 && actualAspectRatio > 1) doRatiosMatch = Math.abs((1 / usedAspectRatio) - (1 / actualAspectRatio))...`

ThisIzKp:
@patrickhulce Travis fails.
why `yarn smoke` failed in node 7? any idea about that. It says `unable to connect to chrome`.

patrickhulce:
re: test failure, yeah that's a flake issue, I restarted the job. It's one of the downsides of having such a large build matrix, even if it only happens 1% of the time we have 5 jobs x 15 smoke tests = 75 runs = .99^75 chance of success = ~50% chance of success :(

ThisIzKp:
@patrickhulce skipping css images & images that use `object-fit: contain or cover`.

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -0,0 +1,107 @@
...
+    let debugString;
+    const results = [];
+    images.filter(image => {
+      // filter out images that don't have following properties
+      // networkRecord, clientWidth, clientHeight
+      // css images, images that use `object-fit`: `cover` or `contain`
+      return image.networkRecord &&
+        image.clientWidth &&
+        image.clientHeight &&
+        image.isCss === false &&
```
patrickhulce: nit: let's just do `!image.isCss`

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -43,6 +43,8 @@ function collectImageElementInfo() {
       naturalHeight: element.naturalHeight,
       isCss: false,
       isPicture: element.parentElement.tagName === 'PICTURE',
+      usesObjectFit: window.getComputedStyle(element).getPropertyValue('object-fit') === 'cover'
```
patrickhulce: nit: let's try to make only one call to `window.getComputedStyle` and clean this line up a bit

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -78,6 +80,8 @@ function collectImageElementInfo() {
       naturalHeight: Number.MAX_VALUE,
       isCss: true,
       isPicture: false,
+      usesObjectFit: window.getComputedStyle(element).getPropertyValue('object-fit') === 'cover'
```
patrickhulce: for CSS background images it doesn't actually matter since the equivalent is expressed as part of the background sizing condition, let's just return `false` here and avoid the extra cost

ThisIzKp:
@patrickhulce did the changes 👍

patrickhulce:
@ThisIzKp just an FYI that I haven't forgotten about this :)

I've been exploring some of the issues here that make this a lot more problematic than it seems on the surface. @samdutton could you provide some example URLs of the worst offenders that motivated this? 

Despite the various mitigations @ThisIzKp has added, most of the sites I try are all false positives related to padding on small images which is much more difficult to detect properly (would need to essentially compute the box model of padding to determine how big the image actually is and would greatly add to the runtime). Not to mention all these mitigations will become much more difficult after #3211 lands for speed improvements :/

samdutton:
I'm struggling to find anything really problematic on a popular site.

Some news sites appear to have minor discrepancies between natural and display size, but nothing really extreme: [bbc.com](https://bbc.com), [wsj.com](https://www.wsj.com/) [economist.com](https://economist.com), etc.

I'm checking with the Chrome extension I wrote, [Image Checker](https://chrome.google.com/webstore/detail/image-checker/bacnicogfgpigmmenfiplfiofpkocpii).

patrickhulce:
thanks for your great work here @ThisIzKp! 🎉  🥇

paulirish:
(hat tip to patrick discovering that `elem.width` solves all our problems. so rad.)

@ThisIzKp killer first pull request, seriously. This is awesome stuff.. we really appreciate it.

samdutton:
Thanks @ThisIzKp and everyone! Appreciated.

PR #3087, opened 2017-08-23
perf(optimized-images): use Audits.getEncodedResponse

patrickhulce:
Uses the new Audits.getEncodedResponse method to determine the size savings. Time savings varies since the canvas method was more variable than outright slow but some runs went from ~6s to ~500ms. Also enables fast checking of cross-origin images which was previously impossible.

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -94,7 +94,7 @@ class OptimizedImages extends Gatherer {
   /**
    * @param {!Object} driver
    * @param {string} requestId
-   * @param {string} encoding
+   * @param {string} encoding Either webp or jpeg.
    * @return {!Promise<{encodedSize: number}>}
    */
   _getEncodedResponse(driver, requestId, encoding) {
@@ -106,7 +106,7 @@ class OptimizedImages extends Gatherer {
   /**
    * @param {!Object} driver
    * @param {{url: string, isBase64DataUri: boolean, resourceSize: number}} networkRecord
-   * @return {!Promise<?{originalSize: number, jpegSize: number, webpSize: number}>}
+   * @return {!Promise<?{fromProtocol: boolean, originalSize: number, jpegSize: number, webpSize: number}>}
    */
   calculateImageStats(driver, networkRecord) {
     // TODO(phulce): remove this dance of trying _getEncodedResponse with a fallback when Audits
@@ -116,6 +116,7 @@ class OptimizedImages extends Gatherer {
       return this._getEncodedResponse(driver, requestId, 'jpeg').then(jpegData => {
         return this._getEncodedResponse(driver, requestId, 'webp').then(webpData => {
           return {
+            fromProtocol: true,
             originalSize: networkRecord.resourceSize,
             jpegSize: jpegData.encodedSize,
             webpSize: webpData.encodedSize,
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -88,24 +93,49 @@ class OptimizedImages extends Gatherer {
 
   /**
    * @param {!Object} driver
+   * @param {string} requestId
+   * @param {string} encoding
```
brendankenny: maybe a note that this will either be `jpeg` or `webp`?

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -88,24 +93,49 @@ class OptimizedImages extends Gatherer {
...
+      return this._getEncodedResponse(driver, requestId, 'jpeg').then(jpegData => {
+        return this._getEncodedResponse(driver, requestId, 'webp').then(webpData => {
+          return {
+            originalSize: networkRecord.resourceSize,
+            jpegSize: jpegData.encodedSize,
+            webpSize: webpData.encodedSize,
+          };
+        });
+      }).catch(err => {
+        if (/wasn't found/.test(err.message)) {
```
brendankenny: re-throw error otherwise?

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -88,24 +93,49 @@ class OptimizedImages extends Gatherer {
...
+          // Mark non-support so we don't keep attempting the protocol method over and over
+          this._getEncodedResponseUnsupported = true;
```
brendankenny: can we include something like `_getEncodedResponseUnsupported` on the `extendedInfo`? Not the most elegant thing, but when looking at old reports (e.g. http archive) it would be good to have an easy way to know when the results include all images, including cross origin ones. Other ways of doing that would also be good (lighthouse version would work in a pinch, but nice to have something audit-local)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -88,24 +93,49 @@ class OptimizedImages extends Gatherer {
...
         }
+      });
+    }).then(result => {
+      if (result) return result;
+
+      // Take the slower/same-origin-limited fallback path if getEncodedResponse isn't available yet
```
brendankenny: maybe an additional note that cross-origin images aren't supported so it's clear until we delete?

lighthouse-core/test/gather/gatherers/dobetterweb/optimized-images-test.js
```diff
@@ -22,49 +22,49 @@ const traceData = {
     {
       _url: 'http://google.com/image.jpg',
       _mimeType: 'image/jpeg',
-      _resourceSize: 10,
+      _resourceSize: 10000,
```
brendankenny: what's the deal with this change? Just more realistic sizes/getting above the 4k cutoff?

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: good call, done

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: it still seemed reasonable to try the fallback path while we have it if this failed, but I'd rather know about failures for now at least until we release so sgtm :)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: yeah sure seem good to track this

patrickhulce: added a `fromProtocol` property to the image result

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
patrickhulce: split the comment to make it more explicit

lighthouse-core/test/gather/gatherers/dobetterweb/optimized-images-test.js
patrickhulce: yeah just needed to go above the minimum image size

PR #3125, opened 2017-08-25
feat(cli): proper blockedUrlPatterns support

### LINKED ISSUE #2869: Lighthouse blocked urls ignored by WPT
Filling the same bug overhere as well:
https://github.com/WPO-Foundation/webpagetest/issues/913

Lighthouse needs to have better support for setting blocked urls from the command line (currently intended for use through the node API only and isn't very smooth, setting just one blocked URL breaks things for example)

Force yargs to always fallback to an array even it has one item.

patrickhulce:
fixes #2869

opens the door for my forthcoming WPT PR :)


wardpeet:
LGTM

Issue #3110, opened 2017-08-24
chrome-launcher: add support for `pollInterval` or `pollRetries` as an option


alekseykulikov:
Currently `pollInterval = 500 ms` and `pollRetries = 10` are hard-coded. It means, that launcher waits approximately 5s and then fails.
Sometimes, this is time not enough to start a Chrome, especially on slow machines or first boot inside new env.
I'd like to provide a PR, but first I'd like to hear feedback, should it be only `pollRetries` or both.

PR #3129, opened 2017-08-25
enhancement(chrome-launcher): add support for `portCheckInterval` and `maxPortCheckRetries`

alekseykulikov:
This PR addresses #3110
I decided to use `portCheckInterval` and `maxPortCheckRetries`, because this names are more explicit.
Also this may fix #2616, because the main goal is to avoid `ECONNREFUSED`. In my experiments, this issue was related with not enough time in 100% cases, just increase from 5s to 25s fixed `ECONNREFUSED`.

### DIFF
--- a/chrome-launcher/chrome-launcher.ts
+++ b/chrome-launcher/chrome-launcher.ts
@@ -34,8 +34,8 @@ export interface Options {
   userDataDir?: string;
   logLevel?: string;
   enableExtensions?: boolean;
-  portCheckInterval?: number;
-  maxPortCheckRetries?: number;
+  connectionPollInterval?: number;
+  maxConnectionRetries?: number;
 }
 
 export interface LaunchedChrome {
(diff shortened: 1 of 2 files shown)


chrome-launcher/chrome-launcher.ts
```diff
@@ -112,6 +115,8 @@ export class Launcher {
     this.requestedPort = defaults(this.opts.port, 0);
     this.chromePath = this.opts.chromePath;
     this.enableExtensions = defaults(this.opts.enableExtensions, false);
+    this.portCheckInterval = defaults(this.opts.portCheckInterval, 500);
+    this.maxPortCheckRetries = defaults(this.opts.maxPortCheckRetries, 500);
```
patrickhulce: let's keep this default at `10` :)

alekseykulikov:
@patrickhulce thank you for review! I completely agree with new names suggestions. I've renamed arguments and fixed defaults typo.

patrickhulce:
hey @alekseykulikov sorry for the staleness of this but we were in the middle of splitting out chrome-launcher to its own [repo](https://github.com/GoogleChrome/chrome-launcher)! hate to do this, but would you mind re-opening your PR over there so we can get it merged? 

on the bright side it should make it easier and more accessible for external contributors such as yourself going forward :)

alekseykulikov:
hey @patrickhulce, sure I will resubmit it to new repo. it's good idea to keep chrome-launcher separately.

PR #3136, opened 2017-08-25
Improve travis-ci builds

wardpeet:
This has been a journey. Hopefully it's faster than before XD.

travis.yml is split in 3 stages
- prepare cache
   we build our project and upload the artifacts to gcloud
- tests
   We run all our tests in 3 steps (UNIT, SMOKE & LINT). We download all builds from gcloud and run 3 parallel tasks.
- cleanup
   We remove the glcoud artifacts

We only build PR's on node 6, master will still be build node 6,7,8

If we could cache gcloud installation as well this would help us speed up the build a bit more as well.


.travis.yml
```diff
@@ -1,14 +1,12 @@
...
-matrix:
-  include:
-    - node_js: "6.9.1"
-    - node_js: "7"
-    - node_js: "8"
 dist: trusty
 env:
   global:
     - ARTIFACTS_AWS_REGION=us-west-2
     - ARTIFACTS_S3_BUCKET=lighthouse-screenshots-debugging
```
patrickhulce: we can totally nix that screenshot saving too btw if we want some extra wins :)

.travis.yml
```diff
@@ -20,34 +18,119 @@ cache:
...
-  - yarn bundlesize
-  - yarn test-cli-formatting
-  - yarn test-launcher-formatting
-  - yarn lint
-  - yarn unit
-  - yarn closure
-  - yarn smoke
-  - yarn smokehouse
-  - yarn compile-devtools
+script: "echo \"do nothing\""
```
patrickhulce: maybe brief comment about the real scripts being decided per job?

.travis.yml
```diff
@@ -20,34 +18,119 @@ cache:
...
+jobs:
+  include:
+    - &prep-cache
```
patrickhulce: I had no idea this was a [thing](http://www.yaml.org/spec/1.2/spec.html#id2760395) that's amazing 😱 🎉

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "install-all": "npm-run-posix-or-windows install-all:task",
-    "install-all:task": "yarn & yarn install-launcher & yarn install-cli & yarn install-extension & yarn install-viewer & wait",
+    "install-all:task": "yarn; yarn install-launcher & yarn install-cli & yarn install-extension & yarn install-viewer & wait",
```
patrickhulce: sidenote while you're in here: can we move `yarn;` to happen in install-all while we're at it so it doesn't fail with "npm-run-posix-or-windows not installed" :)

wardpeet: good point :)

.travis.yml
wardpeet: @see https://docs.travis-ci.com/user/build-stages/using-yaml-aliases/

.travis.yml
wardpeet: good point :)

.travis.yml
wardpeet: I only install it when we want to send it over so not a big issue.

package.json
paulirish: yah that just means we gotta do those two sequentially...

     "install-all": "yarn; npm-run-posix-or-windows install-all:task",

wardpeet:
Travis got updated to only run node 6 on PRs I believe this is the best thing we can improve so far so I'm closing this one.

PR #3154, opened 2017-08-28
Add 'link' type for markdown links

tkadlec:
Adds a "link" type for content to be displayed in tables in the report. This type looks for a markdown link and renders it.

Example of a link type displayed in audit table results

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -35,6 +35,8 @@ class DetailsRenderer {
         return this._renderText(details);
       case 'url':
         return this._renderURL(details);
+      case 'link':
+        return this._renderLink(details);
       case 'thumbnail':
         return this._renderThumbnail(/** @type {!DetailsRenderer.ThumbnailDetails} */ (details));
       case 'filmstrip':
@@ -86,6 +88,17 @@ class DetailsRenderer {
       element.title = url;
     }
 
+    return element;
+  }
+
+  /**
+   * @param {!DetailsRenderer.DetailsJSON} text
+   * @return {!Element}
+   */
+  _renderLink(text) {
+    const snippet = text.text || '';
+    const element = this._dom.convertMarkdownLinkSnippets(snippet);
+    element.classList.add('lh-link');
     return element;
   }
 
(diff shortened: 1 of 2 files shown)


tkadlec:
@patrickhulce I opted for a brand new type, instead of including this in the `renderURL` method by default. What say you?

tkadlec:
I signed it.

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -93,6 +95,17 @@ class DetailsRenderer {
    * @param {!DetailsRenderer.DetailsJSON} text
    * @return {!Element}
    */
+  _renderLink(text) {
+    const snippet = text.text || '';
+    const element = this._dom.convertMarkdownLinkSnippets(snippet);
```
patrickhulce: rather than using markdown here, what about explicitly using `text` for the content and `url` for the URL? 

We originally were avoiding using the markdown conversion for the content that would be coming from the user's site (it was restricted to just the descriptions in the config, helpText, etc), so I don't know if we want to open that door now. If we do go the markdown route, we might as well add the functionality the `text` type and call it a day instead :)

tkadlec: Just want to clarify before proceeding....

You want something like: `_renderLink(text, url)` ?

patrickhulce: well we really just want this to be in details, the variable you've named `text` here is really `details` so it would be something like

```
_renderLink(details) {
  const snippet = details.text || '';
  const url = details.url;
  ...
}
```

Usage: `detailsRenderer.render({type: 'link', text: 'Learn more', url: 'https://developers.google.com/web/...'})`

tkadlec: Ah, gotcha. Ok...that makes sense.

brendankenny:
@tkadlec if you want we can just use #3165 which does what you two are discussing here, plus it cuts off additional rounds of dealing with Closure so you can get back to the important part of this work :)

tkadlec:
@brendankenny Works for me! I'll wait eagerly for it to be merged, then pull it into #2372

paulirish:
it was now merged. :)

PR #3162, opened 2017-08-28
feat(predictive-perf): add CPU estimation

patrickhulce:
in the name of fast mode #2703 

broken out just the CPU estimation portion from #2720

### DIFF
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -15,6 +15,9 @@ const Node = require('../gather/computed/dependency-graph/node.js');
 const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
 const SCORING_MEDIAN = 10000;
 
+// Any CPU task of 20 ms or more will end up being a critical long task on mobile
+const CRITICAL_LONG_TASK_THRESHOLD = 20;
+
 class PredictivePerf extends Audit {
   /**
    * @return {!AuditMeta}
@@ -40,8 +43,7 @@ class PredictivePerf extends Audit {
     const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
     return dependencyGraph.cloneWithRelationships(node => {
       if (node.endTime > fmp) return false;
-      if (node.type !== Node.TYPES.NETWORK) return true;
-      return node.record.priority() === 'VeryHigh'; // proxy for render-blocking
+      return node.isRenderBlocking();
     });
   }
 
@@ -53,7 +55,11 @@ class PredictivePerf extends Audit {
   static getPessimisticFMPGraph(dependencyGraph, traceOfTab) {
     const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
     return dependencyGraph.cloneWithRelationships(node => {
-      return node.endTime <= fmp;
+      if (node.endTime > fmp) return false;
+      // Include CPU tasks that performed a layout
+      if (node.type === Node.TYPES.CPU) return node.didPerformLayout();
+      // Include every request before FMP that wasn't an image
+      return node.resourceType !== 'image';
     });
   }
 
@@ -62,8 +68,14 @@ class PredictivePerf extends Audit {
    * @return {!Node}
    */
   static getOptimisticTTCIGraph(dependencyGraph) {
+    // Adjust the critical long task threshold for microseconds
+    const minimumCpuTaskDuration = CRITICAL_LONG_TASK_THRESHOLD * 1000;
+
     return dependencyGraph.cloneWithRelationships(node => {
-      return node.record._resourceType && node.record._resourceType._name === 'script' ||
+      // Include everything that might be a long task
+      if (node.type === Node.TYPES.CPU) return node.event.dur > minimumCpuTaskDuration;
+      // Include all scripts and high priority requests
+      return node.resourceType === 'script' ||
           node.record.priority() === 'High' ||
           node.record.priority() === 'VeryHigh';
     });
@@ -75,6 +87,18 @@ class PredictivePerf extends Audit {
    */
   static getPessimisticTTCIGraph(dependencyGraph) {
     return dependencyGraph;
+  }
+
+  /**
+   * @param {!Map<!Node, {startTime, endTime}>} nodeTiming
+   * @return {number}
+   */
+  static getLastLongTaskEndTime(nodeTiming) {
+    return Array.from(nodeTiming.entries())
+        .filter(([node, timing]) => node.type === Node.TYPES.CPU &&
+            timing.endTime - timing.startTime > 50)
+        .map(([_, timing]) => timing.endTime)
+        .reduce((max, x) => Math.max(max, x), 0);
   }
 
   /**
@@ -98,7 +122,22 @@ class PredictivePerf extends Audit {
       let sum = 0;
       const values = {};
       Object.keys(graphs).forEach(key => {
-        values[key] = PageDependencyGraph.computeGraphDuration(graphs[key]);
+        const estimate = PageDependencyGraph.estimateGraph(graphs[key]);
+        const lastLongTaskEnd = PredictivePerf.getLastLongTaskEndTime(estimate.nodeTiming);
+
+        switch (key) {
+          case 'optimisticFMP':
+          case 'pessimisticFMP':
+            values[key] = estimate.timeInMs;
+            break;
+          case 'optimisticTTCI':
+            values[key] = Math.max(values.optimisticFMP, lastLongTaskEnd);
+            break;
+          case 'pessimisticTTCI':
+            values[key] = Math.max(values.pessimisticFMP, lastLongTaskEnd);
+            break;
+        }
+
         sum += values[key];
       });
 
--- a/lighthouse-core/gather/computed/dependency-graph/cpu-node.js
+++ b/lighthouse-core/gather/computed/dependency-graph/cpu-node.js
@@ -7,64 +7,65 @@
 
 const Node = require('./node');
 
-class NetworkNode extends Node {
+class CPUNode extends Node {
   /**
-   * @param {!WebInspector.NetworkRequest} networkRecord
+   * @param {!TraceEvent} parentEvent
+   * @param {!Array<TraceEvent>=} children
    */
-  constructor(networkRecord) {
-    super(networkRecord.requestId);
-    this._record = networkRecord;
+  constructor(parentEvent, children = []) {
+    super(`${parentEvent.tid}.${parentEvent.ts}`);
+    this._event = parentEvent;
+    this._children = children;
   }
 
   /**
    * @return {string}
    */
   get type() {
-    return Node.TYPES.NETWORK;
+    return Node.TYPES.CPU;
   }
 
   /**
    * @return {number}
    */
   get startTime() {
-    return this._record.startTime * 1000 * 1000;
+    return this._event.ts;
   }
 
   /**
    * @return {number}
    */
   get endTime() {
-    return this._record.endTime * 1000 * 1000;
+    return this._event.ts + this._event.dur;
   }
 
   /**
-   * @return {!WebInspector.NetworkRequest}
+   * @return {!TraceEvent}
    */
-  get record() {
-    return this._record;
+  get event() {
+    return this._event;
   }
 
   /**
-   * @return {string}
+   * @return {!TraceEvent}
    */
-  get resourceType() {
-    return this._record._resourceType && this._record._resourceType._name;
+  get children() {
+    return this._children;
   }
 
   /**
    * @return {boolean}
    */
-  isRenderBlocking() {
-    const priority = this._record.priority();
-    return priority === 'VeryHigh' || (priority === 'High' && this.resourceType === 'script');
+  didPerformLayout() {
+    return this._children.some(evt => evt.name === 'Layout');
   }
 
   /**
-   * @return {!NetworkNode}
+   * @return {!CPUNode}
    */
   cloneWithoutRelationships() {
-    return new NetworkNode(this._record);
+    return new CPUNode(this._event, this._children);
   }
 }
 
-module.exports = NetworkNode;
+module.exports = CPUNode;
--- a/lighthouse-core/gather/computed/dependency-graph/cpu-node.js
+++ b/lighthouse-core/gather/computed/dependency-graph/cpu-node.js
@@ -10,12 +10,14 @@ const Node = require('./node');
 class CPUNode extends Node {
   /**
    * @param {!TraceEvent} parentEvent
-   * @param {!Array<TraceEvent>=} children
+   * @param {!Array<TraceEvent>=} childEvents
    */
-  constructor(parentEvent, children = []) {
-    super(`${parentEvent.tid}.${parentEvent.ts}`);
+  constructor(parentEvent, childEvents = []) {
+    const nodeId = `${parentEvent.tid}.${parentEvent.ts}`;
+    super(nodeId);
+
     this._event = parentEvent;
-    this._children = children;
+    this._childEvents = childEvents;
   }
 
   /**
@@ -49,22 +51,22 @@ class CPUNode extends Node {
   /**
    * @return {!TraceEvent}
    */
-  get children() {
-    return this._children;
+  get childEvents() {
+    return this._childEvents;
   }
 
   /**
    * @return {boolean}
    */
   didPerformLayout() {
-    return this._children.some(evt => evt.name === 'Layout');
+    return this._childEvents.some(evt => evt.name === 'Layout');
   }
 
   /**
    * @return {!CPUNode}
    */
   cloneWithoutRelationships() {
-    return new CPUNode(this._event, this._children);
+    return new CPUNode(this._event, this._childEvents);
   }
 }
 
--- a/lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
+++ b/lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
@@ -14,6 +14,7 @@ const DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 10;
 const DEFAULT_FALLBACK_TTFB = 30;
 const DEFAULT_RTT = emulation.TYPICAL_MOBILE_THROTTLING_METRICS.targetLatency;
 const DEFAULT_THROUGHPUT = emulation.TYPICAL_MOBILE_THROTTLING_METRICS.targetDownloadThroughput * 8;
+const DEFAULT_CPU_MULTIPLIER = 5;
 
 const TLS_SCHEMES = ['https', 'wss'];
 
@@ -43,6 +44,7 @@ class Estimator {
         throughput: DEFAULT_THROUGHPUT,
         fallbackTTFB: DEFAULT_FALLBACK_TTFB,
         maximumConcurrentRequests: DEFAULT_MAXIMUM_CONCURRENT_REQUESTS,
+        cpuMultiplier: DEFAULT_CPU_MULTIPLIER,
       },
       options
     );
@@ -54,6 +56,7 @@ class Estimator {
       TcpConnection.maximumSaturatedConnections(this._rtt, this._throughput),
       this._options.maximumConcurrentRequests
     );
+    this._cpuMultiplier = this._options.cpuMultiplier;
   }
 
   /**
@@ -84,10 +87,7 @@ class Estimator {
    */
   _initializeNetworkConnections() {
     const connections = new Map();
-    const recordsByConnection = groupBy(
-      this._networkRecords,
-      record => record.connectionId
-    );
+    const recordsByConnection = groupBy(this._networkRecords, record => record.connectionId);
 
     for (const [connectionId, records] of recordsByConnection.entries()) {
       const isTLS = TLS_SCHEMES.includes(records[0].parsedURL.scheme);
@@ -124,22 +124,71 @@ class Estimator {
    */
   _initializeAuxiliaryData() {
     this._nodeTiming = new Map();
+    this._nodesUnprocessed = new Set();
     this._nodesCompleted = new Set();
     this._nodesInProgress = new Set();
     this._nodesInQueue = new Set(); // TODO: replace this with priority queue
     this._connectionsInUse = new Set();
+    this._numberInProgressByType = new Map();
+  }
+
+  /**
+   * @param {string} type
+   * @return {number}
+   */
+  _numberInProgress(type) {
+    return this._numberInProgressByType.get(type) || 0;
   }
 
   /**
    * @param {!Node} node
+   * @param {!Object} values
    */
-  _enqueueNodeIfPossible(node) {
-    const dependencies = node.getDependencies();
-    if (
-      !this._nodesCompleted.has(node) &&
-      dependencies.every(dependency => this._nodesCompleted.has(dependency))
-    ) {
-      this._nodesInQueue.add(node);
+  _setTimingData(node, values) {
+    const timingData = this._nodeTiming.get(node) || {};
+    Object.assign(timingData, values);
+    this._nodeTiming.set(node, timingData);
+  }
+
+  /**
+   * @param {!Node} node
+   * @param {number} queuedTime
+   */
+  _markNodeAsInQueue(node, queuedTime) {
+    this._nodesInQueue.add(node);
+    this._nodesUnprocessed.delete(node);
+    this._setTimingData(node, {queuedTime});
+  }
+
+  /**
+   * @param {!Node} node
+   * @param {number} startTime
+   */
+  _markNodeAsInProgress(node, startTime) {
+    this._nodesInQueue.delete(node);
+    this._nodesInProgress.add(node);
+    this._numberInProgressByType.set(node.type, this._numberInProgress(node.type) + 1);
+    this._setTimingData(node, {startTime});
+  }
+
+  /**
+   * @param {!Node} node
+   * @param {number} endTime
+   */
+  _markNodeAsComplete(node, endTime) {
+    this._nodesCompleted.add(node);
+    this._nodesInProgress.delete(node);
+    this._numberInProgressByType.set(node.type, this._numberInProgress(node.type) - 1);
+    this._setTimingData(node, {endTime});
+
+    // Try to add all its dependents to the queue
+    for (const dependent of node.getDependents()) {
+      // Skip this node if one of its dependencies hasn't finished yet
+      const dependencies = dependent.getDependencies();
+      if (dependencies.some(dependency => !this._nodesCompleted.has(dependency))) continue;
+
+      // Otherwise add it to the queue
+      this._markNodeAsInQueue(dependent, endTime);
     }
   }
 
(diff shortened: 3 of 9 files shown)


lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,202 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+          case 'TimerFire': {
+            const installer = timers.get(evt.args.data.timerId);
+            if (!installer) break;
+            installer.addDependent(node);
+            break;
+          }
+
+          case 'InvalidateLayout':
+          case 'ScheduleStyleRecalculation':
+            stackTraceUrls.forEach(url => addDependencyOnUrl(node, url));
```
paulirish: does it makes sense to generally do this to all events that have a stack trace?

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,202 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+      if (!minCandidate) return;
+      cpuNode.addDependency(minCandidate);
+    }
+
+    const timers = new Map();
+    for (const node of cpuNodes) {
+      for (const evt of node.children) {
+        if (!evt.args.data) continue;
+
+        const url = evt.args.data.url;
```
paulirish: seems like it's possible for this to be undefined in some cases? (does `ScheduleStyleRecalculation` always a url? or EvaluateScript for eval()?)
i guess you could early return in `addDependencyOnUrl` to avoid some work?

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,202 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+        const stackTraceUrls = (evt.args.data.stackTrace || []).map(l => l.url).filter(Boolean);
+
+        switch (evt.name) {
```
paulirish: A few other events that mighttttttttt be worthwhile?  


actually here's a list of main thread trace events with a url in the data:

       1 "name":"CommitLoad"
       8 "name":"EvaluateScript"
      26 "name":"FunctionCall"
       1 "name":"InvalidateLayout"
       2 "name":"Layout"
     179 "name":"PaintImage"
       7 "name":"ParseHTML"
       1 "name":"RenderFrameImpl::didStartProvisionalLoad"
      61 "name":"ResourceSendRequest"
       3 "name":"ScheduleStyleRecalculation"
      12 "name":"TimerInstall"
       7 "name":"TimerRemove"
       1 "name":"TracingStartedInPage"
       2 "name":"UpdateLayoutTree"
       1 "name":"XHRLoad"
       4 "name":"XHRReadyStateChange"
       9 "name":"v8.compile"

    # via: ag '61815.*?775.*?"name":".*?".*?"url":".*?"' news.google.com_2017-09-01_18-14-51-0.trace.json -o  | ag -o '"name":".*?"' | sort | uniq -c


* `ParseHTML` seems good
* `PaintImage` seems good?
* `UpdateLayoutTree` sometimes has a stack trace, which indicates a forced layout. Not sure if the semantics of that matches...
* DecodeImage and ResizeImage i think happen off the main thread..

lighthouse-core/audits/predictive-perf.js
```diff
@@ -51,7 +53,11 @@ class PredictivePerf extends Audit {
   static getPessimisticFMPGraph(graph, traceOfTab) {
     const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
     return graph.cloneWithRelationships(node => {
-      return node.endTime <= fmp;
+      if (node.endTime > fmp) return false;
+      // Include CPU tasks that performed a layout
+      if (node.type === Node.TYPES.CPU) return node.didPerformLayout();
```
paulirish: btw what's special about these?

lighthouse-core/audits/predictive-perf.js
```diff
@@ -15,6 +15,9 @@ const Node = require('../gather/computed/dependency-graph/node.js');
 const SCORING_POINT_OF_DIMINISHING_RETURNS = 1700;
 const SCORING_MEDIAN = 10000;
 
+// Any CPU task of 20 ms or more will end up being a critical long task on mobile
+const CRITICAL_LONG_TASK_THRESHOLD = 20000;
```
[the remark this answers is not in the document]
patrickhulce: sure thing, done

lighthouse-core/audits/predictive-perf.js
patrickhulce: they're included in pessimistic estimate because they frequently capture large CPU work that was needed to render something meaningful and the resources that might not have been render-blocking that were required to perform it. example: a client-side rendered page with a large script in the body

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -13,6 +13,7 @@ const DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 10;
 const DEFAULT_RESPONSE_TIME = 30;
 const DEFAULT_RTT = 150;
 const DEFAULT_THROUGHPUT = 1600 * 1024; // 1.6 Mbps
+const DEFAULT_CPU_MULTIPLIER = 5;
```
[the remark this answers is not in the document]
patrickhulce: it was the multiplier from emulation adjusted upwards to compensate for not including all relevant tasks, but it's moved down closer to the emulation value in #3163

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: yep will do

lighthouse-core/gather/computed/dependency-graph/network-node.js
```diff
@@ -45,6 +45,21 @@ class NetworkNode extends Node {
   }
 
   /**
+   * @return {string}
+   */
+  get resourceType() {
+    return this._record._resourceType && this._record._resourceType._name;
```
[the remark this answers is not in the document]
patrickhulce: yep will do

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: It does, but I wanted to explicitly call out which events we're relying on and do some experimentation as to which are the most informative for when we reduce the set. 

tl;dr 90% of the value comes from EvaluateScript, XHRReadyStateChange, and ResourceSendRequest

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: I'm hesitant to add reliance on PaintImage since it seems to have a larger "unthrottled view obfuscates relationships" effect, i.e. the most common reason PaintImage is included in this task is because the image happened to finish downloading then not because this task necessarily *relied* on the image being available.

UpdateLayoutTree, ParseHTML I can add 👍

lighthouse-core/audits/predictive-perf.js
paulirish: gotcha.. so i guess this is more semantically like `node.containsMeaningfulWork()` 

.. and in which case i'm wondering if there are any other work payloads besides layout that also fit into the definition. Style/Paint probably do, but I can imagine including them wouldn't change the results so much...

lighthouse-core/gather/computed/page-dependency-graph.js
paulirish: sg. what about a `default` case for the switch statement?

lighthouse-core/gather/computed/page-dependency-graph.js
paulirish: SGTM. as discussed offline, right now the graph is identifying which work is DEFINITELY required before each node.  Images don't really fall into this definition. 
Later work could introduce a pessimistic graph where relationships like these images being painted are included. But i agree it's unclear that would add value.

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: I'm on board for #3163 once I measure impact :)

patrickhulce:
FYI it's rebased and easier to read now :)

lighthouse-core/gather/computed/page-dependency-graph.js
paulirish: k. i think there are ~3 we agreed to punt to 3163.. i forget what they were tho.. hopefully you got a hold of em. :)

patrickhulce: 1. Refactor the signature of simulateDownloadUntil to accept an object
2. Add a _nodeState Node -> ProgressEnum map to track the nodes through estimator
3. The resourceType business
4. Network initiators improvements

paulirish: hero. 👍  💯

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+    return nodes;
+  }
+
+  /**
+   * @param {!Node} rootNode
+   * @param {!NetworkNodeOutput} networkNodeOutput
+   */
+  static linkNetworkNodes(rootNode, networkNodeOutput) {
+    networkNodeOutput.nodes.forEach(node => {
+      const initiators = PageDependencyGraphArtifact.getNetworkInitiators(node.record);
```
paulirish: so there is a case where these initiators kind of lie to you, as far as dependencies are concerned.

example:  
* a page has a stylesheet. in that stylesheet are background images and webfonts that will be required.
* the styles have not be calculated yet, but they are downloaded. (yes, chrome should probably try to recalc style/scan/preload these resources immediately after download, but it doesn't)
* an unrelated script on the page runs and forces layout. style recalc is forced because layout.
* recalc runs and the requests for images and fonts begin.
* their initiators have a stack trace including this unrelated script (and whatever other unrelated scripts are in the call stack)

unfortunately this happens pretty often.

----------

to avoid misattributing the dependency i think we'd want to exclude a stack trace if the request began during a StyleRecalc main thread event.  :)

the good news is brendan and i couldn't think of another case where the call stack shouldn't be considered dependencies.

lighthouse-core/audits/predictive-perf.js
```diff
@@ -98,7 +122,22 @@ class PredictivePerf extends Audit {
       let sum = 0;
       const values = {};
       Object.keys(graphs).forEach(key => {
-        values[key] = PageDependencyGraph.computeGraphDuration(graphs[key]);
+        const estimate = PageDependencyGraph.estimateGraph(graphs[key]);
+        const lastLongTaskEnd = PredictivePerf.getLastLongTaskEndTime(estimate.nodeTiming);
+
+        switch (key) {
```
brendankenny: nit: I would personally prefer `if`/`else if` here since it's not really a loop (though really the `Object.keys` may not be needed at all since it's four fixed, known cases that have to be enumerated above and below anyways :P)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -254,22 +321,28 @@ class Estimator {
 
   /**
    * Estimates the time taken to process all of the graph's nodes.
-   * @return {number}
+   * @return {{timeInMs: number, nodeTiming: !Map<!Node, Object>}}
```
brendankenny: `Object`?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -254,22 +321,28 @@ class Estimator {
...
+    const rootNode = this._graph.getRootNode();
+    rootNode.traverse(node => nodesUnprocessed.add(node));
 
     let depth = 0;
     let totalElapsedTime = 0;
+
+    // add root node to queue
+    this._markNodeAsInQueue(rootNode, totalElapsedTime);
+
+    // loop as long as we have nodes in the queue or currently in process
```
brendankenny: progress :)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -148,21 +197,31 @@ class Estimator {
    * @param {number} totalElapsedTime
    */
   _startNodeIfPossible(node, totalElapsedTime) {
-    if (node.type !== Node.TYPES.NETWORK) return;
+    if (node.type === Node.TYPES.CPU) {
+      // Start a CPU task if there's no other CPU task in process
```
brendankenny: for clarity, does this mean for now e.g. all timer callbacks will be run as soon as possible as their dependency URLs are loaded, regardless of when they're actually scheduled? along with other tasks scheduled by chrome itself

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -221,9 +288,17 @@ class Estimator {
    * @param {number} totalElapsedTime
    */
   _updateProgressMadeInTimePeriod(node, timePeriodLength, totalElapsedTime) {
+    const timingData = this._nodeTiming.get(node);
+    const isFinished = timingData.estimatedTimeElapsed === timePeriodLength;
```
brendankenny: feels like one and only one node should be finishing per `estimateWithDetails` loop to make reasoning about progress easier (if thinking of it as a sweep line), but you may be thinking differently

one option: `_findNextNodeCompletionTime` could be `_findNextCompletedNode`, and `_updateProgressMadeInTimePeriod` could finish only that one and use its timing for `totalElapsedTime`

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -124,22 +124,71 @@ class Estimator {
...
+   */
+  _markNodeAsComplete(node, endTime) {
+    this._nodesCompleted.add(node);
+    this._nodesInProgress.delete(node);
+    this._numberInProgressByType.set(node.type, this._numberInProgress(node.type) - 1);
+    this._setTimingData(node, {endTime});
+
+    // Try to add all its dependents to the queue
+    for (const dependent of node.getDependents()) {
+      // Skip this node if one of its dependencies hasn't finished yet
```
brendankenny: `// Skip dependent node if one...`

brendankenny:
The result of things seems like a mixture of parallel data structures (e.g. nodes and their timing information). Would it makes sense to have timing information right on the nodes? And then the different estimated dependency graphs would be clones with their own timing information from each parallel reality of different network/cpu conditions.

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: I'm a little lost on which initiators you're talking about here, so you're saying that image and font network records have in their initiator information a stack trace that's not relevant and so I should cross-reference that against CPU task information to see if we're in a stylerecalc? This feels like the initiator should be fixed instead though because the initiator isn't the script it's the stylesheet, right?

lighthouse-core/gather/computed/dependency-graph/cpu-node.js
```diff
@@ -0,0 +1,71 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Node = require('./node');
+
+class CPUNode extends Node {
+  /**
+   * @param {!TraceEvent} parentEvent
+   * @param {!Array<TraceEvent>=} children
```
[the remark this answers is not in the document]
patrickhulce: `childEvents` sg 👍

lighthouse-core/gather/computed/dependency-graph/cpu-node.js
```diff
@@ -0,0 +1,71 @@
...
+   */
+  constructor(parentEvent, children = []) {
+    super(`${parentEvent.tid}.${parentEvent.ts}`);
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -124,22 +124,71 @@ class Estimator {
...
+   * @param {string} type
+   * @return {number}
+   */
+  _numberInProgress(type) {
+    return this._numberInProgressByType.get(type) || 0;
   }
 
   /**
    * @param {!Node} node
+   * @param {!Object} values
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: correct, timer callbacks are available to be processed as soon as their dependencies have been met (including their installing task), opportunity for future improvement could be to require a minimum time between the installer and the callback

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: because the network logic happens in essentially round trip blocks rather than on a packet simulation level, it's possible to have multiple requests finishing at exactly the same time and get updated all at once and this simplified things to not have to worry about 0 elapsed time

if you think that's confusing though I'm open to a future refactor since that would line up nicely with information stored on each node

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: fixed

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: heh done :)

lighthouse-core/gather/computed/page-dependency-graph.js
paulirish: >  image and font network records have in their initiator information a stack trace that's not relevant

yes correct

>  I should cross-reference that against CPU task information to see if we're in a stylerecalc

yah that's the solution that comes to mind.

> This feels like the initiator should be fixed instead though because the initiator isn't the script it's the stylesheet, right?

If we want to be really specific there are two _initiator-type_ like ideas:  1) **the URL source resource** (bad name sorry). There is a specific resource that contains that URL (like a stylesheet). Obviously a request can't be made to it without this file being present. 2) **the initiating task.** - the main thread activity that caused the browser to request this file. There is sometimes a JS callstack associated with this, but other times it's an async task like the browser just catching up to styles it hasn't computed yet.

Oftentimes 1 and 2 point to the same url. (a script requesting another script). But this forced layout case is the primary example of when they diverge.  IMO developers want both pieces of information. 

I'd tweak your statement to say the initiator should be fixed via breaking into the two components. Then we could clearly ask for the URL source resource and it'd give us a stylesheet rather than whatever JS forced a stylerecalc to happen.


starbucks.com is an example of this behavior. see the first google webfont's initiator is a line of JS that calls `offsetWidth`. 

screen shot 2017-09-12 at 10 37 44 am

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -7,7 +7,11 @@
 
 const ComputedArtifact = require('./computed-artifact');
 const NetworkNode = require('./dependency-graph/network-node');
+const CPUNode = require('./dependency-graph/cpu-node');
 const GraphEstimator = require('./dependency-graph/estimator/estimator');
+const TracingProcessor = require('../../lib/traces/tracing-processor');
+
+const MINIMUM_TASK_DURATION = 10 * 1000;
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
   }
 
   /**
-   * @param {!TraceOfTabArtifact} traceOfTab
    * @param {!Array<!WebInspector.NetworkRequest>} networkRecords
-   * @return {!Node}
+   * @return {!NetworkNodeOutput}
    */
-  static createGraph(traceOfTab, networkRecords) {
+  static getNetworkNodes(networkRecords) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
     networkRecords.forEach(record => {
       const node = new NetworkNode(record);
+      nodes.push(node);
       idToNodeMap.set(record.requestId, node);
 
       if (urlToNodeMap.has(record.url)) {
-        // If duplicate requests have been made to this URL we can't be certain which node is being
-        // referenced, so act like we don't know the URL at all.
-        urlToNodeMap.set(record.url, undefined);
+        const list = urlToNodeMap.get(record.url);
```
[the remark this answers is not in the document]
patrickhulce: heh this evolved over time to eventually become replaceable by that :)

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+  static linkNetworkNodes(rootNode, networkNodeOutput) {
+    networkNodeOutput.nodes.forEach(node => {
+      const initiators = PageDependencyGraphArtifact.getNetworkInitiators(node.record);
       if (initiators.length) {
         initiators.forEach(initiator => {
-          const parent = urlToNodeMap.get(initiator) || rootNode;
+          const parentCandidates = networkNodeOutput.urlToNodeMap.get(initiator) || [rootNode];
+          // Only add the initiator relationship if the initiator is unambiguous
+          const parent = parentCandidates.length === 1 ? parentCandidates[0] : rootNode;
           parent.addDependent(node);
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
         rootNode.addDependent(node);
       }
     });
+  }
+
+  /**
+   * @param {!Node} rootNode
+   * @param {!NetworkNodeOutput} networkNodeOutput
+   * @param {!Array<!CPUNode>} cpuNodes
+   * @param {!TraceOfTabArtifact} traceOfTab
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+   */
+  static linkCPUNodes(rootNode, networkNodeOutput, cpuNodes) {
+    function addDependentNetworkRequest(cpuNode, reqId) {
```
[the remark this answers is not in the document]
patrickhulce: sure, I'm open to clearer names though the larger issue for me has just been spelling dependent not keeping them straight 😆

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+      const networkNode = networkNodeOutput.idToNodeMap.get(reqId);
+      if (!networkNode || networkNode.resourceType !== 'xhr') return;
+      networkNode.addDependency(cpuNode);
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+    }
+  }
+
+  /**
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @param {!Array<!WebInspector.NetworkRequest>} networkRecords
+   * @return {!Node}
+   */
+  static createGraph(traceOfTab, networkRecords) {
+    const networkNodeOutput = this.getNetworkNodes(networkRecords);
```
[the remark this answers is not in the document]
patrickhulce: `Estimator` stores and tracks lots of state but these do not :), done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -34,52 +38,203 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+    this.linkNetworkNodes(rootNode, networkNodeOutput, networkRecords);
+    this.linkCPUNodes(rootNode, networkNodeOutput, cpuNodes, traceOfTab);
 
     return rootNode;
   }
 
   /**
    * @param {!Node} rootNode
-   * @return {number}
+   * @return {}
```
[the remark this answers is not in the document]
patrickhulce: yeah along with the other host of refactors for post #3163, estimator is just wrong at this point, simulator makes a lot more sense based on what it's evolved into since the time is easily achieved from the node information

lighthouse-core/gather/computed/dependency-graph/node.js
```diff
@@ -109,6 +121,14 @@ class Node {
   }
 
   /**
+   * @param {!Node}
+   * @return {boolean}
+   */
+  hasDependency(node) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/audits/predictive-perf.js
patrickhulce: yeah but those duplicated 3 lines at the top really look pretty bad 😆

patrickhulce:
> The result of things seems like a mixture of parallel data structures (e.g. nodes and their timing information). Would it makes sense to have timing information right on the nodes? And then the different estimated dependency graphs would be clones with their own timing information from each parallel reality of different network/cpu conditions.

Yeah Paul and I talked this over a bit in the network PR. The primary reason for avoiding attaching the information to the node was that the timing data is an artifact of the specific simulation and not an intrinsic property of the graph which is meant to reflect the actual page load that was observed. The practicality of throwing the info onto the node though is pretty compelling, so maintaining that distinction with a `get/setSimulatedData` might be worth it in a refactor PR

lighthouse-core/gather/computed/page-dependency-graph.js
patrickhulce: Ahhhhhh I see, gotcha. Yeah then let's dig into this with the rest of the pending initiator improvements for #3163

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
brendankenny: > if you think that's confusing though I'm open to a future refactor since that would line up nicely with information stored on each node

yeah, I think it would generally be easier to reason about if there was only one node finishing at any one point (where that point is defined by actual time + some ordering of the nodes)

PR #3165, opened 2017-08-29
report: add LinkDetailsJSON to details renderer

brendankenny:
alternative to #3154

this does unfortunately leave us with `renderUrl` and `renderLink`, but they were different enough that I didn't really want to merge them

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-/* globals self CriticalRequestChainRenderer Util */
+/* globals self CriticalRequestChainRenderer Util URL */
 
 class DetailsRenderer {
   /**
@@ -35,6 +35,8 @@ class DetailsRenderer {
         return this._renderText(details);
       case 'url':
         return this._renderURL(details);
+      case 'link':
+        return this._renderLink(/** @type {!DetailsRenderer.LinkDetailsJSON} */ (details));
       case 'thumbnail':
         return this._renderThumbnail(/** @type {!DetailsRenderer.ThumbnailDetails} */ (details));
       case 'filmstrip':
@@ -87,6 +89,27 @@ class DetailsRenderer {
     }
 
     return element;
+  }
+
+  /**
+   * @param {!DetailsRenderer.LinkDetailsJSON} details
+   * @return {!Element}
+   */
+  _renderLink(details) {
+    const allowedProtocols = ['https:', 'http:'];
+    const url = new URL(details.url);
+    if (!allowedProtocols.includes(url.protocol)) {
+      // Fall back to text if protocol not allowed.
+      return this._renderText(details);
+    }
+
+    const a = /** @type {!HTMLAnchorElement} */ (this._dom.createElement('a'));
+    a.rel = 'noopener';
+    a.target = '_blank';
+    a.textContent = details.text;
+    a.href = url.href;
+
+    return a;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -90,6 +92,27 @@ class DetailsRenderer {
   }
 
   /**
+   * @param {!DetailsRenderer.LinkDetailsJSON} details
+   * @return {!Element}
+   */
+  _renderLink(details) {
+    const allowedProtocols = ['https:', 'http:'];
```
brendankenny: only allows `http` and `https` links, like [`dom. convertMarkdownLinkSnippets `](https://github.com/GoogleChrome/lighthouse/blob/8c3c3b50ae9bf2d83dcc797d0e5063768f6157aa/lighthouse-core/report/v2/renderer/dom.js#L98) does already.

If we need more supported protocols we can add them in the future, but better to err on the side of caution as these links will be coming from things like the hard-to-inspect snyk blob in #2372

paulirish: in `is-on-https` audit we have a `SECURE_SCHEMES` array that i'd love to use and then add 'http' to, as our starting point. :)

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -90,6 +92,27 @@ class DetailsRenderer {
   }
 
   /**
+   * @param {!DetailsRenderer.LinkDetailsJSON} details
+   * @return {!Element}
+   */
+  _renderLink(details) {
```
paulirish: let's call `_renderURL` `_renderURLText` instead to clarify it's never linkified.

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -90,6 +92,27 @@ class DetailsRenderer {
...
+    const allowedProtocols = ['https:', 'http:'];
+    const url = new URL(details.url);
+    if (!allowedProtocols.includes(url.protocol)) {
```
patrickhulce: can we do startsWith if we're tryin' to be secure here? otherwise people can just throw http: into their malicious URL ;)

lighthouse-core/report/v2/renderer/details-renderer.js
patrickhulce: I'm not necessarily sold on using secure schemes here, just because it's transported securely doesn't mean it won't be malicious :) but seems like you've already highlighted that haha

lighthouse-core/report/v2/renderer/details-renderer.js
paulirish: it is only checking the `.protocol` property, but yeah startsWith sgtm.. might as well just do `===`

paulirish: wait no.. we're doing `array.includes(string)`.. this uses a `===` under the hood so we're good here.

patrickhulce: oh whoops yeah I misread as `url.includes(allowedProcotol)` haha, nevermind carryon :)

lighthouse-core/report/v2/renderer/details-renderer.js
brendankenny: > let's call _renderURL _renderURLText instead to clarify it's never linkified.

done

lighthouse-core/report/v2/renderer/details-renderer.js
brendankenny: there may come a time when the report needs to link to a web socket resource, but it's not today :P blob/data URLs are more plausible, but nothing needs them yet.

My vote is for adding what we need now (linking to snyk vulnerability reports) and add others as they become needed (they haven't been needed yet by `convertMarkdownLinkSnippets`, for instance).

PR #3187, opened 2017-08-30
feat(predictive-perf): add network estimation

patrickhulce:
~blocked on #2720~

split out from #2720 for easier reviewing

### DIFF
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -102,7 +102,7 @@ class PredictivePerf extends Audit {
         sum += values[key];
       });
 
-      const rawValue = sum / 4;
+      const rawValue = sum / Object.keys(values).length;
       const score = Audit.computeLogNormalScore(
         rawValue,
         SCORING_POINT_OF_DIMINISHING_RETURNS,
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -32,13 +32,13 @@ class PredictivePerf extends Audit {
   }
 
   /**
-   * @param {!Node} graph
+   * @param {!Node} dependencyGraph
    * @param {!TraceOfTabArtifact} traceOfTab
    * @return {!Node}
    */
-  static getOptimisticFMPGraph(graph, traceOfTab) {
+  static getOptimisticFMPGraph(dependencyGraph, traceOfTab) {
     const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
-    return graph.cloneWithRelationships(node => {
+    return dependencyGraph.cloneWithRelationships(node => {
       if (node.endTime > fmp) return false;
       if (node.type !== Node.TYPES.NETWORK) return true;
       return node.record.priority() === 'VeryHigh'; // proxy for render-blocking
@@ -46,23 +46,23 @@ class PredictivePerf extends Audit {
   }
 
   /**
-   * @param {!Node} graph
+   * @param {!Node} dependencyGraph
    * @param {!TraceOfTabArtifact} traceOfTab
    * @return {!Node}
    */
-  static getPessimisticFMPGraph(graph, traceOfTab) {
+  static getPessimisticFMPGraph(dependencyGraph, traceOfTab) {
     const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
-    return graph.cloneWithRelationships(node => {
+    return dependencyGraph.cloneWithRelationships(node => {
       return node.endTime <= fmp;
     });
   }
 
   /**
-   * @param {!Node} graph
+   * @param {!Node} dependencyGraph
    * @return {!Node}
    */
-  static getOptimisticTTCIGraph(graph) {
-    return graph.cloneWithRelationships(node => {
+  static getOptimisticTTCIGraph(dependencyGraph) {
+    return dependencyGraph.cloneWithRelationships(node => {
       return node.record._resourceType && node.record._resourceType._name === 'script' ||
           node.record.priority() === 'High' ||
           node.record.priority() === 'VeryHigh';
@@ -70,11 +70,11 @@ class PredictivePerf extends Audit {
   }
 
   /**
-   * @param {!Node} graph
+   * @param {!Node} dependencyGraph
    * @return {!Node}
    */
-  static getPessimisticTTCIGraph(graph) {
-    return graph;
+  static getPessimisticTTCIGraph(dependencyGraph) {
+    return dependencyGraph;
   }
 
   /**
@@ -102,17 +102,17 @@ class PredictivePerf extends Audit {
         sum += values[key];
       });
 
-      const rawValue = sum / Object.keys(values).length;
+      const meanDuration = sum / Object.keys(values).length;
       const score = Audit.computeLogNormalScore(
-        rawValue,
+        meanDuration,
         SCORING_POINT_OF_DIMINISHING_RETURNS,
         SCORING_MEDIAN
       );
 
       return {
         score,
-        rawValue,
-        displayValue: Util.formatMilliseconds(rawValue),
+        rawValue: meanDuration,
+        displayValue: Util.formatMilliseconds(meanDuration),
         extendedInfo: {value: values},
       };
     });
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -8,6 +8,7 @@
 const Audit = require('./audit');
 const Util = require('../report/v2/renderer/util.js');
 const PageDependencyGraph = require('../gather/computed/page-dependency-graph.js');
+const Node = require('../gather/computed/dependency-graph/node.js');
 
 // Parameters (in ms) for log-normal CDF scoring. To see the curve:
 //   https://www.desmos.com/calculator/rjp0lbit8y
@@ -31,14 +32,77 @@ class PredictivePerf extends Audit {
   }
 
   /**
+   * @param {!Node} graph
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {!Node}
+   */
+  static getOptimisticFMPGraph(graph, traceOfTab) {
+    const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
+    return graph.cloneWithRelationships(node => {
+      if (node.endTime > fmp) return false;
+      if (node.type !== Node.TYPES.NETWORK) return true;
+      return node.record.priority() === 'VeryHigh'; // proxy for render-blocking
+    });
+  }
+
+  /**
+   * @param {!Node} graph
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {!Node}
+   */
+  static getPessimisticFMPGraph(graph, traceOfTab) {
+    const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
+    return graph.cloneWithRelationships(node => {
+      return node.endTime <= fmp;
+    });
+  }
+
+  /**
+   * @param {!Node} graph
+   * @return {!Node}
+   */
+  static getOptimisticTTCIGraph(graph) {
+    return graph.cloneWithRelationships(node => {
+      return node.record._resourceType && node.record._resourceType._name === 'script' ||
+          node.record.priority() === 'High' ||
+          node.record.priority() === 'VeryHigh';
+    });
+  }
+
+  /**
+   * @param {!Node} graph
+   * @return {!Node}
+   */
+  static getPessimisticTTCIGraph(graph) {
+    return graph;
+  }
+
+  /**
    * @param {!Artifacts} artifacts
    * @return {!AuditResult}
    */
   static audit(artifacts) {
     const trace = artifacts.traces[Audit.DEFAULT_PASS];
     const devtoolsLogs = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
-    return artifacts.requestPageDependencyGraph(trace, devtoolsLogs).then(graph => {
-      const rawValue = PageDependencyGraph.computeGraphDuration(graph);
+    return Promise.all([
+      artifacts.requestPageDependencyGraph(trace, devtoolsLogs),
+      artifacts.requestTraceOfTab(trace),
+    ]).then(([graph, traceOfTab]) => {
+      const graphs = {
+        optimisticFMP: PredictivePerf.getOptimisticFMPGraph(graph, traceOfTab),
+        pessimisticFMP: PredictivePerf.getPessimisticFMPGraph(graph, traceOfTab),
+        optimisticTTCI: PredictivePerf.getOptimisticTTCIGraph(graph, traceOfTab),
+        pessimisticTTCI: PredictivePerf.getPessimisticTTCIGraph(graph, traceOfTab),
+      };
+
+      let sum = 0;
+      const values = {};
+      Object.keys(graphs).forEach(key => {
+        values[key] = PageDependencyGraph.computeGraphDuration(graphs[key]);
+        sum += values[key];
+      });
+
+      const rawValue = sum / 4;
       const score = Audit.computeLogNormalScore(
         rawValue,
         SCORING_POINT_OF_DIMINISHING_RETURNS,
@@ -49,6 +113,7 @@ class PredictivePerf extends Audit {
         score,
         rawValue,
         displayValue: Util.formatMilliseconds(rawValue),
+        extendedInfo: {value: values},
       };
     });
   }
--- a/lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
+++ b/lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
@@ -1,0 +1,284 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Node = require('../node');
+const TcpConnection = require('./tcp-connection');
+
+// see https://cs.chromium.org/search/?q=kDefaultMaxNumDelayableRequestsPerClient&sq=package:chromium&type=cs
+const DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 10;
+const DEFAULT_RESPONSE_TIME = 30;
+const DEFAULT_RTT = 150;
+const DEFAULT_THROUGHPUT = 1600 * 1024; // 1.6 Mbps
+
+function groupBy(items, keyFunc) {
+  const grouped = new Map();
+  items.forEach(item => {
+    const key = keyFunc(item);
+    const group = grouped.get(key) || [];
+    group.push(item);
+    grouped.set(key, group);
+  });
+
+  return grouped;
+}
+
+class Estimator {
+  /**
+   * @param {!Node} graph
+   * @param {{rtt: number, throughput: number, defaultResponseTime: number,
+   *    maximumConcurrentRequests: number}=} options
+   */
+  constructor(graph, options) {
+    this._graph = graph;
+    this._options = Object.assign(
+      {
+        rtt: DEFAULT_RTT,
+        throughput: DEFAULT_THROUGHPUT,
+        defaultResponseTime: DEFAULT_RESPONSE_TIME,
+        maximumConcurrentRequests: DEFAULT_MAXIMUM_CONCURRENT_REQUESTS,
+      },
+      options
+    );
+
+    this._rtt = this._options.rtt;
+    this._throughput = this._options.throughput;
+    this._defaultResponseTime = this._options.defaultResponseTime;
+    this._maximumConcurrentRequests = Math.min(
+      TcpConnection.maximumSaturatedConnections(this._rtt, this._throughput),
+      this._options.maximumConcurrentRequests
+    );
+  }
+
+  /**
+   * @param {!WebInspector.NetworkRequest} record
+   * @return {number}
+   */
+  static getResponseTime(record) {
+    const timing = record._timing;
+    return (timing && timing.receiveHeadersEnd - timing.sendEnd) || Infinity;
+  }
+
+  _initializeNetworkRecords() {
+    const records = [];
+
+    this._graph.getRootNode().traverse(node => {
+      if (node.type === Node.TYPES.NETWORK) {
+        records.push(node.record);
+      }
+    });
+
+    this._networkRecords = records;
+    return records;
+  }
+
+  _initializeNetworkConnections() {
+    const connections = new Map();
+    const recordsByConnection = groupBy(
+      this._networkRecords,
+      record => record.connectionId
+    );
+
+    for (const [connectionId, records] of recordsByConnection.entries()) {
+      const isSsl = records[0].parsedURL.scheme === 'https';
+      let responseTime = records.reduce(
+        (min, record) => Math.min(min, Estimator.getResponseTime(record)),
+        Infinity
+      );
+
+      if (!Number.isFinite(responseTime)) {
+        responseTime = this._defaultResponseTime;
+      }
+
+      const connection = new TcpConnection(
+        this._rtt,
+        this._throughput,
+        responseTime,
+        isSsl
+      );
+
+      connections.set(connectionId, connection);
+    }
+
+    this._connections = connections;
+    return connections;
+  }
+
+  _initializeAuxiliaryData() {
+    this._nodeAuxiliaryData = new Map();
+    this._nodesCompleted = new Set();
+    this._nodesInProcess = new Set();
+    this._nodesInQueue = new Set(); // TODO: replace this with priority queue
+    this._connectionsInUse = new Set();
+  }
+
+  /**
+   * @param {!Node} node
+   */
+  _enqueueNodeIfPossible(node) {
+    const dependencies = node.getDependencies();
+    if (
+      !this._nodesCompleted.has(node) &&
+      dependencies.every(dependency => this._nodesCompleted.has(dependency))
+    ) {
+      this._nodesInQueue.add(node);
+    }
+  }
+
+  /**
+   * @param {!Node} node
+   * @param {number} totalElapsedTime
+   */
+  _startNodeIfPossible(node, totalElapsedTime) {
+    if (node.type !== Node.TYPES.NETWORK) return;
+
+    const connection = this._connections.get(node.record.connectionId);
+
+    if (
+      this._nodesInProcess.size >= this._maximumConcurrentRequests ||
+      this._connectionsInUse.has(connection)
+    ) {
+      return;
+    }
+
+    this._nodesInQueue.delete(node);
+    this._nodesInProcess.add(node);
+    this._nodeAuxiliaryData.set(node, {
+      startTime: totalElapsedTime,
+      timeElapsed: 0,
+      timeElapsedOvershoot: 0,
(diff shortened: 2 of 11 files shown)


lighthouse-core/audits/predictive-perf.js
```diff
@@ -29,14 +30,77 @@ class PredictivePerf extends Audit {
...
+      };
+
+      let sum = 0;
+      const values = {};
+      Object.keys(graphs).forEach(key => {
+        values[key] = PageDependencyGraph.computeGraphDuration(graphs[key]);
+        sum += values[key];
+      });
+
+      const rawValue = sum / 4;
```
paulirish: magic number.  `Object.keys(graphs).length` ?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+'use strict';
+
+const Node = require('../node');
+const TcpConnection = require('./tcp-connection');
+
+// see https://cs.chromium.org/search/?q=kDefaultMaxNumDelayableRequestsPerClient&sq=package:chromium&type=cs
+const DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 10;
+const DEFAULT_RESPONSE_TIME = 30;
+const DEFAULT_RTT = 150;
+const DEFAULT_THROUGHPUT = 1600 * 1024; // 1.6 Mbps
```
paulirish: these two numbers are identical to our throttling presets. should we pull from emulation.js instead?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+   * @param {{rtt: number, throughput: number, defaultResponseTime: number,
+   *    maximumConcurrentRequests: number}=} options
+   */
+  constructor(graph, options) {
+    this._graph = graph;
+    this._options = Object.assign(
+      {
+        rtt: DEFAULT_RTT,
+        throughput: DEFAULT_THROUGHPUT,
+        defaultResponseTime: DEFAULT_RESPONSE_TIME,
```
paulirish: post-discussion:  naming-wise this value is very unclear.   defaultServerResponseTime is still ambiguous wether it includes latency or not. (and 1-way or 2-way?)  

`defaultTTFB` is very clear as a definition, so I'd prefer using that as the argument. That means some adding/subtracting of latency here and there, but i think it's worth it so the terms are clear.

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+      TcpConnection.maximumSaturatedConnections(this._rtt, this._throughput),
+      this._options.maximumConcurrentRequests
+    );
+  }
+
+  /**
+   * @param {!WebInspector.NetworkRequest} record
+   * @return {number}
+   */
+  static getResponseTime(record) {
```
paulirish: yes, this value is the definition of TTFB, so let's s/responseTime/ttfb/  :)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+const Node = require('../node');
+const TcpConnection = require('./tcp-connection');
+
+// see https://cs.chromium.org/search/?q=kDefaultMaxNumDelayableRequestsPerClient&sq=package:chromium&type=cs
+const DEFAULT_MAXIMUM_CONCURRENT_REQUESTS = 10;
+const DEFAULT_RESPONSE_TIME = 30;
+const DEFAULT_RTT = 150;
+const DEFAULT_THROUGHPUT = 1600 * 1024; // 1.6 Mbps
+
+function groupBy(items, keyFunc) {
```
paulirish: I wish we could install this to `Object.groupBy` so its as usable everywhere. the named props and array of obj dances get old. oh well.

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+
+  _initializeNetworkConnections() {
+    const connections = new Map();
+    const recordsByConnection = groupBy(
+      this._networkRecords,
+      record => record.connectionId
+    );
+
+    for (const [connectionId, records] of recordsByConnection.entries()) {
+      const isSsl = records[0].parsedURL.scheme === 'https';
```
paulirish: wss?  if that should be included we have SECURE_SCHEMES elsewhere :)

lighthouse-core/audits/predictive-perf.js

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+   *
+   * See https://hpbn.co/building-blocks-of-tcp/#three-way-handshake and
+   *  https://hpbn.co/transport-layer-security-tls/#tls-handshake for details.
+   *
+   * @param {number} bytesToDownload
+   * @param {number=} timeAlreadyElapsed
+   * @param {number=} maximumTimeToElapse
+   * @return {{timeElapsed: number, roundTrips: number, bytesDownloaded: number, congestionWindow: number}}
+   */
+  calculateTimeToDownload(bytesToDownload, timeAlreadyElapsed = 0, maximumTimeToElapse = Infinity) {
```
paulirish: naming threw me... `simulateDownloadingUntil` is way clearer.

kinda weird that  args 1 and 3 are the constraints but not next to eachother. should we just move to an object instead? Seems like we usually would at this point. strawman:

```js
simulateDownloadingUntil({maxBytes, maxMs, msElapsed})
```

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: heh yeah...I'm still all for a util lib :)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+      timeElapsedOvershoot: 0,
+      bytesDownloaded: 0,
+    });
+
+    this._connectionsInUse.add(connection);
+  }
+
+  _updateNetworkCapacity() {
+    for (const connection of this._connectionsInUse) {
+      connection.setThroughput(this._throughput / this._nodesInProcess.size);
```
paulirish: setThroughput vs availableThroughput makes me think there may be multiple... like currentTHroughput, defaultThroughput, etc... since there's only one should we just use that same name?

lighthouse-core/audits/predictive-perf.js
```diff
@@ -29,14 +30,77 @@ class PredictivePerf extends Audit {
...
+   * @param {!Node} graph
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {!Node}
+   */
+  static getOptimisticFMPGraph(graph, traceOfTab) {
+    const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
+    return graph.cloneWithRelationships(node => {
+      if (node.endTime > fmp) return false;
+      if (node.type !== Node.TYPES.NETWORK) return true;
+      return node.record.priority() === 'VeryHigh'; // proxy for render-blocking
```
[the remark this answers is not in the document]
patrickhulce: yup this is implemented in next PR :) FWIW in the future PR the remaining optimistic/pessimistic split is also excluding render blocking priorities if they were script initiated

lighthouse-core/audits/predictive-perf.js
```diff
@@ -29,14 +30,77 @@ class PredictivePerf extends Audit {
...
+    });
+  }
+
+  /**
+   * @param {!Node} graph
+   * @return {!Node}
+   */
+  static getOptimisticTTCIGraph(graph) {
+    return graph.cloneWithRelationships(node => {
+      return node.record._resourceType && node.record._resourceType._name === 'script' ||
```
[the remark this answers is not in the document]
patrickhulce: heh, do we want to go with the public accessors here?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: renamed to `fallbackTTFB` and added latency subtraction 👍

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: done

lighthouse-core/audits/predictive-perf.js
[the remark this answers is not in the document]
patrickhulce: this particular condition is removed in next PR but will update latest 👍

lighthouse-core/audits/predictive-perf.js
```diff
@@ -29,14 +30,77 @@ class PredictivePerf extends Audit {
...
+          node.record.priority() === 'High' ||
```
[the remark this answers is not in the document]
patrickhulce: future version has comments explaining each condition when more thought started going into them :) but will flesh out more

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+    const records = [];
+
+    this._graph.getRootNode().traverse(node => {
+      if (node.type === Node.TYPES.NETWORK) {
+        records.push(node.record);
+      }
+    });
+
+    this._networkRecords = records;
+    return records;
```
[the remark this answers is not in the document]
patrickhulce: nope it doesn't, done.

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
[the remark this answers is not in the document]
patrickhulce: yeah they won't be I keep track of number of in process by type separately 👍

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+    let handshakeAndRequest = oneWayLatency;
+    if (!this._warmed) {
+      handshakeAndRequest =
+        // SYN
+        oneWayLatency +
+        // SYN ACK
+        oneWayLatency +
+        // ACK + Application Data
+        oneWayLatency +
+        // ClientHello/ServerHello with TLS False Start
```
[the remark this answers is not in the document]
patrickhulce: sg

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+   * @return {{timeElapsed: number, roundTrips: number, bytesDownloaded: number, congestionWindow: number}}
+   */
+  calculateTimeToDownload(bytesToDownload, timeAlreadyElapsed = 0, maximumTimeToElapse = Infinity) {
+    const twoWayLatency = this._rtt;
+    const oneWayLatency = twoWayLatency / 2;
+    const maximumCongestionWindow = this._computeMaximumCongestionWindowInSegments();
+
+    let handshakeAndRequest = oneWayLatency;
+    if (!this._warmed) {
+      handshakeAndRequest =
```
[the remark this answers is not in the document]
patrickhulce: I'm not a huge fan of splitting it since the handshake is in fact concluding in the same trip as the request is being made, adding a comment with `// technically the ACK is still happening blah blah` seems a tad more to process than a variable name indicating the two are conjoined IMO. It also slightly complicates the later http2 case where neither are required once the connection has been warmed.

Do you think addressing the other TTFB issues makes this clearer or you'd still like to see the split?

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
patrickhulce: done, how about switching to an object param in the "incremental PR" after I can rebase against the name change?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,284 @@
...
+    if (
+      this._nodesInProcess.size >= this._maximumConcurrentRequests ||
+      this._connectionsInUse.has(connection)
+    ) {
+      return;
+    }
+
+    this._nodesInQueue.delete(node);
+    this._nodesInProcess.add(node);
+    this._nodeAuxiliaryData.set(node, {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: ok hang on, reopening the discussion :)

i'm gonna push back against this latency subtraction from TTFB deal a bit, I cherrypicked the change to my final fixes branch and it regressed accuracy by 0.3%

given we're significantly under reporting rtt in most cases i'm not sure we should be further reducing the download time estimates, how about as future exploration we try to do something fancy to identify/separate connection latency from server response time instead (I can imagine trying to determine the built-in connection latency by looking at all the requests, or comparing the TCP handshake time to the TTFB time etc but its all more researchy) and leave it be for now?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: yep done

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+        // SYN
+        oneWayLatency +
+        // SYN ACK
+        oneWayLatency +
+        // ACK + Application Data
+        oneWayLatency +
+        // ClientHello/ServerHello assuming TLS Flase Start is enabled (https://istlsfastyet.com/#server-performance).
```
paulirish: False

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+        (this._ssl ? twoWayLatency : 0);
+    }
+
+    let roundTrips = Math.ceil(handshakeAndRequest / twoWayLatency);
+    const timeToFirstByte = handshakeAndRequest + this._responseTime + oneWayLatency;
```
paulirish: _responseTime => _serverLatency

everywhere? :)

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
```diff
@@ -0,0 +1,147 @@
...
+    const maximumCongestionWindow = this._computeMaximumCongestionWindowInSegments();
+
+    let handshakeAndRequest = oneWayLatency;
+    if (!this._warmed) {
+      handshakeAndRequest =
+        // SYN
+        oneWayLatency +
+        // SYN ACK
+        oneWayLatency +
+        // ACK + Application Data
```
paulirish: ACK + initial request

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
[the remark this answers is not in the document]
paulirish: agreed to keep `handshakeAndRequest` as a unit as it's more accurate to the reality of TCP. 👍

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
paulirish: as discsussed offline, i think it'd be great to explore identifing per-server RTT later. this comment sg. let's add it. :)

lighthouse-core/audits/predictive-perf.js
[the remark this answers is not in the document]
paulirish: this is same as `node.record._resourceType === Common.resourceTypes.Script`  (which is fine).  The benefit is just dont have to reach all the way to _name

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
paulirish: >  how about switching to an object param in the "incremental PR" after I can rebase against the name change?

sure that's fine.

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
paulirish: update: saw the comment is added already. 👍

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/tcp-connection.js
patrickhulce: done

lighthouse-core/audits/predictive-perf.js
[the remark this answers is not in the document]
patrickhulce: ah ok, will re-open in the final version then where i've used `node.resourceType` then 👍

patrickhulce:
I think everything's addressed here, so back to you @paulirish build failures seem to be `bundlesize` step failing due to a github 404

brendankenny:
sorry, I'd still like to take a look :D:D

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    this._connections = connections;
+    return connections;
+  }
+
+  _initializeAuxiliaryData() {
+    this._nodeTiming = new Map();
+    this._nodesCompleted = new Set();
+    this._nodesInProcess = new Set();
```
brendankenny: in the same process? if not, may be a bit confusing naming-wise. `nodesInProgress`?

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
...
+    this._connectionsInUse = new Set();
+  }
+
+  /**
+   * @param {!Node} node
+   */
+  _enqueueNodeIfPossible(node) {
+    const dependencies = node.getDependencies();
+    if (
+      !this._nodesCompleted.has(node) &&
```
brendankenny: if all dependencies must have been completed to be enqueued, shouldn't it not be possible for a node to already be completed here (since it's being added as a dependent of another node)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: hahaha, good point, done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
...
+    for (const [connectionId, records] of recordsByConnection.entries()) {
+      const isTLS = TLS_SCHEMES.includes(records[0].parsedURL.scheme);
+
+      // We'll approximate how much time the server for a connection took to respond after receiving
+      // the request by computing the minimum TTFB time for requests on that connection.
+      //    TTFB = one way latency + server response time + one way latency
+      // Even though TTFB is greater than server response time, the RTT is underaccounted for by
+      // not varying per-server and so the difference roughly evens out.
+      // TODO(patrickhulce): investigate a way to identify per-server RTT
+      let estimatedResponseTime = records.reduce(
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -79,26 +79,37 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
-        depthByNodeId.set(node.id, maxDepth + 1);
-      });
+    return new GraphEstimator(rootNode).estimate();
+  }
+
+  /**
+   *
+   * @param {!Node} rootNode
+   */
+  static printGraph(rootNode, widthInCharacters = 100) {
```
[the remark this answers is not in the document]
patrickhulce: correct

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
...
+      // update how far each node will progress until that point
+      for (const node of nodesInProcess) {
+        this._updateProgressMadeInTimePeriod(
+          node,
+          minimumTime,
+          totalElapsedTime
+        );
+      }
+
+      if (depth > 10000) {
```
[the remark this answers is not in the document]
patrickhulce: While it was also very useful for early debugging with a small value, this is just a sensibility check to avoid looping forever and instead fail in the event of illogical network record or trace event relationships (or bug in graph construction though at this point it's synonymous with illogic event relationships given the interpretations we're taking).

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
+/**
```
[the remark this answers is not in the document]
paulirish: I was thinking the same, but came to the same conclusion that we move them _after_ this series of PRs land. 👍

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: This is cleaned up a little bit in #3162 but the summary here is that it was the only check against duplicate processing of a node. By definition, if a node has been completed, all of its dependencies have been completed, so while we're enqueing don't start on a node we've already finished (really should be if a node we've already started which is fixed later but it didn't have an impact when only network records were considered). 

Once again, this logic is re-arranged in the next PR to be clearer.

lighthouse-core/audits/predictive-perf.js
```diff
@@ -31,14 +32,77 @@ class PredictivePerf extends Audit {
   }
 
   /**
+   * @param {!Node} graph
```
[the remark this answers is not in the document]

lighthouse-core/audits/predictive-perf.js
```diff
@@ -31,14 +32,77 @@ class PredictivePerf extends Audit {
...
+      };
+
+      let sum = 0;
+      const values = {};
+      Object.keys(graphs).forEach(key => {
+        values[key] = PageDependencyGraph.computeGraphDuration(graphs[key]);
+        sum += values[key];
+      });
+
+      const rawValue = sum / Object.keys(values).length;
```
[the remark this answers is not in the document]

lighthouse-core/gather/computed/dependency-graph/node.js
```diff
@@ -36,6 +35,27 @@ class Node {
   }
 
   /**
+   * @return {string}
+   */
+  get type() {
+    return Node.TYPES.UNKNOWN;
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/node.js
```diff
@@ -56,8 +76,14 @@ class Node {
...
   getRootNode() {
     let rootNode = this;
-    while (rootNode._dependencies.length) {
+    let maxDepth = 1000;
+    while (rootNode._dependencies.length && maxDepth) {
       rootNode = rootNode._dependencies[0];
+      maxDepth--;
+    }
+
+    if (!maxDepth) {
```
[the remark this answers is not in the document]
patrickhulce: It means something is wrong with graph that's been created, see https://github.com/GoogleChrome/lighthouse/pull/3187#discussion_r137120833 I think it's still more useful than having a fatal infinite loop on bad input.

lighthouse-core/gather/computed/dependency-graph/node.js
```diff
@@ -121,6 +147,11 @@ class Node {
       if (!shouldIncludeNode(originalNode)) return;
       const clonedNode = originalNode.cloneWithoutRelationships();
       idToNodeMap.set(clonedNode.id, clonedNode);
+    });
+
+    rootNode.traverse(originalNode => {
```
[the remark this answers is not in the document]
patrickhulce: it's not guaranteed that a node will be visited by `traverse` before all of its dependencies, it's just doing a traversal of the edges not necessarily a topological sort ordered traversal

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -79,26 +79,37 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
+    function padRight(str, target, padChar = ' ') {
+      while (str.length < target) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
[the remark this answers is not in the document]
patrickhulce: happy to have a global "where should things live" discussion post #3163 :)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
```diff
@@ -0,0 +1,291 @@
...
+    const key = keyFunc(item);
+    const group = grouped.get(key) || [];
+    group.push(item);
+    grouped.set(key, group);
+  });
+
+  return grouped;
+}
+
+class Estimator {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/computed/dependency-graph/node.js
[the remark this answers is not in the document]
brendankenny: yeah, which I guess is good :) (especially during ongoing development) but it does feel like a check for a properly formed graph at construction time is the right solution (plus we could probably give better error messages for what went wrong with graph construction)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
brendankenny: > By definition, if a node has been completed, all of its dependencies have been completed, so while we're enqueing don't start on a node we've already finished

right, I was trying to say that always checking `dependencies.every(dependency => this._nodesCompleted.has(dependency))` before enqueuing should be sufficient, because it means if a dependent of a newly-completed node is being enqueued,  it can't have even been in the queue before (let alone completed), because one of its dependencies wasn't completed until just now.

We can wait as it changes, but it feels like it's a `// should never happen` situation and it should throw rather than ignoring just in case it does ever happen :)

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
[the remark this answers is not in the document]
brendankenny: > this is just a sensibility check to avoid looping forever and instead fail in the event of illogical network record or trace event relationships

same as above, maybe we can move to an `assertValidDepGraph` method in the future

lighthouse-core/gather/computed/dependency-graph/estimator/estimator.js
patrickhulce: Oh oh I see my confusion here now :) 

In another version, this method was called in a loop over all nodes, not just the dependents of recently completed nodes. Yup, yup in this snapshot it's a 100% superfluous check.

lighthouse-core/gather/computed/dependency-graph/node.js
```diff
@@ -177,4 +208,9 @@ class Node {
   }
 }
 
+Node.TYPES = {
+  UNKNOWN: 'unknown',
```
brendankenny: remove?

Issue #1822, opened 2017-03-06
Extension: support disable device emulation flag


sowmyamanusani:
When I do generate report , it is always opening mobile version of the page however I entered page in desktop version. How can I change the settings to open a desktop version but not mobile

patrickhulce:
@sowmyamanusani you can use the `--disable-device-emulation` flag on the CLI to run lighthouse without mobile emulation

sowmyamanusani:
I am using google chrome plugin. I do not see any option to disable that mobile emulation

patrickhulce:
Not all options available on the CLI are available in the extension yet, though it's a good first bug if you're so inclined :)

ZZhaoTireless:
I would like to fix this.

ZZhaoTireless:
@patrickhulce Could you tell me which file is responsible for the flag in extension folder? I did try to find the way to add flag in extension but I had no clue. I will be glad to help if there is a little hint. Thanks!

Pines-Cheng:
That's cool!

patrickhulce:
closing as dupe of #2838

PR #3209, opened 2017-08-31
Add flags to chrome lighthouse extension

mlucool:
Fixes https://github.com/GoogleChrome/lighthouse/issues/1822


brendankenny:
Code has changed quite a bit since this was opened, so closing for now.

It feels kind of ridiculous to suggest this since it's been so long @mlucool, but if you're still interested in this feature we can always open an issue to discuss!

mlucool:
No worries. I am really happy with the built in version in chrome that this doesn't seem needed anymore.

PR #3227, opened 2017-09-01
SEO - description audit

### LINKED ISSUE #3175: [SEO Audits] Document has a meta description
**Description:** Document has a meta description
**Failure description:** Document does not have a meta description
**Help text:** Meta descriptions may be included in search results to concisely summarize page content. Read more in the [Search Console Help page](https://support.google.com/webmasters/answer/35624?hl=en#1).

**Success condition:** Query selector `head > meta[name=description]` exists and has a non-empty content attribute.

Notes
- If short-circuiting the audit when `nosnippet` exists, use the audit description field to explain that the meta description is unused. Other `robots` directives that affect the meta description like `noindex` are covered in separate audits. _(not used)_

kdzwinel:
Closes #3175

### DIFF
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -29,8 +29,22 @@ class Description extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
+    if (artifacts.MetaDescription === null) {
+      return {
+        rawValue: false,
+        debugString: 'Description meta tag is missing.'
+      };
+    }
+
+    if (artifacts.MetaDescription.trim().length === 0) {
+      return {
+        rawValue: false,
+        debugString: 'Description text is empty.'
+      };
+    }
+
     return {
-      rawValue: (artifacts.MetaDescription !== null && artifacts.MetaDescription.length > 0)
+      rawValue: true
     };
   }
 }
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -30,7 +30,7 @@ class Description extends Audit {
    */
   static audit(artifacts) {
     return {
-      rawValue: (artifacts.Description !== null && artifacts.Description.length)
+      rawValue: (artifacts.Description !== null && artifacts.Description.length > 0)
     };
   }
 }
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -31,8 +31,7 @@ class Description extends Audit {
   static audit(artifacts) {
     if (artifacts.MetaDescription === null) {
       return {
-        rawValue: false,
-        debugString: 'Description meta tag is missing.'
+        rawValue: false
       };
     }
 
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -20,7 +20,7 @@ class Description extends Audit {
       helpText: 'Meta descriptions may be included in search results to concisely summarize ' +
           'page content. Read more in the ' +
           '[Search Console Help page](https://support.google.com/webmasters/answer/35624?hl=en#1).',
-      requiredArtifacts: ['Description']
+      requiredArtifacts: ['MetaDescription']
     };
   }
 
@@ -30,7 +30,7 @@ class Description extends Audit {
    */
   static audit(artifacts) {
     return {
-      rawValue: (artifacts.Description !== null && artifacts.Description.length > 0)
+      rawValue: (artifacts.MetaDescription !== null && artifacts.MetaDescription.length > 0)
     };
   }
 }
--- a/lighthouse-core/config/seo.js
+++ b/lighthouse-core/config/seo.js
@@ -7,6 +7,20 @@
 
 module.exports = {
   extends: 'lighthouse:default',
+  passes: [{
+    passName: 'seoPass',
+    recordTrace: true,
+    pauseAfterLoadMs: 5250,
+    networkQuietThresholdMs: 5250,
+    cpuQuietThresholdMs: 5250,
+    useThrottling: true,
+    gatherers: [
+      'seo/description',
+    ]
+  }],
+  audits: [
+    'seo/description'
+  ],
   groups: {
     'seo-mobile': {
       title: 'Mobile Friendly',
@@ -28,8 +42,9 @@ module.exports = {
       description: 'These ensure your app is able to be understood by search engine crawlers.',
       audits: [
         {id: 'meta-viewport', weight: 1, group: 'seo-mobile'},
-        {id: 'document-title', weight: 1, group: 'seo-content'}
+        {id: 'document-title', weight: 1, group: 'seo-content'},
+        {id: 'meta-description', weight: 1, group: 'seo-content'}
       ]
     }
-  }
+  },
 };
--- a/lighthouse-core/config/seo.js
+++ b/lighthouse-core/config/seo.js
@@ -38,7 +38,7 @@ module.exports = {
       audits: [
         {id: 'meta-viewport', weight: 1, group: 'seo-mobile'},
         {id: 'document-title', weight: 1, group: 'seo-content'},
-        {id: 'meta-description', weight: 1, group: 'seo-content'}
+        {id: 'meta-description', weight: 1, group: 'seo-content'},
       ]
     }
   }
--- a/lighthouse-core/config/seo.js
+++ b/lighthouse-core/config/seo.js
@@ -9,17 +9,12 @@ module.exports = {
   extends: 'lighthouse:default',
   passes: [{
     passName: 'seoPass',
-    recordTrace: true,
-    pauseAfterLoadMs: 5250,
-    networkQuietThresholdMs: 5250,
-    cpuQuietThresholdMs: 5250,
-    useThrottling: true,
     gatherers: [
       'seo/description',
     ]
   }],
   audits: [
-    'seo/description'
+    'seo/meta-description'
   ],
   groups: {
     'seo-mobile': {
@@ -46,5 +41,5 @@ module.exports = {
         {id: 'meta-description', weight: 1, group: 'seo-content'}
       ]
     }
-  },
+  }
 };
--- a/lighthouse-core/config/seo.js
+++ b/lighthouse-core/config/seo.js
@@ -8,13 +8,13 @@
 module.exports = {
   extends: 'lighthouse:default',
   passes: [{
-    passName: 'seoPass',
+    passName: 'defaultPass',
     gatherers: [
-      'seo/description',
+      'seo/meta-description',
     ]
   }],
   audits: [
-    'seo/meta-description'
+    'seo/meta-description',
   ],
   groups: {
     'seo-mobile': {
--- a/lighthouse-core/test/audits/seo/meta-description-test.js
+++ b/lighthouse-core/test/audits/seo/meta-description-test.js
@@ -13,19 +13,19 @@ const assert = require('assert');
 describe('SEO: description audit', () => {
   it('fails when HTML does not contain a description meta tag', () => {
     return assert.equal(Audit.audit({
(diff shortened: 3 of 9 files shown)


lighthouse-core/config/seo.js
```diff
@@ -7,6 +7,15 @@
 
 module.exports = {
   extends: 'lighthouse:default',
+  passes: [{
+    passName: 'seoPass',
```
patrickhulce: if you set `passName` to 'defaultPass' then the gatherer should just be added to the list to run, if not it's a bug and I can look into it :)

paulirish: patrick says you can just `passName: 'defaultPass'` and get this gather merged into that one.

so no need for an extra pass

lighthouse-core/audits/seo/meta-description.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+  }
+
+  /**
+   * @param {!Artifacts} artifacts
+   * @return {!AuditResult}
+   */
+  static audit(artifacts) {
+    return {
+      rawValue: (artifacts.Description !== null && artifacts.Description.length > 0)
```
kdzwinel: 1) Should we also `.trim()` it?
2) Should we include additional text explaining that either description was not found or it was empty?
3) BTW I've spotted that returning a numeric value here causes test to fail but green tick to be shown next to the test:
screen shot 2017-09-01 at 21 21 45
Is this a bug?

lighthouse-core/audits/seo/meta-description.js
```diff
@@ -0,0 +1,38 @@
...
+  static get meta() {
+    return {
+      category: 'Content Best Practices',
+      name: 'meta-description',
+      description: 'Document has a meta description',
+      failureDescription: 'Document does not have a meta description',
+      helpText: 'Meta descriptions may be included in search results to concisely summarize ' +
+          'page content. Read more in the ' +
+          '[Search Console Help page](https://support.google.com/webmasters/answer/35624?hl=en#1).',
+      requiredArtifacts: ['Description']
```
patrickhulce: `Description` is rather generic, what about `MetaDescription` or start namespacing with `SEODescription` if there are going to be many more seo specific ones?

kdzwinel:
@paulirish @patrickhulce thank you for your comments!

1) defaultPass - it works! However, it looks like my gatherer is now painfully slow for some reason:
```
  status Retrieving: WebSQL +2ms
  status Retrieving: MetaDescription +500ms
```
Interestingly, the problem goes away if it's not running last:
```
  status Retrieving: MetaDescription +3ms
  status Retrieving: WebSQL +3ms
```
🤔

2) I've renamed `Description` artefact to `MetaDescription` and updated gatherer's name accordingly.

3) I'll appreciate someone helping me with these questions: https://github.com/GoogleChrome/lighthouse/pull/3227#discussion_r136660967

lighthouse-core/audits/seo/meta-description.js
patrickhulce: 1. trim sgtm
2. debugString for those cases would be nice
3. yes this seems like a bug :) there's some magic that happens if the rawValue is a number to auto-populate the score but even if that's producing something weird the UI should still be consistent.

patrickhulce:
> It seems to me that 'MetaDescription' taking 500ms has something to do with WebSQL having a 500ms timeout

Yeah the debug timestamps are confusing to read, the `x ms` is reporting to you how many milliseconds have passed since the previous statement, since we print the name of the gatherer before we call it, the time for a gatherer is actually printed on the *next* line

lighthouse-core/audits/seo/meta-description.js
kdzwinel: Thanks for help! Regarding no.3 - should I create an issue for it?

ali999abas:
Good job

kdzwinel:
> `x ms` is reporting to you how many milliseconds have passed since the previous statement
thanks for explanation, it makes sense (although I'll have to agree that it's not very intuitive)

I've added description trimming and debugStrings. PTAL

lighthouse-core/audits/seo/meta-description.js
brendankenny: Inconsistency does look like a bug. In the meantime, should be able to set `scoringMode: Audit.SCORING_MODES.NUMERIC` in the meta information (otherwise everything other than 100 is interpreted as failing)

kdzwinel: @brendankenny thanks! For this audit rawValue is a boolean so we are fine, but I'll keep your advice in mind for the next ones.

lighthouse-core/audits/seo/meta-description.js
```diff
@@ -0,0 +1,52 @@
...
+
+const Audit = require('../audit');
+
+class Description extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Content Best Practices',
```
rviscomi: Meta question for the LH team. `category` here is referring to the title of the audit _group_, not to be confused with the top-level audit _category_, SEO, defined in the config file?

https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/seo.js#L25-L27

lighthouse-core/audits/seo/meta-description.js
rviscomi: If the meta description exists but is invalid, I'd also recommend including the markup in the debugString. For example:

```
Description text is empty: `<meta name="description" content="     ">`
```

It'll be helpful for developers to know where to look to fix the issue.

lighthouse-core/audits/seo/meta-description.js
patrickhulce: @rviscomi yeah the field here has an ambiguous name and is not used anywhere anymore 😆 

I'm in favor of removing entirely now that we have real categories that control quite a bit of report building.

lighthouse-core/audits/seo/meta-description.js
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * @param {!Artifacts} artifacts
+   * @return {!AuditResult}
+   */
+  static audit(artifacts) {
+    if (artifacts.MetaDescription === null) {
+      return {
+        rawValue: false,
+        debugString: 'Description meta tag is missing.'
```
patrickhulce: Actually now that you've brought it up in the other issue, I think I might have a slight preference to put these in the `displayValue` and keep `debugString` for errors like the gatherer failing (which is handled automatically)

lighthouse-core/audits/seo/meta-description.js
kdzwinel: @rviscomi I think that's a good idea, but I wonder if it won't be confusing to show `<meta name="description" content="     ">` while the actual tag on page may look totally different (e.g. `<META YADA="yada" CONTENT="     " name="description" />`)?

lighthouse-core/audits/seo/meta-description.js
brendankenny: > Actually now that you've brought it up in the other issue, I think I might have a slight preference to put these in the displayValue and keep debugString for errors like the gatherer failing (which is handled automatically)

agree about confusion between user vs system errors in `debugString`, but don't we still want the red string for failure here? (We still use the guidance "Optional error string for helping the user figure out why they failed here." in our docs)

lighthouse-core/test/audits/seo/meta-description-test.js
```diff
@@ -0,0 +1,37 @@
...
+'use strict';
+
+const Audit = require('../../../audits/seo/meta-description.js');
+const assert = require('assert');
+
+/* eslint-env mocha */
+
+describe('SEO: description audit', () => {
+  it('fails when HTML does not contain a description meta tag', () => {
+    return assert.equal(Audit.audit({
```
brendankenny: it would be good to assert that the audit is also returning a `debugString` (or a `displayValue`, depending on how the comment above shakes out :)

kdzwinel: will do 👍

lighthouse-core/audits/seo/meta-description.js
kdzwinel: IMO `debugString` is more readable and, as @brendankenny mentioned, in this case it matches the description in the docs:

https://github.com/GoogleChrome/lighthouse/blob/fadaef95137181aa41ed44fd78f17d3a8b85a2c3/lighthouse-core/closure/typedefs/AuditResult.js#L31-L35

@patrickhulce WDYT? Can we keep the `debugString` here?

lighthouse-core/config/seo.js
```diff
@@ -28,7 +37,8 @@ module.exports = {
       description: 'These ensure your app is able to be understood by search engine crawlers.',
       audits: [
         {id: 'meta-viewport', weight: 1, group: 'seo-mobile'},
-        {id: 'document-title', weight: 1, group: 'seo-content'}
+        {id: 'document-title', weight: 1, group: 'seo-content'},
+        {id: 'meta-description', weight: 1, group: 'seo-content'}
```
brendankenny: nit: go for the trailing comma here so future diffs will be cleaner

lighthouse-core/test/audits/seo/meta-description-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+/* eslint-env mocha */
+
+describe('SEO: description audit', () => {
+  it('fails when HTML does not contain a description meta tag', () => {
+    const auditResult = Audit.audit({
+      MetaDescription: null
+    });
+    assert.equal(auditResult.rawValue, false);
+    assert.ok(auditResult.debugString);
```
brendankenny: it would probably be (slightly) better to check which string is getting returned as the `debugString` since there's two different cases (could go for an exact string match or a `debugString.includes()` on `'empty'` and `'missing'`

lighthouse-core/audits/seo/meta-description.js
patrickhulce: Yeah sgtm, I'm sort of off the reservation when it comes to my hopes for debugString anyway :)

lighthouse-core/test/audits/seo/meta-description-test.js
patrickhulce: +1 to a quick includes check

lighthouse-core/audits/seo/meta-description.js
paulirish: I'll put the title and debugString in the order you'd see them in the report:

> #### Document does not have a meta description
> Description meta tag is missing

Given that the debugString duplicates so much, i think we should exclude it.

lighthouse-core/audits/seo/meta-description.js
```diff
@@ -0,0 +1,52 @@
...
+      };
+    }
+
+    if (artifacts.MetaDescription.trim().length === 0) {
+      return {
+        rawValue: false,
+        debugString: 'Description text is empty.'
```
paulirish: 👍  Seems okay to keep this, as it's an unexpected case that'd otherwise be hard to debug.

lighthouse-core/config/seo.js
kdzwinel: I saw trailing comma being used here and there, but since there is no 100% consistency, I left it out. The cleaner diffs argument SGTM, maybe we should make eslint to check for that?

kdzwinel:
I removed the 'tag missing' `debugString`, added the missing trailing comma and, updated tests to check `debugString` with `includes` (there is only one `debugString` now, but IMO it still makes sense to have `includes`).

PTAL

lighthouse-core/config/seo.js
patrickhulce: AFAIK there wasn't any effort to do trailing commas other than occasionally sneaking them in as we touched code but huge +1 to turning on the eslint rule :)

kdzwinel: I've opened an issue for it: https://github.com/GoogleChrome/lighthouse/issues/3304 👍

PR #3283, opened 2017-09-08
Add Lighthouse-assert

denar90:
This PR solves #1832 and also adds more sugar to `smokehouse` and simplify using Lighthouse with CI.
Changes:
 - `smokehouse` depends on `lighthouse-assert` for comparing results
 - integrated into CLI. passing just expectations config will pass/fail CI
 - supports different reporters, hence users can write their own
 - right now it compares just `audits`, will be more in future

I'm open to concerns/suggestions/questions.

cc @paulirish @brendankenny


lighthouse-assert/test/fixtures/expectations/default-expectations.ts
```diff
@@ -0,0 +1,45 @@
...
+      'first-contentful-paint': {
+        score: {
+          "warn": '>=65',
+          "error": '<=85',
+        },
+        rawValue: '>3000',
+      },
+      'first-meaningful-paint': {
+        score: {
+          "warn": '<=65',
```
patrickhulce: I'm a little confused by the values that have `warn` less than the `error` value, shouldn't it be the other way around?

lighthouse-assert/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  private logger: IReporter;
+
+  constructor(reporter?: IReporter) {
+    reporter = reporter || new DefaultReporter();
+    this.logger = new Logger(reporter);
+  }
+
+  assert(results: Array<IResult>, expectations: Array<IExpectation>) {
+    this.assertInstance = new Assert(results, expectations);
+    this.assertInstance.collate();
```
patrickhulce: the usage pattern of this feels a bit weird to me, is this modeling something pre-existing that I'm not aware of?

I ask because I would expect collate to return some object that contains the results found rather than it modifying the object and needing to call other methods to find out what happened.

lighthouse-assert/assert/assert.ts
```diff
@@ -0,0 +1,405 @@
+export interface IResult {
```
patrickhulce: this feels a bit like a monolith file to have so many types, wdyt about breaking out the results to an object that would be returned by an `Assert` instance and moving some of the logic there?

lighthouse-assert/assert/assert.ts
```diff
@@ -0,0 +1,405 @@
...
+
+  /**
+   * Checks if the actual value in warning and error range
+   *
+   * @param {number} actual
+   * @param {INormalizedExpectedScore} expected
+   * @return {boolean}
+   */
+  private inRange(actual: number, expected: INormalizedExpectedScore): boolean {
+    return actual >= expected.error || actual >= expected.warn && actual < expected.error;
```
patrickhulce: it seems that greater the greater than/less than operators in the error and warn are thrown away then?

lighthouse-assert/test/fixtures/expectations/default-expectations.ts
denar90: Yeah, of course. My bad. Thx 👍

lighthouse-assert/index.ts
denar90: No, it's not smth pre-existing :) , I just wanted to compare results not during instantiating class but with other action, that's why I added `collate` method. Another idea was to reuse `collatedResults` in future e.g. for comparing like `equal`, but right now I realized that was not a good idea...

lighthouse-assert/assert/assert.ts
denar90: Totally agree, I'll split it up.

> breaking out the results to an object that would be returned by an Assert instance and moving some of the logic there

do you mean all 'collate' logic?

lighthouse-assert/assert/assert.ts
denar90: I'm not sure I got your point correct. Are you concerned that original `expected.score.warn` and `expected.score.error` are changed during normalizing?

denar90:
@patrickhulce thank you 😃 I'm not expecting that it will be merged asap because there are really a looot stuff for review. Your plan sgtm 👍 
P.S. The reason I did it in scope of one PR is making sure everything works together :)

lighthouse-assert/assert/assert.ts
patrickhulce: Well it looks like you're setting the warn and expectations as strings with `>=` or `<=` but then the actual comparison that happens here doesn't seem to use that operator information since the operators are hard-coded

lighthouse-assert/assert/assert.ts
patrickhulce: yeah basically the same idea as https://github.com/GoogleChrome/lighthouse/pull/3283#discussion_r137894453

if `collate()` returned some `ExpectationsResult` object that had the `isEqual()` and `getDifferenceCount()` methods on it

patrickhulce:
> P.S. The reason I did it in scope of one PR is making sure everything works together :)

no worries I like to do the same, just makin' sure the expectation was some smaller reviews 😄

patrickhulce:
> I'd be interested in how others feel

I'll propose moving *everything* to typescript :)

lighthouse-assert/assert/assert.ts
denar90: now I got it) thx

denar90:
@brendankenny thanks :) 
1. The idea was to have a strict format like `score: {error: '<75', warn: '<75'}` and I understand your confusion because after merging latest `master` expectations in smokehouse were updated, that's why some scores look inconsistent. So I dunno what to do, either left this behavior or stick to your one?
2. I know that my point doesn't count but wanna left a comment about typescript. I've chosen it just because of logic kinda hard so having different interfaces for different cases sounded reasonable (SOLID shmolid, you know ;) ). Other than that LH has other packages used ts, so thought for smoother integration appeared. In any case, if you guys in a scope of a core team come up with solution dropping ts then I'm open to rewriting it.

lighthouse-assert/assert/difference/object.ts
```diff
@@ -0,0 +1,80 @@
...
+   * @return {boolean}
+   */
+  matchesExpectation(): boolean {
+    const actualValue = this.normalize(this.actual.score);
+    const normalizedExpected: INormalizedExpectedScore = {
+      warn: this.normalize(this.expected.score.warn),
+      error: this.normalize(this.expected.score.error)
+    };
+    // @todo check inRange due to symbols like '<=', '>=' etc.
+    return inRange(normalizedExpected.error, 100, actualValue);
```
denar90: what I've been thinking is getting `<=, >=` during normalize, but then understood, why should we allow users to do crazy setups like `score: { warn: '>=70', 'error': '<70'}` or `score: { warn: '<80', error: '>90'}` etc. 
What if we say that they allowed setting expectation lower/equal than result like `score: { warn: '<=70', 'error': '<60'}`, `score: { warn: '70', 'error': '60'}`, '`score: true`', `score: 100`, `score: '<70'`,skippinga cases above?

denar90:
I've updated do to comments, but some smokehouse tests fail because of not decided `inRange` behavior. I was lazy and didn't wanna fix them right now and later after decision about all `inRange` logic :)

lighthouse-assert/assert/difference/object.ts
patrickhulce: it's useful to have assertions on the length of details, rawValue, etc which was the original motivation for the `>=` I'm not the biggest fan of the warn/error object (what do we do if there's an error property of something we want to check?) 

have we considered other possible approaches for expressing warn vs. error?
what do we think the most common use case is? I'm assuming it's going to be failing this in CI which brings up the question of how valuable warn is and if it'd be ok to have a more roundabout usage for warn. It seems like we'll need to make a decision about whether we want to easily support warn/error and go an eslint-like direction that only asserts score or if we want to easily support assertions on arbitrary properties (like smokehouse does now) and force a separate config if you want warn assertions.

denar90: Approach having warn and error borrowed from pwmetrics. Why have warn? 
Sometimes lighthouse can return not consistent results for several runs (also other people faced this kind of problem, reff https://medium.com/@trueadm/this-looks-awesome-and-is-something-similar-to-what-i-built-for-react-ebdd77192258). Having warn will just show message and CI pass.

I think we can consider using more than score, and change all `difference` logic to work with abstract thing from inside, e.g. not
`findDifference(auditName, actualResult, expectedResult)` 
but 
`findDifference(auditName, resultValue, expectedResultValue);` 
where `expectedResultValue` -> `{warn: '<=70', error: '<60' } or smth from cases above`. On this inside level we can isolate what this `expectedResultValue` value will be, either `score`, or `rawValue`, or something else (we can have a list of availiable things).

But your idea about config aka eslint can also give us a good opportunity. It can be simpler than current config config aka
```
"first-meaningful-paint": {
  "expected": {
    "warn": "<=85",
    "error": "<=65"
  },
  "compareWith": "rawValue"
},
"redirects-http": {
   "expected": true,
  "compareWith": "score"
},
```

we should choose which approach to use, so I can continue work on that...

denar90:
Guys, are you still interested in this one?

paulirish:
We are, but we don't have consensus on adding more TS to the codebase. We're interested in seeing if the `checkJs` approach will provide as much benefit. That [approach](https://github.com/GoogleChrome/puppeteer/pull/219) in short: source is all JS, type annotations are provided as comments, a stub d.ts is generated.

denar90:
Ok, got your point, I'll try to refactor things, wanna do it asap, but it will not be immediate... Thanks for the reply.

denar90:
Finally, have an opportunity to look at it one more time and work on rewriting.

@patrickhulce @brendankenny @paulirish is it still can be useful, should I spend time improving this PR?

denar90:
Does it make sense continue/finishing before v3 release? Or do you prefer to have it after v3 release?

brendankenny:
@denar90 I think we should continue to approach this with external utilities for now (there are a few now in addition to pwmetrics) and let things settle a bit before thinking again about integrating directly into Lighthouse. And we should definitely coordinate with the discussion in #6053 which wants very similar things to what this PR was doing.

Going to close for now, as regardless of approach, it's been so long that it's probably going to require a lot of rewriting :/

PR #3287, opened 2017-09-10
Lighthouse is in Chrome 60

rootulp:
The Audits panel was added in Chrome 60: https://developers.google.com/web/updates/2017/05/devtools-release-notes


rootulp:
I signed it!

PR #3267, opened 2017-09-06
test(smokehouse): add basic smoke test for SEO audits

brendankenny:
@kdzwinel FYI

this adds a starting point for SEO audit integration tests. Right now this tests only the two a11y audits that are also being used as SEO audits, but since they're a11y audits they don't have many details to test beyond the score

as new SEO audits get added:
- add audits to the expectations file (obviously :)
- new audits may have a lot more detail in `extendedInfo`/`details` that can be tested ([example](https://github.com/GoogleChrome/lighthouse/blob/a9b02a948b579be1a0667b69627f6d1f831d30ca/lighthouse-cli/test/smokehouse/pwa-expectations.js#L40-L56))
- could add a second html tester file that fails some/all of these audits and asserts proper failure output in those cases

PR #3309, opened 2017-09-12
Apply small license header to last remaining files

paulirish:
Noticed in https://github.com/GoogleChrome/lighthouse/pull/3267/files#r138413155


lighthouse-cli/test/smokehouse/byte-efficiency/run-tests.sh
```diff
@@ -13,4 +13,4 @@ exit_code=$?
 # kill test servers
 kill $(jobs -p)
 
-exit "$exit_code"
+exit 1; # "$exit_code"
```
patrickhulce: this seems unintentional?

paulirish: indeed. thx

PR #3310, opened 2017-09-12
test(smokehouse): add TTFB smoke test

brendankenny:
@wardpeet

followup to #2231

### DIFF
--- a/lighthouse-cli/test/smokehouse/perf/expectations.js
+++ b/lighthouse-cli/test/smokehouse/perf/expectations.js
@@ -32,6 +32,10 @@ module.exports = [
       'consistently-interactive': {
         score: '>=90',
       },
+      'time-to-first-byte': {
+        // Can be flaky, so test float rawValue instead of boolean score
+        rawValue: '<1000',
+      },
     },
   },
 ];


wardpeet:
thanks @brendankenny!
LGTM, not much changed 👍

lighthouse-cli/test/smokehouse/perf/expectations.js
```diff
@@ -32,6 +32,10 @@ module.exports = [
       'consistently-interactive': {
         score: '>=90',
       },
+      'time-to-first-byte': {
+        // Can be flaky, so test float rawValue instead of boolean score
```
patrickhulce: we can add a `?delay=650` or something to the main URL to force it to fail if we want to check the boolean score?

brendankenny: do we prefer failure? either way we're only checking one execution path

brendankenny: this does check that `rawValue` is a number and that it's less than 1000. I could tighten that up based on what passes on Travis, though

patrickhulce: I'm fine either way just wanted to bring it up since the comment suggests that it was *trying* to check the boolean but couldn't, at first I did misread this as score < 1000 though 😆

brendankenny: Oh I see, yeah, the delay would push `score` firmly over to the fail side so we could test reliably.

I'm ok with the current approach rather than slowing it down (which would make me have to figure out new values for all the other numbers in this file :)

PR #3377, opened 2017-09-20
Docs: Eliminate integration with CI

denar90:
Since e.g. Travis has chrome under the hood https://docs.travis-ci.com/user/chrome so this section is no longer needed. I think ...

### DIFF
--- a/chrome-launcher/README.md
+++ b/chrome-launcher/README.md
@@ -129,3 +129,5 @@ before_script:
 addons:
   chrome: stable
 ```
+
+> Since Travis can install Chrome at a run time you no longer need to install it manually. Read [more](https://docs.travis-ci.com/user/chrome).
--- a/chrome-launcher/README.md
+++ b/chrome-launcher/README.md
@@ -125,5 +125,7 @@ before_script:
   - export DISPLAY=:99.0
   - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
   - sh -e /etc/init.d/xvfb start
-  - curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash
+  - sleep 3 # wait for xvfb to boot
+addons:
+  chrome: stable
 ```
--- a/chrome-launcher/README.md
+++ b/chrome-launcher/README.md
@@ -111,9 +111,10 @@ chromeLauncher.launch({
 
 ### Continuous Integration
 
-In a CI environment like Travis, Chrome may not be installed. If you want to use `chrome-launcher`, you can install Chrome using Lighthouse's `download-chrome.sh` script:
+In a CI environment, Chrome may not be installed. If you want to use `chrome-launcher`, e.g. Travis can [install Chrome at run time with an addon](https://docs.travis-ci.com/user/chrome).
 
-`curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash`
+> You can also install Chrome manually using Lighthouse's `download-chrome.sh` script:
+> `curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash`
 
 Then in `.travis.yml`, use it like so:
 
@@ -129,5 +130,3 @@ before_script:
 addons:
   chrome: stable
 ```
-
-> Since Travis can install Chrome at a run time you no longer need to install it manually. Read [more](https://docs.travis-ci.com/user/chrome).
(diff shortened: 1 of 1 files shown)


chrome-launcher/README.md
```diff
@@ -108,22 +108,3 @@ chromeLauncher.launch({
...
-Then in `.travis.yml`, use it like so:
-
-```yaml
-language: node_js
-install:
-  - yarn install
-before_script:
-  - export DISPLAY=:99.0
-  - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
-  - sh -e /etc/init.d/xvfb start
```
brendankenny: still need xvfb, right? Also what do you think of using the `addons` lines in this example (and maybe linking to those docs)?

For new users it's nice to have a simple but complete solution, especially since there's probably lots of outdated answers out there about running Chrome on Travis (like this one, until your change lands :)

denar90: Yes, you are right 👍  Looks like I went too far after reading docs section in Travis...

chrome-launcher/README.md
```diff
@@ -125,5 +125,9 @@ before_script:
   - export DISPLAY=:99.0
   - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
   - sh -e /etc/init.d/xvfb start
-  - curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash
+  - sleep 3 # wait for xvfb to boot
+addons:
+  chrome: stable
 ```
+
+> Since Travis can install Chrome at a run time you no longer need to install it manually. Read [more](https://docs.travis-ci.com/user/chrome).
```
brendankenny: maybe this line makes more sense right below the `curl` command?

brendankenny: actually, it's a little confusing to have the `curl` command above then not have it in the `travis.yml` snippet. What do  you think about rearranging to something like

```
In a CI environment like Travis, Chrome may not be installed. If you want to use `chrome-launcher`, Travis can [install Chrome at run time with an addon](https://docs.travis-ci.com/user/chrome).

> You can also install Chrome manually using Lighthouse's `download-chrome.sh` script:
> `curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash`
```

then just the snippet changes you have currently:

````
Then in `.travis.yml`, use it like so:

```yaml
language: node_js
install:
  - yarn install
before_script:
  - export DISPLAY=:99.0
  - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
  - sh -e /etc/init.d/xvfb start
  - sleep 3 # wait for xvfb to boot

addons:
  chrome: stable
```
````

File: chrome-launcher/README.md [hunk 3]
```diff
@@ -111,9 +111,10 @@ chromeLauncher.launch({
 
 ### Continuous Integration
 
-In a CI environment like Travis, Chrome may not be installed. If you want to use `chrome-launcher`, you can install Chrome using Lighthouse's `download-chrome.sh` script:
+In a CI environment, Chrome may not be installed. If you want to use `chrome-launcher`, e.g. Travis can [install Chrome at run time with an addon](https://docs.travis-ci.com/user/chrome).
```
denar90: @brendankenny I changed your message a bit... wdyt?

denar90: After rereading that, I realized that your one is better here, so I pushed new commit https://github.com/GoogleChrome/lighthouse/pull/3377/commits/4cb66353fede765e11ee45bbae8b6a85a44b9d60. Thanks :)

paulirish:
the download script will be moving to chrome-launcher so the path here isn't necessarily reliable. I just updated the chrome-launcher readme with a variant of this PR: https://github.com/GoogleChrome/chrome-launcher/blob/master/README.md#continuous-integration

might as well have the same thing?

chrome-launcher/README.md
```diff
@@ -125,5 +126,7 @@ before_script:
   - export DISPLAY=:99.0
   - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
   - sh -e /etc/init.d/xvfb start
-  - curl -L https://raw.githubusercontent.com/GoogleChrome/lighthouse/v2.1.0/lighthouse-core/scripts/download-chrome.sh | bash
+  - sleep 3 # wait for xvfb to boot
```
patrickhulce: I believe @paulirish recently removed this line as unnecessary, not sure if we should encourage using it more?

thoughts paul?

paulirish: i removed it because we dont need to use chrome in the next line of travis `script`... we have at least 3 seconds worth of work before we get there.

but i can imagine for other projects using chrome-launcher that they'd want this bit to be more reliable.

denar90:
Cool 👍  Closing then.

PR #3392, opened 2017-09-22
re-enable bundlesize reporting

paulirish:
The real fix is a permissions issue documented here: https://github.com/siddharthkp/bundlesize/issues/148

this is some cleanup.

Also removing a 3s sleep since it doesn't appear to be necessary.


patrickhulce:
a wild green travis appears 😃

PR #3413, opened 2017-09-26
Don't wait for Page.navigate's resolution

paulirish:
to repro:

```sh
yarn start -- --disable-device-emulation --disable-cpu-throttling --disable-network-throttling "http://example.com" --verbose
```

this will always timeout at 30 waiting for example.com

that's because it fires loadEventFired before page.navigate ever resolves:

image

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -622,9 +622,13 @@ class Driver {
     /* eslint-enable max-len */
 
     return this._beginNetworkStatusMonitoring(url)
-      .then(_ => this.sendCommand('Page.enable'))
-      .then(_ => this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS}))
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -622,9 +622,13 @@ class Driver {
     /* eslint-enable max-len */
 
     return this._beginNetworkStatusMonitoring(url)
-      .then(_ => this.sendCommand('Page.enable'))
-      .then(_ => this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS}))
-      .then(_ => this.sendCommand('Page.navigate', {url}))
+      .then(_ => {
+        // These can 'race' and that's OK.
+        // We don't want to wait for Page.navigate's resolution, as it can happen _after_ onload.
```
brendankenny: maybe add a bug reference

paulirish:
> nice detective work!

i noticed example.com was timing out. patrick picked up on the race issue here.

PR #3464, opened 2017-10-03
fix: update plots-config to use newer quiet parameters

brendankenny:
we missed these when updating in #2220, #2473, et al, which leads to lots of null TTCIs in plots runs :S


brendankenny:
Haha, I suspected as much

PR #3489, opened 2017-10-05
core(predictive-perf): refactor simulation logic

patrickhulce:
Addresses the remaining non-accuracy impacting items from code review.

* Moves to use WebInspector.resourceTypes
* Moves the bulk of the logic into lib/
* Renames estimator -> simulator
* Eliminates estimateGraph method from PageDependencyGraph artifact class
* Renames the node states in Simulator
* Refactors signature of simulateDownloadUntil

**TODO from https://github.com/GoogleChrome/lighthouse/pull/3162#discussion_r137640353** (Separately with analysis of impact to accuracy)
* Use all trace events that have a stack trace
* Improve the network initiator logic
* Remove safeguards in favor of `isValidDepGraph`


lighthouse-core/lib/dependency-graph/simulator/tcp-connection.js
```diff
@@ -89,11 +89,15 @@ class TcpConnection {
...
    *
    * @param {number} bytesToDownload
-   * @param {number=} timeAlreadyElapsed
-   * @param {number=} maximumTimeToElapse
+   * @param {{timeAlreadyElapsed: number, maximumTimeToElapse}=} options
    * @return {{timeElapsed: number, roundTrips: number, bytesDownloaded: number, congestionWindow: number}}
    */
-  simulateDownloadUntil(bytesToDownload, timeAlreadyElapsed = 0, maximumTimeToElapse = Infinity) {
+  simulateDownloadUntil(bytesToDownload, options) {
+    options = Object.assign({
```
paulirish: can we destructure AND do default values?

based on http://node.green/ i think we're good.

patrickhulce: totes

patrickhulce:
@brendankenny did you want a whack at this too, or you good?

lighthouse-core/audits/predictive-perf.js
```diff
@@ -123,13 +126,13 @@ class PredictivePerf extends Audit {
       let sum = 0;
       const values = {};
       Object.keys(graphs).forEach(key => {
-        const estimate = PageDependencyGraph.estimateGraph(graphs[key]);
-        const lastLongTaskEnd = PredictivePerf.getLastLongTaskEndTime(estimate.nodeTiming);
+        const results = new Simulator(graphs[key]).simulate();
```
brendankenny: I'd personally still prefer `estimate` for the local var here, as `results` is information-less :) or `simulatedResults` or something. `results` could be anything

lighthouse-core/audits/predictive-perf.js
```diff
@@ -7,8 +7,9 @@
 
 const Audit = require('./audit');
 const Util = require('../report/v2/renderer/util.js');
-const PageDependencyGraph = require('../gather/computed/page-dependency-graph.js');
-const Node = require('../gather/computed/dependency-graph/node.js');
+const Simulator = require('../lib/dependency-graph/simulator/simulator.js');
```
brendankenny: I guess `simulator.js`'s context can be figured out from its path, but maybe `LoadSimulator` or something for the local handle to it?

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -132,16 +139,17 @@ class Estimator {
...
   _initializeAuxiliaryData() {
     this._nodeTiming = new Map();
-    this._nodesUnprocessed = new Set();
-    this._nodesCompleted = new Set();
-    this._nodesInProgress = new Set();
-    this._nodesInQueue = new Set(); // TODO: replace this with priority queue
     this._connectionsInUse = new Set();
     this._numberInProgressByType = new Map();
+
+    this._nodes = {};
```
brendankenny: should be an array now?

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -320,8 +328,10 @@ class Estimator {
     const connection = this._connections.get(node.record.connectionId);
     const calculation = connection.simulateDownloadUntil(
       node.record.transferSize - timingData.bytesDownloaded,
-      timingData.timeElapsed,
-      timePeriodLength - timingData.timeElapsedOvershoot
+      {
```
brendankenny: wait, why is this moving to an options object? There's only three and having them named doesn't seem better in this case...

patrickhulce: I feel ya @brendankenny, but @paulirish was persistent

executive summary: it's easier for the caller to understand what the params are

lighthouse-core/audits/predictive-perf.js
patrickhulce: sg done

lighthouse-core/audits/predictive-perf.js

lighthouse-core/lib/dependency-graph/simulator/simulator.js
patrickhulce: wait why? I thought the whole point was to key these by their state 😆

brendankenny: > EDIT: oh just that the keys are numbers now so you'd prefer to have array?

yah

brendankenny:
rebase to pick up test fixes?

lighthouse-core/lib/dependency-graph/simulator/simulator.js
patrickhulce: done

PR #3492, opened 2017-10-06
report(redirects): reformat results, incl all requests and wasted time,

paulirish:
##### Changes:

* The cost of each redirect is shown next to the new/redirected URL rather than initial URL. (Makes more sense to my brain). This also means we now show the final URL in the table, whereas before it wasn't there.
* The very first initial url is shown for completeness. 
* (If there's just 1 redirect we currently pass them. Nothing changes here.)
* All costs of redirects are kept and summed. We don't skip the first one. 
* Even if there's 1 redirect and you pass, we will still show the cost of the redirects to be clear.
* I updated the test cases to be slightly more realistic.

##### screenshots

failing with 2 redirects: 
image

passing with 1 redirect: 
image

passing with 0 redirects:
image

### DIFF
--- a/lighthouse-core/audits/redirects.js
+++ b/lighthouse-core/audits/redirects.js
@@ -68,7 +68,7 @@ class Redirects extends Audit {
 
         return {
           // We award a passing grade if you only have 1 redirect
-          score: redirectRequests.length === 2 ? 100 : UnusedBytes.scoreForWastedMs(totalWastedMs),
+          score: redirectRequests.length <= 2 ? 100 : UnusedBytes.scoreForWastedMs(totalWastedMs),
           rawValue: totalWastedMs,
           displayValue: Util.formatMilliseconds(totalWastedMs, 1),
           extendedInfo: {
--- a/lighthouse-core/audits/redirects.js
+++ b/lighthouse-core/audits/redirects.js
@@ -32,38 +32,48 @@ class Redirects extends Audit {
       .then(mainResource => {
         // redirects is only available when redirects happens
         const redirectRequests = Array.from(mainResource.redirects || []);
+
         // add main resource to redirectRequests so we can use it to calculate wastedMs
         redirectRequests.push(mainResource);
+
         let totalWastedMs = 0;
-
         const pageRedirects = [];
-        for (let i = 1; i < redirectRequests.length; i++) {
-          const request = redirectRequests[i - 1];
-          const nextRequest = redirectRequests[i];
-          const wastedMs = (nextRequest.startTime - request.startTime) * 1000;
 
-          // We skip the first redirect in our calculations but show it in the table below.
-          // We allow 1 redirect (www. => m.)
-          if (i > 1) {
-            totalWastedMs += wastedMs;
+        for (let i = 0; i < redirectRequests.length; i++) {
+          const initialRequest = redirectRequests[i - 1];
+          const redirectedRequest = redirectRequests[i];
+
+          // Only create a table (with the initial request) if there are > 1 redirects
+          if (!initialRequest) {
+            if (redirectRequests.length > 1) {
+              pageRedirects.push({
+                url: `(Initial: ${redirectedRequest.url})`,
+                wastedMs: 'n/a',
+              });
+            }
+            continue;
           }
 
+          const wastedMs = (redirectedRequest.startTime - initialRequest.startTime) * 1000;
+          totalWastedMs += wastedMs;
+
           pageRedirects.push({
-            url: request.url,
-            wastedMs: Util.formatMilliseconds(wastedMs),
+            url: redirectedRequest.url,
+            wastedMs: Util.formatMilliseconds(wastedMs, 1),
           });
         }
 
         const headings = [
-          {key: 'url', itemType: 'text', text: 'URL'},
+          {key: 'url', itemType: 'text', text: 'Redirected URL'},
           {key: 'wastedMs', itemType: 'text', text: 'Time for Redirect'},
         ];
         const details = Audit.makeTableDetails(headings, pageRedirects);
 
         return {
-          score: UnusedBytes.scoreForWastedMs(totalWastedMs),
+          // We award a passing grade if you only have 1 redirect
+          score: redirectRequests.length === 2 ? 100 : UnusedBytes.scoreForWastedMs(totalWastedMs),
           rawValue: totalWastedMs,
-          displayValue: Util.formatMilliseconds(totalWastedMs),
+          displayValue: Util.formatMilliseconds(totalWastedMs, 1),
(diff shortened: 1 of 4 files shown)


lighthouse-core/audits/redirects.js
```diff
@@ -33,38 +33,48 @@ class Redirects extends Audit {
...
-          // We allow 1 redirect (www. => m.)
-          if (i > 1) {
-            totalWastedMs += wastedMs;
+        for (let i = 0; i < redirectRequests.length; i++) {
+          const initialRequest = redirectRequests[i - 1];
+          const redirectedRequest = redirectRequests[i];
+
+          // Only create a table (with the initial request) if there are > 1 redirects
+          if (!initialRequest) {
+            if (redirectRequests.length > 1) {
```
paulirish: This `>1` and a `=== 2` below could instead use a `ACCEPTED_REDIRECTS` const if folks prefer.

lighthouse-core/test/audits/redirects-test.js
```diff
@@ -9,41 +9,54 @@ const Audit = require('../../audits/redirects.js');
 const assert = require('assert');
 
 /* eslint-env mocha */
-const FAILING_REDIRECTS = {
+const FAILING_THREE_REDIRECTS = {
   startTime: 17,
+  url: 'http://exampel.com/',
```
wardpeet: typo example.com?

lighthouse-core/test/audits/redirects-test.js
```diff
@@ -9,41 +9,54 @@ const Audit = require('../../audits/redirects.js');
...
-      responseReceivedTime: 15,
       startTime: 12,
       url: 'https://m.example.com/',
     },
   ],
 };
 
+const FAILING_TWO_REDIRECTS = {
+  startTime: 446.286,
+  url: 'http://lisairish.com/',
```
wardpeet: perhaps better to just use example.com as an url so it's consistent with the rest?

Or is this a marketing stunt for your mom's website? 😆

brendankenny:
> was that expected?

ah, yes, it was:

> All costs of redirects are kept and summed. We don't skip the first one.

lighthouse-core/audits/redirects.js
```diff
@@ -33,38 +33,48 @@ class Redirects extends Audit {
...
-        for (let i = 1; i < redirectRequests.length; i++) {
-          const request = redirectRequests[i - 1];
-          const nextRequest = redirectRequests[i];
-          const wastedMs = (nextRequest.startTime - request.startTime) * 1000;
 
-          // We skip the first redirect in our calculations but show it in the table below.
-          // We allow 1 redirect (www. => m.)
-          if (i > 1) {
-            totalWastedMs += wastedMs;
+        for (let i = 0; i < redirectRequests.length; i++) {
```
brendankenny: the `if (!initialRequest)` part of this ends up being a little confusing. What about pulling it out, something like


```js
let totalWastedMs = 0;
const pageRedirects = [];

if (redirectRequests.length > 1) {
  pageRedirects.push({
    url: `(Initial: ${redirectRequests[0].url})`,
    wastedMs: 'n/a',
  });
}

for (let i = 1; i < redirectRequests.length; i++) {
  const initialRequest = redirectRequests[i - 1];
  const redirectedRequest = redirectRequests[i];

  const wastedMs = (redirectedRequest.startTime - initialRequest.startTime) * 1000;
  totalWastedMs += wastedMs;

  pageRedirects.push({
    url: redirectedRequest.url,
    wastedMs: Util.formatMilliseconds(wastedMs, 1),
  });
}
```
(I believe that should be equivalent)

patrickhulce: +1

paulirish: sg thx

lighthouse-core/audits/redirects.js
```diff
@@ -33,38 +33,48 @@ class Redirects extends Audit {
...
-          {key: 'url', itemType: 'text', text: 'URL'},
+          {key: 'url', itemType: 'text', text: 'Redirected URL'},
           {key: 'wastedMs', itemType: 'text', text: 'Time for Redirect'},
         ];
         const details = Audit.makeTableDetails(headings, pageRedirects);
 
         return {
-          score: UnusedBytes.scoreForWastedMs(totalWastedMs),
+          // We award a passing grade if you only have 1 redirect
+          score: redirectRequests.length === 2 ? 100 : UnusedBytes.scoreForWastedMs(totalWastedMs),
```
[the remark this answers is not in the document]
paulirish: With the new logic you can get `totalWastedMs > 0` and still get a passing grade.  Take http://www.lisairish.com which i was seeing it take 500ms to redirect to https://www.lisairish.com. We should still tell them about the 500ms even if we pass them.

lighthouse-core/test/audits/redirects-test.js
paulirish: Well i tested this with these actual URLs, which have this set of redirects setup.

lighthouse-core/test/audits/redirects-test.js
paulirish: this was deliberate. :)

paulirish:
updated! @brendankenny ptal.

brendankenny:
will you rebase to pick up the passive event listener fix?

paulirish:
@brendankenny thar u go

brendankenny:
@patrickhulce if you have any last issues

lighthouse-core/test/audits/redirects-test.js
```diff
@@ -60,19 +73,28 @@ const mockArtifacts = (mockChain) => {
...
 describe('Performance: Redirects audit', () => {
-  it('fails when more than one redirect detected', () => {
-    return Audit.audit(mockArtifacts(FAILING_REDIRECTS)).then(output => {
+  it('fails when 3 redirects detected', () => {
+    return Audit.audit(mockArtifacts(FAILING_THREE_REDIRECTS)).then(output => {
       assert.equal(output.score, 0);
+      assert.equal(output.details.items.length, 4);
+      assert.equal(output.rawValue, 17000);
+    });
+  });
```
patrickhulce: nit: add newline

lighthouse-core/audits/redirects.js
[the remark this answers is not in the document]
patrickhulce: could we add a passing test case for a resource with 0 redirects too? I had to pause for a second on this, alternatively switch to `redirectRequests.length <= 2`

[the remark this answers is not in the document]
paulirish: already have passing test case for 0 redirects.
changed to ` <= 2`.

PR #3497, opened 2017-10-06
 web-inspector: fall back to page's Runtime and queryParam()

paulirish:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "lighthouse",
-  "version": "2.5.0",
+  "version": "2.5.1",
```
brendankenny: need extension bump and changelog.md, too :)

paulirish: changelog updated. whitespace sorry

PR #3498, opened 2017-10-06
test(smokehouse): Passive event listener violation doesn't report on passive:false now

paulirish:
Fixes appveyor breakage


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -158,7 +158,8 @@ module.exports = [
           value: {
             // Note: This would normally be 7 but M56 defaults document-level
             // listeners to passive. See https://www.chromestatus.com/features/5093566007214080
-            length: 4,
+            // Note: Explicitly passive:false doesn't get a warning as of M63: crbug.com/770208
```
brendankenny: mention in the comment that this should be changed to 3 when m63 hits stable?

PR #3500, opened 2017-10-06
core(driver): use execution context isolation when necessary

### LINKED ISSUE #2505: Audit error: Required Accessibility gatherer encountered an error when using almond.js
We are using https://github.com/requirejs/almond in our builds instead of the whole requirejs package and this seems to break lighthouse accessibility tests. I've tried it both via the node package and via the Audit tab in Chrome Developer Tools.

    lighthouse https://bibliotek.jonkoping.se/

The result is a 0 score for accessibility and all tests have failed with the below message:

    Audit error: Required Accessibility gatherer encountered an error: See almond README: incorrect module build, no module name

Almond has 30 000 downloads the last week on npm and is used in lots of places around the web so a solution or at least a workaround that we can use would be great!

### LINKED ISSUE #2511: Move `evaluateAsync` calls to content scripts
A new protocol method `createIsolatedWorld` [landed in 59](https://chromium.googlesource.com/chromium/src/+blame/6ebcb95f8c4fadb29b69d2fbcbf8c964974e76aa/third_party/WebKit/Source/core/inspector/browser_protocol.json#535) that creates an execution sandbox that we can use to isolate ourselves from the page mucking with natives.

This should allow us to remove the `cacheNatives` method in driver and prevent miscellaneous bugs from strange interactions with the page.

patrickhulce:
closes #2511 
fixes #2505

### DIFF
--- a/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
+++ b/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
@@ -120,7 +120,7 @@
     <section>
       <form>
         <input
-          id="label" 
+          id="label"
           type="text">
       </form>
     </section>
@@ -197,5 +197,13 @@
     <section>
       <video id="video-description"></video>
     </section>
+    <script>
+      // axe fails if `define` is present but misbehaves, create a bogus implementation to exercise our
+      // execution context isolation workaround, see https://github.com/GoogleChrome/lighthouse/issues/2505
+      window.define = function () {
+        throw new Error('No AMD modules allowed');
+      };
+      window.define.amd = true;
+    </script>
   </body>
 </html>
--- a/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
+++ b/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
@@ -198,8 +198,8 @@
       <video id="video-description"></video>
     </section>
     <script>
-      // axe fails if `define` is present but misbehaves, create a bogus implementation to exercise our
-      // execution context isolation workaround, see https://github.com/GoogleChrome/lighthouse/issues/2505
+      // axe fails if a different UMD loader like almond.js is used on the page, see https://github.com/GoogleChrome/lighthouse/issues/2505
+      // We can simulate this behavior by defining a similarly misbehaving `define` function.
       window.define = function () {
         throw new Error('No AMD modules allowed');
       };
--- a/lighthouse-cli/test/smokehouse/a11y/a11y-config.js
+++ b/lighthouse-cli/test/smokehouse/a11y/a11y-config.js
@@ -1,0 +1,18 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/**
+ * Config file for running PWA smokehouse audits for axe.
+ */
+module.exports = {
+  extends: 'lighthouse:default',
+  settings: {
+    onlyCategories: [
+      'accessibility',
+    ],
+  },
+};
--- a/lighthouse-cli/test/smokehouse/run-all-tests.sh
+++ b/lighthouse-cli/test/smokehouse/run-all-tests.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+for d in "$DIR"/*/ ; do
+  bash "${d}run-tests.sh"
+  if [ $? -ne 0 ]
+  then
+    echo "Error: $d smoke test failed"
+    exit 1
+  fi
+done
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -41,6 +41,7 @@ class Driver {
     this._devtoolsLog = new DevtoolsLog(/^(Page|Network)\./);
     this.online = true;
     this._domainEnabledCounts = new Map();
+    this._isolatedExecutionContextId = undefined;
 
     /**
      * Used for monitoring network status events during gotoURL.
@@ -219,9 +220,11 @@ class Driver {
    * Evaluate an expression in the context of the current page.
    * Returns a promise that resolves on the expression's value.
    * @param {string} expression
+   * @param {{useIsolation: boolean}=} options
    * @return {!Promise<*>}
    */
-  evaluateAsync(expression) {
+  evaluateAsync(expression, options) {
+    const {useIsolation = true} = options || {};
     return new Promise((resolve, reject) => {
       // If this gets to 60s and it hasn't been resolved, reject the Promise.
       const asyncTimeout = setTimeout(
@@ -229,7 +232,7 @@ class Driver {
         60000
       );
 
-      this.sendCommand('Runtime.evaluate', {
+      const evaluationParams = {
         // We need to explicitly wrap the raw expression for several purposes:
         // 1. Ensure that the expression will be a native Promise and not a polyfill/non-Promise.
         // 2. Ensure that errors in the expression are captured by the Promise.
@@ -247,7 +250,13 @@ class Driver {
         includeCommandLineAPI: true,
         awaitPromise: true,
         returnByValue: true,
-      }).then(result => {
+      };
+
+      if (useIsolation && typeof this._isolatedExecutionContextId === 'number') {
+        evaluationParams.contextId = this._isolatedExecutionContextId;
+      }
+
+      this.sendCommand('Runtime.evaluate', evaluationParams).then(result => {
         clearTimeout(asyncTimeout);
         const value = result.result.value;
 
@@ -431,10 +440,12 @@ class Driver {
 
     let lastTimeout;
     let cancelled = false;
+
+    const checkForQuietExpression = `(${checkTimeSinceLastLongTask.toString()})()`;
     function checkForQuiet(driver, resolve) {
       if (cancelled) return;
 
-      return driver.evaluateAsync(`(${checkTimeSinceLastLongTask.toString()})()`)
+      return driver.evaluateAsync(checkForQuietExpression, {useIsolation: false})
         .then(timeSinceLongTask => {
           if (cancelled) return;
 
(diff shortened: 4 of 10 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -247,7 +250,13 @@ class Driver {
         includeCommandLineAPI: true,
         awaitPromise: true,
         returnByValue: true,
-      }).then(result => {
+      };
+
+      if (useIsolation && typeof this._isolatedExecutionContextId === 'number') {
```
paulirish: if isolation is requested but there is no contextId... should we throw?

lighthouse-cli/test/smokehouse/run-all-tests.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+for d in "$DIR"/*/ ; do
+  bash "${d}run-tests.sh"
```
paulirish: there are a few other ways to do this, but the benefit of this approach is the failures won't be silent. :)

sgtm

lighthouse-core/gather/driver.js
patrickhulce: I had this originally to check for leaks and it seemed ok, but it's definitely possible and seemed bad to tank the run because it since it won't matter most of the time

perhaps something to add to sentry though when it lands?

lighthouse-cli/test/smokehouse/run-all-tests.sh
brendankenny: can we just manually list the files in here? If I'm messing with some smoke tests it's a lot easier to comment out lines in one place by name than look up where the individual run-tests file is

also it'd probably be better to just get rid of all the `run-tests.sh` files since they're all exactly the same except for the expectations and config filename strings :)

lighthouse-core/gather/driver.js
```diff
@@ -219,17 +220,19 @@ class Driver {
    * Evaluate an expression in the context of the current page.
    * Returns a promise that resolves on the expression's value.
    * @param {string} expression
+   * @param {{useIsolation: boolean}=} options
    * @return {!Promise<*>}
    */
-  evaluateAsync(expression) {
+  evaluateAsync(expression, options) {
+    const {useIsolation = true} = options || {};
```
brendankenny: the double defaults here is kind of confusing. Maybe move the `{}` default for `options` to the params?

lighthouse-core/gather/driver.js
brendankenny: Seems like we should be able to figure out definitively if it's possible to reach here without having created a new context? Then future violations of this can complain loudly by throwing (alternatively, see other comment about lazily calling `_createIsolatedWorld()`)

lighthouse-core/gather/driver.js
```diff
@@ -633,6 +661,7 @@ class Driver {
       })
       .then(_ => waitForLoad && this._waitForFullyLoaded(pauseAfterLoadMs,
           networkQuietThresholdMs, cpuQuietThresholdMs, maxWaitMs))
+      .then(_ => this._createIsolatedWorld())
```
brendankenny: Would there be downsides to clearing on all navigations (as above) but calling `_createIsolatedWorld` lazily on next `evaluateAsync()` when `_isolatedExecutionContextId` isn't defined?

lighthouse-core/gather/driver.js
patrickhulce: going the lazy route 👍

lighthouse-core/gather/driver.js
patrickhulce: done

lighthouse-core/gather/driver.js
patrickhulce: done

lighthouse-cli/test/smokehouse/run-all-tests.sh
patrickhulce: will nuke 'em all in a follow up PR and improve this file here 👍

patrickhulce:
> Is there anything we can do for these non-isolated ones (js-libraries, function callsites, and check for quiet) to make them isolated? I guess it's just unclear what's isolated and what's not if the axe checks work but js-library-detector does not

discussed offline, decided to switch the default to useIsolation=false and just opt-in a11y for now

patrickhulce:
PTAL :)

lighthouse-cli/test/fixtures/a11y/a11y_tester.html
```diff
@@ -197,5 +197,13 @@
     <section>
       <video id="video-description"></video>
     </section>
+    <script>
+      // axe fails if `define` is present but misbehaves, create a bogus implementation to exercise our
```
paulirish: I don't totally get this. I'll ask you in person if i'm interpreting it correctly, but it could probably use a slightly more extended comment. :)

lighthouse-cli/test/smokehouse/a11y/a11y-config.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Config file for running PWA smokehouse audits for axe.
+ */
+module.exports = {
+  extends: 'lighthouse:default',
+  settings: {
+    onlyCategories: [
+      'accessibility',
+    ],
```
paulirish: since we were talking about it yesterday... is it possible to reuse an existing smoketest?

lighthouse-core/gather/driver.js
```diff
@@ -216,20 +217,37 @@ class Driver {
...
+   * will be evaluated in a content script that has access to the page's DOM but whose JavaScript state
+   * is completely separate.
    * Returns a promise that resolves on the expression's value.
    * @param {string} expression
+   * @param {{useIsolation: boolean}=} options
    * @return {!Promise<*>}
    */
-  evaluateAsync(expression) {
+  evaluateAsync(expression, options = {}) {
+    const contextIdPromise = options.useIsolation ?
```
paulirish: what about 

```js
    const fn = options.useIsolation ? this._getOrCreateIsolatedContextId : Promise.resolve;
    return fn().then(contextId => this._evaluateInContext(expression, contextId));
```

not a huge deal . w/e

lighthouse-core/gather/driver.js
```diff
@@ -596,6 +619,26 @@ class Driver {
   }
 
   /**
+   * @return {!Promise<number>}
+   */
+  _getOrCreateIsolatedContextId() {
```
paulirish: `_acquireIsolatedContextId()`  ?

lighthouse-core/gather/driver.js
```diff
@@ -596,6 +619,26 @@ class Driver {
...
+    if (typeof this._isolatedExecutionContextId === 'number') {
+      return Promise.resolve(this._isolatedExecutionContextId);
+    }
+
+    return this.sendCommand('Page.getResourceTree')
+      .then(data => {
+        const frameId = data.frameTree.frame.id;
+        return this.sendCommand('Page.createIsolatedWorld', {frameId});
```
paulirish: let's give it a `worldName` while we're at it.

lighthouse-core/gather/driver.js
```diff
@@ -596,6 +619,26 @@ class Driver {
...
+   * @return {!Promise<number>}
+   */
+  _getOrCreateIsolatedContextId() {
+    if (typeof this._isolatedExecutionContextId === 'number') {
+      return Promise.resolve(this._isolatedExecutionContextId);
+    }
+
+    return this.sendCommand('Page.getResourceTree')
+      .then(data => {
+        const frameId = data.frameTree.frame.id;
```
paulirish: frameId => mainFrameId

lighthouse-core/gather/driver.js
paulirish: nevermind this one. bad idea.

lighthouse-core/gather/driver.js
```diff
@@ -216,20 +217,37 @@ class Driver {
   }
 
   /**
-   * Evaluate an expression in the context of the current page.
+   * Evaluate an expression in the context of the current page. If useIsolation is true, the expression
+   * will be evaluated in a content script that has access to the page's DOM but whose JavaScript state
+   * is completely separate.
```
brendankenny: 👍 👍

lighthouse-core/gather/driver.js
```diff
@@ -216,20 +217,37 @@ class Driver {
...
+        this._getOrCreateIsolatedContextId() :
+        Promise.resolve(undefined);
+    return contextIdPromise.then(contextId => this._evaluateInContext(expression, contextId));
+  }
+
+  /**
+   * Evaluate an expression in the given execution context; an undefined contextId implies the main
+   * page without isolation.
+   * @param {string} expression
+   * @param {number|undefined} contextId
```
brendankenny: can be `{number=}` FWIW

lighthouse-cli/test/smokehouse/a11y/a11y-config.js
patrickhulce: I know we were literally just talking about not adding more yesterday but of all the testers that should probably get their own one, I think a11y is one of them 😆  

maybe we can start by reducing DBW's 6 passes instead? :)

lighthouse-core/gather/driver.js
brendankenny: maybe a jsdoc comment here that the contextId won't persist past a page load (so should be cleared and recreated)?

lighthouse-cli/test/fixtures/a11y/a11y_tester.html
patrickhulce:
```
// axe fails if a different UMD loader like almond.js is used on the page, see https://github.com/GoogleChrome/lighthouse/issues/2505
// We can simulate this behavior by defining a similarly misbehaving `define` function.
```

sounds better?

lighthouse-core/gather/driver.js
patrickhulce: discussed offline, Paul's a huge fan of `getOrCreate` now 🚗  🎏  🎉

lighthouse-core/gather/driver.js
patrickhulce: done

lighthouse-core/gather/driver.js
patrickhulce: done

lighthouse-core/gather/driver.js
patrickhulce: yeah just seemed more explicit since something is always passed in but doesn't matter to me

lighthouse-core/gather/driver.js
patrickhulce: good call, done

lighthouse-cli/test/fixtures/a11y/a11y_tester.html
paulirish: YESSS so good

lighthouse-core/gather/driver.js
paulirish: lol

PR #3501, opened 2017-10-07
Update readme.md

vinamratasingal:
Added Treo to the list of Lighthouse based projects


brendankenny:
Should we combine with #3484?

vinamratasingal:
Yup. Feel free to ignore this PR. We can go with the other one.

PR #3520, opened 2017-10-10
new_audit(mainthread-work-breakdown): Audit for page-execution-timings

### LINKED ISSUE #3105: audit: JavaScript boot-up time is high ⏱
**Description:** JavaScript boot-up time is high (> 4s)
**Failure text:** JavaScript boot-up time was 12s. This is too high.
**Help text:** Consider reducing the time spent parsing, compiling and executing JS. You may find delivering smaller JS payloads helps with this.

Overall: Highlight issues with parse/compile/execution time and point to the Performance panel for digging into this deeper.

Related issues: #2535 
Could be a proper version of/redux of: https://github.com/GoogleChrome/lighthouse/pull/1695

wardpeet:
Fixes #3105

Credits go to @addyosmani to do the heavy lifting https://github.com/GoogleChrome/lighthouse/pull/1695


wardpeet:
I'm not sure how to unit test this? should I record a trace and run it against it or mock devtoolsModel?

Mocking devtoolsModel looks easier

wardpeet:
> I'd prefer to make sure this includes everything. But in order to help clarify script costs.. I see two solutions here:
1) Add a TaskCategory column that would help clarify Script vs Paint, etc.
2) Add a color chip next to the titles.. All yellow/orange would be quick to notice.

How do I add a color ship? Change the category renderer?

wardpeet:
Example report:
paulirish.com
image

m.nieuwsblad.be
image

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * @param {!Artifacts} artifacts
+   * @return {!AuditResult}
+   */
+  static audit(artifacts) {
+    const trace = artifacts.traces[PageExecutionTimings.DEFAULT_PASS].traceEvents;
+    const executionTimings = PageExecutionTimings.getExecutionTimingsByCategory(trace);
+    let totalExecutionTime = 0;
+
+    const results = timelineCategories.map(category => {
```
paulirish: feels better to loop over your real results instead of iterating the lookup.
with the categories i shared above.. you have constant time lookup:

```js
Array.from(executionTimings).map(([eventName, duration]) => {
    totalExecutionTime += duration;
    return { 
      category: eventName,
      group: taskToGroup[eventName],
      duration: Util.formatMilliseconds(duration, 1),
    };
});
```

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,161 @@
...
+ * @fileoverview Audit a page to see if it does not use <link> that block first paint.
+ */
+
+'use strict';
+
+const Audit = require('./audit');
+const Util = require('../report/v2/renderer/util');
+const DevtoolsTimelineModel = require('../lib/traces/devtools-timeline-model');
+
+const timelineCategories = [
```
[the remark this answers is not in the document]
paulirish: I've just updated https://docs.google.com/spreadsheets/d/1jqbSfdkB5RMtuifJwQsGb70WUUhWvzir7hq41wYHXsU/edit#gid=0 with the latest and greatest.

and for convenience.. here's the categories lookup:

```js
const group = {
  loading: 'Network request loading',
  parseHTML: 'Parsing DOM',
  styleLayout: 'Style & Layout',
  compositing: 'Compositing',
  painting: 'Paint',
  gpu: 'GPU',
  scripting: 'Script Evaluation',
  scriptParseCompile: 'Script Parsing & Compile',
  scriptGC: 'Garbage collection',
  other: 'Other',
  images: 'Images',
};

const taskToGroup = {
  'Animation': group.painting,
  'Async Task': group.other,
  'Frame Start': group.painting,
  'Frame Start (main thread)': group.painting,
  'Cancel Animation Frame': group.scripting,
  'Cancel Idle Callback': group.scripting,
  'Compile Script': group.scriptParseCompile,
  'Composite Layers': group.compositing,
  'Console Time': group.scripting,
  'Image Decode': group.images,
  'Draw Frame': group.painting,
  'Embedder Callback': group.scripting,
  'Evaluate Script': group.scripting,
  'Event': group.scripting,
  'Animation Frame Fired': group.scripting,
  'Fire Idle Callback': group.scripting,
  'Function Call': group.scripting,
  'DOM GC': group.scriptGC,
  'GC Event': group.scriptGC,
  'GPU': group.gpu,
  'Hit Test': group.compositing,
  'Invalidate Layout': group.styleLayout,
  'JS Frame': group.scripting,
  'Input Latency': group.scripting,
  'Layout': group.styleLayout,
  'Major GC': group.scriptGC,
  'DOMContentLoaded event': group.scripting,
  'First paint': group.painting,
  'FMP': group.painting,
  'FMP candidate': group.painting,
  'Load event': group.scripting,
  'Minor GC': group.scriptGC,
  'Paint': group.painting,
  'Paint Image': group.images,
  'Paint Setup': group.painting,
  'Parse Stylesheet': group.parseHTML,
  'Parse HTML': group.parseHTML,
  'Parse Script': group.scriptParseCompile,
  'Other': group.other,
  'Rasterize Paint': group.painting,
  'Recalculate Style': group.styleLayout,
  'Request Animation Frame': group.scripting,
  'Request Idle Callback': group.scripting,
  'Request Main Thread Frame': group.painting,
  'Image Resize': group.images,
  'Finish Loading': group.loading,
  'Receive Data': group.loading,
  'Receive Response': group.loading,
  'Send Request': group.loading,
  'Run Microtasks': group.scripting,
  'Schedule Style Recalculation': group.styleLayout,
  'Scroll': group.compositing,
  'Task': group.other,
  'Timer Fired': group.scripting,
  'Install Timer': group.scripting,
  'Remove Timer': group.scripting,
  'Timestamp': group.scripting,
  'Update Layer': group.compositing,
  'Update Layer Tree': group.compositing,
  'User Timing': group.scripting,
  'Create WebSocket': group.scripting,
  'Destroy WebSocket': group.scripting,
  'Receive WebSocket Handshake': group.scripting,
  'Send WebSocket Handshake': group.scripting,
  'XHR Load': group.scripting,
  'XHR Ready State Change': group.scripting,
}
```

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,161 @@
...
+        'You may find delivering smaller JS payloads helps with this.',
+      requiredArtifacts: ['traces'],
+    };
+  }
+
+  /**
+   * @param {!Array<TraceEvent>=} trace
+   * @param {!WebInspector.TimelineProfileTree.Node} A grouped and sorted tree
+   */
+  static getTimingsByCategory(trace) {
```
paulirish: this is pretty straightforward. you might as well inline this stuff into your `getExecutionTimingsByCategory` method

lighthouse-core/audits/page-execution-timings.js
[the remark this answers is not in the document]
addyosmani: That is one super-handy category break-down!

lighthouse-core/audits/page-execution-timings.js
wardpeet: will switch back to the old way, I tried override that methode to mock DevtoolsTimeLinemodel

lighthouse-core/audits/page-execution-timings.js
wardpeet: nice! much better actually 👍

lighthouse-core/audits/page-execution-timings.js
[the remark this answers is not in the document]
wardpeet: where do you get that list when I capture a trace I can only find a handful of them

lighthouse-core/test/audits/page-execution-timings-test.js
```diff
@@ -0,0 +1,76 @@
...
+'use strict';
+
+const PageExecutionTimings = require('../../audits/page-execution-timings.js');
+const fs = require('fs');
+const assert = require('assert');
+
+// sadly require(file) is not working correctly.
+// traceParser parser returns preact trace data the same as JSON.parse
+// fails when require is used
+const readFile = (file, cb) => {
```
patrickhulce: and `readFileSync` doesn't work either!? geez what's going on with these files 😆

wardpeet: Readfilesync prolly works too. I bet devtoolstimeline changes the trace

patrickhulce: Ah ok, `require` caches the same object so you're probably right, could we switch to readfilesync then for brevity?

wardpeet:
I think this audit is good to go, only have my doubts about the meta data.
score is always false so we show it as a perf Diagnostic
image

benschwarz:
@wardpeet - “executiontime” should be two words 👍

Looking really, really good otherwise.

wardpeet:
@brendankenny should I add it to any of the real url smoketests?

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,171 @@
...
+      requiredArtifacts: ['traces'],
+    };
+  }
+
+  /**
+   * @param {!Array<TraceEvent>=} trace
+   * @return {!Map<string, Number>}
+   */
+  static getExecutionTimingsByCategory(trace) {
+    const timelineModel = new DevtoolsTimelineModel(trace);
```
brendankenny: is it possible to limit to events before TTCI? Not sure how to do this with `DevtoolsTimelineModel`

lighthouse-core/test/audits/page-execution-timings-test.js
```diff
@@ -0,0 +1,65 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/* eslint-env mocha */
+const PageExecutionTimings = require('../../audits/page-execution-timings.js');
+const fs = require('fs');
+const assert = require('assert');
+
+// sadly require(file) is not working correctly.
```
brendankenny: this is a bit worrying because it means DevtoolsTimelineModel will be changing the main trace as well. Looking at the test diff, it looks like it's only changing column/line numbers (to make them 0-indexed) but that's still not great :)

Talking here we were thinking about either `Object.freeze` or cloning options. Not sure what's best from a performance perspective, though.

brendankenny:
> @brendankenny should I add it to any of the real url smoketests?

That would be interesting :) but flakiness would definitely be an issue. I wonder if any of our existing tester pages have reliable amounts of time in different categories (at least `>100` or whatever).

If that's not super reliable, though, we do already have some decent smoke coverage that traces are being taken and are valid, so I'd say throwing in four or five more of our other saved traces into the unit test would cover it pretty well.

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,171 @@
...
+};
+
+class PageExecutionTimings extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
+      name: 'page-execution-timings',
```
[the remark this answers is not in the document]
paulirish: mainthread-work-breakdown ?

lighthouse-core/audits/page-execution-timings.js
```diff
@@ -0,0 +1,171 @@
...
+      description: 'Page execution time is high',
```
paulirish: Main thread work breakdown

lighthouse-core/audits/page-execution-timings.js
paulirish: discussed offline. this could be done in a followup.. it wouldn't make a huge difference though it will be nice.

lighthouse-core/test/audits/page-execution-timings-test.js
wardpeet: yes exactly just numbers are changing.

cloning would probably take a while as you'll need a deep clone which can be pretty heavy. Will object.freeze work with timeline as they won't be able to edit the values.

I can poke around to see what works

lighthouse-core/audits/mainthread-work-breakdown.js
```diff
@@ -0,0 +1,172 @@
...
+        duration: Util.formatMilliseconds(duration, 1),
+      };
+    });
+
+    const headings = [
+      {key: 'category', itemType: 'text', text: 'Category'},
+      {key: 'group', itemType: 'text', text: 'Task Category'},
+      {key: 'duration', itemType: 'text', text: 'Time spent'},
+    ];
+    const tableDetails = PageExecutionTimings.makeTableDetails(headings, results);
```
paulirish: can you sort this data?

the bottomupbyname was already sorted by Duration so we dont need to do that, but let's sort by each `group`'s total time spent.

so basically we'll have all the Script Evaluation work (as that will likely be the top one), with its tasks then sorted in decreasing order of time, then the rest of the groups.

paulirish: before and after: 
image

(different data)

paulirish: here's the patch for all this work:

```diff
diff --git a/lighthouse-core/audits/mainthread-work-breakdown.js b/lighthouse-core/audits/mainthread-work-breakdown.js
index 0dd563e4..c3fd7461 100644
--- a/lighthouse-core/audits/mainthread-work-breakdown.js
+++ b/lighthouse-core/audits/mainthread-work-breakdown.js
@@ -139,22 +139,31 @@ class PageExecutionTimings extends Audit {
     let totalExecutionTime = 0;
 
     const extendedInfo = {};
+    const categoryTotals = {};
+
     const results = Array.from(executionTimings).map(([eventName, duration]) => {
       totalExecutionTime += duration;
       extendedInfo[eventName] = duration;
+      const groupName = taskToGroup[eventName];
+
+      const catTotal = categoryTotals[groupName] || 0;
+      categoryTotals[groupName] = catTotal + duration;
 
       return {
         category: eventName,
-        group: taskToGroup[eventName],
+        group: groupName,
         duration: Util.formatMilliseconds(duration, 1),
       };
     });
 
     const headings = [
-      {key: 'category', itemType: 'text', text: 'Category'},
-      {key: 'group', itemType: 'text', text: 'Task Category'},
+      {key: 'group', itemType: 'text', text: 'Category'},
+      {key: 'category', itemType: 'text', text: 'Work'},
       {key: 'duration', itemType: 'text', text: 'Time spent'},
     ];
+
+    results.stableSort((a, b) => categoryTotals[b.group] - categoryTotals[a.group]);
     const tableDetails = PageExecutionTimings.makeTableDetails(headings, results);
 
     return {

```

i swear at some point ill stop editing your PRs but i enjoy them so much i want to try some things out! :)  sorry!

wardpeet: no problem! Makes my job much easier anyway so keep editing away! 😄 💯

addyosmani:
Let's get a check in on whether this is good to go from @paulirish when he returns. I _think_ this was good to go the last sprint review we completed.

brendankenny:
We should definitely prioritize #3702

PR #3536, opened 2017-10-12
core(emulation): remove use of deprecated Emulation.setVisibleSize

### LINKED ISSUE #2831: Do not use Emulation.setVisibleSize
It's not needed anymore, since setDeviceMetricsOverride does it automatically.
We are going to remove setVisibleSize soon, so we better stop calling it from lighthouse.

------

EDIT: dmitry indicates this was removed in m61, but we need to verify that.
verified that the protocol changes are in 61.0.3155.0

paulirish:
This protocol method was made redundant in 61.0.3155.0

It's a noop for us now. Trace screenshots are no different.

fixes #2831

### DIFF
--- a/lighthouse-core/lib/emulation.js
+++ b/lighthouse-core/lib/emulation.js
@@ -104,11 +104,6 @@ function enableNexus5X(driver) {
 
   return Promise.all([
     driver.sendCommand('Emulation.setDeviceMetricsOverride', NEXUS5X_EMULATION_METRICS),
-    // required for screenshotting emulated page size, rather than full size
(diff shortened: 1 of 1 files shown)


patrickhulce:
Is 61 fully out now? It's been pretty slow going

paulirish:
> Is 61 fully out now? It's been pretty slow going


Shipped on sept 5th! 
stable is really only a concern for our extension users, but not our devtools people. and typically not CLI, though we don't know for sure.

chrome doesn't support stable-1 releases and i don't think we should either. 😀

PR #3537, opened 2017-10-12
docs(assets): Add example scoring graph

vinamratasingal:
Add graph that explains median and optimal value on the distribution.


brendankenny:
didn't we want to drop the "optimal value" phrase? Not sure if we have an alternative, though. "Point of diminishing returns" still has a certain ring to it :)

patrickhulce:
do we want to show the graph anywhere or is this just putting it there for later use?

+1 to brendan's point on moving away from "optimal value". I'm not a huge fan of "diminishing returns" externally either though. Suggesting there's no point getting better seems bad, if only there were a short pithy phrase for "you're already doing really really well, so you shouldn't beat yourself up about needing to get better. if there's more important work for you to be doing, you could stop here, but if you want to keep improving anyway that'd be totally awesome :D"

vinamratasingal:
Hmm good points team. I guess if we want to move away from optimal values, why don't we not even explain them within the doc?

PR #3552, opened 2017-10-13
core(image-aspect-ratio): pass audit when no images are missized

### LINKED ISSUE #3551: Audit - aspect ratio is always throwing an error
lighthouse 2.5.1 audit tests.

the Does not uses Images with appropriate aspect ratio is always failing. This test does not tell you which file is failing the test.  We then took all all of the images from the page and we still get this error.  We had to turn off our lighthouse test for the time because of this.  Here is an example page
https://jcdalton2201.github.io/progressive-app-test/ we have one img and its 80 X 80  the nat is 150 X 150.

brendankenny:
also add smoke test for audit

fixes #3551


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -36,13 +36,13 @@ module.exports = [
         extendedInfo: {
           value: {
             results: {
-              length: 14,
+              length: 15,
```
brendankenny: additional request for image in page

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -185,6 +185,20 @@ module.exports = [
           },
         },
       },
+      'image-aspect-ratio': {
```
brendankenny: dbw seemed like the best place for this since it's already running all of Best Practices, but happy to move

PR #3555, opened 2017-10-13
core(service-workers): sub-target message passing enabled. WIP

### LINKED ISSUE #709: Update throttling/cache/js state on the SW target
In [Developer Diary #Day1 - YouTube](https://www.youtube.com/watch?annotation_id=annotation_3128105329&feature=iv&src_vid=L50YZrpcEqA&v=CubrYp4cuyY) @paullewis ran into a weird behavior that the Service Worker wasnt going offline, despite Lighthouse making the page offline.  It was filed as [crbug 648208](https://bugs.chromium.org/p/chromium/issues/detail?id=648208), and duped into a WontFix, as the devtools frontend handles this case decently well. The frontend will connect to both targets and send the network throttling commands to each.

In our current behavior ...
- sw network requests are unaffected by the throttling
- sw doesn't go offline when the page goes offline.

Currently Lighthouse only sends the throttling stuff to the page's target. It'd make sense to also connect to the SW and flip it offline etc.

paulirish:
For early review. I have a bit more to do, but in case you're interested in big changes.. speak now. :)

fixes #709


lighthouse-core/gather/connections/sw.js
```diff
@@ -0,0 +1,47 @@
...
+
+class SWSession extends Connection {
+  constructor(sessionId, targetInfo, parentConnection) {
+    super();
+    this._targetInfo = targetInfo;
+    this._sessionId = sessionId;
+    this._parent = parentConnection;
+
+    this._parent.on('notification', this._onEvent.bind(this));
+    // now we repeat and do this again. we need to autoattach and grab a type: `worker` within
```
patrickhulce: I thought I understood what's going on until I read this comment, isn't the attaching within happening in the `onTargetAttached` over in manager where you create manager in a manager?

lighthouse-core/gather/connections/sw.js
```diff
@@ -0,0 +1,47 @@
...
+const Connection = require('./connection.js');
+
+class SWSession extends Connection {
+  constructor(sessionId, targetInfo, parentConnection) {
+    super();
+    this._targetInfo = targetInfo;
+    this._sessionId = sessionId;
+    this._parent = parentConnection;
+
+    this._parent.on('notification', this._onEvent.bind(this));
```
patrickhulce: my only gripe with the overall setup is this `notification` listener bit since it feels like a hacky way to override a method, can it not be done with regular override and super call?

lighthouse-core/gather/connections/sw-manager.js
```diff
@@ -0,0 +1,62 @@
...
+    if (targetInfo.type !== this._requiredType) return;
+
+    // setup message passing to wrap/unwrap Target messages
+    const session = new SWSession(sessionId, targetInfo, this._connection);
+
+    // `service_worker` targets are phantom pages. They have a `worker` target within that we want
+    let manager;
+    if (this._requiredType === 'service_worker') {
+      manager = new SWManager(session, {requiredType: 'worker'});
+      manager.listen();
```
brendankenny: my main question is if we could flatten this out and lose the recursive step.

If I'm reading/understanding this correctly (big if), it seems like it's always `service_worker` -> `worker`, and then the only thing the `service_worker` target is kept around (on the LH side) is for dispatching to the `worker`. Can we ditch a reference to the `service_worker` manager and instead listen for the follow up `worker`, and then only keep that around?

lighthouse-core/gather/connections/sw-manager.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  sendCommandToSubTargets(method, params) {
+    const _flatten = arr => [].concat(...arr);
+
+    const p = this._subTargets.map(({session, manager}) => {
+      if (manager) return manager.sendCommandToSubTargets(method, params);
+      else return session.sendCommand(method, params);
+    });
+    return Promise.all(p).then(arr => _flatten(arr));
```
brendankenny: it seems like this is set up to return an array of results...but what are all the results going to be coming from? Can we only attach to an active SW (so there should only be one)?

lighthouse-core/gather/connections/sw-manager.js
```diff
@@ -0,0 +1,97 @@
...
+      .sendCommand('Target.setAutoAttach', opts)
+      .then(_ => this._connection.sendCommand('Runtime.runIfWaitingForDebugger'));
+  }
+
+  sendCommandToSubtargets(method, params) {
+    if (this._subtargets.length === 0) {
+      console.warn(`No subtargets yet for ${method}`);
+    }
+    // Some protocol domains are not supported by the dedicated worker but by the SW page
+    // Note: Log and Network are technically supported in both, but not completely,
```
wardpeet: any documentation about this? What and what not is supported?

paulirish:
Shelving this until the new target selection support is landed in Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=775132

PR #3563, opened 2017-10-14
new_audit(bootup-time): Bootup time per script

wardpeet:
show bootup time per script

Fixes part of #3105


wardpeet:
Devtools is not giving me parse time of scripts :(
I  can look in the trace for v8.parseOnBackground?

Also when scripts are not moved to v8.parseOnBackground than I can't seem to find any parse event.

any idea @paulirish ?

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,97 @@
...
+    const timelineModel = new DevtoolsTimelineModel(trace);
+    const bottomUpByName = timelineModel.bottomUpGroupBy('URL');
+    const result = new Map();
+    bottomUpByName.children.forEach((value, url) => {
+      // when url is "", we skip it
+      if (!url) {
+        return;
+      }
+
+      const evaluateTime = value.children.get('EvaluateScript:@' + url) || {};
```
paulirish: we could do this but i'd rather iterate over the `perUrlNode` children (which would be `perTaskPerUrlNode` i guess)...

so iterate over them.. then grab  `perTaskPerUrlNode.event.name`. This should give you `v8.compile` and `EvaluateScript`. We can then reuse this stuff: https://github.com/GoogleChrome/lighthouse/blob/5fbff87ef9c8c18d51d05e6a07f5631d6edde8b6/lighthouse-core/audits/page-execution-timings.js#L17-L98

and we'll want any children with events of name in the 3 `script*`categories. (also note: many of those task names are for instant events so they have 0 duration. ;)

paulirish: we can exclude GC from the report for now but might as well collect it here and put in extInfo ?  or ignore it and leave a comment saying so..

paulirish:
> Also when scripts are not moved to v8.parseOnBackground than I can't seem to find any parse event.

perhaps they are on a different thread? (`tid`)? that would usually exclude them.

wardpeet:
yes they are so that's why I can't find it but do I need show parse time?

wardpeet:
> perhaps they are on a different thread? (tid)? that would usually exclude them.
yes indeed, not on the main thread which is good but should I also try to include it in the report?

Not all scripts are parsed on the parser thread (only the once that have async/defer). When they're not it seems like there is no parse time registered in the trace, it's probably included in compile time of the script?

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,97 @@
...
+
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
+      {key: 'evaluate', itemType: 'text', text: 'Time To Evaluate'},
+      {key: 'compile', itemType: 'text', text: 'Time To Compile'},
+    ];
+    const tableDetails = BootupTime.makeTableDetails(headings, results);
+
+    return {
+      score: totalBootupTime < 4000,
```
patrickhulce: 4s is quite generous, did we consider using a numeric instead of binary score to allow nuance?

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,97 @@
...
+      requiredArtifacts: ['traces'],
+    };
+  }
+
+  /**
+   * @param {!Array<TraceEvent>=} trace
+   * @return {!Map<string, Number>}
+   */
+  static getExecutionTimingsByURL(trace) {
+    const timelineModel = new DevtoolsTimelineModel(trace);
```
patrickhulce: are we doing this in a lot of places/should it be a computed artifact?

wardpeet: Are you referring to make a computed artifact `requestExecutionTimingsByUrl(trace)` or rather have a computed artifact `requestExecutionTimings(trace, 'URL')`

paulirish: the idea is to make DTM a computed artifact.

we were also discussing doing a deepFreeze (from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) on the trace data before sending it into devtools.

so it wont mutate.

wardpeet: should i do it in this PR? or just a followup?

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,203 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Audit = require('./audit');
+const DevtoolsTimelineModel = require('../lib/traces/devtools-timeline-model');
+const WebInspector = require('../lib/web-inspector');
+const Util = require('../report/v2/renderer/util.js');
+
+const group = {
```
brendankenny: could pull these two objects into a `lib/traces/trace-event-groups.js` or something so that they can be used across audits

lighthouse-core/test/audits/bootup-time-test.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+    };
+
+    const output = BootupTime.audit(artifacts);
+    assert.equal(output.details.items.length, 8);
+    assert.equal(output.score, true);
+    assert.equal(Math.round(output.rawValue), 176);
+
+    const valueOf = name => output.extendedInfo.value[name];
+    assert.deepEqual(valueOf('https://www.google-analytics.com/analytics.js'), {'Evaluate Script': 40.1, 'Compile Script': 9.6, 'Recalculate Style': 0.2});
```
paulirish: It was kind of confusing that our extendedInfo reported tasks, but the UI reported taskGroups. I don't have a strong opinion on what should be exposed in extendedInfo. Maybe we just leave `extendedInfo` empty this time? In the tests we could assert against the table details instead.

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,203 @@
...
+    const bootupTimings = BootupTime.getExecutionTimingsByURL(trace);
+
+    let totalBootupTime = 0;
+    const extendedInfo = {};
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
+    ];
+
+    // Group tasks per url
+    const groupsPerUrl = Array.from(bootupTimings).map(([url, durations]) => {
```
paulirish: this loop groups individual tasks into taskGroups. I figured this could be a lot simpler if we just do the same thing up inside the inner `getExecutionTimingsByURL` loop. that's what you'll see in #3703

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,203 @@
...
+      // when url is "" or about:blank, we skip it
+      if (!url || url === 'about:blank') {
+        return;
+      }
+
+      const tasks = {};
+      perUrlNode.children.forEach((perTaskPerUrlNode) => {
+        const taskGroup = WebInspector.TimelineUIUtils.eventStyle(perTaskPerUrlNode.event);
+        tasks[taskGroup.title] = tasks[taskGroup.title] || 0;
+        tasks[taskGroup.title] += Number((perTaskPerUrlNode.selfTime || 0).toFixed(1));
```
paulirish: i felt weird about this rounding here as we're summing, so i moved it to outside of the loop.

lighthouse-core/audits/bootup-time.js
```diff
@@ -0,0 +1,203 @@
...
+    };
+  }
+
+  /**
+   * @param {!Array<TraceEvent>=} trace
+   * @return {!Map<string, Number>}
+   */
+  static getExecutionTimingsByURL(trace) {
+    const timelineModel = new DevtoolsTimelineModel(trace);
+    const bottomUpByName = timelineModel.bottomUpGroupBy('URL');
```
paulirish: bottomUpByURL

lighthouse-core/test/audits/bootup-time-test.js
wardpeet: it might be confusing indeed but isn't it handy to show the raw information somewhere in the json report so people who want to use it can use it? Or isn't that why extendedInfo exists?

paulirish:
I own landing this now. No more work needed from you, @wardpeet. :)

Still need a little more work on #3703 then we're good with both.

lighthouse-core/audits/bootup-time.js
paulirish: filed here https://github.com/GoogleChrome/lighthouse/issues/3702

PR #3578, opened 2017-10-16
report(report-ui-features): Disable auto expansion of print option via report ui

### LINKED ISSUE #2782: Request: print summary/collapsed view
The default view I'm presented with is about 2 pages and is perfect.  When I click print from the share menu, it is a 28 page document.

I would like an option to print the main/collapsed view.

karanjthakkar:
Fixes #2782

### DIFF
--- a/lighthouse-core/report/v2/renderer/report-ui-features.js
+++ b/lighthouse-core/report/v2/renderer/report-ui-features.js
@@ -177,7 +177,8 @@ class ReportUIFeatures {
       case 'copy':
         this.onCopyButtonClick();
         break;
-      case 'print':
+      case 'print-summary':
+        this.collapseAllDetails();
         this.closeExportDropdown();
         self.print();
         break;
--- a/lighthouse-core/report/v2/renderer/report-ui-features.js
+++ b/lighthouse-core/report/v2/renderer/report-ui-features.js
@@ -178,6 +178,10 @@ class ReportUIFeatures {
         this.onCopyButtonClick();
         break;
       case 'print':
+        this.closeExportDropdown();
+        self.print();
+        break;
+      case 'print-expanded':
         this.expandAllDetails();
         this.closeExportDropdown();
         self.print();
--- a/lighthouse-core/report/v2/renderer/report-ui-features.js
+++ b/lighthouse-core/report/v2/renderer/report-ui-features.js
@@ -257,7 +257,6 @@ class ReportUIFeatures {
    */
   printShortCutDetect(e) {
     if ((e.ctrlKey || e.metaKey) && e.keyCode === 80) { // Ctrl+P
-      this.expandAllDetails();
       this.closeExportDropdown();
     }
   }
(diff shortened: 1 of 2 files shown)


karanjthakkar:
@patrickhulce From what I understand, the code you outlined is the one that runs when you click on the dropdown in the top right corner. For this we want to preserve the expanded functionality so I've just modified the text for the dropdown option. There will no longer be a print option in the dropdown. The user can do a normal print (which does not auto expand) using Ctrl + P or the Print option in File menu.

patrickhulce:
Oh oops I misread/didn't notice you were removing the print option entirely from the dropdown! That's fine with me, but I'll defer to report owners if we should have both "Print" and "Print Expanded" in the dropdown, in that case, my comment becomes relevant again :)

to @paulirish and @vinamratasingal

vinamratasingal:
To make sure I understand what's going on: we are changing the default print option to "print expanded" and then also ensuring that when a user does the system print, that they only print the summary view (i.e. 2 page report) and not the 28 page report, yes?

If the above is true, I would love to see in the share menu both a "print summary view" (or "print current view") and "print expanded view." What additional information gets printed out in the expanded view @patrickhulce ?

patrickhulce:
Expanded view just expands all the disclosure triangle sections in the report for you which is what balloons the size.

Sounds like there's desire to have both "Print" (same as triggering system print) and "Print Expanded" (the current logic for the print dropdown with just the text change) options in the dropdown, does that make sense to you too @karanjthakkar?

karanjthakkar:
If I understand correctly Print, Print Summary and Print Expanded are three different options:

1. Print: Will defer to the system print option. May or may not be summary view. Depends on whether the user has expanded any triangles on their own
2. Print Summary: Manually collapse all triangles and then print
3. Print Expanded: Manually expand all triangles and then print

In the old implementation, `Print` was actually printing the expanded report so I just updated the text to `Print Expanded`. Do we also want to keep the old `Print` option and make it do what Pt 1. says?

patrickhulce:
That's exactly what I'm thinking too @karanjthakkar, let's have pt 1 and pt 3 as options in the dropdown and punt on pt 2 for now 👍

karanjthakkar:
@patrickhulce done!

karanjthakkar:
Bump!

patrickhulce:
wait @vinamratasingal you were planning on having 3 distinct print options in the dropdown? Did you want all the options @karanjthakkar described?

I took
> I would love to see in the share menu both a "print summary view" (or "print current view") and "print expanded view." 

To mean that you were onboard with having a "Print" which prints the current view as is and a "Print Expanded" which explodes to the full report (current functionality)

vinamratasingal:
Er let me clarify. I would like 2 options: Print Summary and Print Expanded. That's it. 

With Print, it's just changing what the system print should do. System print should map to printing the current view of the page (which I believe it's already doing, so no need to change).

karanjthakkar:
So I removed the `Print...` option from the dropdown and added a summary option instead. Here are some screenshots of how it looks like:

- Dropdown screenshot
screen shot 2017-10-30 at 11 15 42 am

- Print Summary screenshot
screen shot 2017-10-30 at 11 15 50 am

- Print Expanded screenshot
screen shot 2017-10-30 at 11 15 55 am

Is that along the lines of what you were expecting, @vinamratasingal?

vinamratasingal:
Yup, this looks great, thanks Karan!

PR #3582, opened 2017-10-16
core(config): add silent seo audits to default config

### LINKED ISSUE #3521: Enable SEO audits without reporting in default config
We'd like to be able to have an audit that is not surfaced in the UI but available in the JSON results. For example, the SEO audits have a separate config and are not enabled by default. But we'd like to be able to get SEO results in HTTP Archive without necessarily displaying the incomplete audit category to users.

brendankenny:
fixes #3521

These audit results will show up in the `audits` object on the LHR, but won't show up in the HTML report as no category includes them. They will be filtered out by any config that subsets the default config because of this, however.

We can add the other SEO audits as they land (probably as long as they don't have substantial cost) and/or make them visible in the report by bringing in the SEO report category if we're ready.

@kdzwinel @rviscomi

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -35,6 +35,7 @@ module.exports = {
       'dobetterweb/response-compression',
       'dobetterweb/tags-blocking-first-paint',
       'dobetterweb/websql',
+      'seo/meta-description',
     ],
   },
   {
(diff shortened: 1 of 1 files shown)


kdzwinel:
:+1: I'll put all new audits into the default config from now on.

rviscomi:
👍 thanks @brendankenny!

PR #3589, opened 2017-10-17
misc(jsconfig): Enable type checking for JavaScript

kdzwinel:
Highlights errors in IDE, doesn't affect automatic linters.

checkJS:true

Doc: https://code.visualstudio.com/docs/languages/javascript
Can be disabled per file via `// @ts-nocheck`.

We could also add `strict: true`. I have hard time finding the docs for this, but from what I can see it doesn't allow implicit 'any' type and is able to recognize that function may return undefined while jsdoc says it should return a string.

checkJS:true; strict:true

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -6,7 +6,8 @@
   "compilerOptions": {
     "target": "ES6",
     "noEmit": true,
-    "diagnostics": true
+    "diagnostics": true,
+    "checkJs": true
   },
   "exclude": [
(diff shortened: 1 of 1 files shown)


brendankenny:
> We could also add `strict: true`

obvs this is non-binding for now, but this seems like a good idea. *If* we have type checking some day we'll want to be strict with nulls/undefined

kdzwinel:
@brendankenny 👍 added

PR #3591, opened 2017-10-17
misc(codereview): add CODEOWNERS file

paulirish:
#### More: 
* https://github.com/blog/2392-introducing-code-owners
* https://help.github.com/articles/about-codeowners/


paulirish:
> seems to request review of all owners 

yah. i think it'll probably be a bit overkill, but let's give it a whirl for a few days.. facebooks' mention bot has a little more configurability but it'll take a while to get up and running right.

and i was unable to find a tool that automates our waiting4reviewer/committer labels. so we'll probably need to write our own bot for that.

patrickhulce:
> so we'll probably need to write our own bot for that.

it shouldn't be too hard to press git3po into service there

PR #3592, opened 2017-10-17
core(dependency-graph): add cycle sanity check

### LINKED ISSUE #2703: Implement FastMode MVP

patrickhulce:
closes #2703

### DIFF
--- a/lighthouse-core/gather/computed/page-dependency-graph.js
+++ b/lighthouse-core/gather/computed/page-dependency-graph.js
@@ -231,6 +231,10 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
     PageDependencyGraphArtifact.linkNetworkNodes(rootNode, networkNodeOutput, networkRecords);
     PageDependencyGraphArtifact.linkCPUNodes(rootNode, networkNodeOutput, cpuNodes);
 
+    if (NetworkNode.hasCycle(rootNode)) {
+      throw new Error('Invalid dependency graph created, cycle detected');
+    }
+
     return rootNode;
   }
 
--- a/lighthouse-core/lib/dependency-graph/node.js
+++ b/lighthouse-core/lib/dependency-graph/node.js
@@ -81,14 +81,8 @@ class Node {
    */
   getRootNode() {
     let rootNode = this;
-    let maxDepth = 1000;
-    while (rootNode._dependencies.length && maxDepth) {
+    while (rootNode._dependencies.length) {
       rootNode = rootNode._dependencies[0];
-      maxDepth--;
-    }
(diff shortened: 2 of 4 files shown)


paulirish:
sgtm. i'll let brendan take this.

PR #3611, opened 2017-10-18
docs(api): generate API docs from source

### LINKED ISSUE #2463: generate html API docs for artifacts, driver, etc
generated API docs could be very helpful, preventing the need to dive into multiple gatherers just to see what they provide in their artifacts, for instance. Ideally these would be generated from source with a single `npm` command.

- It may not be necessary (or even helpful) to have html docs for everything, but docs for key pieces would aid those developing Lighthouse core and/or their own extensions of it. `Driver`, the artifacts returned by all the core gatherers, and maybe things like the [`Audit`](https://github.com/GoogleChrome/lighthouse/blob/1bb1cfe9a44a201c71f9631daa93a2528ee53b45/lighthouse-core/audits/audit.js) and [`Gatherer`](https://github.com/GoogleChrome/lighthouse/blob/1bb1cfe9a44a201c71f9631daa93a2528ee53b45/lighthouse-core/gather/gatherers/gatherer.js) base classes would be a great first start. OTOH, things like `GatherRunner` may not be helpful to many (but it does have that sweet execution outline)

- One issue is that the existing jsdoc declarations are a little spread out and not always complete. [Driver](https://github.com/GoogleChrome/lighthouse/blob/1bb1cfe9a44a201c71f9631daa93a2528ee53b45/lighthouse-core/gather/driver.js) is more or less documented in place, but the [artifacts returned by gatherers often aren't](https://github.com/GoogleChrome/lighthouse/blob/1bb1cfe9a44a201c71f9631daa93a2528ee53b45/lighthouse-core/gather/gatherers/html-without-javascript.js), and instead you'd have to [look to the closure externs for those types](https://github.com/GoogleChrome/lighthouse/blob/1bb1cfe9a44a201c71f9631daa93a2528ee53b45/lighthouse-core/closure/typedefs/Artifacts.js#L26-L27). Hopefully efforts to document all of this will help suggest how to better document the different parts of our architecture in a way that's maintainable and stays current.

- It's not clear how the docs should be laid out. It probably makes sense to do more of a logical grouping than one based on file location...e.g. you don't want to have to navigate to 20 different pages to see all the different artifact types, but `Driver` should probably be a page on its own. Again we want something maintainable, but also something that humans will be able to easily browse.

manekinekko:
FIX #2463

The generated documentation looks like this:
image

You can generate the api doc using the npm command: `npm run generate-docs`

I set the output folder in `docs/api` (this is configurable in `.jsdoc.json`)

However, it seems that some jsdocs comments are valid like this:

```
ERROR: Unable to parse a tag's type expression for source file 
lighthouse/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js in line 181 with tag title "return" and text "{!Promise<!Array<!Object>}": Invalid type expression "!Promise<!Array<!Object>": Expected ",", "=", ">" or "|" but end of input found.
```

TBH, I'm not familiar with this syntax. any idea how to fix this?

### DIFF
--- a/.jsdoc.json
+++ b/.jsdoc.json
@@ -1,0 +1,35 @@
+{
+    "tags": {
+        "allowUnknownTags": true,
+        "dictionaries": ["jsdoc", "closure"]
+    },
+    "source": {
+        "include": [
+            "lighthouse-cli/commands",
+            "lighthouse-core/audits",
+            "lighthouse-core/closue",
+            "lighthouse-core/gather",
+            "lighthouse-logger",
+            "package.json",
+            "README.md"
+        ],
+        "includePattern": ".+\\.(js|ts)$",
+        "excludePattern": "(node_modules/|docs|test|types)"
+    },
+    "plugins": [
+      "plugins/markdown"
+    ],
+    "templates": {
+        "cleverLinks": true,
+        "monospaceLinks": true,
+        "useLongnameInNav": false,
+        "showInheritedInNav": true
+    },
+    "opts": {
+        "destination": "./docs/api/",
+        "encoding": "utf8",
+        "private": true,
+        "recurse": true,
+        "template": "./node_modules/minami"
+    }
+}
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "smokehouse": "node lighthouse-cli/test/smokehouse/smokehouse.js",
     "deploy-viewer": "cd lighthouse-viewer && gulp deploy",
     "bundlesize": "bundlesize",
-    "plots-smoke": "bash plots/test/smoke.sh"
+    "plots-smoke": "bash plots/test/smoke.sh",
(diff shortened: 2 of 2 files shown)


patrickhulce:
WOW thanks @manekinekko!! the error you're getting is from some of our invalid [closure types](https://github.com/google/closure-compiler/wiki/Types-in-the-Closure-Type-System), just missing a `>` :)

paulirish:
I just staged this up on my gh-pages: http://paulirish.github.io/lighthouse/docs/api/lighthouse/2.5.1/

paulirish:
Thanks very much for doing this.

After some discussion we're not going to include this. Some of our recent typing work will definitely assist with the goals of #2463

Cheers!

manekinekko:
Sounds good 👍

PR #3617, opened 2017-10-19
docs(new-audits.md): Docs on new audits

### LINKED ISSUE #2624: Brainstorm on criteria for adding a new audit
All new audits bring overhead—in runtime performance, development maintenance, and report overload—so it would be good to have a list of criteria that allow weighing tradeoffs before adding a new audit or gatherer to Lighthouse core. Not necessarily for basing inclusion on a threshold that must be passed, but ideally most of the items below should be answered and documented somewhere before an audit is accepted.

Here are a few starting criteria I'd like us to enforce. Feel free to add to it/clarify things/strike things out as too onerous.

- Audit goal
  - Does it measure something important?
   Easily measurable things are often not important, and many important things are very difficult to measure. Is it worth bringing to a developers attention? If a metric is being used as a proxy for the real goal of an audit, they should correlate well, demonstrated with real data.
  - What are the specific pass/fail states?
  - If it's going to be a perf metric, it must be quantifiable:
   How much does thing X hurt page performance? If a performance best practice doesn't actually have an effect we can measure, it may not be worth including (and may not actually be a performance best practice :)
- Runtime performance overhead
  - How much does it slow down a LH run on a typical page? What about on pathological pages (ones with many images or many event listeners or many long tasks, etc)?
  - Does it require a new gatherer pass?
   It can be difficult to schedule new gatherers if they have specific offline/throttling/serviceworker-initialized requirements, but adding a new pass (and an additional page load) is the very last resort, and may often be worse than not including the audit.
- Does it call out to third-party servers?
  - This is currently a strict no.
- Report noise
   The report needs to communicate clearly what Lighthouse thinks are top priorities, and we only have so much developer patience when reading our results.
  - Does it overlap with existing audits?
   Maybe we should refactor the existing audits so that we can have a single audit that provides more insight.
  - On how many sites is it relevant?
   If it only applies to a small subset of sites, is it just noise in the report for most people running Lighthouse? We've discussed one way of dealing with this, an audit that just won't show up in the report if it wasn't relevant (e.g. don't show `<video>` audit results if there were no `<video>`s in the page)

Some of these aren't always relevant. A best-practices audit may not have good answers for e.g. scorability but could still be included if it has minimal overhead. If a set of audits almost always passes, the extra noise in the report may be fine since they'll be hidden away in a "passed items" section.

If an audit/gatherer isn't appropriate for the default config, it still could live in the newly-created [full config](https://github.com/GoogleChrome/lighthouse/blob/1bf0e07cb568a1d54fe19f8a216115da47e05a4d/lighthouse-core/config/full-config.js), where for now our performance requirements are less strict.

We also shouldn't be afraid of starting to point new audit ideas at Lighthouse plugins, which can have their own domain-specific requirements for audits. However, plugins have the downside that they can't currently be loaded in Lighthouse in DevTools or in the extension.

vinamratasingal:
closes #2624

### DIFF
--- a/docs/new-audits.md
+++ b/docs/new-audits.md
@@ -12,19 +12,19 @@ Lighthouse audits that surface in the report should be:
 - Measurable (especially for performance audits) or have clear pass/fail states. 
 - Not use 3rd party APIs for completing the audit check. 
 
-*Note: it's possible to submit a new audit proposal for something that you as a contributor feel very passionate about, but the 
-Lighthouse team feels that it wouldn't add value to a significant portion of the Lighthouse users. In this case, we'd be open to 
-having your audit be a part of the report when run with the full config on the Command Line. * 
+*Note: it's possible to submit a new audit proposal for something that you as a contributor feel very passionate about, but the Lighthouse team feels that it wouldn't add value to a significant portion of the Lighthouse users. In this case, we'd be open to having your audit be a part of the report when run with the full config on the Command Line.* 
 
 ## Process for creating a new audit
-- Briefly scan the criteria we’ve laid out above. If you think the principles match with your proposed new audit, then proceed! 
-- Next step is to create an issue on Github with answers to the following questions: 
+1. Briefly scan the criteria we’ve laid out above. If you think the principles match with your proposed new audit, then proceed! 
+2. Next step is to create an issue on Github with answers to the following questions: 
    - Description of audit and audit category (please include pass/fail states, and how it might be measured)
    - Explanation of how it’s different from other audits
    - What % of developers/pages will this impact (estimates OK, data points preferred) 
    - How is the new audit making a better web for end users? (data points preferred)
-   - What is the resourcing situation (who will create the audits, maintain the audits, and write/maintain the documentation) 
+   - What is the resourcing situation (who will create the audits, maintain the audits, and write/maintain the documentation)
+   - Do you envision this audit in the Lighthouse report or the full config in the CLI? If in the report, which section? 
    - How much support is needed from the Lighthouse team?
-- Once the proposal is submitted, then Lighthouse team will take a look and followup. We will discuss possible implementation approaches, and associated runtime overhead.
+   - Any other links or documentation that we should check out?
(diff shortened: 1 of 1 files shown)


docs/new-audits.md
```diff
@@ -0,0 +1,30 @@
...
+- Contribute significantly towards making the mobile web experience better for end users. 
+- Not have a significant impact on our runtime performance or bundle size. 
+- Something that is new, and not something that is already measured by existing audits. 
+- Important for our strategic goals as a product (i.e. could other Lightrider customers benefit from this? What about other GBO teams?)
+- Measurable (especially for performance audits) or have clear pass/fail states. 
+- Not use 3rd party APIs for completing the audit check. 
+
+*Note: it's possible to submit a new audit proposal for something that you as a contributor feel very passionate about, but the 
+Lighthouse team feels that it wouldn't add value to a significant portion of the Lighthouse users. In this case, we'd be open to 
+having your audit be a part of the report when run with the full config on the Command Line. *
```
paulirish: nuke the space before your last * to get the italics

docs/new-audits.md
```diff
@@ -0,0 +1,30 @@
...
+- Important for our strategic goals as a product (i.e. could other Lightrider customers benefit from this? What about other GBO teams?)
+- Measurable (especially for performance audits) or have clear pass/fail states. 
+- Not use 3rd party APIs for completing the audit check. 
+
+*Note: it's possible to submit a new audit proposal for something that you as a contributor feel very passionate about, but the 
+Lighthouse team feels that it wouldn't add value to a significant portion of the Lighthouse users. In this case, we'd be open to 
+having your audit be a part of the report when run with the full config on the Command Line. * 
+
+## Process for creating a new audit
+- Briefly scan the criteria we’ve laid out above. If you think the principles match with your proposed new audit, then proceed!
```
paulirish: lets use numbered bullets for this one instead ?  you can just use `1.` every time (and it'll increment)

docs/new-audits.md
vinamratasingal: done!

docs/new-audits.md
vinamratasingal: done!

PR #3633, opened 2017-10-21
Add new console error-logs audit and tests

Chris011:


paulirish:
github spam is getting weird.

PR #3657, opened 2017-10-26
new_audit(blocked-from-indexing): page is blocked from indexing

### LINKED ISSUE #3182: [SEO Audits] Page is not blocked from indexing
**Audit group:** [Crawling and indexing](https://github.com/GoogleChrome/lighthouse/issues/3118)
**Description:** Page is not blocked from indexing
**Failure description:** Page is blocked from indexing
**Help text:** The “Robots” directives tell crawlers how your content should be indexed. Directives like `noindex` and `none` prevent indexing. Learn more about the [Robots meta tag and header](https://developers.google.com/search/reference/robots_meta_tag).

**Success conditions:**
- No tag matching the query selector `meta[name=robots]` has a content attribute with a comma-separated value in the blocklist below
- X-Robots-Tag response header does not have a value in the blocklist below

Blocklist:
- noindex
- none

Notes:
- Site owners can block crawlers via: meta tags, robots.txt rules, or X-Robots-Tag response header.
  - In next version, check that location.href is not disallowed by robots.txt
- The robots meta name attribute is interchangeable with other bot names including:
  - msnbot
  - bingbot
  - google
  - yandex

kdzwinel:
Closes #3182

Failing:
screen shot 2017-10-26 at 02 40 53

Passing:
screen shot 2017-10-26 at 02 42 27

### DIFF
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -79,7 +79,7 @@ function requestHandler(request, response) {
 
         extraHeaders.forEach(header => {
           const parts = header.split(':');
-          headers[parts[0]] = parts[1];
+          headers[parts[0]] = parts.slice(1).join(':');
         });
       }
 
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -51,13 +51,14 @@ function requestHandler(request, response) {
   }
 
   function sendResponse(statusCode, data) {
-    let headers;
+    const headers = {};
+
     if (filePath.endsWith('.js')) {
-      headers = {'Content-Type': 'text/javascript'};
+      headers['Content-Type'] = 'text/javascript';
     } else if (filePath.endsWith('.css')) {
-      headers = {'Content-Type': 'text/css'};
+      headers['Content-Type'] = 'text/css';
     } else if (filePath.endsWith('.svg')) {
-      headers = {'Content-Type': 'image/svg+xml'};
+      headers['Content-Type'] = 'image/svg+xml';
     }
 
     let delay = 0;
@@ -70,6 +71,16 @@ function requestHandler(request, response) {
       // set delay of request when present
       if (typeof queryString.delay !== 'undefined') {
         delay = parseInt(queryString.delay, 10) || 2000;
+      }
+
+      if (typeof queryString.extra_header !== 'undefined') {
+        let extraHeaders = queryString.extra_header;
+        extraHeaders = Array.isArray(extraHeaders) ? extraHeaders : [extraHeaders];
+
+        extraHeaders.forEach(header => {
+          const parts = header.split(':');
+          headers[parts[0]] = parts[1];
+        });
       }
 
       // redirect url to new url if present
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -12,6 +12,7 @@ const path = require('path');
 const fs = require('fs');
 const parseQueryString = require('querystring').parse;
 const parseURL = require('url').parse;
+const HEADER_SAFELIST = new Set(['x-robots-tag']);
 
 const lhRootDirPath = path.join(__dirname, '../../../');
 
@@ -78,8 +79,11 @@ function requestHandler(request, response) {
         extraHeaders = Array.isArray(extraHeaders) ? extraHeaders : [extraHeaders];
 
         extraHeaders.forEach(header => {
-          const parts = header.split(':');
-          headers[parts[0]] = parts.slice(1).join(':');
+          const [headerName, ...headerValue] = header.split(':');
+
+          if (HEADER_SAFELIST.has(headerName.toLowerCase())) {
+            headers[headerName] = headerValue.join(':');
+          }
         });
       }
 
--- a/lighthouse-core/audits/seo/is-crawlable.js
+++ b/lighthouse-core/audits/seo/is-crawlable.js
@@ -12,6 +12,16 @@ const BLOCKLIST = new Set([
 ]);
 const ROBOTS_HEADER = 'x-robots-tag';
 
+/**
+ * @param {string} directives
+ * @returns {boolean}
+ */
+function hasBlockingDirective(directives) {
+  return directives.split(',')
+    .map(d => d.toLowerCase().trim())
+    .some(d => BLOCKLIST.has(d));
+}
+
 class IsCrawlable extends Audit {
   /**
    * @return {!AuditMeta}
@@ -22,7 +32,7 @@ class IsCrawlable extends Audit {
       description: 'Page isn’t blocked from indexing',
       failureDescription: 'Page is blocked from indexing',
       helpText: 'The “Robots” directives tell crawlers how your content should be indexed. ' +
-        '[Learn more](https://developers.google.com/search/reference/robots_meta_tag).',
+      '[Learn more](https://developers.google.com/search/reference/robots_meta_tag).',
       requiredArtifacts: ['MetaRobots'],
     };
   }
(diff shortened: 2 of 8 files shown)


lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,109 @@
...
+ * @returns {boolean}
+ */
+function isUnavailable(directive) {
+  const parts = directive.split(':');
+
+  if (parts.length <= 1 || parts[0] !== UNAVAILABLE_AFTER) {
+    return false;
+  }
+
+  const date = Date.parse(parts[1]);
```
patrickhulce: seems like we would want to do a `parts.slice(1).join(':')`, no? losing the time and timezone

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * Returns false if robots header specifies user agent (e.g. googlebot)
+ * @param {string} directives
+ * @returns {boolean}
+ */
+function hasUA(directives) {
+  const parts = directives.split(':');
+  return parts.length > 1 && parts[0] !== UNAVAILABLE_AFTER;
```
patrickhulce: should it be looking for a `GOOGLEBOT_USER_AGENT` const maybe?

let's also rename this then, if it returns false when a UA is specified how about `hasNoUserAgent` or something

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,109 @@
...
+              source: {
+                type: 'node',
+                snippet: `<meta name="robots" content="${artifacts.MetaRobots}" />`,
+              },
+            });
+          }
+        }
+
+        mainResource.responseHeaders
+          .filter(h => h.name.toLowerCase() === ROBOTS_HEADER && !hasUA(h.value) &&
```
patrickhulce: so to be clear, we're looking for the robots header that has a user agent specified *and* is specified to block?

seems like this might miss a few cases maybe I'm misunderstanding

can there not be multiple directives in the header?

lighthouse-core/audits/seo/is-crawlable.js
kdzwinel: Oh, good catch! Tests didn't catch that because `Date.parse` is very forgiving.

lighthouse-core/audits/seo/is-crawlable.js
kdzwinel: we are looking for robots header that *doesn't* have UA specified (we don't support UA specific headers) and is blocking indexing

kdzwinel:
@patrickhulce Thank you for a review! With all the tests I'm pretty sure it's working as intended, but I made the `UA` part very confusing - sorry for that. We want to ignore all user agent specific tags (we are only looking for `<meta type="robots"`) and headers (all directives prefixed with `somebot:`). There is one edge case for that last one though `unavailable_after: ` can be a first directive and may be confused for a bot name, that's why I have `!== UNAVAILABLE_AFTER` check.

I left additional comment. Please let me know if that makes sense to you.

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,112 @@
...
+ * @returns {boolean}
+ */
+function hasBlockingDirective(directives) {
+  return directives.split(',')
+    .map(d => d.toLowerCase().trim())
+    .some(d => BLOCKLIST.has(d) || isUnavailable(d));
+}
+
+/**
+ * Returns false if robots header specifies user agent (e.g. `googlebot: noindex`)
```
patrickhulce: Ah, I see is this comment a typo then? `Returns *true* if robots header specifies a user agent`?

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,112 @@
...
+    return false;
+  }
+
+  const date = Date.parse(parts.slice(1).join(':'));
+
+  return !isNaN(date) && date < Date.now();
+}
+
+/**
+ * Returns false if any of provided directives blocks page from being indexed
```
patrickhulce: same here, doesn't this return `true` when any of the directives blocks the page from being indexed?

lighthouse-core/test/audits/seo/is-crawlable-test.js
```diff
@@ -0,0 +1,184 @@
...
+
+    return IsCrawlableAudit.audit(artifacts).then(auditResult => {
+      assert.equal(auditResult.rawValue, true);
+    });
+  });
+
+  it('ignores UA specific directives', () => {
+    const mainResource = {
+      responseHeaders: [
+        {name: 'x-robots-tag', value: 'googlebot: unavailable_after: 25 Jun 2007 15:00:00 PST'},
```
patrickhulce: ah ok, I was confused about how multiple user agent + a default would be expressed. I didn't realize it'd be duplicate headers rather than a csv

would you mind adding a default value here that's valid just for future readers
i.e.
```
responseHeaders: [
  {name: 'x-robots-tag', value: 'googlebot: unavailable_after: 25 Jun 2007 15:00:00 PST'},
  {name: 'x-robots-tag', value: 'unavailable_after: 25 Jun 2027 15:00:00 PST'},
]
```

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,112 @@
...
+const ROBOTS_HEADER = 'x-robots-tag';
+const UNAVAILABLE_AFTER = 'unavailable_after';
+
+/**
+ * Checks if given directive is a valid unavailable_after directive with a date in the past
+ * @param {string} directive
+ * @returns {boolean}
+ */
+function isUnavailable(directive) {
+  const parts = directive.split(':');
```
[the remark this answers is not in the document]
kdzwinel: Yeah, I agree that'd be much more elegant, but in this case it won't work:

```javascript
const [key, value] = 'unavailable_after: 12 Jun 2017 12:30:00'.split(':');
```
value in this case would be ` 12 Jun 2017 12` instead of ` 12 Jun 2017 12:30:00`. I could do:
```javascript
const [key, ...value] = 'unavailable_after: 12 Jun 2017 12:30:00'.split(':');
```
But then, `value` is an array and I still have to `.join(':')` it, so not much different from a current solution :(

lighthouse-core/audits/seo/is-crawlable.js
kdzwinel: Right, I forgot to update both comments. Thanks!

lighthouse-core/audits/seo/is-crawlable.js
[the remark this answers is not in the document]
rviscomi: Would `split(':', 1)` resolve your concern?

kdzwinel:
@patrickhulce I've addressed your comments 👍 PTAL

lighthouse-core/audits/seo/is-crawlable.js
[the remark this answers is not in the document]
kdzwinel: TIL about second parameter of `.split`!
This gives me access to `unavailable_after` but doesn't give me the date:

```javascript
const [key, value] = 'unavailable_after: 12 Jun 2017 12:30:00'.split(':', 1);
```

`value` will be empty. Am I missing something here? 🤔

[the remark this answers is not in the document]
rviscomi: Yeah you're right, it doesn't do what I thought it would. Carry on!

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+class IsCrawlable extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      name: 'is-crawlable',
+      description: 'Page isn’t blocked from indexing',
```
patrickhulce: I'd prefer to word this more affirmatively (i.e. `Page can be indexed` or `Page is indexable`), but I'm guessing we can't because it'd be misleading and there are many other ways a page could be prevented from indexing?

patrickhulce:
@brendankenny there are smokehouse server changes FYI if you wanted to review :)

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -72,6 +73,16 @@ function requestHandler(request, response) {
         delay = parseInt(queryString.delay, 10) || 2000;
       }
 
+      if (typeof queryString.extra_header !== 'undefined') {
+        let extraHeaders = queryString.extra_header;
+        extraHeaders = Array.isArray(extraHeaders) ? extraHeaders : [extraHeaders];
+
+        extraHeaders.forEach(header => {
+          const parts = header.split(':');
```
brendankenny: I actually would prefer `const [key, ...value]

brendankenny: I actually would prefer `const [key, ...value]`, but that's just down to preference.

You could also do `header.split(/:(.+)/);`, which should give the correct split (captured groups also appear in the resulting array)

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -72,6 +73,16 @@ function requestHandler(request, response) {
...
+          headers[parts[0]] = parts.slice(1).join(':');
```
brendankenny: this might be complete overkill, but can we make a set of allowed headers and only add to `headers` if found in there? We block hidden files and anything outside of the working directory, but you never know...

kdzwinel: One can't be too careful!

kdzwinel:
@brendankenny header safelist added PTAL

kdzwinel:
@brendankenny thanks for merging 🙌

PR #3670, opened 2017-10-27
tests(web-inspector): add test for setImmediate polyfill

patrickhulce:
reminded by gitter to add test for this :)

### DIFF
--- a/lighthouse-core/test/lib/web-inspector-test.js
+++ b/lighthouse-core/test/lib/web-inspector-test.js
@@ -21,6 +21,21 @@ describe('Web Inspector lib', function() {
     assert.ok(WebInspector.Color);
   });
 
+  it('does not destroy setImmediate', done => {
+    const fakeArg = {
+      foo() {},
+    };
+
+    setImmediate(function(arg) {
+      try {
+        arg.foo();
+        done();
+      } catch (err) {
+        done(err);
+      }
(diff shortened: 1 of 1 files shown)


lighthouse-core/test/lib/web-inspector-test.js
```diff
@@ -21,6 +21,21 @@ describe('Web Inspector lib', function() {
     assert.ok(WebInspector.Color);
   });
 
+  it('does not destroy setImmediate', done => {
```
brendankenny: not immediately clear what's being tested here. Is this checking for the specific WebInspector (bad) `setImmediate` implementation?

patrickhulce: yes

`does not break setImmediate`
`does not polyfill setImmediate incorrectly`

any better?

brendankenny: yeah, that helps, though maybe a comment about why `arg.foo();` would throw, too :)

brendankenny: also TIL `setImmediate` also exists on `require('timers').setImmediate`, so could check `setImmediate === require('timers').setImmediate` in addition to this functional test.

In theory someone could overwrite that as well, but they may deserve to screw things up by the time they reach that level of effort.

PR #3676, opened 2017-10-28
core(tags blocking first-paint): exclude script type=module

### LINKED ISSUE #3359: Render-blocking script warnings should exclude <script type="module">
The logic here should be updated to check for the module attribute:
https://github.com/GoogleChrome/lighthouse/blob/63b4ac14d0a871ade0630db2885edd7848843243/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js#L32-L36

ThisIzKp:
Fixes #3359

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
@@ -27,13 +27,13 @@ const Gatherer = require('../gatherer');
 function collectTagsThatBlockFirstPaint() {
   return new Promise((resolve, reject) => {
     try {
-      const tagList = [...document.querySelectorAll('link, head script[src], script[type]')]
+      const tagList = [...document.querySelectorAll('link, head script[src]')]
         .filter(tag => {
           if (tag.tagName === 'SCRIPT') {
             return !tag.hasAttribute('async') &&
                 !tag.hasAttribute('defer') &&
                 !/^data:/.test(tag.src) &&
-                !(tag.getAttribute('type') === 'module');
+                tag.getAttribute('type') !== 'module';
           }
 
           // Filter stylesheet/HTML imports that block rendering.
--- a/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
@@ -27,12 +27,13 @@ const Gatherer = require('../gatherer');
 function collectTagsThatBlockFirstPaint() {
   return new Promise((resolve, reject) => {
     try {
-      const tagList = [...document.querySelectorAll('link, head script[src]')]
+      const tagList = [...document.querySelectorAll('link, head script[src], script[type]')]
         .filter(tag => {
           if (tag.tagName === 'SCRIPT') {
             return !tag.hasAttribute('async') &&
                 !tag.hasAttribute('defer') &&
-                !/^data:/.test(tag.src);
+                !/^data:/.test(tag.src) &&
+                !(tag.getAttribute('type') === 'module');
           }
 
           // Filter stylesheet/HTML imports that block rendering.
--- a/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
+++ b/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
@@ -36,7 +36,7 @@
 <script src="./dbw_tester.js"></script>
 
 <!-- PASS: 'modules' are deferred by default and don't block rendering -->
-<script type="module" src="./sample-module.js?delay=500"></script>
+<script type="module" src="./empty_module.js?delay=500"></script>
 
 <!-- FAIL(errors-in-console): exception thrown -->
 <script type="text/javascript">throw new Error('An error');</script>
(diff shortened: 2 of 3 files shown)


lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
```diff
@@ -27,12 +27,13 @@ const Gatherer = require('../gatherer');
 function collectTagsThatBlockFirstPaint() {
   return new Promise((resolve, reject) => {
     try {
-      const tagList = [...document.querySelectorAll('link, head script[src]')]
+      const tagList = [...document.querySelectorAll('link, head script[src], script[type]')]
```
patrickhulce: Why was this added?

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
```diff
@@ -27,12 +27,13 @@ const Gatherer = require('../gatherer');
...
         .filter(tag => {
           if (tag.tagName === 'SCRIPT') {
             return !tag.hasAttribute('async') &&
                 !tag.hasAttribute('defer') &&
-                !/^data:/.test(tag.src);
+                !/^data:/.test(tag.src) &&
+                !(tag.getAttribute('type') === 'module');
```
patrickhulce: let's do `tag.getAttribute('type') !== 'module'`

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
ThisIzKp: To detect the inline modules, type="module" may not be always followed by the `src` attribute.

```html
<script type="module">
import module from '../example.js';
module.test();
</script>
```

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
ThisIzKp: is this just for readability or anything I'm missing over?
I just wrote like this to make it consistent with the above checks

patrickhulce: well the others are like that only because there not using an operator, so yeah just easier to read :)

lighthouse-core/gather/gatherers/dobetterweb/tags-blocking-first-paint.js
patrickhulce: sure, but I mean we're trying to ignore modules so we'd be pulling them in just to ignore them

ThisIzKp:
@patrickhulce did those minor fixes, not sure about adding module entry to smoke tests. Can you help me out with any pointers to docs, examples or any resources about doing that?

patrickhulce:
The two lines I referenced are the best example. Basically you want to copy those two lines in that file, add `type="module"`, and change the comment to say it should pass because it's a module and is not blocking

https://github.com/GoogleChrome/lighthouse/blob/76e51f3821dad783763873e860e486949146d402/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html#L35-L36

ThisIzKp:
Oh, I was thinking may be I should create a new file separately to test `module` thing and I'm not sure about what to test out in those files. Thanks for that 👍

ThisIzKp:
Smoke tests are failing, need some more time to fix that issue.

ThisIzKp:
After adding this line
```
<script type="module" src="dbw_tester.js?module"></script>
```

Smoke tests are failing for `uses-http2` and `mutation-events`. I'm puzzled about how this change can affect that. Any help

patrickhulce:
You haven't pushed your changes so I'm not sure exactly what's failing 😄 but you've added a new URL so you'll probably need to update the `uses-http2` expectation to include 1 more request.

https://github.com/GoogleChrome/lighthouse/blob/fbe3fda0d7203039e378c101e482a7974c2f9949/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js#L35-L43

(bump line 40 to `16`)

mutation events is probably reporting the mutation events fired in dbw_tester.js so let's just change the URL to something that'll 404 like `src="sample-module.js"`, keep it simpler for you :)

ThisIzKp:
but if it's just a 404, the test will not give accurate results right?

patrickhulce:
a blocking script that 404s is still a blocking script so it should work fine for our purposes (you can add `src="sample-module.js?delay=500"` to make sure it takes a while to return and it shouldn't impact any of the other expectations other than `uses-http2`

ThisIzKp:
blocked at this - `<script type="module" src="./sample-module.js?delay=500"></script>` line is being triggered twice. 

https://github.com/GoogleChrome/lighthouse/blob/fbe3fda0d7203039e378c101e482a7974c2f9949/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js#L35-L43

If I bump up this number to 17 instead of 16, it works. But I'm curious about what is triggering it.

patrickhulce:
> If I bump up this number to 17 instead of 16, it works. But I'm curious about what is triggering it.

Seems like chrome double requests 404 modules, so I went ahead and fixed up the smoketests since I've had you churn enough 😄

ThisIzKp:
thank you for helping me out, got stuck so many times... but somehow finally merged 🎉

PR #3700, opened 2017-10-31
misc: use undated Apache 2 LICENSE file

ebidel:


ebidel:
You're supposed to append to the  generic license and modify it with your own info:

``` APPENDIX: How to apply the Apache License to your work.

      To apply the Apache License to your work, attach the following
      boilerplate notice, with the fields enclosed by brackets "[]"
      replaced with your own identifying information. (Don't include
      the brackets!)  The text should be enclosed in the appropriate
      comment syntax for the file format. We also recommend that a
      file or class name and description of purpose be included on the
      same "printed page" as the copyright notice for easier
      identification within third-party archives.

brendankenny:
right, but it's telling you how to do that when attaching that header to your work, not to do it to the text below. The internet seems mixed on this, but most of the Apache projects seem to do it this way, and presumably they know how to do it right (e.g. [couchdb](https://github.com/apache/couchdb/blob/0ce05d26a565188aa4e2b0181788b244fe385e01/LICENSE#L190), though notably [this one does not](https://github.com/apache/incubator-wave/blob/800fbc87a0a0d1655d497b54f87bf007064fca38/LICENSE#L189))

Issue #3716, opened 2017-11-01
Aspect ratio audit fails on SVG


peterjanes:
The image aspect ratio audit appears to fail for SVG images. If you run a "best practices" audit on [the standalone HTML file in this zip](https://github.com/GoogleChrome/lighthouse/files/1435521/svg-test.zip) (unzip it and put it on a webserver) you'll see it includes

```
Does not uses Images with appropriate aspect ratio
Image displayed sizes should match their natural aspect ratio.
```

I've tried it with different combinations of width/height attributes and CSS, and none can make it pass the audit.

patrickhulce:
Thanks for reporting @peterjanes! We ignore SVGs for responsive image savings but we should ignore them here too 👍 

For any potential first-time bug fixers, we just need to filter out based on mimeType here
https://github.com/GoogleChrome/lighthouse/blob/fbe3fda0d7203039e378c101e482a7974c2f9949/lighthouse-core/audits/image-aspect-ratio.js#L73-L78

TejasQ:
On it!

patrickhulce:
Thanks so much for volunteering @TejasQ but Peter beat you to the punch 😄  feel free to take a look at some of our other [good first issues](https://github.com/GoogleChrome/lighthouse/labels/good%20first%20issue)!

TejasQ:
Thanks @patrickhulce! Missed the reference from @peterjanes!

leonardorey1970:
Hi all, do you have any idea about when this changes will deployed to prod?.

Cheers!
Leo

patrickhulce:
@leonardorey1970 should go out with our next release by end of this week in extension and CLI. Chrome stable will get it around end of January.

leonardorey1970:
Thanks patrickhulce  for your reply!

PR #3722, opened 2017-11-01
core(aspect-ratio): skip aspect ratio audit for svg

peterjanes:
I'm pretty confident that the code matches the suggestion in #3716. I made an attempt at a new test, but it isn't correct (passes with and without the new check); suggestions welcome.

### DIFF
--- a/lighthouse-core/audits/image-aspect-ratio.js
+++ b/lighthouse-core/audits/image-aspect-ratio.js
@@ -73,6 +73,7 @@ class ImageAspectRatio extends Audit {
       // filter out images that don't have following properties
       // networkRecord, width, height, images that use `object-fit`: `cover` or `contain`
       return image.networkRecord &&
+        image.networkRecord.mimeType !== 'image/svg+xml' &&
         image.width &&
         image.height &&
         !image.usesObjectFit;
--- a/lighthouse-core/audits/image-aspect-ratio.js
+++ b/lighthouse-core/audits/image-aspect-ratio.js
@@ -70,8 +70,9 @@ class ImageAspectRatio extends Audit {
     let debugString;
     const results = [];
     images.filter(image => {
-      // filter out images that don't have following properties
-      // networkRecord, width, height, images that use `object-fit`: `cover` or `contain`
+      // - filter out images that don't have following properties:
+      //   networkRecord, width, height, images that use `object-fit`: `cover` or `contain`
+      // - filter all svgs as they have no natural dimensions to audit
       return image.networkRecord &&
         image.networkRecord.mimeType !== 'image/svg+xml' &&
         image.width &&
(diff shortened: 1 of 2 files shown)


peterjanes:
I've added my GitHub email to my Google account.

lighthouse-core/test/audits/image-aspect-ratio-test.js
```diff
@@ -32,7 +32,7 @@ describe('Images: aspect-ratio audit', () => {
           generateImage(
             {width: data.clientSize[0], height: data.clientSize[1]},
             {naturalWidth: data.naturalSize[0], naturalHeight: data.naturalSize[1]},
-            generateRecord(),
+            generateRecord({mimeType: data.mimeType}),
```
patrickhulce: it's expecting two parameters so I believe this should look like `generateRecord(undefined, data.mimeType)`

then again, it doesn't really look like `generateRecord` is very useful for this test at all and was copied from another file so you can probably just ditch the function entirely and pass in the object directly to `generateImage`

lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -73,6 +73,7 @@ class ImageAspectRatio extends Audit {
       // filter out images that don't have following properties
       // networkRecord, width, height, images that use `object-fit`: `cover` or `contain`
       return image.networkRecord &&
+        image.networkRecord.mimeType !== 'image/svg+xml' &&
```
patrickhulce: want to update the comment real quick to explain svg's don't have valid natural dimensions :)

paulirish:
huge! thanks so much @peterjanes  😁

piotr-cz:
Does it fix inline SVG images?
I'm on Chrome Version 63.0.3239.108 (Official Build) (64-bit)

patrickhulce:
@piotr-cz this PR won't show up in Chrome until M65 unfortunately

piotr-cz:
No, I don't but it should be easy to reproduce with just any inline SVG image

patrickhulce:
@piotr-cz it shouldn't be possible, and I'm not able to reproduce with any inline SVG image. That's why I'm asking if you remember where you saw it :)

piotr-cz:
@patrickhulce The app I'm working on is quite complex and it would pretty hard to create test case out of it. It's not public, but I can send you a link by email to investigate if that's ok.

patrickhulce:
sure that'd be great, email should be available on my GitHub profile

piotr-cz:
@patrickhulce I've just send the link to email at Your GitHub profile.

Would be helpful, if Lighthouse would add more info, for example which image doesn't pass the test.

patrickhulce:
@piotr-cz thanks for sending, I see the confusion now. You're referring to a different bug, #3551, which was fixed in 2.6 and should also be in Chrome 65 (use Chrome Canary, CLI, or extension for immediate workaround).

PR #3729, opened 2017-11-02
core(meta-tag-gatherers): meta tag search should be case-insensitive

kdzwinel:
screen shot 2017-11-02 at 22 00 20


kdzwinel:
@patrickhulce good point! I've updated one of my previous tests (and somehow pulled @brendankenny into this - sorry!).

PR #3730, opened 2017-11-02
core(predictive-perf): predict FCP

### LINKED ISSUE #3593: Lantern: Implement proper redirect handling
carried over from #2703, adding proper redirect handling regressed TTCI prediction performance so it requires digging into the regressions to see what's going on there (likely accounting for the true redirect cost increases the penalty for erroneously included iframes or something)

### LINKED ISSUE #3671: Lantern: Research performance of predicting FCP instead of FMP
Currently FMP is what's targeted by the predictive-perf audit but the graphs generated should apply equally to FCP and might be even more accurate. Since FMP is being deprioritized, we should investigate how well FCP fares.

patrickhulce:
closes #3671 
closes #3593 

Minimal changes required here for converting FMP to FCP, basically just move the time boundary to FCP, removing the CPU layout tasks, and adding in the EvaluateScript tasks of blocking scripts (which really could have been part of FMP prediction too).

I also added redirects support since FCP was more radically impacted by missing redirect support than FMP was. I dug into a lot of the larger error cases (where error was >1s) and at this point small differences should not be used for decision making. 

Some notable examples:

* DevTools throttling limitations and approximation of just adding 600ms delay to everything greatly exaggerates impact of redirects (i.e. 3 redirects = 1.8 s instead of the correct result of ~
* Server response time was measured in seconds and had variability i.e. returned in 5s for unthrottled case but only 3 seconds in throttled case

Overall, we're about as accurate as we were with FMP, higher error at the tail but lower error for most sites. Further progress on accuracy will need higher fidelity throttling like WPT.

### Accuracy comparison
| Metric | Value on 99% | Value on 95% | Value on 90% |
| -- | -- | -- | -- |
| FMP - spearman's rho | .805 | .833 | .854 |
| FCP - spearman's rho | .792 | .832 | .850 |
| FMP - MAPE % | 27.44% | 24.11% | 21.94% |
| FCP - MAPE % | 29.03% | 21.72% | 19.65% |

### DIFF
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -97,6 +97,36 @@ class PredictivePerf extends Audit {
 
   /**
    * @param {!Node} dependencyGraph
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {!Node}
+   */
+  static getOptimisticFMPGraph(dependencyGraph, traceOfTab) {
+    const fmp = traceOfTab.timestamps.firstMeaningfulPaint;
+    return dependencyGraph.cloneWithRelationships(node => {
+      if (node.endTime > fmp || node.type === Node.TYPES.CPU) return false;
+      // Include non-script-initiated network requests with a render-blocking priority
+      return node.hasRenderBlockingPriority() && node.initiatorType !== 'script';
+    });
+  }
+
+  /**
+   * @param {!Node} dependencyGraph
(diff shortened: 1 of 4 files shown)


patrickhulce:
> I'm interested in keeping the FMP prediction, for now.

I added back the prediction functions so we don't have to dig those up again, but I'm inclined to reduce churn in exposing until we have a concrete plan for what we expose ala #3761

PR #3741, opened 2017-11-03
core(driver): increase default timeout to 45s

### LINKED ISSUE #3733: Core: increase page load timeout to 45s
As mentioned in #3639

patrickhulce:
closes #3733 

things to consider:
* this might break folks who were running on bad hardware and needed to keep traces small (counter: it currently breaks for plenty of these sites at 30s already)
* scores will change from 0 to ~8 at sites with TTI at 30s


vinamratasingal:
Huge fan of this- increasing timeouts will (I think) reduce our error rates.

brendankenny:
we got a verbal 👍 from @paulirish so let's do this!

PR #3743, opened 2017-11-03
core(lifecycle): allow gathering & auditing to run separately

### LINKED ISSUE #1806: Split Gather phase from Audit phase
We currently can [support auditing from saved artifacts](https://github.com/GoogleChrome/lighthouse#lighthouse-as-trace-processor), but we haven't really generalized this for all artifacts, just for performance stuff. 

Splitting these two up would allow for folks who want to run Lighthouse against 1000s of URLs to gather on one machine and audit on another (or farmed out too many).

I _think_ this involves saving the artifacts to disk as each gatherer finishes, and then allowing LH to pick up the disk artifacts later and run the remainder of the analysis.

As a side benefit, it'll be nice to do `lighthouse --process-last-run` (or whatever) during development of an audit/formatter/report rather than doing the entire round trip each iteration. 

Off the top of my head, we probably have to figure out:

1. What this looks like at the CLI
2. How we save/store/delete the disk artifacts
3. How the artifacts are retained in the devtools/extension case
4. How we adapt the existing config-based approach to this

---

This is open for ideas, discussion, and anyone's help in moving this forward. :)

paulirish:
fixes #1806

The mystical -GAR feature.  Actually its just -GA for now. 

Here's how this works:

```sh
lighthouse -G http://example.com
# launches browser, collects data, saves to disk (in `./latest-run/`) and quits

lighthouse -A http://example.com
# skips browser interaction, loads artifacts from disk (in `./latest-run/`), runs audits on them, generates report

lighthouse -GA http://example.com
# Normal gather + audit run, but also saves artifacts to disk
```

* `--gather-mode` and `--audit-mode` are the long versions of `-G` and `-A`.
* `config.auditResults` is removed. it was weird.
* assetSaver got a new impl of `saveArtifacts`, and added `loadArtifacts`
* --save-artifacts CLI flag was removed, as it was never really used and makes less sense in -G world. 

-------------------

##### Todo:
* Decide if we want to rename these, as "gather" hasn't been really user-visible before. 

##### Future work:
* Add `-G=./artifacts-save-path/`
* Add `-A=./artifacts-load-path/`

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -270,6 +270,8 @@ class Config {
    * @param {string=} configPath The absolute path to the config file, if there is one.
    */
   constructor(configJSON, configPath) {
+    debugger;
+
     if (!configJSON) {
       configJSON = defaultConfig;
       configPath = path.resolve(__dirname, defaultConfigPath);
@@ -304,12 +306,14 @@ class Config {
     }
 
     // Generate a limited config if specified
+    debugger;
+
     if (configJSON.settings &&
         (Array.isArray(configJSON.settings.onlyCategories) ||
-        Array.isArray(configJSON.settings.auditModes) ||
+        Array.isArray(configJSON.settings.onlyAudits) ||
         Array.isArray(configJSON.settings.skipAudits))) {
       const categoryIds = configJSON.settings.onlyCategories;
-      const auditIds = configJSON.settings.auditModes;
+      const auditIds = configJSON.settings.onlyAudits;
       const skipAuditIds = configJSON.settings.skipAudits;
       configJSON = Config.generateNewFilteredConfig(configJSON, categoryIds, auditIds,
           skipAuditIds);
@@ -365,6 +369,7 @@ class Config {
    * @return {!Object} A new config
    */
   static generateNewFilteredConfig(oldConfig, categoryIds, auditIds, skipAuditIds) {
+    debugger;
     // 0. Clone config to avoid mutating it
     const config = deepClone(oldConfig);
     // 1. Filter to just the chosen categories
@@ -521,6 +526,7 @@ class Config {
    * @return {!Object} fresh passes object
    */
   static generatePassesNeededByGatherers(oldPasses, requiredGatherers) {
+    debugger;
     const auditsNeedTrace = requiredGatherers.has('traces');
     const passes = JSON.parse(JSON.stringify(oldPasses));
     const filteredPasses = passes.map(pass => {
@@ -601,6 +607,11 @@ class Config {
   /** @type {Array<!Audit>} */
   get audits() {
     return this._audits;
+  }
+
+  /** @type {Array<!AuditResult>} */
+  get auditResults() {
+    return this._auditResults;
   }
 
   /** @type {Array<!Artifacts>} */
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -306,10 +306,10 @@ class Config {
     // Generate a limited config if specified
     if (configJSON.settings &&
         (Array.isArray(configJSON.settings.onlyCategories) ||
-        Array.isArray(configJSON.settings.onlyAudits) ||
+        Array.isArray(configJSON.settings.auditModes) ||
         Array.isArray(configJSON.settings.skipAudits))) {
       const categoryIds = configJSON.settings.onlyCategories;
-      const auditIds = configJSON.settings.onlyAudits;
+      const auditIds = configJSON.settings.auditModes;
       const skipAuditIds = configJSON.settings.skipAudits;
       configJSON = Config.generateNewFilteredConfig(configJSON, categoryIds, auditIds,
           skipAuditIds);
@@ -584,6 +584,10 @@ class Config {
     });
   }
 
+  removePasses() {
+    this._passes = undefined;
+  }
+
   /** @type {string} */
   get configDir() {
     return this._configDir;
@@ -597,11 +601,6 @@ class Config {
   /** @type {Array<!Audit>} */
   get audits() {
     return this._audits;
-  }
-
-  /** @type {Array<!AuditResult>} */
-  get auditResults() {
-    return this._auditResults;
   }
 
   /** @type {Array<!Artifacts>} */
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -140,14 +140,12 @@ function validatePasses(passes, audits, rootPath) {
   });
 }
 
-function validateCategories(categories, audits, auditResults, groups) {
+function validateCategories(categories, audits, groups) {
   if (!categories) {
     return;
   }
 
-  const auditIds = audits ?
-      audits.map(audit => audit.meta.name) :
-      auditResults.map(audit => audit.name);
+  const auditIds = audits.map(audit => audit.meta.name);
   Object.keys(categories).forEach(categoryId => {
     categories[categoryId].audits.forEach((audit, index) => {
       if (!audit.id) {
@@ -323,10 +321,6 @@ class Config {
     this._configDir = configPath ? path.dirname(configPath) : undefined;
 
     this._passes = configJSON.passes || null;
-    this._auditResults = configJSON.auditResults || null;
-    if (this._auditResults && !Array.isArray(this._auditResults)) {
-      throw new Error('config.auditResults must be an array');
-    }
 
     this._audits = Config.requireAudits(configJSON.audits, this._configDir);
     this._artifacts = expandArtifacts(configJSON.artifacts);
@@ -335,7 +329,7 @@ class Config {
 
     // validatePasses must follow after audits are required
     validatePasses(configJSON.passes, this._audits, this._configDir);
-    validateCategories(configJSON.categories, this._audits, this._auditResults, this._groups);
+    validateCategories(configJSON.categories, this._audits, this._groups);
   }
 
   /**
@@ -607,11 +601,6 @@ class Config {
   /** @type {Array<!Audit>} */
   get audits() {
     return this._audits;
-  }
-
-  /** @type {Array<!AuditResult>} */
-  get auditResults() {
-    return this._auditResults;
   }
 
   /** @type {Array<!Artifacts>} */
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -268,8 +268,6 @@ class Config {
    * @param {string=} configPath The absolute path to the config file, if there is one.
    */
   constructor(configJSON, configPath) {
-    debugger;
-
     if (!configJSON) {
       configJSON = defaultConfig;
       configPath = path.resolve(__dirname, defaultConfigPath);
@@ -304,8 +302,6 @@ class Config {
     }
 
     // Generate a limited config if specified
-    debugger;
-
     if (configJSON.settings &&
         (Array.isArray(configJSON.settings.onlyCategories) ||
         Array.isArray(configJSON.settings.onlyAudits) ||
@@ -363,7 +359,6 @@ class Config {
    * @return {!Object} A new config
    */
   static generateNewFilteredConfig(oldConfig, categoryIds, auditIds, skipAuditIds) {
-    debugger;
     // 0. Clone config to avoid mutating it
     const config = deepClone(oldConfig);
     // 1. Filter to just the chosen categories
@@ -520,7 +515,6 @@ class Config {
    * @return {!Object} fresh passes object
    */
   static generatePassesNeededByGatherers(oldPasses, requiredGatherers) {
-    debugger;
     const auditsNeedTrace = requiredGatherers.has('traces');
     const passes = JSON.parse(JSON.stringify(oldPasses));
     const filteredPasses = passes.map(pass => {
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -578,10 +578,6 @@ class Config {
     });
   }
 
-  removePasses() {
-    this._passes = undefined;
-  }
-
   /** @type {string} */
   get configDir() {
     return this._configDir;
--- a/lighthouse-core/index.js
+++ b/lighthouse-core/index.js
@@ -46,7 +46,7 @@ function lighthouse(url, flags = {}, configJSON) {
 
     // kick off a lighthouse run
     return Runner.run(connection, {url, flags, config})
-      .then(lighthouseResults => {
+      .then((lighthouseResults = {}) => {
         // Annotate with time to run lighthouse.
         const endTime = Date.now();
         lighthouseResults.timing = lighthouseResults.timing || {};
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -142,7 +142,7 @@ function saveArtifacts(artifacts, basePath) {
 
   // save everything else
   promise = promise.then(_ => {
-    fs.writeFileSync(`${basePath}/${artifactsFilename}`, JSON.stringify(artifacts), 'utf8');
+    fs.writeFileSync(`${basePath}/${artifactsFilename}`, JSON.stringify(artifacts, 0, 2), 'utf8');
     log.log('Artifacts saved to disk in folder:', basePath);
   });
   return promise;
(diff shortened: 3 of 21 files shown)


lighthouse-core/config/config.js
```diff
@@ -318,10 +316,6 @@ class Config {
     this._configDir = configPath ? path.dirname(configPath) : undefined;
 
     this._passes = configJSON.passes || null;
-    this._auditResults = configJSON.auditResults || null;
-    if (this._auditResults && !Array.isArray(this._auditResults)) {
-      throw new Error('config.auditResults must be an array');
```
patrickhulce: Is anyone using this should we wait for a major version? I'm fine nuking, but seems like it could still be supported

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +58,58 @@ img {
   `;
 }
 
+function loadArtifacts(basePath) {
+  log.log('Reading artifacts from disk:', basePath);
+  const artifacts = JSON.parse(fs.readFileSync(path.join(basePath, 'artifacts.json'), 'utf8'));
+  artifacts.traces = {};
+
+  const filenames = fs.readdirSync(basePath);
+  const promises = filenames.filter(filename => filename.endsWith('-trace.json')).map(filename => {
```
patrickhulce: should we do both traces and devtoolsLogs this way? It'd be nice to be consistent with how the assets are saved to disk in non-GAR mode but also less useful since there's nothing you can do with just the devtoolslog

nit: we also liked `.trace.json`/`.devtoolslog.json` if we want to stick with it here too :)

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +58,58 @@ img {
...
 // Set to ignore because testing it would imply testing fs, which isn't strictly necessary.
 /* istanbul ignore next */
-function saveArtifacts(artifacts, pathWithBasename) {
-  const fullPath = `${pathWithBasename}.artifacts.log`;
-  // The networkRecords artifacts have circular references
-  fs.writeFileSync(fullPath, stringifySafe(artifacts));
-  log.log('artifacts file saved to disk', fullPath);
+function saveArtifacts(artifacts, basePath) {
+  assert.notEqual(basePath, '/');
+  rimraf.sync(basePath);
```
patrickhulce: since we actually know about all the files we create, can we just delete those instead of the whole directory?

lighthouse-core/runner.js
```diff
@@ -56,83 +59,68 @@ class Runner {
...
-        }
-
-        // And add computed artifacts.
-        return Object.assign({}, artifacts, Runner.instantiateComputedArtifacts());
-      });
+    // User can run -G solo, -A solo, or -GA together
+    const shouldGatherAndQuit = opts.flags.gatherMode && !opts.flags.auditMode;
+    const shouldOnlyAudit = opts.flags.auditMode && !opts.flags.gatherMode;
+    const shouldDefaultRunButSaveArtifacts = opts.flags.auditMode && opts.flags.gatherMode;
+    const shouldDoTypicalRun = !opts.flags.gatherMode && !opts.flags.auditMode;
```
patrickhulce: can we pick either `Typical` *or* `Default` :)

lighthouse-core/runner.js
```diff
@@ -56,83 +59,68 @@ class Runner {
...
-          }
-        }
+    const shouldSaveArtifactsToDisk = shouldGatherAndQuit || shouldDefaultRunButSaveArtifacts;
+    const shouldLoadArtifactsFromDisk = shouldOnlyAudit;
+    const shouldGatherFromBrowser = config.passes && !config.artifacts &&
+        (shouldGatherAndQuit || shouldDefaultRunButSaveArtifacts || shouldDoTypicalRun);
 
-        return artifacts;
+    if (shouldLoadArtifactsFromDisk) {
+      config.removePasses();
```
patrickhulce: seems like this should be unnecessary given we've exploded the various should* branches into variables above?

lighthouse-core/runner.js
```diff
@@ -56,83 +59,68 @@ class Runner {
...
+      run = run.then(_ => {
+        return assetSaver.loadArtifacts(basePath).then(artifacts => config._artifacts = artifacts);
```
patrickhulce: could we collect config's artifacts into a local variable and just manipulate that instead of relying on them being set on the config down below?

lighthouse-core/runner.js
```diff
@@ -11,11 +11,14 @@ const ReportGeneratorV2 = require('./report/v2/report-generator');
...
+const assetSaver = require('./lib/asset-saver');
 const fs = require('fs');
 const path = require('path');
 const URL = require('./lib/url-shim');
 const Sentry = require('./lib/sentry');
 
+const basePath = path.join(process.cwd(), 'latest-run');
+
 class Runner {
   static run(connection, opts) {
```
patrickhulce: this method is a bit big for its 👖  these days, it seems like it can roughly be broken down into

* loadOrGatherArtifacts
* potentiallySaveArtifacts
* potentiallyComputeAudits
* combine results

do you think it'd be easier to handle the `config.gatherMode` type flags individually in each of these 4 hypothetical methods rather than exploding the matrix of choices? the plain english variable names of "this is what is happening" is kinda nice right now, but its quite a long function

lighthouse-core/test/index-test.js
```diff
@@ -87,7 +87,7 @@ describe('Module Tests', function() {
       });
   });
 
-  it('should return formatted audit results when given no categories', function() {
```
patrickhulce: oh wow you don't even get any category results? yeah let's nuke this

lighthouse-core/runner.js
```diff
@@ -56,83 +59,68 @@ class Runner {
...
       });
+    }
 
-      run = run.then(artifacts => {
-        log.log('status', 'Analyzing and running audits...');
-        return artifacts;
-      });
+    // Entering: Gather phase
+    if (!config.passes && !config.artifacts && !shouldLoadArtifactsFromDisk) {
+      const err = new Error('You must require either gather passes or provide saved artifacts.');
```
patrickhulce: "You must require" seems a bit odd, "You must provide either ..." or "The config requires ... " perhaps?

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -547,9 +547,8 @@ describe('GatherRunner', function() {
 
     return GatherRunner.run(passes, options)
       .then(artifacts => {
-        // todo, trash these
-        assert.equal(artifacts.networkRecords['firstPass'], undefined);
-        assert.equal(artifacts.networkRecords['secondPass'], undefined);
+        assert.equal(artifacts.networkRecords, undefined);
+        assert.equal(artifacts.networkRecords, undefined);
```
patrickhulce: nit: remove one of these

lighthouse-core/test/runner-test.js
```diff
@@ -6,16 +6,83 @@
...
 const path = require('path');
+const sinon = require('sinon');
+
 const computedArtifacts = Runner.instantiateComputedArtifacts();
 
 /* eslint-env mocha */
 
 describe('Runner', () => {
+  const saveArtifactsSpy = sinon.spy(assetSaver, 'saveArtifacts');
+  const loadArtifactsSpy = sinon.spy(assetSaver, 'loadArtifacts');
```
patrickhulce: oh these are straight spies, I thought you were stubbing

I'd have a slight preference for stubbing `loadArtifacts` and asserting the LHR looks like we'd expect it to than 100% relying on 'was called' assertions

lighthouse-cli/run.ts
```diff
@@ -141,23 +142,28 @@ export async function runLighthouse(
...
+      flags.port = launchedChrome.port;
+    }
     const results = await lighthouse(url, flags, config);
 
     const artifacts = results.artifacts;
     delete results.artifacts;
-
-    await saveResults(results, artifacts!, flags);
-    await launchedChrome.kill();
+    if (shouldSaveResults){
```
brendankenny: can this be combined with the logic of whether or not to save already in `saveResults`?

lighthouse-cli/run.ts
```diff
@@ -141,23 +142,28 @@ export async function runLighthouse(
...
+      await saveResults(results, artifacts!, flags);
+    }
 
     return results;
   } catch (err) {
+    return handleError(err);
```
brendankenny: does this work? `handleError` calls `process.exit()`

lighthouse-core/config/config.js
brendankenny: Ideally it would stick around, at least until 3.0. `auditResults` can skip everything and go right to the scoring (basically `-R`)

lighthouse-core/index.js
```diff
@@ -39,7 +39,7 @@ module.exports = function(url, flags = {}, configJSON) {
 
     // kick off a lighthouse run
     return Runner.run(connection, {url, flags, config})
-      .then(lighthouseResults => {
+      .then((lighthouseResults = {}) => {
```
brendankenny: what execution path leads to this being undefined?

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +58,58 @@ img {
...
+  const traces = artifacts.traces;
+  const savePromies = Object.keys(traces).map(passName => {
+    return saveTrace(traces[passName], `${basePath}/${passName}-trace.json`);
+  });
+
+  const p = Promise.all(savePromies).then(_ => {
+    // do everything else
+    delete artifacts.traces;
+    // The networkRecords artifacts have circular references
+    fs.writeFileSync(`${basePath}/artifacts.json`, stringifySafe(artifacts), 'utf8');
```
brendankenny: `stringifySafe` will be broken on re-import if we actually have circular references, so should just use `JSON.stringify` (and hopefully we've done things correctly :)

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +58,58 @@ img {
...
+
+  // do traces
+  const traces = artifacts.traces;
+  const savePromies = Object.keys(traces).map(passName => {
+    return saveTrace(traces[passName], `${basePath}/${passName}-trace.json`);
+  });
+
+  const p = Promise.all(savePromies).then(_ => {
+    // do everything else
+    delete artifacts.traces;
```
brendankenny: can't delete `traces` on `artifacts`, maybe make a copy of the object?

lighthouse-core/runner.js
brendankenny: Agreed on no real need to make a variable for each possible state. Rather than just four, can't we just have
- loadArtifacts
- gatherArtifacts
- saveArtifacts
- runAudits
- score
- createLHR
and just have run() run the ones that are needed for each flag?

lighthouse-core/test/runner-test.js
```diff
@@ -276,47 +345,19 @@ describe('Runner', () => {
...
     const config = new Config({
-      auditResults: [{
-        name: 'content-width',
-        rawValue: true,
-        score: true,
-        displayValue: 'display',
+      passes: [{
+        gatherers: ['viewport-dimensions'],
       }],
+      audits: [
```
brendankenny: how does this test audit output without running a gatherer or audit?

lighthouse-core/config/config.js
brendankenny: (it's also helpful for tests where you just want to test the -R part)

paulirish: I'd prefer to just kill auditResults. Turns out we dont really have tests using it (except one).

auditResults is this weird inbetween value that's only different from the LHR because of these few lines: https://github.com/GoogleChrome/lighthouse/blob/gar/lighthouse-core/runner.js#L128-L155

if we REALLY want to support -R then we'd definitely not put auditResults on the config instance because it's bizarre. (obv that would break backcompat (for those mystery users)).   if we want to do this i'd prefer to do it in a followup

lighthouse-core/index.js
paulirish: -G does. no lighthouse results because we saved artifacts and are resolving early

i could pass an object back there, but it seems slightly more awkward. open to ideas.

wardpeet:
will there be docs about this? This feature is really great if you run it in a cloud environment!

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +56,93 @@ img {
...
 // Set to ignore because testing it would imply testing fs, which isn't strictly necessary.
 /* istanbul ignore next */
-function saveArtifacts(artifacts, pathWithBasename) {
-  const fullPath = `${pathWithBasename}.artifacts.log`;
-  // The networkRecords artifacts have circular references
-  fs.writeFileSync(fullPath, stringifySafe(artifacts));
-  log.log('artifacts file saved to disk', fullPath);
+function saveArtifacts(artifacts, basePath) {
+  if (!fs.existsSync(basePath)) {
+    fs.mkdirSync(basePath);
```
wardpeet: we should use `mkdirp.sync` here to make sure we can create dirs recursively  (we should already have it as a dependency)

lighthouse-core/lib/asset-saver.js
```diff
@@ -54,19 +56,93 @@ img {
...
+  }
+  rimraf.sync(`${basePath}/*${traceSuffix}`);
```
wardpeet: I believe chrome launcher had a lot of issues on windows when using sync rimraf. Should we move to async?

paulirish: rimraf's bumps since then were around this issue so i am hoping we're good now. https://github.com/isaacs/rimraf/commits/master

lighthouse-core/lib/asset-saver.js
paulirish: sg. done

lighthouse-cli/run.ts
paulirish: fair. changed it.

lighthouse-cli/run.ts
paulirish: done

lighthouse-cli/cli-flags.ts
```diff
@@ -66,6 +69,9 @@ export function getFlags(manualArgv?: string) {
         'disable-device-emulation': 'Disable Nexus 5X emulation',
         'disable-cpu-throttling': 'Disable CPU throttling',
         'disable-network-throttling': 'Disable network throttling',
+        'gather-mode':
+            'Collect artifacts from a connected browser, save, & quit. However, if audit-mode is also enabled, then the run will complete after saving artifacts to disk.',
```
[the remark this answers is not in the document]
paulirish: rewrote it

lighthouse-core/runner.js
paulirish: sg. say hello to

_loadArtifactsFromDisk
_gatherArtifactsFromBrowser
_saveArtifacts
_runAudits
_scoreAndCategorize

lighthouse-core/runner.js
paulirish: how about 

> No browser artifacts are either provided or requested.

lighthouse-core/test/index-test.js
paulirish: added a new test to assert things finish even if no categories in the config

lighthouse-core/test/runner-test.js
paulirish: i think we'll have to discuss this in person, as all runner tests do actually run gatherrunner.run, runAudits, generateReport, etc. So only going with stubs for these seems inconsistent. 
Also, we'll have to mimic what these methods do on the test side if we want to see that everything works (for example: does -GA save to disk?)

lighthouse-core/test/runner-test.js
paulirish: it does run a gatherer and an audit. :o

paulirish:
Okay folks.. A fresh update here.

The big changes to runner are in this commit: https://github.com/GoogleChrome/lighthouse/pull/3743/commits/a3da663e1679a6074f1c0e33a4fa533a5477f525
There's no way to make that diff super straightforward, but I hope it's a bit easier to read now.

Otherwise, the other feedback has been addressed as well.

lighthouse-core/runner.js
patrickhulce: sg, even further?

> No browser artifacts provided or requested.

lighthouse-core/test/runner-test.js
patrickhulce: I'm all for having at least some integration testing of this 👍 

Maybe we'll chat in person but seems like there's room for both

patrickhulce:
hey @devtools-bot why'd you change this :P

paulirish:
Ready for another look.

PR #3774, opened 2017-11-07
core: remove optimalValue

### LINKED ISSUE #2812: Documentation: How is the optimal and the median value calculated?

patrickhulce:
closes #2812 

I'd normally oppose removing this without a major version bump, but the value was almost always an english string like `< 1,000` rather than a number that people likely relied on, we also just removed the similarly useless `category` property without a major bump so blocking this seems inconsistent

I kept the existing target references in dom-size audit since those are explicitly in the description as to why you care and its just for making sense of the data in the cards. Still removed the overall optimalValue display.

### DIFF
--- a/docs/understanding-results.md
+++ b/docs/understanding-results.md
@@ -61,7 +61,6 @@ An object containing the results of the audits, keyed by their name.
 | extendedInfo | `Object` | Extra information found by the audit. The structure of this object varies from audit to audit and is generally for programmatic consumption and debugging, though there is typically overlap with `details`. *WARNING: The structure of this object is not stable and cannot be trusted to follow semver* |
 | manual | `boolean` | Indicator used for display that the audit does not have results and is a placeholder for the user to conduct manual testing. |
 | informative | `boolean` | Indicator used for display that the audit is intended to be informative only. It cannot be passed or failed. |
-| category | `string` | No longer used. *WARNING: Deprecated, will be removed in Lighthouse 3.0* |
 
 ### Example
 ```json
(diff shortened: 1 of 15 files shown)


docs/understanding-results.md
```diff
@@ -61,7 +61,6 @@ An object containing the results of the audits, keyed by their name.
 | extendedInfo | `Object` | Extra information found by the audit. The structure of this object varies from audit to audit and is generally for programmatic consumption and debugging, though there is typically overlap with `details`. *WARNING: The structure of this object is not stable and cannot be trusted to follow semver* |
 | manual | `boolean` | Indicator used for display that the audit does not have results and is a placeholder for the user to conduct manual testing. |
 | informative | `boolean` | Indicator used for display that the audit is intended to be informative only. It cannot be passed or failed. |
-| category | `string` | No longer used. *WARNING: Deprecated, will be removed in Lighthouse 3.0* |
```
patrickhulce: This should've been cleaned up when category was removed, no references to optimalValue since we didn't want anyone to use it :)

PR #3408, opened 2017-09-25
adopt yargsParser directly for chromeFlags parsing

paulirish:
yargs has some overzealous parsing. by using yargs-parser directly, we can turn off boolean & camelCasing. Still have to add back quotes, but it doesn't seem too bad.

basically the alternative impl of #2754 (which was for #2753)

fixes #3003

PR #3775, opened 2017-11-07
cli: do not double quote chromeFlags

### LINKED ISSUE #3744: Overriding user agent using --chrome-flags adds additional quotes to user agent
For example:
Using --chrome-flags="--user-agent=testing123"
Will result a request with the following user-agent: "testing123" (including the quotes), basically every useragent will be automatically surrounded with quotes.

patrickhulce:
closes #3744

### DIFF
--- a/lighthouse-cli/run.ts
+++ b/lighthouse-cli/run.ts
@@ -40,7 +40,10 @@ export function parseChromeFlags(flags: string = '') {
       // Avoid '=true', then reintroduce quotes
       .map(key => {
         if (parsed[key] === true) return `--${key}`;
-        return `--${key}="${parsed[key]}"`;
+        // ChromeLauncher passes flags to Chrome as atomic arguments, so do not double quote
+        // i.e. `lighthouse --chrome-flags="--user-agent='My Agent'"` becomes `chrome "--user-agent=My Agent"`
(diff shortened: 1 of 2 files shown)


patrickhulce:
Link just takes me to the commit so not entirely sure if you're talking about https://github.com/GoogleChrome/lighthouse/pull/3408#discussion_r140888688, but the short story here is that quoting *any* value is indeed bad, not that quoting single words is bad but spaced values is fine.

> So basically it's taken care of since spawn passes in arguments as an array, so they don't need their values quoted? 

Correct

> This is confusing because the fix in #2754 needed quotes, but maybe the spawn method or our handling of chromeFlags has changed since then.

Nothing has changed, but it is confusing. #2754 was about us needing to respect quoted values when *parsing* because we just did a basic `s.split(' ')`. This issue is about us needing to not add superfluous quotes when sending along the parsed values. 

All of this would go away if we weren't using spawn and just doing a shell `chrome ${theRawFlags}`, but because we parse and pass as array it raises all these problems

> maybe the tests can do with some cleaning up?

ya done

brendankenny:
appveyor is being dumb missing that offscreen svg again. Ignoring for now

dickeylth:
When could this merge be published? I've encountered the `--user-data-dir` wrapped with quotes, which breaks my program.

Issue #3731, opened 2017-11-02
ImageUsage gatherer: Cannot read property 'tagName' of null


patrickhulce:
Encountered in #2293 on https://www.victoriassecret.com/

Audit error: Required ImageUsage gatherer encountered an error: Cannot read property 'tagName' of null

![image](https://user-images.githubusercontent.com/2301202/32355407-a95b406c-bfeb-11e7-8042-0afad79b1495.png)

comes from line 48 where we grab the parentElement of the image 

should be `!!element.parentElement && element.parentElement.tagName === 'PICTURE'`
https://github.com/GoogleChrome/lighthouse/blob/cfc2afe07eb8e23b8ff07d90760b0cfbe24be85c/lighthouse-core/gather/gatherers/image-usage.js#L34-L52

ThisIzKp:
@patrickhulce I'm picking up this
Hopefully.. I'll not get stuck this time 😜

AkshayIyer12:
@patrickhulce I want to work on the issue.

patrickhulce:
great thanks for volunteering @AkshayIyer12! Let us know if you run into any issues :)

AkshayIyer12:
@patrickhulce I have done the change locally on my system
```!!element.parentElement && element.parentElement.tagName === 'PICTURE' ``` 
But I'm not sure how to proceed further.

patrickhulce:
correct @AkshayIyer12 thanks for fixing!

next steps for you are to...

1. commit your code with appropriate message, `core(image-usage): add null check for parentElement` is fine
2. push your code to your fork on github, assuming you cloned from your fork the command would be something like `git push -u origin name_of_your_branch`, GitHub has some [intro git tutorials](https://try.github.io/levels/1/challenges/1) if you're unsure what the branch, fork, and push terms are
3. open a pull request on our repo pointing to the branch on your fork
4. wait for us to review and address any comments that come up
5. dance and celebrate your first 💡 🏠  contribution! 🎉  😄

patrickhulce:
fixed in #3779

atifwaqar:
I am getting the following error on Auditing my page : 
![image](https://user-images.githubusercontent.com/4388707/35200289-79443ad2-ff0b-11e7-88b8-9122d66cc4c0.png)

To me it looks like similar or same issue. And I ran the Audit from Chrome Dev tools. 

[...]

PR #3779, opened 2017-11-08
core(image-usage): add null check for parentElement

AkshayIyer12:


AkshayIyer12:
I signed it!

wardpeet:
@AkshayIyer12 Thanks for fixing this bug! Do you have a webpage where this audit fails? is it possible to create a unit test for this one?

patrickhulce:
@wardpeet yeah the details are in #3731 :)

unfortunately a unit test is tricky since its eval'd in the browser, but a smoke test would be nice

PR #3804, opened 2017-11-11
tests(smoke): add long task to byte-efficiency tester to deflake appveyor

brendankenny:
`large.svg` starts loading almost exactly at FI on appveyor, leading to it to be pretty flakey since `offscreen-images` doesn't count images that start loading after FI. This PR just adds an artificial long task to push FI out a few more seconds to deflake things.

So far it's been passing consistently on appveyor over a few loads; will rerun multiple more times.

Also: since the `extendedInfo` is sorted by image size, added more useful expectations using the image file names, which should hopefully make the diff a little clearer for any future errors.


Bgshanoams:
Vvms is slow of shit...

PR #3826, opened 2017-11-15
docs: add lantern accuracy data

### LINKED ISSUE #3691: Compare Lantern results to WPT throttled output
To continue our analysis on Lantern results, let's compare Lantern output to the WPT throttled output over the 1.5K URL dataset that we've created.

patrickhulce:
closes #3691 

before rushing off to improve accuracy, the next AI is on @vinamratasingal to draw up document on what we actually want to achieve accuracy with

### DIFF
--- a/docs/lantern.md
+++ b/docs/lantern.md
@@ -32,7 +32,9 @@ Stats were collected using the [trace-evaluation](https://github.com/patrickhulc
 ## Conclusions
 
 ### Lantern Accuracy Conclusions
-Definitive conclusions on repeat view accuracy require much more data for the same URLs (i.e. more than 1 run for each URL per environment), but for the single view use case, Lantern is roughly as accurate at predicting the rank of a website the next time you visit it as the metrics themselves which is the highest goal we set out to achieve. As a sanity check, we also see that using the unthrottled metrics to predict the rank of throttled performance has a significantly lower rank correlation than Lantern.
+
+* For the single view use case, we conclude that Lantern is roughly as accurate at predicting the rank of a website the next time you visit it as the metrics themselves. That is to say, the average error we observe between a Lantern performance score and a LH on DevTools performance score is within the expectation for standard deviation, which is the highest goal we set out to achieve. As a sanity check, we also see that using the unthrottled metrics to predict throttled performance has a significantly lower correlation than Lantern does.
+* For the repeat view use case, we require more data to reach a conclusion, but the high correlation of the single view use case suggests the accuracy meets our correlation requirements even if some sites may diverge.
 
 ### Metric Variability Conclusions
 The reference stats demonstrate that there is high degree of variability with the user-centric metrics and strengthens the position that every load is just an observation of a point drawn from a distribution and to understand the entire experience, multiple draws must be taken, i.e. multiple runs are needed to have sufficiently small error bounds on the median load experience.
--- a/docs/lantern.md
+++ b/docs/lantern.md
@@ -1,0 +1,32 @@
+# Lantern
+
+## Overview
+
+Project Lantern is an ongoing effort to reduce the run time of Lighthouse and improve audit quality by modeling page activity and simulating browser execution.
+
+## Accuracy
+
+All of the following accuracy stats are reported excluding the 10% tail as the initial research found approximately ~10% of sites will radically vary simply by visiting the page a second time through no fault of the metrics or prediction logic. This means the accuracy is slightly overstated but should still hold for the  controlled-enivornment/repeated view use case.
+
+Stats were collected using the [trace-evaluation](https://github.com/patrickhulce/lighthouse-trace-evaluations) scripts. Table cells contain [Spearman's rho](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient) and [MAPE](https://en.wikipedia.org/wiki/Mean_absolute_percentage_error) for the respective metric.
+
+| Comparison | FCP | FMP | TTI |
+| -- | -- | -- | -- |
+| Lantern predicting Default LH | .850 : 19.6% | .866 : 21.0% | .907 : 26.9% |
+| Lantern predicting LH on WPT | .764 : 34.4% | .795 : 32.5% | .879 : 33.1% |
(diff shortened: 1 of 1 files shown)


docs/lantern.md
```diff
@@ -0,0 +1,32 @@
...
+
+All of the following accuracy stats are reported excluding the 10% tail as the initial research found approximately ~10% of sites will radically vary simply by visiting the page a second time through no fault of the metrics or prediction logic. This means the accuracy is slightly overstated but should still hold for the  controlled-enivornment/repeated view use case.
+
+Stats were collected using the [trace-evaluation](https://github.com/patrickhulce/lighthouse-trace-evaluations) scripts. Table cells contain [Spearman's rho](https://en.wikipedia.org/wiki/Spearman%27s_rank_correlation_coefficient) and [MAPE](https://en.wikipedia.org/wiki/Mean_absolute_percentage_error) for the respective metric.
+
+| Comparison | FCP | FMP | TTI |
+| -- | -- | -- | -- |
+| Lantern predicting Default LH | .850 : 19.6% | .866 : 21.0% | .907 : 26.9% |
+| Lantern predicting LH on WPT | .764 : 34.4% | .795 : 32.5% | .879 : 33.1% |
+| Lantern w/adjusted settings predicting LH on WPT<sup>1</sup> | .769 : 32.9% | .808 : 31.1% | .879 : 32.6% |
```
vinamratasingal: What settings were adjusted here? Was this the timeouts?

vinamratasingal: What settings were adjusted here?

docs/lantern.md
```diff
@@ -0,0 +1,32 @@
...
+| Default LH correlation with LH on WPT<sup>2</sup> | .808 : 30.0% | .818 : 31.3% | .819 : 39.5% |
+| Unthrottled LH correlation with LH on WPT | .643 : 36.3% | .625 : 40.1% | .731 : 58.4% |
+
+<sup>1</sup> 320 ms RTT, 1.3 mbps, 5x CPU
+
+<sup>2</sup> Default LH traces and WPT traces were captured several weeks apart, so some site changes may have occurred that skew these stats
+
```
vinamratasingal: Can we add a section explaining what conclusions we are drawing from this data, and potential reasons for why Lantern is correlating TTI on WPT but not FMP/FCP?

docs/lantern.md
patrickhulce: RTT, throughput, and CPU multipliers

I'll move the footnote that explains this to be on `adjusted settings` instead of the end of the line 👍

docs/lantern.md
```diff
@@ -0,0 +1,32 @@
+# Lantern
+
+## Overview
+
+Project Lantern is an ongoing effort to reduce the run time of Lighthouse and improve audit quality by modeling page activity and simulating browser execution.
+
+## Accuracy
+
+All of the following accuracy stats are reported excluding the 10% tail as the initial research found approximately ~10% of sites will radically vary simply by visiting the page a second time through no fault of the metrics or prediction logic. This means the accuracy is slightly overstated but should still hold for the  controlled-enivornment/repeated view use case.
```
vinamratasingal: Can we make it explicit that this was calculated based on an analysis of 1500 URLs run only *once*

patrickhulce:
@vinamratasingal I believe I have addressed your concerns, mind taking another look?

docs/lantern.md
patrickhulce: done

docs/lantern.md
patrickhulce: done-ish :) with the additional reference stats the TTI/FMP/FCP isn't necessarily an outlier that needs explaining anymore IMO, but let me know if you still think it needs some hypotheses on that

docs/lantern.md
```diff
@@ -0,0 +1,50 @@
...
+| Unthrottled LH correlation with LH on WPT | .643 : 36.3% | .625 : 40.1% | .731 : 58.4% |
+
+<sup>1</sup> 320 ms RTT, 1.3 mbps, 5x CPU
+
+<sup>2</sup> Two trace sets were captured several weeks apart, so some site changes may have occurred that skew these stats
+
+## Conclusions
+
+### Lantern Accuracy Conclusions
+Definitive conclusions on repeat view accuracy require much more data for the same URLs (i.e. more than 1 run for each URL per environment), but for the single view use case, Lantern is roughly as accurate at predicting the rank of a website the next time you visit it as the metrics themselves which is the highest goal we set out to achieve. As a sanity check, we also see that using the unthrottled metrics to predict the rank of throttled performance has a significantly lower rank correlation than Lantern.
```
vinamratasingal: Maybe I'm just being silly, but can you help me understand what "Lantern is roughly as accurate at predicting the rank of a website the next time you visit it as the metrics themselves which is the highest goal we set out to achieve." means?

patrickhulce: This sentence is just pointing out that the rank correlation of Lantern with LH is roughly the same as LH with LH, meaning that running lantern once on a URL gives you just as good of a clue as to what the next load time will be as you loading the site for real, i.e. the accuracy of Lantern is smaller than or equal to the natural deviation of load timing.

The jury is still out on how inaccurate the estimate would be if you could run it 100 times, which is identified in future work. 

Let me know which snippets from here you think are worth including or if it's still clear as mud :)

vinamratasingal: Hmmm yeah the language feels a bit unclear to me. For me, what would be helpful is to reframe this section into two bullets: 

- For the single view use case, we conclude that the rank correlation of Lantern with LH is roughly the same as LH with LH. [add 1 sentence explaining what this means based on what you said above] 

- For repeat view accuracy, we need to do more work.

patrickhulce: How's this



* For the single view use case, we conclude that the rank correlation of Lantern with standard LH is roughly the same as any the rank correlation between any two arbitrary LH runs. That is to say, the average error we observe between a Lantern performance score and a LH on DevTools performance score is within the expectated natural deviation. As a sanity check, we also see that using the unthrottled metrics to predict throttled performance has a significantly lower correlation than Lantern does.
* For the repeat view use case, we require more data to reach a conclusion, but the high correlation of the single view use case suggests the accuracy meets our correlation requirements even if some sites may diverge.

vinamratasingal: Sounds great! LGTM coming your way :)

PR #3841, opened 2017-11-17
docs: demo flags in example of programmatic use

nhodges:
In the current documentation, the example given does not pass the flags argument to `chromeLauncher.launch` which caused some initial confusion when trying to leverage any of the chromeFlags. It is not intuitive that flags should also be passed into the launch method.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -11,20 +11,23 @@ assumes you've installed Lighthouse as a dependency (`yarn add --dev lighthouse`
 const lighthouse = require('lighthouse');
 const chromeLauncher = require('chrome-launcher');
 
-function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
-  return chromeLauncher.launch(flags).then(chrome => {
-    flags.port = chrome.port;
-    return lighthouse(url, flags, config).then(results =>
+function launchChromeAndRunLighthouse(url, opts, config = null) {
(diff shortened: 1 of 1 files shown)


vinamratasingal:
Will let Paul make the final call here, but this is OK with me :)

docs/readme.md
```diff
@@ -12,7 +12,7 @@ const lighthouse = require('lighthouse');
 const chromeLauncher = require('chrome-launcher');
 
 function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
-  return chromeLauncher.launch().then(chrome => {
+  return chromeLauncher.launch(flags).then(chrome => {
```
paulirish: launch() takes an opts object

a prop of that object is `chromeFlags` .. so perhaps rename `flags` to `launchOpts` and add in a `chromeFlags: ['--disable-popup-blocking']` into it for a clear example?

paulirish:
@nhodges thanks for this PR.. but there's an error in it. do you have the time to update it? thx! 🏖

nhodges:
@paulirish Hey, thanks for the feedback! Sorry I haven't gotten to it sooner -- was in the middle of a move. (Thanks for the nudge.) I'll gladly update my PR, probably tomorrow evening.

paulirish:
@nhodges nudge ;)

PR #3844, opened 2017-11-17
report: add help link icon

### LINKED ISSUE #3696: Fix discoverability of drop downs and learn more
Based on CDS feedback, improve discoverability of drop downs and learn more in the audit report.

patrickhulce:
closes #3696 

first pass at adding some help icons, a book might be better than a question mark?
not in love with the current placement, moving it right next to the text makes them jump around and feels a bit weird, at the beginning of the line seemed to increase the noise and hurts ability to easily see what failed. definitely open to other ideas

screenshot below, first one is in hover state
image

### DIFF
--- a/lighthouse-core/report/v2/renderer/category-renderer.js
+++ b/lighthouse-core/report/v2/renderer/category-renderer.js
@@ -80,7 +80,22 @@ class CategoryRenderer {
     this._dom.find('.lh-score__description', element)
         .appendChild(this._dom.convertMarkdownLinkSnippets(description));
 
+    this._populateHelpLink(element, description);
+
     return /** @type {!Element} **/ (element);
+  }
+
+  /**
+   * @param {!DocumentFragment|!Element} element
+   * @param {string} helpText
+   */
+  _populateHelpLink(element, helpText) {
+    const firstLinkMatch = helpText.match(/]\((http[^\)]+)/);
+    const helpLink = element.querySelector('.lh-help-link');
+    if (helpLink && firstLinkMatch) {
+      helpLink.href = firstLinkMatch[1];
+      helpLink.classList.add('lh-help-link--visible');
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


paulirish:
i missed the original discussion, and have some concerns about this UI. (IMO adds too much clutter to the UI) But i understand the core problem we're trying to solve (i think)

two alternatives to this i see.

1. try to make this UI work. lower contrast with these icons.. make them gray. slide them next to the title text.
2. don't add new UI. instead follow what PSI does in making their "learn more" links more obvious by incorporating as part of the details rather than at the end of the nobody-reads description:
image


i slightly favor 2, but i have no idea which will work better for users.

vinamratasingal:
Patrick- I agree with Paul that the UI looks a *lot* more cluttered with the question marks, and I don't think will help solve our discoverability issues given that they're on the right side (where the eye will probably not naturally follow, so users will still be confused) 

Paul- I like the second suggestion, and I think we should do it, but I don't think it's going to solve to discoverability issue to begin with (because it will still require users to discover the drop down, which is not happening atm). Is there anyway we can setup an A/B experiment within the UI to test multiple options and see whether our discoverability improves? Additionally, can we try just making the arrows bigger, or maybe move their placement after the word (so like "Does not register a service worker [insert arrow]")

patrickhulce:
@vinamratasingal what do you think of the updated screenshot?

paulirish:
I can live with the updated UI. Still skeptical, but not enough to block it. :)

vinamratasingal:
Errr I think you misunderstood what I said Patrick. I actually think we should get rid of the question marks completely :)  I decided to make a screenshot to explain what I am envisioning: 

screen shot 2017-11-18 at 1 34 41 pm


Does this make sense? Can we make the dropdown a little bit more gray/darker, so there's more emphasis to it?

patrickhulce:
@vinamratasingal I didn't misunderstand. The current screenshot is the 2nd take on the UI after iterating with Paul (it was already done before you had commented just didn't update the screenshot).  I have no dog in this fight, just was doing what was assigned that we decided in a meeting that seemed reasonable :) that being said...

Stuff like moving the arrows after the sentence and tweaking shades of gray feels like a futile exercise. If we're not taking any substantial steps like moving the link outside the dropdown, I'd vote to abandon until a designer or research is involved here.

paulirish:
We can't move those arrows anytime soon.

+1 to deferring this in general.

PR #3850, opened 2017-11-18
tests: Add precommit linter hook

robdodson:
Feel free to reject this, I just did it for fun after one of my PRs failed the Travis linter :)

This adds a precommit hook that runs eslint against any staged `.js` files. If the files fail eslint then the commit fails.

precommit

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "build-extension": "cd ./lighthouse-extension && yarn build",
     "build-viewer": "cd ./lighthouse-viewer && yarn build",
     "clean": "rimraf *.report.html *.report.dom.html *.report.json *.screenshots.html *.devtoolslog.json *.trace.json || true",
+    "precommit": "lint-staged",
     "lint": "[ \"$CI\" = true ] && eslint --quiet -f codeframe . || eslint .",
     "smoke": "bash lighthouse-cli/test/smokehouse/run-all-tests.sh",
     "coverage": "istanbul cover -x \"**/third_party/**\" _mocha -- $(find */test -name '*-test.js') --timeout 10000 --reporter progress --report lcovonly",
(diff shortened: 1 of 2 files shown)


wardpeet:
I like this change!

package.json
```diff
@@ -49,6 +50,9 @@
     "plots-smoke": "bash plots/test/smoke.sh",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile changelog.md --same-file"
   },
+  "lint-staged": {
+    "*.js": ["eslint"]
```
wardpeet: what about
```
"*.js": [
  "eslint --fix"
  "git add"
]
```
so we try to autofix the issues and if it's possible we also add the changed files to the git commit

robdodson: Sure I can add that.

wardpeet:
@brendankenny I'm a fan of magic ✨ 
Whenever eslint fails you probably do an eslint --fix anyway so don't see much damage in doing it on precommit.

patrickhulce:
As someone who keeps lots and lots of untracked files around, I'm also personally not a fan, but I can see how it'd make things easier for casual/first-time contributors.

Is there an opt-out mechanism the core team could use but provide it as the default for everyone else?

robdodson:
> and with the update this is changing code on my behalf without any notice

This is why I initially didn't add the `--fix` flag. I'm curious how you feel about it if it just fails if eslint fails?

> As someone who keeps lots and lots of untracked files around, I'm also personally not a fan

This should only run against files that are staged for commit. Would that still be an issue for you?

> Is there an opt-out mechanism the core team could use but provide it as the default for everyone else?

Not that I know of...

robdodson:
Apparently you can bypass the hooks!

```
      -n, --no-verify
           This option bypasses the pre-commit and commit-msg hooks. See also
           githooks(5).
```

Just do `git commit -am 'Whatever' -n

paulirish:
I'm -1 on this as well. agree that it sucks to wait for travis to learn you didn't lint, but postcommit hooks opens up a bunch of new questions i'd rather avoid for the moment. :/

wardpeet:
@paulirish @patrickhulce 
lint-staged only lint the files that are tracked so if you have unstaged changes they are not linted. It's actually a pretty good plugin. Husky is to enable the git hooks but lint-staged just checks the files.

PR #3855, opened 2017-11-19
core(dtm-gatherer): Stop DevtoolsTimelineModel from mutating trace

### LINKED ISSUE #3702: Stop DevtoolsTimelineModel from mutating trace; move to computed artifact
This is a followup from #3563 & #3520

DevtoolsTimelineModel mutates the trace. It should not be able to. One option is to deep `Object.freeze` the trace events, but we should examine the perf overhead.

It'd also make sense to move the big taskToGroup data into a shared `/lib/` file.

Also can you rename "Parsing DOM" to "Parsing HTML & CSS"

wardpeet:
Fixes #3702 

I did a few benchmarks with and without object.freeze
```
const time = process.hrtime();
const frozenTrace = DevtoolsTimelineModel.deepFreeze(trace);
const diff = process.hrtime(time);

console.log(`Benchmark took ${diff[0] * 1000 + diff[1] / 1000000} miliseconds`);
```
and
```
const time = process.hrtime();
const frozenTrace = trace;
const diff = process.hrtime(time);

console.log(`Benchmark took ${diff[0] * 1000 + diff[1] / 1000000} miliseconds`);
```

https://www.paulirish.com:
With Object.freeze:
```
    // Benchmark took 43.8227 miliseconds
    // Benchmark took 34.0374 miliseconds
    // Benchmark took 34.77622 miliseconds
```
Without Object.freeze:
```
    // Benchmark took 0.036716 miliseconds
    // Benchmark took 0.014257 miliseconds
    // Benchmark took 0.013771 miliseconds
```

http://m.nieuwsblad.be:
With Object.freeze:
```
    // Benchmark took 103.858498 miliseconds
    // Benchmark took 284.155426 miliseconds
    // Benchmark took 305.366242 miliseconds
```
Without Object.freeze:
```
    // Benchmark took 0.014729 miliseconds
    // Benchmark took 0.018531 miliseconds
    // Benchmark took 0.014248 miliseconds
```


lighthouse-core/audits/bootup-time.js
```diff
@@ -144,46 +61,47 @@ class BootupTime extends Audit {
...
+      const headings = [
+        {key: 'url', itemType: 'url', text: 'URL'},
+        {key: 'scripting', itemType: 'text', text: groups.scripting},
+        {key: 'scriptParseCompile', itemType: 'text', text: groups.scriptParseCompile},
+      ];
 
-    let totalBootupTime = 0;
-    const extendedInfo = {};
+      // map data in correct format to create a table
+      const results = Array.from(executionTimings).map(([url, groups]) => {
```
patrickhulce: a little confusing that this has the same name as the outer scoped `groups`

wdyt about naming the `groups` in lib/task-groups to something like `groupIdToName` or similar?

lighthouse-core/test/audits/bootup-time-test.js
```diff
@@ -7,59 +7,63 @@
 
 /* eslint-env mocha */
 const BootupTime = require('../../audits/bootup-time.js');
-const fs = require('fs');
+const DevtoolsTimelineGather = require('../../gather/computed/devtools-timeline-model');
 const assert = require('assert');
+const {groups} = require('../../lib/task-groups');
+
+const devtoolsTimelineGather = new DevtoolsTimelineGather({});
```
patrickhulce: how about we use a real `Artifacts` object to test the real flow 

i.e. similar to 
https://github.com/GoogleChrome/lighthouse/blob/ac2bc69936b2e7160f9a62b08195829762070e14/lighthouse-core/test/audits/consistently-interactive-test.js#L32-L39

paulirish:
here's a test for the computed artifact. @wardpeet can you add this?

```js
const fs = require('fs');
const pwaTrace = require('../../fixtures/traces/progressive-app.json');
const Runner = require('../../../runner.js');

const mochaReporter = require('mocha/lib/reporters/base');

describe('Speedline gatherer', () => {
  let computedArtifacts;

  beforeEach(() => {
    computedArtifacts = Runner.instantiateComputedArtifacts();
  });

  it('measures the pwa.rocks example', () => {
    return computedArtifacts.requestDevtoolsTimelineModel({traceEvents: pwaTrace}).then(model => {
      assert.equal(model.timelineModel().mainThreadEvents().length, 3157);
    });
  });

  it('does not change the orignal trace events', () => {
    // Use fresh trace in case it has been altered by other require()s.
    const pwaJson = fs.readFileSync(__dirname +
        '/../../fixtures/traces/progressive-app.json', 'utf8');
    const pwaTrace = JSON.parse(pwaJson);
    return computedArtifacts.requestDevtoolsTimelineModel({traceEvents: pwaTrace})
      .then(_ => {
        // assert.deepEqual has issue with diffing large array, so manually loop.
        const freshTrace = JSON.parse(pwaJson);
        assert.strictEqual(pwaTrace.length, freshTrace.length);
        for (let i = 0; i < pwaTrace.length; i++) {
          try {
            assert.deepStrictEqual(pwaTrace[i], freshTrace[i]);
          } catch (e) {
            // hack so we can see all diff output not fail on the first one.
            mochaReporter.list([{err: e, fullTitle: _ => ''}]);
          }
        }
      });
  });
});
```

when i disable the deepfreeze, here's the output:

image


 we can see exactly what changed in the trace. mostly linenumbers in stack traces, but also sometimes url's are removed from `FunctionCall` events. Odd.

brendankenny:
I think we need to benchmark all of Lighthouse to be sure (or at least the parts that touch the trace or trace events). It's not just creation but all code that touches frozen Arrays/Objects that will be affected (since all property setting/creation will need to be checked). I believe this has gotten better recently but may require v8 6.2+, which is Node 9+ only (good news is that DevTools/extension should be fine in Chrome 62+, though).

lighthouse-core/test/audits/bootup-time-test.js
```diff
@@ -7,59 +7,63 @@
...
 
 // sadly require(file) is not working correctly.
```
brendankenny: should update these comments since no longer need to `fs.readFileSync`

brendankenny:
> I think we need to benchmark all of Lighthouse to be sure 

#3745 will help here. We'll be able to evaluate before/after of the execution segments that touch trace events more easily after that lands

paulirish:
Added `'use strict'` to all of devtools so it would throw when attempting to change a deepfreeze object.
Tracked down 6 lines in TimelineModel.js. here they are without any context:

```js
--event.stackTrace[i].lineNumber;
--event.stackTrace[i].columnNumber;

eventData["url"] = eventData["scriptName"];
eventData["lineNumber"] = eventData["scriptLine"];

--eventData["lineNumber"];
--eventData["columnNumber"];
```

I think we can be comfortable with these mutations. as long as they're the only ones.

-------------

here's a proposal:

1. remove the object.freeze completely
1. take my test above, add it,  
1. in the test, allow differences in lineNo/colNo, or addition of `.url` property. so we're asserting trace is the same with the exception of those 3 things.  

sg everyone?

patrickhulce:
sgtm @paulirish

lighthouse-core/audits/bootup-time.js
wardpeet: what about something like `groupsEnum`?

wardpeet:
> in the test, allow differences in lineNo/colNo, or addition of .url property. so we're asserting trace is the same with the exception of those 3 things.

You use deepstrictEqual how should I fix this? Just try to set the trace values of url & linecols back to the original ones before I do the equal? If I have to loop all properties and sub properties I believe it would be too much work but maybe I should just do it

wardpeet:
@paulirish any suggestions here? :)

paulirish:
> If I have to loop all properties and sub properties I believe it would be too much work but maybe I should just do it

I think this. In the test above that's actually exactly what we're doing.  (And this was mostly copied from another existing test)

paulirish:
maybe: loop over all trace events. make a copy of each trace event. then delete those properties. then deepstrictequal the trimmed copy.

lighthouse-core/test/gather/computed/dtm-model-test.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('does not change the orignal trace events', () => {
+    // Use fresh trace in case it has been altered by other require()s.
+    const pwaJson = fs.readFileSync(__dirname +
+        '/../../fixtures/traces/progressive-app.json', 'utf8');
+    let pwaTrace = JSON.parse(pwaJson);
+    return computedArtifacts.requestDevtoolsTimelineModel({traceEvents: pwaTrace})
+      .then(_ => {
+        // assert.deepEqual has issue with diffing large array, so manually loop.
```
paulirish: nuke this. we actually need to assert at the item level.

lighthouse-core/audits/bootup-time.js
paulirish: +1 on groupsEnum.

lighthouse-core/test/gather/computed/dtm-model-test.js
```diff
@@ -0,0 +1,84 @@
...
+        const freshTrace = removeMutatedDataFromDevtools(JSON.parse(pwaJson));
+        assert.strictEqual(pwaTrace.length, freshTrace.length);
+
+        pwaTrace = removeMutatedDataFromDevtools(pwaTrace);
+        // for loop is faster than doing a deepStrictEqual on the whole trace
```
paulirish: can nuke this comment

lighthouse-core/test/gather/computed/dtm-model-test.js
```diff
@@ -0,0 +1,84 @@
...
+        // Mocha fails as it has a timeout of 2 seconds
```
paulirish: you can add a timeout to this  test like so: https://github.com/GoogleChrome/lighthouse/blob/3bdb56e80a17c2a02f73e678b906232f9a7eef02/lighthouse-cli/test/cli/run-test.js#L45

paulirish:
Also some failing tests. looks like `traceParser parser returns preact trace data the same as JSON.parse`. Interesting.

PR #3857, opened 2017-11-20
core(connection): add support to use existing tab

### LINKED ISSUE #3837: Lighthouse/Puppeteer integration
I am using lighthouse from javascript to check a few pages from a website. I would like to be able to tell lighthouse to use a specific tab that is already opened in my chrome to do that. Maybe by giving it at link to a ws endpoint. That's because I create the browser using puppeteer and I want to do some operations before to run lighthouse (like to set the useragent or some request interception configuration) and once the lighthouse check is done (like to get the html of the page or to interact with the page).

Is it possible to tell lighthouse to use a tab already existing?

Khady:
fix #3837

### DIFF
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -24,13 +24,24 @@ class CriConnection extends Connection {
     super();
     this.port = port;
     this.hostname = hostname;
+    this.existingTab = false;
   }
 
   /**
    * @override
    * @return {!Promise}
    */
-  connect() {
+  connect(tabId) {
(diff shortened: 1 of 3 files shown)


paulirish:
Awesome PR. Nice work digging into the internals here. ;)  

Let's followup some discussion in 3837

brendankenny:
I'm sorry we've let this linger for so long without progress.

Going to close since the current patch is so out of sync with master by now, but discussion continues in #3837 and #6038 -- which will hopefully end up giving us what we need for this -- or we can take this back up.

PR #3864, opened 2017-11-21
core(driver): add driver.wsEndpoint()

paulirish:
for assistance when working with puppeteer

ref #3837

### DIFF
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -32,7 +32,7 @@ class Connection {
 
 
   /**
-   * @return {!Promise}
+   * @return {!Promise<string>}
    */
   wsEndpoint() {
     return Promise.reject(new Error('Not implemented'));
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -30,6 +30,15 @@ class Connection {
     return Promise.reject(new Error('Not implemented'));
   }
 
+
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/connections/connection.js
```diff
@@ -30,6 +30,15 @@ class Connection {
     return Promise.reject(new Error('Not implemented'));
   }
 
+
+  /**
+   * @return {!Promise}
```
patrickhulce: can we still say `!Promise<string>` here just it'll always be rejected?

lighthouse-core/gather/driver.js
```diff
@@ -104,6 +104,10 @@ class Driver {
     return this._connection.disconnect();
   }
 
+  wsEndpoint() {
```
patrickhulce: copy/paste jsdoc or hopefully tsc smart enough?

brendankenny: add docs for which endpoint this is? And I can't imagine who would think otherwise, but probably worth noting that it's CRI-only

lighthouse-core/gather/driver.js
```diff
@@ -104,6 +104,15 @@ class Driver {
     return this._connection.disconnect();
   }
 
+  /**
+   * Get the browser WebSocket endpoint for devtools protocol clients like Puppeteer
```
brendankenny: maybe add some punctuation :P

PR #3876, opened 2017-11-21
docs(error-reporting): improve clarity for opt-out folks

### LINKED ISSUE #3854: Document --no-enable-error-reporting and CI env var
https://github.com/GoogleChrome/lighthouse/blob/master/docs/error-reporting.md mentions the `--no-enable-error-reporting` but the readme and `lighthouse --help` do not list this flag.

That page also mentions as `CI` env variable. Needs examples and mention of that in  `--help`.

It may also be better to rename the flag `--disable-enable-error-reporting` to be consistent with the other disable flags.

paulirish:
fixes #3854 

* docs clarity around "opt-out" and all that.
* include docs link from --help
* show docs link before providing CLI prompt

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -59,7 +59,7 @@ function getFlags(manualArgv) {
         'Configuration:')
       .describe({
         'enable-error-reporting':
-            'Enables error reporting (prompts once by default, setting this flag will force error reporting to that state).',
+            'Force enables error reporting. --no-enable-error-reporting will force disable. More: https://git.io/vFFTO',
         'blocked-url-patterns': 'Block any network requests to the specified URL patterns',
         'disable-storage-reset':
             'Disable clearing the browser cache and other storage APIs before a run',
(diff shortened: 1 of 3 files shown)


lighthouse-cli/cli-flags.js
```diff
@@ -59,7 +59,7 @@ function getFlags(manualArgv) {
         'Configuration:')
       .describe({
         'enable-error-reporting':
-            'Enables error reporting (prompts once by default, setting this flag will force error reporting to that state).',
+            'Force enables error reporting. --no-enable-error-reporting will force disable. More: https://git.io/vFFTO',
```
ebidel: "Force enables" sounds like a Jedi move. How about "Enables error reporting, overriding any saved preference...."

paulirish: > Enables error reporting, overriding any saved preference. --no-enable-error-reporting will do the opposite. More: https://git.io/vFFTO

wardpeet:
should we never prompt again or after x weeks or after each new release?

paulirish:
> should we never prompt again or after x weeks or after each new release?

I'm into the idea of prompting again if we hit that 20sec timeout, but also it may not matter with enough scale.

paulirish:
Should we ship a quick 2.6.1 with this and Brendans new PR?

PR #3886, opened 2017-11-22
misc(error-reporting): report unhandled promise rejections

paulirish:
Since we have a catch() way back in CLI, it's unlikely we'll actually have an uncaught rejection, but this is insurance anyhow.

[Here's some reported unhandled rejections](https://github.com/GoogleChrome/lighthouse/issues?q=is%3Aissue+%22Unhandled+promise+rejections+are+deprecated%22+sort%3Acreated-desc)

Q: do we want to add any extra metadata to indicate this was unhandled?

### DIFF
--- a/lighthouse-core/lib/sentry.js
+++ b/lighthouse-core/lib/sentry.js
@@ -50,6 +50,9 @@ sentryDelegate.init = function init(opts) {
         Sentry.captureException(...args, () => resolve());
       });
     };
+
+    // Report any rejections that are never caught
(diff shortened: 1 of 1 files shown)


wardpeet:
seems solid :) 👍

paulirish:
Turns out I was reading the docs for browser javascript so this is implemented the wrong way. That said, it works and generally doesn't break things.

https://docs.sentry.io/clients/node/usage/#promises
https://github.com/getsentry/raven-node/blob/master/lib/client.js#L117-L124

whoops

PR #3890, opened 2017-11-22
misc(error-reporting): tweak sentry levels and ignore list

patrickhulce:
now that we've had sentry in WPT and the volume shot up this makes some tweaks to make it a little more manageable

* adjusts error levels to more clearly see which are failed audits/which are warnings/which are fatal errors
* adds fingerprint to distinguish different protocol errors
* samples a few known errors that happen *a lot*

### DIFF
--- a/lighthouse-core/lib/sentry.js
+++ b/lighthouse-core/lib/sentry.js
@@ -20,9 +20,11 @@ const sentryApi = {
   getContext: noop,
 };
 
-const IGNORED_ERRORS = [
-  /Unable to load/, // see https://github.com/GoogleChrome/lighthouse/issues/2784
-  /Tracing already started/, // see https://github.com/GoogleChrome/lighthouse/issues/1091
+const SAMPLED_ERRORS = [
+  {pattern: /Unable to load/, rate: 0.1},
+  {pattern: /Failed to decode/, rate: 0.1},
+  {pattern: /No resource with given id/, rate: 0.01},
+  {pattern: /No node with given id/, rate: 0.01},
 ];
 
 /**
@@ -37,7 +39,6 @@ sentryDelegate.init = function init(opts) {
   }
 
   const environmentData = opts.environmentData || {};
-  const isDevelopment = environmentData.environment === 'development';
   const sentryConfig = Object.assign({}, environmentData, {allowSecretKey: true});
 
   try {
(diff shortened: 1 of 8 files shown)


lighthouse-core/lib/sentry.js
```diff
@@ -20,6 +20,11 @@ const sentryApi = {
   getContext: noop,
 };
 
+const IGNORED_ERRORS = [
```
paulirish: afaict these two aren't even being reported to sentry. why start excluding them?

lighthouse-core/lib/sentry.js
```diff
@@ -44,7 +50,19 @@ sentryDelegate.init = function init(opts) {
...
     // Special case captureException to return a Promise so we don't process.exit too early
     sentryDelegate.captureException = (...args) => {
-      if (args[0] && args[0].expected) return Promise.resolve();
+      const empty = Promise.resolve();
+      // Ignore expected errors
+      if (args[0] && args[0].expected) return empty;
+      // Ignore errors matching our shortlist
+      if (args[0] && IGNORED_ERRORS.some(regex => regex.test(args[0].message))) return empty;
+      // Only report 25% of production errors
+      if (!isDevelopment && Math.random() > 0.25) return empty;
```
paulirish: we have a fairly long tail of errors reported. and i'm seeing quite a few with < 3 reports that look worthwhile.  this makes me hesitant to throw so many away.

lighthouse-core/lib/sentry.js
```diff
@@ -44,7 +50,19 @@ sentryDelegate.init = function init(opts) {
 
     // Special case captureException to return a Promise so we don't process.exit too early
     sentryDelegate.captureException = (...args) => {
-      if (args[0] && args[0].expected) return Promise.resolve();
+      const empty = Promise.resolve();
+      // Ignore expected errors
+      if (args[0] && args[0].expected) return empty;
```
paulirish: feels like it's about time we name the first argument (and maybe second) and spread the remaining.

lighthouse-core/lib/sentry.js
paulirish: could we sample just the 'info' level events?

lighthouse-core/lib/sentry.js
patrickhulce: I marked the unable to load the page ones as ignore in sentry, is there a reason you want to track them?

lighthouse-core/lib/sentry.js
patrickhulce: fair

lighthouse-core/lib/sentry.js
paulirish: ah  unable to load is ignored.  

but tracingalreadystarted?  is that even reported?  getting a repro via CLI would be useful as it seems the only repros are with devtools/ extension

lighthouse-core/lib/sentry.js
patrickhulce: > could we sample just the 'info' level events?

yes good idea, but FWIW nothing is actually info yet :)

> we have a fairly long tail of errors reported. and i'm seeing quite a few with < 3 reports that look worthwhile. this makes me hesitant to throw so many away.

true, I might argue that given our current outstanding issues if something is <0.1% right now we likely shouldn't be investing time to fix it

seems like we need to pick at least 1 of the following:
* ignore more known issues (like falling out of memory cache)
* sample events
* upgrade our plan

lighthouse-core/lib/sentry.js
patrickhulce: alright, fair enough

just seemed like information we wouldn't actually learn anything from as we have no reason to believe it has anything to do with the URL being audited

lighthouse-core/lib/sentry.js
```diff
@@ -49,23 +50,24 @@ sentryDelegate.init = function init(opts) {
...
+      // Ignore if there wasn't an error
+      if (!err) return empty;
       // Ignore expected errors
-      if (args[0] && args[0].expected) return empty;
-      // Ignore errors matching our shortlist
-      if (args[0] && IGNORED_ERRORS.some(regex => regex.test(args[0].message))) return empty;
-      // Only report 25% of production errors
-      if (!isDevelopment && Math.random() > 0.25) return empty;
+      if (err.expected) return empty;
+      // Sample known errors that occurr at a high frequency
```
paulirish: occur

patrickhulce: heh, done

PR #3901, opened 2017-11-23
core(help-text): improve help text

daannijkamp:
Update help text meta-description audit to be more consistent with the other audits.

### DIFF
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -17,7 +17,7 @@ class Description extends Audit {
       description: 'Document has a meta description',
       failureDescription: 'Document does not have a meta description',
       helpText: 'Meta descriptions may be included in search results to concisely summarize ' +
-          'page content. Read more in the ' +
+          'page content. ' +
           '[Learn more](https://support.google.com/webmasters/answer/35624?hl=en#1).',
       requiredArtifacts: ['MetaDescription'],
     };
--- a/lighthouse-core/audits/seo/meta-description.js
+++ b/lighthouse-core/audits/seo/meta-description.js
@@ -18,7 +18,7 @@ class Description extends Audit {
       failureDescription: 'Document does not have a meta description',
       helpText: 'Meta descriptions may be included in search results to concisely summarize ' +
           'page content. Read more in the ' +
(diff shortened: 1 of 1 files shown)


patrickhulce:
Thanks so much for the first-time contribution @daannijkamp! 

The `Learn more` links aren't actually a commonality we're particularly fond of though (it also ironically breaks the SEO audit on descriptive link text 😆) The more descriptive help text of the SEO audits is quite nice IMO.

cc @rviscomi @kdzwinel I'm assuming the break from the norm on these was intentional? :)

daannijkamp:
Some more information why I suggested this:
I'm using Lighthouse to inform customers about how they can improve their website (in Dutch).
That is why I translated most audits into Dutch. Many "Learn More" links (developers.google.com) are already also available in Dutch. 
If there is a new audit (that I have not yet translated), then I can already display the "Learn More" link.
Most ideal improvement would be something like this:
```
return {
      name: 'meta-description',
      description: 'Document has a meta description',
      failureDescription: 'Document does not have a meta description',
      helpText: 'Meta descriptions may be included in search results to concisely summarize ' +
          'page content.',
      learnMore: 'https://support.google.com/webmasters/answer/35624?hl=en#1',
      requiredArtifacts: ['MetaDescription'],
    };
```

rviscomi:
If you do want to change the link to "Learn more", you also need to rewrite the sentence. As this PR is currently written, it will say "Read more in the Learn more". 🙈 

> I'm assuming the break from the norm on these was intentional?

Yes, I'd prefer more descriptive text. But I realize that most of the LH audits are written as "Learn more" so I'd also be fine with making it more consistent.

patrickhulce:
Ah thanks for the additional context @daannijkamp. Recently there came up a different need (#3844) to have dedicated docs link property on the audit too, and the additional translation benefit point you make is a good one.

We'll have to weigh that against the concern of polluting the audit property list with an unmanageable set of names, but a good idea to consider.

As for this PR specifically, if Rick is fine with consistency here and this is really the only standout, this change (sentence fixing aside) seems reasonable to me until we make all help text descriptive.

cc @paulirish who more strongly felt we should move away from "Learn more"

patrickhulce:
Assuming the thumbs up on Rick's comment means you're cool with this @paulirish :)

thanks again @daannijkamp!

PR #3941, opened 2017-11-29
tests(smokehouse): adopt URLSearchParams for querystring manipulation

paulirish:
This is what I was thinking about for #3815. 

wdyt?

### DIFF
--- a/lighthouse-cli/test/cli/run-test.js
+++ b/lighthouse-cli/test/cli/run-test.js
@@ -23,7 +23,7 @@ const getFlags = require('../../cli-flags').getFlags;
 
 describe('CLI run', function() {
   it('runLighthouse completes a LH round trip', () => {
-    const url = 'chrome://version/';
+    const url = 'chrome://version';
     const filename = path.join(process.cwd(), 'run.ts.results.json');
     const flags = getFlags(`--output=json --output-path=${filename} ${url}`);
     return run.runLighthouse(url, flags, fastConfig).then(passedResults => {
--- a/lighthouse-cli/test/cli/run-test.js
+++ b/lighthouse-cli/test/cli/run-test.js
@@ -23,7 +23,7 @@ const getFlags = require('../../cli-flags').getFlags;
 
 describe('CLI run', function() {
   it('runLighthouse completes a LH round trip', () => {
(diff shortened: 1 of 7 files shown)


lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -5,20 +5,20 @@
  */
 'use strict';
 const BASE_URL = 'http://localhost:10200/seo/';
+const URLSearchParams = require('../../../../lighthouse-core/lib/url-shim').URLSearchParams;
 
 function headersParam(headers) {
-  return headers
-    .map(({name, value}) => `extra_header=${name}:${encodeURI(value)}`)
```
patrickhulce: oh nice catch we should've been using encodeURIComponent here 👍

lighthouse-cli/test/cli/run-test.js
```diff
@@ -23,7 +23,7 @@ const getFlags = require('../../cli-flags').getFlags;
 
 describe('CLI run', function() {
   it('runLighthouse completes a LH round trip', () => {
-    const url = 'chrome://version';
+    const url = 'chrome://version/';
```
patrickhulce: hang on, I just realized the comment around this. Are we doing something wrong here?

just tried this out in console, seems like a bad result if everything gets a trailing slash now
```
const URL = require('whatwg-url').URL
console.log(new URL('chrome://version').href) // 'chrome://version'
```

paulirish: Okay looks like special urls had a trailing slash for a little bit in the spec but it was changed.

https://github.com/w3c/web-platform-tests/pull/4586
https://github.com/whatwg/url/pull/213
https://github.com/jsdom/whatwg-url/commit/9ec56188362b5775c53072ba0cfeb437c1e02709

However Chrome doesn't yet implement this change. As a result, it adds the trailing slash when it delivers the canonicalized URL over the devtools protocol. In the future this'll probably change, but for now we'll have to normalize between the spec-violating Chrome and the correct whatwg-url.

Doesn't look like there's a crbug for this, though there's a sprinkling of other issues documenting Chrome's URL failures on webplatformtests.

exciting. 🤒

Issue #3459, opened 2017-10-02
Audit: Resources are served minified


paulirish:
(Broken out from #909)

PSI currently considers HTML as well as CSS & JS. We can start with JS & CSS and later add support for considering HTML minification.

Required data:
* URLs lacking minification
* For each, potential byte savings 

More: https://developers.google.com/speed/docs/insights/MinifyResources

(Will likely take ~2 sprint to complete)

vinamratasingal:
Patrick will research the JS part of this, Brendan will finish research/start implementation on the HTML/CSS part of this audit.

patrickhulce:
looked into what's necessary for PSI-parity here. Bad news: they've built their own custom JS lexer and minifier that does some variation of what jsmin and uglify do. I see a few options here in increasing order of appeal...

* Build and maintain a PSI-equivalent JS parser and minifier - really don't want to have to do this, we're stretched thin enough as it is
* Use uglify2/babili/some other library for minification - this is going to be **really** slow, usually around 100-300kb/second so the average site with 1MB of JS will spend ~3-10 seconds in this audit alone. cnn.com for example has ~2MB of gzipped JS which explodes to ~6MB of raw JS which corresponds to ~18-60 seconds spent in this audit.
* Use some heuristics/simplified lexer to identify unminified JS - this is going to be much less accurate but if all we're really looking for is ballpark figures anyway we can probably get close enough here. A super simple lexer that just counts whitespace, comments, and local variable declarations might be good enough here.

I'd prefer to explore option 3 and see how it compares to uglify2 on some popular JS libraries.

paulirish:
status: 

* #3950 handles JS
* we plan on doing CSS with whitespace stripping only. we'll evaluate how close these numbers are
* we currently don't plan to continue with HTML minification.

PR #3950, opened 2017-11-29
new-audit(unminified-javascript): detect savings from minifcation

patrickhulce:
addresses js part of #3459 using the 3rd approach outlined

strategy: estimate minification savings by determining the ratio of the length of js tokens to overall string length

this was surprisingly accurate at identifying if a script was already minified, but is only ~6 lines using esprima and **takes ~30ms/MB to tokenize compared to ~2000ms/MB for uglify** and even longer for babel-minify

below is a table outlining the observed savings, `w/gzip` denotes the % savings after accounting for gzip, which is usually lower because minification tends to remove things that compress well

| library | uglify | uglify w/gzip  | babel-minify |  babel-minify w/gzip | our estimate | old calculation |
| - | - | - | - | - | - | - |
| angular | 86.1% | 80.3% | 85.8% | 80.3% | 74.5%  | `(74.5 + 88.8 ) / 2` |
| caltrainschedule.io | -- | -- | 50% | 50% | 31.3% | `(31.3 + 63.1) / 2` |
| react | 71.4% | 50% | 71.4% | 50% | 49.4% | `(49.4% + 75.2%) / 2` |
| lodash | 86.3% | 75% | 86.3% | 70.8% | 73.3% | `(73.3% + 87.9%) / 2` |
| jquery | 66.6% | 60% | 66.6% | 60% | 49.1% | `(49.1% + 74.4%) / 2` |
| modernizr | 76.9% | 50% | 76.9% | 50% | 67.9% | `(67.9% + 82.1%) / 2` |

identifying the already minified scripts is as easy as checking if the savings is low as no production minified script had more than 5% savings

if this all looks good, I'll go ahead and add tests, remove WIP 👍 

relevant code section for estimating minification is
https://github.com/GoogleChrome/lighthouse/blob/321df679b60210236dd8d017781317a6c3c7b748/lighthouse-core/audits/byte-efficiency/unminified-javascript.js#L34-L42


patrickhulce:
results from cnn.com flag some unminified js from turner cdn, and possibly a bug since the files even have "ugly" in the name 😆 

image

addyosmani:
I was surprised how close the margin of error was using your simplified lexer for minified savings here. Nice work, @patrickhulce!

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -0,0 +1,87 @@
...
+    let tokenLengthWithMangling = 0;
+
+    const tokens = esprima.tokenize(scriptContent);
+    for (const token of tokens) {
+      tokenLength += token.value.length;
+      // assume all identifiers could be reduced to a single character
+      tokenLengthWithMangling += token.type === 'Identifier' ? 1 : token.value.length;
+    }
+
+    if (1 - tokenLength / contentLength < IGNORE_THRESHOLD_IN_PERCENT) return null;
```
paulirish: let's add a comment to indicate this is for handling pre-minified code. \o/

patrickhulce:
aw, it finally happened :( 

image

image

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -0,0 +1,85 @@
...
+      name: 'unminified-javascript',
+      description: 'Minify JavaScript',
+      informative: true,
+      helpText: 'Minifying JavaScript files can reduce payload sizes and script parse time.' +
+        '[Learn more](https://developers.google.com/speed/docs/insights/MinifyResources).',
+      requiredArtifacts: ['Scripts', 'devtoolsLogs'],
+    };
+  }
+
+  /**
```
paulirish: can you add a comment here explaining the basic approach?  using the description from this PR works for me.

let's also call out that inline scripts are not evaluated. (i dont think it matters in practice, but since other audits of ours include inline scripts we should just be explicit)

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -0,0 +1,42 @@
...
+      .filter(record => record.resourceType() === WebInspector.resourceTypes.Script);
+
+    return scriptRecords.reduce((promise, record) => {
+      return promise
+        .then(() => {
+          return driver.getRequestContent(record.requestId)
+            .catch(_ => null)
+            .then(content => {
+              if (!content) return;
+              scriptContentMap.set(record.url, content);
```
paulirish: since we're definitely dealing with networkRecords i'd rather be using requestIds here as the key.  

over in the audit we could could then use `WebInspector.NetworkLog.requestForId()` to grab the request and pull the URL off that. wdyt?

i suppose this'll make testing slightly harder, so curious what you think.

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
patrickhulce: yeah sounds good 👍 

done

lighthouse-core/gather/gatherers/scripts.js
patrickhulce: yeah that's fair, if same URL was requested multiple times with different content we should surface that 👍

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -8,9 +8,19 @@
 const ByteEfficiencyAudit = require('./byte-efficiency-audit');
 const esprima = require('esprima');
 
-const IGNORE_THRESHOLD_IN_PERCENT = .1;
+const IGNORE_THRESHOLD_IN_PERCENT = 10;
```
paulirish: lol looks like past audits use a mix:  image

would be nice to have types for this  and milliseconds v seconds.  damn.

patrickhulce: lol, yeah I ended up switching because the `wastedPercent` value is x/100, it's the `wastedRatio` thats x/1

PR #4010, opened 2017-12-07
core(web-inspector): keep all experiments disabled

paulirish:
this fixes a conflict when LH runs in the devtools worker:

image

### DIFF
--- a/lighthouse-core/lib/web-inspector.js
+++ b/lighthouse-core/lib/web-inspector.js
@@ -30,10 +30,10 @@ module.exports = (function() {
   const _setImmediate = global.setImmediate;
 
   global.Runtime = global.Runtime || {};
-
-  // Required for devtools-timeline-model
   global.Runtime.experiments = global.Runtime.experiments || {};
-  global.Runtime.experiments.isEnabled = global.Runtime.experiments.isEnabled || (_ => false);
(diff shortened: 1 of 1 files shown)


lighthouse-core/lib/web-inspector.js
```diff
@@ -30,10 +30,10 @@ module.exports = (function() {
   const _setImmediate = global.setImmediate;
 
   global.Runtime = global.Runtime || {};
-
-  // Required for devtools-timeline-model
   global.Runtime.experiments = global.Runtime.experiments || {};
-  global.Runtime.experiments.isEnabled = global.Runtime.experiments.isEnabled || (_ => false);
+  // DevTools runtime doesn't know about some experiments that DTM looks for
+  // To avoid exceptions, we assume all experiments are disabled
+  global.Runtime.experiments.isEnabled = (_ => false);
```
patrickhulce: the fix we needed last time is just that this function existed? or did we really need to use the built-in?

patrickhulce: seems like we just needed this function to exist before https://github.com/GoogleChrome/lighthouse/pull/3497/files#diff-daa5294aa1f14b31b082e6155d79f2d3L29

PR #4020, opened 2017-12-09
report(perf-audits): adjust presentation of runtime cost audits

paulirish:
I noticed and fixed some inconsistencies

* main thread work breakdown was always showing in diagnostics, even if total cost was just 40ms. 
* bootup time audit was saying "is high" but only when  the audit was passing.
* now the failure threshold for the two audits is adjusted. 
   - just JS bootup times costing more than 2s, show it. total main thread time costing more than 4s, show that.

### DIFF
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -18,7 +18,7 @@ class BootupTime extends Audit {
     return {
       category: 'Performance',
       name: 'bootup-time',
-      description: 'JavaScript boot-up time is high (> 4s)',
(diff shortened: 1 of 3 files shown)


vinamratasingal:
Silly question- are there any UX changes to the report itself? Otherwise, I don't have any input on the logic changes.

paulirish:
* i'm changing when both audits show up vs get put into "Passed Audits"
* The title of "JavaScript boot-up time is high (> 4s)" (which you see in the report), i'm changing to just "JavaScript boot-up time".

vinamratasingal:
LGTM

PR #4032, opened 2017-12-12
docs(readme): Update wording around Calibre being a paid product

benschwarz:
The existing wording that was contributed was a bit awkward.
I think this is better, still highlights that it's a paid product by saying there's a free 14-day trial.


benschwarz:
Not sure what's wrong with commit-lint there?

PR #4035, opened 2017-12-12
core: upstream patches from downstream consumer (WIP)

paulirish:
For early review. Initial patches by jzfeng.

I expect we'll make some changes to these before landing. Let's take a look to see how we want to upstream these.

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js
@@ -16,6 +16,8 @@ const gzip = require('zlib').gzip;
 const compressionTypes = ['gzip', 'br', 'deflate'];
 const CHROME_EXTENSION_PROTOCOL = 'chrome-extension:';
 
+const GETREQUESTCONTENT_MAXWAIT = 1000;
+
 class ResponseCompression extends Gatherer {
   /**
    * @param {!NetworkRecords} networkRecords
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/gatherers/dobetterweb/response-compression.js
```diff
@@ -58,7 +58,11 @@ class ResponseCompression extends Gatherer {
 
     const driver = options.driver;
     return Promise.all(textRecords.map(record => {
-      return driver.getRequestContent(record.requestId).then(content => {
+      const contentPromise = driver.getRequestContent(record.requestId);
+      const timeoutPromise = new Promise((resolve,reject)=>{
```
patrickhulce: we should util a promise race timeout like this and do it everywhere

this brings up a good point, should we race all gatherer afterPasses at some max timeout?

lighthouse-core/lib/asset-saver.js
```diff
@@ -198,9 +198,29 @@ function saveAssets(artifacts, audits, pathWithBasename) {
...
+ * Log trace(s) and associated screenshot(s) to console.
+ * @param {!Artifacts} artifacts
+ * @param {!Audits} audits
+ * @return {!Promise}
+ */
+function logAssets(artifacts, audits) {
+  return prepareAssets(artifacts, audits).then(assets => {
+    const data = assets[0];
+    log.log('devtoolslog.json', JSON.stringify(data.devtoolsLog, null, 2));
+    log.log('screenshots.html', data.screenshotsHTML);
```
patrickhulce: this hurts a little, haha, do we want to perpetuate the screenshots html everywhere? if they don't need it I'd vote to leave out and remove the file entirely in v3 :)

lighthouse-extension/gulpfile.js
```diff
@@ -119,7 +119,6 @@ gulp.task('browserify-lighthouse', () => {
       // Do the additional transform to convert references of devtools-timeline-model
       // to the modified version internal to Lighthouse.
       bundle.transform('./dtm-transform.js', {global: true})
-      .ignore('../lighthouse-core/lib/asset-saver.js') // relative from gulpfile location
```
patrickhulce: I thought this might pull in a scary amount of stuff into the bundle, but bundleSize doesn't seem to think so, so yay! :)

paulirish:
@jianzhoufeng

paulirish:
Closing this as @jianzhoufeng will open a PR from his fork. ;)  

(Shoulda done it this way before. my bad! :)

PR #4045, opened 2017-12-13
core(optimized-images): skip mismatched mimeTypes

patrickhulce:
workaround fix for https://bugs.chromium.org/p/chromium/issues/detail?id=787701 in addition to the chrome-side fix


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -73,7 +73,9 @@ class OptimizedImages extends Gatherer {
       }
 
       seenUrls.add(record._url);
-      const isOptimizableImage = /image\/(png|bmp|jpeg)/.test(record._mimeType);
+      const isOptimizableImage = record._resourceType &&
```
wardpeet: I was thinking about adding a comment here that we should remove it later on but TBH this makes it more robus anyway so LGTM :)

PR #4048, opened 2017-12-13
core(gather-runner): covert assertPageLoaded into soft failure

### LINKED ISSUE #1978: Site Redirects HTTP Traffic to HTTPS shouldn't stop the execution if failling

patrickhulce:
closes #1978 according to feedback in #2461 

converts the assertPageLoaded into a soft failure that just fails every gatherer afterPass

if more than half of all artifacts failed due to page load failure then the error becomes fatal (like the case where you try to audit http://total-junk-hostname).

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -138,14 +138,12 @@ class GatherRunner {
   }
 
   /**
-   * Throws an error if the original network request failed or wasn't found.
+   * Returns an error if the original network request failed or wasn't found.
    * @param {string} url The URL of the original requested page.
-   * @param {{online: boolean}} driver
    * @param {!Array<WebInspector.NetworkRequest>} networkRecords
+   * @return {?Error}
    */
-  static assertPageLoaded(url, driver, networkRecords) {
-    if (!driver.online) return;
-
+  static getPageLoadError(url, networkRecords) {
     const mainRecord = networkRecords.find(record => {
       // record.url is actual request url, so needs to be compared without any URL fragment.
       return URL.equalWithExcludedFragments(record.url, url);
@@ -162,7 +160,7 @@ class GatherRunner {
       log.error('GatherRunner', errorMessage, url);
       const error = new Error(`Unable to load page: ${errorMessage}`);
       error.code = 'PAGE_LOAD_ERROR';
-      throw error;
+      return error;
     }
   }
 
@@ -260,6 +258,7 @@ class GatherRunner {
     const passData = {};
 
     let pass = Promise.resolve();
+    let pageLoadError;
 
     if (config.recordTrace) {
       pass = pass.then(_ => {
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -280,9 +279,18 @@ class GatherRunner {
       log.log('status', status);
       const devtoolsLog = driver.endDevtoolsLog();
       const networkRecords = NetworkRecorder.recordsFromLogs(devtoolsLog);
-      GatherRunner.assertPageLoaded(options.url, driver, networkRecords);
       log.verbose('statusEnd', status);
 
+      pageLoadError = GatherRunner.getPageLoadError(options.url, networkRecords);
+      // If the driver was offline, a page load error is expected, so do not save it.
+      if (!driver.online) pageLoadError = null;
```
paulirish: can we do this check within `getPageLoadError`? seems reasonable.

lighthouse-core/gather/gather-runner.js
```diff
@@ -319,8 +329,10 @@ class GatherRunner {
     const artifacts = {};
 
     // Nest LighthouseRunWarnings, if any, so they will be collected into artifact.
-    gathererResults.LighthouseRunWarnings = [gathererResults.LighthouseRunWarnings];
+    const uniqueWarnings = Array.from(new Set(gathererResults.LighthouseRunWarnings));
```
paulirish: would it be worthwhile to change `LighthouseRunWarnings` to be a set instead?

lighthouse-core/gather/gather-runner.js
```diff
@@ -336,9 +348,16 @@ class GatherRunner {
           // To reach this point, all errors are non-fatal, so return err to
           // runner to handle turning it into an error audit.
           artifacts[gathererName] = err;
+          // Track page load errors separately, so we can fail loudly if needed.
+          if (err.code === 'PAGE_LOAD_ERROR') pageLoadFailures.push(err);
         });
       });
     }, Promise.resolve()).then(_ => {
+      // Fail the run if more than 50% of all audits failed due to page load failure.
```
paulirish: "50% of all audits" => "50% of all artifacts"

patrickhulce: oh oops, yeah done :)

lighthouse-core/gather/gather-runner.js
patrickhulce: eh, it's not clear when we would do the flip back to array needed for JSON.stringify, so I'm not necessarily sold. since we were already manipulating it here and aggregating across multiple passes, seems reasonable to de-dupe but let audits keep track of it on their own

lighthouse-core/gather/gather-runner.js
patrickhulce: heh, I explicitly moved it out of there since it felt like it was in the wrong place over there :) 

in fact I'd eventually want to remove this special check entirely to make the passes on similar footing

PR #4052, opened 2017-12-14
report(a11y): Don't count non-applicable a11y audits toward score

robdodson:
This updates the accessibility report so it no longer counts non-applicable tests toward the user's score. Currently lighthouse gives folks an artificial "boost" to their a11y score because aXe was not filtering out tests which matched 0 nodes. That should be fixed in the next release of aXe (https://github.com/dequelabs/axe-core/pull/637).

This PR also updates how the number of passed/failed/notApplicable audits are reported by counting any grouped audits toward the total. Previously it was just counting the number of groups. So even though we have 35 non-manual a11y audits, it looked like we had far fewer. Here's the new look (note the 12 passed, 15 not applicable):

image

### DIFF
--- a/lighthouse-core/report/v2/renderer/category-renderer.js
+++ b/lighthouse-core/report/v2/renderer/category-renderer.js
@@ -226,12 +226,32 @@ class CategoryRenderer {
   }
 
   /**
+   * Find the total number of audits contained within a section.
+   * Accounts for nested subsections like Accessibility.
+   * @param {!Array<!Element>} elements
+   * @return {number}
+   */
+  _getTotalAuditsLength(elements) {
+    // Create a scratch element to append sections to so we can reuse querySelectorAll().
+    const scratch = this._dom.createElement('div');
+    elements.forEach(function(element) {
+      scratch.appendChild(element);
+    });
+    const subAudits = scratch.querySelectorAll('.lh-audit-group .lh-audit');
+    if (subAudits.length) {
+      return subAudits.length;
+    } else {
+      return elements.length;
+    }
+  }
+
+  /**
    * @param {!Array<!Element>} elements
    * @return {!Element}
    */
   _renderPassedAuditsSection(elements) {
     const passedElem = this._renderAuditGroup({
-      title: `${elements.length} Passed Audits`,
+      title: `${this._getTotalAuditsLength(elements)} Passed Audits`,
     }, {expandable: true});
     passedElem.classList.add('lh-passed-audits');
     elements.forEach(elem => passedElem.appendChild(elem));
@@ -244,7 +264,7 @@ class CategoryRenderer {
    */
   _renderNotApplicableAuditsSection(elements) {
     const notApplicableElem = this._renderAuditGroup({
-      title: `${elements.length} Not Applicable Audits`,
+      title: `${this._getTotalAuditsLength(elements)} Not Applicable Audits`,
     }, {expandable: true});
     notApplicableElem.classList.add('lh-passed-audits');
     elements.forEach(elem => notApplicableElem.appendChild(elem));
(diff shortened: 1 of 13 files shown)


robdodson:
I'm not sure if that travis failure is related to my changes.

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -225,19 +225,52 @@ class CategoryRenderer {
...
+   * @param {!Array<!Element>} elements
+   * @return {number}
+   */
+  _getTotalAuditsLength(elements) {
+    // Create a scratch element to append sections to so we can reuse querySelectorAll().
+    const scratch = this._dom.createElement('div');
+    elements.forEach(function(element) {
+      scratch.appendChild(element);
+    });
+    const subAudits = scratch.querySelectorAll('.lh-audit-group .lh-audit');
```
patrickhulce: can we just query for `.lh-audit` and always return that number?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -446,21 +479,27 @@ class CategoryRenderer {
...
-      const groups = auditsGroupedByGroup[groupId] || {passed: [], failed: []};
+      const groups = auditsGroupedByGroup[groupId] || {passed: [], failed: [], notApplicable: []};
 
-      if (audit.score === 100) {
-        groups.passed.push(audit);
+      if (audit.result.notApplicable) {
+        groups.notApplicable.push(audit);
       } else {
-        groups.failed.push(audit);
+        if (audit.score === 100) {
```
patrickhulce: nit: let's collapse this with an `else if`

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -484,6 +529,12 @@ class CategoryRenderer {
     const passedElem = this._renderPassedAuditsSection(passedElements);
     element.appendChild(passedElem);
 
+    // don't create a not applicable section if there are no not applicables
+    if (!notApplicableElements.length) return element;
```
patrickhulce: seems like we need to move these `if (!array.length)` checks inside some braces instead of early return

might be a current bug with no manual audits if everything fails :)

lighthouse-core/report/v2/report-generator.js
```diff
@@ -50,6 +50,7 @@ class ReportGeneratorV2 {
    * @return {number}
    */
   static arithmeticMean(items) {
+    // debugger;
```
patrickhulce: leftover :)

lighthouse-core/report/v2/report-generator.js
```diff
@@ -99,6 +100,15 @@ class ReportGeneratorV2 {
         if (typeof result.score === 'boolean') {
           auditScore = result.score ? 100 : 0;
         }
+        // If a result was not applicable, meaning its checks did not run against anything on
+        // the page, force it's weight to 0. It will not count during the arithmeticMean() but
+        // will still be included in the final report json and displayed in the report as
+        // "Not Applicable".
+        if (result.notApplicable) {
+          audit.weight = 0;
+          result.informative = true;
```
patrickhulce: super nit: can we move this line to be the last of the three? just something about me expects to see score, weight, other...in that order 😆

lighthouse-core/report/v2/renderer/category-renderer.js
robdodson: sure! would you like me to keep the failure check that returns `elements.length` if subAudits is 0?

robdodson:
_oof_, getting those tests in was rough...

Maybe I was doing something silly, but it seemed like I needed to manually edit the `sample_v2.json`. This file is a bit stale (doesn't include manual a11y audits) and even though my PR adds a new `notApplicable` property to all of the audits, I didn't want to go through and manually add it to each entry. Is it possible to generate this file from a test page and cache it?

For instance, [this test](https://github.com/GoogleChrome/lighthouse/pull/4052/files#diff-b307b0a52dbb987448f7eefda8831afcL252), as previously written, should actually fail once the manual audits are added because it just looks at the number of audit-groups and assumes it will be a certain size. It might be nicer to add a class to failed audit groups so they can specifically be selected in tests. As it stands, the only way I could differentiate failing audit groups from the others is by selecting for `div` instead of `details`. Pretty brittle :\

lighthouse-core/report/v2/renderer/category-renderer.js
patrickhulce: querying for `.lh-audit` *should* just work if I'm remember correctly, if they return different results though I'm probably misremembering and you can ignore me :)

robdodson:
I tried regenerating the sample data but it broke an unrelated performance test and i'm not entirely sure why. Maybe we can look at it when folks are back in the office.

robdodson:
looks like axe 2.6.1 is out with the notapplicable fix. let me rev and add it to this pr

robdodson:
@patrickhulce Just rebased against master because the scoring module was changed a bit. I think this is ready to be squashed and merged once travis comes back.

robdodson:
blargh why are these smoke house tests failing all of a sudden?

PR #4057, opened 2017-12-14
core(config): show SEO audits in the UI

### LINKED ISSUE #3772: Enable SEO audits in the UI
When the rest of the [SEO issues for Sprint Tres](https://github.com/GoogleChrome/lighthouse/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Sprint+Tres%3A+Oct+30+-+Nov+12%22+project%3AGoogleChrome%2Flighthouse%2F3) are completed, we should be in a good place to enable the SEO audits in the LH UI. We will continue to iterate on the remaining audits and add them to the UI as they become available.

The [SEO-only extension](https://github.com/GoogleChrome/lighthouse/issues/3656) also intended to make the SEO audits publicly accessible, so this change would replace it. The extension could still be useful in some cases when we want to get feedback on the SEO audits. The Audits panel of DevTools will still take a while to update and users can install the special purpose extension (or the full LH extension to demo all audits... but this would focus their attention only on SEO audits).

kdzwinel:
Latest copy taken from the ["Lighthouse SEO Audit Text"](https://docs.google.com/document/d/1OavQffvvYpRKAPHofIUmzHnFyYmuk_QG7xm1AoE1qs0/edit#heading=h.nphfjvuxwlo) doc.

Tested in the extension:

---

screen shot 2017-12-14 at 19 31 22
screen shot 2017-12-14 at 19 25 57

---

I made sure that all "learn more" links work and link to the right docs (that's how I found that one of the links was broken ).

Closes #3772


rviscomi:
:+1:

lighthouse-core/audits/seo/http-status-code.js
```diff
@@ -20,7 +20,7 @@ class HTTPStatusCode extends Audit {
       failureDescription: 'Page has unsuccessful HTTP status code',
       helpText: 'Pages with unsuccessful HTTP status codes may not be indexed properly. ' +
       '[Learn more]' +
-      '(https://developers.goole.com/web/tools/lighthouse/audits/successful-http-code).',
+      '(https://developers.google.com/web/tools/lighthouse/audits/successful-http-code).',
```
kdzwinel: BTW http://gooleisnotgoogle.weebly.com/

lighthouse-core/config/default.js
```diff
@@ -224,6 +224,19 @@ module.exports = {
...
+      description: 'Make sure your pages are mobile friendly so users don’t have to pinch or zoom ' +
+          'in order to read the content pages. [Learn more](https://developers.google.com/search/mobile-sites/).',
+    },
+    'seo-content': {
+      title: 'Content Best Practices',
+      description: 'Format your HTML in a way that enables crawlers to better understand your app’s content.',
+    },
+    'seo-crawl': {
+      title: 'Crawling and Indexing',
+      description: 'To appear in search results, crawlers need access to your app.',
```
kdzwinel: Hm… I wonder why these groups are not shown in the report as they are in e.g. accessibility:

screen shot 2017-12-14 at 19 56 39

screen shot 2017-12-14 at 19 56 31

patrickhulce:
a11y category has its own render logic at the moment, we'll need to refactor to make it part of default logic to render groups if they're present :)

https://github.com/GoogleChrome/lighthouse/blob/9ef858ac5121cdfee5638a1a7622f32632b98a4b/lighthouse-core/report/v2/renderer/category-renderer.js#L440-L461

lighthouse-core/audits/seo/font-size.js
```diff
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-const parseURL = require('url').parse;
+const URL = require('../../lib/url-shim');
```
kdzwinel: I missed that `require('url').parse` won't work in the extension. Thankfully it was easy to replace.

kdzwinel:
@patrickhulce thanks! I guess we don't want to create another custom renderer ATM? It doesn't make sense to make this refactor here IMO, it deserves its own PR. @rviscomi are we OK with launching w/o the categories?

rviscomi:
Yes, categories would be great to have but shouldn't block this PR.

kdzwinel:
Sorry that I kept adding stuff after opening the PR.  I should have put [WIP] in the title. Anyway, I have nothing more to add here. Thank you for reviewing it!

patrickhulce:
> Sorry that I kept adding stuff after opening the PR. I should have put [WIP] in the title. Anyway, I have nothing more to add here. Thank you for reviewing it!

no worries we all find things after the PR goes up :) this should go out with 2.7 today then!

kdzwinel:
🎉 Fantastic! Thanks a bunch.

PR #4102, opened 2017-12-20
core(network-recorder): use findNetworkQuietPeriods for networkIdle

### LINKED ISSUE #3734: Network timestamps differ significantly from NetworkRecorder observed time
Discovered in #2293, on https://www.instagram.com/p/BNQ98s4BTEa/ but probably affects the other sites as well

Summary: we knew we received network events over the protocol later than the request started in cases where main thread was busy (and thus thought waiting for CPU idle would be sufficient), but it appears that network recorder routinely receives events significantly after the request has started. This leads to network recorder believing that fewer network requests are inflight than there really are (as end events are still delivered in a timely manner).

Simplified Example from Instagram

Network record A real start 9544ms `1 inflight, 0 observed inflight`
Network record A observed start: 10111ms `1 inflight, 1 observed inflight`
Network record B real start 10115ms `2 inflight, 1 observed inflight`
Network record A real end 10135ms  `1 inflight, 1 observed inflight`
Network record A observed end: 10137ms  `1 inflight, 0 observed inflight`
Network record B observed start: 10138ms  `1 inflight, 1 observed inflight`
Network record B real end 10139ms  `0 inflight, 1 observed inflight`
Network record B observed end: 10140ms  `0 inflight, 0 observed inflight`

patrickhulce:
fixes #3734

### DIFF
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -151,18 +151,10 @@ class NetworkRecorder extends EventEmitter {
 
   onRequestWillBeSent(data) {
     // NOTE: data.timestamp -> time, data.type -> resourceType
-    this.networkManager._dispatcher.requestWillBeSent(
-      data.requestId,
-      data.frameId,
-      data.loaderId,
-      data.documentURL,
-      data.request,
-      data.timestamp,
-      data.wallTime,
-      data.initiator,
-      data.redirectResponse,
-      data.type
-    );
+    this.networkManager._dispatcher.requestWillBeSent(data.requestId,
+        data.frameId, data.loaderId, data.documentURL, data.request,
+        data.timestamp, data.wallTime, data.initiator, data.redirectResponse,
+        data.type);
   }
 
   onRequestServedFromCache(data) {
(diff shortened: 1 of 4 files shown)


lighthouse-core/lib/network-recorder.js
```diff
@@ -53,22 +130,17 @@ class NetworkRecorder extends EventEmitter {
    */
   onRequestStarted(request) {
     this.startedRequestCount++;
+    request.data._observedNodeStartTime = Date.now();
```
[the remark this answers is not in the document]
patrickhulce: we actually aren't anymore but it was useful for debugging delta between trace and epoch timestamps :)

patrickhulce:
findNetworkQuietPeriods is largely untouched other than 1 new filter that's arguably unnecessary, the meat of the PR is that we use these periods to determine isIdle and isIdle2 instead of the old counts

lighthouse-core/lib/network-recorder.js
```diff
@@ -53,22 +130,17 @@ class NetworkRecorder extends EventEmitter {
...
     const activeCount = this.activeRequestCount();
-    log.verbose('NetworkRecorder', `Request started. ${activeCount} requests in progress` +
-        ` (${this.startedRequestCount} started and ${this.finishedRequestCount} finished).`);
-
-    // If only one request in progress, emit event that we've transitioned from
-    // idle to busy.
-    if (activeCount === 1) {
-      this.emit('networkbusy');
-    }
+    log.verbose(
```
paulirish: do we want to keep these lines attached to activeRequestCount now? since our idling logic isn't related it seems better to drop the logs or have them log out the networkStatus

lighthouse-core/lib/network-recorder.js
```diff
@@ -39,11 +45,82 @@ class NetworkRecorder extends EventEmitter {
...
+    }
+  }
+
+  /**
+   * Finds all time periods where the number of inflight requests is less than or equal to the
+   * number of allowed concurrent requests.
+   * @param {!Array<!WebInspector.NetworkRequest>} networkRecords
+   * @param {number} allowedConcurrentRequests
+   * @param {number=} endTime
+   * @return {!Array<{start: number, end: number, ongoing: boolean|undefined}>}
```
paulirish: naming bikeshed! is ongoing === "in flight"? if so i'd prefer that as a name, but i suspect there's a difference already..... :)

patrickhulce: yeah they're different haha :)

it's a flag to signal that this period had no end, basically that it's `Infinity` even though we mark it at `endTime`, suggestions welcome :)

lighthouse-core/lib/network-recorder.js
patrickhulce: yeah good point I just removed all other active request count tracking since we don't really need it, moved log to network status

paulirish:
@patrickhulce before landing. should we tweak this commit message any? i feel like there's a decent size gap between the actual change and the fix. how about something like this:

> use findNetworkQuietPeriods for networkIdle calculations

land at will.

patrickhulce:
yeah name change sgtm 👍

PR #4127, opened 2017-12-27
 new-audit(unminified-css): identifies savings from unminified CSS

### LINKED ISSUE #3459: Audit: Resources are served minified

patrickhulce:
closes #3459 

don't squash this one please as removing no-old-flexbox is a relatively atomic commit :) I'll manually rebase before merge

I'm honestly not a huge fan of this audit, I'm not sure it would pass our criteria and it was quite difficult to find a site where it was triggered at all as the usual suspects were fine (CNN, The Verge, etc).

A relatively bloated Wordpress theme (http://theme.weartstudio.eu/thewire/) finally got it but only with ~40ms estimated savings, the production minified version of bootstrap only saves you 2KB after gzip which is below our regular threshold.
image

The overall approach is similar to unminified javascript, count the % of the content that is meaningful  which in this case excludes whitespace and comments. It's just that the volume of CSS is usually quite a bit lower than JS on sites. In the process I had to revive the styles gatherer and kill the no-old-flexbox which needed parsed content that made it slow.

**realization that needs to be addressed:** the styles gatherer is basically useless on its own because every time you turn on/off the CSS domain all the stylesheet ids change, so we might as well just base the audit on network records and leave the old flexbox cleanup for another time


lighthouse-core/audits/byte-efficiency/unminified-css.js
```diff
@@ -0,0 +1,143 @@
...
+        if (chars === ESCAPE_SLASH) {
+          totalTokenLength++;
+          i++;
+        } else if (char === stringStarter) {
+          isInString = false;
+          totalTokenLength++;
+          i++;
+        }
+      } else {
+        if (chars === '/*') {
```
paulirish: license headers start with `/*!`. let's leave those alone?

lighthouse-core/audits/byte-efficiency/unminified-css.js
```diff
@@ -0,0 +1,143 @@
...
+  static computeTokenLength(content) {
+    let totalTokenLength = 0;
+    let isInComment = false;
+    let isInString = false;
+    let stringStarter = null;
+
+    for (let i = 0; i < content.length; i++) {
+      const char = content.charAt(i);
+      const nextChar = content.charAt(i + 1);
+      const chars = char + nextChar;
```
paulirish: might be worth doing perf testing of these.

whats the runtime cost of this audit on a decent size site?

lighthouse-core/audits/byte-efficiency/unminified-css.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/unminified-css.js
patrickhulce: test1 https://jsperf.com/charat-master/1 proves winner is 
```
const chars = content.substr(i, 2)
const char = chars.charAt(0)
```

equals is 100x faster than regex for those two lines too 👍 

total runtime on sites with lots of css (i.e. that actually trigger savings) was ~50ms

patrickhulce:
I reverted the commit that messes with styles gatherer and just stuck this in full config for now. My plan is to immediately followup with a PR that brings both unused-css and unminified css into the default config 🎉

Issue #4039, opened 2017-12-13
Text compression test should not run for binary files


tomayac:
When you analyze https://airhorner.com, Lighthouse complains about text compression not being active for ```/sounds/airhorn.mp3```, which is a binary file.

patrickhulce:
Thanks for filing @tomayac good find!

We do currently filter to text type for this audit, but it seems like Chrome's sniffing thinks it's a text file anyway since it's issued as an XHR.
https://github.com/GoogleChrome/lighthouse/blob/5e23dbba1a7129c17d71927b9b039dbbaeedaba4/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js#L28-L34

A simple sanity check of excluding audit/video/image mime types might be a good idea here 👍

reddysridhar53:
Can I work on this?

paulirish:
@reddysridhar53 yup. we'll have to look at the contenttype response header and exclude binary ones

reddysridhar53:
@paulirish thank you, Can you please tell me how to debug ?

reddysridhar53:
@paulirish I have added the following code in the file lighthouse/lighthouse-core/gather/gatherers/dobetterweb/response-compression.js.

const isContentBinary = record.responseHeaders.find(header =>
  header.contentType.toLowerCase() === 'application/octet-stream'
);
const isTextBasedResource = record.resourceType() && record.resourceType().isTextType() && !isContentBinary;

Is my approach correct and how do I debug this?Thank you.

paulirish:
@reddysridhar53  #4144 already started to explore this. I think we'll continue from that one. thank you though!

underbyte:
Looks like this issue was address. Probably good to close.

paulirish:
fixed by #4144

PR #4143, opened 2017-12-30
core(dbw response compression): Exclude non binary files from auditing

FadySamirSadek:
I have followed the recommended approach but I am wondering if there is a better way to know the type of the file and if the extensions I added are enough and if not what can I refer to in order to get a full list of extensions.
#4039


FadySamirSadek:
I already have the CLA signed

PR #4161, opened 2018-01-02
core: extract scoring from reportgenerator into own module

paulirish:
Entirely a copy-paste job except for.... I made `generateReportJson` a static method on the class.

Partial fix for #2014

PR #4162, opened 2018-01-02
core(extension): extract a new ext-bg file, splitting extn/devtools usecases

### LINKED ISSUE #2014: lighthouse-background contains un-minified report-renderer v2

paulirish:
Cut-paste job except for:

* `updateBadgeFn` passed as arg into `runLighthouseForConnection`
* `lighthouseIsRunning` is set within the `updateBadgeUI` method now

fixes #2014 (along with #4161)

### DIFF
--- a/lighthouse-core/gather/connections/extension.js
+++ b/lighthouse-core/gather/connections/extension.js
@@ -144,7 +144,7 @@ class ExtensionConnection extends Connection {
   }
 
   /**
-   * Used by lighthouse-background to kick off the run on the current page
+   * Used by lighthouse-ext-background to kick off the run on the current page
    */
   getCurrentTabURL() {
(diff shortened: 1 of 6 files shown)


paulirish:
After these two PRs, the total lighthouse-background file size drops from 2.75MB to 2.56MB. (which is 7.42% smaller)

And obviously this also means that we have two compiled files:

* lighthouse-background.js: used by devtools
* lighthouse-ext-background.js: used by chrome extension and LRider

I'm open to other filenames if we want to bikeshed that.

patrickhulce:
FYI @jianzhoufeng LRider will need to use the lighthouse-ext-background moving forward after this change

jianzhoufeng:
Got it, thanks!

PR #4163, opened 2018-01-03
new_audit(canonical): document has a valid rel=canonical

### LINKED ISSUE #3178: [SEO Audits] Document has a valid rel=canonical
**Description:** Document has a valid `rel=canonical`
**Failure description:** Document does not have a valid `rel=canonical` `({value})`
**Help text:** Canonical links suggest which URL to show in search results. Read more in [Use canonical URLs](https://support.google.com/webmasters/answer/139066?hl=en).

**Success conditions:**
- Query selector `head > link[rel=canonical]` doesn’t match any elements; otherwise
  - `href` value of canonical link is not set to the root; otherwise
    - origins of `href` and current page are different
    - `location.pathname` is not the root.
- `href` value of canonical link is absolute

Valid examples:
- URL `example.de/`, `<link rel="canonical" href="example.com/de/">`
- URL `example.com/de/`, `<link rel="canonical" href="example.de/">`

Invalid examples:
- URL `example.com/blog/`, `<link rel="canonical" href="example.com/">`
- URL `example.de/`, `<link rel="canonical" href="example.com/">`
- URL `example.com/de/`, `<link rel="canonical" href="/">`

kdzwinel:
Closes #3178 

Success:
screen shot 2018-01-03 at 01 31 40

Failure:
screen shot 2018-01-03 at 01 47 11


lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -80,7 +80,8 @@ function requestHandler(request, response) {
         const extraHeaders = new URLSearchParams(params.get('extra_header'));
         for (const [headerName, headerValue] of extraHeaders) {
           if (HEADER_SAFELIST.has(headerName.toLowerCase())) {
-            headers[headerName] = headerValue;
+            headers[headerName] = headers[headerName] || [];
+            headers[headerName].push(headerValue);
```
kdzwinel: Some headers can appear multiple times in the HTTP response (set-cookie, link, etc.)

lighthouse-core/audits/seo/canonical.js
```diff
@@ -0,0 +1,166 @@
...
+            displayValue: `points to a different domain (${canonicalURL})`,
+          };
+        }
+
+        // another common mistake is to have canonical pointing from all pages of the website to its root
+        if (canonicalURL.origin === baseURL.origin &&
+          canonicalURL.pathname === '/' && baseURL.pathname !== '/') {
+          return {
+            rawValue: false,
+            displayValue: 'points to a root of the same origin',
```
kdzwinel: @rviscomi I made these `displayValue`s very short and while this works for `invalid URL` all others require more explanation. Not sure if `displayValue` is the right place for this explanation though (limited formatting, not much space) - maybe separate documentation page that we discussed would help here?

lighthouse-core/audits/seo/link-text.js
```diff
@@ -30,7 +30,7 @@ class LinkText extends Audit {
       description: 'Links have descriptive text.',
       failureDescription: 'Links do not have descriptive text',
       helpText: 'Descriptive link text helps search engines understand your content. ' +
-      '[Learn more](https://webmasters.googleblog.com/2008/10/importance-of-link-architecture.html)',
+      '[Learn more](https://webmasters.googleblog.com/2008/10/importance-of-link-architecture.html).',
```
kdzwinel: unrelated change, sorry to shove it here, but it was killing me

had_to_fix_it

lighthouse-core/audits/seo/canonical.js
rviscomi: Can the `failureDescription` be overwritten to have failure-specific explanations?

lighthouse-core/audits/seo/link-text.js
rviscomi: 😄 👍

lighthouse-core/audits/seo/canonical.js
```diff
@@ -0,0 +1,166 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
```
rviscomi: 🎊 

2018?

kdzwinel: @paulirish you always check those 😉 so here is a fun riddle - I created this file in 2017, but it will be merged into master in 2018. Should I put 2017 or 2018 here?

paulirish: 2017 sgtm. :)

lighthouse-core/audits/seo/canonical.js
kdzwinel: From what I tested neither `failureDescription` nor `helpText` can be overwritten ☹️ We only have `displayValue` and `details`. Unfortunately, `details` do not support any formatting when it comes to text:

screen shot 2018-01-03 at 02 52 37

lighthouse-core/audits/seo/canonical.js
```diff
@@ -0,0 +1,166 @@
...
+
+        artifacts.Hreflang.forEach(({href}) => hreflangs.push(href));
+
+        hreflangs = hreflangs
+          .filter(href => isValidRelativeOrAbsoluteURL(href))
+          .map(href => (new URL(href, baseURL)).href); // normalize URLs
+
+        if (canonicals.length === 0) {
+          return {
+            rawValue: true,
```
paulirish: when 4052 lands, we'll want to add `notApplicable: true` to this case.

lighthouse-core/audits/seo/canonical.js
rviscomi: Is the immutability of `failureDescription` a bug or a feature? cc @paulirish

lighthouse-core/audits/seo/canonical.js
kdzwinel: I added it right away (it doesn't really matter if this PR will be merge after or before #4052)

kdzwinel:
@patrickhulce we talked with @rviscomi about your idea to split this audit into two (errors with weight 1 and warnings with weight 0). On the one hand, it'd allow us to have separate titles and descriptions for both audits and provide more explanation, but on the other hand, it would create a confusing situation where user gets 100/100 but still sees that one of the audits is failing. So, if you don't have any strong feelings regarding the split we'd like to avoid it.

lighthouse-core/audits/seo/canonical.js
```diff
@@ -0,0 +1,167 @@
...
+        // bing and yahoo don't allow canonical URLs pointing to different domains, it's also
+        // a common mistake to publish a page with canonical pointing to e.g. a test domain or localhost
+        if (getPrimaryDomain(canonicalURL) !== getPrimaryDomain(baseURL)) {
+          return {
+            rawValue: false,
+            displayValue: `points to a different domain (${canonicalURL})`,
+          };
+        }
+
+        // another common mistake is to have canonical pointing from all pages of the website to its root
```
patrickhulce: still somewhat 😢  that this can fail for intentional usage. though I don't have a real life example so it's probably fine :)

lighthouse-core/test/audits/seo/canonical.js
```diff
@@ -0,0 +1,181 @@
...
+
+    return CanonicalAudit.audit(artifacts).then(auditResult => {
+      assert.equal(auditResult.rawValue, false);
+      assert.ok(auditResult.displayValue.includes('domain'));
+    });
+  });
+
+  it('fails when canonical points to the root while current URL is not the root', () => {
+    const mainResource = {
+      url: 'https://example.com/articles/cats-and-you',
```
patrickhulce: 😆

lighthouse-core/audits/seo/canonical.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        const canonicalURL = new URL(canonical);
+
+        // cross-language or cross-country canonicals are a common issue
+        if (hreflangs.includes(baseURL.href) && hreflangs.includes(canonicalURL.href) &&
+          baseURL.href !== canonicalURL.href) {
+          return {
+            rawValue: false,
+            displayValue: `points to another hreflang location (${baseURL.href})`,
```
paulirish: Rather than overloading displayValue I think we should abuse debugString like we already do for the PWA ones.

We can even extend https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/multi-check-audit.js much like the splash screen one.

(yes, debugString is supposed to be exceptions, not details of failing tests. But since the PWA audits are already violating that, we might as well violate it here, too.)

kdzwinel: TIL about the multi-check-audit! Looks cool, but it probably makes most sense for audits returning multiple failures (`Failures: ${result.failures.join(', ')}.`) and this one doesn't, so I'll just switch `displayValue`s to `debugString`s.

kdzwinel:
@paulirish thank you for taking a look at this! I addressed your debugString comment:

screen shot 2018-01-11 at 00 32 18

PR #4176, opened 2018-01-05
cli: remove legacy domhtml output

paulirish:
This was still hanging around. We can nuke it nowadays.


patrickhulce:
technically a breaking change, any rush on it or cool letting it chill? :)

paulirish:
> technically a breaking change, any rush on it or cool letting it chill? :)

found we would break this dude: https://github.com/dschmeh/lighthouseR/blob/56947d2f970ba1df2af38962f7fa5c80d8e03bb4/R/lighthouse.R

no other users i could find that we'd break.

happy to do this whenever.

Issue #3827, opened 2017-11-15
Fix report table display


paulirish:
for example:
![image](https://user-images.githubusercontent.com/39191/32865724-35dc99f0-ca19-11e7-88e9-419004fe98c3.png)

if i remove `table-layout:fixed` the columns get auto-sized and it looks WAY better:
![image](https://user-images.githubusercontent.com/39191/32865743-4b7bde6a-ca19-11e7-9709-12d00182e5ca.png)

however doing this means a long URL can extend the table to be too long:
![image](https://user-images.githubusercontent.com/39191/32865758-5f6975c2-ca19-11e7-986c-2e5a3c864941.png)

We just need some CSS massaging to make these tables more reasonable.

Related issue: Don't truncate actual text (file names) #1975

smileytechguy:
Should also be noted that the table appears incorrectly in dark mode with rows alternating dark background with white text and white background with white text.

paulirish:
@smileytechguy hows this look?

![image](https://user-images.githubusercontent.com/39191/34805786-b023cce6-f634-11e7-934e-d4e56aa17b51.png)

smileytechguy:
@paulirish actually readable! nice job!

PR #4190, opened 2018-01-06
report: more attractive table/URL rendering

paulirish:
### Before and after
image

1. you can see WAY more of the URL because good CSS.
2. meanwhile we're more aggressively eliding non-readable strings in the URL

partial fix for #3827 but its not totally complete.

### DIFF
--- a/lighthouse-core/report/v2/renderer/details-renderer.js
+++ b/lighthouse-core/report/v2/renderer/details-renderer.js
@@ -67,7 +67,7 @@ class DetailsRenderer {
     const url = text.text || '';
 
     let displayedPath;
-    let displayedHost = '';
+    let displayedHost;
     let title;
     try {
       const parsed = Util.parseURL(url);
@@ -85,16 +85,16 @@ class DetailsRenderer {
     element.appendChild(this._renderText({
       text: displayedPath,
     }));
-    const hostElem = this._renderText({
-      text: displayedHost,
-    });
-    hostElem.classList.add('lh-text__url-host');
-    element.appendChild(hostElem);
 
-    if (title) {
-      element.title = url;
+    if (displayedHost) {
+      const hostElem = this._renderText({
+        text: displayedHost,
+      });
+      hostElem.classList.add('lh-text__url-host');
+      element.appendChild(hostElem);
     }
 
+    if (title) element.title = url;
     return element;
   }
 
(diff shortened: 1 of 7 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -66,23 +66,30 @@ class DetailsRenderer {
...
     }
 
-    const element = this._renderText({
-      type: 'url',
-      text: displayedURL,
+    const element = this._dom.createElement('div', 'lh-text__url');
+    element.appendChild(this._renderText({
+      text: displayedPath,
+    }));
+    const hostElem = this._renderText({
```
patrickhulce: does this need to be conditional on `displayedHost`?

lighthouse-core/report/v2/renderer/util.js
```diff
@@ -147,8 +147,15 @@ class Util {
     }
 
     const MAX_LENGTH = 64;
-    // Always elide hash
+    // Always elide git hash
```
patrickhulce: how about just `hex hash` :)

lighthouse-core/report/v2/renderer/util.js
```diff
@@ -147,8 +147,15 @@ class Util {
...
     name = name.replace(/([a-f0-9]{7})[a-f0-9]{13}[a-f0-9]*/g, `$1${ELLIPSIS}`);
+    // Also elide other hash-like mixed-case strings // lol this is crazy
+    name = name.replace(/([a-zA-Z0-9-_]{9})(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])[a-zA-Z0-9-_]{10,}/g,
+        `$1${ELLIPSIS}`);
+    // Also elide long number sequences
+    name = name.replace(/(\d{3})\d{6,}/g, `$1${ELLIPSIS}`);
+    // Merge any adjacent ellipses
+    name = name.replace(/\u2026+/g, '\u2026');
```
patrickhulce: `\u2026` is ellipsis right, so just pass `ELLIPSIS` here like the others? :)

paulirish:
> do you know how this impacts cells that are in the url column but are actually other types like code? i'm thinking of the css usage audit in particular

good call!! it wasn't ideal. but after some fixes:

image
image
image


updated.

paulirish:
_awaiting CI results. when green, land away._

patrickhulce:
geez since when did "Unable to connect to Chrome" start failing 100% of the time??

PR #4192, opened 2018-01-06
deps(chrome-launcher): Upgrade chrome-launcher

wardpeet:
Upgrade chrome-launcher to 0.10.1.

I wasn't sure why chrome-launcher does not have a `^` in package.json so feel free to decline this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,14 +73,14 @@
     "jsdom": "^9.12.0",
     "mocha": "^3.2.0",
     "npm-run-posix-or-windows": "^2.0.2",
-    "typescript": "^2.6.1",
     "sinon": "^2.3.5",
+    "typescript": "^2.6.1",
     "zone.js": "^0.7.3"
   },
   "dependencies": {
     "axe-core": "2.6.1",
     "chrome-devtools-frontend": "1.0.422034",
-    "chrome-launcher": "0.8.1",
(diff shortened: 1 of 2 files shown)


patrickhulce:
uh-oh looks like we broke chrome-launcher typings :(

image

paulirish:
ruh roh. looks like i did indeed break the TS typings with the folder move: https://github.com/GoogleChrome/chrome-launcher/pull/69#issuecomment-350393653

paulirish:
just rebased to get the #4209 fix that was breaking this PR.

why was it breaking this PR out of the blue?  that, we'll never know.  🐩

samccone:
Is there a test we could add to chrome-launcher to prevent this?

paulirish:
> Is there a test we could add to chrome-launcher to prevent this?

yes i would love that. not how to write it, personally. you might know better than most of us. :)

wardpeet:
Not 100% sure but we can write a test like this:
```
npm pack
Unpack zip 
Parse package.json
See if typing field exists
```
By default it checks in the root directory

paulirish:
i was thinking a more functional test. e.g. some module that require()s chrome-launcher successfully typechecks with TS.

paulirish:
even though travis is busted, appveyor is happy. shipping.

PR #4198, opened 2018-01-08
report: Update vulnerability links

tkadlec:
Quick change to vulnerability links so that Snyk can serve up a version of the vuln pages with information specific to the versions Lighthouse found.


paulirish:
@tkadlec cool. are these live now? can you link up an example?

paulirish:
Looks like it's not live yet but we'll get this merged in the meantime.  Eager to see it in action. expect it to help a lot.

PR #4209, opened 2018-01-09
tests(trace-parser): use fs over require

patrickhulce:
fixes https://travis-ci.org/GoogleChrome/lighthouse/jobs/326591475

### DIFF
--- a/lighthouse-core/test/lib/traces/trace-parser-test.js
+++ b/lighthouse-core/test/lib/traces/trace-parser-test.js
@@ -40,7 +40,8 @@ describe('traceParser parser', () => {
     const parser = new TraceParser();
     let bytesRead = 0;
     // FYI: this trace doesn't have a traceEvents property ;)
-    const events = require('../../fixtures/traces/devtools-homepage-w-screenshots-trace.json');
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/lib/traces/trace-parser-test.js
```diff
@@ -76,7 +77,7 @@ describe('traceParser parser', () => {
     const streamedTrace = parser.getTrace();
 
     assert.ok(bytesRead > 256 * 1024 * 1024, `${bytesRead} bytes read`);
-    assert.strictEqual(bytesRead, 270179102, `${bytesRead} bytes read`);
+    assert.strictEqual(bytesRead, 270128965, `${bytesRead} bytes read`);
```
paulirish: yes we were actually overreporting the true size. it was the size AFTER devtools timelinemodel made its adjustments.
those adjustments detailed https://github.com/GoogleChrome/lighthouse/pull/3855#issuecomment-345807234 and in a few subsequent comments.

so now we'll not count 50kb worth of some extra `url` properties. \o/  funny.

PR #4210, opened 2018-01-09
cli(lifecycle): introduce report-only runs

paulirish:
1. If you run lighthouse with `-GA` or `-A` either one of the following, it will also save the LHR results json to disk within the `./latest-run/` folder, alongside the artifacts.
2. When running lighthouse with `-R`, it'll read that `latest-run/lhr.json` file and then generate a report and save that to disk. (It'll abide your output options or our defaults).

Also

* it turned out Printer.write() was more complicated than necessary. 
  * update: okay, [maybe previous paulirish says it _was_ necessary](https://github.com/GoogleChrome/lighthouse/commit/c9ab75dca68c1907ace71ed334db7d88c10487fa#diff-babd7547389c6ecb7aeb1ce64c8aab97L160). Will investigate! :)
* runner.js doesn't need to know about the 'latest-run' path. Moved that to be private to asset-saver.
* Unlike -G and -A, the -R flag is CLI only, and lighthouse-core is unaware of it completely. there's something nice about that.
* todo: readme update, tests

**GAR!**

### DIFF
--- a/lighthouse-cli/printer.js
+++ b/lighthouse-cli/printer.js
@@ -90,22 +90,16 @@ function writeFile(filePath, output, outputMode) {
  * @param {!LH.Results} results
  * @param {string} mode
  * @param {string} path
- * @return {!Promise<!LH.Results>}
+ * @return {!Promise<void>}
  */
 function write(results, mode, path) {
-  return new Promise((resolve, reject) => {
-    const outputPath = checkOutputPath(path);
-    const output = createOutput(results, mode);
+  const outputPath = checkOutputPath(path);
+  const output = createOutput(results, mode);
 
-    if (outputPath === 'stdout') {
-      return writeToStdout(output).then(_ => resolve(results));
-    }
-    return writeFile(outputPath, output, mode)
-      .then(_ => {
-        resolve(results);
-      })
-      .catch(err => reject(err));
-  });
+  if (outputPath === 'stdout') {
+    return writeToStdout(output);
+  }
+  return writeFile(outputPath, output, mode);
 }
 
 /**
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -123,6 +123,9 @@ function saveResults(results, artifacts, flags) {
 
   if (flags.saveAssets) {
     promise = promise.then(_ => assetSaver.saveAssets(artifacts, results.audits, resolvedPath));
+  } else if (flags.gatherMode || flags.auditMode) {
+    const latestPath = path.join(cwd, 'latest-run', 'lhr.json');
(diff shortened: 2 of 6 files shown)


lighthouse-cli/printer.js
```diff
@@ -90,22 +90,16 @@ function writeFile(filePath, output, outputMode) {
...
  * @param {string} mode
  * @param {string} path
- * @return {!Promise<!LH.Results>}
+ * @return {!Promise<void>}
  */
 function write(results, mode, path) {
-  return new Promise((resolve, reject) => {
-    const outputPath = checkOutputPath(path);
-    const output = createOutput(results, mode);
+  const outputPath = checkOutputPath(path);
```
patrickhulce: this looks just like cleanup no functional change right?

lighthouse-cli/run.js
```diff
@@ -123,6 +123,9 @@ function saveResults(results, artifacts, flags) {
 
   if (flags.saveAssets) {
     promise = promise.then(_ => assetSaver.saveAssets(artifacts, results.audits, resolvedPath));
+  } else if (flags.gatherMode || flags.auditMode) {
+    const latestPath = path.join(cwd, 'latest-run', 'lhr.json');
+    promise = promise.then(_ => Printer.write(results, Printer.OutputMode.json, latestPath));
```
patrickhulce: will we be able to do this for just `gatherMode`? seems like we can only do this with auditMode or `gatherMode === auditMode`

lighthouse-cli/run.js
```diff
@@ -159,6 +162,12 @@ function saveResults(results, artifacts, flags) {
  * @return {!Promise<!LH.Results|void>}
  */
 function runLighthouse(url, flags, config) {
+  if (flags.reportMode) {
```
patrickhulce: this means you can't really do `-GAR` for the full thing and existence of `-R` forces report-only mode no matter what, is that what we want? I guess auditMode continues on to do what reportMode would do anyhow

paulirish: hmm that's true. let's talk about what's ideal here.

lighthouse-cli/run.js
paulirish: actually yah this should instead be 

 ```js
else if (flags.auditMode) {
```

lighthouse-cli/printer.js
paulirish: yup. though i remember some bizarre rejection problems were, i think, why we complicated this method. we'll see.

brendankenny:
Definitely need to finish the *GAR*, but this patch is now super out of sync. Closing for now and with the hope that the next approach can be even simpler with our `Runner` revisions since Jan 2018.

PR #4218, opened 2018-01-09
new_audit(plugins): avoid plugins

### LINKED ISSUE #3180: [SEO Audits] Avoids plugins
**Description:** Avoids plugins
**Failure description:** Does not avoid plugins `({plugins...})`
**Help text:** Some types of media or content are not playable on mobile devices, such as license-constrained media or experiences that require Flash or other players that are not broadly supported on mobile devices. Unplayable content, when featured on a page of any website can be very frustrating for users.

**Success condition:** No plugin in the blocklist below is detected on the page.

Blocklist:
- Flash
- Silverlight
- Java

Notes:
- See previous Lighthouse [issue](https://github.com/GoogleChrome/lighthouse/issues/101) and [PR](https://github.com/GoogleChrome/lighthouse/pull/122)
- Use debugString to elaborate on which plugins were detected, also possibly the source code that included them.
- Exact detection methods for each plugin still TBD

kdzwinel:
Fixes #3180 

Success:
plugins_success

Failure:
screen shot 2018-01-09 at 01 36 16

As I found out, world of embedding external content is very messy. There are many non-standard ways of embedding that still work (multiplied by the number of different browsers 😵).

My approach here was to focus on documented ways of embedding (docs from adobe, ms, oracle, mdn) and real-life uses on popular sites (video, games, showcases). It should cover most of the real-life uses, but I'm pretty sure there are some false positives and false negatives.

Some notes:
- it's common to do feature detection when embedding such content (esp. for flash with `swfobject.js`, but there is also `silverlight.js` and some code from oracle doing the same for java). We won't be able to detect that unless user has these plugins enabled.
- flash is often embedded as `<object …><embed …></object>`. I'm listing both the `object` and `embed` as separate entries on the details list as they potentially can have different contents.

### DIFF
--- a/lighthouse-core/audits/seo/plugins.js
+++ b/lighthouse-core/audits/seo/plugins.js
@@ -66,7 +66,6 @@ class Plugins extends Audit {
    */
   static get meta() {
     return {
-      category: 'Content Best Practices',
       name: 'plugins',
       description: 'Document avoids plugins.',
       failureDescription: 'Document uses plugins',
--- a/lighthouse-core/audits/seo/plugins.js
+++ b/lighthouse-core/audits/seo/plugins.js
@@ -12,10 +12,10 @@ const JAVA_APPLET_TYPE = 'application/x-java-applet';
 const JAVA_BEAN_TYPE = 'application/x-java-bean';
 const TYPE_BLOCKLIST = new Set([
   'application/x-shockwave-flash',
-  // https://docs.oracle.com/cd/E19683-01/816-0378/using_tags/index.html
+  // See https://docs.oracle.com/cd/E19683-01/816-0378/using_tags/index.html
   JAVA_APPLET_TYPE,
   JAVA_BEAN_TYPE,
-  // https://msdn.microsoft.com/es-es/library/cc265156(v=vs.95).aspx
+  // See https://msdn.microsoft.com/es-es/library/cc265156(v=vs.95).aspx
   'application/x-silverlight',
   'application/x-silverlight-2',
 ]);
@@ -50,6 +50,7 @@ function isPluginType(type) {
  */
 function isPluginURL(url) {
   try {
+    // in order to support relative URLs we need to provied a base URL
     const filePath = new URL(url, 'http://example.com').pathname;
     const parts = filePath.split('.');
 
@@ -68,10 +69,10 @@ class Plugins extends Audit {
       category: 'Content Best Practices',
       name: 'plugins',
       description: 'Document avoids plugins.',
-      failureDescription: 'Document doesn\'t avoid plugins',
+      failureDescription: 'Document uses plugins',
       helpText: 'Some types of media or content are not playable on mobile devices. ' +
       '[Learn more](https://developers.google.com/speed/docs/insights/AvoidPlugins).',
-      requiredArtifacts: ['ExternalContent'],
+      requiredArtifacts: ['EmbeddedContent'],
     };
   }
 
@@ -80,7 +81,7 @@ class Plugins extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    const plugins = artifacts.ExternalContent
+    const plugins = artifacts.EmbeddedContent
       .filter(item => {
         if (item.tagName === 'APPLET') {
           return true;
--- a/lighthouse-core/audits/seo/plugins.js
+++ b/lighthouse-core/audits/seo/plugins.js
@@ -1,0 +1,147 @@
+/**
+ * @license Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Audit = require('../audit');
+const URL = require('../../lib/url-shim');
+
+const JAVA_APPLET_TYPE = 'application/x-java-applet';
+const JAVA_BEAN_TYPE = 'application/x-java-bean';
+const TYPE_BLOCKLIST = new Set([
+  'application/x-shockwave-flash',
+  // https://docs.oracle.com/cd/E19683-01/816-0378/using_tags/index.html
+  JAVA_APPLET_TYPE,
+  JAVA_BEAN_TYPE,
+  // https://msdn.microsoft.com/es-es/library/cc265156(v=vs.95).aspx
+  'application/x-silverlight',
+  'application/x-silverlight-2',
+]);
+const FILE_EXTENSION_BLOCKLIST = new Set([
+  'swf',
+  'flv',
+  'class',
+  'xap',
+]);
+const SOURCE_PARAMS = new Set([
+  'code',
+  'movie',
+  'source',
+  'src',
+]);
+
+/**
+ * Verifies if given MIME type matches any known plugin MIME type
+ * @param {string} type
+ */
+function isPluginType(type) {
+  type = type.trim().toLowerCase();
+
+  return TYPE_BLOCKLIST.has(type) ||
+    type.startsWith(JAVA_APPLET_TYPE) || // e.g. "application/x-java-applet;jpi-version=1.4"
+    type.startsWith(JAVA_BEAN_TYPE);
+}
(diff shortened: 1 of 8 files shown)


lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+'use strict';
+
+const Audit = require('../audit');
+const URL = require('../../lib/url-shim');
+
+const JAVA_APPLET_TYPE = 'application/x-java-applet';
+const JAVA_BEAN_TYPE = 'application/x-java-bean';
+const TYPE_BLOCKLIST = new Set([
+  'application/x-shockwave-flash',
+  // https://docs.oracle.com/cd/E19683-01/816-0378/using_tags/index.html
```
patrickhulce: nit: at a quick scan I thought these were part of the blocklist, maybe throw a "// See ..." in front of there?

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+    type.startsWith(JAVA_BEAN_TYPE);
+}
+
+/**
+ * Verifies if given url points to a file that has a known plugin extension
+ * @param {string} url
+ */
+function isPluginURL(url) {
+  try {
+    const filePath = new URL(url, 'http://example.com').pathname;
```
patrickhulce: nit: throw a comment here that we need example.com to support relative URLs, almost looks like a bug at first :)

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+class Plugins extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Content Best Practices',
+      name: 'plugins',
+      description: 'Document avoids plugins.',
+      failureDescription: 'Document doesn\'t avoid plugins',
```
patrickhulce: the double negative gets a bit wonky, wdyt about `Document uses plugins` I only found one other case of the "doesn't avoid" we should fix

the overall pattern we favor elsewhere seems to be just `Avoids plugins` and `Uses plugins`

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+        if (item.tagName === 'EMBED' && embedSrc && isPluginURL(embedSrc)) {
+          return true;
+        }
+
+        if (item.tagName === 'OBJECT' && item.data && isPluginURL(item.data)) {
+          return true;
+        }
+
+        const failingParams = item.params.filter(param =>
+          SOURCE_PARAMS.has(param.name.trim().toLowerCase()) && isPluginURL(param.value)
```
patrickhulce: is a param always guaranteed to have a name and value? ~~if not let's either guard these accesses or throw each item into a try/catch we can recover from~~ or maybe we can just do an `|| ''` in the gatherer

lighthouse-core/gather/gatherers/seo/external-content.js
```diff
@@ -0,0 +1,41 @@
...
+              value: el.getAttribute('value'),
+            })),
+        }));
+    })()`;
+
+    return options.driver.evaluateAsync(expression);
+  }
+}
+
+module.exports = ExternalContent;
```
patrickhulce: 🚲 🏠  I'm not 100% sold on `ExternalContent`, wdyt about `EmbeddedContent`?

MDN seems like it goes back and forth but titles it "Multimedia and embedding" https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Other_embedding_technologies

lighthouse-core/audits/seo/plugins.js
kdzwinel: Agreed, and since Rick also OKed it, I changed it to `Document uses plugins` as suggested.

lighthouse-core/audits/seo/plugins.js
kdzwinel: Good catch - these can be nulls. I updated the gatherer.

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,148 @@
...
+  }
+}
+
+class Plugins extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Content Best Practices',
```
patrickhulce: I think we've fully nixed `category` at this point, so should be safe to remove :)

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+        const attributes = ['src', 'data', 'code', 'type']
+          .reduce((result, attr) => {
+            if (plugin[attr] !== null) {
+              result += ` ${attr}="${plugin[attr]}"`;
+            }
+            return result;
+          }, '');
+        const params = plugin.params
+          .filter(param => SOURCE_PARAMS.has(param.name.trim().toLowerCase()))
+          .map(param => `<param ${param.name}="${param.value}" />`)
```
paulirish: technically its not
`<param ${param.name}="${param.value}" />`
but
`<param name="${param.name}" value="${param.value}" />`

but since we're already simplifying the display of these elements, i don't really think it's too important. ;)

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      name: 'plugins',
+      description: 'Document avoids plugins.',
+      failureDescription: 'Document uses plugins',
+      helpText: 'Some types of media or content are not playable on mobile devices. ' +
+      '[Learn more](https://developers.google.com/speed/docs/insights/AvoidPlugins).',
```
paulirish: This URL is a dead 404 now, and I can't find a true replacement. MFT says this: 
image

cached link is https://webcache.googleusercontent.com/search?q=cache:rke54rGwIMcJ:https://developers.google.com/speed/docs/insights/AvoidPlugins+&cd=1&hl=en&ct=clnk&gl=us

also https://developers.google.com/speed/docs/insights/UseLegibleFontSizes is gone, but they set up a redirect to WF.

@rviscomi do you have a preference where this links to? (i cc'd you on the recent CL where these docs were removed)

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+class Plugins extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      name: 'plugins',
+      description: 'Document avoids plugins.',
+      failureDescription: 'Document uses plugins',
+      helpText: 'Some types of media or content are not playable on mobile devices. ' +
```
paulirish: 'Some types of media or content are not playable on mobile devices.'
=>
Most mobile devices do not support plugins, and many desktop browsers restrict them.

lighthouse-core/audits/seo/plugins.js
```diff
@@ -0,0 +1,147 @@
...
+        return {
+          source: {
+            type: 'node',
+            snippet: `<${tagName}${attributes}>${params}</${tagName}>`,
+          },
+        };
+      });
+
+    const headings = [
+      {key: 'source', itemType: 'code', text: 'Source'},
```
paulirish: Source => Element source

lighthouse-core/audits/seo/plugins.js
rviscomi: Ideally we'd make a new doc on https://developers.google.com/web/tools/lighthouse/. I'd be ok removing "Learn more" from the help text for now until we have that doc online.

paulirish: WFM!

kdzwinel: I double checked that URL when I opened this PR (a day ago) and it was up ¯\\\_(ツ)\_/¯

lighthouse-core/audits/seo/plugins.js
kdzwinel: Oops, good catch! I agree that we are already simplifying, but it might be a bit too much. I'd change it to `<param name="${param.name}" value="${param.value}" />`

lighthouse-core/audits/seo/plugins.js
paulirish: yeah i looked and they had made the change this week. :)

lighthouse-core/audits/seo/plugins.js
kdzwinel: Just a note: we are already using 'Source' in `hreflang` and `is-crawlable`

PR #4219, opened 2018-01-09
tests(travis): remove upload artifacts

patrickhulce:
no longer need the screenshots traces, we've got more than enough examples and it's not as pressing anymore :)


paulirish:
can we revert all of https://github.com/GoogleChrome/lighthouse/commit/4f598c501060f05ef0879bfe23d9ea1e4cdfd366 ? there's a few more things (the saveAssetsPath in smokehouse, and the secure travis vars) that catch my eye.

patrickhulce:
aye, sg I suppose we can always revert revert if we need that infra again

PR #4223, opened 2018-01-09
core(audit): Add 10ms threshold to bootup-time audits

### LINKED ISSUE #4220: boot-up time shouldn't show values < 10ms
@paulirish mentioned that bootup-time audit had a list of 0ms or really low ms costs. So we want to filter out all requests that have script evaluation and script parse & compile of less than 10ms.

screenshot 2018-01-09 21 17 50

wardpeet:
Fixes #4220

Result:
screenshot 2018-01-09 22 58 23

### DIFF
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -9,6 +9,7 @@ const Audit = require('./audit');
 const WebInspector = require('../lib/web-inspector');
 const Util = require('../report/v2/renderer/util');
 const {groupIdToName, taskToGroup} = require('../lib/task-groups');
+const THRESHOLD_IN_MS = 10;
 
 class BootupTime extends Audit {
   /**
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/bootup-time.js
```diff
@@ -73,22 +74,25 @@ class BootupTime extends Audit {
...
 
       // map data in correct format to create a table
-      const results = Array.from(executionTimings).map(([url, groups]) => {
-        // Add up the totalBootupTime for all the taskGroups
-        totalBootupTime += Object.keys(groups).reduce((sum, name) => sum += groups[name], 0);
-        extendedInfo[url] = groups;
+      const results = Array.from(executionTimings)
+        .map(([url, groups]) => {
+          // Add up the totalBootupTime for all the taskGroups
+          totalBootupTime += Object.keys(groups).reduce((sum, name) => sum += groups[name], 0);
```
patrickhulce: @paulirish should we create an "Other" entry in the table so the total still adds up?

paulirish: ehhhh we _could_. wdyt?

addyosmani: I'm in favor of "Other" unless there are better ideas

paulirish: I looked into this and the table already doesn't sum to the total time displayed. the total time shown at the top includes all non-script time (layouts/paints/etc)

so if we really wanted to also capture those, we might do something like 

image

but it's kinda wonky.  

I'm in favor of skipping the "other" row(s)

patrickhulce: Alright that's fine we can skip too

wardpeet: So what's the advice here? 😋

PR #4224, opened 2018-01-10
tests: remove global timeout, set timeouts individually

paulirish:
This should make unexpected hangs more obvious and potentially failing/long tests will fail faster.


patrickhulce:
skip CI is on, but it passes? :)

paulirish:
Yeah. It did pass. Then I added the last commit. Is good though.

PR #4231, opened 2018-01-10
report(grammar): do not finalize audit titles with a period

paulirish:
Currently we're very inconsistent when it comes to the trailing period on audit titles:

image

Kayce sez no period on these. So this PR removes them.


robdodson:
Another thing worth considering down the line... Linking phrases like "Learn more" isn't a good practice because if a screen reader user navigates by link it'll just say "learn more, learn more, etc." I can go back and improve my links, but I'd want other audits to do the same so the report feels consistent.

vinamratasingal:
Thanks for taking care of this Paul :)

Issue #4262, opened 2018-01-15
Chrome Extension Options has a scrollbar


coliff:
The Chrome Extension has an options screen with a scrollbar (it didn't appear until the SEO audit was added). Increasing the height by 10 pixels would make scrollbar not appear and look a bit nicer.

![http://oi64.tinypic.com/10hsjec.jpg](http://oi64.tinypic.com/10hsjec.jpg)

PR #4263, opened 2018-01-15
extension: Fix scrollbar from showing on Options page

coliff:
Fixes: https://github.com/GoogleChrome/lighthouse/issues/4262


paulirish:
Coolness. thanks!

Issue #1449, opened 2017-01-11
Performance difference between extension and cli report


pedro93:
Hi 
I was trying out both the lighthouse extension and NodeJS Cli and noticed that for the same external URL, the reports vary greatly between the two. I made sure to clean caches and be logged out of my account when testing the extension on all executions.

The URL in question was https://github.com

### Extension output:
![screen shot 2017-01-11 at 16 42 59](https://cloud.githubusercontent.com/assets/3718654/21857472/772cdd50-d81d-11e6-9e8c-aaad7f29f51e.png)
![screen shot 2017-01-11 at 16 43 08](https://cloud.githubusercontent.com/assets/3718654/21857473/774b2a6c-d81d-11e6-9e42-e88699b46abe.png)

### Cli output:
![screen shot 2017-01-11 at 16 45 22](https://cloud.githubusercontent.com/assets/3718654/21857487/8a70872c-d81d-11e6-8eb7-737cbf846299.png)
![screen shot 2017-01-11 at 16 44 21](https://cloud.githubusercontent.com/assets/3718654/21857488/8a715f4e-d81d-11e6-8e3c-868e1c3d4a50.png)

Note that reports made by the cli and extensions were consistent amongst themselves, I ran 10 tests on each time, these examples are merely representative, no network connections were changed in the process, nor was any flag/option explicitly used on either case.

### Hardware/Software information:
Macbook Pro (Retina 13" Early 2015)
OS: OS X El Captain v.10.11.6
Chrome: 55.0.2883.95 (64-bit)
Node: 6.9.4
Lighthouse: 1.4.1

Is this a matter of varying execution configurations or a potential bug?

patrickhulce:
It's a known issue that the results are different and we have an issue open to warn about that (see #1426), but it's interesting that your CLI results are much *less* performant than the CRX, counter to previous issues that suggest other extensions, scripts, etc interfere only when using the extension... @ebidel any other reports of CLI producing consistently less performant results?

brendankenny:
There are some things Chrome does at startup that can drain performance from whatever page we're testing. #1416 helps disable a few more of those in the CLI and will be in the next release.

An interesting test to see how well we're eliminating these things would be to run `chrome-debug` to launch Chrome, let it settle for a minute or two (maybe even load an unrelated page and let it settle there, e.g. https://example.com), then see how the numbers compare.

ebidel:
[...]

PR #4310, opened 2018-01-19
core(gather-runner): load a blank data URI, rather than about:blank

### LINKED ISSUE #3707: [perf] reduce wasted time on about:blank
In total we spend ~1.2 seconds just sitting on about:blank. If we switched to an empty data URI as previously attempted, we could wait for `load` event and ditch the arbitrary setTimeout.

paulirish:
fixes #3707

This is a lot nicer. This page definitely fires onload and DCL, and does so fast. (~35ms after navigation on my machine)

In total this saves 500ms on this run: `time lighthouse -G http://example.com`

This also deprecates config.blankDuration, which [WPT is using](https://github.com/WPO-Foundation/webpagetest/blob/4234baa351eb1cb9d6b00dc513133c24f1d5a990/agent/js/src/wd_server.js#L1051)..  
Also fwiw their blankPage is this: `<html><head><style>body {background-color: #DE640D;}</style></head></html>`

@patrickhulce how you want to handle the private method? shall we expose with the same API?


lighthouse-core/gather/gather-runner.js
```diff
@@ -64,8 +64,13 @@ class GatherRunner {
    * @param {number=} duration
    * @return {!Promise}
    */
-  static loadBlank(driver, url = 'about:blank', duration = 300) {
-    return driver.gotoURL(url).then(_ => new Promise(resolve => setTimeout(resolve, duration)));
+  static loadBlank(driver, url) {
+    // The real about:blank doesn't fire onload and is full of mysteries
+    //   https://github.com/whatwg/html/issues/816#issuecomment-288931753
+    // To improve speed and avoid anomalies, we use a basic data uri blank page
+    url = url || 'data:text/html,<!doctype html><title>Resetting Page...</title>';
```
patrickhulce: I actually kind of love that WPT's is a non-white color to signal when Chrome gets the first byte by turning white but not sure if that will cause issues :)

paulirish:
K. Let me wrestle with the smoke tests first. :)

paulirish:
Alright... new version updated. and no longer is blank, but branded! ;)


<a href="https://gfycat.com/ifr/TanFluffyArieltoucan">
Click to view the GIF:

</a>

wardpeet:
The screenshot is amazing! We should have done this earlier

lighthouse-core/gather/blank-page.html
```diff
@@ -0,0 +1,23 @@
...
...dd835"/><circle class="cls-11" cx="328.678" cy="9.394" r="3.53" id="circle69" fill="#fdd835"/><circle class="cls-11" cx="640.898" cy="179.204" r="3.53" id="circle73" fill="#fdd835"/><circle class="cls-11" cx="747.868" cy="90.754" r="3.53" id="circle75" fill="#fdd835"/></g><g id="layer3" transform="translate(-111.07 296.27)"><path class="cls-2" d="M475.488-83.896l-334.62-47.79a3.65 3.65 0 0 0-1-.14c-8.67 0-16 31.89-16 71.15 0 39.26 7.33 71.07 16 71.07a3.66 3.66 0 0 0 .93-.13l334.64-47.88v-46.28z" id="path11" opacity=".5" fill="#ffe082"/></g><g id="layer6"><path id="polygon37" transform="translate(-42.202 118.024)" opacity=".5" fill="#304ffe" d="M493.27 173.5h-42.9V393L502 416l-8.73-242.5z"/></g><g id="layer5" transform="translate(-111.07 296.27)"><path d="M413.768 275.484c52.32 0 52.32-33.94 104.63-33.94 52.31 0 52.32 33.94 104.63 33.94 44.42 0 51.13-24.46 84.16-31.84-45.13-24.66-112.53-40.33-187.84-40.33-75.63 0-143.28 15.79-188.41 40.64 31.93 7.73 39.02 31.53 82.83 31.53z" id="path79" fill="#00c853"/><path d="M413.868 275.014c52.32 0 52.32-33.94 104.63-33.94h1.1l-.58-37.32c-74.89 0-142 15.49-187.08 39.91 31.16 7.98 38.55 31.35 81.93 31.35z" id="path81" fill="#64dd17"/></g></svg>'>
```
kdzwinel: nit: we can have an svg instead of img here: https://gist.github.com/kdzwinel/3912e659a13629bf0de7b0611953affe

vinamratasingal:
Yup this LGTM'es to me.

lighthouse-core/gather/gather-runner.js
```diff
@@ -64,8 +66,12 @@ class GatherRunner {
    * @param {number=} duration
    * @return {!Promise}
    */
-  static loadBlank(driver, url = 'about:blank', duration = 300) {
-    return driver.gotoURL(url).then(_ => new Promise(resolve => setTimeout(resolve, duration)));
+  static loadBlank(driver, url) {
+    // The real about:blank doesn't fire onload and is full of mysteries
+    //   https://github.com/whatwg/html/issues/816#issuecomment-288931753
+    // To improve speed and avoid anomalies, we use a basic data uri blank page
```
patrickhulce: can probably reference https://github.com/GoogleChrome/lighthouse/issues/1449#issuecomment-318230353 too?

paulirish: done

lighthouse-core/gather/blank-page.html
paulirish: done

paulirish:
Updated. 

I got to wrestle with a phantom/nonexisting onload event. The result is what you see in the PR now. Two separate data uri pages. 

Using `about:blank` for the first one could have worked, too. But I really didn't like the flash of white.  

PTAL

lighthouse-core/gather/gather-runner.js
```diff
@@ -55,17 +58,27 @@ const NetworkRecorder = require('../lib/network-recorder.js');
...
+  static loadBlank(driver, url) {
+    // The real about:blank doesn't fire onload and is full of mysteries (https://goo.gl/mdQkYr)
+    // To improve speed and avoid anomalies (https://goo.gl/Aho2R9), we use a basic data uri page
+    url = url || `data:text/html,${logoPageSource}`;
+    const blankPageUrl = `data:text/html,${blankPageSource}`;
+
+    // Only navigating to a single data-uri doesn't reliably trigger onload. (Why? Beats me.)
+    // Two data uris work, however the two need to be sufficiently different (Why? Beats me.)
+    // If they are too similar, Chrome considers the latter to be as superficial as a pushState
+    // Lastly, it's possible for two navigations to be racy, so we await onload inbetween.
```
patrickhulce: Maybe throw a note in here that it's still not that long since onload happens real fast for these suckers.

Also, if we eliminate the double `loadBlank` call at the beginning we would've been fine too, right?

paulirish: > Also, if we eliminate the double loadBlank call at the beginning we would've been fine too, right?

Hmmm interesting. I gotta look into why we added it: https://github.com/GoogleChrome/lighthouse/pull/850

patrickhulce: well I meant introducing something more like "if this is the first pass, don't bother trying to navigate to blank since we know we're already on it" not necessarily nuking the first one :)

paulirish: is https://github.com/GoogleChrome/lighthouse/pull/4310/commits/df175bc06409fe29b7eb24a76e37d89a816d3703 what you were thinking?

paulirish:
Before: 4 calls of `loadBlank` with a slightly higher timeout. ~1270ms per run
After: 3 calls of `loadBlank` with a shorter wait. ~630ms per run

With those 600ms of savings, I'll apply the OYB label. 😀

lighthouse-core/gather/gather-runner.js
patrickhulce: ah yes, it is! though I believe in LR they had to switch the order of passes so we can't necessarily rely on `defaultPass` being first :/ an index approach seems annoying though. 

I'm cool to revisit later if you think that's too much trouble since we've already reduced the time considerably.

patrickhulce:
I'm a little surprised it still takes 200ms for `onload` of a data URI but I guess we're doing 2 now, so really more like ~100ms?

looks like tests might need a bit of massaging still

wardpeet:
does the onload changes when the data-uri is smaller or doesn't it really matter?

lighthouse-core/gather/gather-runner.js
paulirish: yeah sg. i'll do a followup PR to use the index.

paulirish: included passIndex in this PR. 2 line change. :)

paulirish:
> I'm a little surprised it still takes 200ms for onload of a data URI but I guess we're doing 2 now, so really more like ~100ms?

yes

FWIW i did test a 10kb datauri png instead of this 3kb datauri svg and the png is slower. (~320ms  vs  ~230ms each loadBlank)

and I did trace the entire thing to attempt to understand whats happening during the 200ms, but it looks like mostly browser process work that I don't see much opportunity in. 
[trace-of-entire-run.json.zip](https://github.com/GoogleChrome/lighthouse/files/1700834/omg.trace.json.zip)

Issue #4333, opened 2018-01-23
☂️ 👣 Things to fix on next breaking change


patrickhulce:
Non-Comprehensive List of Things We Should Fix on Next Breaking Change 

- [ ] rename "uses-request-compression" to "uses-response-compression" :)
- [ ] Use LHError for all thrown errors when possible
- [ ] Remove unused/unsupported config flags
- [ ] Remove cache-start-url audit (unused)

wardpeet:
Should we upgrade to node 8 as well?

patrickhulce:
@ebidel yup for sure, both these target #1999
*  "Kill off" extendedInfo and make details real and good.
* de-dupe all audit results

brendankenny:
we should remove `chrome-launcher/`. Kept only to preserve various places we suggested calling deep into our directory structure

PR #4344, opened 2018-01-24
cli(yarn): Add mixed-content yarn script

christhompson:
This adds a new yarn script (as discussed in #3953) which simplifies running the custom configuration required for the mixed-content audit.

@paulirish Could you PTAL to see if this is what you had in mind?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "bundlesize": "bundlesize",
     "plots-smoke": "bash plots/test/smoke.sh",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile changelog.md --same-file",
-    "type-check": "tsc -p ."
+    "type-check": "tsc -p .",
+    "mixed-content": "./lighthouse-cli/index.js --chrome-flags='--headless' --config-path=./lighthouse-core/config/mixed-content.js"
   },
   "devDependencies": {
     "@types/configstore": "^2.1.1",


christhompson:
FYI I don't have write access to merge this, so if this looks good to go could a core member merge it? Thanks!

paulirish:
As for someone consuming this after a global install of LH... i guess they would do:

```sh
npm explore lighthouse -- npm run mixed-content
```

paulirish:
@christhompson  I published a prerelease so we can test out this flow:

```sh
npm install --global lighthouse@canary
npm explore --global lighthouse -- npm run mixed-content http://bbc.com
```

This works for me, though it saves the result in a funny spot (because of the explore call resetting CWD, i guess):

>   Printer html output written to /Users/paulirish/.homebrew/lib/node_modules/lighthouse/www.bbc.com_2018-02-06_12-46-36.report.html  

But aside from that it seems roughly fine.

christhompson:
Thanks for publishing the canary release -- that will help with testing the final documentation.

The working-dir issue seems unfortunate but maybe okay overall. It might be best if the example invocation we give includes `--view` to automatically open it. The concern then would be that it is storing them in a directory where the user would have trouble clearing them out.

It might work better to recommend the flow, which would leave the user's shell in the lighthouse directory after running?
```
npm install --global lighthouse
npm explore lighthouse
npm run mixed-content http://www.bbc.com
```

paulirish:
@christhompson hmmm that sounds cool, but it's not working out so hot on my machine:

image


i'm gonna see if i can trick LH into saving the file locally.

paulirish:
okay got a solution:

```sh
npm explore --global lighthouse -- npm run mixed-content -- --output-path=$PWD http://bbc.com
```

this ALMOST works and will work once #4369 is landed. but still, the experience here is pretty ugly.

maybe a new commandline flag `--mixed-content` is the right answer?

christhompson:
If you're okay with using flags for this, it would definitely make invoking it a lot simpler. I don't want to make precedent for adding a laundry list of flags though.

I've uploaded a quick PR for adding a flag to do this, but there may a cleaner way to handle the alternate config than what I did in lighthouse-core/index.js. See #4441

patrickhulce:
> I don't want to make precedent for adding a laundry list of flags though.

We've already added rethinking how we invoke non-default configs to our thoughts for #4333 on next breaking change, so if you're cool with this fix in 2.x releases then sgtm :)

christhompson:
sgtm

PR #4362, opened 2018-01-25
new_audit(preconnect): add preconnect audit

wardpeet:
- [x] Filter same origin
- [x] Filter all records requested by main resource
- [x] Filter all non http(s) records (urls with no origin)
- [x] Filter out allready connected origins (connection time is 0)
- [x] Filter out all records that took place after 10s
- [x] Added unit tests
- [x] Added smoketests

image


paulirish:
> Initiate key requests early

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+    const devtoolsLogs = artifacts.devtoolsLogs[UsesRelPreconnectAudit.DEFAULT_PASS];
+    let maxWasted = 0;
+
+    const [networkRecords, mainResource] = await Promise.all([
+      artifacts.requestNetworkRecords(devtoolsLogs),
+      artifacts.requestMainResource(devtoolsLogs),
+    ]);
+
+    const preconnectResults = {};
+    networkRecords
```
wardpeet: should I put all filters in 1?

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+      }).filter(record => {
+        return !!URL.getOrigin(record.url);
+      // filter out all resources where origins are already resolved
+      }).filter(record => {
+        return !UsesRelPreconnectAudit.hasAlreadyConnectedToOrigin(record);
+      }).forEach(record => {
+        const requestDelay = record._startTime - mainResource._endTime;
+        const recordOrigin = URL.getOrigin(record.url);
+
+        if (preconnectResults[recordOrigin]) {
```
wardpeet: better way to find unique origins?

paulirish: IMO use a Set for `preconnectResults` instead

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+    ]);
+
+    const preconnectResults = {};
+    networkRecords
+      // filter out all resources that have a different origin
+      .filter(record => !URL.originsMatch(mainResource.url, record.url))
+      // filter out all resources that are not loaded by the document
+      .filter(record => {
+        return record.initiatorRequest() !== mainResource && record !== mainResource;
+      // filter out urls that do not have an origin
```
paulirish: seems like L62 would already have taken care of these.

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+      }).filter(record => {
+        return !!URL.getOrigin(record.url);
+      // filter out all resources where origins are already resolved
```
paulirish: since we're only displaying origins and not full request URLs, I feel like this is already taken care of by reduction down to unique origins below.

(The assumption being that connect is always a cost for new origins... which seems a reasonable assumption. right @patrickhulce ?)

lighthouse-core/audits/uses-rel-preconnect.js
paulirish: yah a single filter method would look more straightforward IMO

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+
+'use strict';
+
+const Audit = require('./audit');
+const Util = require('../report/v2/renderer/util');
+const UnusedBytes = require('./byte-efficiency/byte-efficiency-audit');
+const URL = require('../lib/url-shim');
+const PRECONNECT_SOCKET_MAX_IDLE = 10;
```
paulirish: Can you add some comments?

> Preconnect establishes a "clean" socket. Chrome's socket manager will keep an unused socket 
> around for 10s. Meaning, the time delta between processing preconnect a request should be <10s, otherwise it's wasted

paulirish: also I think we could actually use 12-15 for our threshold, tbh. we'd include slightly more connections and there's a decent chance in real loads, the preconnect would be used.

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+      }).forEach(record => {
+        const requestDelay = record._startTime - mainResource._endTime;
+        const recordOrigin = URL.getOrigin(record.url);
+
+        if (preconnectResults[recordOrigin]) {
+          return;
+        }
+
+        // make sure the requests are below the PRECONNECT_SOCKET_MAX_IDLE (10s) mark
+        if (Math.max(0, requestDelay) < PRECONNECT_SOCKET_MAX_IDLE) {
```
paulirish: style wise it'd probably be better to check the inverse and return, like the other filters. 

that way at the end we take everything that made it through the gauntlet and add to the set.

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+      maxWasted = Math.max(wastedMs, maxWasted);
+
+      return {
+        url: new URL(record.url).origin,
+        wastedMs: Util.formatMilliseconds(wastedMs),
+      };
+    });
+
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
```
paulirish: `text: 'Origin'`

paulirish: based on your screenshot i'd say our url renderer is a little awkward in handling this origin-only URLs. 

let's fix that. :)
we can check for pathname of `/` and if that's the case, not separate the two.... probably want most of that in `Util.parseURL` and small adjustments in the `_renderTextURL`

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+    return {
+      category: 'Performance',
+      name: 'uses-rel-preconnect',
+      description: 'Avoid multiple, costly round trips to any origin',
+      informative: true,
+      helpText: 'Consider using<link rel="preconnect dns-prefetch"> to set up early connections ' +
+        ' before an HTTP request is actually sent to the server. This will reduce multiple, ' +
+        `costly round trips to any origin. [Learn more](${learnMoreUrl}).`,
+      requiredArtifacts: ['devtoolsLogs'],
+      scoringMode: Audit.SCORING_MODES.NUMERIC,
```
patrickhulce: scoring display mode

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+const learnMoreUrl =
+  'https://developers.google.com/web/fundamentals/performance/resource-prioritization#preconnect';
+
+class UsesRelPreconnectAudit extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      category: 'Performance',
```
patrickhulce: no more category :)

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+      .filter(record => !URL.originsMatch(mainResource.url, record.url))
+      // filter out all resources that are not loaded by the document
+      .filter(record => {
+        return record.initiatorRequest() !== mainResource && record !== mainResource;
+      // filter out urls that do not have an origin
+      }).filter(record => {
+        return !!URL.getOrigin(record.url);
+      // filter out all resources where origins are already resolved
+      }).filter(record => {
+        return !UsesRelPreconnectAudit.hasAlreadyConnectedToOrigin(record);
```
patrickhulce: might want to explicitly call out invalid timing information as well, i.e. if dnsStart/etc are not finite positive numbers

lighthouse-core/audits/uses-rel-preconnect.js
patrickhulce: yeah but he's trying to find the first representative network record for the origin so you do need to filter out the network records that were already connected

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        // make sure the requests are below the PRECONNECT_SOCKET_MAX_IDLE (10s) mark
+        if (Math.max(0, requestDelay) < PRECONNECT_SOCKET_MAX_IDLE) {
+          preconnectResults[recordOrigin] = record;
+        }
+      });
+
+    const results = Object.values(preconnectResults).map(record => {
+      const wastedMs = record.timing.connectEnd - record.timing.dnsStart;
```
patrickhulce: we'll want to `Math.min` this with the delta between `connectStart` and main document resource end

lighthouse-core/audits/uses-rel-preconnect.js
patrickhulce: yeah I might flip this around a bit ward

1. create a set of all origins
2. filter out the bad origins
3. for each origin, find the earliest record that meets the other criteria

lighthouse-core/audits/uses-rel-preconnect.js
wardpeet: call out? Do you mean log to sentry?

wardpeet:
@paulirish @patrickhulce 
I think I have all your review comments done, not sure if I implemented them like you guys wanted 🙄.

i'm not really sure about the displayUrl helpers though.. 

Also smoketests are failing on travis as it's such a low number. Could this be a docker issue? that host is has these origins cached, should I just check for items than?

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+
+'use strict';
+
+const Audit = require('./audit');
+const Util = require('../report/v2/renderer/util');
+const UnusedBytes = require('./byte-efficiency/byte-efficiency-audit');
+const URL = require('../lib/url-shim');
+// Preconnect establishes a "clean" socket. Chrome's socket manager will keep an unused socket
```
patrickhulce: throw a link to reference?

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      name: 'uses-rel-preconnect',
+      description: 'Avoid multiple, costly round trips to any origin',
+      informative: true,
+      helpText:
+        'Consider using<link rel="preconnect dns-prefetch"> to set up early connections ' +
```
patrickhulce: @paulirish may have opinions here too, but WDYT about?

> Consider adding preconnect or dns-prefetch resource hints to establish early connections to important third-party origins. [Learn more](https://developers.google.com/web/fundamentals/performance/resource-prioritization#preconnect).

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+    let maxWasted = 0;
+
+    const [networkRecords, mainResource] = await Promise.all([
+      artifacts.requestNetworkRecords(devtoolsLogs),
+      artifacts.requestMainResource(devtoolsLogs),
+    ]);
+
+    const origins = networkRecords
+      .filter(record => {
+        // filter out all resources that have the same origin
```
patrickhulce: nit: move this comment down a line like the others :)

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+          UsesRelPreconnectAudit.socketStartTimeIsBelowThreshold(record, mainResource)
+        );
+      })
+      .map(record => URL.getOrigin(record.url));
+
+    const preconnectOrigins = new Set(origins);
+    const results = [];
+    preconnectOrigins.forEach(origin => {
+      const records = networkRecords.filter(record => URL.getOrigin(record.url) === origin);
+      if (!records.length) {
```
patrickhulce: this seems impossible, you didn't run across this IRL did you?

seems like the same checks from above also belong in this spot though?

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+        return;
+      }
+
+      // Sometimes requests are done simultaneous and the connection has not been made
+      // chrome will try to connect for each network record, we get the first record
+      let firstRecordOfOrigin;
```
patrickhulce: @paulirish is this an acceptable occasion for reduce in your 👀 ? :)

this works too

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,143 @@
...
+        }
+      });
+
+      const connectionTime =
+        firstRecordOfOrigin.timing.connectEnd - firstRecordOfOrigin.timing.dnsStart;
+      const timeBetweenMainResourceAndConnectStart =
+        firstRecordOfOrigin._startTime * 1000 -
+        mainResource._endTime * 1000 +
+        firstRecordOfOrigin.timing.connectStart;
+      // calculate delta between connectionTime and timeToConnectionStart from main resource
```
patrickhulce: this comment is a bit confusing, didn't we already compute the delta and the line below is ensuring we just cap the `connectionTime` savings?

lighthouse-core/report/v2/renderer/crc-details-renderer.js
```diff
@@ -110,7 +110,7 @@ class CriticalRequestChainRenderer {
     const {file, hostname} = Util.parseURL(segment.node.request.url);
     const treevalEl = dom.find('.crc-node__tree-value', chainsEl);
     dom.find('.crc-node__tree-file', treevalEl).textContent = `${file}`;
-    dom.find('.crc-node__tree-hostname', treevalEl).textContent = `(${hostname})`;
+    dom.find('.crc-node__tree-hostname', treevalEl).textContent = hostname ? `(${hostname})` : '';
```
patrickhulce: nice find :) if it doesn't have a hostname though should we be showing it? what type of request is that, data URI?

lighthouse-core/test/audits/users-rel-preconnect-test.js
```diff
@@ -0,0 +1,206 @@
...
+
+const UsesRelPreconnect = require('../../audits/uses-rel-preconnect.js');
+const assert = require('assert');
+const Util = require('../../report/v2/renderer/util');
+
+const mainResource = {
+  url: 'https://www.example.com/',
+  _endTime: 1,
+};
+const mainResourceRecord = {
```
patrickhulce: why do these two need to be different? :)

lighthouse-core/test/audits/users-rel-preconnect-test.js
```diff
@@ -0,0 +1,206 @@
...
+      requestNetworkRecords: () => Promise.resolve(networkRecords),
+      requestMainResource: () => Promise.resolve(mainResource),
+    };
+
+    const {rawValue, extendedInfo} = await UsesRelPreconnect.audit(artifacts);
+    assert.equal(rawValue, 400);
+    assert.equal(extendedInfo.value.length, 2);
+    assert.deepStrictEqual(extendedInfo.value, [
+      {url: 'https://cdn.example.com', wastedMs: Util.formatMilliseconds(200)},
+      {url: 'https://othercdn.example.com', wastedMs: Util.formatMilliseconds(400)},
```
patrickhulce: would you mind explaining this result? seems like it should be min of `500 (600 - 100)` and `300 (1000 * (1.2 - 1) + 100)`

wardpeet: when looking in code I use connectStart where I should have used dnsStart 🤦‍♂️ 
```
const connectionTime = 600 - 100;
const timeBetweenMainResourceAndConnectStart =
  1.2 * 1000 -
   1 * 1000 +
  200;
// calculate delta between connectionTime and timeToConnectionStart from main resource
const wastedMs = Math.min(500, 400);
```

lighthouse-core/test/audits/users-rel-preconnect-test.js
wardpeet: good question 😛

lighthouse-core/test/audits/users-rel-preconnect-test.js
patrickhulce: ah yeah I missed that too :) seems like `timeBetweenMainResourceAndDnsStart` should fix it 👍 (or fallback to connect start I guess if dns start is invalid, is that a real case?)

lighthouse-core/audits/uses-rel-preconnect.js
wardpeet: fine by me i most of the time juts copy what you guys say 😄 but I think <link tags shouldn't be in the description like you describe

lighthouse-core/report/v2/renderer/crc-details-renderer.js
wardpeet: no it comes back from the parseUrl where a request might be the root so I don't return a hostname.

it actually looks weird so I have to do the origin display inside the report and not in the parseUrl 
image

wardpeet: Updated to
critical request chain:
image


preconnect table:
image

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,139 @@
...
+    const [networkRecords, mainResource] = await Promise.all([
+      artifacts.requestNetworkRecords(devtoolsLogs),
+      artifacts.requestMainResource(devtoolsLogs),
+    ]);
+
+    const origins = networkRecords
+      .filter(record => {
+        return (
+          // filter out all resources that have the same origin
+          !URL.originsMatch(mainResource.url, record.url) &&
```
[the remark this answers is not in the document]
wardpeet: should we use webinspector parsedurl instead?
`mainResource.parsedURL.securityOrigin() === record.parsedURL.securityOrigin()`

lighthouse-cli/test/fixtures/perf/preload_tester.js
```diff
@@ -7,3 +7,6 @@
 /* eslint-disable */
 
 document.write('<script src="level-2.js?delay=500"></script>')
+
+// load another origin
+fetch('http://localhost:10503/preload.html', () => {});
```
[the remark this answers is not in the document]
wardpeet: not sure why I even added it as it returns a promise

lighthouse-core/audits/uses-rel-preconnect.js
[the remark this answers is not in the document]
patrickhulce: yeah even better! 👍

wardpeet:
@patrickhulce @paulirish @brendankenny 
fixed all comments

lighthouse-core/report/v2/renderer/util.js
```diff
@@ -184,9 +193,10 @@ class Util {
     if (name.length > MAX_LENGTH) {
       const dotIndex = name.lastIndexOf('.');
       if (dotIndex >= 0) {
-        name = name.slice(0, MAX_LENGTH - 1 - (name.length - dotIndex)) +
-            // Show file extension
-            `${ELLIPSIS}${name.slice(dotIndex)}`;
+        name =
```
patrickhulce: think we can revert some of this diff noise?

lighthouse-core/test/results/sample_v2.json
```diff
@@ -5347,6 +5373,6 @@
     }
   },
   "timing": {
-    "total": 830
+    "total": 2654
```
patrickhulce: yikes! this wasn't from added preconnect time was it? ;)

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,150 @@
...
+          // make sure the requests are below the PRECONNECT_SOCKET_MAX_IDLE (15s) mark
+          UsesRelPreconnectAudit.socketStartTimeIsBelowThreshold(record, mainResource)
+        );
+      })
+      .map(record => record.parsedURL.securityOrigin());
+
+    const preconnectOrigins = new Set(origins);
+    let results = [];
+    preconnectOrigins.forEach(origin => {
+      const records = networkRecords.filter(record => record.parsedURL.securityOrigin() === origin);
```
paulirish: we need all records that survived the above filtering. 

So one way we could do it is to use a `Map` for this `origins` variable above. The key is the securityOrigin, and the value is an an array of networkRecords (that survived the filter).

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -0,0 +1,150 @@
...
+
+      // Sometimes requests are done simultaneous and the connection has not been made
+      // chrome will try to connect for each network record, we get the first record
+      let firstRecordOfOrigin;
+      records.forEach(record => {
```
paulirish: yeah style-wise I like patrick's idea of using a reduce to find the `firstRecordOfOrigin`. kinda fancy.

typings/web-inspector.d.ts
```diff
@@ -36,6 +36,7 @@ declare global {
       localizedFailDescription?: string;
 
       _initiator: NetworkRequestInitiator;
+      initiatorRequest(): NetworkRequest,
```
brendankenny: in `NetwokRequest` it looks like this can also be `null`

typings/web-inspector.d.ts
```diff
@@ -48,13 +49,16 @@ declare global {
...
       scheme: string;
       host: string;
+      securityOrigin(): string;
     }
 
     export interface NetworkRequestInitiator {
       type: 'script' | 'parser';
     }
 
     export interface NetworkRequestTiming {
```
brendankenny: we should drop `NetworkRequestTiming` altogether since (@paulirish just pointed out) this is actually just `LH.Crdp.Network.ResourceTiming` and is already typed for us :)

You can switch or I can take it later

paulirish: @brendankenny how about you take this later, yah

paulirish:
lgtm. ready to merge when green.

addyosmani:
\o/ Thanks for the diligent reviews getting this audit firmed up @paulirish @patrickhulce and @brendankenny. Big thanks for all your work getting this audit landed, @wardpeet *high-fives*

PR #4369, opened 2018-01-26
cli(output-path): Add ability to use directory for output-path

codingmatty:
This allows the `--output-path` flag to be a directory where result files are placed. 

This will gives users the option to run lighthouse at a root directory and to keep resulting files cleanly in a directory that can be easily ignored, instead of needing to change directories or clutter up the working directory.

Ex. `lighthouse <url> --output-path ./lighthouse-results` will place the result output file within `./lighthouse-results` _if_ it is a directory.


paulirish:
This is good stuff. I've been thinking our concept of output-path is a little odd.

mouafa:
@paulirish is there any reason to not support this?

brendankenny:
moving discussion to #7408

PR #4370, opened 2018-01-26
misc(travis): Run node 6.9.1 as only process for PRs

wardpeet:
Travis has new features. Much easier to only run specific tasks for a specific branch.

I also disabled the yarn install from the travis script

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: required
 language: node_js
 matrix:
+  fast_finish: true
   include:
     - node_js: "6.9.1"
       if: type IN (push, pull_request)
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,11 @@ language: node_js
 matrix:
   include:
     - node_js: "6.9.1"
+      if: type IN (push, pull_request)
     - node_js: "8"
+      if: branch = master
     - node_js: "9"
+      if: branch = master
 dist: trusty
 cache:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,10 +1,14 @@
 sudo: required
 language: node_js
 matrix:
+  fast_finish: true
```
paulirish: i dont think this works the way you think it does.

i only know this because everytime i see this property i get excited, but then read the docs and get disappointed: https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing

it needs `allow_failures` marked on some jobs

wardpeet: actually I do :p but I saw so many projects using fast_finish without allow_failures that I thought travis might haven't documented everything :p but i'll revert the commit.

wardpeet:
It will run whenever something is merged to master. so Master will still run all node

wardpeet:
I think I fixed it

PR build:
https://travis-ci.org/GoogleChrome/lighthouse/builds/335377013

master build:
https://travis-ci.org/wardpeet/lighthouse/builds/335377485

paulirish:
Cool. the only node 6x will reduce flakiness and get us results faster. \o/


------

Let's make one more change. Let's disable travis (and appveyor) from building `push`/`branch` builds (unless it's master branch).

Those are pretty irrelevant IMO.

PR #4372, opened 2018-01-27
PWA related metrics with jsonpretty cli output

eeshdarthvader:
Changes in json ouput to cater to PWA react app metrics with json pretty CLI ouput
screen shot 2018-01-27 at 5 40 01 pm

### DIFF
--- a/lighthouse-cli/printer.js
+++ b/lighthouse-cli/printer.js
@@ -8,6 +8,8 @@
 const fs = require('fs');
 const ReportGenerator = require('../lighthouse-core/report/v2/report-generator');
 const log = require('lighthouse-logger');
+const prettyjson = require('prettyjson');
+
 
 /**
  * An enumeration of acceptable output modes:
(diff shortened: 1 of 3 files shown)


wardpeet:
@eeshdarthvader thanks for creating this PR.

I don't really know what the rest of the team thinks of this but honestly I don't think we need this. JSON.stringify does an ok job + you can easily pipe it through a file. The Json output isn't really made for readability.

patrickhulce:
Thanks for the PR @eeshdarthvader! This also doesn't seem to render valid JSON output, which is being relied upon by several systems when the output type is `json` :)

I don't think we can approve this particular change in core, but if you're interested in helping out we'd love a hand on figuring out what's up with some of the [open bugs that need investigation](https://github.com/GoogleChrome/lighthouse/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+label%3A%22needs-investigation%22+) ❤️

PR #4380, opened 2018-01-29
core(a11y): aXe perf: only collect provided resultTypes

paulirish:
[A new axe document on performance](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#section-4-performance) suggests safelisting the resultTypes you're interested in. This saves time generating unique selectors (which on mlb.com can take 4s). 

cc @dylanb

This reduces about a 1 second of runtime cost of axe on mlb.com. (13s to 12s with the https://github.com/dequelabs/axe-core/pull/699 PR)

### DIFF
--- a/lighthouse-core/gather/gatherers/accessibility.js
+++ b/lighthouse-core/gather/gatherers/accessibility.js
@@ -25,6 +25,7 @@ function runA11yChecks() {
         'wcag2aa',
       ],
     },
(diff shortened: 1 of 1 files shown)


dylanb:
It might require you to change your score calculator if you take the exact number of passes into account. We are considering adding the count into the results structure in the future.

PR #4394, opened 2018-01-30
core(config): add support for audit/gatherer options

### LINKED ISSUE #4206: Add support for audit/gatherer options
To support LRider we need to allow customization of score control points, Lantern device types, etc. For this, we need to allow audit and gatherer options as proposed in config v2 😄

patrickhulce:
big change coming in! closes #4206 

introduces support for audit and gatherer options while maintaining backcompat with existing config options, internally all audits and gatherers are mapped to the new format

additionally, the very poorly named `options` input to gatherers I'm now calling `context` as it more roughly captures what's being passed in, also we avoid `options.options` :)

feedback very very welcome while it's still early please! :D 

## Usage - before
```js
module.exports = {
  passes: [{
    gatherers: [
      'builtin-gatherer',
      class MyGatherer extends Gatherer { },
  }],
  audits: [
    'builtin-audit',
    class MyAudit extends Audit { },
  ]
}
```

## Usage - after
```js
module.exports = {
  passes: [{
    gatherers: [
      'builtin-gatherer', // legacy
      {path: 'seo/other-gatherer', options: {foo: 1}}, // new

      class MyGatherer extends Gatherer { }, // legacy
      {implementation: class MyGatherer {}}, // new
      {instance: new (class MyGatherer extends Gatherer { })()}, // new
  }],
  audits: [
    'builtin-audit',
    {path: 'byte-efficiency/other-audit', options: {foo: 1}},
    class MyAudit extends Audit { },
  ]
}
```


lighthouse-core/config/config.js
```diff
@@ -145,7 +146,7 @@ function validateCategories(categories, audits, groups) {
     return;
   }
 
-  const auditIds = audits.map(audit => audit.meta.name);
+  const auditIds = audits.map(audit => audit.implementation.meta.name);
```
wardpeet: I find `implementation` such a weird name I wanted to call it instance but I see below that it's indeed the implementation but still find it weird 😛 but maybe that's because i'm no native speaker.

wardpeet:
I really like this PR! not much to say about it. I haven't tested any of it but the approach looks good. I only think we need to document the option parameter a bit. We don't want it to be used as an escape hatch to smuggle custom settings in.

lighthouse-core/gather/gather-runner.js
```diff
@@ -491,14 +501,21 @@ class GatherRunner {
...
-      pass.gatherers = pass.gatherers.map(gatherer => {
+      pass.gatherers = pass.gatherers.map(gathererDefn => {
         // If this is already instantiated, don't do anything else.
-        if (typeof gatherer !== 'string') {
-          return gatherer;
+        if (gathererDefn.instance) {
+          return gathererDefn;
+        }
+
+        if (typeof gathererDefn.beforePass === 'function') {
```
paulirish: this is for backwards compat? if so can you add a comment with the case this covers?

lighthouse-core/gather/gather-runner.js
```diff
@@ -235,14 +237,16 @@ class GatherRunner {
...
       // Navigate.
-      .then(_ => GatherRunner.loadPage(driver, options))
+      .then(_ => GatherRunner.loadPage(driver, context))
       .then(_ => log.log('statusEnd', status));
 
-    return gatherers.reduce((chain, gatherer) => {
+    return gatherers.reduce((chain, gathererDefn) => {
       return chain.then(_ => {
-        const artifactPromise = Promise.resolve().then(_ => gatherer.pass(options));
+        const gatherer = gathererDefn.instance;
```
paulirish: feels like pass/beforePass shouldn't have to do this resolving on their own. can we provide instantiated gatherers instead of the `config.gatherers` or something?

lighthouse-core/gather/gather-runner.js
```diff
@@ -491,14 +501,21 @@ class GatherRunner {
...
+
+        if (typeof gathererDefn.beforePass === 'function') {
+          return {instance: gathererDefn, options: {}};
         }
 
-        const GathererClass = GatherRunner.getGathererClass(gatherer, rootPath);
-        return new GathererClass();
+        const gathererPath = typeof gathererDefn === 'string' ? gathererDefn : gathererDefn.path;
+        const GathererClass = GatherRunner.getGathererClass(
+          gathererDefn.implementation || gathererPath, rootPath);
```
paulirish: should we move all this into config.js? seems like it spiritually belongs over there.

There's also a lot of cases (6+?) being served by these lines: https://github.com/GoogleChrome/lighthouse/blob/c2652e6851a0b9b1a33fab978bfc4cd99bb4db75/lighthouse-core/gather/gather-runner.js#L505-L516

I'm hoping by moving it into config, we're going to end up with something simpler. Open to ideas.

lighthouse-core/gather/gather-runner.js
```diff
@@ -235,14 +237,16 @@ class GatherRunner {
...
-      .then(_ => GatherRunner.loadPage(driver, options))
+      .then(_ => GatherRunner.loadPage(driver, context))
       .then(_ => log.log('statusEnd', status));
 
-    return gatherers.reduce((chain, gatherer) => {
+    return gatherers.reduce((chain, gathererDefn) => {
       return chain.then(_ => {
-        const artifactPromise = Promise.resolve().then(_ => gatherer.pass(options));
+        const gatherer = gathererDefn.instance;
+        context.options = gathererDefn.options || {};
```
paulirish: at the very least let's comment that this is terrible. :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -491,14 +501,21 @@ class GatherRunner {
 
   static instantiateGatherers(passes, rootPath) {
```
paulirish: let's do instantiateGatherers back in config so by the time `gatherrunner.run` is called, everything has an instance. 
this'll require updating some tests. sorry 'bout that.

lighthouse-core/gather/gather-runner.js
patrickhulce: resolved in person, outcome: gather-runner will still need access to `.instance` and `.options` but will be oblivious to everything else

lighthouse-core/gather/gather-runner.js
patrickhulce: done

lighthouse-core/gather/gather-runner.js
patrickhulce: yeah just lots of test changes 😢 I'm on it though

lighthouse-core/gather/gather-runner.js
patrickhulce: I'll create a new context so it won't be terrible :)

lighthouse-core/gather/gather-runner.js
patrickhulce: on it

lighthouse-core/gather/gather-runner.js
```diff
@@ -203,31 +203,34 @@ class GatherRunner {
...
+        .then(() => passContext.driver.setExtraHTTPHeaders(passContext.flags.extraHeaders));
 
-    return options.config.gatherers.reduce((chain, gatherer) => {
+    return passContext.config.gatherers.reduce((chain, gathererDefn) => {
       return chain.then(_ => {
-        const artifactPromise = Promise.resolve().then(_ => gatherer.beforePass(options));
+        const gatherer = gathererDefn.instance;
+        // Abuse the passContext to pass through gatherer options
+        passContext.options = gathererDefn.options || {};
+        const artifactPromise = Promise.resolve().then(_ => gatherer.beforePass(passContext));
```
paulirish: wdyt about a followup PR to rename `options` in all the gatherers to `passContext`?

patrickhulce: sg

patrickhulce:
> this is gtg, though I think we should land it after we cut the 2.9.1. sgty? 

yeah good plan

Issue #4395, opened 2018-01-31
Move any clientside JS functions to their own files


paulirish:
I noticed that during the `yarn coverage` run, there'd always be a timeout. ([example](https://travis-ci.org/GoogleChrome/lighthouse/jobs/334950696#L886)). Turns out its the cli's `run-test.js` that does full roundtrip of lighthouse. Digging in, it timed out in `driver.js` waiting for CPU idle... and eventually hits the 45s time limit.

So the current impact of this bug:
* incorrect code coverage stats for CLI
* an extra ~42 seconds of build time

----

## Details

I poked at the behavior and turns out the code running via `driver.evaluateAsync` has istanbul instrumentation:

![image](https://user-images.githubusercontent.com/39191/35600430-b26337f4-05e2-11e8-900b-bcb372b30920.png)

Also amusingly https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md says adding `istanbul ignore next` doesn't exclude it from instrumentation. So it's fair to assume istanbul instrumentation will break most of our `evaluateAsync` calls.

### Bonus details
1. We had a typo: `instanbul` -> `istanbul`
1. istanbul's `ignore next` regex doesn't match the statement if it's at the end of a jsdoc (or other flowerbox) comment. [it has to be at the start of a comment.](https://github.com/gotwarlost/istanbul/blob/bc84c315271a5dd4d39bcefc5925cfb61a3d174a/lib/instrumenter.js#L16)
1. Fixing both of these didn't fix the problem. But you can imagine I expected them to. :disappointed:

## Solutions

* A) Move any clientside JS functions to their own files. Maybe `xxxx-clientside.js` and then we can add a `--exclude *-clientside.js` to the `istanbul cover` command.
* B) Switch to using v8/devtools code coverage. Right now, the leading impl of this, [c8](https://github.com/bcoe/c8), only creates istanbul lcov json files. The project needs to mature a bit before it'll be a real option for istanbul/nyc. But I'm sure it will. 
    * Regardless, this will only work in node 9, which we're about to stop testing on travis. 😁 
* C) A hackier quick fix which I'm about to PR... 😛

paulirish:
As discussed in #4396, the next action is to implement solution A.

> it might be nice to collect all our client-side run functions in a single place so it becomes clear what gets run where

tollmanz:
[...]

PR #4396, opened 2018-01-31
tests(coverage): minimize impact of timeout due to istanbul's instrumentation

paulirish:
Issue #4395 has the backstory. This PR has a few things going on:

* **primary fix:** wait only 9s load during run-test's roundtrip. No need to wait 45s if there's a problem loading the page.
* **driveby:** remove explicit timeout from istanbul's mocha call. (finishing #4224)
* **moot:** fix `istanbul ignore` statements. as detailed in #4395 they must be standalone (and spelled right). 

As a result of this PR, `yarn coverage` should be 36s faster, and no timeout error in the travis log. However the code coverage stats for the CLI are still incorrect and the roundtrip run is still not truly  successful.

This leads to a followup question: Should we have other success criteria for the run-test.js roundtrip test, like calculating TTI, etc?

Ref #4395 but doesn't close it.

### DIFF
--- a/lighthouse-cli/test/cli/run-test.js
+++ b/lighthouse-cli/test/cli/run-test.js
@@ -25,7 +25,8 @@ describe('CLI run', function() {
   it('runLighthouse completes a LH round trip', () => {
     const url = 'chrome://version';
     const filename = path.join(process.cwd(), 'run.ts.results.json');
-    const flags = getFlags(`--output=json --output-path=${filename} ${url}`);
+    const timeoutFlag = `--max-wait-for-load=${9000}`;
+    const flags = getFlags(`--output=json --output-path=${filename} ${timeoutFlag} ${url}`);
     return run.runLighthouse(url, flags, fastConfig).then(passedResults => {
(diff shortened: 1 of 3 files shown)


lighthouse-cli/test/cli/run-test.js
```diff
@@ -25,7 +25,8 @@ describe('CLI run', function() {
   it('runLighthouse completes a LH round trip', () => {
     const url = 'chrome://version';
     const filename = path.join(process.cwd(), 'run.ts.results.json');
-    const flags = getFlags(`--output=json --output-path=${filename} ${url}`);
+    const timeoutFlag = `--max-wait-for-load=${9000}`;
```
patrickhulce: I'm curious, why put 9000 in template? :)

paulirish: for the funsies.  🕺

paulirish:
> really annoying that the only way to eliminate instrumentation is separate files, then again it might be nice to collect all our client-side run functions in a single place so it becomes clear what gets run where :)

Yup totally agree.

PR #4408, opened 2018-02-01
docs(puppeteer.md): examples of combining the two

ebidel:
R: @paulirish

### DIFF
--- a/docs/puppeteer.md
+++ b/docs/puppeteer.md
@@ -1,44 +1,45 @@
 # Recipes Puppeteer with Lighthouse
 
 **Note**: https://github.com/GoogleChrome/lighthouse/issues/3837 tracks the discussion for making Lighthouse work in concert with Puppeteer.
-Some things are possible today (login to a page using Puppeteer, audit it using Lighthouse) while others (A/B testing the perf of UI changes) are not.
+Some things are possible today (login to a page using Puppeteer, audit it using Lighthouse) while others (A/B testing the perf of UI changes) are trickier or not yet possible.
 
-### Custom network throttle settings using Puppeteer
+### Inject JS/CSS before the page loads
 
-By default, Lighthouse uses a mobile 3G connecton for it's network throttlings. It's possible to use Puppeter
-to launch Chrome and customize these settings. 
-
-Flow: 
-1. disable the throttling settings in Lighthouse.
-2. Launch Chrome using Puppeteer and tell Lighthouse to reuse Puppeteer's chrome instance.
-3. Tell lighthouse to programmatically load the page.
-4. Watch for the page to open and set the emulation conditions:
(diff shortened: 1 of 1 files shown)


docs/puppeteer.md
```diff
@@ -0,0 +1,101 @@
+# Recipes Puppeteer with Lighthouse
+
+**Note**: https://github.com/GoogleChrome/lighthouse/issues/3837 tracks the discussion for making Lighthouse work in concert with Puppeteer.
+Some things are possible today (login to a page using Puppeteer, audit it using Lighthouse) while others (A/B testing the perf of UI changes) are not.
+
+### Custom network throttle settings using Puppeteer
```
patrickhulce: Could we advertise a different example of using puppeteer? Perhaps just logging in to your page with the second example you've already got? 

Our preferred method of customizing throttling is already [documented](https://github.com/GoogleChrome/lighthouse/blob/e8a1a7fb82c979f0298d8a7f752c61de467f7bcb/docs/throttling.md) and jumping in to muck with those settings in puppeteer is bound to result in over confidence in the accuracy of the performance metrics for those conditions.

docs/puppeteer.md
```diff
@@ -0,0 +1,101 @@
...
+})();
+```
+
+### Connecting Puppeteer to a browser instance launched by chrome-launcher.
+
+When using Lighthouse programmatically, you'll often use chrome-launcher to launch Chrome.
+Puppeteer can reconnect to this existing browser instance like so:
+
+```js
+onst chromeLauncher = require('chrome-launcher');
```
pedro93: Const typo

docs/puppeteer.md
ebidel: Yea that's fair. I can cook up a slightly different example of doing something in pptr then use lighthouse.

ebidel:
Changed to an example that injects css. As noted, it uses the protocol api instead of the pptr API to do so, but we can update the snippet when that's fixed.

wardpeet:
does it make sense to indent the async functions? not that it really matters but I just like that more 🙄

ebidel:
I usually don't b/c it creates unnecessary indentation. But happy to if that's what people want.

docs/puppeteer.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Custom network throttle settings using Puppeteer
+
+By default, Lighthouse uses a mobile 3G connecton for it's network throttlings. It's possible to use Puppeter
+to launch Chrome and customize these settings. 
+
+Flow: 
+1. disable the throttling settings in Lighthouse.
+2. Launch Chrome using Puppeteer and tell Lighthouse to reuse Puppeteer's chrome instance.
+3. Tell lighthouse to programmatically load the page.
```
coliff: 'Lighthouse' rather than 'lighthouse' I think...

Issue #3618, opened 2017-10-19
Lantern: Improve performance opportunity estimates based on graphs


patrickhulce:
The potential savings estimates currently just compute the milliseconds based solely on bytes saved. This leads to some confusion and misrepresentation since all ms are not created equal, i.e. shaving 1s off FMP means a lot more than saving 1s off the end of "all images have finished downloading"

We can use the fast mode graphs to predict how much time would be saved for each metric as well as overall savings

patrickhulce:
blocked on UX thinking here

patrickhulce:
I am *so* *stoked* about how this is shaping up early. It's reranking things much more in line with how we intuitively think of biggest bang-for-buck performance. Improvements to files that are loaded early and dependencies of lots of resources now have the biggest impact whereas improvements to lazily-loaded images that are already as late as they should be have little to no impact.

![image](https://user-images.githubusercontent.com/2301202/36328599-8b749a6a-1317-11e8-9af6-bcfa91c88aa2.png)

PR #4409, opened 2018-02-01
core(computed-artifact): use deep equality over strict

patrickhulce:
needed by #3618 and the introduction of audit options (multiple different audits will request the same lantern simulation settings for example)


lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -5,13 +5,15 @@
...
+const CustomMap = require('../../lib/custom-map');
+
 class ComputedArtifact {
   /**
    * @param {!ComputedArtifacts} allComputedArtifacts
    */
   constructor(allComputedArtifacts) {
     /** @private {!Map} */
-    this._cache = new Map();
+    this._cache = new CustomMap(CustomMap.deepEquals);
```
paulirish: instead of passing it into the constructor i'm thinking it might be helpful to clarify the intent of the arg.   

how about:

```js
this._cache = new CustomMap();
this._cache.setEqualityFn(CustomMap.deepEquals);
```

without it being set, it's just using `a === b`

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -5,13 +5,15 @@
  */
 'use strict';
 
+const CustomMap = require('../../lib/custom-map');
```
paulirish: this impl doesn't support `new CustomMap([[x, 1], [y, 2]])`. 

given that maybe call it `CacheMap`?    really anything to give it some distance appearing as a true subclass of `Map`

patrickhulce: ah yeah, good call I'll add support :)

lighthouse-core/gather/computed/computed-artifact.js
patrickhulce: good plan 👍

lighthouse-core/gather/computed/computed-artifact.js
paulirish: hmmm. I don't think we have a real usecase for supporting that constructor API. can we keep it out of scope?

wardpeet:
It's a really stupid question probably but why does this help us for project Lantern? Which use case do we want to tackle here? (not sure why we need deep equal yet)

lighthouse-core/gather/computed/computed-artifact.js
patrickhulce: alright alright, cache map it is

lighthouse-core/lib/custom-map.js
```diff
@@ -0,0 +1,61 @@
...
+  _findIndexOf(key) {
+    for (let i = 0; i < this._entries.length; i++) {
+      if (this._equalsFn(key, this._entries[i].key)) return i;
+    }
+
+    return -1;
+  }
+
+  static deepEquals(objA, objB, maxDepth = 5) {
+    if (objA === objB) return true;
```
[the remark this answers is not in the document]
patrickhulce: discussed offline, basically exceptions are a dog, the "is not hella slow" test takes 55s a.k.a it's hella slow so we'll just use jdalton's isEqual from lodash instead :)

lighthouse-core/test/lib/custom-map-test.js
```diff
@@ -0,0 +1,87 @@
...
+    map.set('bar', 4);
+
+    assert.equal(map.has(1), false);
+    assert.equal(map.has(false), true);
+    assert.equal(map.has(''), true);
+    assert.equal(map.has({x: 1}), true);
+    assert.equal(map.get('foo'), 4);
+  });
+
+  it('is not hella slow', () => {
```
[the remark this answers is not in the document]
patrickhulce: sure

patrickhulce:
> why does this help us for project Lantern? Which use case do we want to tackle here?

not stupid question, it's hard/impossible to tell from just this PR :) 

basically simulator.js accepts options for what type of connection/cpu/etc to simulate but there's no mechanism for a LH user to control those options, audit/gatherer options gives us this

however, because computed artifacts cache their contents based on strict equality the options objects won't match even if they all request lantern on 3G and so the entire graph would be re-computed for every performance opportunity, introducing caching based on deep equality allows us to maintain the status quo computing the graph just once per setting. make sense?

lighthouse-core/lib/cache-map.js
```diff
@@ -0,0 +1,62 @@
...
+'use strict';
+
+const isEqual = require('lodash.isequal');
+
+/**
+ * @fileoverview This class is designed to allow maps with arbitrary equality functions.
+ * It is not meant to be performant and is well-suited to use cases where the number of entries is
+ * likely to be small (like computed artifacts).
+ */
+module.exports = class CacheMap {
```
patrickhulce: for the record, I'm still not a huge fan of this name, something like `ArbitraryEqualityMap` would make more sense 😛

wardpeet:
> however, because computed artifacts cache their contents based on strict equality the options objects won't match even if they all request lantern on 3G and so the entire graph would be re-computed for every performance opportunity, introducing caching based on deep equality allows us to maintain the status quo computing the graph just once per setting. make sense?

Yes but where does the cpu options come in? Do we store it in the computed artifacts?

patrickhulce:
> Yes but where does the cpu options come in? Do we store it in the computed artifacts?

Not 100% sure what you're asking, but each audit will have default options that can be overridden from config. Audit will pass the blended final options to the computed artifact and if it hasn't been requested with the same options before will compute it otherwise return the corresponding one from cache.

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -5,13 +5,16 @@
...
+
 class ComputedArtifact {
   /**
    * @param {!ComputedArtifacts} allComputedArtifacts
    */
   constructor(allComputedArtifacts) {
     /** @private {!Map} */
-    this._cache = new Map();
+    this._cache = new CacheMap();
+    this._cache.setEqualsFn(CacheMap.deepEquals);
```
paulirish: `setEqualityFn`

lighthouse-core/lib/cache-map.js
paulirish: I'm fine with calling it ArbitraryEqualityMap.

paulirish:
lgtm just waiting for CI

PR #4437, opened 2018-02-06
misc(package): scripts don't require "--" for options to be forwarded

paulirish:
This isn't required in Yarn 1.0+ or NPM.

And now we kill off this warning:
image


wardpeet:
i'm pretty sure npm still needs it but we aren't using it in our scripts so we can just ship this

PR #4439, opened 2018-02-06
deps(snyk): update snyk snapshot

paulirish:
saw sonarwhal did an update, so figured it was time for one over here, too. :)

@tkadlec


wardpeet:
does snyk provide something like greenkeeper.io to keep this up to date or perhaps we should look for a new snyk release every time we want to bump lighthouse?

paulirish:
good question. @tkadlec what do you think makes most sense here?

PR #4441, opened 2018-02-06
cli(lighthouse): Add --mixed-content flag for triggering the mixed content audit

christhompson:
Per discussion on #4344, it may be simpler to add a flag to lighthouse itself for the mixed content scan configuration rather than trying to do it through yarn/npm (as the yarn/npm invocations get unwieldy very fast).

This is an initial attempt at adding a `--mixed-content` flag that loads the mixed-content.js config if passed (essentially simplifying passing a custom config path to an included configuration file).

@paulirish WDYT?

### DIFF
--- a/lighthouse-core/index.js
+++ b/lighthouse-core/index.js
@@ -40,8 +40,8 @@ function lighthouse(url, flags = {}, configJSON) {
     log.setLevel(flags.logLevel);
 
     // Use ConfigParser to generate a valid config file
-    const config = new Config(configJSON, flags.configPath);
-
+    const config = flags.mixedContent ? new Config(require('./config/mixed-content.js'), null)
+      : new Config(configJSON, flags.configPath);
     const connection = new ChromeProtocol(flags.port, flags.hostname);
 
     // kick off a lighthouse run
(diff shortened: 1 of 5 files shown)


lighthouse-core/index.js
```diff
@@ -40,8 +40,8 @@ function lighthouse(url, flags = {}, configJSON) {
     log.setLevel(flags.logLevel);
 
     // Use ConfigParser to generate a valid config file
-    const config = new Config(configJSON, flags.configPath);
-
+    const config = flags.mixedContent ? new Config(require('./config/mixed-content.js'), null)
```
patrickhulce: how about we handle this over in bin.js where we handle `--perf` :)

https://github.com/GoogleChrome/lighthouse/blob/3030b4f0651fc4eb2634600d7cf0dbca43972b32/lighthouse-cli/bin.js#L57-L59

christhompson: Good idea. That's a better place than the awkward ternary I came up with.

paulirish:
I think we also want to add "headless" into the chromeFlags in this case, yeah?

christhompson:
Yep, missed that when converting it over to a flag (and didn't do enough testing). I'll push another commit adding that in.

paulirish:
I added some commits. @patrickhulce lgty?

PR #4444, opened 2018-02-07
core(help-text-fix): Add link to docs for mixed-content audit

christhompson:
This adds a link to the documentation in the helpText for the new mixed-content audit (#3953). The link should be live once the docs PR (google/WebFundamentals#5758) lands.


paulirish:
@christhompson docs-wise where do you want to describe how to use this audit (basically the global install plus the new flag)?

i figure that might as well go in the WF page. Does that seem right to you?

christhompson:
That seems reasonable. I can work with @kaycebasques to add a section for that using the new `--mixed-content` flag.

(The [docs page is live](https://developers.google.com/web/tools/lighthouse/audits/mixed-content), but we can do another PR to update it with the how-to text.)

paulirish:
sweet. sounds great to me.

Issue #4469, opened 2018-02-08
Lighthouse is not Generating Any Report


taimoor1982:
The lighthouse is not generating any report since yesterday. Please help me.

See the snapshot below for better understanding:

![snaphsot](https://user-images.githubusercontent.com/36256897/35967782-be574afc-0ce3-11e8-9d63-bac4facdfbe8.jpg)

tvld:
Same problem here, full freeze ``Lighthouse is debugging this browser```  Same on Ubuntu, Windows etc.

GijsMolsbergen:
Having the same problem here

wardpeet:
closing :) duplicate of #4466

Issue #4467, opened 2018-02-08
Issue encountered while performing the SEO audit from the Lighthouse extension!


AsmaMohd:
Hi,

![check](https://user-images.githubusercontent.com/36260045/35963112-22baa71c-0ccd-11e8-9527-b5282a867de5.png)
I am reaching out to you because I was trying to assess the features of a Lighthouse Chrome extension and I want to conduct an SEO audit of my webpage but all in vain. I am unable to see any reports or prompts on my screen after selecting the SEO audit option on the installed chrome extension. 

Please let me know what could be the possible issue and How can I fix it? 

I have attached the screenshot for you to review.

Waiting for your positive response.

Cheers,
Asma.

AsmaMohd:
URL: http://zoetalentsolutions.com/hr-courses/

tvld:
Seems duplicate of  #4466

wardpeet:
closing :) duplicate of #4466

PR #4470, opened 2018-02-08
extension: work around Data URI resetPage limitation

### LINKED ISSUE #4466: Lighthouse 2.9.0 Chrome extension is broken
Hi,
Lighthouse does not stop showing "Lighthouse is debugging this browser" I get no report either.

Anyone else facing this issue?

kdzwinel:
~Work in Progress 👷~

Attempt to fix #4466 #4467 #4469 

## Issue

2.9 replaced  `about:blank` with data URIs (#4310). Unfortunately, it looks like the data URIs can't be controlled by a Chrome extension. Once we navigate to the blank-page.html data URI we can't navigate away from it using debugger protocol.

screen shot 2018-02-08 at 11 50 49

## Fix

What I tried doing so far:

### blob object URL
```javascript
const blankPageBlob = new Blob([blankPageSource], {type: 'text/html'});
const blankPageUrl = window.URL.createObjectURL(blankPageBlob);
```

I created blob object URL from the blank-page HTML, but this URL doesn't seem to load neither via `driver.goToURL` nor `chrome.tabs.update` 🤔

### embedded html files
I embedded `blank-page.html` and `logo-page.html` into the extension and got their URLs via `chrome.runtime.getURL`. This works, but the Performance category fails with a `NO_NAVSTART` error (other categories pass):

screen shot 2018-02-08 at 11 37 29

[EDIT] It looks like the logo-page.html ends up in the trace, gets picked by `TraceOfTab` as a main resource and all the legitimate trace data are thrown away.

[EDIT] Filtering extension specific info from the trace seems to fix the issue.


lighthouse-core/gather/gather-runner.js
```diff
@@ -70,8 +70,16 @@ class GatherRunner {
   static loadBlank(driver, url) {
     // The real about:blank doesn't fire onload and is full of mysteries (https://goo.gl/mdQkYr)
     // To improve speed and avoid anomalies (https://goo.gl/Aho2R9), we use a basic data uri page
-    url = url || `data:text/html,${logoPageSource}`;
-    const blankPageUrl = `data:text/html,${blankPageSource}`;
+
+    let blankPageUrl;
+
+    if (typeof chrome !== 'undefined') {
```
wardpeet: I don't mind this as a quickfix but could we maybe move this up to lighthouse options so we don't have to do a dirty check inside our gatherer for if we are a chrome extension

kdzwinel: This PR is more of a proof of concept than a final solution 👍 It definitely needs a clean up

paulirish:
Okay I think I see what failed here.. 

The error is:

> Protocol error (Emulation.setScriptExecutionDisabled): Cannot access contents of the page. Extension manifest must request permission to access the respective host.

Apparently the extension believe it cannot run `chrome.debugger.sendCommand` on this page. I tried adding `"<all_urls>"` to the manifest's `permissions` to no avail. But looking in the chromium source I found a good lead.

image

The error is emitted after seeing if the host page passes `URLPattern::IsValidSchemeForExtensions`. Also hilariously, it checks ` && document_url.spec() != url::kAboutBlankURL`, which is why we never encountered this before. We got safelisted! 

Okay so what's a valid scheme?

image

This list doesn't include our friend `data:` but we have a [9-year-old](https://chromium.googlesource.com/chromium/src/+/0b7c092f7f5b196fd9085f1ab796a0c9ac9473a6%5E%21/#F6) comment from chrome extensions' original architect, aaron boodman:

> TODO(aa): What about more obscure schemes like data: and javascript: ?
 
YAHHHHHHHHHH what about those. :)

---

Alright, well `chrome-extension:` scheme is OK, so I'm going to try using that first.

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -51,8 +51,17 @@ class TraceOfTab extends ComputedArtifact {
...
     // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
     // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
-    const startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+    const startedInPageEvt = keyEvents.find(e => {
+      if (e.name !== 'TracingStartedInPage') return false;
+      // In the extension, the first TracingStartedInPage is on chrome-extension://…/logo.html
+      // We'll await the next, which is the page/frame of the target URL
+      const url = e.args.data && e.args.data.frames && e.args.data.frames[0] &&
+          e.args.data.frames[0].url;
+      return !(url.startsWith('chrome-extension:') || url.startsWith('data:'));
```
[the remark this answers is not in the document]
paulirish: Yes, it does mean that.   Well, at least you won't get performance metrics. But conceivably the rest should work on chrome-extension. It won't on 'data:' via the LH extension for above reasons, but `data:` is hypothetically fine for CLI and devtools.   

...

yeah data uri works on the CLI:

```sh
lighthouse 'data:text/html;utf-8,<h1>omsdf</h1><img src="https://www.google.com/logos/doodles/2018/doodle-snow-games-day-1-4597519198715904.3-s.png">'
```
creates this one: https://googlechrome.github.io/lighthouse/viewer/?gist=7324cd97b0c25394274a169bef13b25a

that's amusing.

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -51,8 +51,17 @@ class TraceOfTab extends ComputedArtifact {
...
     // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
     // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
-    const startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+    const startedInPageEvt = keyEvents.find(e => {
+      if (e.name !== 'TracingStartedInPage') return false;
+      // In the extension, the first TracingStartedInPage is on chrome-extension://…/logo-page.html
+      // We'll await the next, which is the page/frame of the target URL
+      const url = e.args.data && e.args.data.frames && e.args.data.frames[0] &&
+          e.args.data.frames[0].url;
+      return !(url.startsWith('chrome-extension:') || url.startsWith('data:'));
```
patrickhulce: can we limit this to our `logo-page.html` somehow? maybe with `/^chrome-extension:.*\/(logo|blank)-page.html$/`? It would be nice to keep intact the CLI use case of auditing chrome extension pages

paulirish: ya! sg

kdzwinel: I agree with Patrick that we could be a bit more specific here, we can even do `chrome-extension://${chrome.runtime.id}/logo-page.html`. Also, I don't quite understand why we exclude `data:` here because AFAIK only `chrome-extension:` was problematic. Was `data:` also showing up in the trace?

paulirish:
FYI: ToT Chromium now is fine with these data uri's. 🎉  https://chromium-review.googlesource.com/919984

~April 17 that ships to stable.

kdzwinel:
@paulirish great work 👏

kdzwinel:
This was fixed long time ago. Closing!

PR #4475, opened 2018-02-08
core(main-resource): adjust main resource identification logic

### LINKED ISSUE #4440: rel="canonical" different domain false negative
States "Document does not have a valid rel=canonical. Points to a different domain."

This is not true for the sites I ran it for.

There appears to be a bug.

### LINKED ISSUE #4454: Main resource computed artifact does not match finalUrl
In our main-resource computed artifact we use different logic from how our final URL is computed. Significantly, it considers 304 to be redirect. We should unify this logic so issues like #4440 don't happen.

Final URL Update Logic - works pretty well, let's align on this for now
https://github.com/GoogleChrome/lighthouse/blob/3030b4f0651fc4eb2634600d7cf0dbca43972b32/lighthouse-core/gather/driver.js#L611-L622

Main Resource Logic - fails in cases like #4440 
https://github.com/GoogleChrome/lighthouse/blob/3030b4f0651fc4eb2634600d7cf0dbca43972b32/lighthouse-core/gather/computed/main-resource.js#L21-L28

kdzwinel:
this PR aligns main-resource gatherer logic with final URL logic from driver.js

https://github.com/GoogleChrome/lighthouse/blob/3030b4f0651fc4eb2634600d7cf0dbca43972b32/lighthouse-core/gather/driver.js#L611-L622

Fixes #4454
Fixes #4440

### DIFF
--- a/lighthouse-core/test/gather/fake-driver.js
+++ b/lighthouse-core/test/gather/fake-driver.js
@@ -16,7 +16,7 @@ module.exports = {
     return Promise.resolve();
   },
   gotoURL() {
-    return Promise.resolve('https://example.com');
+    return Promise.resolve('https://www.reddit.com/r/nba');
   },
   waitForLoadEvent() {
     return Promise.resolve();
--- a/lighthouse-core/test/runner-test.js
+++ b/lighthouse-core/test/runner-test.js
@@ -330,6 +330,9 @@ describe('Runner', () => {
       ],
 
       artifacts: {
+        URL: {
+          finalUrl: 'https://www.reddit.com/r/nba',
+        },
         devtoolsLogs: {
           defaultPass: path.join(__dirname, '/fixtures/perflog.json'),
         },
@@ -473,7 +476,7 @@ describe('Runner', () => {
     const config = new Config({
       passes: [{
         passName: 'firstPass',
-        gatherers: ['viewport-dimensions'],
+        gatherers: ['url', 'viewport-dimensions'],
       }],
 
       audits: [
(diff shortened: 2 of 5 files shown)


kdzwinel:
@patrickhulce @paulirish PTAL

kdzwinel:
@patrickhulce no worries! Reusing `URL.finalURL` makes much more sense, not sure how I missed that. I went ahead and updated the code. PTAL

lighthouse-core/test/gather/fake-driver.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
     return Promise.resolve();
   },
   gotoURL() {
-    return Promise.resolve('https://example.com');
+    return Promise.resolve('https://www.reddit.com/r/nba');
```
kdzwinel: that's the correct URL for the `../fixtures/traces/progressive-app.json` returned by fake-driver

lighthouse-core/test/runner-test.js
```diff
@@ -330,6 +330,9 @@ describe('Runner', () => {
       ],
 
       artifacts: {
+        URL: {
+          finalUrl: 'https://www.reddit.com/r/nba',
```
kdzwinel: critical-request-chains now depends on URL via mainResource

patrickhulce:
the appveyor failure is hopefully unrelated @kdzwinel ?

image

kdzwinel:
@patrickhulce fun bug alert!

Hm… It fails on AppVeyor but passes on Travis? Yesterday I couldn't reproduce it locally, but today I can?? It fails on master???

Mystery! So I started debugging and it looks like `mainResource.responseHeaders` suddenly misses one of the `Link` headers 🤔Someone messed with the static-server and commited a broken PR? Nope! It's Chrome:

DevTools in Chrome 63 (stable):
two-63

DevTools in Chrome 66 (one used for testing):
one-66

patrickhulce:
awesome sleuthing and thanks for filing the crbug @kdzwinel!

PR #4480, opened 2018-02-08
core(tracing-processor): fix scheduleable task logic

patrickhulce:
makes appveyor green again!

this [CL](https://chromium-review.googlesource.com/c/901322/) changed the name of scheduleable tasks from `TaskQueueManager::ProcessTaskFromWorkQueue` to `ThreadControllerImpl::DoWork` we now support both and throw loudly if we don't find any tasks at all


paulirish:
This fixes a bug with Chrome Canary: https://bugs.chromium.org/p/chromium/issues/detail?id=783309&desc=2#c22

Issue #3508, opened 2017-10-09
Handling lazy loaded images


midzer:
Hi all,
using LH v2.5 lazy loaded images above the fold are reported as "Properly size images" with 100% saving opportunity right now. I think this is misleading (or a bug) because those images are in correct dimension/size.

Furthermore those images result in a rather bad PSI pushing down overall performance rating. I read its value is calculated with all layout changes being done after initial page load. Maybe lazy loading elements above the fold has no point at all performance-wise? On the other hand, using lazy loading every heavy element FMP, FI and CI are fine :+1: 

Here is an example site showing the issue https://feuerwehr-eisolzried.de/

Thanks for this awesome tool and have a nice day
midzer

patrickhulce:
Thanks for verifying @philipwalton. I believe we originally threw around the idea of calling this "invisible images" but thought that would be even more confusing 😆 we'll have to dig into some of theses edges and decide how we should handle them.

tomayac:
This causes a lot of confusion, we need to rename this *urgently*. How about being very clear and just say “images that were loaded, but that aren’t being used on the page”?

dtomaszewski:
I think it caused additional problem see on https://memoteka.pl/ that lazy loaded images are marked as "offscreen"

kenhoward:
Like @neodigm, I set the lazy load to an absurd amount of time without change in results. I'm targeting the correct images that are in question.

patrickhulce:
@kenhoward are you still seeing problems with 3.0.0-beta? If so, please file a new issue with repro steps :)

neodigm:
I am no longer seeing this issue in the extension.

PR #4487, opened 2018-02-08
core(responsive-images): move images with no dimensions to offscreen audit

patrickhulce:
fixes https://github.com/GoogleChrome/lighthouse/issues/3508

previously, images that had dimensions of 0 were listed in both the "Properly size images" audit (uses-responsive-images) and the "Offscreen images" audit (offscreen-images).

This PR changes the name of offscreen to "Non-visible images" and "Properly size images" does not show an image if the area of the image was 0.

@midzer @tomayac @philipwalton @wardpeet does this sound good to you?


midzer:
Sounds awesome!
Looking forward for new version being released :)

Issue #4486, opened 2018-02-08
Fonts gatherer can take a very long time


patrickhulce:
While debugging the offscreen images issue on https://feuerwehr-eisolzried.de/ I noticed `Fonts` gatherer took 13 seconds and https://us.norton.com/ took more than 60s (!!).

@wardpeet maybe we should put a timeout on that loadStylesheetWithCORS function? or find a way to do it over protocol without rerequesting if possible :)

wardpeet:
Yeah I find a solution to make it faster. It's pretty slow when there are some CORS stylesheets. I'll fix that in a proper fix within the #4464 issue so we also load imported stylesheets

patrickhulce:
SG! I'm also not sure what was going on haha, I've re-run both of those after adding some logging and now I can't get it to take more than 1s so maybe just something odd with my machine too 🤷‍♂️

kdzwinel:
It's not just you 😉 I've run the extension quite a few times today and 'Fonts' was taking significant amount of time on non-trivial websites.

paulirish:
fixed by #4488, we believe. let's reopen immediately if we're seeing Fonts gatherer take > 1000ms

PR #4488, opened 2018-02-08
core(fonts): fix infinite loop

patrickhulce:
partial fix to address #4486 and prevent an infinite loop when getting an exception


wardpeet:
not sure that's the cause of infinite loop but LGTM :)

patrickhulce:
> not sure that's the cause of infinite loop but LGTM :)

it certainly was in my case :)

`getFontFaceFromStylesheets` calls `loadStylesheetWithCORS` on all errors, `loadStylesheetWithCORS` calls `getFontFaceFromStylesheets` and so the cycle repeats inifinitely if the error isn't solved by CORS fetching

PR #4537, opened 2018-02-14
report: add docs link to bootup-time audit

kaycebasques:
Doc isn't live yet, but should be soon...


kaycebasques:
Doc is live: https://developers.google.com/web/tools/lighthouse/audits/bootup

PR #4540, opened 2018-02-14
docs(programmatic): add caveat about removing artifacts

### LINKED ISSUE #4346: Update programmatic example to delete results.artifacts
I checked every 'report.html', file size of each report is larger than 20M bytes. It is impossible to serve on websites. Is there any way to shrink the size?

patrickhulce:
closes #4346

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -14,10 +14,11 @@ const chromeLauncher = require('chrome-launcher');
 function launchChromeAndRunLighthouse(url, opts, config = null) {
   return chromeLauncher.launch({chromeFlags: opts.chromeFlags}).then(chrome => {
     opts.port = chrome.port;
-    return lighthouse(url, opts, config).then(results =>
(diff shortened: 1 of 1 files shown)


patrickhulce:
does this help @ganlanyuan?

ganlanyuan:
Yes, that's helpful!
Thanks

ganlanyuan:
But this doesn't work for me, I just get the results of `undefined`:
```javascript
return lighthouse(url, opts, config).then(results => {
  delete results.artifacts;
  chrome.kill().then(() => {
    return new ReportGenerator().generateReportHtml(results);
  })});
```
While this works:
```javascript
return lighthouse(url, opts, config).then(results =>
  chrome.kill().then(() => {
    delete results.artifacts;
    return new ReportGenerator().generateReportHtml(results);
  }));
```

paulirish:
@ganlanyuan this is true. implicit return. updated the code to be more obvious about this. :)

Issue #4356, opened 2018-01-25
[SEO Audits] Integrate robots.txt analysis


rviscomi:
Using a JS-based robots.txt parser (like [this one](https://github.com/samclarke/robots-parser)), validate the file itself and apply existing SEO audits whenever applicable.

This integration has two parts:

- [ ] robots.txt is valid
- [ ] Page is not blocked from indexing

# robots.txt is valid (new audit)

**Audit group:** [Crawling and indexing](https://github.com/GoogleChrome/lighthouse/issues/3118)
**Description:** robots.txt is valid
**Failure description:** robots.txt is not valid
**Help text:** If your robots.txt file is malformed, crawlers may not be able to understand how you want your website to be crawled or indexed. Learn more.

**Success conditions:**

- robots.txt doesn't exist, otherwise:
  - it can be parsed without errors
  - it does not contain conflicting directives, eg `all, noindex`

# Page is not blocked from indexing

Add the following success condition:

- The robots.txt file does not include directives in the [blocklist](https://github.com/GoogleChrome/lighthouse/issues/3182).

Note that directives may be applied to the site as a whole or a specific page. Only fail if the current page is blocked from indexing (directly or indirectly).

kdzwinel:
Do we want to let user know if `/robots.txt` fails with something like HTTP 500? IMO if the response code is in HTTP 500 - 600 range we can safely report it as an issue.

kdzwinel:
```
User-agent: Googlebot
Disallow: / # everything is blocked for googlebot

User-agent: *
Disallow: # but allowed for everyone else
```

Should we fail in such case? How about robots.txt that is only blocking e.g. `Googlebot-Image`, or Bing, Yandex, DuckDuckGo? 🤔

rviscomi:
For consistency with https://github.com/GoogleChrome/lighthouse/issues/3182 let's try to avoid distinguishing between crawlers. If the common case is `*` UAs, then that's the one we should check. If possible, it would be great to warn saying "you passed, but you're blocking googlebot".

The alternative is to fail the audit when seeing anything resembling noindex, which seems too strict.

I'd also love to see the contents echoed back in the extra info table or similar. Just showing it to users is a sort of manual validation, even if the audit passes. As a secondary benefit, this would be great for data mining later.

PR #4548, opened 2018-02-15
core(is-crawlable): make sure that page is not blocked by robots.txt file

kdzwinel:
First part of #4356

Extends is-crawlable audit, so that it also verifies robots.txt file (besides meta tag and header).

screen shot 2018-02-15 at 14 37 51

This change adds a new dependency - [robots-parser](https://github.com/samclarke/robots-parser). It's a lightweight, well tested robots.txt parser with no dependencies.

### DIFF
--- a/lighthouse-core/gather/gatherers/seo/robots-txt.js
+++ b/lighthouse-core/gather/gatherers/seo/robots-txt.js
@@ -7,7 +7,7 @@
 
 const Gatherer = require('../gatherer');
 
-/* global fetch, URL */
+/* global fetch, URL, location */
 
 function getRobotsTxtContent() {
   return fetch(new URL('/robots.txt', location.href))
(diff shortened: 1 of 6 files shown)


lighthouse-core/gather/gatherers/seo/robots-txt.js
```diff
@@ -0,0 +1,36 @@
...
+  return fetch(new URL('/robots.txt', location.href))
+    .then(response => {
+      if (!response.ok) {
+        return {status: response.status, content: null};
+      }
+
+      return response.text()
+        .then(content => ({status: response.status, content}));
+    })
+    .catch(_ => ({status: null, content: null}));
```
kdzwinel: Second part of #4356 is a new audit that checks if robots.txt file is valid. That audit should fail if fetch returns HTTP 500+ and that's why we collect it.

paulirish:
Also... looking at the robots-parser test suite, do you think there are any additional cases they should add? 

i'm looking at https://github.com/ChristopherAkroyd/robots-txt-parser/tree/master/test/test-data https://github.com/python/cpython/blob/master/Lib/test/test_robotparser.py https://github.com/meanpath/robots/tree/master/test

kdzwinel:
@paulirish we do show where the blocking directive comes from (header, meta, robots.txt) - check out the screenshot at the top. I'd be great to show which robots.txt directive is blocking, but `robots-parser` doesn't tell us that.

Regarding tests - I quickly scanned test suites that you linked and `robots-parser` test file and I don't see any prominent test cases missing. Are you concerned about some specific case?

PR #4552, opened 2018-02-15
 core(lantern): improve RTT estimates

patrickhulce:
### changes
* estimates the observed RTT of each origin and takes this into account when simulating
* estimates the server response time of each origin using *all* available records, instead of just the records involved in the graph (prior behavior responsible for 1st bug in list below)
* ignores observed connectionId from Chrome when simulating and creates its own set of connections instead

### consequences
* fixes a bug where the pessimistic estimate could be smaller than the optimistic estimate due to different connections being used
* should substantially improve accuracy observed in LR when the connectionIds are all `0`
* reduces absolute estimate error by ~4% across the board (25% error -> 21% error, not 25% -> 24% 😉)
* marginally increases rank correlation on FCP, marginally decreases rank correlation on FMP, no impact to rank correlation of TTI

### DIFF
--- a/lighthouse-core/audits/predictive-perf.js
+++ b/lighthouse-core/audits/predictive-perf.js
@@ -76,26 +76,32 @@ class PredictivePerf extends Audit {
    * @param {!Node} dependencyGraph
    * @return {!Object}
    */
-  static computeOptions(dependencyGraph) {
+  static computeRTTAndServerResponseTime(dependencyGraph) {
     const records = [];
     dependencyGraph.traverse(node => {
       if (node.type === Node.TYPES.NETWORK) records.push(node.record);
     });
 
-    const rttSummaries = Array.from(NetworkAnalyzer.estimateRTTByOrigin(records).entries());
-    const rttByOrigin = new Map(rttSummaries.map(item => [item[0], item[1].min]));
+    // First pass compute the estimated observed RTT to each origin's servers.
+    const rttByOrigin = new Map();
+    for (const [origin, summary] of NetworkAnalyzer.estimateRTTByOrigin(records).entries()) {
+      rttByOrigin.set(origin, summary.min);
+    }
+
+    // We'll use the minimum RTT as the assumed connection latency since we care about how much addt'l
+    // latency each origin introduces as Lantern will be simulating with its own connection latency.
+    const minimumRtt = Math.min(...Array.from(rttByOrigin.values()));
+    // We'll use the observed RTT information to help estimate the server response time
     const responseTimeSummaries = NetworkAnalyzer.estimateServerResponseTimeByOrigin(records, {
       rttByOrigin,
     });
 
-    const baselineRtt = Math.min(...Array.from(rttByOrigin.values()));
-    const additionalRttByOrigin = new Map(
-      Array.from(rttByOrigin.entries()).map(item => [item[0], item[1] - baselineRtt])
-    );
-
-    const serverResponseTimeByOrigin = new Map(
-      Array.from(responseTimeSummaries.entries()).map(item => [item[0], item[1].median])
-    );
+    const additionalRttByOrigin = new Map();
+    const serverResponseTimeByOrigin = new Map();
+    for (const [origin, summary] of responseTimeSummaries.entries()) {
+      additionalRttByOrigin.set(origin, rttByOrigin.get(origin) - minimumRtt);
+      serverResponseTimeByOrigin.set(origin, summary.median);
+    }
 
     return {additionalRttByOrigin, serverResponseTimeByOrigin};
   }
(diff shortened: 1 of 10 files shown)


lighthouse-core/audits/predictive-perf.js
```diff
@@ -71,6 +72,34 @@ class PredictivePerf extends Audit {
     return scriptUrls;
   }
 
+  /**
+   * @param {!Node} dependencyGraph
+   * @return {!Object}
+   */
+  static computeOptions(dependencyGraph) {
```
paulirish: i appreciate the thought you put into the names in this file, but i think we can upgrade this one a bit. 
`computeRTTandSR` ?

lighthouse-core/audits/predictive-perf.js
```diff
@@ -220,9 +252,14 @@ class PredictivePerf extends Audit {
       };
 
       const values = {};
+      const options = PredictivePerf.computeOptions(graph);
       Object.keys(graphs).forEach(key => {
-        const estimate = new LoadSimulator(graphs[key]).simulate();
-        const lastLongTaskEnd = PredictivePerf.getLastLongTaskEndTime(estimate.nodeTiming);
+        const estimate = new LoadSimulator(graphs[key], options).simulate();
+        const longTaskThreshold = /optimistic/.test(key) ? 100 : 50;
```
paulirish: startsWith?

lighthouse-core/audits/predictive-perf.js
```diff
@@ -71,6 +72,34 @@ class PredictivePerf extends Audit {
...
+   * @param {!Node} dependencyGraph
+   * @return {!Object}
+   */
+  static computeOptions(dependencyGraph) {
+    const records = [];
+    dependencyGraph.traverse(node => {
+      if (node.type === Node.TYPES.NETWORK) records.push(node.record);
+    });
+
+    const rttSummaries = Array.from(NetworkAnalyzer.estimateRTTByOrigin(records).entries());
```
paulirish: later on you use this pattern, which i think is a lot more readable:

```js
      for (const [origin, summary] of rttByOrigin.entries()) {
        rttByOrigin.set(origin, summary.min);
      }
```
does take more lines, but worth it IMO

lighthouse-core/audits/predictive-perf.js
patrickhulce: sg done

lighthouse-core/audits/predictive-perf.js
```diff
@@ -71,6 +72,34 @@ class PredictivePerf extends Audit {
...
+    const rttByOrigin = new Map(rttSummaries.map(item => [item[0], item[1].min]));
```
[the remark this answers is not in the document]
patrickhulce: yeah I cleaned this whole function up

lighthouse-core/audits/predictive-perf.js
```diff
@@ -71,6 +72,34 @@ class PredictivePerf extends Audit {
...
+    const responseTimeSummaries = NetworkAnalyzer.estimateServerResponseTimeByOrigin(records, {
```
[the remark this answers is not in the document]
patrickhulce: yeah I cleaned this whole function up

lighthouse-core/audits/predictive-perf.js

lighthouse-core/audits/predictive-perf.js
patrickhulce: yeah cleaned this whole thing up

patrickhulce:
> In the non-LR case, we have the original data. So can we see how well our simulation did against observed connectionIds?

yes there's even a test for that ;) 
https://github.com/GoogleChrome/lighthouse/blob/23e6778d68a0989a14d36c4c71454b8f864bb3b6/lighthouse-core/test/lib/dependency-graph/simulator/network-analyzer-test.js#L147-L156

during simulation though this is a slightly different concern, essentially the connection IDs that were used during the fast load aren't necessarily the ones you want to force when simulating a very small subset. i.e. we happened to pick 3 assets that all shared the same connection ID when loading with 100 other resources but would've obviously been put onto different connections when loaded alone. In this sense simulator becomes a little more rational/browser-like.

> as an aside, think now's a good time to get type-checking enabled for lib/dependency-graph?

yes, but the error list was ~200 lines long so I'll save that for a followup :)

paulirish:
nice cleanup! thxx

-------

>> In the non-LR case, we have the original data. So can we see how well our simulation did against observed connectionIds?

> yes there's even a test for that ;)

nice nice. looked at the `'should approximate well with either method'` test and the results.

```
min
diff:  11.084 result:  2.621 approx:  13.705
avg
diff:  28.577 result:  10.712 approx:  39.29
median
diff:  10.237 result:  3.666 approx:  13.903
```

the gap between the result and approx is pretty big for all of these. I would have expected <= 20% difference. i'm not sure how we should quantify these are reasonable approximations; what do you think?

patrickhulce:
> the gap between the result and approx is pretty big for all of these. I would have expected <= 20% difference. i'm not sure how we should quantify these are reasonable approximations; what do you think?

~10ms gap is about as good as we can hope for given we trying to blindly guess what % of ~100ms request was due to RTT and what was server response time

it was more impressive when run on traces that actually had some latency, but all ours are pretty slim. want me to add some beefy ones to fixtures from the throttled dataset?

PR #4553, opened 2018-02-15
docs(headless-chrome): switch to prioritize headless

ebidel:
Prioritizing headless now that it supports emulation.

### DIFF
--- a/docs/headless-chrome.md
+++ b/docs/headless-chrome.md
@@ -2,9 +2,6 @@
 
 ## CLI (headless)
 
-> **Note**: Headless Chrome still has a few bugs to work out. For example, [network emulation](https://bugs.chromium.org/p/chromium/issues/detail?id=728451) is not supported yet.
-This can affect the accuracy of performance scores returned by Lighthouse.
-
 Setup:
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/headless-chrome.md
```diff
@@ -1,10 +1,33 @@
 # Running Lighthouse using headless Chrome
 
-For now, we recommend running Chrome with xvfb. See below.
+## CLI (headless)
+
+> **Note**: Headless Chrome still has a few bugs to work out. For example, [network emulation](https://bugs.chromium.org/p/chromium/issues/detail?id=728451) is not supported yet.
```
patrickhulce: can we remove this caveat now? :D

might be worth still noting there may be some graphics-related perf differences

docs/headless-chrome.md
```diff
@@ -1,10 +1,33 @@
...
+Kick off run of Lighthouse using headless Chrome:
+
+```sh
+lighthouse --chrome-flags="--headless" https://github.com
+```
 
 ## CLI (xvfb)
 
-Chrome + xvfb is the stable solution we recommend. These steps worked on Debian Jessie:
+Alternativeluy, you can run full Chrome + xvfb instead of headless mode. These steps worked on Debian Jessie:
```
patrickhulce: s/Alternativeluy/Alternatively

docs/headless-chrome.md
paulirish: 👍 . offline and throttling are implemented.  we can nuke the note.

there should probably be a list somewhere which documents known differences between headless and headful. but there isn't one yet.

paulirish: ping @ebidel

ebidel:
ready to go

PR #4601, opened 2018-02-21
core(byte-efficiency): use lantern for opportunity estimates

### LINKED ISSUE #3618: Lantern: Improve performance opportunity estimates based on graphs

patrickhulce:
closes #3618

This moves all byte efficiency audits to use lantern estimates. Render blocking resources and preload opportunities still use bespoke methods, but IMO anything else we do until we surface opportunities by metric would be wasted work anyhow. If others feel differently that it's important to update those too, happy to take a crack at it.

Overall approach: simulate the pessimistic TTCI graph as-is, modify all the resources to have the specified transfer size, re-simulate and report the difference


lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -28,15 +30,14 @@ module.exports = [
         score: '<100',
         extendedInfo: {
           value: {
-            wastedKb: 14,
+            wastedKb: 45,
```
brendankenny: as discussed, can we add more `wastedMs` values to attempt to catch regressions on our side or Chrome's?

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -89,47 +91,85 @@ class UnusedBytes extends Audit {
...
+        ])
+      )
+      .then(([result, graph]) => this.createAuditResult(result, graph));
   }
 
   /**
-   * @param {!Audit.HeadingsResult} result
-   * @param {number} networkThroughput
-   * @return {!AuditResult}
+   * @param {!Array<{url: string, wastedByte: number}>} results
```
brendankenny: string to explain method/`result`?

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -89,47 +91,85 @@ class UnusedBytes extends Audit {
...
+        item.totalKb = this.bytesToKbString(item.totalBytes);
+        item.potentialSavings = this.toSavingsString(item.wastedBytes, wastedPercent);
+        return item;
+      })
+      .sort((itemA, itemB) => itemB.wastedBytes - itemA.wastedBytes);
 
     const wastedBytes = results.reduce((sum, item) => sum + item.wastedBytes, 0);
     const wastedKb = Math.round(wastedBytes / KB_IN_BYTES);
-    const wastedMs = Math.round(wastedBytes / networkThroughput * 100) * 10;
+    const wastedMs = this.computeWasteWithTTIGraph(results, graph);
```
brendankenny: `results` don't have `url`s? (needed by `computeWasteWithTTIGraph`)

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
patrickhulce: done

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
patrickhulce: but they do? :)

brendankenny: Ah, I also missed that it was `result.results.map(...)`

patrickhulce:
anything else this is waiting on @brendankenny @paulirish ?

paulirish:
> Render blocking resources and preload opportunities still use bespoke methods, but IMO anything else we do until we surface opportunities by metric would be wasted work anyhow. If others feel differently that it's important to update those too, happy to take a crack at it.

I think we should do those as well. I'm fine with it being a separate PR, but I'd like to have our opportunities move to lantern at the same time (ie. same release). Also have a feeling that opps-by-metric will be landing a while from now.

patrickhulce:
> I think we should do those as well. I'm fine with it being a separate PR, but I'd like to have our opportunities move to lantern at the same time (ie. same release). 

Definitely same release (3.0)

> Also have a feeling that opps-by-metric will be landing a while from now.

Not as confident 😄we've got 4 weeks 'til due date for this and with this PR inflight ~3 weeks that puts me needing to finish it this sprint 😆

PR #4607, opened 2018-02-23
(core) response-compression: also check x-original-content-encoding

jianzhoufeng:
@patrickhulce 

PTAL.

Thanks!


patrickhulce:
one more question though @jianzhoufeng, does this mean the `transferSize` properties on network records are also the decoded length or are those alright?

jianzhoufeng:
@patrickhulce: The transferSize set from encodedDataLength of the Network.loadingFinished event is still the encoded size. If the transferSize is set from other devtools events, it might be the decoded length, though I need to check.

Issue #4614, opened 2018-02-23
Improve LHR API shape for downstream consumers


paulirish:
We have downstream consumers of Lighthouse data that want to use the details within `extendedInfo` and/or `details`. 

However there are a few problems that make it difficult:

* `details` currently ships mostly strings like "42 ms". consumers gotta parse a number out of there. :/
* grabbing data from `details` requires relying on array indices and looking up those locations from a `itemHeaders` (which also doesn't offer reliable key names)
* `extendedInfo` we've never made promises on, but much of the good stuff is there. (we know, as we use it almost exclusively for unit tests)
* the two overlap awkwardly.
* probably some more.

Fixing this is a breaking change (_Hello #4333!_), but it's worth it.

Background doc: [AuditResults Data format discussion](https://docs.google.com/document/d/1G_5jVIlxnRz4vhQDEzuVqGakR4UmgFP1vGouvY0HEOw/edit)

paulirish:
Current Plan:

1. https://github.com/GoogleChrome/lighthouse/pull/4384 ships fewer strings in the LHR
1. https://github.com/GoogleChrome/lighthouse/pull/4616 makes `details` accessible by property
1. A subsequent PR will remove `extendedInfo` (and update all those tests)
   * We'll also need to take inventory of extendedInfo data that's currently not in the associated `details` block, but should be. (here's [what HA uses](https://github.com/HTTPArchive/beta.httparchive.org/search?utf8=%E2%9C%93&q=audits&type=), what [brendan uses](https://github.com/brendankenny/bq-har/search?utf8=%E2%9C%93&q=extendedInfo&type=))
1. We're discussing a `load-metrics` audit that contains all timings/timestamps in its `details`, so all the things don't need to reach into `FMP.extInfo`

paulirish:
First build from this 3.0 effort. The build is based on [this branch](https://github.com/GoogleChrome/lighthouse/tree/newdetails-noextinfo), which is 3 commits beyond #4616.

[lighthouse-ext-background.js.zip](https://github.com/GoogleChrome/lighthouse/files/1774664/lighthouse-ext-background.js.zip)

PR #4616, opened 2018-02-23
core(lhr): overhaul LHR details, introduce details.summary

paulirish:
Alright, #4384 laid the groundwork here and I've built off that for round two in the goal of _**"making the juicy details in LHR easy to consume."**_

(This is a super-breaking change. Yes.)

### Changes
* completely new shape of `details`:
  * `result.details.headings` describes the expected report columns
  * `result.details.items` contains row information, and may include _more_ data than referenced by the headings. it's the new `extendedInfo`. :)
* a new `result.details.summary` obj that includes whatever toplevel numeric summary we have. So far just `wastedMs` and `wastedKb`, but I could see `displayValue` migrating into it.

I think this is easiest to evaluate by looking at the LHR json. Here's one audit result, before and after:

🔎  Also here's a complete LHR: [tinyhouse.json.txt](https://github.com/GoogleChrome/lighthouse/files/1776823/tinyhouse.json.txt)

@patrickhulce @brendankenny  Thoughts on the new LHR shape?

cc @benschwarz @denar90 

------- 

This is a chain of PRs. Landing order is... 1st: `newdetails` (#4616). 2nd: `shallowCategories` (#4711). 3rd: `scoring2.0` (#4690)

ref #4614

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
@@ -7,6 +7,7 @@
 
 const assert = require('assert');
 const parseCacheControl = require('parse-cache-control');
+const Audit = require('../audit');
 const ByteEfficiencyAudit = require('./byte-efficiency-audit');
 const WebInspector = require('../../lib/web-inspector');
 const URL = require('../../lib/url-shim');
@@ -18,7 +19,7 @@ const IGNORE_THRESHOLD_IN_PERCENT = 0.925;
 const SCORING_POINT_OF_DIMINISHING_RETURNS = 4; // 4 KB
 const SCORING_MEDIAN = 768; // 768 KB
 
-class CacheHeaders extends ByteEfficiencyAudit {
+class CacheHeaders extends Audit {
   /**
    * @return {!AuditMeta}
    */
@@ -186,13 +187,7 @@ class CacheHeaders extends ByteEfficiencyAudit {
 
         const url = URL.elideDataURI(record._url);
         const totalBytes = record._transferSize;
-        const totalKb = ByteEfficiencyAudit.bytesDetails(totalBytes);
         const wastedBytes = (1 - cacheHitProbability) * totalBytes;
-        const cacheLifetimeDisplay = {
-          type: 'ms',
-          value: cacheLifetimeInSeconds,
-          displayUnit: 'duration',
-        };
 
         totalWastedBytes += wastedBytes;
         if (url.includes('?')) queryStringCount++;
@@ -201,9 +196,7 @@ class CacheHeaders extends ByteEfficiencyAudit {
           url,
           cacheControl,
           cacheLifetimeInSeconds,
-          cacheLifetimeDisplay,
           cacheHitProbability,
-          totalKb,
           totalBytes,
           wastedBytes,
         });
@@ -225,11 +218,13 @@ class CacheHeaders extends ByteEfficiencyAudit {
 
       const headings = [
         {key: 'url', itemType: 'url', text: 'URL'},
-        {key: 'cacheLifetimeDisplay', itemType: 'text', text: 'Cache TTL'},
-        {key: 'totalKb', itemType: 'text', text: 'Size (KB)'},
+        {key: 'cacheLifetimeInSeconds', itemType: 'ms', text: 'Cache TTL', displayUnit: 'duration'},
+        {key: 'totalBytes', itemType: 'bytes', text: 'Size (KB)', displayUnit: 'kb',
+          granularity: 1},
       ];
 
-      const tableDetails = ByteEfficiencyAudit.makeTableDetails(headings, results);
+      const summary = {wastedBytes: totalWastedBytes};
+      const details = ByteEfficiencyAudit.makeTableDetails(headings, results, summary);
 
       return {
         score,
@@ -241,7 +236,7 @@ class CacheHeaders extends ByteEfficiencyAudit {
             queryStringCount,
           },
         },
-        details: tableDetails,
+        details,
       };
     });
   }
--- a/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
@@ -8,7 +8,6 @@
 const assert = require('assert');
 const parseCacheControl = require('parse-cache-control');
 const Audit = require('../audit');
-const ByteEfficiencyAudit = require('./byte-efficiency-audit');
 const WebInspector = require('../../lib/web-inspector');
 const URL = require('../../lib/url-shim');
 
@@ -32,7 +31,7 @@ class CacheHeaders extends Audit {
       helpText:
         'A long cache lifetime can speed up repeat visits to your page. ' +
         '[Learn more](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#cache-control).',
-      scoringMode: ByteEfficiencyAudit.SCORING_MODES.NUMERIC,
+      scoringMode: Audit.SCORING_MODES.NUMERIC,
       requiredArtifacts: ['devtoolsLogs'],
     };
   }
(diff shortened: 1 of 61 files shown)


lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -315,7 +337,7 @@ class DetailsRenderer {
    */
   _renderCode(details) {
     const pre = this._dom.createElement('pre', 'lh-code');
-    pre.textContent = details.text;
+    pre.textContent = details.text || details.value; // TODO, pick a winner
```
patrickhulce: +1 for `value`

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -165,15 +169,16 @@ class DetailsRenderer {
    * @param {!DetailsRenderer.ThumbnailDetails} value
    * @return {!Element}
    */
-  _renderThumbnail(value) {
-    if (/^image/.test(value.mimeType) === false) {
-      return this._dom.createElement('span');
-    }
+  _renderThumbnail(details) {
+    // TODO: use some empty details test instead? to handle data uris?
```
patrickhulce: yeah let's strip out all data URIs at the audit base-class, they can make the report super beefy anyway

lighthouse-core/report/v2/renderer/performance-category-renderer.js
```diff
@@ -45,13 +45,13 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
    * @return {!Element}
    */
   _renderPerfHintAudit(audit, scale) {
-    const extendedInfo = /** @type {!PerformanceCategoryRenderer.PerfHintExtendedInfo}
-        */ (audit.result.extendedInfo);
+    const summaryInfo = /** @type {!PerformanceCategoryRenderer.PerfHintSummaryInfo}
+        */ (audit.result.summary);
```
patrickhulce: I might be missing something, I thought we were nesting `summary` inside of `details`?

lighthouse-core/scoring.js
```diff
@@ -53,15 +53,14 @@ class ReportScoring {
           result.informative = true;
         }
 
-        return Object.assign({}, audit, {result, score: auditScore});
```
patrickhulce: while we're in the spirit of breaking changes, let's go whole hog and do it at the audit class level :D

denar90:
Wow, big amount of work! Thanks for noticing 👍
Is there any v3 roadmap issue? I just found only 2 pr's but no doubt there will be more)

paulirish:
> Is this what we currently have?

Yes, that looks mostly accurate. I'm almost done with typechecking the JSON against an iteration of these interfaces. ;)

----------

In the case of PSI audits we have an even simpler details story:

```ts
interface AuditDetails {
  summary?: AuditDetailsSummary  // could be also be a sibling to details..
  headings?: AuditTableHeading[]
  items?: AuditDetailsItem[]
}

interface AuditDetailsSummary {
  wastedMs?: number // maybe these two within a `performanceOpportunity` object?
  wastedKb?: number
}

interface AuditTableHeading {
  key: string
  text: string
  itemType: string
  granularity?: string
  displayUnit?: string
}

interface AuditDetailsItem {
  [key: string]: string | number
}
```

paulirish:
@patrickhulce this is ready for review.

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -18,7 +18,7 @@ const IGNORE_THRESHOLD_IN_PERCENT = 0.925;
 const SCORING_POINT_OF_DIMINISHING_RETURNS = 4; // 4 KB
 const SCORING_MEDIAN = 768; // 768 KB
 
-class CacheHeaders extends ByteEfficiencyAudit {
+class CacheHeaders {
```
patrickhulce: @paulirish let's make sure this extends audit

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -133,11 +137,16 @@ class DetailsRenderer {
    * @return {!Element}
    */
   _renderLink(details) {
+    console.assert(('text' in details), 'link details must have .url');
```
patrickhulce: seems like this should be `'url' in details`?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -209,26 +214,46 @@ class DetailsRenderer {
...
     const tbodyElem = this._dom.createChildOf(tableElem, 'tbody');
     for (const row of details.items) {
       const rowElem = this._dom.createChildOf(tbodyElem, 'tr');
-      for (const columnItem of row) {
-        const classes = `lh-table-column--${columnItem.type}`;
-        this._dom.createChildOf(rowElem, 'td', classes).appendChild(this.render(columnItem));
+      for (const heading of details.headings) {
+        const value = /** @type {number|string|!DetailsRenderer.DetailsJSON} */ (row[heading.key]);
+        if (typeof value === 'undefined') {
+          continue; // e.g. no lineNumber in this row item
```
patrickhulce: don't we still want to create a td element even if it's empty?

lighthouse-core/runner.js
```diff
@@ -93,9 +93,10 @@ class Runner {
       const resultsById = {};
       for (const audit of runResults.auditResults) resultsById[audit.name] = audit;
 
-      let report;
+      let cats;
```
patrickhulce: cats

can we go `categories` though :)

paulirish:
updated and squashed (to resolve silly conflicts)

paulirish:
@patrickhulce lgty?

lighthouse-core/report/v2/renderer/category-renderer.js
```diff
@@ -203,9 +203,10 @@ class CategoryRenderer {
     const auditsGroupedByGroup = /** @type {!Object<string,
         !Array<!ReportRenderer.AuditJSON>>} */ ({});
     manualAudits.forEach(audit => {
-      const group = auditsGroupedByGroup[audit.group] || [];
+      const groupId = audit.group || 'ungrouped';
```
brendankenny: what is this doing?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -70,29 +72,31 @@ class DetailsRenderer {
...
 
   /**
-   * @param {!DetailsRenderer.DetailsJSON} text
+   * @param {!DetailsRenderer.StringDetailsJSON} text
    * @return {!Element}
    */
   _renderTextURL(text) {
-    const url = text.text || '';
+    console.assert(!('text' in text), 'url details should not have .text');
+    console.assert('value' in text, 'url details must have .value');
```
brendankenny: these should be checked by the type system?

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -133,11 +137,16 @@ class DetailsRenderer {
    * @return {!Element}
    */
   _renderLink(details) {
+    console.assert(('url' in details), 'link details must have .url');
+    console.assert(('text' in details), 'link details must have .text');
```
brendankenny: same as above

lighthouse-core/report/v2/renderer/details-renderer.js
```diff
@@ -425,3 +456,11 @@ DetailsRenderer.LinkDetailsJSON; // eslint-disable-line no-unused-expressions
...
  */
 DetailsRenderer.FilmstripDetails; // eslint-disable-line no-unused-expressions
+
+
+/** @typedef {{
+ *     wastedMs: (number|undefined),
+ *     wastedBytes: (number|undefined),
+ * }}
+ */
+DetailsRenderer.DetailsSummary; // eslint-disable-line no-unused-expressions
```
brendankenny: this seems more specific than a just a "summary"?

brendankenny:
if we want to get really fancy (and if we do TS checking of the report renderer), we can use [discriminated unions](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) which will type check based on the value of `type` in a particular branch. Will need to be in an externs file, though, because closure will choke on `type: "text"` in the typedef

paulirish:
> also, it looks like we won’t be able to non-awkwardly be able to get read of text.text since so many other details types still include a text?

fwiw the current shape of details objects looks like this:



### detail types with 1 required field
type `text` with `.value`
type `code` with `.value`
type `thumbnail` w/  `.value`

type `ms`  with `.value`
type `bytes` w  `.value`
type `url`   w  `.value`


### detail types with >1 required field
type `link` with `.text` and `.url`
type `node` with `.snippet` `.selector` `.path`
type `card` with `.title` `.value` `.target` `.snippet`  

---

Additionally, the second block of 1-field types (ms/bytes/url) are currently never expressed in the LHR explicitly; they are all created via `headings..itemType`.

PR #4631, opened 2018-02-27
core(network-requests): add network requests audit

### LINKED ISSUE #4517: Audit: details of all network requests
An audit we want to expose mostly for diagnostic purposes. Should include:

### props:

* url
* start time (network OR renderer-level)
* end time (not duration)
* transfer size
* status code

### exposed as:

lhr.audits['network-requests'].details  // type table, etc.

patrickhulce:
closes #4517


lighthouse-core/audits/network-requests.js
```diff
@@ -0,0 +1,67 @@
...
+          startTime: (record.startTime - earliestStartTime) * 1000,
+          endTime: (record.endTime - earliestStartTime) * 1000,
+          transferSize: record.transferSize,
+          statusCode: record.statusCode,
+        };
+      });
+
+      const headings = [
+        {key: 'url', itemType: 'url', text: 'URL'},
+        {key: 'startTime', itemType: 'ms', text: 'Start Time'},
```
patrickhulce: @paulirish not sure how we should handle these I was straddling the fence between current method and your next-gen method since we don't really care how the user sees them but......it feels like we should only really expect this to be consumed after your mega-awesome PR lands?

paulirish: yes i'd assume so.  
tbh i think should branch off a 2.x branch and keep master as the v3. we can ship both easily and continue to land whatever on both sides.

lighthouse-core/config/default.js
```diff
@@ -285,6 +286,7 @@ module.exports = {
         {id: 'uses-long-cache-ttl', weight: 0, group: 'perf-info'},
         {id: 'dom-size', weight: 0, group: 'perf-info'},
         {id: 'critical-request-chains', weight: 0, group: 'perf-info'},
+        {id: 'network-requests', weight: 0, group: 'perf-info'},
```
paulirish: leave off the group name and it'll be excluded from the report, a la `screenshot-thumbnails`?

patrickhulce: yeah sure sg

lighthouse-core/audits/network-requests.js
paulirish: image
yeah. samesies.

wardpeet: i'm all in for a release 2.0 branch

lighthouse-core/audits/network-requests.js
```diff
@@ -0,0 +1,67 @@
...
+        {key: 'endTime', itemType: 'ms', text: 'End Time'},
+        {key: 'transferSize', itemType: 'bytes', text: 'Transfer Size'},
```
wardpeet: shouldn't we set itemType to 'text' and convert it to kb? as the whole report shows kb?

Issue #4479, opened 2018-02-08
Integration testing for the Chrome extension


paulirish:
Having this would help prevent bugs like #4466 #4467 #4469

Current best idea to pull this off:

1. Use puppeteer to drive chrome headful
1. Probably use some aspects of [this guide from eric](https://stackoverflow.com/questions/48089670/detect-and-test-chrome-extension-using-puppeteer)
1. Issue [`Input.dispatchKeyEvent`](https://chromedevtools.github.io/devtools-protocol/tot/Input#method-dispatchKeyEvent) and set a `nativeVirtualKeyCode`. Joel believes this sends the keys to browser and thus we could tab to the pageaction button. (perhaps puppeteer handles the details, if not we can fall back to `_client.send()` ourselves)
1. Assuming we can trigger and interact with the popup, we can test the entire flow. If we cannot, we can at least interact with the background page and trigger the run from there.
1. Assert the report page opened, has the right details, etc.

wardpeet:
I can pick up this one :)

ebidel:
I'd be willing to pick this up too. We need a good example of crx testing using pptr!

If the dispatch key stuff doesn't work out, you could try opening the page directly and test them:

``js
// Can we navigate to a chrome-extension page? YES!
const page = await browser.newPage();
// Create a stable id: https://stackoverflow.com/a/23877974/274673
await page.goto('chrome-extension://ipfiboohojhbonenbbppflmpfkakjhed/popup.html');
// click buttons, check boxes, etc.
```

...Although the extension would need to be modified a bit to accept a test url from an `env` var or something. The [extension looks at the current tab](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-extension/app/src/popup.js#L228) for the test url, and if the tab is the popup page, you're not going to have a test URL :(

`_client.send()`...we've got https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-cdpsession now.

wardpeet:
Feel free to take it if you have the time for it :)

ebidel:
Yea :\ Probably best if someone else tackles it. I've got too much going on right now. I'm happy to consult or review PRs. The SO answer should mostly get you cooking.

paulirish:
After discussing with the pptr team (and a small prototype) it doesn't appear we can open the pop-up.html w

paulirish:
[...]

PR #4640, opened 2018-02-27
extension(tests): Add extension smoketest

wardpeet:
This PR will run the extension in chrome and checks if the report has all necessary fields listed.

the manifest change is needed because puppeteer can't detect the active tab. We need to add tabs to the permission list, we could temporary add it for each run so we don't have to update our manifest.

Also a caveat, we need to run puppeteer in non headless mode to let the extensions work


wardpeet:
Fixes #4479

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,91 @@
...
+                awaitPromise: true,
+                returnByValue: true,
+              }
+            );
+          })
+          .then(() => {
+            return browser.pages().then(pages => {
+              const page = pages.find(page => page.url().includes('blob:chrome-extension://'));
+              const assertAudits = (category, expected, selector) =>
+                page.$eval(`#${category}`,
```
paulirish: trying this out to see if it's more clear:

```js
function assertAuditElements(category, expected, selector) {
  page.$evaluate(({category, selector}) => 
    document.querySelector(`#${category}`).parentNode.querySelectorAll(selector).length,
    {category, selector}
  ).then(elementCount => {
    assert.equal(expected, elementCount);
  });
}

```

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,91 @@
...
+      return page.goto('https://www.paulirish.com', {waitUntil: 'networkidle2'});
+    })
+    .then(() => browser.targets())
+    .then(targets => {
+      const extensionTarget = targets.find(target => {
+        return target._targetInfo.title === 'Lighthouse';
+      });
+
+      if (extensionTarget) {
+        extensionTarget.createCDPSession()
```
ebidel: Any way to use puppeteer's `evaluate` method(s) here 

https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#executioncontextevaluatepagefunction-args maybe?

```js
await ctx.evaluate(() => {
  return runLighthouseInExtension({restoreCleanState: true}, ['performance', 'pwa', 'accessibility', 'best-practices', 'seo']);
});
```

Whatever that looks like. Not sure if you can get the crx context, other than what you're doing using an private api.

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,91 @@
...
+const config = require(path.resolve(__dirname, '../../lighthouse-core/config/default.js'));
+
+puppeteer.launch({
+  headless: false,
+  executablePath: process.env.CHROME_PATH ? process.env.CHROME_PATH : undefined,
+  args: [
+    `--disable-extensions-except=${lighthouseExtensionPath}`,
+    `--load-extension=${lighthouseExtensionPath}`,
+  ],
+}).then(browser => {
```
paulirish: Let's do async/await for this file. I'm sorry we didn't sort that out beforehand.

But IMO that'll make this guy a lot more readable.

lighthouse-extension/test/extension.js
ebidel: Having a hard time parsing this. Maybe just use `function`. And I think selector can just be passed as a string, not in in an obj:

```js
const assertNumAuditsEqual  = function(el, selector) {
  return page.$eval(`#${category}`, (el, selector) => el.parentNode.querySelectorAll(selector).length, selector));
}
```

paulirish: also, let's use a keyed object for the method signature?   


```js
assertAudits({
	category: 'performance',
	count: config.categories.performance.audits.length,
	selector: '.lh-audit,.lh-timeline-metric,.lh-perf-hint,.lh-filmstrip'
});
```

lighthouse-extension/app/manifest.json
```diff
@@ -18,6 +18,7 @@
   },
   "permissions": [
     "activeTab",
+    "tabs",
```
paulirish: i remember you telling me, but remind me again why we now need this permission?

.travis.yml
```diff
@@ -19,24 +19,24 @@ cache:
...
-  - echo $TRAVIS_EVENT_TYPE;
-  - echo $TRAVIS_BRANCH
   - yarn bundlesize
   - yarn lint
   - yarn unit
   - yarn type-check
   - yarn closure
   - yarn smoke
   - yarn smokehouse
+  - node lighthouse-extension/test/extension.js
```
paulirish: let's add an npm scripts entry to the extensions package.json so it's easier to run on demand

    cd lighthouse-extension; yarn test:e2e

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,91 @@
...
+                .then(categories => {
+                  assert.equal(categories.length, 5);
+
+                  return assertAudits(
+                    'performance',
+                    config.categories.performance.audits.length,
+                    '.lh-audit,.lh-timeline-metric,.lh-perf-hint,.lh-filmstrip'
+                  );
+                })
+                .then(
```
ebidel: the formatting of these then's should be the same. I may also break this testing section into a separate function if you can.

lighthouse-extension/app/manifest.json
patrickhulce: let's definitely add this on demand as you suggested @wardpeet, grabbing a new permission might still disable the extension on update?

ebidel: I don't think we need it.  In https://stackoverflow.com/questions/48089670/detect-and-test-chrome-extension-using-puppeteer, I needed it to get the title of tab. We're not doing anything special with the tabs api here.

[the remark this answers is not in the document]
wardpeet: I could change the config just during the test so we don't need to change it for installations?

.travis.yml
```diff
@@ -19,24 +19,24 @@ cache:
...
     - /home/travis/.rvm/gems/
 install:
+  - export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1
   - yarn
   # travis can't handle the parallel install (without caches)
   - yarn run install-all:task:windows
 before_script:
   - export DISPLAY=:99.0
-  - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
+  - export CHROME_PATH="$(which google-chrome-stable)"
```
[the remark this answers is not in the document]
wardpeet: I just thought it was a bit stupid to install it twice and we moved to chrome stable to test the lowest version we support. I could enable chromium again.

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,91 @@
...
+const path = require('path');
+const assert = require('assert');
+const puppeteer = require('puppeteer');
+
+const lighthouseExtensionPath = path.resolve(__dirname, '../app');
+const config = require(path.resolve(__dirname, '../../lighthouse-core/config/default.js'));
+
+puppeteer.launch({
+  headless: false,
+  executablePath: process.env.CHROME_PATH ? process.env.CHROME_PATH : undefined,
```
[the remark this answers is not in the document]
wardpeet: Thanks!

lighthouse-extension/app/manifest.json
wardpeet: The reason why I need the extension is the following function
```
 const queryOpts = {
        active: true,
        currentWindow: true,
      };

      chrome.tabs.query(queryOpts, (tabs => {
```
without it I couldn't get the url of the tab

lighthouse-extension/test/extension.js
wardpeet: what about node 6 which we run our PR checks on? it doesn't support async/await right?

lighthouse-extension/test/extension.js
wardpeet: sounds neat, will try :)

.travis.yml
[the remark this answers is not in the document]
paulirish: well, it does _appear_ to work fine with chrome stable. 
I'm also fine with using the bundled Chromium, but yes it does seem wasteful to d/l each time. 

If it works fine with chrome stable as of now, then moving forward (assuming we infrequently bump the pptr version), it shouldn't regress. So I'm +1 to using the stable.

lighthouse-extension/app/manifest.json
paulirish: Yessssss. I repro'd this.
Basically I believe activeTab isn't enough because of however we're bypassing the popup and such. It's hitting some corner case where `activeTab` just isn't enough. +1 to dynamically adding tabs permission just for this case.

lighthouse-extension/test/extension.js
paulirish: we're about to drop node 6 support. :) we'll land this patch in the 3.0 side of things.

lighthouse-extension/test/extension.js
paulirish: yup +1 to using this rather than dropping to the protocol. good call.

wardpeet: @ebidel not sure how I can get that execution context from the Target class. I tried `.page().mainFrame().executionContext()` but page() is of course null.

Any ideas? 😛

lighthouse-extension/package-lock.json
```diff
@@ -0,0 +1,5979 @@
+{
```
paulirish: we're still yarn folk so we should be updating the yarn lock and not adding this, right?
or did you already point out why we need this and i'm forgetting?

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,104 @@
...
+
+  const page = await browser.newPage();
+  await page.goto('https://www.paulirish.com', {waitUntil: 'networkidle2'});
+  const targets = await browser.targets();
+  const extensionTarget = targets.find(({_targetInfo}) => {
+    return _targetInfo.title === 'Lighthouse' &&
+      _targetInfo.type === 'background_page';
+  });
+
+  if (extensionTarget) {
```
paulirish: early exit instead

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,104 @@
...
+const path = require('path');
+const assert = require('assert');
+const fs = require('fs');
+const puppeteer = require('puppeteer');
+
+const lighthouseExtensionPath = path.resolve(__dirname, '../app');
+const config = require(path.resolve(__dirname, '../../lighthouse-core/config/default.js'));
+const lighthouseCategories = Object.keys(config.categories);
+
+const manifestLocation = path.join(lighthouseExtensionPath, 'manifest.json');
```
paulirish: can you pull this stuff into a little named function? might as well create another for putting the orig back.

paulirish: actually if we use mocha to run this test, then we can use before/after methods for these things.

seems fairly reasonable to use mocha at this point

lighthouse-extension/test/extension.js
```diff
@@ -0,0 +1,104 @@
...
+          throw new Error(lighthouseResult.exceptionDetails.exception.description);
+        }
+
+        throw new Error(lighthouseResult.exceptionDetails.text);
+      }
+
+      extensionPage = (await browser.pages())
+        .find(page => page.url().includes('blob:chrome-extension://'));
+
+      const categories = await extensionPage.$$(`#${lighthouseCategories.join(',#')}`);
```
paulirish: i'd pull 80-94 into a separate function, so it's clear that this is the test assertions against the report page

lighthouse-extension/package-lock.json
wardpeet: no no, it's my mistake . It shouldn't be here. I use yarn in WSL but npm on powershell because i'm to lazy to install yarn on windows as I only need it to run puppeteer (i'll probably create a bug for that in the puppeteer repository).

lighthouse-extension/test/extension.js
wardpeet: any idea how I should name this test? describe and it text😛.

lighthouse-extension/test/extension.js
ebidel: Asking...

lighthouse-extension/package-lock.json
paulirish: fwiw `npm install -g yarn` is what we do on appveyor. probably not too bad.

paulirish: also i think pptr is moving to package-lock soon.

lighthouse-extension/test/extension.js
paulirish: describe('lighthouse chrome extension', ...
it('completes an end-to-end run', ...

lighthouse-extension/package.json
```diff
@@ -22,6 +23,7 @@
     "gulp-util": "^3.0.7",
     "gulp-zip": "^3.2.0",
     "package-json-versionify": "^1.0.4",
+    "puppeteer": "^1.1.1",
```
brendankenny: what do folks think about moving puppeteer up to a root-level dev dependency? As I was testing viewer yesterday I was thinking it would be great to have a similar (but should be much simpler) e2e test for viewer, and it seems a bit silly to have a copy here and in viewer when `yarn install-all` would install both anyway...

ebidel: Great idea. Viewer UI testing is the perfect use case for 🤹‍♀️

paulirish: sgtm

wardpeet:
@paulirish yarn compile-devtools is failing. How can I fix that?

paulirish:
>  yarn compile-devtools is failing. How can I fix that?

You can't. :) I filed #4735 to look into it. I'll sort it out with brendan.

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,130 @@
...
+  assert.equal(categories.length, lighthouseCategories.length,
+    `${categories.join(' ')} does not match ${lighthouseCategories.join(' ')}`);
+
+  for (const category of lighthouseCategories) {
+    let selector = '.lh-audit';
+    let expected = config.categories[category].audits.length;
+    if (category === 'performance') {
+      selector = '.lh-audit,.lh-timeline-metric,.lh-perf-hint,.lh-filmstrip';
+      // we deducted the expected by 1 because network-requests audit is not showing in the report
+      expected = expected - 1;
```
paulirish: i think what we want here is more like

```js
expected = config.categories[category].audits.map(a => !!a.group).length;
```

lighthouse-core/scripts/run-mocha.sh
```diff
@@ -18,10 +18,13 @@ elif [ "$flag" == '--cli' ]; then
   _runmocha 'lighthouse-cli'
 elif [ "$flag" == '--viewer' ]; then
   _runmocha 'lighthouse-viewer'
+elif [ "$flag" == '--extension' ]; then
```
brendankenny: I missed the earlier conversation, but do we want this in the unit tests? Considerably slower than (most? all?) the rest of them. It's also not really a unit test...

`yarn test-extension` or something instead?

brendankenny: having it under `test` in the extension `package.json` would make sense as well

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,130 @@
...
+const fs = require('fs');
+const puppeteer = require('puppeteer');
+
+const lighthouseExtensionPath = path.resolve(__dirname, '../app');
+const config = require(path.resolve(__dirname, '../../lighthouse-core/config/default.js'));
+const lighthouseCategories = Object.keys(config.categories);
+
+let extensionPage;
+
+const assertAuditElements = async ({category, expected, selector}) => {
```
brendankenny: nit: any reason to have these two as arrow functions expressions? e.g. could just be
`async function assertAuditElements({category, expected, selector}) {`

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,130 @@
...
+          throw new Error(lighthouseResult.exceptionDetails.exception.description);
+        }
+
+        throw new Error(lighthouseResult.exceptionDetails.text);
+      }
+
+      extensionPage = (await browser.pages())
+        .find(page => page.url().includes('blob:chrome-extension://'));
+    } catch (err) {
+      assert.ok(false, err.message);
```
brendankenny: any reason for the try/catch in this case if it's always going to fail on this? (shouldn't mocha treat the exception as a test failure, with message `err.message` anyways?)

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    if (extensionPage) {
+      await assertReport();
+    }
+
+    if (browser) {
+      await browser.close();
```
brendankenny: if the test throws, do we want this pulled out into `after` so that the browser always gets closed afterwards?

lighthouse-extension/test/extension-test.js
wardpeet: no real reason :) whatever you guys prefer

lighthouse-extension/test/extension-test.js
wardpeet: we could also to the lighthouse run in before and do regular test cases 

```
describe('lighthouse extension' () => {
it ('should contain all categories', () => {

});

it('should contain all audits of category performance', () => {
})
```

lighthouse-extension/test/extension-test.js
wardpeet: filmstrip doesn't have a group either so I'll lose that one too. I could test for that separately.`

```
{id: 'screenshot-thumbnails', weight: 0},
{id: 'network-requests', weight: 0},
```

paulirish: good catch. IMO let's remove filmstrip from the selector above and use this `.length`.

lighthouse-extension/test/extension-test.js
paulirish: arrow fn style makes sense if its a 1liner or 1.5 liner. past that I agree it should get a function declaration.

lighthouse-extension/test/extension-test.js
paulirish: 👍 let's do that!

lighthouse-extension/test/extension-test.js
wardpeet: let me memorize this rule for the next time :p

lighthouse-core/scripts/run-mocha.sh
wardpeet: something as `"unit-extension": "cd lighthouse-extension && yarn test"`?

paulirish: bingo. so remove it from run-mocha. add `test-extension` to root manifest. add `test` to extension manifest. put `test-extension` in travis.yml

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,130 @@
...
+
+      const elementCount = await getAuditElementsCount({category, selector});
+
+      assert.equal(expected, elementCount,
+        `${category} does not have the correct amount of audits`);
+    }
+  });
+
+  it('should contain a filmstrip', async () => {
+    const filmstrip = await extensionPage.$('lh-filmstrip');
```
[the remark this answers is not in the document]
paulirish: i think this sorts it out:

```js
  it('should contain a filmstrip', async () => {
    const filmstrip = await extensionPage.$('.lh-filmstrip');
    assert.equal(true, !!filmstrip, `filmstrip is not available`);
  });
```

lighthouse-extension/test/extension-test.js
```diff
@@ -0,0 +1,138 @@
...
+
+    assert.equal(null, filmstrip,
+      `filmstrip is not available`);
+  });
+
+  it('should contain no failed audits', async () => {
+    const auditErrors = await extensionPage.$$('.lh-debug');
+    const failedAudits = auditErrors.filter(error =>
```
paulirish: heh. I ran into a similar API misunderstanding. I was doing `$$eval(selector, el => ...`) but its actually `elems` instead of just `el`. Whoopsie daisy.

-----------

try something like this out:


```js
  it('should not have any audit errors', async () => {
    function getDebugStrings(elems) {
      return elems.map(el => ({
        debugString: el.textContent,
        title: el.closest('.lh-audit').querySelector('.lh-score__title').textContent,
      }));
    }

    const auditErrors = await extensionPage.$$eval('.lh-debug', getDebugStrings);
    const errors = auditErrors.filter(item => item.debugString.includes('Audit error:'));
    assert.deepStrictEqual(errors, [], 'Audit errors found within the report');
  });
```

gives a nice purdy error: 
image

paulirish:
🎉 this is epic, @wardpeet. thx for making this happen. :D

PR #4676, opened 2018-03-02
PROPOSAL: new structure for metrics + lantern

patrickhulce:
UPDATE: closing this now that separate PRs to land have been opened #4764 #4766 

This PR is an example of a couple proposed changes we've been discussing that, if still acceptable to everyone, I'll make in separate PRs, but thought it'd be easier to see the whole thing come together.

I've demo'd the following changes with how it would look for `first-contentful-paint`

* A global `settings.throttling` object in the config controls whether the throttling method is `devtools`, `lantern`, or `provided` and the relevant settings. `devtools` and `provided` will make audits surface things based on what was observed (current method), while lantern will make audits surface things based on the lantern graph with specified settings (what we need for LR, default, etc)
* Runner passes the config to the audit (settings.throttling included)
* Each metric will have an audit and computed artifact
* Each metric computed artifact will either build the graphs for lantern and simulate, or find the metric in the trace depending on the global throttling setting
* Predictive-perf will go away and all the shared logic between metrics will become their own computed artifacts (like network-analysis in this example).
* Remove the `requiredNumberOfArtifacts` now that we have deep equality

### DIFF
--- a/lighthouse-core/config/default.js
+++ b/lighthouse-core/config/default.js
@@ -7,8 +7,24 @@
 
 /* eslint-disable max-len */
 
+const emulation = require('../lib/emulation');
+
 module.exports = {
-  settings: {},
+  settings: {
+    onlyAudits: ['first-contentful-paint'],
+    throttling: {
+      // method: 'devtools',
+      // requestLatency: emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.latency,
+      // downloadThroughput: emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.downloadThroughput,
+      // uploadThroughput: emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.uploadThroughput,
+
+      method: 'lantern',
+      rtt: 160,
+      throughput: 128 * 1000,
+
+      cpuSlowdownMultiplier: emulation.settings.CPU_THROTTLE_METRICS.rate,
+    },
+  },
   passes: [{
     passName: 'defaultPass',
     recordTrace: true,
@@ -72,6 +88,7 @@ module.exports = {
     ],
   }],
   audits: [
+    'first-contentful-paint',
     'is-on-https',
     'redirects-http',
     'service-worker',
@@ -264,6 +281,7 @@ module.exports = {
       name: 'Performance',
       description: 'These encapsulate your web app\'s current performance and opportunities to improve it.',
       audits: [
+        {id: 'first-contentful-paint', weight: 5, group: 'perf-metric'},
         {id: 'first-meaningful-paint', weight: 5, group: 'perf-metric'},
         {id: 'first-interactive', weight: 5, group: 'perf-metric'},
         {id: 'consistently-interactive', weight: 5, group: 'perf-metric'},
--- a/lighthouse-core/gather/computed/first-contentful-paint.js
+++ b/lighthouse-core/gather/computed/first-contentful-paint.js
@@ -1,0 +1,137 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const ComputedArtifact = require('./computed-artifact');
+const Node = require('../../lib/dependency-graph/node');
+const Simulator = require('../../lib/dependency-graph/simulator/simulator');
+
+const COEFFICIENTS = {
+  intercept: 1440,
+  optimistic: -1.75,
+  pessimistic: 2.73,
+};
+
+class FirstContentfulPaint extends ComputedArtifact {
+  get name() {
+    return 'FirstContentfulPaint';
+  }
+
+  /**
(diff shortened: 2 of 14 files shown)


lighthouse-core/gather/computed/first-contentful-paint.js
```diff
@@ -0,0 +1,137 @@
...
+      pessimisticGraph,
+    };
+  }
+
+  /**
+   * @param {{trace: Object, devtoolsLog: Object, throttling: Object}} data
+   * @return {Object}
+   */
+  async compute_(data, artifacts) {
+    if (data.throttling.method !== 'lantern') {
```
wardpeet: perhaps we should move this to the gatherer runner? this way we don't need to explicitly do if statements inside the compute_.

```
const throttlingMethod = data.throttling.method;
const computedGathererFunc = throttlingMethod.charAt(0).toUpperCase() + throttlingMethod.slice(1);
if (gatherer[computedGathererFunc]) {
  gatherer[computedGathererFunc]();
} else {
  gatherer.compute_();
}
```

lighthouse-core/gather/computed/first-contentful-paint.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    return FirstContentfulPaint.computeLantern(data, artifacts);
+  }
+}
+
+/**
+ * @typedef MetricResult
+ * @property {number} timing
+ * @property {number|undefined} timestamp
```
wardpeet: i know it's just an example but shouldn't this be `@property {number=} timestamp`

patrickhulce: nah it should be more like `@property {number} [timestamp]` [docs](http://usejsdoc.org/tags-property.html) but point taken yes will spend more time on the real typedef when trying to land 😛

lighthouse-core/gather/computed/first-contentful-paint.js
patrickhulce: Hm I feel like that introduces a decent amount of extra complexity to where the computed artifacts are coming from when they are already pretty hard to follow. what's the primary concern with having the ifs here in the compute?

patrickhulce:
@brendankenny @paulirish any other broader feedback on this structure? I'll need to get moving on all of this ASAP

lighthouse-core/gather/computed/first-contentful-paint.js
wardpeet: no real reason :) just thought it was better to not think about putting ifs everywhere but maybe it's better to make it explicit 😄

lighthouse-core/gather/computed/first-contentful-paint.js
```diff
@@ -0,0 +1,137 @@
...
+      if (node.record._resourceType !== WebInspector.resourceTypes.Script) return;
+      if (condition && !condition(node)) return;
+      scriptUrls.add(node.record.url);
+    });
+
+    return scriptUrls;
+  }
+
+  /**
+   * @param {!Node} dependencyGraph
```
paulirish: not relevant to the architecture, but we should document how we compute this estimated FCP.

lighthouse-core/gather/computed/first-contentful-paint.js
```diff
@@ -0,0 +1,137 @@
...
+      return node.hasRenderBlockingPriority() && node.initiatorType !== 'script';
+    });
+  }
+
+  /**
+   * @param {!Node} dependencyGraph
+   * @param {!TraceOfTabArtifact} traceOfTab
+   * @return {!Node}
+   */
+  static getPessimisticGraph(dependencyGraph, traceOfTab) {
```
paulirish: theres so much similarity between these two fns (optimistic & pessimistic). could we DRY it up?   dunno if adding `&& optionalPred(node)` to the end of both would be terrible or not.

lighthouse-core/gather/computed/first-contentful-paint.js
```diff
@@ -0,0 +1,137 @@
...
+  static async computeLantern(data, artifacts) {
+    const {trace, devtoolsLog} = data;
+    const graph = await artifacts.requestPageDependencyGraph({trace, devtoolsLog});
+    const traceOfTab = await artifacts.requestTraceOfTab(trace);
+    const networkAnalysis = await artifacts.requestNetworkAnalysis(devtoolsLog);
+
+    const optimisticGraph = FirstContentfulPaint.getOptimisticGraph(graph, traceOfTab);
+    const pessimisticGraph = FirstContentfulPaint.getPessimisticGraph(graph, traceOfTab);
+
+    const options = {...networkAnalysis, ...data.throttling};
```
paulirish: looks like data.throttling will override the rtt/throughput provided by networkAnalysis, right?
would you ever want just rtt from data.throttling and take throughput from networkAnalysis?

lighthouse-core/config/default.js
```diff
@@ -7,8 +7,24 @@
...
 
+const emulation = require('../lib/emulation');
+
 module.exports = {
-  settings: {},
+  settings: {
+    onlyAudits: ['first-contentful-paint'],
+    throttling: {
+      // method: 'devtools',
+      // requestLatency: emulation.settings.TYPICAL_MOBILE_THROTTLING_METRICS.latency,
```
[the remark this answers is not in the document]
patrickhulce: I named them differently specifically because they operate at different levels and the amount you'd want to apply to one *should* be very different from the amount you'd apply to the other to get the same experience :)

lighthouse-core/gather/computed/first-contentful-paint.js
patrickhulce: right, and not that I can think of. Network analysis artifact is meant to provide you a glimpse into what was *observed* while throttling is what you *want* it to be. 

In lantern's case we don't really care about what was observed, other than the additional latency observed per origin

PR #4677, opened 2018-03-02
report(description): update helpText for document.title audit

kaycebasques:
Since the audit occurs in both the a11y and SEO categories, the new helpText now mentions both. The helpText links to a new reference on d.g.c/web that discusses both the a11y and SEO aspects of the audit. The previously-linked-to aXe doc only mentioned the a11y side. But that aXe doc is itself mentioned in the new reference. I also updated the failureDescription while I was here, since we agreed in group chat that contractions are more in-tune with the Lighthouse voice.

#polish


rviscomi:
LGTM 👍 
blocked on https://github.com/google/WebFundamentals/pull/5855

PR #4703, opened 2018-03-07
misc: drop support of Node 6

### LINKED ISSUE #3742: Move to Node 8
Node 8 has now [hit LTS](https://github.com/nodejs/Release), so we should consider moving over to having that as the minimum supported version. 

We're of course already forward compatible, so anyone wanting e.g. perf improvements from Node 8 and 9 can already run Lighthouse there and get the benefits. However the language improvements [are significant](http://node.green/). Namely:
- `async`/`await` (maybe worth it for the clarity brought by this alone)
- `Object.entries` and `Object.values` (no more `Object.keys(obj).forEach(key => const value = obj[key];...` :)

along with a few other nice things like [object rest/spread](https://github.com/tc39/proposal-object-rest-spread), `String.padStart`, `String.padEnd`, the exponentiation operator, etc.

A mostly organic transition to async/await would probably be best, like we did when we moved to Node v6 and only really added destructuring where it made sense (of course) and where we had some other reason to touch code than just updating to new syntax. We have some tricky promise tricks we use at times and a mass migration could easily miss subtleties (though hopefully our tests won't). And promises and async/await play together nicely, so mixing them for the near future won't be much of an issue.

I'm personally very eager to move over, but it is worth noting that [last time](https://github.com/GoogleChrome/lighthouse/issues/449) we waited until January after the October LTS switch to move up versions, so prudence might have us wait.

paulirish:
Moving to Node 8 as lowest supported version.
This is a breaking change, and we've already branched the `2.x` branch, as master is now our 3.0. ref #4333 

(Breaking  ecc3d06 out of #4640 for clarity)

fixes #3742

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,11 +9,11 @@ branches:
 environment:
   fast_finish: true
   matrix:
-    - nodejs_version: "6"
+    - nodejs_version: "8"
       platform: x86
-    # node 7 is skipped, as appveyor only allows 1 concurrent job
+    # other Node versions are skipped, as appveyor only allows 1 concurrent job
     # and we want appveyor finishing ASAP. see #2382
-    # - nodejs_version: "7"
+    # - nodejs_version: "9"
     #   platform: x86
 
 build: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ branches:
   - master
 matrix:
   include:
-    - node_js: "6.9.1"
     - node_js: "8"
-      if: head_branch IS blank AND branch = master
     - node_js: "9"
       if: head_branch IS blank AND branch = master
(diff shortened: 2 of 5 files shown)


wardpeet:
@paulirish do we have a policy on new code style? Write everything with async/await from now on? what to do with old gatherers/audits/core, ...

paulirish:
@wardpeet 

> do we have a policy on new code style? Write everything with async/await from now on? what to do with old gatherers/audits/core, ...

I think what @brendankenny spelled out here is perfect: https://github.com/GoogleChrome/lighthouse/issues/3742#issue-271079651

"A mostly organic transition to async/await would probably be best". 👍 👍 

We'll refactor old code as we have reason to work on the files.

PR #4710, opened 2018-03-07
core(start_url): fix start_url audit while offline

### LINKED ISSUE #2576: 'Unable to retrieve manifest' when manifest file is available and valid
https://pwa.banggood.com/manifest.json
Audit error: Required StartUrl gatherer encountered an error: Unable to retrieve manifest at https://pwa.banggood.com/manifest.json.

Lighthouse shows this error, but this json file is avaliable and it is a valid json file. We can't fix this error ,so help me fix it.

wardpeet:
Tests are broken but I want some feedback on this one.

is this a good approach? It seems to work on some urls I've tested but should test more of them

It fixes #2576

### DIFF
--- a/lighthouse-cli/test/fixtures/manifest.json
+++ b/lighthouse-cli/test/fixtures/manifest.json
@@ -1,0 +1,38 @@
+{
+  "name": "App Install",
+  "short_name": "App Install",
+  "icons": [
+    {
+      "src": "launcher-icon-0-75x.png",
+      "sizes": "36x36",
+      "type": "image/png"
+    },
+    {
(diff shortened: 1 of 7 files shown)


wardpeet:
Some test urls that have failed before
- https://www.mlamannadev.com/
- https://silent-things.surge.sh/
- https://pwa.banggood.com/
- https://preact-cli-workbox-buggy.surge.sh/

brendankenny:
@wardpeet: not sure if @paulirish mentioned this to you somewhere else, but we'll have to stick to promises in this file for now because we want it in the 2.x branch, which doesn't drop Node 6

PR #4690, opened 2018-03-05
core: refactor scoring to be numeric only. introduce scoreDisplayMode

paulirish:
1. Ensure all audits return a numeric `score` (0-1), no boolean nonsense. 
1. `scoreDisplayMode` forces how it's displayed (renamed from `scoringMode`)

PR #4711, opened 2018-03-07
core(lhr): make reportCategories shallow; move audit scores to AuditResultJSON

### LINKED ISSUE #1999: Reduce size of JSON results: remove duplication

paulirish:
### Changes

* `reportCategories` no longer contains real auditResults (aka _de-dupe all audit results_ from #4333)
   * The categories still keep their `.score` prop. Eg: `reportCategories.performance.score`
  * Each audit's score moves from `reportCategories.performance['audit-name']` (which now just holds `weight` and `group`) over to `audits['audit-name']`
* Remove `LHR.score`, the overallScore

---------------

This is a chain of PRs. Landing order is... 1st: `newdetails` (#4616). 2nd: `shallowCategories` (#4711). 3rd: `scoring2.0` (#4690)

ref #4614

fixes #1999


lighthouse-core/runner.js
```diff
@@ -93,9 +93,11 @@ class Runner {
       const resultsById = {};
       for (const audit of runResults.auditResults) resultsById[audit.name] = audit;
 
-      let report;
+      let categories;
       if (opts.config.categories) {
-        report = Runner._scoreAndCategorize(opts, resultsById);
+        Runner._scoreAndCategorize(opts, resultsById);
+        // COMPAT: after dropping Node 6, upgrade to use Object.values
```
patrickhulce: does this mean we can do it now? :)

lighthouse-core/scoring.js
```diff
@@ -13,30 +13,33 @@ class ReportScoring {
...
   static scoreAllCategories(config, resultsByAuditId) {
-    const categories = Object.keys(config.categories).map(categoryId => {
+    Object.keys(config.categories).forEach(categoryId => {
       const category = config.categories[categoryId];
       category.id = categoryId;
 
-      const audits = category.audits.map(audit => {
+      category.audits.forEach(audit => {
         const result = resultsByAuditId[audit.id];
         // Cast to number to catch `null` and undefined when audits error
```
patrickhulce: should all of these line just be moved into `audit.js` then to keep all the score mapping in the same place?

patrickhulce: more a comment for #4690 I suppose

brendankenny:
smokehouse failures?

lighthouse-core/runner.js
brendankenny: do it!

lighthouse-core/scoring.js
paulirish: Aye. It actually already happens in that PR: https://github.com/GoogleChrome/lighthouse/pull/4690/files#diff-488f5da33160a651526463bdb951e5deL31

paulirish:
> As discussed, we'll want to smoosh the audit results into a copy of the LHR so the canonical one still exists untouched.

Done. added a test.

> something for the future: as we add more typechecking, we may want to split AuditJSON to a version with result (only used inside the report renderer) and a version without (used everywhere else) so that we can more easily reason about which we have at some point in time and so we don’t have to put in a bunch of checks to keep tsc happy :)

YESSSSSSS. So want this.

Here was my typechecking spike, validating the outgoing LHR against some real types. Works! https://github.com/GoogleChrome/lighthouse/compare/95e19a56503c702a5077c62dda1112cd39c4d705...typechecking


---------

This and #4690 are updated and ready

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -155,24 +157,39 @@ class ReportRenderer {
     perfCategoryRenderer.setTemplateContext(this._templateContext);
 
     const categories = reportSection.appendChild(this._dom.createElement('div', 'lh-categories'));
+
+    ReportRenderer.smooshAuditResultsIntoCategories(report.audits, report.reportCategories);
```
brendankenny: does this make more sense up in `renderReport` before passing into `_renderReport`? Then internally it'll always be the smooshed version and we won't have to remember not to use `reportCategories` above this line

lighthouse-core/report/v2/renderer/report-renderer.js
```diff
@@ -181,6 +198,16 @@ if (typeof module !== 'undefined' && module.exports) {
...
 
+/**
+ * @typedef {{
+ *     id: string,
+ *     weight: number,
+ *     group: (string|undefined),
+ *     result: (ReportRenderer.AuditResultJSON|undefined)
+ * }}
+ */
+ReportRenderer.AuditJSON; // eslint-disable-line no-unused-expressions
```
brendankenny: defined below (merge conflict, I think)

lighthouse-core/runner.js
```diff
@@ -108,8 +109,7 @@ class Runner {
         audits: resultsById,
         artifacts: runResults.artifacts,
         runtimeConfig: Runner.getRuntimeConfig(opts.flags),
-        score: report ? report.score : 0,
-        reportCategories: report ? report.categories : [],
+        reportCategories: categories,
```
brendankenny: before this defaulted to an empty array. Will it cause problems anywhere as `undefined` (and removed from the JSON) instead?

lighthouse-core/scoring.js
```diff
@@ -13,35 +13,49 @@ class ReportScoring {
...
    * Returns the report JSON object with computed scores.
-   * @param {{categories: !Object<string, {id: string|undefined, weight: number|undefined, audits: !Array<{id: string, weight: number|undefined}>}>}} config
+   * @param {{categories: !Object<string, {id: string|undefined, weight: number|undefined, score: number|undefined, audits: !Array<{id: string, weight: number|undefined}>}>}} config
    * @param {!Object<string, {score: ?number|boolean|undefined, notApplicable: boolean, informative: boolean}>} resultsByAuditId
    * @return {{score: number, categories: !Array<{audits: !Array<{score: number, result: !Object}>}>}}
    */
   static scoreAllCategories(config, resultsByAuditId) {
-    const categories = Object.keys(config.categories).map(categoryId => {
-      const category = config.categories[categoryId];
+    for (const [categoryId, category] of Object.entries(config.categories)) {
```
brendankenny: 🎉

paulirish:
merge when green.

PR #1407, opened 2017-01-05
fix: remove extending es6 class from url-shim to support es5 transpilation

patrickhulce:

PR #4712, opened 2018-03-08
core: make url-shim extend native URL module; add type checking

### LINKED ISSUE #1186: Use native Node URL parsing after Node 7.2.1 errors resolved
I'm getting URL parsing errors in `audits/dobetterweb/uses-http2.js` that weren't present before https://github.com/GoogleChrome/lighthouse/pull/997.

When running:

    npm run fast -- --output=html --output-path=results.html http://www.univision.com

Think it's failing on `http://5321212.fls.doubleclick.net/activityi;src=5321212;type=unvsn_un;cat=unvsn_uv;ord=7762287885264.98?`

screen shot 2016-12-20 at 2 52 39 pm

brendankenny:
adds type checking to URLShim. removes `whatwg-url` now that Node 8 is required (fixes #1186)

This moves us back to extending `URL`, which was removed in #1407. If anyone can remember *why* we did that  at the timeplease let me know, but for now CLI and extension seem to work fine with this change.


brendankenny:
diff is much better at https://github.com/GoogleChrome/lighthouse/pull/4712/files?w=1. PR is primarily indentation into a class body :)

brendankenny:
that's some nice archaeology!

PR #4718, opened 2018-03-09
core(driver): add timeout to getRequestContent

### LINKED ISSUE #4258: Lighthouse hangs indefinitely on Retrieving Scripts
First of all, thank you for an excellent tool.

I am using the chrome extension of lighthouse for sometime. And being benefited by the insights that it gives on page analysis. But all of a sudden it started hanging at the "Retrieving Scripts" stage and never proceeds further. I have stopped all the other extensions on chrome and tried again. But still the same issue.

The site is publicly available: https://itsroo.com
Attached the complete log file here
[itsroo_com.log](https://github.com/GoogleChrome/lighthouse/files/1629963/itsroo_com.log)

paulirish:
fixes #4258

I was able to repro on all the URLs mentioned in the ticket: 

* https://openvegemap.netlib.re/#zoom=16&lat=48.85659&lon=2.35149
* https://dreamteamtrips.com/
* https://itsroo.com/
* https://anketa.alfabank.ru/land/pil-refin/

`Network.getResponseBody` is taking >5s to respond when provided particular requestIds. But it's quite odd. If i have devtools open at the same time, the devtools frontend can issue the exact same protocol method with the same requestId and it has no problem. (and apparently devtools and CLI also have no problem)

image

So this is particular to the extension.. sort of.

**Update:** the root bug appears to be an encoding issue in how the extension handles `chrome.debugger` payloads. Below is a bunch of details that supports that argument:

<details>

here's the 4 scripts that will reliably cause a problem: 

* https://openvegemap.netlib.re/dist/bundle.js 
* https://smartsupp-widget-161959.c.cdn77.org/build/smartchat-2.1.16.min.js
* https://itsroo.com/bower_components/firebase/firebase-firestore.js
* https://anketa.alfabank.ru/land/pil-refin/assets/index.592fac88f8cec24d8b0e.js

I believe these files all have some characters in common that are mishandled somewhere in encoding.

I ran this across these 4 files: 

```js
var codepoints = []; 
for (const char of filesource){ 
	var code = char.codePointAt(0); 
	if (code > 30000) codepoints.push(code) 
}
```

And they all have in common some fairly high codepoints:

```
bundle.js 
[ 127828, 127860, 127860, 127861, 127864, 127866, 128722, 129366, 65306, 65306, 65374, 65374, 65534 ]

smartchat
[ 65311, 65311, 65311, 65311, 65520, 65521, 65522, 65523, 65524, 65525, 65526, 65527, 65528, 65529, 65530, 65531, 65532, 65533, 65534, 65535 ]

firebase firestore
[ 65535 ]

alfabank
[ 65533, 65533, 65533, 65533, 65534 ]
```

They all have in common 65534 or 65535. 

**And...** check this out. Run this in any DevTools console:

```js
String.fromCodePoint(65535)
```

Now take that resulting string, copy/paste it back and try to evaluate it. DevTools won't. :)
</details>

----

So, since this appears to be rather an involved bug. My easy workaround is to add a 2s timeout. 

I did observe a 300ms successful response time for this method on a fairly big file so 2000ms seems an OK maximum.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -717,10 +717,20 @@ class Driver {
    * @return {string}
    */
   getRequestContent(requestId) {
-    return this.sendCommand('Network.getResponseBody', {
-      requestId,
-    // Ignoring result.base64Encoded, which indicates if body is already encoded
-    }).then(result => result.body);
+    return new Promise((resolve, reject) => {
+      // If this takes more than 2s, reject the Promise.
+      const err = new Error('Fetching resource content has exceeded the allotted time of 2s');
+      const asyncTimeout = setTimeout((_ => reject(err)), 2000);
+
+      this.sendCommand('Network.getResponseBody', {requestId}).then(result => {
+        clearTimeout(asyncTimeout);
+        // Ignoring result.base64Encoded, which indicates if body is already encoded
+        resolve(result.body);
+      }).catch(e => {
+        clearTimeout(asyncTimeout);
+        reject(e);
+      });
+    });
   }
 
   /**
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -715,7 +715,7 @@ class Driver {
   /**
    * Return the body of the response with the given ID.
    * @param {string} requestId
-   * @param {?number} timeout
+   * @param {number|undefined} timeout
    * @return {string}
    */
   getRequestContent(requestId, timeout = 1000) {
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -9,6 +9,7 @@
 const NetworkRecorder = require('../lib/network-recorder');
 const emulation = require('../lib/emulation');
 const Element = require('../lib/element');
+const LHError = require('../lib/errors');
 const EventEmitter = require('events').EventEmitter;
 const URL = require('../lib/url-shim');
 const TraceParser = require('../lib/traces/trace-parser');
@@ -718,9 +719,9 @@ class Driver {
    */
   getRequestContent(requestId) {
     return new Promise((resolve, reject) => {
-      // If this takes more than 2s, reject the Promise.
-      const err = new Error('Fetching resource content has exceeded the allotted time of 2s');
-      const asyncTimeout = setTimeout((_ => reject(err)), 2000);
+      // If this takes more than 1s, reject the Promise.
+      const err = new LHError(LHError.errors.REQUEST_CONTENT_TIMEOUT);
+      const asyncTimeout = setTimeout((_ => reject(err)), 1000);
 
       this.sendCommand('Network.getResponseBody', {requestId}).then(result => {
         clearTimeout(asyncTimeout);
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -717,10 +717,20 @@ class Driver {
...
    */
   getRequestContent(requestId) {
-    return this.sendCommand('Network.getResponseBody', {
-      requestId,
-    // Ignoring result.base64Encoded, which indicates if body is already encoded
-    }).then(result => result.body);
+    return new Promise((resolve, reject) => {
+      // If this takes more than 2s, reject the Promise.
+      const err = new Error('Fetching resource content has exceeded the allotted time of 2s');
+      const asyncTimeout = setTimeout((_ => reject(err)), 2000);
```
patrickhulce: I'd be cool going down to 1s too to keep things moving

lighthouse-core/gather/driver.js
```diff
@@ -717,10 +717,20 @@ class Driver {
    * @return {string}
    */
   getRequestContent(requestId) {
-    return this.sendCommand('Network.getResponseBody', {
-      requestId,
-    // Ignoring result.base64Encoded, which indicates if body is already encoded
-    }).then(result => result.body);
+    return new Promise((resolve, reject) => {
+      // If this takes more than 2s, reject the Promise.
+      const err = new Error('Fetching resource content has exceeded the allotted time of 2s');
```
patrickhulce: should we try to give this thing an error code and be consistent about timeout stuff moving forward?

lighthouse-core/gather/driver.js
```diff
@@ -717,10 +717,20 @@ class Driver {
    * @return {string}
    */
   getRequestContent(requestId) {
-    return this.sendCommand('Network.getResponseBody', {
-      requestId,
-    // Ignoring result.base64Encoded, which indicates if body is already encoded
-    }).then(result => result.body);
+    return new Promise((resolve, reject) => {
+      // If this takes more than 2s, reject the Promise.
```
brendankenny: maybe add a note with link back to your investigation for why this method in particular can be problematic?

lighthouse-core/test/gather/driver-test.js
```diff
@@ -138,6 +141,14 @@ describe('Browser Driver', () => {
     });
   });
 
+  it('throws if getRequestContent takes too long', () => {
+    return driverStub.getRequestContent(0, MAX_WAIT_FOR_PROTOCOL).then(_ => {
+      assert.ok(false, 'long-running getRequestContent supposed to reject');
+    }).catch(e => {
+      assert.equal(e.code, 'REQUEST_CONTENT_TIMEOUT');
```
brendankenny: will the above assert message get eaten by this comparison? could also do the two function `then()`
```js
return driverStub.getRequestContent(0).then(value => {
  assert.ok(false, 'long-running getRequestContent supposed to reject');
}, e => {
  assert.equal(e.code, 'REQUEST_CONTENT_TIMEOUT');
});
```

lighthouse-core/gather/driver.js
```diff
@@ -714,13 +715,25 @@ class Driver {
   /**
    * Return the body of the response with the given ID.
    * @param {string} requestId
+   * @param {?number} timeout
```
brendankenny: `{number=}`

PR #4723, opened 2018-03-09
core(is-https): use network record's securityState

### LINKED ISSUE #4722: mixed-content/https audits missing a insecure requests
https://www.paulirish.com/2006/think-you-can-recycle-the-single-serving-coffee-packets-think-again/ has a `<img src=http://....` insecure image request.

But I don't see anything in the report:

image

It is a little odd the score of the particular audit is 92, but it doesn't list any problematic requests.

cc @christhompson do you know what's going on here?

And fwiw I do have 'Upgrade Insecure Requests' enabled on the site, but i'm not sure that's related to the issue so far.

paulirish:
The `mixed-content` audit uses it. 

By adopting it in `is-https`, we can correctly include some `http` resources that were upgraded via "Upgrade Insecure Requests" but failed regardless.

repro: `lighthouse --mixed-content https://www.paulirish.com/2006/think-you-can-recycle-the-single-serving-coffee-packets-think-again/`

Before: 
image

After: 
image

fixes #4722 i think?

### DIFF
--- a/lighthouse-core/audits/is-on-https.js
+++ b/lighthouse-core/audits/is-on-https.js
@@ -31,8 +31,10 @@ class HTTPS extends Audit {
 
   /**
    * Checks whether the resource was securely loaded.
-   * We special-case data: URLs, as they inherit the security state of their
-   * referring document url. Localhost is also safelisted.
+   * While `Security.securityStateChanged` is broken on Android, the networkResponse.securityState
+   * property is functional, so we use that instead of checking the URL scheme.
+   * We special-case data: URLs, as they inherit the security state of their referring document url.
+   * Localhost is also safelisted.
    *
    * @param {{scheme: string, domain: string, protocol: string, securityState: function}} record
    * @return {boolean}
--- a/lighthouse-core/audits/is-on-https.js
+++ b/lighthouse-core/audits/is-on-https.js
@@ -31,13 +31,18 @@ class HTTPS extends Audit {
   }
 
   /**
-   * @param {{scheme: string, domain: string}} record
+   * Checks whether the resource was securely loaded.
+   * We special-case data: URLs, as they inherit the security state of their
+   * referring document url, and so are trivially "upgradeable" for mixed-content purposes.
+   *
+   * @param {{scheme: string, domain: string, protocol: string, securityState: function}} record
    * @return {boolean}
    */
   static isSecureRecord(record) {
-    return SECURE_SCHEMES.includes(record.scheme) ||
-           SECURE_SCHEMES.includes(record.protocol) ||
-           SECURE_DOMAINS.includes(record.domain);
+    return record.securityState() === 'secure' ||
+      SECURE_SCHEMES.includes(record.scheme) ||
+      SECURE_SCHEMES.includes(record.protocol) ||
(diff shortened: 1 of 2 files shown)


paulirish:
On second thought, I'm not sure how to handle the semantics here.

These _upgraded_ requests failed for various certificate reasons: 
image

Despite the failures the lock icon is still green: 
image

So if `is-https` is meant to match the green lock, then this PR deviates from that.

But it does flag a few problematic resources that are otherwise a blind spot.

I'm not sure what makes the most sense. @christhompson do you have an opinion?

lighthouse-core/audits/is-on-https.js
```diff
@@ -31,13 +30,16 @@ class HTTPS extends Audit {
...
+   * referring document url. Localhost is also safelisted.
+   *
+   * @param {{scheme: string, domain: string, protocol: string, securityState: function}} record
    * @return {boolean}
    */
   static isSecureRecord(record) {
-    return SECURE_SCHEMES.includes(record.scheme) ||
-           SECURE_SCHEMES.includes(record.protocol) ||
-           SECURE_DOMAINS.includes(record.domain);
+    return record.securityState() === 'secure' || record.protocol === 'data' ||
```
patrickhulce: does this work on android? IIRC the original reason for not trusting all the security info from devtools and following back to URLs was all the security fails on android devices

paulirish: thanks for calling this out. after double-checking, we are good for android. 

image

turns out its the more global `Security.securityStateChanged`  that's broken on android, but this secState within network is just fine.

paulirish: @patrickhulce now to test a page that embed 'about:blank' to see if securityState is set correctly....

uno momento.

christhompson:
Sorry, was OOO and playing catch-up now.

> So if is-https is meant to match the green lock, then this PR deviates from that.
> But it does flag a few problematic resources that are otherwise a blind spot.
> I'm not sure what makes the most sense. @christhompson do you have an opinion?

Yeah, I think discouraging Upgrade-Insecure-Requests by counting failed upgrade requests isn't quite right either. From my comment on #4722 (and looking at [paul's comment above][1]), I think the problem is that `insecureRecords` is defined to be any record that doesn't have `securityState == 'secure'`, but HTTPS requests with failed cert validation appear to have `securityState == 'secure' && failed = true`. These failed subresource requests should not cause mixed content errors (and would be non-bypassable errors in Chrome).

I think it would be better to try to fix the false-insecures in the Mixed Content audit.

(Using `securityState` may make more sense for main frames only rather than for subframes or subresources, since for subframes/resources they will just get blocked on a security error rather than allowing other security states.)

[1]: https://github.com/GoogleChrome/lighthouse/pull/4723#issuecomment-371938541

brendankenny:
> Mar 14, 2018  
> uno momento

:)

> So if is-https is meant to match the green lock, then this PR deviates from that. But it does flag a few problematic resources that are otherwise a blind spot.

I think we do want to match the green lock. These cases *are* caught by the Best Practices `errors-in-console` audit, though they are kind of far down in there and disassociated with the https check.

I'm not sure if we want to do more than that, but it's been long enough that I'm closing for now :)

PR #4737, opened 2018-03-12
tests(closure): avoid failing closure compilation

### LINKED ISSUE #4735: compile-devtools is failing
The `yarn compile-devtools` script has been failing for the past few days.

Here's two recent builds from master:

PASSING: https://travis-ci.org/GoogleChrome/lighthouse/jobs/351077847#L828
FAILING: https://travis-ci.org/GoogleChrome/lighthouse/jobs/351130910#L826

Both appear to check out the exact same hash of -frontend and -protocol. Which indicates it's only Lighthouse that's changing.

Here's the error:

```txt
devtools frontend compilation output:
/tmp/tmp6f4UuQ/front_end/bindings/FileUtils.js:146: WARNING - Property result never defined on EventTarget
    const buffer = event.target.result;
                                ^^^^^^
```

I'm not really sure why this started cropping up after bc9efba9b7b5c40a2886d6f1d366e272f420afb2, but a bisect confirmed that this is the commit where it started.

paulirish:
This is most definitely a closure compiler bug. But this reordering fixes #4735

### DIFF
--- a/lighthouse-core/report/v2/renderer/report-renderer.js
+++ b/lighthouse-core/report/v2/renderer/report-renderer.js
@@ -183,17 +183,6 @@ if (typeof module !== 'undefined' && module.exports) {
 
 /**
  * @typedef {{
- *     id: string,
- *     score: number,
- *     weight: number,
- *     group: (string|undefined),
- *     result: (ReportRenderer.AuditResultJSON|undefined)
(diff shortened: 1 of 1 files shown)


brendankenny:
(waiting on CI)

PR #4738, opened 2018-03-12
core: add type checking to connection

brendankenny:
no significant changes, mostly typings (e.g. `runJsonCommand` returns an object, not a string).

The biggest shifts were just e.g. assuring the compiler we wouldn't emit events after the event emitter was disposed.

For the future, there are a lot of opportunities for better typing on these methods that take a string but return a bunch of different things via TS's conditional types (or plain overloading with string literal arguments)

### DIFF
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -3,7 +3,6 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
-// @ts-nocheck
 'use strict';
 
 const Connection = require('./connection.js');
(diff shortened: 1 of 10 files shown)


lighthouse-core/scoring.js
```diff
@@ -28,7 +28,7 @@ class ReportScoring {
   /**
    * Returns the report JSON object with computed scores.
    * @param {{categories: !Object<string, {id: string|undefined, weight: number|undefined, audits: !Array<{id: string, weight: number|undefined}>}>}} config
-   * @param {!Object<{score: ?number|boolean|undefined}>} resultsByAuditId
+   * @param {!Object<string, {score: ?number|boolean|undefined, notApplicable: boolean, informative: boolean}>} resultsByAuditId
```
brendankenny: not sure why this wasn't failing before. Possibly from the update to ts 2.7.2?

lighthouse-core/gather/connections/cri.js
```diff
@@ -25,34 +24,37 @@ class CriConnection extends Connection {
...
-          const firstTab = tabs[0];
-          if (!Array.isArray(tabs) || !firstTab) {
+          if (!Array.isArray(tabs) || tabs.length === 0) {
             return Promise.reject(new Error('Cannot create new tab, and no tabs already open.'));
           }
+          const firstTab = tabs[0];
           // first, we activate it to a foreground tab, then we connect
           return this._runJsonCommand(`activate/${firstTab.id}`)
-              .then(_ => this._connectToSocket(firstTab));
+              .then(() => this._connectToSocket(firstTab));
```
patrickhulce: 😄

PR #4748, opened 2018-03-13
tests(smokehouse): run smoketests in parallel

paulirish:
From 149s (2.5 minutes!) down to 45s!

Mostly a matter of running them in parallel.

### Changes
* Replaced those gnarly `run-test.sh` scripts with a flexible JS-based smoketest runner. Runs each `smokehouse.js` invocation in its own child-process.
* Moved a run of `dbw-expectations` into `offline`, as it was testing the exact same URL, just asserting different aspects out the output.
* `yarn smoke` will run all, but you can do `yarn smoke byte` and `yarn smoke byte ttci` to run individual smokehouse expectations
* Split the PWA smoketests into two files (`pwa` and `pwa2`), as they were some of the slowest smoketests we did.
* `yarn smokehouse` was removed. it's effectively replaced with `yarn smoke pwa && yarn smoke pwa2`
* Restore the `shop.polymer-project.org` smoketest. 
  * Add a workaround for #4898, since it reproduces in this test.


paulirish:
> I am concerned about the debuggability if something does go wrong and the output is all interleaved together :)

Aye. I added an extra line of output where "checking <url>" is emitted right before the assertions are checked. IMO i think that lowers the probability of confusing results a LOT. But it's still possible and to fix that means buffering output which means any failures you won't see until ALL smoketests are done. Also requires lots more work. :)

> also, whats the feasibility of getting rid of all the run-tests.sh files for just one :)

And going to JS? Mmmmmmm that's also an option.

paulirish:
@patrickhulce (no rush, but..) thoughts? right now it looks like this work will go into ❄️ _**chillin**_ state.

brendankenny:
> And going to JS? Mmmmmmm that's also an option.

shelljs is somewhat active again so could work ok for this (and get us easier windows support)

paulirish:
TBH it could move to JS without shelljs, even. really not too terrible.

wardpeet:
100% js would be nice.

For now we could still buffer the output inside a variable and output it where needed if we are afraid of debugging.

paulirish:
> For now we could still buffer the output inside a variable and output it where needed if we are afraid of debugging.

Yeah.. Now with basic pattern of https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/scripts/update-report-fixtures.js available, we can much more easily do this in JS. And yeah we can avoid interleaving much easier. 👍

package.json
```diff
@@ -23,7 +23,11 @@
     "build-viewer": "cd ./lighthouse-viewer && yarn build",
     "clean": "rimraf *.report.html *.report.dom.html *.report.json *.screenshots.html *.screenshots.json *.devtoolslog.json *.trace.json || true",
     "lint": "[ \"$CI\" = true ] && eslint --quiet -f codeframe . || eslint .",
-    "smoke": "bash lighthouse-cli/test/smokehouse/run-all-tests.sh",
+
+    "smoke": "node lighthouse-cli/test/smokehouse/run-smoke.js",
+    "coverage": "istanbul cover -x \"**/third_party/**\" _mocha -- $(find */test -name '*-test.js') --reporter progress --report lcovonly",
```
brendankenny: rebase problem?

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -288,8 +285,9 @@ const expectations = require(resolveLocalOrCwd(cli['expectations-path']));
 let passingCount = 0;
 let failingCount = 0;
 expectations.forEach(expected => {
-  console.log(`Checking '${expected.initialUrl}'...`);
+  console.log(`Running '${expected.initialUrl}'...`);
   const results = runLighthouse(expected.initialUrl, configPath);
+  console.log(`Comparing results of '${expected.initialUrl}':`);
```
brendankenny: maybe ``console.log(`Auditing findings of '${expected.initialUrl}':`);``?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,109 @@
...
+  'ttci': {
+    expectations: 'tricky-ttci/expectations.js',
+    config: 'lighthouse-core/config/default-config.js',
+  },
+  'pwa': {
+    expectations: smokehouseDir + 'pwa-expectations.js',
+    config: smokehouseDir + 'pwa-config.js',
+  },
+};
+
```
brendankenny: extra line

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,109 @@
...
+
+function displayOutput(cp) {
```
brendankenny: Add jsdoc?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,109 @@
...
+    console.log(cp.message);
+  }
+  process.stdout.write(cp.stdout);
+  process.stderr.write(cp.stderr);
+  console.log(`End of results for: ${cp.id}`);
+  console.log('\n\n')
+}
+
+/**
+ * Update the report artifacts
```
brendankenny: old comment?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,109 @@
...
+ */
+async function run() {
+  // start webservers
+  server.listen(10200, 'localhost');
+  serverForOffline.listen(10503, 'localhost');
+
+  const cmdPromises = [];
+  for (const [id, {expectations, config}] of Object.entries(smokes)) {
```
brendankenny: add way to specify how many to run at the same time? But it sounds like you may be planning that already

lighthouse-cli/test/smokehouse/smokehouse.js
patrickhulce: 😱

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  const parallelSmokes = SMOKETESTS.filter(t => !t.perfSensitive);
+  const serialSmokes = SMOKETESTS.filter(t => t.perfSensitive);
+
+  const smokeResults = await runSmokehouse(parallelSmokes);
+  const serialSmokeResults = await runSmokehouse(serialSmokes);
+  smokeResults.push(...serialSmokeResults);
+
+  await new Promise(res => server.close(res));
```
patrickhulce: nit: this really throws me just because `res` is the canonical express `response` shorthand and then later you use it for result, `resolve` or `done`? :)

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+  await new Promise(res => serverForOffline.close(res));
+
+  const failingTests = smokeResults.filter(res => !!res.error);
```
patrickhulce: nit: s/res/anything else :)

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -317,8 +316,10 @@ const expectations = require(resolveLocalOrCwd(cli['expectations-path']));
 let passingCount = 0;
 let failingCount = 0;
 expectations.forEach(expected => {
-  console.log(`Checking '${expected.initialUrl}'...`);
+  console.log(`Doing a run of '${expected.initialUrl}'...`);
   const results = runLighthouse(expected.initialUrl, configPath, cli.debug);
+
+  console.log(`Auditing findings of '${expected.initialUrl}':`);
```
patrickhulce: 🤦‍♂️

package.json
```diff
@@ -23,7 +23,9 @@
     "build-viewer": "cd ./lighthouse-viewer && yarn build",
     "clean": "rimraf *.report.html *.report.dom.html *.report.json *.screenshots.html *.screenshots.json *.devtoolslog.json *.trace.json || true",
     "lint": "[ \"$CI\" = true ] && eslint --quiet -f codeframe . || eslint .",
-    "smoke": "bash lighthouse-cli/test/smokehouse/run-all-tests.sh",
+
```
brendankenny: these newlines won't survive a `yarn add`, fwiw

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+  perfSensitive: true,
+}, {
+  id: 'ttci',
+  expectations: 'tricky-ttci/expectations.js',
+  config: 'lighthouse-core/config/default-config.js',
+}];
+
+/**
+ * Display smokehouse output from child process
+ * @param {{id: string, process?: NodeJS.Process, code?: number}} cp
```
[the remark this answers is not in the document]
brendankenny: it would also be something like `@param {{id: string, process: NodeJS.Process} || {id: string, error: Error & {stdout : NodeJS.WriteStream, stderr: NodeJS.WriteStream}}} result`

but that might be getting a bit silly, especially with no type checking on this file :)

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+    process.stdout.write(result.process.stdout);
+    process.stderr.write(result.process.stderr);
+  }
+  console.log(`${purpleify(result.id)} smoketest complete. \n`);
+}
+
+/**
+ * Run smokehouse in child processes for selected smoketests
+ * Display output from each as soon as they finish, but resolve function when ALL are complete
+ * @param {*} smokes
```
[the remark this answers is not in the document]
brendankenny: er, and `perfSensitive?: boolean`

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+  let smokes = [];
+  const argv = process.argv.slice(2);
+  if (argv.length === 0) {
+    smokes = SMOKETESTS;
+    console.log('Running ALL smoketests. Equivalent to:');
+    console.log(`    ${log.dim}yarn smoke ${smokes.map(t => t.id).join(' ')}${log.reset}\n`);
+  } else {
+    smokes = SMOKETESTS.filter(test => argv.includes(test.id));
+    console.log(`Running ONLY smoketests for: ${smokes.map(t => t.id).join(' ')}\n`);
+  }
```
brendankenny: maybe add an error case that prints `smokes` ids?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,152 @@
...
+
+  const parallelSmokes = SMOKETESTS.filter(t => !t.perfSensitive);
+  const serialSmokes = SMOKETESTS.filter(t => t.perfSensitive);
```
brendankenny: maybe I'm missing it, but aren't these still run in parallel?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,191 @@
...
+    cmdPromises.push(p);
+  }
+
+  return Promise.all(cmdPromises);
+}
+
+/**
+ *
+ * @param {string[]} argv
+ * @return {Map<string, Array<SmoketestDfn>>}
```
brendankenny: `string|undefined`? or give a default batch name/symbol so it doesn't fall back to undefined

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  const unmatchedIds = argv.filter(requestedId => !SMOKETESTS.map(t => t.id).includes(requestedId));
+  if (unmatchedIds.length) {
+    console.log(log.redify(`Smoketests not found for: ${unmatchedIds.join(' ')}`));
+    console.log(usage);
+  }
+
+  // Split into serial batches that will run their tests concurrently
+  const batches = smokes.reduce((map, test) => {
```
brendankenny: is this still running the two perf batches in parallel?

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Run smokehouse in child processes for selected smoketests
+ * Display output from each as soon as they finish, but resolve function when ALL are complete
+ * @param {Array<SmoketestDfn>} smokes
+ */
+async function runSmokehouse(smokes) {
+  const cmdPromises = [];
+  for (const {id, expectations, config} of smokes) {
+    console.log(`${purpleify(id)} smoketest starting…`);
```
brendankenny: for some reason, just for the perf batch, Travis does 

screen shot 2018-04-10 at 16 36 19

not sure why :)

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -0,0 +1,191 @@
...
+    // The promise ensures we output immediately, even if the process errors
+    const p = execAsync(cmd, {timeout: 3 * 60 * 1000, encoding: 'utf8'}).then(cp => {
+      const ret = {id: id, process: cp};
+      displaySmokehouseOutput(ret);
+      return ret;
+    }).catch(err => {
+      const ret = {id: id, error: err};
+      displaySmokehouseOutput(ret);
+      return ret;
+    });
```
brendankenny: just to code golf, could also be
```js
const p = execAsync(cmd, {timeout: 3 * 60 * 1000, encoding: 'utf8'}).then(cp => {
  return {id: id, process: cp};
}).catch(err => {
  return {id: id, error: err};
}).then(result => displaySmokehouseOutput(result));
```

lighthouse-cli/test/smokehouse/run-smoke.js
paulirish: currently there's a default batch (with like 7 smoketests) and a perf batch (with 2).

when perf batch goes, the 2 smoketests in there do run in parallel.

I was hoping that style was a little more clear with this "batch" framing.

PR #4749, opened 2018-03-14
core(is-on-https): consider about:* secure

patrickhulce:
fixes [crbug.com/810709](https://bugs.chromium.org/p/chromium/issues/detail?id=810709)

### DIFF
--- a/lighthouse-core/audits/is-on-https.js
+++ b/lighthouse-core/audits/is-on-https.js
@@ -9,7 +9,7 @@ const Audit = require('./audit');
 const URL = require('../lib/url-shim');
 const Util = require('../report/v2/renderer/util');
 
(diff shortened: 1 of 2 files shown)


paulirish:
A more involved fix is flipping nonApplicable to true in this case. 

Also it looks like `about:` is not a secure origin.. 


image

lol that this was reported. :)

patrickhulce:
@paulirish how about when on an HTTPS page?

elawrence's comment seemed like this was legit. I'm not sure we should flag abouts as there's no action item to convert about:blank into an HTTPS URL is there? does an about:blank iframe on an https page do anything bad?

> Contexts that inherit a security context (about:, javascript:) should probably be exempted from the audit.

paulirish:
I see, okay. So we're focused on about:blank as a resource/iframe on the page rather than the root document itself. Okay, fine then. :)

This makes sense.  Also good luck trying to audit a root 'about:blank' page. :p

PR #4751, opened 2018-03-14
core: look up custom gatherer relative to the config file path

echopi:


echopi:
I signed it!

paulirish:
Thanks very much for the contribution! Appreciate it. 
It's 7pm so we'll take a look tomorrow.

patrickhulce:
awesome thanks for the fix @echopi! would you mind adding a test for your use case? ours are covered by the existing ones but we want to make sure not to break you in the future :)

https://github.com/GoogleChrome/lighthouse/blob/ce8ed566e994ccfac32f1f24a4909a78130e53a5/lighthouse-core/test/config/config-test.js#L642-L684

echopi:
Done

PR #4762, opened 2018-03-14
core: add type checking to audit and gatherer base classes

brendankenny:
having typing here helps in a number of places and was easy to pull out since they're relatively isolated. This is just the base classes...all the implementing classes are untouched (and unchecked)

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -218,6 +218,8 @@ module.exports = Audit;
 // TODO: placeholder typedefs until Details are typed
 /**
  * @typedef {void} Audit.DetailsRenderer.DetailsSummary
+ * @property {number=} wastedMs
+ * @property {number=} wastedKb
  */
 
 /**
@@ -225,5 +227,5 @@ module.exports = Audit;
  * @property {'table'} type
  * @property {Array<Audit.Heading>} headings
  * @property {Array<Object<string, string>>} items
- * @property {void} summary
+ * @property {Audit.DetailsRenderer.DetailsSummary} summary
  */
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -190,10 +190,10 @@ module.exports = Audit;
  * @property {string} description
  * @property {string} helpText
  * @property {Array<string>} requiredArtifacts
- * @property {string=} failureDescription
- * @property {boolean=} informative
- * @property {boolean=} manual
- * @property {Audit.ScoringModeValues=} scoreDisplayMode
+ * @property {string|undefined} failureDescription
+ * @property {boolean|undefined} informative
+ * @property {boolean|undefined} manual
+ * @property {Audit.ScoringModeValues|undefined} scoreDisplayMode
  */
 
 /**
@@ -212,14 +212,14 @@ module.exports = Audit;
  * @property {number} results
  * @property {Audit.Headings} headings
  * @property {boolean} passes
- * @property {string=} debugString
+ * @property {string|undefined} debugString
  */
 
 // TODO: placeholder typedefs until Details are typed
 /**
  * @typedef {void} Audit.DetailsRenderer.DetailsSummary
- * @property {number=} wastedMs
- * @property {number=} wastedKb
+ * @property {number|undefined} wastedMs
+ * @property {number|undefined} wastedKb
  */
 
 /**
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -3,7 +3,6 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
-// @ts-nocheck
 'use strict';
 
 const statistics = require('../lib/statistics');
@@ -19,14 +18,14 @@ const clampTo2Decimals = val => Math.round(val * 100) / 100;
 
 class Audit {
   /**
-   * @return {!string}
+   * @return {string}
    */
   static get DEFAULT_PASS() {
     return DEFAULT_PASS;
   }
 
   /**
-   * @return {{NUMERIC: string, BINARY: string}}
+   * @return {Audit.ScoringModes}
    */
   static get SCORING_MODES() {
     return {
@@ -36,14 +35,14 @@ class Audit {
   }
 
   /**
-   * @throws {Error}
+   * @return {Audit.Meta}
    */
   static get meta() {
     throw new Error('Audit meta information must be overridden.');
   }
 
   /**
-   * Computes a clamped score between 0 and 100 based on the measured value. Score is determined by
+   * Computes a clamped score between 0 and 1 based on the measured value. Score is determined by
    * considering a log-normal distribution governed by the two control points, point of diminishing
    * returns and the median value, and returning the percentage of sites that have higher value.
    *
@@ -65,9 +64,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
+   * @param {typeof Audit} audit
    * @param {string} debugString
-   * @return {!AuditFullResult}
+   * @return {LH.AuditFullResult}
    */
   static generateErrorAuditResult(audit, debugString) {
     return Audit.generateAuditResult(audit, {
@@ -78,10 +77,10 @@ class Audit {
   }
 
   /**
-   * @param {!Audit.Headings} headings
-   * @param {!Array<!Object<string, string>>} results
-   * @param {!DetailsRenderer.DetailsSummary} summary
-   * @return {!DetailsRenderer.DetailsJSON}
+   * @param {Audit.Headings} headings
+   * @param {Array<Object<string, string>>} results
+   * @param {Audit.DetailsRenderer.DetailsSummary} summary
+   * @return {Audit.DetailsRenderer.DetailsJSON}
    */
   static makeTableDetails(headings, results, summary) {
     if (results.length === 0) {
@@ -102,9 +101,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
-   * @param {!AuditResult} result
-   * @return {{score: number, scoreDisplayMode: string}}
+   * @param {typeof Audit} audit
+   * @param {LH.AuditResult} result
+   * @return {{score: number, scoreDisplayMode: Audit.ScoringModeValues}}
    */
   static _normalizeAuditScore(audit, result) {
     // Cast true/false to 1/0
@@ -125,9 +124,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
-   * @param {!AuditResult} result
-   * @return {!AuditFullResult}
+   * @param {typeof Audit} audit
+   * @param {LH.AuditResult} result
+   * @return {LH.AuditFullResult}
    */
   static generateAuditResult(audit, result) {
     if (typeof result.rawValue === 'undefined') {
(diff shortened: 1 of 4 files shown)


lighthouse-core/audits/audit.js
```diff
@@ -78,10 +77,10 @@ class Audit {
   }
 
   /**
-   * @param {!Audit.Headings} headings
-   * @param {!Array<!Object<string, string>>} results
-   * @param {!DetailsRenderer.DetailsSummary} summary
-   * @return {!DetailsRenderer.DetailsJSON}
+   * @param {Audit.Headings} headings
+   * @param {Array<Object<string, string>>} results
+   * @param {Audit.DetailsRenderer.DetailsSummary} summary
```
brendankenny: stuck these on `Audit` to keep them local since they're going to be pulled in by the larger LHR/lite effort. `summary` is defined as void for now to encourage adding types :)

typings/externs.d.ts
```diff
@@ -4,139 +4,141 @@
...
-
-export interface DevToolsJsonTarget {
-  description: string;
-  devtoolsFrontendUrl: string;
-  id: string;
-  title: string;
-  type: string;
-  url: string;
-  webSocketDebuggerUrl: string;
+declare namespace LH {
```
brendankenny: turns out `export as namespace` was really only intended for UMD modules. This is the "proper" way to declare a global type namespace (in a d.ts file)

lighthouse-core/gather/gatherers/gatherer.js
```diff
@@ -28,28 +28,42 @@ class Gatherer {
 
   /**
    * Called before navigation to target url.
-   * @param {!Object} options
+   * @param {Gatherer.PassContext} passContext
+   * @return {*|!Promise<*>}
    */
-  beforePass(options) { }
+  beforePass(passContext) { }
```
patrickhulce: 😃

typings/externs.d.ts
brendankenny: https://github.com/GoogleChrome/lighthouse/pull/4762/files?w=1#diff-518173330fe66b5974ceeb9a14a7b215 to see actual changes to externs file

lighthouse-core/gather/gatherers/gatherer.js
```diff
@@ -28,28 +28,42 @@ class Gatherer {
...
    * @return {*|!Promise<*>}
    */
-  afterPass(options, loadData) { }
+  afterPass(passContext, loadData) { }
 
   /* eslint-enable no-unused-vars */
 }
 
+/**
+ * @typedef {Object} Gatherer.PassContext
```
patrickhulce: 🎉

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "lighthouse-cli/**/*.js",
     "lighthouse-core/lib/dependency-graph/**/*.js",
     "lighthouse-core/gather/connections/**/*.js",
+    "lighthouse-core/gather/gatherers/gatherer.js",
```
patrickhulce: add audit.js too?

brendankenny: yeah, better to be explicit

lighthouse-core/audits/audit.js
```diff
@@ -65,9 +64,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
+   * @param {typeof Audit} audit
```
paulirish: what's this mean btw?

lighthouse-core/audits/audit.js
```diff
@@ -193,3 +214,16 @@ module.exports = Audit;
  * @property {boolean} passes
  * @property {string=} debugString
  */
+
+// TODO: placeholder typedefs until Details are typed
+/**
+ * @typedef {void} Audit.DetailsRenderer.DetailsSummary
```
paulirish: we could go with this for now:

```ts
 wastedMs?: number
 wastedKb?: number
```

lighthouse-core/audits/audit.js
brendankenny: `Audit` is the name of the class in js-land, but in type-land it's the name of the type of *instances* of `Audit`.

To refer to the type of the class `Audit` itself (since we use all static methods) you have to use `typeof`, which is kind of unfortunately overloaded in TS but gets the job done and isn't *too* obscure

lighthouse-core/audits/audit.js
brendankenny: done

lighthouse-core/audits/audit.js
```diff
@@ -175,6 +174,28 @@ class Audit {
...
+ * @typedef {Audit.ScoringModes[keyof Audit.ScoringModes]} Audit.ScoringModeValues
+ */
+
+/**
+ * @typedef {Object} Audit.Meta
+ * @property {string} name
+ * @property {string} description
+ * @property {string} helpText
+ * @property {Array<string>} requiredArtifacts
+ * @property {string=} failureDescription
```
paulirish: how do you feel about doing `foo|undefined` instead of `foo=`? i can almost promise i'll forget what `foo=` means in another 3 weeks.

lighthouse-core/audits/audit.js
```diff
@@ -125,9 +124,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
-   * @param {!AuditResult} result
-   * @return {!AuditFullResult}
+   * @param {typeof Audit} audit
+   * @param {LH.AuditResult} result
```
paulirish: Something that's been bugging me lately.. In our external LHR docs we want to basically call `LH.AuditFullResult` the "Audit Result". So one alternative here is...

* `LH.AuditResult` => something.. (like LH.AuditReturnValue) and 
* `LH.AuditFullResult` => `LH.AuditResult`

typings/externs.d.ts
```diff
@@ -4,139 +4,141 @@
...
+  }
+
+  export interface AuditResults {
+    [metric: string]: AuditResult;
+  }
+
+  export interface AuditFullResult extends AuditResult {
+    displayValue: string;
+    score: number;
+    scoreDisplayMode: 'numeric' | 'binary';
```
paulirish: Audit.ScoringModes?

lighthouse-core/audits/audit.js
```diff
@@ -102,9 +101,9 @@ class Audit {
   }
 
   /**
-   * @param {!Audit} audit
-   * @param {!AuditResult} result
-   * @return {{score: number, scoreDisplayMode: string}}
+   * @param {typeof Audit} audit
+   * @param {LH.AuditResult} result
+   * @return {{score: number, scoreDisplayMode: Audit.ScoringModeValues}}
```
paulirish: `Audit.ScoringModeValues` vs `@typedef {Object} Audit.ScoringModes` below?

typings/externs.d.ts
```diff
@@ -4,139 +4,141 @@
...
+  export interface AuditResult {
+    rawValue: boolean | number | null;
+    displayValue?: string;
+    debugString?: string;
+    score?: number;
+    extendedInfo?: {value: string};
+    notApplicable?: boolean;
+    error?: boolean;
+    // TODO: define details
+    details?: object;
```
paulirish: (for when it's time to take on `details`...) 

i did an initial spike of `details`'s types in https://github.com/GoogleChrome/lighthouse/compare/95e19a56503c702a5077c62dda1112cd39c4d705...typechecking#diff-518173330fe66b5974ceeb9a14a7b215


But then https://docs.google.com/document/d/1TumV5t3xtJ_e2b8y9PTwh4j2RaT-lmaQI9lwmF5tD2U/edit#heading=h.y9rlfi2u13u9 has a more future-forward definition, which is even stricter.

typings/externs.d.ts
```diff
@@ -4,139 +4,141 @@
...
+    perf: boolean;
+    mixedContent: boolean;
+    verbose: boolean;
+    quiet: boolean;
+  }
+
+  export interface Config {}
+
+  export interface AuditResult {
+    rawValue: boolean | number | null;
```
paulirish: we need null?

lighthouse-core/audits/audit.js
paulirish: huh! thanks for explaining.
wish it was `instanceof` instead.

lighthouse-core/audits/audit.js
brendankenny: I don't mind the explicitness, especially because it's only allowed in some circumstances, but on the other hand it can get pretty unwieldy and there's no jsdoc equivalent to the nice `?:` on the ts side

lighthouse-core/audits/audit.js
brendankenny: yeah, the problem used to be we needed a nice name in each audit and we didn't care so much what came out the other side. I don't mind the switch but we should come up with a good alternative for the first one (and maybe punt on this because there are a lot of `AuditResult`s that would need to change

typings/externs.d.ts
brendankenny: `Audit.ScoringModes` is scoped to `audit.js`, so the externs would need to import `Audit` to have access. I think that's in our future anyways, but not worth it for this (it changes other things when our `.d.ts` file imports things because then it too becomes a module...)

typings/externs.d.ts
brendankenny: yeah, audit error results have `rawValue` of null (`generateErrorAuditResult`). Scoring seems to not really care, but we have unit tests asserting it so I just left it as is

lighthouse-core/audits/audit.js
brendankenny: Fortunately you almost always the normal thing. It's just here we're not just using all static methods, we want the static methods on subclasses of `Audit`, so we have to pass the subclass in to `Audit`'s static methods... 🤪

lighthouse-core/audits/audit.js
brendankenny: `ScoringModeValues` is one of the string values of the `Audit.ScoringModes` object (so just `'numeric' | 'binary'`). We could probably fix this up better, maybe if we pull most of the audit stuff into here and/or an `audit.d.ts` file

lighthouse-core/audits/audit.js
paulirish: k can we do `foo|undefined`?

lighthouse-core/audits/audit.js
paulirish: fine with punting for today.

typings/externs.d.ts
paulirish: mmmmmmmmmmmkkkkkkkkkkkkkkkkkkkkayyyyyyyyyyy.

paulirish:
lgtm once foo|undefined

lighthouse-core/audits/audit.js
patrickhulce: oh but there is! http://usejsdoc.org/tags-property.html :)

```js
/**
 * @property {Array<string>} requiredArtifacts
 * @property {string} [failureDescription]
 */
```

typings/externs.d.ts
patrickhulce: The future plan is to much as much as external facing stuff as possible to a .d.ts file anyway right?

lighthouse-core/audits/audit.js
brendankenny: that...does not seem better than `=` :P

typings/externs.d.ts
brendankenny: I'm not sure. There's also what we want global vs scoped. What we produce for others to link against our public API may or may not be tied to that :)

lighthouse-core/audits/audit.js
paulirish: it seems slightly better to me, but not as good as ? or |undefined

brendankenny: you guys are crazy, optional is part of the type not the name!

luckily we can all agree on good old boring dependable `|undefined`

patrickhulce: > you guys are crazy, optional is part of the type not the name!

but the semantics are different!

`|undefined` implies the key is always there but just potentially with undefined, the `[prop]` syntax more correctly suggests that the property name is optional rather than the type of the value is sometimes undefined :)

brendankenny: typescript treats `{type=}` and `[propName]` the same, but the distinction from `|undefined` *is* important...and I was wrong before about `|undefined`. tsc isn't going to let us use it anyways with `--strict` on.

The difference is erased when you get something that is this type, but not when declaring that something is one of these types...e.g. with `@property {boolean|undefined} informative` you better have a property `informative` that's one of the two. A missing `informative` won't suffice, so the property is no longer optional.

I like this reasoning of `[prop]`, it offers a nice way of expressing the defined/not-defined vs value/undefined mixup of JS, so let's go with that.

typings/externs.d.ts
brendankenny: I think I was also wrong here. All the typedefs in `Audit` appear to be entirely local to the file and aren't exported, even though they're on an object that was, so all of these may need to be in d.ts files unless only used inside this file (that includes subclasses, so right now no audits can see the typedef of `Audt.Meta` even though they all import `Audit`

brendankenny:
fixed up optional properties. I'll follow up with `d.ts` changes later

PR #4778, opened 2018-03-15
core: convert diagnostics to numeric scores

### LINKED ISSUE #4521: Re-evaluate runtime perf thresholds
For example mainthread-work-breakdown fails if work takes > 3s. Is that a reasonable result on all hardware, including production google-scale infra?

patrickhulce:
closes #4521 

I picked the thresholds so that the score turns yellow at whatever the previous threshold was
this also changes the base audit class so that the `failureDescription` is only used if we mark the audit not in green as it was strange to see "JavaScript bootup time is too high: 99"

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -7,6 +7,7 @@
 'use strict';
 
 const statistics = require('../lib/statistics');
+const Util = require('../report/v2/renderer/util');
 
 const DEFAULT_PASS = 'defaultPass';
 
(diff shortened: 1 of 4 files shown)


patrickhulce:
> Because it's been slightly fun to juggle the non-ByteEfficiency audits that are within opportunities.

Yeah that sounds good for all opportunities to share some common base to help with graph estimations 👍 

Not 100% sold on the same for the diagnostics since the only thing they really share is scoring based on a sum of time that changes for each class and the meta changes which we know from manual are ugly to share :) haha, what are the primary shared components you're thinkin'?

paulirish:
Well for opps, i think we have WastedBytes and WastedTime as the two options. I don't see other ones right now.

TBH i'm not sure about diagnostics right now.  here was a quick categorization i tried: 
image

sooo given it appears quite messier than opportunities, let's solve that problem another time. :)

paulirish:
exciting closure failure waiting for you :)

image

PR #4779, opened 2018-03-15
core(emulation): update Chrome userAgent to 66

paulirish:
overdue. :)

i'm also open to updating the Android version. wdyt?


paulirish:
> updating android version

image

so right now we're on the most common android version. so actually that seems reasonable. we have a 3-4 year old device on the most common android version. WFM, personally. 👍 

sg?

PR #4783, opened 2018-03-16
core(runner): rename generatedTime to fetchedAt

### LINKED ISSUE #4753: Fetch the gatheredTimestamp
Right now we add a `.generatedTime` timestamp to the LHR that marks when [runner is packaging up the LHR](https://github.com/GoogleChrome/lighthouse/blob/1fc455786ab30dfad8fabfb100e3ed34b7bbf4d2/lighthouse-core/runner.js#L105).

IMO, a more useful timestamp is when we start gathering; basically when we loaded the page. Beyond semantics, this probably only matters in the situations where gathering&auditing are done separately. Recently, I've been repeatedly regenerating the sample_v2.json, doing `-A` runs from the exact same artifacts. I'd argue it doesn't really make sense for the timestamp in the LHR to change. 

(Implementation-wise: Collecting in the gatherrunner right [when we grab the userAgent](https://github.com/GoogleChrome/lighthouse/blob/1fc455786ab30dfad8fabfb100e3ed34b7bbf4d2/lighthouse-core/gather/gather-runner.js#L103-L105) makes sense to me). 

A few questions:

1. Do ya'll agree this is a more useful timestamp?
1. If we start exposing this, should we stop exposing the current `.generatedTime`? (including for the LHR-Lite)
1. If we do that, we should probably rename this property. We could do just `.startTime`. And then the timings in #3745 would be treated just like navStart & resource timing. 

thoughts?

ev1stensberg:
Fixes #4753 

# Summary

Refactored some code to fit the new convention. `fetchedAt` gets mounted to the options object and pulled in the runner. Default value/fallback is set in the formatter, so no need to set default value on an empty Date object (doesn't ever resolve to null either). 

Ping me if there's anything that needs to be revised. :octocat:

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -100,13 +100,11 @@ class GatherRunner {
     const resetStorage = !options.flags.disableStorageReset;
     // Enable emulation based on flags
     return driver.assertNoSameOriginServiceWorkerClients(options.url)
-      .then(_ => {
-        gathererResults.fetchedAt = [(new Date()).toJSON()];
-        return driver.getUserAgent();
-      })
+      .then(_ => driver.getUserAgent())
       .then(userAgent => {
         gathererResults.UserAgent = [userAgent];
         GatherRunner.warnOnHeadless(userAgent, gathererResults);
+        gathererResults.fetchedAt = [(new Date()).toJSON()];
       })
       .then(_ => driver.beginEmulation(options.flags))
       .then(_ => driver.enableRuntimeEvents())
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -101,7 +101,7 @@ class GatherRunner {
     // Enable emulation based on flags
     return driver.assertNoSameOriginServiceWorkerClients(options.url)
       .then(_ => {
-        gathererResults.fetchedAt = (new Date()).toJSON();
+        gathererResults.fetchedAt = [(new Date()).toJSON()];
         return driver.getUserAgent();
       })
       .then(userAgent => {
@@ -351,7 +351,6 @@ class GatherRunner {
     // Nest LighthouseRunWarnings, if any, so they will be collected into artifact.
     const uniqueWarnings = Array.from(new Set(gathererResults.LighthouseRunWarnings));
     gathererResults.LighthouseRunWarnings = [uniqueWarnings];
-    gathererResults.fetchedAt = [gathererResults.fetchedAt];
 
     const pageLoadFailures = [];
     return Object.keys(gathererResults).reduce((chain, gathererName) => {
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -101,7 +101,7 @@ class GatherRunner {
     // Enable emulation based on flags
     return driver.assertNoSameOriginServiceWorkerClients(options.url)
       .then(_ => {
-        options.fetchedAt = (new Date()).toJSON();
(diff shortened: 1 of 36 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -100,7 +100,10 @@ class GatherRunner {
     const resetStorage = !options.flags.disableStorageReset;
     // Enable emulation based on flags
     return driver.assertNoSameOriginServiceWorkerClients(options.url)
-      .then(_ => driver.getUserAgent())
+      .then(_ => {
+        options.fetchedAt = (new Date()).toJSON();
+        return driver.getUserAgent();
```
ev1stensberg: The other promise expects the user agent in return, throws else. Can convert to one-liner if you want to.

ev1stensberg:
Looks like the Chrome Infra is using generatedTime, should I change that?

lighthouse-core/gather/gather-runner.js
```diff
@@ -100,7 +100,10 @@ class GatherRunner {
     const resetStorage = !options.flags.disableStorageReset;
     // Enable emulation based on flags
     return driver.assertNoSameOriginServiceWorkerClients(options.url)
-      .then(_ => driver.getUserAgent())
+      .then(_ => {
+        options.fetchedAt = (new Date()).toJSON();
```
patrickhulce: think we can stick it on gathererResults instead like we do for user agent?

ev1stensberg: It's gonna be an artifact though? Replaced with `gatherRunner` and when running the tests ( 2 failed ), this is what came up:

skjermbilde 2018-03-17 kl 00 48 26

Lighthouse has `result.fetchedAt` as a prop above the artifact, but it resolves to undefined

ev1stensberg:
Also, noticed you're replacing Nexus in favor for Pixel in emulation. Could I take a crack at that?

lighthouse-core/gather/gather-runner.js
patrickhulce: it makes a little more semantic sense to have it as an artifact like UserAgent as it is produced through the gathering process. also I think we'll have to to get this to work with the `lighthouse -G && lighthouse -A` workflow that paul flagged as part of #4753 since we don't save options to disk, just artifacts

patrickhulce: it might need some updating of tests as you noticed though :) sorry!

patrickhulce:
> Also, noticed you're replacing Nexus in favor for Pixel in emulation. Could I take a crack at that?

I think that proposal needs a little more discussion before we move to impl ;) love the enthusiasm though!

lighthouse-core/gather/gather-runner.js
ev1stensberg: Makes sense now as you say it, a bit busy now, but going to address it during the weekend 👍

lighthouse-core/gather/gather-runner.js
```diff
@@ -348,6 +351,7 @@ class GatherRunner {
     // Nest LighthouseRunWarnings, if any, so they will be collected into artifact.
     const uniqueWarnings = Array.from(new Set(gathererResults.LighthouseRunWarnings));
     gathererResults.LighthouseRunWarnings = [uniqueWarnings];
+    gathererResults.fetchedAt = [gathererResults.fetchedAt];
```
ev1stensberg: Works now, sorry for delaying. Extremely hungover yesterday 😄

ev1stensberg: Another thing, if people use this through the CLI, do we need to expose name etc in order to save it to disk using `-A` or `-G`?

PR #4791, opened 2018-03-17
misc(eslint): update to es2017 to allow async/await

paulirish:


.eslintrc.js
```diff
@@ -68,7 +68,7 @@ module.exports = {
     'arrow-parens': 0,
   },
   parserOptions: {
-    ecmaVersion: 6,
+    ecmaVersion: 2017,
```
patrickhulce: how about 8 to keep easier with our node version :)

PR #4792, opened 2018-03-17
core(runner): add custom folder support to -G/-A

paulirish:
Enable this:

```
lighthouse -G=./cnn_artifacts http://cnn.com
...
lighthouse -A=./cnn_artifacts http://cnn.com
```

Path is resolved via pwd, but absolute paths are allowable, too.

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -111,7 +111,7 @@ function getFlags(manualArgv) {
         'disable-storage-reset', 'disable-device-emulation', 'disable-cpu-throttling',
         'disable-network-throttling', 'save-assets', 'list-all-audits',
         'list-trace-categories', 'perf', 'view', 'verbose', 'quiet', 'help',
-        'gather-mode', 'audit-mode', 'mixed-content',
+        'mixed-content',
       ])
       .choices('output', printer.getValidOutputOptions())
       // force as an array
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -73,8 +73,8 @@ function getFlags(manualArgv) {
         'disable-cpu-throttling': 'Disable CPU throttling',
         'disable-network-throttling': 'Disable network throttling',
         'gather-mode':
-            'Collect artifacts from a connected browser and save to disk. If audit-mode is not also enabled, the run will quit early.',
-        'audit-mode': 'Process saved artifacts from disk',
+            'Collect artifacts from a connected browser and save to disk. (Artifacts folder path may optionally be provided). If audit-mode is not also enabled, the run will quit early.',
+        'audit-mode': 'Process saved artifacts from disk. (Artifacts folder path may be provided, otherwise defaults to ./latest-run/)',
         'save-assets': 'Save the trace contents & screenshots to disk',
         'list-all-audits': 'Prints a list of all available audits and exits',
         'list-trace-categories': 'Prints a list of all required trace categories and exits',
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -134,19 +134,20 @@ class Runner {
    * @param {*} connection
    * @return {!Promise<!Artifacts>}
    */
-  static _gatherArtifactsFromBrowser(opts, connection) {
+  static async _gatherArtifactsFromBrowser(opts, connection) {
     if (!opts.config.passes) {
       return Promise.reject(new Error('No browser artifacts are either provided or requested.'));
     }
 
     opts.driver = opts.driverMock || new Driver(connection);
-    return GatherRunner.run(opts.config.passes, opts).then(artifacts => {
-      const flags = opts.flags;
-      const shouldSave = flags.gatherMode;
+    const artifacts = await GatherRunner.run(opts.config.passes, opts);
+
+    const flags = opts.flags;
+    if (flags.gatherMode) {
       const path = Runner._getArtifactsPath(flags);
-      const p = shouldSave ? Runner._saveArtifacts(artifacts, path): Promise.resolve();
-      return p.then(_ => artifacts);
-    });
+      await Runner._saveArtifacts(artifacts, path);
+    }
+    return artifacts;
   }
 
   /**
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -18,8 +18,6 @@ const path = require('path');
 const URL = require('./lib/url-shim');
 const Sentry = require('./lib/sentry');
 
-const basePath = path.join(process.cwd(), 'latest-run');
-
 class Runner {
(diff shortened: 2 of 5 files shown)


lighthouse-core/runner.js
```diff
@@ -144,18 +143,20 @@ class Runner {
     return GatherRunner.run(opts.config.passes, opts).then(artifacts => {
       const flags = opts.flags;
       const shouldSave = flags.gatherMode;
-      const p = shouldSave ? Runner._saveArtifacts(artifacts): Promise.resolve();
+      const path = Runner._getArtifactsPath(flags);
+      const p = shouldSave ? Runner._saveArtifacts(artifacts, path): Promise.resolve();
```
patrickhulce: since we just bumped our `ecmaVersion` how about some `async/await` to kill this dumb `p` variable :D

paulirish: 👍  sgtm!

paulirish: done

lighthouse-core/runner.js
```diff
@@ -68,7 +66,7 @@ class Runner {
     // Gather phase
     // Either load saved artifacts off disk, from config, or get from the browser
     if (opts.flags.auditMode && !opts.flags.gatherMode) {
-      run = run.then(_ => Runner._loadArtifactsFromDisk());
+      run = run.then(_ => Runner._loadArtifactsFromDisk(Runner._getArtifactsPath(opts.flags)));
```
brendankenny: nit: pull into a var

lighthouse-core/runner.js
```diff
@@ -135,27 +134,30 @@ class Runner {
    * @param {*} connection
    * @return {!Promise<!Artifacts>}
    */
-  static _gatherArtifactsFromBrowser(opts, connection) {
+  static async _gatherArtifactsFromBrowser(opts, connection) {
```
brendankenny: fun!

lighthouse-core/runner.js
```diff
@@ -135,27 +134,30 @@ class Runner {
...
-      const shouldSave = flags.gatherMode;
-      const p = shouldSave ? Runner._saveArtifacts(artifacts): Promise.resolve();
-      return p.then(_ => artifacts);
-    });
+    const artifacts = await GatherRunner.run(opts.config.passes, opts);
+
+    const flags = opts.flags;
+    if (flags.gatherMode) {
+      const path = Runner._getArtifactsPath(flags);
+      await Runner._saveArtifacts(artifacts, path);
```
brendankenny: inherited, but this would probably make more sense back up in `run()`

lighthouse-core/runner.js
```diff
@@ -410,6 +412,20 @@ class Runner {
...
+    // This enables usage like: -GA=./custom-folder
+    if (typeof flags.auditMode === 'string') return path.join(process.cwd(), flags.auditMode);
+    if (typeof flags.gatherMode === 'string') return path.join(process.cwd(), flags.gatherMode);
+    return path.join(process.cwd(), 'latest-run');
+  }
 }
 
 module.exports = Runner;
+
+
```
brendankenny: extra line?

lighthouse-cli/cli-flags.js
```diff
@@ -111,7 +111,7 @@ function getFlags(manualArgv) {
         'disable-storage-reset', 'disable-device-emulation', 'disable-cpu-throttling',
         'disable-network-throttling', 'save-assets', 'list-all-audits',
         'list-trace-categories', 'perf', 'view', 'verbose', 'quiet', 'help',
-        'gather-mode', 'audit-mode', 'mixed-content',
+        'mixed-content',
```
brendankenny: mention path option in description?

lighthouse-core/runner.js
```diff
@@ -410,6 +412,20 @@ class Runner {
       extraHeaders: flags.extraHeaders || {},
     };
   }
+
+  /**
+   * Get path to use for -G and -A modes. Defaults to $CWD/latest-run
+   * @param {*} flags
```
brendankenny: more specific type pls

brendankenny: also need to update externs.d.ts

lighthouse-core/test/runner-test.js
```diff
@@ -44,9 +45,15 @@ describe('Runner', () => {
       }],
       audits: ['content-width'],
     });
+    const artifactsPath = '.tmp/test_artifacts';
+
+    after(() => {
+      const path = Runner._getArtifactsPath({auditMode: artifactsPath});
```
brendankenny: test to verify that artifacts do exist at that location?  And preferably since `Runner._getArtifactsPath` is private no need to call it for rimraf if verified that runner is saving at that location

lighthouse-core/runner.js
paulirish: true. but how about we do that when runner goes async/await?

brendankenny: Well you're already refactoring it to async/await so you brought it on yourself :P

It doesn't look like there are any tests of this method (since it's private, yay), so should pop into the surrounding else branch with no side effects?

lighthouse-core/runner.js
```diff
@@ -410,6 +413,18 @@ class Runner {
...
   }
+
+  /**
+   * Get path to use for -G and -A modes. Defaults to $CWD/latest-run
+   * @param {Flags} flags
+   * @return {string}
+   */
+  static _getArtifactsPath(flags) {
+    // This enables usage like: -GA=./custom-folder
+    if (typeof flags.auditMode === 'string') return path.join(process.cwd(), flags.auditMode);
```
brendankenny: I think you want `path.resolve`? To handle absolute paths

lighthouse-core/test/runner-test.js
```diff
@@ -36,17 +38,23 @@ describe('Runner', () => {
     resetSpies();
   });
 
-  describe('Gather Mode & Audit Mode', () => {
+  describe.only('Gather Mode & Audit Mode', () => {
```
brendankenny: remove

lighthouse-core/test/runner-test.js
```diff
@@ -36,17 +38,23 @@ describe('Runner', () => {
...
     const url = 'https://example.com';
     const generateConfig = _ => new Config({
       passes: [{
         gatherers: ['viewport-dimensions'],
       }],
       audits: ['content-width'],
     });
+    const artifactsPath = '.tmp/test_artifacts';
+    const resolvedPath = Runner._getArtifactsPath({auditMode: artifactsPath});
```
brendankenny: since the expected behavior is that we support a path relative to `cwd`, I think we want an impl independent of the thing itself to test it (e.g. `const resolvedPath = path.resolve(process.cwd(), artifactsPath)` or whatever works from this file)

PR #4793, opened 2018-03-17
tests(fixtures): introduce script to update the report fixtures

paulirish:
This includes my previous two PRs. sorry. [Compare link without those PRs in the mix.](https://github.com/GoogleChrome/lighthouse/pull/4793/files/6c66034bc25030070d589f5460895a03148654a6..c940da871dc0c572125140b2871cf990b2a82fc3)

#### Additions / changes: 

* `yarn update:json`: refreshes the sample_v2.json from static artifacts. Great as it's the exact same load, so minimal json changes.
* `yarn update:artifacts`: needed less often, but necessary when changes/additions are made to gatherers
* static-server is a bit more scriptable now, and this PR lays the groundwork for porting the smokehouse run-test scripts into JS (yo #4748!).

This _does_ check in 3.9MB of artifacts to the repo. But I think it's worth it.


paulirish:
> maybe I'm being blinded, but not entirely what this gets us over a shell script...

sure sg. for the -A side of things, I've updated the npm script to use this direct approach.

The -G side of things still needs some special handling because of the web server.

.gitignore
```diff
@@ -24,6 +24,9 @@ last-run-results.html
 *.report.pretty
 *.artifacts.log
 
+!lighthouse-cli/test/sample_artifacts/*.trace.json
```
brendankenny: should these live somewhere with `sample_v2.json`?

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -0,0 +1,31 @@
...
+    gatherMode: 'lighthouse-cli/test/sample_artifacts',
+  };
+  // @ts-ignore Remove when we fix Flags typing
+  await cli.runLighthouse(url, flags, undefined);
+  await new Promise(res => server.close(res));
+}
+
+(async function() {
+  await update();
+})();
```
brendankenny: just any reason to wrap `update()`?

package.json
```diff
@@ -48,6 +48,8 @@
     "plots-smoke": "bash plots/test/smoke.sh",
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile changelog.md --same-file",
     "type-check": "tsc -p .",
+    "update:artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
```
brendankenny: Need something more specific. Maybe `update:sample-artifacts` or something? (same with `json` below)

tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "lighthouse-core/lib/dependency-graph/**/*.js",
     "lighthouse-core/gather/connections/**/*.js",
     "lighthouse-core/gather/gatherers/gatherer.js",
+    "lighthouse-core/scripts/*.js",
```
brendankenny: 🎉

.gitignore
paulirish: sure!

how about  `lighthouse-core/test/results/artifacts/`?

package.json
paulirish: sg. updated.

lighthouse-core/runner.js
```diff
@@ -429,3 +429,5 @@ class Runner {
 }
 
 module.exports = Runner;
+
```
patrickhulce: revert? :)

paulirish:
> I wonder if we should start printing devtoolsLog in the style of traces so we can get better diffs on them in artifacts

good idea.

PR #4809, opened 2018-03-19
tests(smokehouse): adjust byte efficiency CPU multiplier

### LINKED ISSUE #4808: Byte-efficiency smoketests fail on appveyor
since moving to lantern for byte-efficiency savings estimates the smoketests are failin on appveyor. It's  not a canary v. stable issues since they all pass locally and on travis. 

My current guess is that because appveyor is astronomically slower than travis/local, the CPU tasks x4 are waaaaaay longer than expected and the savings from any network just gets washed out.

patrickhulce:
fixes #4808

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -136,6 +136,7 @@ class UnusedBytes extends Audit {
    */
   static createAuditResult(result, graph) {
     const simulatorOptions = PredictivePerf.computeRTTAndServerResponseTime(graph);
+    Object.assign(simulatorOptions, {cpuTaskMultiplier: 1, layoutTaskMultiplier: 1});
     const simulator = new LoadSimulator(graph, simulatorOptions);
 
     const debugString = result.debugString;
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -136,6 +136,8 @@ class UnusedBytes extends Audit {
    */
   static createAuditResult(result, graph) {
     const simulatorOptions = PredictivePerf.computeRTTAndServerResponseTime(graph);
+    // TODO: calibrate multipliers, see https://github.com/GoogleChrome/lighthouse/issues/820
+    Object.assign(simulatorOptions, {cpuTaskMultiplier: 1, layoutTaskMultiplier: 1});
```
paulirish: why reset these multipliers to 1 rather than their defaults?

patrickhulce: well in our pre-lantern case the throttling is usually being applied by devtools, so pumping it through another 4x multiplier is sorta unfair. This shouldn't really matter once #4766 and follow-ups land

paulirish: aye true. k.

PR #4813, opened 2018-03-19
deps(browserify): update deep transitive dependency to use recent acorn

paulirish:
without this, #4792 fails in `yarn build-extension`

see https://github.com/substack/astw/pull/14

### DIFF
--- a/lighthouse-extension/package.json
+++ b/lighthouse-extension/package.json
@@ -25,5 +25,8 @@
     "run-sequence": "^1.1.5",
     "through2": "^2.0.1"
   },
+  "resolutions": {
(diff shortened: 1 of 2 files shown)


lighthouse-extension/package.json
```diff
@@ -25,5 +25,8 @@
     "run-sequence": "^1.1.5",
     "through2": "^2.0.1"
   },
+  "resolutions": {
+    "browserify/insert-module-globals/lexical-scope/astw": "github:goto-bus-stop/astw#new-acorn"
```
patrickhulce: yikes haha

critical goto-bus-stop repo essential to Chrome build process 😆

patrickhulce:
that linked issue was bumping it from 8 to 9, can we just use the proper released version that uses 8 as the default instead?

Issue #4818, opened 2018-03-20
Report: empty details table row when showing URLs


kdzwinel:
<img width="805" alt="screen shot 2018-03-20 at 02 09 26" src="https://user-images.githubusercontent.com/985504/37630319-348c27c8-2be4-11e8-9b06-3dde7db9dd95.png">

It looks like https://github.com/GoogleChrome/lighthouse/commit/bc9efba9b7b5c40a2886d6f1d366e272f420afb2#diff-99cea3b0b135c7e5a46c068661e7cf5eR96 updated only `type: 'text'` entries to use `value` instead of `text`, but `type: 'url'` should also be updated. 

Replacing `text` with `value` here:

https://github.com/GoogleChrome/lighthouse/blob/adf4f86dcef08d263bb6d77e117659a53d3c2002/lighthouse-core/audits/seo/is-crawlable.js#L110

fixes the is-crawlable audit, but potentially more audits are affected.

*[EDIT]* It looks like https://github.com/GoogleChrome/lighthouse/pull/4616 was created before is-crawlable was merged, so that's probably why is-crawlable wasn't updated to use `value` and it's probably only affected audit.

PR #4819, opened 2018-03-20
deps(robots-parser): patch robots-parser to work in both node and browser env

### LINKED ISSUE #4794: Page is blocked from indexing (libUrl.parse is not a function)
Hi,

Get this fail on every website. 
By the way clicking "Learn more" becomes 404.
indexing

kdzwinel:
I've tried creating a browserify transform that replaces `request('url')` calls with `request('path/to/url-shim.js')`. It worked, but uncovered another issue - robots-parser is using `url.path` that doesn't exists on the web and is deprecated in node. So I decied to go with https://github.com/ds300/patch-package as suggested by @patrickhulce and @wardpeet.

Fixes #4794, but nothing will show up in the details table due to #4818

### DIFF
--- a/patches/robots-parser+1.0.2.patch
+++ b/patches/robots-parser+1.0.2.patch
@@ -1,0 +1,36 @@
+patch-package
+--- a/node_modules/robots-parser/Robots.js
++++ b/node_modules/robots-parser/Robots.js
+@@ -1,4 +1,4 @@
+-var libUrl   = require('url');
++var URL   = (typeof self !== 'undefined' && self.URL) || require('url').URL;
+ var punycode = require('punycode');
+ 
+ /**
+@@ -176,7 +176,7 @@ function isPathAllowed(path, rules) {
+ 
+ 
+ function Robots(url, contents) {
+-	this._url = libUrl.parse(url);
++	this._url = new URL(url);
+ 	this._url.port = this._url.port || 80;
+ 	this._url.hostname = punycode.toUnicode(this._url.hostname);
+ 
+@@ -262,7 +262,7 @@ Robots.prototype.setPreferredHost = function (url) {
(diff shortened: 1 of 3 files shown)


patches/robots-parser+1.0.2.patch
```diff
@@ -0,0 +1,36 @@
+patch-package
+--- a/node_modules/robots-parser/Robots.js
```
kdzwinel: a diff of a diff 😵

paulirish:
Presumably this also needs some post install script to apply the patch?

kdzwinel:
@paulirish I also thought so, but …

> Patches created by patch-package are automatically and gracefully applied when you use npm(>=5) or yarn.

🕺

wardpeet:
Yeah patch-package is awesome!

kdzwinel:
Yeah, I agree it's not perfect. I thought about opening a PR with this patch against the original repo. We can also pull that library into LH or fork it.

patrickhulce:
>  I thought about opening a PR with this patch against the original repo

PR to have robots-parser use real `URL` class and switch to have us swap `require('url')` with our shim sounds like most reasonable long term solution to me

paulirish:
+1 to giving @samclarke a PR. 
Hi Sam, big fan. :)

kdzwinel:
Perfect, I'll open a PR👍 should I create an issue to track this?

brendankenny:
> PR to have robots-parser use real URL class and switch to have us swap require('url') with our shim sounds like most reasonable long term solution to me

SGreatTM

brendankenny:
> should I create an issue to track this?

that SG too

kdzwinel:
Issue created (#4825) 👌

PR #4820, opened 2018-03-20
core(is-crawlable): fix empty row in the details table when showing an URL

### LINKED ISSUE #4818: Report: empty details table row when showing URLs

kdzwinel:
Fixes #4818 

screen shot 2018-03-20 at 03 01 56

It looks like #4616 was created before is-crawlable was merged causing this file to be overlooked. I searched for other similar issues, but found none.

### DIFF
--- a/lighthouse-core/audits/seo/is-crawlable.js
+++ b/lighthouse-core/audits/seo/is-crawlable.js
@@ -107,7 +107,7 @@ class IsCrawlable extends Audit {
             blockingDirectives.push({
               source: {
                 type: 'url',
-                text: robotsFileUrl.href,
+                value: robotsFileUrl.href,
               },
             });
           }


kdzwinel:
@paulirish IMO this should also get a "to merge:2.x" tag

kdzwinel:
@patrickhulce gotcha 👍

Issue #495, opened 2016-07-06
Remove standalone / fullscreen requirement


triblondon:
The requirement for apps to have a display property of 'standalone' or 'fullscreen', implemented as a result of #143, is problematic.  Many in the web community object to this being considered a web best practice, notably articulated by Jeremy Keith in his [post](https://adactio.com/journal/10708), though I'm not endorsing all the sentiments he expresses there.

For me the question is whether Lighthouse is a community project designed to judge conformance with universally agreed best practices, or whether it is specifically designed to express the proprietary requirements of Google Chrome in respect to offering an install prompt.  If the latter, then since the heuristics/requirements for installability are not subject to a standard, it's totally valid for Chrome to set whatever requirements it likes - though care should probably be taken to ensure that the intersection of all vendors' installability requirements don't leave developers with insufficient control over their own application design.

However, if Lighthouse is a community tool, then this rule could be seen to be narrowing a recent specification that offers developers a range of options for good reasons.  If options like 'browser' and 'minimal-ui' are bad, why are they in the manifest spec?

If this is deemed justfiable on the grounds that 'native apps don't have a URL bar so PWAs shouldn't either' I would point to zoomability as a similar case - we consider sites that disable zoom to be bad practice, despite pinch-zooming the whole UI not typically being a feature offered by native apps.

This topic came up as part of a discussion of PWAs on a [TAG thread](https://github.com/w3ctag/spec-reviews/issues/123).

cc @marcoscaceres @torgo

paulirish:
> For me the question is whether Lighthouse is a community project designed to judge conformance with universally agreed best practices, or whether it is specifically designed to express the proprietary requirements of Google Chrome in respect to offering an install prompt.

Really, it's both. And when there is a conflict between these two goals, then we need to resolve it as Chrome shouldn't require something that violates best practices. 

I think it'd be good if Lighthouse required _a value_ for the `display` property, but not requiring any particular value. We certainly want the developer to decide what makes the most sense for their users.  

I believe there's an outstanding Chrome bug regarding its treatment of `display` in determining install promptability (e.g. [crbug/596729](https://bugs.chromium.org/p/chromium/issues/detail?id=596729#c11) and [crbug/471623](https://bugs.chromium.org/p/chromium/issues/detail?id=471623), but I'm supportive of separating that from Lighthouse's evaluation here. Until the issue is resolved, we can note it as a documentation note in Lighthouse.

samccone:
[...]

PR #4822, opened 2018-03-20
browser is valide `display` property

oliviertassinari:
### The issue

capture d ecran 2018-03-20 a 10 00 11

### The specification

capture d ecran 2018-03-20 a 09 59 39
[reference](https://developer.mozilla.org/en-US/docs/Web/Manifest)

### DIFF
--- a/lighthouse-core/gather/computed/manifest-values.js
+++ b/lighthouse-core/gather/computed/manifest-values.js
@@ -8,7 +8,7 @@
 const ComputedArtifact = require('./computed-artifact');
 const icons = require('../../lib/icons');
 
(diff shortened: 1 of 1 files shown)


paulirish:
Hi @oliviertassinari 

A valid PWA requires display of fullscreen/standalone/minimal-ui. Currently `display:standalone` isn't considered a PWA. 

Many teams are fairly happy with `minimal-ui` instead of `browser.

cc #495

oliviertassinari:
@paulirish Thank, I will update @Material-UI documentation to use `minimal-ui` 👍 .

PR #4838, opened 2018-03-21
core(start-url): reorganize start_url gatherer

paulirish:
(This is a merge into #4710) 

This is reusing the exact same new logic from 4710, just separated a bit.

Only interesting things:

1. using `isReadFailure` boolean to indicate if we should bail and fail before trying a `fetch`
1. added a timeout to make sure we're not waiting forever for a matching network request
1. more comments & jsdoc

### DIFF
--- a/lighthouse-core/gather/gatherers/start-url.js
+++ b/lighthouse-core/gather/gatherers/start-url.js
@@ -9,54 +9,71 @@ const Gatherer = require('./gatherer');
 const manifestParser = require('../../lib/manifest-parser');
 
 class StartUrl extends Gatherer {
-  readManifest(manifest) {
+  /**
+   * Grab the manifest, extract it's start_url, attempt to `fetch()` it while offline
+   * @param {*} options
+   * @return {{statusCode: number, debugString: ?string}}
+   */
+  afterPass(options) {
+    return options.driver.goOnline(options)
+      .then(() => options.driver.getAppManifest())
+      .then(response => options.driver.goOffline(options).then(() => response))
+      .then(response => response && manifestParser(response.data, response.url, options.url))
+      .then(manifest => {
+        const {isReadFailure, reason, startUrl} = this.readManifestStartUrl(manifest);
+        if (isReadFailure) {
+          return {statusCode: -1, debugString: reason};
+        }
+
+        return this.attemptManifestFetch(options, startUrl);
+      }).catch(() => {
+        return {statusCode: -1, debugString: 'Unable to fetch start URL via service worker'};
+      });
+  }
+
+  /**
+   * Read the parsed manifest and return failure reasons or the startUrl
+   * @param {Manifest} manifest
+   * @return {{isReadFailure: !boolean, reason: ?string, startUrl: ?startUrl}}
+   */
+  readManifestStartUrl(manifest) {
     if (!manifest || !manifest.value) {
       const detailedMsg = manifest && manifest.debugString;
 
       if (detailedMsg) {
-        const debugString = `Error fetching web app manifest: ${detailedMsg}`;
-        return {
-          statusCode: -1,
-          debugString,
-        };
+        return {isReadFailure: true, reason: `Error fetching web app manifest: ${detailedMsg}`};
       } else {
-        const debugString = `No usable web app manifest found on page`;
-        return {
-          statusCode: -1,
-          debugString,
-        };
+        return {isReadFailure: true, reason: `No usable web app manifest found on page`};
       }
     }
 
+    // Even if the start URL had an error, the browser will still supply a fallback URL.
+    // Therefore, we only set the debugString here and continue with the fetch.
     if (manifest.value.start_url.debugString) {
-      // Even if the start URL had an error, the browser will still supply a fallback URL.
-      // Therefore, we only set the debugString here and continue with the fetch.
-      return {
-        statusCode: -1,
-        debugString: manifest.value.start_url.debugString,
-      };
+      return {isReadFailure: true, reason: manifest.value.start_url.debugString};
     }
 
-    return {
-      statusCode: 1,
-      startUrl: manifest.value.start_url.value,
-    };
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -9,76 +9,90 @@ const Gatherer = require('./gatherer');
...
+
+  /**
+   * Try to `fetch(start_url)`, return true if fetched by SW
+   * @param {*} options
+   * @param {!string} startUrl
+   * @return {Promise<{statusCode: ?number, debugString: ?string}>}
+   */
+  _attemptManifestFetch(options, startUrl) {
+    // Wait up to 3s to get a matched network request from the fetch() to work
+    const timeoutPromise = new Promise(resolve => setTimeout(_ =>
```
wardpeet: why `setTimeout(_ =>` and not `setTimeout(() =>` not sure if we have a rule on this one 🙄

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -9,76 +9,90 @@ const Gatherer = require('./gatherer');
...
+      resolve({statusCode: -1, debugString: 'Timed out waiting for fetched start_url'})
+      , 3000));
```
patrickhulce: this is some strange indentation haha is this prettier's handiwork?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -9,76 +9,90 @@ const Gatherer = require('./gatherer');
...
+
+    const fetchPromise = new Promise(resolve => {
+      const driver = options.driver;
```
wardpeet: should we do ´attemptManifestFetch({driver}, startUrl) {´ ?
than we could nuke this line

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -9,76 +9,90 @@ const Gatherer = require('./gatherer');
...
+      driver.on('Network.responseReceived', onResponseReceived);
+
+      // Deliberately not returning, as _attemptManifestFetch resolves when we have a matched network request
+      driver.evaluateAsync(`fetch('${startUrl}')`);
```
patrickhulce: do we want to have a catch though to prevent unhandled promise rejection?

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -9,76 +9,90 @@ const Gatherer = require('./gatherer');
...
+        return this._attemptManifestFetch(options, startUrl);
+      }).catch(() => {
+        return {statusCode: -1, debugString: 'Unable to fetch start URL via service worker'};
+      });
+  }
+
+  /**
+   * Read the parsed manifest and return failure reasons or the startUrl
+   * @param {Manifest} manifest
+   * @return {{isReadFailure: !boolean, reason: ?string, startUrl: ?startUrl}}
```
patrickhulce: don't need any of the `!` anymore 🎉  @brendankenny are we going with closure or ts style `?`?

lighthouse-core/gather/gatherers/start-url.js
wardpeet: probably, was thinking the same but didn't dare to comment 😛

lighthouse-core/gather/gatherers/start-url.js
brendankenny: no more `!`! Assume everything is non-nullable and add `?` (or `|null`) if you really want something to be nullable

brendankenny: also I think you want `reason?: string, startUrl?: startUrl` here since they're optional and not null? If you want to get *real* fancy you can do
```
@return {{isReadFailure: true, reason: string}|{isReadFailure: false, startUrl: string}}
```

lighthouse-core/gather/gatherers/start-url.js
brendankenny: do we want to hold off on adding this and just have a general "add generalized timeout method to driver commands" issue? Only bring this up because we've never had reports of this timing out, so not sure if it's worth adding in various implementations of timeouts spread out over a few different gatherers

brendankenny: we've historically allowed a `_` as a short param name that's always ignored, but with tsc type checking it's flagged as an implicit `any` typed variable if the compiler can't figure out what it is, at which point it's usually easier just to switch to `()`. I'm not sure if there's any way to get typescript to ignore those cases, so we may end up just ditching it. Not sure yet :)

paulirish: i'm ok with delaying.

up till now the starturl gatherer uses `fetch('${url}').then(response => response.status)`. That's pretty direct and I wouldn't expect it to hang forever.

Now we're fetching and observing all network requests until we see one with a matching url. it's different and i'm not 100% we'll always have a match. redirects _shouldn't_ time out but ... it probably will mess up.

lighthouse-core/gather/gatherers/start-url.js
brendankenny: can this be pulled out and do
```js
return driver.evaluateAsync(`fetch('${startUrl}')`)
  .then(_ => Promise.race([fetchPromise, timeoutPromise]));
```
at the end?

lighthouse-core/gather/gatherers/start-url.js
patrickhulce: oo, oo ditch it! ditch it! ditch it! 😃

lighthouse-core/gather/gatherers/start-url.js
paulirish: > can this be pulled out and do 

goooooooood calllllll. sorry didnt understand. yah this is good.

lighthouse-core/gather/gatherers/start-url.js
paulirish: good idea! much better.

Issue #4355, opened 2018-01-25
[SEO Audits] Write documentation


rviscomi:
Some SEO audits point to existing support documentation and other external sources. We want to have documentation of all of Lighthouse’s audits in one canonical place (d.g.co/tools/lighthouse or jsdoc). Additionally, we should use these LH-specific docs to describe the exact pass/fail conditions for each audit and guidance on how to fix common failure reasons.

@kaycebasques, feel free to work with @kdzwinel on any implementation details of the audits and reach out to Lizzi for peer reviews.

**Documentation checklist:**
- [x] Document has a `<title>`
- [x] Images have `alt` text
- [x] Has a `<meta viewport>`
- [x] Page has successful HTTP status code
- [x] Document has a `<meta description>`
- [x] Document has a valid `hreflang` code
- [x] Document has a `rel=canonical`
- [x] Anchors have descriptive text
- [x] Avoids plugins
- [x] Document uses legible font sizes
- [x] Page is not blocked from indexing
- [x] [robots.txt is valid](https://github.com/GoogleChrome/lighthouse/issues/4356)
- [x] [Tap targets are sized appropriately](https://github.com/GoogleChrome/lighthouse/issues/4358)
- [ ] [JSON-LD structured data is valid](https://github.com/GoogleChrome/lighthouse/issues/4359)

This list will be updated as we add more SEO audits.

rviscomi:
@kaycebasques I see there were PRs to add docs for the [legible text](https://github.com/google/WebFundamentals/pull/5810) and [descriptive anchors](https://github.com/google/WebFundamentals/pull/5813) audits. Have these audits' descriptions also been updated to point to the new docs? Just trying to get a sense of progress on this issue.

rviscomi:
@kaycebasques looking at the checklist in the first comment, there are two items unchecked:

The [robots.txt audit](https://github.com/GoogleChrome/lighthouse/issues/4356) is done and ready for docs. Are you able to work on this?

@kdzwinel is still working on the [tap targets audit](https://github.com/GoogleChrome/lighthouse/issues/4358) but we'll also need docs for that when ready.

kaycebasques:
Yes, I can work on the robots.txt doc

rviscomi:
[...]

PR #4845, opened 2018-03-22
new_audit(robots-txt): /robots.txt validation

### LINKED ISSUE #4356: [SEO Audits] Integrate robots.txt analysis

kdzwinel:
Closes #4356 

I've run my validator on the top 1000 domains, it found issues on 39 of them: https://gist.github.com/kdzwinel/b791967eb66d0e2925ea22c8ca14233a

Example websites with errors in their robots.txt: netflix.com , chase.com
Example websites with no robots.txt: office.com
Example websites with valid robots.txt: brainly.com

Example failure:
screen shot 2018-03-22 at 16 16 44


lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,195 @@
...
+  static audit(artifacts) {
+    const {
+      status,
+      content,
+    } = artifacts.RobotsTxt;
+
+    if (!status) {
+      return {
+        rawValue: false,
+        debugString: 'Lighthouse was unable to download your robots.txt file',
```
patrickhulce: should we mark this as not applicable instead?

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,195 @@
...
+  'clean-param', 'host', // yandex
+  'request-rate', 'visit-time', 'noindex', // not officially supported, but used in the wild
+]);
+const SITEMAP_VALID_PROTOCOLS = new Set(['https:', 'http:', 'ftp:']);
+
+/**
+ * @param {!string} line single line from a robots.txt file
+ * @returns {!{directive: string, value: string}}
+ */
+function parseLine(line) {
```
patrickhulce: this function is a tad long, think we can break this up into a few subchunks of work?  maybe ~45-80 gets split out to parseDirective or something?

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,195 @@
...
+  if (directiveName === DIRECTIVE_USER_AGENT && !directiveValue) {
+    throw new Error('No user-agent specified');
+  }
+
+  if (directiveName === DIRECTIVE_ALLOW || directiveName === DIRECTIVE_DISALLOW) {
+    if (directiveValue !== '' && directiveValue[0] !== '/' && directiveValue[0] !== '*') {
+      throw new Error('Pattern should either be empty, start with "/" or "*"');
+    }
+
+    const dolarIndex = directiveValue.indexOf('$');
```
patrickhulce: nit: dollar :)

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,195 @@
...
+      } catch (e) {
+        return {
+          index: index + 1,
+          line: line,
+          error: e.message,
+        };
+      }
+
+      if (parsedLine.directive === DIRECTIVE_USER_AGENT) {
+        inGroup = true;
```
patrickhulce: once you enter a group you never leave it? man robots files are weird 😆 

a comment or two explaining how this works would be nice for us less seo-expert maintainers :)

lighthouse-core/audits/seo/robots-txt.js
kdzwinel: This will only happen when a `fetch` call fails which, AFAIK, can only happen on a network failure (or a CORS failure, but we are fetching a local resource here). IMO that's an extraordinary situation which prevents us from finishing the audit, ergo the error. Let me know if that makes sense to you!

lighthouse-core/audits/seo/robots-txt.js
kdzwinel: 💸good catch!

lighthouse-core/audits/seo/robots-txt.js
kdzwinel: I added a short comment with explanation and doc link in this code 👍 

BTW I wrote down list of rules and resources that I've used [here](https://github.com/GoogleChrome/lighthouse/issues/4356#issuecomment-375489925)

lighthouse-core/audits/seo/robots-txt.js
kdzwinel: good call, I extracted the part that was validating the directives to a separate function

lighthouse-core/audits/seo/robots-txt.js
patrickhulce: that's fair, sounds fine to me 👍

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,207 @@
...
+  DIRECTIVE_USER_AGENT, DIRECTIVE_DISALLOW, // standard
+  DIRECTIVE_ALLOW, DIRECTIVE_SITEMAP, // universally supported
+  'crawl-delay', // yahoo, bing, yandex
+  'clean-param', 'host', // yandex
+  'request-rate', 'visit-time', 'noindex', // not officially supported, but used in the wild
+]);
+const SITEMAP_VALID_PROTOCOLS = new Set(['https:', 'http:', 'ftp:']);
+
+/**
+ * @param {!string} directiveName
```
patrickhulce: no `!` necessary with our tsc typechecking anymore :)

mind adding this file to the typechecked files in `tsconfig` while we're at it?

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,207 @@
...
+        return {
+          index: index + 1,
+          line: line,
+          error: 'No user-agent specified',
+        };
+      }
+
+      return null;
+    })
+    .filter(error => error !== null);
```
patrickhulce: nit: can we rename either this variable or the `error.error` to `error.message`

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,207 @@
...
+class RobotsTxt extends Audit {
+  /**
+   * @return {!AuditMeta}
+   */
+  static get meta() {
+    return {
+      name: 'robots-txt',
+      description: 'robots.txt is valid',
+      failureDescription: 'robots.txt is not valid',
+      helpText: 'If your robots.txt file is malformed, crawlers may not be able to understand ' +
```
patrickhulce: do we have a link we can give them? perhaps your favorite validator/resource :)

kdzwinel: Since we are trying to stay crawler agnostic I think we shouldn't link to any of the official google/bing/yandex/yahoo docs here. Only independent resource that comes to mind is http://www.robotstxt.org/ but TBH it wasn't that useful to me. Proper docs for this audit are coming (#4355) to developers.google.com, so should wait for them.

patrickhulce:
@kdzwinel were you able to give the typechecking a try by chance?

kdzwinel:
@patrickhulce yeah, I'm wrestling with it (so far type checking: 1, Konrad: 0). It will be the first audit in tsconfig, so the path isn't paved yet.

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -108,44 +108,55 @@ function parseLine(line) {
...
+    .forEach((line, index) => {
       let parsedLine;
 
       try {
         parsedLine = parseLine(line);
       } catch (e) {
-        return {
-          index: index + 1,
+        errors.push({
+          index: (index + 1).toString(),
```
kdzwinel: `Audit.makeTableDetails` type definition forces all of these fields to be strings

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -108,44 +108,55 @@ function parseLine(line) {
...
+  /**
+   * @type Array<{index: string, line: string, message: string}>
+   */
+  const errors = [];
   let inGroup = false;
 
-  return content
+  content
     .split(/\r\n|\r|\n/)
-    .map((line, index) => {
```
kdzwinel: I couldn't figure out how to explain to tsc that:

```
arr.map(() => {
  if(x) { return {} }
  return null;
}
.filter(x => x !== null)
```

can't possibly return `null`s because they are filtered out, so I rewrote it using `forEach`.

kdzwinel:
@patrickhulce ok, it looks like I figured it out 👍 I had to rewrite parts of `validateRobots` function though - PTAL

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -0,0 +1,225 @@
...
+      failureDescription: 'robots.txt is not valid',
+      helpText: 'If your robots.txt file is malformed, crawlers may not be able to understand ' +
+      'how you want your website to be crawled or indexed.',
+      requiredArtifacts: ['RobotsTxt'],
+    };
+  }
+
+  /**
+   * @param {{RobotsTxt: {status: number, content: string}}} artifacts
+   * @return {LH.Audit.Product}
```
patrickhulce: heh somewhat strange to see this in real usage :)

PR #4851, opened 2018-03-22
core: add node 6 support to (patched) robots-parser

paulirish:
Our new patch leaves out node 6 out in cold:
https://travis-ci.org/GoogleChrome/lighthouse/jobs/357025749#L634

node 6.13 has `require('url').URL` but its a month old and no other node 6 has it. 
so we'll add this backwards-compat support for the rest of node 6's out there.

This PR is for branch2.x only.

Konrad has an upcoming PR to roll latest `robots-parser` (which will now be node 7+ only).

Audience-wise: this patch only matters for our Node 6 (>6.0.0 and [<6.13.0](https://github.com/nodejs/Release/issues/208#issuecomment-364697107)) CLI users

### DIFF
--- a/patches/robots-parser+1.0.2.patch
+++ b/patches/robots-parser+1.0.2.patch
@@ -1,36 +1,54 @@
 patch-package
 --- a/node_modules/robots-parser/Robots.js
 +++ b/node_modules/robots-parser/Robots.js
-@@ -1,4 +1,4 @@
+@@ -1,4 +1,22 @@
 -var libUrl   = require('url');
-+var URL   = (typeof self !== 'undefined' && self.URL) || require('url').URL;
++
++var getUrlPath;
++var parse;
++if (typeof self !== 'undefined' && self.URL) {
++	// Extra safety for invalid URLs
++	parse = url => {
++		try {
++			return new URL(url);
++		} catch(e) {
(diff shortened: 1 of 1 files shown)


patches/robots-parser+1.0.2.patch
```diff
@@ -30,7 +40,7 @@ patch-package
  	var rules = this._rules[userAgent] || this._rules['*'] || [];
  
 -	return isPathAllowed(parsedUrl.path, rules);
-+	return isPathAllowed(parsedUrl.pathname + parsedUrl.search, rules);
++	return isPathAllowed(getPath(parsedUrl), rules);
```
patrickhulce: s/getPath/getUrlPath/

paulirish:
PTAL

I'm committed to making this work. Even if it hates me. 👿 

This fixes:

1. node 6.0 - 6.12  support
1. gracefully handling invalid URLs discovered by robots-parser

Konrad's forthcoming PR will also handle the latter, whereas current master fails on it. Honestly, not a huge deal because the invalid URLs would have to be either A) the robots.txt resolved URL or B) the main resource URL. Both of these will always be valid so it's not a huge deal. Was just something the robots-parser tests caught.

PR #4855, opened 2018-03-23
tests: transpile extension test for node 6

paulirish:
I brought the pptr extension smoketest back to 2.x branch as it's a nice assurance. But node 6 doesn't like the pretty async await. This was a 10 minute transpile and it looks p decent. 

wdyt?

for 2.x branch only


patrickhulce:
other question: how much longer are we going to be shipping on 2.x? :)

paulirish:
I don't see anything upcoming that we'd ship on the 2.x branch. But if any more sentry bugs are addressed, then those would qualify.   
So I'm guessing we may have 1 more release.

Issue #686, opened 2016-09-21
"Manifest contains short_name" audit is misleading


kaycebasques:
It checks for `short_name` or `name`. Is the goal to just make sure that some name is provided for the app? In which case maybe the audit should be renamed to something more generic?

kaycebasques:
Actually, just noticed that we also have an audit for `name`...

brendankenny:
We could clarify this, but for some background, see #348 (and linked twitter thread)

Onset:
In addition `Manifest's short_name will be truncated when displayed on homescreen` is misleading too if you haven't any manifest.

paulirish:
duping into  #2698

Issue #23, opened 2016-03-17
audit: Can prompt recurring visitor to install to homescreen


paulirish:
Manual evaluation: manifest has short_name,  144x144 png icon

##### input:
* parsed manifest - #22

##### audit:
- has a registered SW
- valid start_url
- valid name
- valid short_name
- icon of size >= 144x144 and png (either type `image/png` or filename ending in `.png`)

_update_: ^  Copied from brendan's comment below.

See also "PWA validator" bookmarklet - #17

docs:  [install-to-homescreen on /web](https://developers.google.com/web/updates/2015/03/increasing-engagement-with-app-install-banners-in-chrome-for-android?hl=en)  

##### when: 
* page loaded and in idle state

paulirish:
from the docs: "Your icon declarations should include a mime type of image/png"

@brendankenny is going to verify in the source

brendankenny:
currently required:
- manifest
- valid start_url
- valid name
- valid short_name
- icon of png (either type `image/png` or filename ending in `.png`)

paulirish:
"valid". for url it's valid by chromium's gURL standards, right?   and then the two names it checks for exists and not empty?

brendankenny:
for valid URL, it does appear that's correct. For the name checks, you may technically be able to get away with empty strings (it checks they aren't null), but I'm not sure. The manifest parse step may just covert empty strings to null.

paulirish:
k cool​

paulirish:
Nearly done, but we need to check for 144px icons instead of 192px. (But having 192 icons would allow the test to pass)

paulirish:
@brendan  i have this alsmost complete in a branch on my fork. 1% battery. see you guys.

brendankenny:
To add to this, on a macbook pro, I get a "256px square icon is required" in Chrome. The 256px appears to come from it not taking the emulated screen into account, but we may need to investigate the "square" part of that.

This seems to discard non-square icons:
https://code.google.com/p/chromium/codesearch#chromium/src/chrome/browser/manifest/manifest_icon_selector.cc&rcl=1459883706&l=43

but this explicitly tests non-square is OK:
https://code.google.com/p/chromium/codesearch#chromium/src/chrome/browser/banners/app_banner_data_fetcher_unittest.cc&sq=package:chromium&type=cs&q=%22Non-square%20is%20okay%22%20file:%5Esrc/chrome/browser/banners/

[...]

PR #4860, opened 2018-03-23
core(pwa): revamp and move short_name_length audit

### LINKED ISSUE #348: manifest's "short_name" isn't required if "name" is short enough
See this thread for details: https://twitter.com/nekrtemplar/status/734483891071922182

Not sure about what is exactly "short enough", but 4 symbols seems to be.

### LINKED ISSUE #2698: short_name shouldn't be required
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/webapp-install-banner.js#L49

If the `name` property is short enough, you don't need `short_name`. The test should reflect this.

This is either a regression of https://github.com/GoogleChrome/lighthouse/issues/348, or wasn't fixed.

paulirish:
Two separate, but related changes in here:

1. webapp-install-banner audit doesn't check for a defined `short_name` any more. It's never technically been required, so we shouldn't assert it.
1. Just straight up remove short-name-length. The length of 12 characters was picked out of the air. And having a whole audit to verify `str.length <= 12` is kinda annoying. 
   * If we didn't remove it, I'd move it to PWA, set weight to 0, and flip it to `notApplicable: true` if no manifest is present. But honestly I don't think it's worth keeping it around. Open to arguments tho.

---------

Fixes #2698. See also https://github.com/GoogleChrome/lighthouse/issues/23#issuecomment-270461938
Fixes https://github.com/GoogleChrome/lighthouse/issues/686#issuecomment-339308030
Fixes #4860

Doesn't fix #348 because it makes a different conclusion.


paulirish:
Welcome back!

* Now it's in PWA section
* Weight of 0 always, since it isn't included in PWA checklist requirements
* Failure of no manifest is communicated

### screenshot where there's no manifest at all

image

### screenshot where there's a failing manifest

image

patrickhulce:
got some lint failures though

paulirish:
> aside: is it just me or could multicheckaudit and byteefficiencyaudit use a bit of a refresh cleanup post I/O :)

yUPPPPPPPPP

lighthouse-core/audits/manifest-short-name-length.js
```diff
@@ -28,26 +27,24 @@ class ManifestShortNameLength extends Audit {
...
-      const hasShortName = manifestValues.allChecks.find(i => i.id === 'hasShortName').passing;
-      if (!hasShortName) {
-        return {
-          rawValue: false,
-          debugString: 'No short_name found in manifest.',
-        };
-      }
+      const shortNameCheckIds = ['hasShortName', 'shortNameLength'];
+      manifestValues.allChecks.filter(item => shortNameCheckIds.includes(item.id)).forEach(item => {
+        if (!item.passing) failures.push(item.failureText);
```
brendankenny: won't this have the other problem where if you don't have a short name you'll get both ``'Manifest does not have `short_name`'`` and ``'Manifest `short_name` will be truncated when displayed on the homescreen'``

glittle:
As for the limit of 12, does this consider that Android word-wraps the text and shows on two lines?  So, two words, both 12 long, would be fine.

paulirish:
> does this consider that Android word-wraps the text and shows on two lines?

It doesn't, but I don't see this on my phone. Android 8.1.0 on a Pixel 2. I've never seen the text below an icon get word-wrapping. :/

glittle:
Hmm... it looks like on the desktop it does not, but in a folder it does!  I keep almost everything in folders, so didn't notice. (Sorry these screen captures are so big!)

screenshot_20180426-134730
screenshot_20180426-134953

lighthouse-core/gather/computed/manifest-values.js
```diff
@@ -64,8 +64,10 @@ class ManifestValues extends ComputedArtifact {
       },
       {
         id: 'shortNameLength',
-        failureText: 'Manifest `short_name` will be truncated when displayed on the homescreen',
-        validate: manifest => !!manifest.value.short_name.value &&
+        failureText: `Manifest's \`short_name\` is too long (>${SUGGESTED_SHORTNAME_LENGTH} ` +
+          `characters) to be displayed on a homescreen without truncation`,
+        // Pass if there's no short_name. Don't want to report a non-existent string is too long
```
JoelEinbinder: I think that this should fail if there is no short_name and the name is longer than SUGGESTED_SHORTNAME_LENGTH. Or another audit like `hasSomeShortName` should fail. I wouldn't know that `short_name` exists without this audit.

paulirish:
@patrickhulce @brendankenny ptal

brendankenny:
should update the PR description (looks like you ended up doing the opposite)

lighthouse-core/audits/manifest-short-name-length.js
```diff
@@ -26,28 +25,37 @@ class ManifestShortNameLength extends Audit {
...
+   * @return {Promise<{failures: Array<string>, manifestValues: LH.Artifacts.ManifestValues, notApplicable?: boolean}>}
    */
-  static audit(artifacts) {
+  static audit_(artifacts) {
+    /** @type {Array<string>} */
+    const failures = [];
+    /** @type {Array<string>} */
+    const warnings = [];
+
     return artifacts.requestManifestValues(artifacts.Manifest).then(manifestValues => {
```
brendankenny: await

lighthouse-core/audits/manifest-short-name-length.js
```diff
@@ -26,28 +25,37 @@ class ManifestShortNameLength extends Audit {
...
-        };
+      if (shortNameLengthCheck && !shortNameLengthCheck.passing) {
+        failures.push(shortNameLengthCheck.failureText);
       }
 
-      const isShortEnough = manifestValues.allChecks.find(i => i.id === 'shortNameLength');
-      return {
-        rawValue: !!isShortEnough && isShortEnough.passing,
-      };
+      return result;
```
brendankenny: there's no warnings and only one possible failure, so why not drop mulitcheckAudit and do like
```js
const manifestValues = await artifacts.requestManifestValues(artifacts.Manifest);
// If there's no valid manifest, this audit is not applicable
if (manifestValues.isParseFailure) {
  return {
    rawValue: true,
    notApplicable: true,
  };
}

const shortNameCheck = manifestValues.allChecks.find(i => i.id === 'hasShortName');
const shortNameLengthCheck = manifestValues.allChecks.find(i => i.id === 'shortNameLength');

// If there's no short_name present, this audit is not applicable
if (shortNameCheck && !shortNameCheck.passing) {
  return {
    rawValue: true,
    notApplicable: true,
  };
}

if (shortNameLengthCheck && !shortNameLengthCheck.passing) {
  return {
    rawValue: false,
    explanation: `Failure: ${shortNameLengthCheck.failureText}.`,
  };
}

return {
  rawValue: true,
};
```

lighthouse-core/test/audits/manifest-short-name-length-test.js
```diff
@@ -30,29 +30,30 @@ describe('Manifest: short_name_length audit', () => {
     artifacts.Manifest = null;
 
     return ManifestShortNameLengthAudit.audit(artifacts).then(result => {
-      assert.strictEqual(result.rawValue, false);
-      assert.strictEqual(result.explanation, undefined);
```
brendankenny: update test description (doesn't fail)

lighthouse-core/test/audits/manifest-short-name-length-test.js
```diff
@@ -30,29 +30,30 @@ describe('Manifest: short_name_length audit', () => {
...
+      assert.strictEqual(result.notApplicable, true);
     });
   });
 
   it('fails when an empty manifest is present', () => {
     const artifacts = generateMockArtifacts();
     artifacts.Manifest = manifestParser('{}', EXAMPLE_MANIFEST_URL, EXAMPLE_DOC_URL);
     return ManifestShortNameLengthAudit.audit(artifacts).then(result => {
-      assert.equal(result.rawValue, false);
-      assert.equal(result.explanation, 'No short_name found in manifest.');
```
brendankenny: sames

lighthouse-core/test/audits/manifest-short-name-length-test.js
```diff
@@ -30,29 +30,30 @@ describe('Manifest: short_name_length audit', () => {
...
+      assert.strictEqual(result.rawValue, true);
+      assert.strictEqual(result.notApplicable, true);
     });
   });
 
-  it('fails when a manifest contains no short_name and too long name', () => {
```
brendankenny: sames

lighthouse-core/test/audits/manifest-short-name-length-test.js
```diff
@@ -66,6 +67,8 @@ describe('Manifest: short_name_length audit', () => {
     artifacts.Manifest = manifestParser(manifestSrc, EXAMPLE_MANIFEST_URL, EXAMPLE_DOC_URL);
     return ManifestShortNameLengthAudit.audit(artifacts).then(result => {
       assert.equal(result.rawValue, false);
+      assert.ok(result.explanation.includes('without truncation'), result.explanation);
+      assert.ok(!result.explanation.includes('Manifest does not have `short_name`'));
```
brendankenny: delete this line as it's not a possibility any longer? Want to assert not not applicable, though?

patrickhulce:
:shipit:

noyainrain:
Thanks for fixing this @paulirish :blush: 

I know it's been over a year, just a quick question related to this change. From the description of the PR, the original goal was to remove the `hasShortName` check:

>     1. webapp-install-banner audit doesn't check for a defined `short_name` any more. It's never technically been required, so we shouldn't assert it.

If I understand the discussion correctly, it was then agreed to make it **informative** instead:

> I'd propose turning it informative and marking not applicable when no manifest rather than killing it off entirely.

But in the end, the check still is mandatory:

```js
      // Technically shortname isn't required (if name is defined):
      //   https://cs.chromium.org/chromium/src/chrome/browser/installable/installable_manager.cc?type=cs&q=IsManifestValidForWebApp+f:cc+-f:test&sq=package:chromium&l=473
      // Despite this, we think it's better to require it anyway.
      // short_name is preferred for the homescreen icon, but a longer name can be used in
      // the splash screen and app title. Given the different usecases, we'd like to make it clearer
      // that the developer has two possible strings to work with.
      'hasShortName',
```

Is there a particular reason for keeping it required in the `installable-manifest` audit, instead of making it informative somehow?

(Problem background: Most PWA directories use the result of lighthouse `pwa-installable` to inform the user if a PWA is installable. For a manifest with a short `name` and no (redundant) `shortname` this produces a false-negative.)

PR #4875, opened 2018-03-26
core: update robots-parser, extension: use browser-specific impl of url library

### LINKED ISSUE #4825: Clean up patch-package hack after robots-parser update
Clean up after #4819

1. Open a PR against robots-parser
1. ⏳
1. Update robots-parser in LH
1. Have a browserify transform that swaps `require('url')`  with url-shim
1. Remove patch-package (+ robots-parser patch and a postinstall script)

kdzwinel:
- patch-package removed
- robots-parser updated from 1.0.2 to [2.0.1](https://github.com/samclarke/robots-parser#version-210)
- ~browserify transform added to replace `require('url')` with `require('path/to/url-shim')`~
- allow browserify to do its default thing and replace `require('url')` with a browser-specific implementation

Closes #4825


lighthouse-extension/gulpfile.js
```diff
@@ -100,6 +100,8 @@ gulp.task('chromeManifest', () => {
 function applyBrowserifyTransforms(bundle) {
   // Fix an issue with imported speedline code that doesn't brfs well.
   return bundle.transform('./fs-transform', {global: true})
+  // Replace url module with url-shim
+  .transform('./url-transform', {global: true})
```
patrickhulce: is it not possible to get the `bundle.require` style working like we do for audits/gatherers? I don't know if it's possible really just curious :)

kdzwinel:
@patrickhulce oh my, things got interesting. Your question got me to dig deeper into browserify docs and it turns out that "[browserify provides many browser-specific implementations of node core libraries](https://github.com/browserify/browserify-handbook#builtins)" and that list includes… url module. So why it didn't work for us?

https://github.com/GoogleChrome/lighthouse/blob/f7efaa5468f0ab939259d3b742930125119c0957/lighthouse-extension/gulpfile.js#L124

🙀Turns out that removing that one line is THE fix that we should have applied in the first place. Oh well, at least I learned about patch-package and browserify transforms 👨‍🎓

patrickhulce:
> Turns out that removing that one line is THE fix that we should have applied in the first place.

😱 nice sleuthing though! does this mean we can nuke the patch in this PR? I don't see it removed

kdzwinel:
@patrickhulce good catch, removed 🗑

PR #4910, opened 2018-04-02
core(lantern): cleanup Simulator construction

patrickhulce:
~~final touches waiting on #4894 but mostly independent~~ touched up!

this PR cleans up the creation of Simulator objects and their usage by creating a new computed artifact that reads throttling info from the new `settings` object


patrickhulce:
@brendankenny thoughts here?

lighthouse-core/runner.js
```diff
@@ -182,7 +182,7 @@ class Runner {
     let promise = Promise.resolve();
     for (const auditDefn of opts.config.audits) {
       promise = promise.then(_ => {
-        return Runner._runAudit(auditDefn, artifacts).then(ret => auditResults.push(ret));
+        return Runner._runAudit(auditDefn, artifacts, opts).then(ret => auditResults.push(ret));
```
brendankenny: re: conversation on settings === artifact settings, not sure of the best place, but rather than buried in the specific audits that use them (since all audits have access to settings) could be a few lines above here after `artifacts` is defined?

PR #4914, opened 2018-04-03
core(tsc): add type checking to remote protocol commands

brendankenny:
This PR adds type checking/inference for `Driver.sendCommand` (to go along with typing events in #4886).

command-completion-gif

Like #4886, this PR has changes to `scripts/extract-crdp-mapping.js` to extract all the protocol commands which are then written to `typings/crdp-mapping.d.ts`. Again, hopefully most of this can be upstreamed into `devtools-protocol`. This time the types need both the options `params` object (if any) associated with the command but also the command's return type (if any).

The other set of changes are in `Driver`, `Connection`, and `ExtensionConnection` (everything else is just minor fixes for any type errors that popped up after all the above changes). This time `sendCommand` is pulled out of the class bodies for stricter typing. Getting the typing correct is a little harder than it was for the events because this time we aren't tying together an event name and payload type, but two function parameters and a return type.

Typescript types (in typescript or jsdoc) can express that one property of an object is defined only for certain values of a different property (or similarly, for an event payload in a callback and an event name), but typescript can't yet express that a function parameter is not to be provided depending on the value of another parameter (there's a [proposal to enable this](https://github.com/Microsoft/TypeScript/issues/5453), which is now on the typescript roadmap, but under the "Future" heading with no planned implementation schedule).

As a result, the solution is a little more awkward, but not too bad. The function is defined with somewhat loose argument types, but when added to the class prototype it's cast to a much stricter type as the external interface. This allows casting with an overloaded type signature, which can create a one argument version (for commands like `DOM.enable`) or a two argument version so options can be passed (for e.g. `DOM.querySelector`), depending on the first argument (the command string).

One last wrinkle: for commands that have all optional options that we don't use, we have often just dropped the second argument when calling (e.g. for `Network.enable`). I've added `void` as one option for the `params` object for those commands, so they actually show up in both the one and two argument versions of `sendCommand`

### DIFF
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -11,12 +11,14 @@ const LHError = require('../../lib/errors');
 
 /**
  * @typedef {LH.StrictEventEmitter<{'protocolevent': LH.Protocol.RawEventMessage}>} CrdpEventMessageEmitter
+ * @typedef {LH.CrdpCommands[keyof LH.CrdpCommands]['paramsType']} CommandParamsTypes
+ * @typedef {LH.CrdpCommands[keyof LH.CrdpCommands]['returnType']} CommandReturnTypes
  */
 
 class Connection {
   constructor() {
     this._lastCommandId = 0;
-    /** @type {Map<number, {resolve: function(Promise<*>), method: string, options: {silent?: boolean}}>}*/
+    /** @type {Map<number, {resolve: function(Promise<CommandReturnTypes>), method: keyof LH.CrdpCommands, options: {silent?: boolean}}>} */
     this._callbacks = new Map();
 
     /** @type {?CrdpEventMessageEmitter} */
@@ -42,23 +44,6 @@ class Connection {
    */
   wsEndpoint() {
     return Promise.reject(new Error('Not implemented'));
-  }
-
-  /**
-   * Call protocol methods
-   * @param {string} method
-   * @param {object=} params
-   * @param {{silent?: boolean}=} cmdOpts
-   * @return {Promise<*>}
-   */
-  sendCommand(method, params = {}, cmdOpts = {}) {
-    log.formatProtocol('method => browser', {method, params}, 'verbose');
-    const id = ++this._lastCommandId;
-    const message = JSON.stringify({id, method, params});
-    this.sendRawMessage(message);
-    return new Promise(resolve => {
-      this._callbacks.set(id, {resolve, method, options: cmdOpts});
-    });
   }
 
   /* eslint-disable no-unused-vars */
@@ -95,6 +80,8 @@ class Connection {
     if (callback) {
       this._callbacks.delete(object.id);
 
+      // @ts-ignore since can't convince compiler that callback.resolve's return
+      // type and object.result are matching since only linked by object.id.
       return callback.resolve(Promise.resolve().then(_ => {
         if (object.error) {
           const logLevel = callback.options.silent ? 'verbose' : 'error';
@@ -111,7 +98,7 @@ class Connection {
       // just log these occurrences.
       const error = object.error && object.error.message;
       log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
-          {method: object.method, params: error || object.result}, 'verbose');
+          {method: 'UNKNOWN', params: error || object.result}, 'verbose');
     }
   }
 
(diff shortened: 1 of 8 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -656,11 +637,9 @@ class Driver {
       return this._isolatedExecutionContextId;
     }
 
-    /** @type {LH.Crdp.Page.GetResourceTreeResponse} */
```
brendankenny: typed responses lets us get rid of a lot of these explicit types :)

lighthouse-core/lib/emulation.js
```diff
@@ -86,7 +86,7 @@ function enableNexus5X(driver) {
     // Network.enable must be called for UA overriding to work
     driver.sendCommand('Network.enable'),
     driver.sendCommand('Network.setUserAgentOverride', NEXUS5X_USERAGENT),
-    driver.sendCommand('Emulation.setTouchEmulationEnabled', {
+    driver.sendCommand('Page.setTouchEmulationEnabled', {
```
brendankenny: first real bug caught with type checks? Haven't confirmed that it was a problem, but there is both [`Emulation.setTouchEmulationEnabled`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation#method-setTouchEmulationEnabled) and [`Page.setTouchEmulationEnabled`](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-setTouchEmulationEnabled), but only the `Page` one takes these params

typings/protocol.d.ts
```diff
@@ -20,18 +20,30 @@ declare global {
...
      * responses.
      */
     export type RawMessage = RawCommandMessage | RawEventMessage;
+
+    /**
+     * Strict overloaded typing for calling protocol methods with and without
+     * params object.
+     */
+    export type SendCommand = {
+      <C extends VoidParamsKeys>(method: C, params?: void, cmdOpts?: {silent?: boolean}): Promise<CrdpCommands[C]['returnType']>;
```
patrickhulce: a little unfortunate that types make us do an `extends` and we can't do `in`

lighthouse-core/gather/driver.js
```diff
@@ -809,11 +785,9 @@ class Driver {
    * @return {Promise<Array<Element>>} The found elements, or [], resolved in a promise
    */
   async querySelectorAll(selector) {
-    /** @type {LH.Crdp.DOM.GetDocumentResponse} */
-    const documentResponse = await this.sendCommand('DOM.getDocument');
+    const documentResponse = await this.sendCommand('DOM.getDocument', {});
```
patrickhulce: I thought the `| void` would let us omit the parameters if they're all optional?

lighthouse-core/gather/connections/connection.js
```diff
@@ -111,7 +98,7 @@ class Connection {
       // just log these occurrences.
       const error = object.error && object.error.message;
       log.formatProtocol(`disowned method <= browser ${error ? 'ERR' : 'OK'}`,
-          {method: object.method, params: error || object.result}, 'verbose');
+          {method: 'UNKNOWN', params: error || object.result}, 'verbose');
```
patrickhulce: do we not get back a method from disowned messages? seems like `|| 'UNKOWN'` would work if we do :)

lighthouse-core/gather/driver.js
brendankenny: whoops, yes, those were from an earlier version of the PR

lighthouse-core/gather/connections/connection.js
brendankenny: > do we not get back a method from disowned messages?

we can check on the Chrome side if it's possible, but I did a bunch of runs watching the properties of all messages sent back over the protocol and never got a `method` on a non-event response

lighthouse-core/lib/emulation.js
paulirish: ha! yeah I think this might have been a small bug that normally didn't matter.
I think we want [`Emulation.setEmitTouchEventsForMouse`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation#method-setEmitTouchEventsForMouse) instead.

that's what devtools uses when you toggle device type:
image

lighthouse-core/scripts/extract-crdp-mapping.js
```diff
@@ -147,39 +178,205 @@ function getEventMap(sourceRoot, domainName) {
...
+  });
+}
+
+/**
+ * Returns a Map of events to params and return type for given domain.
+ * @param {ts.Node} sourceRoot
+ * @param {string} domainName
+ * @return {Map<string, {paramsType: string, returnType: string}>}
+ */
+function getCommandMap(sourceRoot, domainName) {
```
paulirish: do you want to upstream all this into `devtools-protocol`?

brendankenny: > do you want to upstream all this into devtools-protocol?

the event map (just `'Domain.eventName': Crdp.EventPayloadType` I think could easily live upstream, it's so simple. The command map seems much more specialized to our use case, so I'm not sure. We could propose it to everyone involved in https://github.com/ChromeDevTools/devtools-protocol/pull/90 and see what they think

paulirish: cc @@JoelEinbinder whatchu think homeboy

brendankenny: he sounded 👍 for upstreaming (if not for using in pptr)

brendankenny:
yah?

lighthouse-core/lib/emulation.js
paulirish: seems like you're fixing this in a separate PR?

can you write up the list of changes that are on their way?

brendankenny: sorry, I can fix this :)

brendankenny:
> can you write up the list of changes that are on their way?

I believe the only (direct) next steps are to
- upstream `CrdpEvents` and `CrdpCommands` to `devtools-protocol`
- delete our local `scripts/extract-crdp-mappings.js` and `typings/crdp-mappings.d.ts` and remove `vscode-chrome-debug-core` as a dep
- switch over to `devtools-protocol` for typing the protocol

brendankenny:
@patrickhulce also gave the 👍

PR #4926, opened 2018-04-04
core(network-recorder): fix typo in once() call on super

brendankenny:
Introduced in #4918, `networkRecorder.once()` calls are being rerouted to `networkRecorder.on()`, which isn't great :)


brendankenny:
I think yeah, just leaking. The listeners just listen forever, equivalent if the original call sites had accidentally called `on`.

brendankenny:
@patrickhulce: @paulirish said you touched this last so you have to approve :P

PR #4932, opened 2018-04-05
core(fast-config): bring back a11y and SEO categories

patrickhulce:
We want a11y and SEO for fast-mode LR on 2.x now that we're not prioritizing PSI-parity speed.


patrickhulce:
yo @paulirish @brendankenny, @vinamratasingal wants this one bad for our hungry users :)

can we get a quick review?

PR #4933, opened 2018-04-05
core(lantern): more flexible graph edge creation

### LINKED ISSUE #4891: Appveyor smoke tests are flaky
Appveyor only passes the byte-efficiency smoke tests ~80% of the time (2 of last 10 builds on master have failed, and anecdotally it feels like lots of my PRs get red status checks).

It's going to be difficult to get consistent results on appveyor since we discovered how crazy variable their CPU performance is, we might want to loosen our expectations on smokehouse just in appveyor/run multiple times/adjust estimate if we detect very slow CPU/etc.

patrickhulce:
closes #4891 

this should fix the appveyor flakiness we've been seeing which was 3-pronged

* generic variability on the image wastedMs assertions, if Chrome doesn't load the images fast enough the graph won't necessarily have them loading at the same time either. adjusting some of the expectations fixes this
* appveyor is *helllllla* slow already and throttling made this even worse, CPU throttling is now disabled for appveyor
* appveyor is so slow that it takes too long for chrome to mark that a network request finished such that the script may have already started executing and the edge was excluded from the graph. the lantern side fix addresses this along with some tests

also I made smokehouse a bit more debuggable by using output-path for the JSON instead of stdout and adding a ENV variable to print stdout


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -43,19 +44,30 @@ function resolveLocalOrCwd(payloadPath) {
...
     '--quiet',
     '--port=0',
   ];
 
+  if (isDebug || process.env.SMOKEHOUSE_DEBUG) {
+    args.push('-GA');
+  }
+
+  if (process.env.APPVEYOR) {
+    // Appveyor is hella slow already, disable CPU throttling so we're not 16x slowdown
```
brendankenny: maybe link to #4891?

patrickhulce: 👍 done

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -43,19 +44,30 @@ function resolveLocalOrCwd(payloadPath) {
...
  * @return {!LighthouseResults}
  */
-function runLighthouse(url, configPath) {
+function runLighthouse(url, configPath, isDebug) {
   const command = 'node';
   const args = [
     'lighthouse-cli/index.js',
     url,
     `--config-path=${configPath}`,
+    `--output-path=smokehouse.report.json`,
```
paulirish: I often run `yarn smokehouse`  and `yarn smoke` simultaneously which could conflict in this scenario. 

also my smokehouse revamp PR means we'll have a lot of these running at the same time.

how about something like:  
```js
const filename = `smokehouse${Math.round(Math.random()*1000))}.report.json`;
```

PR #4936, opened 2018-04-06
extension: Pass flags from options in runLighthouseForConnection

jianzhoufeng:
@patrickhulce 
We pass flags like disableDeviceEmulation from lightrider in options.flags, which used to work. It is broken now due to "unify config and CLI settings".

PTAL.

Thanks!

### DIFF
--- a/lighthouse-extension/app/src/lighthouse-background.js
+++ b/lighthouse-extension/app/src/lighthouse-background.js
@@ -22,10 +22,10 @@ const log = require('lighthouse-logger');
 window.runLighthouseForConnection = function(
     connection, url, options, categoryIDs,
     updateBadgeFn = function() { }) {
-  const config = options && options.fastMode ? new Config(fastConfig) : new Config({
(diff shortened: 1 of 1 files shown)


patrickhulce:
oh are you trying to roll off of master @jianzhoufeng? there have been quite a few breaking changes which is why we've been trying to keep the [branch2.x](https://github.com/GoogleChrome/lighthouse/tree/branch2.x) compatible, thanks for the PR though! this seems good either way

jianzhoufeng:
@patrickhulce 
Oh, I am not aware of there is a branch2.x. Yes, I just did a roll for master. 
It looks good so far. Will there be more breaking changes in master?

patrickhulce:
Yes there are few more coming (some in #4333) along with quite a few output structure changes, there are discussion docs for, sorry we didn't keep you updated! the comms have been scattered in different channels

jianzhoufeng:
I see. Next time I will keep this in mind. Thanks!

PR #4945, opened 2018-04-10
cli(chrome-launcher): remove chrome-launcher forwarding stubs

brendankenny:
Part of #4333, the long awaited followup to #3410 :)

With 3.0 coming up, we can get rid of any remaining chrome-launcher stubs (forwarded to the `chrome-launcher` dep) previously maintained to be compatible with various docs and tutorials we put out in the past.

`manual-chrome-launcher.js` (used by `chrome-debug` and `yarn chrome`) is moved to `scripts/`.


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "./lighthouse-core/index.js",
   "bin": {
     "lighthouse": "./lighthouse-cli/index.js",
-    "chrome-debug": "./chrome-launcher/manual-chrome-launcher.js"
+    "chrome-debug": "./lighthouse-core/scripts/manual-chrome-launcher.js"
```
paulirish: one option:

we remove this from the lh package and setup the bin over in chrome-launcher.

brendankenny: Since we have `yarn chrome` now I'm fine removing `chrome-debug`, but I also don't install lighthouse globally anyways. It probably makes sense to not install with lighthouse? cc @patrickhulce

patrickhulce: I'm fine either way, I don't use chrome-debug these days

brendankenny: after discussion, `chrome-debug` is [still useful](https://github.com/GoogleChrome/lighthouse/blob/9823a6579e59b8ea86620bf3dec59a7fd66110c4/docs/readme.md#testing-on-a-site-with-authentication) for testing a site requiring authentication with a global lighthouse install, so leaving for now

PR #4949, opened 2018-04-10
core(tsc): add type checking to asset-saver

brendankenny:
another step on the way to typing the LHR pipeline, but was self contained enough that I split it out for easier review. Adds types to `asset-saver.js` and a few places changes due to that touch, like e.g. making a type alias for `LH.DevtoolsLog` (instead of (`Array<LH.Protocol.RawEventMessage>` :).

`asset-saver.js` is mostly the same except for indenting, so probably easier to review with [?w=1](https://github.com/GoogleChrome/lighthouse/pull/4949/files?w=1).

With the use of object spread, I moved our minimum node engine to 8.9 (the [first Node 8 LTS release](https://nodejs.org/en/blog/release/v8.9.0/)) and bumped our eslint to support what it calls `ecmaVersion` 2018.

### DIFF
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -129,7 +129,13 @@ async function saveArtifacts(artifacts, basePath) {
   rimraf.sync(`${basePath}/*${traceSuffix}`);
   rimraf.sync(`${basePath}/${artifactsFilename}`);
 
-  const {traces, devtoolsLogs, ...restArtifacts} = artifacts;
+  // TODO: when https://github.com/GoogleChrome/lighthouse/issues/4952 is fixed
+  // const {traces, devtoolsLogs, ...restArtifacts} = artifacts;
+  const traces = artifacts.traces;
+  const devtoolsLogs = artifacts.devtoolsLogs;
+  const restArtifacts = Object.assign({}, artifacts);
+  delete restArtifacts.traces;
+  delete restArtifacts.devtoolsLogs;
 
   // save traces
   for (const [passName, trace] of Object.entries(traces)) {
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -3,7 +3,6 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
-// @ts-nocheck
 'use strict';
 
 const fs = require('fs');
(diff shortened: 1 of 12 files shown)


typings/externs.d.ts
```diff
@@ -102,8 +102,17 @@ declare global {
...
+     * A record of DevTools Debugging Protocol events.
+     */
+    export type DevtoolsLog = Array<Protocol.RawEventMessage>;
+
     export interface Trace {
       traceEvents: TraceEvent[];
+      metadata?: {
+        'cpu-family'?: number;
+      };
+      [futureProps: string]: any;
```
paulirish: do we need this added? afaik there aren't any other toplevel items in a trace than these two.

typings/externs.d.ts
```diff
@@ -102,8 +102,17 @@ declare global {
...
 
+    /**
+     * A record of DevTools Debugging Protocol events.
+     */
+    export type DevtoolsLog = Array<Protocol.RawEventMessage>;
+
     export interface Trace {
       traceEvents: TraceEvent[];
+      metadata?: {
+        'cpu-family'?: number;
```
paulirish: just curious why you called out this but not the rest..

paulirish: also... worth pointing out that this metadata object has some really useful stuff..

image

lighthouse-core/lib/asset-saver.js
```diff
@@ -104,91 +112,85 @@ function loadArtifacts(basePath) {
...
-/* istanbul ignore next */
-function saveArtifacts(artifacts, basePath) {
+async function saveArtifacts(artifacts, basePath) {
   mkdirp.sync(basePath);
   rimraf.sync(`${basePath}/*${traceSuffix}`);
   rimraf.sync(`${basePath}/${artifactsFilename}`);
 
-  // We don't want to mutate the artifacts as provided
-  artifacts = Object.assign({}, artifacts);
+  const {traces, devtoolsLogs, ...restArtifacts} = artifacts;
```
paulirish: So hot. 🔥

brendankenny:
object spread support is currently blocked (and making the tests fail) by brfs. I've filed an [upstream bug](https://github.com/browserify/brfs/issues/87) but maybe I'll back out the spread line for now

typings/externs.d.ts
brendankenny: we support saving `metadata` in `asset-saver`, so we want to put it in there, but we don't really use it anywhere and I don't know what's actually required in there vs optional. However if we just put a `metadata?: {}` it treats it as `any`. Putting a single property triggers excess property checking, so could signal "hey, add some types if you want to use this". Could also set to `never` or something for the same effect...

typings/externs.d.ts
brendankenny: > do we need this added?

we support [outputting arbitrary keys in the trace object](https://github.com/GoogleChrome/lighthouse/blob/aec55bd59b39aeca7a84f51b760fcd4626897b25/lighthouse-core/lib/asset-saver.js#L211-L218) in `saveTrace()` and `logAssets()`, which seems worth keeping?

But then tsc needs to know there are other properties to loop over in the first place, so it needs to be either ts-ignored in place or a generic "every other property name" entry needs to be added here.

So...what do you think we should do

paulirish: ah ok. well that's fine. can you just leave a comment that this is for our flexible traceWriter method?

typings/externs.d.ts
paulirish: lol ok

paulirish:
lgtm

PR #4950, opened 2018-04-10
core: de-dupe URLs in is-on-http, uses-http2

patrickhulce:
should fix our recent CI flakes too :)

### DIFF
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -35,13 +35,13 @@ module.exports = [
         extendedInfo: {
           value: {
             results: {
-              length: 18,
+              length: '>15',
             },
           },
         },
         details: {
           items: {
-            length: 18,
+            length: '>15',
           },
         },
       },
--- a/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
+++ b/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
@@ -35,13 +35,13 @@ module.exports = [
         extendedInfo: {
           value: {
             results: {
-              length: '>15',
+              length: 17,
             },
           },
         },
         details: {
           items: {
-            length: '>15',
+            length: 17,
           },
         },
       },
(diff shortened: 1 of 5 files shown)


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -35,13 +35,13 @@ module.exports = [
         extendedInfo: {
           value: {
             results: {
-              length: 18,
```
paulirish: i kinda like being so explicit, though ideally in failure we see exactly where the difference is.

can we avoid dropping back to `>15` ?

lighthouse-core/audits/dobetterweb/uses-http2.js
```diff
@@ -47,11 +48,14 @@ class UsesHTTP2Audit extends Audit {
...
       }).map(record => {
+        if (urls.has(record._url)) return;
+        urls.add(record._url);
+
         return {
           protocol: record.protocol,
-          url: record.url, // .url is a getter and not copied over for the assign.
+          url: record._url,
         };
-      });
```
paulirish: instead of this dance should we just do this instead:

```js
const uniqueUrls = Array.from(new Set(resources));
```

lighthouse-core/audits/is-on-https.js
```diff
@@ -47,30 +47,28 @@ class HTTPS extends Audit {
...
-      }));
+      const items = Array.from(new Set(insecureURLs)).map(url => ({url}));
 
       const headings = [
         {key: 'url', itemType: 'url', text: 'Insecure URL'},
       ];
 
       return {
-        rawValue: insecureRecords.length === 0,
+        rawValue: insecureURLs.length === 0,
```
paulirish: why not `items`

lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
patrickhulce: :( alright

lighthouse-core/audits/dobetterweb/uses-http2.js
brendankenny: > const uniqueUrls = Array.from(new Set(resources));

unfortunately won't dedupe because it's an object, so they'll all be unique.

We could add it to the filter instead, the Boolean thing is pretty ugly :)

patrickhulce: we're moving this to replace `.filter(Boolean)`

lighthouse-core/audits/is-on-https.js
patrickhulce: sure, doesn't matter :)

patrickhulce:
how's it lookin' now @paulirish ?

PR #4958, opened 2018-04-12
Runtime error encountered: Cannot change settings between gathering and auditing

wardpeet:
I couldn't run lighthouse-cli with -A anymore. I would always get the following error:
```
Runtime error encountered: Cannot change settings between gathering and auditing
Error: Cannot change settings between gathering and auditing
```
It seems that lodash isEqual checks the length of our objects which are not the same. The saved artifact removes all undefined values as it uses JSON.stringify. I convert all undefined values to null so it keeps the keys intact.


wardpeet:
Sounds good. Feel free to close :)

patrickhulce:
cool beans, it's over in #4960

PR #4960, opened 2018-04-12
core(config): clean flags for settings

patrickhulce:
alternative to #4958

this brings back -A, it didn't come out quite as clean as I had hoped, but it is typesafe which is kinda nice :)

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -246,8 +246,24 @@ function expandArtifacts(artifacts) {
   return artifacts;
 }
 
+/**
+ * @param {LH.Flags=} flags
+ * @return {Partial<LH.Config.Settings>}
+ */
+function cleanFlagsForSettings(flags = {}) {
+  const settings = {};
+  for (const key of Object.keys(flags)) {
+    if (typeof constants.defaultSettings[key] !== 'undefined') {
+      settings[key] = flags[key];
+    }
+  }
+
+  return settings;
+}
+
 function merge(base, extension) {
-  if (typeof base === 'undefined') {
+  // If the default value doesn't exist or is explicitly null, defer to the extending value
+  if (typeof base === 'undefined' || base === null) {
     return extension;
   } else if (Array.isArray(extension)) {
     if (!Array.isArray(base)) throw new TypeError(`Expected array but got ${typeof base}`);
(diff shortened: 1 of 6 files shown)


typings/config.d.ts
```diff
@@ -24,11 +24,11 @@ declare global {
         settings?: SettingsJson;
         passes?: PassJson[];
       }
-
```
patrickhulce: someone's got a diff trailing space editor setting it seems :)

lighthouse-core/config/config.js
```diff
@@ -246,8 +246,24 @@ function expandArtifacts(artifacts) {
   return artifacts;
 }
 
+/**
+ * @param {LH.Flags=} flags
+ * @return {Partial<LH.Config.Settings>}
+ */
+function cleanFlagsForSettings(flags = {}) {
```
paulirish: semantically this is `removedUndefinedProperties` right?

could we call it that instead?

patrickhulce: not sure I'm following you, this is looking at what properties are not on `constants.defaultSettings` and dropping them from flags regardless of their value.

the JSON.parse(JSON.stringify()) approach would address the specific undefined yargs case but not an underlying problem that having miscellaneous junk present that isn't part of settings throws off our isDeepEqual later

paulirish: oh got it. 

>  what properties are not on constants.defaultSettings

ahhh.. mmhmmm. 

k lgtm

lighthouse-core/config/constants.js
```diff
@@ -26,10 +26,24 @@ const throttling = {
   },
 };
 
+/** @type {LH.Config.Settings} */
```
brendankenny: 👍

typings/config.d.ts
brendankenny: ah, damn, eslint isn't checking .ts files, obvs. I'll find the setting :P

lighthouse-core/config/config.js
```diff
@@ -246,8 +246,24 @@ function expandArtifacts(artifacts) {
   return artifacts;
 }
 
+/**
+ * @param {LH.Flags=} flags
```
brendankenny: maybe add a function description that this is dropping any flag not found on settings?

PR #4962, opened 2018-04-13
report(bootup-time): fix learn more link

underbyte:

PR #4964, opened 2018-04-13
tests: exclude audit helpText from 'yarn diff:sample-json' assertion

paulirish:
tested on the initial #4962


paulirish:
> is it just cause it's annoying to get an extra file in just a helpText PR? 

yeah. adds way more friction to the person who wants to do a driveby `patch-1` PR.

> I'm fine with this, but it seems like it's just going to cause unrelated PRs to get the helpText changes dumped in there instead :)

I'm expecting the helptext in the golden json to be updated whenever the next PR that changes the JSON is submitted.  Don't feel like it'll be anything additional

PR #4965, opened 2018-04-13
core(screenshots): align filmstrip to observed metrics

patrickhulce:
followup fix as promised in #4951 

(easier with https://github.com/GoogleChrome/lighthouse/pull/4965/files?w=1)

### DIFF
--- a/lighthouse-core/audits/screenshot-thumbnails.js
+++ b/lighthouse-core/audits/screenshot-thumbnails.js
@@ -72,12 +72,17 @@ class ScreenshotThumbnails extends Audit {
 
     const speedline = await artifacts.requestSpeedline(trace);
 
-    let alternativeMaximumValue = 0;
+    let minimumTimelineDuration = 0;
+    // Ensure thumbnails cover the full range of the trace (TTI can be later than visually complete)
     if (context.settings.throttlingMethod !== 'simulate') {
       const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
       const metricComputationData = {trace, devtoolsLog, settings: context.settings};
       const ttci = artifacts.requestConsistentlyInteractive(metricComputationData);
-      alternativeMaximumValue = (await ttci.catch(() => ({timing: 0}))).timing;
+      try {
+        minimumTimelineDuration = (await ttci).timing;
+      } catch (_) {
+        minimumTimelineDuration = 0;
+      }
     }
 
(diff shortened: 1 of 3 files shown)


lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -67,62 +66,70 @@ class ScreenshotThumbnails extends Audit {
...
-          timestamp: targetTimestamp * 1000,
-          data: base64Data,
+    const speedline = await artifacts.requestSpeedline(trace);
+
+    let alternativeMaximumValue = 0;
+    if (context.settings.throttlingMethod !== 'simulate') {
+      const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
+      const metricComputationData = {trace, devtoolsLog, settings: context.settings};
+      const ttci = artifacts.requestConsistentlyInteractive(metricComputationData);
+      alternativeMaximumValue = (await ttci.catch(() => ({timing: 0}))).timing;
```
brendankenny: ha, I'm not sure I'm wild about this approach. try/catch is ugly, but this makes it look like it's doing something much more complicated than it is :S

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -67,62 +66,70 @@ class ScreenshotThumbnails extends Audit {
...
-          jpeg.encode(thumbnailImageData, 90).data.toString('base64');
-
-        cachedThumbnails.set(frameForTimestamp, base64Data);
-        thumbnails.push({
-          timing: Math.round(targetTimestamp - speedline.beginning),
-          timestamp: targetTimestamp * 1000,
-          data: base64Data,
+    const speedline = await artifacts.requestSpeedline(trace);
+
+    let alternativeMaximumValue = 0;
```
brendankenny: add a comment why there needs to be an `alternativeMaximumValue`? (there's a comment below but maybe some variant should be up here so you don't spend too long thinking about it :)

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: fixed :P

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: yeah good call, done

PR #4966, opened 2018-04-13
core(metrics): add first CPU idle lantern metric

patrickhulce:
I realized that we can realistically carry over all of our existing metrics in lantern world, and really FCPUI is the only one that I didn't have an impl of yet

this just reuses the graphs and coefficients of TTI combined with the existing algorithm for first interactive that does shortening window size/task clusters/etc

by reusing the graphs and coefficients of TTI we will also always give consistent answers on TTI/TTFCPUI because in the worst case, the start of the quiet window will be the last long task (which is equal to TTI estimate)

accuracy is really quite good, spearman's rho of ~0.82, which is slightly worse than TTI but better than FCP/FMP. MAPE of ~30%, which is nearly identical to TTI absolute error

### DIFF
--- a/lighthouse-core/gather/computed/metrics/lantern-first-cpu-idle.js
+++ b/lighthouse-core/gather/computed/metrics/lantern-first-cpu-idle.js
@@ -1,0 +1,56 @@
+/**
+ * @license Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Node = require('../../../lib/dependency-graph/node');
+const CPUNode = require('../../../lib/dependency-graph/cpu-node'); // eslint-disable-line no-unused-vars
+const NetworkNode = require('../../../lib/dependency-graph/network-node'); // eslint-disable-line no-unused-vars
+
+const FirstInteractive = require('../first-interactive');
+const LanternConsistentlyInteractive = require('./lantern-consistently-interactive');
+
+class FirstCPUIdle extends LanternConsistentlyInteractive {
+  get name() {
+    return 'LanternFirstCPUIdle';
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/computed/metrics/lantern-first-cpu-idle.js
```diff
@@ -0,0 +1,55 @@
...
+'use strict';
+
+const Node = require('../../../lib/dependency-graph/node');
+const CPUNode = require('../../../lib/dependency-graph/cpu-node'); // eslint-disable-line no-unused-vars
+const NetworkNode = require('../../../lib/dependency-graph/network-node'); // eslint-disable-line no-unused-vars
+
+const FirstInteractive = require('../first-interactive');
+const LanternConsistentlyInteractive = require('./lantern-consistently-interactive');
+
+class FirstCPUIdle extends LanternConsistentlyInteractive {
```
brendankenny: > this just reuses the graphs and coefficients of TTI combined with the existing algorithm for first interactive

one downside to this approach is it's getting difficult to layer all three (four?) of these in my head to follow along :)

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
```diff
@@ -0,0 +1,28 @@
...
+    const result = await artifacts.requestLanternFirstCPUIdle({trace, devtoolsLog});
+
+    assert.equal(Math.round(result.timing), 5308);
+    assert.equal(Math.round(result.optimisticEstimate.timeInMs), 2451);
+    assert.equal(Math.round(result.pessimisticEstimate.timeInMs), 2752);
+    assert.equal(result.optimisticEstimate.nodeTiming.size, 19);
+    assert.equal(result.pessimisticEstimate.nodeTiming.size, 79);
+    assert.ok(result.optimisticGraph, 'should have created optimistic graph');
+    assert.ok(result.pessimisticGraph, 'should have created pessimistic graph');
+  });
```
brendankenny: it feels like there should be more tests here, but I don't have any special insight as to what boundaries they should be checking. Difficult traces or whatever. They can build up as we come across the difficult cases, but we don't have much infrastructure to spot those :) We should probably get some lantern smoke tests set up.

patrickhulce: 👍 right now its just me manually running my trace-evaluations scripts, post I/O I'll try and stick a lot of my traces onto GCP and get a travis script running to make sure accuracy doesn't regress

patrickhulce:
> whats the relative error? That's probably the better number to give here :P

sure it's ~30% which is similar to TTI

patrickhulce:
@paulirish did the single-line if's win you over enough to insta-approve? :)

PR #4967, opened 2018-04-13
tests(smoke): add smoke test code coverage

brendankenny:
Since:
- #4748 moved the smoke test runner to pure js
- `nyc` supports merging code coverage from any tests run in child processes
- `codecov` [supports uploading multiple coverage reports](https://docs.codecov.io/v4.3.6/docs/flags)

it's now easy to add smoke test coverage :)

Run locally with `yarn coverage:smoke` and on codcov you get little `unit`/`smoke` tabs at the top of file views (e.g. [`report-generator.js`](https://codecov.io/gh/GoogleChrome/lighthouse/src/2a9a772fa8bb57ff6b308251789cd0ad58c79355/lighthouse-core/report/v2/report-generator.js)). In codecov you seem to only get those tabs in individual files, both tabs are selected at first by default, and the coverage numbers on the left may be the sum of the two? So the view is a little shaky if you're interested in comparing unit vs smoke coverage, but it's nice that if you want the union of the two you can see that as well without uploading yet another coverage file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ script:
   - yarn type-check
   - yarn closure
   - yarn diff:sample-json
-  - yarn smoke
+  - yarn smoke:silentcoverage
   - yarn test-extension
   # _JAVA_OPTIONS is breaking parsing of compiler output. See #3338.
   - unset _JAVA_OPTIONS
@@ -49,5 +49,6 @@ before_cache:
   - rm -rf ./node_modules/temp-devtoolsprotocol/
 after_success:
   - yarn coveralls
+  - yarn codecov
 addons:
   chrome: stable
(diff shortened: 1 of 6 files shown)


brendankenny:
oh, the big gotcha here is that if a function will be `toString()`ed and then passed to `driver.evaluateAsync()` or `driver.evaluteScriptOnNewDocument()` it *must* have a `/* istanbul ignore next */` above it or `nyc` will instrument it and it will throw when then stringified and run in the browser.

Not an issue for (almost all of) the unit tests, but the smokehouse expectations will fail with gatherer errors if the code can't run. I can't think of a great way to automate doing this or checking it, so we'll rely on vigiliance?

PR #4968, opened 2018-04-13
core(image-usage): use min of resourceSize/transferSize

patrickhulce:
should fix https://twitter.com/dougsillars/status/984786614156189697

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -79,14 +79,15 @@ class OptimizedImages extends Gatherer {
       const isSameOrigin = URL.originsMatch(pageUrl, record._url);
       const isBase64DataUri = /^data:.{2,40}base64\s*,/.test(record._url);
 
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -79,14 +79,15 @@ class OptimizedImages extends Gatherer {
       const isSameOrigin = URL.originsMatch(pageUrl, record._url);
       const isBase64DataUri = /^data:.{2,40}base64\s*,/.test(record._url);
 
-      if (isOptimizableImage && record._resourceSize > MINIMUM_IMAGE_SIZE) {
+      const actualResourceSize = Math.min(record._resourceSize, record._transferSize);
```
brendankenny: can `_transferSize` be larger than `_resourceSize`? Seems possible, especially with image data, but even in that case wouldn't we still want to use the transfer size because they're sending too many bytes (it's just in that case one of the culprits is a bad choice to gzip?)

brendankenny:
looks like these tests might be helped by #4988?

PR #4973, opened 2018-04-15
core(canonical): pass when there are multiple identical canonical links

### LINKED ISSUE #3178: [SEO Audits] Document has a valid rel=canonical

kdzwinel:
We were failing when multiple canonical links were provided (as suggested by the docs). [John says](https://github.com/GoogleChrome/lighthouse/issues/3178#issuecomment-381181762) that we shound only fail if URLs are different.

oh no

Closes #3178

### DIFF
--- a/lighthouse-core/audits/seo/canonical.js
+++ b/lighthouse-core/audits/seo/canonical.js
@@ -110,7 +110,7 @@ class Canonical extends Audit {
         if (canonicals.length > 1) {
           return {
             rawValue: false,
-            debugString: `Multiple URLs (${canonicals.join(', ')})`,
+            debugString: `Multiple conflicting URLs (${canonicals.join(', ')})`,
           };
         }
 
--- a/lighthouse-core/audits/seo/canonical.js
+++ b/lighthouse-core/audits/seo/canonical.js
@@ -90,6 +90,8 @@ class Canonical extends Audit {
           });
 
         canonicals = canonicals.concat(artifacts.Canonical);
+        // we should only fail if there are multiple conflicting URLs
+        // see: https://github.com/GoogleChrome/lighthouse/issues/3178#issuecomment-381181762
         canonicals = Array.from(new Set(canonicals));
 
         artifacts.Hreflang.forEach(({href}) => hreflangs.push(href));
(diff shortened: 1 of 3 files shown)


patrickhulce:
@paulirish FYI weird parallel smoke failures here :/ is this what we should expect from the output or is something squirrelly going on here
image

lighthouse-core/audits/seo/canonical.js
```diff
@@ -90,6 +90,7 @@ class Canonical extends Audit {
           });
 
         canonicals = canonicals.concat(artifacts.Canonical);
+        canonicals = Array.from(new Set(canonicals));
```
brendankenny: maybe add comment with link to discussion? Looks like some of it happened out of band, but https://github.com/GoogleChrome/lighthouse/issues/3178#issuecomment-381181762 is at least a start if you're wondering why it's deduping here :)

kdzwinel:
@patrickhulce good point, we may need a small update in [the docs](https://developers.google.com/web/tools/lighthouse/audits/canonical)

> There is more than 1 canonical link.

should probably become sth like "There is more than one unique canonical link." or "There are multiple conflicting canonical links."

(cc @kaycebasques @rviscomi )

-------

@brendankenny thanks for suggestion! I've added same canonical URL we have in seo-tester.html `<head>` to its HTTP header.

lighthouse-core/audits/seo/canonical.js
kdzwinel: 👍 comment with explanation and link added

brendankenny:
smoke failure is unrelated and hopefully fixed by #4988 :)

PR #4983, opened 2018-04-17
core(lhr): lhr-lite type declaration

brendankenny:
ready for feedback on design/implementation of the design :)

### DIFF
--- a/typings/lhr-lite.d.ts
+++ b/typings/lhr-lite.d.ts
@@ -75,54 +75,49 @@ declare global {
          * 'informative': the audit is an FYI only, and can't be interpreted as pass/fail. Ignore the score.
          * 'notApplicable': the audit turned out to not apply to the page. Ignore the score.
          * 'manual': The audit exists only to tell you to review something yourself. Ignore the score.
+         * 'error': There was an error while running the object (check `errorMessage` for details). Ignore the score.
          */
-        scoreDisplayMode: 'binary' | 'numeric' | 'informative' | 'notApplicable' | 'manual';
+        scoreDisplayMode: 'binary' | 'numeric' | 'informative' | 'notApplicable' | 'manual' | 'error';
+        /** An explanation of audit-related issues encountered on the test page. */
+        explanation?: string;
         /** Extra information provided by some types of audits. */
-        details?: AuditDetails;
+        details?: Audit.MetricDetails | Audit.OpportunityDetails;
         /** Error message from any exception thrown while running this audit. */
         errorMessage?: string;
       }
 
-      export type AuditDetails = AuditMetricDetails | AuditOpportunityDetails | AuditTableDetails;
+      export module Audit {
+        export interface MetricDetails {
+          type: 'metric';
+          /** The value of the metric expressed in milliseconds. */
+          timespanMs?: number;
+        }
 
-      export interface AuditMetricDetails {
-        type: 'metric';
-        /** The value of the metric expressed in milliseconds. */
-        timespanMs?: number;
-      }
+        export interface OpportunityDetails {
+          type: 'opportunity';
+          wastedMs: number
+          wastedBytes?: number
+          headings: ColumnHeading[];
+          items: (WastedBytesDetailsItem | WastedTimeDetailsItem)[];
+        }
 
-      export interface AuditOpportunityDetails {
-        type: 'opportunity';
-        wastedMs: number
-        wastedBytes?: number
-        headings: AuditColumnHeading[];
-        rows: AuditTableRow[];
-      }
+        export interface ColumnHeading {
+          key: string;
+          label: string;
+          valueType: 'url' | 'timespanMs' | 'bytes';
+        }
 
-      export interface AuditTableDetails {
-        type: 'table';
-        headings: AuditColumnHeading[];
-        rows: AuditTableRow[];
-      }
+        export interface WastedBytesDetailsItem {
+          url: string;
+          wastedBytes?: number;
+          totalBytes?: number;
+        }
 
-      export interface AuditColumnHeading {
-        key: string;
-        label: string;
-        valueType: 'url' | 'text' | 'link' | 'timespanMs';
-      }
-
-      export type AuditTableRow = WastedBytesDetailsRow | WastedTimeDetailsRow;
-
-      export interface WastedBytesDetailsRow {
-        url: string;
-        wastedBytes?: number;
-        totalBytes?: number;
-      }
-
-      export interface WastedTimeDetailsRow {
-        url: string;
-        wastedMs: number;
-        totalBytes?: number;
+        export interface WastedTimeDetailsItem {
+          url: string;
+          wastedMs: number;
+          totalBytes?: number;
+        }
       }
     }
   }
--- a/typings/lhr-lite.d.ts
+++ b/typings/lhr-lite.d.ts
@@ -86,24 +86,24 @@ declare global {
       export type AuditDetails = AuditMetricDetails | AuditOpportunityDetails | AuditTableDetails;
 
       export interface AuditMetricDetails {
-        type: 'opportunity';
+        type: 'metric';
         /** The value of the metric expressed in milliseconds. */
         timespanMs?: number;
       }
 
       export interface AuditOpportunityDetails {
-        type: 'metric';
+        type: 'opportunity';
         wastedMs: number
         wastedBytes?: number
       }
 
       export interface AuditTableDetails {
         type: 'table';
-        headings: AuditTableHeadings[];
+        headings: AuditColumnHeading[];
         rows: AuditTableRow[];
       }
 
-      export interface AuditTableHeadings {
+      export interface AuditColumnHeading {
         key: string;
         label: string;
         valueType: 'url' | 'text' | 'link' | 'timespanMs';
--- a/typings/lhr-lite.d.ts
+++ b/typings/lhr-lite.d.ts
@@ -1,0 +1,130 @@
+/**
+ * @license Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+
+declare global {
+  module LH {
+    /**
+     * The lightweight version of Lighthouse results.
+     */
+    export interface ResultLite {
+      /** The URL that was supplied to Lighthouse and initially navigated to. */
+      requestedUrl: string;
+      /** The post-redirects URL that Lighthouse loaded. */
+      finalUrl: string;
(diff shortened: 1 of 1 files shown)


typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,130 @@
...
+        score: number;
+        /**
+         * A string identifying how the score should be interpreted:
+         * 'binary': pass/fail audit (0 and 1 are only possible scores).
+         * 'numeric': scores of 0-1 (map to displayed scores of 0-100).
+         * 'informative': the audit is an FYI only, and can't be interpreted as pass/fail. Ignore the score.
+         * 'notApplicable': the audit turned out to not apply to the page. Ignore the score.
+         * 'manual': The audit exists only to tell you to review something yourself. Ignore the score.
+         */
+        scoreDisplayMode: 'binary' | 'numeric' | 'informative' | 'notApplicable' | 'manual';
```
brendankenny: should we have an `'error'` `scoreDisplayMode`? That brings us (basically) back to a boolean and a string, which isn't ideal, but it also makes sense as it affects interpretation of the score in the same way that `'informative'` `'notApplicable'`, and `'manual'` do (you should ignore whatever is in `score`) and you won't have to look in multiple places to see how you should interpret `score` (like you do currently)

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+        /** The value of the metric expressed in milliseconds. */
+        timespanMs?: number;
+      }
+
+      export interface AuditOpportunityDetails {
+        type: 'opportunity';
+        wastedMs: number
+        wastedBytes?: number
+        headings: AuditColumnHeading[];
+        rows: AuditTableRow[];
```
paulirish: renaming `items` to `rows`, eh? :)

I prefer `items` personally because it doesn't imply these table semantics. Table-ness is a  visual/report thing, which i don't think is relevant to most consumers of the LHR

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+      /** The ISO-8601 timestamp of when the results were generated. */
+      fetchedAt: string;
+      /** The version of Lighthouse with which these results were generated. */
+      lighthouseVersion: string;
+      /** An object containing the results of the audits. */
+      audits: Record<string, ResultLite.Audit>;
+      /** An object containing the top-level categories, their overall scores, and reference to member audits. */
+      categories?: Record<string, ResultLite.Category>;
+      /** Descriptions of the audit groups referenced by AuditRefs. */
+      categoryGroups?: Record<string, ResultLite.CategoryGroup>;
```
patrickhulce: Did we decide of `groups` for this? I don't remember at this point 😆 or `auditGroups` aka "groups of related audits"

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+      /** The URL that was supplied to Lighthouse and initially navigated to. */
+      requestedUrl: string;
+      /** The post-redirects URL that Lighthouse loaded. */
+      finalUrl: string;
+      /** The ISO-8601 timestamp of when the results were generated. */
+      fetchedAt: string;
+      /** The version of Lighthouse with which these results were generated. */
+      lighthouseVersion: string;
+      /** An object containing the results of the audits. */
+      audits: Record<string, ResultLite.Audit>;
```
patrickhulce: since we don't have any references to what `auditId` really is how do you feel about aliases like `type AuditId = string`?

typings/lhr-lite.d.ts
patrickhulce: I like it 👍

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+        rows: AuditTableRow[];
+      }
+
+      export interface AuditColumnHeading {
+        key: string;
+        label: string;
+        valueType: 'url' | 'text' | 'link' | 'timespanMs';
+      }
+
+      export type AuditTableRow = WastedBytesDetailsRow | WastedTimeDetailsRow;
```
paulirish: can u inline this into the two uses above?

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+      export interface AuditMetricDetails {
+        type: 'metric';
+        /** The value of the metric expressed in milliseconds. */
+        timespanMs?: number;
+      }
+
+      export interface AuditOpportunityDetails {
+        type: 'opportunity';
+        wastedMs: number
+        wastedBytes?: number
```
brendankenny: should we rename these `totalWastedMs`/`totalWastedBytes`? To make it clear what they are

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+      export interface AuditTableDetails {
+        type: 'table';
+        headings: AuditColumnHeading[];
+        rows: AuditTableRow[];
+      }
+
+      export interface AuditColumnHeading {
+        key: string;
+        label: string;
+        valueType: 'url' | 'text' | 'link' | 'timespanMs';
```
patrickhulce: also have `bytes` and `duration`

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,132 @@
...
+      export interface AuditOpportunityDetails {
+        type: 'opportunity';
+        wastedMs: number
+        wastedBytes?: number
+        headings: AuditColumnHeading[];
+        rows: AuditTableRow[];
+      }
+
+      export interface AuditTableDetails {
+        type: 'table';
```
patrickhulce: do we actually have any of these that we want to expose? seems like we're only wanting to expose the opportunities which all use the table headings/rows types

typings/lhr-lite.d.ts
brendankenny: > renaming items to rows, eh? :)

hehe, caught me. I can switch back, but "headings" doesn't really make sense without some matrix/table-ness structure imposed on the data anyways

typings/lhr-lite.d.ts
paulirish: Crazy idea. since these are opportunities, we could stay on the optimistic side.. something like.. `"estimatedSavingsMs"` ?

typings/lhr-lite.d.ts
brendankenny: duration -> timespanMs I thought? Also do we actually have text or links if it's just WastedBytesDetailsItem or WastedTimeDetailsItem?

brendankenny:
oh yeah, a big one is `explanation`. Currently it's on `details`...should it just be on the audit result?

typings/lhr-lite.d.ts
brendankenny: > estimatedSavingsMs

I'm on board with that @patrickhulce

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,127 @@
...
+      /** The post-redirects URL that Lighthouse loaded. */
+      finalUrl: string;
+      /** The ISO-8601 timestamp of when the results were generated. */
+      fetchedAt: string;
+      /** The version of Lighthouse with which these results were generated. */
+      lighthouseVersion: string;
+      /** An object containing the results of the audits. */
+      audits: Record<string, ResultLite.Audit>;
+      /** An object containing the top-level categories, their overall scores, and reference to member audits. */
+      categories?: Record<string, ResultLite.Category>;
```
brendankenny: should be required, even if an empty object (API responses should always have categories, so no need to make people jump through hoops checking for `categories` existence every time)

brendankenny:
I think the last decision is `categoryGroups`/`groups`/`auditGroups`/etc? (I believe with the split on where the groups should live we decided on the status quo of leaving at the top level).

Any thoughts on that and last changes before I commit us to this? :)

patrickhulce:
> I think the last decision is categoryGroups/groups/auditGroups/etc? (I believe with the split on where the groups should live we decided on the status quo of leaving at the top level).
Any thoughts on that and last changes before I commit us to this? :)

Aw shoot sorry I forgot to submit my review!! Yes I have a thought. Do we even need these in LHR lite? 
I realized they're a display-only concern and my favorite names for them are just `groups`/`displayGroups`. Seems like we can punt from LHR lite but maybe I'm missing something?

brendankenny:
if it were `groups` vs `displayGroups` I'd tip slightly toward `displayGroups`.

I can't remember exactly why we added them to LHR-lite but I believe @paulirish thought it would be desired by consumers/trivial to add.

paulirish:
okay i'm on board with leaving groups out of LHR-lite. we're not 100% on them and there's not a high priority reason to include them. (i think i started with "hey, might as well.")

* metrics & opps are already distinguishable from `details.type`.
* and manual audits are also distinguishable.
* so that just leaves the grouping within `a11y` and `seo`. 

The only one there that could be useful is `seo-mobile`, our mobile friendly group. but.. eh.

k let's boot groups.

patrickhulce:
cool cool +1 to brendan's fav for `displayGroups` and `displayGroupId` for our real LHR :)

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,116 @@
...
+    export module ResultLite {
+      export interface Category {
+        /** The human-friendly name of the category. */
+        title: string;
+        /** A description of what this category is about (e.g. these help you validate your PWA). */
+        description: string;
+        /** The overall score of the category, the weighted average of all its audits. */
+        score: number;
+        /** An array of references to all the audit members of this category. */
+        auditRefs: AuditRef[];
```
patrickhulce: 😬

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,116 @@
...
+        /** An optional description for manual audits within this category. */
+        manualDescription?: string;
```
patrickhulce: 😬

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,116 @@
...
+        weight: number;
+      }
+
+      export interface Audit {
+        /** The brief description of the audit. The text can change depending on if the audit passed or failed. */
+        title: string;
+        /** A more detailed description that describes why the audit is important and links to Lighthouse documentation on the audit; markdown links supported. */
+        description: string;
+        /** The scored value determined by the audit, in the range `0-1`, or null if `scoreDisplayMode` indicates not scored. */
+        score: number | null;
```
patrickhulce: FWIW I'd prefer `score?: number`

typings/lhr-lite.d.ts
```diff
@@ -0,0 +1,116 @@
...
+        details?: Audit.MetricDetails | Audit.OpportunityDetails;
+        /** Error message from any exception thrown while running this audit. */
+        errorMessage?: string;
+      }
+
+      export module Audit {
+        export interface MetricDetails {
+          type: 'metric';
+          /** The value of the metric expressed in milliseconds. */
+          timespanMs?: number;
```
patrickhulce: I'm still 😬, but I think we mostly exhausted options. last-ditch effort anyone like `ms` `timeMs` `timingMs` more?

paulirish: i consider timespanMs/ms/timingMs to be equally 😬  and i'd be fine with any of those.

brendankenny:
@paulirish up to you now

PR #4986, opened 2018-04-18
core(config): remove config.artifacts; always use auditMode

brendankenny:
part of #4333

remove support for artifacts loaded in the config and the kind of wonky way it loaded traces and devtoolsLogs from there. Now the official way to load artifacts and skip the gather step is with `--audit-mode`/`auditMode` in the config.

Required creating some artifacts fixtures to port over old tests that loaded with `config.artifacts`, but on the plus side it means more tests for `--audit-mode`.


lighthouse-core/runner.js
```diff
@@ -115,7 +113,8 @@ class Runner {
       const resultsById = {};
       for (const audit of auditResults) resultsById[audit.name] = audit;
 
-      let reportCategories;
+      /** @type {Array<LH.Result.Category>} */
+      let reportCategories = [];
```
brendankenny: part of the discussed change to default to having an empty `categories` array, even if none in the config (we could maybe start defaulting to `[]` in the config, but we'll see)

lighthouse-core/test/fixtures/artifacts/perflog/defaultPass.trace.json
```diff
@@ -0,0 +1,14 @@
+[
```
brendankenny: copied over from `fixtures/traces/trace-user-timings.json` cause it was small and we really only need to test the loading part :)

lighthouse-core/test/gather/fake-driver.js
```diff
@@ -60,7 +60,7 @@ module.exports = {
   },
   beginDevtoolsLog() {},
   endDevtoolsLog() {
-    return require('../fixtures/perflog.json');
+    return require('../fixtures/artifacts/perflog/defaultPass.devtoolslog.json');
```
brendankenny: this was big, so just moved over to the artifacts fixtures and moved the two tests using it to point at the new path

lighthouse-core/test/index-test.js
```diff
@@ -87,18 +87,17 @@ describe('Module Tests', function() {
       });
   });
 
-  it.skip('should return formatted audit results when given no categories', function() {
+  it.only('should return formatted LHR when given no categories', function() {
```
brendankenny: this was skipped when `-GA` was added. Reenabled :P

lighthouse-core/test/runner-test.js
```diff
@@ -148,25 +148,6 @@ describe('Runner', () => {
       });
   });
 
-  it('accepts existing artifacts', () => {
```
brendankenny: overlap with lots of other -GA and `auditMode` tests, so removed

lighthouse-core/test/index-test.js
```diff
@@ -87,18 +87,17 @@ describe('Module Tests', function() {
       });
   });
 
-  it.skip('should return formatted audit results when given no categories', function() {
+  it('should return formatted LHR when given no categories', function() {
```
brendankenny: this was skipped when `-GA` was added. Reenabled :P

lighthouse-core/config/config.js
```diff
@@ -17,89 +17,6 @@ const path = require('path');
 const Audit = require('../audits/audit');
 const Runner = require('../runner');
 
-// cleanTrace is run to remove duplicate TracingStartedInPage events,
-// and to change TracingStartedInBrowser events into TracingStartedInPage.
-// This is done by searching for most occuring threads and basing new events
-// off of those.
-function cleanTrace(trace) {
```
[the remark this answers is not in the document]
paulirish: aye. this removes support for traces captured by webdriver, but at this point, i'm just fine with that. :)

PR #4989, opened 2018-04-18
core(estimated-input-latency): use a 5s rolling window

patrickhulce:
chose a 5s window because that's what TTI uses and it gave the right "feel" on several test sites, CNN gets ~1s and is punished harshly, pwa.rocks gets ~17ms, 1s was a bit too small and harsh 10s is forgiving of 1s long tasks which is just too lenient


paulirish:
Is what we're computing here functionally equivalent to 
 
> the largest 90pctl EQT value of 5s windows between FMP to end of the trace

patrickhulce:
> Is what we're computing here functionally equivalent to
the largest 90pctl EQT value of 5s windows between FMP to end of the trace

yes

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -18,11 +20,12 @@ class EstimatedInputLatency extends Audit {
     return {
       name: 'estimated-input-latency',
       description: 'Estimated Input Latency',
-      helpText: 'The score above is an estimate of how long your app takes to respond to user ' +
-          'input, in milliseconds. There is a 90% probability that a user encounters this amount ' +
-          'of latency, or less. 10% of the time a user can expect additional latency. If your ' +
-          'latency is higher than 50 ms, users may perceive your app as laggy. ' +
-          '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/estimated-input-latency).',
+      helpText:
```
brendankenny: possible to revert these? don't appear to have changed :)

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -45,23 +48,34 @@ class EstimatedInputLatency extends Audit {
...
+    const events = TracingProcessor.getMainThreadTopLevelEvents(tabTrace, startTime);
+
+    const candidateStartEvts = events.filter(evt => evt.duration >= 10);
+
+    let worst90thPercentileLatency = 16;
+    for (const startEvt of candidateStartEvts) {
+      const latencyPercentiles = TracingProcessor.getRiskToResponsiveness(
+        events,
+        startEvt.start,
+        startEvt.start + ROLLING_WINDOW_SIZE
```
brendankenny: if there's a last argument of `[0.9]` then no need for the `find()` below (since dropping the `extendedInfo`...though if we keep EIL, there's arguably usefulness in keeping other percentiles :)

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -18,11 +20,12 @@ class EstimatedInputLatency extends Audit {
...
-          'input, in milliseconds. There is a 90% probability that a user encounters this amount ' +
-          'of latency, or less. 10% of the time a user can expect additional latency. If your ' +
-          'latency is higher than 50 ms, users may perceive your app as laggy. ' +
-          '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/estimated-input-latency).',
+      helpText:
+        'The score above is an estimate of how long your app takes to respond to user ' +
+        'input, in milliseconds. There is a 90% probability that a user encounters this amount ' +
+        'of latency, or less. 10% of the time a user can expect additional latency. If your ' +
+        'latency is higher than 50 ms, users may perceive your app as laggy. ' +
+        '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/estimated-input-latency).',
```
brendankenny: we might want to tweak this, though

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -75,8 +89,9 @@ class EstimatedInputLatency extends Audit {
   static audit(artifacts, context) {
     const trace = artifacts.traces[this.DEFAULT_PASS];
 
-    return artifacts.requestTraceOfTab(trace)
-        .then(traceOfTab => EstimatedInputLatency.calculate(traceOfTab, context));
+    return artifacts
```
brendankenny: revert? :D

patrickhulce: done

lighthouse-core/audits/estimated-input-latency.js
patrickhulce: tweaked!

lighthouse-core/audits/estimated-input-latency.js
patrickhulce: done

lighthouse-core/audits/estimated-input-latency.js
patrickhulce: done

patrickhulce:
> what's performance like? We could add an iterative version of getMainThreadTopLevelEventDurations() if it's slow as it's doing a lot of redundant work per window

yeah it could definitely be better (~200ms for a CNN trace), but it's marginally better than our old TTI, and the filter on possible start points helps a lot

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -45,23 +46,36 @@ class EstimatedInputLatency extends Audit {
       throw new LHError(LHError.errors.NO_FMP);
     }
 
-    const latencyPercentiles = TracingProcessor.getRiskToResponsiveness(tabTrace, startTime);
-    const ninetieth = latencyPercentiles.find(result => result.percentile === 0.9);
-    const rawValue = parseFloat(ninetieth.time.toFixed(1));
+    const events = TracingProcessor.getMainThreadTopLevelEvents(tabTrace, startTime)
+      .filter(evt => evt.duration >= 1);
```
paulirish: since a blank trace gets a value of 16, can't we safely remove any events with a duration < 16?

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -45,23 +46,36 @@ class EstimatedInputLatency extends Audit {
...
+
+    const candidateStartEvts = events.filter(evt => evt.duration >= 10);
```
paulirish: similarly, can we raise 10 to something like 50?  Seems like we'd lose precision in the "you're good" area of scores, but that doesn't seem too bad.

lighthouse-core/audits/estimated-input-latency.js
patrickhulce: well a trace with constant 16ms events would technically get a EQT of 24ms since we arbitrarily add 16ms to whatever value we compute, I'm all for doing more filtering here though, who cares about your small events

lighthouse-core/audits/estimated-input-latency.js
patrickhulce: yeah sgtm 👍 much faster too

brendankenny: doesn't that make the max score the PODR?

brendankenny: oh, only filtering the start event times

brendankenny:
if we're only concerned with speed there are easier ways of making this faster than sprinkling filters everywhere :P

patrickhulce:
> if we're only concerned with speed there are easier ways of making this faster than sprinkling filters everywhere :P

Speed is a major concern right now. In the unit test which has 20s of CPU time split between 1000 start times and 10000 small tasks it takes ~40-50ms. Of the 200ms for CNN 180 of those are spent in trace-of-tab, not here. We're not filtering because we absolutely have to, it's more why do useless work. I don't really care either way. 1ms and 10ms don't seem to hurt anything speed-wise or have any meaningful impact accuracy-wise.

patrickhulce:
friendly ping for re-review!

PR #4995, opened 2018-04-19
core: merge render blocking audits to lantern

patrickhulce:
So this turned out to be quite a doozy 😩 

**Changes in this PR**:
* Fixes a bug in byte-efficiency-audit base where the default lantern settings were always used regardless of settings
* Adds HTML imports to Lantern's render blocking definition
* Merges render blocking links/scripts into single audit
* Changes the savings calculation to be based on lantern
    * **What's the same?** We still pass our smoketests and identify all the same things as render blocking! 🎉 

**What's different about the savings calculation?** 

* Previously we would say that you save Math.max(...downloadTimes) but this wasn't really accurate since you pretty much have to remove all render blocking resource to see this benefit. 
* Additionally, this assumed that the code is just deferred and you don't need it at all. The new approach is computing "how much faster would this page reach FCP if we inlined all the used CSS from render blocking stylesheets and deferred all scripts" as a result the savings we report are slightly more conservative but much more accurate. 
* This also means that H2 sites will basically never be flagged since the simulation treats H2 in essentially this way. Doesn't seem so bad though considering the only savings you're seeing with H2 is  ~1 RT which in our settings is just 150ms. 
* Our fancy amp-style detection is less relevant since we don't actually identify savings from addressing a single resource, just all in aggregate and this should be much more conservative. It does mean though that we might slightly overstate savings if AMP-style mitigation didn't kick in on unthrottled but would on throttled. 

ref #4333


paulirish:
> Merges render blocking links/scripts into single audit

I think we should keep them separate. :) 

Handling RB stylesheets is done via reducing, inlining and the critical CSS dance.Handling RB scripts involved a bunch of different techniques.  I think they're quite distinct in how you have to reason about them, even though at a browser level they are similar. So i feel its still valuable to report on them individually. Also getting to avoid the super generic "resources" is great..

paulirish:
Discussed this more. Here are the options that we could do:

1. Start with what's in master. Use lantern for the download timing calculation.
2. Use this PR, mostly as is. single RB audit instead of the current 2.
3. Use this PR, but show it as two (CSS & JS). Savings will be the same number. Docs optionally explain that you need to fix both to see the savings.
4. Take it to the next level. This audit becomes a "Inline critical render-blocking stuff and defer the rest". We use code coverage to calculate how much of the JS/CSS is needed and then calculate if that was inlined and then the rest are deferred. Most sophisticated, but accurate estimation and the user recommendation is very clear.

patrickhulce:
@paulirish says
> let's do 4

on it.

patrickhulce:
gentle reminder this is awaitin' review once again :)

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+      const nodeTiming = nodeTimingMap.get(node);
+      // TODO(phulce): beacon these occurences to Sentry to improve FCP graph
+      if (!node) continue;
+
+      // Mark this node and all it's dependents as deferrable
+      // TODO(phulce): make this slightly more surgical
+      // i.e. the referenced font asset won't become inlined just because you inline the CSS
+      node.traverse(node => deferredNodeIds.add(node.id));
+
+      const wastedMs = nodeTiming.endTime - nodeTiming.startTime;
```
paulirish: > The wasted cost of each request is from requestSent to responseReceived

paulirish: btw do you want to round this?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+   */
+  static estimateSavingsFromInlining(simulator, fcpGraph, deferredIds, wastedBytesMap) {
+    const originalEstimate = simulator.simulate(fcpGraph).timeInMs;
+
+    let totalChildNetworkBytes = 0;
+    const graphWithoutChildren = fcpGraph.cloneWithRelationships(node => {
+      const willDefer = deferredIds.has(node.id);
+      if (willDefer && node.type === Node.TYPES.NETWORK &&
+          node.record._resourceType === WebInspector.resourceTypes.Stylesheet) {
+        const wastedBytes = wastedBytesMap.get(node.record.url) || 0;
```
paulirish: > Assume we defer the network request till later, and mark it non-blocking

right?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+        totalChildNetworkBytes += node.record._transferSize - wastedBytes;
+      }
+
+      // Include all nodes that couldn't be deferred
+      return !willDefer;
+    });
+
+    graphWithoutChildren.record._transferSize += totalChildNetworkBytes;
```
paulirish: > Add inlined bytes back to the HTML response

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+    let totalChildNetworkBytes = 0;
+    const graphWithoutChildren = fcpGraph.cloneWithRelationships(node => {
+      const willDefer = deferredIds.has(node.id);
+      if (willDefer && node.type === Node.TYPES.NETWORK &&
+          node.record._resourceType === WebInspector.resourceTypes.Stylesheet) {
+        const wastedBytes = wastedBytesMap.get(node.record.url) || 0;
+        totalChildNetworkBytes += node.record._transferSize - wastedBytes;
+      }
+
+      // Include all nodes that couldn't be deferred
```
paulirish: > If a node can be deferred till later (non-render-blocking), it's not part of the FCP graph

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+   * @param {Set<string>} deferredIds
+   * @param {Map<string, number>} wastedBytesMap
+   * @return {number}
+   */
+  static estimateSavingsFromInlining(simulator, fcpGraph, deferredIds, wastedBytesMap) {
+    const originalEstimate = simulator.simulate(fcpGraph).timeInMs;
+
+    let totalChildNetworkBytes = 0;
+    const graphWithoutChildren = fcpGraph.cloneWithRelationships(node => {
+      const willDefer = deferredIds.has(node.id);
```
paulirish: seems more like `canDeferRequest`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+        const wastedBytes = wastedBytesMap.get(node.record.url) || 0;
+        totalChildNetworkBytes += node.record._transferSize - wastedBytes;
+      }
+
+      // Include all nodes that couldn't be deferred
+      return !willDefer;
+    });
+
+    graphWithoutChildren.record._transferSize += totalChildNetworkBytes;
+    const estimateAfterInlineA = simulator.simulate(graphWithoutChildren);
```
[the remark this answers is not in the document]
paulirish: `estimateAfterInlining`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  /**
+   * @param {Simulator} simulator
+   * @param {Node} fcpGraph
+   * @param {Set<string>} deferredIds
+   * @param {Map<string, number>} wastedBytesMap
+   * @return {number}
+   */
+  static estimateSavingsFromInlining(simulator, fcpGraph, deferredIds, wastedBytesMap) {
```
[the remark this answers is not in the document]
paulirish: > Estimate savings from deferring render-blocking requests and inlining used CSS into the HTML response.

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+    const originalEstimate = simulator.simulate(fcpGraph).timeInMs;
+
+    let totalChildNetworkBytes = 0;
+    const graphWithoutChildren = fcpGraph.cloneWithRelationships(node => {
+      const willDefer = deferredIds.has(node.id);
+      if (willDefer && node.type === Node.TYPES.NETWORK &&
```
paulirish: extract condition to `canInlineCSSBytes`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
paulirish: also since the inlining only applies to CSS but this affects both JS and CSS it should be named something that's more balanced.  `estimateSavingsFromDeferring` sg. wdyt?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+   * @param {Node} fcpGraph
+   * @param {Set<string>} deferredIds
+   * @param {Map<string, number>} wastedBytesMap
+   * @return {number}
+   */
+  static estimateSavingsFromInlining(simulator, fcpGraph, deferredIds, wastedBytesMap) {
+    const originalEstimate = simulator.simulate(fcpGraph).timeInMs;
+
+    let totalChildNetworkBytes = 0;
+    const graphWithoutChildren = fcpGraph.cloneWithRelationships(node => {
```
paulirish: why is it called this btw? can't really make sense of it.

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
paulirish: And can you also add a comment that we don't inline JS because we assume they can get a [defer] attribute

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+    const simulator = await artifacts.requestLoadSimulator(simulatorData);
+    const wastedBytesMap = await RenderBlockingResources.computeWastedCSSBytes(artifacts, context);
+
+    const metricSettings = {throttlingMethod: 'simulate'};
+    const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
+    const fcpSimulation = await artifacts.requestFirstContentfulPaint(metricComputationData);
+    const fcpTsInMs = traceOfTab.timestamps.firstContentfulPaint / 1000;
+
+    const nodeTimingMap = fcpSimulation.optimisticEstimate.nodeTiming;
+    const nodesByUrl = keyByUrl(Array.from(nodeTimingMap.keys()));
```
paulirish: can we instead just have a helper function of like `getNodeTimingForUrl(nodeTimingMap, url){ ..`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,188 @@
...
+    const traceOfTab = await artifacts.requestTraceOfTab(trace);
+    const simulator = await artifacts.requestLoadSimulator(simulatorData);
+    const wastedBytesMap = await RenderBlockingResources.computeWastedCSSBytes(artifacts, context);
+
+    const metricSettings = {throttlingMethod: 'simulate'};
+    const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
+    const fcpSimulation = await artifacts.requestFirstContentfulPaint(metricComputationData);
+    const fcpTsInMs = traceOfTab.timestamps.firstContentfulPaint / 1000;
+
+    const nodeTimingMap = fcpSimulation.optimisticEstimate.nodeTiming;
```
paulirish: side note i wish this was called `nodeTimings` or `nodeTimingMap` to differentiate from the singular case within it.

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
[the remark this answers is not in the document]
paulirish: Also a lot of what you wrote in the [awesome PR description](https://github.com/GoogleChrome/lighthouse/pull/4995#issue-182851927) can be moved into comments. 😀

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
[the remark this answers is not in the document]
patrickhulce: just leftover from debugging, fixed

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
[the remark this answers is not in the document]
patrickhulce: 👍 done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: since it's doing both, I just renamed to `estimateSavingsWithGraphs` and added better description 👍

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: it was leftover from when everything was inlined, changed to `minimalFCPGraph`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: and the helper function loops through the keys to find the matching record?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: yeah we can rename in separate PR 👍

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: ya, comment added

patrickhulce:
@paulirish all done 'cept the helper method business, not 100% I understood what that wins us

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,213 @@
...
+    const deferredNodeIds = new Set();
+    for (const resource of artifacts.TagsBlockingFirstPaint) {
+      // Ignore any resources that finished after observed FCP (they're clearly not render-blocking)
+      if (resource.endTime * 1000 > fcpTsInMs) continue;
+      // TODO(phulce): beacon these occurences to Sentry to improve FCP graph
+      if (!nodesByUrl[resource.tag.url]) continue;
+
+      const {node, nodeTiming} = nodesByUrl[resource.tag.url];
+
+      // Mark this node and all it's dependents as deferrable
```
brendankenny: its

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * @param {!Artifacts} artifacts
+   * @param {LH.Audit.Context} context
+   * @return {Map<string, number>}
+   */
+  static async computeWastedCSSBytes(artifacts, context) {
+    const wastedBytesByUrl = new Map();
+    try {
+      const results = await UnusedCSS.audit(artifacts, context);
```
brendankenny: maybe a TODO that this should be pulled out into a computed artifact?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,213 @@
...
+      for (const item of results.details.items) {
+        wastedBytesByUrl.set(item.url, item.wastedBytes);
+      }
+    } catch (_) {}
```
brendankenny: whats the failure mode here?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -0,0 +1,213 @@
...
+   * @param {Artifacts} artifacts
+   * @param {LH.Audit.Context} context
+   */
+  static async computeResults(artifacts, context) {
+    const trace = artifacts.traces[Audit.DEFAULT_PASS];
+    const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
+    const simulatorData = {devtoolsLog, settings: context.settings};
+    const traceOfTab = await artifacts.requestTraceOfTab(trace);
+    const simulator = await artifacts.requestLoadSimulator(simulatorData);
+    const wastedBytesMap = await RenderBlockingResources.computeWastedCSSBytes(artifacts, context);
```
brendankenny: `wastedCssBytesMap`? Or it's more like `cssBytesByUrl` or something in function

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: assume all bytes were used and inline all CSS

PR #4997, opened 2018-04-19
tests: split up parallel smoketests into 3 batches (WIP)

paulirish:
variant of #4996

### DIFF
--- a/lighthouse-cli/test/smokehouse/run-smoke.js
+++ b/lighthouse-cli/test/smokehouse/run-smoke.js
@@ -28,34 +28,42 @@ const SMOKETESTS = [{
   id: 'ally',
   config: smokehouseDir + 'a11y/a11y-config.js',
   expectations: 'a11y/expectations.js',
+  batch: 'parallel-first',
 }, {
   id: 'pwa',
   expectations: smokehouseDir + 'pwa-expectations.js',
   config: smokehouseDir + 'pwa-config.js',
+  batch: 'parallel-second',
 }, {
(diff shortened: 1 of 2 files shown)


lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -70,6 +78,7 @@ const SMOKETESTS = [{
   id: 'ttci',
   expectations: 'tricky-ttci/expectations.js',
   config: 'lighthouse-core/config/default-config.js',
+  batch: 'parallel-third',
```
patrickhulce: this one seems particularly important to run on its own, if it's always really really slow then the expectation it's asserting (that TTI is at least X seconds) becomes meaningless

paulirish:
based on the data in https://github.com/GoogleChrome/lighthouse/pull/4996, 2 batches is faster.

lighthouse-cli/test/smokehouse/run-smoke.js
paulirish: mmmmmmmhmmmmmmm have we seen it flake yet? i kinda dont want to break it into a solo if we don't have to. :)

patrickhulce: it's not that it will flake, it's that we shouldn't both running it at all if it's going to be running very slowly because the only assertion it makes is that the TTI is at least some value based on tricky properties of the page, but if it always passes because travis so doggone slow then no point in having it

PR #5000, opened 2018-04-19
cli: update documented throttling flags

patrickhulce:
doing some cleanup on the remaining references to the old `disable-X-throttling` flags, we should probably explore an automated way of keeping those flags in sync with our `SettingsJson` typedefs now that all the options are available to CLI

not sure if it's really better to piecemeal update readme like this or do a pass over our docs and re-evaluate if we still want to tell people to be doing things like we are

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -88,7 +88,7 @@ $ adb devices -l
 
 $ adb forward tcp:9222 localabstract:chrome_devtools_remote
 
-$ lighthouse --port=9222 --disable-device-emulation --disable-cpu-throttling https://example.com
+$ lighthouse --port=9222 --disable-device-emulation --throttling.cpuSlowdownMultiplier=1 https://example.com
 ```
 
 ## Lighthouse as trace processor
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -76,7 +76,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling` and potentially `--disable-network-throttling`.
+You'll likely want to use the CLI flags `--disable-device-emulation --throttling.cpuSlowdownMultiplier`.
 
 ```sh
 $ adb kill-server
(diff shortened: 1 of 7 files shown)


docs/throttling.md
```diff
@@ -50,7 +50,7 @@ Currently, `comcast` will also throttle the websocket port that Lighthouse uses
 comcast --latency=150 --target-bw=1600
 
 # Run Lighthouse with it's own throttling disabled
```
brendankenny: will you fix the `it's` while youre in here?

docs/readme.md
```diff
@@ -76,7 +76,7 @@ instance with an open debugging port.
 
 Lighthouse can run against a real mobile device. You can follow the [Remote Debugging on Android (Legacy Workflow)](https://developer.chrome.com/devtools/docs/remote-debugging-legacy) up through step 3.3, but the TL;DR is install & run adb, enable USB debugging, then port forward 9222 from the device to the machine with Lighthouse.
 
-You'll likely want to use the CLI flags `--disable-device-emulation --disable-cpu-throttling` and potentially `--disable-network-throttling`.
+You'll likely want to use the CLI flags `--disable-device-emulation --throttling.cpuSlowdownMultiplier`.
```
brendankenny: update example below?

lighthouse-cli/cli-flags.js
```diff
@@ -33,8 +33,8 @@ function getFlags(manualArgv) {
           'lighthouse <url> --output=json --output-path=./report.json --save-assets',
           'Save trace, screenshots, and named JSON report.')
       .example(
-          'lighthouse <url> --disable-device-emulation --disable-network-throttling',
-          'Disable device emulation')
+          'lighthouse <url> --disable-device-emulation --throttling-method=provided',
```
brendankenny: thoughts for another day: `--throttling-method=provided` -- this is kind of unfortunate wording if you really do just want throttling off. We could maybe make `none` or something that's basically like `provided`, though we probably don't want to make it a strict alias...

paulirish: yeah I also think this is awkward. Now with our laternification do we need our user to inform us if the throttling is off or provided?  If not, then I think we should rename `provided` (and maybe even `throttling-method`). 

Either way it's possible we have to introduce some CLI sugar that may be inconsistent with the config..

docs/throttling.md
patrickhulce: haha done

docs/readme.md
patrickhulce: done

lighthouse-cli/cli-flags.js
patrickhulce: Yeah I agree it's odd, there's no difference between `off` and `provided` I'm fine to rename to whatever in followup 👍

Issue #1902, opened 2017-03-24
Audit: Avoid large JavaScript bundle sizes


addyosmani:
We currently have an audit for enormous network payloads (byte efficiency) which provides a baseline for understanding opportunities to reduce overall payload size:

![image](https://cloud.githubusercontent.com/assets/110953/24308954/335cd7f4-1087-11e7-8cbe-4ff3e33a48df.png)

**How would we feel about introducing an audit (or sub-audit) for large JS bundle sizes too?** 

If you're shipping more than ~170KB of script then we flag you. I've used some higher budgets just for illustrative purposes below since our understanding of what the medians and 99th percentiles were including JS wise have improved over time.

If we wanted to accomplish this as a UI treatment in the existing audit, we could highlight per-script budget breakers in red:

![image](https://cloud.githubusercontent.com/assets/110953/24309111/c321dd26-1087-11e7-8d27-fe767cd9e321.png)

..or we could look at the total sum of bytes for scripts with a '.js' extension and flag those:

![image](https://cloud.githubusercontent.com/assets/110953/24309213/29d1895e-1088-11e7-8edd-ec52f0434d3e.png)

and include some messaging about why we're flagging those scripts:

![image](https://cloud.githubusercontent.com/assets/110953/24309373/da77610c-1088-11e7-9f08-afe2cd2903b6.png)

Thoughts?

Side: Would totally love to have an audit like this for my I/O talk as bundle sizes (in particular for folks reusing bits in the React ecosystem) can be a challenge. 

Audit impacted: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/byte-efficiency/total-byte-weight.js

patrickhulce:
I think it'd also be nice to express the cost here in terms of parse time in addition to network time. Can't really get that with CPU emulation today but maybe with calibration + trace we can estimate how long parse time would have been?

ebidel:
Doing a sum of all the JS bytes might be confusing/tricky since we don't know what should be consider the user's "bundle". For example, we wouldn't want to include analytics.js.

That said, I think the key point we should raise to the user is that too much JS (> 450KB) is too much. We could highlight that target in the help text? Or, have another audit specific to JS size and use the amazing cards formatter. Thoughts?

[...]

PR #5010, opened 2018-04-22
new-audit(large-bundle-size): highlight large js bundles

wardpeet:
I updated the total-byte-size audit to highlight big js bundles.

image

I added some remarks on the issue #1902


wardpeet:
Locally when I run yarn smoke dbw when fixing travis I get the following:
image

so unsure what to do here. If I change it smoketests stop working locally but work on travis.

patrickhulce:
> so unsure what to do here. If I change it smoketests stop working locally but work on travis.

let's loosen the assertion a bit `>30000` seems sufficient

hwikyounglee:
Here's a recommended visual style. An intention is to provide a visual cue with both shape and color. 

large-js-bundle

addyosmani:
After spending a while reviewing how existing performance budget features work in other products, @vinamratasingal, @hwikyounglee and I felt that in the short term we could possibly move ahead with this feature using the above treatment. 

We would defer anything more intricate to a proper perf budget feature (if Lighthouse wanted to adopt one in the future). 

I would be curious to hear if @paulirish @patrickhulce etc. have any strong opinions on direction here either way. Would be great to let @wardpeet know next steps :)

paulirish:
hwi's mock in https://github.com/GoogleChrome/lighthouse/pull/5010#issuecomment-395228656 WFM. 

I think we'll need to add text in the description which says "JavaScript files exceeding ____ are marked below, as they surpass [recommended thresholds]." or something to that effect. 

👍

addyosmani:
> I think we'll need to add text in the description which says "JavaScript files exceeding ____ are marked below, as they surpass [recommended thresholds]." or something to that effect.

👍 

@wardpeet wdyt about moving ahead based on hwi's mock?

wardpeet:
@addyosmani @paulirish @hwikyounglee 
Just to make sure I got it right:
When exceeding the bundle size we get
* transfertime in red
* warning icon
* add extra description why we put things in red

addyosmani:
Did we incorporate "JavaScript files exceeding ____ are marked below, as they surpass [recommended thresholds]." into the description?

paulirish:
Let's link to slightlyoff's post:
```md
JavaScript files exceeding 170kb are marked below, as they surpass [recommended thresholds](https://infrequently.org/2017/10/can-you-afford-it-real-world-web-performance-budgets/).
```

wardpeet:
@paulirish @addyosmani @hwikyounglee 
are we happy with this? Not sure if it's good enough
image

I could also implement the new line markdown feature `  ` 2 spaces as described [here](https://github.github.com/gfm/#hard-line-break) to put it on a new line

wardpeet:
updated
image

addyosmani:
UI wise, I'm pretty happy with what you have now. You'll want to circle back on the unused `networkThroughput`(https://github.com/GoogleChrome/lighthouse/blob/78ba0c15880b616d74fd0dab9d439ee60bae67bd/lighthouse-core/audits/byte-efficiency/total-byte-weight.js#L63) currently being flagged by CI.

jest.config.js
```diff
@@ -11,8 +11,8 @@ module.exports = {
   collectCoverageFrom: [
     '**/lighthouse-core/**/*.js',
     '**/lighthouse-cli/**/*.js',
-    '!**/test/',
-    '!**/scripts/',
+    '!**/test/**/*',
```
wardpeet: this wasn't working before now it excludes the tests and script folder. I noticed it because jest hanged on the dbw-large-file.js which was in test/fixtures

patrickhulce: oh nice! 👍

wardpeet: and the build speed is amazing! 10s less on travis 👏 😆

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
```diff
@@ -60,7 +73,7 @@ class TotalByteWeight extends ByteEfficiencyAudit {
       const result = {
         url: record.url,
         totalBytes: record.transferSize,
-        totalMs: ByteEfficiencyAudit.bytesToMs(record.transferSize, networkThroughput),
+        flagged: TotalByteWeight.hasExceededJSBundleSize(record),
```
paulirish: Would kinda prefer to only add the boolean in the true case. @brendankenny type-wise is that OK?

wardpeet:
I'll close this PR in favor of the new lightwallet/budgets issue/PR:
issue: https://github.com/GoogleChrome/lighthouse/issues/6053
PR: https://github.com/GoogleChrome/lighthouse/pull/7176

PR #5013, opened 2018-04-23
misc(lighthouse-docker): Fix metadata

frvge:
Fixes requests from reviewers.

FYI: @ebidel


frvge:
@ebidel , I'm not sure if this PR to your branch for getting lighthouse to work in docker is still valuable, now that we're at Lighthouse 3. It takes care of some reviews on your branch.

paulirish:
@frvge thanks very much for these! sorry for the long delay.

Issue #5008, opened 2018-04-20
[meta] 2018 I/O Burndown


paulirish:
Channeling our ancestors, we introduce this year's abbreviated I/O burndown list. Also, heads up there are some duplicate checkboxes between here and #4333.

### LHR / Config / API [bckenny]
- [X] decide on groups/tags/scoredisplaymode
- [x] Submit lhr-lite CL to oneplatform #4867 
- [x] node module return is `{lhr, artifacts}` #4999
- [x] support 'html'/'csv' output from node module #4999
- [x] String/printf `displayValue` [@phulce] #5099 
- [ ] Introduce new `details`'s `type:'metric'` 
- [X] Introduce new `details`'s `type:'opportunity'` #5296
- [x] Update scoreDisplayMode, remove booleans, support null scores #5105 
- [x] rename perf-hint => `opportunities`, perf-info => `diagnostics`, perf-metric => `metrics` #5102 
- [x] Introduce `manualDescription` #5100 
- [x] update metrics.js output to be consumable from HTTPArchive. #5101 
- [ ] LHR extends LHR-lite
- [x] Remove `extendedInfo` #4614
   - [x] Initial: exclude it from LHR via `createAuditResult` but audits still report (so our audit tests don't need massive updates
   - [ ] Take inventory of extendedInfo data that's currently not in the associated `details` block, but should be.  (here's [what HA uses](https://github.com/HTTPArchive/beta.httparchive.org/search?utf8=%E2%9C%93&q=audits&type=), what [brendan uses](https://github.com/brendankenny/bq-har/search?utf8=%E2%9C%93&q=extendedInfo&type=))
   - [ ] Update all those tests
- [ ] Apply necessary LHR-lite naming changes #4983:
  * [x] `initialUrl` -> `requestedUrl`, `url` -> `finalUrl` #5127
  * [x] `fetchedAt` -> `fetchTime`, remove `generatedTime` #5112
  * [x] `reportCategories` -> `categories`, `categories` from arrays to keyed objects #5155 
  * [x] `reportGroups` -> `categoryGroups` #5155 
  * [x] `runtimeConfig` -> simplified settings in LHR and create runtimeConfig in report renderer #5122
  * [ ] remove `Category.id` (maybe don't need)
  * [x] `category`: `name` -> `title`, `audits` -> `auditRefs` #5157
  * [x] `audit`: `name` => `id`, `description` -> `title`, `helpText` -> `description`
[...]

Issue #4629, opened 2018-02-26
Perf Metrics & Scoring vNext


paulirish:
Some small changes we plan on making to the metrics we expose and how we score them

* [x] Add FCP to the LH results #3761
* [x] Metrics: rename "First Interactive" and "Time To Consistently Interactive" #3655
* [ ] Redo the weighting
* [x] TTI [moved to depending on FCP](https://docs.google.com/document/d/1yE4YWsusi5wVXrnwhR61j-QyjK9tzENIzfxrCjA1NAk/edit#heading=h.cdpn5719a7nc). we should follow
* [x] [Network Idle was adjusted](https://docs.google.com/document/d/1yE4YWsusi5wVXrnwhR61j-QyjK9tzENIzfxrCjA1NAk/edit#heading=h.3mu7z0sv880k). We can also filter out non-GET and Failed requests

proposal doc: [LH Performance Score 2018Q1 Revision](https://docs.google.com/document/d/1XiKdkLgWul0l40slhkCmQufNdjUiTlq3H_EAjrBH5Cw/edit)

cc #4333

vinamratasingal:
Based on feedback from the LH scoring committee meeting, we have an overall preliminary LGTM but a) need to dig into the feedback a bit more and b) need to see how the new score shapes up on a few websites and dig into the traces a bit more. We're hoping to have this ready before the next scoring committee meeting in late March.

paulirish:
* [ ] TTI [moved to depending on FCP](https://docs.google.com/document/d/1yE4YWsusi5wVXrnwhR61j-QyjK9tzENIzfxrCjA1NAk/edit#heading=h.cdpn5719a7nc). we should follow
* [ ] [Network Idle was adjusted](https://docs.google.com/document/d/1yE4YWsusi5wVXrnwhR61j-QyjK9tzENIzfxrCjA1NAk/edit#heading=h.3mu7z0sv880k). We can also filter out non-GET and Failed requests

PR #5022, opened 2018-04-23
core(scoring): tweak scoring thresholds based on HTTPArchive data

patrickhulce:
tweaks all the metric scoring points according to our agreed upon scheme, 90th percentile becomes a 75, 75th percentile becomes a 50. Pretty much everything got much harsher except for speedIndex.

I also updated 2 non-metric numeric audits, long-cache-ttl and dom-size which were insanely out of whack with the distributions, for them I picked the quantiles which still seemed reasonable (i.e. the real 75th percentile for dom-size was like 30 nodes, or something ridiculously low)

I didn't update bootup-time/mainthread since something seemed fishy with the data, like ~30% of values are 0, which seems impossible that that many sites have 0ms of javascript on mobile.

ref #4629 #4333 #5008

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
@@ -37,9 +37,11 @@ class CacheHeaders extends Audit {
    */
   static get defaultOptions() {
     return {
-      // see https://www.desmos.com/calculator/zokzso8umm
+      // 50th and 75th percentiles HTTPArchive -> 50 and 75
+      // https://bigquery.cloud.google.com/table/httparchive:lighthouse.2018_04_01_mobile?pli=1
+      // see https://www.desmos.com/calculator/8meohdnjbl
       scorePODR: 4 * 1024,
-      scoreMedian: 768 * 1024,
+      scoreMedian: 128 * 1024,
     };
   }
 
--- a/lighthouse-core/audits/consistently-interactive.js
+++ b/lighthouse-core/audits/consistently-interactive.js
@@ -35,9 +35,11 @@ class ConsistentlyInteractiveMetric extends Audit {
    */
   static get defaultOptions() {
     return {
-      // see https://www.desmos.com/calculator/uti67afozh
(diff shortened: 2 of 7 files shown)


lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -37,9 +37,11 @@ class CacheHeaders extends Audit {
    */
   static get defaultOptions() {
     return {
-      // see https://www.desmos.com/calculator/zokzso8umm
+      // 50th and 75th percentiles HTTPArchive -> 50 and 75
+      // https://bigquery.cloud.google.com/table/httparchive:lighthouse.2018_04_01_mobile?pli=1
```
brendankenny: are there links for any of these in https://httparchive.org/reports? Nicer experience than just a link to the bigquery table, but not sure if @rviscomi has one for each LH audit

patrickhulce:
> did you tweak these by hand until the 75th percentile lined up with the desired value?

yeah I had already done the metric ones for our meeting to propose it, and there really aren't that many others so programmatic seemed like overkill at that point

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
patrickhulce: I can link to the state of javascript/byte weight when we do that, but yeah if there are premade ones that'd be even better :D

patrickhulce: I also thought I remembered httparchive having # dom elements but I couldn't find the report where I expected and ended up querying ours instead

rviscomi: The legacy site tracks average DOM element count: https://legacy.httparchive.org/trends.php#bytesHtmlDoc&numDomElements.

If there are any stats you'd like to see added to the redesigned site, please [file an issue](https://github.com/HTTPArchive/httparchive.org/issues) and consider contributing, I'd be happy to help with the setup.

PR #5024, opened 2018-04-24
core(metrics): add lantern EIL

patrickhulce:
Adds a lantern implementation of estimated input latency and moves the metric to a computed artifact. The coefficients are currently best guess. I'll do one final pass for coeffs on all other metrics since the 3 original haven't been tweaked in many iterations, but accuracy stats unfortunately unavailable since we just introduced this rolling window metric in v3 as well. This is the only non-lantern metric ATM and seems worth giving a try to evaluate later

ref #5008


lighthouse-core/test/gather/computed/metrics/estimated-input-latency-test.js
```diff
@@ -0,0 +1,68 @@
...
+
+/* eslint-env mocha */
+
+describe('Metrics: EIL', () => {
+  it('should compute a simulated value', async () => {
+    const artifacts = Runner.instantiateComputedArtifacts();
+    const settings = {throttlingMethod: 'simulate'};
+    const result = await artifacts.requestEstimatedInputLatency({trace, devtoolsLog, settings});
+
+    assert.equal(Math.round(result.timing), 104);
```
patrickhulce: this is a legitimate difference, basically the original trace has ~400ms of mainthread work that escapes in the freebie 500ms of a 5s window, in Lantern this gets 4x'd and isn't hidden anymore

lighthouse-core/audits/estimated-input-latency.js
```diff
@@ -7,16 +7,7 @@
 
 const Audit = require('./audit');
 const Util = require('../report/v2/renderer/util');
-const TracingProcessor = require('../lib/traces/tracing-processor');
```
brendankenny: I think the conflict is just `v2/` -> `html/`

PR #5035, opened 2018-04-24
core(lhr): savingsMs, savingsBytes

paulirish:
proposal to rename.

Helps to differentiate between the item-level `wastedMs`. Got mixed up between the two just a moment ago. :)


paulirish:
k sg. 

@brendankenny we can update this in the v1alpha2 CL the next time you're in there.

PR #5039, opened 2018-04-25
core(render-blocking): address followup feedback

patrickhulce:
followup to #4995

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
+++ b/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
@@ -49,7 +49,9 @@ class RenderBlockingResources extends Audit {
         'Resources are blocking the first paint of your page. Consider ' +
         'delivering critical JS/CSS inline and deferring all non-critical ' +
         'JS/styles. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
-      requiredArtifacts: ['CSSUsage', 'URL', 'TagsBlockingFirstPaint', 'traces'],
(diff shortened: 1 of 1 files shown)


lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -49,7 +49,9 @@ class RenderBlockingResources extends Audit {
         'Resources are blocking the first paint of your page. Consider ' +
         'delivering critical JS/CSS inline and deferring all non-critical ' +
         'JS/styles. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
-      requiredArtifacts: ['CSSUsage', 'URL', 'TagsBlockingFirstPaint', 'traces'],
+      // This audit also looks at CSSUsage but has a graceful fallback if it failed, so do not mark
+      // it as a "requiredArtifact".
+      requiredArtifacts: ['URL', 'TagsBlockingFirstPaint', 'traces'],
```
brendankenny: I like this for working around gatherer errors, but won't this be undesirable if doing an `--only-categories`/`audits`? The `CSSUsage` gatherer could be pruned. It won't be a failure but probably not what you want.

Time to introduce a `desiredArtifacts` field? :D

patrickhulce: This is also true. I have a slight preference for eliminating the failure case in exchange for making the `onlyAudits: ['render-blocking-resource']` case a little more forgiving.

I'll add a todo to look into updating after an `optionalArtifacts` :D

PR #5041, opened 2018-04-25
core(config): switch to lantern by default, add presets

patrickhulce:
## Changes
* converts the default throttling method to lantern `'simulate'`
* reduces the default network quiet/cpu quiet to 1000ms, this is a bit less aggressive than our old original values but still much improved, and we could increase this if needed, but the full 5s should definitely be unnecessary
* replaces `--perf` and `--mixed-content` with `--preset=perf/mixed-content`, there's been some opposition to this before, it wouldn't be incredibly laborious to keep so lmk
* old defaults are runnable with `--throttling-method=devtools`

## Yet to Change

The perfy smoke tests made some assumptions about the network/CPU quiet thresholds that don't hold up anymore, so left those testing the devtools path for now and will update those as I finish up the opportunity parity, on the plus side rest of smoketest time went from ~100s to ~80s :)

Also left the golden LHR diffing on observed for now to reduce the noise, seems like we'll want to be checking both in our tests anyhow once there's 100% parity?


lighthouse-cli/bin.js
```diff
@@ -54,12 +52,13 @@ if (cliFlags.configPath) {
   // Resolve the config file path relative to where cli was called.
   cliFlags.configPath = path.resolve(process.cwd(), cliFlags.configPath);
   config = /** @type {LH.Config.Json} */ (require(cliFlags.configPath));
-} else if (cliFlags.perf) {
-  config = perfOnlyConfig;
-} else if (cliFlags.mixedContent) {
-  config = mixedContentConfig;
-  // The mixed-content audits require headless Chrome (https://crbug.com/764505).
-  cliFlags.chromeFlags = `${cliFlags.chromeFlags} --headless`;
+} else if (cliFlags.configPreset) {
```
paulirish: how about just `--preset` :D

lighthouse-core/config/perf-config.js
```diff
@@ -6,7 +6,7 @@
 'use strict';
 
 module.exports = {
-  extends: 'lighthouse:default',
+  extends: 'lighthouse:observed',
```
paulirish: I feel like the "observed" name refers to observed things (metrics, etc), regardless of throttling method.

Since this config is primarily about changing throttling method I would have expected a name like 'throttled'

lighthouse-cli/bin.js
patrickhulce: wfm, @brendankenny ?

lighthouse-core/config/perf-config.js
patrickhulce: I started out with `devtools` as the name but the primary value of this thing existing is having the 5s+ wait times in the defaultPass which applies to all observed methods. It does indeed change the throttlingMethod as well, but that's easily done without presets/extension. `throttled` overly communicates the value in changing the throttling IMO, i.e. why would I want `--preset=throttled --throttlingMethod=off/provided`?

The most consistent might be to make `observed` not actually apply the throttling by default, and require `--throttlingMethod=devtools` but that seems to discourage throttling which isn't great.

lighthouse-cli/bin.js
brendankenny: > `--preset`

yeah, I think it reads well in context

lighthouse-core/config/perf-config.js
paulirish: I just think its confusing that observedMetrics all exist outside of the `lighthouse:observed` config.



better question: when is observed used? for --preset=perf and for the ttci smoketest, right? why doesn't ttci smoke just use/extend perf config?

lighthouse-core/config/config.js
```diff
@@ -187,7 +188,7 @@ function deepClone(json) {
   // injection of plugins.
   if (Array.isArray(json.passes)) {
     cloned.passes.forEach((pass, i) => {
-      pass.gatherers = cloneArrayWithPluginSafety(json.passes[i].gatherers);
+      pass.gatherers = cloneArrayWithPluginSafety(json.passes[i].gatherers || []);
```
brendankenny: what happened here? Don't we require at least an empty array in the passes?

lighthouse-core/config/perf-config.js
brendankenny: agree that `observed` is a little confusing here. It does sound like it's fairly neutral, observed regardless of throttling method, but if I was told it affected the throttling method I would guess it was closer to `none`, not using devtools throttling.

But it sounds like you may be rearranging things, so maybe I'll wait to comment further :)

lighthouse-core/config/config.js
patrickhulce: Well IMO, we shouldn't and the new defaults for passes make sure it's at least an empty array, but that doesn't work with the cloning, I can revert if you feel strongly all passes should declare empty gatherers.

patrickhulce:
OK ready for re-review I believe! no more observed config or extension, config.js just forces the defaultPass quiet thresholds to 5.25s+ if the throttling method is devtools or provided

patrickhulce:
friendly ping here, review please! :)

lighthouse-core/config/config.js
```diff
@@ -381,6 +383,26 @@ class Config {
...
+  static adjustDefaultPassForThrottling(config) {
+    if (config.settings.throttlingMethod !== 'devtools' &&
+        config.settings.throttlingMethod !== 'provided') {
+      return;
+    }
+
+    const defaultPass = config.passes.find(pass => pass.passName === 'defaultPass');
+    if (!defaultPass) return;
+
+    defaultPass.pauseAfterLoadMs = Math.max(5250, defaultPass.pauseAfterLoadMs);
```
paulirish: feels like this 5250 magic number should be defined in constants.js

package.json
```diff
@@ -60,7 +60,7 @@
     "changelog": "conventional-changelog --config ./build/changelog-generator/index.js --infile changelog.md --same-file",
     "type-check": "tsc -p .",
     "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
-    "update:sample-json": "node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --output=json --output-path=./lighthouse-core/test/results/sample_v2.json http://localhost/dobetterweb/dbw_tester.html && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing",
+    "update:sample-json": "node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --throttling-method=devtools --output=json --output-path=./lighthouse-core/test/results/sample_v2.json http://localhost/dobetterweb/dbw_tester.html && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing",
```
paulirish: doing this to avoid the mixed settings error, right?
next run of update-report-fixtures will be throttling-method: 'simulate', right?

so a followup PR to update both makes sense to me.

paulirish: but speaking of. perhaps we should also have 1 golden LHR for simulate and 1 for throttled.  Feels like we have a blind spot otherwise.

lighthouse-core/config/config.js
patrickhulce: good call done

package.json
patrickhulce: > Also left the golden LHR diffing on observed for now to reduce the noise, seems like we'll want to be checking both in our tests anyhow once there's 100% parity?

totally agree ;)

patrickhulce:
if everyone is afraid of this one we should get it in sooner rather than later to iron out the kinks ;)

lighthouse-cli/test/smokehouse/run-smoke.js
```diff
@@ -63,7 +63,7 @@ const SMOKETESTS = [{
   id: 'offline',
   expectations: 'offline-local/offline-expectations.js',
   config: smokehouseDir + 'offline-config.js',
-  batch: 'parallel-second',
+  batch: 'offline',
```
patrickhulce: @paulirish I made this change after too many flakes on travis, now that things aren't throttled it seems too much is trying to happen at once and at least one navstart was getting dropped on ~70% of runs :(

paulirish: aiight.

lighthouse-core/config/constants.js
```diff
@@ -60,8 +60,15 @@ const defaultPassConfig = {
   gatherers: [],
 };
 
+const observedPassConfigOverrides = {
```
paulirish: `throttledPassConfigDefaults`
or
`nonSimulatedPassConfigOverrides`

lighthouse-core/config/config.js
```diff
@@ -381,6 +383,30 @@ class Config {
...
+    if (config.settings.throttlingMethod !== 'devtools' &&
+        config.settings.throttlingMethod !== 'provided') {
+      return;
+    }
+
+    const defaultPass = config.passes.find(pass => pass.passName === 'defaultPass');
+    if (!defaultPass) return;
+
+    const overrides = constants.observedPassConfigOverrides;
+    defaultPass.pauseAfterLoadMs =
```
paulirish: i wouldn't yell bloody murder if you renamed this to `dPass` or something to avoid all the wrapppppppppppppppping

paulirish:
🍫  :yum:

lighthouse-core/config/config.js
patrickhulce: heh, yeah I spent more time than I should to save this one, basically nothing saves the last one (triple `networkQuietThresholdMs` is just too much` so I gave up, at least here it's consistent and easy to understand

readme.md
```diff
@@ -58,7 +58,6 @@ Configuration:
                                  CHROME_PATH: Explicit path of intended Chrome binary. If set must point to an executable of a build of
                                  Chromium version 54.0 or later. By default, any detected Chrome Canary or Chrome (stable) will be launched.
                                                                                                                                        [default: ""]
-  --perf                         Use a performance-test-only configuration                                                                 [boolean]
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -22,7 +21,7 @@ const log = require('lighthouse-logger');
 window.runLighthouseForConnection = function(
     connection, url, options, categoryIDs,
     updateBadgeFn = function() { }) {
-  const config = options && options.fastMode ? new Config(fastConfig, options.flags) : new Config({
```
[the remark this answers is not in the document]
patrickhulce: it's just set by lightrider, all runs for them will be simulated now anyway

lighthouse-core/test/config/config-test.js
```diff
@@ -454,6 +454,43 @@ describe('Config', () => {
...
+    const config = new Config({
+      extends: 'lighthouse:default',
+      settings: {
+        throttlingMethod: 'devtools',
+        onlyCategories: ['performance'],
+      },
+    });
+
+    assert.equal(config.settings.throttlingMethod, 'devtools');
+    assert.ok(config.passes[0].pauseAfterLoadMs >= 5000, 'did not adjust load quiet ms');
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/config/config.js
```diff
@@ -381,6 +383,30 @@ class Config {
...
+   *
+   * @param {LH.Config.Json} config
+   */
+  static adjustDefaultPassForThrottling(config) {
+    if (config.settings.throttlingMethod !== 'devtools' &&
+        config.settings.throttlingMethod !== 'provided') {
+      return;
+    }
+
+    const defaultPass = config.passes.find(pass => pass.passName === 'defaultPass');
```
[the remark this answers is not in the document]
patrickhulce: discussed offline, it's an impl detail of TTI that we need 5s+ which is hardcoded to `defaultPass` so we're sticking to defaultPass for now 👍

patrickhulce:
@brendankenny LGTY?

brendankenny:
ju9dxm5

PR #5051, opened 2018-04-26
core(tsc): add type checking to computed artifacts

brendankenny:
Important first step to refactoring computed artifacts to be just regular requires. Big PR, but almost all type stuff.

Introduces the lie that the computed artifact `request*()` methods are part of the `Artifacts` type, which isn't usually true except when calling `myaudit.audit(artifacts, auditContext)`. Follow ups will untangle that.

Only real code changes:
- The only time that the `computedArtifacts` object realized it also had all the artifacts on it was in `MainResource`, where `artifacts.URL` was accessed. Instead, `URL` is now passed in as an artifact dependency (as it should be), but that requires `CriticalRequestChains` also needing `URL` passed in. This needed a lot of small test tweaks.
- The lantern simulator needs a settings object passed in, even if it's empty, to keep tsc happy about destructuring. This required some scattered `settings: {}` around.
- the `ComputedArtifact.compute_` method now *has* to be async. It was too annoying dealing with it either being sync (`URL`, `TraceOfTab`, maybe one more?) or async, so just easier this way. Required a good number of additional async/await scattered around, especially in tests.

Pretty much everything else is just adding type annotations or making code a little more explicit for the compiler.

### DIFF
--- a/lighthouse-core/audits/seo/is-crawlable.js
+++ b/lighthouse-core/audits/seo/is-crawlable.js
@@ -68,7 +68,7 @@ class IsCrawlable extends Audit {
       helpText: 'Search engines are unable to include your pages in search results ' +
           'if they don\'t have permission to crawl them. [Learn ' +
           'more](https://developers.google.com/web/tools/lighthouse/audits/indexing).',
-      requiredArtifacts: ['MetaRobots', 'RobotsTxt'],
+      requiredArtifacts: ['MetaRobots', 'RobotsTxt', 'URL'],
     };
   }
 
@@ -77,7 +77,9 @@ class IsCrawlable extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    return artifacts.requestMainResource(artifacts.devtoolsLogs[Audit.DEFAULT_PASS])
+    const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
+
+    return artifacts.requestMainResource({devtoolsLog, URL: artifacts.URL})
       .then(mainResource => {
         const blockingDirectives = [];
 
--- a/lighthouse-core/gather/computed/computed-artifact.js
+++ b/lighthouse-core/gather/computed/computed-artifact.js
@@ -3,35 +3,44 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
-// @ts-nocheck
 'use strict';
 
 const ArbitraryEqualityMap = require('../../lib/arbitrary-equality-map');
 
 class ComputedArtifact {
   /**
-   * @param {!ComputedArtifacts} allComputedArtifacts
+   * @param {LH.ComputedArtifacts} allComputedArtifacts
    */
   constructor(allComputedArtifacts) {
-    /** @private {!Map} */
-    this._cache = new ArbitraryEqualityMap();
-    this._cache.setEqualityFn(ArbitraryEqualityMap.deepEquals);
+    const cache = new ArbitraryEqualityMap();
+    cache.setEqualityFn(ArbitraryEqualityMap.deepEquals);
 
-    /** @private {!ComputedArtifacts} */
+    /** @type {Map<FirstParamType<this['compute_']>, Promise<ReturnType<this['compute_']>>>} */
+    // @ts-ignore cache is close enough to a Map for our purposes (but e.g. no [Symbol.toStringTag])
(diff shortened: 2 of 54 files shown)


typings/artifacts.d.ts
```diff
@@ -254,16 +317,32 @@ declare global {
...
+        first: number;
+        speedIndex: number;
+        frames: {
+          getProgress(): number;
+          getTimeStamp(): number;
+          isProgressInterpolated(): number;
+        }[]
+      }
+
+      // TODO(bckenny): all but navigationStart could actually be undefined.
```
brendankenny: these should be updated in a follow up, but it had too many places where we're treating them as always defined. Whether we were correct or not (and we'll get to fix some bugs) remains to be seen.

lighthouse-core/gather/computed/computed-artifact.js
```diff
@@ -3,53 +3,67 @@
...
+   * `compute_` method on classes derived from ComputedArtifact.
+   * @param {FirstParamType<this['compute_']>} requiredArtifacts
+   * @return {Promise<ReturnType<this['compute_']>>}
    */
-  request(artifacts) {
-    if (this._cache.has(artifacts)) {
-      return Promise.resolve(this._cache.get(artifacts));
+  async request(requiredArtifacts) {
+    const computed = this._cache.get(requiredArtifacts);
+    if (computed) {
```
patrickhulce: kind of a pain that TS can't process `.has` in the same way, that's kind of the point of `.has` :)

lighthouse-core/gather/computed/manifest-values.js
```diff
@@ -23,86 +25,96 @@ class ManifestValues extends ComputedArtifact {
     return ['hasManifest', 'hasParseableManifest'];
   }
 
+  /** @typedef {(val: NonNullable<LH.Artifacts.Manifest['value']>) => boolean} validator */
```
patrickhulce: lowercase typedef name? 😮

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -163,7 +172,7 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
       let minCandidate = null;
       let minDistance = Infinity;
       // Find the closest request that finished before this CPU task started
-      candidates.forEach(candidate => {
```
patrickhulce: does this really change things type interference wise? or just drive-by perf/style?

lighthouse-core/gather/computed/page-dependency-graph.js
```diff
@@ -173,26 +182,29 @@ class PageDependencyGraphArtifact extends ComputedArtifact {
...
       cpuNode.addDependency(minCandidate);
     }
 
+    /** @type {Map<string, CPUNode>} */
     const timers = new Map();
     for (const node of cpuNodes) {
       for (const evt of node.childEvents) {
         if (!evt.args.data) continue;
 
-        const url = evt.args.data.url;
```
patrickhulce: dataUrl makes me think of a data URI, which other url is this prefix disambiguating?

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -77,7 +77,9 @@ class IsCrawlable extends Audit {
    * @return {!AuditResult}
    */
   static audit(artifacts) {
-    return artifacts.requestMainResource(artifacts.devtoolsLogs[Audit.DEFAULT_PASS])
+    const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
+
+    return artifacts.requestMainResource({devtoolsLog, URL: artifacts.URL})
```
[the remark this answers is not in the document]
brendankenny: yeah, I thing simplifying it is a good idea. This PR I guess was mostly about making everything explicit so it'll be easier to refactor :)

lighthouse-core/gather/computed/computed-artifact.js
brendankenny: https://github.com/Microsoft/TypeScript/issues/18781 but they don't sound very enthusiastic :)

lighthouse-core/gather/computed/manifest-values.js
brendankenny: fixed

lighthouse-core/gather/computed/page-dependency-graph.js
brendankenny: yeah, tsc bug (or limitation). `minCadidate` lost its `null | NetworkNode` type after the `forEach()`, and with the if eliminating falsey values it ended up thinking it was `cpuNode.addDependency(never);`

lighthouse-core/gather/computed/page-dependency-graph.js
brendankenny: purely style. I was wondering where `url` came from below, then got briefly confused by the various `forEach`es. Changed to `argsUrl`...wdyt? `evtUrl`? Can also revert.

brendankenny:
@patrickhulce you can land #5024 and I'll update this

brendankenny:
updated!

brendankenny:
scoped types and rebased :)

PR #5052, opened 2018-04-26
tests: fix cli handling empty runnerResult during gatherMode

paulirish:
the destructure in `saveResults` broke cases where gatherMode existed runner early with an `undefined` runnerResult

plenty of others ways to solve this. happy to do them, though this seemed like the most non-invasive.

-----

driveby fix: avoid comparing `output` much like gather/auditMode as the `output` is irrelevant to a gather run.

### DIFF
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -100,12 +100,12 @@ function handleError(err) {
  * @return {Promise<void>}
  */
 function saveResults(runnerResult, flags) {
-  const {lhr, artifacts} = runnerResult;
   const cwd = process.cwd();
   let promise = Promise.resolve();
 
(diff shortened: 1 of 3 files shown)


lighthouse-cli/run.js
```diff
@@ -100,12 +100,12 @@ function handleError(err) {
  * @return {Promise<void>}
  */
 function saveResults(runnerResult, flags) {
-  const {lhr, artifacts} = runnerResult;
   const cwd = process.cwd();
   let promise = Promise.resolve();
 
   const shouldSaveResults = flags.auditMode || (flags.gatherMode === flags.auditMode);
   if (!shouldSaveResults) return promise;
+  const {lhr, artifacts} = runnerResult;
```
brendankenny: clearly we need typechecking in here :)

PR #5058, opened 2018-04-27
report: add auditIds to DOM. update pptr tests to assert against them.

paulirish:
audit ids making it far more clear:
image

Failure that ward, brendan and me faced before:
image

New failure:
image


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -30,6 +30,7 @@ class CategoryRenderer {
   renderAudit(audit) {
     const tmpl = this.dom.cloneTemplate('#tmpl-lh-audit', this.templateContext);
     const auditEl = this.dom.find('.lh-audit', tmpl);
+    auditEl.id = audit.result.name;
```
brendankenny: this won't be unique if we do the key/name split, but we can probably cross that bridge when we come to it since we don't have an alternative name at this point

PR #5062, opened 2018-04-28
core(preload): use lantern to compute savings

patrickhulce:
this ended up turning into quite the behemoth, I was tempted to break up the lantern/connection-pool test changes, but it's kinda hard to grok without the motivating example.

most of the lines are tests, so that's the good news :)

### DIFF
--- a/lighthouse-core/audits/uses-rel-preload.js
+++ b/lighthouse-core/audits/uses-rel-preload.js
@@ -51,54 +51,117 @@ class UsesRelPreloadAudit extends Audit {
   }
 
   /**
+   * Computes the estimated effect of preloading all the resources.
+   * @param {Set<string>} urls The array of byte savings results per resource
+   * @param {Node} graph
+   * @param {Simulator} simulator
+   * @param {LH.WebInspector.NetworkRequest} mainResource
+   * @return {{wastedMs: number, results: Array<{url: string, wastedMs: number}>}}
+   */
+  static computeWasteWithGraph(urls, graph, simulator, mainResource) {
+    if (!urls.size) {
+      return {wastedMs: 0, results: []};
+    }
+
+    const simulationBeforeChanges = simulator.simulate(graph, {ignoreObserved: true});
+
+    const modifiedGraph = graph.cloneWithRelationships();
+
+    const nodesToPreload = [];
+    /** @type {Node|null} */
+    let mainDocumentNode = null;
+    modifiedGraph.traverse(node => {
+      if (node.record && urls.has(node.record.url)) {
+        nodesToPreload.push(node);
+      }
+
+      if (node.record && node.record.url === mainResource.url) {
+        mainDocumentNode = node;
+      }
+    });
+
+    if (!mainDocumentNode) {
+      // Should always find the main document node
+      throw new Error('Could not find main document node');
+    }
+
+    // Preload has the effect of moving the resource's only dependency to the main HTML document
+    // Remove all dependencies of the nodes
+    for (const node of nodesToPreload) {
+      node.removeAllDependencies();
+      node.addDependency(mainDocumentNode);
+    }
+
+    const simulationAfterChanges = simulator.simulate(modifiedGraph, {ignoreObserved: true});
+    const originalNodesByRecord = Array.from(simulationBeforeChanges.nodeTimings.keys())
+        .reduce((map, node) => map.set(node.record, node), new Map());
+
+    const results = [];
+    for (const node of nodesToPreload) {
+      const originalNode = originalNodesByRecord.get(node.record);
+      const timingAfter = simulationAfterChanges.nodeTimings.get(node);
+      const timingBefore = simulationBeforeChanges.nodeTimings.get(originalNode);
+      const wastedMs = Math.round(timingBefore.endTime - timingAfter.endTime);
+      if (wastedMs < THRESHOLD_IN_MS) continue;
+      results.push({url: node.record.url, wastedMs});
+    }
+
+    if (!results.length) {
+      return {wastedMs: 0, results};
+    }
+
+    return {
+      // Preload won't necessarily impact the deepest chain/overall time
+      // We'll use the maximum endTime improvement for now
+      wastedMs: Math.max(...results.map(item => item.wastedMs)),
+      results,
+    };
+  }
+
+  /**
    * @param {!Artifacts} artifacts
    * @return {!AuditResult}
    */
-  static audit(artifacts) {
-    const devtoolsLogs = artifacts.devtoolsLogs[UsesRelPreloadAudit.DEFAULT_PASS];
+  static audit(artifacts, context) {
+    const trace = artifacts.traces[UsesRelPreloadAudit.DEFAULT_PASS];
+    const devtoolsLog = artifacts.devtoolsLogs[UsesRelPreloadAudit.DEFAULT_PASS];
+    const simulatorOptions = {trace, devtoolsLog, settings: context.settings};
 
     return Promise.all([
-      artifacts.requestCriticalRequestChains(devtoolsLogs),
-      artifacts.requestMainResource(devtoolsLogs),
-    ]).then(([critChains, mainResource]) => {
-      const results = [];
-      let maxWasted = 0;
+      // TODO(phulce): eliminate dependency on CRC
+      artifacts.requestCriticalRequestChains(devtoolsLog),
+      artifacts.requestMainResource(devtoolsLog),
+      artifacts.requestPageDependencyGraph({trace, devtoolsLog}),
+      artifacts.requestLoadSimulator(simulatorOptions),
+    ]).then(([critChains, mainResource, graph, simulator]) => {
       // get all critical requests 2 + mainResourceIndex levels deep
       const mainResourceIndex = mainResource.redirects ? mainResource.redirects.length : 0;
 
       const criticalRequests = UsesRelPreloadAudit._flattenRequests(critChains,
         3 + mainResourceIndex, 2 + mainResourceIndex);
-      criticalRequests.forEach(request => {
-        const networkRecord = request;
+      const urls = new Set();
+      for (const networkRecord of criticalRequests) {
         if (!networkRecord._isLinkPreload && networkRecord.protocol !== 'data') {
-          // calculate time between mainresource.endTime and resource start time
-          const wastedMs = Math.min(request._startTime - mainResource._endTime,
-            request._endTime - request._startTime) * 1000;
-
-          if (wastedMs >= THRESHOLD_IN_MS) {
-            maxWasted = Math.max(wastedMs, maxWasted);
-            results.push({
-              url: request.url,
-              wastedMs: Util.formatMilliseconds(wastedMs),
-            });
-          }
+          urls.add(networkRecord._url);
         }
-      });
+      }
 
+      const {results, wastedMs} = UsesRelPreloadAudit.computeWasteWithGraph(urls, graph, simulator,
+          mainResource);
       // sort results by wastedTime DESC
       results.sort((a, b) => b.wastedMs - a.wastedMs);
 
       const headings = [
         {key: 'url', itemType: 'url', text: 'URL'},
-        {key: 'wastedMs', itemType: 'text', text: 'Potential Savings'},
+        {key: 'wastedMs', itemType: 'ms', text: 'Potential Savings', granularity: 10},
       ];
-      const summary = {wastedMs: maxWasted};
+      const summary = {wastedMs};
       const details = Audit.makeTableDetails(headings, results, summary);
 
       return {
-        score: UnusedBytes.scoreForWastedMs(maxWasted),
-        rawValue: maxWasted,
-        displayValue: Util.formatMilliseconds(maxWasted),
+        score: UnusedBytes.scoreForWastedMs(wastedMs),
+        rawValue: wastedMs,
+        displayValue: Util.formatMilliseconds(wastedMs),
         extendedInfo: {
           value: results,
         },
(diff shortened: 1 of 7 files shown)


lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -281,10 +292,13 @@ class Simulator {
...
    * Estimates the time taken to process all of the graph's nodes.
    * @param {Node} graph
+   * @param {{ignoreObserved?: boolean}=} options
    * @return {LH.Gatherer.Simulation.Result}
    */
-  simulate(graph) {
+  simulate(graph, options) {
+    options = Object.assign({ignoreObserved: false}, options);
     // initialize the necessary data containers
+    this._ignoreObserved = options.ignoreObserved;
```
patrickhulce: I'm really not sure what to call this thing, basically it's "don't try and stick just to how the page loaded in the trace, feel free to load it as efficiently as you think possible"

paulirish: yeah i don't love this term. :)  as you know i don't love reusing it in different places in diff contexts. 

is `reconstructTCPConnectionPooling` accurate?

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -82,15 +82,21 @@ module.exports = class ConnectionPool {
...
+      // Make sure each origin has 6 connections available
+      // https://cs.chromium.org/chromium/src/net/socket/client_socket_pool_manager.cc?type=cs&q="int+g_max_sockets_per_group"
+      while (connections.length < 6) connections.push(connections[0].clone());
+      while (connections.length > 6) connections.pop();
+
       this._connectionsByOrigin.set(origin, connections);
     }
   }
 
   /**
```
paulirish: can we add a few lines to describe what's happening in here?

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -99,16 +105,29 @@ module.exports = class ConnectionPool {
...
-    if (!connection) return null;
-    this._connectionsInUse.add(connection);
-    this._connectionsByRecord.set(record, connection);
-    return connection;
+    const needsColdConnection = !options.ignoreObserved &&
+        !this._connectionReusedByRequestId.get(record.requestId);
+    const needsWarmConnection = !options.ignoreObserved &&
+        this._connectionReusedByRequestId.get(record.requestId);
+
+    let connectionToUse = availableWarmConnection || availableColdConnection;
```
paulirish: still not sure exactly what logic you're trying to apply here, but it does look odd there's no `else` case in all 4 lines here.

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -82,15 +82,21 @@ module.exports = class ConnectionPool {
         throw new Error(`Could not find a connection for origin: ${origin}`);
       }
 
+      // Make sure each origin has 6 connections available
+      // https://cs.chromium.org/chromium/src/net/socket/client_socket_pool_manager.cc?type=cs&q="int+g_max_sockets_per_group"
+      while (connections.length < 6) connections.push(connections[0].clone());
+      while (connections.length > 6) connections.pop();
```
paulirish: how would we end up with >6?  our network records indicate we have >6 even though that should be impossible?

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -99,16 +105,29 @@ module.exports = class ConnectionPool {
...
+    const availableColdConnection = sortedConnections.find(connection => {
+      return !connection.isWarm() && !this._connectionsInUse.has(connection);
     });
 
-    if (!connection) return null;
-    this._connectionsInUse.add(connection);
-    this._connectionsByRecord.set(record, connection);
-    return connection;
+    const needsColdConnection = !options.ignoreObserved &&
+        !this._connectionReusedByRequestId.get(record.requestId);
```
paulirish: meta point: it's really funky to be mixing this "connection reused" state that refers to observed network activity with what's "available" and "needed" for the simulation. I wish these were clearer

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
patrickhulce: If Chrome killed connections and needed to reopen them, we can kill this line though

lighthouse-core/audits/uses-rel-preload.js
```diff
@@ -50,56 +50,119 @@ class UsesRelPreloadAudit extends Audit {
...
+    }
+
+    if (!results.length) {
+      return {wastedMs: 0, results};
+    }
+
+    return {
+      // Preload won't necessarily impact the deepest chain/overall time
+      // We'll use the maximum endTime improvement for now
+      wastedMs: Math.max(...results.map(item => item.wastedMs)),
```
brendankenny: seems like it gets most of the way to an opportunity but then stops here? Is measuring impact on TTI or whatever problematic?

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -9,56 +9,105 @@
...
+    const mainDocumentNode = buildNode(2, 'http://www.example.com');
+    const scriptNode = buildNode(3, 'http://www.example.com/script.js');
+    const scriptAddedNode = buildNode(4, 'http://www.example.com/script-added.js');
+
+    mainDocumentNode.addDependency(rootNode);
+    scriptNode.addDependency(mainDocumentNode);
+    scriptAddedNode.addDependency(scriptNode);
+
+    mockGraph = rootNode;
+    mockSimulator = {
```
brendankenny: is this reused for the other tests?

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -9,56 +9,105 @@
...
-    },
-    requestNetworkRecords: () => networkRecords,
-    requestMainResource: () => {
-      return Promise.resolve(mainResource);
-    },
+describe('Performance: uses-rel-preload audit', () => {
+  let mockGraph;
+  let mockSimulator;
+
+  const mockArtifacts = (networkRecords, mockChain, mainResource = defaultMainResource) => {
```
brendankenny: the audit has some non-trivial uses of the graph and simulator, which means we're doing a lot of testing of the mocks :) Some tests using the full real thing would be 👍 👍

lighthouse-core/lib/dependency-graph/simulator/tcp-connection.js
```diff
@@ -88,6 +102,14 @@ class TcpConnection {
     this._h2OverflowBytesDownloaded = bytes;
   }
 
+  /**
+   * @return {TcpConnection}
+   */
+  clone() {
+    // @ts-ignore
+    return Object.assign(new TcpConnection(), this);
```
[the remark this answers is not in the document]
patrickhulce: done 👍

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -212,7 +223,7 @@ class Simulator {
 
     const record = /** @type {NetworkNode} */ (node).record;
     const timingData = this._nodeTimings.get(node);
-    const connection = /** @type {TcpConnection} */ (this._connectionPool.acquire(record));
+    const connection = /** @type {TcpConnection} */ (this._acquireConnection(record));
```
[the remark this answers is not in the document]
patrickhulce: commented

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -255,7 +266,7 @@ class Simulator {
     if (node.type !== Node.TYPES.NETWORK) throw new Error('Unsupported');
 
     const record = /** @type {NetworkNode} */ (node).record;
-    const connection = /** @type {TcpConnection} */ (this._connectionPool.acquire(record));
+    const connection = /** @type {TcpConnection} */ (this._acquireConnection(record));
```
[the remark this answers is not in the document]
patrickhulce: commented

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -82,15 +82,27 @@ module.exports = class ConnectionPool {
         throw new Error(`Could not find a connection for origin: ${origin}`);
       }
 
+      // Make sure each origin has 6 connections available
+      // https://cs.chromium.org/chromium/src/net/socket/client_socket_pool_manager.cc?type=cs&q="int+g_max_sockets_per_group"
+      while (connections.length < 6) connections.push(connections[0].clone());
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/audits/uses-rel-preload.js
```diff
@@ -50,56 +50,119 @@ class UsesRelPreloadAudit extends Audit {
...
+      artifacts.requestLoadSimulator(simulatorOptions),
+    ]).then(([critChains, mainResource, graph, simulator]) => {
       // get all critical requests 2 + mainResourceIndex levels deep
       const mainResourceIndex = mainResource.redirects ? mainResource.redirects.length : 0;
 
       const criticalRequests = UsesRelPreloadAudit._flattenRequests(critChains,
         3 + mainResourceIndex, 2 + mainResourceIndex);
-      criticalRequests.forEach(request => {
-        const networkRecord = request;
+      const urls = new Set();
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/audits/uses-rel-preload.js
```diff
@@ -50,56 +50,119 @@ class UsesRelPreloadAudit extends Audit {
...
+    }
+
+    // Preload has the effect of moving the resource's only dependency to the main HTML document
+    // Remove all dependencies of the nodes
+    for (const node of nodesToPreload) {
+      node.removeAllDependencies();
+      node.addDependency(mainDocumentNode);
+    }
+
+    const simulationAfterChanges = simulator.simulate(modifiedGraph, {flexibleOrdering: true});
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -281,10 +292,13 @@ class Simulator {
   /**
    * Estimates the time taken to process all of the graph's nodes.
    * @param {Node} graph
+   * @param {{flexibleOrdering?: boolean}=} options
```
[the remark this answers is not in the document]
patrickhulce: done 👍

lighthouse-core/test/audits/uses-rel-preload-test.js
patrickhulce: nope, just scoped for the `mockArtifacts` to be able to return it

lighthouse-core/test/audits/uses-rel-preload-test.js
patrickhulce: totally agreed, but it requires a trace and devtoolslog of a site that would benefit from preloading which can be quite large.

my proposal:
* we rely on the smoke test + unit tests for now
* for cases like this where we need to exercise complicated lantern paths with real traces we use my patent-pending post-I/O approach of storing ~100 traces/devtoolslogs on GCP that can run on-demand/on commits that touch lantern files to assert that opportunities and metrics look decent

WDYT?

lighthouse-core/audits/uses-rel-preload.js
patrickhulce: Yeah fair, I was mostly retaining parity with the old way, but we can do max(TTI, load) like the others 👍

patrickhulce: having had the hour to think about it, I've reconvinced myself that this is a bad idea and we should keep it to max(...wastedMs), our selection of preloaded items is fairly narrow and in a busy page it will rarely impact the longest chain

given our super ambiguous no-longer-just-on-a-particular-metric approach, it seems OK for now

lighthouse-core/test/audits/uses-rel-preload-test.js
brendankenny: wait, but how are the others running with an undefined simulator? (obvs I'm too lazy to debug here :)

patrickhulce: the other tests are asserting that *no* savings are found, so a simulator isn't necessary

brendankenny: oh, lol

patrickhulce:
FYI @brendankenny I rebased against your changes and fixed the remaining type check items

lighthouse-core/test/audits/uses-rel-preload-test.js
brendankenny: it wouldn't give good coverage, but even a simple devtoolslog/trace with little or no savings would be a start until the GCP smoke tests. there's `lighthouse-core/tests/fixtures/artifacts/perflog/` and of course `lighthouse-core/test/results/artifacts/`

patrickhulce: sure

brendankenny: we should maybe make a tool to take real artifacts and strip out a bunch of stuff to make nice artifact fixtures and maybe only a few hundred KB at most. e.g. for things like this that are mostly networking we can get rid of a lot of trace events (enough that it was the same page load but with a lot less busy CPU)

(we have a lot of old traces for testing, but without paired devtools logs they're becoming increasingly useless for tests)

Issue #4425, opened 2018-02-05
Preload audit improvements


paulirish:
A few followups from #3450 

* [x] Restrict results to same-origin (as a rough filter for predictable and unchanging URLs)
* [ ] Expand which requests are critical enough to be included. (XHRs, etc)

We could do these in separate PRs.

wardpeet:
> Restrict results to same-origin (as a rough filter for predictable and unchanging URLs)
same-origin? as in the origin of url? what about cdn urls? could we use the third party badges? Or just take all origins that are in the document?

> Expand which requests are critical enough to be included. (XHRs, etc)
what definition do we use? Just keep requests 2 levels deep but use all requests? (xhr, images, ...)

patrickhulce:
> same-origin? as in the origin of url? what about cdn urls? could we use the third party badges? Or just take all origins that are in the document?

I vote we include all subdomains but exclude 3rd party. i.e. for cnn.com, we could flag us2.cdn.cnn.com URLs but not cnn-cdn.amazonaws.com.

It's possible they know the URL ahead of time for 3rd party but we'll flag those URLs in the preconnect audit anyhow.

paulirish:
> I vote we include all subdomains but exclude 3rd party

ya, sounds good. lets put this lil domain comparison fn in our url-shim.

addyosmani:
> I vote we include all subdomains but exclude 3rd party. i.e. for cnn.com, we could flag us2.cdn.cnn.com URLs but not cnn-cdn.amazonaws.com.

I pretty much agree. Let's roll with this.  

I'm moving this to P1 (can P2) so that the preload audit is as sharp as we can get it when we highlight it at events and folks try it out.

wardpeet:
> Expand which requests are critical enough to be included. (XHRs, etc)
@paulirish @patrickhulce are we still looking into this one?

wardpeet:
@paulirish @patrickhulce @addyosmani @brendankenny 
any feedback ? :)

patrickhulce:
Whoops, sorry Ward!

> Expand which requests are critical enough to be included 

Seems a bit open-ended and researchy to me, but valuable for lantern too. I'm not sure if it's the highest priority thing to be doing though. Maybe as a start we see if the simplest rule "XHRs completed before FMP" actually finds anything on real sites and if it looks promising we continue?

Not sure if @addyosmani or @paulirish has other plans for you though :)

paulirish:
[...]

PR #5065, opened 2018-04-29
core(preload): only allow same origin (domain + subdomains)

wardpeet:
Fixes nr 1 of #4425 
Restrict results to same-origin (as a rough filter for predictable and unchanging URLs). Same origin as in subdomains and rootdomains.

### DIFF
--- a/lighthouse-core/lib/url-shim.js
+++ b/lighthouse-core/lib/url-shim.js
@@ -97,12 +97,12 @@ class URLShim extends URL {
     const urlAInfo = new URL(urlA);
     const urlBInfo = new URL(urlB);
 
-    if (!urlAInfo.host || !urlBInfo.host) {
+    if (!urlAInfo.hostname || !urlBInfo.hostname) {
       return false;
     }
 
-    const urlARootDomain = urlAInfo.host.split('.').slice(-2).join('.');
-    const urlBRootDomain = urlBInfo.host.split('.').slice(-2).join('.');
+    const urlARootDomain = urlAInfo.hostname.split('.').slice(-2).join('.');
+    const urlBRootDomain = urlBInfo.hostname.split('.').slice(-2).join('.');
 
     return urlARootDomain === urlBRootDomain;
   }
--- a/lighthouse-core/lib/url-shim.js
+++ b/lighthouse-core/lib/url-shim.js
@@ -17,23 +17,12 @@ const Util = require('../report/html/renderer/util.js');
 const URL = /** @type {!Window["URL"]} */ (typeof self !== 'undefined' && self.URL) ||
     require('url').URL;
 
-const tldPlusOne = {
-  ar: ['com', 'edu', 'gob', 'int', 'mil', 'mar', 'net', 'org', 'tur', 'musica'],
-  at: ['co', 'or', 'priv', 'ac'],
-  fr: ['avocat', 'aeroport', 'veterinaire'],
-  nz: ['ac', 'co', 'school', 'cri', 'govt', 'mil', 'parliament'],
-  il: ['org', 'k12', 'gov', 'muni', 'idf'],
-  ru: ['com', 'edu', 'gob', 'int', 'mil', 'mar', 'net', 'org', 'tur', 'musica'],
-  za: ['ac', 'gov', 'law', 'mil', 'nom', 'school', 'net'],
-  kr: ['ac', 'co', 'es', 'go', 'hs', 'kg', 'mil', 'ms', 'ne', 'or', 'pe', 're', 'sc', 'busan',
-    'chungbuk', 'chungnam', 'daegu', 'daejeon', 'gangwon', 'gwangju', 'gyeongbuk', 'gyeonggi',
-    'gyeongnam', 'incheon', 'jeju', 'jeonbuk', 'jeonnam', 'seoul', 'ulsan'],
-  es: ['org', 'gob'],
-  tr: ['com', 'info', 'biz', 'net', 'org', 'web', 'gen', 'tv', 'av', 'dr', 'bbs', 'name', 'tel',
-    'gov', 'bel', 'pol', 'mil', 'k12', 'edu', 'kep', 'nc', 'gov.nc'],
-  ua: ['gov', 'com', 'in', 'org', 'net', 'edu'],
-  uk: ['co', 'org', 'me', 'ltd', 'plc', 'net', 'sch', 'ac', 'gov', 'mod', 'mil', 'nhs', 'police'],
-};
+const tldPlusOneRegex = new RegExp(`\\.(com|co|gov|org|ac|or|edu|go|nic|in|gouv|gob|web|spb|net|` +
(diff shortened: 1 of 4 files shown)


patrickhulce:
@wardpeet FYI I had to rework considerable portions of this for lantern (#5062), wdyt about holdin' tight for a cycle to get that in first? :)

wardpeet:
sure thing no worries

lighthouse-core/lib/url-shim.js
```diff
@@ -87,6 +87,26 @@ class URLShim extends URL {
     }
   }
 
+  /**
+   * Check if rootDomains matches
+   *
+   * @param {string} urlA
+   * @param {string} urlB
+   */
+  static rootDomainsMatch(urlA, urlB) {
```
paulirish: can you add some tests in url-shim-test?

lighthouse-core/lib/url-shim.js
```diff
@@ -87,6 +87,26 @@ class URLShim extends URL {
...
+    const urlAInfo = new URL(urlA);
+    const urlBInfo = new URL(urlB);
+
+    if (!urlAInfo.hostname || !urlBInfo.hostname) {
+      return false;
+    }
+
+    const urlARootDomain = urlAInfo.hostname.split('.').slice(-2).join('.');
```
paulirish: this is called "tld plus one". let's use that term for the method name and here.

let's add a few whitelisted exceptions like '.co.uk'. (country-code second level domains).

lighthouse-core/lib/url-shim.js
```diff
@@ -87,6 +87,26 @@ class URLShim extends URL {
...
+    const urlBRootDomain = urlBInfo.hostname.split('.').slice(-2).join('.');
```
patrickhulce: https://en.wikipedia.org/wiki/Second-level_domain is a good starting point

lighthouse-core/lib/url-shim.js
```diff
@@ -17,6 +17,24 @@ const Util = require('../report/html/renderer/util.js');
 const URL = /** @type {!Window["URL"]} */ (typeof self !== 'undefined' && self.URL) ||
     require('url').URL;
 
+const tldPlusOne = {
```
paulirish: let's hardcode just the keys of this list (and add .co.uk)

add comment saying that https://publicsuffix.org/ is the real source of this data. and hypothetically we could pull in a dep like https://npms.io/search?q=public+suffix+list to do this forreal, but at significant byte cost.

and admit we're taking the lazy way out.

patrickhulce: shared a sheet with you on the most popular 2-level TLDs, but summary: if we did a regex with /\.(co|com|org|edu|gov)\.TLD$/ that should cover most cases

wardpeet: I added a new list to the regex coming from http archive. Not sure how I should comment it in the code. Also i'm not matching all 1000 entries

lighthouse-core/lib/url-shim.js
```diff
@@ -17,6 +17,13 @@ const Util = require('../report/html/renderer/util.js');
 const URL = /** @type {!Window["URL"]} */ (typeof self !== 'undefined' && self.URL) ||
     require('url').URL;
 
+const tldPlusOneRegex = new RegExp(`\\.(com|co|gov|org|ac|or|edu|go|nic|in|gouv|gob|web|spb|net|` +
```
patrickhulce: I think we can trim down this list a bit based on the coverage percentages in that sheet, doing just the top 10 already gives us 94% coverage

```
com       2321232
co        1714813
gov       435538
edu       304490
ac        222746
org       202231
go        82772
gob       73721
or        72214
net       43587
----------------
sum       5473344
total     5841097
coverage  ~94%

File: lighthouse-core/lib/url-shim.js [hunk 6]
```diff
@@ -87,6 +94,55 @@ class URLShim extends URL {
...
+  /**
+   * Gets the tld of a domain
+   *
+   * @param {string} hostname
+   * @return {string} tld
+   */
+  static getTld(hostname) {
+    const tld = hostname.split('.').slice(-1)[0];
+
+    const tldRegex = new RegExp(`${tldPlusOneRegex.source}\\.${tld}$`);
```
patrickhulce: let's just do the `.split` and a `tlds.has` instead :)

wardpeet:
Updated not sure this is what you had in mind

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -50,14 +50,18 @@ describe('Performance: uses-rel-preload audit', () => {
 
   it('should suggest preload resource', () => {
     const rootNode = buildNode(1, 'http://example.com');
-    const mainDocumentNode = buildNode(2, 'http://www.example.com');
+    const mainDocumentNode = buildNode(2, 'http://www.example.com:3000');
```
patrickhulce: we're gonna need some graph creation helpers soon to make all this stuff :)

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -69,41 +73,63 @@ describe('Performance: uses-rel-preload audit', () => {
...
+          nodeTimings.set(scriptOtherNodeLocal, {startTime: 1000, endTime: 2500});
+        }
+
+        return {timeInMs: 3500, nodeTimings};
       },
     };
 
     const mainResource = Object.assign({}, defaultMainResource, {
-      url: 'http://www.example.com',
+      url: 'http://www.example.com:3000',
```
patrickhulce: is this just to exercise extra hostname checks?

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -131,7 +165,7 @@ describe('Performance: uses-rel-preload audit', () => {
     return UsesRelPreload.audit(mockArtifacts(networkRecords, chains, mainResource), {}).then(
       output => {
         assert.equal(output.rawValue, 1250);
-        assert.equal(output.details.items.length, 1);
+        assert.equal(output.details.items.length, 2);
```
patrickhulce: maybe overkill but assert the URL matches too?

lighthouse-core/lib/url-shim.js
```diff
@@ -17,6 +17,10 @@ const Util = require('../report/html/renderer/util.js');
 const URL = /** @type {!Window["URL"]} */ (typeof self !== 'undefined' && self.URL) ||
     require('url').URL;
 
+const listOfTlds = [
```
patrickhulce: Add a comment that somewhat documents what happened here re: coverage and whatnot? maybe just link to the best comment thread on this PR :)

lighthouse-core/test/audits/uses-rel-preload-test.js
wardpeet: yes it's to make sure that origins with a port number are handled correct by the rootDomainMatch

patrickhulce:
so I just realized in #5471, we don't want to hide URLs based on if *they* were 3rd party, we want to hide them if *their initiator* was a 3rd party 🤔 nearly all of this should still apply, but I'll prob make an update soon :)

PR #5068, opened 2018-04-30
core(metrics): rename ConsistentlyInteractive -> Interactive

patrickhulce:
renames `consistently-interactive` -> `interactive`

no other metric has `time to` in its implementation even though all of the `first*` metrics are also `time to` metrics, for consistency I'm proposing just `interactive`. Feel free to continue to use TTI to describe it, but seems silly to be the only one to do so when referring to its full name :)

ref #4629


lighthouse-core/audits/interactive.js
```diff
@@ -14,16 +14,15 @@ const Util = require('../report/html/renderer/util');
...
    * @return {!AuditMeta}
    */
   static get meta() {
     return {
-      name: 'consistently-interactive',
-      description: 'Consistently Interactive (beta)',
-      helpText: 'Consistently Interactive marks the time at which the page is ' +
-          'fully interactive. ' +
+      name: 'interactive',
+      description: 'Interactive',
```
paulirish: IMO this should be called the full "Time to Interactive" in the UI. even if there's inconsistency, that's the name.

lighthouse-core/config/default-config.js
```diff
@@ -84,7 +84,7 @@ module.exports = {
     'errors-in-console',
     'time-to-first-byte',
     'first-cpu-idle',
-    'consistently-interactive',
+    'interactive',
```
paulirish: i do prefer 'time-to-interactive' personally, but i could could live with this slug if i had to.

lighthouse-core/audits/interactive.js
patrickhulce: done!

PR #5069, opened 2018-04-30
core(tsc): add type checking to dbw audits

brendankenny:
the only major cleanups are in the `no-vulnerable-libraries.js.js` (lots of mutations) and `event-helpers.js` (simplified so that the type checker could follow along).

The golden lhr changes are eliminating the duplicate vulnerability listings (because the original library objects are no longer mutated) and taking out the `label` entry on the event listeners since it appears to only be set, not read, and it's just a duplicate of the line/col info anyways.

### DIFF
--- a/lighthouse-core/audits/dobetterweb/no-vulnerable-libraries.js
+++ b/lighthouse-core/audits/dobetterweb/no-vulnerable-libraries.js
@@ -15,13 +15,17 @@
 const Audit = require('../audit');
 const Sentry = require('../../lib/sentry');
 const semver = require('semver');
+// @ts-ignore - json require
 const snykDatabase = require('../../../third-party/snyk/snapshot.json');
 
 const SEMVER_REGEX = /^(\d+\.\d+\.\d+)[^-0-9]+/;
 
+/** @typedef {{npm: Object<string, Array<{id: string, severity: string, semver: {vulnerable: Array<string>}}>>}} SnykDB */
+/** @typedef {{severity: string, numericSeverity: number, library: string, url: string}} Vulnerability */
+
 class NoVulnerableLibrariesAudit extends Audit {
   /**
-   * @return {!AuditMeta}
+   * @return {LH.Audit.Meta}
    */
   static get meta() {
     return {
@@ -37,14 +41,14 @@ class NoVulnerableLibrariesAudit extends Audit {
   }
 
   /**
-   * @return {{npm: !Object<string, !Array<{id: string, severity: string, semver: {vulnerable: !Array<string>}}>>}}
+   * @return {SnykDB}
    */
   static get snykDB() {
     return snykDatabase;
   }
 
   /**
-   * @return {!Object<string, number>}
+   * @return {Object<string, number>}
    */
   static get severityMap() {
     return {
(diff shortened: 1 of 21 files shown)


lighthouse-core/audits/dobetterweb/no-vulnerable-libraries.js
```diff
@@ -66,86 +70,103 @@ class NoVulnerableLibrariesAudit extends Audit {
...
-      if (lib.vulns.length > 0) {
-        lib.vulnCount = lib.vulns.length;
-        lib.highestSeverity = this.highestSeverity(lib.vulns).replace(/^\w/, l => l.toUpperCase());
-        totalVulns += lib.vulnCount;
-        lib.detectedLib = {};
-        lib.detectedLib.text = lib.name + '@' + lib.version;
-        lib.detectedLib.url = lib.pkgLink;
-        lib.detectedLib.type = 'link';
+    /** @type {Array<{highestSeverity: string, vulnCount: number, detectedLib: LH.Audit.DetailsRendererLinkDetailsJSON}>} */
+    const vulnerableResults = [];
```
patrickhulce: vulnerabilities? vulnerabilityResults?

lighthouse-core/audits/dobetterweb/password-inputs-can-be-pasted-into.js
```diff
@@ -23,15 +23,19 @@ class PasswordInputsCanBePastedIntoAudit extends Audit {
...
    */
   static audit(artifacts) {
     const passwordInputsWithPreventedPaste = artifacts.PasswordInputsWithPreventedPaste;
 
-    const items = passwordInputsWithPreventedPaste.map(input => ({
-      node: {type: 'node', snippet: input.snippet},
-    }));
+    /** @type {Array<{node: LH.Audit.DetailsRendererNodeDetailsJSON}>} */
+    const items = [];
+    passwordInputsWithPreventedPaste.forEach(input => {
```
patrickhulce: 😢 TS really can't deal with a map + assertion?

typings/audit.d.ts
```diff
@@ -7,7 +7,7 @@
 declare global {
   module LH.Audit {
     export interface Context {
-      options: Object; // audit options
+      options: Record<string, any>; // audit options
```
patrickhulce: what's the difference between `Record<K, V>` and `Object<K, V>`

typings/audit.d.ts
```diff
@@ -47,7 +47,7 @@ declare global {
       debugString?: string;
     }
 
-    // TODO: placeholder typedefs until Details are typed
+    // TODO: placeholders typedefs until Details are typed
```
patrickhulce: revert? :)

typings/audit.d.ts
```diff
@@ -57,27 +57,36 @@ declare global {
     export interface DetailsRendererDetailsJSON {
       type: 'table';
       headings: Array<Audit.Heading>;
-      items: Array<{[x: string]: string|number|DetailsRendererNodeDetailsJSON}>;
+      items: Array<{[x: string]: DetailsItem}>;
       summary?: DetailsRendererDetailsSummary;
     }
 
+    export type DetailsItem = string | number | DetailsRendererNodeDetailsJSON |
```
patrickhulce: DetailsItemMemberChildInfo

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3941,26 +3904,8 @@
         ],
         "items": [
           {
-            "name": "jQuery",
```
paulirish: lgtm on nuking this data. it's pretty low-value anyway

lighthouse-core/audits/dobetterweb/no-vulnerable-libraries.js
brendankenny: went with vulnerabilityResults :)

typings/audit.d.ts
brendankenny: > DetailsItemMemberChildInfo

...Summary

typings/audit.d.ts
brendankenny: done

lighthouse-core/audits/dobetterweb/password-inputs-can-be-pasted-into.js
brendankenny: it loses the type `'node'` (it widens to type `string`) :( The heuristics are wonky on widening like that, but people do want very different things at times (even we would probably usually want `string` in this case if it weren't for wanting to tell the difference between details types).

It's basically this or put a cast in there in addition to the declared type above, I believe, so this seemed better.

typings/audit.d.ts
brendankenny: > what's the difference between Record<K, V> and Object<K, V>

`Object<K, V>` only works in jsdoc. It's not parameterized in typescript code

brendankenny: ...we should probably come up with some sort of style rule for this vs `{[p: string]: any}` vs `Object<>` on the jsdoc side

PR #5070, opened 2018-04-30
core(preconnect): use lantern to compute savings

patrickhulce:
~~blocked on #5071~~

Relatively simple change, approximates the savings using just the TCP handshake time from lantern. The simulator itself doesn't have a notion of preconnect, so we won't estimate savings on a particular metric and bringing in the whole graph isn't necessary.

### DIFF
--- a/lighthouse-core/audits/uses-rel-preconnect.js
+++ b/lighthouse-core/audits/uses-rel-preconnect.js
@@ -15,6 +15,8 @@ const UnusedBytes = require('./byte-efficiency/byte-efficiency-audit');
 // @see https://github.com/GoogleChrome/lighthouse/issues/3106#issuecomment-333653747
 const PRECONNECT_SOCKET_MAX_IDLE = 15;
 
+const IGNORE_THRESHOLD_IN_MS = 50;
+
 class UsesRelPreconnectAudit extends Audit {
   /**
    * @return {LH.Audit.Meta}
@@ -134,6 +136,8 @@ class UsesRelPreconnectAudit extends Audit {
         firstRecordOfOrigin._timing.dnsStart;
 
       const wastedMs = Math.min(connectionTime, timeBetweenMainResourceAndDnsStart);
+      if (wastedMs < IGNORE_THRESHOLD_IN_MS) return;
+
       maxWasted = Math.max(wastedMs, maxWasted);
       results.push({
         url: securityOrigin,
--- a/lighthouse-core/audits/uses-rel-preconnect.js
+++ b/lighthouse-core/audits/uses-rel-preconnect.js
@@ -65,17 +65,22 @@ class UsesRelPreconnectAudit extends Audit {
 
   /**
    * @param {LH.Artifacts} artifacts
+   * @param {LH.Audit.Context} context
    * @return {Promise<LH.Audit.Product>}
    */
-  static async audit(artifacts) {
+  static async audit(artifacts, context) {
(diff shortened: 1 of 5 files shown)


lighthouse-core/test/results/sample_v2.json
```diff
@@ -2894,27 +2894,27 @@
           {
             "url": "http://localhost:10200/dobetterweb/dbw_tester.css?delay=100",
             "totalBytes": 821,
-            "wastedMs": 873
+            "wastedMs": 330
```
patrickhulce: No clue why these changed I'm looking into it...

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -113,17 +118,24 @@ class UsesRelPreconnectAudit extends Audit {
...
 
-      const connectionTime =
-        firstRecordOfOrigin._timing.connectEnd - firstRecordOfOrigin._timing.dnsStart;
+      const securityOrigin = firstRecordOfOrigin.parsedURL.securityOrigin();
+
+      // Approximate the connection time with the duration of TCP (+potentially SSL) handshake
+      // DNS time can be large but can also be 0 if a commonly used origin that's cached, so make
+      // no assumption about DNS.
+      const additionalRtt = additionalRttByOrigin.get(securityOrigin) || 0;
+      let connectionTime = rtt + additionalRtt;
```
brendankenny: is `rtt` not `number | undefined`? (based on `LH.Gatherer.Simulation.Options`, at least)

typings/web-inspector.d.ts
```diff
@@ -17,7 +17,6 @@ declare global {
       url: string;
       _url: string;
       protocol: string;
-      origin: string | null;
```
brendankenny: I know this was from the last PR, but should we put a warning on `origin` to not use it? like
```js
/** use `parsedURL.securityOrigin()` instead */
origin: never;
```
or something. It could probably sneak back in in the future otherwise (just a comment might also work :)

lighthouse-core/audits/uses-rel-preconnect.js
brendankenny: oh, the file isn't type checked yet :) It would be good to handle it now, though

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -113,17 +118,24 @@ class UsesRelPreconnectAudit extends Audit {
...
+      if (firstRecordOfOrigin.parsedURL.scheme === 'https') connectionTime = connectionTime * 2;
```
brendankenny: comment on where the *2 comes from (is that the TLS mentioned above? Why *2?)

lighthouse-core/audits/uses-rel-preconnect.js
patrickhulce: fixed on the simulator side 👍

lighthouse-core/audits/uses-rel-preconnect.js
patrickhulce: done

patrickhulce: done!

typings/web-inspector.d.ts
patrickhulce: yeah good call done

PR #5071, opened 2018-04-30
core(lantern): use securityOrigin on record

patrickhulce:
found a good bug while working on #5070 that's relatively independent

basically if you try to call network analysis before you create a network node from the network record, all the origins will be undefined and so the `additionalRttByOrigin` and `serverResponseTimeByOrigin` are totally useless for the future cached runs that *do* properly create a network node from the network record. We originally did this because calling `new URL()` to get the origin was super expensive and these are somewhat hot paths, while the function call + getter is a bit more expensive, it's still not nearly as bad as `new URL`. Seems OK until it crops up as a bottleneck

this PR reverts the `.origin` dance we do on network records and just uses `.securityOrigin()` on the record, if it becomes a big bottleneck again we can further optimize

### DIFF
--- a/lighthouse-core/lib/dependency-graph/network-node.js
+++ b/lighthouse-core/lib/dependency-graph/network-node.js
@@ -42,13 +42,6 @@ class NetworkNode extends Node {
    * @return {LH.WebInspector.NetworkRequest}
    */
   get record() {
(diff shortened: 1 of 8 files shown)


brendankenny:
Can you add a description of the bug for future git blaming? :)

PR #5072, opened 2018-04-30
core(tsc): add type checking to most byte efficiency audits

brendankenny:
`uses-optimized-images.js` and `uses-webp-images.js` only get the surface treatment for now as I noticed I was lazy typing the `OptimizedImages` in the error case. Fixing that will make their changes really easy.

The only annoying change here is having to explicitly cast `Node` to `NetworkNode` when wanting to access its network record. The cast is somewhat reasonable, but since they already have a readonly `type` property to differentiate, we could make them a discriminated union in the future. That way the compiler knows a node could only be a CPU or network node (+ whatever the future brings), and checking `node.type === 'network'` would establish to it that the following code was dealing with a `NetworkNode`.

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
+++ b/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
@@ -8,6 +8,9 @@
 const Audit = require('../audit');
 const Interactive = require('../../gather/computed/metrics/lantern-interactive'); // eslint-disable-line max-len
 const Simulator = require('../../lib/dependency-graph/simulator/simulator'); // eslint-disable-line no-unused-vars
+const Node = require('../../lib/dependency-graph/node.js'); // eslint-disable-line no-unused-vars
+
+const NetworkNode = require('../../lib/dependency-graph/network-node.js'); // eslint-disable-line no-unused-vars
 
 const KB_IN_BYTES = 1024;
 
@@ -33,7 +36,7 @@ class UnusedBytes extends Audit {
   /**
    * @param {number} bytes
    * @param {number} networkThroughput measured in bytes/second
-   * @return {string}
+   * @return {number}
    */
   static bytesToMs(bytes, networkThroughput) {
     const milliseconds = bytes / networkThroughput * 1000;
@@ -42,9 +45,9 @@ class UnusedBytes extends Audit {
 
   /**
    * Estimates the number of bytes this network record would have consumed on the network based on the
-   * uncompressed size (totalBytes), uses the actual transfer size from the network record if applicable.
+   * uncompressed size (totalBytes). Uses the actual transfer size from the network record if applicable.
    *
-   * @param {?LH.WebInspector.NetworkRequest} networkRecord
+   * @param {LH.WebInspector.NetworkRequest=} networkRecord
    * @param {number} totalBytes Uncompressed size of the resource
    * @param {string=} resourceType
    * @param {number=} compressionRatio
@@ -58,19 +61,21 @@ class UnusedBytes extends Audit {
       return Math.round(totalBytes * compressionRatio);
     } else if (networkRecord._resourceType && networkRecord._resourceType._name === resourceType) {
       // This was a regular standalone asset, just use the transfer size.
-      return networkRecord._transferSize;
+      return networkRecord._transferSize || 0;
     } else {
       // This was an asset that was inlined in a different resource type (e.g. HTML document).
       // Use the compression ratio of the resource to estimate the total transferred bytes.
-      const compressionRatio = networkRecord._transferSize / networkRecord._resourceSize || 1;
+      const transferSize = networkRecord._transferSize || 0;
+      const resourceSize = networkRecord._resourceSize;
+      const compressionRatio = resourceSize !== undefined ? (transferSize / resourceSize) : 1;
       return Math.round(totalBytes * compressionRatio);
     }
   }
 
   /**
-   * @param {Artifacts} artifacts
-   * @param {LH.Audit.Context=} context
-   * @return {Promise<AuditResult>}
+   * @param {LH.Artifacts} artifacts
+   * @param {LH.Audit.Context} context
+   * @return {Promise<LH.Audit.Product>}
    */
   static audit(artifacts, context) {
     const trace = artifacts.traces[Audit.DEFAULT_PASS];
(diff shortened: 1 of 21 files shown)


lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -128,15 +130,12 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
...
 
   /**
-   * @param {!Object} stylesheetInfo The stylesheetInfo object.
+   * @param {StyleSheetInfo} stylesheetInfo The stylesheetInfo object.
    * @param {string} pageUrl The URL of the page, used to identify inline styles.
-   * @return {?{url: string, wastedBytes: number, totalBytes: number}}
+   * @return {LH.Audit.ByteEfficiencyResult}
    */
   static mapSheetToResult(stylesheetInfo, pageUrl) {
-    if (stylesheetInfo.isDuplicate) {
```
brendankenny: it appears this is a holdover from old code. It doesn't appear to be set anywhere

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -161,6 +175,8 @@ class UnusedBytes extends Audit {
       wastedMs,
       wastedBytes,
     };
+
+    // @ts-ignore - TODO(bckenny): unify details types. items shouldn't be an indexed type.
```
brendankenny: we need a unified theory of details item types rather than some with `{[x: string]: DetailsItem}` and some not (if it doesn't have a generic index you can't assign to a type that does). I'd rather punt here than figure that out, though, since we're just going to have to figure it out all over again when we switch to our new details types

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -116,6 +120,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
     // TODO(phulce): move this to always use lantern
     const settings = context.settings;
     return artifacts.requestFirstCPUIdle({trace, devtoolsLog, settings}).then(firstInteractive => {
+      // @ts-ignore - see above TODO.
```
brendankenny: right now this is actually an `LH.Artifacts.Metric`, so `firstInteractive.timestamp` does exist. When it switches to lantern (see above), it's going to have to switch off of `timestamp` anyways, so didn't seem worth fixing

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -70,6 +81,7 @@ class RenderBlockingResources extends Audit {
 
     const metricSettings = {throttlingMethod: 'simulate'};
     const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
+    // @ts-ignore - TODO(bckenny): allow optional `throttling` settings
```
brendankenny: @patrickhulce maybe we want a new type for settings required for metrics rather than the full config settings

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -91,6 +103,7 @@ class RenderBlockingResources extends Audit {
       node.traverse(node => deferredNodeIds.add(node.id));
 
       // "wastedMs" is the download time of the network request, responseReceived - requestSent
+      // @ts-ignore - TODO(phulce): nodeTiming.startTime/endTime shouldn't be optional by this point?
```
brendankenny: `nodeTimings` is going to need something to differentiate between the in progress object with properties still being defined (and therefore optional) vs what's exposed externally where presumably most (all?) properties are defined. I added @patrickhulce here but I can also take it eventually :)

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -135,38 +148,43 @@ class RenderBlockingResources extends Audit {
...
    * @param {LH.Audit.Context} context
-   * @return {Map<string, number>}
+   * @return {Promise<Map<string, number>>}
    */
   static async computeWastedCSSBytes(artifacts, context) {
     const wastedBytesByUrl = new Map();
     try {
       // TODO(phulce): pull this out into computed artifact
       const results = await UnusedCSS.audit(artifacts, context);
+      // @ts-ignore - TODO(bckenny): details types.
```
brendankenny: we *could* have a more specific type returned from `UnusedCSS.audit` to avoid the `any` type, but it unfortunately passes through `audit` in the super class, making the return type generic as well. We could figure this out with details types or the above TODO about moving to a computed artifact would fix it as well.

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: yeah we'll probably need to do an `IntermediateNodeTiming` vs. `NodeTiming` split, I can do that in followup 👍

lighthouse-core/audits/byte-efficiency/total-byte-weight.js
```diff
@@ -50,11 +50,12 @@ class TotalByteWeight extends ByteEfficiencyAudit {
       artifacts.requestNetworkThroughput(devtoolsLogs),
     ]).then(([networkRecords, networkThroughput]) => {
       let totalBytes = 0;
+      /** @type {Array<{url: string, totalBytes: number, totalMs: number}>} */
       let results = [];
       networkRecords.forEach(record => {
         // exclude data URIs since their size is reflected in other resources
         // exclude unfinished requests since they won't have transfer size information
-        if (record.scheme === 'data' || !record.finished) return;
+        if (record.parsedURL.scheme === 'data' || !record.finished) return;
```
patrickhulce: 😆

typings/audit.d.ts
```diff
@@ -38,15 +38,24 @@ declare global {
       key: string;
       itemType: string;
       text: string;
+      displayUnit?: string;
+      granularity?: number;
     }
 
-    export interface HeadingsResult {
```
patrickhulce: so looks like this was just nonsense before? `results: number` :)

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: yeah let's just do the right computed artifact fix later 👍

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: heh, didn't see this but yeah bingo

PR #5073, opened 2018-04-30
report: tweak tooltips

paulirish:
@hwikyounglee asked for two changes. 1) tooltip arrow points to the number. 2) hover background color.

Also made a few changes:

* tooltip doesnt start showing up for 900ms now (instead of 500ms) 
* tooltip trigger boundary is smaller. It matches the gray background (and doesn't extend to the line)
* left/right padding on these boxes so their container extends a little farther than the text. I'll be applying this padding to other elements in the report, as the mocks suggest. see "padding illustration" below.

before and after gif:
kapture 2018-04-30 at 15 46 02

<details>
<summary>
padding illustration:
</summary>

</details>

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -22,6 +22,7 @@
   --section-indent: 16px;
   --audit-group-indent: 16px;
   --audit-indent: 16px;
+  --text-indent: 8px;
   --expandable-indent: 20px;
   --secondary-text-color: #565656;
   /*--accent-color: #3879d9;*/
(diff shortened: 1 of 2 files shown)


paulirish:
@hwikyounglee we should probably just sit down for an hour and go through things with a fine-toothed comb, but im not yet ready for that. probably later this week.

vinamratasingal:
Should we setup some time for Thursday/Friday? I'd love to join as well just to see the state of where things are at.

patrickhulce:
❤️ the enthusiasm @vinamratasingal 

in general, we like holdin' off on mergin' til the tests go through at least once and we can 🎃


🎃  = squash, as in squash 'n merge vs. create merge commit :)
image

vinamratasingal:
My b, got a little too eager there...  thanks for the tip Paul! 

LMK if you want me to schedule some time to do a run through of the UI.

paulirish:
Turns out that was Patrick...  but hey we could pull a Chrome DevRel and turn everyone into a Paul.   @paulhulce and @paulkenny have a good ring to them. 🤣

vinamratasingal:
Whooppssss i blame this on the jetlag my b, thanks for catching that hehe. 

Updated my original comment :P

PR #5077, opened 2018-05-01
report: error'd audits get 'Error!' treatment

paulirish:
### changes

* what the title said
* the report's ❌  and ✅ use standard fail/pass colors
* populateScore no longer sets title/description. Just handles the score gauge.
* removed right column score/icon for informative/manual audits (no more blue '100's)

### screenshots

image

image

### DIFF
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -55,15 +55,15 @@ class CategoryRenderer {
       auditEl.classList.add('lh-audit--manual');
     }
 
-    this._populateScore(auditEl, audit.result.score, scoreDisplayMode);
+    this._populateScore(auditEl, audit.result.score, scoreDisplayMode, audit.result.error);
 
     if (audit.result.error) {
       auditEl.classList.add(`lh-audit--error`);
       const valueEl = this.dom.find('.lh-score__value', auditEl);
       valueEl.textContent = 'Error';
       valueEl.classList.add('tooltip-boundary');
-      const content = this.dom.createChildOf(valueEl, 'div', 'lh-error-tooltip-content tooltip');
-      content.textContent = audit.result.debugString || 'Report error: no audit information';
+      const tooltip = this.dom.createChildOf(valueEl, 'div', 'lh-error-tooltip-content tooltip');
+      tooltip.textContent = audit.result.debugString || 'Report error: no audit information';
     } else if (audit.result.debugString) {
       const debugStrEl = auditEl.appendChild(this.dom.createElement('div', 'lh-debug'));
       debugStrEl.textContent = audit.result.debugString;
@@ -75,14 +75,16 @@ class CategoryRenderer {
    * @param {!DocumentFragment|!Element} element DOM node to populate with values.
    * @param {number} score
    * @param {string} scoreDisplayMode
+   * @param {?boolean} isError
    * @return {!Element}
    */
-  _populateScore(element, score, scoreDisplayMode) {
+  _populateScore(element, score, scoreDisplayMode, isError) {
     const scoreOutOf100 = Math.round(score * 100);
     const valueEl = this.dom.find('.lh-score__value', element);
     valueEl.textContent = Util.formatNumber(scoreOutOf100);
-    valueEl.classList.add(`lh-score__value--${Util.calculateRating(score)}`,
-        `lh-score__value--${scoreDisplayMode}`);
+    // FIXME(paulirish): this'll have to deal with null scores and scoreDisplayMode stuff..
+    const rating = isError ? 'error' : Util.calculateRating(score);
+    valueEl.classList.add(`lh-score__value--${rating}`, `lh-score__value--${scoreDisplayMode}`);
 
     return /** @type {!Element} **/ (element);
   }
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -60,10 +60,10 @@ class CategoryRenderer {
     if (audit.result.error) {
       auditEl.classList.add(`lh-audit--error`);
       const valueEl = this.dom.find('.lh-score__value', auditEl);
-      valueEl.textContent = 'Error!';
+      valueEl.textContent = 'Error';
       valueEl.classList.add('tooltip-boundary');
       const content = this.dom.createChildOf(valueEl, 'div', 'lh-error-tooltip-content tooltip');
-      content.textContent = audit.result.debugString || 'Report error: no metric information';
+      content.textContent = audit.result.debugString || 'Report error: no audit information';
     } else if (audit.result.debugString) {
       const debugStrEl = auditEl.appendChild(this.dom.createElement('div', 'lh-debug'));
       debugStrEl.textContent = audit.result.debugString;
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -42,11 +42,6 @@ class CategoryRenderer {
     this.dom.find('.lh-audit__description', auditEl)
       .appendChild(this.dom.convertMarkdownLinkSnippets(audit.result.helpText));
 
-    if (audit.result.debugString) {
-      const debugStrEl = auditEl.appendChild(this.dom.createElement('div', 'lh-debug'));
-      debugStrEl.textContent = audit.result.debugString;
-    }
-
     // Append audit details to header section so the entire audit is within a <details>.
     const header = /** @type {!HTMLDetailsElement} */ (this.dom.find('.lh-audit__header', auditEl));
     if (audit.result.details && audit.result.details.type) {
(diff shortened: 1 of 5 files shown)


vinamratasingal:
Hmmm question: why do warnings appear in red? Or is that out of scope for this PR? 

Wondering if we should offer similar tooltip visual treatment for warnings + errors?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -16,21 +16,24 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
     const descriptionEl = this.dom.find('.lh-perf-metric__description', tmpl);
     descriptionEl.appendChild(this.dom.convertMarkdownLinkSnippets(audit.result.helpText));
 
-    if (typeof audit.result.rawValue !== 'number') {
-      const debugStrEl = this.dom.createChildOf(element, 'div', 'lh-debug');
-      debugStrEl.textContent = audit.result.debugString || 'Report error: no metric information';
-      return element;
+    if (audit.result.error) {
+      element.classList.add(`lh-perf-metric--error`);
+      descriptionEl.innerHTML = '';
```
patrickhulce: is `.textContent = ''` not sufficient?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -58,30 +55,35 @@ class CategoryRenderer {
...
-    return this._populateScore(auditEl, audit.result.score, scoreDisplayMode, title, description);
+    this._populateScore(auditEl, audit.result.score, scoreDisplayMode);
+
+    if (audit.result.error) {
+      auditEl.classList.add(`lh-audit--error`);
+      const valueEl = this.dom.find('.lh-score__value', auditEl);
+      valueEl.textContent = 'Error!';
+      valueEl.classList.add('tooltip-boundary');
+      const content = this.dom.createChildOf(valueEl, 'div', 'lh-error-tooltip-content tooltip');
+      content.textContent = audit.result.debugString || 'Report error: no metric information';
```
patrickhulce: isn't this one just for generic audits?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -58,30 +55,35 @@ class CategoryRenderer {
       auditEl.classList.add('lh-audit--manual');
     }
 
-    return this._populateScore(auditEl, audit.result.score, scoreDisplayMode, title, description);
+    this._populateScore(auditEl, audit.result.score, scoreDisplayMode);
+
+    if (audit.result.error) {
+      auditEl.classList.add(`lh-audit--error`);
+      const valueEl = this.dom.find('.lh-score__value', auditEl);
+      valueEl.textContent = 'Error!';
```
patrickhulce: not sure I love the `!` on the end, save that sucker for the category score ;)

lighthouse-core/report/html/renderer/util.js
```diff
@@ -38,6 +39,9 @@ class Util {
     } else if (score >= RATINGS.AVERAGE.minScore) {
       rating = RATINGS.AVERAGE.label;
     }
+    if (score === null) {
```
patrickhulce: this is a noop until the `FIXME`? if so save for that PR?

lighthouse-core/report/html/report-styles.css
```diff
@@ -953,13 +949,14 @@ summary.lh-passed-audits-summary {
   animation: fadeInTooltip 150ms;
   animation-fill-mode: forwards;
   animation-delay: 900ms;
-  min-width: 15em;
+  min-width: 23em;
```
patrickhulce: just felt right?

paulirish:
> Hmmm question: why do warnings appear in red? 

Good question.

We do have some warning-ish text right now that we show. We'd prefer to not hide this stuff behind a tooltip.

Here's some of the cases:
image
image
image
image


I'm open to changing their color. 

I tried the 'average-color' orange, but the color contrast is too low:
image

> Or is that out of scope for this PR?

yah basically. i think we need a design decision. The followup eng work doesn't really retrace steps so we're good.

lighthouse-core/report/html/renderer/category-renderer.js
paulirish: I thought felt a little less naked with the bang.

but i'm with the errors being bang-less.

image
image

lighthouse-core/report/html/report-styles.css
paulirish: Yup. In the new audit-error case the tooltip boundary is super narrow so the tooltip was being displayed at its min-width. 

Behold!

image

vinamratasingal:
Got it. Not needed for this PR, but I do think that changing the warning color would be good to avoid confusion between warning/error. @hwikyounglee can you help suggest a different color? 

Otherwise, LGTM from my side.

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -16,21 +16,24 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     const tmpl = this.dom.cloneTemplate('#tmpl-lh-perf-metric', this.templateContext);
     const element = this.dom.find('.lh-perf-metric', tmpl);
     element.id = audit.result.name;
+    // FIXME(paulirish): currently this sets a 'lh-perf-metric--undefined' class on error'd audits
     element.classList.add(`lh-perf-metric--${Util.calculateRating(audit.result.score)}`);
```
brendankenny: just add a conditional on `const rating = Util.calculateRating(audit.result.score); if (rating !== undefined) {element.classList.add(...);}`? Since `lh-perf-metric--error` is added below that seems like it's fine (unless you have more specific future plans)

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -16,21 +16,24 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
     descriptionEl.appendChild(this.dom.convertMarkdownLinkSnippets(audit.result.helpText));
 
-    if (typeof audit.result.rawValue !== 'number') {
-      const debugStrEl = this.dom.createChildOf(element, 'div', 'lh-debug');
-      debugStrEl.textContent = audit.result.debugString || 'Report error: no metric information';
-      return element;
+    if (audit.result.error) {
+      element.classList.add(`lh-perf-metric--error`);
+      descriptionEl.textContent = '';
+      valueEl.textContent = 'Error';
```
brendankenny: ha, well I actually liked the bangful message. Agree that `Error` just sitting on its own looks incomplete or like the string accidentally leaked into the html there rather than intentionally placed :)

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -58,30 +55,35 @@ class CategoryRenderer {
...
 
-    return this._populateScore(auditEl, audit.result.score, scoreDisplayMode, title, description);
+    this._populateScore(auditEl, audit.result.score, scoreDisplayMode);
+
+    if (audit.result.error) {
+      auditEl.classList.add(`lh-audit--error`);
+      const valueEl = this.dom.find('.lh-score__value', auditEl);
+      valueEl.textContent = 'Error';
+      valueEl.classList.add('tooltip-boundary');
+      const content = this.dom.createChildOf(valueEl, 'div', 'lh-error-tooltip-content tooltip');
```
brendankenny: bikeshed: if it's a tooltip, name the variable `tooltip`?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -16,21 +16,24 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
+      const content = this.dom.createChildOf(descriptionEl, 'span', 'lh-error-tooltip-content');
```
brendankenny: tooltip?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -58,30 +55,35 @@ class CategoryRenderer {
...
-  _populateScore(element, score, scoreDisplayMode, title, description) {
-    // Fill in the blanks.
+  _populateScore(element, score, scoreDisplayMode) {
     const scoreOutOf100 = Math.round(score * 100);
     const valueEl = this.dom.find('.lh-score__value', element);
     valueEl.textContent = Util.formatNumber(scoreOutOf100);
     valueEl.classList.add(`lh-score__value--${Util.calculateRating(score)}`,
         `lh-score__value--${scoreDisplayMode}`);
 
-    this.dom.find('.lh-audit__title, .lh-category-header__title', element).appendChild(
```
brendankenny: 👏 👏

lighthouse-core/report/html/renderer/performance-category-renderer.js
patrickhulce: if it were small caps'd and had a hover state ala material design would you feel differently? :)

brendankenny: ha, I think it's making it feel deliberate rather than any particular way of doing that :) Right now it kind of looks like we had a error string of `'Error\nsomething something'` and the rest of the string was cut off by the template or overflow hidden or something

brendankenny: `¡Error!`

patrickhulce: > ¡Error!

this 💯

lighthouse-core/report/html/renderer/performance-category-renderer.js
paulirish: Right now scores are 0 in the case of error's. :( 
I split that into a second PR because it opens a big can of worms.

paulirish:
ready to land when green.

PR #5081, opened 2018-05-01
core(redirects): use lantern to compute savings

patrickhulce:
should type-check pending #5072 🌮 

mostly whitespace changes, does the exact same thing as before but using the lantern graphs `startTime` instead of observed

### DIFF
--- a/lighthouse-core/audits/redirects.js
+++ b/lighthouse-core/audits/redirects.js
@@ -11,7 +11,7 @@ const UnusedBytes = require('./byte-efficiency/byte-efficiency-audit');
 
 class Redirects extends Audit {
   /**
-   * @return {!AuditMeta}
+   * @return {LH.Audit.Meta}
    */
   static get meta() {
     return {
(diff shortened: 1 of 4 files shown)


lighthouse-core/audits/redirects.js
```diff
@@ -20,68 +20,92 @@ class Redirects extends Audit {
...
+    const devtoolsLog = artifacts.devtoolsLogs[Audit.DEFAULT_PASS];
+    const simulatorData = {devtoolsLog, settings: context.settings};
+    const simulator = await artifacts.requestLoadSimulator(simulatorData);
+
+    // TODO(phulce): split MetricComputationInput type and avoid this cast
+    const metricSettings = /** @type {LH.Config.Settings} */ ({throttlingMethod: 'simulate'});
+    const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
+
+    const mainResource = await artifacts.requestMainResource({URL: artifacts.URL, devtoolsLog});
+    const metricResult = /** @type {LH.Artifacts.LanternMetric} */ (
```
brendankenny: any reason to not just ask for `artifacts.requestLanternInteractive()` directly?

lighthouse-core/audits/redirects.js
```diff
@@ -20,68 +20,92 @@ class Redirects extends Audit {
...
+      });
+    }
+
+    for (let i = 1; i < redirectRequests.length; i++) {
+      const initialRequest = redirectRequests[i - 1];
+      const redirectedRequest = redirectRequests[i];
+
+      const initialTiming = nodeTimingsByUrl.get(initialRequest.url);
+      const redirectedTiming = nodeTimingsByUrl.get(redirectedRequest.url);
+      if (!initialTiming || !redirectedTiming) continue;
```
brendankenny: wouldn't this be a error with the graph if any of the redirects aren't in there? Maybe we should throw or log an error in that (unlikely?) case?

lighthouse-core/audits/redirects.js
patrickhulce: yeah I suppose at this point it's not that different 👍 done

lighthouse-core/audits/redirects.js
patrickhulce: fair, I'll switch to error

PR #5088, opened 2018-05-02
docs(readme): remove ts command for lighthouse-cli dev

underbyte:
Looks like `lighthouse-cli` was converted over to commonjs in #3747 and the `yarn dev` was removed.


paulirish:
@underbyte good catch thank you!

PR #5089, opened 2018-05-02
core(tsc): more type checking of top-level audits

brendankenny:
multi-check audits, screenshots and viewport


brendankenny:
added unit test for new failure case in time-to-first-byte where *somehow* we end up with a finalUrl without a network record for it, which should be impossible considering finalUrl comes from the network records :)

Maybe some -GA mismatch situation

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -30,8 +32,8 @@ class ScreenshotThumbnails extends Audit {
    * Scales down an image to THUMBNAIL_WIDTH using nearest neighbor for speed, maintains aspect
    * ratio of the original thumbnail.
    *
-   * @param {{width: number, height: number, data: !Array<number>}} imageData
-   * @return {{width: number, height: number, data: !Array<number>}}
```
paulirish: do we want to reference the speedlineoutput.frames type instead?

brendankenny: SG, done

PR #5083, opened 2018-05-01
core(scoring): tweak performance weights

patrickhulce:
closes #4629 

tweaks the performance score weights according to https://docs.google.com/document/d/1XiKdkLgWul0l40slhkCmQufNdjUiTlq3H_EAjrBH5Cw/edit#

I ran the weights and new thresholds for what we have in HTTPArchive and got the distribution, it's pretty aggressive.

Median: 27
75th Percentile: 43
90th Percentile: 69
95th Percentile: 83
99th Percentile: 98

Roughly 3% of sites get a 90+
Roughly 0.08% of sites get a 100

I'm disappointed new rolling window EIL won't get any credit, but that'll give us time to take a look at the distribution of the new definition and adjust score thresholds without impacting the overall perf score, and this is aggressive enough as it is :)

PR #5092, opened 2018-05-02
core(scoring): loosen metric thresholds

patrickhulce:
As we just discussed, the new scoring thresholds based on 50/75 were making the requirements for 99/100 too unachievable. The root issue was that our PODR was not considered with our policy. The new policy instead...

Median === 75th percentile
PODR = 95th percentile

This makes a 100 roughly equal to the 98th percentile and keeps the green/orange boundary of 75 to the ~87th percentile which is pretty close to our old policy. Checkout the new scoring calculator sheet https://docs.google.com/spreadsheets/d/1Cxzhy5ecqJCucdf1M0iOzM8mIxNc7mmx107o5nj38Eo/edit#gid=0 to play around with the values.

### DIFF
--- a/docs/scoring.md
+++ b/docs/scoring.md
@@ -1,10 +1,10 @@
 # Goal
 The goal of this document is to explain how scoring works in Lighthouse and what to do to improve your Lighthouse scores across the four sections of the report.
 
-Note 1: if you want a **nice spreadsheet** version of this doc to understand weighting and scoring, check out the [scoring spreadsheet](https://docs.google.com/spreadsheets/d/1dXH-bXX3gxqqpD1f7rp6ImSOhobsT1gn_GQ2fGZp8UU/edit?ts=59fb61d2#gid=0)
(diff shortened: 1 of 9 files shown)


patrickhulce:
> does this affect the decisions that went into #5083?

for better or for worse, the score thresholds/curves of the metrics themselves were not considered at all as part of weighting committee rulings, so this doesn't affect it :)

kaycebasques:
@patrickhulce what does PODR mean?

connorjclark:
> It's roughly a score of ~97

FYI for future code archaeologists: PODR is not necessarily ~97, it varies based on what the median is set as. Everything else @patrickhulce said is true though.

patrickhulce:
Yes, future code archaeologists please do not interpret "roughly a score of ~97" to mean that it in any way is trying to target a score of 97. It just so happened that for many of our curves the PODR is 99-95, so roughly ~97 felt appropriate to say. Everything else @connorjclark said is true 😉

PR #5099, opened 2018-05-03
core(lhr): support printf displayValues

patrickhulce:
adds support for array-based display values in a printf style combined with our `granularity` for formatting numbers, I converted the display values that were most likely to be affected by this (numbers that can reach 1000+), everything else doesn't necessarily need to change but the breaking change has been made and we're free to move between them now :)

I wanted to do the conversion in the report at the beginning so everything else can assume it's a string, but there were only 3 callers, so it didn't seem like a big deal. LMK if others prefer that approach

ref #5008 #4333


lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -53,4 +53,16 @@ describe('util helpers', () => {
     assert.equal(Util.calculateRating(0.80), 'pass');
     assert.equal(Util.calculateRating(1.00), 'pass');
   });
+
+  it('formats display values', () => {
+    const format = (...args) => Util.formatDisplayValue(...args);
+    assert.equal(format(undefined), '');
+    assert.equal(format([1]), 'UNKNOWN');
+    assert.equal(format(['%s %s', 'Hello', 'Paul']), 'Hello Paul');
```
paulirish: Hello formatDisplayValue

lighthouse-core/report/html/renderer/util.js
```diff
@@ -26,6 +26,40 @@ class Util {
...
+    const template = /** @type {string} */ displayValue.shift();
+    if (typeof template !== 'string') {
+      // First value should always be the format string, but we don't want to fail to build
+      // a report, return a placeholder.
+      return 'UNKNOWN';
+    }
+
+    let output = template;
+    while (displayValue.length) {
+      const replacement = /** @type {number|string} */ (displayValue.shift());
```
paulirish:
```js
    assert.equal(format(['%s']), '%s', 'not enough replacements');
    assert.equal(format(['%s', 'a', 'a']), 'a', 'extra replacements');
```

i think we may want to console.warn() in these cases. Allow them, but raise a warning flag.

lighthouse-core/report/html/renderer/util.js
```diff
@@ -26,6 +26,40 @@ class Util {
...
+      output = output.replace(/%([0-9.]+)?(d|s)/, match => {
+        const granularity = Number(match.match(/[0-9.]+/)) || 1;
+        return match === '%s' ?
```
paulirish:
```js
assert.equal(format(['%10s', 123.456]), '120');
```

currently this _does_ pass, fwiw. shoudl we also throw a warning if you are matching 's' and have a granularity?

lighthouse-core/report/html/renderer/util.js
```diff
@@ -26,6 +26,40 @@ class Util {
...
+    return `%10d${NBSP}ms`;
+  }
+
+  /**
+   * @param {string|Array<string|number>=} displayValue
+   * @return {string}
+   */
+  static formatDisplayValue(displayValue) {
+    if (typeof displayValue === 'string' || typeof displayValue === 'undefined') {
+      return displayValue || '';
```
paulirish: i'd prefer being super clear

```js
    if (typeof displayValue === 'string') return displayValue;
    if (typeof displayValue === 'undefined') return '';
```

lighthouse-core/test/audits/estimated-input-latency-test.js
```diff
@@ -29,7 +30,7 @@ describe('Performance: estimated-input-latency audit', () => {
     return Audit.audit(artifacts, {options, settings}).then(output => {
       assert.equal(output.debugString, undefined);
       assert.equal(Math.round(output.rawValue * 10) / 10, 17.1);
-      assert.equal(output.displayValue, '17\xa0ms');
+      assert.equal(Util.formatDisplayValue(output.displayValue), '17\xa0ms');
```
brendankenny: test the real object here like `dom-size-test.js` above?

lighthouse-core/test/audits/first-meaningful-paint-test.js
```diff
@@ -26,7 +27,7 @@ describe('Performance: first-meaningful-paint audit', () => {
     const fmpResult = await FMPAudit.audit(artifacts, context);
 
     assert.equal(fmpResult.score, 1);
-    assert.equal(fmpResult.displayValue, '780\xa0ms');
+    assert.equal(Util.formatDisplayValue(fmpResult.displayValue), '780\xa0ms');
```
brendankenny: same with these

lighthouse-core/test/audits/interactive-test.js
```diff
@@ -33,7 +34,7 @@ describe('Performance: interactive audit', () => {
     return Interactive.audit(artifacts, {options, settings}).then(output => {
       assert.equal(output.score, 1);
       assert.equal(Math.round(output.rawValue), 1582);
-      assert.equal(output.displayValue, '1,580\xa0ms');
+      assert.equal(Util.formatDisplayValue(output.displayValue), '1,580\xa0ms');
```
brendankenny: and these :)

typings/audit.d.ts
```diff
@@ -23,6 +23,8 @@ declare global {
 
     export type ScoringModeValue = Audit.ScoringModes[keyof Audit.ScoringModes];
 
+    export type DisplayValue = string | Array<string|number>
```
brendankenny: there might be an even more elegant way to do this, but you can do
```ts
interface DisplayValueArray extends Array<string | number> {
  0: string;
}

export type DisplayValue = string | DisplayValueArray;
```
to enforce a string in the first position (and a length of at least 1)

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -53,4 +53,16 @@ describe('util helpers', () => {
     assert.equal(Util.calculateRating(0.80), 'pass');
     assert.equal(Util.calculateRating(1.00), 'pass');
   });
+
+  it('formats display values', () => {
+    const format = (...args) => Util.formatDisplayValue(...args);
```
brendankenny: don't need to rest and then spread since they're all arrays below?

lighthouse-core/report/html/renderer/util.js
brendankenny: just switch to regular indexing and avoid these shift assertions? Or could be a reduce

lighthouse-core/report/html/renderer/util.js
```diff
@@ -26,6 +26,40 @@ class Util {
...
+    if (typeof template !== 'string') {
+      // First value should always be the format string, but we don't want to fail to build
+      // a report, return a placeholder.
+      return 'UNKNOWN';
+    }
+
+    let output = template;
+    while (displayValue.length) {
+      const replacement = /** @type {number|string} */ (displayValue.shift());
+      output = output.replace(/%([0-9.]+)?(d|s)/, match => {
```
brendankenny: this allows more than one decimal point? and no support for granularity when it's a string?

lighthouse-core/report/html/renderer/util.js
patrickhulce: done

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -172,9 +172,10 @@ class UnusedBytes extends Audit {
     const wastedKb = Math.round(wastedBytes / KB_IN_BYTES);
     const wastedMs = this.computeWasteWithTTIGraph(results, graph, simulator);
 
+    /** @type {LH.Audit.DisplayValue} */
     let displayValue = result.displayValue || '';
     if (typeof result.displayValue === 'undefined' && wastedBytes) {
-      displayValue = `Potential savings of ${wastedBytes} bytes`;
+      displayValue = ['Potential savings of %d\xa0KB', wastedKb];
```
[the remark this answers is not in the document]
patrickhulce: done

typings/audit.d.ts
patrickhulce: done 👍

lighthouse-core/report/html/renderer/util.js
patrickhulce: correct if you mess up your format string it'll put `NaN`
correct %s is just straight up replacement

lighthouse-core/report/html/renderer/util.js
patrickhulce: done

lighthouse-core/test/report/html/renderer/util-test.js
patrickhulce: done

lighthouse-core/test/audits/interactive-test.js
patrickhulce: done

lighthouse-core/report/html/renderer/util.js
brendankenny: I meant should it be more like `/%([0-9]*(?:\.[0-9]*))?d|s/` or whatever (you covered the no granularity + string case now)

lighthouse-core/test/audits/first-meaningful-paint-test.js
```diff
@@ -39,7 +40,7 @@ describe('Performance: first-meaningful-paint audit', () => {
     const fmpResult = await FMPAudit.audit(artifacts, context);
 
     assert.equal(fmpResult.score, 0.79);
-    assert.equal(fmpResult.displayValue, '2,850\xa0ms');
+    assert.equal(Math.round(fmpResult.displayValue[1]), 2851);
```
brendankenny: lol I meant the whole display value to ensure the whole thing matched what's expected (including format string) but can also revert to the last commit for these if you'd prefer that :)

patrickhulce: will do (revert that is) :)

lighthouse-core/report/html/renderer/util.js
patrickhulce: sure, tweaked it a bit

PR #5100, opened 2018-05-03
core(category): add manualDescription

patrickhulce:
makes the change for manualDescription instead of a manual group

ref #5008 #4333

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -62,7 +62,8 @@ function validateCategories(categories, audits, groups) {
         throw new Error(`could not find ${audit.id} audit for category ${categoryId}`);
       }
 
-      if (categoryId === 'accessibility' && !audit.group) {
+      const auditImpl = audits.find(a => a.implementation.meta.name === audit.id).implementation;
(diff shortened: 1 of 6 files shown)


lighthouse-core/config/config.js
```diff
@@ -62,7 +62,8 @@ function validateCategories(categories, audits, groups) {
         throw new Error(`could not find ${audit.id} audit for category ${categoryId}`);
       }
 
-      if (categoryId === 'accessibility' && !audit.group) {
+      const auditImpl = audits.find(a => a.implementation.meta.name === audit.id).implementation;
```
brendankenny: maybe get rid of `auditIds` above and use `auditImpl` for both this and the check of `audit.id` above?

patrickhulce: done

patrickhulce:
@brendankenny feedback addressed!

patrickhulce:
> should we add a check to validateCategories that any manual audit has a weight of 0?

ya mind if I do that after #5105 ? conflicting junk

brendankenny:
> ya mind if I do that after #5105 ? conflicting junk

sure, but maybe add a TODO?

PR #5101, opened 2018-05-03
core(metrics): consumable metrics audit output

patrickhulce:
after much ado about nothing and the repeated creation/deletion of summary, we're just making a single item inside details that has all the metrics keyed by name and `Ts`. @paulirish approved of this approach

it means that querying will look like `audits.metrics.details.items[0].firstContentfulPaint`

we'll just commit to not breaking that contract with uber typechecking 🎉 

ref #5008

### DIFF
--- a/lighthouse-core/audits/metrics.js
+++ b/lighthouse-core/audits/metrics.js
@@ -83,9 +83,7 @@ class Metrics extends Audit {
 
     for (const [name, value] of Object.entries(metrics)) {
       const key = /** @type {keyof UberMetricsItem} */ (name);
-      if (typeof value === 'undefined') {
-        delete metrics[key];
-      } else {
+      if (typeof value !== 'undefined') {
         metrics[key] = Math.round(value);
       }
     }
--- a/lighthouse-core/audits/metrics.js
+++ b/lighthouse-core/audits/metrics.js
@@ -39,48 +39,104 @@ class Metrics extends Audit {
     const interactive = await artifacts.requestInteractive(metricComputationData);
     const speedIndex = await artifacts.requestSpeedIndex(metricComputationData);
     const estimatedInputLatency = await artifacts.requestEstimatedInputLatency(metricComputationData); // eslint-disable-line max-len
-    const metrics = {};
 
-    // Include the simulated/observed performance metrics
-    const metricsMap = {
-      firstContentfulPaint,
-      firstMeaningfulPaint,
-      firstCPUIdle,
-      interactive,
-      speedIndex,
-      estimatedInputLatency,
+    /** @type {UberMetricsItem} */
+    const metrics = {
+      // Include the simulated/observed performance metrics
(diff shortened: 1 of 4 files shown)


paulirish:
@denar90 @pedro93  @benschwarz @alekseykulikov hows this look to you? you'll get all metrics within this audit instead of relying on `lighthouse-core/lib/traces/pwmetrics-events.js` or pulling diff numbers from various audit's extendedInfo

lighthouse-core/audits/metrics.js
```diff
@@ -39,66 +39,104 @@ class Metrics extends Audit {
...
+    };
 
-    const headings = [
-      {key: 'metricName', itemType: 'text', text: 'Name'},
-      {key: 'timing', itemType: 'ms', granularity: 10, text: 'Value (ms)'},
-    ];
+    for (const [name, value] of Object.entries(metrics)) {
+      const key = /** @type {keyof UberMetricsItem} */ (name);
+      if (typeof value === 'undefined') {
+        delete metrics[key];
```
brendankenny: maybe copy defined values to a new `metrics` object to return rather than delete from the original?

lighthouse-core/lib/traces/pwmetrics-events.js
```diff
@@ -8,13 +8,13 @@
...
-function findValueInMetricsAuditFn(metricName, timingOrTimestamp) {
+function findValueInMetricsAuditFn(metricName) {
   return auditResults => {
     const metricsAudit = auditResults.metrics;
     if (!metricsAudit || !metricsAudit.details || !metricsAudit.details.items) return;
 
-    const values = metricsAudit.details.items.find(item => item.metricName === metricName);
-    return values && values[timingOrTimestamp];
+    const values = metricsAudit.details.items[0];
+    return values && values[metricName];
```
brendankenny: this can be a TODO, but seems like we can just get rid of this whole `findValueInMetricsAuditFn` thing now. If there's no metrics audit, pwmetrics can't really do much, so it could check once at the beginning. Then all the gets are just property dereferencing with the new object layout.

(`metricsDefinitions` might still be useful, haven't looked into it)

lighthouse-core/audits/metrics.js
patrickhulce: any particular reason for this? I'll end up just creating an empty object with no type and casting at the end which seems sad 😢

lighthouse-core/lib/traces/pwmetrics-events.js
patrickhulce: sure TODO added

patrickhulce:
last 📞 , any objections?

lighthouse-core/audits/metrics.js
patrickhulce: yeah I guess it's not the end of the world to leave them undefined, will do

PR #5103, opened 2018-05-03
core(tsc): add type checking for seo audits

brendankenny:
with this, all of `lighthouse-core/**/*.js` is type checked :) (with carve outs for tests and the html renderer, for now).

The only remaining files this adds are these seo audits and the all-important `empty-stub.js` (which turns out was already type safe! Good job `empty-stub`).

### DIFF
--- a/lighthouse-core/audits/seo/font-size.js
+++ b/lighthouse-core/audits/seo/font-size.js
@@ -43,10 +43,10 @@ function getUniqueFailingRules(fontSizeArtifact) {
 }
 
 /**
- * @param {Array<string>} attributes
+ * @param {Array<string>=} attributes
  * @returns {Map<string, string>}
  */
-function getAttributeMap(attributes) {
+function getAttributeMap(attributes = []) {
   const map = new Map();
 
   for (let i = 0; i < attributes.length; i += 2) {
@@ -86,7 +86,8 @@ function getSelector(node) {
  * @return {{type: 'node', selector: string, snippet: string}}
  */
 function nodeToTableNode(node) {
-  const attributesString = node.attributes.map((value, idx) =>
+  const attributes = node.attributes || [];
+  const attributesString = attributes.map((value, idx) =>
     (idx % 2 === 0) ? ` ${value}` : `="${value}"`
   ).join('');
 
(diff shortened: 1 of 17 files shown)


lighthouse-core/audits/seo/font-size.js
```diff
@@ -34,17 +39,17 @@ function getUniqueFailingRules(fontSizeArtifact) {
     }
   });
 
-  return failingRules.valuesArray();
+  return [...failingRules.values()];
```
patrickhulce: I'd have an extraordinarily small preference for `Array.from`, mostly commenting out of curiosity if this is your preference? :)

lighthouse-core/audits/seo/is-crawlable.js
```diff
@@ -89,14 +90,14 @@ class IsCrawlable extends Audit {
           if (isBlocking) {
             blockingDirectives.push({
               source: {
-                type: 'node',
+                type: /** @type {'node'} */ ('node'),
```
patrickhulce: 🤦‍♂️

tsconfig.json
```diff
@@ -16,20 +16,14 @@
...
-    "lighthouse-core/audits/dobetterweb/**/*.js",
-    "lighthouse-core/audits/byte-efficiency/**/*.js",
-    "lighthouse-core/audits/manual/**/*.js",
-    "lighthouse-core/audits/seo/manual/*.js",
-    "lighthouse-core/audits/seo/robots-txt.js",
-    "lighthouse-core/lib/dependency-graph/**/*.js",
-    "lighthouse-core/lib/emulation.js",
-    "lighthouse-core/gather/**/*.js",
-    "lighthouse-core/scripts/*.js",
+    "lighthouse-core/**/*.js",
```
patrickhulce: 🎉

lighthouse-core/audits/seo/font-size.js
brendankenny: hmm, I used to prefer `Array.from` since it felt more deliberate and less like an action in place or something, but now I have spread blindness and see them as the same thing, so I don't think I care anymore...

PR #5108, opened 2018-05-03
core(lhr): lhr type for 3.0

brendankenny:
proposed type for LHR3.0. The goal would be to either move this into `lhr.d.ts` eventually or gradually copy over until the files are identical.

Note that it's strictly derived from LHR-lite so that it can always be cast to it or pruned down to an LHR-lite (any property path that's valid in LHR-lite will lead in the full LHR to a type that's compatible with the LHR-lite type at the same point (it just might be more specific or have more information on it))

This still punts on the details and defers to the LHR-lite ones for now. I can fill those in or we can do it piecemeal.

### DIFF
--- a/typings/lhr-lite.d.ts
+++ b/typings/lhr-lite.d.ts
@@ -15,7 +15,7 @@ declare global {
       /** The post-redirects URL that Lighthouse loaded. */
       finalUrl: string;
       /** The ISO-8601 timestamp of when the results were generated. */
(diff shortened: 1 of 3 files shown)


brendankenny:
(also updated #5008 with all the naming changes needed)

brendankenny:
this is out of date. The LHR types have kept up with v3, what's left is reorganization which can happen elsewhere :)

PR #5112, opened 2018-05-04
core(lhr): s/fetchedAt/fetchTime

brendankenny:
and remove `generatedTime`. From API feedback.

cf. #5008


patrickhulce:
travis is unhappy with closure though :(

brendankenny:
with drive by fix for `displayValue.shift()` warning

PR #5113, opened 2018-05-04
core(tsc): fix ImageUsage artifact type and gather bug

brendankenny:
The type for the `ImageUsage` artifact accidentally left off the optional image `width` and `height`, which made using it in the `image-aspect-ratio` audit impossible before. Fixed the artifact and removed the `@ts-nocheck` from the audit.

In the process, uncovered a bug in the gatherer. At some point shifting types made the extra fetch of size information never run for css or `<picture>` images. As a result, the `naturalWidth` and `naturalHeight` was being left at `Number.MAX_VALUE` for those and so the `wastedPercent` for those images was 100% in the `uses-responsive-images` audit :)

There was a [clue in the smoke code coverage](https://codecov.io/gh/GoogleChrome/lighthouse/src/aad3aa6/lighthouse-core/gather/gatherers/image-usage.js#L131), but we hadn't noticed yet :)

Fixed the bug and added more specific expectations for that audit.

### DIFF
--- a/lighthouse-core/gather/gatherers/image-usage.js
+++ b/lighthouse-core/gather/gatherers/image-usage.js
@@ -15,7 +15,7 @@ const Driver = require('../driver.js'); // eslint-disable-line no-unused-vars
 
 /* global window, getElementsInDocument, Image */
 
-/** @typedef {MakeOptional<LH.Artifacts.SingleImageUsage, 'networkRecord'>} ImageSizeInfo */
+/** @typedef {Omit<LH.Artifacts.SingleImageUsage, 'networkRecord'>} ImageSizeInfo */
 
 /** @return {Array<ImageSizeInfo>} */
 /* istanbul ignore next */
@@ -145,7 +145,7 @@ class ImageUsage extends Gatherer {
    * @param {LH.Gatherer.LoadData} loadData
    * @return {Promise<LH.Artifacts['ImageUsage']>}
    */
-  afterPass(passContext, loadData) {
+  async afterPass(passContext, loadData) {
     const driver = passContext.driver;
     const indexedNetworkRecords = loadData.networkRecords.reduce((map, record) => {
       if (/^image/.test(record._mimeType) && record.finished) {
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -167,27 +167,24 @@ class ImageUsage extends Gatherer {
...
 
-    /** @type {Promise<Array<ImageSizeInfo>>} */
-    const evaluatePromise = driver.evaluateAsync(expression);
-    return evaluatePromise.then(elements => {
-      return elements.reduce((promise, element) => {
-        return promise.then(collector => {
-          // Images within `picture` behave strangely and natural size information isn't accurate,
-          // CSS images have no natural size information at all.
-          // Try to get the actual size if we can.
-          const elementPromise = (element.isPicture || element.isCss) && element.networkRecord ?
```
brendankenny: this was the bug. `element.networkRecord` was never defined at this point

lighthouse-core/gather/gatherers/image-usage.js
```diff
@@ -167,27 +167,24 @@ class ImageUsage extends Gatherer {
...
+    /** @type {Array<ImageSizeInfo>} */
+    const elements = await driver.evaluateAsync(expression);
+
+    /** @type {LH.Artifacts['ImageUsage']} */
+    const imageUsage = [];
+    for (let element of elements) {
+      // Images within `picture` behave strangely and natural size information isn't accurate,
+      // CSS images have no natural size information at all.
+      // Try to get the actual size if we can.
+      if (element.isPicture || element.isCss) {
```
patrickhulce: should we maintain the original concern and only fetch when we can find the network record? `fetchElementWithSize` is the expensive path

brendankenny:
> good news is looks like this was introduced just 11 days ago 

lol, hadn't looked at the blame yet :) Shows the danger of trying to do a `Partial<>` without looking deeper at the logic (or living without tests). `networkRecords` isn't actually used in `fetchElementWithSizeInformation` so it seemed fine to put first.

lighthouse-core/gather/gatherers/image-usage.js
brendankenny: > should we maintain the original concern and only fetch when we can find the network record? 

Only `uses-responsive-images` seems to use `naturalWidth`/`naturalHeight` from css/`<picture>` images, and it requires a networkRecord to work, so makes sense

brendankenny: > when we can find the network record?

...actually this conversation clearly shows that it should have been an optional property...

brendankenny:
sorry for the expanding scope, but this should properly reflect the `ImageUsage` artifact now. Logic was already in place to check if there wasn't a `networkRecord`, but had to do some slight rearranging to convince the compiler of this.

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -55,11 +57,15 @@ class OffscreenImages extends ByteEfficiencyAudit {
   }
 
   /**
-   * @param {!Object} image
```
brendankenny: not sure how this slipped by :)

PR #5123, opened 2018-05-04
report: rename audit title 'Efficiently encode images' => 'Serve optimized images'

paulirish:
bikeshed time.

patrick sez "optimized" is too vague"
paul thinks everyone considers this operation to be "image optimization"

how about.. just a strawman...

"Re-encode images for max compression"


patrickhulce:
> paul thinks everyone considers this operation to be "image optimization"

I propose a twitter poll to settle this 😛 

"When you're told to 'optimize your images', what does that mean?

A) Resize your images to their displayed size
B) Re-encode images at a lower quality setting
C) Lazily load your images
D) All of the above
E) Don't know
"

Claim A: There will not be a preference for B in that list
Claim B: Even if there were, it is not going to be sufficiently clear (i.e. 80%+ pick B) such that we should not be specific here :)

PR #5124, opened 2018-05-04
core(lantern): never exclude main document from graphs

patrickhulce:
should fix the travis flake we've been seeing on extension tests `cannot find node ID 02u309rjfja0fsdjflajsfd`

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const INITIAL_CWD = 14 * 1024;
+const WebInspector = require('../../web-inspector');
 
 class NetworkAnalyzer {
   /**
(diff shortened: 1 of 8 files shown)


patrickhulce:
@paulirish @brendankenny no longer WIP, it's 9/9 so far on passing extension test so lookin' good

lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
```diff
@@ -319,6 +320,17 @@ class NetworkAnalyzer {
     const estimatesByOrigin = NetworkAnalyzer._estimateResponseTimeByOrigin(records, rttByOrigin);
     return NetworkAnalyzer.summarize(estimatesByOrigin);
   }
+
+  /**
+   * @param {Array<LH.WebInspector.NetworkRequest>} records
+   * @return {LH.WebInspector.NetworkRequest}
+   */
+  static findMainDocument(records) {
+    // TODO(phulce): handle more edge cases like client redirects, or plumb through finalUrl
```
patrickhulce: adding `finalUrl` to all calls of request* is just insane work atm, should be easier after a refactor to pass in artifacts and make that dance a little easier

patrickhulce:
if you fellas wanna take a look we can have a lot more green builds ;)

PR #5127, opened 2018-05-04
core(lhr): s/initialUrl/requestedUrl, s/url/finalUrl

brendankenny:
also removes the `URL` gatherer and collects the `{requestedUrl, finalUrl}` artifact in GatherRunner itself (alongside `fetchTime`, `UserAgent`, etc). It will now always be collected (which makes sense).

This allows us to finally break the mutable `options.url` from outside `Runner.run` all the way down into every gatherer :):) `passContext.url` is still mutable and follows the same behavior as before (to allow things like the `http-redirect` gatherer to work), but the connection is broken outside of that except for the single explicit write to the artifact.


lighthouse-core/runner.js
```diff
@@ -119,12 +119,14 @@ class Runner {
         reportCategories = ReportScoring.scoreAllCategories(opts.config.categories, resultsById);
       }
 
+      const finalUrl = artifacts.URL ? artifacts.URL.finalUrl : '';
```
patrickhulce: when do we not have one? isn't it always set now?

lighthouse-core/test/fixtures/artifacts/empty-artifacts/artifacts.json
```diff
@@ -1 +1,6 @@
-{}
+{
```
patrickhulce: `empty-ish-artifacts`

lighthouse-core/test/index-test.js
```diff
@@ -88,7 +88,7 @@ describe('Module Tests', function() {
   });
 
   it('should return formatted LHR when given no categories', function() {
-    const exampleUrl = 'https://example.com/';
+    const exampleUrl = 'https://www.reddit.com/r/nba';
```
patrickhulce: lol, any particular reason?

lighthouse-core/runner.js
paulirish: lol i'm reading the opposite. since this PR nukes the URL gatherer how is `artifacts.URL` set?

brendankenny: good catch. Left over from getting un-updated tests to run :)

only future issue for those of us using saved artifacts (aka @paulirish) is that this is a breaking saved artifacts change, and since it'll be "cannot read property 'finalUrl' of undefined", this will throw on any `-A` run with previously saved artifacts instead of producing an LHR with an empty `finalUrl`

lighthouse-core/test/index-test.js
brendankenny: I think thats what the devtoolsLog was for those fixtures, which spread to the artifacts, and now to the input URL (though I guess technically only the `finalUrl` assertion would have to change :)

lighthouse-core/runner.js
paulirish: yup! no worries. :)

PR #5105, opened 2018-05-03
core(lhr): drop informative/manual, moving them to scoreDisplayMode

patrickhulce:
removes informative/manual/notApplicable booleans from `LH.Audit.Result`, uses `scoreDisplayMode` for all of these now

keeps `notApplicable` on the `LH.Audit.Product` though which `createAuditResult` uses to set the scoreDisplayMode to `not-applicable`, it felt a little more natural to me as the runtime split, but curious what other folks think

@paulirish FYI this probably has a few implications for your design changes i.e. an audit can no longer be both binary informative (made some blue 0s show up)

ref #5008 #4333

PR #5128, opened 2018-05-05
core(lhr): support null scores

patrickhulce:
followup to #5105, add support for null scores and forces all errors/informative/manual/not-applicable to be null

when category errors, the gauge shows '?' ~~I'll try to find a screenshot~~ :)

image


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -275,12 +274,12 @@ class CategoryRenderer {
         group = auditsUngrouped;
       }
 
-      if (audit.result.scoreDisplayMode === 'not-applicable') {
-        group.notApplicable.push(audit);
-      } else if (audit.result.score === 1 && !audit.result.debugString) {
```
paulirish: that logic on debugString seems rather incorrect tbh. glad it's gone.

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -17,7 +17,8 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     const element = this.dom.find('.lh-metric', tmpl);
     element.id = audit.result.name;
     // FIXME(paulirish): currently this sets a 'lh-metric--fail' class on error'd audits
```
paulirish: can you also address this?

i'd prefer do not set a --fail class on these audits.

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -228,14 +225,16 @@ class CategoryRenderer {
...
     // https://codepen.io/xgad/post/svg-radial-progress-meters
     // score of 50: `stroke-dasharray: 164.5 329`;
-    this.dom.find('.lh-gauge-arc', gauge).style.strokeDasharray = `${category.score * 329} 329`;
+    this.dom.find('.lh-gauge-arc', gauge).style.strokeDasharray = `${numericScore * 329} 329`;
 
-    const scoreOutOf100 = Math.round(category.score * 100);
-    this.dom.find('.lh-gauge__percentage', tmpl).textContent = scoreOutOf100;
+    const scoreOutOf100 = Math.round(numericScore * 100);
+    this.dom.find('.lh-gauge__percentage', tmpl).textContent = category.score === null ?
+      '?' : scoreOutOf100;
```
paulirish: can you also apply a title attribute to that explains why its ?

lighthouse-core/scoring.js
```diff
@@ -18,17 +18,22 @@ const clampTo2Decimals = val => Math.round(val * 100) / 100;
...
   /**
    * Computes the weighted-average of the score of the list of items.
-   * @param {Array<{score: number, weight: number}>} items
-   * @return {number}
+   * @param {Array<{score: number|null, weight: number}>} items
+   * @return {number|null}
    */
   static arithmeticMean(items) {
+    // If there is 1 null score, return a null average
+    if (items.some(item => item.score === null && item.weight !== 0)) return null;
```
paulirish: wdyt about an early filter in this method to get rid of the weight===0 items. then we don't need the check in here PLUS L32.
ndb, tho

lighthouse-core/report/report-generator.js
```diff
@@ -73,7 +73,7 @@ class ReportGenerator {
       return category.audits.map(catAudit => {
         const audit = lhr.audits[catAudit.id];
         return [category.name, audit.name, audit.description, audit.scoreDisplayMode, audit.score]
-          .map(value => value.toString())
+          .map(value => value === null ? '0' : value.toString())
```
paulirish: i dont know what the CSV consumers want but i feel like getting 0 is a little weird. empty string seems slightly better.

lighthouse-core/report/html/renderer/util.js
```diff
@@ -70,12 +70,39 @@ class Util {
...
+    switch (audit.scoreDisplayMode) {
+      case 'manual':
+      case 'not-applicable':
+        return true;
+      case 'error':
+      case 'informative':
+        return false;
+      case 'numeric':
+      case 'binary':
+        return audit.score === 1;
```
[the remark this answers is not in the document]
patrickhulce: yeah good call 👍

lighthouse-core/report/html/renderer/util.js
```diff
@@ -70,12 +70,39 @@ class Util {
...
+   *
+   * @param {{score: number|null, scoreDisplayMode: string}} audit
+   */
+  static didAuditPass(audit) {
+    switch (audit.scoreDisplayMode) {
+      case 'manual':
+      case 'not-applicable':
+        return true;
+      case 'error':
+      case 'informative':
```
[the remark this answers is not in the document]
patrickhulce: so they show up as failing, if an informative audit has nothing to tell you it marks itself as `not-applicable` ;)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,19 +144,20 @@ describe('CategoryRenderer', () => {
...
-      categoryDOM.querySelectorAll('.lh-audit-group--notapplicable .lh-audit').length,
+      categoryDOM.querySelectorAll('.lh-audit-group--not-applicable .lh-audit').length,
       notApplicableCount,
       'score shows informative and dash icon'
     );
 
-    const perfCategory = sampleResults.reportCategories.find(cat => cat.id === 'performance');
-    const categoryDOM2 = renderer.render(perfCategory, sampleResults.reportGroups);
-    assert.ok(!categoryDOM2.querySelector('.lh-audit-group--notapplicable'));
+    const bestPracticeCat = sampleResults.reportCategories.find(cat => cat.id === 'best-practices');
```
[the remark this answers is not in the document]
patrickhulce: yeah bingo

lighthouse-core/report/html/renderer/performance-category-renderer.js
patrickhulce: haha sure

lighthouse-core/report/html/renderer/category-renderer.js
patrickhulce: done, text suggestion welcome right now "Errors occurred while auditing"

lighthouse-core/scoring.js
patrickhulce: yeah sure sg 👍

lighthouse-core/report/report-generator.js
patrickhulce: csv-validator requires that all values in the column are numbers, I have no preference here we can ditch csv-validator if we want to deviate?  

cc @nourikhalass in case you have a preference for how `null` scores show up in CSV

nourikhalass: I agree with @paulirish that 0 is a bit weird so empty string would be better. But as @patrickhulce says: csv-validator does not play along with that.  Maybe its better to use -1 or omit the result entirely if the score is null.

patrickhulce: I'll go with -1 for now since you're our primary CSV consumer, PR welcome in the future if you think of a better canonical CSV-way to represent this :)

lighthouse-core/report/html/renderer/util.js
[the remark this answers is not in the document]
paulirish: aiight sg.

paulirish:
@brendankenny you wanna take a look?

brendankenny:
looks like manual expectations need updating

lighthouse-core/audits/audit.js
```diff
@@ -162,18 +163,26 @@ class Audit {
 
     // If the audit was determined to not apply to the page, we'll reset it as informative only
     if (result.notApplicable) {
-      score = 1;
       scoreDisplayMode = Audit.SCORING_MODES.NOT_APPLICABLE;
       result.rawValue = true;
     }
 
+    if (result.error) {
```
brendankenny: is it possible to rework this code to be a flow through a set of mutually exclusive states so it's easy to see the possibilities?

e.g. score is null, if `error`, it's always scoring mode error, else if it's been marked `notApplicable`, it's mode `notApplicable`, then `manual`, and then only if it's binary or numeric is a score calculated?

`description` seems like it might not fit the mold, but then maybe it's worth calling separately if it sits across the display modes differently than `score` does

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -228,14 +225,21 @@ class CategoryRenderer {
     wrapper.href = `#${category.id}`;
     wrapper.classList.add(`lh-gauge__wrapper--${Util.calculateRating(category.score)}`);
 
+    const numericScore = Number(category.score);
```
brendankenny: should call out casting `null` to 0 here in a comment

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -259,11 +263,11 @@ class CategoryRenderer {
       notApplicable: !Array<!ReportRenderer.AuditJSON>}>} */ ({});
     const auditsUngrouped = {passed: [], failed: [], notApplicable: []};
 
-    nonManualAudits.forEach(audit => {
+    nonManualAudits.forEach(auditRef => {
```
brendankenny: 🎉

lighthouse-core/report/html/renderer/util.js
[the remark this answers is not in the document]
brendankenny: should record this idea somewhere (to the lhr/lhr-lite description of these, maybe?)

lighthouse-core/report/html/renderer/util.js
```diff
@@ -70,12 +71,42 @@ class Util {
     return output;
   }
 
+  /**
+   * @param {{score: (number|null), scoreDisplayMode: string}} audit
+   * @return {boolean}
+   */
+  static didAuditPass(audit) {
```
brendankenny: maybe this shouldn't be considered if the audit "passed" (we even call out informative audits as not being able to be interpreted as pass/fail lol), but I can't think of a better word and this is the way we divide the sections in the report. Maybe add a comment that this is specifically for that distinction?

lighthouse-core/scoring.js
```diff
@@ -56,9 +62,12 @@ class ReportScoring {
         // will still be included in the final report json and displayed in the report as
         // "Not Applicable".
         const result = resultsByAuditId[member.id];
-        if (result.scoreDisplayMode === Audit.SCORING_MODES.NOT_APPLICABLE) {
+        if (result.scoreDisplayMode === Audit.SCORING_MODES.NOT_APPLICABLE ||
+            result.scoreDisplayMode === Audit.SCORING_MODES.INFORMATIVE ||
+            result.scoreDisplayMode === Audit.SCORING_MODES.MANUAL) {
           member.weight = 0;
         }
+
```
brendankenny: I think these members are `auditRef`s now if you want to do a drive by in this function :)

lighthouse-core/audits/audit.js
```diff
@@ -162,18 +163,26 @@ class Audit {
 
     // If the audit was determined to not apply to the page, we'll reset it as informative only
```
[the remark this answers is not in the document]
patrickhulce: fixed

lighthouse-core/audits/audit.js
patrickhulce: this gets reworked immediately as part of #5132, fair to take up there?

lighthouse-core/report/html/renderer/category-renderer.js
patrickhulce: done

lighthouse-core/report/html/renderer/util.js
patrickhulce: yeah sure sg, I went back and forth on name, basically "shouldHideAudit" but not always

patrickhulce: actually, I could flip and do "shouldConsiderFailing" or something?

patrickhulce:
> looks like manual expectations need updating

whoops yeah fixed them on the wrong branch 😆

lighthouse-core/report/html/renderer/util.js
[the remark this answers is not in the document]
patrickhulce: It's more an audit author concern since it's an immediate conclusion from the scoreDisplayMode that if an audit isn't `not-applicable` it means it was applicable :)

maybe in the custom audit docs?

lighthouse-core/report/report-generator.js
nourikhalass: I think that's fine since the validation library leaves a lot to be desired. Maybe in the future I'll look at a better / more robust way of validating the CSV . So I think you'll be OK if you don't validate that scores are numeric.

Also thanks for including me in the loop!

lighthouse-core/audits/audit.js
brendankenny: > fair to take up there?

yeah, SG

PR #5129, opened 2018-05-05
core(tsc): fix OptimizedImages type; type check dep audits

brendankenny:
make `OptimizedImages` a discriminated union on `failed` or not. Keeps the artifact type simpler and the use of it really simple in these two audits.

The only remaining `// @ts-nocheck` audit is `mixed-content.js`, which should make refactors a little more easier.

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-webp-images.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-webp-images.js
@@ -6,7 +6,6 @@
 /*
  * @fileoverview This audit determines if the images could be smaller when compressed with WebP.
  */
-// @ts-nocheck - TODO(bckenny)
 'use strict';
 
 const ByteEfficiencyAudit = require('./byte-efficiency-audit');
@@ -47,7 +46,9 @@ class UsesWebPImages extends ByteEfficiencyAudit {
   static audit_(artifacts) {
     const images = artifacts.OptimizedImages;
 
+    /** @type {Array<LH.Artifacts.OptimizedImageError>} */
     const failedImages = [];
+    /** @type {Array<{url: string, fromProtocol: boolean, isCrossOrigin: boolean, totalBytes: number, wastedBytes: number}>} */
     const results = [];
     images.forEach(image => {
       if (image.failed) {
--- a/lighthouse-core/audits/byte-efficiency/uses-webp-images.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-webp-images.js
@@ -46,16 +46,15 @@ class UsesWebPImages extends ByteEfficiencyAudit {
   static audit_(artifacts) {
     const images = artifacts.OptimizedImages;
 
-    /** @type {Array<LH.Artifacts.OptimizedImageError>} */
-    const failedImages = [];
     /** @type {Array<{url: string, fromProtocol: boolean, isCrossOrigin: boolean, totalBytes: number, wastedBytes: number}>} */
     const results = [];
-    images.forEach(image => {
+    const failedImages = [];
+    for (const image of images) {
       if (image.failed) {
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -169,36 +173,42 @@ class OptimizedImages extends Gatherer {
...
+      try {
+        const stats = await this.calculateImageStats(driver, record);
+        if (stats === null) {
+          continue;
+        }
 
-  /** @typedef {{isSameOrigin: boolean, isBase64DataUri: boolean, requestId: string, url: string, mimeType: string, resourceSize: number}} SimplifiedNetworkRecord */
+        results.push({
+          failed: false,
+          ...stats,
```
patrickhulce: 🎉

lighthouse-core/audits/byte-efficiency/uses-webp-images.js
```diff
@@ -47,7 +46,9 @@ class UsesWebPImages extends ByteEfficiencyAudit {
   static audit_(artifacts) {
     const images = artifacts.OptimizedImages;
 
+    /** @type {Array<LH.Artifacts.OptimizedImageError>} */
     const failedImages = [];
+    /** @type {Array<{url: string, fromProtocol: boolean, isCrossOrigin: boolean, totalBytes: number, wastedBytes: number}>} */
```
paulirish: since this is used in multiple audits can we share it? i'm okay either way.

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -169,36 +173,42 @@ class OptimizedImages extends Gatherer {
...
-            return results.concat(Object.assign(stats, record));
-          });
-      });
-    }, Promise.resolve(result));
-  }
+    const results = [];
+
+    for (const record of imageRecords) {
+      try {
+        const stats = await this.calculateImageStats(driver, record);
```
paulirish: throwing this out there... maybe we should let these race and promise.all them instead. would probably be faster.

but i cant remember if I've seen this gatherer being particularly slow..

paulirish: k sg. then yah nvm. we'll keep it running serially.

lighthouse-core/audits/byte-efficiency/uses-webp-images.js
brendankenny: it's almost but not quite the same as we use in other audits, enough to be annoying to redefine. `url` has to be a string and there's the extra `isCrossOrigin` and `fromProtocol` properties. I don't think it's worth defining externally for just this and `uses-optimized-images`, though, since it just becomes a `ByteEfficiencyResult` to the rest of the pipeline

brendankenny: I can make the `failedImages` type be inferred, though

brendankenny:
Yet another browserify syntax error. It's on object spread again, this time in [lexical-scope](https://www.npmjs.com/package/lexical-scope), called by [insert-module-globals](https://www.npmjs.com/package/insert-module-globals). It appears it might be an actual bug, though, and not another parser version that needs to be piped through because you can actually use object spread in some files and not others. e.g. not in `optimized-images` or `accessibility` gatherers, but it's fine in the `websql` or `css-usage` gatherers. I haven't figured out what the error cases have in common yet or how to fix them.

brendankenny:
gave up on fixing object spread for now and filed #5152 :)

Have to `ts-ignore` these `Object.assign()`s because tsc widens the `failed` from their specific values to `boolean`

PR #5132, opened 2018-05-05
core(lhr): remove debugString, add explantion/errorMessage

patrickhulce:
I am *SUPER* jazzed about this change. I think it cleans up the intention behind everything *A LOT* and is extra nice. There is no more debug string exposed at the audit level (manifest-parser still returns a debug string from gatherer but we can clean that up later). Existing debug strings got recategorized into one of 

1. displayValue - for cases like appcache where we just communicated the name of the thing we found
2. errorMessage - when the message was a true error message/failure case
3. warnings - for when the message was just a warning. This is the biggest behavioral change in the PR, the warnings are now collected and displayed in the top-level warnings instead of popping the specific audit into failing. **I request that you only consider if warnings should be exposed per-audit in the LHR for this PR and not discuss the display mechanism** we can always change how the warnings manifest in the report later, so let's just focus on the breaking-LHR-relevant change.
4. explanation - the default case if the debugString didn't obviously fit into the other 3, these are currently displayed exactly as debugString was before, so no real impact other than name change. I have ideas for how to change the display but we can 🚲 🏠 that another day


vinamratasingal:
Not sure if my approval here is really needed but I'll give a LGTM anyways :P

lighthouse-core/runner.js
```diff
@@ -111,7 +111,14 @@ class Runner {
 
       /** @type {Object<string, LH.Audit.Result>} */
       const resultsById = {};
-      for (const audit of auditResults) resultsById[audit.name] = audit;
+      for (const audit of auditResults) {
+        resultsById[audit.name] = audit;
+
+        if (audit.warnings.length) {
+          const prefixedWarnings = audit.warnings.map(msg => `${audit.description}: ${msg}`);
+          lighthouseRunWarnings.push(...prefixedWarnings);
```
paulirish: all warnings are pushed to the toplevel warnings?

relatedly: i haven't seen a top level warning in the report in a long time. :)

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -99,7 +99,8 @@ describe('PerfCategoryRenderer', () => {
     assert.ok(oppSparklineElement.title, 'did not set tooltip on sparkline');
   });
 
-  it('renders the performance opportunities with a debug string', () => {
+  // TODO(phulce): verify these don't happen anymore
+  it.skip('renders the performance opportunities with a debug string', () => {
```
paulirish: in a branch of mine, i handle this case.

lighthouse-core/test/results/sample_v2.json
```diff
@@ -4356,30 +4466,31 @@
...
     },
     "robots-txt": {
       "score": null,
-      "displayValue": "",
       "rawValue": null,
-      "error": true,
-      "debugString": "Audit error: Required RobotsTxt gatherer did not run.",
+      "displayValue": "",
+      "errorMessage": "Audit error: Required RobotsTxt gatherer did not run.",
+      "warnings": [],
```
paulirish: can we leave warnings undefined if there are none?
same thing with displayValue, tbh. :)

lighthouse-core/runner.js
patrickhulce: > all warnings are pushed to the toplevel warnings?

correct, I figure we can always change this without breaking LHR changes later if we don't like it/prefer to flag the individual audits, buuuuuut as you point out...

> relatedly: i haven't seen a top level warning in the report in a long time. :)

agreed I can't remember seeing one naturally, so seems like a good thing to try this out

lighthouse-core/test/results/sample_v2.json
patrickhulce: yeah sgtm 👍 @brendankenny agree?

patrickhulce:
Hmmm strange @paulirish they're working fine for me, is this merged with some of your branches perhaps and mixing with the single-use displayValue?

image

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
patrickhulce: cool beans

patrickhulce:
@paulirish diff is ~100 lines smaller with undefined warnings/displayValue 🎉

brendankenny:
travis/appveyor seem real confused

lighthouse-core/test/results/sample_v2.json
brendankenny: @brendankenny agree? yeah, sounds good

patrickhulce:
> travis/appveyor seem real confused

yeah I don't think they rebuild on change of base, just force pushed

lighthouse-core/audits/seo/robots-txt.js
```diff
@@ -182,7 +182,7 @@ class RobotsTxt extends Audit {
     if (!status) {
       return {
         rawValue: false,
-        debugString: 'Lighthouse was unable to download your robots.txt file',
+        explanation: 'Lighthouse was unable to download your robots.txt file',
```
brendankenny: is this a failure case on our part or the site's?

lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
```diff
@@ -40,9 +41,8 @@ class ExternalAnchorsUseRelNoopenerAudit extends Audit {
         try {
           return new URL(anchor.href).host !== pageHost;
         } catch (err) {
-          debugString = 'Lighthouse was unable to determine the destination ' +
-              'of some anchor tags. If they are not used as hyperlinks, ' +
-              'consider removing the _blank target.';
+          // TODO(phulce): make this message better with anchor.outerHTML
```
brendankenny: old message had a lot more explanation :)

lighthouse-core/audits/audit.js
```diff
@@ -167,7 +166,7 @@ class Audit {
       result.rawValue = true;
     }
 
-    if (result.error) {
+    if (result.errorMessage) {
```
brendankenny: time to rework re: https://github.com/GoogleChrome/lighthouse/pull/5128#discussion_r186307641  ?

brendankenny:
some of the TODOs clearly seem like for the future (e.g. `debugString` in gatherer output), but others seem like the needed to be decided soonish/in this PR? Sorry if i misclassified :)

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
patrickhulce: yes indeed, removed

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -113,7 +113,8 @@ describe('CategoryRenderer', () => {
     category.description = prevDesc;
   });
 
-  it('renders audits with debugString as failed', () => {
+  // TODO(phulce): decide what to do about these cases
```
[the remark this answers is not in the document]
patrickhulce: updated the comment. this is what's up for debate re: toplevel warnings

currently the test simply doesn't apply anymore since all warnings are stuck into toplevel for display

lighthouse-core/audits/seo/robots-txt.js
patrickhulce: depends on what you mean by "our part" it could be because the server aborted the request/bad server/etc or it could be because the network failed, either way it seems to belong in explanation instead of errorMessage since there's nothing we would change code-wise, it's expected behavior

lighthouse-core/audits/dobetterweb/external-anchors-use-rel-noopener.js
patrickhulce: fixed

patrickhulce:
> some of the TODOs clearly seem like for the future (e.g. debugString in gatherer output), but others seem like the needed to be decided soonish/in this PR? Sorry if i misclassified :)

yeah mostly just called out for discussion which it has sparked so 👍 

only one remaining is in response to your other comments which we just discussed in person 

> It's definitely harder to scan across to the value for the audit vs just having it right below

We've definitely entered highly subjective territory here because I disagree 😄 Either way, we can *always* change where we put/color/highlight displayValue later on so we'll pickup discussion of best display value designs in a separate issue.

patrickhulce:
filed https://github.com/GoogleChrome/lighthouse/issues/5153 @brendankenny 👍

lighthouse-core/audits/seo/robots-txt.js
brendankenny: I guess this is the nuance we should expose so the user has some recourse (either file a bug on lighthouse or know how to start debugging their site). Really a bug on robots-txt gatherer so fine to leave

brendankenny:
> we can always change where we put/color/highlight displayValue later on so we'll pickup discussion of best display value designs in a separate issue.

yeah, it was just a comment on the display of `displayValue`. Out of scope with this PR

PR #5136, opened 2018-05-06
report: DRY up audit & opportunity rendering

paulirish:
Gets rid of a lot of duplicated work, and brings a lovely consistency to the styles.

https://lighthouse-oppoasanaudit.now.sh/

PR #5137, opened 2018-05-06
report: animated chevrons

paulirish:
Thx to @hwikyounglee for putting up with me as I was like "OH HECK NO".
Because I enjoyed making this work. :D 

seagulls2

This is dependent on #5136

-----

https://lighthouse-chevrons.now.sh/


lighthouse-core/report/html/report-styles.css
```diff
@@ -916,6 +884,47 @@ summary.lh-passed-audits-summary {
...
+  margin-top: calc((var(--body-line-height) - 12px) / 2);
+}
+
+.lh-chevron__lines {
+  transition: transform 0.4s;
+  transform: translateY(20px);
+}
+.lh-chevron__line {
+ stroke: hsl(216, 5%, 39%);
+ transform-origin: 50px 50px;
```
patrickhulce: can any of this be set to a variable to scale the chevron or is it all very precarious?

hwikyounglee:
Thank you @paulirish 😊🙃🕺💃

paulirish:
> my hesitation with approving is setting the expectation we can do fancy stuff like this left and right 😆

you can try this out on https://sweltering-ground.surge.sh/cnn.html
when its not zoomed in, i think the effect is pretty subtle.

regarless, in general, i think that with hwi's guidance we won't overdo things. :)

paulirish:
oooooooooooh gotcha. :D

lighthouse-core/report/html/report-styles.css
paulirish: good call. done.

PR #5154, opened 2018-05-07
core(runner): support multiple output modes

patrickhulce:
returns support for multiple outputs

I know there's some ugliness in config to get this to work with unified config/CLI settings, but that'll need some good cleanup for type checking soon anyhow

biggest question for 3.0: **do we want to have report/output types be `string|string[]` or `string[]`** 

I did the `|` because it's easier impl wise and won't break common `output: 'html'` case, but it's new functionality anyhow so now's our chance

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ last-run-results.html
 *.report.html
 *.report.dom.html
 *.report.json
+*.report.csv
 *.report.pretty
 *.artifacts.log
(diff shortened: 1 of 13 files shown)


brendankenny:
I haven't looked at the code yet, but agree on `string|Array<string>`. I think the common case will still be the single output mode, so no need to make it harder for that one

PR #5155, opened 2018-05-07
core(lhr): convert reportCategories to categories object

patrickhulce:
* renames reportCategories -> categories
* converts categories to a keyed object instead of array
* renames reportGroups -> categoryGroups

ref #4333 #5008


lighthouse-core/runner.js
```diff
@@ -119,10 +119,10 @@ class Runner {
...
-      /** @type {Array<LH.Result.Category>} */
-      let reportCategories = [];
+      /** @type {Object<string, LH.Result.Category>} */
+      let categories = {};
       if (opts.config.categories) {
-        reportCategories = ReportScoring.scoreAllCategories(opts.config.categories, resultsById);
+        categories = ReportScoring.scoreAllCategories(opts.config.categories, resultsById);
       }
 
       const lhr = {
```
brendankenny: we should probably put a `/** @type {LH.Audit.Result} */` above this so it's checked for excess properties not in the def (`reportGroups` -> `categoryGroups` isn't caught by the un-updated type in `lhr.d.ts` since `reportGroups` was optional)

typings/lhr.d.ts
```diff
@@ -21,7 +21,7 @@ declare global {
       /** An object containing the results of the audits, keyed by the audits' `id` identifier. */
       audits: Record<string, Audit.Result>;
       /** The top-level categories, their overall scores, and member audits. */
-      reportCategories: Result.Category[];
+      categories: Record<string, Result.Category>;
       /** Descriptions of the groups referenced by CategoryMembers. */
       reportGroups?: Record<string, Result.Group>;
```
brendankenny: `categoryGroups`

brendankenny: and do we want the interface of the values to be `ReportGroup` to match our usual plural/singular thing?

patrickhulce: yeah good call fixed!

lighthouse-core/runner.js
patrickhulce: yeah good call done

PR #5080, opened 2018-05-01
deps(snyk): prettify snyk snapshot

patrickhulce:
should make future updates to snyk much more readable

PR #5162, opened 2018-05-09
core(vulns): snyk snapshot will be updated by PRs

aviadatsnyk:
Removing this script since it might be misleading, creating the impression the snapshot is updated by it, which is not the case.


patrickhulce:
Hey there @aviadatsnyk! The script is what downloads and updates the file to make PRs easier though, #5080 makes the output even cleaner. Is there an alternative, even easier approach you'd recommend? :)

paulirish:
see https://github.com/GoogleChrome/lighthouse/pull/5239#issuecomment-391112662

PR #5190, opened 2018-05-11
core(lhr): audit id, title, description

patrickhulce:
phase 1 of audit property renames, this PR changes the LHR property names but not the source ones in the audits (we can do that piecemeal if need be since that'll be huge)


patrickhulce:
> Is this just in meta objects

mostly just in meta objects but it touches every audit and a few places that consume direct from meta

PR #5191, opened 2018-05-11
tests(viewer): upgrade pptr to handle new CSSOM use in the report

brendankenny:
`computedStyleMap` in the report fails in the Chromium version that our current version of puppeteer ships with. This revs the version so that we get a more recent Chromium.

Drive by change to viewer and the viewer pptr test so that when there's an in-page error like this, the test can pick it up and print it to console instead of just timing out in `before()` waiting for the report to load.

### DIFF
--- a/lighthouse-core/report/html/renderer/logger.js
+++ b/lighthouse-core/report/html/renderer/logger.js
@@ -49,6 +49,11 @@ class Logger {
    */
   error(msg) {
     this.log(msg);
+
+    // Rethrow to make sure it's auditable as an error.
+    setTimeout(_ => {
+      throw new Error(msg);
+    }, 0);
   }
 
   /**
--- a/lighthouse-core/report/html/renderer/logger.js
+++ b/lighthouse-core/report/html/renderer/logger.js
@@ -50,7 +50,8 @@ class Logger {
   error(msg) {
     this.log(msg);
 
-    // Rethrow to make sure it's auditable as an error.
+    // Rethrow to make sure it's auditable as an error, but in a setTimeout so page
+    // recovers gracefully and user can try loading a report again.
     setTimeout(_ => {
       throw new Error(msg);
     }, 0);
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/html/renderer/logger.js
```diff
@@ -49,6 +49,11 @@ class Logger {
    */
   error(msg) {
     this.log(msg);
+
+    // Rethrow to make sure it's auditable as an error.
+    setTimeout(_ => {
+      throw new Error(msg);
```
ebidel: Why the settimeout?

package.json
```diff
@@ -97,7 +97,7 @@
     "npm-run-posix-or-windows": "^2.0.2",
     "nyc": "^11.6.0",
     "postinstall-prepare": "^1.0.1",
-    "puppeteer": "^1.1.1",
+    "puppeteer": "1.4.0",
```
ebidel: Should the version be more flexible?

lighthouse-viewer/test/viewer-test-pptr.js
```diff
@@ -66,6 +66,10 @@ describe('Lighthouse Viewer', function() {
   });
 
   after(async function() {
+    // Log any page load errors encountered in case before() failed.
+    // eslint-disable-next-line no-console
+    console.log(pageErrors);
```
paulirish: console.error

lighthouse-core/report/html/renderer/logger.js
brendankenny: > Why the settimeout?

it's a hack that lets logger catch it so the report still acts nicely (with a toast) even if the report is invalid or whatever, but now it also async throws the error so that the problem can be picked up in the console and (more importantly) in the puppeteer `pageerror` listener we add in the test :)

brendankenny: added to the comment to clarify

lighthouse-viewer/test/viewer-test-pptr.js
brendankenny: > console.error

done

PR #5204, opened 2018-05-14
misc(viewer): support 3.x and legacy (2.x) reports in viewer

### LINKED ISSUE #5170: Viewer: support lhr v3.0
Currently errors out on the new lhr:

screen shot 2018-05-09 at 12 04 35 pm

paulirish:
#### Basic setup:

1. The 2.x viewer is published to https://googlechrome.github.io/lighthouse/viewer2x/
2. The 3.x viewer will live at https://googlechrome.github.io/lighthouse/viewer/ (but doesn't yet)
3. User opens report in 3.x viewer app (via however: drag, paste, click menu item)
4. The 3.x viewer app notices report isn't v3 and opens another tab with `viewer2x/` and has it render the report.
4. The 3.x viewer attempts to close it's own tab (but this doesn't always work (because user gestures?))
1. The 3.x viewer tells the user what's happening with some toaster notifications.

#### Screencast:
2xviewr

This also relies on a few commits added to 2.x branch: https://github.com/GoogleChrome/lighthouse/compare/branch2.x...viewer2x-deploy

fixes #5170

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -349,6 +349,7 @@ class ReportUIFeatures {
    * to the online viewer using postMessage.
    * @param {!ReportRenderer.ReportJSON} reportJson
    * @param {string} viewerPath
+   * @suppress {reportUnknownTypes}
    * @protected
    */
   static openTabAndSendJsonReport(reportJson, viewerPath) {
@@ -375,8 +376,8 @@ class ReportUIFeatures {
     });
 
     // The popup's window.name is keyed by version+url+fetchTime, so we reuse/select tabs correctly
-    const fetchTime = /** @type {!string} */ (json.fetchTime || json.generatedTime);
-    const windowName = `${json.lighthouseVersion}-${json.initialUrl}-${fetchTime}`;
+    const fetchTime = json.fetchTime || json.generatedTime;
+    const windowName = `${json.lighthouseVersion}-${json.requestedUrl}-${fetchTime}`;
     const popup = /** @type {!Window} */ (window.open(`${VIEWER_ORIGIN}${viewerPath}`, windowName));
 
     return p;
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -368,7 +368,10 @@ class ReportUIFeatures {
       onMessageCallback(messageEvent.data);
     });
 
-    const popup = /** @type {!Window} */ (window.open(`${VIEWER_ORIGIN}${viewerPath}`, 'zblank'));
+    // The popup's window.name is keyed by version+url+fetchTime, so we reuse/select tabs correctly
+    const fetchTime = json.fetchTime || json.generatedTime;
+    const windowName = `${json.lighthouseVersion}-${json.initialUrl}-${fetchTime}`;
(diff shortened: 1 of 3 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -344,30 +347,31 @@ class ReportUIFeatures {
...
+    const json = reportJson;
     window.addEventListener('message', function msgHandler(/** @type {!Event} */ e) {
       const messageEvent = /** @type {!MessageEvent<{opened: boolean}>} */ (e);
       if (messageEvent.origin !== VIEWER_ORIGIN) {
         return;
       }
-
       if (messageEvent.data.opened) {
         popup.postMessage({lhresults: json}, VIEWER_ORIGIN);
-        window.removeEventListener('message', msgHandler);
```
patrickhulce: Is this because we need to listen for the render message to close? will this cause duplicate handler problems if they click "Open in viewer" twice?

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -164,12 +169,30 @@ class LighthouseReportViewer {
...
+   * @private
+   */
+  _loadInLegacyViewerVersion(json) {
+    const warnMsg = `Version mismatch between viewer and JSON.
+    Opening new tab with compatible viewer.`;
+    const features = new ViewerUIFeatures(new DOM(document));
+    const viewerPath = '/lighthouse/viewer2x/';
+
+    logger.log(warnMsg, false);
+    features.sendJsonReport(json, viewerPath, onMessage);
```
patrickhulce: impl is a bit confusing that we create a temporary instance of the ViewerUIfeatures to trigger the render in viewer, feels a bit hacky

I think it's mostly just the fake flexibility of `sendJsonReport` that confuses me. WDYT of making this a bit more explicit with a `sendJsonReportToV2` or something that both sure the same underlying logic?

they could just resolve a promise when they get confirmation that the report has been rendered and nix the onMessage piece

lighthouse-core/report/html/renderer/report-ui-features.js
paulirish: good question. @ebidel is that the situation this is handling?

brendankenny: I remember one issue is if you open viewer from a report, then use viewer as a viewer (looking at other reports or something), if you refresh it will send back an `opened` message back to the original report and load the first report, not whatever you've been doing since.

ebidel: I think what @brendankenny described is the sitch. Although I'm not sure why Chrome does that, but you sometimes get another `message` event when the page refreshes.

paulirish:
> If we really want to do it right it feels like we should be bundling the report generator, loading the v3 bundle by default, and then loading the v2 bundle if the report is older. Then it should be pretty isolated to _replaceReportHtml(), where it'll just have to be calls to new bundle.DOM(), new bundle.ReportRenderer, and new bundle.ViewerUIFeatures() (or pull the init code into the bundle) and everything else is pretty much just saving/loading reports to pass into the versioned bundle

I tried this but it's real nasty. See https://github.com/GoogleChrome/lighthouse/compare/viewer2x-open-versioned-samedpage and https://github.com/GoogleChrome/lighthouse/compare/branch2.x...viewer2x-deploy-versioned

removeEvtListener restored. 

we good here?

PR #5205, opened 2018-05-14
misc(none): removed extra comma to keep the json valid

xRahul:
I executed a validator on the project and found this extra comma in the json file.
It shouldn't be here.


patrickhulce:
Thanks for the PR @xRahul! Is there value in not having the trailing comma? 

The tsconfig still has js comments, so it's still not real JSON, and TS doesn't seem to care either way. Seems like we'd favor consistency with the rest of the codebase if there's not a feature we're unlocking here.

xRahul:
It's cool either way. It's just my IDE was throwing invalid json for the trailing comma, and I thought it would be better to remove it.

PR #5207, opened 2018-05-15
report: avoid paint storms on scrolly header.

paulirish:
And animate the clouds on hover

Here's the diff on the paint storms. Before and after:

layerslh

Check out the deployment to try the animated clouds.

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -730,7 +730,7 @@
   justify-content: center;
   overflow-x: hidden;
   position: relative;
-  padding-bottom: calc(var(--section-indent) * 2);
+  padding: var(--section-indent) calc(var(--section-indent) / 2) calc(var(--section-indent) * 2);
 }
 .lh-scores-header__solo {
   padding: 0;
@@ -739,9 +739,10 @@
 
 .lh-scorescale {
   color: var(--medium-75-gray);
-  padding: 0 var(--section-indent) 0 0;
+  padding: 0 calc(var(--section-indent) * 1.5) 0;
   text-align: right;
   transform-origin: bottom right;
+  will-change: opacity; /* opacity is changed on scroll */
 }
 
 .lh-scorescale-range {
--- a/lighthouse-core/report/html/templates.html
+++ b/lighthouse-core/report/html/templates.html
@@ -331,9 +331,72 @@
         margin-left: 0;
       }
     }
+    .lh-lighthouse__clouds {
+      animation: panacross 15s linear infinite;
+      animation-play-state: paused;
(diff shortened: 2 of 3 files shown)


lighthouse-core/report/html/templates.html
```diff
@@ -325,9 +331,72 @@
...
+            <path d="M107 22a10 10 0 0 1 10 10v1H97v-1a10 10 0 0 1 10-10zM96.77 66.23l20.97-10.7.63 11.87-22.28 11.87zM95 94.8L119.1 82l.9 14H95z"
+              fill="#EC5548" />
+            <path d="M0 132a177.07 177.07 0 0 1 108.5-37c40.78 0 78.38 13.78 108.5 37H0z" fill="#E8EAED" />
+            <rect fill="#FEF0C8" x="98" y="33" width="10" height="10" rx="5" />
+            <path fill="url(#c)" d="M7 0l91 33.18v10.05L7 77z" />
+          </g>
+        </g>
+        <g mask="url(#b)" fill="#FFF">
+          <g transform="translate(96 86)">
+            <circle cx="87.5" cy=".5" r="1" />
```
patrickhulce: these are the little stars right? should we just remove them? they look a little weird, deleting this whole block looks better IMO

| before | after |
| - | - |
| image | image
 |

lighthouse-core/report/html/report-styles.css
```diff
@@ -739,9 +739,10 @@
 
 .lh-scorescale {
   color: var(--medium-75-gray);
-  padding: 0 var(--section-indent) 0 0;
+  padding: 0 calc(var(--section-indent) * 1.5) 0;
   text-align: right;
   transform-origin: bottom right;
+  will-change: opacity; /* opacity is changed on scroll */
```
patrickhulce: see the score scales look fine with this now, how are you supposed to tell when the text will be messed up by this?

lighthouse-core/report/html/templates.html
```diff
@@ -325,9 +331,72 @@
         margin-left: 0;
       }
     }
+    .lh-lighthouse__clouds {
+      animation: panacross 30s linear infinite;
+      animation-play-state: paused;
+    }
+    .lh-header-sticky:hover .lh-lighthouse__clouds {
```
patrickhulce: the `:hover` threw me for a while since the hitbox includes the dropdown of the scores, how much do you like the hover effect vs. always? :)

brendankenny: I'm personally ok with always, too :)

Something something lighthouse is always looking out for you metaphor

brendankenny:
whats the deal

PR #5221, opened 2018-05-15
core(lantern): handle disk cache simulation

### LINKED ISSUE #5193: Lantern: work with warm loads
right now Lantern simulations assume cold cache, but some work likely needs to be done to get it to work as expected with warm loads/no cache clearing

necessary for allowing `--disable-storage-reset` to work seamlessly under lantern

patrickhulce:
adds support to lantern for handling network files that were loaded from disk cache instead of network (necessary for working with `--disable-storage-reset` and in DevTools)

this also makes an adjustment to all lantern metrics predicted to be <1s that scales the intercept linearly until 1s. It only affects sites that were predicted to be very fast.

for example a site predicted to have FCP at 500ms
Before: 600 + 500 = 1100
After: 500/1000 * 600 + 500 = 800

for example a site predicted to have FCP at 10ms (for example cached example.com)
Before: 600 + 10 = 610
After: 10/1000 * 600 + 10 = 16

lastly, it adjusts speed index pessimistic estimate to use the pessimistic FCP as the floor rather than the coefficient predicted one. It was similarly unfair to very fast sites to double apply the coefficient in this case.

closes #5193

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/simulator.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/simulator.js
@@ -213,23 +213,38 @@ class Simulator {
    */
   _estimateTimeRemaining(node) {
     if (node.type === Node.TYPES.CPU) {
-      const timingData = this._nodeTimings.get(node);
-      const multiplier = /** @type {CpuNode} */ (node).didPerformLayout()
-        ? this._layoutTaskMultiplier
-        : this._cpuSlowdownMultiplier;
-      const totalDuration = Math.min(
-        Math.round(/** @type {CpuNode} */ (node).event.dur / 1000 * multiplier),
-        DEFAULT_MAXIMUM_CPU_TASK_DURATION
-      );
-      const estimatedTimeElapsed = totalDuration - timingData.timeElapsed;
-      this._setTimingData(node, {estimatedTimeElapsed});
-      return estimatedTimeElapsed;
+      return this._estimateCPUTimeRemaining(/** @type {CpuNode} */ (node));
+    } else if (node.type === Node.TYPES.NETWORK) {
+      return this._estimateNetworkTimeRemaining(/** @type {NetworkNode} */ (node));
+    } else {
+      throw new Error('Unsupported');
     }
+  }
 
-    if (node.type !== Node.TYPES.NETWORK) throw new Error('Unsupported');
+  /**
+   * @param {CpuNode} cpuNode
+   * @return {number}
+   */
+  _estimateCPUTimeRemaining(cpuNode) {
+    const timingData = this._nodeTimings.get(cpuNode);
+    const multiplier = cpuNode.didPerformLayout()
+      ? this._layoutTaskMultiplier
+      : this._cpuSlowdownMultiplier;
+    const totalDuration = Math.min(
+      Math.round(cpuNode.event.dur / 1000 * multiplier),
+      DEFAULT_MAXIMUM_CPU_TASK_DURATION
+    );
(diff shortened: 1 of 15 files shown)


lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -224,16 +228,27 @@ class Simulator {
...
-      {timeAlreadyElapsed: timingData.timeElapsed, maximumTimeToElapse: Infinity}
-    );
 
-    const estimatedTimeElapsed = calculation.timeElapsed + timingData.timeElapsedOvershoot;
+    let timeElapsed = 0;
+    if (networkNode.fromDiskCache) {
+      // Rough access time for seeking to location on disk and reading sequentially
+      // @see http://norvig.com/21-days.html#answers
+      const sizeInMb = (networkNode.record._resourceSize || 0) / 1024 / 1024;
+      timeElapsed = 8 + 20 * sizeInMb - timingData.timeElapsed;
```
paulirish: I think this is slower in Chrome. UMA must have some timings we could look at.

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -224,16 +228,27 @@ class Simulator {
...
-    // If we're estimating time remaining, we already acquired a connection for this record, definitely non-null
-    const connection = /** @type {TcpConnection} */ (this._acquireConnection(record));
-    const calculation = connection.simulateDownloadUntil(
-      record.transferSize - timingData.bytesDownloaded,
-      {timeAlreadyElapsed: timingData.timeElapsed, maximumTimeToElapse: Infinity}
-    );
 
-    const estimatedTimeElapsed = calculation.timeElapsed + timingData.timeElapsedOvershoot;
+    let timeElapsed = 0;
+    if (networkNode.fromDiskCache) {
```
paulirish: Given the new logic I think we should spilt `_estimateTimeRemaining` into a few pieces -- it's gettin a little long.

patrickhulce: aye, prefer that happen now? wanted to keep the diff a little simpler though I suppose with separate commits that's still the case

lighthouse-core/lib/dependency-graph/simulator/simulator.js
patrickhulce: yeah we're not really accounting for the full Chrome IPC RT time, just the OS/disk access, but the same could be said for network really

perhaps we add an IPC RT latency factor in the future where we account for more based on the chunk size of all IPC-based traffic?

lighthouse-core/lib/dependency-graph/simulator/simulator.js
paulirish: ya let's do it now. i can follow these changes pretty ez.

lighthouse-core/lib/dependency-graph/simulator/simulator.js
paulirish: okay yah sg. `DiskCache.0.AsyncIOTime` is the best one but it doesn't give us much as its measured from the cache thread and no cost to browser IO thread or over the IPC to the renderer.

patrickhulce:
> presumably you're seeing better accuracy for your trace corpus?

yeah there's really negligible impact here though since all the traces were collated with LH (and thus had the cache entirely disabled)

the only thing impacting accuracy is the SI change which was relatively minor since it's just the floor component and the intercept scaling adjustment which only impacts sites with estimates <1s (slight improvement in accuracy to those sites)

PR #5227, opened 2018-05-16
docs(readme): add lighthouse-lambda to related projects

baocancode:
Hi, we made this library to run Lighthouse on AWS Lambda easily. Headless Chrome (stable desktop releases) is built by https://github.com/joytocode/headless-chrome-builder and tested to ensure working with Lighthouse on Lambda. It would be nice if you could add it to related projects.


baocancode:
I signed it!

baocancode:
Thanks @patrickhulce, we will integrate with v3 when the docs are ready.

PR #5233, opened 2018-05-16
core(pwa): adjust score weights

patrickhulce:
adjust the PWA score weights to give more credit to speed/offline, less to adding a manifest/1-line meta

You can play around with how this affects score at [the calculator](https://docs.google.com/spreadsheets/d/1Cxzhy5ecqJCucdf1M0iOzM8mIxNc7mmx107o5nj38Eo/edit#gid=283330180)

Basically... 

Adding manifest alone used to be worth 18, now worth 8
Adding noop service worker used to be worth 9, now worth 4
Adding service worker to make site work offline used to be worth 18, now worth 23
Performance used to be worth 9, now worth 27

### DIFF
--- a/lighthouse-core/config/default-config.js
+++ b/lighthouse-core/config/default-config.js
@@ -289,16 +289,20 @@ module.exports = {
           '[PWA Checklist](https://developers.google.com/web/progressive-web-apps/checklist) but are ' +
           'not automatically checked by Lighthouse. They do not affect your score but it\'s important that you verify them manually.',
       auditRefs: [
+        // Most difficult and critical for good UX
+        {id: 'works-offline', weight: 5},
+        {id: 'load-fast-enough-for-pwa', weight: 5},
+        // Encompasses most of the other checks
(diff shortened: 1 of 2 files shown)


paulirish:
cc @tjsavage

patrickhulce:
> still difficult for established ones with legacy content/serving/ad systems

sure but a very different kind of difficult, the technical challenge of getting a cert and flipping a server bit is still "not too difficult", wrangling all your third parties and getting all of them to go https is an ecosystem challenge. one could make the claim that making **any** changes to a legacy site can be supremely challenging, so might as well not classify them at all 🤷‍♀️

patrickhulce:
@vinamratasingal thoughts?

brendankenny:
so it seems like PWA badging or whatever is independent of these changes. The score re-weighting seems good and should definitely land before the full 3.0 release, so should we move forward with this?

paulirish:
yes and yes 👍

patrickhulce:
SGTM 👍

PR #5256, opened 2018-05-17
core(network-recorder): handle QUIC requests

### LINKED ISSUE #5254: H2/QUIC requests don't fire RequestFinished
while investigating #5249, I noticed that the page always timed out while waiting for load because of 3 network requests to a Google origin that were using QUIC/H2, it seems that they never "finish" in the traditional sense and that the end of the request is simply the receivedHeadersEnd.

URL: https://www.guidetoonlineschools.com/online-schools
QUIC requests: https://apis.google.com/se/0/_/+1/fastbutton?usegapi=1&size=medium&origin=https%3A%2F%2Fwww.guidetoonlineschools.com&url=https%3A%2F%2Fwww.guidetoonlineschools.com%2Fonline-schools&gsrc=3p&ic=1&jsh=m%3B%2F_%2Fscs%2Fapps-static%2F_%2Fjs%2Fk%3Doz.gapi.en_US.YrtA5yEj3Q0.O%2Fm%3D__features__%2Fam%3DQQE%2Frt%3Dj%2Fd%3D1%2Frs%3DAGLTcCN0sMuxpka3R31qPm4lRzkEwQAp8Q

patrickhulce:
closes #5254

### DIFF
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -107,9 +107,15 @@ class NetworkRecorder extends EventEmitter {
         return;
       }
 
+      // QUIC network requests don't always "finish" even when they're done loading data
+      // Use receivedHeaders instead, see https://github.com/GoogleChrome/lighthouse/issues/5254
+      const isQUIC = record._responseHeaders && record._responseHeaders
+          .find(header => header.name.toLowerCase() === 'alt-svc' && /quic/.test(header.value));
+      const receivedHeaders = record._timing && record._timing.receiveHeadersEnd > 0;
+
       // convert the network record timestamp to ms
       timeBoundaries.push({time: record.startTime * 1000, isStart: true});
-      if (record.finished) {
+      if (record.finished || (isQUIC && receivedHeaders && record.endTime)) {
         timeBoundaries.push({time: record.endTime * 1000, isStart: false});
       }
     });
@@ -119,7 +125,7 @@ class NetworkRecorder extends EventEmitter {
       .sort((a, b) => a.time - b.time);
 
     let numInflightRequests = 0;
-    let quietPeriodStart = 0;
+    let quietPeriodStart = -Infinity;
     /** @type {Array<{start: number, end: number}>} */
     const quietPeriods = [];
     timeBoundaries.forEach(boundary => {
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -90,6 +90,19 @@ class NetworkRecorder extends EventEmitter {
   }
 
   /**
+   * QUIC network requests don't always "finish" even when they're done loading data, use recievedHeaders
+   * @see https://github.com/GoogleChrome/lighthouse/issues/5254
+   * @param {LH.WebInspector.NetworkRequest} record
+   * @return {boolean}
+   */
+  static _isQUICAndFinished(record) {
+    const isQUIC = record._responseHeaders && record._responseHeaders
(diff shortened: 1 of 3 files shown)


lighthouse-core/lib/network-recorder.js
```diff
@@ -107,9 +107,15 @@ class NetworkRecorder extends EventEmitter {
         return;
       }
 
+      // QUIC network requests don't always "finish" even when they're done loading data
+      // Use receivedHeaders instead, see https://github.com/GoogleChrome/lighthouse/issues/5254
+      const isQUIC = record._responseHeaders && record._responseHeaders
```
brendankenny: would be nice pulled out into a function to separate determining if request is QUIC vs what to do with QUIC requests

lighthouse-core/lib/network-recorder.js
```diff
@@ -107,9 +107,15 @@ class NetworkRecorder extends EventEmitter {
...
+          .find(header => header.name.toLowerCase() === 'alt-svc' && /quic/.test(header.value));
```
brendankenny: how certain is the `'alt-svc'` check? Any other way to tell?

lighthouse-core/lib/network-recorder.js
brendankenny: it could also be a function that's like `isQuicAndFinished(record)` to absorb logic below, too :)

lighthouse-core/lib/network-recorder.js
```diff
@@ -119,7 +125,7 @@ class NetworkRecorder extends EventEmitter {
       .sort((a, b) => a.time - b.time);
 
     let numInflightRequests = 0;
-    let quietPeriodStart = 0;
+    let quietPeriodStart = -Infinity;
```
brendankenny: what's the advantage of a [-Infinity, 0] period? Time starting at 0 also seems like a reasonable state of things

brendankenny:
is this something we should also fix in the upstream `NetworkManager` so these are marked as finished? Or is it actually finished and we just haven't updated our devtools frontend in so long...

patrickhulce:
> is this something we should also fix in the upstream NetworkManager so these are marked as finished? Or is it actually fixed and we just haven't updated our devtools frontend in so long...

DevTools frontend handles this case correctly, I didn't dive into the code there to see how since its ~2 years or so diff :)

lighthouse-core/lib/network-recorder.js
patrickhulce: that's fine just made tests have [0, 0] quiet period so we either handle the empty period case, move test cases to past 0, or -Infinity, -Infinity seemed like most reasonable conceptually but I do not have strong feelings

lighthouse-core/lib/network-recorder.js
patrickhulce: seems to be fairly certain, in the spec it's not technically required to announce it buuuuuut...

> While HTTP/QUIC doesn't formally require a client to implement Alt-Svc, there's no discovery mechanism other than Alt-Svc provided, so you're not going to get very far without 
it. 

from https://github.com/quicwg/base-drafts/issues/253

lighthouse-core/lib/network-recorder.js
patrickhulce: sure

lighthouse-core/lib/network-recorder.js
brendankenny: it may be fine, and I haven't spent enough time here recently to have a good gut instinct, but I think I would expect the earliest quiet period to start at 0.

I guess I don't understand what this is addressing. Seems independent of the QUIC change?

PR #5258, opened 2018-05-18
core(tsc): add type checking to viewer

brendankenny:
fully replaces previous closure type checking, which was also checking the viewer files with the report ones. Should help catch things like those fixed in #5251 and the `Viewer._replaceReportHTML()` call (which should be a camel-cased `Html`) fixed in this PR.

These types play a little loose with the values coming out of firebase, `idb-keyval`, and the responses from github, but that can be improved someday with better `idb-keyval` types and if we could switch to async/await (tsc really gets confused in the gist-fetching code). However the code interacting with LHRs and the DOM is still solid and checked, which is the important improvement.


brendankenny:
oh, I added a separate `tsconfig.json` file for viewer. It made sense not to pollute the normal lighthouse scope with all these global antics in `viewer.d.ts`.

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -40,30 +58,40 @@ class LighthouseReportViewer {
    * @private
    */
   _addEventListeners() {
+    // @ts-ignore - tsc thinks document can't listen for `copy`
```
patrickhulce: can't listen for `paste` ;)

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -40,30 +58,40 @@ class LighthouseReportViewer {
...
+    const gistUrlInput = find('.js-gist-url', document);
     gistUrlInput.addEventListener('change', this._onUrlInputChange);
 
     // Hidden file input to trigger manual file selector.
-    const fileInput = document.querySelector('#hidden-file-input');
+    const fileInput = find('#hidden-file-input', document);
     fileInput.addEventListener('change', e => {
-      this._onFileLoad(e.target.files[0]);
-      e.target.value = null;
+      if (!e.target) {
```
patrickhulce: wow yeah no kidding this is a pain

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -254,7 +283,7 @@ class LighthouseReportViewer {
     try {
       const json = JSON.parse(e.clipboardData.getData('text'));
       this._reportIsFromGist = false;
-      this._replaceReportHTML(json);
```
patrickhulce: lol, so this was just broken?

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -285,12 +317,12 @@ class LighthouseReportViewer {
   /**
    * Loads report json from gist URL, if valid. Updates page URL with gist ID
    * and loads from github.
-   * @param {string} url Gist URL.
+   * @param {string} urlStr Gist URL.
    * @private
    */
-  _loadFromGistURL(url) {
+  _loadFromGistURL(urlStr) {
```
patrickhulce: yay 🇭🇺!

lighthouse-viewer/package.json
```diff
@@ -28,7 +33,7 @@
     "run-sequence": "^1.1.5",
     "streamqueue": "^1.1.1",
     "tsify": "^2.0.3",
-    "typescript": "^2.1.4",
+    "typescript": "2.9.1-insiders.20180516",
```
patrickhulce: does it need it's own, can we share with root?

lighthouse-viewer/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
```
patrickhulce: should we extend our base or not worth it cause we're just checking the src here?

lighthouse-viewer/package.json
brendankenny: > does it need it's own, can we share with root?

have to do `"../node_modules/.bin/tsc -p ."`, but that's not so bad (and probably easier than remembering to keep them in sync). Done

lighthouse-viewer/app/src/lighthouse-report-viewer.js
brendankenny: done

lighthouse-viewer/app/src/lighthouse-report-viewer.js
brendankenny: > lol, so this was just broken?

ha, yes. No idea for how long...given that it is (or appears to be) a change from `replaceReportHTML` -> `replaceReportHtml`, I would give good odds it was my fault when converting to 2.0 last year...

lighthouse-viewer/tsconfig.json
brendankenny: > should we extend our base or not worth it cause we're just checking the src here?

it doesn't buy us a whole lot because we have to overwrite `typeRoots` and `include`, but it's kind of nice to share the other options. Done.

PR #5267, opened 2018-05-18
report: restore missing non-applicable icon

paulirish:
This icon was missing because of all the rebasing:

image


lighthouse-core/report/html/report-styles.css
```diff
@@ -325,10 +325,6 @@
   padding: 11px var(--text-indent);
 }
 
-.lh-metric__header {
```
paulirish: driveby removal of this unused selector

PR #5270, opened 2018-05-19
report: audit warnings are not top-level

### LINKED ISSUE #5153: Redesign displayValue/warnings
@brendankenny expressed some concerns over our displayValue/warnings display methods over in https://github.com/GoogleChrome/lighthouse/pull/5132#pullrequestreview-118109533

two things

### 1. visually associating of displayValue + audit title

> A few major things for how we handle these different outputs:

> It's great to start using displayValue for things that are display values, but it does expose a weakness with how we display them. It's definitely harder to scan across to the value for the audit vs just having it right below

old:
screen shot 2018-05-07 at 12 31 10

new:
screen shot 2018-05-07 at 12 31 18
(and it's even wider across when not shrunk by github)

### 2. warnings at the top of the report

> Separating the warnings from the audits seems like a regression from the status quo in most of the warnings cases added here. Having a list of images we can't reencode at the top of the report or (in the case of sample_v2.json) a warning that some target=_blank anchors may be outgoing but we can't tell, seems like noise at the top of the report and is lost as valuable context at the individual audit level (where you probably want to be warned that the audit may have been unable to comprehensively run its test).

paulirish:
They're now reported along with the audit.

Examples:

image
image
image

You can see one example of this in the deployment. (But these other warning cases are pretty esoteric so i've created an artifact bundle that reproduces them.)

fixes  #5153

### DIFF
--- a/lighthouse-core/audits/viewport.js
+++ b/lighthouse-core/audits/viewport.js
@@ -40,10 +40,10 @@ class Viewport extends Audit {
     const parsedProps = Parser.parseMetaViewPortContent(artifacts.Viewport);
 
     if (Object.keys(parsedProps.unknownProperties).length) {
-      warnings.push(`Invalid properties found: ${JSON.stringify(parsedProps.unknownProperties)}.`);
+      warnings.push(`Invalid properties found: ${JSON.stringify(parsedProps.unknownProperties)}`);
     }
     if (Object.keys(parsedProps.invalidValues).length) {
-      warnings.push(`Invalid values found: ${JSON.stringify(parsedProps.invalidValues)}.`);
+      warnings.push(`Invalid values found: ${JSON.stringify(parsedProps.invalidValues)}`);
     }
 
     const viewportProps = parsedProps.validProperties;
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -85,8 +85,12 @@ class CategoryRenderer {
       const tooltip = this.dom.createChildOf(textEl, 'div', 'tooltip lh-debug');
       tooltip.textContent = audit.result.errorMessage || 'Report error: no audit information';
     } else if (audit.result.explanation) {
-      const explanationEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug');
-      explanationEl.textContent = audit.result.explanation;
+      const explEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--explanation');
+      explEl.textContent = audit.result.explanation;
+    }
+    if (audit.result.warnings && audit.result.warnings.length > 0) {
+      const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--warnings');
+      warningsEl.textContent = 'Warnings: ' + audit.result.warnings.join(', ');
     }
     return auditEl;
   }
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -88,9 +88,20 @@ class CategoryRenderer {
       const explEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--explanation');
       explEl.textContent = audit.result.explanation;
     }
-    if (audit.result.warnings && audit.result.warnings.length > 0) {
-      const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--warnings');
-      warningsEl.textContent = 'Warnings: ' + audit.result.warnings.join(', ');
+    const warnings = audit.result.warnings;
+    if (!warnings || warnings.length === 0) return auditEl;
+
+    // Add list of warnings or singular warning
+    const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--warnings');
+    if (warnings.length === 1) {
+      warningsEl.textContent = `Warning: ${warnings.join('')}`;
+    } else {
+      warningsEl.textContent = 'Warnings: ';
+      const warningsUl = this.dom.createChildOf(warningsEl, 'ul');
+      for (const warning of warnings) {
+        const item = this.dom.createChildOf(warningsUl, 'li');
+        item.textContent = warning;
+      }
     }
     return auditEl;
   }
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -90,7 +90,7 @@ class CategoryRenderer {
     if (!warnings || warnings.length === 0) return auditEl;
 
     // Add list of warnings or singular warning
-    const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--warnings');
+    const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-warnings');
     if (warnings.length === 1) {
       warningsEl.textContent = `Warning: ${warnings.join('')}`;
     } else {
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -60,10 +60,8 @@ class CategoryRenderer {
 
     const titleEl = this.dom.find('.lh-audit__title', auditEl);
     titleEl.appendChild(this.dom.convertMarkdownCodeSnippets(audit.result.title));
-    if (audit.result.description) {
-      this.dom.find('.lh-audit__description', auditEl)
+    this.dom.find('.lh-audit__description', auditEl)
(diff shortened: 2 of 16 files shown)


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -85,8 +85,23 @@ class CategoryRenderer {
...
-      const explanationEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug');
-      explanationEl.textContent = audit.result.explanation;
+      const explEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--explanation');
+      explEl.textContent = audit.result.explanation;
+    }
+    const warnings = audit.result.warnings;
+    if (!warnings || warnings.length === 0) return auditEl;
+
+    // Add list of warnings or singular warning
+    const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-debug lh-debug--warnings');
```
patrickhulce: do we have any cases of warning + explanation? if so, does it still look okish :)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -114,23 +125,6 @@ describe('CategoryRenderer', () => {
     category.description = prevDesc;
   });
 
-  // TODO(phulce): revisit if top-level warnings approach is too noisy
-  it.skip('renders audits with warnings as failed', () => {
```
patrickhulce: do we want to render an audit with warnings as failed? or keep it in passed?

lighthouse-core/test/report/html/renderer/report-renderer-test.js
```diff
@@ -135,8 +135,10 @@ describe('ReportRenderer', () => {
       const output = renderer.renderReport(sampleResults, container);
 
       const warningEls = output.querySelectorAll('.lh-run-warnings > ul > li');
-      assert.strictEqual(warningEls.length, 1);
-      assert.ok(/Links.*unsafe/.test(warningEls[0].textContent), 'did not add warning text');
+      assert.strictEqual(warningEls.length, sampleResults.runWarnings.length);
+      for (const warningEl of warningEls) {
+        assert.ok(/Links.*unsafe/.test(warningEl.textContent), 'did not add warning text');
```
patrickhulce: shouldn't this just be deleted now that warningEls should be length 0?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -33,7 +33,9 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     valueEl.textContent = Util.formatDisplayValue(audit.result.displayValue);
 
     const descriptionEl = this.dom.find('.lh-metric__description', tmpl);
-    descriptionEl.appendChild(this.dom.convertMarkdownLinkSnippets(audit.result.description));
+    if (audit.result.description) {
```
patrickhulce: a little confused by removing the `?` from `description` but making this conditional... what's up with that?

lighthouse-core/test/report/html/renderer/category-renderer-test.js
paulirish: good policy question!

Let's see. here's the possible warnings we have right now (that are placed into Passed)

image
image
image

I see the argument for these being visible by default, however in the case of timing/decode, there's not much a user can do about them. Either way, to be visible by default, they'd have to be placed in "Failed" and either we keep the green checkmark or fail them. Both are pretty weird.

So I'm comfy with saying that warnings are orthogonal to passing/failing and default report visibility. wdyt?

lighthouse-core/test/report/html/renderer/report-renderer-test.js
paulirish: good call.

lighthouse-core/report/html/renderer/performance-category-renderer.js
paulirish: good catch. it's assumed to be there now.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
patrickhulce: Yeah I don't think there's much anyone can do about them either, I'm fine to hide especially now that we've got Sentry tracking 'em for Httparchive runs 👍

lighthouse-core/audits/viewport.js
```diff
@@ -40,10 +40,10 @@ class Viewport extends Audit {
     const parsedProps = Parser.parseMetaViewPortContent(artifacts.Viewport);
 
     if (Object.keys(parsedProps.unknownProperties).length) {
-      warnings.push(`Invalid properties found: ${JSON.stringify(parsedProps.unknownProperties)}.`);
+      warnings.push(`Invalid properties found: ${JSON.stringify(parsedProps.unknownProperties)}`);
     }
     if (Object.keys(parsedProps.invalidValues).length) {
-      warnings.push(`Invalid values found: ${JSON.stringify(parsedProps.invalidValues)}.`);
+      warnings.push(`Invalid values found: ${JSON.stringify(parsedProps.invalidValues)}`);
```
brendankenny: tests for these will need to be updated (they assert a `.` in there :)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -87,6 +87,17 @@ describe('CategoryRenderer', () => {
...
 
+  it('renders audits with warnings', () => {
+    const auditResult = {
+      title: 'Audit',
+      description: 'Learn more',
+      warnings: ['It may not have worked!'],
+      score: 1,
+    };
+    const auditDOM = renderer.renderAudit({id: 'foo', score: 1, result: auditResult});
+    assert.ok(auditDOM.querySelector('.lh-debug--warnings'), 'did not render debug message');
```
brendankenny: want to test at least that the warning text is in there as well? And probably want a plural/bulleted test in addition

lighthouse-core/test/report/html/renderer/report-renderer-test.js
```diff
@@ -135,8 +135,7 @@ describe('ReportRenderer', () => {
       const output = renderer.renderReport(sampleResults, container);
 
       const warningEls = output.querySelectorAll('.lh-run-warnings > ul > li');
-      assert.strictEqual(warningEls.length, 1);
-      assert.ok(/Links.*unsafe/.test(warningEls[0].textContent), 'did not add warning text');
+      assert.strictEqual(warningEls.length, sampleResults.runWarnings.length);
```
brendankenny: there's already a "no `lighthouseRunWarnings`" test above this, so probably want to make a `const warningResults = Object.assign({}, sampleResults, {runWarnings: ['Warnings are great!']});` or whatever and pass that in

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -194,29 +194,13 @@ if (typeof module !== 'undefined' && module.exports) {
...
- */
-
 /**
  * @typedef {{
       id: string,
       score: (number|null),
       weight: number,
       group?: string,
-      result: AuditResultJSON
+      result: LH.Audit.Result
```
brendankenny: 👍

lighthouse-core/test/report/html/renderer/category-renderer-test.js
rviscomi: My suggestion in https://github.com/GoogleChrome/lighthouse/issues/5273#issuecomment-390777039 was to use warnings as a way to get attention to an issue without penalizing the score. So I would definitely prefer for the warning to be visible by default (not collapsed). I understand that warnings may also be used for things outside of the developer's control so collapsing may make more sense. Is this something that can be configured by the audit at runtime, rather than forcing all warnings to behave the same?

I'd also love to see different UI treatment for warnings; not a green check and not a red exclamation. Could it be a yellow/orange exclamation to distinguish it that something's amiss?

hwikyounglee: Here're some ideas to visually distinguish the orthogonal warnings from audits.

inlinewarning-ideas

paulirish: Sg. broke this into it's own issue: https://github.com/GoogleChrome/lighthouse/issues/5327 ("make sure warnings within Passed Audits are visible")

paulirish:
@brendankenny PTAL
@hwikyounglee does this look right to you?

here's a report with a bunch of errors: http://lh-warningsredux0.surge.sh/errors.html Open up Perf > Passed to see more warnings.

FYI Explanations & warnings are now the same font-size as audit title text:
image

brendankenny:
> FYI Explanations & warnings are now the same font-size as audit title text:

this is a little hard to read now to me, but fine with whatever is decided. It looks like in the earlier screenshots there were still bullets and indentation, though?

brendankenny:
oh, that one was a single warning, not a multi-warning

PR #5271, opened 2018-05-19
core: support traces with TracingStartedInBrowser event

a1ph:
Since version 67 Chrome uses a new schema to describe page frame tree.
The tree is encoded within TracingStartedInBrowser trace event.
The patch makes lighthouse to support the new event.
Legacy TracingStartedInPage event is going to the removed soon.

### DIFF
--- a/lighthouse-core/gather/computed/trace-of-tab.js
+++ b/lighthouse-core/gather/computed/trace-of-tab.js
@@ -42,11 +42,12 @@ class TraceOfTab extends ComputedArtifact {
     // *must* be stable to keep events correctly nested.
     /** @type Array<LH.TraceEvent> */
     const keyEvents = trace.traceEvents
-      .filter(e =>
-          e.cat.includes('blink.user_timing') ||
+      .filter(e => {
+        return e.cat.includes('blink.user_timing') ||
           e.cat.includes('loading') ||
           e.cat.includes('devtools.timeline') ||
-          e.cat === '__metadata')
+          e.cat === '__metadata'
+      })
       // @ts-ignore - stableSort added to Array by WebInspector.
       .stableSort((event0, event1) => event0.ts - event1.ts);
 
@@ -54,14 +55,13 @@ class TraceOfTab extends ComputedArtifact {
     /** @type {LH.TraceEvent|undefined} */
     let startedInPageEvt;
     const startedInBrowserEvt = keyEvents.find(e => e.name === 'TracingStartedInBrowser');
+    // `persistentIds` is a signal that the frame data will be attached, prefer it when available.
     if (startedInBrowserEvt && startedInBrowserEvt.args.data &&
         startedInBrowserEvt.args.data.persistentIds) {
       const mainFrame = (startedInBrowserEvt.args.data.frames || []).find(frame => !frame.parent);
       const pid = mainFrame && mainFrame.processId;
       const threadNameEvt = keyEvents.find(e => e.pid === pid && e.ph === 'M' &&
-        e.cat === '__metadata' && e.name === 'thread_name' &&
-        // @ts-ignore - property chain exists for 'thread_name' event.
-        e.args.name === 'CrRendererMain');
+        e.cat === '__metadata' && e.name === 'thread_name' && e.args.name === 'CrRendererMain');
       startedInPageEvt = mainFrame && threadNameEvt ?
         Object.assign({}, startedInBrowserEvt, {
           pid, tid: threadNameEvt.tid, name: 'TracingStartedInPage',
--- a/lighthouse-core/gather/computed/trace-of-tab.js
+++ b/lighthouse-core/gather/computed/trace-of-tab.js
@@ -44,9 +44,9 @@ class TraceOfTab extends ComputedArtifact {
     const keyEvents = trace.traceEvents
       .filter(e => {
         return e.cat.includes('blink.user_timing') ||
-          e.cat.includes('loading') ||
-          e.cat.includes('devtools.timeline') ||
-          e.cat === '__metadata';
+            e.cat.includes('loading') ||
+            e.cat.includes('devtools.timeline') ||
+            e.cat === '__metadata';
       })
       // @ts-ignore - stableSort added to Array by WebInspector.
       .stableSort((event0, event1) => event0.ts - event1.ts);
@@ -55,10 +55,9 @@ class TraceOfTab extends ComputedArtifact {
     /** @type {LH.TraceEvent|undefined} */
     let startedInPageEvt;
     const startedInBrowserEvt = keyEvents.find(e => e.name === 'TracingStartedInBrowser');
-    // `persistentIds` is a signal that the frame data will be attached, prefer it when available.
     if (startedInBrowserEvt && startedInBrowserEvt.args.data &&
-        startedInBrowserEvt.args.data.persistentIds) {
-      const mainFrame = (startedInBrowserEvt.args.data.frames || []).find(frame => !frame.parent);
+        startedInBrowserEvt.args.data.frames) {
+      const mainFrame = startedInBrowserEvt.args.data.frames.find(frame => !frame.parent);
       const pid = mainFrame && mainFrame.processId;
       const threadNameEvt = keyEvents.find(e => e.pid === pid && e.ph === 'M' &&
         e.cat === '__metadata' && e.name === 'thread_name' && e.args.name === 'CrRendererMain');
--- a/lighthouse-core/gather/computed/trace-of-tab.js
+++ b/lighthouse-core/gather/computed/trace-of-tab.js
@@ -42,21 +42,42 @@ class TraceOfTab extends ComputedArtifact {
     // *must* be stable to keep events correctly nested.
     /** @type Array<LH.TraceEvent> */
     const keyEvents = trace.traceEvents
-      .filter(e => {
-        return e.cat.includes('blink.user_timing') ||
+      .filter(e =>
+          e.cat.includes('blink.user_timing') ||
           e.cat.includes('loading') ||
           e.cat.includes('devtools.timeline') ||
-          e.name === 'TracingStartedInPage';
-      })
+          e.cat === '__metadata')
       // @ts-ignore - stableSort added to Array by WebInspector.
       .stableSort((event0, event1) => event0.ts - event1.ts);
 
-    // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
-    // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
-    const startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+    // Find out the inspected page frame.
+    /** @type {LH.TraceEvent|undefined} */
+    let startedInPageEvt;
+    const startedInBrowserEvt = keyEvents.find(e => e.name === 'TracingStartedInBrowser');
+    if (startedInBrowserEvt && startedInBrowserEvt.args.data &&
+        startedInBrowserEvt.args.data.persistentIds) {
+      const mainFrame = (startedInBrowserEvt.args.data.frames || []).find(frame => !frame.parent);
+      const pid = mainFrame && mainFrame.processId;
+      const threadNameEvt = keyEvents.find(e => e.pid === pid && e.ph === 'M' &&
+        e.cat === '__metadata' && e.name === 'thread_name' &&
+        // @ts-ignore - property chain exists for 'thread_name' event.
+        e.args.name === 'CrRendererMain');
+      startedInPageEvt = mainFrame && threadNameEvt ?
+        Object.assign({}, startedInBrowserEvt, {
+          pid, tid: threadNameEvt.tid, name: 'TracingStartedInPage',
+          args: {data: {page: mainFrame.frame}}}) :
+        undefined;
+    }
+    // Support legacy browser versions that do not emit TracingStartedInBrowser event.
+    if (!startedInPageEvt) {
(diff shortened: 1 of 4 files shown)


lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -45,18 +45,30 @@ class TraceOfTab extends ComputedArtifact {
       .filter(e => {
         return e.cat.includes('blink.user_timing') ||
           e.cat.includes('loading') ||
-          e.cat.includes('devtools.timeline') ||
-          e.name === 'TracingStartedInPage';
```
brendankenny: think you don't want to filter this out to support the legacy case? Or in the fallback case need to search for the first `TracingStartedInPage` in the original `trace.traceEvents`

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -45,18 +45,30 @@ class TraceOfTab extends ComputedArtifact {
...
 
-    // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
-    // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
-    const startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
-    if (!startedInPageEvt) throw new LHError(LHError.errors.NO_TRACING_STARTED);
-    // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
+    // Find out the inspected page frame.
+    let startedInPageEvt;
+    const startedInBrowserEvt = keyEvents.find(e => e.name === 'TracingStartedInBrowser');
+    if (startedInBrowserEvt && startedInBrowserEvt.args.data.persistentIds) {
```
brendankenny: what's the significance of `persistentIds`? doesn't appear to be used anywhere?

lighthouse-core/gather/computed/trace-of-tab.js
brendankenny: also, does the filter now need to let through `e.name === 'TracingStartedInBrowser'` events?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -45,18 +45,30 @@ class TraceOfTab extends ComputedArtifact {
...
+      const mainFrame = (startedInBrowserEvt.args.data.frames || []).find(frame => !frame.parent);
+      const mainFrameId = mainFrame && mainFrame.frame || '';
```
brendankenny: what's the use of the fallback `''` value here if there's no `mainFrame` or `mainFrameId` found? Seems like we need those and should just throw if not found

(or could set `startedInPageEvt` to `undefined` and let the legacy fallback below try to find a backup `TracingStartedInPage` event)

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -45,18 +45,30 @@ class TraceOfTab extends ComputedArtifact {
...
+      startedInPageEvt =
+        {...startedInBrowserEvt, name: 'TracingStartedInPage', args: {data: {page: mainFrameId}}};
+    }
+    // Support legacy browser versions that do not emit TracingStartedInBrowser event.
+    if (!startedInPageEvt) {
+      // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
+      // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
+      startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+      if (!startedInPageEvt) throw new LHError(LHError.errors.NO_TRACING_STARTED);
+      // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
```
brendankenny: two lines above should be pulled out of the conditional (throw if no `startedInPageEvt` found, `// @ts-ignore ` comment is for the `const frameId = ` line below)

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: Both TracingStartedInPage and TracingStartedInBrowser are in devtools.timeline category which is already allowed.

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: We used to have a TracingStartedInBrowser event before, but with no frames information. persistentIds is a marker that frames are provided.

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: Agree

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: It doesn't really matter as when we didn't enter the `if` block the `startedInPageEvt` is already there. Anyway, I will move it.

lighthouse-core/test/gather/computed/trace-of-tab-test.js
```diff
@@ -16,6 +16,8 @@ const preactTrace = require('../../fixtures/traces/preactjs.com_ts_of_undefined.
 const noFMPtrace = require('../../fixtures/traces/no_fmp_event.json');
 const noFCPtrace = require('../../fixtures/traces/airhorner_no_fcp');
 const backgroundTabTrace = require('../../fixtures/traces/backgrounded-tab-missing-paints');
+const tracingStartedInBrowserTrace =
+  require('../../fixtures/traces/tracing-started-in-browser.json');
```
patrickhulce: this whole fixture is just to exercise that one event right?

could we use a reduced trace or even just a fake array of events for this? 6k lines for the 1 relevant event is a lot :)

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: It seems like it was in toplevel once upon a time based on the fixture change below?

my traces from ~December have it in `disabled-by-default-devtools.timeline` so I guess it's a question of how far back we want to go

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -106,12 +121,13 @@ class TraceOfTab extends ComputedArtifact {
     // stable-sort events to keep them correctly nested.
     /** @type Array<LH.TraceEvent> */
     const processEvents = trace.traceEvents
+      // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
```
patrickhulce: hmmm, we shouldn't need the ts-ignore for .pid, what was the error you're suppressing?

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: AFAICT TracingStartedInPage was introduced 4 years ago with devtools.timeline category. See https://codereview.chromium.org/254613002/diff/90001/Source/core/inspector/InspectorTracingAgent.cpp
I wonder where the traces with it being in toplevel come from?

patrickhulce: Hm, no clue then 🤷‍♂️ 4 years is plenty good enough for me 😆

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: error TS2532: Object is possibly 'undefined'.

patrickhulce: Oh because startedInPageEvt is `LH.TraceEvent|undefined` and we're accessing inside a different block :/

mind just casting instead then? `/** @type {LH.TraceEvent} */ (startedInPageEvt).pid` (same for .tid below)

a1ph: Sure, I can annotate. I surprised the compiler hasn't noticed `if (!startedInPageEvt) throw`

brendankenny: yeah, they still do a lot simpler CFG work than closure, focusing almost entirely locally, so things like this still trip them up even while being more sophisticated in other ways :)

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -45,18 +45,34 @@ class TraceOfTab extends ComputedArtifact {
...
+    }
+    // Support legacy browser versions that do not emit TracingStartedInBrowser event.
+    if (!startedInPageEvt) {
+      // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
+      // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
+      startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+    }
     if (!startedInPageEvt) throw new LHError(LHError.errors.NO_TRACING_STARTED);
     // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
     const frameId = startedInPageEvt.args.data.page;
```
brendankenny: if the casts are annoying, could also do something like 
```js
 /** @type {LH.TraceEvent|undefined} */
let possibleStartedInPageEvt;
...
```
above, and then here
```js
const startedInPageEvt = possibleStartedInPageEvt;
```
because the compiler will know that `possibleStartedInPageEvt` is definitely defined at this point, the type for `startedInPageEvt` will be inferred correctly as `LH.TraceEvent`, and then it won't need any casts below this.

a1ph: I considered that, but I didn't want to introduce more code to workaround compiler issues. I'm fine with casts.

lighthouse-core/test/gather/computed/trace-of-tab-test.js
```diff
@@ -112,6 +112,55 @@ describe('Trace of Tab computed artifact:', () => {
...
+        'pid': 69920,
+        'tid': 1,
+        'ts': 0,
+        'ph': 'M',
+        'cat': '__metadata',
+        'name': 'thread_name',
+        'args': {'name': 'CrRendererMain'},
+      }]};
+    const trace = await traceOfTab.compute_(tracingStartedInBrowserTrace);
+    assert.equal(trace.startedInPageEvt.ts, 2193564729582);
```
patrickhulce: want to assert the navStart is found and in the trace.mainThreadEvents since I assume that's what the `thread_name` addition was for?

a1ph: Done

patrickhulce:
@brendankenny does this look good to go to you?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -42,21 +42,42 @@ class TraceOfTab extends ComputedArtifact {
     // *must* be stable to keep events correctly nested.
     /** @type Array<LH.TraceEvent> */
     const keyEvents = trace.traceEvents
-      .filter(e => {
-        return e.cat.includes('blink.user_timing') ||
+      .filter(e =>
+          e.cat.includes('blink.user_timing') ||
```
brendankenny: generally for multi-line arrow function we keep the curly braces/`return` for clarity. Any reason to drop it here?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -42,21 +42,42 @@ class TraceOfTab extends ComputedArtifact {
...
 
-    // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
-    // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
-    const startedInPageEvt = keyEvents.find(e => e.name === 'TracingStartedInPage');
+    // Find out the inspected page frame.
+    /** @type {LH.TraceEvent|undefined} */
+    let startedInPageEvt;
+    const startedInBrowserEvt = keyEvents.find(e => e.name === 'TracingStartedInBrowser');
+    if (startedInBrowserEvt && startedInBrowserEvt.args.data &&
+        startedInBrowserEvt.args.data.persistentIds) {
```
brendankenny: can you add a comment about why `persistentIds` is significant?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -42,21 +42,42 @@ class TraceOfTab extends ComputedArtifact {
...
+      const mainFrame = (startedInBrowserEvt.args.data.frames || []).find(frame => !frame.parent);
+      const pid = mainFrame && mainFrame.processId;
+      const threadNameEvt = keyEvents.find(e => e.pid === pid && e.ph === 'M' &&
+        e.cat === '__metadata' && e.name === 'thread_name' &&
+        // @ts-ignore - property chain exists for 'thread_name' event.
+        e.args.name === 'CrRendererMain');
```
brendankenny: since it's just a single level into `args`, add `name?: string` to `args` in `externs.d.ts` instead of `@ts-ignore` here?

patrickhulce: done

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: done

lighthouse-core/gather/computed/trace-of-tab.js
a1ph: I talked to dgozman, it was added to ensure this is a new version of the event that carries frames. I think we can just check for frames field being there.

patrickhulce:
oops you got some conflicts now @a1ph! I had fixed the feedback before you pushed sorry!

patrickhulce:
thanks @a1ph!

PR #5293, opened 2018-05-21
extension: expose URL shim

### LINKED ISSUE #5273: SEO - Page is blocked from indexing - robots.txt
I wrote about this bug before, then someone correctly mentioned an issue on one of the sites in question, which made me question if it was a bug, however, I have since corrected that issue, and after more testing, every single site I try and analyze with Lighthouse 3.0.0-alpha.2 says that "Page is blocked from indexing" and applies an SEO penalty.  Older versions of Lighthouse seem to be fine.  This is the version of Lighthouse that automatically downloads / installs when a user visits the Chrome webstore to add the Lighthouse plugin as an extension for easier one click testing.  Other reviews seem to indicate that perhaps this version was pushed to the Chrome webstore a little too early?

Version info on the Chrome webstore indicates it is 2.10.0.3002 (May 16, 2018) however what actually downloads and installs as a browser plugin/extension is 3.0.0-alpha.2.

image

image

patrickhulce:
**Related Issues/PRs**
fixes #5273

### DIFF
--- a/lighthouse-core/lib/url-shim.js
+++ b/lighthouse-core/lib/url-shim.js
@@ -132,6 +132,7 @@ class URLShim extends URL {
   }
 }
 
+URLShim.URL = URL;
 URLShim.URLSearchParams = (typeof self !== 'undefined' && self.URLSearchParams) ||
     require('url').URLSearchParams;
 
--- a/lighthouse-extension/gulpfile.js
+++ b/lighthouse-extension/gulpfile.js
@@ -141,6 +141,11 @@ gulp.task('browserify-lighthouse', () => {
         bundle = bundle.require(artifact, {expose: artifact.replace(corePath, './')});
       });
 
+      // browerify's url shim doesn't work with .URL in node_modules, so expose our own
+      // @see https://github.com/GoogleChrome/lighthouse/issues/5273
+      const pathToURLShim = require.resolve('../lighthouse-core/lib/url-shim.js');
+      bundle = bundle.require(pathToURLShim, {expose: 'url'});
+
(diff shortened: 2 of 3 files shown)


lighthouse-extension/gulpfile.js
```diff
@@ -141,6 +141,11 @@ gulp.task('browserify-lighthouse', () => {
         bundle = bundle.require(artifact, {expose: artifact.replace(corePath, './')});
       });
 
+      // browerify's url shim doesn't work with .URL in node_modules, so expose our own
+      // @see https://github.com/GoogleChrome/lighthouse/issues/5273
+      const pathToURLShim = require.resolve('../lighthouse-core/lib/url-shim.js');
+      bundle = bundle.require(pathToURLShim, {expose: 'url'});
```
paulirish: where is this lowercase usage of it used? i'm not following what the fix is here.

lighthouse-core/lib/url-shim.js
```diff
@@ -132,6 +132,7 @@ class URLShim extends URL {
   }
 }
 
+URLShim.URL = URL;
```
paulirish: who uses URL.URL?

lighthouse-extension/gulpfile.js
patrickhulce: inside robots-parser it does `var URL = require('url').URL`

lighthouse-core/lib/url-shim.js
patrickhulce: robots parser, see above :)

PR #5295, opened 2018-05-22
report: move runtime settings to footer

paulirish:
screenshots
image

image

deployment should have it, too.


lighthouse-core/report/html/templates.html
```diff
@@ -432,24 +399,47 @@
...
+    }
+    .lh-env__items {
+      padding-left: 16px;
+    }
+    span.lh-env__name {
+      font-weight: bold;
+    }
+    span.lh-env__description {
+      font-family: var(--monospace-font-family);
+      font-size-adjust: 0.5;
```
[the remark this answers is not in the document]
paulirish: yup whoops. i had on experimental web platform features. i'll do this without FSA.

paulirish:
>  I feel like a question we often get is "What device/network configuration is this Lighthouse run reflective of"

aye, though this will rarely be changing. so people just need to see it like once. IMO at the bottom / visible by default is already a lot more visible than previously, so I think this audience is probably satisfied by the current PR.

>  is there anyway we can have just the device and network settings in the header?  (or a summary of it)

It's a lot of text, so it'd have to be pretty minimal. Hwi had something along these lines before. It works for me.

image

paulirish:
okay summary line added to the report header, it can be these strings:

```
    Emulated Nexus 5X, No throttling applied
    Emulated Nexus 5X, Throttled 3G
    Emulated Nexus 5X, Simulated 3G
    No emulation, Simulated 3G
    No emulation, No throttling applied
```

also it's a subtle link and can be clicked to see all the details in the bottom.

tweaked styles in the bottom

vinamratasingal:
Small nit: Can the summary say "3G network" 

Otherwise, LGTM

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -65,23 +65,12 @@ class ReportRenderer {
...
-    const environment = Util.getEnvironmentDisplayValues(report.configSettings || {});
-    environment.forEach(runtime => {
-      const item = this._dom.cloneTemplate('#tmpl-lh-env__items', env);
-      this._dom.find('.lh-env__name', item).textContent = runtime.name;
-      this._dom.find('.lh-env__description', item).textContent = runtime.description;
-      env.appendChild(item);
-    });
+    url.href = url.textContent = toolbarUrl.href = toolbarUrl.textContent = report.finalUrl;
 
+    const emuDescs = Util.getEmulationDescriptions(report.configSettings || {});
```
patrickhulce: the `lation` is just too much, huh?

let's use emulation and nix the `/** @type HTMLAnchorElement`

lighthouse-core/report/html/renderer/util.js
```diff
@@ -314,59 +314,66 @@ class Util {
...
         networkThrottling = 'Provided by environment';
+        summary = 'No throttling applied';
         break;
       case 'devtools': {
         const {cpuSlowdownMultiplier, requestLatencyMs} = throttling;
         cpuThrottling = `${Util.formatNumber(cpuSlowdownMultiplier)}x slowdown (DevTools)`;
         networkThrottling = `${Util.formatNumber(requestLatencyMs)}${NBSP}ms HTTP RTT, ` +
           `${Util.formatNumber(throttling.downloadThroughputKbps)}${NBSP}Kbps down, ` +
           `${Util.formatNumber(throttling.uploadThroughputKbps)}${NBSP}Kbps up (DevTools)`;
+        summary = 'Throttled Fast–3G network';
```
patrickhulce: I don't think we style this as `Fast-3G` anywhere else, just `Fast 3G`?

patrickhulce:
tests need some work though

PR #5296, opened 2018-05-22
core(lhr): migrate opportunity details to new format

brendankenny:
part of #5008

Moves opportunity details over to the new format. Wanted to get feedback before changing the many tests that are failing due to the slightly different details shape.

Most of the byte efficiency audit changes are just name changes, e.g. `results`->`items`, `itemType`->`valueType`.

Shape of the new opportunity `details` is what's in lhr-lite: https://github.com/GoogleChrome/lighthouse/blob/804e9b8c9703fac356474ac1cf6012dc13133376/typings/lhr-lite.d.ts#L85-L91
with the addition of more possibilities in `OpportunityDetailsItem` than is in the more limited `lhr-lite`.

### DIFF
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -126,6 +126,23 @@ class Audit {
   }
 
   /**
+   * @param {Array<LH.ResultLite.Audit.ColumnHeading>} headings
+   * @param {Array<LH.ResultLite.Audit.WastedBytesDetailsItem>|Array<LH.ResultLite.Audit.WastedTimeDetailsItem>} items
+   * @param {number} overallSavingsMs
+   * @param {number=} overallSavingsBytes
+   * @return {LH.Result.Audit.OpportunityDetails}
+   */
+  static makeOpportunityDetails(headings, items, overallSavingsMs, overallSavingsBytes) {
+    return {
+      type: 'opportunity',
+      headings,
+      items,
+      overallSavingsMs,
+      overallSavingsBytes,
+    };
+  }
+
+  /**
    * @param {typeof Audit} audit
    * @param {LH.Audit.Product} result
    * @return {{score: number|null, scoreDisplayMode: LH.Audit.ScoreDisplayMode}}
--- a/lighthouse-core/audits/audit.js
+++ b/lighthouse-core/audits/audit.js
@@ -135,7 +135,7 @@ class Audit {
   static makeOpportunityDetails(headings, items, overallSavingsMs, overallSavingsBytes) {
     return {
       type: 'opportunity',
-      headings,
+      headings: items.length === 0 ? [] : headings,
       items,
       overallSavingsMs,
       overallSavingsBytes,
--- a/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
+++ b/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
@@ -207,20 +207,14 @@ class RenderBlockingResources extends Audit {
       displayValue = `${results.length} resource delayed first paint by ${wastedMs}ms`;
     }
 
+    /** @type {LH.Result.Audit.OpportunityDetails['headings']} */
     const headings = [
-      {key: 'url', itemType: 'url', text: 'URL'},
-      {
-        key: 'totalBytes',
-        itemType: 'bytes',
-        displayUnit: 'kb',
-        granularity: 0.01,
-        text: 'Size (KB)',
-      },
-      {key: 'wastedMs', itemType: 'ms', text: 'Download Time (ms)', granularity: 1},
+      {key: 'url', valueType: 'url', label: 'URL'},
+      {key: 'totalBytes', valueType: 'bytes', label: 'Size (KB)'},
+      {key: 'wastedMs', valueType: 'timespanMs', label: 'Download Time (ms)'},
     ];
 
-    const summary = {wastedMs};
-    const details = Audit.makeTableDetails(headings, results, summary);
+    const details = Audit.makeOpportunityDetails(headings, results, wastedMs);
 
     return {
       displayValue,
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -299,7 +299,7 @@ class DetailsRenderer {
           }
           case 'bytes': {
             const numValue = /** @type {number} */ (value);
-            itemElement = this._renderBytes({value: numValue});
+            itemElement = this._renderBytes({value: numValue, granularity: 1});
             break;
           }
           case 'thumbnail': {
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -59,7 +59,7 @@ class DetailsRenderer {
         // @ts-ignore - TODO(bckenny): Fix type hierarchy
         return this._renderTable(/** @type {TableDetailsJSON} */ (details));
       case 'code':
-        return this._renderCode(/** @type {StringDetailsJSON} */ (details));
+        return this._renderCode(/** @type {DetailsJSON} */ (details));
       case 'node':
         return this.renderNode(/** @type {NodeDetailsJSON} */(details));
       case 'criticalrequestchain':
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -9,6 +9,10 @@
 
 /** @typedef {import('./dom.js')} DOM */
 /** @typedef {import('./crc-details-renderer.js')} CRCDetailsJSON */
+/** @typedef {LH.Result.Audit.OpportunityDetails} OpportunityDetails */
+
+/** @type {Array<string>} */
+const URL_PREFIXES = ['http://', 'https://', 'data:'];
 
 class DetailsRenderer {
   /**
@@ -29,7 +33,7 @@ class DetailsRenderer {
   }
 
   /**
-   * @param {DetailsJSON} details
+   * @param {DetailsJSON|OpportunityDetails} details
    * @return {Element}
    */
   render(details) {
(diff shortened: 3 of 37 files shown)


lighthouse-core/audits/audit.js
```diff
@@ -125,6 +125,23 @@ class Audit {
...
   }
 
+  /**
+   * @param {Array<LH.ResultLite.Audit.ColumnHeading>} headings
+   * @param {Array<LH.ResultLite.Audit.WastedBytesDetailsItem>|Array<LH.ResultLite.Audit.WastedTimeDetailsItem>} items
+   * @param {number} overallSavingsMs
+   * @param {number=} overallSavingsBytes
+   * @return {LH.Result.Audit.OpportunityDetails}
+   */
+  static makeOpportunityDetails(headings, items, overallSavingsMs, overallSavingsBytes) {
```
brendankenny: there's no real reason this needs to be a thing, we just need a place to check that audits are returning the correct shape. If we start enforcing the return type of audits based on the details they provide (e.g. `@return {LH.Audit.Result<OpportunityDetails>}` or whatever) that could work just as well.

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -242,6 +244,79 @@ class DetailsRenderer {
     return tableElem;
   }
 
+  /**
```
brendankenny: this is called `_renderOpportunityTable` for now, but I'd like to migrate all audits with table details to this one and eventually just have it be `_renderTable`. The changes from the original mean we only have so many types of things that can fit in the table cells, so just call them out explicitly instead of having nested details that can themselves contain more details, etc.

E.g. here we get rid of the "URLs" that are actually `{type: 'code', value: txt}`, and instead just say if it doesn't look like a URL, render it in `<pre></pre>` tags

patrickhulce: Does this mean we'd eliminate the `fallbackType/fallbackKey` approach? Also this method feels a bit long, want to break it up? :)

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -55,31 +59,34 @@ class DetailsRenderer {
...
+        return this._renderCode(/** @type {StringDetailsJSON} */ (details));
       case 'node':
         return this.renderNode(/** @type {NodeDetailsJSON} */(details));
       case 'criticalrequestchain':
         return CriticalRequestChainRenderer.render(this._dom, this._templateContext,
           // @ts-ignore - TODO(bckenny): Fix type hierarchy
           /** @type {CRCDetailsJSON} */ (details));
+      case 'opportunity':
+        // @ts-ignore - TODO(bckenny): Fix type hierarchy
+        return this._renderTable(/** @type {OpportunityDetails} */ (details));
```
patrickhulce: render opportunity table?

lighthouse-core/test/results/sample_v2.json
```diff
@@ -788,12 +789,21 @@
         0
       ],
       "details": {
-        "type": "table",
-        "headings": [],
+        "type": "opportunity",
+        "headings": [
```
patrickhulce: seems like we still don't want to populate headings if there are no items

typings/lhr.d.ts
```diff
@@ -81,6 +81,23 @@ declare global {
...
       }
+
+      export module Audit {
+        export interface OpportunityDetailsItem {
+          url: string;
+          wastedBytes?: number;
+          totalBytes?: number;
+          wastedMs?: number;
+          wastedPercent?: number;
+          fromProtocol?: boolean;
```
patrickhulce: do we want to just nuke these?

lighthouse-core/report/html/renderer/details-renderer.js
brendankenny: > I like this. Does this mean we'd eliminate the fallbackType/fallbackKey approach?

I guess it comes down to how flexible we need it to end up being. It doesn't seem like we have a ton of cases where we need fallbacks, so we can probably get by with using hybrid `valueType`s, like `url` is here (maybe we'd want to rename them based on their actual combined type? e.g. `'urlOrCode'`, but something better :)

typings/lhr.d.ts
brendankenny: > do we want to just nuke these?

yeah, I was building them up to make sure every case was covered, but probably should just be an index type

typings/lhr.d.ts
```diff
@@ -81,6 +81,23 @@ declare global {
         blockedUrlPatterns: string[];
         extraHeaders: Crdp.Network.Headers;
       }
+
+      export module Audit {
```
paulirish: curious why this is in lhr.d.ts instead of audit.d.ts

lighthouse-core/report/html/renderer/details-renderer.js
paulirish: can you add a comment with this todo?

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -55,31 +59,34 @@ class DetailsRenderer {
...
       default: {
         throw new Error(`Unknown type: ${details.type}`);
       }
     }
   }
 
   /**
-   * @param {NumericUnitDetailsJSON} details
+   * @param {{value: number, granularity?: number}} details
```
paulirish: bytes defaults to granularity of 0.1 in util, so removing the granularity in the audits means these will now show as more precise, right?

dont' we want to keep the granularity the same?

typings/lhr.d.ts
brendankenny: > curious why this is in lhr.d.ts instead of audit.d.ts

Like `lhr-lite.d.ts`, it seems like all the LHR stuff should be defined in `lhr.d.ts`, instead of spread over `lhr.d.ts` and `audit.d.ts`. That would leave `audit.d.ts` with types internal to operating an audit, while `Audit.Result` stuff would all be in `lhr.d.t.s` (with a good chunk of that just inherited from `lhr-lite.d.ts`).

This starts on that path, plus is a nice dividing line between this finalized-ish `details` definition vs the handwavey/just-make-it-work definitions in `audit.d.ts`

lighthouse-core/report/html/renderer/details-renderer.js
brendankenny: > bytes defaults to granularity of 0.1 in util, so removing the granularity in the audits means these will now show as more precise, right? dont' we want to keep the granularity the same?

I forget what we decided on granularity for these. Always to `1` or was bytes going to be one of the auto-formatting options?

paulirish: happy with defaulting opportunity bytes granularity to 1.

typings/lhr.d.ts
brendankenny: > but probably should just be an index type

turns out it wasn't that simple because we want them to extend lhr-lite, but now an index type + minimum set of properties

brendankenny:
turns out tests weren't bad, mostly `s/results/items` and flattening the smoke expectations a bit. Ready for really real review

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -55,31 +59,34 @@ class DetailsRenderer {
         // @ts-ignore - TODO(bckenny): Fix type hierarchy
         return this._renderTable(/** @type {TableDetailsJSON} */ (details));
       case 'code':
-        return this._renderCode(details);
+        return this._renderCode(/** @type {StringDetailsJSON} */ (details));
```
patrickhulce: if we're killing string details json elsewhere should we be moving this to it?

lighthouse-core/report/html/report-styles.css
```diff
@@ -842,6 +842,7 @@ summary.lh-passed-audits-summary {
 
 .lh-table-column--text,
 .lh-table-column--bytes,
+.lh-table-column--timespanMs,
```
patrickhulce: 😢 I guess a _.toKebabCase is a bit much huh?

lighthouse-core/test/audits/byte-efficiency/unused-css-rules-test.js
```diff
@@ -109,9 +109,9 @@ describe('Best Practices: unused css rules audit', () => {
     });
 
     it('correctly computes url', () => {
-      const expectedPreview = {type: 'code', value: 'dummy'};
-      assert.deepEqual(map({header: {sourceURL: ''}}).url, expectedPreview);
-      assert.deepEqual(map({header: {sourceURL: 'a'}}, 'http://g.co/a').url, expectedPreview);
+      const expectedPreview ='dummy';
```
patrickhulce: nit: space after =

lighthouse-core/test/results/sample_v2.json
patrickhulce: bump

lighthouse-core/report/html/renderer/details-renderer.js
brendankenny: > if we're killing string details json elsewhere should we be moving this to it?

I can leave it as `DetailsJSON`? Basically I need to tell it that it's not an `OpportunityDetails`, and until all the details types are known to the compiler with `type` defined it isn't narrowing it down here

lighthouse-core/report/html/report-styles.css
brendankenny: > I guess a _.toKebabCase is a bit much huh?

ha, that's a good point, but it'd also be pretty ugly. Is the tear because you're accepting living with it or do you really want this?

lighthouse-core/test/audits/byte-efficiency/unused-css-rules-test.js
brendankenny: what are you doing eslint

brendankenny:
updated

lighthouse-core/report/html/report-styles.css
patrickhulce: 💦 because I'm living with it :)

lighthouse-core/report/html/renderer/details-renderer.js
patrickhulce: Oh I meant elsewhere it seems like we're deleting `StringDetailsJSON` and going with literal `{text: string}` typedefs, so should we just do that instead

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -242,6 +244,81 @@ class DetailsRenderer {
     return tableElem;
   }
 
+  /**
+   * TODO(bckenny): migrate remaining table rendering to this function, then rename
+   * back to _renderTable and replace the original.
+   * @param {OpportunityDetails} details
+   * @return {Element}
+   */
+  _renderOpportunityTable(details) {
```
patrickhulce: this fn is a bit long, you foresee breaking it up in the future? add a TODO if so :)

brendankenny: done!

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -207,20 +207,14 @@ class RenderBlockingResources extends Audit {
...
-        key: 'totalBytes',
-        itemType: 'bytes',
-        displayUnit: 'kb',
-        granularity: 0.01,
-        text: 'Size (KB)',
-      },
-      {key: 'wastedMs', itemType: 'ms', text: 'Download Time (ms)', granularity: 1},
+      {key: 'url', valueType: 'url', label: 'URL'},
+      {key: 'totalBytes', valueType: 'bytes', label: 'Size (KB)'},
+      {key: 'wastedMs', valueType: 'timespanMs', label: 'Download Time (ms)'},
```
paulirish: we'll be dropping granularity of 1 and getting default of 10 afaict.
that is probably fine, but just FYI

PR #5311, opened 2018-05-22
core(bootup-time): recalculate total time in JS bootup

paulirish:
It's likely this was my fault, but the bootup-time summarizes the total amount of bootup time, but it's ALL work, not just Javascript.

image

But that's at odds with the title of the audit "javascript boot-up time"

so now it only is scoped to JS work.

### DIFF
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -101,13 +101,13 @@ class BootupTime extends Audit {
         // Add up the totalBootupTime for all the taskGroups
         for (const [name, value] of Object.entries(groups)) {
           groups[name] = value * multiplier;
-          totalBootupTime += value * multiplier;
         }
 
         extendedInfo[url] = groups;
 
         const scriptingTotal = groups[groupIdToName.scripting] || 0;
         const parseCompileTotal = groups[groupIdToName.scriptParseCompile] || 0;
+        totalBootupTime += scriptingTotal + parseCompileTotal;
         return {
           url: url,
           sum: scriptingTotal + parseCompileTotal,
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -73,18 +73,54 @@ class BootupTime extends Audit {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/bootup-time.js
```diff
@@ -72,19 +72,55 @@ class BootupTime extends Audit {
     return result;
   }
 
+  /**
+   * Summarize JS costs by URL
+   * @param {Map<string, Object<string, number>>} executionTimings
+   * @param {LH.Audit.Context} context
+   * @return {{results: Object, totalBootupTime: number, extendedInfo: Object}}
+   */
+  static execCostsByURL(executionTimings, context) {
```
brendankenny: should be `executionCostsByURL`, but from the outside at least that doesn't appear to be any different than `getExecutionTimingsByURL`?

lighthouse-core/audits/bootup-time.js
```diff
@@ -72,19 +72,55 @@ class BootupTime extends Audit {
     return result;
   }
 
+  /**
+   * Summarize JS costs by URL
+   * @param {Map<string, Object<string, number>>} executionTimings
+   * @param {LH.Audit.Context} context
+   * @return {{results: Object, totalBootupTime: number, extendedInfo: Object}}
```
brendankenny: `Object` is just `any`, so these should be updated. They have types on master (though as local variables), so you can just use those

lighthouse-core/audits/bootup-time.js
```diff
@@ -72,19 +72,55 @@ class BootupTime extends Audit {
...
+      // Add up the totalBootupTime for all the taskGroups
+      for (const [name, value] of Object.entries(groups)) {
+        groups[name] = value * multiplier;
+      }
+
+      extendedInfo[url] = groups;
+
+      const scriptingTotal = groups[groupIdToName.scripting] || 0;
+      const parseCompileTotal = groups[groupIdToName.scriptParseCompile] || 0;
+      totalBootupTime += scriptingTotal + parseCompileTotal;
```
brendankenny: is the only change moving this line out of the for loop? Might be easier to just do that :)

PR #5330, opened 2018-05-23
misc: remove checkboxes from bug report template

patrickhulce:
Just created a bug report from the template and it doesn't make sense to use checkboxes for the environment since then it shows up in the issues UI as "progress" toward completion.


patrickhulce:
SGTM @paulirish done

PR #5341, opened 2018-05-24
extension: close the popup once the report has opened

### LINKED ISSUE #5340: extension: "Generating results" hangs around after report opns
to be fair this is the same session as #5339, but I've seen it before in the 3.0 era.

1. Run LH extension on paulirish.com
2. dont switch to another window while it's doing stuff.
3. it opens new window with the report, but the popup stays open on 'generating results...'

image

paulirish:
fixes #5340 

thanks https://stackoverflow.com/questions/4924615/chrome-extension-force-popup-html-to-close <3

Tested error -> reporting flow and it lg, too.

### DIFF
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -117,7 +117,7 @@ function createOptionItem(text, id, isChecked) {
  * @param {!Window} background Reference to the extension's background page.
  * @param {{selectedCategories: !Object<boolean>, useDevTools: boolean}} settings
  */
-function onGenerateReportButtonClick(background, settings) {
+async function onGenerateReportButtonClick(background, settings) {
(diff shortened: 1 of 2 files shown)


lighthouse-extension/app/src/popup.js
```diff
@@ -117,18 +117,27 @@ function createOptionItem(text, id, isChecked) {
...
+      flags: {throttlingMethod: useDevTools ? 'devtools' : 'simulate'},
+    }, selectedCategories);
+
+    // Close popup once report is opened in a new tab
+    window.close();
+  } catch (err) {
+    handleError(err);
+  }
+
+  function handleError(err) {
```
brendankenny: any reason to make this a function? I'd say either spin off as it's own function or just nest in the catch{}

PR #5342, opened 2018-05-24
core(tsc): migrate renderer & viewer off typedefs to .d.ts

paulirish:


lighthouse-viewer/app/src/viewer-ui-features.js
```diff
@@ -44,7 +42,7 @@ class ViewerUIFeatures extends ReportUIFeatures {
    * @override
    */
   getReportHtml() {
-    // @ts-ignore - TODO(bckenny): remove ignore when this.json is an LHR instead of ReportJSON.
+    // @ts-ignore - TODO(bckenny): remove ignore when this.json is an LHR instead of LH.ReportResult.
```
patrickhulce: everything extends LH.Result now, right? so we should be good to remove this?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -339,14 +336,16 @@ class CategoryRenderer {
     const passedElements = /** @type {Array<Element>} */ ([]);
     const notApplicableElements = /** @type {Array<Element>} */ ([]);
 
-    auditsUngrouped.failed.forEach((/** @type {AuditJSON} */ audit, i) =>
+    auditsUngrouped.failed.forEach((/** @type {LH.ReportResult.AuditRef} */ audit, i) =>
```
brendankenny: do these work without the cast now?

typings/html-renderer.d.ts
```diff
@@ -35,6 +35,21 @@ declare global {
     ReportUIFeatures: typeof _ReportUIFeatures;
     Util: typeof _Util;
   }
+
+  module LH {
+    export interface ReportResult extends Result {
```
brendankenny: maybe add a comment that this is for the bastardized report json we create within the report?

lighthouse-core/report/html/renderer/category-renderer.js
paulirish: yup. removed.

typings/html-renderer.d.ts
paulirish: done

paulirish:
@brendankenny ptal

lighthouse-viewer/app/src/viewer-ui-features.js
brendankenny: is this ignore still necessary? Since `LH.ReportResult` extends `LH.Result`, I would assume you could pass it into `ReportGenerator.generateReportHtml`?

PR #5025, opened 2018-04-24
tests(viewer): add pptr test for viewer

brendankenny:
As we start major work again on the report, it'd be good to have an easy way to verify we aren't breaking viewer. While viewer does already have compile and (a few) basic unit tests, this PR rips off @wardpeet's hard work from #4640 to make a pptr integration test for viewer as well.

Ideally we'd extract some sort of "use puppeteer to test a live report interface" test and have each individual directory file only in charge of setup and teardown, but I think(?) this is sufficient for now as we have more pressing matters.

PR #5345, opened 2018-05-24
docs(releasing): minor updates

paulirish:
_update_: see https://github.com/GoogleChrome/lighthouse/pull/5345#issuecomment-424825309

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -24,10 +24,8 @@ lighthouse-logger/
 node_modules/
 results/
 lantern-data/
-
-plots/
-
 latest-run/
+plots/
 
 # generated files
 **/pages/scripts/lighthouse-report.js
@@ -39,6 +37,7 @@ closure-error.log
 last-run-results.html
 latest.artifacts.log
 npm-debug.log
+yarn-error.log
 results.html
 
 # dev files
--- a/.npmignore
+++ b/.npmignore
@@ -7,7 +7,7 @@ assets/
 build/
 coverage/
 
-lighthouse-core/scripts/
+lighthouse-core/scripts/*
 lighthouse-core/test/
 lighthouse-core/third_party/src/
 lighthouse-viewer/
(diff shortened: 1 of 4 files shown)


brendankenny:
wait, it's still not ignored? https://github.com/GoogleChrome/lighthouse/blob/520f9ac06420634ccc03d0326e8e42fe83610a42/.npmignore#L23

paulirish:
@brendankenny blast from the past. ;)

##### changes

* lighthouse-core/scripts wasn't ignored correctly. their syntax is odd if you want most files in a dir ignored but one allowed. this works though.
* keep all of lighthouse-cli/tests. most notably all the fixtures to support smokehouse
* ignore latest-run
* remove mentions of `yarn smokehouse` since [that died here](https://github.com/GoogleChrome/lighthouse/commit/419b023680a51080d078b0c782ebae23b34fefdc).
* remove the ["drive-by-tightening"](https://github.com/GoogleChrome/lighthouse/pull/5025/files#r183588493) that doesnt seem to help and actually prevents the `npm explore lighthouse -- npm run smoke` from working.

.npmignore
```diff
@@ -18,23 +18,17 @@ lighthouse-logger/
...
-!lighthouse-cli/test/smokehouse
-!lighthouse-cli/test/fixtures/
-
-!lighthouse-cli/index.js
-lighthouse-cli/*.map
-lighthouse-cli/commands/*.map
-lighthouse-cli/types/*.js
-lighthouse-cli/types/*.map
+# keep smokehouse tests, etc
+!lighthouse-cli/test
```
brendankenny: how much bigger does our package end up?

paulirish: oh this was not fun. turns out `yarn pack` builds things VERY differently than `npm pack`. 
and `npm pack` gives basically the same package before and after this PR. wonderful.

paulirish: turns out it's smaller now. because the npmignore in master includes all of lighthouse-cli already.

so aint that somethin

PR #5346, opened 2018-05-24
extension(tsc): add type checking to extension entry points

brendankenny:
to do this, needed to do a little import/export doctoring.

`lighthouse-background.js`
- exports functions to `window` if not in node (this is how devtools uses it), otherwise exports function as a node module

`lighthouse-ext-background.js`
- requires `lighthouse-background.js` in module form, allowing cross-file references to work in type checking and IDE navigation. Browserify handles the require normally (before it was still required but not as a module, but wrote its exports to `window` so they were available there)
- exports functions to `window`, since this file is only ever used where there is a window, but *also* to a module so `popup.js` can access for type checking

`popup.js`
- has a typedef import of `lighthouse-ext-background.js` *only* for types, so still works as before, but now types of calls to the background page are checked and IDE nav, etc works

~git isn't picking up that `lighthouse-ext-background.js` moved to `extension-background.js`, so it's treating it as a totally new file. Sorry about that. I can revert the move and do that afterwards if that's easier to review.~ reverted file renaming for now


lighthouse-core/gather/connections/raw.js
```diff
@@ -10,22 +10,11 @@ const Connection = require('./connection.js');
 /* eslint-disable no-unused-vars */
 
 /**
- * @interface
```
brendankenny: turns out this was never working as an interface for tsc (@interface has closure support only). Making a typedef so can be imported elsewhere

lighthouse-extension/app/src/bundle-entry.js
```diff
@@ -39,30 +43,46 @@ window.runLighthouseForConnection = function(
...
+}
 
-window.listenForStatus = function(listenCallback) {
+/** @param {(status: [string, string, string]) => void} listenCallback */
+function listenForStatus(listenCallback) {
   log.events.addListener('status', listenCallback);
-};
+}
+
+if (typeof module !== 'undefined' && module.exports) {
```
patrickhulce: can we just comment that this is for browserify and that it's not really executed in node? this threw me a bit :)

lighthouse-extension/app/src/popup.js
```diff
@@ -45,17 +44,38 @@ function getLighthouseCommitHash() {
...
 }
 
+/**
+ * Guaranteed context.querySelector. Always returns an element or throws if
+ * nothing matches query.
+ * @param {string} query
+ * @param {ParentNode} context
+ * @return {HTMLElement}
+ */
+function find(query, context) {
```
patrickhulce: we seem to always use `document`, do we need context?

brendankenny:
ok, yeah, that's way better without the file renames :) will propose that in a future PR

lighthouse-extension/app/manifest_canary.json
```diff
@@ -12,8 +12,7 @@
   "default_locale": "en",
   "background": {
     "scripts": [
-      "scripts/chromereload.js",
```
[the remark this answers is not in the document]
brendankenny: I'm good with deleting, but we should probably delete the actual extension in that case? @paulirish

lighthouse-core/index.js
```diff
@@ -30,16 +29,17 @@ const Config = require('./config/config');
  * @param {string} url
  * @param {LH.Flags} flags
  * @param {LH.Config.Json|undefined} configJSON
- * @return {Promise<LH.RunnerResult>}
+ * @return {Promise<LH.RunnerResult|undefined>}
  */
-function lighthouse(url, flags = {}, configJSON) {
```
patrickhulce: this is a breaking change for `lighthouse('http://example.com')` in node, do we really need to do this?

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -70,117 +47,101 @@ function updateBadgeUI(optUrl) {
...
-      // return enableOtherChromeExtensions(true).then(_ => {
-      throw err;
-      // });
-    });
-};
+  const url = await connection.getCurrentTabURL();
+  const runnerResult = await background.runLighthouseForConnection(connection, url, options,
+    categoryIDs, updateBadgeUI);
+  if (runnerResult) {
+    // For now, should always be a runnerResult as the extension can't do `gatherMode`
```
patrickhulce: should we throw if no runner result? seems like a fatal condition

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -5,46 +5,20 @@
...
+
 const STORAGE_KEY = 'lighthouse_audits';
 const SETTINGS_KEY = 'lighthouse_settings';
 
-// let installedExtensions = [];
-let disableExtensionsDuringRun = false;
 let lighthouseIsRunning = false;
-let latestStatusLog = [];
-
-// /**
```
brendankenny: this was disabled in February 2017. I don't think we need to leave it in anymore just in case :)

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -190,53 +151,67 @@ window.loadSettings = function() {
...
   chrome.runtime.onInstalled.addListener(details => {
     if (details.previousVersion) {
       // eslint-disable-next-line no-console
       console.log('previousVersion', details.previousVersion);
     }
   });
 }
+
+if (typeof module !== 'undefined' && module.exports) {
+  // We don't want tsc to infer an index type, so use exports intest of module.exports.
```
patrickhulce: who consumes this via `require`?

lighthouse-core/index.js
brendankenny: do we support fully optional `flags`? I hadn't dug into it so I didn't pursue it (`{}` isn't currently castable to `LH.Flags` which is why I dropped it), but if so, agreed we should support that here.

I can add a manual cast of a `{}` here, maybe, and then follow up with solidifying the `flags` pipeline in a follow up?

lighthouse-extension/app/src/lighthouse-ext-background.js
brendankenny: SG

lighthouse-extension/app/src/lighthouse-ext-background.js
brendankenny: > who consumes this via require?

popup.js imports this for a typedef only. Added a comment to explain

patrickhulce: oh just for typedefs gotcha 👍  yeah comment is good

lighthouse-extension/app/src/popup.js
brendankenny: added default of `document`. WDYT?

lighthouse-core/index.js
patrickhulce: would `Partial<LH.Flags>` cut it? I guess we could differentiate LH.Flags from LH.CLIFlags which is basically what's defined now

lighthouse-extension/app/src/popup.js
patrickhulce: sure that sg 👍

lighthouse-core/index.js
brendankenny: > would Partial<LH.Flags> cut it? I guess we could differentiate LH.Flags from LH.CLIFlags which is basically what's defined now

yeah, I think that's basically what we need to do. Handy to have output from yargs, but then our optional flags fed into Config is a different beast (and it looks like completely optional is fine).

Ok if we do a future "Make Flags Good" PR while leaving this with the `{}` default?

patrickhulce: 👍

lighthouse-extension/app/manifest_canary.json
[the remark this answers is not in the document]
paulirish: ill delete canary extension now

lighthouse-extension/app/src/chromereload.js
```diff
@@ -1,29 +0,0 @@
-/**
```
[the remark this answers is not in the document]
paulirish: i do sometimes but dont most of the time. happy to get rid of it.

lighthouse-extension/app/src/lighthouse-ext-background.js
paulirish: HERO

lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -39,30 +43,47 @@ window.runLighthouseForConnection = function(
...
+}
 
-window.listenForStatus = function(listenCallback) {
+/** @param {(status: [string, string, string]) => void} listenCallback */
+function listenForStatus(listenCallback) {
   log.events.addListener('status', listenCallback);
-};
+}
+
+if (typeof module !== 'undefined' && module.exports) {
```
paulirish: really dig this. thanks.

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -70,117 +47,103 @@ function updateBadgeUI(optUrl) {
...
-window.runLighthouseAsInCLI = function(connection, url, options, categoryIDs) {
+async function runLighthouseAsInCLI(connection, url, options, categoryIDs) {
   log.setLevel('info');
-  const startTime = Date.now();
   options.flags = Object.assign({}, options.flags, {output: options.outputFormat});
 
-  return window.runLighthouseForConnection(connection, url, options, categoryIDs)
-    .then(results => {
-      const endTime = Date.now();
-      results.timing = {total: endTime - startTime};
```
paulirish: currently LR continues to expect a .timing.total prop.

lighthouse-extension/app/src/popup.js
```diff
@@ -173,85 +201,88 @@ function generateOptionsList(background, selectedCategories) {
...
-    // To prevent visual hiccups when opening the popup, we default the subpage
-    // to the "running" view and switch to the default view once we're sure
-    // Lighthouse is not already auditing the page. This change was necessary
-    // now that fetching the background event page is async.
-    if (background.isRunning()) {
-      showRunningSubpage();
-    } else {
-      hideRunningSubpage();
+async function initPopup() {
+  chrome.tabs.query({active: true, lastFocusedWindow: true}, function(tabs) {
```
paulirish: is changing the order here deliberate?

lighthouse-extension/app/src/lighthouse-ext-background.js
brendankenny: > currently LR continues to expect a .timing.total prop.

this is [added in runner now](https://github.com/GoogleChrome/lighthouse/blob/c5743e9a376b709d4a0e74d8cec7f02095a8a573/lighthouse-core/runner.js#L135), so doesn't need to be done here

lighthouse-extension/app/src/popup.js
brendankenny: > is changing the order here deliberate?

yeah, it used to fire off the promise to retrieve the background page and immediately run this query in parallel. Since I switched the `promise.then` to `await`, the query wouldn't run until after the background page had been found and the rest of the init stuff had run. Since at least `siteURL` is used for the pre-populated github message in case of a thrown error, it made sense to make sure it was set before the stuff

paulirish:
👍 to a file-rename PR afterwards. :)

brendankenny:
@patrickhulce second (third?) look?

PR #5347, opened 2018-05-24
core(simulator): improved timing typedef

patrickhulce:
cleans up a lot of the TODOs left based on the internal/external timing discrepancy

### DIFF
--- a/lighthouse-core/audits/uses-rel-preload.js
+++ b/lighthouse-core/audits/uses-rel-preload.js
@@ -112,7 +112,6 @@ class UsesRelPreloadAudit extends Audit {
       const originalNode = originalNodesByRecord.get(node.record);
       const timingAfter = simulationAfterChanges.nodeTimings.get(node);
       const timingBefore = simulationBeforeChanges.nodeTimings.get(originalNode);
-      // @ts-ignore TODO(phulce): fix timing typedef
       const wastedMs = Math.round(timingBefore.endTime - timingAfter.endTime);
       if (wastedMs < THRESHOLD_IN_MS) continue;
       results.push({url: node.record.url, wastedMs});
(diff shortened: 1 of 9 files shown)


lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -405,9 +431,20 @@ class Simulator {
 
     return {
       timeInMs: totalElapsedTime,
-      nodeTimings: this._nodeTimings,
+      nodeTimings: this._computeFinalNodeTimings(),
```
paulirish: since we're now emitting a diff shape at the end should we use diff property names to keep totally clear?   nodeTimeSpans  vs nodeTimings? eh

patrickhulce: I would prefer to rename the internal since that's the one that carries more than just timings data

paulirish:
wfm​

lighthouse-core/audits/uses-rel-preload.js
```diff
@@ -112,7 +112,6 @@ class UsesRelPreloadAudit extends Audit {
       const originalNode = originalNodesByRecord.get(node.record);
       const timingAfter = simulationAfterChanges.nodeTimings.get(node);
       const timingBefore = simulationBeforeChanges.nodeTimings.get(originalNode);
-      // @ts-ignore TODO(phulce): fix timing typedef
       const wastedMs = Math.round(timingBefore.endTime - timingAfter.endTime);
```
brendankenny: ha, looks like the `@ts-ignore` was also hiding the possibly undefined (to tsc) response from `nodeTimings.get()`. I'm good with casting or a runtime check. It would be nice to avoid `@ts-ignore` since it masks everything in the line, but it also wouldn't be the end of the world

patrickhulce: whoops yeah, added a check for it! 👍

lighthouse-core/lib/dependency-graph/simulator/simulator.js
patrickhulce: Which name would you prefer since you were the original ✍️ of `nodeTiming` ;) https://github.com/GoogleChrome/lighthouse/pull/3187#discussion_r136454948

PR #5358, opened 2018-05-25
core(image-aspect-ratio): loosen ratio check

### LINKED ISSUE #5356: Image aspect ratio audit is too strict
Running Lighthouse as a plugin in the latest version of Firefox (Lighthouse 3.0.0-beta.0).

   How can I correct this??  I have tried everything I can think of - and I don't understand why the image is not being displayed correctly.

Thanks,
Todd

RESULTS FROM LIGHTHOUSE SCAN (nrsmart.com):
Displays images with incorrect aspect ratio
Image display dimensions should match natural aspect ratio.
URL
Aspect Ratio (Displayed)
Aspect Ratio (Actual)
	
/assets/nrsmart-logo-longtext.png(www.nrsmart.com)
412 x 36 (11.44)
800 x 69 (11.59)

patrickhulce:
**Related Issues/PRs**
closes #5356

### DIFF
--- a/lighthouse-core/audits/image-aspect-ratio.js
+++ b/lighthouse-core/audits/image-aspect-ratio.js
@@ -14,7 +14,8 @@
 const Audit = require('./audit');
 
 const URL = require('../lib/url-shim');
-const THRESHOLD = 0.05;
+const THRESHOLD_IN_PERCENT = 0.05;
+const THRESHOLD_IN_PX = 2;
 
 /** @typedef {Required<LH.Artifacts.SingleImageUsage>} WellDefinedImage */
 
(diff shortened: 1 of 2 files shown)


lighthouse-core/audits/image-aspect-ratio.js
```diff
@@ -40,7 +41,10 @@ class ImageAspectRatio extends Audit {
     const url = URL.elideDataURI(image.src);
     const actualAspectRatio = image.naturalWidth / image.naturalHeight;
     const displayedAspectRatio = image.width / image.height;
-    const doRatiosMatch = Math.abs(actualAspectRatio - displayedAspectRatio) < THRESHOLD;
+
+    const targetDisplayHeight = image.width / actualAspectRatio;
+    const doRatiosMatch = Math.abs(actualAspectRatio - displayedAspectRatio) < THRESHOLD_PERCENT ||
```
brendankenny: is there value in maintaining the ratio difference check, then? It seems like it would either be less lenient than the pixel check or the images would be big enough that they would have the flexibility to change the image size to get under the limit.

I haven't completely worked out the problem, though, so I may be missing cases :)

brendankenny: a quick check with this branch shows that dropping it for the pixel check would still pass all our tests, at least :)

patrickhulce: yeah sgtm, let's just go with targetDisplayHeight

PR #5362, opened 2018-05-25
core(simulator): start nodes in observed order

patrickhulce:
~~I'm gonna let this one chill for a bit until I do more investigation, but basically I'm seeing graphs with some strange order and want to explore if this helps.~~

Big improvements to TTI tail error and not much change elsewhere. This should also make the smoketest savings much more reproducible/understandable.

image


patrickhulce:
alright I rebased this against the DNS change and things are still looking positive 👍

lighthouse-core/gather/computed/metrics/lantern-estimated-input-latency.js
```diff
@@ -94,7 +94,7 @@ class LanternEstimatedInputLatency extends LanternMetricArtifact {
       });
     }
 
-    return events;
```
paulirish: I don't follow the flow perfectly but it seems like the new sorting in simulator would make this one redundant. But no?
(if not, could this one be done earlier within a more common location?)

lighthouse-core/gather/computed/metrics/lantern-first-cpu-idle.js
```diff
@@ -56,6 +56,7 @@ class LanternFirstCPUIdle extends LanternInteractive {
       longTasks.push({start: timing.startTime, end: timing.endTime});
     }
 
+    longTasks.sort((a, b) => a.start - b.start);
```
paulirish: same question. can we do the sort earlier in a more generic location?

I want to avoid the situation where someone writes a new audit and forgets they gotta sort on their own.

lighthouse-core/gather/computed/metrics/lantern-estimated-input-latency.js
patrickhulce: it's redundant if the nodes all start in the order that they were queued (which used to always be true), but necessary in other cases. Even if it were true though, the fact that the `.entries` iterator on a map returns this in insertion order feels like an impl detail I wouldn't want important pieces of the codebase to rely on knowing since it bit me here once it changed slightly :/

I'm generally comfortable with the idea that consumers of a Map who need the results sorted going forward need to sort it themselves. As a safeguard though, I'm cool with *also* sorting the map before we return it based on start time. Sound good compromise?

patrickhulce:
@paulirish I rebased and added the double safeguard for insertion order 👍

patrickhulce:
All lantern-impacting stats remained the same after rebasing since there have been no other lantern changes since then

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -354,18 +365,23 @@ class Simulator {
...
+      nodeTimingEntries.push([node, {
         startTime: timing.startTime,
         endTime: timing.endTime,
         duration: timing.endTime - timing.startTime,
-      });
+      }]);
     }
 
-    return nodeTimings;
+    // Most consumers will want the entries sorted by startTime, so insert them in that order
```
brendankenny: ❤️

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
```diff
@@ -8,6 +8,7 @@
 const LanternFirstCPUIdle = require('../../../../gather/computed/metrics/lantern-first-cpu-idle.js'); // eslint-disable-line max-len
 const assert = require('assert');
 
+const LanternFCPUI = require('../../../../gather/computed/metrics/lantern-first-cpu-idle');
```
brendankenny:
```suggestion
const LanternFCpuI = require('../../../../gather/computed/metrics/lantern-first-cpu-idle');
```

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
```diff
@@ -28,4 +29,20 @@ describe('Metrics: Lantern TTFCPUI', () => {
...
+        [{type: 'cpu'}, {startTime: 0, endTime: 100, duration: 100}],
+        [{type: 'cpu'}, {startTime: 100, endTime: 200, duration: 100}],
+        [{type: 'cpu'}, {startTime: 500, endTime: 600, duration: 100}],
+        [{type: 'cpu'}, {startTime: 200, endTime: 300, duration: 100}],
+        [{type: 'cpu'}, {startTime: 400, endTime: 500, duration: 100}],
+      ]);
+
+      assert.equal(LanternFCPUI.getFirstCPUIdleWindowStart(tasks, 0), 700);
+    });
+  });
```
brendankenny: it might be harder to set up, but maybe a `Simulator` test that reads in artifacts and similarly asserts that `nodeTimings` are in `startTime` order (and wouldn't have been before this change)? I think our `stableSort` test for a trace has caught at least one regression that was accidentally reordering traces :)

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -181,6 +181,17 @@ class Simulator {
     });
   }
 
+  /**
+   * @param {Node[]} nodes
+   * @return {Node[]}
+   */
+  _sortNodesByStartTime(nodes) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -354,18 +365,23 @@ class Simulator {
     }
   }
 
+  /**
+   * @return {Map<Node, LH.Gatherer.Simulation.NodeTiming>}
+   */
   _computeFinalNodeTimings() {
-    /** @type {Map<Node, LH.Gatherer.Simulation.NodeTiming>} */
-    const nodeTimings = new Map();
+    /** @type {Array<[CpuNode | NetworkNode, LH.Gatherer.Simulation.NodeTiming]>} */
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -419,7 +435,7 @@ class Simulator {
     // loop as long as we have nodes in the queue or currently in progress
     while (nodesReadyToStart.size || nodesInProgress.size) {
       // move all possible queued nodes to in progress
-      for (const node of nodesReadyToStart) {
+      for (const node of this._sortNodesByStartTime(Array.from(nodesReadyToStart))) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
patrickhulce: never!!!! ⚔️

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
brendankenny: (or maybe there's an easier way to unit test it without calling private stuff; I didn't look very hard :)

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
brendankenny: ☠️😁😀 they need a "reject suggestion with prejudice" button

lighthouse-core/test/gather/computed/metrics/lantern-first-cpu-idle-test.js
patrickhulce: done

PR #5366, opened 2018-05-27
misc(externs): import crdp from root node_modules

denar90:
**Summary**
Import crdp directly from package.
Fix for using LH types.
If LH installed as dependency, then `vscode-chrome-debug-core` isn't and can't be found via `../node_modules/vscode-chrome-debug-core/lib/crdp/crdp` path. Then I as user adding it manually  to my package.json will be able to use `externs.d.ts` types.


denar90:
thx 💪

Issue #5383, opened 2018-05-30
Explore `jest` for unit tests


patrickhulce:
We discussed exploring moving to jest for our unit tests and leveraging snapshot testing for audit results.

We should be able to give this a try with next to no code changes since jest supports the same describe/it and see how it goes.

patrickhulce:
dunzo

PR #5386, opened 2018-05-30
tests: move cli/core to jest

patrickhulce:
in pursuit of #5383 

| Before | After |
| - | - |
| image | image |

only real code changes are before -> beforeAll, after -> afterAll, and moving .timeout

lots of noise from `eslint-env mocha` though :/

it also caught a bug that's not clear at all how the mocha tests were passing in CLI case where requiring in the yargs getFlags invokes process.exit (!!!!!). there's another psuedobug where we pass in 'Unknown' strings to our parseURL function which can throw

I'm not using jest's built-in expectations/coverage/anything yet to keep the diff small and easily revertable if folks don't like it.


paulirish:
Also this is mad cool. Excited for this. :D

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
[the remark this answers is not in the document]
patrickhulce: done

jest.config.js
```diff
@@ -0,0 +1,7 @@
...
+  testEnvironment: 'node',
+  testMatch: [
+    '**/lighthouse-core/**/*-test.js',
+    '**/lighthouse-cli/**/*-test.js',
+  ],
+};
```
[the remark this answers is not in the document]
patrickhulce: how important is this one? do we think 2s was a good feature? :)

in jest you do this via `jest.setTimeout` which requires a bootstrap file, and AFAICT cannot be done through config https://facebook.github.io/jest/docs/en/configuration.html

package.json
```diff
@@ -31,9 +31,8 @@
     "test": "yarn lint --quiet && yarn unit && yarn type-check && yarn diff:sample-json",
     "test-extension": "cd lighthouse-extension && yarn test",
     "test-viewer": "cd lighthouse-viewer && yarn pptr-test",
-
-    "unit-core": "mocha --reporter dot \"lighthouse-core/test/**/*-test.js\"",
-    "unit-cli": "mocha --reporter dot \"lighthouse-cli/test/**/*-test.js\"",
+    "unit-core": "jest \"lighthouse-core/\"",
```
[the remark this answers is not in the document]
patrickhulce: yeah not sure how we should do this one, since we reuse lots of commands, suggestions?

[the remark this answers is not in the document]
patrickhulce: I just went with a `:ci` copy of `unit-core` 👍

jest.config.js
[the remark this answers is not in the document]
paulirish: oic. hmmmmmmmm well nevermind then.

paulirish:
> coverage

for whatever reason this appears to drop our coverage % down quite a bit. not sure what's happening, though. https://codecov.io/gh/GoogleChrome/lighthouse/pull/5386

patrickhulce:
OK I sorted out the coverage problem thanks to  https://github.com/facebook/jest/issues/3190#issuecomment-354758036 😃 🎉

lighthouse-core/report/html/renderer/util.js
```diff
@@ -288,12 +288,16 @@ class Util {
    * @return {{file: string, hostname: string, origin: string}}
    */
   static parseURL(url) {
-    const parsedUrl = new URL(url);
-    return {
-      file: Util.getURLDisplayName(parsedUrl),
-      hostname: parsedUrl.hostname,
-      origin: parsedUrl.origin,
-    };
+    try {
```
brendankenny: why was this needed?

lighthouse-core/test/gather/computed/manifest-values-test.js
```diff
@@ -28,7 +28,7 @@ function noUrlManifestParser(manifestSrc) {
   return manifestParser(manifestSrc, EXAMPLE_MANIFEST_URL, EXAMPLE_DOC_URL);
 }
 
-/* eslint-env mocha */
+/* eslint-env jest */
```
brendankenny: delete this one?

package.json
```diff
@@ -31,20 +31,22 @@
...
     "core-unit": "yarn unit-core",
     "cli-unit": "yarn unit-cli",
     "viewer-unit": "yarn unit-viewer",
     "watch": "yarn unit-core --watch",
 
-    "unit:silentcoverage": "nyc --silent yarn unit && nyc report --reporter text-lcov > unit-coverage.lcov",
-    "coverage": "nyc yarn unit && nyc report --reporter html",
-    "smoke:silentcoverage": "nyc --silent yarn smoke && nyc report --reporter text-lcov > smoke-coverage.lcov",
-    "coverage:smoke": "nyc yarn smoke && nyc report --reporter html",
+    "unit:silentcoverage": "nyc --silent --clean yarn unit:ci && nyc report --reporter text-lcov > unit-coverage.lcov",
```
brendankenny: one unfortunate side of this (at least judging from that comment in the jest issue) is that subprocesses don't automatically collect coverage. I don't *believe* we do that in any unit tests, but it is a handy thing to lose

brendankenny: since jest automatically collects coverage with `--coverage`, are the lcov files written anywhere so we drop the `nyc` calls (for non-smoke tests) altogether?

package.json
```diff
@@ -31,20 +31,22 @@
...
+    "coverage": "nyc --clean yarn unit:ci && nyc report --reporter html",
+    "smoke:silentcoverage": "nyc --silent --clean yarn smoke && nyc report --reporter text-lcov > smoke-coverage.lcov",
+    "coverage:smoke": "nyc --clean yarn smoke && nyc report --reporter html",
```
brendankenny: FWIW I believe `--clean` is the default

lighthouse-core/report/html/renderer/util.js
patrickhulce: well we only *need* this or a fix on the `_renderTextURL` to be looser on its try/catch (we check instanceof TypeError), but it seemed like a good catch on passing invalid URLs to `parseURL`. We pass through several `'Unknown'` strings to this function which throw

lighthouse-core/test/gather/computed/manifest-values-test.js
patrickhulce: done

package.json
patrickhulce: if we use jest everywhere we won't need nyc for the unit tests at all and we can use the lcov/html files they generate in `./coverage`

right now, jest coverage output is just disabled and it lets nyc collect/combine everything since we still need it for the mocha cases

package.json
patrickhulce: removed

jest.config.js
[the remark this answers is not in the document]
wardpeet: we could go for eslint-env jest

patrickhulce:
so what's the story here, are we sold on jest? sticking with mocha? undecided? :)

wardpeet:
@patrickhulce I'm not hearing any people saying no that means we can go for Jest, right? 😛

jest.config.js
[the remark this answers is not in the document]
patrickhulce: ya @wardpeet I made that change 👍 :)

patrickhulce:
I like the way you think @wardpeet, I'm thinking auto-merge in 48 hours if no more objections? ;)

wardpeet:
They might not even notice 😄

paulirish:
The advantages that I'm jazzed about:

* Way faster locally. 
* Display is very concise. You can see failing tests immediately, but passing tests don't scroll the view.
* Update expectations snapshot with a keystroke sounds soooooo good.

The disadvantages that I'm bummed about:

* A bit slower on travis.


----------

On the whole though i think it'll be an improved experience.

package.json
```diff
@@ -93,6 +96,7 @@
     "gulp": "^3.9.1",
     "gulp-replace": "^0.5.4",
     "gulp-util": "^3.0.7",
+    "jest": "^23.0.1",
```
paulirish: can now use 23.2.0

patrickhulce: done

lighthouse-core/report/html/renderer/util.js
brendankenny: hmm, but why was it needed for this PR? If it's throwing shouldn't it have caused a failure before, too? Or is the problem something else.

patrickhulce: well again it isn't strictly *needed*, but seemed like a good idea anyhow.

full details:
in jest's jsdom env, the URL class throws an error that isn't instanceof TypeError
`_renderTextURL` which normally catches TypeErrors thrown by this method was then throwing

this isn't needed anymore because along the way I turned off default jsdom environment, but it seemed like a good idea to keep, it seems like you don't think so though so I'll revert :)

Issue #1685, opened 2017-02-10
RangeError: Invalid string length at onChunkRead


mourawaldson:
**Lighthouse Version**: 1.5.0
**Chrome Version**: 56.0.2924.87
**URL**: https://app.ft.com/index_page/home
**Error Message**: scripts/popup.js:153 RangeError: Invalid string length at onChunkRead (lighthouse-background.js:9579)
background.runLighthouseInExtension.catch.err @ scripts/popup.js:153
**Log**:
```
lighthouse-background.js:36680 status Disconnecting from browser... +6s
scripts/popup.js:153 RangeError: Invalid string length
    at onChunkRead (lighthouse-background.js:9579)
background.runLighthouseInExtension.catch.err @ scripts/popup.js:153
lighthouse-background.js:36680 <= Network.requestWillBeSent +6ms Object {documentURL: "https://app.ft.com/index_page/home", frameId: "18752.18", initiator: Object, loaderId: "18752.23", request: Object…}
```

Note: Since the log is too big I've only added the final lines, let me know if the full log is needed. Also, this was taken from the console log since the app just stopped running without raising any issue to the client level (which in my opinion should be done if any error occur).

brendankenny:
This is interesting. Happening from the trace being streamed back over the protocol.

I'm not able to reproduce, but the page load is timing out for me, so the trace will be fairly large. Maybe it's some kind OOM issue?

paulirish:
Okay fun one. @caseq and I looked into this.

I'm pretty sure it's coming from `  result += response.data;` in onChunkRead.

It's coming from V8 when a string gets to be too big.
What is the max size of a string in V8? 
Glad you asked. The answer is 2^28 bytes.. which is 256 megs.

---

Now you may think, maybe we should build this string in a fashion more efficient than `+=`? Good idea, bucko, and while that _may_ decrease the memory usage, it won't end up with a smaller string length and we'll still hit this bug.

So, if we want to handle a trace this big, we will need to parse it incrementally.

DevTools already has code to do that and we've already loaded all those modules so it shouldn't be too hard. :)

see `WebInspector.TimelineLoader` and `WebInspector.TextUtils.BalancedJSONTokenizer`.

mourawaldson:
@paulirish and @brendankenny,

When trying to reproduce now I get this:
```
status Evaluating: Estimated Input Latency +18s
[...]

PR #5387, opened 2018-05-30
core: faster saveTrace by streaming 500 events at a time

paulirish:
asset-saver-test.js goes from 12.5s to 7.9s.

cc #2593 #1685

### DIFF
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -196,7 +196,7 @@ async function prepareAssets(artifacts, audits) {
 /**
  * Generates a JSON representation of traceData line-by-line to avoid OOM due to very large traces.
  * COMPAT: As of Node 9, JSON.parse/stringify can handle 256MB+ strings. Once we drop support for
- * Node 8, we can 'revert' PR #2593
+ * Node 8, we can 'revert' PR #2593. See https://stackoverflow.com/a/47781288/89484
  * @param {LH.Trace} traceData
  * @return {IterableIterator<string>}
  */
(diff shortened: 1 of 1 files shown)


lighthouse-core/lib/asset-saver.js
```diff
@@ -211,9 +213,18 @@ function* traceJsonGenerator(traceData) {
     // Emit first item manually to avoid a trailing comma.
     const firstEvent = eventsIterator.next().value;
     yield `  ${JSON.stringify(firstEvent)}`;
+
+    let eventsRemaining = EVENTS_PER_ITERATION;
+    let eventsJSON = '';
     for (const event of eventsIterator) {
-      yield `,\n  ${JSON.stringify(event)}`;
+      eventsJSON += `,\n  ${JSON.stringify(event)}`;
+      if (--eventsRemaining === 0) {
```
patrickhulce: inline `--x ===` always makes me think too much, WDYT about a separate statement :)

lighthouse-core/lib/asset-saver.js
```diff
@@ -194,12 +194,14 @@ async function prepareAssets(artifacts, audits) {
 }
 
 /**
- * Generates a JSON representation of traceData line-by-line to avoid OOM due to
- * very large traces.
+ * Generates a JSON representation of traceData line-by-line to avoid OOM due to very large traces.
+ * COMPAT: As of Node 9, JSON.parse/stringify can handle 256MB+ strings. Once we drop support for
+ * Node 8, we can 'revert' PR #2593. See https://stackoverflow.com/a/47781288/89484
```
brendankenny: > Once we drop support for Node 8, we can 'revert' PR #2593

not sure if we actually want to anymore...we like the event-per-line formatting, right?

(We could consider shipping the `pretty-json-stringify` we just used in the snyk snapshot thing—no dependencies and 2.43KB before minification!—but it looks like using it on large traces will [depend on Node 9 as well](https://github.com/sergets/pretty-json-stringify/blob/5e81f3108c840d7ec7e967b02a0e4b6b17b98726/index.js#L20))

Issue #5397, opened 2018-06-01
Extension Error: Protocol error (Emulation.setDeviceMetricsOverride): Invalid


dhp-denero:
**Lighthouse Version**: 2.10.1.3000
**Lighthouse Commit**: 996c5d7ea19e1f7622fd4b2b2b85d6e60460a8d1
**Chrome Version**: 60.0.3112.78
**Initial URL**: https://www.sasmar.com/
**Error Message**: Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters
**Stack Trace**:
```
Error: Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters
    at Object.chrome.debugger.sendCommand.result [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17113:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

patrickhulce:
Old Chrome there friend, update to stable :)

Issue #5395, opened 2018-05-31
Extension Error: Protocol error (Emulation.setEmitTouchEventsForMouse): 'Emul


pash1:
**Lighthouse Version**: 2.10.1.3000
**Lighthouse Commit**: 996c5d7ea19e1f7622fd4b2b2b85d6e60460a8d1
**Chrome Version**: 61.0.3163.100
**Initial URL**: https://www.hopkinsmedicine.org/dev/xtest_pete/ModuleTest
**Error Message**: Protocol error (Emulation.setEmitTouchEventsForMouse): 'Emulation.setEmitTouchEventsForMouse' wasn't found
**Stack Trace**:
```
Error: Protocol error (Emulation.setEmitTouchEventsForMouse): 'Emulation.setEmitTouchEventsForMouse' wasn't found
    at Object.chrome.debugger.sendCommand.result [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17113:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

patrickhulce:
Old Chrome my friend upgrade to stable for Lighthouse :)

Issue #5305, opened 2018-05-22
Extension Error: Protocol error (Emulation.setDeviceMetricsOverride): Invalid


manishkumbhkar:
**Lighthouse Version**: 2.10.0.3002
**Lighthouse Commit**: 632eb32688cf3fd0b2dd0158a712fd1d12ddd021
**Chrome Version**: 60.0.3112.101
**Initial URL**: https://www.google.co.in/search?q=Emulation.setDeviceMetricsOverride)%3A+Invalid+parametersReport+ErrorOptions+Generate+report&oq=Emulation.setDeviceMetricsOverride)%3A+Invalid+parametersReport+ErrorOptions+Generate+report&aqs=chrome..69i57.51595j1j7&sourceid=chrome&ie=UTF-8
**Error Message**: Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters
**Stack Trace**:
```
Error: Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters
    at Object.chrome.debugger.sendCommand.result [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17101:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

patrickhulce:
You'll need to update your Chrome version, 60 is old :)

PR #5403, opened 2018-06-01
extension: update minimum Chrome version to 66

brendankenny:
**Summary**
This should hopefully get some of those seeing emulation protocol errors to update Chrome and get the debugger protocol methods we need.

Stable is m67 now, but that seemed a bit harsh (only happened three days ago), and 66 should be sufficient for our uses.

**Related Issues/PRs**
#5398, #5397, #5395, #5334, #5307, #5305...


brendankenny:
> nice! I think our docs already say stable right so no need there?

I believe so. I can't find any strings that seem to match, either :)

PR #5409, opened 2018-06-02
core(preload): remove blob protocol from preload audit

### LINKED ISSUE #5380: Preload key requests potential bug. Suggestion of preloading lighthouse blob: document
Lighthouse has suggested preloading key requests. The example appears to be very similar to the lighthouse report URL generated by the chrome extension. 

Report URL (Lighthouse 3.0.0-beta.0):

`blob:chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/37639f89-65c7-46a1-b15e-d5ba5f50916f`

Preload suggestion:

`blob:https://the-dots.com/337f5ca1-7a07-4ce3-bd72-985a13d750d3` 

alt

The URL I was testing was: [https://the-dots.com/users/search](https://the-dots.com/users/search)

Potentially a bug?

Tom

wardpeet:
**Summary**
Removes the blob protocol of preload requests like we do for data

**Related Issues/PRs**
Fixes #5380


lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -196,6 +196,36 @@ describe('Performance: uses-rel-preload audit', () => {
     });
   });
 
+  it(`shouldn't suggest preload for protocol blob`, () => {
```
patrickhulce: does this test fail without the fix? seems like it wouldn't have wastedMs anyhow

patrickhulce: maybe let's just leave a TODO to make creating fake lantern graphs easier, it's inline with the other tests in the file for now

lighthouse-core/audits/uses-rel-preload.js
```diff
@@ -159,7 +159,8 @@ class UsesRelPreloadAudit extends Audit {
     /** @type {Set<string>} */
     const urls = new Set();
     for (const networkRecord of criticalRequests) {
-      if (!networkRecord._isLinkPreload && networkRecord.protocol !== 'data') {
+      if (!networkRecord._isLinkPreload && networkRecord.protocol !== 'data'
```
paulirish: can we move this to an array in url-shim?  

`NON_NETWORK_PROTOCOLS` or something?

There's a few audits where we're checking these individually and i think we can generalize it.

PR #5410, opened 2018-06-03
docs(readme): lighthouse-ci

andreasonny83:
**Summary**
Add lighthouse-ci to the list of Lighthouse's related projects.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -282,6 +282,7 @@ Other awesome open source projects that use Lighthouse.
 * **[lighthouse-lambda](https://github.com/joytocode/lighthouse-lambda)** - Run Lighthouse on AWS Lambda with prebuilt stable desktop Headless Chrome.
 * **[lighthouse-security](https://github.com/voorhoede/lighthouse-security#readme)** - Run a set of security audits along with Lighthouse.
 * **[Garie](https://github.com/boyney123/garie)** — An open source tool for monitoring performance using Lighthouse,  PageSpeed Insights, [Prometheus](https://prometheus.io/), [Grafana](https://grafana.com/) and [Docker](https://www.docker.com/).
+* **[lighthouse-ci](https://github.com/andreasonny83/lighthouse-ci)** - Run Lighthouse and assert scores satisfy your custom thresholds.
 
 ## FAQ
 


readme.md
```diff
@@ -282,6 +282,7 @@ Other awesome open source projects that use Lighthouse.
 * **[lighthouse-lambda](https://github.com/joytocode/lighthouse-lambda)** - Run Lighthouse on AWS Lambda with prebuilt stable desktop Headless Chrome.
 * **[lighthouse-security](https://github.com/voorhoede/lighthouse-security#readme)** - Run a set of security audits along with Lighthouse.
 * **[Garie](https://github.com/boyney123/garie)** — An open source tool for monitoring performance using Lighthouse,  PageSpeed Insights, [Prometheus](https://prometheus.io/), [Grafana](https://grafana.com/) and [Docker](https://www.docker.com/).
+* **[lighthouse-ci](https://github.com/andreasonny83/lighthouse-ci)** - Integrate Google Lighthouse in your favourite Continuous Integration solution with zero configuration.
```
paulirish: wdyt about this description:

> Run Lighthouse and assert scores satisfy your custom thresholds.

paulirish: i will admit we're working on making some CI stuff for lighthouse so we will likely ask you for this npm module name soon. :) (i've sent you some of our plans)  

wdyt about renaming? something like `lighthouse-score-assert` seems cool.

--------------------

and lastly an alternative implementation could be something like...

```
lighthouse http://mysite.com --output=json | lighthouse-score-assert --performance=90 --pwa=100
```

this way you don't have to support all of `lighthouse` CLI's flags in your own app, but can still support people who change throttling settings, pass in config-path's, etc

this seems mostly reasonable with your project except for the chromeFlags you add on. but that's not too bad for users. we plan to publish a docker container that will take care of all those sandboxing issues etc and make running lighthouse in CI a straightforward affair.

andreasonny83:
Thanks for your feedback @paulirish, I updated the README according to your suggestion.

Unfortunately, I don't think npm allows renaming or deleting already published npm packages. So my project's name is still `lighthouse-ci` for now. 
If you know of any way of renaming npm packages, please let me know and I will be more than happy to change my package name.

I can see the value in passing the Lighthouse generated JSON report to my module just to process the information, however, I prefer to keep that logic inside my package responsibility for now and make sure the version is always in sync with the latest Lighthouse version.

PR #5412, opened 2018-06-03
extension(popup): integration test for popup

wardpeet:
**Summary**
The extension is broken by 92728c9725318cffd14a5918b83be084b88661fd when adding type checking.

:white_check_mark: Integration tests added for popup.html
:ambulance: fix popup js errors so it works again


brendankenny:
d'oh, thanks @wardpeet

wardpeet:
:tada: I'll add some more tests later on so we also test some button interactions :)

Issue #5424, opened 2018-06-04
environment variable CHROME_PATH must be set to executable of a build of Chromium


sillpa-polasani:
Jenkins-lighthouse setup instructions:
https://kitconcept.com/blog/continuous-performance-analysis-with-lighthouse-and-jenkins/

Issue: 
I am trying to setup lighthouse on jenkins machine.
I am getting following error:

Runtime error encountered: The environment variable CHROME_PATH must be set to executable of a build of Chromium version 54.0 or later.
Error: The environment variable CHROME_PATH must be set to executable of a build of Chromium version 54.0 or later.
at Object.linux (/xxxxx/node_modules/chrome-launcher/dist/chrome-finder.js:108:15)

#### Environment Information
* Affected Channels: 
* Lighthouse version: 2.9.4
* Node.js version: v6.10.0
* Operating System: linux

patrickhulce:
They seem to have left out the step where you install Chrome :)

Take a look at https://github.com/GoogleChrome/lighthouse/blob/master/docs/headless-chrome.md#running-lighthouse-using-headless-chrome

ray-kim-12:
Hi, I am getting the following error when trying to `yum install chromium-browser`

`Error: Package: chromium-browser-38.0.2125.111-1.el6.centos.x86_64 (chromium-el6)
           Requires: libudev.so.0()(64bit)`

and when I try to install it with `yum install chromium-browser --skip-broken`, I get the following:
`Packages skipped because of dependency problems:
    chromium-browser-38.0.2125.111-1.el6.centos.x86_64 from chromium-el6`

typing `rpm -Va --nofiles --nodigest`, produces no output, but the chromium-browser is not downloaded.

Red Hat Enterprise Linux 7.4

ray-kim-12:
Found the Solution:

same as @patrickhulce 's link, except `apt-get install chromium` for those using CentOS-7 (instead of 6)

For Red Hat Enterprise Linux 7.4: `yum install chromium`

ray-kim-12:
Typing `which chromium-browser` outputs:  `Chromium 65.0.3325.181 Fedora Project`
but still can't run lighthouse in my Jenkins as it fails with: `Unable to connect to Chrome  \n error Command failed with exit code 1.` .
More specifically fails when I run the script `node_modules/lighthouse/lighthouse-cli/index.js --output-path=./lighthouse-report.html --quiet --chrome-flags='--headless' APP_URL`

[...]

PR #5425, opened 2018-06-04
docs: update required chrome version

patrickhulce:
#5424 got me thinkin' :)


paulirish:
i wonder if we should just point to https://googlechromelabs.github.io/current-versions/ and say "stable"

patrickhulce:
Meh, I'm somewhat ambivalent. We don't really *need* to update it that often, but it is a pain to remember.

paulirish:
sure.

related: 
#5424 looks like it was with Chromium 65, which is recent enough to not have headless issues, id expect.
so i dont know if a bump to 66 really woulda solved it for him

patrickhulce:
#5424 got me thinking not because it had anything to do with his problem but because the bug contained the error message "The environment variable CHROME_PATH must be set to executable of a build of Chromium version 54.0 or later" which is definitely wrong at this point.

Issue #2736, opened 2017-07-25
Audit error: Required EventListeners gatherer encountered an error: Protocol error (DOM.resolveNode): No node with given id found


cyberixae:
Audit error: Required EventListeners gatherer encountered an error: Protocol error (DOM.resolveNode): No node with given id found

I get this error while auditing https://dwst.github.io/

patrickhulce:
Thanks for the report @cyberixae! I'm unable to reproduce on the latest version, but it seems like it'd be a good idea to not fail the whole audit if a single DOM node ceased existing :)

Future bug fixers can probably just add a catch to the inner loop here
https://github.com/GoogleChrome/lighthouse/blob/bc6783b6d0f4ab94b0b90c2058d7dffa8f592db8/lighthouse-core/gather/gatherers/dobetterweb/all-event-listeners.js#L103-L108

webcarrot:
Same error on: https://www.cyclosport.pl
Lighthouse 3.0.0-beta.0
Chrome:  67.0.3396.62
OS: Linux

brendankenny:
we don't have an `all-event-listeners.js` anymore, we moved to using console violations for this, so it's no longer an issue :)

PR #5427, opened 2018-06-05
core: Handle DOM.resolveNode errors

webcarrot:
**Summary**
Catch ```DOM.resolveNode``` errors.

**Related Issues/PRs**
Issue: #2736

### DIFF
--- a/lighthouse-core/lib/element.js
+++ b/lighthouse-core/lib/element.js
@@ -63,7 +63,8 @@ class Element {
           return null;
         }
         return this.driver.getObjectProperty(resp.object.objectId, propName);
-      });
+      })
+      .catch(() => null);
   }
 }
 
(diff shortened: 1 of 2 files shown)


webcarrot:
@paulirish 
It works for me but I'm not sure if it should not be handled elsewhere - I do not know this project and flows in it.

lighthouse-core/lib/element.js
```diff
@@ -63,7 +63,8 @@ class Element {
           return null;
         }
         return this.driver.getObjectProperty(resp.object.objectId, propName);
-      });
+      })
+      .catch(() => null);
```
brendankenny: FWIW I'm not sure we should fail silently here because the semantics end up being a bit weird (pretending the property doesn't exist when it's actually the element that doesn't). It seems like the caller should handle the exception (as it now does above).

patrickhulce: This is consistent with the current behavior of the method though, returning `null` when the object does not exist. If we keep throwing here, we should convert the above to throw and change the typedef

brendankenny: Ah, good point, I missed the first check. I wonder if we ever actually fail here then. The issue seems like it's been the event listener gatherer that's been failing. All good regardless.

PR #5428, opened 2018-06-05
[WIP] Generate HAR 

benschwarz:
**Work in progress** 

I sat down with @paulirish at JSConf EU to see how difficult it would be to generate a har now that lantern simulates hars. The code we spiked out is mostly commented out. Keeping it until I can understand if the current method is viable or not. 

**What does this aim to do?**

I figure the only difference between lantern and non-lantern is timing information. I believe we can grab `devtoolsLogs`, run it through a har generation process, then go through each har entry and update the timing information. 

**Open questions**

* Is that a viable plan? 
* The `nodeTimings` timing hash that comes from `pessimisticEstimate` has only three keys: `startTime`, `endTime` and `duration` - Can `duration` be added to `sendEnd` in the har? 
* Does the nodeTimings duration allow for related / chained / timing delays? Eg, if fonts.woff is loaded by style.css and that's loaded by index.html, will fonts.woff have a duration entry that can be used to shift it's timing along? Or is there more to it than that? 

**To be done?**

- [ ] Ensure this still works when we're not using simulated runs
- [ ] Remove uncommented code
- [ ] Write a test or two


patrickhulce:
> I figure the only difference between lantern and non-lantern is timing information. I believe we can grab devtoolsLogs, run it through a har generation process, then go through each har entry and update using the lantern timing information.

> Is that a viable plan?

Yep, that sounds pretty solid 👍 

> The nodeTimings timing hash that comes from pessimisticEstimate has only three keys: startTime, endTime and duration - Can duration be added to sendEnd in the har?

Not sure I understand this question. Duration here is just `endTime - startTime`, there's going to be some missing information (i.e. what was the breakdown like for response received, DNS resolved, handshake, etc). It'd make sense to start exposing that info once we've got the final story nailed down.


> Does the nodeTimings duration allow for related / chained / timing delays? Eg, if fonts.woff is loaded by style.css and that's loaded by index.html, will fonts.woff have a duration entry that can be used to shift it's timing along? Or is there more to it than that?

Not 100% sure I understand this question either. The startTime/endTime already capture the dependency information you're referring to. i.e. for the example chain the start/end times might look like

```
NetworkNode index.html -> {startTime: 0, endTime: 600, duration: 600}
NetworkNode style.css -> {startTime: 600, endTime: 900, duration: 300}
NetworkNode fonts.woff -> {startTime: 900, endTime: 1900, duration: 1000}
```

### Alternative Approach
One other approach to consider (which it seems like you may have explored already?) would be instead of creating a har from the real trace and mutating, you could run the node timings through `convertNodeTimingsToTrace ` and then just `Page.processMessage` those guys instead, you'll probably need to augment with a few extra events, so it might be more of a hassle than the first approach, but we can improve the `convertNodeTimingsToTrace` in the process so it'd be a win :)

benschwarz:
> Not sure I understand this question. Duration here is just endTime - startTime, there's going to be some missing information (i.e. what was the breakdown like for response received, DNS resolved, handshake, etc). It'd make sense to start exposing that info once we've got the final story nailed down.

Do you see all the response, dns, handshake etc metrics being added as part of your existing pull request @patrickhulce? Or is it more of a future thing?

patrickhulce:
> Do you see all the response, dns, handshake etc metrics being added as part of your existing pull request @patrickhulce? Or is it more of a future thing?

Definitely a future thing, DNS isn't even implemented yet. Once we have the regression testing turned on per commit, we'll be able to do that, then it makes sense to try to expose those more granular timings.

benschwarz:
> The startTime/endTime already capture the dependency information you're referring to. i.e. for the example chain the start/end times might look like

> ```
> NetworkNode index.html -> {startTime: 0, endTime: 600, duration: 600}
> NetworkNode style.css -> {startTime: 600, endTime: 900, duration: 300}
> NetworkNode fonts.woff -> {startTime: 900, endTime: 1900, duration: 1000}
> ```

Right, so given a non simulated `Network.responseReceived` event had timings like: 

```
{
  "requestTime": 98387.309451,
  "proxyStart": -1,
  "proxyEnd": -1,
  "dnsStart": 34.805,
  "dnsEnd": 34.816,
  "connectStart": 34.816,
  "connectEnd": 101.225,
  "sslStart": 49.383,
  "sslEnd": 101.219,
  "workerStart": -1,
  "workerReady": -1,
  "sendStart": 101.994,
  "sendEnd": 586.988,
  "pushStart": 0,
  "pushEnd": 0,
  "receiveHeadersEnd": 199.294
}
```

and the corresponding lantern timing had `startTime: 140, endTime: 500, duration: 360`, how would you suggest I manipulate the timings?

patrickhulce:
> and the corresponding lantern timing had startTime: 140, endTime: 500, duration: 360, how would you suggest I manipulate the timings?

Well we don't have almost all of this information. You'd set everything expect `requestTime` to -1, and then `.requestTime = firstNetworkRequest.timing.requestTime + lanternTiming.startTime / 1000`.

Then you find the corresponding `Network.loadingFinished` and update that `.timestamp = firstNetworkRequest.timing.requestTime + lanternTiming.endTime / 1000`.

Once we have all the other goodies we can worry about the DNS/SSL/Connect/Send start/end.

> So we'd have to manually create the network request started events, as well as a couple timeline metrics like ondomcontent loaded, right?

We're already creating them for the fake trace, so it wouldn't be an *insane* amount of work to flesh them out a bit for the devtools log. We don't have fake domContentLoaded timings right now anyway, so I'm not sure what you'd do with them in the regular trace case, but we'd eventually want to do this in the synthetic trace.

I think this would be the more sustainable approach in the long run since we'll need to invest in those scripts for better debugging anyhow, and they can become the single source of truth. If you want something that works *now* though then it might not be worth the investment for you and your build-HAR then patch (with just network requests) seems promising.

benschwarz:
@patrickhulce I _think_ I followed along with your suggestions for the "something that works now" solution, but the har that gets calculated ends up with negative request timings 😳

Do you spot anything glaringly incorrect about what I've done so far?

lighthouse-core/audits/generate-har.js
```diff
@@ -0,0 +1,137 @@
...
+
+      const [networkRecord, timings] = simulatedRequest;
+
+      return initialDocumentRequestTime + timings[timing] / 1000;
+    };
+
+    const events = devtoolsLog.map(event => {
+      switch (event.method) {
+        case 'Network.requestWillBeSent':
+          event.params.timestamp = shiftedRequestTimestamp(event, 'startTime');
```
patrickhulce: there's probably a few more you'll need to fix within the `.redirectResponse` property

lighthouse-core/audits/generate-har.js
```diff
@@ -0,0 +1,137 @@
...
+          break;
+        case 'Network.loadingFinished':
+          event.params.timestamp = shiftedRequestTimestamp(event, 'endTime');
+          break;
+        case 'Network.loadingFailed':
+          event.params.timestamp = shiftedRequestTimestamp(event, 'endTime');
+          break;
+        case 'Network.responseReceived':
+          event.params.response.timing.requestTime = shiftedRequestTimestamp(event, 'startTime');
```
patrickhulce: you'll probably want to zero-out/`-1`-out the other times in timing and fix the timestamp here too

lighthouse-core/audits/generate-har.js
```diff
@@ -0,0 +1,137 @@
...
+      const simulatedRequest = simulatedNetworkNodes.find(
+        ([key]) => key._id === networkEvent.params.requestId
+      );
+
+      const [networkRecord, timings] = simulatedRequest;
+
+      return initialDocumentRequestTime + timings[timing] / 1000;
+    };
+
+    const events = devtoolsLog.map(event => {
```
patrickhulce: I'm not entirely sure what HAR actually uses but you might want to either drop or fix the dataRecieved messages too

lighthouse-core/audits/generate-har.js
```diff
@@ -0,0 +1,137 @@
...
+
+    const initialDocumentRequestTime = initialDocumentResponse.params.response.timing.requestTime;
+
+    const simulatedNetworkNodes = Array.from(timings.entries()).filter(
+      ([key]) => key.type === 'network'
+    );
+
+    const shiftedRequestTimestamp = (networkEvent, timing) => {
+      const simulatedRequest = simulatedNetworkNodes.find(
+        ([key]) => key._id === networkEvent.params.requestId
```
patrickhulce: you might run into issues with redirected requests this way, the network nodes use the front-end style `:redirect` ids while raw events have duplicated request IDs

benschwarz:
No sweat @brendankenny. Do you foresee DNS timings coming into lantern any time in the near future?

PR #5435, opened 2018-06-05
tests(lantern): add lantern regression test scripts

patrickhulce:
Adds lantern regression testing against ~100 URLs to travis (only when lantern files are affected).

A few things this PR highlighted:
1. Our scripts are getting sophisticated and probably a candidate for some refactoring into libs and such
2. We duplicate a decent bit of logic in our scripts, perhaps moving more to js and aforementioned refactoring would help.
3. Names are hard and suggestions welcome. It felt weird to call the lantern expectations file we checkin a "golden" file since it's not really golden, it's where we are right now and WPT results are the golden expectations we're trying to reach. I sorta went with `computed` vs. `expected` but these aren't great and not 100% consistent.


patrickhulce:
I think this is the last thing standing in my way before #5362 #5482 #5483 etc 😃

paulirish:
>  I sorta went with computed vs. expected but these aren't great and not 100% consistent.

what do each of these groups represent? i'm not sure between WPT data, 'golden' lantern data, and freshly computed lantern data.. what we're dealing with here.

patrickhulce:
> what do each of these groups represent? i'm not sure between WPT data, 'golden' lantern data, and freshly computed lantern data.. what we're dealing with here.


we have a few main things

1. `lantern-data/lantern-expectations.json` this contains URLs and the **WPT metric values**, this is our target against which accuracy is judged
2. `lantern-data/lantern-computed.json` this contains the URLs and **computed Lantern values for the current build**
3. `core/test/fixtures/lantern-expectations.json` this contains the URLs and **computed Lantern values for master** which is diff'd similar to our golden LHR. this is the one that is most in need of a name change, but I'm not sure what to call it :)

paulirish:
perfect. this helps enormously. thanks!

some name ideas below:

> 1. `lantern-data/lantern-expectations.json` this contains URLs and the **WPT metric values**, this is our target against which accuracy is judged

`wpt-baseline-data`

> 2. `lantern-data/lantern-computed.json` this contains the URLs and **computed Lantern values for the current build**

`lantern-fresh-data`

> 3. `core/test/fixtures/lantern-expectations.json` this contains the URLs and **computed Lantern values for master** which is diff'd similar to our golden LHR. this is the one that is most in need of a name change, but I'm not sure what to call it :)

`lantern-master-data` 

-----

I'm also fine with switching the order of baseline/fresh/master and wpt/lantern.

lighthouse-core/scripts/test-lantern.sh
```diff
@@ -0,0 +1,34 @@
...
+DIRNAME="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+LH_ROOT="$DIRNAME/../.."
+cd $LH_ROOT
+
+set -e
+
+# Testing lantern can be expensive, we'll only run the tests if we touched files that affect the simulations.
+CHANGED_FILES=""
+if [[ "$CI" ]]; then
+  CHANGED_FILES=$(git --no-pager diff --name-only $TRAVIS_COMMIT_RANGE)
```
brendankenny: > `$TRAVIS_COMMIT_RANGE`

cooool

patrickhulce:
@paulirish I oversimplified the `lantern-expectations` one a bit which I only realized once `wpt-baseline-data` sounded wrong to me. 

It contains the list of all the URLs we should compare, the WPT metric values, and the paths to the devtools log and trace of the unthrottled runs. Just WPT feels a bit weird in that scenario since it's an index of the unthrottled data around it. If I were going verbose I'd probably say something like 
`lantern-3g-golden-expectations-index.json`
`lantern-3g-HEAD-computed-values.json`
`lantern-3g-master-computed-values.json`

lighthouse-core/scripts/lantern/diff-expectations.js
```diff
@@ -0,0 +1,58 @@
...
+  try {
+    execSync(`git --no-pager diff --color=always --no-index ${TMP_EXPECTED} ${TMP_COMPUTED}`);
+    console.log('✅  PASS    No changes between expected and computed!');
+  } catch (err) {
+    console.log('❌  FAIL    Changes between expected and computed!\n');
+    console.log(err.stdout.toString());
+    exitCode = 1;
+  }
+} finally {
+  fs.unlinkSync(TMP_COMPUTED);
```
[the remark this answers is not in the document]
paulirish: maybe these should have `-for-diff` in the filename? (if i'm understanding why they're separate)

lighthouse-core/scripts/lantern/update-expectations.js
```diff
@@ -0,0 +1,44 @@
...
+const fs = require('fs');
+const path = require('path');
+const execFileSync = require('child_process').execFileSync;
+const prettyJSONStringify = require('pretty-json-stringify');
+
+const INPUT_PATH = process.argv[2] || './lantern-data/lantern-expectations.json';
+const EXPECTATIONS_PATH = path.resolve(process.cwd(), INPUT_PATH);
+const EXPECTATIONS_DIR = path.dirname(EXPECTATIONS_PATH);
+const COMPUTED_PATH = path.join(EXPECTATIONS_DIR, 'lantern-computed.json');
+const RUN_ALL_SCRIPT_PATH = path.join(__dirname, 'run-all-expectations.js');
```
paulirish: seems like this isn't _running expectations_ but "collecting computed values". wdyt?

lighthouse-core/scripts/lantern/download-traces.sh
```diff
@@ -1,13 +1,21 @@
 #!/bin/bash
 
+set -e
+
 DIRNAME="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 LH_ROOT_PATH="$DIRNAME/../../.."
 cd $LH_ROOT_PATH
 
+if [[ -f lantern-data/lantern-expectations.json ]] && ! [[ "$FORCE" ]]; then
+  echo "Lantern data already detected, done."
```
paulirish: within the `lantern-data/lantern-expectations.json` file....

image

can these metrics be within an object? like `wptObserved`?

lighthouse-core/scripts/lantern/evaluate-results.js
```diff
@@ -8,16 +8,19 @@
...
+const fs = require('fs');
 const path = require('path');
 
 const GOOD_ABSOLUTE_THRESHOLD = 0.2;
 const OK_ABSOLUTE_THRESHOLD = 0.5;
 
 const GOOD_RANK_THRESHOLD = 0.1;
 
-if (!process.argv[2]) throw new Error('Usage $0 <computed summary file>');
+const COMPUTATIONS_INPUT_ARG = process.argv[2] || './lantern-data/lantern-computed.json';
```
paulirish: one of these things doesn't look like the others.  COMPUTATIONS?

patrickhulce:
@paulirish how you feelin' about this now :)

patrickhulce:
let's do it! 🎉

PR #5439, opened 2018-06-06
core(a11y): add back html/tags/impact

### LINKED ISSUE #5402: Lighthouse 3 Regression in Reporting Data
Run Lighthouse 2.9.4 on any website using the cli.

For my example I am running a simple report on http://cats.com. 

To simply the example, you can run this, but it is not necessary:

```js
'use strict';

/**
 * Config file for running A11y audits.
 */
module.exports = {
  settings: {
    onlyAudits: ['color-contrast'],
  },
};
```

If there is a contrast error, Lighthouse 2.9 provides information about the current failing element. 
 For my example I am showing the contrast ratio.  Notice in `failureSummary` it calculates the current ratio and advises what contrast ratio needs met to pass.

> "failureSummary": "Fix any of the following:\n  **Element has insufficient color contrast of 2.32 (foreground color: #ffffff, background color: #aaaaaa, font size: 9.8pt, font weight: normal). Expected contrast ratio of 4.5:1**",

Example result:

```json
  "extendedInfo": {
        "value": {
          "id": "color-contrast",
          "impact": "serious",
          "tags": [
            "cat.color",
            "wcag2aa",
            "wcag143"
          ],
          "description": "Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds",
          "help": "Elements must have sufficient color contrast",
          "helpUrl": "https://dequeuniversity.com/rules/axe/3.0/color-contrast?application=axeAPI",
          "nodes": [
            {
              "impact": "serious",
              "html": "<input type=\"submit\" value=\"Subscribe\" name=\"subscribe\" id=\"mc-embedded-subscribe\" class=\"button\">",
              "target": [
                "#mc-embedded-subscribe"
              ],
              "failureSummary": "Fix any of the following:\n  Element has insufficient color contrast of 2.32 (foreground color: #ffffff, background color: #aaaaaa, font size: 9.8pt, font weight: normal). Expected contrast ratio of 4.5:1",
              "path": "0,HTML,2,BODY,0,DIV,0,FORM,0,DIV,4,DIV,0,INPUT",
              "snippet": "<input type=\"submit\" value=\"Subscribe\" name=\"subscribe\" id=\"mc-embedded-subscribe\" class=\"button\">"
            }
          ]
        }
      },
```
[Full Gist here](https://gist.github.com/joelhsmith/5e5af940060b53a680b6bdca6e58eb44)

Yay! Lots of important meta.  

#### But that meta has been removed in Lighthouse 3

Run Lighthouse 3.0.0-alpha.2 on any website. To simplify, you can run my Gist above.

```json
  "details": {
        "type": "table",
        "headings": [
          {
            "key": "node",
            "itemType": "node",
            "text": "Failing Elements"
          }
        ],
        "items": [
          {
            "node": {
              "type": "node",
              "selector": "#mc-embedded-subscribe",
              "path": "0,HTML,2,BODY,0,DIV,0,FORM,0,DIV,4,DIV,0,INPUT",
              "snippet": "<input type=\"submit\" value=\"Subscribe\" name=\"subscribe\" id=\"mc-embedded-subscribe\" class=\"button\">"
            }
          }
        ]
      }
```

[View Full Gist here](https://gist.github.com/joelhsmith/32c590c2ec609c55594dcd7446a89805)

All gone.

This seems like a major regression. I was using that data to consume the json and create markdown files with all the data, not just what is shown on the default html Lighthouse report.

I know it says that `extendedInfo` has changed to `details` in the migration doc. But this is more than renaming something. This is removing major functionality.

patrickhulce:
**Related Issues/PRs**
closes #5402

### DIFF
--- a/lighthouse-core/audits/accessibility/axe-audit.js
+++ b/lighthouse-core/audits/accessibility/axe-audit.js
@@ -45,7 +45,7 @@ class AxeAudit extends Audit {
           selector: Array.isArray(node.target) ? node.target.join(' ') : '',
           path: node.path,
           snippet: node.html || node.snippet,
-          axeFailureSummary: node.failureSummary,
+          explanation: node.failureSummary,
         }),
       }));
     }
--- a/lighthouse-core/audits/accessibility/axe-audit.js
+++ b/lighthouse-core/audits/accessibility/axe-audit.js
@@ -33,6 +33,8 @@ class AxeAudit extends Audit {
 
     const violations = artifacts.Accessibility.violations || [];
     const rule = violations.find(result => result.id === this.meta.name);
+    const impact = rule && rule.impact;
+    const tags = rule && rule.tags;
 
     /** @type {Array<{node: LH.Audit.DetailsRendererNodeDetailsJSON}>}>} */
     let items = [];
@@ -42,7 +44,8 @@ class AxeAudit extends Audit {
           type: 'node',
           selector: Array.isArray(node.target) ? node.target.join(' ') : '',
           path: node.path,
-          snippet: node.snippet,
+          snippet: node.html || node.snippet,
+          axeFailureSummary: node.failureSummary,
         }),
       }));
     }
@@ -56,7 +59,7 @@ class AxeAudit extends Audit {
       extendedInfo: {
         value: rule,
       },
-      details: Audit.makeTableDetails(headings, items),
+      details: {...Audit.makeTableDetails(headings, items), impact, tags},
     };
   }
 }
--- a/lighthouse-core/test/results/sample_v2.json
+++ b/lighthouse-core/test/results/sample_v2.json
@@ -1225,7 +1225,7 @@
               "selector": "div > h2",
               "path": "3,HTML,1,BODY,0,DIV,0,H2",
               "snippet": "<h2>Do better web tester page</h2>",
-              "axeFailureSummary": "Fix any of the following:\n  Element has insufficient color contrast of 1.32 (foreground color: #ffc0cb, background color: #eeeeee, font size: 18.0pt, font weight: bold). Expected contrast ratio of 3:1"
+              "explanation": "Fix any of the following:\n  Element has insufficient color contrast of 1.32 (foreground color: #ffc0cb, background color: #eeeeee, font size: 18.0pt, font weight: bold). Expected contrast ratio of 3:1"
             }
           },
           {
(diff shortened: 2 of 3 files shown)


lighthouse-core/test/results/sample_v2.json
```diff
@@ -1450,17 +1488,29 @@
...
             }
           },
           {
             "node": {
               "type": "node",
               "selector": "body > a[href$=\"mailto:inbox@email.com\"]",
               "path": "3,HTML,1,BODY,29,A",
-              "snippet": "<a href=\"mailto:inbox@email.com\" target=\"_blank\">"
+              "snippet": "<a href=\"mailto:inbox@email.com\" target=\"_blank\"></a>",
+              "axeFailureSummary": "Fix all of the following:\n  Element is in tab order and does not have accessible text\n\nFix any of the following:\n  Element does not have text that is visible to screen readers\n  aria-label attribute does not exist or is empty\n  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty or not visible\n  Element's default semantics were not overridden with role=\"presentation\"\n  Element's default semantics were not overridden with role=\"none\""
```
patrickhulce: @joelhsmith it seems like in several cases these can be static

Do you have a sense of which cases they're needed per-node? Is it just color-contrast?

joelhsmith: Looking through [axe-core](https://github.com/dequelabs/axe-core/tree/020fd150ed0bf0702a89d19712bd38ea0c3817d8/lib/checks), I found these:

aria-allowed-attr
aria-required-children
aria-required-parent
aria-valid-attr-value
aria-valid-attr
color-contrast
fieldset
group-labelledby
duplicate-id
non-empty-if-present

I would not be surprised in the future if even more of them became contextual in the future. Plus people can write their own checks which also might be contextual.

lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -56,7 +59,7 @@ class AxeAudit extends Audit {
       extendedInfo: {
         value: rule,
       },
-      details: Audit.makeTableDetails(headings, items),
+      details: {...Audit.makeTableDetails(headings, items), impact, tags},
```
brendankenny: should `impact` and `tags` be nested or something? I don't really have an opinion, this seems fine and nesting might not be any better, but it may be putting us on a path toward `details` becoming the new grab bag :)

joelhsmith: FWIW, I like it where it is. 

If it were nested, I can't think of what it would be called. On Deque's site "impact" and "tags" meta is extrapolated to create the "[Compliance Data & User Impact](https://dequeuniversity.com/rules/axe/2.2/color-contrast?application=lighthouse)" column. "complianceUserImpact:" seems a little long anyway ;-)

lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -42,7 +44,8 @@ class AxeAudit extends Audit {
           type: 'node',
           selector: Array.isArray(node.target) ? node.target.join(' ') : '',
           path: node.path,
-          snippet: node.snippet,
+          snippet: node.html || node.snippet,
+          axeFailureSummary: node.failureSummary,
```
paulirish: how about just `summary`

patrickhulce: discussed in person `explanation` to match with our audit-level naming 👍

PR #5442, opened 2018-06-06
core(user-timings): add back startTime

### LINKED ISSUE #5418: lighthouse 3 regression in user-timings details
## Bug report

Lighthouse v3 does not provide `startTime` value for user timing measures. This info is the part of `extendedInfo` but it was removed.

1. `lighthouse https://treo.sh`
2. Check JSON report:

```json
  "user-timings": {
      "id": "user-timings",
      "title": "User Timing marks and measures",
      "description": "Consider instrumenting your app with the User Timing API to create custom, real-world measurements of key user experiences. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/user-timing).",
      "score": null,
      "scoreDisplayMode": "informative",
      "rawValue": false,
      "displayValue": "2",
      "details": {
        "type": "table",
        "headings": [{
          "key": "name",
          "itemType": "text",
          "text": "Name"
        }, {
          "key": "timingType",
          "itemType": "text",
          "text": "Type"
        }, {
          "key": "time",
          "itemType": "ms",
          "granularity": 0.01,
          "text": "Time"
        }],
        "items": [{
          "name": "home did mount",
          "timingType": "Measure",
          "time": 49
        }, {
          "name": "boot",
          "timingType": "Mark",
          "time": 716.685
        }]
      }
    }
```

The report contains only `time` value.

The report should contain `startTime` value for measures.

#### Environment Information
* Affected Channels: All (report)
* Lighthouse version: 3.0.0-beta.0

**Related issues**

#5402

patrickhulce:
**Related Issues/PRs**
closes #5418

### DIFF
--- a/lighthouse-core/audits/user-timings.js
+++ b/lighthouse-core/audits/user-timings.js
@@ -118,6 +118,7 @@ class UserTimings extends Audit {
         const time = item.isMark ? item.startTime : item.duration;
         return {
           name: item.name,
+          startTime: item.startTime,
           timingType: item.isMark ? 'Mark' : 'Measure',
           time,
         };
--- a/lighthouse-core/audits/user-timings.js
+++ b/lighthouse-core/audits/user-timings.js
@@ -142,11 +142,20 @@ class UserTimings extends Audit {
 
       const details = Audit.makeTableDetails(headings, tableRows);
 
+      /** @type {LH.Audit.Product['displayValue']} */
+      let displayValue;
+      if (userTimings.length) {
+        displayValue = [
+          userTimings.length === 1 ? '%d user timing' : '%d user timings',
+          userTimings.length,
+        ];
+      }
+
       return {
         // mark the audit as notApplicable if there were no user timings
         rawValue: userTimings.length === 0,
         notApplicable: userTimings.length === 0,
-        displayValue: userTimings.length ? `${userTimings.length}` : '',
+        displayValue,
         extendedInfo: {
           value: userTimings,
         },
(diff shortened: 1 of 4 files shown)


lighthouse-core/audits/user-timings.js
```diff
@@ -118,6 +118,7 @@ class UserTimings extends Audit {
         const time = item.isMark ? item.startTime : item.duration;
         return {
           name: item.name,
+          startTime: item.startTime,
```
paulirish: can you assert this in user-timing-test?

lighthouse-core/audits/user-timings.js
```diff
@@ -118,6 +118,7 @@ class UserTimings extends Audit {
...
           timingType: item.isMark ? 'Mark' : 'Measure',
           time,
```
paulirish: just noticed.. this name seems pretty weak. duration? timeInMs?

lighthouse-core/audits/user-timings.js
patrickhulce: done

lighthouse-core/audits/user-timings.js
patrickhulce: well that name is weak because it's either or and just for display, I added duration separately, but programmatic folks should probably just ignore `time`

maybe update to `timeToDisplay`?

paulirish: it's always been a bit odd that we share a column for two diff types of values

how about we add a second column, then we can be really clear with `Start time` and `Duration`. wdyt?

patrickhulce:
> add additional column? does it mess too much with our types to have undefined/null durations?

done

PR #5449, opened 2018-06-07
fix: short_name shouldn't be required if the name is short

### LINKED ISSUE #2698: short_name shouldn't be required

JoelEinbinder:
fixes #2698


JoelEinbinder:
When searching for how to name this, apparently this PR already exists as https://github.com/GoogleChrome/lighthouse/pull/4860

PR #5451, opened 2018-06-08
core: remove dependency on devtools-frontend NetworkRequest

patrickhulce:
title speaks for itself, big change :)

part of the broader goal of removing our dependency on devtools-frontend

tests should be passing now! 🎉  I want to do more comparisons with my lantern traces and land our safety checks before this one


typings/web-inspector.d.ts
```diff
@@ -31,28 +29,33 @@ declare global {
...
+      initiatorRequest(): NetworkRequest | undefined;
+      _responseHeaders?: HeaderValue[];
 
       _fetchedViaServiceWorker?: boolean;
-      _frameId: Crdp.Page.FrameId;
+      _frameId?: Crdp.Page.FrameId;
       _isLinkPreload?: boolean;
-      initiatorRequest(): NetworkRequest | null;
-      redirects?: NetworkRequest[];
+    }
```
brendankenny: shouldn't we be able to get rid of this and use the class definition?

lighthouse-core/lib/web-inspector.js
```diff
@@ -317,5 +317,7 @@ module.exports = (function() {
   // Restore setImmediate, see comment at top.
   global.setImmediate = _setImmediate;
 
+  // TODO(phulce): replace client requires to not need this
+  WebInspector.resourceTypes = require('../../third-party/devtools/ResourceType').TYPES;
```
brendankenny: can more be removed from this file?

patrickhulce:
> is it more removing dependency on devtools-frontend to create network records?

er, yes! depends on what we're calling a network record, I guess that's somewhat ambiguous

lighthouse-core/lib/web-inspector.js
patrickhulce: eventually yes!

typings/web-inspector.d.ts
patrickhulce: yes, that jump seemed a bit bigger in the same go since I'm using this to typecheck the other class, but we definitely will be able to

lighthouse-core/test/fixtures/wikipedia-redirect.critical-request-chains.json
```diff
@@ -1,11 +1,11 @@
 {
-  "33097.1:redirected.0": {
```
patrickhulce: open question if we want to match exactly the old API instead?

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
```diff
@@ -165,6 +166,7 @@ class UnusedBytes extends Audit {
       const networkNode = /** @type {NetworkNode} */ (node);
       const originalTransferSize = originalTransferSizes.get(networkNode.record.requestId);
       if (originalTransferSize === undefined) return;
+      networkNode.record.transferSize = originalTransferSize;
       networkNode.record._transferSize = originalTransferSize;
```
paulirish: should we just use one of these? this PR is switching to use the other already. 
Also happy to fix the `_prop` use in a followup to minimize the diff.

lighthouse-core/audits/uses-rel-preconnect.js
```diff
@@ -90,7 +91,7 @@ class UsesRelPreconnectAudit extends Audit {
           // filter out all resources where timing info was invalid
           !UsesRelPreconnectAudit.hasValidTiming(record) ||
           // filter out all resources that are loaded by the document
-          record.initiatorRequest() === mainResource ||
+          record._initiator.url === mainResource.url ||
```
paulirish: these aren't totally equivalent, fwiw. but the difference may not matter. And I suppose devtools was making this assumption previously. :p

lighthouse-core/lib/network-recorder.js
```diff
@@ -34,6 +38,10 @@ class NetworkRecorder extends EventEmitter {
     );
   }
 
+  getRecords() {
+    return this._records.slice();
```
paulirish: Array.from() ?

lighthouse-core/lib/network-recorder.js
```diff
@@ -16,12 +17,15 @@ const IGNORED_NETWORK_SCHEMES = ['data', 'ws'];
...
+  constructor() {
     super();
 
-    this._records = recordArray;
+    /** @type {NetworkRequest[]} */
+    this._records = [];
+    /** @type {Map<string, NetworkRequest>} */
+    this._recordsById = new Map();
+
     this.networkManager = NetworkManager.createWithFakeTarget();
```
paulirish: do we still need this around? followup?

lighthouse-core/lib/network-recorder.js
```diff
@@ -259,29 +294,87 @@ class NetworkRecorder extends EventEmitter {
...
+    const records = networkRecorder.getRecords();
+
+    // create a map of all the records by URL to link up initiator
+    const recordsByURL = new Map();
+    for (const record of records) {
+      if (recordsByURL.has(record.url)) continue;
+      recordsByURL.set(record.url, record);
+    }
+
+    // set the initiator and redirects array
```
paulirish: at first glance, it seems odd that this happens only when we're hydrating the records from logs. what about when records are created via `this._networkStatusMonitor.dispatch(event);` in driver?

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,244 @@
...
+ * @see https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/sdk/NetworkManager.js
+ */
+
+const URL = require('./url-shim');
+const resourceTypes = require('../../third-party/devtools/ResourceType').TYPES;
+
+module.exports = class NetworkRequest {
+  constructor() {
+    this.requestId = '';
+    this._requestId = '';
```
paulirish: as mentioned, i'd love if a followup normalized these paths.

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,244 @@
...
+    this._mimeType = '';
+    this.priority = () => /** @type {LH.Crdp.Network.ResourcePriority} */ ('Low');
+    this.initiatorRequest = () => /** @type {NetworkRequest|undefined} */ (undefined);
+    this._responseHeaders = /** @type {LH.WebInspector.HeaderValue[]} */ ([]);
+
+    this._fetchedViaServiceWorker = false;
+    this._frameId = /** @type {string|undefined} */ ('');
+    this._isLinkPreload = false;
+
+    // Make sure we're compitable with NetworkRequest
```
paulirish: compatible

paulirish: which NetworkRequest ? modern SDK.NetworkRequest ? or the one within the frontend lighthouse currently depends on.

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,244 @@
...
+
+    this._fetchedViaServiceWorker = !!response.fromServiceWorker;
+
+    if (response.timing) this._recomputeTimesWithResourceTiming(response.timing);
+  }
+
+  /**
+   * @param {LH.Crdp.Network.ResourceTiming} timing
+   */
+  _recomputeTimesWithResourceTiming(timing) {
```
paulirish: add a comment referencing `SDK.NetworkRequest.timing` setter and link to https://github.com/ChromeDevTools/devtools-frontend/blob/56a99365197\/front_end/sdk/NetworkRequest.js#L485-L502

lighthouse-core/test/fixtures/wikipedia-redirect.critical-request-chains.json
paulirish: i dun care. this wfm.

lighthouse-core/lib/network-recorder.js
```diff
@@ -259,29 +294,87 @@ class NetworkRecorder extends EventEmitter {
...
     switch (event.method) {
-      case 'Network.requestWillBeSent': return this.onRequestWillBeSent(event.params);
-      case 'Network.requestServedFromCache': return this.onRequestServedFromCache(event.params);
-      case 'Network.responseReceived': return this.onResponseReceived(event.params);
-      case 'Network.dataReceived': return this.onDataReceived(event.params);
-      case 'Network.loadingFinished': return this.onLoadingFinished(event.params);
-      case 'Network.loadingFailed': return this.onLoadingFailed(event.params);
-      case 'Network.resourceChangedPriority': return this.onResourceChangedPriority(event.params);
-      default: return;
+      case 'Network.requestWillBeSent':
```
paulirish: revert the ws change

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,244 @@
...
+
+  _updateResponseReceivedTimeIfNecessary() {
+    this._responseReceivedTime = Math.min(this.endTime, this._responseReceivedTime);
+  }
+
+  /**
+   * @param {LH.Crdp.Network.Headers} headersMap
+   * @return {Array<LH.WebInspector.HeaderValue>}
+   */
+  static _headersMapToHeadersArray(headersMap) {
```
paulirish: headersDict or headersObj or headersHash seem fine..

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,244 @@
...
+    const result = [];
+    for (const name of Object.keys(headersMap)) {
+      const values = headersMap[name].split('\n');
```
paulirish: which method in sdk.networkrequest is this disguised as? `_computeHeaderValue`

lighthouse-core/lib/network-recorder.js
patrickhulce: haha, man between you and dgozman I'm gonna lose myself! 😆 

done

lighthouse-core/lib/network-recorder.js
patrickhulce: that one's totally safe to go right now actually 👍

lighthouse-core/lib/network-recorder.js
patrickhulce: yeah it's a bit weird, AFAICT we actually only ever use the ones created by the recordsFromLogs, the ones created incrementally are just for the status notifications

https://github.com/GoogleChrome/lighthouse/blob/996c5d7ea19e1f7622fd4b2b2b85d6e60460a8d1/lighthouse-core/gather/gather-runner.js#L301
(and obvs the computed artifact uses this)

long-term though I hope to eliminate the initiatorRequest logic here entirely and move it to the page-dependency-graph since it's just CRC-based ones that need it right now

redirects can also move to direct usage of source/destination which I think becomes more explicit and clear

lighthouse-core/lib/network-request.js
patrickhulce: yeah for sure 👍

lighthouse-core/lib/network-request.js
patrickhulce: lol whoops, at least it was one of the more amusing ones :)

patrickhulce: old one we currently depend on, updated the comment

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-request.js
patrickhulce: done

lighthouse-core/lib/network-request.js
patrickhulce: done

lighthouse-core/lib/network-request.js
patrickhulce: Not even disguised verbatim s/in/of/ ;)
https://github.com/ChromeDevTools/devtools-frontend/blob/3415ee28e86a3f4bcc2e15b652d22069938df3a6/front_end/sdk/NetworkManager.js#L285-L297

added a comment

lighthouse-core/audits/uses-rel-preconnect.js
patrickhulce: yeah I was kinda shocked that the whole initiatorRequest logic is literally just first one with matching URL and manager 😮

lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js
patrickhulce: yeah eliminated `_transferSize`, the dance was necessary for transferSize since the weird way the getter interacts with our simulation but the rest will be followup 👍

patrickhulce:
Alright I think I'm comfortable taking this out of WIP, tests look good, Paul's initial feedback addressed, added some better coverage around the redirect cases

I'm still going to put this through the lantern wringer before being 100% cool to merge 👍

patrickhulce:
so more fun discoveries as part of this, mixed-content has been broken for some time :) #5127 removed the URL gatherer, so the mixed content config tried to treat the `url` npm module as a gatherer and config craps out.

on the plus side it seems that chrome changes to interception also broke it (it seems to think example.com is not upgradeable?) so we didn't break anything very useful? 🤷‍♀️

brendankenny:
lol we really needed an integration test on that one... `Config` was hopefully catching this and we just weren't seeing it?

lighthouse-core/lib/network-recorder.js
```diff
@@ -96,8 +93,11 @@ class NetworkRecorder extends EventEmitter {
    * @return {boolean}
    */
   static _isQUICAndFinished(record) {
-    const isQUIC = record._responseHeaders && record._responseHeaders
-        .some(header => header.name.toLowerCase() === 'alt-svc' && /quic/.test(header.value));
+    const isQUIC =
```
brendankenny: revert?

lighthouse-core/lib/network-recorder.js
```diff
@@ -189,64 +191,86 @@ class NetworkRecorder extends EventEmitter {
    * @param {LH.Crdp.Network.RequestWillBeSentEvent} data
    */
   onRequestWillBeSent(data) {
-    // NOTE: data.timestamp -> time, data.type -> resourceType
-    this.networkManager._dispatcher.requestWillBeSent(data.requestId,
-        data.frameId, data.loaderId, data.documentURL, data.request,
-        data.timestamp, data.wallTime, data.initiator, data.redirectResponse,
-        data.type);
+    const originalRequest = this._findRealRequest(data.requestId);
+    if (originalRequest) {
```
brendankenny: a comment on what this case is would be 👍👍

lighthouse-core/lib/network-recorder.js
```diff
@@ -189,64 +191,86 @@ class NetworkRecorder extends EventEmitter {
...
+
+      originalRequest.redirectDestination = redirectRequest;
+      redirectRequest.redirectSource = originalRequest;
+
+      this.onRequestStarted(redirectRequest);
+      this.onRequestFinished(originalRequest);
+      return;
+    }
+
+    const request = new NetworkRequest();
```
brendankenny: maybe move this case to the early return as it's so much simpler/shorter?

lighthouse-core/lib/network-recorder.js
```diff
@@ -270,18 +294,66 @@ class NetworkRecorder extends EventEmitter {
     }
   }
 
+  /**
+   * @param {string} requestId
```
brendankenny: comment on what a "real" request is?

For my own confusion, the IDs of a chain of redirected requests are all the same? Hence the `redirectDestination` chain traversal? And the assumption is that if there's more data for a request it can only be the last in the redirect chain because the rest are finished?

lighthouse-core/lib/network-recorder.js
brendankenny: is there a way to do this incrementally as (presumably) NetworkManager or whatever does? If the same thing is available in the page dependency graph I guess that's fine too, but at that point do we need NetworkRecords...

lighthouse-core/lib/web-inspector.js
```diff
@@ -100,67 +100,13 @@ module.exports = (function() {
     return this._moduleSettings[settingName];
   };
 
-  // Enum from chromium//src/third_party/WebKit/Source/core/loader/MixedContentChecker.h
-  global.NetworkAgent = {
```
brendankenny: 🎉 🎈 🎉

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,251 @@
...
+    this.parsedURL = /** @type {LH.WebInspector.ParsedURL} */ ({scheme: ''});
+    this.documentURL = '';
+
+    this.startTime = -1;
+    /** @type {number} */
+    this.endTime = -1;
+    /** @type {number} */
+    this._responseReceivedTime = -1;
+
+    this.transferSize = 0;
```
brendankenny: if you just went with `this._transferSize` I feel like this PR would shrink by 30%, but presumable there was a reason :)

patrickhulce: well transferSize is the only one we have to standardize in the middle of this PR to get everything to work thanks to the way that the old network requests worked, since we want to end up on `transferSize` in the end, makes more sense to just go straight to it :)

lighthouse-core/lib/network-recorder.js
patrickhulce: I really, really want to nix these entirely so network request can be trivially serializable. Because of that, I don't really want to do unnecessary work to make them nice only to remove them

third-party/devtools/ResourceType.js
```diff
@@ -0,0 +1,326 @@
+/*
```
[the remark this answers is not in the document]
patrickhulce: sure done

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-recorder.js
patrickhulce: > For my own confusion, the IDs of a chain of redirected requests are all the same?

Yup

> And the assumption is that if there's more data for a request it can only be the last in the redirect chain because the rest are finished?

Yup, after another `requestWillBeSent` is issued signaling a redirect, that is the final message of the original request and the rest are about the redirected request.

patrickhulce: added a comment basically saying this 👍

patrickhulce:
@brendankenny would you be able to give some post-initial thoughts? :)

patrickhulce:
alright I'm pretty confident in this at this point, I've run it against the lantern expectations found, 2 more subtle places where we weren't matching DT behavior (though arguably we could've left it as is), and the only 8 sites that changed were small ordering differences that should be eliminated by #5362 

the actual network record objects that were created were identical across all critical properties

third-party/devtools/ResourceType.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * @return {boolean}
+   */
+  isTextType() {
+    return this._isTextType;
+  }
+};
+
+/**
+ * Keep these in sync with WebCore::InspectorPageAgent::resourceTypeJson
```
brendankenny: you could maybe type this with `Record<LH.Crdp.Page.ResourceType, ResourceType>` to make sure they stay in sync (at least with our types for the protocol)

brendankenny: hmm, but this has more keys than that. I would have expected them to match. The [types in the protocol repo](https://github.com/ChromeDevTools/devtools-protocol/blob/master/types/protocol.d.ts#L11202) have more of them (I should update us to use that), but still missing `SourceMapScript` and `SourceMapStyleSheet`

lighthouse-core/test/lib/network-recorder-test.js
```diff
@@ -17,6 +18,33 @@ describe('network recorder', function() {
     assert.equal(records.length, 76);
   });
 
+  it('handles redirects properly', () => {
```
brendankenny: it feels like we should be adding a lot more tests here, though I realize our tests for working network request creation were just as implicit before :) WDYT? Can we cheat and bring over devtools tests or are they all in the form of crazy layout tests?

lighthouse-core/lib/network-recorder.js
```diff
@@ -162,91 +159,123 @@ class NetworkRecorder extends EventEmitter {
...
+      return;
+    }
+
+    // TODO(phulce): log these to sentry?
+    if (!data.redirectResponse) {
+      return;
+    }
+
+    // On redirect, another requestWillBeSent message is fired for the same requestId.
+    // Update the previous network request and create a new one for the redirect.
```
brendankenny: maybe "Update and finish the previous network request and..."

lighthouse-core/lib/network-recorder.js
```diff
@@ -162,91 +159,123 @@ class NetworkRecorder extends EventEmitter {
...
+    const modifiedData = {
+      ...data,
+      // Copy over the initiator as well to match DevTools behavior
+      // TODO(phulce): abandon this DT hack and update Lantern graph to handle it
+      initiator: originalRequest._initiator,
+      requestId: `${originalRequest.requestId}:redirect`,
+    };
+    const redirectRequest = new NetworkRequest();
```
brendankenny: bikeshed: `redirectRequest` to me sounds like the original request (it returned with instructions to redirect). Maybe `redirectedRequest`?

lighthouse-core/lib/network-recorder.js
```diff
@@ -162,91 +159,123 @@ class NetworkRecorder extends EventEmitter {
...
+
+    redirectRequest.onRequestWillBeSent(modifiedData);
+    originalRequest.onRedirectResponse(data);
```
brendankenny: I assume the interleaved order of the original and redirected mostly don't matter, at least for the filling of the `NetworkRequest` objects, but should things like marking the original finished (and thus sending out the event for it) be before the redirected request is started or is this ordering of events as things are now?

lighthouse-core/lib/network-recorder.js
brendankenny: > yeah it's a bit weird, AFAICT we actually only ever use the ones created by the recordsFromLogs, the ones created incrementally are just for the status notifications

ha, I think it was in anticipation of exactly this sort of situation. Since audits can only access network records created by `recordsFromLogs(devtoolsLogs)`, we wanted to make sure gatherers also saw exactly the same thing (in `afterPass()`) and so switched everything to using it.

So this impl and then migration from it sounds good. Nothing should be able to observe the difference since driver never actually inspects the records while listening to the quiet events

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    this._resourceSize = 0;
+    this._fromDiskCache = false;
+    this._fromMemoryCache = false;
+
+    this.finished = false;
+    this.requestMethod = '';
+    this.statusCode = -1;
+    this.redirectSource = /** @type {NetworkRequest|undefined} */ (undefined);
```
brendankenny: wouldn't these be fine as just `@type` and not casts? (could also add them comments I mention below in those jsdoc blocks, then)

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    this.redirectDestination = /** @type {NetworkRequest|undefined} */ (undefined);
+    this.redirects = /** @type {NetworkRequest[]|undefined} */ (undefined);
+    this.failed = false;
+    this.localizedFailDescription = '';
+
+    this._initiator = /** @type {LH.Crdp.Network.Initiator} */ ({type: 'other'});
```
brendankenny: same with these

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    this._timing = /** @type {LH.Crdp.Network.ResourceTiming|undefined} */ (undefined);
+    this._resourceType = /** @type {LH.WebInspector.ResourceType|undefined} */ (undefined);
+    this._mimeType = '';
+    this.priority = () => /** @type {LH.Crdp.Network.ResourcePriority} */ ('Low');
```
brendankenny: any reason not to use an internal `_priority` and then have a regular function member to get it? (same for `initiatorRequest`)

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    this.initiatorRequest = () => /** @type {NetworkRequest|undefined} */ (undefined);
+    this._responseHeaders = /** @type {LH.WebInspector.HeaderValue[]} */ ([]);
+
+    this._fetchedViaServiceWorker = false;
+    this._frameId = /** @type {string|undefined} */ ('');
+    this._isLinkPreload = false;
+
+    // Make sure we're compatible with old WebInspector.NetworkRequest
+    // eslint-disable-next-line no-unused-vars
+    const record = /** @type {LH.WebInspector.NetworkRequest} */ (this);
```
brendankenny: haha, this is great. Since we don't have `@implements` this could simplify quite a few things...

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  /**
+   * @return {NetworkRequest}
+   */
+  clone() {
```
brendankenny: used anywhere?

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    if (data.encodedDataLength !== -1) {
+      this.transferSize += data.encodedDataLength;
+    }
+  }
+
+  /**
+   * @param {LH.Crdp.Network.LoadingFinishedEvent} data
+   */
+  onLoadingFinished(data) {
+    if (this.finished) return;
```
brendankenny: when does this happen?

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+   * @param {LH.Crdp.Network.LoadingFailedEvent} data
+   */
+  onLoadingFailed(data) {
+    if (this.finished) return;
+
+    this.finished = true;
+    this.endTime = data.timestamp;
+
+    this.failed = true;
+    this._resourceType = data.type && resourceTypes[data.type];
```
brendankenny: why updated here?

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+  /**
+   * Resolve differences between conflicting timing signals. Based on the property setters in DevTools.
+   * @see https://github.com/ChromeDevTools/devtools-frontend/blob/56a99365197b85c24b732ac92b0ac70feed80179/front_end/sdk/NetworkRequest.js#L485-L502
+   * @param {LH.Crdp.Network.ResourceTiming} timing
+   */
+  _recomputeTimesWithResourceTiming(timing) {
+    // Take startTime and responseReceivedTime from timing data for better accuracy.
+    // Timing's requestTime is a baseline in seconds, rest of the numbers there are ticks in millis.
+    this.startTime = timing.requestTime;
+    const headersReceivedTime = timing.requestTime + timing.receiveHeadersEnd / 1000.0;
```
brendankenny: 1000

lighthouse-core/lib/network-request.js
```diff
@@ -0,0 +1,260 @@
...
+    if (!this._responseReceivedTime || this._responseReceivedTime < 0) {
+      this._responseReceivedTime = headersReceivedTime;
+    }
+
+    this._responseReceivedTime = Math.min(this._responseReceivedTime, headersReceivedTime);
+    this._responseReceivedTime = Math.max(this._responseReceivedTime, this.startTime);
+    this.endTime = Math.max(this.endTime, this._responseReceivedTime);
+  }
+
+  _updateResponseReceivedTimeIfNecessary() {
```
brendankenny: add a comment for what this is doing/when it's "necessary"?

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-recorder.js
patrickhulce: done

lighthouse-core/lib/network-recorder.js
patrickhulce: yeah mostly doesn't matter, the only thing that matters is making sure we don't have a fake quiet period on status messages, so starting the new one before finishing the old one is what we want 👍

lighthouse-core/lib/network-request.js
patrickhulce: yeah sure

lighthouse-core/lib/network-request.js
patrickhulce: done-ish, some still won't work

lighthouse-core/lib/network-request.js
patrickhulce: yes, don't want to add machinery for things I immediately want to get rid of :)

lighthouse-core/lib/network-request.js
patrickhulce: nah, I'll remove

lighthouse-core/lib/network-request.js
patrickhulce: it seems to be expected for a certain class of request failures (aborted redirected requests)

not sure how to best comment it...maybe?
`// On cancelled requests, DevTools can send duplicate erroneous events, prefer the first one for best timing data`

lighthouse-core/lib/network-request.js
patrickhulce: because DT does :) https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/sdk/NetworkManager.js?type=cs&q=%22setResourceType%22+-f:out+lang:js&sq=package:chromium&g=0&l=600

I think it's because resourceTypes can always lie, so you can always get more information when the response comes back, also probably has to do with signed-exchanges and some corner cases we don't really care about

lighthouse-core/lib/network-request.js
patrickhulce: done

lighthouse-core/lib/network-request.js
patrickhulce: done

lighthouse-core/test/lib/network-recorder-test.js
patrickhulce: > Can we cheat and bring over devtools tests

there are devtools tests!? 😉

there's really nothing portable, I have a proposal for "supplemental" testing based on the lantern-data directory and using jest we can discuss tomorrow at eng sync, but our extensive use of network records throughout the rest of the codebase was really quite good coverage

third-party/devtools/ResourceType.js
patrickhulce: FWIW that was a devtools comment I didn't add that :)

we only need those ones that match so I'm cool with that, and honestly this entire file can be deleted after we stop that `WebInspector.resourceTypes ===` nonsense :D

typings/web-inspector.d.ts
```diff
@@ -6,8 +6,8 @@
 
 declare global {
   module LH.WebInspector {
-    // TODO(bckenny): standardize on underscored internal API
-    // externs for chrome-devtools-frontend/front_end/sdk/NetworkRequest.js
+    // TODO(phulce): standardize on non-underscored property names
```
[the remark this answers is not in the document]
patrickhulce: done

typings/web-inspector.d.ts
```diff
@@ -18,41 +18,44 @@ declare global {
...
 
       finished: boolean;
       requestMethod: string;
       statusCode: number;
-      redirectSource?: {
-        url: string;
-      }
+      redirectSource?: {url: string;};
+      redirectDestination?: {url: string;};
+      redirects?: NetworkRequest[];
```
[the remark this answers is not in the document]
patrickhulce: done over there 👍

typings/web-inspector.d.ts
```diff
@@ -62,16 +65,10 @@ declare global {
     }
 
     export interface ResourceType {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/lib/network-recorder-test.js
brendankenny: > but our extensive use of network records throughout the rest of the codebase was really quite good coverage

yeah, they are pretty fundamental. I'm ok with that + incrementally adding as we hit issues/update

lighthouse-core/lib/network-recorder.js
brendankenny: maybe add a comment so that anyone rearranging in the future won't accidentally do that?

lighthouse-core/lib/network-request.js
brendankenny: > yes, don't want to add machinery for things I immediately want to get rid of :)

ha, oh right. SG

lighthouse-core/lib/network-request.js
brendankenny: I think that's 👍

lighthouse-core/lib/network-request.js
brendankenny: ah, yeah, good to mirror the real deal. I was also wondering because `onLoadingFinished` doesn't do the same

typings/web-inspector.d.ts
```diff
@@ -61,17 +66,7 @@ declare global {
...
-      name(): string;
-      _name: string;
-      title(): string;
-      isTextType(): boolean;
-    }
-
-    export interface ResourceCategory {
-      title: string;
-    }
+    export type ResourceType = _ResourceType;
```
brendankenny: it doesn't seem like anything is referring to this type except the `NetworkRequest` interface, so can drop this declaration?

patrickhulce: done

PR #5462, opened 2018-06-09
core: convert gather-runner.js to async/await

JoelEinbinder:
I was reading through how this file worked, and decided to convert it to async/await as I did. Clears up some dastardly uses of `array.reduce`.

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -70,12 +70,13 @@ class GatherRunner {
    * @param {number=} duration
    * @return {Promise<void>}
    */
-  static loadBlank(
+  static async loadBlank(
       driver,
       url = constants.defaultPassConfig.blankPage,
       duration = constants.defaultPassConfig.blankDuration
   ) {
-    return driver.gotoURL(url).then(_ => new Promise(resolve => setTimeout(resolve, duration)));
+    await driver.gotoURL(url);
+    await new Promise(resolve => setTimeout(resolve, duration));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -248,31 +242,25 @@ class GatherRunner {
     const status = 'Loading page & waiting for onload';
     log.log('status', status, gatherernames);
 
-    const pass = Promise.resolve()
-      // Clear disk & memory cache if it's a perf run
-      .then(_ => {
-        if (isPerfRun) driver.cleanBrowserCaches();
```
JoelEinbinder: this code wasn't be awaited, but it looks like it should be. Same with `beginTrace` down below.

lighthouse-core/gather/gather-runner.js
```diff
@@ -248,31 +242,25 @@ class GatherRunner {
...
+    // Clear disk & memory cache if it's a perf run
+    if (isPerfRun) await driver.cleanBrowserCaches();
+    // Always record devtoolsLog
+    await driver.beginDevtoolsLog();
+    // Begin tracing if requested by config.
+    if (recordTrace) await driver.beginTrace(settings);
+
+    // Navigate.
+    await GatherRunner.loadPage(driver, passContext);
+    await log.log('statusEnd', status);
```
patrickhulce: this one shouldn't need to be awaited

lighthouse-core/gather/gather-runner.js
brendankenny: maybe https://palantir.github.io/tslint/rules/no-floating-promises/, which isn't perfect but would have caught this

JoelEinbinder: I think just using async/await makes these stand out a lot more. That lint rule might be more trouble than its worth.

lighthouse-core/gather/gather-runner.js
JoelEinbinder: done.

PR #5474, opened 2018-06-11
core: add option to disable navigation

JoelEinbinder:
When using Lighthouse with Puppeteer, I want to be able to set the page up with Puppeteer by clicking things and waiting for the exact moment loading finished, and then run Lighthouse's accessibility audits on the current state of the page. This requires the ability to run Lighthouse audits without clearing the page state by reloading.


lighthouse-core/gather/gather-runner.js
```diff
@@ -106,7 +106,9 @@ class GatherRunner {
     log.log('status', 'Initializing…');
     const resetStorage = !options.settings.disableStorageReset;
     // Enable emulation based on settings
-    await driver.assertNoSameOriginServiceWorkerClients(options.requestedUrl);
+    if (!options.settings.dontNavigate) {
```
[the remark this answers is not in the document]
JoelEinbinder: Done.

typings/externs.d.ts
```diff
@@ -63,6 +63,7 @@ declare global {
       onlyAudits?: string[] | null;
       onlyCategories?: string[] | null;
       skipAudits?: string[] | null;
+      skipNavigation?: boolean;
```
patrickhulce: our tests will hopefully force you to add this to `core/config/constants.js` as well

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -453,7 +453,7 @@ describe('GatherRunner', function() {
       ],
     };
 
-    return GatherRunner.afterPass({url, driver, passConfig}, {TestGatherer: []}).then(passData => {
```
patrickhulce: could you add a quick test for the `skipNavigation` use case? it might make sense to do it at the highest `run` level to get most bang for buck

lighthouse-core/gather/gather-runner.js
```diff
@@ -106,7 +106,9 @@ class GatherRunner {
     log.log('status', 'Initializing…');
     const resetStorage = !options.settings.disableStorageReset;
     // Enable emulation based on settings
-    await driver.assertNoSameOriginServiceWorkerClients(options.requestedUrl);
```
patrickhulce: I wonder if it might make more sense long term to pass the settings into `Driver` and just have all the methods that only make sense for navigation be noops.

patching the holes here seems easy to miss one, OTOH it might be confusing to have `loadPage` do nothing based on some hidden context

lighthouse-core/gather/gather-runner.js
```diff
@@ -106,7 +106,9 @@ class GatherRunner {
...
+    if (!options.settings.skipNavigation) {
+      await driver.assertNoSameOriginServiceWorkerClients(options.requestedUrl);
+    }
     const userAgent = await driver.getUserAgent();
     gathererResults.UserAgent = [userAgent];
     GatherRunner.warnOnHeadless(userAgent, gathererResults);
```
patrickhulce: there's other setup stuff that we can probably skip in this block, but if it's not necessary for you I guess we can pick that up later 👍

lighthouse-core/test/gather/gather-runner-test.js
JoelEinbinder: Yeah I was going to ask where the test for this could go. I was having trouble figuring out how fake the fake-driver was.

brendankenny:
we should probably write a guide for how to add a new flag

PR #5481, opened 2018-06-12
core(tsc): add initial trivial type info to config.js

brendankenny:
Part 1 of 2 to add type checking to `config.js`

This is an attempt to reduce noise around the more major changes by pulling out the trivial type annotations/changes needed without changing any functionality. Some of these types are slightly white lies, but will be correct with part 2.

Main trickiness here is some of the generic functions, like `merge()` and `deepClone()`, but they maintain their earlier behavior (except `deepCloneConfigJson()` is split out), it's mostly just convincing the compiler that they do what they say they do.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -17,10 +17,18 @@ const path = require('path');
 const Audit = require('../audits/audit');
 const Runner = require('../runner');
 
+/** @typedef {typeof import('../gather/gatherers/gatherer.js')} GathererConstructor */
+/** @typedef {InstanceType<GathererConstructor>} Gatherer */
+
+/**
+ * @param {LH.Config['passes']} passes
+ * @param {LH.Config['audits']} audits
+ */
 function validatePasses(passes, audits) {
   if (!Array.isArray(passes)) {
     return;
   }
+
   const requiredGatherers = Config.getGatherersNeededByAudits(audits);
 
   // Log if we are running gathers that are not needed by the audits listed in the config
@@ -28,7 +36,7 @@ function validatePasses(passes, audits) {
     pass.gatherers.forEach(gathererDefn => {
       const gatherer = gathererDefn.instance || gathererDefn.implementation;
       const isGatherRequiredByAudits = requiredGatherers.has(gatherer.name);
-      if (isGatherRequiredByAudits === false) {
+      if (!isGatherRequiredByAudits) {
         const msg = `${gatherer.name} gatherer requested, however no audit requires it.`;
         log.warn('config', msg);
       }
@@ -46,6 +54,11 @@ function validatePasses(passes, audits) {
   });
 }
 
+/**
+ * @param {LH.Config['categories']} categories
+ * @param {LH.Config['audits']} audits
+ * @param {LH.Config['groups']} groups
+ */
 function validateCategories(categories, audits, groups) {
   if (!categories) {
     return;
@@ -57,7 +70,7 @@ function validateCategories(categories, audits, groups) {
         throw new Error(`missing an audit id at ${categoryId}[${index}]`);
       }
 
-      const audit = audits.find(a => a.implementation.meta.name === auditRef.id);
+      const audit = audits && audits.find(a => a.implementation.meta.name === auditRef.id);
       if (!audit) {
         throw new Error(`could not find ${auditRef.id} audit for category ${categoryId}`);
       }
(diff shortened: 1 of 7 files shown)


typings/config.d.ts
```diff
@@ -7,29 +7,30 @@
...
 import Audit = require('../lighthouse-core/audits/audit.js');
 
-
 declare global {
   module LH {
     /**
      * The full, normalized Lighthouse Config.
      */
-    export interface Config {
+    export interface Config extends Config.Json {
```
brendankenny: this will give the very nice property that a `Config` is a valid `Config.Json`. Will allow easy serialization, `--print-config` or whatever, and means you don't have to keep track of what you have, a real config or a configJson. Either way, just pass in to the `Config` constructor (followup will ensure that a `Config` passed into `new Config()` will end up unchanged).

typings/config.d.ts
```diff
@@ -7,29 +7,30 @@
...
       settings: Config.Settings;
-      passes?: Config.Pass[];
-      audits?: Config.AuditDefn[];
-      categories?: Record<string, Config.Category>;
-      groups?: Record<string, Config.Group>;
+      passes: Config.Pass[] | null;
+      audits: Config.AuditDefn[] | null;
+      categories: Record<string, Config.Category> | null;
+      groups: Record<string, Config.Group> | null;
```
brendankenny: we had a mixture of optional and `null` used when one of these wasn't provided (e.g. don't need `audits` when in `gatherMode`). Went with the null as otherwise we'd need to be deleting things when e.g. everything ends up filtered out or whatever

typings/config.d.ts
```diff
@@ -107,6 +117,13 @@ declare global {
         auditRefs: AuditRef[];
       }
       export interface Group extends GroupJson {}
+
+      export type MergeOptionsOfItems = <T extends {path?: string, options: Record<string, any>}>(items: T[]) => T[];
```
brendankenny: good news is with the latest tsc changes, these will be able to live in jsdocs soon

typings/config.d.ts
```diff
@@ -7,29 +7,30 @@
...
+      categories: Record<string, Config.Category> | null;
+      groups: Record<string, Config.Group> | null;
     }
 
     module Config {
       /**
        * The pre-normalization Lighthouse Config format.
        */
       export interface Json {
+        extends?: 'lighthouse:default' | 'lighthouse:full' | string | boolean;
```
patrickhulce: what's the benefit of doing `'lighthouse:default' | string` ?

lighthouse-core/config/config.js
```diff
@@ -172,32 +199,73 @@ function merge(base, extension, overwriteArrays = false) {
...
+ * @param {Array<T>} array
+ * @return {Array<T>}
+ */
 function cloneArrayWithPluginSafety(array) {
   return array.map(item => {
-    return typeof item === 'object' ? Object.assign({}, item) : item;
+    if (typeof item === 'object') {
+      // Return copy of instance and prototype chain (in case item is instantiated class).
+      return Object.assign(
+        Object.create(
```
patrickhulce: did we need this? if we do, add a test for it?

lighthouse-core/config/config.js
```diff
@@ -172,32 +199,73 @@ function merge(base, extension, overwriteArrays = false) {
...
+ * @return {LH.Config.Json}
+ */
+function deepCloneConfigJson(json) {
+  const cloned = deepClone(json);
 
   // Copy arrays that could contain plugins to allow for programmatic
   // injection of plugins.
-  if (Array.isArray(json.passes)) {
-    cloned.passes.forEach((pass, i) => {
+  if (Array.isArray(cloned.passes) && Array.isArray(json.passes)) {
```
patrickhulce: could this ever not be true if one of them is true? just for TS?

typings/config.d.ts
brendankenny: > what's the benefit of doing `'lighthouse:default' | string`?

no benefit, just documentation. I can remove or it might not matter :) (originally it was just the specific strings, but it gets automatically widened to `string` when requiring json files, so default-config et al were failing the type check)

lighthouse-core/config/config.js
brendankenny: > just for TS?

just for TS and pretty annoying :)

lighthouse-core/config/config.js
brendankenny: > did we need this? if we do, add a test for it?

we need it for supporting live gatherer instances or `{instance: liveGathererIntance, ...}`, which seems reasonable if we support live `implementation`s.

The reason I want it most is so `Config` is derived from `Config.Json`, but another approach to do that would be for shorthand expansion to drop `{instance: ...}`...we'd just need to make sure there was a `path` or `implementation` in that case.

I'll add a test but let me know if you'd rather just not support that

patrickhulce: that makes sense seems reasonable! just test case then :)

PR #5486, opened 2018-06-13
core(tsc): refactor and add type checking to config.js

brendankenny:
even with #5481 out of the way, this is still kind of a doozy to review, *but*

- With most of the rest of lighthouse type checked, this involved significantly fewer changes than a few months ago
- With @patrickhulce's work to unify CLI flags and the config and adding defaults for everything, a lot of simplification was already done and more was possible

Main starting change is eliminating `generateNewFilteredConfig()` as kind of an externally-facing API. We have a way of expressing that in the config itself now (`onlyCategories`, etc), so it makes sense to not expose a function for it anymore (and the extension, etc moved on to using `onlyCategories`, etc a while ago).

With that gone, the constructor is really the only entry point, and control flow becomes just a simple tree (ignoring `deepClone` and `mergeOptionsOfItems`):
  - make a copy of input config json
  - (optional) extend from a base config
  - clean settings and set defaults
  - clean passes and require gatherers
  - require audits
  - filter if needed
  - validate internal connections between audits, gatherers, etc

This means configs are always in one of two states, a `ConfigJson` -- possibly missing values and needing to extend something or be filtered -- and a `Config` -- all defaults set, filtered and extended. As mentioned in #5481, conveniently a `Config` is also a `ConfigJson`, enforced by the type system and some tests here.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -3,7 +3,6 @@
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
-// @ts-nocheck
 'use strict';
 
 const defaultConfigPath = './default-config.js';
@@ -34,7 +33,7 @@ function validatePasses(passes, audits) {
   // Log if we are running gathers that are not needed by the audits listed in the config
   passes.forEach(pass => {
     pass.gatherers.forEach(gathererDefn => {
-      const gatherer = gathererDefn.instance || gathererDefn.implementation;
+      const gatherer = gathererDefn.instance;
       const isGatherRequiredByAudits = requiredGatherers.has(gatherer.name);
       if (!isGatherRequiredByAudits) {
         const msg = `${gatherer.name} gatherer requested, however no audit requires it.`;
@@ -165,10 +164,14 @@ function assertValidGatherer(gathererInstance, gathererName) {
  * @return {Partial<LH.Config.Settings>}
  */
 function cleanFlagsForSettings(flags = {}) {
+  /** @type {Partial<LH.Config.Settings>} */
   const settings = {};
+
   for (const key of Object.keys(flags)) {
+    // @ts-ignore - intentionally testing some keys not on defaultSettings to discard them.
     if (typeof constants.defaultSettings[key] !== 'undefined') {
-      settings[key] = flags[key];
+      const safekey = /** @type {keyof LH.SharedFlagsSettings} */ (key);
+      settings[safekey] = flags[safekey];
     }
   }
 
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/config/config-test.js
```diff
@@ -623,90 +683,171 @@ describe('Config', () => {
     });
   });
 
-  describe('generateNewFilteredConfig', () => {
+  describe('filterConfigIfNeeded', () => {
```
brendankenny: the only changes to existing tests were
- to use `onlyCategories` (etc) instead of `Config.generateNewFilteredConfig` directly
- to use `requireAudits`/`requireGatherers` which now call the shorthand expansion methods instead of `expandAuditShorthandAndMergeOptions` directly
caused a little code churn to maintain the same test, but the intention was for test path/result to be exactly the same

lighthouse-core/config/config.js
```diff
@@ -327,46 +330,31 @@ class Config {
...
-    this._configDir = configPath ? path.dirname(configPath) : undefined;
+    // Augment passes with necessary defaults and require gatherers.
+    const passesWithDefaults = Config.augmentPassesWithDefaults(configJSON.passes);
+    Config.adjustDefaultPassForThrottling(settings, passesWithDefaults);
+    const passes = Config.requireGatherers(passesWithDefaults, configDir);
 
     /** @type {LH.Config['settings']} */
-    this._settings = configJSON.settings || {};
+    this.settings = settings;
     /** @type {LH.Config['passes']} */
```
paulirish: i think these casts arent necessary any longer?

brendankenny: Well, they're kind of casts, but they're not coercive, so if the lhs can't be treated as these things there will be an error. That's a nice thing to have because if future changes to `config.js` accidentally change the lhs type, the error will be here. If not, the class's property type will just be whatever the accidental type is, and the error will instead be deep somewhere else where the property is actually used. It's also nice for file-level documentation since these are pretty much the Config instance API.

Some of this (and some of `config.d.ts`) can be deleted when typescript supports `@implements`, though (https://github.com/Microsoft/TypeScript/issues/17498)

lighthouse-core/config/config.js
```diff
@@ -395,40 +383,59 @@ class Config {
...
-    if (config.passes) {
-      config.passes = config.passes.map(pass => merge(deepClone(defaultPassConfig), pass));
+  static augmentPassesWithDefaults(passes) {
+    if (!passes) {
+      return null;
     }
 
-    return config;
+    const {defaultPassConfig} = constants;
+    return passes.map(pass => merge(deepClone(defaultPassConfig), pass));
```
patrickhulce: ~~I think I may have missed this last time, but don't we need safe deepClone here for `.gatherers` ala deepCloneConfigJson?~~

nevermind we only clone the defaults, right? 👍

lighthouse-core/config/config.js
```diff
@@ -395,40 +383,59 @@ class Config {
...
     if (!audits) {
       return null;
     }
 
     const newAudits = audits.map(audit => {
       if (typeof audit === 'string') {
         return {path: audit, options: {}};
-      } else if (audit && typeof audit.audit === 'function') {
+      } else if ('implementation' in audit && typeof audit.implementation.audit === 'function') {
+        return audit;
```
patrickhulce: the example comments for gatherer are 💯 want to add them here too?

lighthouse-core/config/config.js
```diff
@@ -491,70 +501,56 @@ class Config {
...
-  static generateNewFilteredConfig(oldConfig, categoryIds, auditIds, skipAuditIds) {
-    // 0. Clone config to avoid mutating it
-    const config = deepCloneConfigJson(oldConfig);
-    config.audits = Config.expandAuditShorthandAndMergeOptions(config.audits);
-    config.passes = Config.expandGathererShorthandAndMergeOptions(config.passes);
-    config.passes = Config.requireGatherers(config.passes);
+  static filterConfigIfNeeded(config) {
+    const settings = config.settings;
+    if (!settings.onlyCategories && !settings.onlyAudits && !settings.skipAudits) {
+      return config;
```
patrickhulce: want to add a return typedef? seems like we don't actually want to return anything

lighthouse-core/config/config.js
```diff
@@ -568,17 +564,17 @@ class Config {
...
       if (!foundCategory) {
         const parentKeyName = skipAuditIds.includes(auditId) ? 'skipAudits' : 'onlyAudits';
         log.warn('config', `unrecognized audit in '${parentKeyName}': ${auditId}`);
-      }
-
-      if (auditIds.includes(auditId) && categoryIds.includes(foundCategory)) {
-        log.warn('config', `${auditId} in 'onlyAudits' is already included by ` +
-            `${foundCategory} in 'onlyCategories'`);
+      } else {
+        if (auditIds.includes(auditId) && categoryIds.includes(foundCategory)) {
```
patrickhulce: want to smoosh into else if?

lighthouse-core/config/config.js
```diff
@@ -714,28 +690,38 @@ class Config {
...
+    if (!expandedAudits) {
       return null;
     }
 
     const coreList = Runner.getAuditList();
-    return audits.map(auditDefn => {
-      if (!auditDefn.implementation) {
-        const path = auditDefn.path;
+    const auditDefns = expandedAudits.map(audit => {
+      let implementation;
```
patrickhulce: does this become `any`? or does `TS` infer based on the later assignments?

lighthouse-core/config/config.js
```diff
@@ -752,63 +738,54 @@ class Config {
...
-
-            let requirePath = `../gather/gatherers/${name}`;
-            if (!coreGatherer) {
-              // Otherwise, attempt to find it elsewhere. This throws if not found.
-              requirePath = Runner.resolvePlugin(name, configPath, 'gatherer');
-            }
-
-            GathererClass = require(requirePath);
+    const fullPasses = passes.map(pass => {
+      const gathererDefns = Config.expandGathererShorthand(pass.gatherers).map(gathererDefn => {
```
patrickhulce: this has become quite the inner loop, WDYT about extracting a method for the actual `requireing` part as it was before?

lighthouse-core/test/config/config-test.js
```diff
@@ -623,90 +683,171 @@ describe('Config', () => {
...
+      const extended = {
+        extends: 'lighthouse:default',
+        settings: {
+          onlyCategories: ['performance'],
+        },
+      };
+      const extendedConfig = new Config(extended);
+
+      assert.equal(config.passes.length, 1, 'did not filter config');
+      assert.deepStrictEqual(config, extendedConfig, 'no mutations');
```
patrickhulce: should the message be "had mutations"?

brendankenny:
> a bit heavier on impl changes for a type checking PR...

good point. Updated the PR name :)

lighthouse-core/config/config.js
brendankenny: good idea, done

lighthouse-core/config/config.js
brendankenny: > seems like we don't actually want to return anything

ah, yes, good call

lighthouse-core/config/config.js
brendankenny: done

lighthouse-core/config/config.js
brendankenny: it's `any` here, but by the time it's out of the if/else blocks it's `typeof Audit` from the assignments

lighthouse-core/config/config.js
brendankenny: > WDYT about extracting a method for the actual requireing part as it was before?

ok, attempted. LMKWYT

lighthouse-core/test/config/config-test.js
brendankenny: yes, done

lighthouse-core/config/config.js
brendankenny: yeah, the default is an empty `gatherers` array so it's fine, though I guess it could be fragile if we ever change that in the future...

brendankenny:
🔥🔥 ❓ 🔥🔥

Issue #5493, opened 2018-06-13
Parsing trace file created using chrome driver performance logging mode


mangeshkakad:
Lighthouse GitHub page shows that it can be use as trace parser as well. I am struggling to make it work and always get empty report back.

Command: lighthouse --config-path=config.json --view http://dummy.com

patrickhulce:
1. What does your config.json look like? 
The critical part of [the docs](https://github.com/GoogleChrome/lighthouse/tree/master/docs#lighthouse-as-trace-processor) to get this to work is the config file not the command :)

2. What audits are you trying to run?
There are only a few that can work with *just* a trace, trace + DevTools log can take you quite a bit further.

mangeshkakad:
I am not 100% sure what config setting i should be using. Is there an example which i can refer to try with my trace file or to test with i can try with json trace file generated by lighthouse only.

My objective is logging all traces offline and then run lighthouse command offline to generate and store all reports.

patrickhulce:
If that's what you're after, it'd be simplest to just use LH for all of this.

1. Run LH on the site and save the results `lighthouse https://example.com -G=./save-results-here`
2. Store the `./save-results-here` folder wherever you need to
3. Run LH on the folder `lighthouse https://example.com -A=./path-to-saved-results-folder`. Note that the while you still need to pass in a URL, LH is only checking to make sure it matches the artifacts, it won't load the site.

mangeshkakad:
Thanks. I will try this tomorrow and will update comments.

patrickhulce:
Yes, but you'll be able to do run the audits whose only artifacts are `traces`, which is just 6 audits. You can move your trace file to a folder and call it `my-other-tool-run/defaultPass.trace.json` now you should be able to pass `lighthouse -A=./my-other-tool-run`

PR #5495, opened 2018-06-13
core(audit-mode): do not require a URL

patrickhulce:
realized in #5493 that requiring a URL in audit mode is silly, this removes that requirement

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -138,7 +138,9 @@ function getFlags(manualArgv) {
       .check(/** @param {!LH.Flags} argv */ (argv) => {
         // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
         // or --list-trace-categories. If not, stop the program and ask for a url
-        if (!argv.listAllAudits && !argv.listTraceCategories && argv._.length === 0) {
+        const isListMode = argv.listAllAudits || argv.listTraceCategories;
+        const isOnlyAuditMode = !!argv.auditMode && !argv.gatherMode;
+        if (!isListMode && !isOnlyAuditMode && argv._.length === 0) {
           throw new Error('Please provide a url');
         }
 
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -136,8 +136,10 @@ function getFlags(manualArgv) {
       .default('port', 0)
       .default('hostname', 'localhost')
       .check(/** @param {!LH.Flags} argv */ (argv) => {
-        // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
-        // or --list-trace-categories. If not, stop the program and ask for a url
+        // Lighthouse doesn't need a URL if...
+        //   - We're in auditMode (and we have artifacts already)
+        //   - We're just listing the available options.
+        // If one of these don't apply, stop the program and ask for a url.
         const isListMode = argv.listAllAudits || argv.listTraceCategories;
         const isOnlyAuditMode = !!argv.auditMode && !argv.gatherMode;
         if (!isListMode && !isOnlyAuditMode && argv._.length === 0) {
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -23,7 +23,7 @@ const Connection = require('./gather/connections/connection.js'); // eslint-disa
 class Runner {
   /**
    * @param {Connection} connection
-   * @param {{config: LH.Config, url: string, driverMock?: Driver}} opts
+   * @param {{config: LH.Config, url?: string, driverMock?: Driver}} opts
    * @return {Promise<LH.RunnerResult|undefined>}
    */
   static async run(connection, opts) {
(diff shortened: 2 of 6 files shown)


lighthouse-cli/cli-flags.js
```diff
@@ -138,7 +138,9 @@ function getFlags(manualArgv) {
       .check(/** @param {!LH.Flags} argv */ (argv) => {
         // Make sure lighthouse has been passed a url, or at least one of --list-all-audits
         // or --list-trace-categories. If not, stop the program and ask for a url
```
brendankenny: update comment too

lighthouse-core/runner.js
```diff
@@ -59,27 +46,49 @@ class Runner {
...
       if (settings.auditMode && !settings.gatherMode) {
         // No browser required, just load the artifacts from disk.
         const path = Runner._getArtifactsPath(settings);
         artifacts = await assetSaver.loadArtifacts(path);
+        requestedUrl = artifacts.URL.requestedUrl;
+
+        if (!requestedUrl) {
+          throw new Error('Cannot run audit mode on empty URL');
+        }
+        if (opts.url && opts.url !== requestedUrl) {
```
brendankenny: might throw a wrench in the type checking works, but `opts.url` should now be optional in the function params?

lighthouse-core/runner.js
```diff
@@ -59,27 +46,49 @@ class Runner {
...
+          throw new Error('Cannot run audit mode on different URL');
+        }
       } else {
+        // save the requestedUrl provided by the user
```
brendankenny: we can ditch this comment now since we don't actually save it anymore (can also just drop it and use `opts.url` in its place)

lighthouse-core/runner.js
```diff
@@ -59,27 +46,49 @@ class Runner {
...
+          throw new Error('The url provided should have a proper protocol and hostname.');
+        }
+
+        // If the URL isn't https and is also not localhost complain to the user.
+        if (parsedURL.protocol !== 'https:' && parsedURL.hostname !== 'localhost') {
+          log.warn('Lighthouse', 'The URL provided should be on HTTPS');
+          log.warn('Lighthouse', 'Performance stats might be skewed redirecting to HTTPS.');
+        }
+
+        // canonicalize URL with any trailing slashes neccessary
```
brendankenny: this comment is just weird now. Add an `if`? Drop the `necessary`? Something else?

lighthouse-core/runner.js
```diff
@@ -59,27 +46,49 @@ class Runner {
...
       } else {
+        // save the requestedUrl provided by the user
+        const rawRequestedUrl = opts.url;
+        if (typeof rawRequestedUrl !== 'string' || rawRequestedUrl.length === 0) {
+          throw new Error(`You must provide a url to the runner. '${rawRequestedUrl}' provided.`);
+        }
+
+        let parsedURL;
+        try {
+          parsedURL = new URL(opts.url);
```
paulirish: rawRequestedUrl

lighthouse-core/runner.js
```diff
@@ -59,27 +46,49 @@ class Runner {
...
+        } catch (e) {
+          throw new Error('The url provided should have a proper protocol and hostname.');
+        }
+
+        // If the URL isn't https and is also not localhost complain to the user.
+        if (parsedURL.protocol !== 'https:' && parsedURL.hostname !== 'localhost') {
```
paulirish: I _was_ gonna say two things...

1. we could use `is-on-https`'s `isSecureRecord` method here..
1. lets merge the strings and just say `'The URL provided should be on HTTPS. Performance stats might be skewed redirecting to HTTPS.'`

but tbh.. 

how about we nuke this check? Our redirects audit already catches perf cost of redirects. and if someone wants to audit a http page, they should be able to.

lighthouse-core/runner.js
paulirish: how about .. move this comment to  L77 and drop the use of `parsedURL` and s/parsedURL/requestedUrl/ ?

lighthouse-core/runner.js
paulirish: also, in the future, i'd like to support `example.com` and just toss `https://` in front of it. psi supports this. i think wpt, too.

patrickhulce: yeah sg :)

brendankenny: a bigger issue is service workers and how do you notice you aren't getting one because you typed in `http`, but agree there could be a better flow (we have top level warnings for that kind of thing now)

lighthouse-core/runner.js
patrickhulce: done

lighthouse-core/runner.js
patrickhulce: N/A

lighthouse-core/runner.js
patrickhulce: done

lighthouse-core/runner.js
patrickhulce: done

lighthouse-cli/cli-flags.js
patrickhulce: done

lighthouse-core/test/runner-test.js
```diff
@@ -92,6 +92,17 @@ describe('Runner', () => {
       }
     });
 
+    it('-A throws if the URL changes', async () => {
+      const settings = {auditMode: artifactsPath, disableDeviceEmulation: true};
+      const opts = {url: 'https://example.com', config: generateConfig(settings), driverMock};
+      try {
+        await Runner.run(null, opts);
+        assert.fail('should have thrown');
```
brendankenny: won't this be caught by the try/catch and so never fail?

patrickhulce: yeah but the message won't match, it's the same weird pattern observed elsewhere in the tests :) jest has some much niftier matchers FWIW ;)

brendankenny: oh right. That's confusing but ok :)

PR #5504, opened 2018-06-14
core: enable async stacks

### LINKED ISSUE #3375: Lantern: turn on async stacks to get more reliable initiator data
`Debugger.setAsyncCallStackDepth` ftw.

As evidenced in https://github.com/ChromeDevTools/devtools-protocol/issues/56

### LINKED ISSUE #5494: Lighthouse report on AMP page - Offscreen Images
1. Run LH on https://storage.googleapis.com/cs-emea-assets/antoine/Safestore/AMP%20POC/index.html

The report highlighted "Offscreen images" as the top opportunity.

My tested URL being an AMP page, lazy load of background images is a feature coming out of the box and is a major USP of AMP, which Lighthouse reports against here. Therefore this could send the wrong signals to users who have built similar AMP pages and would not expect this feedback from a performance analysis.

Comments from Malte Ubl (AMP PM) on this issue:

AMP's algorithm is:
Prioritize resources (like images) based on distance to viewport
Eventually (when there is nothing better to do) download resources within 3 viewports of the current viewport.
This was chosen as a trade-off between bandwidth usage and avoiding users having to wait for resources to download as they scroll.
Basically, it seems like lighthouse would prefer a more conservative strategy (less bandwidth used, more waiting for users). I'd appreciate filing a lighthouse bug for them to evaluate whether they shouldn't accept AMP's strategy. My suspicion is that they don't use a large enough doc to notice that AMP stops are 3 viewports.

#### Environment Information
* Affected Channels: 
* Lighthouse version: Latest from Chrome DevTools
* Node.js version: N/A
* Operating System: Mac OS

patrickhulce:
**Summary**
Initiator information can fallback to `parser` in cases when elements were added inside a microtask/Promise chain. This hurts the fidelity of the lantern graph.

Even though async stacks are enabled by default in stable Chrome now in v8 (https://v8.dev/blog/v8-release-73), it doesn't apply to the initiator information we get from Chrome which still requires explicitly setting the async stack depth.

I would love to get this in for 5.0 as it enables a lot of performance accuracy improvements and better attribution for bootup-time/third-party-web

**Related Issues/PRs**
fixes #5494 
closes #3375

### DIFF
--- a/lighthouse-core/computed/page-dependency-graph.js
+++ b/lighthouse-core/computed/page-dependency-graph.js
@@ -30,9 +30,17 @@ class PageDependencyGraph {
   static getNetworkInitiators(record) {
     if (!record.initiator) return [];
     if (record.initiator.url) return [record.initiator.url];
-    if (record.initiator.type === 'script' && record.initiator.stack) {
-      const frames = record.initiator.stack.callFrames;
-      return Array.from(new Set(frames.map(frame => frame.url))).filter(Boolean);
+    if (record.initiator.type === 'script') {
+      /** @type {Set<string>} */
+      const scriptURLs = new Set();
+      for (let stack = record.initiator.stack; stack; stack = stack.parent) {
+        const callFrames = stack.callFrames || [];
+        for (const frame of callFrames) {
+          if (frame.url) scriptURLs.add(frame.url)
+        }
+      }
+
+      return Array.from(scriptURLs);
     }
 
     return [];
--- a/lighthouse-core/computed/page-dependency-graph.js
+++ b/lighthouse-core/computed/page-dependency-graph.js
@@ -36,7 +36,7 @@ class PageDependencyGraph {
       for (let stack = record.initiator.stack; stack; stack = stack.parent) {
         const callFrames = stack.callFrames || [];
         for (const frame of callFrames) {
-          if (frame.url) scriptURLs.add(frame.url)
+          if (frame.url) scriptURLs.add(frame.url);
         }
       }
 
--- a/lighthouse-core/computed/page-dependency-graph.js
+++ b/lighthouse-core/computed/page-dependency-graph.js
@@ -31,13 +31,19 @@ class PageDependencyGraph {
     if (!record.initiator) return [];
     if (record.initiator.url) return [record.initiator.url];
     if (record.initiator.type === 'script') {
+      // Script initiators have the stack of callFrames from all functions that led to this request.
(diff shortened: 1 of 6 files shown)


patrickhulce:
this is essentially waiting on #6152 to validate the impact of this

patrickhulce:
DZL, do a barrel roll!

brendankenny:
is it weird none of the `yarn test-lantern` numbers changed?

patrickhulce:
@brendankenny no that's expected. Those are based on static traces/devtoolslogs, and this changes the devtoolslog for the full effect.

patrickhulce:
The performance metrics charts look pretty good from DZL on this, so I think we should proceed. Our primary concern would be if we saw substantial increases in TTI and that's not the case. The fact that zero-cost async stacks should already be enabled and this is just propagating that information over to network record initiators makes this a little less concerning as well.

Review away 😃

lighthouse-core/gather/driver.js
```diff
@@ -1401,6 +1398,15 @@ class Driver {
     return this.sendCommand('Runtime.enable');
   }
 
+  /**
```
brendankenny: not sure if this is the best place for this since it's more just turning it on than using it, but it would be nice if there were a comment somewhere to answer "why is this being turned on?" (or maybe more accurately, sometime in the future, "was this important or can I remove this?")

lighthouse-core/computed/page-dependency-graph.js
```diff
@@ -30,9 +30,17 @@ class PageDependencyGraph {
...
+      /** @type {Set<string>} */
+      const scriptURLs = new Set();
+      for (let stack = record.initiator.stack; stack; stack = stack.parent) {
+        const callFrames = stack.callFrames || [];
+        for (const frame of callFrames) {
+          if (frame.url) scriptURLs.add(frame.url);
+        }
+      }
+
+      return Array.from(scriptURLs);
```
brendankenny: unit test for this?

patrickhulce:
@brendankenny if we had jest for smoketests (#8190) it'd be a piece of cake, but it's extremely non-intuitive to understand and the failure mode would be very difficult to debug. As this is mostly an *enablement* thing and isn't meant to suddenly have a big impact on its own (we need to add the logic to start taking advantage of async stacks in the places I mentioned), I think it makes sense to land without it.

I have the smoketest ready to go locally post #8190 or something like it.

PR #5506, opened 2018-06-14
core: split out BaseArtifacts, those always provided by LH itself

brendankenny:
A small refactor of GatherRunner to split the types of the artifacts always provided by LH (`URL`, `fetchTime`, `LighthouseRunWarnings`, etc) from the artifacts provided by gatherers.

This gets rid of the awkward nested arrays around the base artifacts and allows tightening of the types a bit so that gatherers now only return `undefined` or the values of `LH.GathererArtifacts`.

Side effect: with the recent simplification of `passContext`, it seemed to make more sense to add `LighthouseRunWarnings` to the context rather than also pass in `baseArtifacts` to `beforePass`/`pass`/`afterPass`. This has the added benefit of solving how gatherers can raise top-level warnings...they can now just push to `passContext.LighthouseRunWarnings` :)

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -236,8 +236,10 @@ class GatherRunner {
       // Abuse the passContext to pass through gatherer options
       passContext.options = gathererDefn.options || {};
       const artifactPromise = Promise.resolve().then(_ => gatherer.pass(passContext));
+
       const gathererResult = gathererResults[gatherer.name] || [];
       gathererResult.push(artifactPromise);
+      gathererResults[gatherer.name] = gathererResult;
       await GatherRunner.recoverOrThrow(artifactPromise);
     }
   }
@@ -302,8 +304,10 @@ class GatherRunner {
         Promise.reject(pageLoadError) :
         // Wrap gatherer response in promise, whether rejected or not.
         Promise.resolve().then(_ => gatherer.afterPass(passContext, passData));
+
       const gathererResult = gathererResults[gatherer.name] || [];
       gathererResult.push(artifactPromise);
+      gathererResults[gatherer.name] = gathererResult;
       await GatherRunner.recoverOrThrow(artifactPromise);
       log.verbose('statusEnd', status);
     }
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -13,14 +13,14 @@ const constants = require('../config/constants');
 
 const Driver = require('../gather/driver.js'); // eslint-disable-line no-unused-vars
 
+/** @typedef {import('./gatherers/gatherer.js').PhaseResult} PhaseResult */
 /**
  * Each entry in each gatherer result array is the output of a gatherer phase:
  * `beforePass`, `pass`, and `afterPass`. Flattened into an `LH.Artifacts` in
  * `collectArtifacts`.
- * @typedef {{LighthouseRunWarnings: Array<string>, [artifactName: string]: Array<*>}} GathererResults
+ * @typedef {Record<keyof LH.GathererArtifacts, Array<PhaseResult|Promise<PhaseResult>>>} GathererResults
  */
-/** @typedef {{traces: Object<string, LH.Trace>, devtoolsLogs: Object<string, LH.DevtoolsLog>}} TracingData */
-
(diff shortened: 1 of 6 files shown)


brendankenny:
> and allows tightening of the types a bit so that gatherers now only return undefined or the values of LH.GathererArtifacts.

for those writing custom gatherers and using typescript checking, just need to [augment the interface](http://typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation) with their artifact from their `.d.ts` file/module:
```ts
declare global {
  module LH.GathererArtifacts {
    MyFantasticArtifact: {such: 'a', good: 'type'};
  }
}
```

lighthouse-core/gather/gather-runner.js
```diff
@@ -337,52 +336,64 @@ class GatherRunner {
...
+      if (pageLoadFailures.length > Object.keys(gathererArtifacts).length * 0.5) {
         throw pageLoadFailures[0];
       }
     }
 
-    return /** @type {LH.Artifacts } */ (artifacts);
+    // Take only unique LighthouseRunWarnings.
+    baseArtifacts.LighthouseRunWarnings = Array.from(new Set(baseArtifacts.LighthouseRunWarnings));
+
+    return /** @type {LH.Artifacts} */ ({...baseArtifacts, ...gathererArtifacts});
```
brendankenny: can drop this cast when computed artifacts aren't part of the `Artifacts` interface anymore

lighthouse-core/gather/gather-runner.js
```diff
@@ -98,18 +98,16 @@ class GatherRunner {
...
-  static async setupDriver(driver, gathererResults, options) {
+  static async setupDriver(driver, baseArtifacts, options) {
     log.log('status', 'Initializing…');
     const resetStorage = !options.settings.disableStorageReset;
     // Enable emulation based on settings
     await driver.assertNoSameOriginServiceWorkerClients(options.requestedUrl);
-    const userAgent = await driver.getUserAgent();
-    gathererResults.UserAgent = [userAgent];
-    GatherRunner.warnOnHeadless(userAgent, gathererResults);
+    GatherRunner.warnOnHeadless(baseArtifacts);
```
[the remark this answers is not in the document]
patrickhulce: +1 to dropping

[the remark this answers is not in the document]
brendankenny: deleted! Much better line ratio now :)

paulirish:
juicy. lgtm

lighthouse-core/gather/gather-runner.js
```diff
@@ -258,7 +236,8 @@ class GatherRunner {
       // Abuse the passContext to pass through gatherer options
       passContext.options = gathererDefn.options || {};
       const artifactPromise = Promise.resolve().then(_ => gatherer.pass(passContext));
-      gathererResults[gatherer.name].push(artifactPromise);
+      const gathererResult = gathererResults[gatherer.name] || [];
```
patrickhulce: can we ts-ignore if this is not a real thing/really fix if it is? adding the promise to an array that doesn't go anywhere seems worse haha :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -323,7 +302,8 @@ class GatherRunner {
         Promise.reject(pageLoadError) :
         // Wrap gatherer response in promise, whether rejected or not.
         Promise.resolve().then(_ => gatherer.afterPass(passContext, passData));
-      gathererResults[gatherer.name].push(artifactPromise);
+      const gathererResult = gathererResults[gatherer.name] || [];
+      gathererResult.push(artifactPromise);
```
patrickhulce: ditto

lighthouse-core/gather/gather-runner.js
brendankenny: > adding the promise to an array that doesn't go anywhere seems worse haha :)

ha...that is much worse. I'll fix

brendankenny: ok, slightly more verbose, but ts-ignoring would unfortunately handle not only the defined case but also wrong types for `push()`, and these are two out of three places we push to this array, so that seemed like it was defeating the purpose of better types :)

Unfortunately the compiler won't recognize the one-line check `if (!gathererResults[gatherer.name]) throw new Error(...)` as properly guarding the undefined case.

patrickhulce: ah well we tried! fix you got now wfm :)

PR #5509, opened 2018-06-15
core: remove no-mutation-events audit

patrickhulce:
**Summary**
- The `no-mutation-events` audit is the only audit using a large chunk of heavy/slow gatherer code.
- Only 0.8% (3,370/435,697) of sites in HTTPArchive have even a single mutation event listener.
- The gatherer would need fixing to land #5504 
- All the similar DBW audits were replaced by Chrome violations

If this becomes a priority again after passing our new audit process standards, then we still have this diff to look back on or it can become a Chrome violation to be picked up for free :)


patrickhulce:
> you have a few more lines to remove in dbw_tester.js

done

patrickhulce:
pushed now

What were your mutation event listener findings? :)

patrickhulce:
@brendankenny you were 👍 on this, right?

patrickhulce:
> it looks like event-helpers might be only used in the audit as well?

the things I can still learn exist 😮

PR #5517, opened 2018-06-18
misc(scripts): more useful lantern debugging output

patrickhulce:
so lantern tests have been great canary but not all that helpful for debugging :) the git diff didn't always include the URL you needed to look at, it included a lot of lines that weren't relevant, and didn't give clear priority to the most affected URLs

this makes a few tweaks and adds a debug script

new output of the `assert-*` script:
image

### DIFF
--- a/lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
+++ b/lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
@@ -10,50 +10,62 @@
 
 const fs = require('fs');
 const path = require('path');
-const execSync = require('child_process').execSync;
 const constants = require('./constants');
+
+// @ts-ignore
+const chalk = require('chalk');
 
 const INPUT_PATH = process.argv[2] || constants.SITE_INDEX_WITH_GOLDEN_WITH_COMPUTED_PATH;
 const HEAD_PATH = path.resolve(process.cwd(), INPUT_PATH);
 const MASTER_PATH = constants.MASTER_COMPUTED_PATH;
 
-const TMP_DIR = path.join(__dirname, '../../../.tmp');
-const TMP_HEAD_PATH = path.join(TMP_DIR, 'HEAD-for-diff.json');
-const TMP_MASTER_PATH = path.join(TMP_DIR, 'master-for-diff.json');
-
-if (!fs.existsSync(TMP_DIR)) fs.mkdirSync(TMP_DIR);
-
 if (!fs.existsSync(HEAD_PATH) || !fs.existsSync(MASTER_PATH)) {
   throw new Error('Usage $0 <computed file>');
 }
 
-let exitCode = 0;
+const computedResults = require(HEAD_PATH);
+const expectedResults = require(MASTER_PATH);
 
-try {
-  const computedResults = require(HEAD_PATH);
-  const expectedResults = require(MASTER_PATH);
+/** @type {{url: string, maxDiff: number, diffsForSite: DiffForSite[]}[]} */
+const diffs = [];
+for (const entry of computedResults.sites) {
+  // @ts-ignore - should always exist
+  const expectedLantern = expectedResults.sites.find(candidate => entry.url === candidate.url);
+  const actualLantern = entry.lantern;
 
-  const sites = [];
-  for (const entry of computedResults.sites) {
-    const lanternValues = entry.lantern;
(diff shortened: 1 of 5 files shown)


lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
```diff
@@ -10,50 +10,62 @@
 
 const fs = require('fs');
 const path = require('path');
-const execSync = require('child_process').execSync;
 const constants = require('./constants');
 
+// @ts-ignore
+const chalk = require('chalk');
```
brendankenny: chalk isn't in our deps, is that on purpose? (chalk also added its own types at some point, it looks like, though our transitive dep chalk is stuck on an old version)

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
```diff
@@ -10,50 +10,62 @@
...
-    console.log(err.stdout.toString());
-    exitCode = 1;
-  }
-} finally {
-  fs.unlinkSync(TMP_HEAD_PATH);
-  fs.unlinkSync(TMP_MASTER_PATH);
+const computedResults = require(HEAD_PATH);
+const expectedResults = require(MASTER_PATH);
+
+/** @type {{url: string, maxDiff: number, diffsForSite: DiffForSite[]}[]} */
```
brendankenny: this seems like a case for old-fashioned `Array<>`, at least in the surrounding block :)

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
```diff
@@ -10,50 +10,62 @@
...
+const diffs = [];
+for (const entry of computedResults.sites) {
+  // @ts-ignore - should always exist
```
brendankenny: what should always exist?

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
```diff
@@ -10,50 +10,62 @@
...
-  } catch (err) {
-    console.log('❌  FAIL    Changes between expected and computed!\n');
-    console.log(err.stdout.toString());
-    exitCode = 1;
-  }
-} finally {
-  fs.unlinkSync(TMP_HEAD_PATH);
-  fs.unlinkSync(TMP_MASTER_PATH);
+const computedResults = require(HEAD_PATH);
+const expectedResults = require(MASTER_PATH);
```
brendankenny: since master is constant, would it be better to require it directly so the compiler can pick it up?

lighthouse-core/scripts/lantern/debug-url.js
```diff
@@ -0,0 +1,30 @@
...
+
+const INPUT_URL = process.argv[2];
+if (!INPUT_URL) throw new Error('Usage $0: <url>');
+
+const SITE_INDEX_PATH = path.resolve(process.cwd(), constants.SITE_INDEX_WITH_GOLDEN_PATH);
+const SITE_INDEX_DIR = path.dirname(SITE_INDEX_PATH);
+const RUN_ONCE_PATH = path.join(__dirname, 'run-once.js');
+
+const siteIndex = require(SITE_INDEX_PATH);
+// @ts-ignore
```
brendankenny: whats being ignored?

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
patrickhulce: I don't think so unless I'm misunderstanding. TS just seems to put on a compiler error on the JSON for not being a module and then we have to copy/paste the path around...?

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
patrickhulce: heh, alright done :)

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
patrickhulce: fixed

lighthouse-core/scripts/lantern/assert-master-lantern-values-unchanged.js
patrickhulce: done

PR #5523, opened 2018-06-19
misc(scripts): better print-correlations output

patrickhulce:
basically a total re-write of the `print-correlations` script, example new output looks like the below and compares the current state to the baseline master state as well! 🎉 

image


paulirish:
@brendankenny any last thoughts on this one?

PR #5524, opened 2018-06-19
add links to lighthouse docs

kaycebasques:
Updates `helpText` links to point to our docs.

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
+++ b/lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
@@ -27,7 +27,7 @@ class CacheHeaders extends Audit {
       failureDescription: 'Uses inefficient cache policy on static assets',
       helpText:
         'A long cache lifetime can speed up repeat visits to your page. ' +
-        '[Learn more](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#cache-control).',
+        '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/cache-policy).',
       scoreDisplayMode: Audit.SCORING_MODES.NUMERIC,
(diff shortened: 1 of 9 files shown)


lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -27,7 +27,7 @@ class CacheHeaders extends Audit {
       failureDescription: 'Uses inefficient cache policy on static assets',
       helpText:
         'A long cache lifetime can speed up repeat visits to your page. ' +
-        '[Learn more](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching#cache-control).',
+        '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/cache-policy).',
```
patrickhulce: we might want to update this one, we generally want folks to cache for much longer when possible

1 year for immutable static assets, i.e. configure webpack/gulp to embed a hash in the filename so they're unique (ideal)
explicitly add `no-cache` if the resource changes and freshness matters

patrickhulce:
Hey @kaycebasques try to make sure the build is green before mergin', we also have a commit message convention we try to stick to for PR titles and prefer "squash 'n merge" over a merge commit :)

kaycebasques:
Got it, sorry. I figured you would have done that during review.

I think you can enforce squash and merge as the default by disabling the other ones:

image

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
kaycebasques: FYI https://github.com/google/WebFundamentals/issues/6285 is tracking the work to address this feedback in the doc

PR #5525, opened 2018-06-20
core(tsc): use Config class to define Config type

brendankenny:
Previously there was the `Config` class (in `config.js`) and the `LH.Config` type (in `config.d.ts`). The `LH.Config` type implemented `LH.Config.Json`, ensuring that any `LH.Config` was also a valid `LH.Config.Json`.

When passing around an instance of `Config` we called it an `LH.Config`, which meant the compiler was enforcing that it implemented both interfaces, but that means two sources of truth for what a "Config" is and indirect errors if those interfaces were violated.

Based on @patrickhulce's idea in #5451 to enforce interfaces until tsc [supports `@implements` tags](https://github.com/Microsoft/TypeScript/issues/17498), we can eliminate the `d.ts` declaration of `LH.Config` and just use the `Config` class directly, just asserting that the object is a valid `LH.Config.Json` near the end of the constructor in a throwaway variable.

(the rest of the config interfaces staying separate for now because `@typedef` [doesn't support inheritance yet](https://github.com/Microsoft/TypeScript/issues/20077) 🙄 and they aren't clearly defined anywhere with a constructor or single object literal)

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -20,8 +20,8 @@ const Runner = require('../runner');
 /** @typedef {InstanceType<GathererConstructor>} Gatherer */
 
 /**
- * @param {LH.Config['passes']} passes
- * @param {LH.Config['audits']} audits
+ * @param {Config['passes']} passes
+ * @param {Config['audits']} audits
  */
 function validatePasses(passes, audits) {
   if (!Array.isArray(passes)) {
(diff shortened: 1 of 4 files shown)


lighthouse-core/config/config.js
```diff
@@ -303,6 +303,7 @@ const mergeOptionsOfItems = (function(items) {
 class Config {
   /**
    * @constructor
+   * @implements {LH.Config.Json}
```
brendankenny: non-functional for now

PR #5546, opened 2018-06-22
docs(headless-chrome): need to install npm

klick-barakgall:
Following documentation will result in an error when running any `npm` command as it appears like it's not installed along with node

**Summary**
Fixes an issue when following documentation for headless chrome.

Makes docs easier to follow

**Related Issues/PRs**
None that I know of.


klick-barakgall:
I signed it!

PR #5548, opened 2018-06-22
core(tsc): make CPUNode and NetworkNode a discriminated union

brendankenny:
@patrickhulce this is what I mentioned a while back (in #5072). The Node type passed around becomes a union of `CPUNode | NetworkNode`, which means things inside e.g. an `if (node.type !== 'network')` block that `node` becomes a `CPUNode` with no cast or anything.

The downsides are
a) naming (but we can tweak) and
b) double casts of `this` within the `Node` class since the compiler doesn't *know* that `this` isn't an incompatible subclass of `Node` (so first cast to `Node`, then cast to `CPUNode | NetworkNode`.

### DIFF
--- a/lighthouse-core/gather/computed/metrics/lantern-first-contentful-paint.js
+++ b/lighthouse-core/gather/computed/metrics/lantern-first-contentful-paint.js
@@ -8,7 +8,7 @@
 const MetricArtifact = require('./lantern-metric');
 const BaseNode = require('../../../lib/dependency-graph/base-node');
 
-/** @typedef {BaseNode.NodeType} NodeType */
+/** @typedef {BaseNode.Node} Node */
 
 class FirstContentfulPaint extends MetricArtifact {
   get name() {
@@ -27,9 +27,9 @@ class FirstContentfulPaint extends MetricArtifact {
   }
 
   /**
-   * @param {NodeType} dependencyGraph
+   * @param {Node} dependencyGraph
    * @param {LH.Artifacts.TraceOfTab} traceOfTab
-   * @return {NodeType}
+   * @return {Node}
    */
   getOptimisticGraph(dependencyGraph, traceOfTab) {
     const fcp = traceOfTab.timestamps.firstContentfulPaint;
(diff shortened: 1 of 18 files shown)


brendankenny:
> b) double casts of `this` within the `Node` class

other options:
- `@ts-ignore`, downside is that if `Node` or `NodeType` become incompatible it wouldn't be caught in these spots
- [type guard support just landed for jsdoc](https://github.com/Microsoft/TypeScript/issues/25127), so we could have a function that encapsulates the double cast `if (isNodeType(this)) { /* do something NodeTypey with this */ }`, but that involves functional changes that actually execute at runtime whereas this PR only touches non-functional type annotations...

brendankenny:
er, and last option, we don't do this and keep casting. Not a big deal if not :)

wardpeet:
> type guard support just landed for jsdoc, so we could have a function that encapsulates the double cast if (isNodeType(this)) { /* do something NodeTypey with this */ }, but that involves functional changes that actually execute at runtime whereas this PR only touches non-functional type annotations...
I'm probably not getting it 100% but you would always some kind of `if` statement right?

lighthouse-core/gather/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -7,8 +7,8 @@
 
 const MetricArtifact = require('./lantern-metric');
 const Node = require('../../../lib/dependency-graph/node');
-const CPUNode = require('../../../lib/dependency-graph/cpu-node'); // eslint-disable-line no-unused-vars
-const NetworkNode = require('../../../lib/dependency-graph/network-node'); // eslint-disable-line no-unused-vars
+
+/** @typedef {Node.NodeType} NodeType */
```
patrickhulce: is this a necessary thing, or just prefer it over `Node.NodeType` everywhere?

lighthouse-core/lib/dependency-graph/node.js
```diff
@@ -5,6 +5,12 @@
  */
 'use strict';
 
+/**
+ * A union of all types derived from Node, allowing type check disrimination
```
patrickhulce: discrimination :)

lighthouse-core/lib/dependency-graph/node.js
```diff
@@ -5,6 +5,12 @@
...
+ * based on `node.type`. If a new node type is created, it should be added here.
+ * @typedef {import('./cpu-node.js') | import('./network-node.js')} NodeType
```
patrickhulce: how would you feel about calling this one `Node` and the real base class `BaseNode` or something? it feels weird to say that all those methods accept a thing of type `NodeType` instead of saying they expect a `Node`

PR #5555, opened 2018-06-25
core(lhr): add metrics type to details

paulirish:
ref #5008

```js
metricAudit.details.type // 'metric'
metricAudit.details.timespanMs // numeric value
```

Interestingly we don't have any code that consumes metricAudit.rawValue, so no place to have it start consuming this details.timespanMs.


paulirish:
at the moment we dont have any need for this. We'll see when it comes back up during proto-fitting

PR #5556, opened 2018-06-25
core: remove WebInspector.resourceTypes references

patrickhulce:
first step cleanup after NetworkRequest class, we only do string comparison in basically all cases except 2, those are easy to do as a set of strings, so no need for the complexity of the object.


lighthouse-core/audits/network-requests.js
```diff
@@ -46,7 +46,7 @@ class NetworkRequests extends Audit {
           transferSize: record.transferSize,
           statusCode: record.statusCode,
           mimeType: record._mimeType,
-          resourceType: record._resourceType && record._resourceType._name,
+          resourceType: (record._resourceType || 'other').toLowerCase(),
```
brendankenny: lowercase is confusing here, but I'm assuming this is to keep compatibility with old `ResourceType.name`...is it worth keeping it that way or should we just switch to the same capitalization as `Page.ResourceType`?

typings/web-inspector.d.ts
```diff
@@ -44,7 +44,7 @@ declare global {
 
       _initiator: Crdp.Network.Initiator;
       _timing?: Crdp.Network.ResourceTiming;
-      _resourceType?: ResourceType;
+      _resourceType?: Crdp.Page.ResourceType;
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/lib/network-request.js
```diff
@@ -12,10 +12,26 @@
  */
 
 const URL = require('./url-shim');
-const ResourceType = require('../../third-party/devtools/ResourceType');
 
 const SECURE_SCHEMES = ['data', 'https', 'wss', 'blob', 'chrome', 'chrome-extension', 'about'];
 
+/** @type {Record<LH.Crdp.Page.ResourceType, LH.Crdp.Page.ResourceType>} */
```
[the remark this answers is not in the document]
patrickhulce: is there any situation right now in which the type precision you're proposing helps us?

lighthouse-core/audits/network-requests.js
patrickhulce: yeah this was for compat, but we're pre 3.0 official so I'm game to remove if others think it makes sense :)

patrickhulce:
> Can we also delete third-party/devtools/ResourceType.js?

whoops, yes I meant to do that :)

lighthouse-core/audits/network-requests.js
paulirish: eh. let's nuke the toLowerCase. 🔥

PR #5560, opened 2018-06-26
tests(smoke): remove console.timeline() call

patrickhulce:
@paulirish bisected the regression to a V8 roll that removed console.timeline, we call this method in our smoke tests so the test script was erroring and not actually executing our intentional failures :)

### DIFF
--- a/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
+++ b/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
@@ -302,9 +302,6 @@ function deprecationsTest() {
 
   // TODO: some deprecated API messages are not currently being logged correctly.
   // See: https://crbug.com/680832
-  try {
-    MediaStreamTrack.getSources(); // Should eventually FAIL
-  } catch(e) {}
 
   // FAIL
   const xhr = new XMLHttpRequest();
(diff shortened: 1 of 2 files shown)


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -17,7 +17,7 @@ module.exports = [
         score: 0,
         details: {
           items: {
-            length: '>3',
+            length: 5,
```
paulirish: 6️⃣

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -306,8 +306,6 @@ <h2>Do better web tester page</h2>
     MediaStreamTrack.getSources(); // Should eventually FAIL
   } catch(e) {}
 
-  console.timeline(); // Should eventually FAIL
```
patrickhulce: the `Should eventually FAIL` seems to be a signal that at some point in the future we expected that this would trigger a deprecation? 

we assert that 3 deprecations are found in our expectations, the 3 real ones + 2 fake ones (1 fake one) after this PR

PR #5565, opened 2018-06-26
deps(axe): bump axe to latest

paulirish:
changes here: https://github.com/dequelabs/axe-core/releases

Nothing in particular triggered me to bump, though "Prevent color rules from crashing Chrome 66+" looks certainly worthwhile. :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
     "zone.js": "^0.7.3"
   },
   "dependencies": {
-    "axe-core": "3.0.0-beta.2",
+    "axe-core": "^3.0.3",
     "chrome-devtools-frontend": "1.0.422034",
     "chrome-launcher": "^0.10.2",
     "configstore": "^3.1.1",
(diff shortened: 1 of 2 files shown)


paulirish:
@robdodson questions for you. we have two failures with the latest here.

1. `accesskeys`. https://github.com/dequelabs/axe-core/pull/922 removed it from wcag tags, so its never included. luckily your smoketest caught this. could have been a silent failure for a while otherwise. :)   since its not wcag anymore should we remove it? or explicitly include. up to you.
2. `layout-table`. https://github.com/dequelabs/axe-core/pull/828 changed it's implementation and now the smoketest returns it as nonApplicable suggesting that the new [matches fn](https://github.com/dequelabs/axe-core/pull/828/files#diff-b2ba0e0e11a4b5c98e4873d24a9b6442) doesn't match [our table](https://github.com/GoogleChrome/lighthouse/blob/281582439dbf7a45b00e3673840f0b93542d66aa/lighthouse-cli/test/fixtures/a11y/a11y_tester.html#L129). I don't totally follow why... so what should we do about this?

robdodson:
I think removing `role=presentation` could fix 2. If not, also try removing the `th`. I think it's testing that folks aren't using a table to lay stuff out without putting data inside of it.

brendankenny:
Will also fix #5940

PR #5579, opened 2018-06-28
extension: expose devtools hooks reliably

paulirish:


lighthouse-extension/app/src/lighthouse-background.js
```diff
@@ -79,8 +79,10 @@ if (typeof module !== 'undefined' && module.exports) {
     getDefaultCategories,
     listenForStatus,
   };
-} else {
-  // If not require()d, expose on window for devtools, other consumers of file.
+}
+
+if (window) {
```
patrickhulce: `typeof window !== 'undefined'` maybe? :)

PR #5580, opened 2018-06-28
core: convert requestIds before sending to backend

patrickhulce:
#5451 changed our network request ID such that many more IDs had the `:redirect` scheme than before, the DT backend has no notion of these fake `<requestId>:redirect` IDs and we always should have been cleansing them to be regular requestIds.


patrickhulce:
> looks like we can do this on our Network.getResponseBody call, too. it's used by response-compression and scripts gatherers.

They both use the getRequestContent method on driver, we should be fine?

paulirish:
this should fix these errors on paulirish.com

image

PR #5591, opened 2018-06-29
extension(taburl): extension Error: couldn't resolve current tab

wardpeet:
**Summary**

bugfix

When you run the extension and hit enter immediately afterwards, lighthouse tries to start again which will start on about:blank and fails as it can't get an url from the tab.

**Related Issues/PRs**

https://github.com/GoogleChrome/lighthouse/issues?q=is%3Aissue+Error%3A+Couldn%27t+resolve+current+tab+is%3Aclosed

### DIFF
--- a/lighthouse-extension/app/src/popup.js
+++ b/lighthouse-extension/app/src/popup.js
@@ -257,9 +257,11 @@ async function initPopup() {
   // bind Generate Report button
   const generateReportButton = find('#generate-report');
   generateReportButton.addEventListener('click', () => {
-    background.loadSettings().then(settings => {
-      onGenerateReportButtonClick(background, settings);
-    });
+    if (!background.isRunning()) {
+      background.loadSettings().then(settings => {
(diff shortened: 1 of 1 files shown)


lighthouse-extension/app/src/popup.js
```diff
@@ -257,9 +257,11 @@ async function initPopup() {
   // bind Generate Report button
   const generateReportButton = find('#generate-report');
   generateReportButton.addEventListener('click', () => {
-    background.loadSettings().then(settings => {
-      onGenerateReportButtonClick(background, settings);
-    });
+    if (!background.isRunning()) {
```
patrickhulce: micronit: `if (background.isRunning()) return;` either here or onGenerateReportButtonClick matches a bit more with what I might expect, but *awesome* 🔍

wardpeet:
I used a boolean instead. The smoketest didn't work out so no test :(

PR #5606, opened 2018-07-02
core(network-request): cleanup DevTools property names

patrickhulce:
https://www.youtube.com/watch?v=WJ9uhDzN-rA

* Remove all the `_` from the network request properties
* Updates our existing network record fixtures to have no `_`
* Removes functions that didn't need to be functions
* Removes unused fixture and computed artifact
* Removes web-inspector.d.ts typings

-------
Split into 2 commits for easier (maybe?) reviewing, 1 removes the underscores, the next converts the functions and removes the web inspector types


patrickhulce:
@brendankenny I know this is big, but it can conflict with stuff real soon if we make too many changes, so quick look would be nice :) (most of the line volume changes are to fixtures)

PR #5608, opened 2018-07-02
core: bump to v3.0.1

paulirish:
~This is going to the 3.0.x branch`, only to exclude the jest commit, right now.~

jk. master branch. I dbl-checked the jest commit and feel fine about it rolling into the patch rls.


patrickhulce:
needs a rebase then since branch change?

paulirish:
published 3.0.1 to npm.

PR #5610, opened 2018-07-02
tests: [minor] rename robots.txt test file

paulirish:
the only test suite not following the "*-test.js" naming convention.


patrickhulce:
oops :)

PR #5612, opened 2018-07-03
docs(custom-audit): update custom-audit to 3.0.0 api

eduardosada:
**Summary**
Update custom-audit example as latest version changed the api so custom-audit is currently broken.


paulirish:
@eduardosada my man! thanks for hooking this up. :D

PR #5614, opened 2018-07-03
[WIP] report: i18n spike

paulirish:
There appears to be at least 3 primary libraries for working with the icu messageformat.

1. [`format-message`](https://github.com/format-message/format-message) - which made the very useful [Online ICU Message Editor](http://format-message.github.io/icu-message-format-for-translators/editor.html)
1. [`messageformat`](https://github.com/messageformat/messageformat) - which has these [pretty docs](https://messageformat.github.io/messageformat/)
1. [`intl-messageformat`](https://github.com/yahoo/intl-messageformat) - which is part of the larger [FormatJS](https://formatjs.io/) effort.

Also found [ttag/c-3po](https://c-3po.js.org/) which seems cool, though it has more opinionated integrations, afaict.

I chose `intl-messageformat`, originally made by yahoo. It's API follows the ES proposal, so its a prollyfill. Size/dependency-wise it seems good.  

------------

Nothing terribly interesting in the PR, I think. Not yet at least.

- L174/L417 (_anagram!_ 😮 ) is the only use of plurals.
- I added a `Util.i18n()` method, but am happy to handle it differently.

Next move is to extract strings from the rest of the renderer, especially including the `templates.html` file.

### DIFF
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -75,10 +75,10 @@ class CategoryRenderer {
     if (audit.result.scoreDisplayMode === 'error') {
       auditEl.classList.add(`lh-audit--error`);
       const textEl = this.dom.find('.lh-audit__display-text', auditEl);
-      textEl.textContent = 'Error!';
+      textEl.textContent = Util.i18n(CR.strings.errorBang);
       textEl.classList.add('tooltip-boundary');
       const tooltip = this.dom.createChildOf(textEl, 'div', 'tooltip tooltip--error');
-      tooltip.textContent = audit.result.errorMessage || 'Report error: no audit information';
+      tooltip.textContent = audit.result.errorMessage || Util.i18n(CR.strings.errNoAuditInfo);
     } else if (audit.result.explanation) {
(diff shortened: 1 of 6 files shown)


paulirish:
@patrickhulce @brendankenny

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -75,10 +75,10 @@ class CategoryRenderer {
     if (audit.result.scoreDisplayMode === 'error') {
       auditEl.classList.add(`lh-audit--error`);
       const textEl = this.dom.find('.lh-audit__display-text', auditEl);
-      textEl.textContent = 'Error!';
+      textEl.textContent = Util.i18n(CR.strings.errorBang);
       textEl.classList.add('tooltip-boundary');
       const tooltip = this.dom.createChildOf(textEl, 'div', 'tooltip tooltip--error');
-      tooltip.textContent = audit.result.errorMessage || 'Report error: no audit information';
+      tooltip.textContent = audit.result.errorMessage || Util.i18n(CR.strings.errNoAuditInfo);
```
patrickhulce: we might want to develop some guidelines for naming these things, we only have a few and I'm already sensing some divergence ;)

paulirish:
closing as we have a new plan, mostly enabled by #5655

PR #5644, opened 2018-07-09
[WIP] core: i18n spike

patrickhulce:
a slightly different take on the dynamic approach we ditched earlier.

I wanted to pitch this one more time because I ran into 2 issues that felt 😝


lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -25,6 +26,19 @@ const NetworkRequest = require('../../lib/network-request');
...
+  description: 'Resources are blocking the first paint of your page. Consider ' +
+  'delivering critical JS/CSS inline and deferring all non-critical ' +
+  'JS/styles. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).',
+  displayValue: `{itemCount, plural,
+    one {1 resource}
+    other {# resources}
+    } delayed first paint by {timeInMs, number, milliseconds} ms`
+};
+
+const i18n = i18nUtils.createStringFormatter(__filename, UI_STRINGS);
```
patrickhulce: this is a formatting function that has context bundled in explicitly via `__filename` and `UI_STRINGS`

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -198,17 +209,15 @@ class RenderBlockingResources extends Audit {
...
+    if (results.length > 0) {
+      displayValue = i18n(UI_STRINGS.displayValue, {timeInMs: wastedMs, itemCount: results.length});
     }
 
     /** @type {LH.Result.Audit.OpportunityDetails['headings']} */
     const headings = [
-      {key: 'url', valueType: 'url', label: 'URL'},
-      {key: 'totalBytes', valueType: 'bytes', label: 'Size (KB)'},
-      {key: 'wastedMs', valueType: 'timespanMs', label: 'Download Time (ms)'},
+      {key: 'url', valueType: 'url', label: i18n(i18nUtils.UI_STRINGS.columnURL)},
```
patrickhulce: global shared UI_STRINGS can be used too

lighthouse-core/index.js
```diff
@@ -34,6 +35,7 @@ const Config = require('./config/config');
 async function lighthouse(url, flags, configJSON) {
   // TODO(bckenny): figure out Flags types.
   flags = flags || /** @type {LH.Flags} */ ({});
+  i18n.setLocale(flags.locale || 'en-US');
```
patrickhulce: all it takes to swap locales at runtime is a single call here

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+   */
+  createStringFormatter(filename, localStrings) {
+    const mergedStrings = {...UI_STRINGS, ...localStrings};
+    /** @param {string} msg @param {*} [values] */
+    const format = (msg, values) => {
+      const keyname = Object.keys(mergedStrings).find(key => mergedStrings[key] === msg);
+      if (!keyname) throw new Error(`Could not locate: ${msg}`);
+
+      const parsed = MessageParser.parse(msg);
+      parsed.elements
```
patrickhulce: by forcing the use of our own format function first, we can do things like preserve our fancy rounding logic for milliseconds, automatically convert bytes to KB, etc.

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
...
+  '/node_modules',
+  '/lighthouse-cli',
+  '/renderer',
+  '/test/',
+  '-test.js',
+  'lighthouse-viewer',
+  'lighthouse-extension',
+]
+
+function collectAllStringsInDir(dir) {
```
patrickhulce: simple script recurses the directories looking for strings

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+collectAllStringsInDir(path.join(LH_ROOT, 'lighthouse-core'))
+console.log('Done!')
+
+fs.writeFileSync(path.join(LH_ROOT, 'lighthouse-core/lib/locales/en-US.js'), 'module.exports = ' + JSON.stringify(strings, null, 2) + ';')
+
+const gibberish = {}
+for (const [key, value] of Object.entries(strings)) {
```
patrickhulce: it's easy to generate other locales

lighthouse-core/index.js
paulirish: we can use something like this https://github.com/brandonhorst/javascript-environment-lang/blob/master/index.js before we fall back to english. (well i suppose it wouldn't ever fall back)

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+}
+
+collectAllStringsInDir(path.join(LH_ROOT, 'lighthouse-core'))
+console.log('Done!')
+
+fs.writeFileSync(path.join(LH_ROOT, 'lighthouse-core/lib/locales/en-US.js'), 'module.exports = ' + JSON.stringify(strings, null, 2) + ';')
+
+const gibberish = {}
```
paulirish: wrap into a `createPsuedoLocaleStrings` fn (or 'gibberish')

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+      }
+    }
+  }
+}
+
+collectAllStringsInDir(path.join(LH_ROOT, 'lighthouse-core'))
+console.log('Done!')
+
+fs.writeFileSync(path.join(LH_ROOT, 'lighthouse-core/lib/locales/en-US.js'), 'module.exports = ' + JSON.stringify(strings, null, 2) + ';')
```
paulirish: template literal for this multiline thang

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
...
+
+    if (fs.statSync(fullPath).isDirectory()) {
+      collectAllStringsInDir(fullPath)
+    } else {
+      if (name.endsWith('.js')) {
+        console.log('Collecting from', relativePath)
+        const mod = require(fullPath)
+        if (!mod.UI_STRINGS) continue
+        for (const [key, value] of Object.entries(mod.UI_STRINGS)) {
+          strings[`${relativePath}!#${key}`] = value
```
paulirish: wdyt about just adopting the basic chrome i18n format here: 


```js
        "relativePathAndKey": {
          "message": "stringValue",
          // "description": "..."
        },
```

Like.. we _could_ translate again into that format, but perhaps its nicer to just start with it here.
also it looks like git diff doesnt handle small changes in these strings files well. considers the WHOLE block to be a change.

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -25,6 +26,19 @@ const NetworkRequest = require('../../lib/network-request');
 // to possibly be non-blocking (and they have minimal impact anyway).
 const MINIMUM_WASTED_MS = 50;
 
+const UI_STRINGS = {
```
paulirish: wdyt about `strings`?   the yelling hurts me. :(

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+  },
+};
+
+module.exports = {
+  UI_STRINGS,
+  /**
+   * @param {string} filename
+   * @param {Record<string, string>} localStrings
+   */
+  createStringFormatter(filename, localStrings) {
```
paulirish: localStrings => fileStrings ?  or moduleStrings 

local had me thinking local to the i18n file for a sec..

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+    const mergedStrings = {...UI_STRINGS, ...localStrings};
+    /** @param {string} msg @param {*} [values] */
+    const format = (msg, values) => {
```
paulirish: formatFn (or w/e) to differentiate from all the other 'format's in this file

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+      const parsed = MessageParser.parse(msg);
+      parsed.elements
+        .filter(el => el.format && el.format.style === 'bytes')
+        .forEach(el => (values[el.id] = values[el.id] / 1024));
+
+      const filenameToLookup = keyname in UI_STRINGS ? __filename : filename;
+      const lookupKey = path.relative(LH_ROOT, filenameToLookup) + '!#' + keyname;
+      const localeStrings = LOCALES[locale] || {};
+      const localeForMessageFormat = locale === 'gibberish' ? 'de-DE' : locale;
+      const formatter = new MessageFormat(localeStrings[lookupKey] || msg, localeForMessageFormat, formats);
```
paulirish: so we prefer to pull it from the locale file, but if its not present we fall back to the msg passed in? can you comment why we want that fallback behavior?

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+
+      const parsed = MessageParser.parse(msg);
+      parsed.elements
+        .filter(el => el.format && el.format.style === 'bytes')
+        .forEach(el => (values[el.id] = values[el.id] / 1024));
+
+      const filenameToLookup = keyname in UI_STRINGS ? __filename : filename;
+      const lookupKey = path.relative(LH_ROOT, filenameToLookup) + '!#' + keyname;
+      const localeStrings = LOCALES[locale] || {};
+      const localeForMessageFormat = locale === 'gibberish' ? 'de-DE' : locale;
```
paulirish: g3 uses "en-XA" for psuedo locales. (and also en-XB and en-XC)

```
	"en-XA",  # Accented English
```

i'm okay doing that replacement here, though we could just switch to 'en-XA' everywhere now.

lighthouse-core/lib/i18n.js
paulirish: sg. i like it. let's split into its own fn

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,78 @@
...
+      const keyname = Object.keys(mergedStrings).find(key => mergedStrings[key] === msg);
+      if (!keyname) throw new Error(`Could not locate: ${msg}`);
+
+      const parsed = MessageParser.parse(msg);
+      parsed.elements
+        .filter(el => el.format && el.format.style === 'bytes')
+        .forEach(el => (values[el.id] = values[el.id] / 1024));
+
+      const filenameToLookup = keyname in UI_STRINGS ? __filename : filename;
+      const lookupKey = path.relative(LH_ROOT, filenameToLookup) + '!#' + keyname;
```
paulirish: btw is `!#` a convention?

lighthouse-core/index.js
patrickhulce: oh yeah this was me being lazy for WIP, I'd have `setLocale` just use the default locale determined by `MessageFormat.defaultLocale` when this is undefined

lighthouse-core/scripts/collect-strings.js
```diff
@@ -0,0 +1,66 @@
+const fs = require('fs')
+const path = require('path')
+
+const LH_ROOT = path.join(__dirname, '../../')
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/scripts/collect-strings.js
patrickhulce: done

lighthouse-core/scripts/collect-strings.js
patrickhulce: done

lighthouse-core/scripts/collect-strings.js
patrickhulce: yeah sg, done

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: I'm on board for anything we reasonably think will never be used in any other context, I'd love to add another script for essentially linting to make sure all UI_STRINGS get exported, all get stuck into `i18n` functions, etc.

lighthouse-core/lib/i18n.js
patrickhulce: totally 👍 done

lighthouse-core/lib/i18n.js

lighthouse-core/lib/i18n.js
patrickhulce: perfect, sg 👍

lighthouse-core/lib/i18n.js
patrickhulce: totes

lighthouse-core/lib/i18n.js
patrickhulce: not that I know of, just seemed like some good characters that won't appear in a valid path and look good for separating things :)

lighthouse-core/config/constants.js
```diff
@@ -39,6 +39,7 @@ const defaultSettings = {
 
   // the following settings have no defaults but we still want ensure that `key in settings`
   // in config will work in a typechecked way
+  locale: null, // default determined by the intl library
```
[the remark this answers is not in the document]
paulirish: we don't need to block this PR on this decision IMO.

but we should be sniffing the user's locale somewhere, right?  in both the CLI and browser case.

paulirish:
Shall we land this so we can start expanding our coverage?

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -198,17 +209,15 @@ class RenderBlockingResources extends Audit {
...
+    if (results.length > 0) {
+      displayValue = i18n(UIStrings.displayValue, {timeInMs: wastedMs, itemCount: results.length});
     }
 
     /** @type {LH.Result.Audit.OpportunityDetails['headings']} */
     const headings = [
-      {key: 'url', valueType: 'url', label: 'URL'},
-      {key: 'totalBytes', valueType: 'bytes', label: 'Size (KB)'},
-      {key: 'wastedMs', valueType: 'timespanMs', label: 'Download Time (ms)'},
+      {key: 'url', valueType: 'url', label: i18n(i18nUtils.UIStrings.columnURL)},
```
paulirish: naming-wise, how about: `uiS(i18n.UIStrings.columnURL)`
bikeshedding wise, "i18n" feels most like a namespace so i hesitate to use it for a method name.

patrickhulce:
> Shall we land this so we can start expanding our coverage?

let's!

patrickhulce:
just need an approval ;)

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: chatted outside and went with `str_` for simplicity and easy future refactoring :)

lighthouse-core/lib/i18n.js
```diff
@@ -0,0 +1,99 @@
...
+module.exports = {
+  UIStrings,
+  /**
+   * @param {string} filename
+   * @param {Record<string, string>} fileStrings
+   */
+  createStringFormatter(filename, fileStrings) {
+    const mergedStrings = {...UIStrings, ...fileStrings};
+
+    /** @param {string} msg @param {*} [values] */
```
brendankenny: same

PR #5655, opened 2018-07-11
core(i18n): localize strings at end of run

patrickhulce:
An example of how we might be able to "fix" the strings before shipping the LHR, basically the i18n formatting function just returns a reference to the string to be used which is stored in a map, then at the end before returning, runner asks i18n to replace all the references in the LHR with the string of the appropriate locale. BOOM done! a log is kept of what replacements were made and added to the LHR as `localeLog` so that later the report could be translated again without re-running.

image

### DIFF
--- a/lighthouse-core/config/default-config.js
+++ b/lighthouse-core/config/default-config.js
@@ -5,9 +5,21 @@
  */
 'use strict';
 
-const constants = require('./constants');
-
 /* eslint-disable max-len */
+
+const constants = require('./constants');
+const i18n = require('../lib/i18n');
+
+const UIStrings = {
+  performanceCategoryTitle: 'Performance',
+  metricGroupTitle: 'Metrics',
+  loadOpportunitiesGroupTitle: 'Opportunities',
+  loadOpportunitiesGroupDescription: 'These are opportunities to speed up your application by optimizing the following resources.',
+  diagnosticsGroupTitle: 'Diagnostics',
+  diagnosticsGroupDescription: 'More information about the performance of your application.',
+};
+
+const str_ = i18n.createStringFormatter(__filename, UIStrings);
 
 module.exports = {
   settings: constants.defaultSettings,
@@ -188,15 +200,15 @@ module.exports = {
 
   groups: {
     'metrics': {
-      title: 'Metrics',
+      title: str_(UIStrings.metricGroupTitle),
     },
     'load-opportunities': {
-      title: 'Opportunities',
-      description: 'These are opportunities to speed up your application by optimizing the following resources.',
+      title: str_(UIStrings.loadOpportunitiesGroupTitle),
+      description: str_(UIStrings.loadOpportunitiesGroupDescription),
     },
     'diagnostics': {
-      title: 'Diagnostics',
-      description: 'More information about the performance of your application.',
+      title: str_(UIStrings.diagnosticsGroupTitle),
+      description: str_(UIStrings.diagnosticsGroupDescription),
     },
     'a11y-color-contrast': {
       title: 'Color Contrast Is Satisfactory',
@@ -246,7 +258,7 @@ module.exports = {
   },
   categories: {
     'performance': {
-      title: 'Performance',
+      title: str_(UIStrings.performanceCategoryTitle),
       auditRefs: [
         {id: 'first-contentful-paint', weight: 3, group: 'metrics'},
         {id: 'first-meaningful-paint', weight: 1, group: 'metrics'},
@@ -409,3 +421,9 @@ module.exports = {
     },
   },
 };
+
+// Use `defineProperty` so that the strings are accesible from original but ignored when we copy it
+Object.defineProperty(module.exports, 'UIStrings', {
+  enumerable: false,
+  get: () => UIStrings,
+});
--- a/lighthouse-core/config/default-config.js
+++ b/lighthouse-core/config/default-config.js
@@ -19,7 +19,7 @@ const UIStrings = {
   diagnosticsGroupDescription: 'More information about the performance of your application.',
 };
 
-const str_ = i18n.createStringFormatter(__filename, UIStrings);
+const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
 
 module.exports = {
   settings: constants.defaultSettings,
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -7,25 +7,31 @@
 
 const path = require('path');
 const isDeepEqual = require('lodash.isequal');
+const log = require('lighthouse-logger');
 const MessageFormat = require('intl-messageformat').default;
 const MessageParser = require('intl-messageformat-parser');
 const LOCALES = require('./locales');
 
 const LH_ROOT = path.join(__dirname, '../../');
 
-try {
-  // Node usually doesn't come with the locales we want built-in, so load the polyfill.
-  // In browser environments, we won't need the polyfill, and this will throw so wrap in try/catch.
+(() => {
+  // Node usually doesn't come with the locales we want built-in, so load the polyfill if we can.
 
-  // @ts-ignore
-  const IntlPolyfill = require('intl');
-  if (!IntlPolyfill.NumberFormat) throw new Error('Invalid polyfill');
+  try {
+    // @ts-ignore
+    const IntlPolyfill = require('intl');
+    // In browser environments where we don't need the polyfill, this won't exist
+    if (!IntlPolyfill.NumberFormat) return;
 
-  // @ts-ignore
-  Intl.NumberFormat = IntlPolyfill.NumberFormat;
-  // @ts-ignore
-  Intl.DateTimeFormat = IntlPolyfill.DateTimeFormat;
-} catch (_) {}
+    // @ts-ignore
+    Intl.NumberFormat = IntlPolyfill.NumberFormat;
+    // @ts-ignore
+    Intl.DateTimeFormat = IntlPolyfill.DateTimeFormat;
+  } catch (_) {
+    log.warn('i18n', 'Failed to install `intl` polyfill');
+  }
+})();
+
 
 const UIStrings = {
   ms: '{timeInMs, number, milliseconds}\xa0ms',
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -135,7 +135,8 @@ module.exports = {
       if (!keyname) throw new Error(`Could not locate: ${template}`);
 
       const filenameToLookup = keyname in UIStrings ? __filename : filename;
-      const key = path.relative(LH_ROOT, filenameToLookup) + '!#' + keyname;
+      const unixStyleFilename = path.relative(LH_ROOT, filenameToLookup).replace(/\\/g, '/');
+      const key = unixStyleFilename + '!#' + keyname;
       const keyUsages = formattedStringUsages.get(key) || [];
       keyUsages.push({key, template, values});
       formattedStringUsages.set(key, keyUsages);
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -161,6 +161,7 @@ function replaceLocaleStringReferences(lhr, locale) {
     for (const [property, value] of Object.entries(objectInLHR)) {
       const currentPathInLHR = pathInLHR.concat([property]);
 
+      // Check to see if the value in the LHR looks like a string reference. If it is, replace it.
       if (typeof value === 'string' && /.* \| .* # \d+$/.test(value)) {
         // @ts-ignore - Guaranteed to match from .test call above
         const [_, templateID, usageIndex] = value.match(/(.*) # (\d+)$/);
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -44,15 +44,15 @@ const formats = {
 };
 
 /**
- * @param {string} template
+ * @param {string} icuMessage
  * @param {Record<string, *>} [values]
  */
-function preprocessMessageValues(template, values) {
+function _preprocessMessageValues(icuMessage, values) {
   if (!values) return;
 
   const clonedValues = JSON.parse(JSON.stringify(values));
-  const parsed = MessageParser.parse(template);
-  // Round all milliseconds to 10s place
+  const parsed = MessageParser.parse(icuMessage);
+  // Round all milliseconds to the nearest 10
   parsed.elements
     .filter(el => el.format && el.format.style === 'milliseconds')
     .forEach(el => (clonedValues[el.id] = Math.round(clonedValues[el.id] / 10) * 10));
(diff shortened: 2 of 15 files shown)


lighthouse-core/lib/i18n.js
```diff
@@ -67,24 +99,18 @@ module.exports = {
...
-      const localeForMessageFormat = locale === 'en-XA' ? 'de-DE' : locale;
-
-      const formatter = new MessageFormat(messageForMessageFormat, localeForMessageFormat, formats);
-      return formatter.format(values);
+      const key = path.relative(LH_ROOT, filenameToLookup) + '!#' + keyname;
+      const keyUsages = formattedStringUsages.get(key) || [];
+      keyUsages.push({key, template, values});
+      formattedStringUsages.set(key, keyUsages);
+
+      return `${key}#${keyUsages.length - 1}`;
```
patrickhulce: just a reference is returned here instead of the real string

lighthouse-core/lib/i18n.js
```diff
@@ -96,4 +122,37 @@ module.exports = {
...
+        if (typeof value === 'string' && /.*!#.*#\d+$/.test(value)) {
+          const [_, templateKey, usageIndex] = value.match(/(.*)#(\d+)$/)
+          const templateLogRecord = log[templateKey] || {}
+          const usages = formattedStringUsages.get(templateKey) || []
+          const usage = usages[usageIndex]
+
+          const occurrences = templateLogRecord.occurrences || []
+          occurrences.push({values: usage.values, path: currentPath})
+
+          const {message, template} = formatTemplate(locale, templateKey, usage.template, usage.values);
```
patrickhulce: we do the real formatting here in the specified locale

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3416,5 +3416,144 @@
...
+        {
+          "path": [
+            "audits",
+            "interactive",
+            "description"
+          ]
+        }
+      ]
+    },
+    "lighthouse-core/lib/i18n.js!#ms": {
```
patrickhulce: a log keeps track of all the strings we used

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3416,5 +3416,144 @@
...
+      "occurrences": [
+        {
+          "values": {
+            "timeInMs": 4927.278
+          },
+          "path": [
```
patrickhulce: as well as the path in the LHR of where you can find them to replace them

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3416,5 +3416,144 @@
       "title": "Crawling and Indexing",
       "description": "To appear in search results, crawlers need access to your app."
     }
+  },
+  "localeLog": {
```
patrickhulce: this will get big, but maybe it can be an optional thing?

paulirish: Yah +1 to this being an option that's off by default.

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3416,5 +3416,144 @@
...
+          },
+          "path": [
+            "audits",
+            "render-blocking-resources",
+            "displayValue"
+          ]
+        }
+      ]
+    },
+    "lighthouse-core/lib/i18n.js!#columnURL": {
```
paulirish: how about [JSONPath](http://goessner.net/articles/JsonPath/index.html#e2)?


```js
    "lighthouse-core/lib/i18n.js!#columnURL": {
      "occurrences": [
        {
          "path": [
            "audits",
            "render-blocking-resources",
            "details",
            "headings",
            "0",
            "label"
          ]
        }
      ]
    },
```

could become:

```js
    "lighthouse-core/lib/i18n.js!#columnURL": {
      "occurrences": ["$.audits['render-blocking-resources'].details.headings[0].label"],
    },
```

obviously it's more dense, but I think its a lot more readable. 

(actually looks like [lodash's get](https://lodash.com/docs/4.17.10#get) also supports this style, anyhow.) 
We don't really need ALL of jsonpath, just the basics.

lighthouse-core/config/default-config.js
```diff
@@ -406,3 +414,7 @@ module.exports = {
     },
   },
 };
+
+Object.defineProperty(module.exports, 'UIStrings', {
```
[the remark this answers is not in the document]
patrickhulce: sure :)

it's because config data gets cloned and validated and whatnot so we want a getter and not put the real strings on the actual object

lighthouse-core/test/results/sample_v2.json
patrickhulce: do you want JSONPath or [what lodash get supports](https://github.com/lodash/lodash/blob/4.17.5/lodash.js#L6727-L6736)? ;)

I'm on board for this 👍

patrickhulce: I should've said, "I'm on board for this, so long as we use someone else's getter code :)"

paulirish: sg! based on their docs, the string form with lodash get lgtm

lighthouse-core/config/default-config.js
```diff
@@ -5,9 +5,21 @@
  */
 'use strict';
 
+/* eslint-disable max-len */
+
 const constants = require('./constants');
+const i18n = require('../lib/i18n');
 
-/* eslint-disable max-len */
+const UIStrings = {
```
patrickhulce: FWIW @brendankenny I think these are the only strings we really need for performance section, so it's relatively small

if you want to be noodling on a much better config solution before we start on the other categories, nows a good time :)

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,149 @@ const formats = {
...
+
+/**
+ * @typedef StringUsage
+ * @prop {string} key
+ * @prop {string} template
+ * @prop {*} [values]
+ */
+
+/** @type {Map<string, StringUsage[]>} */
+const formattedStringUsages = new Map();
```
patrickhulce: this is the new singleton, we might want to put some disclaimers into README somewhere about YMMV using LH as a long-lived module not in a worker. CLI/DevTools/Extension all don't really care about these types of concerns so inevitably we'll make some trade-offs not great for node module users (example: #5683)

patrickhulce:
I think this is ready for ready for re-review.

I'm going to throw one more thought out there as I was thinking about report strings. instead of `lhr.localeLog`...

```js
lhr.localization = {
  log: {...},
  uiMessages: {},
}
```

thoughts?

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+   * @param {LH.Locale} locale
    */
-  setLocale(newLocale) {
-    if (!newLocale) return;
-    locale = newLocale;
+  replaceLocaleStringReferences(lhr, locale) {
+    /**
+     * @param {*} objectInLHR
+     * @param {LH.LocaleLog} log
+     * @param {string[]} path
```
paulirish: only for testing?

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
   },
   /**
-   * @param {LH.Locale|null} [newLocale]
+   * @param {LH.Result} lhr
+   * @param {LH.Locale} locale
    */
-  setLocale(newLocale) {
-    if (!newLocale) return;
-    locale = newLocale;
+  replaceLocaleStringReferences(lhr, locale) {
```
paulirish: naming

okay we have:
string IDs?
string references?
object paths?

i'm also seeing: template key, key name, locale string reference, keyusages / usages. 

can we make these names a bit more concrete and consistent?

typings/lhr.d.ts
```diff
@@ -6,6 +6,12 @@
 
 declare global {
   module LH {
+    export type LocaleLogEntry = string | {path: string, values: any};
+
+    export interface LocaleLog {
```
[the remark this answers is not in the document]
patrickhulce: how about 
```
lhr.i18n.messages
```

:)

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+    } else {
+      if (/]|"|'/.test(property)) throw new Error(`Cannot handle "${property}" in i18n`);
+      pathAsString += `[${property}]`;
+    }
+  }
+
+  return pathAsString;
 }
 
 module.exports = {
```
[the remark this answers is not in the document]
patrickhulce: it was whatever is used inside the module gets defined outside everything else is inline, I'll just move it all out though

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+  const valuesForMessageFormat = preprocessMessageValues(template, values);
+
+  const formatter = new MessageFormat(templateForMessageFormat, localeForMessageFormat, formats);
+  const message = formatter.format(valuesForMessageFormat);
+
+  return {message, template: templateForMessageFormat};
+}
+
+/** @param {string[]} path */
+function formatPathAsString(path) {
```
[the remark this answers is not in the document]

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+const formattedStringUsages = new Map();
+
+/**
+ *
+ * @param {LH.Locale} locale
+ * @param {string} templateKey
+ * @param {string} template
+ * @param {*} [values]
+ */
+function formatTemplate(locale, templateKey, template, values) {
```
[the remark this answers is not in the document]

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
-      // fallback to the original english message if we couldn't find a message in the specified locale
-      // better to have an english message than no message at all, in some number cases it won't even matter
-      const messageForMessageFormat = localeString || msg;
-      // when using accented english, force the use of a different locale for number formatting
-      const localeForMessageFormat = locale === 'en-XA' ? 'de-DE' : locale;
-
-      const formatter = new MessageFormat(messageForMessageFormat, localeForMessageFormat, formats);
-      return formatter.format(values);
+      const unixStyleFilename = path.relative(LH_ROOT, filenameToLookup).replace(/\\/g, '/');
+      const key = unixStyleFilename + '!#' + keyname;
```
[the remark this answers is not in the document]
patrickhulce: sg, done

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+      const keyUsages = formattedStringUsages.get(key) || [];
+      keyUsages.push({key, template, values});
```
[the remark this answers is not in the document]
patrickhulce: how did you get that? are you looking at the test output? the tests invoke FCP `.meta` lots of times so that's probably what you're seeing

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
-      // better to have an english message than no message at all, in some number cases it won't even matter
-      const messageForMessageFormat = localeString || msg;
-      // when using accented english, force the use of a different locale for number formatting
-      const localeForMessageFormat = locale === 'en-XA' ? 'de-DE' : locale;
-
-      const formatter = new MessageFormat(messageForMessageFormat, localeForMessageFormat, formats);
-      return formatter.format(values);
+      const unixStyleFilename = path.relative(LH_ROOT, filenameToLookup).replace(/\\/g, '/');
+      const key = unixStyleFilename + '!#' + keyname;
+      const keyUsages = formattedStringUsages.get(key) || [];
```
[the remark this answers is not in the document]
patrickhulce: yeah it'd be unexpected for more than 1 audit title to be used, but for strings that will be used for table cells it will be expected

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+      keyUsages.push({key, template, values});
+      formattedStringUsages.set(key, keyUsages);
+
+      return `${key}#${keyUsages.length - 1}`;
```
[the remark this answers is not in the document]
patrickhulce: sg done

lighthouse-core/lib/i18n.js
patrickhulce: are you referring to why path is an array of strings in this function signature? it's so that we can easily push onto it and convert the name at each step rather than have two functions that build from an array *and* one that builds incrementally

lighthouse-core/lib/i18n.js
patrickhulce: yeah this is a bit confusing, I'll do a pass to try to make them all consistent

patrickhulce:
alright @paulirish I think the names have been sorted out except this area https://github.com/GoogleChrome/lighthouse/pull/5655/files#diff-7e3be24bffdebd4c067d12d0596858f9R171

now that it's not log and we also have usages saved in our `formattedStringUsages` not sure what to call these two different things

lighthouse-core/lib/i18n.js
```diff
@@ -43,57 +41,150 @@ const formats = {
...
+  getDefaultLocale() {
+    const defaultLocale = MessageFormat.defaultLocale;
+    if (defaultLocale in LOCALES) return /** @type {LH.Locale} */ (defaultLocale);
+    return 'en-US';
+  },
   /**
    * @param {string} filename
    * @param {Record<string, string>} fileStrings
    */
   createStringFormatter(filename, fileStrings) {
```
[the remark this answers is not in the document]
paulirish: oh wait actually it doesnt create the formatter at this point at all. that seems to happen much later.

lighthouse-core/lib/i18n.js
```diff
@@ -42,58 +44,164 @@ const formats = {
...
-    locale = newLocale;
-  },
+  function replaceInObject(objectInLHR, messages, pathInLHR = []) {
+    if (typeof objectInLHR !== 'object' || !objectInLHR) return;
+
+    for (const [property, value] of Object.entries(objectInLHR)) {
+      const currentPathInLHR = pathInLHR.concat([property]);
+
+      // Check to see if the value in the LHR looks like a string reference. If it is, replace it.
+      if (typeof value === 'string' && /.* \| .* # \d+$/.test(value)) {
```
paulirish: wanna throw in the param groups that you have on 177 so these look more similar?

paulirish:
@brendankenny u want a pass?

lighthouse-core/lib/i18n.js
```diff
@@ -20,6 +19,8 @@ try {
 
   // @ts-ignore
   const IntlPolyfill = require('intl');
+  if (!IntlPolyfill.NumberFormat) throw new Error('Invalid polyfill');
```
patrickhulce: @paulirish FYI this was my fix to what you discovered in #5702

paulirish: ah cool. nice. 

The msg is a little weird though. Seems like this case is described as: "Polyfill is a no-op due to existing implementation". The fact that we exit via throwing is a little random. can we do a IIFE and return instead?

patrickhulce: yeah IFFE makes a lot more sense, done 👍

paulirish:
👍

PR #5661, opened 2018-07-12
core(emulation): use setTouchEmulationEnabled

### LINKED ISSUE #5659: When running Lighthouse from Webpagetest there is no support of touch device
1. Run LH on <affected url>
Test-URL: https://tolwin.spdns.org/test/test_touch.html

URL of webpagetest run: https://www.webpagetest.org/result/180712_GP_88a851ffa3165b62cc505c566d91d4c0/

URL of Lighthouse report from that WPT run: https://www.webpagetest.org/lighthouse.php?test=180712_GP_88a851ffa3165b62cc505c566d91d4c0&run=1

Running a standard test on webpagetest simulates a mobile device with touch support. That is the correct/expected behavior.
But the Lighthouse report, also run from webpagetest seems to be running as a device WITHOUT touch support.

You can see in the screenshots from WPT and from Lighthouse report that they are different. The testpage itself is not doing much useful, other than trying to detect a device with touch support. 

NOTE: It is not about the question, if my touch detection is correct or not. It is just about the question why WPT is different from Lighthouse. Both tools should be either "touch detected" or "no touch detected".
Personally I do think Lighthouse is wrong here, not WPT.
 
Lighthouse should also run as a device with touch support.

#### Environment Information
* Affected Channels: The version on webpagetest.org, https://www.webpagetest.org/lighthouse (not sure if this is a bug for webpagetest or for Lighthouse then)
* Lighthouse version: 3.0.2
* Node.js version: <unknown>
* Operating System: <unknown>

**Related issues**

patrickhulce:
**Related Issues/PRs**
fixes #5659

### DIFF
--- a/lighthouse-core/lib/emulation.js
+++ b/lighthouse-core/lib/emulation.js
@@ -64,10 +64,7 @@ function enableNexus5X(driver) {
     // Network.enable must be called for UA overriding to work
     driver.sendCommand('Network.enable'),
     driver.sendCommand('Network.setUserAgentOverride', NEXUS5X_USERAGENT),
-    driver.sendCommand('Emulation.setEmitTouchEventsForMouse', {
-      enabled: true,
-      configuration: 'mobile',
-    }),
+    driver.sendCommand('Emulation.setTouchEmulationEnabled', {enabled: true}),
   ]);
 }
 


paulirish:
Yup. this is right. Sorry I messed up back then. DevTools sets both. But `Emulation.setEmitTouchEventsForMouse` is only useful for input, which we dont have any of. :)

PR #5662, opened 2018-07-12
core(traces): Enable V8 CPU Profiler in traces

G018:
I was writing a custom audit to check for specific functions that take a long time blocking resources and I couldn't understand why lighthouse traces didn't match those provided by dev tools performance.

It turns out the traces for V8 profiling were disabled in #1444 because of a [formatting bug in chrome traces](https://bugs.chromium.org/p/v8/issues/detail?id=5820). However, this was fixed over a year ago.

Here is a small change to re-enable them, it didn't seem to affect any tests on my machine.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -92,9 +92,8 @@ class Driver {
       'disabled-by-default-devtools.timeline',
       'disabled-by-default-devtools.timeline.frame',
       'disabled-by-default-devtools.timeline.stack',
-      // Flipped off until bugs.chromium.org/p/v8/issues/detail?id=5820 is fixed in Stable
-      // 'disabled-by-default-v8.cpu_profiler',
-      // 'disabled-by-default-v8.cpu_profiler.hires',
+      'disabled-by-default-v8.cpu_profiler',
+      'disabled-by-default-v8.cpu_profiler.hires',
       'disabled-by-default-devtools.screenshot',
     ];
   }


paulirish:
Last time I looked at this a LH trace of theverge without any V8 sampling
data was 41M. And with high-resolution sampling data it was 57M.

Given that, I think it's worth it to flip this on.

patrickhulce:
Just so my thoughts are clear, my primary concern here is increased WPT crashes and the cost of trace retrieval. I'm not really worried at all about disk space 😃 

> Last time I looked at this a LH trace of theverge without any V8 sampling
data was 41M. And with high-resolution sampling data it was 57M.

AFAIK, Lighthouse won't be leveraging any of this extra information, and we're trying to make the trace events we need *much smaller*, so bumping the default trace size ~50% is a big step backwards. This seems like a prime use case for `--additional-trace-categories`, no?

G018:
> Maybe we just collect some stats on impact to trace size to show it's not a big deal?

I did some test runs locally and it seems to increases an average of 15-30%, however it will likely depend on how much JS is executed by the web page.

> AFAIK, Lighthouse won't be leveraging any of this extra information

Doesn't seemed like it, identical reports were generated from disk data with and without the profiling. However, it would be great to see audits with feedback down to the function in the future.

> This seems like a prime use case for `--additional-trace-categories`, no?

Didn't find that setting before submitting the PR, this would definitely work on my use case. Maybe just update the misleading comment with a mention to the flag?

patrickhulce:
> Maybe just update the misleading comment with a mention to the flag?

Definitely 👍 can probably link this discussion as well 😆 

@paulirish are you cool to hold off making this the default or do you feel strongly we should bring it back?

paulirish:
@patrickhulce i'm cool holding off. :)

PR #5673, opened 2018-07-17
report: add helper method for getting final screenshot datauri

paulirish:
requested by psi

### DIFF
--- a/lighthouse-core/report/html/renderer/performance-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/performance-category-renderer.js
@@ -192,6 +192,19 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     element.appendChild(passedElem);
     return element;
   }
+
+  /**
+   * @param {LH.ReportResult.Category} category
+   * @return {null|string}
+   * @override
+   */
+  static getLastScreenshot(category) {
+    const auditRef = category.auditRefs.find(audit => audit.id === 'screenshot-thumbnails');
+    if (!auditRef || !auditRef.result) return null;
+    const screenshots = auditRef.result.details.items;
+    if (!screenshots.length) return null;
+    return `data:image/jpeg;base64,${screenshots.slice(-1)[0].data}`;
+  }
 }
 
 if (typeof module !== 'undefined' && module.exports) {
--- a/lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
+++ b/lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
@@ -17,6 +17,7 @@ const DetailsRenderer = require('../../../../report/html/renderer/details-render
 const CriticalRequestChainRenderer = require(
     '../../../../report/html/renderer/crc-details-renderer.js');
 const CategoryRenderer = require('../../../../report/html/renderer/category-renderer.js');
+
 const ReportRenderer = require('../../../../report/html/renderer/report-renderer.js');
 const sampleResults = require('../../../results/sample_v2.json');
 
@@ -26,15 +27,14 @@ const TEMPLATE_FILE = fs.readFileSync(__dirname +
 describe('PerfCategoryRenderer', () => {
   let category;
   let renderer;
+  global.CategoryRenderer = CategoryRenderer;
+  const PerformanceCategoryRenderer =
+      require('../../../../report/html/renderer/performance-category-renderer.js');
 
   beforeAll(() => {
     global.URL = URL;
     global.Util = Util;
     global.CriticalRequestChainRenderer = CriticalRequestChainRenderer;
-    global.CategoryRenderer = CategoryRenderer;
-
-    const PerformanceCategoryRenderer =
-        require('../../../../report/html/renderer/performance-category-renderer.js');
 
     const document = jsdom.jsdom(TEMPLATE_FILE);
     const dom = new DOM(document);
@@ -169,6 +169,20 @@ describe('PerfCategoryRenderer', () => {
       };
       const wastedMs = renderer._getWastedMs(auditWithDebug);
       assert.ok(Number.isFinite(wastedMs), 'Finite number not returned by wastedMs');
+    });
+  });
+
+  describe('getLastScreenshot', () => {
+    it('gets a datauri as a string', () => {
+      const datauri = PerformanceCategoryRenderer.getLastScreenshot(category);
+      assert.equal(typeof datauri, 'string');
+      assert.ok(datauri.startsWith('data:image/jpeg;base64,'));
+    });
+
+    it('returns null if there are no screenshots', () => {
+      const fakeCategory = Object.assign({}, category, {auditRefs: []});
+      const datauri = PerformanceCategoryRenderer.getLastScreenshot(fakeCategory);
+      assert.equal(datauri, null);
     });
   });
 });


lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -192,6 +192,19 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
     return element;
   }
+
+  /**
+   * @param {LH.ReportResult.Category} category
+   * @return {null|string}
+   * @override
+   */
+  static getLastScreenshot(category) {
+    const auditRef = category.auditRefs.find(audit => audit.id === 'screenshot-thumbnails');
```
patrickhulce: should we give them a new proper full-res final screenshot audit?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -192,6 +192,19 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     element.appendChild(passedElem);
     return element;
   }
+
+  /**
+   * @param {LH.ReportResult.Category} category
+   * @return {null|string}
+   * @override
```
brendankenny: don't think this overrides anything?

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -26,15 +27,14 @@ const TEMPLATE_FILE = fs.readFileSync(__dirname +
...
   let renderer;
+  global.CategoryRenderer = CategoryRenderer;
+  const PerformanceCategoryRenderer =
+      require('../../../../report/html/renderer/performance-category-renderer.js');
 
   beforeAll(() => {
     global.URL = URL;
     global.Util = Util;
     global.CriticalRequestChainRenderer = CriticalRequestChainRenderer;
-    global.CategoryRenderer = CategoryRenderer;
```
brendankenny: why move?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -192,6 +192,19 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
+   */
+  static getLastScreenshot(category) {
```
brendankenny: maybe we don't want this kind of api on a category renderer? Could it live in utils.js?

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
paulirish: `PerformanceCategoryRenderer` has to be in a higher scope (for the new suite)
and `PerformanceCategoryRenderer` extends this one, so the require will fail if this isn't done beforehand

lighthouse-core/report/html/renderer/performance-category-renderer.js
paulirish: done.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -142,15 +142,23 @@ describe('CategoryRenderer', () => {
...
-    const pwaCategory = sampleResults.reportCategories.find(cat => cat.id === 'pwa');
-    const categoryDOM = renderer.render(pwaCategory, sampleResults.categoryGroups);
-    assert.ok(categoryDOM.querySelector('.lh-audit-group--manual .lh-audit-group__summary'));
-    assert.equal(categoryDOM.querySelectorAll('.lh-audit--manual').length, 3,
-        'score shows informative and dash icon');
-
-    const perfCategory = sampleResults.reportCategories.find(cat => cat.id === 'performance');
-    const categoryDOM2 = renderer.render(perfCategory, sampleResults.categoryGroups);
-    assert.ok(!categoryDOM2.querySelector('.lh-audit-group--manual'));
+    // Performance has its own category renderer
```
patrickhulce: I'm still a bit confused why these test changes had to be done for the final screenshot audit to happen :)

lighthouse-core/test/audits/final-screenshot-test.js
```diff
@@ -0,0 +1,34 @@
...
+  });
+
+  it('should extract a final screenshot from a trace', async () => {
+    const artifacts = Object.assign({
+      traces: {defaultPass: pwaTrace},
+    }, computedArtifacts);
+    const results = await FinalScreenshotAudit.audit(artifacts);
+
+    assert.ok(results.rawValue);
+    assert.equal(results.details.items[0].timing, 818);
```
patrickhulce: assert ok on the screenshot? since that's what we actually care about :)

lighthouse-core/audits/final-screenshot.js
```diff
@@ -0,0 +1,63 @@
...
+    };
+  }
+
+  /**
+   * @param {LH.Artifacts} artifacts
+   * @return {Promise<LH.Audit.Product>}
+   */
+  static async audit(artifacts) {
+    const trace = artifacts.traces[Audit.DEFAULT_PASS];
+    const traceOfTab = await artifacts.requestTraceOfTab(trace);
```
brendankenny: can also use the `screenshots` computed artifact and take the last one (it's already prefaced by `data:image/jpg;base64,`, but that doesn't seem too bad)

lighthouse-core/audits/final-screenshot.js
```diff
@@ -0,0 +1,63 @@
...
+    // The timing isn't too important, so don't fail the audit if there's no navigationStart found
+    // We'll just fall back to a less accurate timing
+    const timing = traceOfTab.navigationStartEvt
+      ? finalScreenshot.ts - traceOfTab.navigationStartEvt.ts
+      : finalScreenshot.ts - ssEvents[0].ts;
+
+    return {
+      rawValue: true,
+      details: {
+        type: 'screenshot',
```
brendankenny: it looks like this is type `filmstrip`. If we're going to make a new type that's screenshot (singular), probably no reason to use `items`, could flatten and do
```js
details: {
  type: 'screenshot',
  timestamp: number,
  timing: number,
  data: string
}
```

lighthouse-core/audits/final-screenshot.js
```diff
@@ -0,0 +1,63 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
```
brendankenny: 2018

lighthouse-core/test/audits/final-screenshot-test.js
```diff
@@ -0,0 +1,34 @@
...
+    assert.equal(results.details.items[0].timestamp, 225414990064);
+  }, 10000);
```
brendankenny: need the adjusted timeout?

lighthouse-core/report/html/renderer/performance-category-renderer.js
brendankenny: > it could live in Util, sure. But it seems more specific to perf.

it is specific to perf, but it's also a utility method and doesn't really match the rest of the API (which is entirely construction and `render()` right now). It feels weird to mix DOM generation and helpful utility methods. And since it's static and not called from within `performance-category-renderer.js`, you have to know to call it with the perf category anyways. We also have other single-use methods in `utils.js`, like `getEnvironmentDisplayValues` or `getEmulationDescriptions`.

Though now that this is an audit, maybe we shouldn't have a method at all and the PSI renderer can just reach into the audit itself?

lighthouse-core/audits/final-screenshot.js
paulirish: ho shit. that's much better. totally forgot. thank you

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -142,4 +146,23 @@ describe('util helpers', () => {
     Util.formatDisplayValue(displayValue);
     assert.deepStrictEqual(displayValue, cloned, 'displayValue was mutated');
   });
+
+  describe('getFinalScreenshot', () => {
+    sampleResults.reportCategories = Object.values(sampleResults.categories);
```
brendankenny: should stringify/parse this to not mutate the module

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -142,4 +146,23 @@ describe('util helpers', () => {
...
+      sampleResults.reportCategories);
+
+    it('gets a datauri as a string', () => {
+      const datauri = Util.getFinalScreenshot(category);
+      assert.equal(typeof datauri, 'string');
+      assert.ok(datauri.startsWith('data:image/jpeg;base64,'));
+    });
+
+    it('returns null if there are no screenshots', () => {
+      const fakeCategory = Object.assign({}, category, {auditRefs: []});
```
brendankenny: can you pass in the `pwa` category or cancel out just the final screenshot audit (or both?)? It's hard to tell which error case this is hitting since the category shape is so smooshed up by this point :)

lighthouse-core/lib/asset-saver.js
```diff
@@ -172,7 +172,6 @@ async function prepareAssets(artifacts, audits) {
     const Runner = require('../runner.js');
     const computedArtifacts = Runner.instantiateComputedArtifacts();
     /** @type {Array<Screenshot>} */
-    // @ts-ignore TODO(bckenny): need typed computed artifacts
```
brendankenny: 👍

lighthouse-core/test/audits/final-screenshot-test.js
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('should extract a final screenshot from a trace', async () => {
+    const artifacts = Object.assign({
+      traces: {defaultPass: pwaTrace},
+    }, computedArtifacts);
+    const results = await FinalScreenshotAudit.audit(artifacts);
+
+    assert.ok(results.rawValue);
+    assert.equal(results.details.timestamp, 225414990.064);
```
brendankenny: `assert.ok(results.details.data.startsWith('data:image/jpeg;base64,whateverthestartis'));`?

brendankenny:
`CategoryRenderer › renders manual audits if the category contains them` is definitely a weird test to be failing on, at least based on the description :)

Does it need to be set as notApplicable or something to not go through that loop?

PR #5674, opened 2018-07-17
misc(typings): convert hardcoded strings to networkRequests.TYPES

wardpeet:
**Summary**
Just fixing some changes we missed for resourceType

**Related Issues/PRs**

### DIFF
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -47,7 +47,7 @@ class BootupTime extends Audit {
     /** @type {Set<string>} */
     const urls = new Set();
     for (const record of records) {
-      if (record.resourceType && record.resourceType === NetworkRequest.TYPES.Script) {
+      if (record.resourceType === NetworkRequest.TYPES.Script) {
         urls.add(record.url);
       }
     }
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -7,6 +7,7 @@
 
 const Audit = require('./audit');
 const Util = require('../report/html/renderer/util');
+const NetworkRequest = require('../lib/network-request');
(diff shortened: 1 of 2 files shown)


wardpeet:
I stumbled upon some other hardcoded strings I'm not sure if they are intentional or not so just checking if I need to change them too.

[estimateTransferSize in the byte-efficiency-audits we call resourceType as a string](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/byte-efficiency/unminified-javascript.js#L59-L60) not sure if we should move to NetworkRecord.TYPES too.

lighthouse-core/audits/bootup-time.js
```diff
@@ -46,7 +47,7 @@ class BootupTime extends Audit {
     /** @type {Set<string>} */
     const urls = new Set();
     for (const record of records) {
-      if (record.resourceType && record.resourceType === 'Script') {
+      if (record.resourceType && record.resourceType === NetworkRequest.TYPES.Script) {
```
patrickhulce: we can simplify this more now too, just `record.resourceType === ` :)

lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -87,7 +87,7 @@ class OptimizedImages extends Gatherer {
 
       seenUrls.add(record.url);
       const isOptimizableImage = record.resourceType &&
-        record.resourceType === 'Image' &&
+        record.resourceType === NetworkRequest.TYPES.Image &&
```
patrickhulce: same here :)

wardpeet:
i'll just leave it be and just fix the ones I already have in this PR.

I thank the all seeing eye of @patrickhulce 😄

Issue #5676, opened 2018-07-17
Error: Could not load stylesheet (timeout)


wardpeet:
https://sentry.io/google-lighthouse/lighthouse/issues/596333623/

```
Error: Could not load stylesheet (timeout)
  File "/usr/lib/node_modules/lighthouse/lighthouse-core/gather/gatherers/fonts.js", line 218, in fontsAndErrors.filter.fontOrError
    const err = new Error(dataError.err.message);
  ?, in Array.filter
  File "/usr/lib/node_modules/lighthouse/lighthouse-core/gather/gatherers/fonts.js", line 213, in fontData.then
    const fontFaces = /** @type {Array<LH.Artifacts.Font>} */ (fontsAndErrors.filter(
  ?, in null.<anonymous>
  File "internal/process/next_tick.js", line 182, in process._tickCallback

Error: Could not load stylesheet (timeout)
```

wardpeet:
@paulirish @brendankenny how do you feel about these 2 options?

exterkamp:
Was this fixed as part of #6166?

PR #5677, opened 2018-07-17
misc(sentry): lower "could not load stylesheet" sampling to 0.01

wardpeet:
**Summary**
We've reached our quote of sentry this month because of this issue. Lowered the sample rate to 0.01

**Related Issues/PRs**
#5676


patrickhulce:
> would be easy to miss if only looking at relative count of errors (unless the sentry UI already accounts for this and scales appropriately)

I wish, this is just our own sampling impl so Sentry has no clue we're seeing these. Built-in sampling of particular errors would be awesome but it directly cuts into their pricing strategy so I don't see it happening 😆

PR #5686, opened 2018-07-18
misc(i18n): add assertion script

patrickhulce:
adds an assertion script to make sure we keep the strings up to date

### DIFF
--- a/lighthouse-core/scripts/i18n/assert-strings-collected.sh
+++ b/lighthouse-core/scripts/i18n/assert-strings-collected.sh
@@ -26,14 +26,14 @@ freshstringsPath="$lh_tmp_path/fresh_strings.json";
 function teardown { rm -f "$currentstringsPath" "$freshstringsPath"; }
 trap teardown EXIT
 
-cp $collectedstringsPath $currentstringsPath
+cp "$collectedstringsPath" "$currentstringsPath"
 
 colorText "Collecting strings..." "$purple"
 set -x
 node "$lhroot_path/lighthouse-core/scripts/i18n/collect-strings.js"
 set +x
 
-cp $collectedstringsPath $freshstringsPath
+cp "$collectedstringsPath" "$freshstringsPath"
 
 colorText "Diff'ing golden strings against the fresh strings" "$purple"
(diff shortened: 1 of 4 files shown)


lighthouse-core/scripts/i18n/assert-strings-collected.sh
```diff
@@ -0,0 +1,50 @@
...
+git --no-pager diff --color=always --no-index "$currentstringsPath" "$freshstringsPath"
+
+# Use the return value from last command
+retVal=$?
+
+if [ $retVal -eq 0 ]; then
+  colorText "✅  PASS. All strings have been collected." "$green"
+else
+  colorText "❌  FAIL. Strings have changed." "$red"
+  echo "Commit the changes to `lighthouse-core/lib/locales/` update the strings."
```
paulirish: in bash, backticks evaluate. ;)

so this actually triggers this error: `assert-strings-collected.sh: line 48: lighthouse-core/lib/locales/: Is a directory`

lighthouse-core/scripts/i18n/assert-strings-collected.sh
```diff
@@ -0,0 +1,50 @@
...
+
+collectedstringsPath="$lhroot_path/lighthouse-core/lib/locales/en-US.json";
+currentstringsPath="$lh_tmp_path/current_strings.json";
+freshstringsPath="$lh_tmp_path/fresh_strings.json";
+
+# always run this before exiting
+function teardown { rm -f "$currentstringsPath" "$freshstringsPath"; }
+trap teardown EXIT
+
+cp $collectedstringsPath $currentstringsPath
```
paulirish: actually wanna throw quotes around these paths? shellcheck prefers it like that in case there are spaces in my project path. (hint: there aren't)

patrickhulce: ya, done

lighthouse-core/scripts/i18n/assert-strings-collected.sh
patrickhulce: haha, whoops! that'll show me to make aesthetic changes after running :)

PR #5692, opened 2018-07-19
core(i18n): extract performance strings to UIStrings

patrickhulce:
this should be all of the audit messages necessary for performance
image

what's left after this
* renderer strings
* audit table cell values
* something else I'm forgetting?

**Related Issues/PRs**
blocked on  #5655

### DIFF
--- a/lighthouse-core/audits/dobetterweb/dom-size.js
+++ b/lighthouse-core/audits/dobetterweb/dom-size.js
@@ -14,10 +14,32 @@
 
 const Audit = require('../audit');
 const Util = require('../../report/html/renderer/util.js');
+const i18n = require('../../lib/i18n');
 
 const MAX_DOM_NODES = 1500;
 const MAX_DOM_TREE_WIDTH = 60;
 const MAX_DOM_TREE_DEPTH = 32;
+
+const UIStrings = {
+  title: 'Avoids an excessive DOM size',
+  failureTitle: 'Uses an excessive DOM size',
+  description: 'Browser engineers recommend pages contain fewer than ' +
+    `~${MAX_DOM_NODES} DOM nodes. The sweet spot is a tree ` +
+    `depth < ${MAX_DOM_TREE_DEPTH} elements and fewer than ${MAX_DOM_TREE_WIDTH} ` +
+    'children/parent element. A large DOM can increase memory usage, cause longer ' +
+    '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
+    'and produce costly [layout reflows](https://developers.google.com/speed/articles/reflow). [Learn more](https://developers.google.com/web/tools/lighthouse/audits/dom-size).',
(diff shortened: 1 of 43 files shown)


lighthouse-core/test/results/sample_v2.json
```diff
@@ -2118,27 +2068,24 @@
...
-        63
-      ],
+      "displayValue": "Potential savings of 63 KB",
       "details": {
         "type": "opportunity",
         "headings": [
           {
             "key": "url",
             "valueType": "url",
-            "label": "Uncompressed resource URL"
```
paulirish: I think it's better to have this full label
otherwise i could interpret this table to include already gzipped stuff

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -14,11 +14,33 @@
...
 
 const MAX_DOM_NODES = 1500;
 const MAX_DOM_TREE_WIDTH = 60;
 const MAX_DOM_TREE_DEPTH = 32;
 
+const UIStrings = {
+  title: 'Avoids an excessive DOM size',
+  failureTitle: 'Uses an excessive DOM size',
+  description: 'Browser engineers recommend pages contain fewer than ' +
+    `~${MAX_DOM_NODES} DOM nodes. The sweet spot is a tree ` +
```
paulirish: plural. 
also can we get thousands separation on this number?

patrickhulce: aw that means we have to make it a template and all 😢

lighthouse-core/test/results/sample_v2.json
patrickhulce: aw I was actually excited by the consistency now

It seems like the same argument could be made for all the other audits though. `Large Image URL`, `Uncached URL`, `Unused Image URL`, that all seems a bit much is there something unique about the compression audit?

paulirish: `Large Image URL`, `Uncached URL`, `Unused Image URL` sounds awesome. :)

wdyt?

lighthouse-core/audits/dobetterweb/dom-size.js
paulirish: how painful is that?

lighthouse-core/test/results/sample_v2.json
patrickhulce: 😭 that's not the direction I wanted this to go, lol

I think it's a bit redundant given the audit title and full description are immediately above the column names. And as @brendankenny as noted, i18n makes me want to consolidate the strings we use not aggressively expand them 😆

lighthouse-core/audits/dobetterweb/dom-size.js
patrickhulce: well it expands its representation at every stage in the process for what seems like zero added value, making it a plural implies that the value could change but the description is static and could be translated just once.

PR #5698, opened 2018-07-20
report: link to our own "unused css" reference doc

kaycebasques:
LMK if `report:` is the correct commit tag to use...


wardpeet:
Report is good 👍 , you can find our commit linting values here
https://github.com/GoogleChrome/lighthouse/blob/master/.cz-config.js

LGTM :)

PR #5702, opened 2018-07-21
core(i18n): keep Intl.* methods in extension build

paulirish:
repro: build the extension from master. the TTI audit will get an error saying Intl.NumberFormat is not a constructor.

it's set to undefined as the `require('intl');` doesnt throw but returns an empty object.


paulirish:
I'll close this in exchange for https://github.com/GoogleChrome/lighthouse/pull/5655/files/23e8efe009c0ab0f4b57c4a50349374bf8ee1fa2#r204202719

PR #5713, opened 2018-07-24
core(i18n): export rendererFormattedStrings

patrickhulce:
brings i18n strings to the renderer! this tackles most of what we need for performance category. strategy is what was outlined in #5655, strings are located at `lhr.i18n.rendererFormattedStrings` which the renderer uses to replace its own UIStrings in `util.js`

the `${n} audits` is definitely going to be a problem we're going to have to figure out, but for performance it only shows up for the "passed audits" section and none of the others which isn't the worst thing to live without at first if we need to

### DIFF
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -75,10 +75,10 @@ class CategoryRenderer {
     if (audit.result.scoreDisplayMode === 'error') {
       auditEl.classList.add(`lh-audit--error`);
       const textEl = this.dom.find('.lh-audit__display-text', auditEl);
-      textEl.textContent = 'Error!';
+      textEl.textContent = Util.UIStrings.errorLabel;
       textEl.classList.add('tooltip-boundary');
       const tooltip = this.dom.createChildOf(textEl, 'div', 'tooltip tooltip--error');
-      tooltip.textContent = audit.result.errorMessage || 'Report error: no audit information';
+      tooltip.textContent = audit.result.errorMessage || Util.UIStrings.errorMissingAuditInfo;
     } else if (audit.result.explanation) {
       const explEl = this.dom.createChildOf(titleEl, 'div', 'lh-audit-explanation');
       explEl.textContent = audit.result.explanation;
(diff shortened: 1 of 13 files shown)


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -89,9 +89,10 @@ class CategoryRenderer {
     // Add list of warnings or singular warning
     const warningsEl = this.dom.createChildOf(titleEl, 'div', 'lh-warnings');
     if (warnings.length === 1) {
-      warningsEl.textContent = `Warning: ${warnings.join('')}`;
+      // TODO(i18n): Maybe need to construct this differently
```
brendankenny: > TODO(i18n): Maybe need to construct this differently

if warnings are full sentences this should be fine

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -35,6 +35,11 @@ class ReportRenderer {
   renderReport(report, container) {
     // If any mutations happen to the report within the renderers, we want the original object untouched
     const clone = /** @type {LH.ReportResult} */ (JSON.parse(JSON.stringify(report)));
+    // Mutate the UIStrings if necessary (while saving originals)
```
brendankenny: should the technical debt comment below be moved up here? :)

It does kind of suck to rely on utils as a mutable singleton. We pass at least a `DOM` into all of our report rendering classes. Should we pass in strings? Or pass the lhr to everything, use `lhr.i18n.rendererFormattedStrings.string` everywhere, and here (since we're mutating the lhr anyways) could set `lhr.i18n.rendererFormattedStrings` to the default strings if not set.

Fine to punt to a future PR, but it would be nice to have a plan before we get too deep.

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -35,6 +35,11 @@ class ReportRenderer {
...
+    ReportRenderer.stashUIStrings();
```
brendankenny: if it's stashing just for this function, can it stored in a local variable instead of stashed in a static property?

lighthouse-core/report/html/renderer/category-renderer.js
patrickhulce: alright, removed

lighthouse-core/report/html/renderer/report-renderer.js
patrickhulce: done

lighthouse-core/report/html/renderer/report-renderer.js
patrickhulce: > Or pass the lhr to everything, use lhr.i18n.rendererFormattedStrings.string everywhere, and here (since we're mutating the lhr anyways) could set lhr.i18n.rendererFormattedStrings to the default strings if not set.

This sounds good to me, we have other LHR reaching needs like locale in config settings and such that would be nice. I accept your offer for a followup :)

brendankenny: add a TODO?

PR #5716, opened 2018-07-24
core(i18n): extract performance strings to UIStrings

patrickhulce:
this should be all of the audit messages necessary for performance
image

what's left after this
* renderer strings (#5713) 
* audit table cell values
* something else I'm forgetting?

**Related Issues/PRs**
~~blocked on  #5655~~
carried over from #5692


lighthouse-core/lib/locales/en-XA.json
```diff
@@ -1,16 +1,238 @@
 {
+  "lighthouse-core/audits/bootup-time.js | title": {
```
brendankenny: we'll probably want to stop generating this file when we get our first translation dump (this locale is automatically generated for us)

lighthouse-core/test/results/sample_v2.json
```diff
@@ -652,10 +630,7 @@
       "score": 0.96,
       "scoreDisplayMode": "numeric",
       "rawValue": 1548.5690000000002,
-      "displayValue": [
-        "%10d ms",
-        1548.5690000000002
-      ],
+      "displayValue": "NaN ms",
```
paulirish: 😮 

how about

```diff
diff --git a/lighthouse-core/lib/i18n.js b/lighthouse-core/lib/i18n.js
index 0c4d6d24..2776af87 100644
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -65,6 +65,14 @@ function _preprocessMessageValues(icuMessage, values) {
 
   const clonedValues = JSON.parse(JSON.stringify(values));
   const parsed = MessageParser.parse(icuMessage);
+
+  // Throw an error if a message's value isn't provided
+  parsed.elements.map(el => el.id).filter(Boolean).forEach(valueId => {
+    if ((valueId in values) === false) {
+      throw new Error('ICU Message contains a value reference that wasn\'t provided');
+    }
+  });
+
```

lighthouse-core/lib/i18n.js
```diff
@@ -58,6 +65,13 @@ function _preprocessMessageValues(icuMessage, values) {
 
   const clonedValues = JSON.parse(JSON.stringify(values));
   const parsed = MessageParser.parse(icuMessage);
+  // Throw an error if a message's value isn't provided
+  parsed.elements.map(el => el.id).filter(Boolean).forEach(valueId => {
```
brendankenny: the type definition seems to imply all the elements have `id`. Is that not true?

patrickhulce: not sure, @paulirish did you find some without ids? :)

lighthouse-core/test/audits/byte-efficiency/byte-efficiency-audit-test.js
```diff
@@ -176,7 +176,7 @@ describe('Byte efficiency base audit', () => {
       ],
     }, graph, simulator);
 
-    assert.ok(result.displayValue.includes(2), 'contains correct KB');
+    assert.ok(result.displayValue);
```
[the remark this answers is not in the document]
patrickhulce: the display values coming out of audits now are just temporary string IDs that reference something stored in i18n, so they don't have anything to do with the audit itself.

I'm somewhat against having so many layers asserting displayValue. It's been a large source of churn that hasn't really caught much. The end result in the LHR is really the only thing we care about which is OKishly captured by golden LHR.

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -12,6 +12,21 @@ const Audit = require('../audit');
...
+const i18n = require('../../lib/i18n');
+
+const UIStrings = {
+  title: 'Uses efficient cache policy on static assets',
+  failureTitle: 'Uses inefficient cache policy on static assets',
+  description:
+    'A long cache lifetime can speed up repeat visits to your page. ' +
+    '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/cache-policy).',
+  displayValue: `{itemCount, plural,
+    one {1 resource found}
```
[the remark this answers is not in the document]
patrickhulce: might as well do now, fixed 👍

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -94,10 +94,6 @@ module.exports = [
         },
       },
       'uses-responsive-images': {
-        displayValue: [
```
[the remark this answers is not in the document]
patrickhulce: ya, my comment below applies here too though for rationale :)

lighthouse-core/lib/i18n.js
patrickhulce: @brendankenny correct the types are wrong, all things with a `format` have `id` but the strings in between don't, we'll switch to filter on `type === 'ArgumentElement'`

lighthouse-core/test/audits/byte-efficiency/byte-efficiency-audit-test.js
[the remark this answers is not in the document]
brendankenny: > the display values coming out of audits now are just temporary string IDs that reference something stored in i18n, so they don't have anything to do with the audit itself

good point. We could do `i18n.replaceIcuMessageInstanceIds(result, 'en-US')`, then test `result.displayValue` here, I guess.

> I'm somewhat against having so many layers asserting displayValue. It's been a large source of churn that hasn't really caught much. The end result in the LHR is really the only thing we care about which is OKishly captured by golden LHR.

I mean, that's pretty much an argument against all unit tests. 90+% of the time none of them will fail, until that one time we drop a closing `}` to a pluralization argument and the display value just becomes a literal string (or whatever).

Generating a `displayValue` is pretty much all several of our audits are doing, the rest is in gatherers or computed artifacts. It doesn't seem unreasonable to assert their actual root source is generating them correctly.

(and `dbw_tester.html`, while a workhorse, can't cover all the cases we'd care about)

lighthouse-core/lib/i18n.js
paulirish: can you also update the dfn in `typings/intl-messageformat-parser/index.d.ts`

patrickhulce:
@brendankenny what did you think of #5720, would that work for you?

typings/intl-messageformat-parser/index.d.ts
```diff
@@ -1,7 +1,7 @@
 declare module 'intl-messageformat-parser' {
   interface Element {
     type: 'messageTextElement'|'argumentElement';
-    id: string
+    id?: string
     value?: string
     format?: null | {type: string; style?: string};
   }
```
brendankenny: if format and id are always defined if the other is, it may make sense to make a discriminated union out of this, but fine for now (won't help with the `.filter()`s anyway)

PR #5717, opened 2018-07-24
core: adjust audit titles for consistency

brendankenny:
from lightsights discussions. Let the bikeshedding commence!

### DIFF
--- a/lighthouse-core/audits/critical-request-chains.js
+++ b/lighthouse-core/audits/critical-request-chains.js
@@ -9,7 +9,7 @@ const Audit = require('./audit');
 const i18n = require('../lib/i18n');
 
 const UIStrings = {
-  title: 'Critical Request Chains',
+  title: 'Minimize Critical Request Chains',
   description: 'The Critical Request Chains below show you what resources are ' +
       'issued with a high priority. Consider reducing ' +
       'the length of chains, reducing the download size of resources, or ' +
--- a/lighthouse-core/lib/locales/en-US.json
+++ b/lighthouse-core/lib/locales/en-US.json
@@ -204,7 +204,7 @@
     "message": "Redirects introduce additional delays before the page can be loaded. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/redirects)."
   },
   "lighthouse-core/audits/time-to-first-byte.js | title": {
-    "message": "Server response times are short (TTFB)"
+    "message": "Server response times are low (TTFB)"
   },
   "lighthouse-core/audits/time-to-first-byte.js | failureTitle": {
     "message": "Reduce server response times (TTFB)"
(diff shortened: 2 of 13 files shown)


lighthouse-core/audits/critical-request-chains.js
```diff
@@ -15,7 +15,7 @@ class CriticalRequestChains extends Audit {
   static get meta() {
     return {
       id: 'critical-request-chains',
-      title: 'Critical Request Chains',
+      title: 'Minimize Critical Request Chains',
```
brendankenny: the audit is marked `notApplicable` if no requests are made, so everyone who gets the title could in theory minimize their chains. There might not be any benefit to doing that, but that's just a fundamental issue with the audit that needs to be addressed elsewhere :)

patrickhulce:
could this be done on top of #5716 :D (formerly #5692)

brendankenny:
> could this be done on top of #5716

that's fine. It should be easy

brendankenny:
updated to live on top of the newly extracted UIStrings

lighthouse-core/lib/locales/en-US.json
```diff
@@ -204,7 +204,10 @@
     "message": "Redirects introduce additional delays before the page can be loaded. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/redirects)."
   },
   "lighthouse-core/audits/time-to-first-byte.js | title": {
-    "message": "Keep server response times low (TTFB)"
+    "message": "Server response times are short (TTFB)"
```
patrickhulce: low

:)

lighthouse-core/scripts/assert-golden-lhr-unchanged.sh
```diff
@@ -46,6 +46,6 @@ if [ $retVal -eq 0 ]; then
   colorText "✅  PASS. No change in LHR." "$green"
 else
   colorText "❌  FAIL. LHR has changed." "$red"
-  echo "Run \`yarn update:sample-json\` to rebaseline the golden LHR."
+  echo "Run \`yarn i18n:collect-strings\` and \`yarn update:sample-json\` to rebaseline the golden LHR."
```
brendankenny: It's definitely confusing to update strings in audits but get no changes when you run `yarn update:sample-json `.  This still isn't the best place for this advice (though it is needed) as there's no change to `sample_v2.json` until you extract the strings, so this test was actually passing (though `yarn i18n:checks` would fail on travis).

Should we add `i18n:checks` to `diff:sample-json` so the check does both? Or add `i18n:collect-strings` to `update:sample-json` so it's automatically done at the same time?

patrickhulce: image

brendankenny: done

PR #5720, opened 2018-07-25
tests: better display value tests with i18n

patrickhulce:
@brendankenny we do this post #5716 ?

### DIFF
--- a/lighthouse-core/test/test-utils.js
+++ b/lighthouse-core/test/test-utils.js
@@ -1,0 +1,32 @@
+/**
+ * @license Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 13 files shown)


typings/externs.d.ts
```diff
@@ -184,3 +184,9 @@ declare global {
     }
   }
 }
+
+declare namespace jest {
```
brendankenny: looks unnecessary?

lighthouse-core/test/test-utils.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+'use strict';
+
+/* eslint-env jest */
+
+const i18n = require('../lib/i18n');
+
+expect.extend({
+  toHaveDisplayValue(received, expected) {
+    const obj = {displayValue: received.displayValue};
```
brendankenny: it looks like this and logging is the only `displayValue` specific stuff in here. Do we want a more generalized `toHaveLocalizedWhatever()`?

patrickhulce:
> This works for me for #5716, if the rest are copied over?

yep 👍

PR #5726, opened 2018-07-25
core(i18n): import psuedo-locale json from TC

paulirish:
_updated_

* ar-XB is psuedo right to left
* en-XA is accented


paulirish:
> also they seem to use _ while I was using - is there a best practice?

found another TC export script that has an option for hyphen rather than underscore. So this is a common thing. :)

Since our build script also prefers hyphen i guess we should go with that..

paulirish:
running 

```sh
lighthouse -GA --preset=perf http://example.com --locale=en-XA
``` 

I currently get this:
image


should I address it in this PR?

patrickhulce:
> I currently get this: 
should I address it in this PR?

:/ ya the extra locales will be useless if we can't actually generate reports with them

paulirish:
ready for review. 

```sh
lighthouse --preset=perf --locale=ar-XB http://example.com/
lighthouse --preset=perf --locale=en-XA http://example.com/
```

image

paulirish:
🏏 🏏

lighthouse-core/test/lib/i18n-test.js
```diff
@@ -58,7 +58,8 @@ describe('i18n', () => {
   describe('#getRendererFormattedStrings', () => {
     it('returns icu messages in the specified locale', () => {
       const strings = i18n.getRendererFormattedStrings('en-XA');
-      expect(strings.passedAuditsGroupTitle).toEqual('P̂áŝśêd́ âúd̂ít̂ś');
+      expect(strings.passedAuditsGroupTitle).toEqual('[Þåššéð åûðîţš one two]');
+      expect(strings.scorescaleLabel).toEqual('[Šçöŕé šçåļé: one two]');
```
brendankenny: we should probably convert this file back to `assert` at some point rather than having a different assertion library just for this file

PR #5727, opened 2018-07-25
core(i18n): always use english for logs

patrickhulce:
fixes what @paulirish found in #5726

### DIFF
--- a/lighthouse-core/lib/i18n.js
+++ b/lighthouse-core/lib/i18n.js
@@ -13,6 +13,7 @@ const MessageParser = require('intl-messageformat-parser');
 const LOCALES = require('./locales');
 
 const LH_ROOT = path.join(__dirname, '../../');
+const MESSAGE_INSTANCE_ID_REGEX = /(.* \| .*) # (\d+)$/;
 
 (() => {
   // Node usually doesn't come with the locales we want built-in, so load the polyfill if we can.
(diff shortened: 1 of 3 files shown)


lighthouse-core/runner.js
```diff
@@ -218,7 +218,7 @@ class Runner {
    */
   static async _runAudit(auditDefn, artifacts, settings, runWarnings) {
     const audit = auditDefn.implementation;
-    const status = `Evaluating: ${audit.meta.title}`;
+    const status = `Evaluating: ${i18n.getDisplayValue(audit.meta.title)}`;
```
paulirish: since `displayValue` is a thing i dont think we should reuse that name for this.

`getOriginalMessage` ?

lighthouse-core/lib/i18n.js
```diff
@@ -251,5 +274,7 @@ module.exports = {
   getDefaultLocale,
   getRendererFormattedStrings,
   createMessageInstanceIdFn,
+  getDisplayValue,
+  resolveIcuMessageInstanceId,
```
paulirish: does this need to be exported and public? could keep it _private to the file.

patrickhulce: yeah sure if you'd prefer :)

lighthouse-core/runner.js
patrickhulce: yeah I hesitated here, original message isn't quite right either. how bout 

`getFormatted`
`toFormatted`
`getFormattedValue`
`getFormattedString`
`toFormattedString`

lighthouse-core/lib/i18n.js
```diff
@@ -196,13 +197,41 @@ function createMessageInstanceIdFn(filename, fileStrings) {
   return getMessageInstanceIdFn;
 }
 
+/**
+ * @param {string} icuMessageIdOrRawString
+ * @param {LH.Locale} [locale]
+ * @return {string}
+ */
+function getDisplayValue(icuMessageIdOrRawString, locale) {
```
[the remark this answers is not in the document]
patrickhulce: yup, that was the goal :)

lighthouse-core/lib/i18n.js
```diff
@@ -196,13 +197,41 @@ function createMessageInstanceIdFn(filename, fileStrings) {
...
+  return MESSAGE_INSTANCE_ID_REGEX.test(icuMessageIdOrRawString) ?
```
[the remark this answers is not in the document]
patrickhulce: works for me, done

PR #5730, opened 2018-07-26
report(i18n): localize CRC renderer strings

patrickhulce:
A lot of the text is actually just the path and origin of each request so there wasn't actually that much here.


brendankenny:
Slight change to make translation easier.

before:
screen shot 2018-07-30 at 15 42 58

after:
screen shot 2018-07-30 at 15 46 15

Issue #5719, opened 2018-07-24
Internationalization of Lighthouse (Phase 1)


paulirish:
Through a series of PRs and CLs we're much closer to having Lighthouse i18n-ready than before. Some relevant PRs: #5644, #5655, #5713, #5716, #5718 

### Known remaining work:

- [x] Dump workflow established. xtb dump -> json -> LH core locale files @paulirish 
- [x] Add `[ICU Syntax]` to all descriptions of fields that use icu message syntax. it can start or end with this label string @paulirish #5736
- [x] Pass locale to toLocaleString calls in renderer @patrickhulce @brendankenny #5734 
- [x] Figure out locale fallback. (PSI user with locale pt-PT should get pt, not en-US). Where does this happen? @brendankenny  #5746
- [x] Add descriptions to ALL strings (enabled by #5718) @patrickhulce @brendankenny @paulirish
- [x] Refactor renderer use of Util.UIStrings in reportRenderer. Instead of setup/breakdown, use a str() method which chooses between provided renderer strings and fallbacks defined in Util.UIStrings.
- [x] (psi) Expose a Util.prepareLHR() method
- [x] fix icumessageinstanceid string in extension category setup
- [x] Roll TC locales back into gh repo #5780
- [x] Address remaining english strings in report:
   - [x] CRC template @patrickhulce @paulirish #5730
   - [x] UserTiming has 4 strings #5806
   - [x] ~runtime settings~ PUNTED
   - [x] ~export dropdown (low pri)~ PUNTED
- [x] Text-expansion (`de`) UI review
- [x] Use placeholders for our use of `origin`? See https://goto.google.com/gnjrf
  - [x] Confirm non-translated strings (eg. "origin", product names) are as expected. URL can be translated.
- [x] address TC locale fallback story 
- [ ] figure out Node's ICU support on Travis/Appveyor (`Number.toLocaleString` appears to always use `'en-US'` - [Travis failure](https://travis-ci.org/GoogleChrome/lighthouse/builds/410104347#L933-L986)) 
   - presumably the node install is [system-icu or small-icu](https://nodejs.org/api/intl.html#intl_options_for_building_node_js)
   - can we resolve with an NVM flag or travis config? 
- [x] temporarily exclude locale json files from ext build

### ship l10n'd lighthouse to extension/devtools
- [ ] localize error messages. #6441
- [ ] RTL layout @paulirish
- [ ] `lang` attribute #5805
- [ ] include locale json files in build

[...]

PR #5734, opened 2018-07-26
report(i18n): use LHR locale for toLocaleString

patrickhulce:
another piece of #5719, localize the numbers according to the locale specified in the LHR

note the 0,3 s

image

### DIFF
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -47,7 +47,12 @@ class ReportRenderer {
     // If any mutations happen to the report within the renderers, we want the original object untouched
     const clone = /** @type {LH.ReportResult} */ (JSON.parse(JSON.stringify(report)));
     // Mutate the UIStrings if necessary (while saving originals)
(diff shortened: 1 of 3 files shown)


paulirish:
Now that #5746 is in, we'll want a test here that confirms we use whatever locale `lookupLocale()` fell back to -- also here in .toLocaleString calls, rather then having diff locales for strings and numbers.

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -61,7 +64,7 @@ class ReportRenderer {
     container.appendChild(this._renderReport(clone));
 
     // put the UIStrings back into original state
-    ReportRenderer.updateAllUIStrings(clonedStrings);
+    ReportRenderer.updateAllUIStrings(originalUIStrings);
```
brendankenny: updated because it was a little confusing with `clone` being the cloned LHR, while `clonedStrings` was the clone of the original `Util.UIStrings`, *not* the translated strings found in `clone` :)

PR #5735, opened 2018-07-27
core(i18n): more default descriptions

patrickhulce:
before our description party thought it might make sense to auto-generate as much as possible, this should eliminate the need for title/failureTitle/displayValue


paulirish:
/me reads https://goto.google.com/uapbn first before reviewing..

patrickhulce:
> /me reads go/MessageDescriptions first before reviewing..

Yeah seems like position of the text is a key missing element from the current descriptions I've written, but when I started with them they all sounded dumb starting with the same phrase. Perhaps that's just a good problem though?

PR #5746, opened 2018-07-28
core(i18n): add locale fallbacks when language not supported

brendankenny:
cf. #5719

Sets up locale fallbacks, using the pretty common method of pruning on the last `-` until some prefix of the original locale is found, falling back to the default locale if no match is found.

Also adds a test to make sure that for any locale we set we also have a fallback for that language. So if e.g. we had `pt-BR` but not `pt-PT` for some reason, we'll be guaranteed to have `pt` to fall back to. Looking at our (eventual) list of supported languages, we shouldn't have to do any artificial mappings to support this like I'm doing in this PR.

Since locale is part of `config.settings`, it makes sense to initialize and check it in `Config`, so this moves it there. This allows removing some of the optional uses that had defaults coming from a few different places depending on the execution path taken.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -407,6 +407,9 @@ class Config {
    * @return {LH.Config.Settings}
    */
   static initSettings(settingsJson = {}, flags) {
+    // Use locale specified in flags or settings, allowing i18n system to select fallback if needed.
+    const locale = i18n.lookupLocale((flags && flags.locale) || settingsJson.locale);
+
     // Fill in missing settings with defaults
     const {defaultSettings} = constants;
     const settingWithDefaults = merge(deepClone(defaultSettings), settingsJson, true);
@@ -414,8 +417,8 @@ class Config {
     // Override any applicable settings with CLI flags
     const settingsWithFlags = merge(settingWithDefaults || {}, cleanFlagsForSettings(flags), true);
 
-    // Use locale specified in flags or settings, allowing i18n system to select fallback if needed.
-    settingsWithFlags.locale = i18n.lookupLocale((flags && flags.locale) || settingsJson.locale);
+    // Locale is special and comes only from flags/settings
+    settingsWithFlags.locale = locale;
 
     return settingsWithFlags;
   }
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -407,7 +407,9 @@ class Config {
    * @return {LH.Config.Settings}
    */
   static initSettings(settingsJson = {}, flags) {
-    // Use locale specified in flags or settings, allowing i18n system to select fallback if needed.
+    // If a locale is requested in flags or settings, use it. A typical CLI run will not have one,
+    // however `lookupLocale` will always determine which of our supported locales to use (falling
+    // back if necessary).
     const locale = i18n.lookupLocale((flags && flags.locale) || settingsJson.locale);
 
     // Fill in missing settings with defaults
@@ -417,7 +419,7 @@ class Config {
     // Override any applicable settings with CLI flags
     const settingsWithFlags = merge(settingWithDefaults || {}, cleanFlagsForSettings(flags), true);
 
-    // Locale is special and comes only from flags/settings
+    // Locale is special and comes only from flags/settings/lookupLocale.
     settingsWithFlags.locale = locale;
 
     return settingsWithFlags;
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -8,7 +8,8 @@
 const defaultConfigPath = './default-config.js';
 const defaultConfig = require('./default-config.js');
 const fullConfig = require('./full-config.js');
-const constants = require('./constants');
+const constants = require('./constants.js');
+const i18n = require('./../lib/i18n.js');
 
 const isDeepEqual = require('lodash.isequal');
 const log = require('lighthouse-logger');
@@ -401,17 +402,20 @@ class Config {
   }
 
   /**
-   * @param {LH.Config.SettingsJson=} settings
+   * @param {LH.Config.SettingsJson=} settingsJson
    * @param {LH.Flags=} flags
    * @return {LH.Config.Settings}
    */
-  static initSettings(settings = {}, flags) {
+  static initSettings(settingsJson = {}, flags) {
     // Fill in missing settings with defaults
     const {defaultSettings} = constants;
-    const settingWithDefaults = merge(deepClone(defaultSettings), settings, true);
+    const settingWithDefaults = merge(deepClone(defaultSettings), settingsJson, true);
 
     // Override any applicable settings with CLI flags
     const settingsWithFlags = merge(settingWithDefaults || {}, cleanFlagsForSettings(flags), true);
+
+    // Use locale specified in flags or settings, allowing i18n system to select fallback if needed.
+    settingsWithFlags.locale = i18n.lookupLocale((flags && flags.locale) || settingsJson.locale);
 
     return settingsWithFlags;
   }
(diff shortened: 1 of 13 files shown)


lighthouse-core/lib/i18n.js
```diff
@@ -66,6 +67,21 @@ const formats = {
...
 };
 
+/**
+ * Look up the best available locale for the requested language, falling back by
+ * looking at progressively shorter prefixes (`de-CH-1996` -> `de-CH` -> `de`),
+ * and finally returning the default locale ('en-US') if no match is found.
+ * @param {string=} locale
+ * @return {LH.Locale}
+ */
+function lookupLocale(locale = MessageFormat.defaultLocale) {
```
brendankenny: `MessageFormat.defaultLocale` is actually `en` when no locale info is found, not  `en-US`. Not sure if we have an issue with that since we assume it's `en-US` everywhere else. As a result, we'll only get the `en-US` default set at the end of this function if a specific locale we don't support is requested. Do we want to change the last line to return `en` instead? Or just live with this

lighthouse-core/lib/i18n.js
```diff
@@ -118,7 +134,7 @@ const _icuMessageInstanceMap = new Map();
  * @return {{formattedString: string, icuMessage: string}}
  */
 function _formatIcuMessage(locale, icuMessageId, icuMessage, values) {
-  const localeMessages = LOCALES[locale] || {};
+  const localeMessages = LOCALES[locale];
```
brendankenny: `locale` is always a `LH.Locale`, so standing behind our type checking here by having no fallback :)

lighthouse-core/lib/locales/index.js
```diff
@@ -3,11 +3,17 @@
...
 'use strict';
 
-module.exports = {
+/** @typedef {Record<string, {message: string}>} LocaleMessages */
+
+/** @type {Record<LH.Locale, LocaleMessages>} */
+const locales = {
+  'ar': require('./ar-XB.json'), // TODO: fallback not needed when ar translation available
+  'ar-XB': require('./ar-XB.json'),
+  'en': require('./en-US.json'), // en-* fallback
```
brendankenny: both `ar` and `en` will exist when we get our full languages back. If we want to do other mappings we can do them in here like this or make a simple map in `i18n.js` internal to `lookupLocale()`

lighthouse-core/config/config.js
```diff
@@ -401,18 +402,21 @@ class Config {
...
     // Fill in missing settings with defaults
     const {defaultSettings} = constants;
-    const settingWithDefaults = merge(deepClone(defaultSettings), settings, true);
+    const settingWithDefaults = merge(deepClone(defaultSettings), settingsJson, true);
 
     // Override any applicable settings with CLI flags
     const settingsWithFlags = merge(settingWithDefaults || {}, cleanFlagsForSettings(flags), true);
 
+    // Use locale specified in flags or settings, allowing i18n system to select fallback if needed.
+    settingsWithFlags.locale = i18n.lookupLocale((flags && flags.locale) || settingsJson.locale);
```
paulirish: can you extract `(flags && flags.locale) || settingsJson.locale` to a const that is defined above L415. between fallbacks and merges i want to be quite clear what is affecting locale and what isnt. (in this case the merge apparently doesnt)

lighthouse-core/config/constants.js
```diff
@@ -39,7 +39,7 @@ const defaultSettings = {
 
   // the following settings have no defaults but we still want ensure that `key in settings`
   // in config will work in a typechecked way
-  locale: null, // default determined by the intl library
+  locale: 'en-US', // actual default determined by Config using lib/i18n
```
paulirish: since `Config.initSettings` will never use this, what is it used for

lighthouse-core/lib/i18n.js
```diff
@@ -66,6 +67,21 @@ const formats = {
...
+/**
+ * Look up the best available locale for the requested language, falling back by
+ * looking at progressively shorter prefixes (`de-CH-1996` -> `de-CH` -> `de`),
+ * and finally returning the default locale ('en-US') if no match is found.
+ * @param {string=} locale
+ * @return {LH.Locale}
+ */
+function lookupLocale(locale = MessageFormat.defaultLocale) {
+  // TODO: could do more work to sniff out defaultLocale
+  const canonicalLocale = Intl.getCanonicalLocales(locale)[0];
```
paulirish: compat-wise, how do we look here across browsers and node. do we need this one conditionally polyfilled too?

lighthouse-core/lib/i18n.js
paulirish: these words wash over me like an ocean wave. 🌊 i don't know what hit me but it was powerful and disorienting.  

😕

lighthouse-core/runner.js
```diff
@@ -218,7 +217,7 @@ class Runner {
    */
   static async _runAudit(auditDefn, artifacts, settings, runWarnings) {
     const audit = auditDefn.implementation;
-    const status = `Evaluating: ${i18n.getFormatted(audit.meta.title)}`;
+    const status = `Evaluating: ${i18n.getFormatted(audit.meta.title, 'en-US')}`;
```
paulirish: nice. good call.

tsconfig.json
```diff
@@ -13,6 +13,7 @@
       "./typings"
     ],
 
+    "resolveJsonModule": true,
```
paulirish: this is for our messages files?

lighthouse-core/lib/i18n.js
brendankenny: > these words wash over me like an ocean wave

haha. Basically everywhere we pretend the default has been `en-US`, but the current `i18n.getDefaultLocale` has been trying to return `en` since that's what the default fallback of `MessageFormat.defaultLocale` is. It just then finds out there is no `en` strings file, so the function falls back to the hardcoded `en-US`. Once we have an `en` strings file,  that will change.

So,
- if you request whatever the default is for where LH is running, you'll get `en`
- if you request a locale that we don't have (and have no fallback for), you'll get `en-US`

I'm wondering if we should
- change default fallback everywhere to `en` to match `MessageFormat`
- change `MessageFormat.defaultLocale` so that if it returns `en`, to be `en-US` and not change everywhere else
- live with the inconsistency

lighthouse-core/config/constants.js
brendankenny: > since Config.initSettings will never use this, what is it used for

since we declare the constants in this file to be a valid `Config.Settings`, unfortunately we have to give *some* valid value here. Unfortunately I can't think of a good way to not do that while still retaining all the benefits we currently have of the type checked `Settings` and making sure every settings property that needs to be defined has a valid value.

lighthouse-core/lib/i18n.js
brendankenny: > compat-wise, how do we look here across browsers and node. do we need this one conditionally polyfilled too?

it comes with `Intl`, and it seems like we're assuming at least that exists? The function doesn't check valid tags or need a list of real locales or anything, it's just the capitalization and...maybe some other formatting stuff I can't remember.

Though from https://nodejs.org/api/intl.html it seems like maybe there's no `Intl` object at all if node was compiled with `--with-intl=none`?

tsconfig.json
brendankenny: > this is for our messages files?

yeah, no reason to keep tsc from not seeing them

lighthouse-core/lib/i18n.js
paulirish: the mdn compat table is out of date, but it at least indicates support was added later after the Intl basics:

image

unfortunately the only polyfill appears to be deep inside of Intl.js so hopefully we don't need it?

lighthouse-core/config/config.js
brendankenny: Hmm, I was putting it down here to *attempt* to make it clear that you can just ignore everything above for `locale`, this is the place that it's determined and set (and that `defaultSettings.locale` can't possibly influence the value).

Do you want me to set it on `settingWithDefaults` or `cleanFlagsForSettings(flags)` and let the merge merge it in?

PR #5761, opened 2018-08-01
core(i18n): always use formatted strings for extension popup

paulirish:
similar to #5727

fixes this:
image


lighthouse-extension/test/popup-test.js
```diff
@@ -92,4 +113,13 @@ describe('Lighthouse chrome popup', function() {
     assert.equal(titleText, 'Lighthouse');
     assert.equal(urlText, 'http://example.com');
   });
+
+
+  // Kinda lame as the mocked data is already good.
```
brendankenny: I guess a unit test (which would still be trivial but at least testing something) would be that the categories passed back are what are in the labels and values of the check boxes? So maybe keep the array passed back by the mocked `getDefaultCategories` in a top level variable and loop over it here to verify that all the members are there?

(rather than checking for the lack of ICU messages, which doesn't seem possible since the mock doesn't supply any of them)

PR #5765, opened 2018-08-01
 report: refactor UI string logic

paulirish:
cf. #5179 

* all uses of Util.UIString in the renderer pass through `Util.str()`
* this fn will use the supplied localized string, if that was provided. otherwise it'll fall back to the original string in the `util.js` source file. (which is en-us)
* no more mutate/restore dance on this data

### DIFF
--- a/lighthouse-core/report/html/renderer/util.js
+++ b/lighthouse-core/report/html/renderer/util.js
@@ -39,20 +39,6 @@ class Util {
   }
 
   /**
-   * @param {string} sourceString
-   * @return {string}
-   */
-  static str_(sourceString) {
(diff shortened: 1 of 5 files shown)


lighthouse-core/report/html/renderer/util.js
```diff
@@ -401,6 +401,21 @@ class Util {
     };
   }
 
+  /**
+   * @param {string} sourceString
+   * @return {string}
+   */
+  static str(sourceString) {
```
patrickhulce: does it have to be `str`? :)

I get that we don't want it to be too similar to i18n ones since it's quite different, but maybe a bit more descriptive since we're already forced to wrap :D

PR #5769, opened 2018-08-02
extension: refactor option/flag passing (breaking change)

paulirish:
disclaimer: This is a breaking change for our integrations with them so all the clients must also update for this to work.

So I'm okay with delaying this PR for a bit.

------

This simplifies how we were juggling "options" coming in from extension/devtools/LR. Basically we were passing around an `options` object in the 2 -background files and all we cared about was its `flags` property. And in the case of LR, 2 other props came in on `options` and eventually are delivered into `Runner.run`, despite our typing.

reviewers: 
runner.js only has a rename. boring.
the background files have all the good stuff.

### DIFF
--- a/lighthouse-extension/app/src/lighthouse-ext-background.js
+++ b/lighthouse-extension/app/src/lighthouse-ext-background.js
@@ -47,18 +47,18 @@ function updateBadgeUI(optUrl) {
 }
 
 /**
- * @param {{flags: LH.Flags}} options Lighthouse options.
+ * @param {LH.Flags} flags Lighthouse flags.
  * @param {Array<string>} categoryIDs Name values of categories to include.
  * @return {Promise<LH.RunnerResult|void>}
  */
-async function runLighthouseInExtension(options, categoryIDs) {
+async function runLighthouseInExtension(flags, categoryIDs) {
   // Default to 'info' logging level.
   log.setLevel('info');
   const connection = new ExtensionProtocol();
-  options.flags = Object.assign({}, options.flags, {output: 'html'});
+  flags.output = 'html';
 
   const url = await connection.getCurrentTabURL();
-  const runnerResult = await background.runLighthouseForConnection(connection, url, options,
+  const runnerResult = await background.runLighthouseForConnection(connection, url, flags,
     categoryIDs, updateBadgeUI);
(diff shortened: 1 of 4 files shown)


paulirish:
let's defer this a tad.

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -73,19 +73,17 @@ async function runLighthouseInExtension(options, categoryIDs) {
...
  * @return {Promise<string|Array<string>|void>}
  */
-async function runLighthouseAsInCLI(connection, url, options, categoryIDs) {
+async function runLighthouseAsInCLI(connection, url, flags, categoryIDs) {
   log.setLevel('info');
-  options.flags = Object.assign({}, options.flags, {output: options.outputFormat});
-
-  const results = await background.runLighthouseForConnection(connection, url, options,
-    categoryIDs);
+  // TODO(paulirish): update LR to provide flags and map outputFormat -> output
```
brendankenny: seems like this TODO shouldn't be in this file? (work here is already done)

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -73,19 +73,17 @@ async function runLighthouseInExtension(options, categoryIDs) {
  * Run lighthouse for connection and provide similar results as in CLI.
  * @param {Connection} connection
  * @param {string} url
- * @param {{flags: LH.Flags} & {outputFormat: string, logAssets: boolean}} options Lighthouse options.
+ * @param {LH.Flags & {logAssets: boolean}} flags Lighthouse flags plus logAssets
```
brendankenny: `logAssets`  could also be a fifth parameter to the function, or we could wait for some future refactor of `saveArtifacts`/`saveAssets`/`logAssets` (or this could `@return {Promise<LH.RunnerResult|void>}` and the caller could pull the `artifacts` off there)

brendankenny:
tsc also flags 

https://github.com/GoogleChrome/lighthouse/blob/06644520f862ce6cbdbf738b8242d8071e038d6f/lighthouse-extension/app/src/popup.js#L167

as needing to be updated to just using `flags`

PR #5770, opened 2018-08-02
core(i18n): expose locale on lhr.i18n

paulirish:
pretty straightforward


paulirish:
>  somewhat problematic to have in two places (here and in configSettings)

this is true. and they currently totally match.
 
in LR i want to return what locale LH used. (mostly so one can validate that a request for `de-CH` was serviced with `de`, etc). 

lhr.configSettings.locale could also work. I can use that in the meantime until we have a moment to bikeshed this.

PR #5772, opened 2018-08-02
tests(jest): disable coverage collection by default

paulirish:
since CI uses custom `*:ci` npm scripts that have `--coverage` included, i'm not sure why we defaulted to coverage collected in the config.


paulirish:
@patrickhulce 

you get outta here

😀

PR #5780, opened 2018-08-03
core(i18n): add localized messages for 52 locales

paulirish:
Well, hellloooo there.

[...]

PR #5781, opened 2018-08-04
core(i18n): incorporate 52 languages into messages pipeline

paulirish:
picking up on #5780...

#### Overview

1. All language messages files are added now. The fallback story is indicated in locales/index.js
   * I'm deleting the message files that are duplicated because fallback. Later, we can exclude them from the BUILD/export.
1. The default locale is changed to 'en'. 
   * AFAICT this is equivalent to 'en-US'. (This seems crazy to me, since all en-* (that's not en-US) do things Dates/Times differently.) But anyway. I like having the fallback be less specific.
1. various en-* locales are called out and mapped to `en-IE`, which we're treating like a `en-GB` (eg. "Minimise" instead of "Minimize"). This makes sure these locales get localized number formatting and not american style.
1. CLI won't validate your chosen locale anymore. We have a fallback story, so we won't block someone trying with 'es-MX'.
1. I don't know how important it is that the `LH.Locale` extern is updated, but.. it is.

---------

FWIW: ~Our bundle size for `lighthouse-background` was 1.5MB before we started i18n. Once we had en-us/en-XA/ar-XB, it was 1.6MB. With all these locales, it's now 2.5MB.~ There's a couple optimizations we could make to improve that.

Oops those were sizes before our 'minification' step. `dist/scripts/lighthouse-background.js` went from 349KB to 504KB.


lighthouse-core/lib/locales/index.js
```diff
@@ -9,11 +9,63 @@
...
+  'sv': require('./sv.json'),
+  'ta': require('./ta.json'),
+  'te': require('./te.json'),
+  'th': require('./th.json'),
+  'tr': require('./tr.json'),
+  'uk': require('./uk.json'),
+  'vi': require('./vi.json'),
+  'zh-CN': require('./zh-CN.json'),
+  'zh-HK': require('./zh-HK.json'),
+  'zh-TW': require('./zh-TW.json'),
```
brendankenny: on looking more, it seems like we should have a `'zh'` that uses `'zh-CN'`, but we can investigate more

lighthouse-cli/cli-flags.js
```diff
@@ -64,6 +64,7 @@ function getFlags(manualArgv) {
         ],
         'Configuration:')
       .describe({
+        // We don't allowlist specific locales. Why? So we can support the user who requests 'es-MX' (unsupported) and we'll fall back to 'es' (supported)
```
brendankenny: we might want a command for listing available locales, but something for another day :)

PR #5800, opened 2018-08-07
deps: depend on speedline-core, rather than the cli

paulirish:
removes a few deps from yarn.lock

`meow` is no longer in the deps tree, though it does still get pulled in for devDeps via `conventional-changelog-cli`.


brendankenny:
error is from also needing to update the speedline reference in `artifacts.d.ts` so the artifact is correctly be typed

PR #5802, opened 2018-08-07
core(i18n): roll latest TC messages

paulirish:
I fixed up the upstream BUILD file to match, so it doesn't output files that are duplicates.

* en-IE -> en-GB for the canonical non-american english
* added a few more aliases, based on the duplicates I got back. The actual file is whatever appears to be a more canonical locale name. https://www.unicode.org/cldr/charts/latest/supplemental/aliases.html
   * tagalog aliased to filipino
   * `iw` aliased to hebrew
   * `in` aliased to 'id' (indonesian)
   * norwegian Bokmål aliased to norwegian
   * moldovan aliased to romanian
   * bosnian aliased to croatian
* serbian w/ latin characters added. (`sr` is cyrillic)

(I apologize in advance if any of this offends someone. Please let me know if any of these mappings are incorrect.)


lighthouse-core/lib/locales/index.js
```diff
@@ -56,15 +60,17 @@ const locales = {
...
   'te': require('./te.json'),
   'th': require('./th.json'),
+  'tl': require('./fil.json'), // Alias of 'fil'
   'tr': require('./tr.json'),
   'uk': require('./uk.json'),
   'vi': require('./vi.json'),
-  'zh': require('./zh-CN.json'),
-  'zh-CN': require('./zh-CN.json'),
+  'zh': require('./zh.json'),
+  'zh-CN': require('./zh.json'), // Alias of 'zh'
```
brendankenny: do you want to include this since it'll fall back to `zh` anyways?

lighthouse-core/lib/locales/index.js
```diff
@@ -12,20 +12,19 @@ const locales = {
...
-  'en-AU': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-  'en-GB': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-  'en-IE': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-  'en-SG': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-  'en-ZA': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-  'en-IN': require('./en-IE.json'), // Don't fallback to en (which -> en-US)
-
-  'gsw': require('./de.json'), // swiss german. identical (for our purposes) to 'de'
+  // TODO: en-GB has just ~10 messages that are different from en-US. We should only ship those.
+  'en-AU': require('./en-GB.json'), // Don't fallback to en (which -> en-US)
```
brendankenny: just use `Alias of 'en-GB'` for all these as for entries below?

PR #5804, opened 2018-08-07
report: add PSI.prepareLabData()

### LINKED ISSUE #5803: Add Util.renderPerformanceCategoryForPSI()
Expose a single API for PSI to use.

something like:
```js
/** @typedef {import('./psi-renderer.js')} PSIReportRenderer */

/**
 * @param {string} LHresponseJsonString
 * @return {Element}
 */
function renderPerformanceCategoryForPSI(lighthouseResponse) {
  var lighthouseJson = JSON.parse(lighthouseResponse);
  const dom = new DOM(document);

  const reportLHR = JSON.parse(JSON.stringify(lighthouseJson));
  reportLHR.reportCategories = Object.values(reportLHR.categories);
  ReportRenderer.smooshAuditResultsIntoCategories(reportLHR.audits, reportLHR.reportCategories);
  const perfCategory = reportLHR.reportCategories.find(cat => cat.id === 'performance');
  const perfRenderer = new PerformanceCategoryRenderer(dom, new DetailsRenderer(dom));
  const perfEl = perfRenderer.render(perfCategory, reportLHR.categoryGroups);

  const scoreGaugeEl = dom.find('.lh-score__gauge', perfEl);
  const scoreGaugeWrapper = dom.find('.lh-gauge__wrapper', scoreGaugeEl);
  scoreGaugeWrapper.classList.add('lh-gauge__wrapper--huge');
  scoreGaugeWrapper.removeAttribute('href');
  dom.find('.lh-category-header', perfEl).remove();
  dom.find('.lh-permalink', perfEl).remove();

  const datauri = PerformanceCategoryRenderer.getLastScreenshot(perfCategory);

  return [scoreGaugeEl, perfEl, datauri];
};
```

paulirish:
and move `getFinalScreenshot` from `Util` to `PSI`

fixes #5803

### DIFF
--- a/lighthouse-core/report/html/renderer/performance-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/performance-category-renderer.js
@@ -112,13 +112,20 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
   /**
    * @param {LH.ReportResult.Category} category
    * @param {Object<string, LH.Result.ReportGroup>} groups
+   * @param {string=} environment
    * @return {Element}
    * @override
    */
-  render(category, groups) {
+  render(category, groups, environment) {
     const element = this.dom.createElement('div', 'lh-category');
-    this.createPermalinkSpan(element, category.id);
-    element.appendChild(this.renderCategoryHeader(category));
+    if (environment === 'PSI') {
+      const gaugeEl = this.dom.createElement('div', 'lh-score__gauge');
+      gaugeEl.appendChild(this.renderScoreGauge(category));
+      element.appendChild(gaugeEl);
+    } else {
+      this.createPermalinkSpan(element, category.id);
+      element.appendChild(this.renderCategoryHeader(category));
+    }
 
     // Metrics
     const metricAudits = category.auditRefs.filter(audit => audit.group === 'metrics');
--- a/lighthouse-core/report/html/renderer/performance-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/performance-category-renderer.js
@@ -112,7 +112,7 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
   /**
    * @param {LH.ReportResult.Category} category
    * @param {Object<string, LH.Result.ReportGroup>} groups
-   * @param {string=} environment
+   * @param {string=} environment 'PSI' and undefined are the only valid values
    * @return {Element}
    * @override
    */
--- a/lighthouse-core/report/html/renderer/psi.js
+++ b/lighthouse-core/report/html/renderer/psi.js
@@ -43,7 +43,7 @@ class PSI {
     if (!reportLHR.categoryGroups) throw new Error(`No category groups found.`);
 
     const perfRenderer = new PerformanceCategoryRenderer(dom, new DetailsRenderer(dom));
-    // PSI environment string will assure the categoryHeader and permalink elements are excluded
+    // PSI environment string will ensure the categoryHeader and permalink elements are excluded
     const perfCategoryEl = perfRenderer.render(perfCategory, reportLHR.categoryGroups, 'PSI');
 
     const scoreGaugeEl = dom.find('.lh-score__gauge', perfCategoryEl);
--- a/lighthouse-core/report/html/renderer/psi.js
+++ b/lighthouse-core/report/html/renderer/psi.js
@@ -54,8 +54,18 @@ class PSI {
     scoreGaugeWrapper.removeAttribute('href');
     dom.find('.lh-permalink', perfCategoryEl).remove();
 
-    const finalScreenshotDataUri = Util.getFinalScreenshot(perfCategory);
+    const finalScreenshotDataUri = PSI.getFinalScreenshot(perfCategory);
     return {scoreGaugeEl, perfCategoryEl, finalScreenshotDataUri};
+  }
+
+  /**
+   * @param {LH.ReportResult.Category} perfCategory
+   * @return {null|string}
+   */
+  static getFinalScreenshot(perfCategory) {
+    const auditRef = perfCategory.auditRefs.find(audit => audit.id === 'final-screenshot');
+    if (!auditRef || !auditRef.result || auditRef.result.scoreDisplayMode === 'error') return null;
+    return auditRef.result.details.data;
   }
 }
 
--- a/lighthouse-core/report/html/renderer/psi.js
+++ b/lighthouse-core/report/html/renderer/psi.js
@@ -21,20 +21,20 @@
 class PSI {
   /**
    * Returns all the elements that PSI needs to render the report
-   * We keep expose this helper method to minimize the 'public' API surface of the renderer
+   * We expose this helper method to minimize the 'public' API surface of the renderer
    * and allow us to refactor without two-sided patches.
    *
    *   const {scoreGaugeEl, perfCategoryEl, finalScreenshotDataUri} = PSI.prepareLabData(
-   *      LHresponseJsonString,
+   *      LHResultJsonString,
    *      document
    *   );
    *
-   * @param {string} LHresponseJsonString
-   * @param {Document} document
+   * @param {string} LHResultJsonString The stringified version of {LH.Result}
+   * @param {Document} document The host page's window.document
    * @return {{scoreGaugeEl: Element, perfCategoryEl: Element, finalScreenshotDataUri: string|null}}
    */
-  static prepareLabData(LHresponseJsonString, document) {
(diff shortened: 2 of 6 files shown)


lighthouse-core/report/html/renderer/psi.js
```diff
@@ -0,0 +1,76 @@
...
+ * limitations under the License.
+ */
+'use strict';
+
+/* globals self DOM PerformanceCategoryRenderer Util DetailsRenderer */
+
+class PSI {
+  /**
+   * Returns all the elements that PSI needs to render the report
+   * We keep expose this helper method to minimize the 'public' API surface of the renderer
```
brendankenny: remove "keep"?

lighthouse-core/report/html/renderer/psi.js
```diff
@@ -0,0 +1,76 @@
...
+   * and allow us to refactor without two-sided patches.
+   *
+   *   const {scoreGaugeEl, perfCategoryEl, finalScreenshotDataUri} = PSI.prepareLabData(
+   *      LHresponseJsonString,
+   *      document
+   *   );
+   *
+   * @param {string} LHresponseJsonString
```
brendankenny: - can we call it an `LHRJsonString` or an `LHresultJsonString`? Or something else. Or do you think this will be the most self explanatory name?
- can always add param descriptions to help with that
- the capitalization scheme here is totally crazy. Capital `LH`, all lowercase `response`, camelcase `Json` :)
- can it just be an `LH.Result` and not a string? (i.e. already parsed)

lighthouse-core/report/html/renderer/psi.js
```diff
@@ -0,0 +1,76 @@
...
+    const perfCategoryEl = perfRenderer.render(perfCategory, reportLHR.categoryGroups);
+
+    const scoreGaugeEl = dom.find('.lh-score__gauge', perfCategoryEl);
+    const scoreGaugeWrapper = dom.find('.lh-gauge__wrapper', scoreGaugeEl);
+    scoreGaugeWrapper.classList.add('lh-gauge__wrapper--huge');
+    // Remove Performance category title/description
+    dom.find('.lh-category-header', perfCategoryEl).remove();
+    // Remove navigation links
+    scoreGaugeWrapper.removeAttribute('href');
+    dom.find('.lh-permalink', perfCategoryEl).remove();
```
brendankenny: these `remove()`s are definitely the hackiest bits. Could we do with CSS or have a method for it on `CategoryRenderer` or something else?

lighthouse-core/test/report/html/renderer/psi-test.js
```diff
@@ -0,0 +1,125 @@
...
+  __dirname + '/../../../../report/html/templates.html',
+  'utf8'
+);
+
+/* eslint-env jest */
+
+describe('DOM', () => {
+  let document;
+  beforeAll(() => {
+    global.URL = URL; // do i need to do this?
```
brendankenny: > do i need to do this?

you tell me :P

lighthouse-core/test/report/html/renderer/psi-test.js
```diff
@@ -0,0 +1,125 @@
...
+        const lhrWithoutPerf = JSON.parse(sampleResultsStr);
+        delete lhrWithoutPerf.categories.performance;
+        const lhrWithoutPerfStr = JSON.stringify(lhrWithoutPerf);
+
+        assert.throws(() => {
+          PSI.prepareLabData(lhrWithoutPerfStr, document);
+        }, /no performance category/i);
+      });
+
+      it('throws if there is no perf category', () => {
```
brendankenny: same as above?

lighthouse-core/report/html/renderer/psi.js
paulirish: name updated to `LHResultJsonString`

> can it just be an LH.Result and not a string? (i.e. already parsed)

I considered that. The fact that it's a string is due to LR. (And me, obv). And at some point we'll fix LR and this won't be a string anymore and will be received on the PSI side as a proper `LH.Result` (or `LHLite.Result`) 

Since that seems like an implementation detail on the LH/LR side, it seemed reasonable that we eat the maintenance cost. Both LR and third_party/lh-renderer will have to roll in coordination, but that's about it.

lighthouse-core/report/html/renderer/psi.js
paulirish: impl-wise it really comes down to [avoiding these two lines](https://github.com/GoogleChrome/lighthouse/blob/e9d9fbf8b304473148462e787f77c9118b9acb91/lighthouse-core/report/html/renderer/performance-category-renderer.js#L120-L121), so yah we can easily avoid these .remove's.

1. option: add extra arg to PerfCatRen.render() that avoids making the permamlink & catHeader elements in the first place. 1 more arg to .render() than the base class, but shrug.
1. option: just like `this._dom.isDevTools()` we add a `this._dom.isPSI()` and drop it into PCR.render().
1. the CSS solution (`display:none`) is possible and technically safe for screenreaders, but it feels a tad messy.
1. what kinda method on CatRen are you thinking of?

lighthouse-core/report/html/renderer/psi.js
brendankenny: > Since that seems like an implementation detail on the LH/LR side, it seemed reasonable that we eat the maintenance cost. Both LR and third_party/lh-renderer will have to roll in coordination, but that's about it.

SG, just wondering

lighthouse-core/report/html/renderer/psi.js
brendankenny: > option: just like `this._dom.isDevTools()` we add a `this._dom.isPSI()` and drop it into `PCR.render()`.

this has its downsides, but since we have precedent with `isDevTools()` it seems (to me) like the best option out of these (besides a `.isPSI` CSS class, but seems like you really don't like that :)

> what kinda method on CatRen are you thinking of?

I was hoping if I suggested it vaguely enough you'd assume I meant something in particular and bring that back as a possible solution :)

lighthouse-core/test/report/html/renderer/psi-test.js
brendankenny: > do i need to do this?

```
  ● DOM › psi prepareLabData helpers › prepareLabData › reports expected data
    ReferenceError: URL is not defined
      127 |         a.target = '_blank';
      128 |         a.textContent = linkText;
    > 129 |         a.href = (new URL(linkHref)).href;
          |                       ^
      130 |         element.appendChild(a);
      131 |       }
      132 |     }
      at DOM.convertMarkdownLinkSnippets (lighthouse-core/report/html/renderer/dom.js:129:23)
      at PerformanceCategoryRenderer._renderMetric (lighthouse-core/report/html/renderer/performance-category-renderer.js:44:40)
      at keyMetrics.forEach.item (lighthouse-core/report/html/renderer/performance-category-renderer.js:135:41)
          at Array.forEach (<anonymous>)
      at PerformanceCategoryRenderer.render (lighthouse-core/report/html/renderer/performance-category-renderer.js:134:16)
      at Function.prepareLabData (lighthouse-core/report/html/renderer/psi.js:46:41)
      at Object.it (lighthouse-core/test/report/html/renderer/psi-test.js:59:28)
PASS lighthouse-core/test/audits/mainthread-work-breakdown-test.js
```

apparently yes :)

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -112,13 +112,20 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
   /**
    * @param {LH.ReportResult.Category} category
    * @param {Object<string, LH.Result.ReportGroup>} groups
+   * @param {string=} environment
```
brendankenny: maybe be specific about what's accepted...I assume `'PSI'=` works but that looks weird :)

lighthouse-core/report/html/renderer/psi.js
```diff
@@ -0,0 +1,74 @@
...
+    const lhResult = /** @type {LH.Result} */ JSON.parse(LHResultJsonString);
+    const dom = new DOM(document);
+
+    const reportLHR = Util.prepareReportResult(lhResult);
+    const perfCategory = reportLHR.reportCategories.find(cat => cat.id === 'performance');
+    if (!perfCategory) throw new Error(`No performance category. Can't make lab data section`);
+    if (!reportLHR.categoryGroups) throw new Error(`No category groups found.`);
+
+    const perfRenderer = new PerformanceCategoryRenderer(dom, new DetailsRenderer(dom));
+    // PSI environment string will assure the categoryHeader and permalink elements are excluded
```
brendankenny: ensure?

PR #5805, opened 2018-08-07
report(i18n): set `lang` attribute according to locale

paulirish:


lighthouse-core/report/report-generator.js
```diff
@@ -40,6 +40,7 @@ class ReportGenerator {
     const sanitizedJavascript = htmlReportAssets.REPORT_JAVASCRIPT.replace(/<\//g, '\\u003c/');
 
     return ReportGenerator.replaceStrings(htmlReportAssets.REPORT_TEMPLATE, [
+      {search: '%%LIGHTHOUSE_LOCALE%%', replacement: lhr.configSettings.locale},
```
brendankenny: should have a fallback for older lhrs without a `locale`?

PR #5831, opened 2018-08-14
report: make red/orange/green score color thresholds tougher

brendankenny:
For harmony with PSI thresholds.

note that `Util.PASS_THRESHOLD` also affects when [`title` vs `failureTitle` is selected for an audit](https://github.com/GoogleChrome/lighthouse/blob/049af195ab33d40718e6fe39a278de1c2ec07ec5/lighthouse-core/audits/audit.js#L192-L196)

@paulirish

### DIFF
--- a/lighthouse-core/report/html/renderer/util.js
+++ b/lighthouse-core/report/html/renderer/util.js
@@ -20,11 +20,11 @@
 
 const ELLIPSIS = '\u2026';
 const NBSP = '\xa0';
-const PASS_THRESHOLD = 0.75;
+const PASS_THRESHOLD = 0.9;
 
 const RATINGS = {
   PASS: {label: 'pass', minScore: PASS_THRESHOLD},
-  AVERAGE: {label: 'average', minScore: 0.45},
+  AVERAGE: {label: 'average', minScore: 0.5},
   FAIL: {label: 'fail'},
   ERROR: {label: 'error'},
 };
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -83,10 +83,10 @@ describe('util helpers', () => {
   it('calculates a score ratings', () => {
     assert.equal(Util.calculateRating(0.0), 'fail');
     assert.equal(Util.calculateRating(0.10), 'fail');
-    assert.equal(Util.calculateRating(0.45), 'average');
+    assert.equal(Util.calculateRating(0.45), 'fail');
     assert.equal(Util.calculateRating(0.55), 'average');
```
paulirish: can you add/tweak the test cases so we have coverage for 0.50 and .90

lighthouse-core/report/html/renderer/util.js
```diff
@@ -20,11 +20,11 @@
 
 const ELLIPSIS = '\u2026';
 const NBSP = '\xa0';
-const PASS_THRESHOLD = 0.75;
+const PASS_THRESHOLD = 0.9;
```
patrickhulce: 🎉

zeman:
@paulirish no problem. We're already running 3.1 on SpeedCurve and noticed scores weren't aligned.

Issue #1091, opened 2016-12-01
 ☔️ Protocol error (Tracing.start): Tracing is already started


maulanaahmadarif:
**NOTICE FOR DUPLICATE BUG FILERS:** The only known fix for this is usually completely restarting the browser, and sometimes with a completely fresh profile.

**Lighthouse Version**: 1.2.2
**Chrome Version**: 57.0.2937.0
**Error Message**: Protocol error (Tracing.start): Tracing is already started
**Stack Trace**:
 ```Error: Protocol error (Tracing.start): Tracing is already started
    at ExtensionConnection.handleRawError (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:7377:11)
    at Object.chrome.debugger.sendCommand.result [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-background.js:7526:33)
    at safeCallbackApply (extensions::sendRequest:21:15)
    at handleResponse (extensions::sendRequest:74:7)```

ebidel:
Can you post the URL you're trying? Is this 100℅ reproducible in stable
chrome or just canary?

maulanaahmadarif:
I'm using canary right now
https://lab.ydigital.asia/precision/

paulirish:
This one is hard to solve. It happens on occasion and the best solution is restarting chrome right now.

If this continues we can look into ways of forcing Chrome to stop tracing even after we've disconnected.

brendankenny:
for errors that are causing this, we could add an attempt to stop tracing in the [gather-runner cleanup](https://github.com/GoogleChrome/lighthouse/blob/59f9f4808eb84467d861baf10f3463c0f161f33c/lighthouse-core/gather/gather-runner.js#L272-L276). If the error caused us to lose connection to devtools it won't work, but it may help in some cases

paulirish:
@caseq tells me that the tracing backend should stop the trace is the protocol client has disconnected.  So everything is lined up to make this work, though there may be a race condition.

I'm having trouble reproducing this bug (though I have hit it before).

If anyone can find a reliable repro for this, please comment with it.

gauntface:
I have a reproducible test case. Will try and reduce so you can either use it for debugging OR add it as part of the integration tests.

ebidel:
[...]

PR #5836, opened 2018-08-15
core(tsc): update debugger protocol type checking

brendankenny:
Blocked on ~https://github.com/ChromeDevTools/devtools-protocol/pull/113~

The new tuple types in TypeScript 3.0 allow us to better represent protocol payloads whether an object, an optional object, or empty. See https://github.com/ChromeDevTools/devtools-protocol/pull/113 for details. This allows us to get rid of a bunch of fiddly bits and simplify our `StrictEventEmitter` type to just a few lines.

We can also now
- defer to `devtools-protocol` for types instead of generating them ourselves, deleting the generator and the type definition file
- define methods like `Driver.on`, `Connection.sendCommand`, etc in the class body again instead of manually on the prototype at the end of the file, finally deleting some of the awkwardness debt left by #4886 and #4914 due to TS2.8 stuff.

One possibly controversial thing: I basically reverted the `silent` option from #2279 as it makes the `sendCommand` interface difficult to define. If we really want to, I think we can make it work (in a super ugly way), but based on the on-going #1091 it didn't seem to help the `Tracing.start` issue. Open to thoughts there.


brendankenny:
✂️ ✂️ ✂️ ✂️ ✂️ 
screen shot 2018-08-14 at 18 12 40

lighthouse-core/lib/network-request.js
```diff
@@ -43,6 +43,9 @@ const RESOURCE_TYPES = {
   WebSocket: 'WebSocket',
   Other: 'Other',
   Manifest: 'Manifest',
+  SignedExchange: 'SignedExchange',
+  Ping: 'Ping',
+  CSPViolationReport: 'CSPViolationReport',
```
paulirish: nice!

while i do see these here https://chromedevtools.github.io/devtools-protocol/tot/Page#type-ResourceType

i don't see them here: https://github.com/ChromeDevTools/devtools-frontend/blob/master/front_end/common/ResourceType.js

but i guess that's because we're now ahead of the frontend. :)

patrickhulce: > but i guess that's because we're now ahead of the frontend. :)

🎉 💣 💥 💯 💃 🕺 🥇 

music to my ears!!!

denar90:
Can we expect to have this merged before next release?

brendankenny:
Yes, when https://github.com/ChromeDevTools/devtools-protocol/pull/113 ships to npm I just need to update and we should be able to land it

denar90:
Thanks for the info :)

brendankenny:
no longer a WIP :)

lighthouse-core/gather/driver.js
```diff
@@ -74,6 +73,10 @@ class Driver {
       if (this._networkStatusMonitor) {
         this._networkStatusMonitor.dispatch(event);
       }
+
+      // @ts-ignore TODO(bckenny): tsc can't type event.params correctly yet,
+      // typing as property of union instead of narrowing from union of
+      // properties. See https://github.com/Microsoft/TypeScript/pull/22348.
```
brendankenny: this is unfortunate to ignore, but will have to wait on ts team for the fix.

Basically `_eventEmitter.emit()` correctly checks that for every protocol event `method` name (e.g. `'Profiler.consoleProfileStarted'`), it will have a matching event `params` (e.g. `LH.Crdp.Profiler.ConsoleProfileStartedEvent`).

The `event` object passed into this function is a giant union of possible `method`/`pair` objects like:
```js
{method: 'Profiler.consoleProfileStarted', params: LH.Crdp.Profiler.ConsoleProfileStartedEvent} |
{method: 'Runtime.bindingCalled', params: LH.Crdp.Runtime.BindingCalledEvent} |
{method: 'CSS.fontsUpdated', params: LH.Crdp.CSS.FontsUpdatedEvent} |
...
```

It's a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions), so in theory the compiler could check each union member separately, which would pass the `_eventEmitter.emit()` check.

Currently, however, the compiler doesn't narrow the type based on `method`, so instead of taking each union member separately, it takes the union of all possible property values. The second argument passed to `_eventEmitter.emit()` ends up being `LH.Crdp.Profiler.ConsoleProfileStartedEvent | LH.Crdp.Runtime.BindingCalledEvent | LH.Crdp.CSS.FontsUpdatedEvent | ...`.

When `_eventEmitter.emit()` starts checking that for every `method` (`'Profiler.consoleProfileStarted'`) there is the correct `params` (`LH.Crdp.Profiler.ConsoleProfileStartedEvent`), it sees that full union instead and so rightly shows an error since only one member of the union would have been the correct type.

As noted, this will hopefully be fixed by https://github.com/Microsoft/TypeScript/pull/22348, currently aimed at the typescript 3.2ish timeframe.

brendankenny: Also, it's slightly confusing, but the similar checks in the surrounding methods (e.g. other uses of `eventEmitter`) *are* correctly type checked. The key is that `event` is an object with `method` and `params` as properties. If `method` and `params` had come in as separate function parameters (as they do elsewhere) or as a tuple, the compiler would be able to correctly type them.

paulirish:
LETS DO THIS

denar90:
thanks 💃

denar90:
@brendankenny @paulirish one more thing
do you have any info about next (maybe small/minor/teeny-tiny) release? :)

denar90:
Not relevant question, thanks it's already released)

PR #5842, opened 2018-08-16
new_audit: add Max Potential FID to JSON

patrickhulce:
#### Summary
Putting this up for discussion. Over email we settled on "Maximum Potential FID (mFID)" as a good lab FID proxy that's easy to understand. In our case, it is the longest task post-FCP and this is its PR... :)

#### Open questions to answer along with my thoughts

1. **Should LH surface this metric?**
Yes, mFID is easily explainable and has a clear tie to our RUM push with FID.
2. **What should we call this metric?**
Max Potential FID is what was somewhat settled on in the thread, but Estimated Input Latency is a lot prettier and legit sounding of a name to compete with, so I'm all ears to suggestions. I'd be so bold as to suggest the title is just "Expected First Input Delay" with the explanation of "maximum potential blah blah" in the description tooltip, but I'm certain others will disagree :)
3. **What do we do with Estimated Input Latency?**
~This PR moves it down to LHR only but we can do other things. Keeping it in right now makes the metrics section a bit weird with 3-4. EIL's implementation with thresholds and boundaries has always been somewhat tenuous. Until we have a clear motivating example for devs, I'd prefer a FID stand-in.~ We're just going to land this in the JSON and see how it compares in HTTPArchive for now.

#### Key Change Introduced By This PR
~This PR also adjusts the EIL estimate multipliers to fully represent the Lantern prediction. Previously the raw EIL estimate was reduced by 20% before being surfaced in the UI.~ This will be split out into a separate PR.

cc @tdresser @philipwalton

-----------

screen shot 2018-08-15 at 5 10 34 pm


tdresser:
@stubbornella 

I think this is reasonable. It does make the story around page load interactivity a bit more confusing though.

Do we have any way of evaluating the success of a change like this?

@addyosmani - should we have some documentation on how mFID / FID / TTI all relate before shipping this in lighthouse?

philipwalton:
I'm OK with calling it "Estimated First Input Delay" (note, I wouldn't use the acronym since we don't use it for other metrics).

I'm also OK with removing estimated input latency entirely since it's not super clear how it's calculated (and thus not super clear how to fix it if you're getting a value higher than you expect).

Overall, I'm happy to see this as part of Lighthouse. If nothing else, it'll help make developers aware of long tasks in their script initialization that they should break up.

lighthouse-core/gather/computed/metrics/max-potential-fid.js
```diff
@@ -0,0 +1,38 @@
...
+      throw new LHError(LHError.errors.NO_FCP);
+    }
+
+    const events = TracingProcessor.getMainThreadTopLevelEvents(
+      data.traceOfTab,
+      firstContentfulPaint
+    ).filter(evt => evt.duration >= 1);
+
+    return Promise.resolve({
+      timing: Math.max(...events.map(evt => evt.duration), 16),
```
paulirish: timing and not timeInMs?

lighthouse-core/audits/metrics/max-potential-fid.js
```diff
@@ -0,0 +1,70 @@
...
+  /** The name of the metric "Maximum Potential First Input Delay" that marks the maximum estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. Shown to users as the label for the numeric metric value. Ideally fits within a ~40 character limit. */
+  title: 'Max Potential FID',
+  /** Description of the Estimated Input Latency metric that estimates the amount of time, in milliseconds, that the app takes to respond to user input. This description is displayed within a tooltip when the user hovers on the metric name to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
+  description: 'The potential First Input Delay that your users could experience is the ' +
+      'duration, in milliseconds, of the longest task.'
+};
+
+const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
+
+class MaxPotentialFID extends Audit {
```
paulirish: can we get some overview comment to explain the basics of how this metric is calculated?

lighthouse-core/gather/computed/metrics/lantern-max-potential-fid.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * @return {LH.Gatherer.Simulation.MetricCoefficients}
+   */
+  get COEFFICIENTS() {
+    return {
+      intercept: 0,
+      optimistic: 0.4,
+      pessimistic: 0.4,
```
paulirish: whats this mean? the two graphs are the same. Both are adjusted?

similarly, lantern-speed-index has two identical graphics but then has different coefficients.

paulirish:
> Should LH surface this metric?

yup

> What should we call this metric?

Def think it should tie into "First Input Delay" closely. 
Right now I'm most fond of _First Input Delay Impact_

> What do we do with Estimated Input Latency?

Comment it out until its glorious return as a rolling EQT diagnostic.

lighthouse-core/audits/metrics/max-potential-fid.js
patrickhulce: done

lighthouse-core/gather/computed/metrics/lantern-max-potential-fid.js
patrickhulce: The graphs aren't the differentiating part of the optimistic/pessimistic equation here, it's actually whether we use optimistic/pessimistic FCP as the start point :)

The 0.4 was to mirror EIL, but honestly the justification for it (that we'll overestimate EIL a tad) hasn't proved to be true in the data, so I'll stick to 0.5 (simple average of optimistic + pessimistic estimates)


For speed index case the "optimistic"/"pessimistic" labels are a real stretch. The "optimistic" one is just normal speed index on the fast trace, the graph is essentially ignored completely here. The "pessimistic" one is our lantern psuedo-speed index based on the TTI graph.

addyosmani:
> should we have some documentation on how mFID / FID / TTI all relate before shipping this in lighthouse?

Our interactivity narrative over the last quarter has focused on TTI in the lab and FID in the field; I'd love to make sure `First Input Delay Impact` or `Estimated First Input Delay` (whichever name we decide on) is well covered in our metrics guidance so it's clear:

1) TTI still matters (eFID compliments TTI without diminishing it)
1) How to think about our two lab metrics for interactivity (eFID is a clear lab corollary for FID, how it differs).

@philipwalton Will you have bandwidth to update our metrics docs to reflect the above?
@kaycebasques would be great to similarly see if we can get a Lighthouse doc up about eFID.

patrickhulce:
The new plan here is to land this without exposing for now, so we can take a look at the HTTPArchive results and how it compares to EIL/TTI/etc

patrickhulce:
bump @brendankenny @paulirish :)

lighthouse-core/computed/metrics/lantern-max-potential-fid.js
```diff
@@ -0,0 +1,98 @@
...
+   * @param {number} fcpTimeInMs
+   */
+  static getEventsAfterFCP(nodeTimings, fcpTimeInMs) {
+    /** @type {Array<{start: number, end: number, duration: number}>} */
+    const events = [];
+    for (const [node, timing] of nodeTimings.entries()) {
+      if (node.type !== BaseNode.TYPES.CPU) continue;
+      if (timing.endTime < fcpTimeInMs) continue;
+
+      events.push({
```
paulirish: you could just filter nodeTimings instead of constructing these new objects...

patrickhulce: good point! done

PR #5846, opened 2018-08-16
new_audit(tap-targets): check that tap targets are big enough and don't overlap

### LINKED ISSUE #4358: [SEO Audits] Tap targets are sized appropriately
**Audit group:** [Mobile friendly](https://github.com/GoogleChrome/lighthouse/issues/3116)
**Description:** Tap targets are sized appropriately
**Failure description:** Tap targets are not sized appropriately
**Help text:** Interactive elements like buttons and links should be large enough, and have enough space around them, to be easy enough to tap without overlapping onto other elements. Learn more.

**Success conditions:**
- all interactive elements are at least 48px tall and wide (content + padding)
- all interactive elements are at least 32px apart (margin)

If possible, the weighting for this audit should be proportional to the number of interactive element failures.

Interactive elements:
- button
- a
- input
- textarea
- select
- option
- [role=<[widget](https://www.w3.org/WAI/PF/aria/roles#widget_roles)>]

For reference, see: https://developers.google.com/web/fundamentals/accessibility/accessible-styles#multi-device_responsive_design

kdzwinel:
**Summary**
This audit checks if all potential tap targets (see 1) are sized appropriately (see 2) and are far enough from each other to be unambiguous (see 3).

**Limitations**
1) we take potential tap targets into account (`a, button, input, …`) - this list may not include some tap targets that are made "tappable" via JS (e.g. some div used as a button) and may include some tap targets that don't cause any action (e.g. dummy button that does nothing)
2) We define finger size to be 48x48px
3) Implemented algorithm is simpler from what Chrome does (e.g. Chrome is often able to figure out which link was clicked even if tap area covers two links). The algorithm boils down to checking if there is some other tap target located < 32px from the center of the currently audited one.

**Demo** (note that visual output is *not* part of this audit):

**TODO**
- [ ] descriptions
- [x] linting
- [ ] tests

**Related Issues/PRs**
Closes #4358


lighthouse-core/gather/gatherers/seo/tap-targets-test.js
```diff
@@ -0,0 +1,152 @@
...
+          height: 10,
+          top: 10,
+          bottom: 20,
+        },
+      ]);
+    });
+
+
+    it('Merges two vertically adjacent client rects', () => {
+      const res = simplifyBCRs([
```
patrickhulce: before you write too many more tests, I might make a helper function that just lets you pass in `{x, y, width, height}` to make these a little easier to understand.

I dunno about you but I've always found width, height much easier to quickly parse than top, left, right, bottom :)

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,419 @@
...
+function allBCRsEmpty(bcrs) {
+  return (
+    bcrs.length === 0 || bcrs.every(bcr => bcr.width === 0 && bcr.height === 0)
+  );
+}
+
+/**
+ * @param {Element} node
+ */
+function isVisible({node, bcrs = getBCRs(node, false)}) {
```
patrickhulce: this is a nice little generic helper! might be nice in `page-functions.js`

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,419 @@
...
+    })()`;
+
+    require('fs').writeFileSync('tt-gatherer.js', expression);
+
+    return passContext.driver.evaluateAsync(expression);
+  }
+}
+
+module.exports = TapTargets;
+// todo: find proper way to do export here
```
patrickhulce: this should be fine :)

lighthouse-core/lib/page-functions.js
```diff
@@ -118,7 +118,7 @@ function getOuterHTMLSnippet(element, ignoreAttrs=[]) {
     clone.removeAttribute(attribute);
   });
 
-  const reOpeningTag = /^.*?>/;
+  const reOpeningTag = /^[\s\S]*?>/;
```
patrickhulce: nice catch!

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: we've also been trying to keep all our in-page functions there, but we're going to have so many for this function it's probably worth creating a dedicated file for them. 

do you think it's easiest for it to be basically one giant in-page evaluation?

I ask because it's generally hard to debug and figure out what's going on in the in-page functions than the node side, but it looks like you've worked out a decent method of development here with the `fs` call. There might also be some opportunity to share some BCR logic between audit and gatherer?

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,240 @@
...
+      };
+    }
+
+    // console.time('tooCloseTargets');
+    const tooClose = tooCloseTargets(artifacts.TapTargets);
+
+
+    // console.timeEnd('tooCloseTargets');
+    /** @type {Array<{node: LH.Audit.DetailsRendererNodeDetailsJSON, issue: string}>} */
+    const failures = [];
```
patrickhulce: so these are each tap target pairs where `targetA` is unique across all `failures` and `targetB` is the target with the most overlap with that `targetA`, correct?

I wonder if it might be a little too confusing to the user given the framing focused on tap targets being too small. WDYT about listing the nodes that failed with a filter on node total area? We could then list the selectors of the elements that are too close to them.

mattzeunert: > so these are each tap target pairs where targetA is unique across all failures and targetB is the target with the most overlap with that targetA, correct?

We report multiple failures if there are multiple overlapping targets, rather than just reporting the biggest overlap. (The logic we have for getting the largest overlap is for different client rects.)

> I wonder if it might be a little too confusing to the user given the framing focused on tap targets being too small. WDYT about listing the nodes that failed with a filter on node total area? We could then list the selectors of the elements that are too close to them.

Good point about it being confusing, I'll talk to Rick about the framing and how we show the results.

Are you suggesting we should surface the node area in the results? Or just filter out failures that have a large tap target?

If a tap target is bigger than 36x36px then it won't fail. (We suggest 48x48px but allow some overlap.)

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: We've mostly been using a [manually synced gist](https://gist.github.com/mattzeunert/67dcea9c97a882248508a81e11adaf4a) for development. It also gives a visual output, which we're hoping to add to the DevTools Audit panel one day.

Will move the BCR logic to a shared file.

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,419 @@
...
+
+  let bcrs;
+
+  // if (display === 'inline' && overflowX === 'visible' && overflowY === 'visible') {
+  //   const range = document.createRange();
+  //   range.selectNodeContents(node);
+
+  //   bcrs = range.getClientRects();
+  // } else {
+  bcrs = node.getClientRects();
```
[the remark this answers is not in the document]
mattzeunert: Yeah we could use getBCR – having a tap target split over multiple lines should be super rare. I don't think that would simplify the logic much though.

There are two bits of information we need for the overlap check:
1. where to put the finger on a tap area
2. where tap areas are

For (2) we should use client rects because it prevents false failures in cases like this:

screen shot 2018-11-08 at 08 10 16

We can't just check the tap target element itself but also need to check its children, because they can be outside the parent BCR. For example, images inside of an `<a>` tag:

screen shot 2018-11-08 at 08 44 25

So even with getBCR we end up with multiple client rects, but we don't want to put a finger on each one. So we merge them together. (In the case above simplifyBCRs merges `<a>` and `` BCRs together.)

Another example that gets merged:

screen shot 2018-11-08 at 10 11 52

We also can't just get the bounding rect of all client rects, because some of the tap target children might, for example, be `position: absolute`.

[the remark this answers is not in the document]
patrickhulce: Thanks for the details and example this helps a lot! Few more questions :)

I'm not sure I totally understood what the `(2)` case was with the "Bizarre penalty decision" example. Are those the green boxes if we use client rects, and gray box is getBCR? Is the issue that the european football finger would bleed into the getBCR box but not getClientRects one?

> but we don't want to put a finger on each one

Just thinking aloud here, what if we convert this constraint a little bit. Right now our logic is basically "if there is a place to tap the target unambiguously, then it passes" (i.e. if you can put your finger in the middle). Given this policy, we should be treating every child outside the BCR of the parent as a new potential tap target. It also looks like the current situation with  extending the BCR into non-rectangular shapes could also lead to putting our finger tap completely outside the bounds of the merged shape. 

What happens if we treated them separately and de-duped them at the end? I'm trying to see if there's a consistent way to avoid the complicated BCR merging. I think it makes the audit hard to reason about and explaining/understanding failures will be difficult to users as well. I think you already know very well--and as I'm starting to suspect--it's impossible to have this not be complicated and confusing, but I at least want to give it my best shot :)

The biggest sticking point to me seems to be knowing that tap targets will lead to the same result. We are using shared parent elements as a proxy for this, but we don't really know for sure. Is that a fair assessment?

lighthouse-core/audits/seo/tap-targets.js
patrickhulce: > If a tap target is bigger than 36x36px then it won't fail. (We suggest 48x48px but allow some overlap.)

Right, but what I'm trying to get at is we're listing pairs, not individual elements. Maybe I misunderstood but it seems like it's possible for either targetA or targetB in the list to be large and the other one to be the target that's too small. I'm trying to see if we can reframe the results a bit so that each row is a tap target that is too small, and then one of the columns can list the elements that are overlapping it.

mattzeunert: `targetA` is always too small. If you try tapping on `targetA` you might accidentally tap `targetB`.

screen shot 2018-11-08 at 15 37 48

(A tap target overlapping another target's finger is highlighted by the orange outline.)

> then one of the columns can list all the elements that are overlapping it.

Yeah that's possible, but might be harder to format with the "extra space needed" info. 95% of the time there's only one overlapping tap target.

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
mattzeunert: > I'm trying to see if there's a consistent way to avoid the complicated BCR merging. I think it makes the audit hard to reason about and explaining/understanding failures will be difficult to users as well.

Agreed. Would be great to find an alternative solution or some trade-off we can make 🙂.

> Are those the green boxes if we use client rects, and gray box is getBCR? Is the issue that the european football finger would bleed into the getBCR box but not getClientRects one?

Yes to both questions.

> The biggest sticking point to me seems to be knowing that tap targets will lead to the same result. We are using shared parent elements as a proxy for this, but we don't really know for sure (Example "Senator reacts" one, do both of those go to the same place? Does the play button take you somewhere different? Can we always tell?). Is that a fair assessment?

No, the issue here is different. There's only one `<a>` tag, which contains some text and the play button. We know they have the same action, the actual issue is with children outside the parent BCR.

We don't report a failure if one tap target is completely inside another one, because that's usually an intentional design decision.

> Given this policy, we should be treating every child outside the BCR of the parent as a new potential tap target.

We used to do that, but changed it because it was failing cases that shouldn't fail:

screen shot 2018-11-08 at 16 15 50

It might be a good idea to pass every element that has at least one unambiguous tap target finger area. That would fail rects that are tall enough collectively but not on their own, though.

mattzeunert:
@patrickhulce For reference, here's a drawing that Rick made for how we detect a failure-causing overlap between two tap targets. The approach is taken from the Mobile-Friendly Test.

screen shot 2018-11-08 at 16 20 08

> The blue dotted box is the finger tap area centered around the top anchor. The bottom anchor overlaps the tap area and that shaded area is more than half the shaded area of the top anchor. So this is a violation.

Also, I think you noticed already, but the code is still very much work-in-progress, so don't spend too much time on the details 😂.

lighthouse-core/audits/seo/tap-targets.js
patrickhulce: Ah gotcha, OK well I'll defer to @rviscomi then just wanted to bring it up to make sure we don't get too impl focused compared to what we told users we'd be showing them :)

mattzeunert: Update after talking to @rviscomi:

- We'll tell people tap targets should be 48x48px
- All tap targets that are too small* are shown in the table
- Score is only impacted if there is an overlapping element nearby
- We don't show "extra space needed"

So columns are: Too Small Element | Size | Overlapping Element

* probably based on the of the largest merged area if we have multiple BCRs

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
mattzeunert: Update after talking with @rviscomi:

We're concerned about not reporting what should be failures when using the "at least one non-overlapping tap finger" approach. For example, if there's another target right next to the "senators react" link but it doesn't reach all the way to the play button then the tap target should fail. We don't want to allow the small extra rect/finger to count as a fix to the problem.

We can fix the "finger tap completely outside the bounds of the merged shape" problem by checking that the center of the merged area is within one of the input rects, and if not cancel the merge.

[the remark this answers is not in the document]
patrickhulce: Alright, gotcha that makes sense! But then I think we should reconsider the center of rect approach then if we're wanting to flag anything that's too close.

WDYT about...

A tap target is "too small" if the non-ambiguous tappable area is <X% of the total area.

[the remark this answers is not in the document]
mattzeunert: I like that idea! Is this what you mean?

screen shot 2018-11-09 at 10 34 16

(Dividing by own area might be better, and would also strongly penalize small tap targets.)

So this would be the steps:

1. For each client rect add 24px finger area on each side
2. Get total area of the merged polygons
3. Get total area of the merged polygons overlapping other targets

I'll implement it in the gist and see how well it works in practice.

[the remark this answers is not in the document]
mattzeunert: Two thoughts on the approach that's not finger-based:

- would have a high ambiguity score for a 47x47px target (it currently doesn't take into account that people will try to tap in the middle of a tap target)
- the `ambiguousAreaRects` or `clientRects` could overlap with each other... so ideally we'd merge them each into one polygon, get a union polygon where the two overlap, and then get the area of that polygon. Feels a bit tricky.
- I think it only avoids the multi-rect issue by considering more are for the score? Multiple small target rects would still cause a large finger area.

Fundamentally it's a good idea to consider the entire tappable area (rather than just the center), but in practice I don't think it really makes much difference. The user will try to tap in the center of a button (because otherwise they might accidentally tap on another button). And usually the surrounding tap targets at the center of a tap target are representative of the whole target.

Here's a screenshot from playing around with it, the score is how much of the target is overlapped by other target's extended area):

screenshot 2018-11-13 at 16 23 10

--- 

Been trying to figure out how all this maps to what we want to achieve.

Problem: Users try to tap on a small tap target but accidentally tap on another target. (@rviscomi I hope this sounds about right 🙂)

A target passes if:
- it's at least 48x48px *or*
- it's too small but missing it doesn't accidentally trigger an unintended action

Complications:
- shape of tappable area isn't a single rectangle (though usually it almost is)
- deciding likelihood of accidentally triggering another action (we're ok with some finger area overlap)

Using the finger approach, these are our options for handling multiple rects:
- merge (and check all after merge)
- check largest rect only
- pass if any passes

Given that we'll never report a tap target that has a rect > 48x48px, and that multiple rects are uncommon, I was quite hopeful that we could avoid the merging. 

But actually there is a meaningful score impact. On the left is with the merge, on the right we pass if any client rect passes.

screenshot 2018-11-13 at 17 12 46

"check largest rect" also would cause problems, for example with multiple lines of text (like the "Winds whip" headline above) where the longest also happens to be closest to other rects.

If we use "any rect passes" alongside merging we could use some slightly simpler merging logic (because we don't need to handle the `a` tag with image inside case). Haven't find any real world score differences between normal merging and simple merging+any rect passes.

lighthouse-core/gather/gatherers/seo/tap-targets-test.js
mattzeunert: Oh, good idea!

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
patrickhulce: Thanks for investigating the alternatives so well! The audit wasn't my idea, so I'll definitely defer to the pre-established approach you think is best :)

I'm not 100% sure in the merging vs. non-merging example what new finger pads are causing the "Brexit: " lines to pass now. Naively they all seem to overlap with roughly the same amount of external client rect area as the original finger pads did.

Assuming I figure that one out though...

-------

Just to recap I think we've decided here the following:

1. Finger-based approach on the center of tap targets is what we'd like to go with.
2. We can't really avoid the merging of client rects.
3. Matt's done a ton more investigation than any developer should need to for a single PR and it's time for me to just review some code? @rviscomi are you happy with these decisions too :)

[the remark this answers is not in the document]
mattzeunert: > Assuming I figure that one out though...

That's just a bug in my code. I thought the reason it didn't fail is because of the extra empty space on the left of the tap target, but that doesn't apply because we detect the quarter finger with the largest overlap.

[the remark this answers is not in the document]
mattzeunert: @patrickhulce Should be good to take another look at the code now.

lighthouse-viewer/test/viewer-test-pptr.js
```diff
@@ -134,7 +134,8 @@ describe('Lighthouse Viewer', function() {
     const auditErrors = await viewerPage.$$eval(errorSelectors, getErrors, selectors);
     const errors = auditErrors.filter(item => item.explanation.includes('Audit error:'));
     const unexpectedErrrors = errors.filter(item => {
-      return !item.explanation.includes('Required RobotsTxt gatherer did not run');
+      return !item.explanation.includes('Required RobotsTxt gatherer did not run') &&
```
mattzeunert: I don't understand what the proper fix for this would be, but from the discussion in the channel I understood it might not be easy and we're fine just ignoring the problem.

patrickhulce: Yup, ignoring is what we wanted here for now :)

rviscomi:
@patrickhulce yes the current iteration of audit logic LGTM based on @mattzeunert's demos. Thank you for your thorough code review!

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,399 @@
...
+
+    /** @type {Array<LH.Audit.TooSmallTapTargetItem>} */
+    const tableItems = [];
+
+    const scorePerElement = new Map();
+    artifacts.TapTargets.forEach(target => {
+      scorePerElement.set(target, 1);
+    });
+
+    tooSmallTargets.forEach(target => {
```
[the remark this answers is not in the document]
mattzeunert: Great idea!

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,399 @@
...
+    });
+
+    tableItems.sort((a, b) => {
+      /**
+       * @param {LH.Audit.TooSmallTapTargetItem} failure
+       */
+      function getFailureSeriousness(failure) {
+        let magnitude = failure.width * failure.height;
+        if (failure.extraDistanceNeeded) {
+          magnitude -= failure.extraDistanceNeeded * 10000;
```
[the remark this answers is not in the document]
mattzeunert: We want to show tap targets with the greatest amount of overlap (i.e. extraDistanceNeeded) at the top. Will simplify, this code used to handle non-failing targets that too small.

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,331 @@
...
+ * @returns {LH.Artifacts.ClientRect[]}
+ */
+function getClientRects(node, includeChildren = true) {
+  const rawClientRects = /** @type {LH.Artifacts.ClientRect[]} */ Array.from(
+    node.getClientRects()
+  );
+
+  let clientRects = Array.from(rawClientRects).map(clientRect => {
+    const {width, height, left, top, right, bottom} = clientRect;
+    return {width, height, left, top, right, bottom};
```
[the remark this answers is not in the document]
mattzeunert: Yeah, will add a comment. I don't really understand why it's needed though, since calling JSON.stringify on a DOMRect is fine. But when we use `Runtime.evaluate` the clientRects come back empty (`{}`).

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,331 @@
...
+}
+
+/**
+ * @returns {LH.Artifacts.TapTarget[]}
+ */
+function gatherTapTargets() {
+  const selector = TARGET_SELECTORS.join(',');
+
+  /** @type Array<LH.Artifacts.TapTargetWithNode> */
+  let targets = Array.from(document.querySelectorAll(selector)).map(node => ({
```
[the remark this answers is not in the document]
mattzeunert: No reason, just didn't think of using it.

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,399 @@
...
+    right: clientRect.right - clientRect.width / 2 + FINGER_SIZE_PX / 2,
+    bottom: clientRect.bottom - clientRect.height / 2 + FINGER_SIZE_PX / 2,
+  });
+}
+
+/**
+ * @param {LH.Artifacts.ClientRect} rect1
+ * @param {LH.Artifacts.ClientRect} rect2
+ */
+function getRectXOverlap(rect1, rect2) {
```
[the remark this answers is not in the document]
mattzeunert: Moved them over now 👍

mattzeunert:
I'm thinking of changing the logic to exclude tap targets that are `position: absolute` or `position: fixed`. This will miss some tap targets though, e.g. see Wikipedia. We had some logic before treating elements with `visibility: hidden` as invisible, but I don't think that's the right behavior.

Any objections to this?

screenshot 2018-12-07 at 15 43 04

patrickhulce:
> I'm thinking of changing the logic to exclude tap targets that are position: absolute or position: fixed.

I can recall having absolutely positioned icons quite a few times (thinking the close X at the top right of modals, etc), so it'd be a shame to miss out on those. I'm not totally sure though how it relates to the `visibility: hidden` and `elementFromPoint` stuff though. Is the problem that you can't tell which rects are on top of other rects and so there's false overlap from rects underneath?

mattzeunert:
> Is the problem that you can't tell which rects are on top of other rects and so there's false overlap from rects underneath?

Yeah.

The `visibility: hidden` logic was accidentally solving the problem on some sites, e.g. on Wikipedia the menu is both `z-index: 0` below the main content and `visibility: hidden`.

patrickhulce:
I know this sucks to hear, but I agree `elementFromPoint` is ideal here and we should probably just scroll down if we need to as we're evaluating elements. Do you think there's a safe way to ignore these cases for now and we can circle back around to improve in a follow up? I'm thinking if the rects *actually* overlap then we exclude them from consideration, and we just focus on rects that are "too close". Would that start excluding way too much to be useful?

mattzeunert:
> I know this sucks to hear, but I agree `elementFromPoint` is ideal here and we should probably just scroll down if we need to as we're evaluating elements. Do you think there's a safe way to ignore these cases for now and we can circle back around to improve in a follow up? I'm thinking if the rects _actually_ overlap then we exclude them from consideration, and we just focus on rects that are "too close". Would that start excluding way too much to be useful?

No I don't think it would exclude many failures. It would introduce new false failures though, where there's a position absolute container that overlaps the visible button but the container has tap targets that are adjacent to the visible tap target.

Could we exclude absolutely positioned elements for now and then switch to `elementFromPoint` in the followup?

patrickhulce:
> It would introduce new false failures though, where there's a position absolute container that overlaps the visible button but the container has tap targets that are adjacent to the visible tap target.

I'm not sure I follow this one. I was proposing that if there are tap targets that have overlapping rects in a way that we can't tell if they are on top of each other or not, we just exclude the tap target entirely. The change I was suggesting would purely eliminate tap targets from consideration so there shouldn't be a way in which that creates *more* failures.

> Could we exclude absolutely positioned elements for now and then switch to elementFromPoint in the followup?
Actually, do you reckon we should just switch to elementFromPoint now? Or will it delay the PR too much/make the review harder?

I'll ultimately defer to your judgement on this one, but my sense is that this PR is already very complicated and adding this layer is going to be tricky/worth reviewing on its own. I'd personally do it in a followup.

mattzeunert:
> I'm not sure I follow this one. I was proposing that if there are tap targets that have overlapping rects in a way that we can't tell if they are on top of each other or not, we just exclude the tap target entirely. The change I was suggesting would purely eliminate tap targets from consideration so there shouldn't be a way in which that creates more failures.

I meant introduce more failures compared to ignoring position absolute elements.

In the audit we just have a list of tap targets, we don't know what their z position is on the page. So to us it would look like this:

```
AAAA
AAAA
BB
```

There's no overlap here – B might be in a container that's behind the main content, but we don't have any info about that container. So we'll treat it as a failure.

patrickhulce:
Oh oh oh I gotcha now, yeah that makes sense. I'm somewhat surprised that an absolute check would fix these cases. Couldn't any random other element we're not checking be absolutely positioned on top of our tap target and we wouldn't know it?

mattzeunert:
Yeah, if a tap target is absolutely positioned over our element we'll get a false pass.

Basically I'm suggesting that we're being overly broad and treat anything that's absolutely positioned as invisible.

patrickhulce:
> we'll get a false pass.

OK I think I finally understand :) We're going to look at all client rects of all elements, not just those of tap targets, and if we see any client rect that is absolutely positioned, we will exclude all targets within that client rect. Is that right?

mattzeunert:
Basically. We can do the same thing we already do for `display: none`, check the tap target and then traverse through the ancestors and check them.

patrickhulce:
> check the tap target and then traverse through the ancestors and check them.

Oh hang on then I think what I'm talking about was slightly different then. I'm talking about other rects that have no tap targets in them, but are still on top of/obscuring real target rects. That would still be a problem after your proposal, right?

mattzeunert:
Yeah, it wouldn't catch that. I suppose we have 4 cases:

1. We think there's a visible tap target somewhere and we might report it as a failure, but it's actually hidden by an absolutely positioned overlay – false failure
2. There's an absolutely positioned tap target adjacent to a normal tap target, both are visible, and they should cause a failure together – could cause false pass
3. We think two tap targets conflict with each other, but actually they are on different z positions and one of them is completely invisible – would cause a false failure (the original issue I was thinking of, like, like on Wikipedia) 
4. A tap target is partially overlapped by an absolutely non-tappable element, so it should fail because it's tappable area has been reduced – false pass, rare and complex to resolve, so I think we can ignore it.

patrickhulce:
Gotcha thanks for the clarification! I'm good with whatever simplifies this first attempt the most, and we can document and deal with the other cases in a followup with getElementFromPoint 👍

mattzeunert:
@patrickhulce Done some more tweaking on this. For now it's just ignoring tap targets that are absolutely positioned (or in an absolutely positioned ancestor).

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,274 @@
...
+  const isHttpOrHttpsLink = /https?:\/\//.test(tapTarget.href);
+  if (isHttpOrHttpsLink && tapTarget.href === maybeOverlappingTarget.href) {
+    // no overlap because same target action
+    return null;
+  }
+
+  /** @type LH.Audit.TapTargetOverlapDetail | null */
+  let greatestFailure = null;
+  for (const targetCR of tappableRects) {
+    if (allRectsContainedWithinEachOther(tappableRects, maybeOverlappingTarget.clientRects)) {
```
patrickhulce: seems like this check can be run just once outside the loop (doesn't use `targetCR` at all)

is this even necessary or will running through the loop yield the same result?

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,274 @@
...
+ * @param {LH.Artifacts.Rect} cr
+ */
+function clientRectMeetsMinimumSize(cr) {
+  return cr.width >= FINGER_SIZE_PX && cr.height >= FINGER_SIZE_PX;
+}
+
+/**
+ * @param {LH.Artifacts.TapTarget} target
+ */
+function targetIsTooSmall(target) {
```
patrickhulce: let's just add a quick JSDoc comment here

```
A target is "too small" if none of it's clientRects are at least the size of a finger.

File: lighthouse-core/audits/seo/tap-targets.js [hunk 15]
```diff
@@ -0,0 +1,274 @@
...
+function getTooSmallTargets(targets) {
+  return targets.filter(targetIsTooSmall);
+}
+
+/**
+ *
+ * @param {LH.Artifacts.TapTarget[]} tooSmallTargets
+ * @param {LH.Artifacts.TapTarget[]} allTargets
+ */
+function getOverlapFailures(tooSmallTargets, allTargets) {
```
patrickhulce: nit: some naming changes would help make it clearer what calls what, throwing some ideas out there but I'm just looking for something that conveys the level of detail

is the below accurate-ish?
`getOverlapFailures` -> `getAllFailures`
`getTooCloseTargets` -> `getAllFailuresForTarget`
`getTargetTooCloseFailure` -> `getFailureForTargetPair`
`getOverlapFailure` -> `getFailureForClientRectPair`

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,274 @@
...
+        }
+      }
+    }
+  }
+  return greatestFailure;
+}
+
+/**
+ * @param {LH.Artifacts.Rect} targetCR
+ * @param {LH.Artifacts.Rect} maybeOverlappingCR
```
patrickhulce: adding explicit return here like you did above would be 👍 really helps me understand that it might not return a failure

```js
 /*
  * @returns {LH.Audit.TapTargetOverlapDetail | null}
  */
```

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,274 @@
...
+      return {
+        rawValue: false,
+        explanation:
+          'Tap targets are too small because of a missing viewport config',
+      };
+    }
+
+    const tooSmallTargets = getTooSmallTargets(artifacts.TapTargets);
+    const overlapFailures = getOverlapFailures(tooSmallTargets, artifacts.TapTargets);
+    const tableItems = getTableItems(overlapFailures);
```
patrickhulce: ❤️

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    const visibleClientRects = getVisibleClientRects(tapTargetNode);
+    if (visibleClientRects.length === 0) {
+      return;
+    }
+
+    targets.push({
+      clientRects: visibleClientRects,
+      snippet: truncate(tapTargetNode.outerHTML, 700),
```
patrickhulce: 700 sounds a bit large, were the smaller numbers you tried just not enough to figure out where it is?

lighthouse-core/test/audits/seo/tap-targets-test.js
```diff
@@ -0,0 +1,172 @@
...
+    const auditResult = auditTapTargets(
+      getBorderlineTapTargets({
+        failSecondClientRect: true,
+      })
+    );
+    assert.equal(auditResult.rawValue, false);
+  });
+
+  it('reports 4 items if the main target is overlapped both vertically and horizontally', () => {
+    // Main is overlapped by right + below, right and below are each overlapped by main
```
patrickhulce: It's probably not a good sign that I've spent so much time with this audit by now and don't know the answer to this question yet buuuuuuuuut 😄 

Isn't this usually going to be the case? If it's symmetrical should we only report one of the directions?

Seems like the cases where it is symmetrical there's not necessarily an obvious winner of who is overlapping who and it's really just one failure. Sincere apologies if we've already covered this and reached this conclusion!

lighthouse-core/audits/seo/tap-targets.js
mattzeunert: Good catch, we don't need it anymore because we now filter out child tap targets in the gatherer.

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Yeah, sometimes the outerHTML contains a lot of noise (lots of classes, children) and cutting the outerHTML off early risks losing information that's needed to identify the element.

So it's a trade-off between result size and ensuring every element is identifiable. Maybe let's go down to 300?

We should look into [showing something other than plain outerHTML](https://github.com/GoogleChrome/lighthouse/issues/6683) at some point.

lighthouse-core/test/audits/seo/tap-targets-test.js
mattzeunert: Thanks for bringing this up – there definitely is a lot of duplication:

screenshot 2019-01-04 at 09 51 40

Possible issues with merging them:
- Fixing a problem will sometimes just result in the table element being flipped (I think that's fine)
- Users won't know if the overlapping target is also failing – but it probably doesn't matter to them (Maybe they are fine with one target being hard to tap but not the other? Maybe they'd like to know that fixing some failures will have more impact on their score?)
- The result JSON won't actually say what tap targets are failing, but if we or a third party wanted to visualize the results that would be good to know. Maybe there should be a `bidrectionaFailure: true` property?

Upsides:
- Less visual noise 
- Cuts tap targets JSON size in half in many cases

Sounds like a great idea to me.

@rviscomi Any thoughts on this?

mattzeunert:
Thanks for the review @patrickhulce, addressed most of those issues now.

Saw [this PR](https://github.com/GoogleChrome/lighthouse/pull/6860) by @exterkamp, will create a new PR for tap targets i18n later on.

lighthouse-core/test/audits/seo/tap-targets-test.js
rviscomi: SGTM

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+  for (const cr of target.clientRects) {
+    if (clientRectMeetsMinimumSize(cr)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ *
```
exterkamp: ultra nit: remove extra `*`

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+ * @param {LH.Artifacts.TapTarget[]} targets
+ */
+function getTooSmallTargets(targets) {
+  return targets.filter(targetIsTooSmall);
+}
+
+/**
+ *
```
exterkamp: same ultra nit

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+      return false;
+    }
+  }
+  return true;
+}
+
+/**
+ *
+ * @param {LH.Artifacts.TapTarget[]} targets
+ */
```
exterkamp: Add typed return?
```javascript
* @returns {LH.Artifacts.TapTarget[]}
```

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+
+  return failures;
+}
+
+/**
+ * @param {LH.Artifacts.TapTarget} tapTarget
+ * @param {LH.Artifacts.TapTarget} maybeOverlappingTarget
+ * @returns {LH.Audit.TapTargetOverlapDetail | null}
+ */
+function getFailureForTargetPair(tapTarget, maybeOverlappingTarget) {
```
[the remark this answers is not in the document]
mattzeunert: Makes sense to standardize it, `getFailureForTargetPair ` should not return a `TapTargetOverlapDetail`.

Not all overlaps are failures, and the overlaps we're talking about are between fingers, not the tap target areas themselves. But we decided to use the "overlap" terminology in the UI, so using it in the code should be good too.

Maybe call them OverlapFailures?

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+    overlappingTargetScore: maybeOverlappingScore,
+  };
+}
+
+/**
+ * Only report one failure if two targets overlap each other
+ * @param {LH.Audit.TapTargetOverlapDetail[]} overlapFailures
+ * @returns {LH.Audit.TapTargetOverlapDetail[]}
+ */
+function mergeSymmetricFailures(overlapFailures) {
```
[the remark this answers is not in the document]
mattzeunert: I wrote a Map based function for this, but it ended up being twice as long, so I stuck with the nested iteration.

For a page with 423 failures `mergeSymmetricFailures` takes 2.4ms on my machine, and `getAllFailures` takes 248ms, so I don't think we need to optimize here.

lighthouse-core/audits/seo/tap-targets.js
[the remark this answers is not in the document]
exterkamp: Hm, `getOverlapFailureForTargetPair` works for me. `getTapTargetOverlapForTargetPair` seems extremely accurate but very long.

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+  const isHttpOrHttpsLink = /https?:\/\//.test(tapTarget.href);
+  if (isHttpOrHttpsLink && tapTarget.href === maybeOverlappingTarget.href) {
+    // no overlap because same target action
+    return null;
+  }
+
+  if (allRectsContainedWithinEachOther(tappableRects, maybeOverlappingTarget.clientRects)) {
+    // If one tap target is fully contained within the other that's
+    // probably intentional (e.g. an item with a delete button inside)
+    return null;
```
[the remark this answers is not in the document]
mattzeunert: We're generally ok with missing problems and err on the site of false passes. I'd say in this case it's 95% intentional. (And we also avoid failures where the child target has the same action as the parent but we can't tell that it's the same action.)

There's definitely room to improve this, rather than outright ignoring all contained tap targets. If the parent contains a very small child or multiple children that are too close together we'll miss those problems. But I think ignoring those edge cases is an acceptable trade-off against the extra complexity.

I'll add some more comments to the cases where we tolerate potential failures.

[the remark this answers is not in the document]
exterkamp: wfm

lighthouse-core/audits/seo/tap-targets.js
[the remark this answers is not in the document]
exterkamp: sgtm

lighthouse-core/test/audits/seo/tap-targets-test.js
mattzeunert: @rviscomi Right now you can't tell from the audit result which elements are failing. It could be that just the `tapTarget` is failing, or both the `tapTarget` and the `overlappingTarget` might be failing. Are we fine with that, or do we need to indicate symmetrical failures in the json somehow?

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,314 @@
...
+    const {overlapScoreRatio: failureOSR} = failure;
+    const {overlapScoreRatio: symmetricOSR} = symmetricFailure;
+    if (failureOSR > symmetricOSR || (
+      // If same score for failure and symmetric failure, pick the first one in the list
+      failureOSR === symmetricOSR &&
+      overlapFailureIndex < overlapFailures.indexOf(symmetricFailure)
+    )) {
+      failuresAfterMerging.push(failure);
+    } else {
+      // The symmetric failure will be pushed instead
```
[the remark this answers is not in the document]
mattzeunert: Good point, thanks for pointing this out 👍

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Changed it to 300 now.

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,319 @@
...
+    const headings = [
+      {key: 'tapTarget', itemType: 'node', text: 'Tap Target'},
+      {key: 'size', itemType: 'text', text: 'Size'},
+      {key: 'overlappingTarget', itemType: 'node', text: 'Overlapping Target'},
+    ];
+
+    const details = Audit.makeTableDetails(headings, tableItems);
+
+    const tapTargetCount = artifacts.TapTargets.length;
+    const failingTapTargetCount = new Set(overlapFailures.map(f => f.tapTarget)).size;
```
patrickhulce: maybe add a quick comment that we intentionally want to use `overlapFailures` and not `uniqueOverlapFailures`? I could see myself in 2 years looking at this code and thinking it was a bug without the context :)

types/audit.d.ts
```diff
@@ -70,6 +70,28 @@ declare global {
...
+      tapTargetScore: number;
+      overlappingTargetScore: number;
+    }
+    
+    export interface TapTargetOverlapFailure extends ClientRectOverlapFailure {
+      tapTarget: Artifacts.TapTarget;
+      overlappingTarget: Artifacts.TapTarget;
+    }
+
+    export type TapTargetTableItem = {
```
patrickhulce: any reason this isn't an `interface`? can we make it an `interface`?

lighthouse-core/audits/seo/tap-targets.js
mattzeunert: Good point! I renamed `uniqueOverlapFailures` to `overlapFailuresForDisplay`, do you think that also works?

lighthouse-core/test/audits/seo/tap-targets-test.js
rviscomi: I'm fine with the deduplication, even with the ambiguity it may introduce.

If tap targets are overlapping I think the natural response is to move them further apart, which solves each underlying failure, as opposed to increasing the dimensions of one target or another.

So let's launch with the simplified results, follow user feedback closely to see if there's any confusion, and revisit the config accordingly.

types/audit.d.ts
mattzeunert: The `makeTableDetails` items need to be `Array<{[x: string]: DetailsItem}>`, so we need to guarantee that the index is always a string. And I guess types automatically do that because they aren't extendable?

Anyway, I've changed it to an interface that extends the `DetailsRendererDetailsJSON` item type.

types/audit.d.ts
```diff
@@ -70,17 +70,43 @@ declare global {
...
+    export interface ClientRectOverlapFailure  {
+      overlapScoreRatio: number;
+      tapTargetScore: number;
+      overlappingTargetScore: number;
+    }
+    
+    export interface TapTargetOverlapFailure extends ClientRectOverlapFailure {
+      tapTarget: Artifacts.TapTarget;
+      overlappingTarget: Artifacts.TapTarget;
+    }
```
brendankenny: are these used only in the audit? If so, should move them as typedefs in that file as they aren't really needed in the audit type definition file (I think you still can't use `extends` for typedefs in js, unfortunately, but it's just three properties so not a big deal to repeat).

`ClientRectOverlapFailure` may not even need to be a typedef if it looks ok inline in that `@return` statement

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,319 @@
...
+   * @return {LH.Audit.Meta}
+   */
+  static get meta() {
+    return {
+      id: 'tap-targets',
+      title: 'Tap targets are sized appropriately',
+      failureTitle: 'Tap targets are not sized appropriately',
+      description:
+        'Interactive elements like buttons and links should be large enough (48x48px), and have enough space around them, to be easy enough to tap without overlapping onto other elements. [Learn more](https://developers.google.com/web/fundamentals/accessibility/accessible-styles#multi-device_responsive_design).',
+      requiredArtifacts: ['Viewport', 'TapTargets'],
```
brendankenny: this will have to be updated now after #7025

lighthouse-core/test/results/sample_v2.json
```diff
@@ -2677,6 +2677,15 @@
...
       "errorMessage": "Required RobotsTxt gatherer did not run."
     },
+    "tap-targets": {
+      "id": "tap-targets",
+      "title": "Tap targets are not sized appropriately",
+      "description": "Interactive elements like buttons and links should be large enough (48x48px), and have enough space around them, to be easy enough to tap without overlapping onto other elements. [Learn more](https://developers.google.com/web/fundamentals/accessibility/accessible-styles#multi-device_responsive_design).",
+      "score": null,
+      "scoreDisplayMode": "error",
+      "rawValue": null,
+      "errorMessage": "Required TapTargets gatherer did not run."
```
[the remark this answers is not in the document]
brendankenny: > we should update the artifacts so we can get results here.

Actually, I see this has been discussed and decided to punt on this. That's fine, we should just update in a follow up

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+    return true;
+  }
+  return nodeHasParentTapTarget(node.parentElement);
+}
+
+/**
+ * Check if node is in a block of text, such as paragraph with a bunch of links in it.
+ * Makes a reasonable guess, but for example gets it wrong if the element is surounded by other
+ * HTML elements instead of direct text nodes.
+ * @param {Node} node
```
brendankenny: I believe this is always an `Element`, since it's called with the results of `getElementsInDocument` (which gets its results from `querySelectorAll`, which always returns elements) and with `node.parentElement`, which is always an `Element` as well (or null, of course))

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+ * @returns {boolean}
+ */
+/* istanbul ignore next */
+function nodeIsInTextBlock(node) {
+  /**
+   * @param {Node} node
```
brendankenny: also an `Element` if the above is true

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+   * @returns {boolean}
+   */
+  function isInline(node) {
+    if (node.nodeType === Node.TEXT_NODE) {
+      return true;
+    }
+    if (node.nodeType !== Node.ELEMENT_NODE) {
```
brendankenny: if an `Element`, these two cases can't ever happen. If we can delete those, might be cleaner to just drop the `isInline` function and inline the `getComputedStyle` check into the if statement where it's used below

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+      return false;
+    }
+    const {display} = getComputedStyle(/** @type {Element} */ (node));
+    return display === 'inline' || display === 'inline-block';
+  }
+
+  /**
+   * @param {Node} node
```
brendankenny: also `Element`

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+      // is not a text block container
+      return false;
+    }
+
+    for (const sibling of node.parentElement.childNodes) {
+      if (sibling === node) {
+        continue;
+      }
+      const siblingTextContent = (sibling.textContent || '').trim();
+      if (sibling.nodeType === Node.TEXT_NODE && siblingTextContent.length > 0) {
```
brendankenny: why doesn't the `parentElement.textContent` check above catch this case? Shouldn't all the text of any siblings be in there?

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+    if ((node.clientWidth === 0 && overflowX === 'hidden') ||
+        (node.clientHeight === 0 && overflowY === 'hidden')) {
+      return false;
+    }
+  }
+
+  const parent = node.parentElement;
+  if (parent && parent.tagName !== 'BODY' && !nodeIsVisible(parent)) {
+    // if a parent is invisible then the current node is also invisible
+    return false;
```
brendankenny: can do `return nodeIsVisible(parent);` here (and make the conditional simpler)

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+  return clientRects;
+}
+
+/**
+ * @param {Element} node
+ * @param {boolean} includeChildren
+ * @returns {LH.Artifacts.Rect[]}
+ */
+/* istanbul ignore next */
+function getClientRects(node, includeChildren = true) {
```
brendankenny: all uses appear to either explicitly pass `true` for `includeChildren` or use the default `true` value, so drop the parameter?

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+  /** @type {LH.Artifacts.Rect[]} */
```
brendankenny: no need for the type annotation

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+  let clientRects = Array.from(
```
brendankenny: nit: can be `const` if the `clientRects.concat` below switches to `clientRects.push(...getClientRects(child))`

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+/**
+ * @param {Element} node
+ */
+/* istanbul ignore next */
+function getVisibleClientRects(node) {
+  if (!nodeIsVisible(node)) {
+    return [];
+  }
+
+  const {
```
brendankenny: move inside `if (allClientRectsEmpty(clientRects)) {` block

brendankenny:
we should think about setting up some sort of jsdom-based `evaluateAsync` tester in the future. Something like this gatherer (only walking and querying the DOM) could be tested really well that way.

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+
+  return clientRects;
+}
+
+/**
+ * @param {Element} node
+ * @returns {boolean}
+ */
+/* istanbul ignore next */
+function nodeHasParentTapTarget(node) {
```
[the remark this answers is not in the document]
mattzeunert: Good points!

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Yes, but parent text content also contains the content of siblings that are elements rather than text nodes.

This should not be detected as a text block:

```
<div>
   <a onClick=()>Do something</a>
   <a onClick=()>Do more stuff</a>
</div>
```

This should be counted as a text block:

```
<div>
   <a>Lighthouse</a> is a tool for web developers.
</div>
```

This should be ideally be detected as a text block but it isn't, and we're ok with that:

```
<div>
   <a>Lighthouse</a> <span>is a tool for web developers.</span>
</div>
```

Will add a comment to the code.

types/audit.d.ts
mattzeunert: Makes sense, they don't need to be in there 👍

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,351 @@
...
+  if (!node.parentElement) {
+    return false;
+  }
+  if (node.parentElement.matches(tapTargetsSelector)) {
```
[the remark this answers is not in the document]
mattzeunert: Cool, changed that! Should we add some kind of test case for when `matches` is overwritten?

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Good catch!

lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -174,4 +174,18 @@ module.exports = [
...
   },
+  {
+    requestedUrl: BASE_URL + 'seo-tap-targets.html',
+    finalUrl: BASE_URL + 'seo-tap-targets.html',
+    audits: {
+      'tap-targets': {
+        score: 0.9, // 10 passing targets/11 total visible targets
+        details: {
+          items: {
+            length: 1,
```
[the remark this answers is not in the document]
mattzeunert: Added an assertion for the item details.

It would be nice to assert the full list of detected tap targets from the artifacts. Currently smokehouse just provides the LHR though.

mattzeunert:
Thanks for the initial feedback @brendankenny!

> Looks good but biiiiiiiig :)

Let me know if you see a good way I could have split the PR into smaller bits. Might make future PRs easier 🙂.

@patrickhulce already told me to split off the two smaller PRs for [rect helpers](https://github.com/GoogleChrome/lighthouse/pull/6703) and [page functions](https://github.com/GoogleChrome/lighthouse/pull/6702).

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
brendankenny: >  Should we add some kind of test case for when matches is overwritten?

We do have [the dbw smoke test check that it's possible to work around the issue](https://github.com/GoogleChrome/lighthouse/blob/8ed55613d41e47e44d2be032e671571177aa5c53/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html#L124), but we'd have to add it to the SEO smoke test as well to catch regressions in this use. However...

> actually is it possible for us to just `useIsolation` here? 

this is a good idea and lets the code skip using the cached version. [Example in `link-elements.js`](https://github.com/GoogleChrome/lighthouse/blob/8ed55613d41e47e44d2be032e671571177aa5c53/lighthouse-core/gather/gatherers/link-elements.js#L32).

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,343 @@
...
+ * HTML elements instead of direct text nodes.
+ * @param {Element} element
+ * @returns {boolean}
+ */
+/* istanbul ignore next */
+function elementIsInTextBlock(element) {
+  /**
+   * @param {Element} element
+   */
+  function hasTextNodeSiblingsFormingTextBlock(element) {
```
brendankenny: `hasTextNodeSiblingsFormingTextBlock` doesn't rely on being closed over by `elementIsInTextBlock`, so move outside to a top-level function?

lighthouse-core/gather/gatherers/seo/tap-targets.js
brendankenny: > Will add a comment to the code.

great, that definitely helps

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -0,0 +1,343 @@
...
+  }
+
+  let clientRects = getClientRects(element);
+
+  if (allClientRectsEmpty(clientRects)) {
+    const {
+      overflowX,
+      overflowY,
+    } = getComputedStyle(element);
+    if ((overflowX === 'hidden' && overflowY === 'hidden') || element.children.length === 0) {
```
brendankenny: if all the client rects are size 0, does it matter whether or not the children would be visible since their client rects are also size 0 (since they're included in the `allClientRectsEmpty(clientRects)` check)?

types/artifacts.d.ts
```diff
@@ -115,6 +115,12 @@ declare global {
       StartUrl: {statusCode: number, explanation?: string};
       /** Information on <script> and <link> tags blocking first paint. */
       TagsBlockingFirstPaint: Artifacts.TagBlockingFirstPaint[];
+      /** Information about tap targets including their position and size. */
+      TapTargets: Artifacts.TapTarget[];
+      /** The value of the <meta name="theme=color">'s content attribute, or null. */
+      ThemeColor: string|null;
+      /** The value of the <meta name="viewport">'s content attribute, or null. */
+      Viewport: string|null;
```
brendankenny: `ThemeColor` and `Viewport` accidentally got merged back in here

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
mattzeunert: Nice, `useIsolation` works. And once we update the sample json it will run the tap targets gatherer there too and would pick it up if we broke this in the future?

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Makes sense, I was trying to avoid having another function that needs to be imported in the evalAsync string, but this is cleaner 👍

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
mattzeunert: And currently there's no full E2E LH run where we could overwrite `matches`? If that's the case I'll add it to the `seo-tap-targets.html`.

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: No it shouldn't matter, and from trying it out on a few sites it doesn't matter. Will remove the if.

lighthouse-core/gather/gatherers/seo/tap-targets.js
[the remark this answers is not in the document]
patrickhulce: The smokehouse tests are the only E2E tests we have, so that's the right place to put such a test 👍

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,334 @@
...
+ */
+function getTableItems(overlapFailures) {
+  const tableItems = overlapFailures.map(
+    ({
+      tapTarget,
+      overlappingTarget,
+      tapTargetScore,
+      overlappingTargetScore,
+      overlapScoreRatio,
+    }) => {
```
brendankenny: style nit: I don't personally find this very readable, not sure how others feel about it. You've lost track of the `map()` by the time you get to the bottom of the destructured props :) I'd rather just have something like
```js
const tableItems = overlapFailures.map(failure => {
  const largestCR = getLargestRect(failure.tapTarget.clientRects);
  // ...
});
```

types/audit.d.ts
```diff
@@ -70,17 +70,32 @@ declare global {
       wastedPercent?: number;
     }
 
+    export interface TapTargetTableItem extends DetailsObject {
```
brendankenny: can this be moved to `tap-targets.js` as well? Since it's a `DetailsObject` and the table renderer can handle it, seems like it's ok to be internal to the audit until/unless we need it externally

lighthouse-core/test/audits/seo/tap-targets-test.js
```diff
@@ -0,0 +1,184 @@
...
+
+const TapTargetsAudit = require('../../../audits/seo/tap-targets.js');
+const assert = require('assert');
+
+function auditTapTargets(tapTargets) {
+  const artifacts = {
+    TapTargets: tapTargets,
+    MetaElements: [{
+      name: 'viewport',
+      content: 'width=device-width',
```
brendankenny: There are a few cases that still need covering, e.g. need viewport failure test(s). See the red lines in https://coveralls.io/builds/21238984/source?filename=lighthouse-core%2Faudits%2Fseo%2Ftap-targets.js#L42

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,334 @@
...
+      requiredArtifacts: ['MetaElements', 'TapTargets'],
+    };
+  }
+
+  /**
+   * @param {LH.Artifacts} artifacts
+   * @return {LH.Audit.Product}
+   */
+  static audit(artifacts) {
+    const hasViewportSet = ViewportAudit.audit(artifacts).rawValue;
```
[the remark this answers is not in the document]
mattzeunert: https://github.com/GoogleChrome/lighthouse/issues/7084

lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -0,0 +1,334 @@
...
+    if (!hasViewportSet) {
+      return {
+        rawValue: false,
+        explanation:
+          'Tap targets are too small because of a missing viewport config',
```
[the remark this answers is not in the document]
mattzeunert: Yeah it should fail if the viewport tag doesn't make it work for mobile. What do you think of this explanation:

> Tap targets are too small because there's no viewport meta tag optimized for mobile screens

Changed it here and in the font size audit.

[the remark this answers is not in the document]
brendankenny: > What do you think of this explanation

sounds great!

PR #5849, opened 2018-08-16
cli(tsc): make LH.Flags type correct and consistent

brendankenny:
I've had a few TODOs sprinkled in the code to `fix Flags type`. I'm not exactly sure what I meant by that, but this PR makes sure our uses of `Flags` are all correct and gets to remove two places (`index.js` and `popup.js`) where we were coercing empty objects into Flags.

We allowed empty objects as flags there, and now the type system actually verifies that that's completely acceptable :)

We already had `LH.Flags` vs `LH.SharedFlagsSettings` (all properties shared by `LH.Flags` and `LH.Config.Settings`), and now I've added another level with `LH.Flags` vs `LH.CliFlags` (all properties *in addition* to `LH.Flags` that control just CLI functionality). This is another step in making `Flags` a little easier to reason about, because once they reach core, all those CLI flags aren't visible anymore (according to the type system). This means code really doesn't have to care if `Flags` originated from the CLI, someone calling the node module, the extension, or devtools.

Also spruces up type annotations in the CLI files (mostly removing `!`s), which haven't been touched in a while.

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -160,18 +160,18 @@ function assertValidGatherer(gathererInstance, gathererName) {
 /**
  * Creates a settings object from potential flags object by dropping all the properties
  * that don't exist on Config.Settings.
- * TODO(bckenny): fix Flags type
  * @param {Partial<LH.Flags>=} flags
- * @return {Partial<LH.Config.Settings>}
- */
+ * @return {RecursivePartial<LH.Config.Settings>}
+*/
 function cleanFlagsForSettings(flags = {}) {
-  /** @type {Partial<LH.Config.Settings>} */
+  /** @type {RecursivePartial<LH.Config.Settings>} */
   const settings = {};
 
   for (const key of Object.keys(flags)) {
     // @ts-ignore - intentionally testing some keys not on defaultSettings to discard them.
     if (typeof constants.defaultSettings[key] !== 'undefined') {
-      const safekey = /** @type {keyof LH.SharedFlagsSettings} */ (key);
+      // Cast since key now must be able to index both Flags and Settings.
+      const safekey = /** @type {Extract<keyof LH.Flags, keyof LH.Config.Settings>} */ (key);
       settings[safekey] = flags[safekey];
     }
   }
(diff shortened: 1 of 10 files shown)


lighthouse-core/config/config.js
```diff
@@ -160,18 +160,18 @@ function assertValidGatherer(gathererInstance, gathererName) {
 /**
  * Creates a settings object from potential flags object by dropping all the properties
  * that don't exist on Config.Settings.
- * TODO(bckenny): fix Flags type
  * @param {Partial<LH.Flags>=} flags
- * @return {Partial<LH.Config.Settings>}
- */
+ * @return {RecursivePartial<LH.Config.Settings>}
```
brendankenny: It's a bug that this wasn't flagged before, but it wasn't until I made any change to `LH.Flags.output` that it started to be flagged (even though the problem is with `LH.Flags.throttling`, weirdly enough).

We just want to copy any `Settings` properties over from an instance of `Flags`, so the result will only have `Settings` properties but some may not be set. So `Partial<LH.Config.Settings>` would be correct if `Settings` were a simple object.

The issue is that one of the properties (`throttling`) is itself an object, and while `Flags.throttling` has all optional properties, on `Settings.throttling` they're all required. This change applies `Partial<>` to all of `Settings` properties recursively. It could special case `throttling` instead, but this will catch any future objects added to `SharedFlagsSettings` as well.

typings/externs.d.ts
```diff
@@ -77,29 +89,36 @@ declare global {
...
+     * for controlling the CLI.
+     */
+    export interface CliFlags extends Flags {
       _: string[];
       chromeFlags: string;
       outputPath: string;
       saveAssets: boolean;
       view: boolean;
-      enableErrorReporting: boolean;
+      enableErrorReporting?: boolean;
```
paulirish: make this a yargs boolean instead. seems weird to have this one optional but the rest not.

brendankenny: done

patrickhulce: wait noooooo, we gotta revert this. it's purposefully not a yargs boolean so that the setting saved to their config will be respected. explicitly setting as false via yargs will always turn error reporting off

brendankenny: whoops, very good point. There *was* a reason it wasn't in `yargs.boolean()` already :) I'll revert and add a comment so we don't do that again

Issue #4403, opened 2018-01-31
Using lighthouse in combination with Puppeteer


Antonhansel:
Hello,

I'm would like to launch a chrome instance, connect `puppeteer` to it to login, and then start `lighthouse`, is that possible ? I couldn't find a way to connect `puppeteer` to a chrome instance started with `chrome-launcher`.

I've tried something like this but with no success

```
function launchChromeAndRunLighthouse(url, opts, config = null) {
	return chromeLauncher.launch({chromeFlags: opts.chromeFlags}).then(async chrome => {
		opts.port = chrome.port;
		const url = 'ws://localhost:' + chrome.port + '/devtools/browser/';
		const browser = await puppeteer.connect({
			browserWSEndpoint: url
		});
                // this throws the following WS error: (node:9000)
                // UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: unexpected server response (404)
```

because I'm missing the browser id after `browser/` and it doesn't seem to be exposed by `lighthouse`

I've seen issues where it says to manually connect before running `lighthouse` on a website where login is mandatory. Is there any other fully automated way to do this ?

ebidel:
This should work:

```js
const chromeLauncher = require('chrome-launcher');
const puppeteer = require('puppeteer');
const lighthouse = require('lighthouse');
const request = require('request');
const util = require('util');

(async() => {

const URL = 'https://www.chromestatus.com/features';

const opts = {
  chromeFlags: ['--headless'],
  logLevel: 'info',
  output: 'json'
};

// Launch chrome using chrome-launcher.
const chrome = await chromeLauncher.launch(opts);
opts.port = chrome.port;

// Connect to it using puppeteer.connect().
const resp = await util.promisify(request)(`http://localhost:${opts.port}/json/version`);
const {webSocketDebuggerUrl} = JSON.parse(resp.body);
const browser = await puppeteer.connect({browserWSEndpoint: webSocketDebuggerUrl});

// Run Lighthouse.
const lhr = await lighthouse(URL, opts, null);
console.log(`Lighthouse score: ${lhr.score}`);

await browser.disconnect();
await chrome.kill();

})();
```

Antonhansel:
Thank you this is exactly what I was looking for. I couldn't find this snippet in the docs, did I miss it ?

Antonhansel:
[...]

PR #5855, opened 2018-08-17
extension: access core through module instead of Runner

brendankenny:
Makes the extension (and devtools/psi) access lighthouse-core through the module interface instead of directly calling `Runner.run()`.  With the last few simplifications to flags/settings/config, this refactor is now really simple and means we have a single core interface for all our platforms.

Works by allowing a `Connection` to be passed into `lighthouse()` instead of always using a desktop `ChromeProtocol` connection. This is something we have previously discussed with the puppeteer team to make #4403 easy, so it seems like a good move.

@paulirish I think this will also ease other refactors you've talked about for all the `runLighthouseIn*` methods


lighthouse-core/index.js
```diff
@@ -18,27 +20,30 @@ const Config = require('./config/config');
...
+ * @param {string=} url The URL to test. Optional if running in auditMode.
+ * @param {LH.Flags=} flags Optional settings for the Lighthouse run. If present,
+ *   they will override any settings in the config.
+ * @param {LH.Config.Json=} configJSON Configuration for the Lighthouse run. If
+ *   not present, the default config is used.
+ * @param {Connection=} connection
  * @return {Promise<LH.RunnerResult|undefined>}
  */
-async function lighthouse(url, flags = {}, configJSON) {
+async function lighthouse(url, flags = {}, configJSON, connection) {
```
brendankenny: if we feel like this is going to become a never-ending list of optional arguments, we could make this an object (`overrides`? or...something)

patrickhulce: 4️⃣. 0️⃣  😄

lighthouse-extension/gulpfile.js
```diff
@@ -131,6 +131,9 @@ gulp.task('browserify-lighthouse', () => {
       .ignore('rimraf')
       .ignore('pako/lib/zlib/inflate.js');
 
+      // Don't include the desktop protocol connection.
+      bundle.ignore(require.resolve('../lighthouse-core/gather/connections/cri.js'));
```
patrickhulce: score!

lighthouse-core/index.js
brendankenny: > 4️⃣. 0️⃣

ha. Just to be clear, ambiguous smiley face, this should be fine since it's just an addition to our API, not a change, right?

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -53,13 +54,21 @@ function updateBadgeUI(optUrl) {
...
   const url = await connection.getCurrentTabURL();
-  const runnerResult = await background.runLighthouseForConnection(connection, url, flags,
-    categoryIDs, updateBadgeUI);
+  const config = background.getDefaultConfigForCategories(categoryIDs);
+
+  updateBadgeUI(url);
+  let runnerResult;
+  try {
+    runnerResult = await lighthouse(url, flags, config, connection);
+  } finally {
```
paulirish: l33t

paulirish:
this is so great.

PR #5857, opened 2018-08-17
core: enhanced noopener output

patrickhulce:
**Summary**
cleans up TODO for including outerHTML in noopener link audit and does some related page-functions/dom-helpers cleanup along the way :)


lighthouse-core/lib/dom-helpers.js
```diff
@@ -1,46 +0,0 @@
...
-      }
-    }
-  };
-  _findAllElements(document.querySelectorAll('*'));
-
-  return results;
-}
-
-module.exports = {
-  getElementsInDocumentFnString: getElementsInDocument.toString(),
```
brendankenny: @paulirish I can't remember if you were an advocate for all the page functions to be exposed as strings or if it was the opposite (or if your opinion has changed over time)

paulirish: hmmm i dont think i cared, but i _like_ the idea of them being exported as strings.

we can do in a followup. 👩‍🚒

patrickhulce: followup sg :)

PR #5858, opened 2018-08-18
core(tsc): type check config files

brendankenny:
This was flagged testing the typescript nightly against the current code base. `LH.Config.Category.description` and `LH.Config.Group.description` should both be optional [as they are in the LHR](https://github.com/GoogleChrome/lighthouse/blob/d1031575f3f79d7bedbe0f8d5e45a51e16189853/typings/lhr.d.ts#L60), and `LH.Config.Category.manualDescription` should also exist.

Also adds type annotations to the files themselves so that they're checked directly, rather than indirectly through their use in `config.js`.

### DIFF
--- a/lighthouse-core/config/full-config.js
+++ b/lighthouse-core/config/full-config.js
@@ -5,7 +5,8 @@
  */
 'use strict';
 
-module.exports = {
+/** @type {LH.Config.Json} */
+const fullConfig = {
   extends: 'lighthouse:default',
   settings: {},
   passes: [
@@ -19,6 +20,7 @@ module.exports = {
   audits: [
     'byte-efficiency/unused-javascript',
   ],
+  // @ts-ignore TODO(bckenny): type extended Config where e.g. category.title isn't required
   categories: {
     'performance': {
       auditRefs: [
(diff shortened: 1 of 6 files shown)


lighthouse-core/config/full-config.js
```diff
@@ -19,6 +20,7 @@ module.exports = {
   audits: [
     'byte-efficiency/unused-javascript',
   ],
+  // @ts-ignore TODO(bckenny): type extended Config where e.g. category.title isn't required
```
brendankenny: If a config is extending another one, there's no reason for it to have *anything* in particular defined, so maybe these could be `RecursivePartial<LH.Config.Json>`, but we don't want to apply that to e.g. `defaultConfig`. So we'll have to figure out a good way to conditionally apply it another day.

patrickhulce: Maybe we have to split the full config typedef from the valid config typedef?

brendankenny: > Maybe we have to split the full config typedef from the valid config typedef?

yeah, that seems like that might be the only decent way to do it without making the full config typedef useless

PR #5859, opened 2018-08-18
core(tsc): add explicit index signature in mainthread-work-breakdown

brendankenny:
Another change anticipating typescript 3.1nightly. Default index signatures are going away in JS ([my fault](https://github.com/Microsoft/TypeScript/pull/25996#issuecomment-408284828) :), so we'll need to migrate anything relying on that to either fully defined objects or explicit index signatures. We use those in a lot of places, so I thought I'd start now :)

Here `TaskGroupIds` just becomes the object's signature.

### DIFF
--- a/lighthouse-core/lib/task-groups.js
+++ b/lighthouse-core/lib/task-groups.js
@@ -5,9 +5,11 @@
  */
 'use strict';
 
+/** @typedef {'parseHTML'|'styleLayout'|'paintCompositeRender'|'scriptParseCompile'|'scriptEvaluation'|'garbageCollection'|'other'} TaskGroupIds */
+
 /**
  * @typedef TaskGroup
- * @property {string} id
+ * @property {TaskGroupIds} id
  * @property {string} label
  * @property {string[]} traceEventNames
  */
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/task-groups.js
```diff
@@ -5,9 +5,11 @@
  */
 'use strict';
 
+/** @typedef {'parseHTML'|'styleLayout'|'paintCompositeRender'|'scriptParseCompile'|'scriptEvaluation'|'garbageCollection'|'other'} TaskGroupIds */
```
patrickhulce: I've been wondering is there not a way to get these programmtically via some `keyof` magic?

patrickhulce: I know we're going the opposite direction by declaring the type of the object as matching this, but ya know just for knowledge :)

brendankenny: > I've been wondering is there not a way to get these programmtically via some keyof magic?

I tried really hard :) but I couldn't find a way to get out of the circularity of needing `TaskGroup.id` to be able to index `taskGroups` in `mainthread-work-breakdown.js`, which means the `taskGroups` declaration needs to be typed, not just inferred (it would go to `string`), which means you can't meaningfully do `keyof` on it

brendankenny: actually, you could do one really terrible thing...
```js
const taskGroups = {
  parseHTML: {
    id: /** @type {'parseHTML'} */ ('parseHTML'),
    // ...
  },
  styleLayout: {
    id: /** @type {'styleLayout'} */ ('styleLayout'),
    // ...
  }, // ...
};
```

patrickhulce: alright thanks good to know!

brendankenny:
@patrickhulce are you 👍 for this then?

Issue #5873, opened 2018-08-20
tsc: don't rely on implicit index signature for objects


brendankenny:
tracking issue for updates needed for the upcoming TypeScript 3.1, [which will no longer](https://github.com/Microsoft/TypeScript/pull/25996#issuecomment-408284828) automatically add `[x: string]: any;` to any type inferred from an object declaration. This is only a change for js; this was never done in ts.

It was helpful in cases where we were mapping specific small objects to a general grab-bag type (e.g. `{type: 'code', snippet: str}` and then saying it was of a more general details item type), and we'll have to explicitly declare a few more types to fix those places.

But it means certain errors will now be caught. e.g. in TS≤3.0 in `lighthouse-ext-background.js`, if you refer to `background.getDefaultCategories()` the return value will be type checked, but if you do `background.getCategoriez()` the compiler will think the call is valid and returns an `any`, because the compiler was automatically adding a `[x: string]: any;` to the [`lighthouse-background.js` export](https://github.com/GoogleChrome/lighthouse/blob/020333031f66684c4c153d080fc42b29c384f732/lighthouse-extension/app/src/lighthouse-background.js#L65-L70).

Before I started we had almost 100 errors from this, and fixing them only makes things more correct (even in 3.0), so I thought I'd start early. In many places a single line fix and multiple errors are resolved, so it shouldn't be too bad.

- [x] config files ~#5858~
- [x] mainthread-work-breakdown/task-groups ~#5859~
- [x] many places ~#5874~
- [x] traceOfTab (and fix optional properties) ~#5887~
- [x] LHError ~#5896~
- [x] `load-simulator`/`simulator.js` ~#5897~
- [x] Viewer ~#5942~
- [x]  `DetailsItem`

brendankenny:
This is complete with #5942 (though detailsItem still needs work independent of this)

PR #5874, opened 2018-08-20
core(tsc): remove more reliance on implicit index signatures

brendankenny:
part of #5873

No real theme here, just scattered fixes where none needed major interventions. Resolves 23 errors when using `typescript@next`.

The other side of some of these fixes is going explicitly `any` or `ts-ignore`, but it didn't seem worth losing refactoring, intellisense, etc for most of these

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
@@ -11,8 +11,11 @@ const NetworkRequest = require('../../network-request');
 // Assume that 40% of TTFB was server response time by default for static assets
 const DEFAULT_SERVER_RESPONSE_PERCENTAGE = 0.4;
 
-// For certain resource types, server response time takes up a greater percentage of TTFB (dynamic
-// assets like HTML documents, XHR/API calls, etc)
+/**
+ * For certain resource types, server response time takes up a greater percentage of TTFB (dynamic
+ * assets like HTML documents, XHR/API calls, etc)
+ * @type {Partial<Record<LH.Crdp.Page.ResourceType, number>>}
+ */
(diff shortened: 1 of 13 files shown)


lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
```diff
@@ -73,9 +76,11 @@ class NetworkAnalyzer {
     return summaryByKey;
   }
 
+  /** @typedef {{record: LH.Artifacts.NetworkRequest, timing: LH.Crdp.Network.ResourceTiming, connectionReused?: boolean}} RequestInfo */
```
brendankenny: I apologize for the terrible name here, @patrickhulce :)

lighthouse-core/runner.js
```diff
@@ -368,6 +368,7 @@ class Runner {
       const ArtifactClass = require('./gather/computed/' + filename);
       const artifact = new ArtifactClass(computedArtifacts);
       // define the request* function that will be exposed on `artifacts`
+      // @ts-ignore - doesn't have an index signature, so can't be set dynamically.
       computedArtifacts['request' + artifact.name] = artifact.request.bind(artifact);
```
brendankenny: `artifact` and `ArtifactClass` are already `any` here, and `computedArtifacts` already has a `TODO` to refactor above this method, so it seems fine to just ignore and keep the issues contained here than to make `computedArtifacts` have a generic index signature and have it leak into audits, etc.

lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
patrickhulce: heh, works for me :)

PR #5878, opened 2018-08-21
core(tracing-processor): throw on no top level events

patrickhulce:
We were throwing on no toplevel events but only in the observed case, this PR adds the check in more places so we should fail all metrics all things that rely on the page dependency graph if there are no toplevel events.

### DIFF
--- a/lighthouse-core/lib/traces/tracing-processor.js
+++ b/lighthouse-core/lib/traces/tracing-processor.js
@@ -19,7 +19,7 @@ const LHError = require('../errors');
 class TraceProcessor {
   /**
    * There should *always* be at least one top level event, having 0 typically means something is
-   * drastically wrong with the trace and would should just give up early and loudly.
+   * drastically wrong with the trace and we should just give up early and loudly.
    *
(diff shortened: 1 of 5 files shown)


lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -17,6 +17,20 @@ const SCHEDULABLE_TASK_TITLE_ALT2 = 'ThreadControllerImpl::RunTask';
 const LHError = require('../errors');
 
 class TraceProcessor {
+  /**
+   * There should *always* be at least one top level event, having 0 typically means something is
+   * drastically wrong with the trace and would should just give up early and loudly.
```
brendankenny: would should
:)

patrickhulce:
> If a trace was collected, assertHasToplevelEvents (similar to checking for a main resource)? Or is that overkill

Seems a bit like overkill to me, this is already the entire performance section basically, so I think we're pretty covered.

brendankenny:
prlin is fixd :)

PR #5880, opened 2018-08-21
i18n: add i18n 'type' for prlint

paulirish:
Seems better to have all that separated from `Core` in the changelog.


patrickhulce:
> prlin

I think that's supposed to be prlint

PR #5885, opened 2018-08-21
extension: fix 'Couldn't resolve current tab'

### LINKED ISSUE #5668: Extension Error: Couldn't resolve current tab. Check your URL, reload, and tr
**Lighthouse Version**: 3.0.2
**Lighthouse Commit**: ff7d162564421094b072ed0b4434a3853bbc6950
**Chrome Version**: 67.0.3396.99
**Initial URL**: http://localhost:8081/app
**Error Message**: Couldn't resolve current tab. Check your URL, reload, and try again. Found 1 tab(s) with id(s) [211].
**Stack Trace**:
```
Error: Couldn't resolve current tab. Check your URL, reload, and try again. Found 1 tab(s) with id(s) [211].
    at Object.chrome.tabs.query.tabs [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17347:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

wardpeet:
**Summary**

bugfix

it fixes the cannot resolve tab issue in the extension
lighthouse-resolve-tab

**Related Issues/PRs**

fixes #5668 

This fix makes it easier to get this bug #5872

### DIFF
--- a/lighthouse-core/gather/connections/extension.js
+++ b/lighthouse-core/gather/connections/extension.js
@@ -128,17 +128,10 @@ class ExtensionConnection extends Connection {
           return reject(new Error(errMessage));
         }
         if (tabs.length > 1) {
-          log.warn('ExtensionConnection', '_queryCurrentTab returned multiple tabs');
+          return reject(new Error(`${errMessage} _queryCurrentTab returned multiple tabs`));
         }
 
-        const firstUrledTab = tabs.find(tab => !!tab.url);
-        if (!firstUrledTab) {
-          const tabIds = tabs.map(tab => tab.id).join(', ');
-          const message = errMessage + ` Found ${tabs.length} tab(s) with id(s) [${tabIds}].`;
-          return reject(new Error(message));
-        }
-
-        resolve(firstUrledTab);
+        resolve(tabs[0]);
       }));
     });
   }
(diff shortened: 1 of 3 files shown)


brendankenny:
Maybe `popup.js` could also query and save the tab ID and the `ExtensionProtocol` constructor could take that as an argument, rather than trying to get it itself. We'd have to think through if there are downsides to that rather than letting it take care of itself...

wardpeet:
Not 100% sure what you're referring too. We fail on the first call `const url = await connection.getCurrentTabURL();` [lighthouse-ext-background.js#L67](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-extension/app/src/lighthouse-ext-background.js#L67).

the tabid can still be resolved just fine, I could send it to the connection to keep consistency maybe.

GodLesZ:
Any progress on this @brendankenny @wardpeet ?
Just getting the error all the time and a fix would be rly nice

wardpeet:
@GodLesZ could you tell us which site you're auditing and what extensions you have ?

GodLesZ:
> @GodLesZ could you tell us which site you're auditing and what extensions you have ?

I'm testing the chrome extension on the latest chrome (already tested on canary & stable) and always get this error.
Maybe this is related to the fact, that the SSL certificate of the (only local accessible) page is not valid?

Currently we're not able to get lighthouse reports of untrusted websites, so I hope this extension is able to generate this?

brendankenny:
At this point, I don't think we're going to move on this :)

PR #5887, opened 2018-08-21
core(tsc): tighten traceOfTab timing types

brendankenny:
#5873 needed fixes to `traceOfTab`'s `timings` and `timestamps` creation, which is a good opportunity to take care of long-standing TODOs to fix the types for those objects. Basically most of their properties may not have been found in the trace but we were pretending they were all there to make dealing with them simpler.

We now throw on `NO_FCP` in so many places that maybe we should just call it required in a trace and throw in `traceOfTab` like we do for no `navigationStart`?

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
+++ b/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
@@ -15,6 +15,7 @@ const BaseNode = require('../../lib/dependency-graph/base-node');
 const ByteEfficiencyAudit = require('./byte-efficiency-audit');
 const UnusedCSS = require('./unused-css-rules');
 const NetworkRequest = require('../../lib/network-request');
+const LHError = require('../../lib/errors');
 
 /** @typedef {import('../../lib/dependency-graph/simulator/simulator')} Simulator */
 /** @typedef {import('../../lib/dependency-graph/base-node.js').Node} Node */
@@ -91,6 +92,9 @@ class RenderBlockingResources extends Audit {
     const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
     // @ts-ignore - TODO(bckenny): allow optional `throttling` settings
     const fcpSimulation = await artifacts.requestFirstContentfulPaint(metricComputationData);
+    if (!traceOfTab.timestamps.firstContentfulPaint) {
+      throw new LHError(LHError.errors.NO_FCP);
+    }
     const fcpTsInMs = traceOfTab.timestamps.firstContentfulPaint / 1000;
 
     const nodesByUrl = getNodesAndTimingByUrl(fcpSimulation.optimisticEstimate.nodeTimings);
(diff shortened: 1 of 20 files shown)


lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -91,6 +92,9 @@ class RenderBlockingResources extends Audit {
     const metricComputationData = {trace, devtoolsLog, simulator, settings: metricSettings};
     // @ts-ignore - TODO(bckenny): allow optional `throttling` settings
     const fcpSimulation = await artifacts.requestFirstContentfulPaint(metricComputationData);
+    if (!traceOfTab.timestamps.firstContentfulPaint) {
```
brendankenny: @patrickhulce this one I was least sure about because it could have been an extra guard vs needed for correct behavior.

This arguably "works" if `fcpTsInMs` ends up as `NaN` as no resources will be ignored. I could change it to make that explicit instead of throwing instead.

lighthouse-core/gather/computed/metrics/interactive.js
```diff
@@ -90,6 +90,9 @@ class Interactive extends MetricArtifact {
    * @return {{cpuQuietPeriod: TimePeriod, networkQuietPeriod: TimePeriod, cpuQuietPeriods: Array<TimePeriod>, networkQuietPeriods: Array<TimePeriod>}}
    */
   static findOverlappingQuietPeriods(longTasks, networkRecords, traceOfTab) {
+    if (!traceOfTab.timestamps.firstContentfulPaint) {
+      throw new LHError(LHError.errors.NO_FCP);
+    }
```
brendankenny: this was already checked down in `computeObservedMetric()`, but this seemed better than passing in the known-defined value as another parameter?

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -133,34 +128,41 @@ class TraceOfTab extends ComputedArtifact {
...
+    /** @type {LH.Artifacts.TraceTimes} */
+    const timings = {
+      navigationStart: 0,
+      firstPaint: getTiming(timestamps.firstPaint),
+      firstContentfulPaint: getTiming(timestamps.firstContentfulPaint),
+      firstMeaningfulPaint: getTiming(timestamps.firstMeaningfulPaint),
+      traceEnd: (timestamps.traceEnd - navigationStart.ts) / 1000,
+      load: getTiming(timestamps.load),
+      domContentLoaded: getTiming(timestamps.domContentLoaded),
+    };
```
brendankenny: happy to code golf on `timestamps`/`timings` creation :)

brendankenny:
> Sprinkling NO_FCP everywhere seems worse than just updating the tests that need a fake FCP event

SG and @paulirish seemed 👍 

Updated.

typings/artifacts.d.ts
```diff
@@ -365,30 +365,43 @@ declare global {
...
         navigationStartEvt: TraceEvent;
-        firstPaintEvt: TraceEvent;
-        firstContentfulPaintEvt: TraceEvent;
-        firstMeaningfulPaintEvt: TraceEvent;
-        loadEvt: TraceEvent;
-        domContentLoadedEvt: TraceEvent;
+        /** The trace event marking firstPaint, if it was found. */
+        firstPaintEvt?: TraceEvent;
+        /** The trace event marking firstContentfulPaint, if it was found. */
+        firstContentfulPaintEvt?: TraceEvent;
```
patrickhulce: this one can be no `?` :)

PR #5893, opened 2018-08-22
core(content-width): not applicable on desktop

patrickhulce:
https://twitter.com/gcmznt/status/1032188854130880512 got me thinking that since the audit is just useless in DevTools without emulation we should mark it as not applicable. The meta tag should still capture this concern on the desktop site, but honestly the whole PWA category should be taken with a grain of salt if you're not auditing with mobile (toplevel warning perhaps? or maybe the justification for category level warning 😄)


lighthouse-core/audits/content-width.js
```diff
@@ -25,13 +25,21 @@ class ContentWidth extends Audit {
...
+   * @param {LH.Audit.Context} context
    * @return {LH.Audit.Product}
    */
-  static audit(artifacts) {
+  static audit(artifacts, context) {
     const viewportWidth = artifacts.ViewportDimensions.innerWidth;
     const windowWidth = artifacts.ViewportDimensions.outerWidth;
     const widthsMatch = viewportWidth === windowWidth;
 
+    if (context.settings.disableDeviceEmulation) {
```
brendankenny: wouldn't this be disabled if you're running on a real phone, though?

patrickhulce: Hm, yes it would. Maybe a `HostUserAgent` plus `disableDeviceEmulation` check? Probably worth rolling up into a generic context `isMobile` boolean

patrickhulce:
Seems like this has been referenced in a few issues as a fix, anyone else wanna take a look? :)

lighthouse-core/audits/content-width.js
paulirish: fwiw this does also not work for the lr-desktop preset. 

but it looks like we're going to use viewport emulation for lr-desktop now, with viewport of `1366x768` and `mobile:false`, to fix a screenshot bug.

So we also have that situation to contend with. :)

paulirish:
discussed a bit in the meeting. 

seems like 

1. we rebase this on top of emulateFormFactor. 
   1. if emulateFormFactor === desktop, then mark as not applicable.
   1. if emulateFormFactor === mobile, continue as planned
   1. if emulateFormFactor is 'none', then we look at the `HostUserAgent` artifact:
      1. if the UA is mobile (contains 'android' (_lol?_) ), continue as planned
      1. if its not mobile, mark as not applicable 

I think this captures the cases we described above. that sound right?

patrickhulce:
> that sound right?

sounds great!

PR #5896, opened 2018-08-22
core(tsc): don't use index signature in LHErrors

brendankenny:
another part of #5873

apologies for the extra line breaks :) it was a lot easier to do it this way, and this file changes so rarely the new (slight) redundancy seems fine

### DIFF
--- a/lighthouse-core/lib/errors.js
+++ b/lighthouse-core/lib/errors.js
@@ -16,8 +16,8 @@ const strings = require('./strings');
 
 class LighthouseError extends Error {
   /**
-   * @param {!LighthouseErrorDefinition} errorDefinition
-   * @param {!Object=} properties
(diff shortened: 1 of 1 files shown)


lighthouse-core/lib/errors.js
```diff
@@ -40,11 +40,11 @@ class LighthouseError extends Error {
   /**
    * @param {string} method
    * @param {{message: string, data?: string|undefined}} protocolError
-   * @return {!Error|LighthouseError}
+   * @return {Error|LighthouseError}
    */
   static fromProtocolMessage(method, protocolError) {
     // extract all errors with a regex pattern to match against.
-    const protocolErrors = Object.keys(ERRORS).filter(k => ERRORS[k].pattern).map(k => ERRORS[k]);
```
patrickhulce: so glad we can use `Object.values` now :)

Issue #5870, opened 2018-08-20
convert page-functions to be exported as strings


patrickhulce:
**Summary**

In https://github.com/GoogleChrome/lighthouse/pull/5857#discussion_r211050626 @paulirish mentioned he liked the idea of page-functions exported as strings instead of functions that are `.toString`'d

danushan1246:
https://github.com/GoogleChrome/lighthouse/pull/5902

Abdulmanahtalib:
Support

patrickhulce:
Fixed by #5902

PR #5902, opened 2018-08-24
core: export any page-functions as string

midzer:
**Summary**

Export all functions in lib page-functions directly as string

Minor coding style i would say

Prettier code due not cluttering implementation files with multiple toString() calls

**Related Issues/PRs**
https://github.com/GoogleChrome/lighthouse/issues/5870

### DIFF
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -145,10 +145,10 @@ function ultradumbBenchmark() {
 }
 
 module.exports = {
-  wrapRuntimeEvalErrorInBrowser,
-  registerPerformanceObserverInPage,
-  checkTimeSinceLastLongTask,
-  getElementsInDocument,
-  getOuterHTMLSnippet,
-  ultradumbBenchmark,
+  wrapRuntimeEvalErrorInBrowser: wrapRuntimeEvalErrorInBrowser.toString(),
+  registerPerformanceObserverInPage: registerPerformanceObserverInPage.toString(),
+  checkTimeSinceLastLongTask: checkTimeSinceLastLongTask.toString(),
+  getElementsInDocument: getElementsInDocument.toString(),
+  getOuterHTMLSnippet: getOuterHTMLSnippet.toString(),
+  ultradumbBenchmark: ultradumbBenchmark.toString(),
 };
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -151,4 +151,5 @@ module.exports = {
   getElementsInDocumentString: getElementsInDocument.toString(),
   getOuterHTMLSnippetString: getOuterHTMLSnippet.toString(),
   ultradumbBenchmark: ultradumbBenchmark,
+  ultradumbBenchmarkString: ultradumbBenchmark.toString(),
 };
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -145,10 +145,10 @@ function ultradumbBenchmark() {
 }
 
 module.exports = {
-  wrapRuntimeEvalErrorInBrowser: wrapRuntimeEvalErrorInBrowser.toString(),
-  registerPerformanceObserverInPage: registerPerformanceObserverInPage.toString(),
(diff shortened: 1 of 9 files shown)


patrickhulce:
thanks @midzer! @paulirish was specifically requesting this too :)

what do we think about adding a suffix to make it clear they are strings? I suppose the danger is less since toString'ing the string won't do much harm 🤷‍♀️ 

- `ultradumbBenchmarkFnString`
- `ultradumbBenchmarkAsString`
- `ultradumbBenchmarkStringified`

midzer:
@patrickhulce thanks for your suggestions. I'm totally up for this change. While looking a bit through LH code base, I've realized there are a couple of situations where variable type is put into it's name like
```
const errorString = 'Unable to locate '
```
for example. For me this resembles to https://en.wikipedia.org/wiki/Hungarian_notation with just a bit different flavour. Dunno whether there exists an exact term for this notation.

In rare cases something like `ultradumbBenchmarkAsString` is used already.

So I'm unsure which way to go to keep code base consistent for this kind of topic.

lighthouse-core/lib/page-functions.js
```diff
@@ -145,10 +145,10 @@ function ultradumbBenchmark() {
...
-  checkTimeSinceLastLongTask,
-  getElementsInDocument,
-  getOuterHTMLSnippet,
-  ultradumbBenchmark,
+  wrapRuntimeEvalErrorInBrowser: wrapRuntimeEvalErrorInBrowser.toString(),
+  registerPerformanceObserverInPage: registerPerformanceObserverInPage.toString(),
+  checkTimeSinceLastLongTask: checkTimeSinceLastLongTask.toString(),
+  getElementsInDocument: getElementsInDocument.toString(),
+  getOuterHTMLSnippet: getOuterHTMLSnippet.toString(),
+  ultradumbBenchmark: ultradumbBenchmark.toString(),
```
patrickhulce: one wrinkle here we actually use this function as a function in `scripts/benchmark.js`

patrickhulce:
> So I'm unsure which way to go to keep code base consistent for this kind of topic.

It's not super consistent throughout the codebase, so I wouldn't put it on yourself to make it so ;)

In the file that a few of these functions were [just moved from](https://github.com/GoogleChrome/lighthouse/pull/5857/files#diff-75d372091df3248365603f67a3806517L45) they were labeled `FnString`, so we can go with that for now?

lighthouse-core/lib/page-functions.js
```diff
@@ -145,10 +145,10 @@ function ultradumbBenchmark() {
...
-  checkTimeSinceLastLongTask,
-  getElementsInDocument,
-  getOuterHTMLSnippet,
-  ultradumbBenchmark,
+  wrapRuntimeEvalErrorInBrowserString: wrapRuntimeEvalErrorInBrowser.toString(),
+  registerPerformanceObserverInPageString: registerPerformanceObserverInPage.toString(),
+  checkTimeSinceLastLongTaskString: checkTimeSinceLastLongTask.toString(),
+  getElementsInDocumentString: getElementsInDocument.toString(),
+  getOuterHTMLSnippetString: getOuterHTMLSnippet.toString(),
+  ultradumbBenchmark: ultradumbBenchmark,
```
paulirish: this one is still toString()d within driver. can we update this one as well?

patrickhulce: I had him change it back because we use the raw function in the scripts

brendankenny: why not both?

brendankenny: .gif

patrickhulce: both

midzer: Doesn't this break https://github.com/GoogleChrome/lighthouse/blob/24a855d325617c3cf330929716c3270dbe165430/lighthouse-core/scripts/benchmark.js#L18 ?

paulirish: Nah. It's still exposed the same way. It's just ALSO being exposed as a string.

patrick didnt like my idea of only exposing as string and `eval`ing in the script/benchmark.js. :) haha

Abdulmanahtalib:
Thankyou,support

PR #5905, opened 2018-08-24
misc(logger): add time/timeEnd methods

patrickhulce:
In order to land #3745, we need logger to be published with the new functionality. This is just the `lighthouse-logger` portion of #3745 with a version bump to `1.1.0`

### DIFF
--- a/lighthouse-logger/index.js
+++ b/lighthouse-logger/index.js
@@ -219,7 +219,7 @@ class Log {
 }
 
 Log.events = new Emitter();
-Log.getTimeEntries = _ => {
+Log.takeTimeEntries = _ => {
   const entries = marky.getEntries();
   marky.clear();
   return entries;
--- a/lighthouse-logger/index.js
+++ b/lighthouse-logger/index.js
@@ -113,7 +113,7 @@ class Log {
 
   static timeEnd({msg, id, args=[]}, level='verbose') {
     Log[level]('statusEnd', msg, ...args);
-    return marky.stop(id);
+    marky.stop(id);
   }
 
   static log(title, ...args) {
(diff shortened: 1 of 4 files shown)


lighthouse-logger/index.js
```diff
@@ -6,6 +6,8 @@
 'use strict';
 
 const debug = require('debug');
+const marky = require('marky');
```
wardpeet: 👌

lighthouse-logger/index.js
```diff
@@ -104,6 +106,16 @@ class Log {
...
   }
 
+  static time({msg, id, args=[]}, level='log') {
+    marky.mark(id);
+    Log[level]('status', msg, ...args);
+  }
+
+  static timeEnd({msg, id, args=[]}, level='verbose') {
+    Log[level]('statusEnd', msg, ...args);
+    return marky.stop(id);
```
brendankenny: is this going to be used? Should update `lighthouse-logger/index.d.ts` or drop the `return` here?

lighthouse-logger/index.js
```diff
@@ -207,5 +219,10 @@ class Log {
 }
 
 Log.events = new Emitter();
+Log.getEntries = _ => {
```
brendankenny: I realize this (appears) to be going for the same `getEntries()` name as the `Performance.getEntries()`, but it's definitely confusing in the context of a logger (where you might expect it to return all log entries).

WDYT about `Log.getTimeEntries()`?

wardpeet: maybe make this a getter instead of a function. No big deal though

patrickhulce: I had the same feeling as well, will do 👍

lighthouse-logger/index.js
patrickhulce: done

lighthouse-logger/index.js
patrickhulce: settled on takeTimeEntries to mimic the semantics and name of PerfObserver 👍

PR #675, opened 2016-09-20
handle computed artifact caching automatically

brendankenny:
@paulirish 

noticed a small bug in the computed artifact caching in `PushedRequests` and thought I'd just make caching taken care of automatically while I was in there.

When implementing a computed artifact in the future you just override `compute_()` and return either a Promise or the computed artifact directly and the `ComputedArtifact` base class will handle caching the response (and subsequently responding from the cache).

PR #5907, opened 2018-08-24
core: initial refactor of computedArtifact import/caching

brendankenny:
Proof of concept of moving computed artifacts off of the `artifacts` object passed into audits and just use regular `require('./path/to/computed-artifact.js')` to use them.

Starts with a really simple computed artifact, `manifest-values.js`, which doesn't need to request any computed artifacts itself, nor does any computed artifact request it, but it is non-trivial and required in four audits. The computed artifact cache is now part of `LH.Audit.Context`, and audits need to pass in the cache rather than it being created and bound behind the scenes.

Nice thing about this approach is non-converted computed artifacts can continue to function as-is and we can convert them one by one (or in bunches) rather than in one massive, unreadable PR.

There are a few things we lose without the automatically created methods, but I think we'll have a huge step up in being able to analyze dependencies, trace logic through, etc. The fact that this was so difficult to untangle just for a super-simple computedArtifact shows the downsides of the earlier magic experience. Using computed artifacts (or moving off them) in an audit is now as easy as pulling in any  other library module.

This was part of #5008 under "Nice to have". I swear we had a discussion about this in a set of comments somewhere, but the earliest reference I could find was https://github.com/GoogleChrome/lighthouse/pull/4961#pullrequestreview-111849395, so maybe it was in person.

### DIFF
--- a/lighthouse-core/audits/splash-screen.js
+++ b/lighthouse-core/audits/splash-screen.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const MultiCheckAudit = require('./multi-check-audit');
+const ManifestValues = require('../gather/computed/manifest-values');
 
 /**
  * @fileoverview
@@ -64,20 +65,20 @@ class SplashScreen extends MultiCheckAudit {
 
   /**
    * @param {LH.Artifacts} artifacts
+   * @param {LH.Audit.Context} context
    * @return {Promise<{failures: Array<string>, manifestValues: LH.Artifacts.ManifestValues}>}
    */
-  static audit_(artifacts) {
+  static async audit_(artifacts, context) {
     /** @type {Array<string>} */
     const failures = [];
 
-    return artifacts.requestManifestValues(artifacts.Manifest).then(manifestValues => {
-      SplashScreen.assessManifest(manifestValues, failures);
+    const manifestValues = await new ManifestValues().request(artifacts.Manifest, context);
+    SplashScreen.assessManifest(manifestValues, failures);
 
-      return {
-        failures,
-        manifestValues,
-      };
-    });
+    return {
+      failures,
+      manifestValues,
+    };
   }
 }
 
--- a/lighthouse-core/audits/splash-screen.js
+++ b/lighthouse-core/audits/splash-screen.js
@@ -72,7 +72,8 @@ class SplashScreen extends MultiCheckAudit {
     /** @type {Array<string>} */
     const failures = [];
 
-    const manifestValues = await new ManifestValues().request(artifacts.Manifest, context);
+    const manifestValues = await new ManifestValues().request(context.computedCaches,
+        artifacts.Manifest);
     SplashScreen.assessManifest(manifestValues, failures);
 
     return {
--- a/lighthouse-core/audits/splash-screen.js
+++ b/lighthouse-core/audits/splash-screen.js
@@ -72,8 +72,7 @@ class SplashScreen extends MultiCheckAudit {
     /** @type {Array<string>} */
     const failures = [];
 
-    const manifestValues = await new ManifestValues().request(context.computedCaches,
-        artifacts.Manifest);
+    const manifestValues = await ManifestValues.request(context.computedCache, artifacts.Manifest);
     SplashScreen.assessManifest(manifestValues, failures);
 
     return {
--- a/lighthouse-core/audits/splash-screen.js
+++ b/lighthouse-core/audits/splash-screen.js
@@ -72,7 +72,7 @@ class SplashScreen extends MultiCheckAudit {
     /** @type {Array<string>} */
     const failures = [];
 
-    const manifestValues = await ManifestValues.request(context.computedCache, artifacts.Manifest);
+    const manifestValues = await ManifestValues.request(context, artifacts.Manifest);
     SplashScreen.assessManifest(manifestValues, failures);
 
     return {
--- a/lighthouse-core/gather/computed/manifest-values.js
+++ b/lighthouse-core/gather/computed/manifest-values.js
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-const ComputedArtifact = require('./computed-artifact');
+const ComputedArtifact = require('./new-computed-artifact');
 const icons = require('../../lib/icons');
 
 const PWA_DISPLAY_VALUES = ['minimal-ui', 'fullscreen', 'standalone'];
--- a/lighthouse-core/gather/computed/manifest-values.js
+++ b/lighthouse-core/gather/computed/manifest-values.js
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-const ComputedArtifact = require('./new-computed-artifact');
+const makeComputedArtifact = require('./new-computed-artifact');
 const icons = require('../../lib/icons');
 
 const PWA_DISPLAY_VALUES = ['minimal-ui', 'fullscreen', 'standalone'];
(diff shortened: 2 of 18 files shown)


brendankenny:
whoops, wanted it to be the cache passed in, not the whole context. Just sec :)

brendankenny:
Done.

Only ergonomic downside is that, since these are essentially single function calls with no state (that isn't cacheable and so needs to be passed in), they really cry out to be static. However, typescript doesn't support polymorphic `this` on static methods, so we can't have both automatic caching *and* type checking (and completion) based on whatever the computed artifact class needs as input/gives as output.

We really want type checking on those, but I think we also want automatic caching. Repeating the boilerplate everywhere would be dumb, and it's really easy to get slightly wrong and not notice it's recomputing some expensive calculation (a bug that did this was the original impetus for automatic caching - #675).

So, as a result, you have to do `const result = new ComputedArtifact().request(cache, input);`. Not ideal, but not that bad, and if typescript ever fixes https://github.com/Microsoft/TypeScript/issues/5863 and these are still around, some future Lighthouse person can do a quick fix for them all :)

patrickhulce:
@brendankenny WDYT about something like [this](http://www.typescriptlang.org/play/#src=export%20function%20makeCacheable%3CI%2C%20R%3E(f%3A%20(input%3A%20I)%20%3D%3E%20Promise%3CR%3E%2C%20name%3A%20string)%3A%20(input%3A%20I%2C%20cache%3A%20Map%3Cany%2C%20any%3E)%20%3D%3E%20Promise%3CR%3E%20%7B%0D%0A%20%20%20%20return%20(input%3A%20I%2C%20cache%3A%20Map%3Cany%2C%20any%3E)%20%3D%3E%20%7B%0D%0A%20%20%20%20%20%20%20%20if%20(cache.has(%7B%20input%2C%20name%20%7D))%20%7B%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20return%20cache.get(%7Binput%2C%20name%7D)%0D%0A%20%20%20%20%20%20%20%20%7D%0D%0A%0D%0A%20%20%20%20%20%20%20%20const%20value%20%3D%20f(input)%0D%0A%20%20%20%20%20%20%20%20cache.set(%7Binput%2C%20name%7D%2C%20value)%0D%0A%20%20%20%20%20%20%20%20return%20value%0D%0A%20%20%20%20%7D%0D%0A%7D) instead?

brendankenny:
> WDYT about something like this instead?

well I spent the last day learning everything about the limits of polymorphic `this` in typescript to make my way work, but I think I like this one better :)

I'll try it out.

brendankenny:
@paulirish how do you feel about this format

lighthouse-core/gather/computed/manifest-values.js
```diff
@@ -14,11 +14,7 @@ const PWA_DISPLAY_VALUES = ['minimal-ui', 'fullscreen', 'standalone'];
 // For more discussion, see https://github.com/GoogleChrome/lighthouse/issues/69 and https://developer.chrome.com/apps/manifest/name#short_name
 const SUGGESTED_SHORTNAME_LENGTH = 12;
 
-class ManifestValues extends ComputedArtifact {
-  get name() {
-    return 'ManifestValues';
-  }
-
+class ManifestValues {
```
patrickhulce: I believe someone once told me that he thought it was silly we made classes with all static members instead of plain objects ;)

FWIW I kinda like them as classes for the flashy `require` name consistency, but I was mostly persuaded back then for plain objects, thoughts?

brendankenny: > I believe someone once told me that he thought it was silly we made classes with all static members instead of plain objects ;)

haha, I didn't even think about it as I changed things. There used to be some tsc benefit due to polymorphic `this` (but not an issue with this implementation) and then from augmenting things in javascript passed out on `module.exports` (but that's all fixed in 3.1).

I'll admit to stockholm syndrome in that I don't see it as ugly and unnecessary any more, but I also don't have any problem with a plain object. The main thing might be the `name` property, which I think we still want, whether explicit or from just the `StaticClass.name` default from declaring a class.

brendankenny:
> Just thinking out loud here, what are the primary benefits of attaching the cache to context? If each computed artifact just kept its cache in its request method closure, would that be enough?

I don't think we can do that because of GC. It would be node-module-only (I believe all the other clients destroy the world after every run), but since we use artifacts as the keys, artifacts will keep accumulating from each run.

Good use for WeakMap, but we don't always have unique parent objects for sets of artifacts, so, I don't think that will work. I can't think of another approach around that.

FWIW, I don't think the churn will end up being too bad outside of tests. Basically add a `context` param to each audit and pass it to each computed artifact request. We also don't have that many computed artifacts when it comes down to it.

Tests need more changes since we make so many calls, but I could reduce the noise there (e.g. just pass in `{computedCache: new Map()}` rather than doing a `getMockContext()` call).

lighthouse-core/gather/computed/manifest-values.js
patrickhulce: true true, this WFM :)

patrickhulce:
> I don't think we can do that because of GC. It would be node-module-only (I believe all the other clients destroy the world after every run), but since we use artifacts as the keys, artifacts will keep accumulating from each run.

Yeah the clearing mechanism would be ugly at best, if you foresee little pain outside tests, then this makes sense to me 👍

lighthouse-core/runner.js
```diff
@@ -269,8 +275,7 @@ class Runner {
       const auditOptions = Object.assign({}, audit.defaultOptions, auditDefn.options);
       const auditContext = {
         options: auditOptions,
-        settings,
-        LighthouseRunWarnings: runWarnings,
+        ...auditsContext,
```
paulirish: dont love these two diff objs with such similar names
how about `sharedAuditsContext` for this one?

lighthouse-core/runner.js
```diff
@@ -375,6 +380,10 @@ class Runner {
       'metrics', // the sub folder that contains metric names
       'metrics/lantern-metric.js', // lantern metric base class
       'metrics/metric.js', // computed metric base class
+
+      // Computed artifacts switching to the new system.
+      'new-computed-artifact.js',
+      'manifest-values.js',
```
paulirish: This fn is for the gulpfile. wont browserify already discover this manifest-values module without being added manually?

patrickhulce: the diff obscures it, but this is the list of filenamesToSkip :)

lighthouse-core/runner.js
brendankenny: sure

brendankenny: done

lighthouse-core/audits/splash-screen.js
```diff
@@ -64,20 +65,20 @@ class SplashScreen extends MultiCheckAudit {
...
    * @return {Promise<{failures: Array<string>, manifestValues: LH.Artifacts.ManifestValues}>}
    */
-  static audit_(artifacts) {
+  static async audit_(artifacts, context) {
     /** @type {Array<string>} */
     const failures = [];
 
-    return artifacts.requestManifestValues(artifacts.Manifest).then(manifestValues => {
-      SplashScreen.assessManifest(manifestValues, failures);
+    const manifestValues = await ManifestValues.request(context, artifacts.Manifest);
```
brendankenny: I'll refrain from going promises->async/await in future changes to cut down the diffs as well :)

Issue #5640, opened 2018-07-09
Replace details tag in html report viewer to support jumping to a specific audit by id anchor


nmchau:
## Feature request

**Summary**

In html report viewer, each audit component has id element but many of them are inside the details tags. So, it's impossible to jump to a specific audit component by its id without expanding the parent details tags. 
For example, report.html#mobile-friendly link can't allow us to jump to 'mobile friendly' audit because this audit stays inside a details tag.

**What is the motivation or use case for changing this?**
Some people are just interested in some specific audits among many audits supported by Lighthouse. 

**How is this beneficial to Lighthouse?**
Links which allow jumping directly to different audits in report viewer can be bookmarked or shared easily.

patrickhulce:
Neat idea! We might want to consider alternatives to support this, perhaps instead we could auto-open the parent with JS on navigation or something.

paulirish:
solution: if targeted audit is within a collapsed details (or multiple of them), we open those with JS. maybe we also highlight background via `:target` CSS.

brendankenny:
let's track in #6668

PR #5915, opened 2018-08-27
report: enable jump to audit in nested details

### LINKED ISSUE #6668: HTML Report: support deep linking
For web.dev, we have the request for users to be able to deep link into sections of the HTML report. Each section, currently has an `id=AUDIT_NAME`, but loading the report does not always scroll properly to that section during page load or the section is hidden under by the top fixed banner. Also need to expand the details element when users are deep linking into a section.

Example: 
https://lighthouse-dot-webdotdevsite.appspot.com/lh/html?url=https://ericbidelman.com#errors-in-console

Items to handle:
- [x] include ids on all audit line items
- [ ] detect anchor target on page load and delay (smooth) scroll to section. Include offset so section isn't hidden under top nav.
- [ ] auto-expand the audit's details dropdown so users can drill right in. 

b/119882633

midzer:
Hi,

**Summary**

In `report.html` jump to respective audit on page load if hash is provided in URL. Collapsed content is opened automatically. Element is highlighted via CSS selector

At first i wanted to use https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView for scrolling to requested item, but somehow this didn't work. Any feedback is welcome.

**Related Issues/PRs**

Fixes https://github.com/GoogleChrome/lighthouse/issues/5640
Fixes #6668


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,27 @@ class ReportUIFeatures {
...
   }
 
+  /**
+   * Jumps to respective audit on page load if hash is provided in URL
+   * Collapsed content is opened automatically
+   * Element is highlighted via CSS selector
+   */
+  _jumpToAudit() {
+    const hash = window.location.hash;
+    if (hash) {
```
patrickhulce: nit: WDYT about reversing these conditions to avoid nesting?

```js

if (!hash) return;

const audit = ...
if (!audit) return;

...

File: lighthouse-core/report/html/renderer/report-ui-features.js [hunk 2]
```diff
@@ -286,6 +287,27 @@ class ReportUIFeatures {
...
+      const audit = document.querySelector(hash);
+      if (audit) {
```
brendankenny: probably want to also at least check that this has class `lh-audit` before we mess with it

lighthouse-core/report/html/report-styles.css
```diff
@@ -213,7 +213,8 @@
   word-break: break-word;
   color: hsl(174, 100%, 27%);
 }
-.lh-node:hover {
+.lh-node:hover,
+div:target {
```
patrickhulce: seems a bit weird to share the style with HTML nodes, let's split this out into its own style with a comment about the scroll to behavior :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,27 @@ class ReportUIFeatures {
...
+        if (audit.firstElementChild) {
```
brendankenny: should probably check that this is a `<details>` (or should we just open any `details` children)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,27 @@ class ReportUIFeatures {
...
+          audit.firstElementChild.setAttribute('open', 'true');
+        }
+        if (audit.parentElement) {
+          audit.parentElement.setAttribute('open', 'true');
```
patrickhulce: in cases like accessibility we'll need to go up a couple levels :/ 

maybe replace this with a quick loop? untested example below :)
 
```js
for (let node = audit.parentElement; node; node = node.parentElement) {
  if (node.tagName === 'DETAILS') node.setAttribtue('open', 'true');
}

File: lighthouse-core/report/html/report-styles.css (hunk 3 above)
Comment (brendankenny) [inline lighthouse-core/report/html/report-styles.css]: make more specific to `.lh-audit`?

File: lighthouse-core/report/html/renderer/report-ui-features.js (hunk 5 above)
Comment (brendankenny) [reply lighthouse-core/report/html/renderer/report-ui-features.js]: we probably also don't want to go anywhere above `lh-container` either. The Lighthouse report might end up embedded in someone else's `<details>` :)

File: lighthouse-core/report/html/renderer/report-ui-features.js [hunk 6]
```diff
@@ -286,6 +287,27 @@ class ReportUIFeatures {
...
+          scroll({top: audit.parentElement.offsetTop, behavior: 'smooth'});
```
paulirish: can you call it as `window.scroll(` to be clear its the global method. I'm not used to seeing it so nude. :)

lighthouse-core/report/html/renderer/report-ui-features.js
paulirish: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest

lighthouse-core/report/html/report-styles.css
midzer: Sounds reasonable to me. I just didn't find a place where to put it at first *g*

lighthouse-core/report/html/renderer/report-ui-features.js
midzer: Yeah, `closest()` worked fine in my last project

lighthouse-core/report/html/renderer/report-ui-features.js
midzer: Good one. Heavy nesting is ugly indeed^^

midzer:
`closest()` returns an `Element`, so I can't use offsetTop any more. I replaced it by `details.getBoundingClientRect().top + window.scrollY`

Thanks for your feedback. Cya later

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,33 @@ class ReportUIFeatures {
...
+   * Jumps to respective audit on page load if hash is provided in URL
+   * Collapsed content is opened automatically
+   * Element is highlighted via CSS selector
+   */
+  _jumpToAudit() {
+    const hash = window.location.hash;
+    if (!hash) {
+      return;
+    }
+    const audit = document.querySelector(hash);
```
brendankenny: `this._document`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,33 @@ class ReportUIFeatures {
     this._dom.resetTemplates();
   }
 
+  /**
+   * Jumps to respective audit on page load if hash is provided in URL
```
brendankenny: nit: can you add periods to the end of these lines since there are multiple lines (we don't enforce this, just my personal preference :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,33 @@ class ReportUIFeatures {
...
+      return;
+    }
+    const audit = document.querySelector(hash);
+    if (!audit || !audit.classList.contains('lh-audit')) {
+      return;
+    }
+    if (audit.firstElementChild && audit.firstElementChild.tagName === 'DETAILS') {
+      audit.firstElementChild.setAttribute('open', 'true');
+    }
+    const details = audit.closest('details')
```
brendankenny: lint: need `;`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -286,6 +287,33 @@ class ReportUIFeatures {
...
+    if (details) {
+      details.setAttribute('open', 'true');
+      window.scroll({
+        top: details.getBoundingClientRect().top + window.scrollY,
+        behavior: 'smooth'
```
brendankenny: lint: need `,`

brendankenny:
> closest() returns an Element, so I can't use offsetTop any more. I replaced it by details.getBoundingClientRect().top + window.scrollY

It returns an Element, but we know more specifically that it's an HTMLElement since we asked for a `'details'`, so there's no harm in treating it as such (in fact, the typescript checker even narrows it down to `HTMLDetailsElement | null` for us, and the conditional rules out the `null`

midzer:
First of all, many thanks for your insight testing and review @brendankenny . I tried to include proposed suggestions in latest commits. Somehow, tests are broken now though. Can't figure out right now whats going havoc there^^

Regarding scrolling to element I've played around with `element.scrollIntoView()` again which does the job as current `window.scroll()` implementation. Too bad, it does not work in all cases yet: for example `#meta-description` element is hidden below `.lh-scores.header` after scroll took place. Manual additions/subtractions to `offsetTop` seem not to have any effect. So I'm kinda stuck on further progress right now.

midzer:
Just to bump this one to the top. I have time to help bring it to merge.

Is this PR out of focus now?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +285,49 @@ class ReportUIFeatures {
     this._dom.resetTemplates();
   }
 
+  /* eslint-env browser */
```
brendankenny: not needed

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +285,49 @@ class ReportUIFeatures {
...
+
+  /* eslint-env browser*/
+  /**
+   * Open all closest <details> elements recursively.
+   * If found, scroll to element in viewport.
+   * @param {Element} element
+   */
+  _openClosestDetails(element) {
+    const details = element.closest('.lh-container details');
+    if (details instanceof HTMLElement) {
```
brendankenny: test `instanceof HTMLDetailsElement` and then the line below this can be the nicer (and type checked) `details.open = true;`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +285,49 @@ class ReportUIFeatures {
...
+    if (!audit || !audit.classList.contains('lh-audit')) {
+      return;
+    }
+    if (audit.firstElementChild && audit.firstElementChild.tagName === 'DETAILS') {
+      audit.firstElementChild.setAttribute('open', 'true');
+    }
+    this._openClosestDetails(audit);
+  }
+
+  /* eslint-env browser*/
```
brendankenny: Let's keep this simpler than adding the whole env and remove this but add `HTMLDetailsElement` to the eslint `/* globals */` block at the top of the file

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +285,49 @@ class ReportUIFeatures {
...
+   * @param {Element} element
+   */
+  _openClosestDetails(element) {
+    const details = element.closest('.lh-container details');
+    if (details instanceof HTMLElement) {
+      details.setAttribute('open', 'true');
+      if (details.parentElement) {
+        this._openClosestDetails(details.parentElement);
+      }
+      window.scroll({
```
brendankenny: should `window.scroll` actually be in the end of `_jumpToAudit` after `_openClosestDetails` returns (and scrolling to `audit` there)? Seems like we want to scroll to the actual audit, not the closest details element containing it?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +285,49 @@ class ReportUIFeatures {
...
+        top: details.offsetTop,
+        behavior: 'smooth',
+      });
+    } else {
+      throw new Error(`Closest ancestor '.lh-container details' of ${element.nodeName} not found`);
```
brendankenny: Since this walks up recursively from the audit until it runs out of `<details>` parents, this will always eventually throw.

But I think we just don't need this at all...some audits may just not be nested in a `<details>` element (like the metrics (e.g. `#first-contentful-paint`)). No harm if it can't find any more closest `details`.

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
```diff
@@ -70,6 +70,8 @@ describe('ReportUIFeatures', () => {
       };
     };
 
+    global.window.location = 'http://localhost';
```
brendankenny: Is this necessary? Things got a little confusing in here with recent updates.

@wardpeet Do you know which `window` we should be using? Especially when looking at how this will merge into the updated version on master https://github.com/GoogleChrome/lighthouse/blob/33ba60cc2baa8704cb9dab92dab88407c25fdcb7/lighthouse-core/test/report/html/renderer/report-ui-features-test.js#L55-L73

Should we be creating our own `window` anymore or just using `document.window` (or `documentReport.window`, both of which provide a `location` for us)?

lighthouse-core/report/html/renderer/report-ui-features.js
midzer: k. Got it removed.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +283,47 @@ class ReportUIFeatures {
...
+    if (!audit || !audit.classList.contains('lh-audit')) {
+      return;
+    }
+    if (audit.firstElementChild && audit.firstElementChild.tagName === 'DETAILS') {
+      audit.firstElementChild.setAttribute('open', 'true');
+    }
+    this._openClosestDetails(audit);
+    if (audit instanceof HTMLElement) {
+      window.scroll({
+        top: audit.offsetTop,
```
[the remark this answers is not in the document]
midzer: Thanks for your feedback @ebidel . I've just pushed a new commit here.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +283,47 @@ class ReportUIFeatures {
...
+    const hash = window.location.hash;
+    if (!hash) {
+      return;
+    }
+    const audit = this._document.querySelector(hash);
+    if (!audit || !audit.classList.contains('lh-audit')) {
+      return;
+    }
+    if (audit.firstElementChild && audit.firstElementChild.tagName === 'DETAILS') {
+      audit.firstElementChild.setAttribute('open', 'true');
```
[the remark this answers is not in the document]
midzer: Done. Now check for `tagName` property should be unnecessary, too

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -282,6 +283,48 @@ class ReportUIFeatures {
...
+    }
+    const audit = this._document.querySelector(hash);
+    if (!audit || !audit.classList.contains('lh-audit')) {
+      return;
+    }
+    if (audit.firstElementChild instanceof HTMLDetailsElement) {
+      audit.firstElementChild.open = true;
+    }
+    this._openClosestDetails(audit);
+    if (audit instanceof HTMLElement) {
```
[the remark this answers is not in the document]
midzer: Without this check, typescript throws `Property 'offsetTop' does not exist on type 'Element'.` in my IDE. I could not find another way around this issue

lighthouse-core/report/html/report-styles.css
```diff
@@ -337,6 +337,10 @@
   display: none;
 }
 
+.lh-audit:target {
```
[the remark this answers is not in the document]
midzer: This has been requested by @paulirish here https://github.com/GoogleChrome/lighthouse/issues/5640#issuecomment-416329468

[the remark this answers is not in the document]
patrickhulce: Oh! OK :)

@paulirish I'm not convinced we need this. It's a bit odd given that the only thing that will make it go away is clicking another category, not just another audit.

lighthouse-core/report/html/renderer/report-ui-features.js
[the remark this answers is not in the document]
patrickhulce: ah ok just for ts then?

let's flip it around then and add a comment to explain


```js
// This should always be an HTMLElement by this point by ts doesn't know that
if (!(audit instanceof HTMLElement)) return;
```

unless @brendankenny has better ideas?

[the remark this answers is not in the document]
brendankenny: > ah ok just for ts then?

yeah, it's a tsc thing (technically it's correct, but we know if this particular `document.querySelector` response isn't null it's an `HTMLElement`).

We can either do your suggestion or do a cast above.
```js
/** @type {?HTMLElement} */
const audit = this._document.querySelector(hash);
if (!audit || !audit.classList.contains('lh-audit')) {
  return;
}
// ...
```

either one sounds good to me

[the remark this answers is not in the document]
midzer: I'm fairly new to TypeScript, so I did not know how to do the cast. Thanks for your suggestion @brendankenny . *pushed.to.branch*

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
brendankenny: > global.window.location = 'http://localhost';

I think we still need to fix this, but it's a bigger fix with `window` in the jsdom tests, so can wait for a future PR.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -285,6 +286,50 @@ class ReportUIFeatures {
...
+    }
+    // This should always be an HTMLElement by this point by ts doesn't know that
+    if (!(audit instanceof HTMLElement)) {
+      return;
+    }
+    if (audit.firstElementChild instanceof HTMLDetailsElement) {
+      audit.firstElementChild.open = true;
+    }
+    this._openClosestDetails(audit);
+    const headerContainer = this._dom.find('lh-header-container', this._document);
```
brendankenny: `'.lh-header-container'`

midzer: Ooooops :) thanks

paulirish:
This scrolling is pretty flaky and I tried experimenting with different combos of rAF/setTimeout/etc and thought I got something good: https://github.com/midzer/lighthouse/pull/1

And then I put my test page online and its no longer scrolling 100% reliably. https://envious-cave.surge.sh/cnn.html#js-libraries (try with reload and also with pasting into a new tab)

I dont think we can guarantee an animated scroll (until some browser bug is fixed?), but we can at least guarantee the page is scrolled to the right spot.

I do like setting focus on the `<summary>` instead of using `:target` though. Seems like a nice solution to the CSS discussion above: https://github.com/midzer/lighthouse/blob/7ce0f63cff7ccd3676d3c6c582b3ab790f7849a7/lighthouse-core/report/html/renderer/report-ui-features.js#L324-L327

brendankenny:
Thanks for all your work @midzer.

I'm going to close for now since the scrolling logic [never entirely panned out](https://github.com/GoogleChrome/lighthouse/pull/5915#issuecomment-452929937) here due to how long it takes the report to construct the entire DOM after load. We're gearing up for a report UI refresh as we close in on May, and I'm hoping we can restructure report rendering and the DOM a little bit to make scrolling to individual audits more feasible. Hopefully then we can revisit this PR and make it work!

PR #5935, opened 2018-08-29
tests: fix getFlags test

paulirish:
started looking into this because of #4395

found that jest was silently exiting on this:

```sh
./node_modules/jest/bin/jest.js lighthouse-cli/test/cli/run-test.js
```

CLI options MUST follow the url because of our use of arrays. you can't put the URL at the end or middle

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
   include:
     - node_js: "8"
     - node_js: "9"
-      if: head_branch IS blank AND branch = master
+    - node_js: "10"
 dist: trusty
 cache:
   yarn: true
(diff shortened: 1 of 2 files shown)


paulirish:
ok. chillin until i have time to reset the travis caches and deal with the fallout.

wardpeet:
@brendankenny node version issue? Could you explain this some more? 😛

brendankenny:
it always fails on @paulirish and my machines, which were running node 10, and passes on @patrickhulce and travis's machines, which were running node 8. So it was a theory, but I believe Paul determined that was a red herring.

wardpeet:
master is failing silently on 8.11.3 & 10.9.0
this one passes on 8.11.3 & 10.9.0

brendankenny:
this is suddenly working for me again :mystery:

Maybe we don't need this after all

brendankenny:
Closing this. I'm pretty sure now we have it correct.

With some experience now with some other zombie deps now, the issue we were seeing may have been an old `lighthouse-cli/node_modules` sticking around after the cli `package.json` was removed.

PR #5936, opened 2018-08-30
cli(run): prevent file:// from running, return error

### LINKED ISSUE #5924: Don't allow running LH against file:// URLs
We should block these entirely

justinribeiro:
Per #5924, this prevents `file://` urls from running in LH. 
image

A few things:

1. The regex doesn't actually look for `file`, rather it checks to verify if the URL is a valid form. To see what the regex matches, I put up https://regexr.com/3um70 for easy glance (which is pretty much just an extended version of Mathias' list from https://mathiasbynens.be/demo/url-regex)
2. The regex does make an exception for `chrome://` urls
3. I placed this in `run` as opposed to running checks in `cli-flags` mostly because it seemed the cleaner route. Re-uses `handleError` in this case (though the error message placement probably needs to move) 
4. I added a specific test that utilizes `jest.spyOn` and a mock for the `process.exit()` call that `handleError` makes when failing on `file:///` (though the console.error will still appear when running tests):
image

Resolves #5924.

### DIFF
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -95,14 +95,6 @@ function handleError(err) {
 }
 
 /**
- * @param {string} url
- */
-function isValidUrl(url) {
-  const urlRegex = /^(https?|chrome):\/\/[^\s$.?#].[^\s]*$/gi;
-  return urlRegex.test(url);
-}
-
-/**
  * @param {LH.RunnerResult} runnerResult
  * @param {LH.CliFlags} flags
  * @return {Promise<void>}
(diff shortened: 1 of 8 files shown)


brendankenny:
we might want to allow `about:blank`, too? It doesn't actually work in Lighthouse, but it's useful to get the `NO_DOCUMENT_REQUEST` error. Maybe that's dumb, though.

@paulirish

brendankenny:
@paulirish confirms that is dumb, so nevermind about the `about:blank` part :)

justinribeiro:
@brendankenny Sounds good! My initial thought was that this could run earlier, but I was unsure if firing up Chrome and then failing out if the URL was invalid was a okay. Since that sounds alright, I'll shuffle this to check earlier and get this PR updated.

justinribeiro:
@brendankenny The one difference in the above implementation from your guidance is that instead of running a regex check on allowed protocols, I broke that out into `URL.isProtocolAllowed` and added the allowed protocols similar to the TLD defs. I felt like this was easier to maintain and test, but if you want that removed and just with a regex check within core/index, I can swap it out.

Other than that, operates as you noted above.
image

PR #5962, opened 2018-09-05
misc: add official lighthouse-docker goodies

### LINKED ISSUE #3715: Define a standard docker image for Lighthouse
I believe this means upstreaming the work done here:

* https://github.com/GoogleChromeLabs/gulliver/tree/master/lighthouse_machine
* https://github.com/ebidel/lighthouse-ci/tree/master/builder

---------

The lighthouse service within should be easily accessible over HTTP.
I kind of preferred the basic REST api within `lighthouse_machine`, but perhaps we should chat more about the API.

@ebidel wanna take a first whack at this?

paulirish:
Fixes #3715

AFAIK, It is NOT yet on dockerhub. (and so any readme mentions assuming that don't work yet)

----

This is a new PR for #4864, redone for easier conversation and review.


justinribeiro:
A few notes/things:

1. If the target is to publish on DockerHub, then making sure that the tagging is either version specific or channel specific may be helpful (I get this request a lot).

2. The use of `dumb-init` within the container shouldn't be required with any recent docker version (1.13+); you can forward/reap processes via passing the `--init` flag to `docker run` to the same effect (they merged tini a while back). See [docs](https://docs.docker.com/engine/reference/commandline/run/).

3. The reliance on `CAP_SYS_ADMIN` in the docs to run the container continues to bother me; it's effectively equivalent to running root. We can't use the default seccomp profile that docker ships since Chrome uses things like `unshare` cap among others, but we could either include a seccomp profile we generate via `strace` or utilize [Jess' version](https://github.com/jessfraz/dotfiles/blob/master/etc/docker/seccomp/chrome.json) and make note of it in the docs.

4. The README should cover the RHEL/CentOS user namespace issue (this came up a lot on the Rendertron repo and other Chrome-related containers).

Willing to help make updates as needed.

ebidel:
FWIW, Google's OS team asks that official Google containers get published to http://gcr.io/ rather than dockerhub.

ebidel:
@justinribeiro I don't see anything in the guidance that suggests we can mirror to Dockerhub. I brought up the gcr.io search+discoverability issues with them in the past and they told me others had mentioned the same. I think we should be in s many places as possible :)

lighthouse-docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
+FROM node:8-slim
```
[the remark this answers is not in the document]
justinribeiro: Would be easier to manage over time. That change would also require the `docker_build.sh` to be updated as well (since the path will have changed).

lighthouse-docker/headful/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
+    && apt-get update \
+    && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
+      --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get purge --auto-remove -y curl \
+    && rm -rf /src/*.deb
+
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.1/dumb-init_1.2.1_amd64 /usr/local/bin/dumb-init
```
ebidel: Get rid of this here too.

lighthouse-docker/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "license": "Apache-2.0",
+  "main": "server.js",
+  "scripts": {
+    "build": "./docker_build.sh",
+    "build:headful": "./docker_build.sh --headful",
+    "serve": "./docker_run.sh",
+    "serve:headful": "./docker_run.sh --headful",
+    "chrome:version": "docker run -it --rm --cap-add=SYS_ADMIN lighthouse_docker https://example.com --fast --quiet --output=json | node -e \"let f = ''; process.stdin.on('data', d => f += d); process.stdin.on('close', () => console.log(JSON.parse(f).userAgent));\""
+  }
+}
```
ebidel: Newline

PR #5963, opened 2018-09-05
report(util): ✅ audits should be in Passed Audits

### LINKED ISSUE #5959: report:  ✅ audits should be in Passed Audits
Can't really explain why the dom-size audit isn't in passed: 

image

i mean i can because i originally wanted it to work this way:

https://github.com/GoogleChrome/lighthouse/blob/e07a94bdd58c3da142dbaf5add15c32142072116/lighthouse-core/report/html/renderer/util.js#L145-L148

but its annoying and just leads to confusion.

 we should just use `RATINGS.PASS.minScore` instead.

justinribeiro:
`Util.showAsPassed()` now default compares to `RATINGS.PASS.minScore`; updated the perf-category render tests to check against the added `PASS_THRESHOLD` const. Resolves #5959 

*With PR*
image

*Without PR*
image

### DIFF
--- a/lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
+++ b/lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
@@ -22,8 +22,6 @@ const sampleResultsOrig = require('../../../results/sample_v2.json');
 const TEMPLATE_FILE = fs.readFileSync(__dirname +
     '/../../../../report/html/templates.html', 'utf8');
 
-const PASS_THRESHOLD = 0.9;
-
 describe('PerfCategoryRenderer', () => {
   let category;
   let renderer;
@@ -142,7 +140,8 @@ describe('PerfCategoryRenderer', () => {
     const diagnosticSection = categoryDOM.querySelectorAll('.lh-category > .lh-audit-group')[2];
 
     const diagnosticAudits = category.auditRefs.filter(audit => audit.group === 'diagnostics' &&
-        audit.result.score < PASS_THRESHOLD && audit.result.scoreDisplayMode !== 'not-applicable');
+        audit.result.score < Util.PASS_THRESHOLD &&
+        audit.result.scoreDisplayMode !== 'not-applicable');
     const diagnosticElements = diagnosticSection.querySelectorAll('.lh-audit');
     assert.equal(diagnosticElements.length, diagnosticAudits.length);
   });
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -140,7 +142,7 @@ describe('PerfCategoryRenderer', () => {
     const diagnosticSection = categoryDOM.querySelectorAll('.lh-category > .lh-audit-group')[2];
 
     const diagnosticAudits = category.auditRefs.filter(audit => audit.group === 'diagnostics' &&
-        audit.result.score !== 1 && audit.result.scoreDisplayMode !== 'not-applicable');
+        audit.result.score < PASS_THRESHOLD && audit.result.scoreDisplayMode !== 'not-applicable');
```
paulirish: can just use `Util.PASS_THRESHOLD` instead of the local declaration above. 

i think that'll be fine despite all the jsdom beforeEach bs. but holler if its stupid.

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -140,7 +140,8 @@ describe('PerfCategoryRenderer', () => {
     const diagnosticSection = categoryDOM.querySelectorAll('.lh-category > .lh-audit-group')[2];
 
     const diagnosticAudits = category.auditRefs.filter(audit => audit.group === 'diagnostics' &&
-        audit.result.score !== 1 && audit.result.scoreDisplayMode !== 'not-applicable');
+        audit.result.score < Util.PASS_THRESHOLD &&
```
brendankenny: @paulirish was there a historical reason for this not just being `!Util.showAsPassed(audit.result)`?

paulirish: nope. that sounds great.

justinribeiro: :+1: Sorted in https://github.com/GoogleChrome/lighthouse/pull/5963/commits/6a1715a1afcf494f5caeca07eb67abfdd71e5aef.

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -140,7 +140,8 @@ describe('PerfCategoryRenderer', () => {
     const diagnosticSection = categoryDOM.querySelectorAll('.lh-category > .lh-audit-group')[2];
 
     const diagnosticAudits = category.auditRefs.filter(audit => audit.group === 'diagnostics' &&
-        audit.result.score !== 1 && audit.result.scoreDisplayMode !== 'not-applicable');
+        !Util.showAsPassed(audit.result) &&
+        audit.result.scoreDisplayMode !== 'not-applicable');
```
brendankenny: oh, sorry, should have put the comment below this line. `showAsPassed()` includes the `not-applicable` check :)

justinribeiro: That's my bad; I should have noticed the check showAsPassed. :+1: Fixed in latest commit.

PR #5975, opened 2018-09-06
docs(gist): example script to post report as gist

mettamatt:
Not sure if this is the best place to include this snippet, but other's may run into this need as a headless setup.


mettamatt:
Thanks, @paulirish !

PR #5981, opened 2018-09-06
i18n: mv locale files

paulirish:


brendankenny:
need to merge/rebase. All sorts of weirdness going on with the file moves, I guess

paulirish:
ptal

PR #5993, opened 2018-09-07
core(tsc): add type checking to sentry usage

### LINKED ISSUE #5971: Sample our sentry errors more heavily
we have a lot of high volume errors that are just happening too often.

Here's our top 25 errors in the last 24hr:

```
"Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules": 305k,
"Could not find any top level events": 249k,
"No firstMeaningfulPaint found, usingfallback: 214k,
"Invalid task timing data": 202k,
"The asynchronous expression exceeded the allotted time of 60s": 18k,
"Could not load stylesheet (timeout)": 104k,
"Invalid Version: <version>": 103k,
"Invalid image sizing information <url>": 76k,
"Cannot create DNS cache with no rtt": 70k,
"finalUrl <url> not found in network records.": 50k,
"SPEEDINDEX_OF_ZERO": 34k,
"NO_TTI_NETWORK_IDLE_PERIOD": 18k,
"NO_FMP": 33k,
"Images have different sizes!": 29k,
"Cannot simulate graph with cycle": 28k,
"Cannot read property 'indexOf' of undefined": 24k,
"Cannot read property 'after' of undefined": 16k,
"Failed to construct 'URL': Invalid URL": 14k,
"navigationStart was not found in the trace": 14k,
"Failed to execute 'matches' on 'Element': '<selector>' is not a valid selector.": 13k,
"Failed to execute 'getComputedStyle' on 'Window': parameter 1 is not of type 'Element'.": 11k,
"NO_FCPUI_IDLE_PERIOD": "106k,
"NO_FCP": "102k
```

in sentry.js you'll see we sample some of these. we could afford to sample more... well actually, it probably makes sense to sample everything. lets start at 1%?

brendankenny:
long-standing TODO.

Also fixes #5971

### DIFF
--- a/lighthouse-core/lib/sentry.js
+++ b/lighthouse-core/lib/sentry.js
@@ -16,16 +16,10 @@ const SENTRY_URL = 'https://a6bb0da87ee048cc9ae2a345fc09ab2e:63a7029f46f74265981
 const SAMPLE_RATE = 0.1;
 
 const SAMPLED_ERRORS = [
-  // Error code based sampling
-  {pattern: /DOCUMENT_REQUEST$/, rate: 0.1},
-  {pattern: /(IDLE_PERIOD|FMP_TOO_LATE)/, rate: 0.1},
-  {pattern: /^NO_.*/, rate: 0.1},
   // Message based sampling
-  {pattern: /Could not load stylesheet/, rate: 0.01},
-  {pattern: /Failed to decode/, rate: 0.1},
-  {pattern: /All image optimizations failed/, rate: 0.1},
-  {pattern: /No.*resource with given/, rate: 0.01},
-  {pattern: /No.*node with given id/, rate: 0.01},
+  {pattern: /Could not load stylesheet/, rate: 0.1},
+  {pattern: /No.*resource with given/, rate: 0.1},
+  {pattern: /No.*node with given id/, rate: 0.1},
 ];
 
 const noop = () => {};
--- a/lighthouse-core/lib/sentry.js
+++ b/lighthouse-core/lib/sentry.js
@@ -13,14 +13,7 @@ const log = require('lighthouse-logger');
 const SENTRY_URL = 'https://a6bb0da87ee048cc9ae2a345fc09ab2e:63a7029f46f74265981b7e005e0f69f8@sentry.io/174697';
 
 // Per-run chance of capturing errors (if enabled).
-const SAMPLE_RATE = 0.1;
-
-const SAMPLED_ERRORS = [
-  // Message based sampling
-  {pattern: /Could not load stylesheet/, rate: 0.1},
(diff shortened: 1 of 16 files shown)


lighthouse-core/lib/sentry.js
```diff
@@ -3,94 +3,106 @@
...
-  {pattern: /^NO_.*/, rate: 0.1},
   // Message based sampling
-  {pattern: /Could not load stylesheet/, rate: 0.01},
-  {pattern: /Failed to decode/, rate: 0.1},
-  {pattern: /All image optimizations failed/, rate: 0.1},
-  {pattern: /No.*resource with given/, rate: 0.01},
-  {pattern: /No.*node with given id/, rate: 0.01},
+  {pattern: /Could not load stylesheet/, rate: 0.1},
+  {pattern: /No.*resource with given/, rate: 0.1},
+  {pattern: /No.*node with given id/, rate: 0.1},
```
brendankenny: we could drop these completely? Also do they fire multiple times per run? We could fire once per audit instead (does sentry support adding a count payload and have a decent aggregation view of them?)

lighthouse-core/lib/sentry.js
```diff
@@ -3,94 +3,95 @@
...
+    // Special case captureException to return a Promise so we don't process.exit too early
+    sentryDelegate.captureException = async (err, opts = {}) => {
       // Ignore if there wasn't an error
-      if (!err) return empty;
+      if (!err) return;
+
       // Ignore expected errors
-      if (err.expected) return empty;
-      // Sample known errors that occur at a high frequency
-      const sampledErrorMatch = SAMPLED_ERRORS.find(sample => sample.pattern.test(err.message));
```
paulirish: can we keep this block?

along with something like this above:

```js
const SAMPLED_ERRORS = [	
  // Error code based sampling	
  // e.g.: {pattern: /No.*node with given id/, rate: 0.01},	
];
```

paulirish:
type stuff and fonts look good!

lighthouse-core/lib/sentry.js
brendankenny: There you go :P

lighthouse-core/lib/sentry.js
```diff
@@ -3,94 +3,105 @@
...
+
 /**
- * We'll create a delegate for sentry so that environments without error reporting enabled will use
+ * A delegate for sentry so that environments without error reporting enabled will use
  * noop functions and environments with error reporting will call the actual Sentry methods.
  */
-const sentryDelegate = Object.assign({}, sentryApi);
-sentryDelegate.init = function init(opts) {
+const sentryDelegate = {
+  init,
```
paulirish: lol the hoist. a lil fancy for my tastes, but ok.. :)

Issue #5881, opened 2018-08-21
Add top-level `error` property on LHR


paulirish:
To communicate if there were any errors during the run. We have to decide which errors should be considered for this. At the very least, a boolean that captures if there were any auditErrors.  

See PSI's error enum: https://goto.google.com/zhian

Property name is available for bikeshedding. ;)  `errorState`?

any takers?

brendankenny:
What kind of error state is it trying to communicate?

- Numbers may not be trustworthy
- Something weird happened, you should try rerunning
- Something happened that rerunning won't fix, re-check how you've configured lighthouse

?

paulirish:
as mentioned in the thread. i think we have 3 classes of errors.

1. the page couldnt be loaded correctly (bad TLS, 500 status, etc).  aka `getPageLoadError` returned truthy.
2. page was loaded but we couldn't collect necessary diagnostics (NO_TRACING_STARTED, NO_NAVSTART).
   * see [screenshot](https://screenshot.googleplex.com/LpKpCsxmdCP). If we're gonna mark ALL perf stuff as 'Error' then we should consider it an audit failure. The two above error codes definitely end up in this state. there may be more.
3. Individual audits had errors
   * some, but not _all_

Looks like we're about to have agreement the 1 single audit error shouldn't tank the run (and so we shouldnt report a toplevel failure). But classes 1 and 2 should both tank the run.

Relatedly, we need a `catch` in `runLighthouseForLR()` so we dont get unhandled rejections. And I guess we'll have to return something in this errorState to LR?

@brendankenny can you put together a PR for this while I'm out?

paulirish:
Fixed. Some small followups done in #6071

PR #6014, opened 2018-09-13
Proposal: core: add top-level runtimeError

brendankenny:
@paulirish

Here's one simple approach to #5881. All artifact errors are caught, and any deemed `runtimeError` worthy are floated to the top-level property in the LHR. From that bug, this handles

> 2. tracing result failures: NO_tracing_start, no_navstart, NO_FCP, no_dcl = FATAL
> 4. Screenshot/speedline errors = all fatal. (unless its a pain. no big deal, really)

It sort of handles

> 1. defaultPass pageload error = runtimeError
>    * 500'd. 400'd. certificate error.  network disconnected / offline. 
>    * FAILED_DOCUMENT__REQUEST, NO_DOCUMENT__REQUEST

but if that pageload error affects enough audits (which it usually does) we throw from the whole Lighthouse run and don't ever return an LHR.

Like the idea mentioned in https://github.com/GoogleChrome/lighthouse/issues/5881#issuecomment-414914507, I'd prefer to continue throwing exceptions for these cases and then checking outside core in `runLighthouseForLR` that the thrown error has `lhrRuntimeError` set on it. Most of our clients just want a thrown error in those cases, not a pseudo-LHR, and dealing with a pseudo-LHR just makes tracking the LHR type and what's on it messier when we really just want an exception floated up.

So I don't think it's worth doing that inside core. In `runLighthouseForLR` we can create that pseudo-LHR with *just* the top-level `runtimeError` property.

Doing it that way will completely handle (1) above and
> 3. tracing recording failures: TRACING_ALREADY_STARTED, PARSING_problem, read_failed, (timeout_during_trace_retreival)
if you're still good with that idea.

If this is 👍 from you I can add that and tests.

### DIFF
--- a/lighthouse-core/lib/lh-error.js
+++ b/lighthouse-core/lib/lh-error.js
@@ -12,18 +12,20 @@ const strings = require('./strings');
  * @property {string} code
  * @property {string} message
  * @property {RegExp} [pattern]
+ * @property {boolean} [lhrRuntimeError] True if it should appear in the top-level LHR.runtimeError property.
  */
 
 class LighthouseError extends Error {
   /**
    * @param {LighthouseErrorDefinition} errorDefinition
-   * @param {Record<string, string|undefined>=} properties
+   * @param {Record<string, string|boolean|undefined>=} properties
    */
   constructor(errorDefinition, properties) {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
     this.friendlyMessage = errorDefinition.message;
+    this.lhrRuntimeError = !!errorDefinition.lhrRuntimeError;
     if (properties) Object.assign(this, properties);
 
     Error.captureStackTrace(this, LighthouseError);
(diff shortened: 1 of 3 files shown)


brendankenny:
lol no

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -111,14 +112,29 @@ async function runLighthouseInLR(connection, url, flags, {lrDevice, categoryIDs,
...
+
+    const runtimeError = {
+      code: err.code,
+      message: err.friendlyMessage ?
+          `${err.friendlyMessage} (${err.message})` :
+          err.message,
+    };
 
-  return results.report;
+    return JSON.stringify(runtimeError, null, 2);
```
paulirish: I was thinking we're returning a psuedo LHR so one more level on top of this:


```js
return JSON.stringify({runtimeError}, null, 2);
```

brendankenny: yeah I missed taht.

brendankenny:
ok, ready for review @paulirish :)

lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -111,14 +112,29 @@ async function runLighthouseInLR(connection, url, flags, {lrDevice, categoryIDs,
...
-    await assetSaver.logAssets(results.artifacts, results.lhr.audits);
-  }
+    if (logAssets) {
+      await assetSaver.logAssets(results.artifacts, results.lhr.audits);
+    }
+    return results.report;
+  } catch (err) {
+    // If an error ruined the entire lighthouse run, attempt to return a meaningful error.
+    if (!(err instanceof LHError) || !err.lhrRuntimeError) {
+      throw err;
```
paulirish: In this situation I think we should not throw but instead return an "UNKNOWN_ERROR" i guess.

tbh i dont know what throwing does in this context. my guess is its not caught and triggers a timeout?

brendankenny: done

PR #6024, opened 2018-09-14
core(icons): Add PNG check to icon validation

### LINKED ISSUE #4876: Lighthouse PWA audits pass even with a JPG icon
Chrome "Improved A2HS" requires a PNG icon in the Web Manifest file, but Lighthouse doesn't mention this in its PWA audits.

To reproduce:
Create a PWA with a JPG icon in its `manifest.json`, like this:

```
{
  "name": "Example",
  "short_name": "Example",
  "icons": [
    {
      "src": "icon.jpg",
      "sizes": "1024x1024",
      "type": "image/jpg"
    }
  ],
  "scope": "https://www.example.com/my-url/",
  "start_url": "https://www.example.com/my-url/",
  "display": "standalone",
  "background_color": "#626160",
  "theme_color": "#626160"
}
```

Expected: Lighthouse PWA audit should flag this manifest as ineligible for Web App Install Banner, because there's no PNG icon in this manifest.

Actual: All audits pass

exterkamps:
fixes #4876 

**Summary**
Adds the additional check of validating if an icon is a PNG when validating that an icon is the correct size.

**Notes** 
Changed the way that artifacts are computed in audit tests so that they can take in a manifest.json as a param to use multiple different manifests for testing.

### DIFF
--- a/lighthouse-core/lib/icons.js
+++ b/lighthouse-core/lib/icons.js
@@ -24,7 +24,7 @@ function doExist(manifest) {
  * @param {NonNullable<LH.Artifacts.Manifest['value']>} manifest
  * @return {Array<string>} Value of satisfactory sizes (eg. ['192x192', '256x256'])
  */
-function sizeAtLeast(sizeRequirement, manifest) {
+function pngSizedAtLeast(sizeRequirement, manifest) {
   // An icon can be provided for a single size, or for multiple sizes.
   // To handle both, we flatten all found sizes into a single array.
   const iconValues = manifest.icons.value;
@@ -57,5 +57,5 @@ function sizeAtLeast(sizeRequirement, manifest) {
 
 module.exports = {
   doExist,
-  sizeAtLeast,
+  pngSizedAtLeast,
 };
--- a/lighthouse-core/lib/icons.js
+++ b/lighthouse-core/lib/icons.js
@@ -52,7 +52,35 @@ function sizeAtLeast(sizeRequirement, manifest) {
       });
 }
 
+/**
+ * @param {NonNullable<LH.Artifacts.Manifest['value']>} manifest
+ * @return {boolean} True/False whether the icons are all PNGs
+ */
+function isPng(manifest) {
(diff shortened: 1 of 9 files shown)


lighthouse-core/lib/icons.js
```diff
@@ -31,7 +31,10 @@ function sizeAtLeast(sizeRequirement, manifest) {
   /** @type {Array<string>} */
   const flattenedSizes = [];
   iconValues.forEach(icon => {
-    if (icon.value.sizes.value) {
+    // check that the icon has a size, src, and is a png
+    if (icon.value.sizes.value &&
+        icon.value.src.value &&
+        icon.value.src.value.endsWith('.png')) {
```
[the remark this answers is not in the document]
exterkamps: Yup!  Modified to look at the typehint if exists, to only analyze 'image/png' typed files.

lighthouse-core/lib/icons.js
```diff
@@ -24,17 +24,26 @@ function doExist(manifest) {
...
-    }
-  });
+  iconValues
+    // filter out icons with a typehint that is not 'image/png'
+    .filter(icon => !icon.value.type.value ||
+      (icon.value.type.value &&
+      icon.value.type.value === 'image/png'))
+    // filter out icons that are not png
+    .filter(icon => icon.value.src.value &&
+      icon.value.src.value.endsWith('.png'))
```
brendankenny: some people add query strings in their manifest (e.g. `https://example.com/img.png?utm_source=manifest` or whatever for analytics), and the spec [doesn't disallow it](https://www.w3.org/TR/appmanifest/#src-member), so I think we want to do `new URL(icon.value.src.value).pathname.endsWith('.png')` here.

(note that `URL` needs to come from `const URL = require('./url-shim.js')`, which gets a different version of the constructor based on if it's running in node or in the browser)

exterkamps: Good point, fixed!

Issue #6022, opened 2018-09-14
Add 'displayValue' property in CSV output report


Scarlett-Truong:
It is great that Lighthouse has the option to export report in csv. But in the csv, there are only few data displayed (category, name, title, type, score) compares to the json file. Is it possible to get all the statistic data results, including the 'displayValue' field in csv as in the json?

brendankenny:
This sounds like it would be easy to add and an easy win. We would just want to coordinate with anyone using the output (@nourikhalass, others?) to make sure it was strictly additive and doesn't end up a breaking change for anyone.

Relevant line would be
https://github.com/GoogleChrome/lighthouse/blob/6eaf51b4ee8df9796d7e1754b005b2aee5d413e9/lighthouse-core/report/report-generator.js#L77
I believe.

brendankenny:
I don't see this in the immediate future of anyone in particular, so, open for any takers.

paulirish:
@Scarlett-Truong whats an example of an audit where you'd like the displayValue?  most of the time those strings are represented elsewhere.. (like `audit.details.overallSavingsMs`)

nourikhalass:
I'm fine with the proposed changes and think they would be a nice addition. I'm not aware of any infrastructure relying on CSV exporting so any changes made, now and in the future, are OK with me!

brendankenny:
Closing since this doesn't seem to be an immediate need :)

Please comment if it's still needed and we can reopen.

liamvictor:
The CSV files are only useful for immediate and isolated use. They don't contain the URL of the page tested so when comparing several CSVs to calculate averages over a site or section it soon becomes onerous.

All of the data is in the json file, but JSON lacks an easy interface to investigate the information. I believe that having that JSON data (especially the requestedUrl & finalUrl ) in the CSV would make the CSV output much more useful. 

In case it matters, I've been using the CLI to generate the files.

axlright:
When using the cli, why does the csv output for Largest Contentful Paint show a score like 0.01 when the html output shows a time like 5.2s? My clients ask for numerous reports that are averaged, and the csv output would be great for easily automating this. My clients ask to see data like they are used to seeing in the web results, so they want to see the 5.2s not a score like 0.01.

Is it possible to add flags to have the csv output the same metrics that the html does?

Thanks!

Spereicle:
[...]

PR #6025, opened 2018-09-14
core(report): add displayValue of an audit to CSV report

midzer:
Hi,
I had some time to jump on this one.

Have a nice weekend
midzer

**Summary**

Add more info to CSV report

Feature

requested data is available and exported to JSON already. Make it availabe for CSV, too

**Related Issues/PRs**

https://github.com/GoogleChrome/lighthouse/issues/6022


brendankenny:
thanks for the PR! But I may have marked the original issue as good to implement too quickly.

I think we don't want to move on this until (if) we have a more concrete use case. I'll close for now, but we can reopen if #6022 moves forward.

PR #6030, opened 2018-09-16
new_audit(a11y): Test axe 3.1 with new css-orientation-lock audit.

robdodson:
**Summary**
Revs axe-core to 3.1 and adds an experimental css-orientation-lock audit to test their new CSSOM support.

@wilcofiers explained their CSSOM support as:
>In axe-core 3.1, we've added an XHR request library which will attempt to fetch and parse stylesheets axe-core can't get access to directly.

Staging this PR so folks can test to see if using this feature has any adverse affects on LH performance.


paulirish:
@robdodson i dont think we're ready to start shipping Lighthouse with axe rules that rely on stylesheet fetching/cssOM'ing.

brendankenny:
Let's close in favor of #5565

Issue #5955, opened 2018-09-04
offscreen-images: fallback to eotrace if there's no TTI


paulirish:
if there's no TTI ([which is not fatal](https://github.com/GoogleChrome/lighthouse/issues/5881#issuecomment-418487688)), then offscreen-images audit will just fail. instead we can fallback to use the end of trace timestamp.  We also supply a warning in the report to say this.

An alternative to doing this is to just really suggest they re-run to get working TTI.

PR #6039, opened 2018-09-17
core: make offscreen-images audit work without TTI

exterkamp:
fixes: #5955 

**Summary**
Makes it so that when TTI derivation throws an error, the offscreen-image audit will fallback to using EndOfTab trace for cutoff instead of erroring out.

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/offscreen-images.js
+++ b/lighthouse-core/audits/byte-efficiency/offscreen-images.js
@@ -165,7 +165,7 @@ class OffscreenImages extends ByteEfficiencyAudit {
    * @param {LH.Audit.Context} context
    * @return {Promise<ByteEfficiencyAudit.ByteEfficiencyProduct>}
    */
-  static audit_(artifacts, networkRecords, context) {
+  static async audit_(artifacts, networkRecords, context) {
     const images = artifacts.ImageUsage;
     const viewportDimensions = artifacts.ViewportDimensions;
     const trace = artifacts.traces[ByteEfficiencyAudit.DEFAULT_PASS];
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -162,6 +177,38 @@ describe('OffscreenImages audit', () => {
     });
   });
 
+  it('disregards images loaded after Trace of Tab when interactive throws error', () => {
```
patrickhulce: nit: did you mean traceEnd instead of Trace of Tab? :)

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -190,4 +237,70 @@ describe('OffscreenImages audit', () => {
...
+      traces: {},
+      devtoolsLogs: {},
+      requestInteractive: async () => ({pessimisticEstimate: {nodeTimings: timings}}),
+      requestTraceOfTab: generateTraceOfTab(2),
+    }, [], context, [], context).then(auditResult => {
+      assert.equal(auditResult.items.length, 0);
+    });
+  });
+
+  it('disregards images loaded after ToT when interactive throws error (Lantern)', () => {
```
patrickhulce: I've never noticed TraceOfTab lines up so well with Tip Of Tree 😆

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -190,4 +237,70 @@ describe('OffscreenImages audit', () => {
       assert.equal(auditResult.items[0].url, 'a');
     });
   });
+
+  it('finds images loaded before last long task (Lantern)', () => {
```
patrickhulce: seems like title of this one is a bit off when we're finding a length of 0 :) which case is this meant to be testing?

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -195,29 +195,39 @@ class OffscreenImages extends ByteEfficiencyAudit {
...
       const lanternInteractive = /** @type {LH.Artifacts.LanternMetric} */ (interactive);
       // Filter out images that were loaded after all CPU activity
-      const items = context.settings.throttlingMethod === 'simulate' ?
+      items = context.settings.throttlingMethod === 'simulate' ?
         OffscreenImages.filterLanternResults(unfilteredResults, lanternInteractive) :
         // @ts-ignore - .timestamp will exist if throttlingMethod isn't lantern
         OffscreenImages.filterObservedResults(unfilteredResults, interactive.timestamp);
+    } catch (err) {
+      // use end of trace as a substitute for finding interactive time
+      items = OffscreenImages.filterObservedResults(unfilteredResults,
```
patrickhulce: seems like this will try to filter observed results even if it was a lantern case here? is this intentional?

exterkamp: Yes, because we wouldn't have a `LH.Artifacts.LanternMetric` object to use since `requestInteractive` is assumed to be the method erroring out, so when we pull the `TraceOfTab` it will be the same for Lantern and for Observed since it is just comparing the wasted artifacts and the timestamp gotten from `TraceOfTab`.

That was my logic at least, dunno if that makes sense?

patrickhulce: Hm, yeah I'm just not sure this case will ever come up in Lantern world since the ByteEfficiencyAudit base class will be needing all the same things that would have errored `requestInteractive` to compute the savings. Maybe worth adding a comment that this is specifically to address the observed case then and bail on lantern?

I might be behind the times though and perhaps there was a specific lantern-case that inspired #5955 I'm not aware of, happy to learn what was going on there too if that's the case :)

exterkamp: @paulirish was this ever including a lantern error?  

I think it was just for the Observed case, but logically it catches both cases and funnels them into the same method since in theory I think both _could_ be mapped to End of Trace?  Not sure if they should, but that's what I imagined & implemented.

exterkamp: Got more context from @brendankenny and made it so that when Lantern throws an error getting interactive the error is rethrown!

lighthouse-core/audits/byte-efficiency/offscreen-images.js
```diff
@@ -195,29 +195,43 @@ class OffscreenImages extends ByteEfficiencyAudit {
...
+      // use interactive to generate items
       const lanternInteractive = /** @type {LH.Artifacts.LanternMetric} */ (interactive);
       // Filter out images that were loaded after all CPU activity
-      const items = context.settings.throttlingMethod === 'simulate' ?
+      items = context.settings.throttlingMethod === 'simulate' ?
         OffscreenImages.filterLanternResults(unfilteredResults, lanternInteractive) :
         // @ts-ignore - .timestamp will exist if throttlingMethod isn't lantern
         OffscreenImages.filterObservedResults(unfilteredResults, interactive.timestamp);
+    } catch (err) {
+      // if the error is during a Lantern run, the result is irrevocable, so rethrow
```
brendankenny: irrecoverable? :)

exterkamp: Fixed :100:

brendankenny:
not sure what we were waiting on here :)

PR #6051, opened 2018-09-18
core(gather-runner): include error status codes in pageLoadError

### LINKED ISSUE #4082: Add top level warning if page returned an error status code
We have some checks in gather-runner to assert we have a network request for the document and it didn't completely fail: 

https://github.com/GoogleChrome/lighthouse/blob/48a129fbb6286846625c10a100f4065070688a47/lighthouse-core/gather/gather-runner.js#L289-L297

completely fail = `Network.loadingFailed` protocol event. 

However 400+ and 500+ status codes will not be caught by this, but are probably an issue. We should add a top level warning in this case.

`mainRecord.hasErrorStatusCode()` just checks if the `mainRecord.statusCode` is `>= 400`. We might as well use this and add the `mainRecord.statusCode` into a toplevelwarning, so the user knows this pageload is not your typical one.

brendankenny:
Adds a pageLoadError for 400/500 http status codes. Fixes #4082; depends on #6014 (see https://github.com/GoogleChrome/lighthouse/commit/dcb468ab16ce71f0566fbca46e7ec41b604ad365 for actual diff).

[Example Report](https://googlechrome.github.io/lighthouse/viewer/?gist=4295b586ef6b9724aea0825ae1a8836d)
screen shot 2018-09-18 at 12 32 48

Good:
- catches the error
- warns user at the top of the report that it happened and they should fix their URL
- issues `runtimeError`

Less good:
- redundant with the `http-status-code` SEO audit (#3311). That audit successfully marks the page as 404ing (or whatever), but all the other audits around it are marked as `Error!` (since no gatherers in that pass return an artifact), so it's pretty much lost in the noise

screen shot 2018-09-18 at 12 31 02

- the [SEO 403 smokehouse test](https://github.com/GoogleChrome/lighthouse/blob/25457c1e0980af7e54ab5958041c46acda6b41ee/lighthouse-cli/test/smokehouse/seo/expectations.js#L81-L139) is pretty much all broken, as another example of this

### DIFF
--- a/lighthouse-cli/test/smokehouse/seo/expectations.js
+++ b/lighthouse-cli/test/smokehouse/seo/expectations.js
@@ -78,8 +78,8 @@ module.exports = [
     },
   },
   {
-    requestedUrl: BASE_URL + 'seo-failure-cases.html?status_code=403&' + failureHeaders,
-    finalUrl: BASE_URL + 'seo-failure-cases.html?status_code=403&' + failureHeaders,
+    requestedUrl: BASE_URL + 'seo-failure-cases.html?' + failureHeaders,
+    finalUrl: BASE_URL + 'seo-failure-cases.html?' + failureHeaders,
     audits: {
       'viewport': {
         score: 0,
@@ -91,8 +91,7 @@ module.exports = [
         score: 0,
       },
       'http-status-code': {
-        score: 0,
-        displayValue: '403',
+        score: 1,
       },
       'font-size': {
         rawValue: false,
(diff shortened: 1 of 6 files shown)


brendankenny:
The redundancy might be fine, and we can just update the smoke test expectations, but it is another example of the tension between automatically failing gatherers (or not) in pageLoadError cases.

On the one hand, why is Lighthouse wasting time auditing stuff that was plainly in error (nothing like getting results for the "There was no service found for the uri requested." page)? On the other, why should Lighthouse care? Just warn me prominently but in the meantime give me the results for the page I asked for.

paulirish:
HUGE. 👏 
This also removes the need for the only remaining patch that LR applies on top of LH.

brendankenny:
After discussion, we're going to maintain the change mostly as is.

I've split the SEO expectations into separate runs for HTTP 403 and other failures. That way we still get the end-to-end request failure test (including that it spawns error artifacts -> audit errors) while not losing all the other useful failure results to `Error!`.

brendankenny:
appveyor failure is a transient chromestatus issue (no manifest found). Other error is bundlesize, which all the PRs are facing.

PR #6056, opened 2018-09-19
core(driver): deliver trace as events rather than a stream

### LINKED ISSUE #5939: Race trace parsing and chunk fetching
On LR (and extension to some degree), we've been seeing trace retrieval taking a lot of time. 

Here's a timing breakdown when getTrace takes 14sec:

image

Brendan's idea is to race our streaming JS parser and fetching the next chunk of the trace. 

Adds some complexity to the async handling, but the perf win would be ~2100ms on this recording.

### LINKED ISSUE #5968: Hanging trace retrieval on LR
We have evidence that the loop of `_readTraceFromStream()` is stalling out and hanging indefinitely. 

1. we might want to add some timeouts
2. we could experiment with [not using a stream](https://github.com/GoogleChrome/lighthouse/blame/0e8982ab037d928a37d8dddd00aa7f1e913f5414/helpers/browser/driver.js) (in the LR case)
3. trace reduction #3596 will help but probably not fix

exterkamps:
**Summary**

This changes the Tracing.start protocol command to use the `ReportEvents` transferMode instead of `ReturnAsStream`.  This makes the trace gathering much faster, as observed in some automated run timings:

| URL | ReturnAsStream | ReportEvents | Difference |
|---|---|---|---|
| http://cnn.com | 3410ms | 1202ms | (2208ms) |
|http://paulirish.com| 174ms | 87ms | (~86ms) |

Note: timings are averages of 5 seperate runs with each method.

**Related Issues/PRs**

fixes #5968
closes #5939
Based on code in [old driver implementation](https://github.com/GoogleChrome/lighthouse/blame/0e8982ab037d928a37d8dddd00aa7f1e913f5414/helpers/browser/driver.js)

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -68,6 +68,13 @@ class Driver {
      */
     this._monitoredUrl = null;
 
+    /**
+     * @type {LH.Trace}
+     */
+    this._traceEvents = {
+      traceEvents: [],
+    };
+
     connection.on('protocolevent', event => {
       this._devtoolsLog.record(event);
       if (this._networkStatusMonitor) {
@@ -929,11 +936,15 @@ class Driver {
       throw new Error('DOM domain enabled when starting trace');
     }
 
+    // setup listener to listen for when trace a data bundle is collected
+    this.on('Tracing.dataCollected', data => {
+      this._traceEvents.traceEvents.push(...data.value);
+    });
+
     // Enable Page domain to wait for Page.loadEventFired
     return this.sendCommand('Page.enable')
       .then(_ => this.sendCommand('Tracing.start', {
         categories: uniqueCategories.join(','),
-        transferMode: 'ReturnAsStream',
         options: 'sampling-frequency=10000', // 1000 is default and too slow.
       }));
   }
@@ -943,10 +954,8 @@ class Driver {
    */
   endTrace() {
     return new Promise((resolve, reject) => {
-      // When the tracing has ended this will fire with a stream handle.
-      this.once('Tracing.tracingComplete', completeEvent => {
-        this._readTraceFromStream(completeEvent)
-            .then(traceContents => resolve(traceContents), reject);
+      this.once('Tracing.tracingComplete', _ => {
+        resolve(this._traceEvents);
       });
 
       // Issue the command to stop tracing.
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -936,7 +936,7 @@ class Driver {
       throw new Error('DOM domain enabled when starting trace');
     }
 
-    // setup listener to listen for when trace a data bundle is collected
+    // setup listener for when a trace bundle is sent
     this.on('Tracing.dataCollected', data => {
       this._traceEvents.traceEvents.push(...data.value);
     });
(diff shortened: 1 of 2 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -68,6 +68,13 @@ class Driver {
      */
     this._monitoredUrl = null;
 
+    /**
```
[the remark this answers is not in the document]
paulirish: your dataCollected listener can be inside the top of endTrace and then it can stay local to that method.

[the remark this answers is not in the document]
paulirish: might as well just collect them in a plain `const traceEvents = []` array.

lighthouse-core/gather/driver.js
```diff
@@ -943,10 +954,8 @@ class Driver {
    */
   endTrace() {
     return new Promise((resolve, reject) => {
-      // When the tracing has ended this will fire with a stream handle.
-      this.once('Tracing.tracingComplete', completeEvent => {
-        this._readTraceFromStream(completeEvent)
-            .then(traceContents => resolve(traceContents), reject);
+      this.once('Tracing.tracingComplete', _ => {
+        resolve(this._traceEvents);
```
paulirish: down here you can resolve with `resolve({traceEvents})` as that's a LH.Trace

lighthouse-core/gather/driver.js
```diff
@@ -942,55 +940,21 @@ class Driver {
...
-      this.once('Tracing.tracingComplete', completeEvent => {
-        this._readTraceFromStream(completeEvent)
-            .then(traceContents => resolve(traceContents), reject);
-      });
+    /** @type {Array<LH.TraceEvent>} */
+    const traceEvents = [];
 
-      // Issue the command to stop tracing.
-      return this.sendCommand('Tracing.end').catch(reject);
+    // setup listener for when a trace bundle is sent
```
paulirish: "trace chunk" but yeah. How about "// dataCollected events fire for each trace chunk, in order"

lighthouse-core/gather/driver.js
```diff
@@ -942,55 +940,21 @@ class Driver {
...
-        }
-
-        return this.sendCommand('IO.read', readArguments).then(onChunkRead);
-      };
+      this.once('Tracing.tracingComplete', _ => {
+        resolve({traceEvents});
+      });
 
-      this.sendCommand('IO.read', readArguments).then(onChunkRead).catch(reject);
+      // Issue the command to stop tracing.
```
paulirish: drop this

lighthouse-core/gather/driver.js
exterkamps: Changed, see if u like it.

lighthouse-core/gather/driver.js
```diff
@@ -942,55 +940,25 @@ class Driver {
...
-        parser.parseChunk(response.data);
-
-        if (response.eof) {
-          isEOF = true;
-          return resolve(parser.getTrace());
-        }
-
-        return this.sendCommand('IO.read', readArguments).then(onChunkRead);
-      };
+      this.once('Tracing.tracingComplete', _ => {
```
patrickhulce: so. much. easier. :D

Issue #5798, opened 2018-08-07
Extension Error from running Lighthouse on a github page


Jeremip11:
**Lighthouse Version**: 3.0.3
**Lighthouse Commit**: d1cae24fda4182406e02d3f4df6309d48878fc50
**Chrome Version**: 70.0.3510.2
**Initial URL**: https://github.com/Jeremip11/thimble.mozilla.org/pull/4
**Error Message**: Cannot attach to this target.
**Stack Trace**:
```
Error: Cannot attach to this target.
    at Object.chrome.debugger.attach [as callback] (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17315:15)
    at safeCallbackApply (extensions::uncaught_exception_handler:27:15)
    at handleResponse (extensions::sendRequest:67:7)
```

brendankenny:
I'm not certain what this bug is exactly. `Cannot attach to this target` should only come from Chrome if trying to run lighthouse on a privileged page (e.g. a `file:///...`).

However, can anyone else run the extension on github pages? (e.g. this page) They run fine in the CLI (master) and devtools (3.0.3) for me, but in the extension they're failing with 
```
Error in event handler for debugger.onDetach: Error: Lighthouse detached from browser: target_closed
    at ExtensionConnection._onUnexpectedDetach (chrome-extension://blipmdconlkpinefehnmjammfjpmpbjk/scripts/lighthouse-ext-background.js:17284:7)
```
for me. The driver then times out and I get `ExtensionConnection:error No tabId set for sendCommand` for subsequent commands, all ending with a sad `ExtensionConnection:warn disconnect() was called without an established connection.`

justinribeiro:
@brendankenny Spot checked some github pages with extension, wasn't able to duplicate.

![image](https://user-images.githubusercontent.com/643503/43978106-04c8da7c-9c9b-11e8-992d-f901116a281e.png)

brendankenny:
thanks for checking! @wardpeet mentioned it may be new behavior in Chrome 69+, but haven't verified

justinribeiro:
@wardpeet is spot on; I can duplicate that with M69+ and the extension.

wardpeet:
[...]

PR #6064, opened 2018-09-19
tests: change extension test url from paulirish.com to google.com

wardpeet:
**Summary**
Changin the test url to google.com

**Related Issues/PRs**
#5798


wardpeet:
how can we make sure the test actually would fail? As on my chrome it works (chrome 69)

wardpeet:
Honestly I don't really care what url we test but I agree that paulirish.com is probably a more stable url and we probably get these edge cases logged by the report button anyway.

Issue #5220, opened 2018-05-15
Audit: new dom nodes table is confusing


ebidel:
I'm not sure what the last line is telling me. It's identified two nodes, but they're not explained or lined up to anything.

<img width="872" alt="screen shot 2018-05-15 at 6 41 07 pm" src="https://user-images.githubusercontent.com/238208/40087206-9f6128de-586f-11e8-9b64-7e4cc05de42f.png">

paulirish:
cc @hwikyounglee for ideas

paulirish:
@exterkamps ive only seen [style] be a problem. my concern is that we'll strip some attributes that are useful. `[data-*]` come to mind.
but there could be others.

if we are subtractive then there's a much smaller risk that we effed up and removed important data.
and i presume it's more straightforward see that we need to remove another attribute than for people to tell us that we've excluded an attribute they need to identify the element.  🤷‍♂️

exterkamps:
That's fair, and easier to implement.  I'll strip out just `style=".*"` tags then.

And did we want to make the stripping of snippets more generic, or put it into some lib/ file so that all snippets can use it? i.e. the accessibility snippets

paulirish:
> And did we want to make the stripping of snippets more generic, or put it into some lib/ file so that all snippets can use it? i.e. the accessibility snippets

yah good call.

Issue #6049, opened 2018-09-18
report details table right-alignment is ugly


paulirish:
e.g. 
![image](https://user-images.githubusercontent.com/39191/45706726-d66c1800-bb31-11e8-9148-f67d286dba7a.png)

first column should always be left aligned. 

but ideally all text columns are left-aligned and all number columns are right-aligned.  not sure if we have the classes for this, but we might.

shane sez we should drop the centered TH's too. let's keep them always the same as the TDs

wardpeet:
what do we assign as a number?
we currently have these types:
* `.lh-table-column--text`
* `.lh-table-column--bytes`
* `.lh-table-column--timespanMs`
* `.lh-table-column--ms`

I guess we align everything right except `.lh-table-column--text`
we end up with
![image](https://user-images.githubusercontent.com/1120926/45777171-53dc7900-bc55-11e8-9400-ea3d3f886208.png)

brendankenny:
fixed? or 3/6 fixed

wardpeet:
Fixed

PR #6065, opened 2018-09-19
core(audit): make dom-size table prettier

exterkamp:
**Summary**
Reformatted the dom-size output table.  Converted the table to be more row-centric vs column centric.  See issue #5220 for discussion and screenshots.  Some CSS changes required from #6049 to format the `code` snippets and the left alignment of the text.

**Related Issues/PRs**
fixes: #5220 
requires: #6049

### DIFF
--- a/lighthouse-core/audits/dobetterweb/dom-size.js
+++ b/lighthouse-core/audits/dobetterweb/dom-size.js
@@ -96,28 +96,46 @@ class DOMSize extends Audit {
     );
 
     const headings = [
-      {key: 'totalNodes', itemType: 'text', text: str_(UIStrings.columnDOMNodes)},
-      {key: 'depth', itemType: 'text', text: str_(UIStrings.columnDOMDepth)},
-      {key: 'width', itemType: 'text', text: str_(UIStrings.columnDOMWidth)},
+      // TODO i18n new UI text
+      {key: 'category', itemType: 'text', text: 'Category'},
+      {key: 'element', itemType: 'code', text: 'Element'},
+      {key: 'observed', itemType: 'text', text: 'Count'},
     ];
+
+    // get identifying snippet parts
+    const divMatcher = /<(\S+).*>/;
+    const identifierMatcher = /(id="\S+"|class="\S+")/;
+
+    const depthNode = divMatcher.exec(stats.depth.snippet);
+    const dIdentifier = identifierMatcher.exec(stats.depth.snippet);
+
+    const widthNode = divMatcher.exec(stats.width.snippet);
+    const wIdentifier = identifierMatcher.exec(stats.width.snippet);
 
     /** @type {Array<Object<string, LH.Audit.DetailsItem>>} */
     const items = [
       {
-        totalNodes: Util.formatNumber(stats.totalDOMNodes),
-        depth: Util.formatNumber(stats.depth.max),
-        width: Util.formatNumber(stats.width.max),
+        category: 'Total DOM Nodes',
+        element: '',
+        observed: Util.formatNumber(stats.totalDOMNodes),
       },
       {
-        totalNodes: '',
-        depth: {
+        category: 'Max Depth',
+        element: {
           type: 'code',
-          value: stats.depth.snippet,
+          value: '<' + (depthNode !== null? depthNode[1]: '') +
+            (dIdentifier !== null? ' ' + dIdentifier[1]: '') + '>',
         },
-        width: {
+        observed: Util.formatNumber(stats.depth.max),
+      },
+      {
+        category: 'Max Children',
+        element: {
           type: 'code',
-          value: stats.width.snippet,
+          value: '<' + (widthNode !== null? widthNode[1]: '') +
+            (wIdentifier !== null? ' ' + wIdentifier[1]: '') + '>',
         },
+        observed: Util.formatNumber(stats.width.max),
       },
     ];
 
--- a/lighthouse-core/audits/dobetterweb/dom-size.js
+++ b/lighthouse-core/audits/dobetterweb/dom-size.js
@@ -32,17 +32,23 @@ const UIStrings = {
     'children/parent element. A large DOM can increase memory usage, cause longer ' +
     '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
     'and produce costly [layout reflows](https://developers.google.com/speed/articles/reflow). [Learn more](https://developers.google.com/web/tools/lighthouse/audits/dom-size).',
-  /** Label for the total number of DOM nodes found in the page. */
-  columnDOMNodes: 'Total DOM Nodes',
-  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
-  columnDOMDepth: 'Maximum DOM Depth',
-  /** Label for the value of the maximum number of children any DOM node in the page has. */
-  columnDOMWidth: 'Maximum Children',
+  /** Table column header for the type of analytic found. */
+  columnCategory: 'Category',
+  /** Table column header for DOM Element's snippet. */
+  columnElement: 'Element',
+  /** Table column header for the observed analytic's value. */
+  columnObserved: 'Observed',
   /** [ICU Syntax] Label for an audit identifying the number of DOM nodes found in the page. */
   displayValue: `{itemCount, plural,
     =1 {1 node}
     other {# nodes}
     }`,
+  /** Label for the total number of DOM nodes found in the page. */
+  categoryDOMNodes: 'Total DOM Nodes',
+  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
+  categoryDOMDepth: 'Maximum DOM Depth',
+  /** Label for the value of the maximum number of children any DOM node in the page has. */
+  categoryDOMWidth: 'Maximum Children',
 };
 
 const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
@@ -96,21 +102,20 @@ class DOMSize extends Audit {
     );
 
     const headings = [
-      // TODO i18n new UI text
-      {key: 'category', itemType: 'text', text: 'Category'},
-      {key: 'element', itemType: 'code', text: 'Element'},
-      {key: 'observed', itemType: 'text', text: 'Count'},
+      {key: 'category', itemType: 'text', text: str_(UIStrings.columnCategory)},
+      {key: 'element', itemType: 'code', text: str_(UIStrings.columnElement)},
+      {key: 'observed', itemType: 'text', text: str_(UIStrings.columnObserved)},
     ];
 
     /** @type {Array<Object<string, LH.Audit.DetailsItem>>} */
     const items = [
       {
-        category: str_(UIStrings.columnDOMNodes),
+        category: str_(UIStrings.categoryDOMNodes),
         element: '',
         observed: Util.formatNumber(stats.totalDOMNodes),
       },
       {
-        category: str_(UIStrings.columnDOMDepth),
+        category: str_(UIStrings.categoryDOMDepth),
         element: {
           type: 'code',
           value: stats.depth.snippet,
@@ -118,7 +123,7 @@ class DOMSize extends Audit {
         observed: Util.formatNumber(stats.depth.max),
       },
       {
-        category: str_(UIStrings.columnDOMWidth),
+        category: str_(UIStrings.categoryDOMWidth),
         element: {
           type: 'code',
           value: stats.width.snippet,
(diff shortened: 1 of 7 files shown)


lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
     'children/parent element. A large DOM can increase memory usage, cause longer ' +
     '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
     'and produce costly [layout reflows](https://developers.google.com/speed/articles/reflow). [Learn more](https://developers.google.com/web/tools/lighthouse/audits/dom-size).',
-  /** Label for the total number of DOM nodes found in the page. */
-  columnDOMNodes: 'Total DOM Nodes',
-  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
-  columnDOMDepth: 'Maximum DOM Depth',
-  /** Label for the value of the maximum number of children any DOM node in the page has. */
-  columnDOMWidth: 'Maximum Children',
+  /** Table column header for the type of analytic found. */
```
exterkamps: @brendankenny @paulirish are these label comments sufficient?

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
+  columnCategory: 'Category',
```
paulirish: bikeshed. wdyt about 'Statistic' or 'Metric'. cc @brendankenny @patrickhulce

lighthouse-core/audits/dobetterweb/dom-size.js
paulirish: keep in mind that ~90 different linguists will be reading these descriptions to translate the text to another language. And they see them in isolation, without knowing necessarily what it looks like in the UI, or without having read any other descriptions of other strings. :)

in this case 'analytic' and 'snippet' aren't well understood terms, so that'd be hard for them.

> Table column header for the type of statistic. These statistics describe how big the DOM is (count of DOM nodes, children, depth).

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
+  /** Table column header for DOM Element's snippet. */
```
paulirish: > Table column header for the DOM element. Each DOM element is described with its HTML representation.

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
+  columnElement: 'Element',
+  /** Table column header for the observed analytic's value. */
```
paulirish: > Table column header for the observed value of the DOM statistic

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
   /** [ICU Syntax] Label for an audit identifying the number of DOM nodes found in the page. */
   displayValue: `{itemCount, plural,
     =1 {1 node}
     other {# nodes}
     }`,
+  /** Label for the total number of DOM nodes found in the page. */
+  categoryDOMNodes: 'Total DOM Nodes',
+  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
+  categoryDOMDepth: 'Maximum DOM Depth',
+  /** Label for the value of the maximum number of children any DOM node in the page has. */
```
paulirish: Might add "This item described is the element that has the most children" just to be safe.

lighthouse-core/lib/page-functions.js
```diff
@@ -107,13 +107,22 @@ function getElementsInDocument(selector) {
 /**
  * Gets the opening tag text of the given node.
  * @param {Element} element
+ * @param {Array<string>} ignore
  * @return {string}
  */
 /* istanbul ignore next */
-function getOuterHTMLSnippet(element) {
+function getOuterHTMLSnippet(element, ignore=[]) {
```
paulirish: ignoreAttrs ?

lighthouse-core/lib/page-functions.js
```diff
@@ -107,13 +107,22 @@ function getElementsInDocument(selector) {
...
   const reOpeningTag = /^.*?>/;
   const match = element.outerHTML.match(reOpeningTag);
-  return (match && match[0]) || '';
+  if (!(match && match[0])) return '';
+
+  let prunedMatch = match[0];
+  ignore.forEach(attribute =>{
+    const ignoreRegex = new RegExp(attribute + '=".*?"');
+    prunedMatch = prunedMatch.split(ignoreRegex).join('');
```
paulirish: wdyt about replacing with `style="..."` instead?

lighthouse-core/lib/page-functions.js
```diff
@@ -107,12 +107,20 @@ function getElementsInDocument(selector) {
 /**
  * Gets the opening tag text of the given node.
  * @param {Element} element
+ * @param {Array<string>} ignore
```
brendankenny: `@param {Array<string>}= ignoreAttrs` (`=` to mark it as optional)

lighthouse-core/lib/page-functions.js
```diff
@@ -107,12 +107,20 @@ function getElementsInDocument(selector) {
 /**
  * Gets the opening tag text of the given node.
```
brendankenny: update the jsdoc string? Something like "Optionally provide an array of attribute names to not include in the tag snippet." (or whatever)

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * @license Copyright 2017 Google Inc. All Rights Reserved.
```
brendankenny: 2018

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const assert = require('assert');
+const fs = require('fs');
+const jsdom = require('jsdom');
+const URL = require( '../../lib/url-shim' );
+const DOM = require('../../report/html/renderer/dom.js');
+const Util = require('../../report/html/renderer/util.js');
```
brendankenny: looks like `URL` and `Util` can be removed (DOM is kind of handy to have around)

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -0,0 +1,82 @@
...
+const pageFunctions = require('../../lib/page-functions');
+
+const TEMPLATE_FILE = fs.readFileSync(__dirname +
```
brendankenny: since all the elements are created dynamically below, I don't believe it needs a template, either (can just do `const document = jsdom.jsdom();`

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -0,0 +1,82 @@
...
+    global.URL = undefined;
+    global.Util = undefined;
+  });
+
+  describe('get outer HTML snippets', () => {
+    it('gets full HTML snippet', () => {
+      const snippet = pageFunctions.getOuterHTMLSnippet(
+        dom.createElement('div', '', {id: '1', style: 'style'})
+      );
+      assert.ok(snippet.includes('div') &&
```
brendankenny: hmm, it seems reasonable to assert the full string of `snippet` since the input element is fully under control of the test. It might make the test more readable as well, and could catch things like if someone messes up the regex at some point and it accidentally includes part of the closing tag or something.

WDYT vs asserting the parts like this is doing?

lighthouse-core/lib/page-functions.js
exterkamps: `@param {Array<string>=}` or `@param {Array<string>}=`  ?

brendankenny: whoops, yes, inside `{Array<string>=} `

lighthouse-core/audits/dobetterweb/dom-size.js
exterkamps: I took the Category name from the `Minimizes main thread work` Audit table.  So we can update the phrasing in every table if we like Metric more (which I agree is more descriptive than Category).

category_screenshot

lighthouse-core/audits/dobetterweb/dom-size.js
exterkamps: Added, might want to check the wording again.

lighthouse-core/lib/page-functions.js
```diff
@@ -107,12 +107,20 @@ function getElementsInDocument(selector) {
...
  * @param {Element} element
+ * @param {Array<string>=} ignoreAttrs Optionally provides an array of attribute tags to not include in HTML the snippet
```
brendankenny: I meant up in the function description, but down by the param is fine too. The word order got a little scrambled here...maybe "An optional array of attribute tags to not include in the HTML snippet."?

lighthouse-core/test/lib/page-functions-test.js
exterkamps: Haha, yeah that's the way I did it originally, but was nervous that might be too explicit.  Changed it back!  Def better and will catch errors in parsing the carrots.  My only concern was with the ordering of elements in the string repr of the element, like style coming before aria-labels, but that is deterministic? maybe? Seems to work fine tho.

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -0,0 +1,53 @@
...
+
+describe('DetailsRenderer', () => {
+  let dom;
+
+  beforeAll(() => {
+    const document = jsdom.jsdom();
+    dom = new DOM(document);
+  });
+
+  afterAll(() => {
```
brendankenny: can delete this block now

lighthouse-core/test/lib/page-functions-test.js
brendankenny: > but that is deterministic

I think at least for jsdom we can assume it is. I have no idea if it's guaranteed in a real browser :)

lighthouse-core/audits/dobetterweb/dom-size.js
brendankenny: heh, in that case I feel like "Category" is better (since it's actually bringing e.g. multiple Script Evaluation tasks together and grouping them into one line), but also see the appeal of consistency

brendankenny:
you'll also need to run `yarn update:sample-json` to get the updated localization descriptions

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
     '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
     'and produce costly [layout reflows](https://developers.google.com/speed/articles/reflow). [Learn more](https://developers.google.com/web/tools/lighthouse/audits/dom-size).',
-  /** Label for the total number of DOM nodes found in the page. */
-  columnDOMNodes: 'Total DOM Nodes',
-  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
-  columnDOMDepth: 'Maximum DOM Depth',
-  /** Label for the value of the maximum number of children any DOM node in the page has. */
-  columnDOMWidth: 'Maximum Children',
+  /** Table column header for the type of statistic. These statistics describe how big the DOM is (count of DOM nodes, children, depth). */
+  columnCategory: 'Category',
```
paulirish: lets rename this guy too

```js
columnStatistic: 'Statistic',
```

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
+  /** Table column header for the DOM element. Each DOM element is described with its HTML representation. */
+  columnElement: 'Element',
+  /** Table column header for the observed value of the DOM statistic. */
+  columnObserved: 'Observed',
```
paulirish: Observed  => Value

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -32,17 +32,23 @@ const UIStrings = {
...
   displayValue: `{itemCount, plural,
     =1 {1 node}
     other {# nodes}
     }`,
+  /** Label for the total number of DOM nodes found in the page. */
+  categoryDOMNodes: 'Total DOM Nodes',
+  /** Label for the numeric value of the maximum depth in the page's DOM tree. */
+  categoryDOMDepth: 'Maximum DOM Depth',
+  /** Label for the numeric value of the maximum number of children any DOM node in the page has. The element described will have the most children in the page. */
+  categoryDOMWidth: 'Maximum Children',
```
paulirish: Children => Child Elements 

(and above in the comment.  to clarify that we mean Elements not nodes.)

paulirish:
**_oochie coochie la la la. oochie coochie._** 🎶

Issue #6067, opened 2018-09-19
a run of theverge.com takes 85s


paulirish:
![image](https://user-images.githubusercontent.com/39191/45783872-e4966300-bc1b-11e8-989a-59a22fd2497a.png)

A few things in here we can look into
* we hit the 45s timeout. though the trace is mostly idle after the first 5s (!!). 
* gatherTrace takes 17s!! @exterkamps's #6056 might be our savior here
* if the trace is shorter  then lantern and speedline _should_ be faster, but those are certainly taking up a lot of time.
* [x] @patrickhulce looking at idle periods not triggering on this run
* [x] speedline's trace sorting took ~1.25 seconds (longer than `calculateVisualProgress`!) even though we've already sorted it https://github.com/paulirish/speedline/pull/70
* [x] `replaceIcuMessageInstanceIds` adds another ~4 seconds due to running a regex twice over every string value in our report #6072

paulirish:
> gatherTrace takes 17s!! @exterkamps's #6056 might be our savior here

from 17s to 5s. OH YEAHHHHHHHH. 🎊🎊

brendankenny:
`replaceIcuMessageInstanceIds` > #6072

patrickhulce:
> idle periods not triggering on this run

Core problem here seems to be that for iframe documents we never end up getting a `loadingFinished` message, just the response received. I'll do a workaround similar to the QUIC situation but should probably look into if this was a recent change or why it was just being masked before.

paulirish:
with #6078 taken care of, we call this closed.

**From 85s ➡️ 16s!**


srs team effort. nice one y'all!

brendankenny:
🎉 🎉

PR #6072, opened 2018-09-20
i18n: speed up replacement regex

brendankenny:
The regex we use for checking if there are strings in the LHR that need to be replaced with their localized version can be really slow against giant strings. Adding a pre-pass with a fast regex is one easy approach to speeding this up without changing our ID format or being too particular about the file paths we allow.

In a trace from the issues we're seeing in #6067, time for localization and report generation goes from ~3500ms to ~33ms.

Even time for localization in the much smaller artifacts in `yarn update:sample-json` goes from 150ms to 22ms :)


paulirish:
NICEEEEEEE 🏁

patrickhulce:
nice easy win 🎉

PR #6078, opened 2018-09-20
core(network-recorder): consider iframe responses finished

patrickhulce:
**Summary**
root requests inside frames never fire a `loadingFinished` devtools event, we'll use the presence of a response as the criteria for finished instead.

**Related Issues/PRs**
#6067

### DIFF
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -107,7 +107,7 @@ class NetworkRecorder extends EventEmitter {
    */
   static _isFrameRootRequestAndFinished(record) {
     const isFrameRootRequest = record.url === record.documentURL;
-    const responseReceived = Number.isFinite(record.responseReceivedTime);
+    const responseReceived = record.responseReceivedTime > 0;
     return !!(isFrameRootRequest && responseReceived && record.endTime);
   }
 
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -100,6 +100,18 @@ class NetworkRecorder extends EventEmitter {
   }
 
   /**
+   * frame root network requests don't always "finish" even when they're done loading data, use responseReceived instead
+   * @see https://github.com/GoogleChrome/lighthouse/issues/6067#issuecomment-423211201
+   * @param {LH.Artifacts.NetworkRequest} record
+   * @return {boolean}
+   */
+  static _isFrameRootRequestAndFinished(record) {
+    const isFrameRootRequest = record.url === record.documentURL;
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/network-recorder.js
```diff
@@ -99,6 +99,18 @@ class NetworkRecorder extends EventEmitter {
...
   }
 
+  /**
+   * frame root network requests don't always "finish" even when they're done loading data, use responseReceived instead
+   * @see https://github.com/GoogleChrome/lighthouse/issues/6067#issuecomment-423211201
+   * @param {LH.Artifacts.NetworkRequest} record
+   * @return {boolean}
+   */
+  static _isFrameRootRequestAndFinished(record) {
+    const isFrameRootRequest = record.url === record.documentURL;
```
brendankenny: do we have to worry about query strings or anything else throwing this check off? (are these fields canonicalized differently, I guess)

lighthouse-core/lib/network-recorder.js
```diff
@@ -99,6 +99,18 @@ class NetworkRecorder extends EventEmitter {
...
+    const responseReceived = record.responseReceivedTime > 0;
```
brendankenny: I really hate trying to remember the difference between data-received/response-received/finished/etc (as well as where things like requestServedFromCache fall into things), but just to make sure I understand, typically `responseReceived` only fires once, we just usually expect a `loadingFinished` to also be fired?

brendankenny:
you'll need to sign the CLA 😢

patrickhulce:
I signed it!

lighthouse-core/lib/network-recorder.js
patrickhulce: From what I could tell this is always the full query string-included URL. There were lots of samples on the verge that had misc query string params.

lighthouse-core/lib/network-recorder.js
patrickhulce: Correct that is my understanding as well, data received can fire lots of times, but we always expect one of response received + [loadingFinished/loadingFailed](https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/sdk/NetworkManager.js?type=cs&q=loadingFailed+f:devtools+lang:js&sq=package:chromium&g=0&l=841-843) to signal the end of the request.

wardpeet: do we need to consider preloads or won't this affect anything?

paulirish:
@patrickhulce i think we can revert this now with the oopif #6922 fix?

patrickhulce:
> i think we can revert this now with the oopif #6922 fix?

Put up #7517 👍

PR #6083, opened 2018-09-20
core: smooth rough edges of pageLoadError display and reporting

### LINKED ISSUE #6071: runtimeError: even more useful
* [ ] Report UI should show the friendlyMessage and code in addition to the typical message.
* [ ] runtimeWarnings and runtimeError both reporting the same sorta thing for 4xx and 5xx cases
* [ ] get rid of throwing if `> 50%`

brendankenny:
fixes #6071

- runtimeWarnings and runtimeError both use the same message for page load errors
- no longer throw on > 50% of audits failing (top level `runtimeError` and individual audit errors now take care of warning the user in that case)
- Show error `friendlyMessage` in HTML report's warning box (`runWarnings`)
screen shot 2018-09-20 at 15 33 18

bonus: no longer have a big red `GatherRunner:error FAILED_DOCUMENT_REQUEST` error log on the pageLoadError we expect in the offline pass when a page doesn't support offline :) @exterkamp

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -333,7 +333,6 @@ class GatherRunner {
     /** @type {Partial<LH.GathererArtifacts>} */
     const gathererArtifacts = {};
 
-    const pageLoadFailures = [];
     const resultsEntries = /** @type {GathererResultsEntries} */ (Object.entries(gathererResults));
     for (const [gathererName, phaseResultsPromises] of resultsEntries) {
       if (gathererArtifacts[gathererName] !== undefined) continue;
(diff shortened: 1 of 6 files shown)


brendankenny:
I will fight you

exterkamp:
Love it.  No more red text.

lighthouse-core/runner.js
```diff
@@ -301,9 +301,7 @@ class Runner {
   static getArtifactRuntimeError(artifacts) {
     for (const possibleErrorArtifact of Object.values(artifacts)) {
       if (possibleErrorArtifact instanceof LHError && possibleErrorArtifact.lhrRuntimeError) {
-        const errorMessage = possibleErrorArtifact.friendlyMessage ?
-            `${possibleErrorArtifact.friendlyMessage} (${possibleErrorArtifact.message})` :
```
patrickhulce: this concatenation is now directly handled by gather runner now, right? was that the only case we fell into this path?

brendankenny: Yeah, in gather runner. And since this was only introduced in #6051, there's nothing depending on it. `message` here is the error code, so it seemed redundant to keep it when the code is sitting right next to it on the object. I can add it back if we want to match the other place(s?) we put together an error string out of an lherror

patrickhulce: Nah, sounds good as-is 👍

lighthouse-core/gather/gather-runner.js
```diff
@@ -156,22 +156,19 @@ class GatherRunner {
...
-    let errorCode;
-    let errorReason;
+    let errorDef;
     if (!mainRecord) {
-      errorCode = LHError.errors.NO_DOCUMENT_REQUEST;
+      errorDef = LHError.errors.NO_DOCUMENT_REQUEST;
     } else if (mainRecord.failed) {
-      errorCode = LHError.errors.FAILED_DOCUMENT_REQUEST;
-      errorReason = mainRecord.localizedFailDescription;
+      errorDef = {...LHError.errors.FAILED_DOCUMENT_REQUEST};
```
paulirish: why so fancy?

brendankenny: > why so fancy?

easier than `Object.assign()`? Dunno :) Just needed a shallow copy so it's not modifying the original message

Issue #6026, opened 2018-09-15
DevTools Error: Cannot read property 'value' of undefined


StructByLightning:
**Initial URL**: https://www.parahumans.net/
**Chrome Version**: 69.0.3497.92
**Error Message**: Cannot read property 'value' of undefined
**Stack Trace**:
```
TypeError: Cannot read property 'value' of undefined
    at sendCommand.then.result (chrome-devtools://devtools/remote/serve_file/@eb2c6d16bcb960cc5c322243c1771713460c4bcf/audits2_worker/audits2_worker_module.js:1008:211)
```

brendankenny:
Something went pretty wrong here. This is the `result` on the response from `sendCommand('Runtime.evaluate', ...)`, which should [always exist](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#method-evaluate), even if it was returning an error.

I don't see any other instances in the error logs for this. Are you still seeing this problem or was it transient?

StructByLightning:
That's interesting. Trying to audit the page still gave me an error. If I clicked retry, again, same error. But if instead if I clicked cancel and went to the console, then went back and tried another audit, it worked. I checked to see if I could duplicate this behavior on a fresh tab but it worked perfectly the first time. There's definitely *something* wrong somewhere but if it can't be duplicated I guess there's nothing to be done.

wardpeet:
@exterkamp is this one fixed? Seems you created a pr for it.

exterkamp:
It is but it is not landed in production Chrome yet so audit panel users will still see this error for the time being.

exterkamp:
Closing, because this should be in stable Chrome M71+ (v3.2.0+).  We can reopen if this still comes up!

PR #6089, opened 2018-09-21
core(driver): add check to make sure Runtime.evaluate result exists

exterkamp:
**Summary**
Added an undefined check for driver results to catch an undefined result coming back from the Protocol, which shouldn't happen, but we check anyway to catch if it does.

[Relevant Protocol](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#method-evaluate)

**Related Issues/PRs**
addresses: #6026

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -316,7 +316,7 @@ class Driver {
 
         if (response.exceptionDetails) {
           // An error occurred before we could even create a Promise, should be *very* rare
-          return reject(new Error('Evaluation exception: ${response.exceptionDetails.text}'));
+          return reject(new Error(`Evaluation exception: ${response.exceptionDetails.text}`));
         }
 
         // Protocol should always return a 'result' object, but it is sometimes undefined.  See #6026.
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -311,15 +311,23 @@ class Driver {
...
           // An error occurred before we could even create a Promise, should be *very* rare
-          reject(new Error('an unexpected driver error occurred'));
-        } if (value && value.__failedInBrowser) {
-          reject(Object.assign(new Error(), value));
+          return reject(new Error(`Evaluation exception: ${response.exceptionDetails.text}`));
+        }
+
+        // Protocol should always return a 'result' object, but it is sometimes undefined.  See #6026.
+        if (response.result === undefined) {
+          return reject(new Error('Runtime.evaluate response omits a result'));
```
brendankenny: I know @paulirish just said this, but "omits" is weird here. At least "omitted", maybe "did not contain"?

exterkamp: I also felt like "a result" was confusing, so clarified it is a "result" object that wasn't there.

PR #6090, opened 2018-09-22
core: remove some trivial uses of WebInspector

brendankenny:
inspired by @patrickhulce's comment in https://github.com/GoogleChrome/lighthouse/pull/5946#issuecomment-423608200, took a look at some of our current uses of WebInspector and found some that could be trivially removed.

### DIFF
--- a/lighthouse-core/lib/web-inspector.js
+++ b/lighthouse-core/lib/web-inspector.js
@@ -32,10 +32,10 @@ module.exports = (function() {
 
   global.CSSAgent = {};
   global.CSSAgent.StyleSheetOrigin = {
-    INJECTED: 'injected',
-    USER_AGENT: 'user-agent',
-    INSPECTOR: 'inspector',
-    REGULAR: 'regular',
+    Injected: 'injected',
+    UserAgent: 'user-agent',
+    Inspector: 'inspector',
+    Regular: 'regular',
   };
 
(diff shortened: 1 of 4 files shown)


lighthouse-core/lib/web-inspector.js
```diff
@@ -32,10 +32,10 @@ module.exports = (function() {
...
   global.CSSAgent = {};
   global.CSSAgent.StyleSheetOrigin = {
-    INJECTED: 'injected',
-    USER_AGENT: 'user-agent',
-    INSPECTOR: 'inspector',
-    REGULAR: 'regular',
+    Injected: 'injected',
+    UserAgent: 'user-agent',
+    Inspector: 'inspector',
+    Regular: 'regular',
```
brendankenny: these shouldn't have been all caps (see e.g. https://github.com/ChromeDevTools/devtools-frontend/blob/51d3d758c57dc51d1864b3003de1fdda82963623/front_end/sdk/CSSRule.js#L70). We only touch the code that checks if a style was injected or came from the user-agent, but unsure if that was affecting the output of our font-size code...

brendankenny:
bundle size form 512.65KB back down to 510.01KB :)

brendankenny:
that's not a review :P

patrickhulce:
❤️ this

PR #6091, opened 2018-09-22
core: use cssstyle to parse CSS colors instead of WebInspector

brendankenny:
Another bit of WebInspector removed. Uses the CSS value parser (just to check if a string is a valid CSS color) from https://github.com/jsakas/CSSStyleDeclaration/blob/master/lib/parsers.js instead of the one in WebInspector.

bundle size is down another 1.4KB on top of #6090 after gzip :)


patrickhulce:
I'm guessing @brendankenny meant 1.4KB net after you factor in that we removed DevTools color? :)

PR #6099, opened 2018-09-24
core: replace WebInspector traceparser with native JSON.parse

brendankenny:
possibly more controversial than #6090 and #6091

The reason we [introduced a streaming trace parser](https://github.com/GoogleChrome/lighthouse/pull/2593) was because we would get `RangeError: Invalid string length` with [traces that were larger than 256MB](https://github.com/GoogleChrome/lighthouse/issues/1685#issuecomment-279115446), the previous maximum string size in V8.

This changed in V8 6.2, which increased the max size to 1GB. This landed in [Node 8.10.0](https://nodejs.org/en/blog/release/v8.10.0/#header-notable-changes) in March, which is after the original LTS release of 8.9. However:
- the LTS branch expects users to keep up with MINOR updates. [8.9.4](https://nodejs.org/en/blog/release/v8.9.4/) was the last release in the 8.9.x branch in January, and major security updates have happened in [8.11.0](https://nodejs.org/en/blog/release/v8.11.0/), [8.11.3](https://nodejs.org/en/blog/release/v8.11.3/), [8.11.4](https://nodejs.org/en/blog/release/v8.11.4/)
- trace parser is now only used for `-A` mode after #6056
- since 3.0, our traces have been significantly smaller (e.g. a full time-out verge run is more like 150MB)

It thus seems reasonable to expect at least Node 8.10 to run Lighthouse if we're supporting at least the LTS branch. I've updated the `package.json` `engines` entry to give a warning on install in `npm` (not sure what the yarn `engines` behavior is these days).


lighthouse-extension/app/src/lighthouse-ext-background.js
```diff
@@ -231,8 +231,8 @@ function isRunning() {
   return lighthouseIsRunning;
 }
 
-// Run when in extension context, but not in devtools.
-if ('chrome' in window && chrome.runtime) {
+// Run when in extension context, but not in devtools or unit tests.
+if (typeof window !== 'undefined' && 'chrome' in window && chrome.runtime) {
```
brendankenny: Before this PR, `lighthouse-ext-background.js` was pulling in `assetSaver`, which was pulling in `WebInspector`, which was setting `global.window = global`, so this file was fine to pull into node unit tests.

With that dependency path removed, this protects us from erroring on the `window` access (in the actual extension the behavior is unchanged since `window` is already defined :)

brendankenny:
removes another 6KB off our gzipped bundle (508.62KB -> 502.56KB) since we can remove some of the other `common/` and `sdk/` files in addition to the parser itself.

brendankenny:
oh and
- it's faster to just `JSON.parse` the string. Loading that 150MB trace takes about 2760ms on master for me, 1618ms after this PR

patrickhulce:
heh, I was going to suggest this too :D nice!!!

Issue #1535, opened 2017-01-25
Update to chrome-devtools-frontend >= 1.0.448737


paulirish:
This will fix the destroySSL bug that's described in https://github.com/GoogleChrome/lighthouse/issues/73#issuecomment-270574112

Might as well update to latest module version.

This upgrade will require some work as devtools dropped the "WebInspector" namespace, and now uses 40-ish global namespaces.  I've already done most of that work over in https://github.com/paulirish/devtools-timeline-model/tree/master/lib but it'll need porting to our use in `lib/web-inspector.js` etc

paulirish:
i have a WIP in `bump-devtools` branch for this.

paulirish:
now 1.0.448737 to also upgrade Gonzalez and fix #1412

XhmikosR:
@paulirish: any updates on this? It's been a long time we hit #1412.

wardpeet:
@paulirish mind if I look at your branch and try to get this working? ^^

paulirish:
@wardpeet go for it.

paulirish:
https://github.com/GoogleChrome/lighthouse/tree/bump-devtools has some work from feb 11th that you can start from. 

thanks for helping me out with this. :)

wardpeet:
@paulirish should we close this? as you mentioned this wasn't necessary anymore

paulirish:
good call!

PR #6101, opened 2018-09-24
core: update chrome-devtools-frontend to latest

brendankenny:
Update is easy now :) Fulfills the dream of #1535, though the problems it would have fixed back then don't exist anymore.

Hopefully this update will make it easier for whoever takes that last webinspector step, either completely isolating the WebInspector global pollution (cc @castilloandres) or reimplementing the font functionality in Lighthouse. With the updates to the fontend, the first should be quite a bit easier and the second should at least be at least a little easier to follow since the implementation will match current DevTools.

### DIFF
--- a/lighthouse-core/lib/web-inspector.js
+++ b/lighthouse-core/lib/web-inspector.js
@@ -7,48 +7,49 @@
 'use strict';
 
 /**
- * Stubbery to allow portions of the DevTools frontend to be used in lighthouse. `WebInspector`
+ * Stubbery to allow portions of the DevTools frontend to be used in lighthouse. `SDK`
  * technically lives on the global object but should be accessed through a normal `require` call.
  */
 module.exports = (function() {
-  if (global.WebInspector) {
-    return global.WebInspector;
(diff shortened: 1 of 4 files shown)


lighthouse-core/lib/web-inspector.js
```diff
@@ -7,64 +7,55 @@
...
-  global.WebInspector = {};
-  const WebInspector = global.WebInspector;
-
-  // Shared Dependencies
-  require('chrome-devtools-frontend/front_end/platform/utilities.js');
+  /**
+   * The single prototype augmentation needed from 'chrome-devtools-frontend/front_end/platform/utilities.js'.
+   * @return {Array<number>}
+   */
+  String.prototype.computeLineEndings = function() { // eslint-disable-line no-extend-native
```
brendankenny: seemed simpler (and more up-front about what we're doing) to explicitly provide the single augmentation we needed rather than all of `utilities.js`. Since we're up to date now, maybe we could upstream putting this on TextUtils or something, instead :)

PR #6107, opened 2018-09-26
cli: add --print-config flag

### LINKED ISSUE #1826: [proposal] harmonize command line flags and config file(s)

brendankenny:
finally the last step fixing #1826 (more or less :)

after `Config` takes the input config JSON and extends it, merges passes, concats audit arrays, filters for `--only-audits`, merges options, etc, it can be difficult to visualize exactly what they end result will be. This flag makes that simple by printing the full, normalized config to stdout.

The only real issue with JSON.stringifying the config is the live `implementation` properties in audits and gatherers (as well as `instance` in gatherers), which can't be stringified. These are skipped.

However, if audits/gatherers were specified with a `path` (like all the core configs do), the output of `--print-config` is a valid `LH.Config.Json` and can be c/p into a new config file and used to create an identical LH run.

Currently not implemented possibilities:
- if audits/gathers don't have a `path`, some kind of string replacement for the live `implementation` and `instance` properties to at least indicate that they were there (and throw an error if you pass it back into `Config`). Basically a more informative `[object Object]`. Live `implementation` seems mostly used for testing Lighthouse itself, though, so this may be necessary.
- absolute paths. Could help debug why e.g. an audit isn't being loaded due to a bad relative path. This limits the copy/pastability of the output, though, and we have pretty good error messages in `config.js` to help in these cases already

I've also eliminated any empty audit/gatherer `options` objects, as they're noisy in the output, they're all empty, and they're optional in a `LH.Config.Json` anyways (they're immediately replaced with the same `{}` when normalized).

Happy to bikeshed on anything!


lighthouse-cli/bin.js
```diff
@@ -94,34 +95,35 @@ if (cliFlags.extraHeaders) {
...
-            channel: 'cli',
-          },
-        },
-      });
+async function run() {
+  if (cliFlags.printConfig) {
+    const config = generateConfig(cliFlags, configJson);
+    process.stdout.write(config.print());
+    return;
+  }
```
brendankenny: this needed to be in the full flow of the cli since it turns out `process.stdout.write` (and by extension `console.log`) isn't flushed or waited on by `process.exit()`, so the exit can come mid write and the JSON output will be truncated. Since `index.js` just calls runs this module and calls `run()`, it needed some way to prevent getting to `run()` without calling `process.exit()`, so...here we are.

(and might as well switch to async/await while in here)

lighthouse-cli/bin.js
```diff
@@ -94,34 +95,35 @@ if (cliFlags.extraHeaders) {
...
 
-      return runLighthouse(url, cliFlags, config);
+  // By default, cliFlags.enableErrorReporting is undefined so the user is
+  // prompted. This can be overriden with an explicit flag or by the cached
+  // answer returned by askPermission().
+  if (typeof cliFlags.enableErrorReporting === 'undefined') {
+    cliFlags.enableErrorReporting = await askPermission();
+  }
+  if (cliFlags.enableErrorReporting) {
```
brendankenny: this check isn't strictly necessary (`Sentry.init` itself checks for `enableErrorReporting`), but added this just to reassure anyone reading this file and not digging in farther

lighthouse-cli/cli-flags.js
```diff
@@ -141,14 +142,18 @@ function getFlags(manualArgv) {
...
         //   - We're in auditMode (and we have artifacts already)
         //   - We're just listing the available options.
+        //   - We're just printing the config.
         // If one of these don't apply, stop the program and ask for a url.
         const isListMode = argv.listAllAudits || argv.listTraceCategories;
         const isOnlyAuditMode = !!argv.auditMode && !argv.gatherMode;
-        if (!isListMode && !isOnlyAuditMode && argv._.length === 0) {
-          throw new Error('Please provide a url');
+        const isPrintConfigMode = argv.printConfig;
+        if (isListMode || isOnlyAuditMode || isPrintConfigMode) {
```
paulirish: i think printConfigMode is actually captured under the spirit of the `isListMode` variable. should be renamed though.   `doPrintAndQuit` ?

lighthouse-core/index.js
```diff
@@ -48,15 +48,27 @@ async function lighthouse(url, flags = {}, configJSON, connection) {
...
 
+/**
+ * Generate a Lighthouse Config.
+ * @param {LH.Flags=} flags Optional settings for the Lighthouse run. If present,
+ *   they will override any settings in the config.
+ * @param {LH.Config.Json=} configJson Configuration for the Lighthouse run. If
+ *   not present, the default config is used.
+ * @return {Config}
+ */
+function generateConfig(flags, configJson) {
```
paulirish: the order here is unexpected to me. i know lighthouse's signature matches this order, but ... can we switch to make it the same as the Config constructor

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -315,4 +319,5 @@ module.exports = {
   createMessageInstanceIdFn,
   getFormatted,
   replaceIcuMessageInstanceIds,
+  MESSAGE_INSTANCE_ID_REGEX,
```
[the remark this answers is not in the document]
paulirish: add comment that it's exported only for testing?

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -265,8 +265,12 @@ function _resolveIcuMessageInstanceId(icuMessageInstanceId, locale) {
 }
 
 /**
- * @param {LH.Result} lhr
+ * Recursively walk the input object, looking for property values that are
+ * string references and replace them with their localized values. Primarily
+ * used with the full LHR as input.
+ * @param {*} lhr
```
paulirish: why not `{LH.Result}` ?

lighthouse-core/config/config.js
```diff
@@ -363,6 +366,45 @@ class Config {
     const configJson = this; // eslint-disable-line no-unused-vars
   }
 
+  /**
+   * Provides a cleaned-up, stringified version of this config. Gatherer and
+   * Audit `implementation` and `instance` do not survive this process.
+   * @return {string}
+   */
+  print() {
```
patrickhulce: 🚲 🏠 : `print` makes me think it will print it not just return the string :) 

`formatAsString` `getPrintString` `stringify` `getDisplayString`

any of those strike your fancy?

lighthouse-core/test/config/config-test.js
```diff
@@ -974,4 +974,76 @@ describe('Config', () => {
...
+        const i18nOrigTitle = i18n.getFormatted(origTitle, origConfig.settings.locale);
+
+        assert.strictEqual(printedCategory.title, i18nOrigTitle);
+      });
+
+      // Should have localized at least one string.
+      assert.ok(localizableCount > 0);
+    });
+
+    it('prints a valid ConfigJson that can make an identical Config', () => {
```
patrickhulce: ❤️ this test

lighthouse-cli/test/cli/index-test.js
```diff
@@ -67,4 +67,33 @@ describe('CLI Tests', function() {
       assert.equal(ret.status, 1);
     });
   });
+
+  describe('print-config', () => {
+    it('should print the default config and exit immediately after', () => {
+      const ret = spawnSync('node', [indexPath, '--print-config'], {encoding: 'utf8'});
+
+      const config = JSON.parse(ret.stdout);
+      assert.strictEqual(config.settings.output[0], 'html');
```
patrickhulce: I think this would actually be a good candidate for snapshot testing, but I know you're not a fan :)

lighthouse-cli/cli-flags.js
brendankenny: done

lighthouse-cli/test/cli/index-test.js
brendankenny: > good candidate for snapshot testing, but I know you're not a fan

I'd make my opinion more nuanced in that if there's constant snapshot churn they lose efficacy as a unit test because we stop paying attention, but I agree this is a good candidate (default config shouldn't change too much...) and less awkward than this style of checking.

We do lose the immediacy of the 'html' here vs the 'json' in the test below (and `auditMode` true vs false), but that might be ok(ish).

I'll add it.

brendankenny: What I'm not a fan of is `expect(result).to.not.be.checked.in.a.ridiculous('manner')` :P:P

lighthouse-core/config/config.js
brendankenny: going with `getPrintString` unless anyone has other votes :) I don't care a whole lot, though it would be nice if whatever we choose communicates that the string result was functional in nature, not just pretty printing or something

lighthouse-core/lib/i18n/i18n.js
brendankenny: > why not `{LH.Result}` ?

I pass in the config now to get a translated one. I can rename the param to make that more obvious

lighthouse-cli/test/cli/index-test.js
patrickhulce: haha alright, very fair. jest isn't big on the dot chains though! at least not nearly as bad as chai

lighthouse-core/lib/i18n/i18n.js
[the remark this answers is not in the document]
brendankenny: > add comment that it's exported only for testing?

I added an `isIcuMessage()` instead, just because we do the same check a couple of times internally, so nice in there and nicely encapsulated externally

PR #6110, opened 2018-09-26
report(psi): remove redundant varience disclaimer

paulirish:
Right now, "Values are estimated and may vary" is shown twice in the PSI case.

image

### DIFF
--- a/lighthouse-core/report/html/renderer/performance-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/performance-category-renderer.js
@@ -112,7 +112,7 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
   /**
    * @param {LH.ReportResult.Category} category
    * @param {Object<string, LH.Result.ReportGroup>} groups
-   * @param {string=} environment 'PSI' and undefined are the only valid values
+   * @param {'PSI'=} environment 'PSI' and undefined are the only valid values
    * @return {Element}
    * @override
    */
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -144,9 +144,13 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     otherMetrics.forEach(item => {
       metricsColumn2El.appendChild(this._renderMetric(item));
     });
-    const estValuesEl = this.dom.createChildOf(metricsColumn2El, 'div',
-        'lh-metrics__disclaimer lh-metrics__disclaimer');
-    estValuesEl.textContent = Util.UIStrings.varianceDisclaimer;
+
+    // 'Values are estimated and may vary' is used as the category description for PSI
+    if (environment !== 'PSI') {
```
patrickhulce: Is this being covered by tsc right now? :)

paulirish: @patrickhulce whatcha mean

patrickhulce: meaning is the magic string  `'PSI'` being checked to make sure we're using it correctly vs. `psi`, typo `PS1` etc

exterkamp:
Would it make more sense to have it removed from the category descriptor so that it is more consistent across environments?

paulirish:
@exterkamp perhaps. but in this case it's a little different. for LS, we have field data at the top and this disclaimer doesnt apply. and then we have the category description that says "hi all this lab data is from lighthouse. it's different data so here's the disclaimer."

whereas in typical LH, we dont have a category description right now. but when we add one, it could make sense to move this disclaimer there. 


@patrickhulce sure yeah. we could but so far we've avoided hiding with CSS and building the DOM how we want to use it.

Issue #3596, opened 2017-10-17
Reduce the size of the trace


patrickhulce:
Tracing causes many sites on WPT to crash Chrome on the phone. Reducing the required number of categories or streaming these few necessary events is the only path forward to making these failure go away.
  

~Lantern relies heavily on quite a few trace events, isolating the key events necessary, and migrating those to Audits.* domain events to eliminate the need for tracing could be helpful since a majority of time spent is in trace parsing~

patrickhulce:
We might want to revisit priority of this since it's responsible for ~10% failure rate on WPT

tdresser:
This feels like it's moving towards a world where catapult and lighthouse are less similar, which is seems like a step backwards to me.

Can we accomplish the same goal by splitting some trace categories into finer grained sub-categories?

patrickhulce:
The long term goal was to enable many of Lighthouse insights on all page runs, not just ones where we can capture a trace. In a way, this is much more aligned with native implementation of TTI.

The short-term goal and realistic version of this is exactly what you describe, moving/copying critical trace events to their own smaller category such that we don't need everything under the sun.

tdresser:
Sound good. For metrics like TTI in particular, long term we hope to just spit out the final values into the trace. Perhaps these final values should be in their own trace category?

paulirish:
> Additionally, tracing causes many sites on WPT to crash Chrome on the phone.

We're also interested in adding another category to the `TaskQueueManager::ProcessTaskFromWorkQueue` trace event (within Chromium) so that Lighthouse doesn't need to request the `toplevel` category, which is usually most of the size of the trace: 
![image](https://user-images.githubusercontent.com/39191/33146986-0c8088de-cf7b-11e7-8e4d-34805f3e8adb.png)

tdresser:
What fraction of toplevel is `TaskQueueManager::ProcessTaskFromWorkQueue`? What else is in toplevel? This might make a bunch of our metrics analysis much easier too.

tdresser:
Should the title of this issue be something like "Shrink set of required trace events", or similar?

paulirish:
> What fraction of toplevel is TaskQueueManager::ProcessTaskFromWorkQueue? 

Here's a 74 mb trace of theverge loading with lighthouse's categories:

![image](https://user-images.githubusercontent.com/39191/33283585-2726d2ee-d361-11e7-8b21-8a996ec1f187.png)

[...]

PR #6117, opened 2018-09-26
core(driver): request smaller trace in m71+

paulirish:
Backstory in #3596. 

Results: cnn.com's trace goes from 69M to 47M.

--

* Using `Browser.getVersion` and parsing out the chrome milestone. We can augment this pattern later for `isMobile` and `os` like @patrickhulce suggested.
* I made a lil sendCommand mock response helper in driver-test. There are similar stubs for `on()` and `once()`. 
* we now default to the new `disabled-by-default-lighthouse` category and fallback to 'toplevel', as @brendankenny very wisely suggested
* ideally i'd like to drop the `SCHEDULABLE_TASK_TITLE` possibilities from _FOUR_ down to 2. But it requires updating a lot of fixture data. So we can do that in a followup.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -918,11 +918,10 @@ class Driver {
   }
 
   /**
-   * @param {LH.Gatherer.PassContext} passContext
+   * @param {{additionalTraceCategories?: string|null}=} settings
    * @return {Promise<void>}
    */
-  async beginTrace(passContext) {
-    const settings = passContext.settings;
+  async beginTrace(settings) {
     const additionalCategories = (settings && settings.additionalTraceCategories &&
         settings.additionalTraceCategories.split(',')) || [];
     const traceCategories = this._traceCategories.concat(additionalCategories);
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -83,7 +83,7 @@ class Driver {
   static get traceCategories() {
     return [
       '-*', // exclude default
-      'toplevel',
+      'disabled-by-default-lighthouse', // used instead of 'toplevel' in Chrome 71+
       'v8.execute',
       'blink.console',
       'blink.user_timing',
@@ -102,11 +102,13 @@ class Driver {
   }
 
   /**
-   * @return {Promise<string>}
+   * @return {Promise<LH.Crdp.Browser.GetVersionResponse & {milestone: number}>}
    */
-  getUserAgent() {
-    // FIXME: use Browser.getVersion instead
-    return this.evaluateAsync('navigator.userAgent');
+  async getBrowserVersion() {
+    const version = await this.sendCommand('Browser.getVersion');
+    const match = version.product.match(/\/(\d+)/);
+    const milestone = match ? parseInt(match[1]) : 0;
+    return Object.assign(version, {milestone});
   }
 
   /**
@@ -919,19 +921,18 @@ class Driver {
    * @param {LH.Gatherer.PassContext} passContext
    * @return {Promise<void>}
    */
-  beginTrace(passContext) {
+  async beginTrace(passContext) {
     const settings = passContext.settings;
     const additionalCategories = (settings && settings.additionalTraceCategories &&
         settings.additionalTraceCategories.split(',')) || [];
     const traceCategories = this._traceCategories.concat(additionalCategories);
 
-    // In Chrome 71+, we trade the chatty 'toplevel' cat for our own, reducing overall trace size
-    // TODO(COMPAT): Once m71 ships to stable, change the traceCategories for real
-    const reChrome71 = /Chrome\/(Headless)?7[1-9]/;
-    if (passContext.hostUserAgent && reChrome71.test(passContext.hostUserAgent)) {
-      traceCategories.forEach((cat, idx) => {
-        if (cat === 'toplevel') traceCategories[idx] = 'disabled-by-default-lighthouse';
-      });
+    // In Chrome <71, gotta use the chatty 'toplevel' cat instead of our own.
+    // TODO(COMPAT): Once m71 ships to stable, drop this section
+    const milestone = (await this.getBrowserVersion()).milestone; // eg 'Chrome/71.0.3577.0'
+    if (milestone < 71) {
+      const toplevelIndex = traceCategories.indexOf('disabled-by-default-lighthouse');
+      traceCategories.splice(toplevelIndex, 1, 'toplevel');
     }
 
     const uniqueCategories = Array.from(new Set(traceCategories));
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -916,13 +916,24 @@ class Driver {
   }
 
   /**
-   * @param {{additionalTraceCategories?: string|null}=} settings
+   * @param {LH.Gatherer.PassContext} passContext
    * @return {Promise<void>}
    */
-  beginTrace(settings) {
+  beginTrace(passContext) {
+    const settings = passContext.settings;
     const additionalCategories = (settings && settings.additionalTraceCategories &&
         settings.additionalTraceCategories.split(',')) || [];
     const traceCategories = this._traceCategories.concat(additionalCategories);
+
+    // In Chrome 71+, we trade the chatty 'toplevel' cat for our own, reducing overall trace size
+    // TODO(COMPAT): Once m71 ships to stable, change the traceCategories for real
+    const reChrome71 = /Chrome\/(Headless)?7[1-9]/;
+    if (passContext.hostUserAgent && reChrome71.test(passContext.hostUserAgent)) {
+      traceCategories.forEach((cat, idx) => {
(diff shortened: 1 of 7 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -402,6 +401,7 @@ class GatherRunner {
           // If the main document redirects, we'll update this to keep track
           url: options.requestedUrl,
           settings: options.settings,
+          hostUserAgent: baseArtifacts.HostUserAgent,
```
patrickhulce: I was parsing the user agent over in #5893, maybe after this we should do a refactor for some useful `host.chromeVersion`, `host.isMobile`, `host.os`?

patrickhulce:
I'm mildly nervous about missing some random lantern-impacting event, but I think we've both been through it multiple times before.

paulirish:
ptal

lighthouse-core/gather/driver.js
```diff
@@ -919,10 +921,19 @@ class Driver {
...
    */
-  beginTrace(settings) {
+  async beginTrace(settings) {
     const additionalCategories = (settings && settings.additionalTraceCategories &&
         settings.additionalTraceCategories.split(',')) || [];
     const traceCategories = this._traceCategories.concat(additionalCategories);
+
+    // In Chrome <71, gotta use the chatty 'toplevel' cat instead of our own.
+    // TODO(COMPAT): Once m71 ships to stable, drop this section
+    const milestone = (await this.getBrowserVersion()).milestone; // eg 'Chrome/71.0.3577.0'
```
patrickhulce: it feels like this comment should be up where the regex is that parses out `71` :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -46,9 +48,23 @@ function createActiveWorker(id, url, controlledClients, status = 'activated') {
   };
 }
 
+function createOnceMethodResponse(method, response) {
```
patrickhulce: someday using jests mock functions would be awesome :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -232,16 +249,34 @@ describe('Browser Driver', () => {
...
+    // m70 doesn't have disabled-by-default-lighthouse, so 'toplevel' is used instead.
+    const m70BrowserVersion = Object.assign({}, browserVersion, {
+      product: 'Chrome/70.0.3577.0',
+      // eslint-disable-next-line max-len
+      userAgent: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3577.0 Safari/537.36',
+    });
+    createOnceMethodResponse('Browser.getVersion', m70BrowserVersion);
+
+    // eslint-disable-next-line max-len
+    return driverStub.beginTrace().then(() => {
```
patrickhulce: want to use `async/await` for new code?

lighthouse-core/test/gather/fake-driver.js
```diff
@@ -5,9 +5,18 @@
  */
 'use strict';
 
-module.exports = {
-  getUserAgent() {
-    return Promise.resolve('Fake user agent');
+const browserVersion = {
+  protocolVersion: '1.3',
```
patrickhulce: this doesn't return `milestone`. Is this filename wrong and it's really not a fake driver but a fake connection?

paulirish: good catch. yeah the separation between protocol response vs driver method response got a little mixed up. 

Clarified this now.

lighthouse-core/test/gather/driver-test.js
paulirish: yeah our lame driver/connection mocking everywhere is ripe for a cleanup.

lighthouse-core/gather/driver.js
```diff
@@ -919,10 +921,19 @@ class Driver {
...
     const additionalCategories = (settings && settings.additionalTraceCategories &&
         settings.additionalTraceCategories.split(',')) || [];
     const traceCategories = this._traceCategories.concat(additionalCategories);
+
+    // In Chrome <71, gotta use the chatty 'toplevel' cat instead of our own.
+    // TODO(COMPAT): Once m71 ships to stable, drop this section
+    const milestone = (await this.getBrowserVersion()).milestone;
+    if (milestone < 71) {
+      const toplevelIndex = traceCategories.indexOf('disabled-by-default-lighthouse');
+      traceCategories.splice(toplevelIndex, 1, 'toplevel');
```
brendankenny: seems like the splice is unnecessary?

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -8,12 +8,16 @@
...
+const SCHEDULABLE_TASK_TITLE_LH = 'RunTask';
+// m69-70 DoWork is different and we now need RunTask, see https://bugs.chromium.org/p/chromium/issues/detail?id=871204#c11
+const SCHEDULABLE_TASK_TITLE_ALT1 = 'ThreadControllerImpl::RunTask';
 // In m66-68 refactored to this task title, https://crrev.com/c/883346
-const SCHEDULABLE_TASK_TITLE_ALT1 = 'ThreadControllerImpl::DoWork';
-// m69+ DoWork is different and we now need RunTask, see https://bugs.chromium.org/p/chromium/issues/detail?id=871204#c11
-const SCHEDULABLE_TASK_TITLE_ALT2 = 'ThreadControllerImpl::RunTask';
+const SCHEDULABLE_TASK_TITLE_ALT2 = 'ThreadControllerImpl::DoWork';
+// m65 and earlier
+const SCHEDULABLE_TASK_TITLE_ALT3 = 'TaskQueueManager::ProcessTaskFromWorkQueue';
```
brendankenny: have you thought about a long-term support plan for these? Supporting old trace versions seems pretty good if it's not much work, but otoh our test traces are increasingly not looking like traces from master

lighthouse-core/test/gather/driver-test.js
```diff
@@ -46,9 +48,23 @@ function createActiveWorker(id, url, controlledClients, status = 'activated') {
   };
 }
 
+function createOnceMethodResponse(method, response) {
+  assert.deepEqual(!!sendCommandMockResponses[method], false, 'stub response already defined');
```
brendankenny: `deepEqual` left over from an earlier revision?

lighthouse-core/test/gather/driver-test.js
```diff
@@ -46,9 +48,23 @@ function createActiveWorker(id, url, controlledClients, status = 'activated') {
...
+  sendCommandMockResponses[method] = response;
+}
+
 connection.sendCommand = function(command, params) {
   sendCommandParams.push({command, params});
+
+  const mockResponse = sendCommandMockResponses[command];
+  if (mockResponse) {
+    delete sendCommandMockResponses[command];
```
brendankenny: should use a Map for things like this

lighthouse-core/gather/driver.js
paulirish: `traceCategories[toplevelIndex] = 'toplevel'` sgtm

lighthouse-core/lib/traces/tracing-processor.js
paulirish: Yes I'd like to redo all our fixture traces. Right now it doesn't seem _too_ bad.. 
.. I _just_ tested the perf downside of checking the evt name four times and its probably a maximum of ~10ms total on a big trace. So we can ignore that bit.

I think it's aiight for now. But we'll put something in the OKRs about freshening up our fixtures/mocking/test infra.

lighthouse-core/test/gather/driver-test.js
paulirish: I wrote the method assuming someone else may try to use it. And just wanted to assert no one was going to use it twice.
I can remove though if you don't see the value.

lighthouse-core/test/gather/driver-test.js
paulirish: I dont see why Map is better, but happy to change it. :)

paulirish:
> since RunTask is the future, we should also update create-test-trace.js to use it

sg will do.

lighthouse-core/lib/traces/tracing-processor.js
brendankenny: yeah, definitely not a performance issue (string compares are fast, especially if they haven't ever been modified), and since the events all more or less cover the subset of events we're interested in (we only have to do the string check, not more convoluted fixes to support each), it's more a case of being able to easily see what the state of the world is in the code.

The downside for updating fixtures (besides being really annoying to do) is we get less coverage of the old versions, so we may want to really phase out support completely for some of these if we actually do that.

lighthouse-core/test/gather/driver-test.js
brendankenny: > I can remove though if you don't see the value.

sorry, I think it's fine, I meant the `deepEqual` is kind of overkill because they're both booleans :)

PR #6127, opened 2018-09-27
misc: tweaks for devtools client

paulirish:
1. match score scale order to LS. (green, orange, red)
2. negative margin of metrics is now scaled against circle-size, which is the intent. 
3. expose lh-background methods appropriately in a worker


brendankenny:
ok, yes, I see the error too

PR #6129, opened 2018-09-27
misc: add cli module diagram

paulirish:


lighthouse-cli/bin.js
```diff
@@ -8,6 +8,19 @@
 const fs = require('fs');
 const path = require('path');
 
+/*
+ * The relationship between these CLI modules:
```
brendankenny: should this be in `index.js` since that's likely the place you'd look first, or here, as the "real" beginning

lighthouse-cli/bin.js
```diff
@@ -8,6 +8,19 @@
...
+ *
+ *   index.js     : only calls bin.js's begin()
+ *   cli-flags.js : leverages yargs to read argv, outputs LH.CliFlags
+ *   bin.js       : CLI args processing. cwd, list/print commands
+ *   run.js       : chrome-launcher bits, calling lighthouse-core, output to Printer
```
patrickhulce: part of me thinks the functionality here should be available from node more easily too, but that's tomorrow's problem :)

PR #6138, opened 2018-09-28
tests: add angular test for JS minification

patrickhulce:
quick followup to #5925

### DIFF
--- a/lighthouse-core/test/lib/minification-estimator-test.js
+++ b/lighthouse-core/test/lib/minification-estimator-test.js
@@ -5,8 +5,14 @@
  */
 'use strict';
 
+const fs = require('fs');
+const path = require('path');
 const assert = require('assert');
 const {computeCSSTokenLength, computeJSTokenLength} = require('../../lib/minification-estimator');
+
+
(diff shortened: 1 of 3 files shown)


lighthouse-core/test/lib/minification-estimator-test.js
```diff
@@ -5,9 +5,15 @@
  */
 'use strict';
 
+const fs = require('fs');
+const path = require('path');
 const assert = require('assert');
 const {computeCSSTokenLength, computeJSTokenLength} = require('../../lib/minification-estimator');
 
+
+const nodeModulesPath = path.join(__dirname, '../../../node_modules');
```
paulirish: `require.resolve` can handle all sorts of funky path issues this'll hit.

```js
const angularJSPath = path.join(require.resolve('angular'), 'angular.js');
const angularFullScript = fs.readFileSync(angularJSPath, 'utf8');
```

patrickhulce: good call. I'll do ya one better, `require.resolve('angular/angular.js')` ;)

PR #6139, opened 2018-09-28
core: throttling=none, formFactor=provided

patrickhulce:
OK, so I don't really believe in this change so much, but I think it's a good forum to discuss what we actually want to happen.

**Backstory**

When we were creating the `throttlingMethod` values I pushed for `provided` instead of `none` to apply a psychological element to the idea that disabling throttling altogether will start to degrade LH performance results and saying "I am now responsible for throttling" might discourage folks from casually messing with throttling method. 'provided' though is functionally identical to 'none' from our perspective.

In emulated form factor, I see no such strong incentive to discourage folks and 'none' seems fine to me.

There is now a disparity between them.

**This PR**

Adds 'provided' as an alias for 'none' to emulatedFormFactor.
Adds 'none' as an alias for 'provided' to throttlingMethod.

**Alternatives**
* Standardize on 'none' (with temporary backcompat)
* Standardize on 'provided' (with temporary backcompat)
* Leave as-is

Thoughts?

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -122,8 +122,8 @@ function getFlags(manualArgv) {
         'list-trace-categories', 'view', 'verbose', 'quiet', 'help', 'print-config',
       ])
       .choices('output', printer.getValidOutputOptions())
-      .choices('emulated-form-factor', ['mobile', 'desktop', 'none'])
-      .choices('throttling-method', ['devtools', 'provided', 'simulate'])
+      .choices('emulated-form-factor', ['mobile', 'desktop', 'provided', 'none'])
+      .choices('throttling-method', ['devtools', 'provided', 'simulate', 'none'])
       .choices('preset', ['full', 'perf', 'mixed-content'])
       // force as an array
       // note MUST use camelcase versions or only the kebab-case version will be forced
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -464,6 +464,10 @@ class Config {
     // Locale is special and comes only from flags/settings/lookupLocale.
     settingsWithFlags.locale = locale;
 
+    // Normalize throttlingMethod
+    settingsWithFlags.throttlingMethod = settingsWithFlags.throttlingMethod === 'none' ?
+      'provided' : settingsWithFlags.throttlingMethod;
+
     return settingsWithFlags;
   }
 
--- a/typings/externs.d.ts
+++ b/typings/externs.d.ts
@@ -87,8 +87,8 @@ declare global {
       gatherMode?: boolean | string;
       disableStorageReset?: boolean;
       disableDeviceEmulation?: boolean;
-      emulatedFormFactor?: 'mobile'|'desktop'|'none';
-      throttlingMethod?: 'devtools'|'simulate'|'provided';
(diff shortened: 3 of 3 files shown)


patrickhulce:
@benschwarz especially curious on your thoughts since you had some good feedback on how confusing emulatedFormFactor was :)

patrickhulce:
> ha, I think you've convinced me we should keep the status quo

❤️

exterkamp:
To me personally, from the outside, `provided` is a little bit confusing.  I think `none` is more clear.  But that's just my 2 cents. Aliasing seems fine though, let people use what makes sense to them.

patrickhulce:
> provided is a little bit confusing

I suppose it's somewhat WAI then 😆 

My hesitation with aliases is two-fold.

1. We lose an element of friction that discourages casually disabling throttling.
2. It introduces a new confusing element of "well what's the difference between provided and none" that makes me think if we're already losing the reason for 'provided' existing in 1, why bother with it shouldn't we just go with only 'none'.

Counterpoint to my hesitations is "Patrick, this friction is silly because it only applies to CLI users who are probably a bit more advanced anyway and we already have a big fat `No throttling` option in DevTools".

My counter-counterpoint is "Yeah, but if we can at least keep encouraging the power users to do advanced, more accurate throttling maybe we'd all live in a better world" 😃

benschwarz:
Hey @patrickhulce, thanks for pinging me. I think these settings are particularly confusing for one main reason -- I can't tell how the flags will interact with each other: 

* If I use `disableDeviceEmulation` or `disableCpuThrottling` does that mean that my custom set `settings.throttling.cpuSlowdownMultiplier` will work? Or will that get no-oped out?
* If I use `emulatedFormFactor: mobile` to indicate that the device is a mobile, will that stomp over other settings? (Answer: No, it means be a nexus 5… which is kind of confusing)
* `emulatedFormFactor: desktop` seems like it'd be the default, fast machine configuration (maybe with a slightly throttled network?)

patrickhulce:
Thanks! Tried to answer your questions so maybe with the clarity you have suggestions for names :D

`disableDeviceEmulation` has always just affected the viewport + user agent, so cpuSlowdownMultiplier will not be affected. disableCpuThrottling isn't a thing in v3, so also not affecting :)

`emulatedFormFactor` is, like `disableDeviceEmulation`, exclusively about the viewport + user agent. It's just meant to replace `disableDeviceEmulation` with a `select` instead of a `checkbox` :)

> emulatedFormFactor: desktop seems like it'd be the default, fast machine configuration (maybe with a slightly throttled network?

Here, desktop is again just referring to the viewport and user agent, not any other changes. I'm curious are you hoping there are bundled up presets that control lots of flags at once? Or do you not want there bundled up presets that control lots of flags at once and are just worried about accidentally using them? :)


> I think it'd be useful to have some kind of logging that confirms the settings that are being used at the beginning of the test - it'd give faster feedback. eg: "Setting viewport to 1300xXXX", "Slowing CPU by Xx".

Neat idea! We could take the runtime settings section printed at the bottom of the report and come up with a plain text version.

patrickhulce:
@paulirish any opinions here?

patrickhulce:
This is one of the stagnant PRs I mentioned today :)

I think the current situation is...

- I feel we can close this with no action taken. 
- @brendankenny seems somewhat convinced of closing this with no action needed. 
- @exterkamp thinks `none` is more clear.
- others ???

Issue #4516, opened 2018-02-12
Idea: add toggle to report to filter all flagged resources to same-origin only / 3rd party / 3P


patrickhulce:
In audits like passive event listeners (#4434) we used to restrict to same-origin URLs but nearly all recent audits flag all resources. It might be nice to add the filtering at the report level instead so you can get a view with "easiest to fix first" and remove it from the audit level across the board.

cc @hwikyounglee in case you have thoughts here :)

hwikyounglee:
Idea - add an indication of source location
![3p](https://user-images.githubusercontent.com/7561395/46123897-c90bf900-c1d5-11e8-9f8b-e31971eda4ec.png)

patrickhulce:
didn't mean to steal this from you @wardpeet! just wanted to investigate what we'll run into before it was assigned to anyone. you're free to still handle it if you'd like :)

wardpeet:
Haha, no worries feel free to take it

patrickhulce:
I really think you should take this back @wardpeet :)

I don't think I'm going to get to it immediately, and it's a nice juicy important piece for ya 😃

wardpeet:
recap before I start:
We will add a toggle to filter out all third party requests from our tables with a toggle button but we won't touch the scores.

This is what i'll add:
1) a toggle button at the top that just hides all tables
2) a toggle button for each table (audit) that has need of it.
3) an indicator in each table to show that we have marked it as third party (i'm thinking of an icon with a tooltip instead of text (wdyt @hwikyounglee ?)
4) We consider a request a third party when it doesn't matches it's root domain, am I correct?

🎉 thanks patrick for the PR!

@patrickhulce @paulirish @brendankenny did I miss anything?

brendankenny:
fixed in #6351!

PR #6142, opened 2018-09-28
[SPIKE] report: add 3rd party filter toggle

patrickhulce:
**Summary**
Once again just a platform for discussion, not a real PR :) I noticed 3p was P1, unassigned, and coming up yet again in issues so I wanted to get a handle on what are the major challenges we'll face.

**The Ultra-Hacky Prototype**

image

**The Cool**
- It seems pretty feasible to do it 100% in the report extra UI features without touching core at all.
- Our shared table structure made adding it everywhere super easy 🎉 

**The Challenges**

- We knew this already, but we won't be able to easily update score, labels saying "N resources found", etc. My "solution" to this is removing the rows but adding a single 3rd party placeholder row to remind folks the totals still include all that junk.
- Our nice CSS odd/even strategy doesn't work with just hiding the divs :(
- There are a few URL tables that don't count and special handling might be a bit awkward (i.e. preconnect to 3rd party origins obvs shouldn't be hiding the 3rd party origins)
- Placement in the score header that comes down feels weird, but it seems potentially real bad if we just hide it all the way at the top/another place a user has to scroll to.

**Other Ideas**
- If we were feeling extremely bold, third party resources would *always* start collapsed and you could click to expand them. in each table
- We could use a whitelist of third parties instead of throwing everything that doesn't match the origin.
- Hwi had a cool mock that just increases the visibility of each element as a 3rd party origin

**Related Issues/PRs**
#4516

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -85,6 +85,7 @@ class ReportUIFeatures {
 
     this.json = report;
     this._setupMediaQueryListeners();
+    this._setupThirdPartyFilter();
     this._setupExportButton();
     this._setUpCollapseDetailsAfterPrinting();
     this._setupHeaderAnimation();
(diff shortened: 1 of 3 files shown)


patrickhulce:
@wardpeet don't feel the need to re-use any of this, but you certainly can if you'd like

Issue #6150, opened 2018-10-01
Helper scripts not found when installed through npm


r0mflip:
#### Provide the steps to reproduce
``` bash
$ npm i -g lighthouse
$ npm explore -g lighthouse -- ./lighthouse-core/scripts/download-chrome.sh
```

Result:
```
/bin/bash: ./lighthouse-core/scripts/download-chrome.sh: No such file or directory
npm ERR! file /bin/bash
npm ERR! code ENOENT
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! enoent spawn ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
npm ERR! enoent Check if the file '/bin/bash' is present.
```

#### What is the current behavior?
Scripts like `download-chrome.sh` cannot be found in the `scripts` directory. I've also downloaded the tar file from https://registry.npmjs.org/lighthouse/-/lighthouse-3.2.1.tgz and can't find any scripts inside it.

#### What is the expected behavior?
I use [this technique](https://github.com/GoogleChrome/devsummit/blob/e1aec9095abe9fdf89f7d95968de8634372b432e/.travis.yml#L12-L18) by @aerotwist in a [similar way](https://github.com/cbitosc/cbitosc.github.io/blob/184c9bfb0f9d5fcfbad33a7931e259700303d96b/.travis.yml#L4-L13) to test my site.

It worked fine for some time(ref: https://travis-ci.org/cbitosc/cbitosc.github.io/builds/433659287, uses lighthouse 3.1.1) but then started to fail (ref: https://travis-ci.org/cbitosc/cbitosc.github.io/builds/433965520 uses lighthouse 3.2.0).

Same behavior persists on my local machine too.

#### Environment Information
* Affected Channels: CLI
* Lighthouse version: 3.2.1
* Node.js version: v10.11.0
* Operating System: Ubuntu 18.04

paulirish:
Ah. This makes sense. we just changed out npmignore to exclude this script.

Yeah we can reintroduce this one. Thanks.

r0mflip:
Closed in #6153

PR #6153, opened 2018-10-02
chore: 🤖 add back lighthouse-core scripts to npm package

r0mflip:
Issues: #6150

**Summary**
Solves #6150

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,7 +7,7 @@ assets/
 build/
 coverage/
 
-lighthouse-core/script/*
+lighthouse-core/scripts/*
 lighthouse-core/test/
 lighthouse-core/third_party/src/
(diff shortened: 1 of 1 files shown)


r0mflip:
I signed it!

.npmignore
```diff
@@ -7,7 +7,6 @@ assets/
 build/
 coverage/
 
-lighthouse-core/scripts/*
```
paulirish: below a bit is this line:

```
!lighthouse-core/scripts/manual-chrome-launcher.js
```

can you add a direct refrence to the download-chrome sh right after it?

i'd rather do that than allowlist this whole folder.

r0mflip: That sounds good, thanks. Please verify.

Issue #6145, opened 2018-09-30
DevTools Error: Failed to construct 'URL': Invalid URL


santamaria7:
**Initial URL**: https://www.bbc.com/sport
**Chrome Version**: 69.0.3497.100
**Error Message**: Failed to construct 'URL': Invalid URL
**Stack Trace**:
```
TypeError: Failed to construct 'URL': Invalid URL
    at new URLShim (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:4878:1)
    at NetworkRequest.onRequestWillBeSent (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1405:67)
    at NetworkRecorder.onRequestWillBeSent (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1384:151)
    at NetworkRecorder.dispatch (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1394:66)
    at devtoolsLog.forEach.message (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1397:126)
    at Array.forEach (<anonymous>)
    at Function.recordsFromLogs (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1397:93)
    at Function.afterPass (chrome-devtools://devtools/remote/serve_file/@8920e690dd011895672947112477d10d5c8afb09/audits2_worker/audits2_worker_module.js:1087:163)
```

patrickhulce:
Hm, I'm not able to reproduce this, but it clearly looks like others have (#6148). Are you still experiencing this @santamaria7 ?

santamaria7:
@patrickhulce  Yes, I am. I ran the Audit on https://www.bbc.com/sport just now for another time and ran into this error again. Because of this, the Audit never finishes and I have to cancel it.

drewmcmillan:
Raising this internally with BBC Sport aswell, apologies

drewmcmillan:
Raising this internally with BBC Sport, apologies

[...]

PR #6157, opened 2018-10-02
core(network): handle invalid NetworkRequests

exterkamp:
**Summary**
Adds `isValid` property to a NetworkRequest and ignores invalid requests when making the list of records to surface to rest of application.

**Related Issues/PRs**
fixes: #6145

### DIFF
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -339,8 +339,8 @@ class NetworkRecorder extends EventEmitter {
     // playback all the devtools messages to recreate network records
     devtoolsLog.forEach(message => networkRecorder.dispatch(message));
 
-    // get out the list of records
-    const records = networkRecorder.getRecords();
+    // get out the list of records & filter out invalid records
+    const records = networkRecorder.getRecords().filter(record => record.isValid);
 
     // create a map of all the records by URL to link up initiator
     const recordsByURL = new Map();
--- a/lighthouse-core/lib/network-recorder.js
+++ b/lighthouse-core/lib/network-recorder.js
@@ -320,7 +320,7 @@ class NetworkRecorder extends EventEmitter {
    */
   _findRealRequest(requestId) {
     let request = this._recordsById.get(requestId);
-    if (!request) return undefined;
+    if (!request || !request.isValid) return undefined;
 
     while (request.redirectDestination) {
       request = request.redirectDestination;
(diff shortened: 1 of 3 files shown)


lighthouse-core/lib/network-recorder.js
```diff
@@ -339,8 +339,8 @@ class NetworkRecorder extends EventEmitter {
     // playback all the devtools messages to recreate network records
     devtoolsLog.forEach(message => networkRecorder.dispatch(message));
 
-    // get out the list of records
-    const records = networkRecorder.getRecords();
+    // get out the list of records & filter out invalid records
+    const records = networkRecorder.getRecords().filter(record => record.isValid);
```
patrickhulce: you might want to add `if (!request || !request.isValid)` to line 323 above too, behavior of the other `on*` methods is unexpected if all the properties are unset

exterkamp: Good call.  Fixed!

lighthouse-core/lib/network-request.js
```diff
@@ -125,27 +126,31 @@ module.exports = class NetworkRequest {
...
-    this.requestMethod = data.request.method;
-
-    this.initiator = data.initiator;
-    this.resourceType = data.type && RESOURCE_TYPES[data.type];
-    this.priority = data.request.initialPriority;
-
-    this.frameId = data.frameId;
-    this.isLinkPreload = data.initiator.type === 'preload' || !!data.request.isLinkPreload;
+    try {
+      // try to construct the url and fill in request
```
brendankenny: Should we unconditionally catch all errors in here? Admittedly there's not anything else in here that will throw if our type checking is correct, but we might add more functionality in the future and any errors will fail silently.

What do you think about limiting the try/catch to the url constructor?

Like
```js
let url;
try {
  // try to construct the url and fill in request
  url = new URL(data.request.url);
} catch (e) {
  return; // isValid left false, all other data is blank
}
this.isValid = true;
this.url = data.request.url;
// ...
```

exterkamp: Yup that was my first instinct!  Reverted.

brendankenny:
@exterkamp should we add this to our smoke tests? Would it be easy to add the @font-face thing to https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html? The expectations shouldn't need to be updated, the test would just be that the run completes and the results still come back as expected before.

brendankenny:
we'll punt on the smoke test to a future change (pending getting it to work in a non-complicated way :)

PR #6163, opened 2018-10-03
report: reframe Fast 3G as Slow 4G

patrickhulce:
Good summary copied from @paulirish below:
> The goal of a good mobile throttling preset is never to accurately reflect the conditions of a cellular telecommunications system. It's to reflect the conditions that real users experience.

> The preset here attempts to capture latency/throughput of ~80 percentile observed connection data from the wild. As people globally acquire faster connections, this preset should be updated to keep pace.

> Right now, we have no changes to make in the latency/throughput numbers.

> However given the recent adoption of 4G, this preset actually now captures a 4G experience moreso than a 3G experience.

--------------------------------------------

We don't actually reference 3G in *that* many places, I think this effort will be mostly about docs and how we talk to folks.

The biggest change I'm not 100% sure on how to navigate is the PWA audit and our docs linking to WPT - Fast 3G preset.

re: PWA check
I took the approach of just saying "mobile networks" and "cellular networks" since that is in fact the motivation of the check. The checklist does specifically call out "first visit on a simulated 3G network." though, so not sure if we should update it there or what :)

### DIFF
--- a/docs/throttling.md
+++ b/docs/throttling.md
@@ -1,13 +1,13 @@
 
-## The network throttling preset
+## The mobile network throttling preset
 
-This is the standard recommendation for Slow 4G throttling:
+This is the standard recommendation for mobile throttling:
 
 - Latency: 150ms
 - Throughput: 1.6Mbps down / 750 Kbps up.
 - Packet loss: none.
 
-These exact figures are used as the [WebPageTest "Mobile 3G - Fast" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample) and [Lighthouse's throttling default](https://github.com/GoogleChrome/lighthouse/blob/8f500e00243e07ef0a80b39334bedcc8ddc8d3d0/lighthouse-core/config/constants.js#L19-L26). These values represent roughly the bottom 25% of 4G connections and top 25% of 3G connections.
+These exact figures are used as [Lighthouse's throttling default](https://github.com/GoogleChrome/lighthouse/blob/8f500e00243e07ef0a80b39334bedcc8ddc8d3d0/lighthouse-core/config/constants.js#L19-L26) and represent roughly the bottom 25% of 4G connections and top 25% of 3G connections. They are identical to the [WebPageTest "Mobile 3G - Fast" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample) and slightly faster for many pages than [WebPageTest "4G" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample).
 
 ## Throttling basics
 
--- a/docs/throttling.md
+++ b/docs/throttling.md
@@ -7,7 +7,7 @@ This is the standard recommendation for mobile throttling:
 - Throughput: 1.6Mbps down / 750 Kbps up.
 - Packet loss: none.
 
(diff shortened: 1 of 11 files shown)


lighthouse-core/report/html/renderer/util.js
```diff
@@ -484,8 +484,8 @@ Util.UIStrings = {
...
   crcLongestDurationLabel: 'Maximum critical path latency:',
 
-  /** Explanation shown to users below performance results to inform them that the test was done with a 3G network connection and to warn them that the numbers they see will likely change slightly the next time they run Lighthouse. 'Lighthouse' becomes link text to additional documentation. */
-  lsPerformanceCategoryDescription: '[Lighthouse](https://developers.google.com/web/tools/lighthouse/) analysis of the current page on emulated 3G. Values are estimated and may vary.',
+  /** Explanation shown to users below performance results to inform them that the test was done with a 4G network connection and to warn them that the numbers they see will likely change slightly the next time they run Lighthouse. 'Lighthouse' becomes link text to additional documentation. */
+  lsPerformanceCategoryDescription: '[Lighthouse](https://developers.google.com/web/tools/lighthouse/) analysis of the current page on emulated 4G. Values are estimated and may vary.',
```
brendankenny: this ends up prominent in the report, so is it worth adding the extra "slow" in front of "4G" for this string?

brendankenny:
also, maybe add a link to previous discussion to the PR description for future digging into this?

lighthouse-core/report/html/renderer/util.js
patrickhulce: I took the approach that if it wasn't worth qualifying `fast` before than why qualify `slow` now, but I see both sides.

brendankenny: yeah, makes sense. I was thinking in terms of easing in users who will still be thinking in terms of "3G is what I should be testing", rather than specific speeds/network characteristics

patrickhulce: yeah that's a good point, I like the more general `emulated mobile network` here too actually :)

docs/throttling.md
```diff
@@ -1,13 +1,13 @@
...
+This is the standard recommendation for mobile throttling:
 
 - Latency: 150ms
 - Throughput: 1.6Mbps down / 750 Kbps up.
 - Packet loss: none.
 
-These exact figures are used as the [WebPageTest "Mobile 3G - Fast" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample) and [Lighthouse's throttling default](https://github.com/GoogleChrome/lighthouse/blob/8f500e00243e07ef0a80b39334bedcc8ddc8d3d0/lighthouse-core/config/constants.js#L19-L26). These values represent roughly the bottom 25% of 4G connections and top 25% of 3G connections.
+These exact figures are used as [Lighthouse's throttling default](https://github.com/GoogleChrome/lighthouse/blob/8f500e00243e07ef0a80b39334bedcc8ddc8d3d0/lighthouse-core/config/constants.js#L19-L26) and represent roughly the bottom 25% of 4G connections and top 25% of 3G connections. They are identical to the [WebPageTest "Mobile 3G - Fast" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample) and slightly faster for many pages than [WebPageTest "4G" preset](https://github.com/WPO-Foundation/webpagetest/blob/master/www/settings/connectivity.ini.sample).
```
paulirish: >  and slightly faster for many pages

and, due to a lower latency, slightly faster for some pages

lighthouse-core/report/html/renderer/util.js
brendankenny: 💥 nice

docs/throttling.md
patrickhulce: alright alright I begrudgingly admit that's better :)

kusler:
LGTM. Makes sense, and +1 to Addy - would love to see this reflected broadly in our tooling.

Issue #5994, opened 2018-09-07
Only capture similar exceptions in sentry once per audit


brendankenny:
Discussed in #5993. We're lowering the overall sample rate, but to get an idea of relative error rates, we need to stop capturing a bunch of exceptions per audit. e.g. if multiple offscreen images have issues with their sizing information, we capture an exception per image.

Instead we should capture a single exception for the whole audit, noting the issue and giving enough information that the specific problem image can be tracked down (e.g. maybe just log the first one, which gets you details, not just that there was a problem in general).

Places to fix:
- [x] `fonts.js` ~#5993~
- [ ] `offscreen-images.js`
- [ ] `uses-responsive-images.js`
- [ ] `no-vulnerable-libraries.js`
- [ ] `optimized-images.js`
- [ ] `response-compression.js`

PR #6166, opened 2018-10-03
core(OYB): remove fonts gatherer

patrickhulce:
don't hate me @wardpeet 😬 

**Summary**
I thought I'd try another approach for the font-display audit when I noticed the `Fonts` gatherer taking ~25% of a perf-only CNN run these days. We're already collecting the stylesheet content for everything for `unminified/unused-css`, so can a simple regex lookup get the job done?

It passes smoke tests, but I'd want to try this on a few more higher profile sites too.


wardpeet:
@patrickhulce I don't hate, I kill 🗡 😄 

 it's actually something I suggested as well https://github.com/GoogleChrome/lighthouse/issues/5676. I'll run it against some sites too, just make sure the results are quite ok.

patrickhulce:
> it's actually something I suggested as well #5676

doh! you totally did my bad!

wardpeet:
I wouldn't have done such a great job though 😊

wardpeet:
Results are the same I tested it on
edition.cnn.com
[master](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/edition.cnn.com_master)
[perf](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/edition.cnn.com_perf_font)

hotelurbano.com
[master](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/hotelurbano.com_master)
[perf](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/hotelurbano.com_perf_font)

tinyhousebuild.com
[master](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/tinyhousebuild.com_master)
[perf](http://wardpeet-filestorage.surge.sh/lh-store/pr-6166/tinyhousebuild.com_perf_font)

lighthouse-core/audits/font-display.js
```diff
@@ -33,59 +36,91 @@ class FontDisplay extends Audit {
...
+        const rawFontURLs = declaration.match(CSS_URL_GLOBAL_REGEX);
+        if (!rawFontURLs) continue;
+
+        const relativeURLs = rawFontURLs
+          // @ts-ignore - guaranteed to match from previous regex, pull URL group out
+          .map(s => s.match(CSS_URL_REGEX)[1].trim())
+          // remove any optional quotes before/after
+          .map(s => {
+            const firstChar = s.charAt(0);
+            if (firstChar === s.charAt(s.length - 1) && (firstChar === '"' || firstChar === '\'')) {
```
brendankenny: better like this than in the regex?

lighthouse-core/audits/font-display.js
```diff
@@ -33,59 +36,91 @@ class FontDisplay extends Audit {
...
+              return s.substr(1, s.length - 2);
+            }
+
+            return s;
+          });
+
+        const absoluteURLs = relativeURLs.map(url => new URL(url, artifacts.URL.finalUrl));
```
brendankenny: how do we want to deal with exceptions here? e.g. invalid font URL probably shouldn't throw the whole audit

lighthouse-core/test/audits/font-display-test.js
```diff
@@ -5,89 +5,110 @@
...
       },
-    ], webFonts)).then(result => {
-      assert.strictEqual(result.rawValue, true);
-    });
+    ];
+
+    const result = await Audit.audit(getArtifacts());
+    assert.strictEqual(result.rawValue, true);
+    assert.deepEqual(result.details.items, []);
   });
```
brendankenny: how do you feel about populating this with some of the `@font-face` blocks we can find out on real sites? (like the sites ward mentioned in https://github.com/GoogleChrome/lighthouse/pull/6166#issuecomment-427505006). We'd have better coverage for the regexes, plus we'd have a ready-made place to put future regression tests if we ever find we're missing weird things people are doing.

If all the examples we can find are pretty similar, could use them but vary new lines, `font-display` value, whitespace, and throw in some invalid URLs. Is there anything interesting for fonts not in the network records or not in stylesheets? (probably not for this audit)

lighthouse-core/test/audits/font-display-test.js
```diff
@@ -5,89 +5,110 @@
...
 
-  it('fails when not all fonts have a correct font-display rule', () => {
-    const webFonts = [
-      Object.assign({}, openSansFont, {display: 'block'}),
-      openSansFontBold,
-    ];
+  it('fails when not all fonts have a correct font-display rule', async () => {
+    stylesheet.content = `
+      @font-face {
+        src: url("./font-a.woff");
```
brendankenny: maybe call out the quote differences here? Either a comment (or comments for each in the css string) or even a test only testing quoting the url? (though it would be identical to this test :)

It's kind of not obvious that this is covering those possibilities if you haven't just read the audit, and I could see them being accidentally dropped in future edits to this file without knowing they were important.

lighthouse-core/audits/font-display.js
patrickhulce: yeah good point, I'll filter on validity

lighthouse-core/test/audits/font-display-test.js
patrickhulce: great idea! will do

lighthouse-core/test/audits/font-display-test.js
patrickhulce: yeah good call, done

lighthouse-core/audits/font-display.js
patrickhulce: I thought this might be more readable than the backreference regex, but I guess not :) switched!

brendankenny: > I thought this might be more readable than the backreference regex

haha, it might be! I was just asking

lighthouse-core/test/audits/font-display-test.js
```diff
@@ -5,89 +5,162 @@
...
+        resourceType: 'Font',
+      },
+    ];
+
+    const result = await Audit.audit(getArtifacts());
+    assert.strictEqual(result.rawValue, true);
+    assert.deepEqual(result.details.items, []);
+  });
+
+  it('should handle real-world font-face declarations', async () => {
```
brendankenny: 👍

lighthouse-core/audits/font-display.js
```diff
@@ -33,59 +37,101 @@ class FontDisplay extends Audit {
...
+            return s;
+          });
+
+        // Convert the relative CSS URL to an absolute URL and add it to the passing set
+        for (const relativeURL of relativeURLs) {
+          try {
+            const absoluteURL = new URL(relativeURL, artifacts.URL.finalUrl);
+            passingURLs.add(absoluteURL.href);
+          } catch (err) {
+            Sentry.captureException(err, {tags: {audit: this.meta.id}});
```
brendankenny: We should probably only capture once per audit run. Kind of annoying to implement, but some sites end up with a ridiculous number of broken URLs (or whatever we're capturing in a loop). See #5994 (which the only one I had actually checked off was the gatherer you're deleting :)...or disagree with the conclusion there :)

patrickhulce: I have a counter-proposal that I will put up in a different PR :)

PR #6171, opened 2018-10-04
tests: add networkRecord-to-devtoolsLog mocking utility

brendankenny:
This came up while refactoring computed artifacts and I've separated it to keep the PRs simpler. Since we'll be requiring files normally, we can't just inject `requestNetworkRecords` into tests anymore (at least not as easily).

We have a lot of places that we want to pass in example/simplified `networkRecords`, however our core artifact for network activity is the `devtoolsLog`. This PR adds a simple test utility that will create a `devtoolsLog` that, when passed back into `NetworkRecorder`, will create `networkRecords` with all the same properties.

Very simple records can be passed in (good for tests! e.g. just `[{url: 'https://example.com}]`) and the rest of the needed properties will be filled in with reasonable-ish defaults.

I've gone through all the existing tests that were mocking `requestNetworkRecords: () => Promise.resolve(records)` to use a generated devtoolsLog instead.

### DIFF
--- a/lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
+++ b/lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
@@ -5,8 +5,10 @@
  */
 'use strict';
 
+const Runner = require('../../../runner.js');
 const UsesHTTP2Audit = require('../../../audits/dobetterweb/uses-http2.js');
 const assert = require('assert');
+const networkRecordsToDevtoolsLog = require('../../network-records-to-devtools-log.js');
 
 const URL = 'https://webtide.com/http2-push-demo/';
 const networkRecords = require('../../fixtures/networkRecords-mix.json');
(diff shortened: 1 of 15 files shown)


lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
```diff
@@ -5,21 +5,31 @@
...
 
 const URL = 'https://webtide.com/http2-push-demo/';
 const networkRecords = require('../../fixtures/networkRecords-mix.json');
 
 /* eslint-env jest */
 
 describe('Resources are fetched over http/2', () => {
+  function getArtifacts(networkRecords, finalUrl) {
+    // networkRecords-mix.json is an old network request format, so don't verify round-trip.
+    const devtoolsLog = networkRecordsToDevtoolsLog(networkRecords, {skipVerification: true});
```
brendankenny: using a real networkRecord as input has a few problems with verification of the output, namely that we've already [snipped a bunch of circular references and replaced them with strings](https://github.com/GoogleChrome/lighthouse/blob/fbcb1e3ea3aa05b0cfdf87ddbc8155bb6fe1783f/lighthouse-core/test/fixtures/networkRecords-mix.json#L7) and we've switched a few properties around since that old version of `NetworkRecorder`. The subset needed for this test generates correctly, so it seemed better to just not verify rather than edit the json file.

benschwarz:
Could this be repurposed to generate hars and other network analysis under simulation mode @patrickhulce?

brendankenny:
it could maybe be a start, but there's probably easier/more robust ways to go about it. This is definitely aimed at when you need to test the accessing of just a few properties, like the main resource by URL in a set of records and you don't care that that request started at timestamp 0 and was completed 1 µs later :)

lighthouse-core/test/network-records-to-devtools-log.js
```diff
@@ -0,0 +1,210 @@
...
+
+    devtoolsLog.push(getResponseReceivedEvent(networkRecord, index));
+    devtoolsLog.push(getDataReceivedEvent(networkRecord, index));
+    devtoolsLog.push(getLoadingFinishedEvent(networkRecord, index));
+  });
+
+  // If in a test, assert that the log will turn into an equivalent networkRecords.
+  if (global.expect && !options.skipVerification) {
+    const roundTrippedNetworkRecords = NetworkRecorder.recordsFromLogs(devtoolsLog);
+    expect(roundTrippedNetworkRecords).toMatchObject(networkRecords);
```
patrickhulce: 🎉 does this mean it's growing on you? :D

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -25,16 +26,15 @@ describe('Performance: uses-rel-preload audit', () => {
   let mockSimulator;
 
   const mockArtifacts = (networkRecords, mainResource = defaultMainResource) => {
-    return {
+    return Object.assign(Runner.instantiateComputedArtifacts(), {
```
patrickhulce: this is a bummer :( more incentive to refactor all the compute artifacts I suppose :)

PR #6195, opened 2018-10-09
tests: remove more mock computed artifacts

brendankenny:
Follow up to #6171, using the `networkRecordsToDevtoolsLog` utility and passing in the resulting devtoolsLog instead of mocking network records directly. Also removed some other low hanging `request*()` mocks in these files.

Sorry for the test noise :)

### DIFF
--- a/lighthouse-core/test/audits/metrics/speed-index-test.js
+++ b/lighthouse-core/test/audits/metrics/speed-index-test.js
@@ -28,19 +28,4 @@ describe('Performance: speed-index audit', () => {
       assert.equal(result.rawValue, 605);
     });
   }, 10000);
-
-  it('scores speed index of 845 as 100', () => {
-    const artifacts = {
-      traces: {},
-      devtoolsLogs: {},
-      requestSpeedIndex() {
-        return Promise.resolve({timing: 845});
-      },
-    };
-
-    return Audit.audit(artifacts, {options}).then(result => {
-      assert.equal(result.score, 1);
-      assert.equal(result.rawValue, 845);
-    });
-  });
 });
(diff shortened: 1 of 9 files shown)


lighthouse-core/test/audits/metrics/speed-index-test.js
```diff
@@ -28,19 +28,4 @@ describe('Performance: speed-index audit', () => {
       assert.equal(result.rawValue, 605);
     });
   }, 10000);
-
-  it('scores speed index of 845 as 100', () => {
```
brendankenny: the test above already tests that a score of 605 yields a passing score. This seemed redundant (and difficult to fake)

lighthouse-core/test/network-records-to-devtools-log.js
```diff
@@ -74,9 +78,12 @@ function getRequestWillBeSentEvent(networkRecord, index) {
...
   const headers = headersArrayToHeadersDict(networkRecord.responseHeaders);
-  const timing = networkRecord.timing;
-  if (timing && timing.requestTime === undefined) {
-    timing.requestTime = networkRecord.startTime || 0;
+  let timing;
+  if (networkRecord.timing) {
+    timing = {...networkRecord.timing};
+    if (timing.requestTime === undefined) {
+      timing.requestTime = networkRecord.startTime || 0;
+    }
```
brendankenny: a few tweaks to get this to work well with existing tests. e.g. not overwriting `timing` or `initiator` as tests sometimes reuse the object and expect it to be unchanged

brendankenny:
> Sorry for the test noise :)

Oh, and remember that all the

```js
const artifacts = Object.assign(Runner.instantiateComputedArtifacts(), {
  // ...
});
```

after the refactor will become just
```js
const artifacts = {
  // ...
};
```
noisy but keeps the tests passing without a 2000 line PR or whatever :)

PR #6196, opened 2018-10-09
tests: add test trace creator

brendankenny:
other half #6195. This PR adds a simple trace creator to easily create test traces that match the needed characteristics. Right now it accepts timestamps for navStart and the end of the trace, and then a set of timestamps and durations for top-level tasks (but the options can easily be expanded).

The allows easy replacement of a bunch of computed artifact mocks relying on a specific TTI or just needed a dummy trace so the computed artifacts don't throw while another aspect of the audit is being tested.

Some of these are a little more sensitive to future changes than what was found in #6195, but we do get a bit more robustness in that these files now test with real computed artifacts rather than just testing the quality of our mocks (which can be particularly risky when testing just simple pass/fail conditions without testing *why*).

### DIFF
--- a/lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
+++ b/lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
@@ -7,10 +7,10 @@
 
 const UnusedImages =
     require('../../../audits/byte-efficiency/offscreen-images.js');
-const NetworkNode = require('../../../lib/dependency-graph/network-node');
-const CPUNode = require('../../../lib/dependency-graph/cpu-node');
 const assert = require('assert');
-const LHError = require('../../../lib/lh-error');
+const Runner = require('../../../runner.js');
+const createTestTrace = require('../../create-test-trace.js');
+const networkRecordsToDevtoolsLog = require('../../network-records-to-devtools-log.js');
 
 /* eslint-env jest */
 function generateRecord(resourceSizeInKb, startTime = 0, mimeType = 'image/png') {
@@ -45,26 +45,6 @@ function generateImage(size, coords, networkRecord, src = 'https://google.com/lo
   return image;
 }
 
-function generateInteractiveFunc(desiredTimeInSeconds) {
-  return () => Promise.resolve({
-    timestamp: desiredTimeInSeconds * 1000000,
-  });
-}
-
-function generateInteractiveFuncError() {
-  return () => Promise.reject(
-    new LHError(LHError.errors.NO_TTI_NETWORK_IDLE_PERIOD)
-  );
-}
-
-function generateTraceOfTab(desiredTimeInSeconds) {
-  return () => Promise.resolve({
-    timestamps: {
-      traceEnd: desiredTimeInSeconds * 1000000,
-    },
-  });
-}
-
 describe('OffscreenImages audit', () => {
   let context;
   const DEFAULT_DIMENSIONS = {innerWidth: 1920, innerHeight: 1080};
@@ -74,15 +54,17 @@ describe('OffscreenImages audit', () => {
   });
 
   it('handles images without network record', () => {
-    return UnusedImages.audit_({
+    const topLevelTasks = [{ts: 1900, duration: 100}];
+    const artifacts = Object.assign(Runner.instantiateComputedArtifacts(), {
       ViewportDimensions: DEFAULT_DIMENSIONS,
       ImageUsage: [
         generateImage(generateSize(100, 100), [0, 0]),
       ],
-      traces: {},
+      traces: {defaultPass: createTestTrace({topLevelTasks})},
       devtoolsLogs: {},
-      requestInteractive: generateInteractiveFunc(2),
-    }, [], context).then(auditResult => {
+    });
+
+    return UnusedImages.audit_(artifacts, [], context).then(auditResult => {
       assert.equal(auditResult.items.length, 0);
     });
   });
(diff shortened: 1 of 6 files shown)


lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -121,189 +106,208 @@ describe('OffscreenImages audit', () => {
...
+      priority: 'High',
+      timing: {receiveHeadersEnd: 1.5},
+    };
+    const devtoolsLog = networkRecordsToDevtoolsLog([recordA, recordB]);
+
+    // Enough tasks to spread out graph.
+    const topLevelTasks = [
+      {ts: 1000, duration: 10},
+      {ts: 1050, duration: 10},
+      {ts: 1975, duration: 50},
```
brendankenny: @patrickhulce I don't entirely understand why I couldn't get the pessimistic graph to download these images in parallel without these extra 10ms tasks. Instead the second image and the CPU node always went in parallel (even if the two images had identical start and end times) and so it was always failing seeing both of these images before the last long task started.

Inserting these extra top level tasks fixed that and doesn't feel too bad (most traces have more than one top level task :), but I do worry about the fragility of my solution

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
```diff
@@ -121,189 +106,208 @@ describe('OffscreenImages audit', () => {
...
+      ViewportDimensions: DEFAULT_DIMENSIONS,
+      ImageUsage: [
+        generateImage(generateSize(0, 0), [0, 0], generateRecord(100, 3), 'a'),
+        generateImage(generateSize(200, 200), [3000, 0], generateRecord(100, 4), 'b'),
+      ],
+      traces: {defaultPass: createTestTrace({traceEnd: 2000})},
+      devtoolsLogs: {},
+    });
 
-  it('finds images loaded before Trace End when interactive throws error (Lantern)', async () => {
```
brendankenny: as far as I can tell this test was identical to the above. The start time of `generateRecord` of 'a' is the only thing that's different, but `throttlingMethod: 'simulate'` uses the network record timing information, not the timing info in the `ImageUsage` artifact.

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -9,191 +9,206 @@
...
       output => {
-        assert.equal(output.rawValue, 1250);
+        assert.equal(output.details.overallSavingsMs, 330);
         assert.equal(output.details.items.length, 2);
-        assert.equal(output.details.items[0].url, 'http://www.example.com/script-added.js');
-        assert.equal(output.details.items[1].url, 'http://sub.example.com/script-sub.js');
+        assert.equal(output.details.items[0].url, scriptSubNodeUrl);
+        assert.equal(output.details.items[0].wastedMs, 330);
+        assert.equal(output.details.items[1].url, scriptAddedNodeUrl);
+        assert.equal(output.details.items[1].wastedMs, 180);
```
brendankenny: @patrickhulce it seemed fine to change these to the timings derived from the generated network records, but I wasn't sure what the `if (scriptOtherNodeLocal.getDependencies()[0] === mainDocumentNodeLocal) {` conditionals were up to, so wanted to double check

lighthouse-core/test/network-records-to-devtools-log.js
```diff
@@ -32,7 +32,8 @@ function getBaseRequestId(record) {
 function headersArrayToHeadersDict(headersArray = []) {
   const headersDict = {};
   headersArray.forEach(headerItem => {
-    const value = headersDict[headerItem.name] ? headersDict[headerItem.name] + '\n' : '';
+    const value = headersDict[headerItem.name] !== undefined ?
```
brendankenny: changes also in #6195 but also needed for this PR

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
patrickhulce: I can't tell for 100% just from glancing, but it's probably because they were inferred to have come from the same connection, if different connectionIds are added and/or the timings show they happened in parallel then it might fix itself?

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
patrickhulce: yeah I think this was very recently added when I had @exterkamp keep fiddling with the tests, probably got duped in the shuffle sorry!

lighthouse-core/test/audits/uses-rel-preload-test.js
patrickhulce: that was just so we'd only be setting the original timings and letting the altered simulation ones be inferred on there own.


this is much nicer :) 👍

lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -9,191 +9,198 @@
...
   it(`shouldn't suggest preload for protocol data`, () => {
-    const networkRecords = [
-      {
-        requestId: '3',
-        protocol: 'data',
-        startTime: 10,
-      },
-    ];
+    const networkRecords = getMockNetworkRecords();
+    networkRecords[2].protocol = 'data';
```
patrickhulce: beautiful these are real tests now! :D

lighthouse-core/test/create-test-trace.js
```diff
@@ -0,0 +1,117 @@
...
+
+  if (options.traceEnd) {
+    // Insert a top level short task to extend trace to requested end.
+    traceEvents.push(getTopLevelTask({ts: options.traceEnd - 1, duration: 1}));
+  }
+
+  return {traceEvents};
+}
+
+module.exports = createTestTrace;
```
patrickhulce: i ❤️this 📄

lighthouse-core/test/audits/byte-efficiency/offscreen-images-test.js
brendankenny: > different connectionIds are added and/or the timings show they happened in parallel 

ah, didn't work. There's probably some subtlety with leaving out real connection timings or something, too. I'll just leave it like this for now and maybe we'll figure it out when a lantern change breaks this test :)

PR #6209, opened 2018-10-10
core: remove dependency on WebInspector

patrickhulce:
**Summary**
A lot going on here but overall this PR removes the DevTools dependency on the font-size gatherer

- Upgrade jest to 23.6 so we can use `toMatchInlineSnapshot`
- Add prettier to devDeps so we can use `toMatchInlineSnapshot`
- Rewrite font-size gatherer for async/await + type-checking (too thorny to try to mess with so much in there the way it was)
- Replace the CSS font rule identification logic with custom handler, we actually start identifying 1 new CSS rule that was being missed on CNN! (probably just so old devtools code it got out of sync)
- Remove the WebInspector dependency in font-size and delete the code!!

witch


.prettierrc
```diff
@@ -0,0 +1,6 @@
+printWidth: 100
```
wardpeet: do we want to include prettier in this PR? also should we enforce it in travis that everything needs to be prettiefied?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +48,128 @@ function nodeInBody(node) {
...
-      return nodes.filter(nodeInBody);
-    });
+async function getAllNodesFromBody(driver) {
+  const nodes = /** @type {DomNodeMaybeWithParent[]} */ (await driver.getNodesInDocument());
+  /** @type {Map<number, LH.Artifacts.FontSize.DomNodeMaybeWithParent>} */
+  const nodeMap = new Map();
+  nodes.forEach(node => nodeMap.set(node.nodeId, node));
+  nodes.forEach(node => (node.parentNode = nodeMap.get(node.parentId)));
+  // @ts-ignore - once passing through the filter, it's guaranteed to have a parent
+  return nodes.filter(nodeInBody);
```
wardpeet: @brendankenny I was struggling with this in my own projects to. Is there no way to fix this with jsdoc?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -144,107 +233,121 @@ function getFontSizeInformation(driver, node) {
...
+  async afterPass(passContext) {
     /** @type {Map<string, LH.Crdp.CSS.CSSStyleSheetHeader>} */
     const stylesheets = new Map();
     /** @param {LH.Crdp.CSS.StyleSheetAddedEvent} sheet */
     const onStylesheetAdd = sheet => stylesheets.set(sheet.header.styleSheetId, sheet.header);
     passContext.driver.on('CSS.styleSheetAdded', onStylesheetAdd);
 
-    const enableDOM = passContext.driver.sendCommand('DOM.enable');
-    const enableCSS = passContext.driver.sendCommand('CSS.enable');
+    await passContext.driver.sendCommand('DOM.enable');
```
wardpeet: not that it's a big deal but we had a Promise.all in the previous version. Probably not worth the hassle
`await Promise.all([passContext.driver.sendCommand('DOM.enable'), passContext.driver.sendCommand('CSS.enable')]);`

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,28 +183,43 @@ function getNodeTextLength(node) {
...
 
 /**
  * @param {Driver} driver
  * @param {LH.Artifacts.FontSize.DomNodeWithParent} node text node
- * @returns {Promise<?{fontSize: number, textLength: number, node: LH.Artifacts.FontSize.DomNodeWithParent}>}
+ * @returns {Promise<?FailingNodeData>}
  */
-function getFontSizeInformation(driver, node) {
-  return driver.sendCommand('CSS.getComputedStyleForNode', {nodeId: node.parentId})
+function fetchComputedFontSize(driver, node) {
```
wardpeet: any reason why we are not asyncing this function?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -144,107 +233,121 @@ function getFontSizeInformation(driver, node) {
...
-          passContext.driver.sendCommand('CSS.disable'),
-        ]).then(_ => ({
-          analyzedFailingNodesData,
-          analyzedFailingTextLength,
-          failingTextLength,
-          visitedTextLength,
-          totalTextLength,
-        }));
-      });
+    await passContext.driver.sendCommand('DOM.disable');
```
wardpeet: same goes for this one that it was a Promise.all before

.prettierrc
patrickhulce: I think that's worth a separate discussion and PR if we decide to move that way.

package.json
```diff
@@ -98,12 +98,13 @@
...
     "gulp-util": "^3.0.7",
     "intl": "^1.2.5",
-    "jest": "^23.2.0",
+    "jest": "^23.6.0",
     "jsdom": "^9.12.0",
     "mocha": "^3.2.0",
     "npm-run-posix-or-windows": "^2.0.2",
     "nyc": "^11.6.0",
     "postinstall-prepare": "^1.0.1",
+    "prettier": "^1.14.3",
```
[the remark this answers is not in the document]
patrickhulce: not used for anything other than `toMatchInlineSnapshot` right now, you know I'd be on board though :)

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: oh no real reason other than I didn't touch any of the logic, small diff whens don't mean much in this PR though 😆

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: sure might as well

package.json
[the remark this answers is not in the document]
wardpeet: yeah sorry I read it to late in your comment above. But I think we should just move 😋 I've found a way to let travis fail without changing code on commit.
`prettier --list-different '**/*.{js,json}`

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +184,175 @@ function getNodeTextLength(node) {
...
+      selectors: sourceRule.parentRule.selectors,
+    },
+  };
 }
 
 /**
  * @param {Driver} driver
  * @param {LH.Artifacts.FontSize.DomNodeWithParent} node text node
- * @returns {Promise<?{fontSize: number, textLength: number, node: LH.Artifacts.FontSize.DomNodeWithParent}>}
+ * @returns {Promise<?FailingNodeData>}
```
brendankenny: I guess it's unfortunate to call this `FailingNodeData` since it might not be failing at this point. Would it lose precision elsewhere to rename it (`NodeFontData` or whatever)?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +184,175 @@ function getNodeTextLength(node) {
...
+    const fontSizeProperty = computedStyle.find(({name}) => name === FONT_SIZE_PROPERTY_NAME);
+
+    return {
+      // @ts-ignore - font size property guaranteed to be returned in getComputedStyle
+      fontSize: parseInt(fontSizeProperty.value, 10),
+      textLength: getNodeTextLength(node),
+      node: /** @type {LH.Artifacts.FontSize.DomNodeWithParent} */ (node.parentNode),
+    };
+  } catch (err) {
+    Sentry.captureException(err);
```
brendankenny: should we debounce this in keeping with #5994?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -15,19 +14,25 @@
...
 const MINIMAL_LEGIBLE_FONT_SIZE_PX = 12;
 // limit number of protocol calls to make sure that gatherer doesn't take more than 1-2s
 const MAX_NODES_VISITED = 500;
 const MAX_NODES_ANALYZED = 50;
 
+/** DevTools uses a numeric enum for nodeType */
+const TEXT_NODE_TYPE = 3;
+
+/** @typedef {LH.Artifacts.FontSize['analyzedFailingNodesData'][0]} FailingNodeData */
+/** @typedef {LH.Artifacts.FontSize.DomNodeMaybeWithParent} DomNodeMaybeWithParent*/
```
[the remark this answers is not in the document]
patrickhulce: https://github.com/Microsoft/TypeScript/issues/20077 seems to suggest inheritance plus recursive type doesn't work so well, I might need some help making it local :)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -15,19 +14,25 @@
...
+
 const Driver = require('../../driver.js'); // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +49,128 @@ function nodeInBody(node) {
...
-function getAllNodesFromBody(driver) {
-  return driver.getNodesInDocument()
-    .then(nodes => {
-      const nodeMap = new Map();
-      nodes.forEach(node => nodeMap.set(node.nodeId, node));
-      nodes.forEach(node => node.parentNode = nodeMap.get(node.parentId));
-      return nodes.filter(nodeInBody);
-    });
+async function getAllNodesFromBody(driver) {
+  const nodes = /** @type {DomNodeMaybeWithParent[]} */ (await driver.getNodesInDocument());
```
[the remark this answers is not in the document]
patrickhulce: yeah didn't even have to do that, just a fallback for parent id

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +49,128 @@ function nodeInBody(node) {
...
  *
- * @param {string} property CSS property name
- * @param {LH.Crdp.DOM.Node} node
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity
+ * @see https://www.smashingmagazine.com/2010/04/css-specificity-and-inheritance/
+ *
+ * @param {string} selector
+ * @return {number}
+ */
+function computeSelectorSpecificity(selector) {
```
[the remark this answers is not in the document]
patrickhulce: I don't know how to interpret this emoji :) haha

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +49,128 @@ function nodeInBody(node) {
...
+    numIDs += ids.length;
+    numClasses += classes.length;
+    numTypes += types.length;
+  }
+
+  return Math.min(9, numIDs) * 100 + Math.min(9, numClasses) * 10 + Math.min(9, numTypes);
+}
+
+/**
+ *
```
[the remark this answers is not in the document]
patrickhulce: 👍 renamed the function a bit too

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: thoughts on #6215 first? :)

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: yeah I think that works fine 👍

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +48,130 @@ function nodeInBody(node) {
...
+  for (const {inlineStyle, matchedCSSRules} of inheritedEntries) {
+    if (hasFontSizeDeclaration(inlineStyle)) return {type: 'Inline', ...inlineStyle};
+
+    const directRule = findMostSpecificMatchedCSSRule(matchedCSSRules);
+    if (directRule) return directRule;
+  }
+}
+
+/**
+ * Returns effective CSS rule for given CSS property.
```
brendankenny: Since this is just matching against `font-size`, maybe mention that instead. "effective" also isn't immediately clear here, instead of something about how it's the css rule currently setting the style for the `font-size` property

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +185,175 @@ function getNodeTextLength(node) {
...
-      return {
-        fontSize: parseInt(fontSizeProperty.value, 10),
-        textLength: getNodeTextLength(node),
-        node: node.parentNode,
-      };
-    })
-    .catch(err => {
-      require('../../../lib/sentry.js').captureException(err);
-      return null;
+async function fetchComputedFontSize(driver, node) {
```
brendankenny: maybe rename `node` to `textNode`? Wasn't immediately clear why you'd want the computed style for the node's *parent* without that info :)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +185,175 @@ function getNodeTextLength(node) {
...
+      nodeId: node.parentId,
     });
+
+    const fontSizeProperty = computedStyle.find(({name}) => name === FONT_SIZE_PROPERTY_NAME);
+
+    return {
+      // @ts-ignore - font size property guaranteed to be returned in getComputedStyle
+      fontSize: parseInt(fontSizeProperty.value, 10),
+      textLength: getNodeTextLength(node),
+      node: /** @type {LH.Artifacts.FontSize.DomNodeWithParent} */ (node.parentNode),
```
brendankenny: cast not needed anymore, I don't think

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +185,175 @@ function getNodeTextLength(node) {
...
+        failingNodes.push(visitedNode);
+        failingTextLength += visitedNode.textLength;
+      }
+    }
+
+    const nodesToAnalyze = failingNodes
+      .sort((a, b) => b.textLength - a.textLength)
+      .slice(0, MAX_NODES_ANALYZED);
+
+    return {totalTextLength, visitedTextLength, failingTextLength, nodesToAnalyze};
```
brendankenny: maybe `failingNodes` to standardize on that terminology instead of `nodesToAnalyze`

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +185,175 @@ function getNodeTextLength(node) {
...
+  }
+
+  /**
+   *
+   * @param {LH.Gatherer.PassContext['driver']} driver
+   * @param {Array<NodeFontData>} nodesToAnalyze
+   */
+  static async analyzeFailingNodes(driver, nodesToAnalyze) {
+    const analysisPromises = nodesToAnalyze.map(async failingNode => {
```
brendankenny: It would be nice if there was an easier way to express the async behavior needed but still use `forEach`

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +185,175 @@ function getNodeTextLength(node) {
...
+    });
+
+    const analyzedFailingNodesData = await Promise.all(analysisPromises);
+
+    const analyzedFailingTextLength = analyzedFailingNodesData.reduce(
+      (sum, {textLength}) => (sum += textLength),
+      0
+    );
+
+    return {analyzedFailingNodesData, analyzedFailingTextLength};
```
brendankenny: just `analyzedFailingNodes`, maybe?

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -36,53 +60,249 @@ describe('Font size gatherer', () => {
...
+
+      inlineStyle = createStyle({id: 1, properties: {'font-size': '1em'}});
+      matchedCSSRules = [{matchingSelectors: [1], rule: fontRule}];
+      inherited = [{matchedCSSRules: [{matchingSelectors: [0], rule: userAgentRule}]}];
+    });
+
+    it('should identify inline styles', () => {
+      const result = FontSizeGather.getEffectiveFontRule({inlineStyle});
+      expect(result).toMatchInlineSnapshot(`
+Object {
```
brendankenny: *IMO*, it seems like
a) it would be shorter and easier to read as an inline object literal and an expect.toEqual
b) if reading isn't important, then could just be a regular snapshot (we have shorter ones :)
c) if these were going to occasionally update, maybe there's an advantage for snapshots' ease of use, but these shouldn't ever change
rather than bringing in prettier just for this feature

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -36,53 +60,249 @@ describe('Font size gatherer', () => {
...
+      expect(compute('h1 > p.other.yeah > span')).toEqual(23);
+    });
+
+    it('should handle ID selectors', () => {
+      expect(compute('h1#awesome.foo')).toEqual(111);
+      expect(compute('h1 > p.other > span#the-text')).toEqual(113);
+    });
+
+    it('should cap the craziness', () => {
+      expect(compute('h1.a.b.c.d.e.f.g.h.i.j.k.l.m.n.o.p')).toEqual(91);
```
brendankenny: 😆

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +48,130 @@ function nodeInBody(node) {
...
+  for (const token of tokens) {
+    const ids = token.match(/\b#[a-z0-9]+/g) || [];
+    const classes = token.match(/\b\.[a-z0-9]+/g) || [];
+    const types = token.match(/^[a-z]+/) ? [1] : [];
+    numIDs += ids.length;
+    numClasses += classes.length;
+    numTypes += types.length;
+  }
+
+  return Math.min(9, numIDs) * 100 + Math.min(9, numClasses) * 10 + Math.min(9, numTypes);
```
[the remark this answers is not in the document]
patrickhulce: I've ignored `!important` so far, think it's worth handling? I'll at least comment it's absence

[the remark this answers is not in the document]
brendankenny: > think it's worth handling?

no idea :) I have used it in anger more than once, so if you really really mean to override a font-size with it, it would be annoying if lighthouse didn't recognize it.

Is it non-trivial to support?

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: done

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: done!

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: fixed

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: well the nodesToAnalyze is the capped subset of `failingNodes` so I like that they're named differently, `failingNodesSubsetToAnalyze`?

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: well that's just based on the artifact shape name, feel strongly enough to change the artifact type and its usages up the chain?

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -36,53 +60,249 @@ describe('Font size gatherer', () => {
...
-              inlineStyle: null,
-              attributesStyle: null,
-              matchedCSSRules: [],
-              inherited: [],
-            };
+  it('returns information about font sizes used on page', async () => {
+    const driver = {
+      on() {},
+      off() {},
+      sendCommand(command, params) {
```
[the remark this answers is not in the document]
patrickhulce: done

.prettierrc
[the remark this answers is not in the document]
patrickhulce: fair enough, only a matter of time now anyway 
image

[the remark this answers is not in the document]
patrickhulce: ;)

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
patrickhulce: > Is it non-trivial to support?

Kinda, we basically have to do this entire thing twice, once looking for !important and once not. The good news is that LH will still recognize the percentage of text going down. If you `!important`'d a valid font size, the risk is that you `!important` an also invalid font-size and we point to a rule that isn't really winning.

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
Hoten: I like snapshot testing. I can't figure out why one would want to do it inline, though... Seems creating the initial snapshot + approving updates to the snapshot become a lot more menial tasks.

patrickhulce: @Hoten the inline snapshot is still managed automatically (that's why we needed to update jest and add a prettier dependency), the benefit is you just get to see the result right next to the test instead of a separate file

wardpeet: Snapshot testing all the way! ♥️

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -36,53 +60,249 @@ describe('Font size gatherer', () => {
...
+          if (params.nodeId === failingNode.nodeId) {
+            throw new Error('This is the failing node');
           }
 
-          return Promise.resolve(result);
-        },
-        getNodesInDocument() {
-          return Promise.resolve(nodes);
-        },
+          result = {
```
brendankenny: nit: can just return since nothing else is being done to `result` (IMO marginally more readable since you have to skip down pretty far to see it's just being returned, but, again, just a nit :)

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
brendankenny: ha, I guess I hadn't completely thought through how this gatherer is working.

Of course we already have the computed style, so the css parsing is just to get the rule that caused that style. It's complicated enough that I had forgotten that by this point 😳 I'll add a comment on `analyzeFailingNodes`

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -44,58 +50,131 @@ function nodeInBody(node) {
...
+      parentRule: {
+        origin: maxSpecificityRule.origin,
+        selectors: maxSpecificityRule.selectorList.selectors,
+      },
+    };
+  }
+}
+
+/**
+ * Finds the most specific directly matched CSS rule from the list.
```
brendankenny: font-size rule

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -109,142 +188,175 @@ function getNodeTextLength(node) {
...
+
+    return {totalTextLength, visitedTextLength, failingTextLength, nodesToAnalyze};
+  }
+
+  /**
+   *
+   * @param {LH.Gatherer.PassContext['driver']} driver
+   * @param {Array<NodeFontData>} nodesToAnalyze
+   */
+  static async analyzeFailingNodes(driver, nodesToAnalyze) {
```
brendankenny: Maybe we could make that clearer than "analyze", since it's not really doing any analysis. `fetchSourceRule` describes itself well, so maybe something like `fetchFailingNodeSourceRules`? `sourceFailingNodes`?  :)

lighthouse-core/gather/gatherers/seo/font-size.js
brendankenny: > well the nodesToAnalyze is the capped subset of failingNodes so I like that they're named differently, failingNodesSubsetToAnalyze?

`failingNodesSubset`? `cappedFailingNodes`? `importantFailingNodes`? `failingNodesToSource`?

lighthouse-core/gather/gatherers/seo/font-size.js
brendankenny: > well that's just based on the artifact shape name, feel strongly enough to change the artifact type and its usages up the chain?

ha, I didn't before, but I think we maybe should. The `Data` part is definitely redundant, but what do you think about `sourcedFailingNodes`?

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -36,53 +60,249 @@ describe('Font size gatherer', () => {
...
+    "selectors": Array [
+      Object {
+        "text": "body",
+      },
+    ],
+  },
+  "styleSheetId": undefined,
+  "type": "Regular",
+}
+`);
```
brendankenny: also
```js
expect(result).toEqual({
  cssProperties: [{
    name: 'font-size',
    value: 12,
  }],
  parentRule: {
    origin: 'user-agent',
    selectors: [{
      text: 'body',
    }],
  },
  styleSheetId: undefined,
  type: 'Regular',
});
```
still way better here :P

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
patrickhulce: done

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: done

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: done

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: done

Hoten:
You can put me on the "against" list. I'm not convinced that small objects are a good candidate for snapshot testing. HTML, for sure, but the entire jest / snapshot overhead for this current use case can just be a deep object equality assertion.

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
paulirish: I think inline snapshots introduce benefits to 1) testing consistency and 2) dev ergonomics

**Consistency**

The alternative we're weighing this against is `expect.toEqual(obj)`. But `obj` could be an object literal or it could just constructed some other way.. (`expected = JSON.parse(JSON.stringify(thing)); expected.change = 42`) 

Adopting inline snapshots brings consistency to this pattern and prioritizes a declarative view of the expected object. I think this encourages more reviewable tests.

**Ergonomics**

First time writing the assertion, just do `expect(result).toMatchInlineSnapshot()` and the test will be updated with the current result. When making any changes that affect the expected value, updating them is just a matter of typing `u`.  

To be fair, we don't spend _a lot_ of time writing expecting values, but these workflow improvements make authoring tests faster and easier. (Which encourages more tests).

------

(I looked across our usage of deepStrictEqual and the large majority are very small expected objects... like 1-3 lines tall)

TBH I think toEqual() continues to make sense if the obj literal is like 1-4ish lines tall. From 5-15 lines tall, inline snapshot I think wins for creating/updating advantages.  And past 15 lines moving to a non-inline snapshot makes the most sense.

wdyt?

wardpeet: > First time writing the assertion, just do expect(result).toMatchInlineSnapshot() and the test will be updated with the current result. When making any changes that affect the expected value, updating them is just a matter of typing u.

this is actually why I think snapshots are great. I agree they might be overkill for smaller objects but should we really care? It's easier to always snapshot test just to make it consistent for contributers? The downside of snapshot testing is that your not always seeing what the result is going to be.

with snapshot:
```
const obj = myCoolFunction();
expect(obj).toMatchInlineSnapshot()
```

without:
```
const obj = myCoolFunction();
assert.deepStrictEqual(obj, { a': 'a', b: { c: 'c'}})
```

nevertheless I still think snapshot testing are great for testing artifacts & audit results

lighthouse-core/gather/gatherers/seo/font-size.js
wardpeet: @patrickhulce shouldn't we remove the ts-ignore if it worked?

patrickhulce: I did. Maybe Github is being too smart for us and showing the original diff for the old convo? 🤷‍♂️

PR #6215, opened 2018-10-12
misc: only allow 1 error from each audit/gatherer

### LINKED ISSUE #5994: Only capture similar exceptions in sentry once per audit

patrickhulce:
**Summary**
WDYT about this @brendankenny ?

**Related Issues/PRs**
closes #5994

### DIFF
--- a/lighthouse-core/lib/sentry.js
+++ b/lighthouse-core/lib/sentry.js
@@ -65,6 +65,8 @@ function init(opts) {
     sentryDelegate.captureBreadcrumb = (...args) => Sentry.captureBreadcrumb(...args);
     sentryDelegate.getContext = () => Sentry.getContext();
 
+    // Keep a record of exceptions per audit/gatherer so we can just report once
+    const sentryExceptionCache = new Map();
     // Special case captureException to return a Promise so we don't process.exit too early
     sentryDelegate.captureException = async (err, opts = {}) => {
       // Ignore if there wasn't an error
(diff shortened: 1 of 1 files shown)


brendankenny:
also, this makes me wish that we had `sentry-test.js` :)

lighthouse-core/lib/sentry.js
```diff
@@ -74,6 +76,19 @@ function init(opts) {
       // @ts-ignore Non-standard property added to flag error as not needing capturing.
       if (err.expected) return;
 
+      const tags = opts.tags || {};
+      if (tags.audit) {
+        const key = `audit-${tags.audit}-${err.message}`;
+        if (sentryExceptionCache.has(key)) return;
```
exterkamp: Just curious:  Could 1 exception have both audit and gatherer tags?  In which case returning here would not allow the gatherer tags to be analyzed?  I don't think so, but wanted to make sure.

lighthouse-core/lib/sentry.js
```diff
@@ -74,6 +76,19 @@ function init(opts) {
...
+        sentryExceptionCache.set(key, true);
+      }
+
+      if (tags.gatherer) {
+        const key = `gatherer-${tags.gatherer}-${err.message}`;
+        if (sentryExceptionCache.has(key)) return;
```
exterkamp: Why check `has`?  Can't it just `set` it and override the old version in the map?  Or is that not readable JS? :confused:

lighthouse-core/lib/sentry.js
patrickhulce: nope, it either happens in an audit, or a gatherer, or neither :)

lighthouse-core/lib/sentry.js
patrickhulce: if it has the key we want to `return` and skip the rest of execution, not just the rest of the block. worth adding a comment since it's confusing already? :)

Issue #6208, opened 2018-10-10
icons.isPNG() should not test extension


feliperohdee:
This issue is related to this code segment:

https://github.com/GoogleChrome/lighthouse/blob/6a10dd46aab1d7f220761afbe2921e1804d3c3c6/lighthouse-core/lib/icons.js#L75

For systems which generates PWA's and uses dynamic icon generation, via image url's arguments (like: https://somecdn.com/iconsrc/192x192), are having their score reduced at PWA section. Even using PNG images, with the right size, but without ".png" extension.

Lighthouse should consider just typehint, but not extension.

patrickhulce:
It seems like this was a recent, purposeful change (#6024), though I agree it seems odd that we say the icon *must* have a `.png` extension. @exterkamp were there some extra circumstances we're missing? :)

exterkamp:
Thanks for calling this out.  In the changes I made I think I tried to match the filtering conditions as they were, but we had made an effort to match Chrome but must have missed this edge case.  Opened a PR to fix!

[Relevant Chrome code](https://cs.chromium.org/chromium/src/chrome/browser/installable/installable_manager.cc?sq=package:chromium&dr=CSs&g=0&l=88)

feliperohdee:
Cool! Thank you guys!

Issue #789, opened 2016-10-17
Ligthouse checks for an icon in the manifest but does not try to load it


juliantoledo:
While collection few PWA examples I have found that many manifest have icon URLs but they are actually broken, 404 or errors in the relative path.

I think that "Manifest contains icons at least 144px" should not be true if the URL is broken, or at least have another way to point it out in the report.

brendankenny:
The DevTools fetches the icons for display in the Application panel. It doesn't appear to go through the page itself, though, as the icon fetch doesn't show up in the network panel like the fetch of the manifest itself does.

@paulirish any idea where the Application panel gets those? Is it something Lighthouse can piggyback on?

paulirish:
Application panel just reads the icon URLs and does a `new URL(iconURL, manifestURL)` and `image.src`'s it:

![image](https://cloud.githubusercontent.com/assets/39191/25461026/81520280-2a9b-11e7-9fe3-c92d4fb96394.png)

but yeah we should validate these icons don't 404.

we probably want to request from inside the host page.. (however devtools doesn't use any fancy auth)

bitterjug:
I ran into this today. Lighthouse says "User can be prompted to Install the Web App" but no A2HS appears on my phone and I ended up inspecting it via remote debug tools and triggering the A2HS to see "The site cannot be installed: icon downloaded from the manifest was empty or corrupted" errors.

docluv:
I am getting a similar error as bitterjug. 

Error while trying to use the following icon from the Manifest: http://localhost:57661/meta/android/android-launchericon-144-144.png (Download error or resource isn't a valid image)
11:00:24.067 localhost/:1 Site cannot be installed: icon downloaded from the manifest was empty or corrupted

The image loads fine in any image editor and browser. The URL returns the image, etc. So I am wondering what triggers this error and how to correct it.

filt3rek:
Hej guys !

I still have this error when launching an audit with devtools : (Download error or resource isn't a valid image) when trying to load the android-144x144 image
Is it a bug or do I miss something please ?

connorjclark:
DT release cycle is long. Will be months before it is fixed there. fix is in master, we are about to do a 6.0 release to npm, but DT is longer b/c of Chromium release cycle

PR #6230, opened 2018-10-12
core(icons): defer to manifest-icon type hint for image type

exterkamp:
**Summary**
Makes the icon pngSizedAtLeast function check the file extension _only if there is no typehint_ like Chrome does.

**Related Issues/PRs**
fixes: #6208

### DIFF
--- a/lighthouse-core/lib/icons.js
+++ b/lighthouse-core/lib/icons.js
@@ -33,13 +33,13 @@ function pngSizedAtLeast(sizeRequirement, manifest) {
   /** @type {Array<string>} */
   const flattenedSizes = [];
   iconValues
+    // filter out icons with no typehint + extension is not .png
     // filter out icons with a typehint that is not 'image/png'
-    .filter(icon => (!icon.value.type.value) ||
+    .filter(icon => (!icon.value.type.value && (icon.value.src.value &&
+      new URL(icon.value.src.value).pathname.endsWith('.png'))) ||
       (icon.value.type.value &&
       icon.value.type.value === 'image/png'))
     // filter out icons that are not png
-    .filter(icon => icon.value.src.value &&
-      new URL(icon.value.src.value).pathname.endsWith('.png'))
     .forEach(icon => {
       // check that the icon has a size
       if (icon.value.sizes.value) {
--- a/lighthouse-core/lib/icons.js
+++ b/lighthouse-core/lib/icons.js
@@ -33,13 +33,16 @@ function pngSizedAtLeast(sizeRequirement, manifest) {
   /** @type {Array<string>} */
   const flattenedSizes = [];
   iconValues
-    // filter out icons with no typehint + extension is not .png
-    // filter out icons with a typehint that is not 'image/png'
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/icons.js
```diff
@@ -33,13 +33,13 @@ function pngSizedAtLeast(sizeRequirement, manifest) {
   /** @type {Array<string>} */
   const flattenedSizes = [];
   iconValues
+    // filter out icons with no typehint + extension is not .png
     // filter out icons with a typehint that is not 'image/png'
-    .filter(icon => (!icon.value.type.value) ||
+    .filter(icon => (!icon.value.type.value && (icon.value.src.value &&
```
brendankenny: this is long enough now that we should just make a function block. It's more verbose, but what about something like
```js
.filter(icon => {
  const typeHint = icon.value.type.value;
  if (typeHint) {
    // if a type hint is present, filter out icons that are not 'image/png'.
    return typeHint === 'image/png';
  }

  // otherwise, fall back to filtering on the icons' extension.
  const src = icon.value.src.value;
  return src && new URL(src).pathname.endsWith('.png');
}).forEach(icon => {
  // ...
```

lighthouse-core/lib/icons.js
```diff
@@ -33,13 +33,13 @@ function pngSizedAtLeast(sizeRequirement, manifest) {
...
+      new URL(icon.value.src.value).pathname.endsWith('.png'))) ||
       (icon.value.type.value &&
       icon.value.type.value === 'image/png'))
     // filter out icons that are not png
```
brendankenny: delete old comment

lighthouse-core/test/lib/icons-test.js
```diff
@@ -278,7 +278,9 @@ describe('Icons helper', () => {
       assert.equal(icons.pngSizedAtLeast(192, manifest.value).length, 0);
     });
 
-    it('fails with an icon that has a png typehint but is not png', () => {
+    it('succeeds with an icon that has a png typehint but is not png', () => {
+      // We will believe your typehints until such time that we can fetch the image and decode it
+      // TODO: fetch images and decode them to check real filetype like in Chrome
```
brendankenny: maybe
`// Note: we will believe your typehints until we can fetch the image and decode it. See https://github.com/GoogleChrome/lighthouse/issues/789`

(TODO is preferably not in the test)

lighthouse-core/test/lib/icons-test.js
```diff
@@ -301,29 +303,5 @@ describe('Icons helper', () => {
       const manifest = manifestParser(manifestSrc, EXAMPLE_MANIFEST_URL, EXAMPLE_DOC_URL);
       assert.equal(icons.pngSizedAtLeast(192, manifest.value).length, 1);
     });
-
-    it('fails with a non-png icon that has query params in url', () => {
```
brendankenny: seems like both of these are still valuable, just changing the descriptions and switching the 0s to 1s (and they'll be ready to switch back and go when we do fix #789)

PR #6233, opened 2018-10-12
docs(new-audits): revise issue template

paulirish:
there was some nonsense in here about the "full config". But I also wanted to prioritize defining what it looks like in the report.

### DIFF
--- a/docs/new-audits.md
+++ b/docs/new-audits.md
@@ -22,8 +22,10 @@ Lighthouse audits that surface in the report should be:
   - What detailed info is shown in all cases, or just failure cases?
   - How would passing/failing be defined?
 #### How is this audit different from existing ones?
-#### What % of developers/pages will this impact? (estimates OK, data points preferred) 
-#### How is the new audit making a better web for end users? (data points preferred)
+#### What % of developers/pages will this impact? 
+_(Estimates OK, data points preferred)_
+#### How is the new audit making a better web for end users?
+_(Data points preferred)_
 #### What is the resourcing situation? 
   - Who will create the audits, write the documentation, and maintain both?
 #### Any other links or documentation that we should check out?
--- a/docs/new-audits.md
+++ b/docs/new-audits.md
@@ -19,15 +19,17 @@ Lighthouse audits that surface in the report should be:
 ```
 #### Provide a basic description of the audit
 #### How would the audit appear in the report? 
-  - What detailed info is shown in all cases, or just failure cases?
-  - How would passing/failing be defined?
+<!-- How would the test look when passing? Would there be additional details available?
+     How would the test look when failing? What additional details are available? 
+     If the details are tabular, what are the columns?
+     If not obvious, how would passing/failing be defined? -->
 #### How is this audit different from existing ones?
 #### What % of developers/pages will this impact? 
-_(Estimates OK, data points preferred)_
+<!-- (Estimates OK, data points preferred) -->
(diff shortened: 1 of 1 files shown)


docs/new-audits.md
```diff
@@ -12,19 +12,24 @@ Lighthouse audits that surface in the report should be:
...
-   - How much support is needed from the Lighthouse team?
-   - Any other links or documentation that we should check out?
+```
+#### Provide a basic description of the audit
+#### How would the audit appear in the report? 
+  - What detailed info is shown in all cases, or just failure cases?
+  - How would passing/failing be defined?
+#### How is this audit different from existing ones?
+#### What % of developers/pages will this impact? 
+_(Estimates OK, data points preferred)_
```
exterkamp: put `<!--`comments`-->` around this instead of _italics_?

docs/new-audits.md
```diff
@@ -12,19 +12,24 @@ Lighthouse audits that surface in the report should be:
...
+#### How is the new audit making a better web for end users?
+_(Data points preferred)_
```
exterkamp: put `<!--`comments`-->` around this instead of _italics_?

docs/new-audits.md
brendankenny: `_`? Trying to format or something else?

docs/new-audits.md
```diff
@@ -12,19 +12,24 @@ Lighthouse audits that surface in the report should be:
...
-   - What % of developers/pages will this impact (estimates OK, data points preferred) 
-   - How is the new audit making a better web for end users? (data points preferred)
-   - What is the resourcing situation (who will create the audits, maintain the audits, and write/maintain the documentation)
-   - Do you envision this audit in the Lighthouse report or the full config in the CLI? If in the report, which section? 
-   - How much support is needed from the Lighthouse team?
-   - Any other links or documentation that we should check out?
+```
+#### Provide a basic description of the audit
+#### How would the audit appear in the report? 
+  - What detailed info is shown in all cases, or just failure cases?
```
brendankenny: this is a weird way of phrasing this, but I don't have a great alternative that's also short. What's the goal here?
- what are all the possible states (pass and fail) of the test results
- what information should be shown in each case to help the user improve
?
(but phrased better :)

docs/new-audits.md
exterkamp: In the gh issue they would paste this into this would show up as `_italics_`

brendankenny: oh I see. Yeah that is confusing if you *aren't* copy/pasting :)

docs/new-audits.md
```diff
@@ -12,19 +12,26 @@ Lighthouse audits that surface in the report should be:
...
-   - Any other links or documentation that we should check out?
+```
+#### Provide a basic description of the audit
+#### How would the audit appear in the report? 
+<!-- How would the test look when passing? Would there be additional details available?
+     How would the test look when failing? What additional details are available? 
+     If the details are tabular, what are the columns?
+     If not obvious, how would passing/failing be defined? -->
+#### How is this audit different from existing ones?
+#### What % of developers/pages will this impact?
```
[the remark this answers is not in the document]
brendankenny: Yeah, the pro/con weighting for it is also different than it once was because now we have `notApplicable` and can just hide it, so visual noise is less of an issue (though still a concern in the limit). Now it's more maintenance/run time cost as the potential negative.

PR #6278, opened 2018-10-15
extension: stop report bug prompt for file:// urls

patrickhulce:
**Summary**
I think someone's already mentioned this but we don't need to prompt folks to file bugs for this one.

image


patrickhulce:
> someone want to update the audits panel as well? :)

audits panel should already be handling this properly :)
https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/audits2/Audits2Controller.js?q=PageAuditabilityChanged+-f:out&dr=CSs&l=99

PR #6282, opened 2018-10-15
extension: simplify bundle building

brendankenny:
(we should probably expand the pr lint to include the other bundles besides `extension` :)

This is a simplification of the building of the bundles (mostly dt/extension).

Primary change is that `extension-entry.js` no longer relies on importing `devtools-entry.js`, so all the entry points are independent of each other (they all just rely on core). There were two functions `extension-entry.js` was still depending on:
- `getDefaultCategories()` (which returns the categories in the default config)
- `getDefaultConfigForCategories()` (which returns a config that extends the default config but sets `onlyCategories` to the passed in categories)

Turns out `getDefaultCategories()` wasn't used in `devtools-entry.js` anyways, so just moved it into `extension-entry.js`. `getDefaultConfigForCategories()`, meanwhile, is so short and straightforward this just duplicates the function in both files.

Things become a lot simpler to understand without the layering/pseudo-inheritance. Things are either set for the consumer of the file (e.g. on `self` for devtools) or exposed at the module level for unit testing.

The other set of changes are a few simple fixes in the extension gulpfile, deleting stuff that wasn't actually doing anything.


lighthouse-extension/app/src/devtools-entry.js
```diff
@@ -45,27 +42,15 @@ function runLighthouseInWorker(port, url, flags, categoryIDs) {
...
-
 /** @param {(status: [string, string, string]) => void} listenCallback */
 function listenForStatus(listenCallback) {
   log.events.addListener('status', listenCallback);
 }
 
 if (typeof module !== 'undefined' && module.exports) {
-  // export for extension-entry to require (via browserify).
+  // export for require()ing (via browserify).
   module.exports = {
```
brendankenny: since there's no unit tests for this bundle, it's not clear we actually want this `module.exports` block anymore. Do we?

lighthouse-extension/app/src/extension-entry.js
```diff
@@ -180,23 +207,23 @@ if (typeof window !== 'undefined' && 'chrome' in window && chrome.runtime) {
 }
 
 if (typeof module !== 'undefined' && module.exports) {
-  // Export for importing types into popup.js, require()ing into unit tests.
+  // Export for importing types into popup.js and require()ing into unit tests.
```
brendankenny: `require()ing into unit tests` is also a lie, but could be true someday :) Meanwhile this block is needed for the mentioned `popup.js` use

lighthouse-extension/fs-transform.js
```diff
@@ -1,37 +0,0 @@
-/**
```
brendankenny: had no effect on the output. Maybe it was required for some ancient version of speedline that pulled in catapult?

lighthouse-extension/gulpfile.js
```diff
@@ -115,27 +114,18 @@ gulp.task('html', () => {
...
 
 gulp.task('chromeManifest', () => {
-  const manifestOpts = {
-    buildnumber: false,
-    background: {
-      target: `scripts/${CONSUMERS.EXTENSION.dist}`,
-    },
-  };
   return gulp.src('app/manifest.json')
-  .pipe(chromeManifest(manifestOpts))
```
brendankenny: this may have come from the original generator. Not entirely sure what it's supposed to do, maybe copy over deps? But since the only target we give is the script bundle, and because we copy that ourselves below after browserifying it, this also has no effect but copying over the manifest file (which we can do more simply)

lighthouse-extension/gulpfile.js
```diff
@@ -198,26 +188,15 @@ gulp.task('browserify-lighthouse', () => {
...
   return gulp.src([
     'app/src/popup.js',
-  ], {read: false})
-    .pipe(tap(file => {
-      let bundle = browserify(file.path); // , {debug: true}); // for sourcemaps
-      bundle = applyBrowserifyTransforms(bundle);
-
-      // Inject the new browserified contents back into our gulp pipeline
-      file.contents = bundle.bundle();
-    }))
```
brendankenny: `popup.js` doesn't `require()` any other files and hasn't in a long time, so no need to run it through browserify.

brendankenny:
Aside from the entry file split, diffing a generated `lighthouse-extension/dist` against one generated by master yields only a difference of the header commit-hash comment and the browserify wrapper of `popup.js` (which does nothing since it calls itself), so not much going on here but cleaning up some cruft

lighthouse-extension/app/src/devtools-entry.js
patrickhulce: Yeah I'm not sure we need it either, but +1 to being paranoid for now :)

lighthouse-extension/app/src/lightrider-entry.js
```diff
@@ -27,7 +27,7 @@ const LR_PRESETS = {
  * @return {Promise<string|Array<string>|void>}
  */
 async function runLighthouseInLR(connection, url, flags, {lrDevice, categoryIDs, logAssets}) {
-  // Certain fixes need to kick-in under LR, see https://github.com/GoogleChrome/lighthouse/issues/5839
+  // Certain fixes need to kick in under LR, see https://github.com/GoogleChrome/lighthouse/issues/5839
```
patrickhulce: 😆

lighthouse-extension/gulpfile.js
```diff
@@ -115,27 +114,18 @@ gulp.task('html', () => {
...
-    },
-  };
   return gulp.src('app/manifest.json')
-  .pipe(chromeManifest(manifestOpts))
   .pipe(gulp.dest(distDir));
 });
 
 function applyBrowserifyTransforms(bundle) {
-  // Fix an issue with imported speedline code that doesn't brfs well.
-  return bundle.transform('./fs-transform', {global: true})
```
patrickhulce: 🎉

PR #6283, opened 2018-10-15
core: save native HTMLElement.prototype.matches function

### LINKED ISSUE #5934: .matches is not a function causes many audits to error
Hi, I was using object HTML tag to exploit element resize detection. Unfortunately, I found through lighthouse that this is detected as using "plugins" a.k.a embedded flash/silverlight/etc even when actually not embedding anything. I  strongly suggest lighthouse/google to check for the data attribute of object tag before flagging it as "plugin" usage since object tag is one of the most efficient/cross-browser compatible ways of detecting element resize events, therefore only count as a use of "plugins" if it actually has a data attribute defined. 

Anyway, after removing all object tags to my surprise I found I was still getting nagged about the use of "plugins". Digging through the code, and actually checking the source of Lighthouse plugin test,  I managed to find the problem, apparently if you define for an element a matches property, this will trigger a false positive for "use of plugins", and maybe other tests too. 

For example 
```
el.matches = "don't hate me :("; 
```
will cause "use of plugins" audit to fail, it will even fail when initialized to null.

My wild guess is that you are using matches internally to flag elements and doing something along the lines of 

```
el.hasOwnProperty("matches") {
 warnUseOfPlugins();
}
```

If this is the case and you are injecting properties to parse the DOM, you should be using a safe scope such as 
```
el._lighthouse_matches
```

I hope this helps to nail down the issue and hope this behavior does not reflect to the actual SEO algorithm.

BTW good job with Lighthouse tool, it's getting more and more awesome each day!

Here is the smallest example I can think of, to prove the concept.
```
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8" />
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1">
  <title>Page Title</title>
</head>
<body>
  I don't use plugins, I swear!
  <div id="test"></div>
  <script>
    document.querySelector("#test").matches = "don't hate me :(";
  </script>
</body>
</html>
```
image

Hoten:
**Summary**

HTMLElement.prototype.matches was being overriden (by userland code), so we now store the native implementation at `HTMLElement.prototype.__matches` and use that instead.

Not sure how to make a test for this. If you use the sample HTML provided in the associated ticket, you'll see that the error no longer occurs.

**Related Issues/PRs**
Fixes #5934

### DIFF
--- a/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
+++ b/lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
@@ -108,6 +108,23 @@
 </head>
 <body>
 
+<!-- Various sites like to assign HTMLElements a custom `matches` property. See issue #5934 -->
+
+<!-- EmbeddedContent will process these elements -->
+<object id="5934a"></object>
+<object id="5934b"></object>
+
+<script>
+  // Ensure gatherers still work when individual elements override '.matches'
+  document.getElementById("5934a").matches = "blahblah";
+  Object.defineProperty(document.getElementById("5934b"), 'matches', {
+    value: "blahblah"
+  });
+  
+  // Ensure gatherers still work when the prototype is messed with
+  HTMLElement.prototype.matches = { value: "blahblah" };
+</script>
+
 <div>
   <h2>Do better web tester page</h2>
   <span>Hi there!</span>
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -1137,7 +1137,8 @@ class Driver {
   async cacheNatives() {
     await this.evaluateScriptOnNewDocument(`window.__nativePromise = Promise;
         window.__nativeError = Error;
-        window.__nativeURL = URL;`);
+        window.__nativeURL = URL;
+        window.__ElementMatches = Element.prototype.matches;`);
   }
 
   /**
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -90,7 +90,7 @@ function getElementsInDocument(selector) {
   /** @param {NodeListOf<Element>} nodes */
   const _findAllElements = nodes => {
     for (let i = 0, el; el = nodes[i]; ++i) {
-      if (!selector || el.matches(selector)) {
+      if (!selector || window.__ElementMatches.call(el, selector)) {
         results.push(el);
       }
       // If the element has a shadow root, dig deeper.


Hoten:
Done, thanks for pointing me towards the smoke tests.

Hoten:
Paul found this err'ing sites:

https://www.fitreisen.de/
http://historie.brann.no/

I confirmed they work after these changes.

Hoten:
OK, the first link fails due to specific elements having a custom property descriptor (overriding `matches`), but the second one fails due to overriding `matches` somewhere on the prototype chain (I assume, did not confirm).

paulirish:
patch looks great. some of the smoketests are being a problem. Perhaps there's an extra console error now?

Hoten:
yeah, seems that the two elements I added produces an extra log statement. It also makes for two fewer render blocking requests ...

I guess these smoke tests are pretty fragile? I'll update the new expected values, let me know if that is incorrect.

Output before changing expectations:

```
lighthouse git:(issue-5934-matches-is-not-a-fn) ✗ yarn smoke dbw
yarn run v1.10.1
$ node lighthouse-cli/test/smokehouse/run-smoke.js dbw
Running ONLY smoketests for: dbw

Smoketest batch: parallel-second
dbw smoketest starting…

dbw smoketest results:
Command failed: node lighthouse-cli/test/smokehouse/smokehouse.js --config-path=lighthouse-cli/test/smokehouse/dbw-config.js --expectations-path=dobetterweb/dbw-expectations.js

Doing a run of 'http://localhost:10200/dobetterweb/dbw_tester.html'...
$ node lighthouse-cli/index.js http://localhost:10200/dobetterweb/dbw_tester.html --config-path=/Users/cjamcl/src/lighthouse/lighthouse-cli/test/smokehouse/dbw-config.js --output-path=smokehouse-32337.report.json --output=json --quiet --port=0
Asserting expected results match those found. (http://localhost:10200/dobetterweb/dbw_tester.html)
  ✓ final url: http://localhost:10200/dobetterweb/dbw_tester.html

  ✘ difference at errors-in-console.details.items.length
              expected: 6
                 found: 7

          found result:
      {
        "id": "errors-in-console",
        "title": "Browser errors were logged to the console",
        "description": "Errors logged to the console indicate unresolved problems. They can come from network request failures and other browser concerns.",
        "score": 0,
        "scoreDisplayMode": "binary",
        "rawValue": 7,
        "details": {
          "type": "table",
          "headings": [
            {
              "key": "url",
              "itemType": "url",
              "text": "URL"
            },
            {
              "key": "description",
              "itemType": "code",
              "text": "Description"
            }
          ],
          "items": [
            {
              "source": "other",
              "description": "Application Cache Error event: Manifest fetch failed (404) http://localhost:10200/dobetterweb/clock.appcache",
              "url": "http://localhost:10200/dobetterweb/dbw_tester.html"
            },
            {
              "source": "network",
              "description": "Failed to load resource: the server responded with a status of 404 (Not Found)",
              "url": "http://localhost:10200/dobetterweb/dobetterweb/dbw_tester.css?delay=100"
            },
            {
              "source": "network",
              "description": "Failed to load resource: the server responded with a status of 404 (Not Found)",
              "url": "http://localhost:10200/dobetterweb/unknown404.css?delay=200"
            },
            {
              "source": "network",
              "description": "Failed to load resource: the server responded with a status of 404 (Not Found)",
              "url": "http://localhost:10200/dobetterweb/fcp-delayer.js?delay=5000"
            },
            {
              "source": "network",
              "description": "Failed to load resource: the server responded with a status of 404 (Not Found)",
              "url": "http://localhost:10200/favicon.ico"
            },
            {
              "source": "Runtime.exception",
              "description": "TypeError: Cannot set property 'media' of null\n    at setTimeout (http://localhost:10200/dobetterweb/dbw_tester.html:24:22)",
              "url": "http://localhost:10200/dobetterweb/dbw_tester.html"
            },
            {
              "source": "Runtime.exception",
              "description": "Error: An error\n    at http://localhost:10200/dobetterweb/dbw_tester.html:63:38",
              "url": "http://localhost:10200/dobetterweb/dbw_tester.html"
            }
          ]
        }
      }


  ✘ difference at render-blocking-resources.details.items.length
              expected: 7
                 found: 5

          found result:
      {
        "id": "render-blocking-resources",
        "title": "Eliminate render-blocking resources",
        "description": "Resources are blocking the first paint of your page. Consider delivering critical JS/CSS inline and deferring all non-critical JS/styles. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).",
        "score": 0.47,
        "rawValue": 986,
        "displayValue": "Potential savings of 990 ms",
        "details": {
          "type": "opportunity",
          "headings": [
            {
              "key": "url",
              "valueType": "url",
              "label": "URL"
            },
            {
              "key": "totalBytes",
              "valueType": "bytes",
              "label": "Size (KB)"
            },
            {
              "key": "wastedMs",
              "valueType": "timespanMs",
              "label": "Potential Savings (ms)"
            }
          ],
          "items": [
            {
              "url": "http://localhost:10200/dobetterweb/dobetterweb/dbw_tester.css?delay=100",
              "totalBytes": 171,
              "wastedMs": 377
            },
            {
              "url": "http://localhost:10200/dobetterweb/unknown404.css?delay=200",
              "totalBytes": 171,
              "wastedMs": 377
            },
            {
              "url": "http://localhost:10200/dobetterweb/dbw_tester.css?delay=2200",
              "totalBytes": 853,
              "wastedMs": 377
            },
            {
              "url": "http://localhost:10200/dobetterweb/dbw_partial_a.html?delay=200",
              "totalBytes": 768,
              "wastedMs": 377
            },
            {
              "url": "http://localhost:10200/dobetterweb/dbw_tester.css?delay=3000&capped",
              "totalBytes": 853,
              "wastedMs": 227
            }
          ],
          "overallSavingsMs": 986
        }
      }

  Correctly passed 15 assertions

15 passing
2 failing
smoketest-dbw: 9636.795ms
dbw smoketest complete.

We have 1 failing smoketests: dbw
error Command failed with exit code 1.
```

Hoten:
My assumption is that some sort of costly processing happens with each element, and it is throwing off the expected result of all those `?delays`.

patrickhulce:
@Hoten yeah those smoketest errors definitely aren't expected. It looks like something is wrong with the HTML that's causing the rest of the elements to be dropped. 

```
            {
              "source": "Runtime.exception",
              "description": "TypeError: Cannot set property 'media' of null\n    at setTimeout (http://localhost:10200/dobetterweb/dbw_tester.html:24:22)",
              "url": "http://localhost:10200/dobetterweb/dbw_tester.html"
            },
```

is an error that shouldn't be there. A different test can't find `amp-style-link` which is just a few lines down from the code you added. Maybe there's something squirrely with the `<object>`s or script tag?

Also for the smoketest I'm thinking we want to redefine matches, not just use it to make sure our workaround is working-around something :)

paulirish:
> Maybe there's something squirrely with the <object>s or script tag?

they're within the `<head>` tag right now. gotta move it down into the `<body>`

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -27,6 +27,15 @@
   }, 1500);
 </script>
 
+<object id="5934a"></object>
+<object id="5934b"></object>
+<script>
+  document.getElementById("5934a").matches = "See #5934";
```
brendankenny: can you add a comment on why these lines are here so we aren't mystified at some point in the future :)

lighthouse-core/gather/driver.js
```diff
@@ -1137,7 +1137,8 @@ class Driver {
   async cacheNatives() {
     await this.evaluateScriptOnNewDocument(`window.__nativePromise = Promise;
         window.__nativeError = Error;
-        window.__nativeURL = URL;`);
+        window.__nativeURL = URL;
+        window.__ElementMatches = Element.prototype.matches;`);
```
brendankenny: 👍👍 on the `__ElementMatches` instead of modifying the prototype

Hoten:
`objects` in `head` was definitely the issue. Doh.

> Also for the smoketest I'm thinking we want to redefine matches, not just use it to make sure our workaround is working-around something :)

@patrickhulce I'll add a line that also redefines `HTMLElement.prototype.matches`. I think we should keep the others, as my first attempt at a solution (`HTMLElement.prototype.__matches = HTMLElement.prototype.matches`) failed to fix scenarios where individual objects (not prototypes) have a `matches` property that hid access to `HTMLElement.prototype.matches`.

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
Hoten: yup, done

patrickhulce:
Oh ya for sure didn't mean to suggest removing what you've already got 👍 :)

Issue #6284, opened 2018-10-15
TTI Metric in PWA & performance reports do not match


mdiblasio:
#### Provide the steps to reproduce
1. Run LH on https://swipelife.tinder.com/

#### What is the current behavior?
TTI in [PWA report (11s) and performance report (5s)](https://www.webpagetest.org/lighthouse.php?test=181002_RA_de2324bbe18d4bbf3f3f5790456d08b9&run=1#pwa) do not match, making it unclear if the metric is measured different in each report (e.g. is throttling different? Does TTI in performance not take into account network requests?)

#### What is the expected behavior?
TTI should match

#### Environment Information
* Affected Channels: 
* Lighthouse version:  3.2.0 
* Node.js version:
* Operating System:

patrickhulce:
Thanks for filing! How did you run lighthouse @mdiblasio? The audit is making sure your TTI is fast enough *on mobile networks* so if you didn't run with one of our approved presets the numbers will be different.

exterkamp:
We might want a visual indication between the two TTI's if they are derived with different methods.  I know I was confused by this since they are different.  Something like this on the audit's TTI explanation:
`Interactive (on Simulated Mobile Network) at 11s` vs just `Interactive at 11s`

cheneytsai:
This was run through WPT (Moto G4, Chrome, 3G, Dulles) which under my understanding should run LH with the same conditions which in this case is a 3G network. This results in the following metrics:

First Interactive (WPT): >5.9s
Time to Interactive (LH - Performance): 5.0s
Interactive on 3G (LH - PWA): 11s

[Webpagetest Link](https://www.webpagetest.org/result/181002_RA_de2324bbe18d4bbf3f3f5790456d08b9/)

patrickhulce:
Thanks for sharing the report @cheneytsai! 

#6162 and #6286 should help with some of the problems here. Locally, simulated performance reflect WPT results quite accurately, but in the load-fast-enough-for-pwa we're double throttling the CPU which is probably what's inflating the numbers. I'll do a followup to #6286 to fix this.

![image](https://user-images.githubusercontent.com/2301202/47055068-34b10880-d17b-11e8-9c02-2990ab9c7e3d.png)

PR #6286, opened 2018-10-16
core(audits): clarify diff between load-fast-enough-for-pwa's TTI and perf TTI

exterkamp:
**Summary**
~In load-fast-enough-for-PWA the interactive when "provided" + mobile throttling is defaulted to force a simulated TTI, not the provided TTI (as shown in the top as the true TTI).  @brendankenny remembers this being purposeful, but it does make inconsistent numbers.  @patrickhulce do you have some context?~

Instead, will provide clearer explanation for why the TTIs might be different.

**TODO:** 
If we do want to revert to this pass through of settings we need to update a test in `lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js`

**Related Issues/PRs**
fixes: #6284

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -54,10 +54,12 @@ class LoadFastEnough4Pwa extends Audit {
 
     // Override settings for interactive if provided throttling was used and network
     // throttling was not consistent with standard `mobile network throttling`
-    const override = context.settings.throttlingMethod === 'provided' &&
-    !isDeepEqual(context.settings.throttling, mobileThrottling);
+    const override = context.settings.throttlingMethod === 'provided' ||
+      !isDeepEqual(context.settings.throttling, mobileThrottling);
 
-    const settings = override ? Object.assign({}, context.settings, settingOverrides): 
+    const displayValueFinal = override?displayValueTextWithOverride: displayValueText;
+
+    const settings = override ? Object.assign({}, context.settings, settingOverrides):
       context.settings;
 
     const metricComputationData = {trace, devtoolsLog, settings};
@@ -70,7 +72,7 @@ class LoadFastEnough4Pwa extends Audit {
     /** @type {string|undefined} */
     let explanation;
     if (!score) {
-      displayValue = [override?displayValueTextWithOverride: displayValueText, tti.timing / 1000];
+      displayValue = [displayValueFinal, tti.timing / 1000];
       explanation = 'Your page loads too slowly and is not interactive within 10 seconds. ' +
         'Look at the opportunities and diagnostics in the "Performance" section to learn how to ' +
         'improve.';
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -16,6 +16,9 @@ const Audit = require('./audit');
 const mobileThrottling = require('../config/constants').throttling.mobileSlow4G;
 const Interactive = require('../gather/computed/metrics/interactive.js');
 
+const displayValueText = `Interactive at %d\xa0s`;
+const displayValueTextWithOverride = `Interactive on simulated mobile network at %d\xa0s`;
+
 // Maximum TTI to be considered "fast" for PWA baseline checklist
 //   https://developers.google.com/web/progressive-web-apps/checklist
 const MAXIMUM_TTI = 10 * 1000;
@@ -48,11 +51,14 @@ class LoadFastEnough4Pwa extends Audit {
     // If throttling was default devtools or lantern slow 4G throttling, then reuse the given settings
     // Otherwise, we'll force the usage of lantern slow 4G.
     const settingOverrides = {throttlingMethod: 'simulate', throttling: mobileThrottling};
-    const settings =
-      context.settings.throttlingMethod !== 'provided' &&
-      isDeepEqual(context.settings.throttling, mobileThrottling)
-        ? context.settings
-        : Object.assign({}, context.settings, settingOverrides);
+
+    // Override settings for interactive if provided throttling was used and network
+    // throttling was not consistent with standard `mobile network throttling`
+    const override = context.settings.throttlingMethod === 'provided' &&
+    !isDeepEqual(context.settings.throttling, mobileThrottling);
+
+    const settings = !override? context.settings:
+      Object.assign({}, context.settings, settingOverrides);
 
     const metricComputationData = {trace, devtoolsLog, settings};
     const tti = await Interactive.request(metricComputationData, context);
@@ -64,7 +70,7 @@ class LoadFastEnough4Pwa extends Audit {
     /** @type {string|undefined} */
     let explanation;
     if (!score) {
-      displayValue = [`Interactive at %d\xa0s`, tti.timing / 1000];
+      displayValue = [override?displayValueTextWithOverride: displayValueText, tti.timing / 1000];
       explanation = 'Your page loads too slowly and is not interactive within 10 seconds. ' +
         'Look at the opportunities and diagnostics in the "Performance" section to learn how to ' +
         'improve.';
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -57,8 +57,8 @@ class LoadFastEnough4Pwa extends Audit {
     const override = context.settings.throttlingMethod === 'provided' &&
     !isDeepEqual(context.settings.throttling, mobileThrottling);
 
-    const settings = !override? context.settings:
-      Object.assign({}, context.settings, settingOverrides);
+    const settings = override ? Object.assign({}, context.settings, settingOverrides): 
+      context.settings;
 
     const metricComputationData = {trace, devtoolsLog, settings};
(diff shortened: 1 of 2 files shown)


patrickhulce:
Yes this is intentional. 

> as shown in the top as the true TTI
True to whom? :)

When using `provided` the TTI returned is probably not 3G/4G, so it's not actually checking the purpose of the audit. If they use `devtools` or `simulated` with our approved settings it uses the same number, but they can't just run it on a fast connection to pass the "Page load is fast enough on *mobile networks*" audit :)

exterkamp:
@patrickhulce modified this to have branching text for when the TTI is overridden.

lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
```diff
@@ -78,4 +78,27 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
...
+
+    const artifacts = {
+      traces: {defaultPass: longTrace},
+      devtoolsLogs: {defaultPass: devtoolsLog},
+    };
+
+    const settings = {throttlingMethod: 'provided', throttling: {rttMs: 40, throughput: 100000}};
+    const result = await FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()});
+    expect(result.displayValue).toContain('Interactive on simulated mobile network at %d\xa0s');
+    expect(result.rawValue).toBeGreaterThan(10000);
```
exterkamp: @patrickhulce I expected this test to work with the same config as `fails a bad TTI value` but I had to bump up the durations to get it to be over 10s.  Thoughts on why?

patrickhulce: without more events/network requests there's no way for the simulator to know that the events couldn't just happen all one after the other, i.e. under simulation it's going to mostly ignore the timestamps in the original trace and just have 6 100ms tasks one after the other.

If you want to simplify this and make it a little more clear, I'd probably move this down to just a single 5s task in the trace. With the multiplier it should be longer than 10s guaranteeing late TTI, without the multiplier it would be well under 10s.

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -48,11 +51,14 @@ class LoadFastEnough4Pwa extends Audit {
...
     const settingOverrides = {throttlingMethod: 'simulate', throttling: mobileThrottling};
-    const settings =
-      context.settings.throttlingMethod !== 'provided' &&
-      isDeepEqual(context.settings.throttling, mobileThrottling)
-        ? context.settings
-        : Object.assign({}, context.settings, settingOverrides);
+
+    // Override settings for interactive if provided throttling was used and network
+    // throttling was not consistent with standard `mobile network throttling`
+    const override = context.settings.throttlingMethod === 'provided' &&
```
patrickhulce: nit: mind naming this `shouldOverride` or something a bit more clear it's a boolean rather than the override itself :)

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -64,7 +70,7 @@ class LoadFastEnough4Pwa extends Audit {
     /** @type {string|undefined} */
     let explanation;
     if (!score) {
-      displayValue = [`Interactive at %d\xa0s`, tti.timing / 1000];
+      displayValue = [override?displayValueTextWithOverride: displayValueText, tti.timing / 1000];
```
patrickhulce: nit: spacing around `?`

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -48,11 +51,14 @@ class LoadFastEnough4Pwa extends Audit {
...
-      isDeepEqual(context.settings.throttling, mobileThrottling)
-        ? context.settings
-        : Object.assign({}, context.settings, settingOverrides);
+
+    // Override settings for interactive if provided throttling was used and network
+    // throttling was not consistent with standard `mobile network throttling`
+    const override = context.settings.throttlingMethod === 'provided' &&
+    !isDeepEqual(context.settings.throttling, mobileThrottling);
+
+    const settings = !override? context.settings:
```
patrickhulce: nit: spacing around `?`, also for the ternary I usually find it easier if there's not another layer of negation, mind switching the order and removing `!`?

lighthouse-core/audits/load-fast-enough-for-pwa.js
brendankenny: is the idea you can be `provided`, but if you *say* you were slow enough in `settings.throttling` you'll pass?

lighthouse-core/audits/load-fast-enough-for-pwa.js
brendankenny: maybe just set it to a variable above this rather than nesting it in the array

lighthouse-core/audits/load-fast-enough-for-pwa.js
patrickhulce: oh whoops thanks @brendankenny I didn't notice the conditions got only half flipped in this

@exterkamp  you meant to `!` this whole condition right? we need to flip the `&&` to `||`

the original idea was that you are fine if you are..
A) Using default simulated settings
OR 
B) Using default devtools settings

i.e. if you are using provided OR you don't have the mobileThrottling settings, you need to be overriden

brendankenny:
@exterkamp I know you have a lot on your plate, but do you want to take another look?

lighthouse-core/audits/load-fast-enough-for-pwa.js
exterkamp: Done and done!  I misread the original if statement.

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -48,11 +51,16 @@ class LoadFastEnough4Pwa extends Audit {
...
-      isDeepEqual(context.settings.throttling, mobileThrottling)
-        ? context.settings
-        : Object.assign({}, context.settings, settingOverrides);
+
+    // Override settings for interactive if provided throttling was used and network
+    // throttling was not consistent with standard `mobile network throttling`
+    const override = context.settings.throttlingMethod === 'provided' ||
+      !isDeepEqual(context.settings.throttling, mobileThrottling);
+
+    const displayValueFinal = override?displayValueTextWithOverride: displayValueText;
```
patrickhulce: nit: spacing :)

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -48,11 +51,16 @@ class LoadFastEnough4Pwa extends Audit {
...
+
+    const settings = override ? Object.assign({}, context.settings, settingOverrides):
```
patrickhulce: nit: spacing before `:` :)

lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
```diff
@@ -78,4 +78,37 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
     expect(result.rawValue).toBeGreaterThan(2000);
     expect(Math.round(result.rawValue)).toMatchSnapshot();
   });
+
+  it('overrides when throttling is modified but method is not provided', async () => {
```
patrickhulce: this test name doesn't seem to line up with what's happening isn't the method provided as devtools here?

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -48,11 +51,16 @@ class LoadFastEnough4Pwa extends Audit {
     // If throttling was default devtools or lantern slow 4G throttling, then reuse the given settings
     // Otherwise, we'll force the usage of lantern slow 4G.
     const settingOverrides = {throttlingMethod: 'simulate', throttling: mobileThrottling};
-    const settings =
-      context.settings.throttlingMethod !== 'provided' &&
-      isDeepEqual(context.settings.throttling, mobileThrottling)
-        ? context.settings
-        : Object.assign({}, context.settings, settingOverrides);
+
+    // Override settings for interactive if provided throttling was used and network
```
brendankenny: *or* network throttling was not consistent...

lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
```diff
@@ -78,4 +78,37 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
...
+      traces: {defaultPass: trace},
+      devtoolsLogs: {defaultPass: devtoolsLog},
+    };
+
+    const settings = {throttlingMethod: 'devtools', throttling: {rttMs: 40, throughput: 100000}};
+    const result = await FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()});
+    expect(result.rawValue).toBeGreaterThan(2000);
+  });
+
+  it('override with simulated result fails a bad simulated TTI value', async () => {
```
brendankenny: maybe something like
`it('overrides when throttling is "provided" and fails the simulated TTI value', async () => {`
?

lighthouse-core/audits/load-fast-enough-for-pwa.js
brendankenny: heh, updated linting is failing travis on this now :P

Issue #6272, opened 2018-10-14
Sites hitting 45s pageload timeout


paulirish:
These URLs pulled from some LR logs. They all reliably hit the 45 timeout

* http://asiaherbassociation.com/en/ (maybe its the 8 youtube embeds?)
* http://www.freedirectory.yonola.com (looks like there's a hanging network request that never times out)
* https://www.expedia.com/Las-Vegas-Hotels-Mandarin-Oriental.h3016186.Hotel-Information?chkin=10/14/2018&chkout=10/20/2018&rm1=a2
* https://www.eltern.de
* http://www.skimmelfrit.dk
* http://www.jessicaadams.com
* http://ouc.com
* http://nosybesesiles.com/galleries/radama-lodge/
* http://ceylonskiychay.ru
* http://autotintexpress.ca
* ~https://ilkayuyarkaba.av.tr/tag/istifaya-zorlanan-iscinin-kidem-tazminati NET::ERR_CERT_SYMANTEC_LEGACY prevents the load. But we don't recognize the problem and we hang permanently at "Retrieving devtoolsLog and network records"~  split out as #6287

patrickhulce:
I've gone through the URLs and think a generic 10s max TTFB timeout would solve most of this class of problems, i.e. if we hear nothing back from a request for 10s we treat it as failed for the purposes of network quiescence. Unfortunately, that doesn't work for the case where a page needs a script for DCL. Not really sure what to do about it other than some separate DCL timeout that's much less than 45 seconds.

Details of each URL found below (lots were not reproducible on my machine on master, @paulirish were these logs taken from the version of LH that didn't have the frame hanging fix already?)


> http://asiaherbassociation.com/en/

Intermittently reproducible. When it does hang, it is indeed the 8 youtube embeds not receiving any response.

> http://www.freedirectory.yonola.com 

Never reaches DCL, http://stats.extrawatch.com/agent.js.php?projectId=3967&env=ExtraWatchNoCMSEnv never responds.

> https://www.expedia.com/Las-Vegas-Hotels-Mandarin-Oriental.h3016186.Hotel-Information?chkin=10/14/2018&chkout=10/20/2018&rm1=a2

Unable to reproduce locally.

> https://www.eltern.de

Unable to reproduce locally, though it is slow loading.

> http://www.skimmelfrit.dk

Unable to reproduce locally.

> http://www.jessicaadams.com

Unable to reproduce locally.

> http://ouc.com

[...]

PR #6293, opened 2018-10-16
core: remove no-websql audit

patrickhulce:
**Summary**
This audit is relevant to only 0.6% of sites on HTTPArchive (only 8k out of ~1.3M) . The `websql` gatherer takes at least 500ms for every site that *passes* the audit, so we're spending 1,200,000 seconds a month just on WPT doing nothing :)

My claim is that this audit would not pass our `new-audit` criteria doc if we were to validate it today and should be removed.

**Related Issues**
thought of while investigating #6272 

<details>

<summary>
See Query
</summary>

```sql
SELECT
  JSON_EXTRACT_SCALAR(report, '$.audits.no-websql.score') as score, COUNT(*)
FROM
  httparchive.lighthouse.2018_09_01_mobile
 GROUP BY score
```
</details>


brendankenny:
screen shot 2018-10-16 at 10 18 43

someone playing around this this?

brendankenny:
oh, maybe it's new actions or whatever

patrickhulce:
> Need to also delete WebSQL in artifacts.d.ts

@brendankenny do you have a way for force vscode show the .d.ts files and have them in find/replace

```
        "**/*.d.ts": {
            "when": "$(basename).ts"
        }
```

hasn't been doing the trick clearly :)

brendankenny:
oh, no, and I hadn't even noticed that was a problem...

I was thinking that maybe we need to reverse `artifacts.d.ts` and have it pull the artifact types off of the gatherers that produce them (which would have flagged the change in this case), but I believe our earlier reasoning was that by having the source of truth in that file, bugs in gatherers would be more easily caught, which still seems sound...

PR #6295, opened 2018-10-16
extension: split browserifying and extension bundling into separate scripts

brendankenny:
This sets the stage for splitting up the extension from the `devtools-entry.js` and `lightrider-entry.js` files. The only thing they have in common is the browserifying of an entry-point script, so it makes sense to pull that out into its own file.

Also removes all the gulp stuff, because all our gulp tasks were either copying files from one location to another (gulp not needed), or tapping into the stream and modifying the files individually (negating the whole point of gulp :). With the `build-extension.js` script now 95% just putting together the extension assets, this is easy to do.

The output of the various bundles and their location doesn't change with this PR. I'll file an issue with ideas for reorganizing so we can all bikeshed :)


lighthouse-extension/build-extension.js
```diff
@@ -0,0 +1,114 @@
...
+ * @return {Promise<void>}
+ */
+async function copyAssets() {
+  await cpy('app/*.html', distDir);
+  await cpy('app/styles/**/*.css', `${distDir}/styles`);
+  await cpy('app/images/**/*', `${distDir}/images`);
+  await cpy('app/manifest.json', distDir);
+
+  // locales (currently non-functional)
+  await cpy('_locales/**', `../${distDir}`, {
```
brendankenny: if we add our locales to the extension, it might still make sense to use these files to do things like localize the webstore descriptions and extension name and stuff, so I left them in

lighthouse-extension/bundle-builder.js
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * @fileoverview Script to bundle lighthouse entry points so that they can be run
+ * in the browser (as long as they have access to a debugger protocol Connection).
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const LighthouseRunner = require('../lighthouse-core/runner');
```
brendankenny: this is all almost exactly the same as it was in the gulpfile, just unindented a few times. One good piece of news: our `astw` hack is no longer necessary because `insert-module-globals` moved off off `lexical-scope` in https://github.com/browserify/insert-module-globals/pull/76 :)

lighthouse-extension/package.json
```diff
@@ -5,31 +5,11 @@
...
-  },
-  "resolutions": {
-    "browserify/insert-module-globals/lexical-scope/astw": "2.2.0",
-    "**/astw/acorn": "5.5.3"
+    "build": "node ./build-extension.js",
+    "test": "mocha test/**/*-test.js",
+    "clean": "rimraf dist",
+    "package": "node ./build-extension.js package"
   },
+  "devDependencies": {},
```
brendankenny: they were all devDeps (so no npm package penalty), we no longer support an environment requiring only one version of any npm module, there were even fewer deps now, and I got a few typescript conflicts between here and the parent directory, so it made sense to move them all up to the root directory

brendankenny:
(vast majority of the line count is `yarn.lock` changes)

lighthouse-extension/build-extension.js
```diff
@@ -0,0 +1,114 @@
...
+
+  await makeDir(`${distDir}/scripts`);
+  fs.writeFileSync(`${distDir}/scripts/popup.js`, popupSrc);
+}
+
+/**
+ * @return {Promise<void>}
+ */
+async function copyAssets() {
+  await cpy('app/*.html', distDir);
```
wardpeet: Promise.all might be faster for all these tasks?

lighthouse-extension/bundle-builder.js
brendankenny: oh, and no objection to future moving to a different bundler or whatever. Left exactly the same to stay out of that question for now :)

It should be even easier since bundling is isolated from the specifics of extension building.

lighthouse-extension/bundle-builder.js
```diff
@@ -0,0 +1,147 @@
...
+
+    bundleStream.pipe(writeStream);
+  });
+}
+
+/**
+ * Minimally minify a javascript file, in place.
+ * @param {string} filePath
+ */
+function minifyScript(filePath) {
```
wardpeet: should we call this minify? 😛

wardpeet:
should we call it `lighthouse-extension/build-extension.js ?` it would seem to me that we just want to build the extension

lighthouse-extension/build-extension.js
brendankenny: it doesn't appear to make much of a difference (around 50ms either way on my machine). I imagine things are disk-bound at that point, since it's browserifying (which touches a whoooole lot of files) at the same time

wardpeet: yeah indeed

brendankenny: buuut, no reason not to do the globbing and copying all in one call (and in theory, cpy will handle load balancing the streams). Fixed

package.json
```diff
@@ -85,15 +91,21 @@
...
+    "babel-plugin-syntax-object-rest-spread": "^6.13.0",
+    "brfs": "^1.6.1",
+    "browserify": "^16.2.3",
     "bundlesize": "^0.14.4",
     "chalk": "^2.4.1",
     "codecov": "^2.2.0",
     "commitizen": "^2.9.6",
     "conventional-changelog-cli": "^1.3.4",
     "coveralls": "^2.11.9",
+    "cpy": "^7.0.1",
```
patrickhulce: not a big deal, but I generally preferred to use https://github.com/shelljs/shelljs instead of bringing in all the individual commands (cpy,del,make-dir,etc) with different APIs to remember

brendankenny:
> should we call it lighthouse-extension/build-extension.js ? it would seem to me that we just want to build the extension

I get this question now :)

I can rename it if you feel strongly, but we'll probably just switch it back when #6299 is fixed and the file only builds the extension. It seems somewhat ok for now because the status quo *is* that "building the extension" (e.g. `yarn build-extension`) builds all of them...

wardpeet:
I don't have strong feelings so feel free to keep it for now

package.json
brendankenny: yeah, I think we should consider that for the future. `del` and `make-dir` bring in a crazy number of deps :) Some of the simpler bash scripts could switch over as well so they work cross platform.

PR #6297, opened 2018-10-16
misc(logging): add hanging request logging in driver

patrickhulce:
**Summary**
Adds `--verbose` logging of the URLs or DCL event that we're currently waiting on.

**Related Issues/PRs**
#6272


lighthouse-core/gather/driver.js
```diff
@@ -484,6 +485,28 @@ class Driver {
     const networkStatusMonitor = this._networkStatusMonitor;
 
     const promise = new Promise((resolve, reject) => {
+      const logStatus = () => {
```
paulirish: can you add a comment to explain what this is for?
the name doesnt totally help, but i agree the name should remain this short. :)

patrickhulce: done :)

lighthouse-core/gather/driver.js
```diff
@@ -499,19 +500,45 @@ class Driver {
...
+          log.verbose('Driver', 'Waiting on DomContentLoaded');
+          return;
+        }
+
+        const inflightRecords = networkStatusMonitor.getInflightRecords();
+        // If there are more than 20 inflight requests, load is still in full swing.
+        // Wait until it calms down a bit to be a little less spammy.
+        if (inflightRecords.length < 20) {
+          for (const record of inflightRecords) {
+            log.verbose('Driver', `Waiting on ${record.url.slice(0, 120)} to finish`);
```
brendankenny: we should probably remove all the `network busy`, `network semi-quiet` logs in `NetworkRecorder`. At least I've never found them useful, and this will give equivalent info (except fully quiet, I guess)

lighthouse-core/gather/driver.js
```diff
@@ -499,19 +500,45 @@ class Driver {
...
 
+      // We frequently need to debug why LH is still waiting for the page.
+      // This listener is added to all network events to verbosely log what URLs we're waiting on.
+      const logStatus = () => {
+        if (!hasDCLFired) {
+          log.verbose('Driver', 'Waiting on DomContentLoaded');
+          return;
+        }
+
+        const inflightRecords = networkStatusMonitor.getInflightRecords();
```
brendankenny: i've been a little concerned about the length of our `NetworkRecord` callbacks. I was seeing 1+ms for some while doing some OYB profiling a while back, and there are a lot of callbacks.

Can we detect whether or not we're in `verbose` mode before doing any of this work? Short of `settings` in here or changing lighthouse-logger again, I'm not sure there is, sadly.

lighthouse-core/gather/driver.js
patrickhulce: FWIW I still found them helpful to track exactly when the idle timeout would be started and cancelled (and to break up the repeated sets of URLs when filtering out the '<= event:verbose' events which are the real crazy chatty ones!), but agreed the wall of them every time we call `recordsFromLogs` is super annoying.

lighthouse-core/gather/driver.js
patrickhulce: +1 to this and another lighthouse-logger change, I want one level in between verbose and default (debug or info perhaps?) that doesn't contain *every* protocol message

paulirish: Good call. Yah not showing events from Network domain would be my number one request.

Issue #6296, opened 2018-10-16
Add timeout to all protocol commands


Hoten:
Some commands already do this (see `getRequestContent` and `_evaluateInContext` in the driver). We could benefit from implementing a timeout for all commands.

where: `sendCommand` in `lighthouse-core/gather/connections/connection.js`

Would prevent issues like #6287

Hoten:
for later
```
// Protocol timeout failures
  PROTOCOL_TIMEOUT: {
    code: 'PROTOCOL_TIMEOUT',
    message: strings.protocolTimeout,
  },
-----
  protocolTimeout: `Waiting for DevTools protocol response has exceeded the allotted time.`,
```

PR #6300, opened 2018-10-17
core: bail if encounter insecure ssl cert, to avoid hanging forever

### LINKED ISSUE #6287: Certificate error hangs Lighthouse
repro:

```
lighthouse https://ilkayuyarkaba.av.tr/tag/istifaya-zorlanan-iscinin-kidem-tazminati
```

The site fails to load and shows the NET::ERR_CERT_SYMANTEC_LEGACY error, but Lighthouse doesn't recognize this. After the 45s timeout it just hangs permanently on "Retrieving devtoolsLog and network records".

We should detect the certificate failure early and fail with a new pageLoadError state. Maybe something like `CERT_ERR_DOCUMENT _REQUEST` ?

I _think_ there's something on the protocol which represents this situation.. Something in the `Security` domain? Will require some investigation.   https://badssl.com/ might be useful though they dont have a symantec cert situation. Just other ones.

-----

Issue broken out from #6272

Hoten:
After encountering an SSL error, some commands to the Network/Page domains either took awhile to timeout, or hung forever (only seemed to happen for `ERR_CERT_SYMANTEC_LEGACY`). You can see this at https://ilkayuyarkaba.av.tr/tag/istifaya-zorlanan-iscinin-kidem-tazminati (use canary).

If one of the passes generates an SSL error, just bail. A report is generated with a useful error message at the top.

Fixes #6287.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -850,6 +850,24 @@ class Driver {
   }
 
   /**
+   * @param {number} [timeout]
+   * @return {Promise<LH.Crdp.Security.SecurityStateChangedEvent>}
+   */
+  getSecurityState(timeout = 1000) {
+    return new Promise((resolve, reject) => {
+      const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
+      const asyncTimeout = setTimeout((_ => reject(err)), timeout);
+
+      this.sendCommand('Security.enable');
+      this.once('Security.securityStateChanged', (e) => {
+        clearTimeout(asyncTimeout);
+        resolve(e);
+        this.sendCommand('Security.disable').catch(reject);
+      });
+    });
+  }
+
+  /**
    * @param {string} name The name of API whose permission you wish to query
    * @return {Promise<string>} The state of permissions, resolved in a promise.
    *    See https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query.
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -859,10 +859,10 @@ class Driver {
       const asyncTimeout = setTimeout((_ => reject(err)), timeout);
 
       this.sendCommand('Security.enable');
-      this.once('Security.securityStateChanged', (e) => {
+      this.once('Security.securityStateChanged', state => {
         clearTimeout(asyncTimeout);
-        resolve(e);
-        this.sendCommand('Security.disable').catch(reject);
+        resolve(state);
+        this.sendCommand('Security.disable');
       });
     });
   }
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -849,22 +849,19 @@ class Driver {
     });
   }
 
-  /**
-   * @param {number} [timeout]
-   * @return {Promise<LH.Crdp.Security.SecurityStateChangedEvent>}
-   */
-  getSecurityState(timeout = 1000) {
-    return new Promise((resolve, reject) => {
-      const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
-      const asyncTimeout = setTimeout((_ => reject(err)), timeout);
-
-      this.once('Security.securityStateChanged', state => {
-        clearTimeout(asyncTimeout);
-        resolve(state);
-        this.sendCommand('Security.disable');
-      });
-      this.sendCommand('Security.enable');
+  async listenForSecurityStateChanges() {
+    this.on('Security.securityStateChanged', state => {
+      this.lastSecurityState = state;
     });
+    await this.sendCommand('Security.enable');
+  }
+
+  /**
+   * @return {LH.Crdp.Security.SecurityStateChangedEvent}
+   */
+  getSecurityState() {
+    // @ts-ignore
+    return this.lastSecurityState;
   }
 
   /**
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -858,12 +858,12 @@ class Driver {
       const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
       const asyncTimeout = setTimeout((_ => reject(err)), timeout);
 
-      this.sendCommand('Security.enable');
       this.once('Security.securityStateChanged', state => {
         clearTimeout(asyncTimeout);
         resolve(state);
         this.sendCommand('Security.disable');
       });
+      this.sendCommand('Security.enable');
     });
   }
 
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -860,7 +860,12 @@ class Driver {
    * @return {LH.Crdp.Security.SecurityStateChangedEvent}
    */
   getSecurityState() {
-    // @ts-ignore
+    if (!this.lastSecurityState) {
+      // happens if 'listenForSecurityStateChanges' is not called,
+      // or if some assumptions about the Security domain are wrong
+      throw new Error('Expected a security state.');
+    }
+
     return this.lastSecurityState;
   }
 
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -78,6 +78,13 @@ class Driver {
       // properties. See https://github.com/Microsoft/TypeScript/pull/22348.
       this._eventEmitter.emit(event.method, event.params);
     });
+
+    /**
+     * Used for monitoring network status events during gotoURL.
+     * @type {?LH.Crdp.Security.SecurityStateChangedEvent}
+     * @private
+     */
+    this._lastSecurityState = null;
   }
 
   static get traceCategories() {
@@ -851,7 +858,7 @@ class Driver {
 
   async listenForSecurityStateChanges() {
     this.on('Security.securityStateChanged', state => {
-      this.lastSecurityState = state;
+      this._lastSecurityState = state;
     });
     await this.sendCommand('Security.enable');
   }
@@ -860,13 +867,13 @@ class Driver {
    * @return {LH.Crdp.Security.SecurityStateChangedEvent}
    */
   getSecurityState() {
-    if (!this.lastSecurityState) {
+    if (!this._lastSecurityState) {
       // happens if 'listenForSecurityStateChanges' is not called,
       // or if some assumptions about the Security domain are wrong
       throw new Error('Expected a security state.');
     }
 
-    return this.lastSecurityState;
+    return this._lastSecurityState;
   }
 
   /**
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -310,12 +310,7 @@ class GatherRunner {
       trace,
     };
 
-    if (!pageLoadError) {
-      // Disable throttling so the afterPass analysis isn't throttled
-      // This will hang if there was a security error. But, there is no
-      // need to throttle if there is such an error. See #6287
-      await driver.setThrottling(passContext.settings, {useThrottling: false});
-    }
+    await driver.setThrottling(passContext.settings, {useThrottling: false});
 
     for (const gathererDefn of gatherers) {
       const gatherer = gathererDefn.instance;
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -174,7 +174,7 @@ class GatherRunner {
   }
 
   /**
-   * Returns an error if the original network request failed or wasn't found.
+   * Returns an error if the security state is insecure.
    * @param {LH.Crdp.Security.SecurityStateChangedEvent} securityState
    * @return {LHError|undefined}
    */
@@ -185,7 +185,7 @@ class GatherRunner {
         .filter(exp => exp.securityState === 'insecure')
         .map(exp => exp.description);
       errorDef.message += ` ${insecureDescriptions.join(' ')}`;
-      return new LHError(errorDef, {fatal: true});
+      return new LHError(errorDef);
     }
   }
 
@@ -310,6 +310,7 @@ class GatherRunner {
       trace,
     };
 
+    // Disable throttling so the afterPass analysis isn't throttled
     await driver.setThrottling(passContext.settings, {useThrottling: false});
 
     for (const gathererDefn of gatherers) {
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -258,7 +258,7 @@ class GatherRunner {
    * object containing trace and network data.
    * @param {LH.Gatherer.PassContext} passContext
    * @param {Partial<GathererResults>} gathererResults
-   * @return {Promise<[LH.Gatherer.LoadData, LH.LighthouseError | undefined]>}
+   * @return {Promise<LH.Gatherer.LoadData>}
    */
   static async afterPass(passContext, gathererResults) {
     const driver = passContext.driver;
@@ -295,6 +295,7 @@ class GatherRunner {
       networkRecords,
       devtoolsLog,
       trace,
+      pageLoadError,
     };
 
     if (!pageLoadError) {
@@ -326,7 +327,7 @@ class GatherRunner {
     }
 
     // Resolve on tracing data using passName from config.
-    return [passData, pageLoadError];
+    return passData;
   }
 
   /**
@@ -423,8 +424,7 @@ class GatherRunner {
         await driver.setThrottling(options.settings, passConfig);
         await GatherRunner.beforePass(passContext, gathererResults);
         await GatherRunner.pass(passContext, gathererResults);
-        const [passData, pageLoadError] =
-          await GatherRunner.afterPass(passContext, gathererResults);
+        const passData = await GatherRunner.afterPass(passContext, gathererResults);
 
         // Save devtoolsLog, but networkRecords are discarded and not added onto artifacts.
         baseArtifacts.devtoolsLogs[passConfig.passName] = passData.devtoolsLog;
@@ -450,6 +450,7 @@ class GatherRunner {
           firstPass = false;
         }
 
+        const pageLoadError = passData.pageLoadError;
         if (pageLoadError && pageLoadError.code === LHError.errors.INSECURE_DOCUMENT_REQUEST.code) {
           // Some protocol commands will hang, so let's just bail. See #6287
           break;
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -176,16 +176,16 @@ class GatherRunner {
   /**
    * Returns an error if the security state is insecure.
    * @param {LH.Crdp.Security.SecurityStateChangedEvent} securityState
-   * @return {LHError|undefined}
+   * @throws {LHError}
    */
-  static checkForSecurityIssue({securityState, explanations}) {
+  static assertNoSecurityIssues({securityState, explanations}) {
     if (securityState === 'insecure') {
       const errorDef = {...LHError.errors.INSECURE_DOCUMENT_REQUEST};
       const insecureDescriptions = explanations
         .filter(exp => exp.securityState === 'insecure')
         .map(exp => exp.description);
       errorDef.message += ` ${insecureDescriptions.join(' ')}`;
-      return new LHError(errorDef);
+      throw new LHError(errorDef);
     }
   }
 
@@ -297,10 +297,7 @@ class GatherRunner {
       passContext.LighthouseRunWarnings.push(pageLoadError.friendlyMessage);
     }
 
-    const securityError = this.checkForSecurityIssue(driver.getSecurityState());
-    if (securityError) {
-      throw securityError;
-    }
+    this.assertNoSecurityIssues(driver.getSecurityState());
 
     // Expose devtoolsLog, networkRecords, and trace (if present) to gatherers
     /** @type {LH.Gatherer.LoadData} */
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -146,12 +146,12 @@ class GatherRunner {
 
   /**
    * Returns an error if the original network request failed or wasn't found.
-   * @param {Driver} driver
    * @param {string} url The URL of the original requested page.
    * @param {Array<LH.Artifacts.NetworkRequest>} networkRecords
-   * @return {Promise<LHError|undefined>}
+   * @param {LH.Crdp.Security.SecurityStateChangedEvent} securityState
+   * @return {LHError|undefined}
    */
-  static async getPageLoadError(driver, url, networkRecords) {
+  static getPageLoadError(url, networkRecords, securityState) {
     const mainRecord = networkRecords.find(record => {
       // record.url is actual request url, so needs to be compared without any URL fragment.
       return URL.equalWithExcludedFragments(record.url, url);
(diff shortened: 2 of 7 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
     });
   }
 
+  /**
+   * @param {number} [timeout]
+   * @return {Promise<LH.Crdp.Security.SecurityStateChangedEvent>}
+   */
+  getSecurityState(timeout = 1000) {
```
Hoten: this will eventually be replaced by #6296

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
...
+    return new Promise((resolve, reject) => {
+      const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
+      const asyncTimeout = setTimeout((_ => reject(err)), timeout);
```
Hoten: this will eventually be replaced by #6296

lighthouse-core/gather/gather-runner.js
```diff
@@ -165,6 +166,20 @@ class GatherRunner {
     } else if (mainRecord.hasErrorStatusCode()) {
       errorDef = {...LHError.errors.ERRORED_DOCUMENT_REQUEST};
       errorDef.message += ` Status code: ${mainRecord.statusCode}.`;
+    } else if (!mainRecord.finished) {
```
Hoten: AFAIK, an insecure security state will always have `.finished` set to false on the network records. I wanted to avoid calling these Security commands if not needed.

lighthouse-core/gather/gather-runner.js
```diff
@@ -313,7 +333,7 @@ class GatherRunner {
     }
 
     // Resolve on tracing data using passName from config.
-    return passData;
+    return [passData, pageLoadError];
```
Hoten: thoughts?

lighthouse-core/gather/gather-runner.js
```diff
@@ -165,6 +166,20 @@ class GatherRunner {
     } else if (mainRecord.hasErrorStatusCode()) {
       errorDef = {...LHError.errors.ERRORED_DOCUMENT_REQUEST};
       errorDef.message += ` Status code: ${mainRecord.statusCode}.`;
+    } else if (!mainRecord.finished) {
+      // could be security error
```
paulirish: I think this may be just certificate errors. Seems like most of the connection errors are `mainRecord.failed === true`.

lighthouse-core/gather/gather-runner.js
```diff
@@ -165,6 +166,20 @@ class GatherRunner {
...
+      const securityState = await driver.getSecurityState();
```
paulirish: Small nit but I think it'd be slightly better to collect this on L288 and pass that into getPageLoadError() rather than all of driver.

lighthouse-core/gather/gather-runner.js
```diff
@@ -165,6 +166,20 @@ class GatherRunner {
...
+      if (securityState.securityState === 'insecure') {
+        errorDef = {...LHError.errors.INSECURE_DOCUMENT_REQUEST};
+        const insecureDescriptions = securityState.explanations
+          .filter(exp => exp.securityState === 'insecure')
+          .map(exp => exp.description)
+          .join(' ');
+        errorDef.message += ` Insecure: ${insecureDescriptions}`;
```
paulirish: Agree the explanation gathering is the best we can do.
For this line i think it could just be 

```suggestion
        errorDef.message += ` ${insecureDescriptions.join(' ')}`;
```

lighthouse-core/gather/gather-runner.js
```diff
@@ -410,7 +430,8 @@ class GatherRunner {
         await driver.setThrottling(options.settings, passConfig);
         await GatherRunner.beforePass(passContext, gathererResults);
         await GatherRunner.pass(passContext, gathererResults);
-        const passData = await GatherRunner.afterPass(passContext, gathererResults);
+        const [passData, pageLoadError] =
```
paulirish: @brendankenny @patrickhulce either of you have an idea on how else to deliver this error?

see also the L462 here

lighthouse-core/lib/strings.js
```diff
@@ -11,7 +11,9 @@ module.exports = {
   badTraceRecording: `Something went wrong with recording the trace over your page load. Please run Lighthouse again.`,
   pageLoadTookTooLong: `Your page took too long to load. Please follow the opportunities in the report to reduce your page load time, and then try re-running Lighthouse.`,
   pageLoadFailed: `Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.`,
+  pageLoadFailedInsecure: `The URL you have provided does not have a valid security certificate.`,
```
paulirish: hmm technically not just certificate, but admittedly most other security errors fall through the ERRORED_DOC_REQ path right now.  how about

> ... does not have valid security credentials.

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
...
+
+      this.sendCommand('Security.enable');
+      this.once('Security.securityStateChanged', (e) => {
```
patrickhulce: somewhat tangential... 

@paulirish are we just paranoid for doing all our listeners before enabling or do we actually need to? I seem to remember dgozman scolding me for not trusting in JS microtasks when we had listeners before `.enable` :)

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
...
+  /**
+   * @param {number} [timeout]
+   * @return {Promise<LH.Crdp.Security.SecurityStateChangedEvent>}
+   */
+  getSecurityState(timeout = 1000) {
+    return new Promise((resolve, reject) => {
+      const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
+      const asyncTimeout = setTimeout((_ => reject(err)), timeout);
+
+      this.sendCommand('Security.enable');
```
patrickhulce: also, @paulirish is Security domain still broken on Android, is this going to break us on real devices again?

lighthouse-core/gather/gather-runner.js
patrickhulce: WDYT about attaching it to `passData` instead? It's a relatively grab-bag object of what a happened during a pass, so it makes some sense :)

lighthouse-core/gather/gather-runner.js
patrickhulce: super nit: WDYT about sprinkling some nice destructuring here, `const {securityState, explanations} =`? somethin' about `securityState.securityState` just spoke to me :)

lighthouse-core/gather/gather-runner.js
brendankenny: `pageLoadError.fatal = true; throw pageLoadError;` :P

lighthouse-core/gather/gather-runner.js
Hoten: Sure. I'll add it as an optional property. I avoided it in the first place b/c I didn't want to modify all usages of this type, but I completely forgot about optional properties :P

lighthouse-core/gather/driver.js
paulirish: hmm good call.

This code here would be a problem if there was an `await` on the `.enable` call above. which seems fairly reasonable.

So yeah +1 to defining the once handler before we flip enable() on

lighthouse-core/gather/driver.js
paulirish: they fixed android like a year ago! yay

image

patrickhulce: well someone's behind the times 😆

lighthouse-core/gather/gather-runner.js
Hoten: I didn't even notice that :O

lighthouse-core/gather/driver.js
Hoten: good point, fixed.

I don't quite understand the execution model here. It's clear to me why `await`ing the enable would schedule registering the listener much too late, but I don't understand how this original code worked (enable without `await`ing, then register a listener via `.once`). Why does enabling the Security domain still occur after the listener is registered?

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
...
+  /**
+   * @param {number} [timeout]
+   * @return {Promise<LH.Crdp.Security.SecurityStateChangedEvent>}
+   */
+  getSecurityState(timeout = 1000) {
+    return new Promise((resolve, reject) => {
+      const err = new LHError(LHError.errors.SECURITY_STATE_TIMEOUT);
+      const asyncTimeout = setTimeout((_ => reject(err)), timeout);
+
+      this.once('Security.securityStateChanged', state => {
```
brendankenny: does this check break if someone else has been listening to the `Security` domain? It would be nice to get a non-event based version of this in the protocol

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,24 @@ class Driver {
...
+        clearTimeout(asyncTimeout);
+        resolve(state);
+        this.sendCommand('Security.disable');
+      });
+      this.sendCommand('Security.enable');
```
brendankenny: If we want this to be a general utility function, it will need to disable the domain and remove the event listener in the rejection case. Otherwise we'll need to think how to make rejections always lead to a program exit, as the state of things might get weird (we'll have to deal with the same thing in #6296)

lighthouse-core/gather/gather-runner.js
```diff
@@ -271,7 +278,9 @@ class GatherRunner {
     const networkRecords = NetworkRecorder.recordsFromLogs(devtoolsLog);
     log.verbose('statusEnd', status);
 
-    let pageLoadError = GatherRunner.getPageLoadError(passContext.url, networkRecords);
+    const securityState = await driver.getSecurityState();
```
brendankenny: Maybe we're making this harder than it needs to be by trying to unify with with the other page load errors.

`getPageLoadError` is trying really hard to not interrupt normal control flow, inserting the promise rejection into the gather results instead of into the executing chain. Meanwhile, this security check is trying to abandon the gathering altogether.

Maybe this should just be a separate function, `checkPageSecurityState()` or whatever, called here-ish. If the page is insecure, throw in there. It'll bail back to here, which will bail up to the `catch` in `run()`. If we really want to still return something in that case, the check against `LHError.errors.INSECURE_DOCUMENT_REQUEST.code` can happen there (or rethrow if it's something different).

This also gets around extending `passData`, which makes sense, because really there isn't any other useful pass data when there's this kind of error.

WDYT?

lighthouse-core/gather/driver.js
brendankenny: > If we want this to be a general utility function, it will need to disable the domain and remove the event listener in the rejection case.

Maybe this is overkill. If it's timing out (unexpectedly), something is pretty wrong and LH probably won't recover. Disabling the domain also might not even work. OTOH, not cleaning up feels wrong :)

lighthouse-core/gather/driver.js
Hoten: I enabled just before calling this function, and yes, it does break. It hangs, and eventually rejects with `SECURITY_STATE_TIMEOUT` (which actually kills Lighthouse. oops)

Besides other users enabling the Security domain, could this fail if a developer opens the Security tab (this enables the domain) before running an audit?

lighthouse-core/gather/gather-runner.js
Hoten: Can't quite figure out how to still display a LHR in `run`'s `catch`. I marked the security LHError fatal (so it would throw), moved `baseArtifacts` up a bit (so it can be accessed in `catch`), but get this silly error:

```
Error: CSSUsage failed to provide an artifact.
    at Function.collectArtifacts (/Users/cjamcl/src/lighthouse/lighthouse-core/gather/gather-runner.js:366:15)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

Hoten:
OK, I believe I resolved all comments. Opened #6330 to discuss how to handle errors being thrown during gather-runner.

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,21 @@ class Driver {
...
+    });
+    await this.sendCommand('Security.enable');
+  }
+
+  /**
+   * @return {LH.Crdp.Security.SecurityStateChangedEvent}
+   */
+  getSecurityState() {
+    // @ts-ignore
+    return this.lastSecurityState;
```
brendankenny: should set this in the `Driver` constructor (I guess to null since `SecurityStateChangedEvent` is a relatively large object so it's probably not worth constructing an `unknown` starting value). Won't need to ts-ignore then

lighthouse-core/gather/gather-runner.js
```diff
@@ -112,6 +112,7 @@ class GatherRunner {
     await driver.cacheNatives();
     await driver.registerPerformanceObserver();
     await driver.dismissJavaScriptDialogs();
+    await driver.listenForSecurityStateChanges();
```
brendankenny: doing it this way, I guess we'll find out if there are issues with listening through the whole page load :)

But it does nicely pave the way for bailing earlier if we want to do that.

lighthouse-core/gather/gather-runner.js
```diff
@@ -172,6 +173,22 @@ class GatherRunner {
...
+   * @return {LHError|undefined}
+   */
+  static checkForSecurityIssue({securityState, explanations}) {
+    if (securityState === 'insecure') {
+      const errorDef = {...LHError.errors.INSECURE_DOCUMENT_REQUEST};
+      const insecureDescriptions = explanations
+        .filter(exp => exp.securityState === 'insecure')
+        .map(exp => exp.description);
+      errorDef.message += ` ${insecureDescriptions.join(' ')}`;
+      return new LHError(errorDef, {fatal: true});
```
brendankenny: rather than returning the error and throwing on the return value, I'd say just throw in here

lighthouse-core/gather/gather-runner.js
```diff
@@ -172,6 +173,22 @@ class GatherRunner {
     }
   }
 
+  /**
+   * Returns an error if the original network request failed or wasn't found.
```
brendankenny: need to update string

lighthouse-core/gather/gather-runner.js
```diff
@@ -172,6 +173,22 @@ class GatherRunner {
...
+   * @param {LH.Crdp.Security.SecurityStateChangedEvent} securityState
+   * @return {LHError|undefined}
+   */
+  static checkForSecurityIssue({securityState, explanations}) {
```
brendankenny: bikeshed: if it does throw in here, maybe rename `assertCertificateError` or something? Or does this catch a larger class of issues

lighthouse-core/gather/gather-runner.js
brendankenny: `{fatal: true}` shouldn't be necessary anymore

lighthouse-core/gather/gather-runner.js
```diff
@@ -288,7 +310,6 @@ class GatherRunner {
       trace,
     };
 
-    // Disable throttling so the afterPass analysis isn't throttled
```
brendankenny: restore?

lighthouse-core/lib/lh-error.js
```diff
@@ -169,6 +175,12 @@ const ERRORS = {
     message: strings.requestContentTimeout,
   },
 
+  // Protocol timeout failures
+  SECURITY_STATE_TIMEOUT: {
```
brendankenny: in the usual case for the security state check, it should just reject on an insecure security state, right? If so, we probably want to make this a more general protocol communication timeout error (anticipating #6296)

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -543,9 +545,9 @@ describe('GatherRunner', function() {
       ],
     };
 
-    return GatherRunner.afterPass({url, driver, passConfig}, {TestGatherer: []}).then(vals => {
+    return GatherRunner.afterPass({url, driver, passConfig}, {TestGatherer: []}).then(passData => {
```
brendankenny: ok to revert these lines now?

lighthouse-core/gather/driver.js
Hoten: It's ts-ignored to coerce a non-nullable value, since I believe this should always be set. I could remove it and allow this getter to be nullable, but would also need to handle that in the calling code (a case that I think will never happen). wdyt?

lighthouse-core/gather/gather-runner.js
Hoten: OK, I can do that. I was thinking it's nicer to make tests that check return values rather than throwing conditions, but just my preference, I could have it either way.

I'll remove fatal. yay, that means the resolve or throw thing really can still be removed.

lighthouse-core/gather/gather-runner.js
Hoten: It catches anytime the latest security state has an insecure property. a certificate error is one example. I don't know if it's just one of, most of them, or all possible scenarios.

lighthouse-core/lib/lh-error.js
Hoten: My plan was to modify that bit in the referenced issue. but now that we have this concrete proto definition it makes sense to make it good sooner rather than later

lighthouse-core/test/gather/gather-runner-test.js
Hoten: I think `passData` is more clear than `vals`.

lighthouse-core/lib/lh-error.js
Hoten: Oh, I actually removed the timeout stuff for security checking. I'll remove this too.

lighthouse-core/gather/driver.js
patrickhulce: If we believe it's always set but don't know for sure, WDYT about throwing an explicit error if it's not? 

This way it'll be clear to us when something unexpected happens and ts is happy by default :)

lighthouse-core/gather/gather-runner.js
patrickhulce: Just triple checking because the protocol name is getting me concerned a bit, but `securityState === 'insecure'` only when the page is HTTPS but actually isn't secure, right? Not just every time the page is served over HTTP without encryption correct? :)

patrickhulce: +1 to the idea of throwing in here with a `assert*` rename too, btw. Seems much easier and removes the cognitive load from the caller

Hoten: Good idea to check HTTP. Tested with `node lighthouse-cli http://www.httpvshttps.com/ --verbose --view` - it still generated a full LHR. I presume "insecure" is only used for HTTPS connection that don't comply with security standards (and I guess they always go hand-in-hand with interstitial security warnings).

brendankenny:
> Just triple checking because the protocol name is getting me concerned a bit, but securityState === 'insecure' only when the page is HTTPS but actually isn't secure, right?

Looking through the chromium source, the values to search for are `blink::kWebSecurityStyleInsecure` and `Security::SecurityStateEnum::Insecure` (that's what ends up setting the field in the event payload in [`security_handler.cc`](https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/security_handler.cc?sq=package:chromium&dr=CSs&g=0&l=186)).

There are two ways it's set. If someone wants to track down which one is actually in our execution path, feel free :)

- The style in [`web_url_loader_impl.cc`](https://cs.chromium.org/chromium/src/content/renderer/loader/web_url_loader_impl.cc?type=cs&q=kWebSecurityStyleInsecure&g=0&l=120-123)
```cc
if (net::IsCertStatusError(cert_status) &&
    !net::IsCertStatusMinorError(cert_status)) {
  return blink::kWebSecurityStyleInsecure;
}
```
This is per-resource, not per-page, so probably not it.
- The check in [`content_utils.cc`](https://cs.chromium.org/chromium/src/components/security_state/content/content_utils.cc?type=cs&q=kWebSecurityStyleInsecure&g=0&l=48-49)
```cc
switch (security_level) {
  // ...
  case security_state::DANGEROUS:
    return blink::kWebSecurityStyleInsecure;
  // ..
}
```

where DANGEROUS is
```
// Attempted HTTPS and failed, page not authenticated, HTTPS with
// insecure active content on the page, malware, phishing, or any other
// serious security issue that could be dangerous.
```

Either way, this assumption seems to be ok. The main extra thing seems to be malware/phishing. Presumably that information will be in the explanation put in the friendly messsage.

brendankenny:
> `HTTPS with insecure active content on the page`

oh, mixed content might be an issue as well? We should check. I'm not sure we want to throw in that case

brendankenny:
yeah, it's flagged as "neutral". Mixed content pages with passwords might trigger this (cf. https://crbug.com/647754), but it seems it will only require a error message tweak in the future if folks run into it

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -679,6 +681,43 @@ describe('GatherRunner', function() {
     });
   });
 
+  describe('#checkForSecurityIssue', () => {
```
brendankenny: `#assertNoSecurityIssues`

lighthouse-core/gather/driver.js
```diff
@@ -849,6 +849,26 @@ class Driver {
     });
   }
 
+  async listenForSecurityStateChanges() {
+    this.on('Security.securityStateChanged', state => {
+      this.lastSecurityState = state;
```
brendankenny: the type checker now allows this (in JS files as of 3.0 or 3.1), but should still declare it in the constructor so there's a clear place to check for what's on Driver, we don't dynamically mutate the object's shape, etc

lighthouse-core/gather/gather-runner.js
```diff
@@ -172,6 +173,22 @@ class GatherRunner {
     }
   }
 
+  /**
+   * Returns an error if the security state is insecure.
```
brendankenny: Throws

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -679,6 +681,43 @@ describe('GatherRunner', function() {
...
+            securityState: 'insecure',
+          },
+        ],
+        securityState: 'insecure',
+      };
+      try {
+        GatherRunner.assertNoSecurityIssues(insecureSecurityState);
+        assert.fail('expected INSECURE_DOCUMENT_REQUEST LHError');
+      } catch (err) {
+        assert.equal(err.message, 'INSECURE_DOCUMENT_REQUEST');
```
brendankenny: maybe check `err.code` as well/instead, since that's what we use for the top level `runtimeError`

lighthouse-core/gather/driver.js
Hoten: yeah this really surprised me, how it was able to infer the type. it was able to see that it could be undefined, since the ctor didn't set it.

Makes sense. This is not good:
image

I guess TSC just infers a permissive union type for `this.property`, based on all setters? What new feature of tsc 3.0/3.1 is this?

lighthouse-core/test/gather/gather-runner-test.js
Hoten: Done. Some other tests were just testing `.message`, so I added checks there too. lmk if I should just remove the `.message` assertions.

PR #6302, opened 2018-10-17
core(driver): stringify protocol error messages

paulirish:
This is an actual log from LR:

> `LHError: UNKNOWN_ERROR Unknown error encountered with message 'Protocol error (Tracing.start): [object Object]'`

The true origin is most likely... (_drumroll..._)

> [**Tracing is already started**](https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/tracing_handler.cc?l=379&rcl=6c1d24539cea713814a538600c5f298f0d7d1e25)

Our best friend. 👨‍❤️‍👨

----------

As to why this PR makes sense.. Well... I can't really explain why `message` here is sometimes an object (and otherwise a string)... but the log above seems to indicate that's precisely what's happening. Could be the flu shot making me crazy though. 🤷‍♂️

### DIFF
--- a/lighthouse-core/lib/lh-error.js
+++ b/lighthouse-core/lib/lh-error.js
@@ -59,7 +59,7 @@ class LighthouseError extends Error {
     }
 
     // otherwise fallback to building a generic Error
-    let errMsg = `(${method}): ${protocolError.message}`;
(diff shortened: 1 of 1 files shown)


patrickhulce:
lol #crushedbydevtoolsbot

Issue #6290, opened 2018-10-16
Ignore incorrect aspect ratio if the image is 1x1 pixel


SebastianEberlein:
**Feature request summary**
 Ignore incorrect aspect ratio if the image is 1x1 pixel.

**What is the motivation or use case for changing this?**
Lazyloading images. The image container has an intrinsic ratio to avoid layout shifting. The placeholder image is a 1x1 pixel transparent base64 gif as suggested [here](https://github.com/aFarkas/lazysizes#modern-transparent-srcset-pattern).

Lighthouse 3.2.0 displays the warning "Displays images with incorrect aspect ratio":

<img width="730" alt="screenshot" src="https://user-images.githubusercontent.com/15672385/47003480-98154900-d12f-11e8-8860-86d90519b5f9.png">

This is technically correct, but should be ignored. A 1x1 pixel image will almost never match the natural aspect ratio. And it is unlikely that a 1x1 pixel image is used for anything other than a placeholder.

**How is this beneficial to Lighthouse?**
Seeing this warning, developers might look for a different solution, maybe a larger placeholder image with the correct aspect ratio. But this could increase the overall page size and thus degrade performance.

patrickhulce:
Thanks for filing @SebastianEberlein! I'm not sure I follow why your placeholder image needs to be so big if you set the actual size to be 1x1 anyhow.

Maybe I'm misunderstanding but it seems like there's two possible paths with this neat trick you're using. Please correct me if I'm wrong :)

1) You use a transparent 1x1 GIF as the source image and stretch it to fill the real aspect ratio of the image it's placeholding. This will fail the audit, but it's efficient, and it seems reasonable for us to ignore images with sufficiently small *source dimensions* in LH.
2) You use a transparent GIF the same size as the image it's placeholding. This will waste bytes, but now there's no need to set the size of the placeholder and the audit will pass naturally.

IIUC, it seems like you're using a transparent GIF the same size as the image it's placeholding, but then forcing it to not take up its natural placeholder size.

With the current situation in the screenshot, there's no way of LH knowing that this seemingly regular sized image isn't being accidentally squished by some CSS rule which is one of the exact reasons the audit exists.

wardpeet:
@patrickhulce I'm pretty sure they use this lazyloading technique
`<img src="data:blankgifdatauri" width="376" height="187" data-src="realimage" />` this way the image will take up it's space and lazyloads the image when needed without getting a page to jump up and down.

[...]

PR #6305, opened 2018-10-17
core(fix): fix incorrect aspect ratio bug if the image is 1x1 pixel

CodeDem:
fixes incorrect aspect ratio bug if the image is 1x1 pixel by filtering the image less than 5 * 5
pixels dimension

**Related Issues/PRs**
#6290

### DIFF
--- a/lighthouse-core/test/audits/image-aspect-ratio-test.js
+++ b/lighthouse-core/test/audits/image-aspect-ratio-test.js
@@ -137,6 +137,16 @@ describe('Images: aspect-ratio audit', () => {
     },
   });
 
+  testImage('is placeholder image', {
+    rawValue: true,
+    clientSize: [300, 220],
+    naturalSize: [1, 1],
+    props: {
+      isCss: false,
+      usesObjectFit: false,
+    },
+  });
+
   it('skips svg images', () => {
     const result = ImageAspectRatioAudit.audit({
       ImageUsage: [
--- a/lighthouse-core/test/audits/image-aspect-ratio-test.js
+++ b/lighthouse-core/test/audits/image-aspect-ratio-test.js
@@ -137,17 +137,6 @@ describe('Images: aspect-ratio audit', () => {
     },
   });
 
-  testImage('has invalid natural sizing information', {
-    rawValue: true,
-    warning: 'Invalid image sizing information https://google.com/logo.png',
-    clientSize: [100, 100],
-    naturalSize: [0, 0],
-    props: {
-      isCss: false,
-      usesObjectFit: false,
-    },
-  });
-
   it('skips svg images', () => {
     const result = ImageAspectRatioAudit.audit({
       ImageUsage: [
--- a/lighthouse-core/audits/image-aspect-ratio.js
+++ b/lighthouse-core/audits/image-aspect-ratio.js
@@ -77,6 +77,8 @@ class ImageAspectRatio extends Audit {
       // - filter all svgs as they have no natural dimensions to audit
       return image.networkRecord &&
         image.networkRecord.mimeType !== 'image/svg+xml' &&
+        image.naturalHeight > 5 &&
+        image.naturalWidth > 5 &&
         image.width &&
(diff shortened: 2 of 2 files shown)


patrickhulce:
Thanks very much for the PR @CodeDem! You'll have to update the unit test for this audit (https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/test/audits/image-aspect-ratio-test.js). 

Right now it's asserting that we warn on invalid sizing information, but we've decided to just filter them out instead so we should update the test case accordingly 👍

patrickhulce:
Good thinking @CodeDem! Honestly though, that `invalid natural sizing information` error has not been very actionable and it's been some overzealous for no good reason over the past year. 

I'm fine with filtering them out altogether unless @brendankenny you might have strong opinions here?

wardpeet:
common is actualy 1x1 i think 5x5 is to high as well at least for lazy loading images but I might be wrong here. I only use 1x1 transparent gifs.

CodeDem:
@brendankenny @benschwarz @patrickhulce 

I agree it makes sense to only ignore  1x1 images since this looks more standard in terms of placeholders for lazyloading and thus saves size.

However, 1x1 pixel can only store single color, using 4x4 size placeholder can represent the rough overview of the image content and colors of what they are about to see.

we should let developer to choose placeholder between 1x1 to 5x5 image as per requirements, correct me if I'm wrong here. 

Eg of 1x1 image placeholder
image

Eg of 4x4 image placeholder
image

Actual image
image

Let me know your thoughts.

patrickhulce:
> is it possible under normal conditions to get a natural size of 0x0? Seems like it would usually indicate that something was wrong

It's not clear under what conditions we'd be getting a size of 0x0, and the last several times I tried looking at the URLs in Sentry for invalid image sizing none of them repro'd. I'm not sure it's worth tracking at this point.

> why < 5x5 and not just allow 1x1? Is it common to use a 4x4 image as a placeholder, for instance?

IMO, this audit is not really all that useful for very small images, there's not much to distort. Heck I'd be willing to say <20x20 could be ignored. There's something to be said for @CodeDem's example of slightly larger placeholder pixelated placeholder that are then blurred to give a better speed index. I'm not sure what our reasonable allowance should be, but I'm in favor of more than just 1x1.

CodeDem:
Alright, @patrickhulce @brendankenny shall I go ahead and commit the changes with removed test case for `invalid natural sizing information`  and image filter for images less than 5x5 ?

lighthouse-core/test/audits/image-aspect-ratio-test.js
```diff
@@ -137,17 +137,6 @@ describe('Images: aspect-ratio audit', () => {
...
-    rawValue: true,
-    warning: 'Invalid image sizing information https://google.com/logo.png',
-    clientSize: [100, 100],
-    naturalSize: [0, 0],
-    props: {
-      isCss: false,
-      usesObjectFit: false,
-    },
-  });
-
```
brendankenny: should add a test (or modify this one) that a 1x1 pixel image passes when stretched to a different aspect ratio, to make sure we don't regress in the future on your use case :)

patrickhulce: @CodeDem we want to test that an image that's similar to the issue that was filed.

i.e. an image with `{clientSize: [300, 200], naturalSize: [1, 1]}` passes the audit

CodeDem: @patrickhulce oh okay got it, so it will be just renaming test to` 'is placeholder image'` and updating it so that it won't expect  `warning: 'Invalid image sizing information https://google.com/logo.png'`  to pass basically removing the warning check!

brendankenny:
over to @patrickhulce if you want a last look

CodeDem:
Thank you @patrickhulce  and @brendankenny  for helping through my first contribution, yaay :)

Issue #6309, opened 2018-10-17
configSettings.output can be list or string


exterkamp:
#### What is the current behavior?
LHR `configSettings.output` can be either a string or a list.  For convergence with proto it needs to be one or the other.

brendankenny:
it's [always an array from the CLI](https://github.com/GoogleChrome/lighthouse/blob/3d5037ade1b0b62b6bfe89642ba0e20904809df2/typings/externs.d.ts#L156), so we'd just need to correct the module uses. It's marginally more annoying to use an array there, but not so bad.

It looks like all the client entry points all use the string version, and we'd need to update the default in `constants.js`.

patrickhulce:
I'd also argue if we're converting it to always array that we should change the name to `reports` :)

PR #6310, opened 2018-10-17
misc(proto): adjust configSetting.output in preprocessor

exterkamp:
**Summary**
Adds configSettings.output processing.

**Related Issues/PRs**
preprocessing for: #6309

### DIFF
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -25,7 +25,7 @@ function processForProto(result) {
   // Clean up the configSettings
   if (reportJson.configSettings) {
     // make sure the 'output' field is an array
-    if ('output' in reportJson.configSettings) {
+    if (reportJson.configSettings.output) {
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -22,6 +22,16 @@ function processForProto(result) {
   /** @type {LH.Result} */
   const reportJson = JSON.parse(result);
 
+  // Clean up the configSettings
+  if ('configSettings' in reportJson) {
```
brendankenny: don't need to guard this
```suggestion
  if (reportJson.configSettings) {
```

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -22,6 +22,16 @@ function processForProto(result) {
...
+    // make sure the 'output' field is an array
+    if ('output' in reportJson.configSettings) {
```
brendankenny:
```suggestion
    if (reportJson.configSettings.output) {
```

Issue #6331, opened 2018-10-17
Benchmark Index can cause a proto unmarshalling error.


exterkamp:
#### What is the current behavior?
Convert JSON -> proto with a non integer `benchmark_index` e.g. 502.485 results in a conversion error.

exterkamp:
Also `rtt_ms` `max_wait_for_load` `cpu_slowdown_multiplier `

PR #6332, opened 2018-10-17
misc(proto): convert int32's into DoubleValue in proto

exterkamp:
**Summary**
Benchmark index is actually a double!  Not an int!

**Related Issues/PRs**
fixes: #6331

### DIFF
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -125,7 +125,7 @@ class Runner {
         environment: {
           networkUserAgent: artifacts.NetworkUserAgent,
           hostUserAgent: artifacts.HostUserAgent,
-          benchmarkIndex: artifacts.BenchmarkIndex,
+          benchmarkIndex: parseInt(artifacts.BenchmarkIndex.toString()),
         },
         lighthouseVersion,
         fetchTime: artifacts.fetchTime,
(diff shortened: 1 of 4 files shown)


proto/sample_v2_round_trip.json
```diff
@@ -3262,7 +3262,7 @@
         "throttlingMethod": "devtools"
     }, 
     "environment": {
-        "benchmarkIndex": 1000, 
+        "benchmarkIndex": 1000.0,
```
brendankenny: since this is JSON, the trailing `.0` ends up in sort of in undefined behavior (it won't survive depending on the language/tool used to parse/stringify the file). Since it's a fake number anyways, should we change the `benchmarkIndex` value in `sample_v2.json` so it's always a float (e.g. 1000.5)?

Or we can leave it and worry about it when it creates a problem (again) :)

brendankenny:
I like @patrickhulce's idea better if the slowest devices really are more on the order of 10-100

lighthouse-core/runner.js
```diff
@@ -125,7 +125,7 @@ class Runner {
         environment: {
           networkUserAgent: artifacts.NetworkUserAgent,
           hostUserAgent: artifacts.HostUserAgent,
-          benchmarkIndex: artifacts.BenchmarkIndex,
+          benchmarkIndex: parseInt(artifacts.BenchmarkIndex.toString()),
```
brendankenny: `Math.round()` :)

But I think we want to do this either in `Driver.getBenchmarkIndex()` or in `ultradumbBenchmark` itself

PR #6343, opened 2018-10-19
core: remove recoverOrThrow / err.fatal

### LINKED ISSUE #6298: Remove recoverOrThrow in gather-runner
I'm told that every problem that could have been solved by throwing errors using this method has had a better solution come up. Could remove this function + any tests that set `.fatal` on an error object.

Hoten:
Fixes #6298

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -327,7 +327,7 @@ class GatherRunner {
         // Typecast pretends artifact always provided here, but checked below for top-level `throw`.
         gathererArtifacts[gathererName] = /** @type {NonVoid<PhaseResult>} */ (artifact);
       } catch (err) {
-        // return error runner to handle turning it into an error audit.
+        // Return error to runner to handle turning it into an error audit.
         gathererArtifacts[gathererName] = err;
       }
 
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -183,7 +183,7 @@ class GatherRunner {
       passContext.options = gathererDefn.options || {};
       const artifactPromise = Promise.resolve().then(_ => gatherer.beforePass(passContext));
       gathererResults[gatherer.name] = [artifactPromise];
-      await artifactPromise;
+      await artifactPromise.catch(() => {});
     }
   }
 
@@ -227,7 +227,7 @@ class GatherRunner {
       const gathererResult = gathererResults[gatherer.name] || [];
       gathererResult.push(artifactPromise);
       gathererResults[gatherer.name] = gathererResult;
-      await artifactPromise;
+      await artifactPromise.catch(() => {});
     }
   }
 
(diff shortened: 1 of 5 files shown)


Hoten:
Hmm.

@brendankenny For some reason, some tests fail now. Related to the gather lifecycle. For example: `'runs gatherer lifecycle methods strictly in sequence'`. Functionally, the only difference I see it one less await for a Promise that immediately resolves. But, I tried adding a dumb `await Promise.resolve();` where `await GatherRunner.recoverOrThrow(artifactPromise)` was, tests still fail.

Hoten:
that promise still needs to be `await`ed :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -197,7 +183,7 @@ class GatherRunner {
       passContext.options = gathererDefn.options || {};
       const artifactPromise = Promise.resolve().then(_ => gatherer.beforePass(passContext));
       gathererResults[gatherer.name] = [artifactPromise];
-      await GatherRunner.recoverOrThrow(artifactPromise);
+      await artifactPromise;
```
brendankenny: Sorry, I looked too quickly :) I should have said it'll have to be something like `await artifactPromise.catch(() => {});` (basically the equivalent of deleting just the `if (err.fatal) {throw err;}` part of `recoverOrThrow`).

`gathererResults[gatherer.name]` needs to be set to the raw promise, which may resolve or reject, which will become either the artifact or an artifact error in `collectArtifacts()`. But the `await` needs to never throw so the failure of one gatherer won't prevent moving on to the next

lighthouse-core/gather/gather-runner.js
```diff
@@ -341,8 +327,7 @@ class GatherRunner {
         // Typecast pretends artifact always provided here, but checked below for top-level `throw`.
         gathererArtifacts[gathererName] = /** @type {NonVoid<PhaseResult>} */ (artifact);
       } catch (err) {
-        // An error result must be non-fatal to not have caused an exit by now,
-        // so return it to runner to handle turning it into an error audit.
+        // return error runner to handle turning it into an error audit.
```
brendankenny:
```suggestion
        // Return error to runner to handle turning it into an error audit.
```

Hoten:
^That happened when I used GH's "apply change" feature ...

lighthouse-core/test/runner-test.js
```diff
@@ -290,7 +290,7 @@ describe('Runner', () => {
 
     // TODO: need to support save/load of artifact errors.
     // See https://github.com/GoogleChrome/lighthouse/issues/4984
-    it.skip('outputs an error audit result when required artifact was a non-fatal Error', () => {
+    it.skip('outputs an error audit result when required artifact was a Error', () => {
```
patrickhulce:
```suggestion
    it.skip('outputs an error audit result when required artifact was an Error', () => {
```

PR #6344, opened 2018-10-19
clients: reorganize file structure and build process

### LINKED ISSUE #6299: Reorganize bundles and buildles
Originally we had only `popup.js`, which split off into the `lighthouse-background.js` extension entry point, which then also hosted a DevTools entry point, which split off into its own file that the extension entry point extended, then LR lived in there and then also split off, and then finally all the entry files were independent.

In all this time, they've all lived in `lighthouse-extension/app/src/`, and when built they all ended up in `lighthouse-extension/dist/scripts/`, even though 2/3 of them now have nothing to do with the extension except they share a build script.

#6295 splits building those files and building the extension, which means we can now do a clean reorg of those files without much fuss. Ideally the result would be easily navigable and obvious in its purpose, with no weirdness based on git history that's long disappeared.

Here's my suggestion, but I don't find it completely satisfying and I'm not particularly wedded to the structure or names.
```
├── clients/
│   ├── devtools-entry.js   (could be in subdirectory but seems unnecessary)
│   ├── lightrider-entry.js
│   ├── extension/   (current top-level lighthouse-extension/)
│   |   ├── scripts/ (currently src/)
│   |   ├── styles/
│   |   ├── manifest.json
│   |   ├── popup.html
│   |   └── etc...
│   └── test/
│       ├── lightrider-entry-test.js
│       └── extension/
│          ├── extension-test.js
│          └── popup-test.js
├── build/
│   ├── build-extension.js
│   ├── bundle-builder.js
│   └── changelog-generator/   (already exists)
├── dist/   (created while building)
│   ├── lighthouse-dt-bundle.js
│   ├── lighthouse-lr-bundle.js
│   ├── extension/
│   |   └── scripts/, styles/, manifest.json, etc
│   └── extension-package
│          ├── lighthouse-3.2.0.zip
│          └── lighthouse-3.2.1.zip
├── lighthouse-cli/
├── lighthouse-core/
└── etc...
```
~"`bundles/`" is a dumb name. `clients/`? `channels/`? It's nice to describe that they are clients/channels, but they only work when bundled and then run somewhere else.~ (now using `clients/`)

That aside, it makes sense to have the entry points in a top-level directory (similar to `cli/`), but also makes sense to group them together because they aren't independent, executable entry points into the module itself (unlike the CLI, they're only useful after building).

It would also be great to move to a top-level output of the build step. `out/` is popular in some parts of the world, but we can take our pick.

Thoughts?

more requirements:
- [ ] logging of build steps
- [ ] browserify without minifying option?

brendankenny:
fixes #6299. New file structure matches the latest version of that issue, recreated here for reference:

```
├── clients/
│   ├── devtools-entry.js   (could be in subdirectory but seems unnecessary)
│   ├── lightrider-entry.js
│   ├── extension/   (current top-level lighthouse-extension/)
│   |   ├── scripts/ (currently src/)
│   |   ├── styles/
│   |   ├── manifest.json
│   |   ├── popup.html
│   |   └── etc...
│   └── test/
│       ├── lightrider-entry-test.js
│       └── extension/
│          ├── extension-test.js
│          └── popup-test.js
├── build/
│   ├── build-extension.js
│   ├── bundle-builder.js
│   └── changelog-generator/   (already exists)
├── dist/   (created while building)
│   ├── lighthouse-dt-bundle.js
│   ├── lighthouse-lr-bundle.js
│   ├── extension/
│   |   └── scripts/, styles/, manifest.json, etc
│   └── extension-package
│          ├── lighthouse-3.2.0.zip
│          └── lighthouse-3.2.1.zip
├── lighthouse-cli/
├── lighthouse-core/
└── etc...
```


brendankenny:
damn, pr lint change didn't work on itself :)

brendankenny:
And not too late to bikeshed if something doesn't look right to you now

brendankenny:
> qq: any reason it is build-all, but then test-clients? 

just dumb reasons. We could follow up by making viewer a client (it arguably is, bundling just the report generator and renderer), then it would be `build-clients`. I don't think we'd want it to be `test-all`, because that would really imply running *all* the tests, including over `core`, which `yarn test` sort of already handles (though it doesn't run integration tests like smokehouse, but these client tests are mostly integration tests via puppeteer...).

Meanwhile, why `build-all` (or `build-clients`) and not just `build`? I think because `install-all` came first, which was an effort to differentiate from `install` because we didn't want to install all deps by default. That's still kind of a concern (though the original, main reason for doing that has long disappeared), but it might be worth revisiting to simplify things.

So, all in all, someone needs to get in there and make our npm scripts sane. 55 down to like 40 or so should maybe be the goal :)

paulirish:
You'll want to manually delete `lighthouse-extension/` from your checkouts.

PR #6347, opened 2018-10-19
core: add timeout to all protocol commands

### LINKED ISSUE #6296: Add timeout to all protocol commands

Hoten:
Fixes #6296


lighthouse-core/gather/driver.js
```diff
@@ -306,61 +333,45 @@ class Driver {
...
-          return reject(new Error(`Evaluation exception: ${response.exceptionDetails.text}`));
-        }
-
-        // Protocol should always return a 'result' object, but it is sometimes undefined.  See #6026.
-        if (response.result === undefined) {
-          return reject(new Error('Runtime.evaluate response did not contain a "result" object'));
-        }
-
-        const value = response.result.value;
+  async _evaluateInContext(expression, contextId) {
```
Hoten: VSCode suggested this could be async, so I let it :) Thoughts on changing asynchronous syntax? FWIW, [V8 just improved the performance of `await` to match Promises](https://docs.google.com/document/d/1kL08cz4lR6gO5b2FATNK3QAfS8t-6K6kdk88U-n8tug/edit#heading=h.bcoo5tv0iksy).

lighthouse-core/gather/driver.js
```diff
@@ -860,27 +871,15 @@ class Driver {
...
-        clearTimeout(asyncTimeout);
-        // Ignoring result.base64Encoded, which indicates if body is already encoded
-        resolve(result.body);
-      }).catch(e => {
-        clearTimeout(asyncTimeout);
-        reject(e);
-      });
-    });
+    // Encoding issues may lead to hanging getResponseBody calls: https://github.com/GoogleChrome/lighthouse/pull/4718
+    // driver.sendCommand will handle timeout after 5s.
```
patrickhulce: Can we keep this at 1s? expanding to 5 could dramatically slow down the failure case when 3 of these go wrong :/

lighthouse-core/gather/driver.js
```diff
@@ -306,61 +333,45 @@ class Driver {
...
-          return reject(Object.assign(new Error(), value));
-        } else {
-          resolve(value);
-        }
-      }).catch(err => {
-        clearTimeout(asyncTimeout);
-        reject(err);
-      });
-    });
+    const response = await this.sendCommand('Runtime.evaluate', evaluationParams);
```
patrickhulce: We're really gonna need more than 5s here, we probably don't need 60s but axe is going to fail on *a lot* of URLs at 5s and on weaker hardware a lot of other audits will start failing too. Unless you guys have already discussed this and really want to put the pressure on all the evaluations to speed up ASAP and are embracing the failures in which case ignore me :)

lighthouse-core/gather/driver.js
```diff
@@ -142,6 +150,13 @@ class Driver {
     return this.evaluateAsync(`(${pageFunctions.ultradumbBenchmarkString})()`);
   }
 
+  /**
+   * @param {number} timeout
+   */
+  setProtocolTimeout(timeout) {
+    this._protocolTimeout = timeout;
```
patrickhulce: I afraid we'll need to make timeout a setting on `sendCommand` with a global default :/

proto/lighthouse-result.proto
```diff
@@ -42,6 +42,10 @@ enum LighthouseError {
   PARSING_PROBLEM = 15;
   // The trace data failed to stream over the protocol.
   READ_FAILED = 16;
+  // Used when security error prevents page load.
+  INSECURE_DOCUMENT_REQUEST = 17;
```
patrickhulce: is this spillover from the other PR? :)

Hoten: Well, it was missed in #6300. nbd since LR's LH bundle hasn't been updated since.

lighthouse-core/gather/driver.js
Hoten: We can move to that if it's really needed. It complicates calling it quite a bit. And due to the unique typing on the method, I think the best option would be this:

```
sendCommand(methodOrOpts, ...params) {
  const {method, timeout} = typeof methodOrOpts === 'object' ? methodOrOpts : {
    method,
    timeout: DEFAULT_PROTOCOL_TIMEOUT
  }
}
```

That way, the calling code for most use cases can remain the same.

proto/lighthouse-result.proto
```diff
@@ -42,6 +42,10 @@ enum LighthouseError {
...
+  // Used when security error prevents page load.
+  PROTOCOL_TIMEOUT = 18;
```
Hoten: I removed REQUEST_CONTENT_TIMEOUT ... instead of making a new enum value for `PROTOCOL_TIMEOUT`, could we just reuse `REQUEST_CONTENT_TIMEOUT`'s enum (rename it)? Would that be OK for proto compat?

lighthouse-core/gather/driver.js
```diff
@@ -255,8 +270,20 @@ class Driver {
...
+        reject(err);
+      }), this._protocolTimeout);
+      this._connection.sendCommand(method, ...params).then(result => {
+        clearTimeout(asyncTimeout);
+        resolve(result);
+      }).catch(err => {
+        clearTimeout(asyncTimeout);
+        reject(err);
+      });
+    });
```
brendankenny: maybe drop keep the promise just to the thing that has to be a promise? e.g. 
```js
  const asyncTimeout = setTimeout((_ => {
    const err = new LHError(LHError.errors.PROTOCOL_TIMEOUT);
    err.message += ` Method: ${method}`;
    reject(err);
  }), this._protocolTimeout);

  try {
    const result = await this._connection.sendCommand(method, ...params);
    clearTimeout(asyncTimeout);
    resolve(result);
  } catch (err) {
    clearTimeout(asyncTimeout);
    reject(err);
  }
});
```

brendankenny:
ok, options:
1) `sendCommand(methodOrOpts, ...params)`
2) call `setNextProtocolTimeout(60000)` right before `sendCommand()` and have it reset itself after that next `sendCommand` call
3) have `sendCommand()` have a 5s default, but have a second `unboundedSendCommand()` that can be called if needed (e.g. by `evaluateAsync`)
4) some kind of sneaky currying that we haven't figured out yet

other options? (1) worries me about exposing complexity for the 5% case that may be with us forever (see: every js API that takes as its first parameter a string, or if it's a number it acts like this, or if it's an object it's an options object, or if it's a stream it...). OTOH it is overhead when writing code, but is fairly self explanatory when reading it, and I take @Hoten's point that it's basically equivalent to (3)

lighthouse-core/gather/driver.js
Hoten: Put it back to 60s.

lighthouse-core/gather/driver.js
Hoten: ok, I put it back to 1s

Hoten:
@patrickhulce I'll test it out on some slower devices today.

Hoten:
Tested on an old Nexus 4 device. Didn't come across any protocol timeouts.

Hoten:
Got some timings. https://gist.github.com/Hoten/de38179b2d941b6d195c3c4383d0bd8d

`Runtime.evaluate` took ~9s on Nexus 4 (this command has an explicit 60s timeout). The next slowest was ~4.1s (brushing up on the 5s timeout). Everything that took longer than 1s seems related to font stuff (`CSS.getComputedStyleForNode`), which I think we are getting rid of soon? If not, we might consider 1) increasing the timeout for those commands or 2) instead of sending them all at once, send them in batches.

paulirish:
> Everything that took longer than 1s seems related to font stuff (CSS.getComputedStyleForNode)

Yup looks like it's the font-size gatherer:
https://github.com/GoogleChrome/lighthouse/blob/879141d18c84a1fb3712affedf14c0bf8efb18f0/lighthouse-core/gather/gatherers/seo/font-size.js#L216-L220

I was hoping it was the `fonts` one that was removed this morning. but alas.

TBH I think that situation is OK. I don't really want the font-size gatherer taking >5s.

lighthouse-core/gather/driver.js
```diff
@@ -860,27 +878,17 @@ class Driver {
    * Return the body of the response with the given ID. Rejects if getting the
    * body times out.
    * @param {string} requestId
-   * @param {number} [timeout]
+   * @param {number} timeout
```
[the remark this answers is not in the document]
Hoten: yeah. how do you mark optional? Are the brackets "clojure style"?

proto/lighthouse-result.proto
paulirish: nope, can't rename. 

But actually looks like we never exposed REQUEST_CONTENT_TIMEOUT in the proto. So I think we're OK. :)

lighthouse-core/gather/driver.js
```diff
@@ -240,23 +248,48 @@ class Driver {
     }
   }
 
+  /**
+   * NOTE: This can eventually be replaced when TypeScript
```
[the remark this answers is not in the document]
Hoten: Oh, this is referring to `sendCommand`. That might be the confusion...

My understanding is that the linked issue will eventually allow types such as:

```
type TypedArray = [string, number];
type ExtendedTypedArray = TypedArray + [string];
const value: ExtendedTypedArray = ['', 12, ''];
```

If so, we'd be able to type that variadic parameter as `LH.CrdpCommands[C]['paramsType'] + [{timeout: number}]`

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
patrickhulce: `{number=}` is closure style :) 

brackets are [jsdoc](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values)

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
patrickhulce: Oooooh, that makes much more sense. Move the comments down there then near that parameter? :)

paulirish:
@Hoten 	proto conflict for ya

Hoten:
@paulirish resolved

brendankenny:
> Yup looks like it's the font-size gatherer

That gatherer sends all those `CSS.getComputedStyleForNode` commands at once, up to 500, depending on the page's node count. It's possible command overhead/contention causes issues, especially on an older phone. We could try doing them in batches instead of all at once to see if that's a solution.

It's also remotely possible we hit some gross style-recalc issue, depending on the page and how the commands are coming in. It's possible it would be more efficient to do all the lookups in the page at once with `evaluateAsync`, since we could guarantee running them all in the same turn of the event loop/without reforcing layout. We'd have to resync node info and debugger `nodeId`s, though. Maybe there's a similar coalesced solution, though.

PR #6351, opened 2018-10-21
report(third party): filter out third party urls

wardpeet:
**Summary**

Hides 3th party urls  from our tables when the checkbox is marked. This do not changes scores, just hides it in the UI.

A request is marked as third party when the TLD + 1 can not be found in the url.
`pageTLDPlusOne = new URL(this.json.finalUrl).origin.split('.').slice(-2).join('.');`

Demo:
http://wardpeet-filestorage.surge.sh/lh-store/pr-6351/report.html

**Related Issues/PRs**

#4516 

need upgraded jsdom:
#6411


wardpeet:
I still need to add a global checkbox to select/unselect all checkboxes. We might also want to show a label on marked elements like @hwikyounglee suggested.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +129,54 @@ class ReportUIFeatures {
...
+        // create input box
+        const filterTemplate = this._dom.cloneTemplate('#tmpl-lh-3p-filter', this._document);
+        const filterInput = filterTemplate.querySelector('input');
+        const id = `lh-3p-filter-label--${index}`;
+        filterTemplate.querySelector('label').setAttribute('for', id);
+        filterInput.setAttribute('id', id);
+
+        filterInput.addEventListener('change', e => {
+          // remove elements from the dom and keep track of them to readd on uncheck
+          // why removing instead of hiding? nth-child(even) background-colors keep working
```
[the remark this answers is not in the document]
wardpeet: would this suffice?
image

hwikyounglee:
3p-show

Bring all things out from my sleeve, hmm, not much hidden! But made up a few things now. Open for suggestions. 

1. Probably okay without background color
2. Probably we could see how the locally scoped checkboxes are used, and reassess a need of a global checkbox.
3. Just to make the checkboxes look _noticeable_, I'm wondering if we might try the checkbox 'checked' and the label it as 'Show' instead of 'Filter-out/hide'.

wardpeet:
2) I think it would be nice to have local scoped checkboxes and a global one? Or maybe the global one is over rated.
3) We could check it by default and rephrase into Show 3th party resources. I don't really have an opinion on this.

lighthouse-core/report/html/templates.html
```diff
@@ -666,8 +666,8 @@
   </style>
   <div class="lh-3p-filter">
     <label class="lh-3p-filter-label">
-      Filter out 3rd-parties:
-      <input type="checkbox" class="lh-3p-filter-input">
+      <input type="checkbox" class="lh-3p-filter-input" checked />
+      Show 3th party resources (<span class="lh-3p-filter-count"></span>)
```
Hoten: 3rd

wardpeet: thanks!

wardpeet:
I'm having an issue with the tests because the initFeatures is ran on the report template which does not contain the <template> strings. I could combine the files with a string replace but other errors occur so not sure what to do here? I'm thinking of moving the report-ui-feature-test.js to end to end tests with puppeteer WDYT?

brendankenny:
there's definitely value in having it be a unit test.

> the report template which does not contain the strings

Which strings do you mean? The travis failure doesn't give much of a clue

wardpeet:
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/html/reporttemplate.html is used in the report features test which means there are no <template> tags to clone.
Templates are stored in https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/html/templates.html. It would make sense to replace the `%%LIGHTHOUSE_TEMPLATES%%` with the templates but somehow it didn't work. I could figure out why that doesn't work

brendankenny:
> It would make sense to replace the %%LIGHTHOUSE_TEMPLATES%% with the templates but somehow it didn't work.

ah, I see. I thought maybe you were talking about the i18n strings and was worried :)

build-viewer.js literally does just `htmlSrc = htmlSrc.replace(/%%LIGHTHOUSE_TEMPLATES%%/, htmlReportAssets.REPORT_TEMPLATES);`, so it should work in theory...

wardpeet:
@brendankenny fixed the issue by changing `const container = reportUIFeatures._dom._document.body` into `const container = reportUIFeatures._dom._document.querySelector('main');` as it's getting cleared by the renderReport function.

Please let me know what is left to move forward.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +129,75 @@ class ReportUIFeatures {
...
+        const filterTemplate = this._dom.cloneTemplate('#tmpl-lh-3p-filter', this._document);
+        const filterInput = /** @type {HTMLInputElement} */ (filterTemplate.querySelector('input'));
+        const id = `lh-3p-filter-label--${index}`;
+
+        if (filterInput) {
+          filterInput.setAttribute('id', id);
+          filterInput.addEventListener('change', e => {
+            // remove elements from the dom and keep track of them to readd on uncheck
+            // why removing instead of hiding? nth-child(even) background-colors keep working
+            if (e.target instanceof HTMLInputElement && e.target.checked) {
```
wardpeet: @brendankenny now I get an eslint issue. What's the proper way to fix this?

wardpeet: @brendankenny thanks! it works

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+    // get all tables with a text url
+    /** @type {Array<HTMLTableElement>} */
+    const tables = Array.from(this._document.querySelectorAll('.lh-table'));
+    tables.filter(el => el.querySelector('td.lh-table-column--url'))
+      .filter(el => /** @type {Element} */ (el.closest('.lh-audit')).id !== 'uses-rel-preconnect')
+      .forEach((el, index) => {
+        /** @type {NodeListOf<HTMLElement>} */
+        const urlItems = el.querySelectorAll('.lh-text__url');
+        /** @type {Record<string, HTMLTableRowElement>} */
+        const thirdPartyRows = {};
```
paulirish: we can pull the first 10ish lines of this out into a `const thirdPRowElems = this._getThirdPartyRows(el, finalUrl);`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+
+        for (const urlItem of urlItems) {
+          const isThirdParty = !urlItem.title.includes(`${pageTLDPlusOne}/`);
+          if (!isThirdParty) {
+            continue;
+          }
+
+          const urlRow = urlItem.closest('tr');
```
paulirish: I'd prefer any var referring to an element has a 'El' or 'Elem' suffix.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+          if (urlRow) {
+            const rowPosition = Array.from(el.tBodies[0].children).indexOf(urlRow);
```
paulirish: i think we could make a helper function since you use this a few times. 

```
const getRowIndex = (rowElem, el) => Array.from(el.tBodies[0].children).indexOf(rowElem);
```

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+            thirdPartyRows[`${rowPosition}`] = urlRow;
+          }
+        }
+
+        // create input box
+        const filterTemplate = this._dom.cloneTemplate('#tmpl-lh-3p-filter', this._document);
+        const filterInput = /** @type {HTMLInputElement} */ (filterTemplate.querySelector('input'));
+        const id = `lh-3p-filter-label--${index}`;
+
+        if (filterInput) {
```
paulirish: i dont think this check is necessary any longer.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+          filterInput.setAttribute('id', id);
+          filterInput.addEventListener('change', e => {
+            // remove elements from the dom and keep track of them to readd on uncheck
+            // why removing instead of hiding? nth-child(even) background-colors keep working
+            if (e.target instanceof HTMLInputElement && e.target.checked) {
+              Object.keys(thirdPartyRows).forEach(position => {
```
paulirish: `Object.entries(thirdPartyRows).forEach(([position, row]) => {` should help you get rid of the next line.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+                const row = thirdPartyRows[position];
+
+                if (position === '0') {
+                  el.tBodies[0].appendChild(row);
+                } else {
+                  Array.from(el.tBodies[0].children)[Number(position) - 1]
+                    .insertAdjacentElement('afterend', row);
```
paulirish: feels like if you use insertBefore you can avoid switching between `append` vs `insertAdjacentElement`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+                }
+              });
+            } else {
+              Object.keys(thirdPartyRows).forEach(position => {
+                const row = thirdPartyRows[position];
+
+                if (row.parentNode) {
+                  row.parentNode.removeChild(row);
```
paulirish: row.remove()

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+                }
+              });
+            }
+          });
+        }
+
+        if (filterTemplate && el.parentNode) {
+          /** @type {Element} */ (filterTemplate.querySelector('label')).setAttribute('for', id);
+          /** @type {Element} */ (filterTemplate.querySelector('.lh-3p-filter-count')).innerHTML =
```
paulirish: danger danger! we can easily use textContent instead.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
 
+  _setupThirdPartyFilter() {
+    const pageTLDPlusOne = new URL(this.json.finalUrl).origin.split('.').slice(-2).join('.');
+
+    // get all tables with a text url
+    /** @type {Array<HTMLTableElement>} */
+    const tables = Array.from(this._document.querySelectorAll('.lh-table'));
+    tables.filter(el => el.querySelector('td.lh-table-column--url'))
+      .filter(el => /** @type {Element} */ (el.closest('.lh-audit')).id !== 'uses-rel-preconnect')
+      .forEach((el, index) => {
```
paulirish: lets call this 'tableEl` since we're dealing with a few different elements in here.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
...
+              });
+            }
+          });
+        }
+
+        if (filterTemplate && el.parentNode) {
+          /** @type {Element} */ (filterTemplate.querySelector('label')).setAttribute('for', id);
+          /** @type {Element} */ (filterTemplate.querySelector('.lh-3p-filter-count')).innerHTML =
+            `${Object.keys(thirdPartyRows).length}`;
+          el.parentNode.insertBefore(filterTemplate, el);
```
paulirish: // Finally, add checkbox to the DOM

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -128,6 +131,75 @@ class ReportUIFeatures {
     dropdown.addEventListener('click', this.onExport);
   }
 
+  _setupThirdPartyFilter() {
+    const pageTLDPlusOne = new URL(this.json.finalUrl).origin.split('.').slice(-2).join('.');
```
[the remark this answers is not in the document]
paulirish: also we have a method in canonical.js that does this, as well. we could move to url-shim.. add some tests...

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: holy! never knew this was possible

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: it's not necessary but always do it for dom elements but i'll remove it

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: i'm only using this once but I do use Array.from a lot 😛

wardpeet:
@paulirish Done & updated example

lighthouse-core/test/lib/url-shim-test.js
```diff
@@ -114,6 +114,17 @@ describe('URL Shim', () => {
     });
   });
 
+  describe('getRootDomain', () => {
+    it('returns the correct rootDomain', () => {
+      assert.equal(URL.getRootDomain('example.com'), 'example.com');
```
paulirish: the input is expected to be a full url, right? with `http://`, etc.. can we use those?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +133,76 @@ class ReportUIFeatures {
     dropdown.addEventListener('click', this.onExport);
   }
 
+  _setupThirdPartyFilter() {
+    // get all tables with a text url
+    /** @type {Array<HTMLTableElement>} */
+    const tables = Array.from(this._document.querySelectorAll('.lh-table'));
+    const thirdPartyFilterAuditExclusions = tables
```
paulirish: heh. i bet this didn't make sense... what i meant to say was.. lets try an array for these ids.. i was thinking like this:

```js
const thirdPartyFilterAuditExclusions = ['uses-rel-preconnect'];
const tablesWithUrls = tables      
      .filter(el => el.querySelector('td.lh-table-column--url'))
      .filter(el => !thirdPartyFilterAuditExclusions.includes( /** @type {Element} */ (el.closest('.lh-audit')).id));
```

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +133,76 @@ class ReportUIFeatures {
...
+      const id = `lh-3p-filter-label--${index}`;
+
+      filterInput.setAttribute('id', id);
+      filterInput.addEventListener('change', e => {
+        // remove elements from the dom and keep track of them to readd on uncheck
+        // why removing instead of hiding? nth-child(even) background-colors keep working
+        if (e.target instanceof HTMLInputElement && e.target.checked) {
+          Object.entries(thirdPartyRows).forEach(([position, row]) => {
+            const childrenArr = Array.from(tableEl.tBodies[0].children);
+            tableEl.tBodies[0].insertBefore(row, childrenArr[Number(position)]);
```
paulirish: this Number cast is neccessary?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +133,76 @@ class ReportUIFeatures {
...
+      const filterInput = /** @type {HTMLInputElement} */ (filterTemplate.querySelector('input'));
+      const id = `lh-3p-filter-label--${index}`;
+
+      filterInput.setAttribute('id', id);
+      filterInput.addEventListener('change', e => {
+        // remove elements from the dom and keep track of them to readd on uncheck
+        // why removing instead of hiding? nth-child(even) background-colors keep working
+        if (e.target instanceof HTMLInputElement && e.target.checked) {
+          Object.entries(thirdPartyRows).forEach(([position, row]) => {
+            const childrenArr = Array.from(tableEl.tBodies[0].children);
```
paulirish: i still count this `Array.from(tableEl.tBodies[0].children);` twice. giving it its own fn just makes it more readable IMO.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +133,76 @@ class ReportUIFeatures {
...
+  }
+
+  /**
+   * @param {HTMLTableElement} el
+   * @param {string} finalUrl
+   */
+  _getThirdPartyRows(el, finalUrl) {
+    /** @type {NodeListOf<HTMLElement>} */
+    const urlItems = el.querySelectorAll('.lh-text__url');
+    const pageTLDPlusOne = new URL(finalUrl).origin.split('.').slice(-2).join('.');
```
paulirish: you were planning to use getRootDomain here?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +133,76 @@ class ReportUIFeatures {
...
+    /** @type {Record<string, HTMLTableRowElement>} */
+    const thirdPartyRows = {};
```
paulirish: feels like a good use of Map or WeakMap instead... just to communicate it's holding elements rather than data.

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: it's a fix for typescript, else i'll need to do something like
```
Object.entries(thirdPartyRows).forEach(([position, row]) => {
            const tmpPosition = /** @type {unknown} */ (position);
            const childrenArr = Array.from(tableEl.tBodies[0].children);
            tableEl.tBodies[0].insertBefore(row, childrenArr[/** @type{number} */ (tmpPosition)]);
```

lighthouse-core/test/lib/url-shim-test.js
wardpeet: hostname is used, if I didn't than I had to create an url object again. I could be smart about it and accept a string or URL object. (this could be a function that we use everywhere in the class so we only have to createn URL once.

```
let hostname;
if (url instanceof URL) {
  hostname = url.hostname;
} else {
  const urlObj = new URL(url);
  hostname = urlObj.hostname;
}
```

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: I added a comment why I can't. UrlShim cannot be imported inside the report.

lighthouse-core/lib/url-shim.js
```diff
@@ -114,6 +114,22 @@ class URLShim extends URL {
...
+   * Returns a primary domain for provided hostname (e.g. www.example.com -> example.com).
+   * @param {string} hostname
+   * @returns {string}
+   */
+  static getRootDomain(hostname) {
+    const tld = URLShim.getTld(hostname);
+
+    // tld is .com or .co.uk which means we means that length is 1 to big
+    // .com => 2 & .co.uk => 3
+    const lengthOfTLD = tld.split('.').length;
```
paulirish:
```suggestion
    const splitTld = tld.split('.');
```

lighthouse-core/lib/url-shim.js
```diff
@@ -114,6 +114,22 @@ class URLShim extends URL {
...
+
+    // get TLD + root domain
+    return hostname.split('.').slice(-lengthOfTLD).join('.');
```
paulirish:
```suggestion
    return hostname.split('.').slice(-splitTld.length).join('.');
```

lighthouse-core/lib/url-shim.js
paulirish: small nit but seemed easier than clarifying this is the lengthOfSplitTLD :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +140,82 @@ class ReportUIFeatures {
...
+    /** @type {Array<HTMLTableElement>} */
+    const tables = Array.from(this._document.querySelectorAll('.lh-table'));
+    const tablesWithUrls = tables
+      .filter(el => el.querySelector('td.lh-table-column--url'))
+      .filter(el =>!thirdPartyFilterAuditExclusions.includes(
+        /** @type {Element} */(el.closest('.lh-audit')).id));
+
+    tablesWithUrls.forEach((tableEl, index) => {
+      const thirdPartyRows = this._getThirdPartyRows(tableEl, this.json.finalUrl);
+
```
paulirish:
```suggestion
  if (!thirdPartyRows.length) return; // no 3rd parties, no checkbox!
```

lighthouse-core/report/html/templates.html
```diff
@@ -891,3 +891,24 @@
...
+    }
+    .lh-3p-filter-label, .lh-3p-filter-input {
+      cursor: pointer;
+      vertical-align: middle;
+    }
+  </style>
+  <div class="lh-3p-filter">
+    <label class="lh-3p-filter-label">
+      <input type="checkbox" class="lh-3p-filter-input" checked />
+      Show 3rd party resources (<span class="lh-3p-filter-count"></span>)
```
paulirish: right now this is hardcoded english. Can you replace the `Show 3rd party resources` here with something you add to `Util.UIStrings`?  

You can leave the `(2)` thing mostly as is. We don't support this completely correctly but i'm going to try to fix that now... :)

lighthouse-core/test/lib/url-shim-test.js
paulirish: sg.

lighthouse-core/report/html/renderer/report-ui-features.js
paulirish: idea... move `getRootDomain`, `getTld` and `listOfTlds` into Util. Add lillte static methods on URL-shim that use the Util impl's... Here in the renderer we can use them from Util.

i'm suggesting this because otherwise we need to match the impl completely. right now they're different. (this implementation doesn't consider public suffixes so sockshop.co.uk and akamaicdn.co.uk are considered the same :)

lighthouse-core/report/html/templates.html
patrickhulce: bump on this :) I think we decided we still want to UIString this even if we don't support plurals

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -25,6 +27,13 @@
 
 /** @typedef {import('./dom.js')} DOM */
 
+/**
+ * @param {HTMLTableElement} tableEl
+ * @return {Array<HTMLTableRowElement>}
+ */
+const getTableRows = tableEl => /** @type {Array<HTMLTableRowElement>} */
+  (Array.from(tableEl.tBodies[0].children));
```
patrickhulce: well TIL `tBodies` 😮

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +140,82 @@ class ReportUIFeatures {
...
 
+  _setupThirdPartyFilter() {
+    const thirdPartyFilterAuditExclusions = ['uses-rel-preconnect'];
+
+    // get all tables with a text url
+    /** @type {Array<HTMLTableElement>} */
+    const tables = Array.from(this._document.querySelectorAll('.lh-table'));
+    const tablesWithUrls = tables
+      .filter(el => el.querySelector('td.lh-table-column--url'))
+      .filter(el =>!thirdPartyFilterAuditExclusions.includes(
```
patrickhulce: nit: space in front of `!`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +140,82 @@ class ReportUIFeatures {
...
+   * @return {Map<number, HTMLTableRowElement>}
+   */
+  _getThirdPartyRows(el, finalUrl) {
+    /** @type {NodeListOf<HTMLElement>} */
+    const urlItems = el.querySelectorAll('.lh-text__url');
+    const rootDomain = Util.getRootDomain(finalUrl);
+    /** @type {Map<number, HTMLTableRowElement>} */
+    const thirdPartyRows = new Map();
+    for (const urlItem of urlItems) {
+      const isThirdParty = !urlItem.title.includes(`${rootDomain}/`);
```
patrickhulce: heh, this feels a bit like a hack :) should we make a note to have a proper data attribute or something with the URL in case we want the title to be something else someday?

lighthouse-core/report/html/templates.html
paulirish: Ah yes. Thanks. :)

lighthouse-core/report/html/renderer/report-ui-features.js
wardpeet: I actually added it to the dataset right away.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -130,6 +140,84 @@ class ReportUIFeatures {
...
+   * @return {Map<number, HTMLTableRowElement>}
+   */
+  _getThirdPartyRows(el, finalUrl) {
+    /** @type {NodeListOf<HTMLElement>} */
+    const urlItems = el.querySelectorAll('.lh-text__url');
+    const rootDomain = Util.getRootDomain(finalUrl);
+    /** @type {Map<number, HTMLTableRowElement>} */
+    const thirdPartyRows = new Map();
+    for (const urlItem of urlItems) {
+      const isThirdParty = !(urlItem.dataset.url || '').includes(`${rootDomain}/`);
```
patrickhulce: ❤️

wardpeet:
Finally everything is green

lighthouse-core/lib/i18n/en-US.json
```diff
@@ -1319,6 +1319,10 @@
     "message": "Expand snippet",
     "description": "Label for button that shows all lines of the snippet when clicked"
   },
+  "lighthouse-core/report/html/renderer/util.js | thirdPartyResourcesLabel": {
+    "message": "Show 3rd party resources",
+    "description": "This label is shown above a table next to the checkbox to hide 3rd party resources"
```
wardpeet: this probably needs to improve 😄

lighthouse-core/lib/url-shim.js
```diff
@@ -11,20 +11,14 @@
 
 /* global self */
 
+/** @typedef {import('url').URL} OriginalURL */
```
wardpeet: @brendankenny wasn't sure how to fix this but `new Window["URL"]` isn't compatible with itself.

Hoten:
gotta merge master back in, to account for the (partial) new ui

wardpeet:
@Hoten I merged and updated the example at the top and see that now also got deployed.

lighthouse-core/report/html/templates.html
```diff
@@ -857,6 +856,26 @@
   </div>
 </template>
 
+<template id="tmpl-lh-3p-filter">
+  <style>
+    .lh-3p-filter {
+      background: var(--lh-table-higlight-bg);
+      color: var(--medium-75-gray);
```
Hoten: `--item-bg` instead to match https://yuinchien.github.io/sandbox/lighthouse/

name isn't great but will fix that later.

lighthouse-core/lib/i18n/en-US.json
patrickhulce: heh :) 

Maybe...

`Label for checkbox that toggles the visibility of "3rd party resources" in the adjacent table. "3rd party resources" are any items in the table that came from another domain that likely isn't owned or controlled by the developer of the webpage.`

lighthouse-core/report/html/renderer/util.js
```diff
@@ -336,6 +350,51 @@ class Util {
...
+  static getRootDomain(url) {
+    const hostname = Util.createOrReturnURL(url).hostname;
+    const tld = Util.getTld(hostname);
+
+    // tld is .com or .co.uk which means we means that length is 1 to big
+    // .com => 2 & .co.uk => 3
+    const splitTld = tld.split('.');
+
+    // get TLD + root domain
+    return hostname.split('.').slice(-splitTld.length).join('.');
```
patrickhulce: I think this new logic breaks on `localhost`, could we add a quick test and fix?

wardpeet: it's actually working :)

```
// hostname = "localhost"
const hostname = Util.createOrReturnURL(url).hostname;
// ".localhost"
const tld = Util.getTld(hostname);

// tld is .com or .co.uk which means we means that length is 1 to big
// .com => 2 & .co.uk => 3
// ['', 'localhost']
const splitTld = tld.split('.');

// get TLD + root domain
// "localhost"
return hostname.split('.').slice(-splitTld.length).join('.');
```

The reason why it works is .slice(-2) or whatever actually just returns array[0] if only 1 element is available.

wardpeet: :o might be the port, let me check that out. Good eye, I only checked real websites.

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -189,6 +190,7 @@ describe('util helpers', () => {
       assert.equal(Util.getRootDomain(new URL('https://sub.example.com')), 'example.com');
       assert.equal(Util.getRootDomain(new URL('https://sub.example.tokyo.jp')), 'tokyo.jp');
       assert.equal(Util.getRootDomain(new URL('http://localhost')), 'localhost');
+      assert.equal(Util.getRootDomain(new URL('http://localhost:8080')), 'localhost');
```
patrickhulce: ahhhhh OK I gotcha now, nice sleuthing! :)

wardpeet: "sleuthing" new word to my vocabulary 😄

lighthouse-core/report/html/renderer/util.js
Hoten: @patrickhulce this is fixed now

Hoten:
@patrickhulce @paulirish tests are passing. any more change requests?

brendankenny:
i'm taking a pass cleaning up URL since we're on Node 10+ now

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -102,6 +112,87 @@ class ReportUIFeatures {
     dropdown.addEventListener('click', this.onExport);
   }
 
+  _setupThirdPartyFilter() {
+    const thirdPartyFilterAuditExclusions = ['uses-rel-preconnect'];
```
brendankenny: need a comment for what this is

brendankenny:
check out 75c6ef5 for the changes I made. Mostly just relying on global `URL` now and eliminating some of the duplication between `Util` and `URL`

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten:
```
List of audits that shouldn't display the 3p filter feature
``` 

That? That's what I'm getting just from the name - which makes me think a comment isn't needed. @wardpeet  is it more nuanced than that?

Hoten: how bout `Some audits should not display the third party filter option.`

patrickhulce: I imagine @brendankenny wanted the *why* in a comment. i.e. this audit is explicitly recommending to you, only third party origins that you should be connecting to so filtering out third parties doesn't make sense

Hoten: @brendankenny all good?

brendankenny:
> No rendering tests, though!

:)

Hoten:
image
highlighted is considered 3P, lol. will add this audit to the ignore list.

Hoten:
> > No rendering tests, though!
> 
> :)
> 
> (sorry)

done. caught a bug :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -191,13 +191,9 @@ class ReportUIFeatures {
...
     for (const urlItem of urlItems) {
-      const datasetUrl = (urlItem.dataset.url || '');
-      // the : check is used for domains that specify a port (google.com:8080/index.html)
-      const isThirdParty = !datasetUrl.includes(`${rootDomain}/`) &&
-        !datasetUrl.includes(`${rootDomain}:`);
-      if (!isThirdParty) {
-        continue;
-      }
+      const datasetUrl = urlItem.dataset.url || '';
+      const isThirdParty = Util.getRootDomain(datasetUrl) !== rootDomain;
```
Hoten: @wardpeet this ok?

brendankenny:
🎊 🎊 🎊

paulirish:
Woo hoooooooooooo!!!!!!!

wardpeet:
@Hoten thanks for getting this out the door! 🙏

Issue #6336, opened 2018-10-18
runTimeError exists when there are NO_ERRORs


ebidel:
Ran into this working on a web server that returns LH results. It's unexpected that `runTimeErrors` exists in the LHR if there were no errors! It also requires users to check for a "NO_ERROR" string code which overcomplicates matters.

```
    if (lhr.runtimeError && lhr.runtimeError.code !== 'NO_ERROR') {
      throw new Error(`${lhr.runtimeError.code} ${lhr.runtimeError.message}`);
    }
```

<img width="480" alt="screen shot 2018-10-17 at 10 55 55 pm" src="https://user-images.githubusercontent.com/238208/47134128-78ca0900-d260-11e8-8f59-2dd02bff9bae.png">

#### Environment Information
* Affected Channels: all
* Lighthouse version: 3.2.1
* Node.js version: 10+

ebidel:
+1 just to echo that, it was unclear in our tool that consumes lhrs that
the results we were saving were actually illegitimate. And you almost never
want a psuedo real result... doesn't tell you much :)

exterkamp:
This has implications for proto enums as well.  @paulirish

paulirish:
#### runtimeError for failed document requests

@ebidel currently in the case of a 500 or a DNS resolving error for the page being tested, we do return an LHR with a populated `runtimeError` object. So you're going to have to check this prop before considering the rest of the LHR valid enough to read. _(But.. more on NO_ERROR below)_


#### On throwing and pseudo-LHRs

I'm not sold that throwing to our node module consumers is useful, but we can leave that be. I'll live with our current approach.

However, for CLI users (and I'm mostly considering CLI use _at scale_ here), I don't think communicating over exit code is tremendously effective. In order to convey any detail about the error we'd need to log something, but that now means the user has to parse stderr, which we've been logging status to this whole time. :/

So I'd argue that CLI users would be better served with a psuedo-LHR with the `runtimeError` object populated. 

#### NO_ERROR

Sounds like nobody likes `lhr.runtimeError` being set if there was no error. I can understand that. While it's more idomatic in JS land to leave it off, in the proto, we do need the field defined and initialized with NO_ERROR. However, we think we can populate it after going JSON->Proto. So that seems like the rough plan here. 

#### API flexibility

[...]

PR #6358, opened 2018-10-22
misc(proto): make NO_ERROR default in proto

exterkamp:
**Summary**
Make NO_ERROR the default enum in proto definition and add runtimeError removal on default from the preprocessor.

**Related Issues/PRs**
addresses, but does not fix totally: #6336


lighthouse-core/test/lib/proto-preprocessor-test.js
```diff
@@ -27,6 +27,30 @@ describe('processing for proto', () => {
...
+        'code': 'NO_ERROR',
+      },
+    };
+
+    const output = processForProto(JSON.stringify(input));
+
+    expect({}).toMatchObject(JSON.parse(output));
+  });
+
+  it('ignores non-default runtimeErrors', () => {
```
patrickhulce: nit: `ignores` was a bit ambiguous to me as it could've meant "drops" WDYT about `passes through`, `leaves untouched`, `preserves`?

lighthouse-core/test/lib/proto-preprocessor-test.js
```diff
@@ -27,6 +27,30 @@ describe('processing for proto', () => {
...
+  it('cleans up default runtimeErrors', () => {
+    const input = {
+      'runtimeError': {
+        'code': 'NO_ERROR',
+      },
+    };
+
+    const output = processForProto(JSON.stringify(input));
+
+    expect({}).toMatchObject(JSON.parse(output));
```
patrickhulce: I think you meant for the output to be inside `expect()` :)

proto/lighthouse-result.proto
```diff
@@ -6,42 +6,41 @@ import "google/protobuf/wrappers.proto";
 
 // Canonical list of errors created by Lighthouse.
 enum LighthouseError {
-  UNDEFINED = 0;
   // No error in Lighthouse; the results are reliable.
-  NO_ERROR = 1;
+  NO_ERROR = 0;
```
patrickhulce: This is kindof a bummer, we can never undo this it seems, so we're 100% happy with it right?

patrickhulce: or rather, do they all need to be sequential? It'd be nice if is there a way for 0 and 1 to both be no error somehow or at least not impact every single other type

exterkamp: They don't have to be sequential at all, we can reserve as much space as we want for anything.  But why would we need to?  I don't follow.

We can also set up `alias`ing for enums.  Like `NO_ERROR = 0;` and `UNDEFINED=0;` and share fieldnumbers for multiple inputs.

exterkamp: https://developers.google.com/protocol-buffers/docs/proto3#assigning-field-numbers

lighthouse-core/test/lib/proto-preprocessor-test.js
patrickhulce: Oh I see what it's doing. I think `toMatchObject` is just not right for the assertion we're trying to make here then.

It's basically [not.toHaveProperty('runtimeError')](https://jestjs.io/docs/en/expect#tohavepropertykeypath-value) we're checking for right?

exterkamp: Good call!  That's what I was looking for!

proto/lighthouse-result.proto
patrickhulce: I just mean that all consumers of the proto would find their values shifted after this change right? We don't have any consumers yet, and I'm sure you've already thought this through :) but I wanted to know what the plan would be if we needed to make another change to the enums like this, all the values wouldn't change right we'd use one of those handy aliases or something?

exterkamp: Yes.  If we had any consumers they'd all be off by one.  Once we start to use this proto we can't change it.  `NO_ERROR` will always be the default, but we can alias more defaults and add new ones and mark numbers as resolved.

PR #6361, opened 2018-10-22
 tests: run node 10 on Travis

brendankenny:
Node 10 is rapidly approaching LTS. We should be running our tests on Node 8 and 10, not 8 and 9 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,8 @@ script:
   - yarn test-viewer
   - yarn test-lantern
   - yarn i18n:checks
+  # _JAVA_OPTIONS is breaking parsing of `compile-devtools` output. See #3338.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -42,7 +42,7 @@ script:
   - yarn test-viewer
   - yarn test-lantern
   - yarn i18n:checks
-  # _JAVA_OPTIONS is breaking parsing of compiler output. See #3338.
+  # _JAVA_OPTIONS is breaking parsing of `compile-devtools` output. See #3338.
```
brendankenny: changed this because I assumed this was old leftover closure stuff and originally deleted it, but #3338 is about a failure on `yarn compile-devtools` (which may eventually run again depending on the FIXME below)

PR #6366, opened 2018-10-23
core: refactor tracing processor. don't load blank page twice in first pass

Hoten:
I started working towards #3707, but a complication arose that made the whole thing a bit too much work to bother with at the moment. Fortunately, couple good things came about, that can be push independently. I removed the stuff specific to #3707, but those changes can be found in this branch: `undo-revert-datauri`. See that issue for more details.

These changes:

1. Refactor how we get the pid/tid/frameId of the current target. The tracing processor was returning a specific event, and even synthesizing it. The callers of this function only grabbed the pid/tid/frameId, so instead of returning an event object I return just the pertinent data.

2. Skip loading the blank page a second time on the first pass. It's already been loaded.


lighthouse-core/gather/gather-runner.js
```diff
@@ -26,20 +26,21 @@ const Driver = require('../gather/driver.js'); // eslint-disable-line no-unused-
...
+ *     v. enableRuntimeEvents
+ *     vi. evaluateScriptOnLoad rescue native Promise from potential polyfill
+ *     vii. register a performance observer
+ *     viii. register dialog dismisser
+ *     iv. clearDataForOrigin
  *
  * 2. For each pass in the config:
  *   A. GatherRunner.beforePass()
- *     i. navigate to about:blank
+ *     i. navigate to a blank page
```
Hoten: these comment changes came from #3707. Although the data uri page isn't going with these changes, I figured the comments are still good. lmk if not.

lighthouse-core/test/gather/computed/main-thread-tasks-test.js
```diff
@@ -14,15 +14,18 @@ const TracingProcessor = require('../../../lib/traces/tracing-processor.js');
 const assert = require('assert');
 
 describe('MainResource computed artifact', () => {
-  const args = {data: {}};
+  const pid = 1;
+  const tid = 2;
+  const frameId = 'BLAH';
+  const args = {data: {}, frame: frameId};
```
Hoten: this test implicitly relied on many fields (tid, pid, ...) being the same (in this case, undefined), so that events would be collected (see `trace-of-tab.js` `mainThreadEvents` for one example). Also, since an error is thrown on a non-truthy id value, these had to be given some positive number.

typings/externs.d.ts
```diff
@@ -219,6 +220,7 @@ declare global {
           styleSheetUrl?: string;
           timerId?: string;
           url?: string;
+          processId?: number;
```
patrickhulce: why did this need to be added aren't we still accessing the same `processId` in `frames` array?

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -198,39 +198,55 @@ class TraceProcessor {
...
+    if (!pid || !tid || !frameId) throw new LHError(LHError.errors.NO_TRACING_STARTED);
 
-    // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
-    const frameId = /** @type {string} */ (startedInPageEvt.args.data.page);
-    return {startedInPageEvt, frameId};
+    return {
+      pid,
+      tid,
+      frameId,
+      // @ts-ignore
```
patrickhulce: any particular reason `ts` isn't in the `if (!` list on 241?

lighthouse-core/gather/gather-runner.js
```diff
@@ -409,6 +418,7 @@ class GatherRunner {
           passConfig,
           // *pass() functions and gatherers can push to this warnings array.
           LighthouseRunWarnings: baseArtifacts.LighthouseRunWarnings,
+          firstPass,
```
patrickhulce: now that we're passing this variable around to other places, WDYT about renaming to `isFirstPass` or something so we know it's a bool and not a ref to the first pass config or something?

lighthouse-core/gather/gather-runner.js
```diff
@@ -105,7 +106,9 @@ class GatherRunner {
   static async setupDriver(driver, options) {
     log.log('status', 'Initializing…');
     const resetStorage = !options.settings.disableStorageReset;
-    // Enable emulation based on settings
+    // In the devtools/extension case, we can't still be on the site while trying to clear state
+    // So we first navigate to a blank page, then apply our emulation & setup
+    await GatherRunner.loadBlank(driver);
```
patrickhulce: another consequence of moving this is that the benchmarkIndex will be computed while on the previous page which may have other junk interfering. If we can't preserve the order maybe add a note to revisit if it becomes a problem or file an issue to investigate?

lighthouse-core/gather/gather-runner.js
```diff
@@ -187,7 +190,14 @@ class GatherRunner {
...
     const blankPage = passContext.passConfig.blankPage;
     const blankDuration = passContext.passConfig.blankDuration;
-    await GatherRunner.loadBlank(passContext.driver, blankPage, blankDuration);
+
+    // On the very first pass we're already on blank
+    const skipLoadBlank = passContext.firstPass;
+    const pass = skipLoadBlank
+      ? Promise.resolve()
+      : GatherRunner.loadBlank(passContext.driver, blankPage, blankDuration);
+    await pass;
```
patrickhulce: WDYT about something like

```js
if (!passContext.isFirstPass) {
  // If we're not on the first pass, we're still on the target page.
  // Navigate to blank for the pass setup.
  await GatherRunner.loadBlank(passContext.driver, blankPage, blankDuration);
}
```

the ternary no-op await is just throwin' me for a loop :)

lighthouse-core/gather/computed/trace-of-tab.js
```diff
@@ -82,10 +82,10 @@ class TraceOfTab {
     });
 
     // Find the inspected frame
-    const {startedInPageEvt, frameId} = TracingProcessor.findTracingStartedEvt(keyEvents);
+    const tracingIds = TracingProcessor.findTracingIds(keyEvents);
```
patrickhulce: I'm not super in love with `tracingIds` though I suppose it is a pretty accurate description of what they are. only other things I'm thinking of are `mainThreadIdentifiers` or `mainFrameIdentifiers` since that's the heart of what we're doing with them

typings/externs.d.ts
Hoten: It isn't needed for this. I forgot to remove this (from the data uri changes).

`findTracingIds` doesn't work when you use a data uri, since the PID changes, and thus there is no metadata for the expected pid for the thread `CrRendererMain`. But, there is an event we can use to match the frameId (which doesn't change!) to the new pid (see https://github.com/GoogleChrome/lighthouse/blob/undo-revert-datauri/lighthouse-core/lib/traces/tracing-processor.js#L225 ). TL;DR - `processId` is just on the event `FrameCommittedInBrowser`, which is not needed for `about:blank`.

lighthouse-core/lib/traces/tracing-processor.js
Hoten: yeah, when ts = 0 things break. I don't recall if it was just from testing data, or from an actual run of LH.

Hoten: I'll add it to the guard but as an undefined check, rather than truthy.

lighthouse-core/gather/gather-runner.js
Hoten: better! I refactored the old code (using this structure), and it was using Promise.then, so that's why this made sense at one point :P

lighthouse-core/gather/computed/trace-of-tab.js
Hoten: yeah, I just couldn't come up with a good name with so little context. maybe one day I can name things as good as the best. I'll defer to you - which do you like better?

lighthouse-core/gather/gather-runner.js
```diff
@@ -394,7 +404,6 @@ class GatherRunner {
     try {
       await driver.connect();
       const baseArtifacts = await GatherRunner.getBaseArtifacts(options);
-      await GatherRunner.loadBlank(driver);
```
Hoten: this probably should stay.

lighthouse-core/gather/gather-runner.js
Hoten: can I just move `baseArtifacts.BenchmarkIndex = await options.driver.getBenchmarkIndex();` to after the call to `setupDriver`?

Hoten:
Yes, they are fairly unrelated. One was already done in the branch I revived attempting to implement #3707. The other came about while tweaking how we get all those ids, and at the same time a nice refactor came out of it.

lighthouse-core/gather/gather-runner.js
patrickhulce: ideally we run it before we start registering all the in-page listeners and whatnot, but that might be better than on existing page 👍

lighthouse-core/gather/computed/trace-of-tab.js
patrickhulce: 😆 we can only dream of being close to the naming master @paulirish 

since we've got `frame` in the mix, I'd go with `TracingProcessor.findMainFrameIds` and `mainFrameIds`, but others might have better ideas

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -198,39 +198,56 @@ class TraceProcessor {
 
   /**
    * @param {LH.TraceEvent[]} events
-   * @return {{startedInPageEvt: LH.TraceEvent, frameId: string}}
+   * @return {{pid: number, tid: number, frameId: string, ts: number}}
```
brendankenny: if the timestamp is *only* used for tests, we should maybe just drop `ts`

lighthouse-core/test/gather/computed/trace-of-tab-test.js
```diff
@@ -54,7 +54,7 @@ describe('Trace of Tab computed artifact:', () => {
   describe('finds correct FMP', () => {
     it('if there was a tracingStartedInPage after the frame\'s navStart', async () => {
       const trace = await TraceOfTab.compute_(lateTracingStartedTrace);
-      assert.equal(trace.startedInPageEvt.ts, 29343544280);
+      assert.equal(trace.tracingIds.ts, 29343544280);
```
brendankenny: these tests are more interested in making sure the correct things were selected from the trace than the particular `startedInPageEvt` timestamps. If dropping `tracingIds.ts`, the replacement for `trace.startedInPageEvt.ts` could be the `pid`, `tid`, or `frameId` (or all of them) with no loss of functionality to the test

Hoten:
Split to https://github.com/GoogleChrome/lighthouse/pull/6369 and https://github.com/GoogleChrome/lighthouse/pull/6370

brendankenny:
Thanks!

PR #6369, opened 2018-10-23
core: don't load blank page twice in first pass

Hoten:
Split from #6366.

Skip loading the blank page a second time on the first pass. It's already been loaded.


lighthouse-core/gather/gather-runner.js
```diff
@@ -105,7 +105,9 @@ class GatherRunner {
   static async setupDriver(driver, options) {
     log.log('status', 'Initializing…');
     const resetStorage = !options.settings.disableStorageReset;
-    // Enable emulation based on settings
+    // In the devtools/extension case, we can't still be on the site while trying to clear state
+    // So we first navigate to a blank page, then apply our emulation & setup
+    await GatherRunner.loadBlank(driver);
```
Hoten: should this use the config for the first pass (for blank page and duration)?

lighthouse-core/gather/gather-runner.js
```diff
@@ -394,12 +394,11 @@ class GatherRunner {
     try {
       await driver.connect();
       const baseArtifacts = await GatherRunner.getBaseArtifacts(options);
-      await GatherRunner.loadBlank(driver);
-      baseArtifacts.BenchmarkIndex = await options.driver.getBenchmarkIndex();
       await GatherRunner.setupDriver(driver, options);
+      baseArtifacts.BenchmarkIndex = await options.driver.getBenchmarkIndex();
```
patrickhulce: wait I take it back this won't work :)

It needs to go before we setup emulation since its purpose is establishing the right multiplier (see #6162)

it doesn't seem like it'd be the end of the world to have the two `loadBlank` calls in this function, was there another reason I'm forgetting that it's inside `setupDriver` now?

lighthouse-core/gather/gather-runner.js
patrickhulce: Nah, I think @paulirish was even borderline on the way to removing the `blankPage` and `blankDuration` params entirely from the passes anyway

lighthouse-core/gather/gather-runner.js
brendankenny: +1 to restoring the original `loadBlank` line

Hoten: done

lighthouse-core/gather/gather-runner.js
patrickhulce: I'm a bit confused, aren't we still loading blank twice now? Seems like this line should go away and the comments moved into the `loadBlank` call in `run`

patrickhulce: sorry if I confused you with my "two blank calls" comment I meant for the pass and for before setup driver, not for setup driver + before setup driver :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -26,22 +26,22 @@ const Driver = require('../gather/driver.js'); // eslint-disable-line no-unused-
  * Execution sequence when GatherRunner.run() is called:
  *
  * 1. Setup
- *   A. navigate to about:blank
  *   B. driver.connect()
```
brendankenny: I think this is now
```
A. driver.connect()
B. navigate to about:blank
```

lighthouse-core/gather/gather-runner.js
```diff
@@ -26,22 +26,22 @@ const Driver = require('../gather/driver.js'); // eslint-disable-line no-unused-
...
  *   C. GatherRunner.setupDriver()
- *     i. assertNoSameOriginServiceWorkerClients
- *     ii. beginEmulation
- *     iii. enableRuntimeEvents
- *     iv. evaluateScriptOnLoad rescue native Promise from potential polyfill
- *     v. register a performance observer
- *     vi. register dialog dismisser
- *     vii. clearDataForOrigin
+ *     i. navigate to a blank page
```
brendankenny: remove line

lighthouse-core/gather/gather-runner.js
```diff
@@ -26,22 +26,22 @@ const Driver = require('../gather/driver.js'); // eslint-disable-line no-unused-
...
+ *     vii. register a performance observer
+ *     viii. register dialog dismisser
+ *     iv. clearDataForOrigin
  *
  * 2. For each pass in the config:
  *   A. GatherRunner.beforePass()
- *     i. navigate to about:blank
- *     ii. Enable network request blocking for specified patterns
- *     iii. all gatherers' beforePass()
+ *     i. Enable network request blocking for specified patterns
```
brendankenny: still need `i. navigate to about:blank` from master (we could maybe add a parenthetical about the first pass? but I don't think it's needed for just the overview up here)

PR #6370, opened 2018-10-23
Refactor tracing processor 2

Hoten:
Split from #6366.

Refactor how we get the pid/tid/frameId of the current target. The tracing processor was returning a specific event, and even synthesizing it. The callers of this function only grabbed the pid/tid/frameId, so instead of returning an event object I return just the pertinent data.


lighthouse-core/test/gather/computed/trace-of-tab-test.js
```diff
@@ -54,7 +54,7 @@ describe('Trace of Tab computed artifact:', () => {
   describe('finds correct FMP', () => {
     it('if there was a tracingStartedInPage after the frame\'s navStart', async () => {
       const trace = await TraceOfTab.compute_(lateTracingStartedTrace);
-      assert.equal(trace.startedInPageEvt.ts, 29343544280);
```
patrickhulce: building on @brendankenny's earlier point do we ever really use the `ts` anywhere other than the tests here? 

feels weird to have `ts` on a property called `*Ids` (my fault I know 😄) and as he pointed out we really just want to make sure that we're identifying the mainFrame correctly

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -198,39 +198,50 @@ class TraceProcessor {
...
+    }
 
-    // @ts-ignore - property chain exists for 'TracingStartedInPage' event.
-    const frameId = /** @type {string} */ (startedInPageEvt.args.data.page);
-    return {startedInPageEvt, frameId};
+    return {
+      pid,
+      tid,
+      frameId,
+    };
```
brendankenny: Style bikeshedding: what about some early returns? That makes it a little easier to see it's either the first conditional matches and it's done, or it's the next conditional and done, or it throws.

Something like
```js
static findMainFrameIds(events) {
  // Prefer the newer TracingStartedInBrowser event first, if it exists
  const startedInBrowserEvt = events.find(e => e.name === 'TracingStartedInBrowser');
  if (startedInBrowserEvt && startedInBrowserEvt.args.data &&
      startedInBrowserEvt.args.data.frames) {
    const mainFrame = startedInBrowserEvt.args.data.frames.find(frame => !frame.parent);
    const frameId = mainFrame && mainFrame.frame;
    const pid = mainFrame && mainFrame.processId;

    const threadNameEvt = events.find(e => e.pid === pid && e.ph === 'M' &&
      e.cat === '__metadata' && e.name === 'thread_name' && e.args.name === 'CrRendererMain');
    const tid = threadNameEvt && threadNameEvt.tid;

    if (pid && tid && frameId) {
      return {
        pid,
        tid,
        frameId,
      };
    }
  }

  // Support legacy browser versions that do not emit TracingStartedInBrowser event.
  // The first TracingStartedInPage in the trace is definitely our renderer thread of interest
  // Beware: the tracingStartedInPage event can appear slightly after a navigationStart
  const startedInPageEvt = events.find(e => e.name === 'TracingStartedInPage');
  if (startedInPageEvt && startedInPageEvt.args && startedInPageEvt.args.data) {
    const frameId = startedInPageEvt.args.data.page;

    if (frameId) {
      return {
        pid: startedInPageEvt.pid,
        tid: startedInPageEvt.tid,
        frameId,
      };
    }
  }

  throw new LHError(LHError.errors.NO_TRACING_STARTED);
}
```

typings/artifacts.d.ts
```diff
@@ -359,8 +359,8 @@ declare global {
         processEvents: Array<TraceEvent>;
         /** The subset of trace events from the page's main thread, sorted by timestamp. */
         mainThreadEvents: Array<TraceEvent>;
-        /** The event marking the start of tracing in the target browser. */
-        startedInPageEvt: TraceEvent;
+        /** Various ids for the current trace context. */
```
brendankenny: maybe something like `/** IDs for the trace's main frame, process, and thread. */`?

PR #6372, opened 2018-10-23
core(report): right-align DOM size table values

### LINKED ISSUE #6084: Left align for numeric columns?
When getting the DOM size table, the Value column is `text` and is therefore left aligned, should it be right aligned?
add_numeric

I suggest adding a `lh-table-column--numeric` type for things that are just numbers with no type that need to be left aligned?

Hoten:
Fixes #6084.

image


proto/sample_v2_round_trip.json
```diff
@@ -460,7 +460,7 @@
                         "text": "Element"
                     }, 
                     {
-                        "itemType": "text", 
+                        "itemType": "numeric",
```
Hoten: @exterkamp  I had to manually edit this one :'(

PR #6374, opened 2018-10-24
misc(proto): align proto whitespace

exterkamp:
**Summary**
Missed tab offset for timing :cry: 

It's really fixed now...

### DIFF
--- a/proto/lighthouse-result.proto
+++ b/proto/lighthouse-result.proto
@@ -223,10 +223,10 @@ message LighthouseResult {
   message Timing {
     // The total duration of Lighthouse's run
     google.protobuf.DoubleValue total = 1;
- }
+  }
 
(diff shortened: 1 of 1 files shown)


Hoten:
Yup, there is [this linter](https://github.com/ckaznocha/protoc-gen-lint), implemented as a `protoc` plugin.

paulirish:
> Any chance there's a public proto linter we can add? :)

right after this PR went up, shane and i tried out the one @Hoten linked. unfortunately, it doesnt care about line length. nor indentation.

the google linter actually also doesn't care about indentation either.

But that all said, I def agree there's an opportunity for some stronger tooling here. thx @Hoten  for filing #6376

Hoten:
Seems [this Ruby linter works](https://codebeautify.org/ruby-formatter-beautifier) :)

PR #6382, opened 2018-10-24
misc: enforce linting for infix-op spacing

brendankenny:
based on this: https://github.com/GoogleChrome/lighthouse/pull/6286#pullrequestreview-167563083 :)

It's surprising we don't enforce spacing around ternary operators, etc already. Enabled the [eslint rule](https://eslint.org/docs/rules/space-infix-ops) to enforce it. The few places where we were breaking seem like things we would normally correct anyways, so it seems reasonable to leave it enforced.


patrickhulce:
I was wondering why eslint wasn't catching those :)

PR #6383, opened 2018-10-24
misc(logger): isVerbose(), getTimeEntries()

### LINKED ISSUE #6313: lighthouse-logger improvments
It's kind of annoying to update lighthouse-logger, but there are some rough edges that we'd like to smooth out. Came up in #6297:

- [x] Some way to query what the logging level is. Not usually needed, but sometimes you really want to skip work for `verbose`-only logs (e.g. querying things, slicing strings, etc) when it's not needed.
- [ ] logging level in between `verbose` and `default` (`debug` or `info` perhaps?) that doesn't contain every protocol message

Hoten:
Fixes #6313. Unblocks #3745.

### DIFF
--- a/lighthouse-logger/index.js
+++ b/lighthouse-logger/index.js
@@ -52,6 +52,7 @@ class Emitter extends EventEmitter {
 
 const loggersByTitle = {};
 const loggingBufferColumns = 25;
+let level_;
 
 class Log {
   static _logToStdErr(title, argsArray) {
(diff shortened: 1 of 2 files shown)


brendankenny:
for background for others, we decided on just doing `isVerbose()` (instead of `isWarning`, `isError`, etc, or some other `getLevel()`/`isLevel(levelStr)`/whatever API) for now since that's the only real need we've had so far and so reimplementing/exposing the log level hierarchy would be more work for no benefit today.

We can always revisit when someone actually needs `logger.isError()` :)

Opinions on the specific choice we made are welcome, of course :)

PR #6385, opened 2018-10-24
deps(logger): bump version for publish

paulirish:
allowing #3745 to proceed with 1.3.0..


paulirish:
> does it need to be bumped again over #6383 bumping it to 1.2.0?

nope.

Hoten:
double bump

PR #6391, opened 2018-10-25
core(font-size): speed up gatherer

### LINKED ISSUE #6386: Speed up the slow font-size gatherer
The font-size gatherer is fairly slow (500ms - 1500ms), especially on text heavy pages.

from #6347 

@hoten: 

> Got some timings. https://gist.github.com/Hoten/de38179b2d941b6d195c3c4383d0bd8d
> .... Everything that took longer than 1s seems related to font stuff (CSS.getComputedStyleForNode), which I think we are getting rid of soon? If not, we might consider 1) increasing the timeout for those commands or 2) instead of sending them all at once, send them in batches.

@brendankenny 

> That gatherer sends all those `CSS.getComputedStyleForNode` commands at once, up to 500, depending on the page's node count. It's possible command overhead/contention causes issues, especially on an older phone. We could try doing them in batches instead of all at once to see if that's a solution.

> It's also remotely possible we hit some gross style-recalc issue, depending on the page and how the commands are coming in. It's possible it would be more efficient to do all the lookups in the page at once with `evaluateAsync`, since we could guarantee running them all in the same turn of the event loop/without reforcing layout. We'd have to resync node info and debugger `nodeId`s, though. Maybe there's a similar coalesced solution, though.

--------

talked with dgozman and appear to be ~3 options, two of which brendan outlined above.

1. batch the current `CSS.getComputedStyleForNode` calls. there can be up to 500. batching we'd avoid the timeout more safely, but this gatherer would still take a LONG time (~300-1000ms). this doesn't seem very compelling.
2. evaluateAsync and do our own `getComputedStyle` inside the page. would be very fast to collect (~25ms). but matching up `nodeId`'s might be a dealbreaker. There's not a good way to do this.
3. Use [`DOMSnapshot.captureSnapshot`](https://chromedevtools.github.io/devtools-protocol/tot/DOMSnapshot#method-captureSnapshot). It collects more data than we need, and its payload is like 300kb. But it gives us computed style for the entire tree (including iframes) in 60ms. Has some potential, but need to explore it more. It's data structure is wild, but it looks like it has the data we need.

To try this out, do devtools-on-devtools and try:  
```js
await Main.sendOverProtocol('DOMSnapshot.captureSnapshot', {computedStyles: ['font-size']})
```

Hoten:
Fixes #6386.

See comments in discussion and code. LMK if anything is unclear :)


lighthouse-core/audits/seo/font-size.js
```diff
@@ -208,7 +208,6 @@ class FontSize extends Audit {
       analyzedFailingNodesData,
       analyzedFailingTextLength,
       failingTextLength,
-      visitedTextLength,
```
Hoten: I removed `visitedTextLength` because we can now easily visit every node (it's one protocol call).

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,16 +254,93 @@ class FontSize extends Gatherer {
...
+      computedStyles: ['font-size'],
+    });
+
+    // Makes the strings access code easier to read.
+    /** @param {number} index */
+    const lookup = (index) => snapshot.strings[index];
+
+    // Locate the document under analysis.
+    // TODO: this needs to use frameId
+    const doc = snapshot.documents.find(doc => lookup(doc.documentURL) === passContext.url);
```
Hoten: I need help here.

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -13,47 +13,83 @@ let fontSizeGather;
...
 const bigText = 'body big text';
 const failingText = 'failing text';
-const bodyNode = {nodeId: 3, nodeName: 'BODY', parentId: 1};
-const failingNode = {nodeId: 10, nodeName: 'P', parentId: 3};
+const bodyNode = {nodeId: 3, backendNodeId: 102, nodeName: 'BODY', parentId: 1, fontSize: '10px'};
+const failingNode = {nodeId: 10, backendNodeId: 109, nodeName: 'P', parentId: 3};
 const nodes = [
-  {nodeId: 1, nodeName: 'HTML'},
-  {nodeId: 2, nodeName: 'HEAD', parentId: 1},
+  {nodeId: 1, backendNodeId: 100, nodeName: 'HTML'},
```
Hoten: @patrickhulce  I hope I maintained the integrity of this test :)

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: Hm, is it not true that "The DOMNode at index 0 corresponds to the root document"? Should poke devtools folks to update that then :)

AFAIK, the only main frame id we have/use is in the trace (the mainFrameIds refactor you just did), and I don't think we're collecting it anywhere else that'd be readily available to a gatherer.

You could query it with `Page.getResourceTree` `resourceTreeResponse.frameTree.frame.id` or use the network records to get at the frame id if we need to

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,16 +254,93 @@ class FontSize extends Gatherer {
...
+      });
+    }
+
+    const nodes = await getAllNodesFromBody(passContext.driver);
+
+    // backendIdsToPartialFontData will include all Nodes,
+    // but nodes will only contain the Body node and its descendants.
+
+    /** @type {NodeFontData[]} */
+    const failingNodes = [];
```
patrickhulce: do you think we could move all this into some functions for the different steps you're adding so `afterPass` will be pretty short like before? I think it might help human-parsers figure out what's going on :)

it looks like there's roughly the following stages

1. get the flattened document + snapshot and make sure its all valid
2. create our mappings to be able to tie the two together
3. go through all the nodes and identify the failing ones

is that roughly accurate?

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -13,47 +13,83 @@ let fontSizeGather;
...
+    backendNodeId: 110,
     nodeValue: failingText,
     nodeType: FontSizeGather.TEXT_NODE_TYPE,
     parentId: 10,
   },
 ];
 
+const stringsMap = {};
+const strings = [];
+const dedupe = (value) => {
```
patrickhulce: WDYT about something like `getOrCreateStringRef`/getOrCreateStringIndex` to indicate the main reason we're doing this is so that we're dealing with that `.strings` property

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
```diff
@@ -13,47 +13,83 @@ let fontSizeGather;
...
+    {
+      nodes: {
+        nodeType: nodes.map(node => node.nodeType),
+        nodeName: nodes.map(node => dedupe(node.nodeName)),
+        nodeValue: nodes.map(node => dedupe(node.nodeValue)),
+        backendNodeId: nodes.map(node => node.backendNodeId),
+        parentIndex: nodes.map(node => nodes.findIndex(pNode => node.parentId === pNode.nodeId)),
+      },
+      layout: {
+        nodeIndex: nodes.map((_, i) => i), // this isn't very accurate
```
patrickhulce: mind elaborating on the comment a bit? :)

is it inaccurate because the nodeIndexes are usually very large? are they usually not sequential? do any of these things affect the fidelity of the tests?

lighthouse-core/audits/seo/font-size.js
```diff
@@ -273,14 +272,7 @@ class FontSize extends Audit {
...
       const percentageOfFailingText = parseFloat((100 - percentageOfPassingText).toFixed(2));
-      let disclaimer = '';
-
-      // if we were unable to visit all text nodes we should disclose that information
-      if (visitedTextLength < totalTextLength) {
-        const percentageOfVisitedText = visitedTextLength / totalTextLength * 100;
-        disclaimer = ` (based on ${percentageOfVisitedText.toFixed()}% sample)`;
-      }
-
+      const disclaimer = '';
```
patrickhulce: can just remove disclaimer now ;)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,16 +254,93 @@ class FontSize extends Gatherer {
...
+      })) continue;
+
+      const styleIndex = nodeIndexToStyleIndex.get(doc.nodes.parentIndex[i]);
+      if (!styleIndex) continue;
+      const parentStyles = doc.layout.styles[styleIndex];
+      const [fontSizeStringId] = parentStyles;
+      const fontSize = parseInt(lookup(fontSizeStringId), 10);
+      backendIdsToPartialFontData.set(doc.nodes.backendNodeId[i], {
+        fontSize,
+        // TODO: trimming this for a second time. maybe don't?
```
patrickhulce: is that because it's internally being trimmed?

lighthouse-core/gather/gatherers/seo/font-size.js
Hoten: Yeah that sounds good. I mostly wanted to wait for PR feedback before chopping up afterPass, just incase the whole thing was moot or something :P

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
Hoten: If you take a look at an actual snapshot, you'll see that there isn't a 1:1 mapping between Nodes and style entries. That's because some elements (like text nodes) don't have styles. Here, I'm being lazy and saying that all nodes have computed styles, and that the mapping is just the identity function.

It is indeed missing out on a small amount of functionality, namely the nodeId => styleIndex mapping in the implementation. it wasn't immediately clear how to generate that part of the fake snapshot so I just punted it. I could look into it a bit more.

lighthouse-core/gather/gatherers/seo/font-size.js
Hoten: because the check before (non empty text node) trims it too.

Hoten: could just make that function call check for a text node, and have the "non empty" part come after. I'll do that. Was mostly trying to keep the previous function as it was.

patrickhulce: but `.trim` doesn't mutate and we're not reassigning `nodeValue` so it should still be needed right? I don't think you have to refactor anything :)

Hoten:
I'll go through the other issue and test some URLs.

lighthouse-core/test/gather/gatherers/seo/font-size-test.js
Hoten: let's trust the protocol documentation

https://chromedevtools.github.io/devtools-protocol/tot/DOMSnapshot#method-captureSnapshot
> The DOMNode at index 0 corresponds to the root document.

If it's wrong then testing will catch it :)

brendankenny:
one key question: is it actually faster? :)

Hoten:
Not sure yet. Need to rebase to get the perf tracing. I'll definitely update w/ my findings

Hoten:
[traces.zip](https://github.com/GoogleChrome/lighthouse/files/2520782/traces.zip)

`lh:gather:afterPass:FontSize`
cnn master 810
cnn pr 472
nashville master 1317
nashville pr 492

Would be useful to have a PR hook (or just print something in travis ci) that shows us diffs on these marks. Maybe collect all the timings from the smoke tests, aggregate the durations of each mark, and pretty print the time deltas against master.

lighthouse-core/gather/gatherers/seo/font-size.js
patrickhulce: ooooooh I see what you were saying now 👍

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,90 @@ class FontSize extends Gatherer {
...
+        parentNodeName: lookup(doc.nodes.nodeName[doc.nodes.parentIndex[i]]),
+      })) continue;
+
+      const styleIndex = nodeIndexToStyleIndex.get(doc.nodes.parentIndex[i]);
+      if (!styleIndex) continue;
+      const textLength = getNodeTextLength(nodeValue);
+      if (!textLength) continue; // ignore empty TextNodes
+      const parentStyles = doc.layout.styles[styleIndex];
+      const [fontSizeStringId] = parentStyles;
+      const fontSize = parseInt(lookup(fontSizeStringId), 10);
```
brendankenny: maybe another comment here that we expect strings like `'16px'`, hence why the `parseInt` is appropriate

brendankenny: also, do we actually want `parseFloat`? I guess 11.5 will still fail the < 12 check, but we might as well store the most accurate info

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,17 +335,28 @@ class FontSize extends Gatherer {
...
+      passContext.driver.sendCommand('DOMSnapshot.enable'),
       passContext.driver.sendCommand('DOM.enable'),
       passContext.driver.sendCommand('CSS.enable'),
     ]);
 
+    // We need to find all TextNodes that do not have legible text. DOMSnapshot.captureSnapshot is the
+    // fastest way to get the computed styles of every Node. Bonus, it allows for whitelisting properties.
+    // Once a bad TextNode is identified, its parent Node is needed. DOMSnapshot.captureSnapshot doesn't
+    // give the entire Node object, so DOM.getFlattenedDocument is used. The only connection between a snapshot
+    // Node and an actual Protocol Node is backendId, so that is used to join the two data structures.
```
brendankenny: maybe just
`// Use DOMSnapshot.captureSnapshot to get the computed font-size style of every node.`
(the rest can be filled in below as it happens)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,17 +335,28 @@ class FontSize extends Gatherer {
...
+    const snapshot = await passContext.driver.sendCommand('DOMSnapshot.captureSnapshot', {
+      computedStyles: ['font-size'],
+    });
+    // backendIdsToPartialFontData will include all Nodes,
```
brendankenny: not really true? (it just contains non-empty text nodes)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,17 +335,28 @@ class FontSize extends Gatherer {
...
+    const backendIdsToPartialFontData = this.calculateBackendIdsToPartialFontData(snapshot);
+    // but nodes will only contain the Body node and its descendants.
+    const nodes = await getAllNodesFromBody(passContext.driver);
```
brendankenny: maybe we now want to differentiate `nodes` from the nodes in the snapshot, somehow? I'm not sure of a great way. `crdpNodes` or something?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,90 @@ class FontSize extends Gatherer {
...
+        fontSize,
+        textLength,
+      });
+    }
+
+    return backendIdsToPartialFontData;
+  }
+
+  /**
+   * @param {BackendIdsToPartialFontData} backendIdsToPartialFontData
```
brendankenny: A jsdoc description of the method would be good here (e.g. the comment from below about matching font data from the snapshot to node references via `backendId`)

brendankenny:
@patrickhulce can you find the discussion # with some test sites? I can't find it now

patrickhulce:
> can you find the discussion # with some test sites? I can't find it now

https://github.com/GoogleChrome/lighthouse/pull/3533

but I'm starting to be afraid my back and forth with konrad with more URLs was in chats instead of that PR, the only ones I found were cnn and flynashville that @Hoten already tested 😞

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,90 @@ class FontSize extends Gatherer {
...
+    // Implementation:
+    // https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/inspector/inspector_dom_snapshot_agent.cc?sq=package:chromium&g=0&l=534
+
+    if (!doc || !doc.nodes.nodeType || !doc.nodes.nodeName || !doc.nodes.backendNodeId
+      || !doc.nodes.nodeValue || !doc.nodes.parentIndex) {
+      throw new Error('Unexpected response from DOMSnapshot.captureSnapshot.');
+    }
+
+    // Not all nodes have computed styles (ex: TextNodes), so doc.layout.* is smaller than doc.nodes.*
+    // doc.layout.nodeIndex maps the index into doc.nodes.* to an index into doc.layout.styles.
```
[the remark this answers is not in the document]
Hoten: yup, nice catch

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,90 @@ class FontSize extends Gatherer {
...
+    // nodeIndexToStyleIndex inverses that mapping.
+    /** @type {Map<number, number>} */
+    const nodeIndexToStyleIndex = new Map();
+    for (let i = 0; i < doc.layout.nodeIndex.length; i++) {
+      nodeIndexToStyleIndex.set(doc.layout.nodeIndex[i], i);
```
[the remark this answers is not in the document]
brendankenny: > just to make sure, are these 1:1? 

yah? nah? :)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,94 @@ class FontSize extends Gatherer {
...
+    }
+
+    /** @type {BackendIdsToFontData} */
+    const backendIdsToFontData = new Map();
+    for (let i = 0; i < doc.nodes.nodeType.length; i++) {
+      const nodeType = doc.nodes.nodeType[i];
+      const nodeValue = strings[doc.nodes.nodeValue[i]];
+      if (!isTextNode({
+        nodeType,
+        parentNodeName: strings[doc.nodes.nodeName[doc.nodes.parentIndex[i]]],
```
brendankenny: still would be good to pull this into a variable (or variables), especially to make the conditional obvious at a glance

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,90 @@ class FontSize extends Gatherer {
...
+        nodeType,
+        parentNodeName: lookup(doc.nodes.nodeName[doc.nodes.parentIndex[i]]),
+      })) continue;
+
+      const styleIndex = nodeIndexToStyleIndex.get(doc.nodes.parentIndex[i]);
+      if (!styleIndex) continue;
+      const textLength = getNodeTextLength(nodeValue);
+      if (!textLength) continue; // ignore empty TextNodes
+      const parentStyles = doc.layout.styles[styleIndex];
+      const [fontSizeStringId] = parentStyles;
```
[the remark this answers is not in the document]
brendankenny: a comment here would still be good too. The type is just `Array<string>`, so it and the protocol docs don't give much help on why this is the way it is

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
Hoten: Correct, AFAIKT from examining the output (printing out layout.nodeIndex) and the source ( https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/inspector/inspector_dom_snapshot_agent.cc?sq=package:chromium&g=0&l=902 ) - these are 1:1

lighthouse-core/gather/gatherers/seo/font-size.js
Hoten: to be clear, do you mean
```
const isTextNode = !isTextNode({
        nodeType,
        parentNodeName: strings[doc.nodes.nodeName[doc.nodes.parentIndex[i]]],
      })
if (!isTextNode) {
  continue;
}
```

or something else?

patrickhulce: I assumed he meant breaking up `strings[doc.nodes.nodeName[doc.nodes.parentIndex[i]]]` but even if he didn't, I think that would be nice if it isn't too much trouble :)

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
Hoten: Add a comment explaining why there's just the one value here.

I'm unsure about adding an assertion for .length === 1 here. This property is inherent to the API for this protocol method - do we assert post conditions like this often for other protocol methods?

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,17 +344,24 @@ class FontSize extends Gatherer {
...
     ]);
 
+    // Use DOMSnapshot.captureSnapshot to get the computed font-size style of every node.
+    const snapshot = await passContext.driver.sendCommand('DOMSnapshot.captureSnapshot', {
+      computedStyles: ['font-size'],
+    });
+    // backendIdsToFontData will include all non-empty TextNodes,
+    const backendIdsToFontData = this.calculateBackendIdsToFontData(snapshot);
+    // but nodes will only contain the Body node and its descendants.
+    const crdpNodes = await getAllNodesFromBody(passContext.driver);
```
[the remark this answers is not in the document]
Hoten: The FontSize artifact contains [entire CRDP Nodes](https://github.com/GoogleChrome/lighthouse/blob/ff3ccde51f95bb7c58ce5ec4207039eb41181ac0/typings/artifacts.d.ts#L192). The snapshot data doesn't have all the data to reconstruct a CRDP Node.

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -304,6 +239,99 @@ class FontSize extends Gatherer {
...
+        nodeType,
+        parentNodeName,
+      })) continue;
+
+      const styleIndex = nodeIndexToStyleIndex.get(parentIndex);
+      if (!styleIndex) continue;
+      const textLength = getNodeTextLength(nodeValue);
+      if (!textLength) continue; // ignore empty TextNodes
+      const parentStyles = doc.layout.styles[styleIndex];
+      // each styles is an array of string indices, one for each property given
```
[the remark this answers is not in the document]
Hoten: `doc.layout.styles is an array of an array of string indices ` ...

Hoten:
I get a different set of results.

image

Differences could be due to plugins, or the site serving different content based on user agent, or the code being wrong :P

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
paulirish: I think we don't need everything that's in CRDP Node.
The audit only uses node.parentNode and node.attributes. It looks like all that data is already in the snapshot.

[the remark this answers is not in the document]
brendankenny: > I think we don't need everything that's in CRDP Node.

we need `CSS.getMatchedStylesForNode` to get the current styles for a node, which needs the nodeId to query for it

Hoten:
Tried this URL: https://www.clumsycrafter.com/the-water-blob-tutorial-using-an-iron-to-seal-the-edges/

The old code reports the "Report this ad" as an illegible node:
image

The new code does not.

For this site, it's obvious that this ad takes awhile to display on the page. I think the difference we are seeing is due to the old code allowing the page more time to do stuff, so the page has time to render things (like the above ad). The code changes here shouldn't affect how early things like snapshot/get flattened document are called, but maybe something else in master (but not this branch) is. I'll merge master into this branch now so I can eliminate that potential difference.

lighthouse-core/gather/gatherers/seo/font-size.js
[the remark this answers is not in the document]
Hoten: @paulirish , given the above (we need the snapshot + the flattened document protocol calls), should we still reduce the scope of artifact?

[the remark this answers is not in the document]
paulirish: Yeah looks like the snapshot doesn't have `nodeId`. only `backendNodeId`. Bummer.

K. We'll keep the current protocol calls. :)

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -316,17 +344,24 @@ class FontSize extends Gatherer {
     passContext.driver.on('CSS.styleSheetAdded', onStylesheetAdd);
 
     await Promise.all([
+      passContext.driver.sendCommand('DOMSnapshot.enable'),
       passContext.driver.sendCommand('DOM.enable'),
       passContext.driver.sendCommand('CSS.enable'),
     ]);
 
+    // Use DOMSnapshot.captureSnapshot to get the computed font-size style of every node.
+    const snapshot = await passContext.driver.sendCommand('DOMSnapshot.captureSnapshot', {
```
paulirish: very small speedup if you run these two in parallel:

```diff
 lighthouse-core/gather/gatherers/seo/font-size.js | 8 +++++---
 1 file changed, 5 insertions(+), 3 deletions(-)

diff --git a/lighthouse-core/gather/gatherers/seo/font-size.js b/lighthouse-core/gather/gatherers/seo/font-size.js
index e6d0dbe0..3ceac9bd 100644
--- a/lighthouse-core/gather/gatherers/seo/font-size.js
+++ b/lighthouse-core/gather/gatherers/seo/font-size.js
@@ -350,13 +350,15 @@ class FontSize extends Gatherer {
     ]);
 
     // Use DOMSnapshot.captureSnapshot to get the computed font-size style of every node.
-    const snapshot = await passContext.driver.sendCommand('DOMSnapshot.captureSnapshot', {
+    const snapshotPromise = passContext.driver.sendCommand('DOMSnapshot.captureSnapshot', {
       computedStyles: ['font-size'],
     });
+    // but nodes will only contain the Body node and its descendants.
+    const allNodesPromise  = getAllNodesFromBody(passContext.driver);
+    const [snapshot, crdpNodes] = await Promise.all([snapshotPromise, allNodesPromise]);
+
     // backendIdsToFontData will include all non-empty TextNodes,
     const backendIdsToFontData = this.calculateBackendIdsToFontData(snapshot);
-    // but nodes will only contain the Body node and its descendants.
-    const crdpNodes = await getAllNodesFromBody(passContext.driver);
     const {
       totalTextLength,
       failingTextLength,
```

Hoten:
One of the missing nodes for the site Paul tried (https://www.haaretz.com/) has 8px font.

I looked up '8px' in `snapshot.strings`, found an index. I look for this index in `doc.layout.styles`, no results. Boo.

I looked up the offending node (from the DOM.getFlattenedDocument method) by class name:
```
crdpNodes.filter(n => n.attributes && n.attributes.length === 2 && n.attributes[1] === 'cbola-banner-structure__advertisement-title cbola-fade-out')
```

and walked up the parentNode hierarchy until I got to a BODY element. Grabbed it's `backendNodeId`, and saw that there is _no_ reference to it in `doc.nodes.backendNodeIds`. Also, this is just one of many (32) `BODY` nodes in `crpdNodes` - but `doc.nodes.nodeName` only references one `BODY` element.

So, I think that `DOMSnapshot.captureSnapshot` splits up the various documents in the page (iframes?) differently that how `DOM.getFlattendDocument` does. Even though the former says `documents[0] is the root` and the latter says `this returns the root document`. Where's the disconnect here?

image
Could it be we need to look at all of these?

Hoten:
Building `backendIdsToFontData` from all the documents in `snapshot` got me this:

image

which is only missing this from master:

image

Hoten:
Some of the documents in the snapshot don't contain any style information (`snapshot.documents[j].layout.styles` has no data). Seeing this for the User Agent Stylesheet node with 11px font size from haaretz.com.

image

So that's not so great.

brendankenny:
Should we
- figure out the difference another day and close, or
- is it in the immediate future and we should mark chillin

Hoten:
I'll give it another shot in the near future.

brendankenny:
Going to close for now since it's not on our immediate roadmap (I believe), but we can keep the branch open :)

PR #6393, opened 2018-10-25
docs: add primer on viewing the timing data

paulirish:
brendan asked for docs when merging #3745 so here we go

[example trace that one can view.](https://ahead-daughter.surge.sh/paulirish.json.timing.trace.html)

Also, fixed a small issue. Deduping by name turned out to be an issue since we call loadBlank and loadPage multiple times. Thus we lost data.

Now we dedupe by startTime, which has sub-millisecond precision, so we're good.

### DIFF
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -175,7 +175,8 @@ class Runner {
     const timingEntriesKeyValues = [
       ...timingEntriesFromArtifacts,
       ...timingEntriesFromRunner,
-    ].map(entry => /** @type {[string, PerformanceEntry]} */ ([entry.name, entry]));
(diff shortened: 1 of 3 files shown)


lighthouse-core/runner.js
```diff
@@ -175,7 +175,8 @@ class Runner {
     const timingEntriesKeyValues = [
       ...timingEntriesFromArtifacts,
       ...timingEntriesFromRunner,
-    ].map(entry => /** @type {[string, PerformanceEntry]} */ ([entry.name, entry]));
+      // As entries can share a name, dedupe based on the startTime timestamp
```
patrickhulce: nice

PR #6396, opened 2018-10-25
core(config): add new PWA category groups

brendankenny:
Starting on #6395

This PR just adds groups to the PWA category and deletes the manifest-short-name-length audit. Feel free to come up with better group descriptions or we can defer, I spent no time on them :)

Issue #6395, opened 2018-10-25
PWA category refactor tracking issue


brendankenny:
#### Fast and Reliable
- [x] Page load is fast enough on mobile networks (TTI < 10s)
- [x] Current URL responds with a 200 when offline
  - [x] rename: "Current page responds with a 200 when offline" #6494
- [x] start_url responds with 200 when offline - #6397

#### Installable
- [x] Uses HTTPS (either localhost or has valid certificate)
- [x] Registers a service worker
  - [x] check that test page URL is in SW scope #6609
  - [x] check that `start_url` is in SW scope #6678
- [x] Manifest requirements (starturl, name, display, icons) #6630

#### PWA Optimized
- [x] rename group #6494
- [x] Redirects HTTP traffic to HTTPS
  - [x] ~should be in a different group?~ (deferring question for now)
- [x] Is not configured for a custom splash screen
- [x] Address bar does not match brand colors
  - [x] needs a rename #6494
- [x] Has a <meta name="viewport"> tag with width or initial-scale
- [x] Content is sized correctly for the viewport
- [x] Contains some content when JavaScript is not available

#### Manual
- [x] Page transitions don't feel like they block on the network
- [x] Site works cross-browser
- [x] Each page has a URL

### Next:
- [x] add groups to default-config - #6396
- [x] Badging!
  - [x] Add grayscale group badges #6486, #6686
  - [x] Change to filled-in badges when group is passing - #6504, 
  - [x] Replace score gauge with badged gauge - #6526, #6670
- [x] Add adequate vertical spacing between audit groups (see [comment](https://github.com/GoogleChrome/lighthouse/issues/6395#issuecomment-439259634)) - #6688
- [x] Score gauge tooltip (see [comment](https://github.com/GoogleChrome/lighthouse/issues/6395#issuecomment-439562286)) - #6689
- [x] No more passed audit hiding 
  - [x] Refactor of current category-renderer to support new use cases #6460, #6482, #6594 
  - [x] Add pwa-category-renderer, don't hide passed audits #6486 

### Lighthouse v5
- ⁅ ⁆ Figure out LHR/scoring story

hwikyounglee:
And we landed on "PWA Optimized" 
cc: @kusler @paulirish  💅

kusler:
What about 'Enhancements'?

kusler:
Refinements
Further Improvements
?

hwikyounglee:
[...]

PR #6397, opened 2018-10-26
new_audit: offline-start-url

brendankenny:
Part of reorganization of the PWA category (#6395).

This splits the `start_url` offline check out of `webapp-install-banner.js` and into its own audit.

Based on top of #6396, so you may want to look at the diff between the two: https://github.com/GoogleChrome/lighthouse/compare/pwacat...start_url

### DIFF
--- a/lighthouse-core/audits/multi-check-audit.js
+++ b/lighthouse-core/audits/multi-check-audit.js
@@ -23,7 +23,7 @@ class MultiCheckAudit extends Audit {
   }
 
   /**
-   * @param {{failures: Array<string>, warnings?: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}} result
+   * @param {{failures: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}} result
    * @return {LH.Audit.Product}
    */
   static createAuditProduct(result) {
@@ -57,7 +57,6 @@ class MultiCheckAudit extends Audit {
     return {
       rawValue: true,
       details,
-      warnings: result.warnings,
     };
   }
 
@@ -66,7 +65,7 @@ class MultiCheckAudit extends Audit {
   /**
    * @param {LH.Artifacts} artifacts
    * @param {LH.Audit.Context} context
-   * @return {Promise<{failures: Array<string>, warnings?: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}>}
+   * @return {Promise<{failures: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}>}
    */
   static audit_(artifacts, context) {
     throw new Error('audit_ unimplemented');
--- a/lighthouse-core/audits/offline-start-url.js
+++ b/lighthouse-core/audits/offline-start-url.js
@@ -1,0 +1,50 @@
+/**
+ * @license Copyright 2018 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+const Audit = require('./audit.js');
+
+class OfflineStartUrl extends Audit {
+  /**
+   * @return {LH.Audit.Meta}
+   */
+  static get meta() {
+    return {
+      id: 'offline-start-url',
+      title: 'start_url responds with a 200 when offline',
+      failureTitle: 'start_url does not respond with a 200 when offline',
+      description: 'If you\'re building a Progressive Web App, consider using a service worker ' +
+          'so that your app can work offline. ' +
+          '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).',
+      requiredArtifacts: ['Manifest', 'StartUrl'],
+    };
+  }
+
+  /**
+   * @param {LH.Artifacts} artifacts
+   * @return {LH.Audit.Product}
+   */
+  static audit(artifacts) {
+    // StartUrl gatherer will give explanations for failures, but need to take manifest parsing
+    // warnings from the manifest itself (e.g. invalid `start_url`, so fell back to document URL).
+    const warnings = [];
+    const manifest = artifacts.Manifest;
+    if (manifest && manifest.value && manifest.value.start_url.warning) {
+      warnings.push('There was an error in the manifest and so the start_url fell back to the ' +
+          `document URL. Error message: '${manifest.value.start_url.warning}'.`);
+    }
+
+    const hasOfflineStartUrl = artifacts.StartUrl.statusCode === 200;
+
+    return {
+      rawValue: hasOfflineStartUrl,
+      explanation: artifacts.StartUrl.explanation,
+      warnings,
+    };
+  }
+}
+
+module.exports = OfflineStartUrl;
--- a/lighthouse-core/audits/offline-start-url.js
+++ b/lighthouse-core/audits/offline-start-url.js
@@ -16,9 +16,7 @@ class OfflineStartUrl extends Audit {
       id: 'offline-start-url',
       title: 'start_url responds with a 200 when offline',
       failureTitle: 'start_url does not respond with a 200 when offline',
-      description: 'If you\'re building a Progressive Web App, consider using a service worker ' +
-          'so that your app can work offline. ' +
-          '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).',
+      description: 'A service worker enables your web app to be reliable in unpredictable network conditions. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).',
       requiredArtifacts: ['Manifest', 'StartUrl'],
     };
   }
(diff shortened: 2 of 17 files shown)


brendankenny:
rebased off master

lighthouse-core/audits/multi-check-audit.js
```diff
@@ -66,7 +65,7 @@ class MultiCheckAudit extends Audit {
   /**
    * @param {LH.Artifacts} artifacts
    * @param {LH.Audit.Context} context
-   * @return {Promise<{failures: Array<string>, warnings?: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}>}
+   * @return {Promise<{failures: Array<string>, manifestValues?: LH.Artifacts.ManifestValues}>}
```
brendankenny: `webapp-install-banner` was the only one using `warnings` for start_url warnings, so no longer needed

lighthouse-core/config/default-config.js
```diff
@@ -334,10 +335,11 @@ const defaultConfig = {
         // Fast and Reliable
         {id: 'load-fast-enough-for-pwa', weight: 7, group: 'pwa-fast-reliable'},
         {id: 'works-offline', weight: 5, group: 'pwa-fast-reliable'},
+        {id: 'offline-start-url', weight: 1, group: 'pwa-fast-reliable'},
```
brendankenny: keeps the scores the same as before (`webapp-install-banner` split in two)

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -29,33 +30,27 @@ class StartUrl extends Gatherer {
 
         return this._attemptManifestFetch(passContext.driver, startUrlInfo.startUrl);
       }).catch(() => {
-        return {statusCode: -1, explanation: 'Unable to fetch start URL via service worker'};
+        return {statusCode: -1, explanation: 'Unable to fetch start URL via service worker.'};
```
brendankenny: periods added to look better in the report :) (this is currently the only audit using this gatherer)

lighthouse-core/gather/gatherers/start-url.js
```diff
@@ -29,33 +30,27 @@ class StartUrl extends Gatherer {
...
     }
 
-    // Even if the start URL had an error, the browser will still supply a fallback URL.
-    // Therefore, we only set the warning here and continue with the fetch.
-    if (manifest.value.start_url.warning) {
-      return {isReadFailure: true, reason: manifest.value.start_url.warning};
-    }
-
-    // @ts-ignore - TODO(bckenny): should actually be testing value above, not warning
+    // Even if the start URL had a parser warning, the browser will still supply a fallback URL.
```
brendankenny: this was wrong before. A manifest parsing warning doesn't mean there was a read failure, since the parser falls back to the document URL

lighthouse-core/lib/manifest-parser.js
```diff
@@ -127,10 +128,18 @@ function parseStartUrl(jsonInput, manifestUrl, documentUrl) {
...
+      raw,
+      value: documentUrl,
+    };
+  }
+  if (typeof raw !== 'string') {
+    return {
+      raw,
+      value: documentUrl,
+      warning: 'ERROR: expected a string.',
+    };
```
brendankenny: equivalent to before, but a lot easier to understand what happens to `undefined` and non strings

lighthouse-core/test/gather/gatherers/start-url-test.js
```diff
@@ -154,20 +154,7 @@ describe('Start-url gatherer', () => {
 
     return startUrlGatherer.afterPass(options, tracingData)
       .then(artifact => {
-        assert.equal(artifact.explanation, 'ERROR: start_url string empty');
-      });
-  });
-
-  it('returns an error when origin is not the same', () => {
```
brendankenny: not an error, a warning

lighthouse-core/audits/offline-start-url.js
```diff
@@ -0,0 +1,50 @@
...
+const Audit = require('./audit.js');
+
+class OfflineStartUrl extends Audit {
+  /**
+   * @return {LH.Audit.Meta}
+   */
+  static get meta() {
+    return {
+      id: 'offline-start-url',
+      title: 'start_url responds with a 200 when offline',
```
patrickhulce: should we be trying to `i18n` new strings in general? feels like we should

unrelated: `start_url` feels so weird in the title like that haha

lighthouse-core/audits/offline-start-url.js
```diff
@@ -0,0 +1,50 @@
...
+      failureTitle: 'start_url does not respond with a 200 when offline',
+      description: 'If you\'re building a Progressive Web App, consider using a service worker ' +
```
patrickhulce: feels a bit out of the ordinary for the PWA, but I like the qualification. Maybe we should prefix the other audit descriptions too?

lighthouse-core/test/gather/gatherers/start-url-test.js
patrickhulce: is it easy to convert to a test for the warning?

lighthouse-core/audits/offline-start-url.js
```diff
@@ -0,0 +1,50 @@
...
+   * @param {LH.Artifacts} artifacts
+   * @return {LH.Audit.Product}
+   */
+  static audit(artifacts) {
+    // StartUrl gatherer will give explanations for failures, but need to take manifest parsing
+    // warnings from the manifest itself (e.g. invalid `start_url`, so fell back to document URL).
+    const warnings = [];
+    const manifest = artifacts.Manifest;
+    if (manifest && manifest.value && manifest.value.start_url.warning) {
+      warnings.push('There was an error in the manifest and so the start_url fell back to the ' +
```
paulirish: What conditions lead to this?

Wording seems a bit odd to me.

> We couldn't read the start_url from the manifest. As a result, the start_url was assumed to be the document's URL.  ${warning}

lighthouse-core/audits/offline-start-url.js
paulirish: There's a text update on the `works-offline` audit. Something like "Current page... responds with a 200"

dunno if you want to do that here but we shouldn't drop it.

lighthouse-core/test/gather/gatherers/start-url-test.js
brendankenny: > is it easy to convert to a test for the warning?

the gatherer was set up to either fail and return an explanation or succeed and have no explanation, so making it not fail in that case means no more warning from the manifest parser. But both `webapp-install-banner` and `offline-start-url` both check the warning straight from the `Manifest` artifact (and both have tests)

lighthouse-core/audits/offline-start-url.js
brendankenny: > should we be trying to i18n new strings in general? feels like we should

I like doing them with everyone all at once so I can avoid responsibility for them, but should we do them as we go?

> unrelated: `start_url` feels so weird in the title like that haha

haha, I felt the same, but it felt worse to just have them be words. "Start URL responds with a 200..."

lighthouse-core/audits/offline-start-url.js
brendankenny: > how about

Sure.

It's also a little weird that both this and the `offline` check (checking the current page) have very similar descriptions, but I'm sure we can work on tweaking that.

lighthouse-core/audits/offline-start-url.js
brendankenny: > What conditions lead to this?

If you do something like `{start_url: 5}` as your manifest.

(more realistically, maybe you have a typo in the URL making it invalid, or it's a URL that's not same origin as the page being tested)

> Wording seems a bit odd to me.

done

lighthouse-core/audits/offline-start-url.js
brendankenny: > here's a text update on the `works-offline` audit. Something like "Current page... responds with a 200"
>
> dunno if you want to do that here but we shouldn't drop it.


The current proposed update is `"Current page responds with a 200 when offline"`. I modeled this after that...not sure what you mean by dropping it, though.

`Page at start_url responds with a 200 when offline`?

brendankenny:
> looks good to me outside i18n question! boy it is going to be so hard to remember all the various states of checking across LH versions

yeah, I don't envy anyone doing that :(

lighthouse-core/audits/offline-start-url.js
exterkamp: Maybe "Service Worker's start_url responds with a 200"?

lighthouse-core/audits/offline-start-url.js
```diff
@@ -0,0 +1,49 @@
...
+class OfflineStartUrl extends Audit {
+  /**
+   * @return {LH.Audit.Meta}
+   */
+  static get meta() {
+    return {
+      id: 'offline-start-url',
+      title: 'start_url responds with a 200 when offline',
+      failureTitle: 'start_url does not respond with a 200 when offline',
+      description: 'A service worker enables your web app to be reliable in unpredictable network conditions. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).',
```
exterkamp: nit: I feel like this maybe isn't specific enough?  Someone might have a service worker but have a bad start_url that would be flagged by this audit.

"A service worker _with a valid start_url_ enables your web app.."

lighthouse-core/test/results/sample_v2.json
```diff
@@ -1048,6 +1048,15 @@
         ]
       }
     },
+    "offline-start-url": {
+      "id": "offline-start-url",
+      "title": "start_url does not respond with a 200 when offline",
+      "description": "If you're building a Progressive Web App, consider using a service worker so that your app can work offline. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline).",
```
exterkamp: yarn update:sample-json

brendankenny: > yarn update:sample-json

done

lighthouse-core/audits/offline-start-url.js
brendankenny: > Someone might have a service worker but have a bad start_url that would be flagged by this audit.

yeah, I'm just not sure how to word it. It's not the service worker that has the start_url, it's the manifest, though the service worker needs to serve it. Maybe we can tweak

lighthouse-core/audits/offline-start-url.js
brendankenny: > Maybe "Service Worker's start_url responds with a 200"?

yeah, it's not the service worker's `start_url`, but the service worker needs to respond to the request for the `start_url`...

PR #6403, opened 2018-10-26
clients: Output setting is always an array 

exterkamp:
**Summary**
Removed string only uses of output, changed to array only.

**Related Issues/PRs**
fixes: #6309


clients/lightrider-entry.js
```diff
@@ -49,7 +49,7 @@ async function runLighthouseInLR(connection, url, flags, {lrDevice, categoryIDs,
     }
 
     // pre process the LHR for proto
-    if (flags.output === 'json' && typeof results.report === 'string') {
+    if (flags.output === ['json'] && typeof results.report === 'string') {
```
Hoten: unfortunately this doesn't work, the equality operator here is checking references not value.

```
> const array = ['json'];
< undefined
> ['json'] === array
< false
> ['json'] == array
< false
```

will need to be more explicit
```
flags.output && flags.output.length === 1 && flags.output[0] === 'json'
```

exterkamp: #saving me from myself

exterkamp:
You guys are re-reviewing too fast! :smiley:   

I have to make other changes on LR before this is ready to merge.  So I'll mark this as chillin' until I can coordinate with @paulirish about that.

PR #6404, opened 2018-10-26
test: use predictable numbers in golden LHR

paulirish:
No more excluding items from the golden LHR.

### DIFF
--- a/lighthouse-core/scripts/cleanup-LHR-for-diff.js
+++ b/lighthouse-core/scripts/cleanup-LHR-for-diff.js
@@ -26,6 +26,11 @@ function cleanAndFormatLHR(lhrString) {
   /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
 
+  // TODO: Include this, but resolve
+  // 1) the string|boolean story for proto
+  // 2)  Travis gets a absolute path during yarn diff:sample-json
+  delete lhr.configSettings.auditMode;
+
   // Set timing values, which change from run to run, to predicable values
   lhr.timing.total = 12345.6789;
   lhr.timing.entries.forEach(entry => {
--- a/lighthouse-core/scripts/cleanup-LHR-for-diff.js
+++ b/lighthouse-core/scripts/cleanup-LHR-for-diff.js
@@ -26,12 +26,12 @@ function cleanAndFormatLHR(lhrString) {
   /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
 
-  // TODO: Include this, but resolve
-  // 1) the string|boolean story for proto
-  // 2)  Travis gets a absolute path during yarn diff:sample-json
+  // TODO: Resolve the below so we don't need to delete it:
+  // 1) The string|boolean story for proto
+  // 2) Travis gets a absolute path during yarn diff:sample-json
   delete lhr.configSettings.auditMode;
 
-  // Set timing values, which change from run to run, to predicable values
+  // Set timing values, which change from run to run, to predictable values
   lhr.timing.total = 12345.6789;
   lhr.timing.entries.forEach(entry => {
     // @ts-ignore - write to readonly property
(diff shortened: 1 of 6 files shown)


Hoten:
Can we remain `sample_v2` to `golden`? :)

paulirish:
> Can we rename sample_v2 to golden? :)

been wanting to for a while.

I'll do it after this PR

brendankenny:
what, it's not obvious what `sample_v2` refers to?? :)

brendankenny:
for test failures, see #6405 :)

lighthouse-core/scripts/cleanup-LHR-for-diff.js
```diff
@@ -23,9 +23,21 @@ writeFileSync(filename, cleanAndFormatLHR(data), 'utf8');
  * @return {string}
  */
 function cleanAndFormatLHR(lhrString) {
+  /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
-  delete lhr.configSettings.auditMode;
```
brendankenny: was there an issue with having this previously? Was it only because we were generating it more manually?

lighthouse-core/scripts/cleanup-LHR-for-diff.js
```diff
@@ -23,9 +23,21 @@ writeFileSync(filename, cleanAndFormatLHR(data), 'utf8');
...
-  delete lhr.timing;
+
+  // Set to predictable and typical value
+  lhr.configSettings.auditMode = false;
+
+  // Set timing values, which change from run-to-run, to predicable values
+  lhr.timing.total = 12345.6789;
+  lhr.timing.entries.forEach((entry, i) => {
+    // @ts-ignore
+    entry.duration = 100;
```
brendankenny: why are these ignored? `readonly`? Maybe we don't want them to descend from `PerformanceEntry` :)

Should put a reason why it's being ignored. e.g. `// @ts-ignore - write to readonly property` or whatever

lighthouse-core/scripts/cleanup-LHR-for-diff.js
```diff
@@ -23,9 +23,21 @@ writeFileSync(filename, cleanAndFormatLHR(data), 'utf8');
...
 function cleanAndFormatLHR(lhrString) {
+  /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
-  delete lhr.configSettings.auditMode;
-  delete lhr.timing;
+
+  // Set to predictable and typical value
+  lhr.configSettings.auditMode = false;
+
+  // Set timing values, which change from run-to-run, to predicable values
```
brendankenny: no `-`

lighthouse-core/scripts/cleanup-LHR-for-diff.js
```diff
@@ -23,9 +23,21 @@ writeFileSync(filename, cleanAndFormatLHR(data), 'utf8');
...
+  lhr.timing.total = 12345.6789;
+  lhr.timing.entries.forEach((entry, i) => {
+    // @ts-ignore
+    entry.duration = 100;
+    // @ts-ignore
+    entry.startTime = 100 * i + i; // 1ms gap between them
```
brendankenny: if we add new entries this is going to cause some noise. What about based on length of the name or some kind of hash of the name or something?

lighthouse-core/scripts/cleanup-LHR-for-diff.js
patrickhulce: if we're jumping on the comments there's also a `t` missing from `predictable` :)

lighthouse-core/scripts/cleanup-LHR-for-diff.js
paulirish: its just this:

```js
auditMode": "./lighthouse-core/test/results/artifacts",
```

hopefully it isn't an absolute path in some cases?

lighthouse-core/scripts/cleanup-LHR-for-diff.js
paulirish: k. behold our new predictable timing trace! 

image

Issue #6405, opened 2018-10-26
Make updating the golden JSON not painful


brendankenny:
This has gotten pretty bad now :)

For many changes, you have to have `protobuf` compiled and installed and run `yarn update:sample-json` and `yarn update:proto`. For any changes to the default config, you also have to run `jest lighthouse-cli/test/cli/index-test.js -u`.

Some ideas:
1) Add `update:proto` to `update:sample-json`. Both json files have to be updated at the same time anyways
2) Sub idea: rename `update:proto` and `build-proto`. It's not updating or building the proto, it's updating the roundtripped sample json :)
3) Remove `protobuf` requirement. Most changes to Lighthouse are going to need to touch the sample json, and asking folks to install this is going to eliminate a lot of potential contributors.
    - @exterkamp mentioned we could have a web service that takes a `sample_v2.json` and returns the proto roundtrip result. We could still verify the valid result in CI (which we really need to do anyways).
    - we could also investigate wasm compiles of protobuf and check the result in. https://github.com/kwonoj/protobuf-wasm has been continually updated to track the latest protobuf release
4) make `sample_v2_round_trip.json` ephemeral, created only to test that it round trips correctly (I believe this was/is the plan for it). We don't generally want to have two identical files checked in, just makes PRs longer and makes it easier to miss stuff.

paulirish:
Great thoughts here.

> Add update:proto to update:sample-json

yes. very much

> Sub idea: rename update:proto and build-proto.

+1 to renaming build-proto to `proto-roundtrip`. 

> make sample_v2_round_trip.json ephemeral

I have a branch where instead of checking in the full pyproto_roundtrip.json we check in the diff of (the pyproto_roundtrip vs samplev2json). It'll be an empty file when the preprocessor is gone.

patrickhulce:
> make sample_v2_round_trip.json ephemeral, created only to test that it round trips correctly

+10 :) can't wait to see the branch @paulirish

exterkamp:
Another idea, we can check-in the generated `_pb2.py` generated proto python file.  Then it wouldn't require installing proto to update the round trip.  Just having python.

brendankenny:
> we can check-in the generated _pb2.py generated proto python file

is that sufficient? We should definitely do that, then! :)

brendankenny:
> is that sufficient?

[...]

PR #6406, opened 2018-10-26
misc: simplify proto npm scripts

brendankenny:
adds proto roundtrip update to `yarn update:sample-json` and updates `build-proto` to `build-proto-roundtrip`

(this is (1) and (2) of #6405)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,14 +61,13 @@
     "i18n:checks": "./lighthouse-core/scripts/i18n/assert-strings-collected.sh",
     "i18n:collect-strings": "node lighthouse-core/scripts/i18n/collect-strings.js",
     "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
-    "update:sample-json": "yarn i18n:collect-strings && node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --throttling-method=devtools --output=json --output-path=./lighthouse-core/test/results/sample_v2.json && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,14 +61,13 @@
     "i18n:checks": "./lighthouse-core/scripts/i18n/assert-strings-collected.sh",
     "i18n:collect-strings": "node lighthouse-core/scripts/i18n/collect-strings.js",
     "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
-    "update:sample-json": "yarn i18n:collect-strings && node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --throttling-method=devtools --output=json --output-path=./lighthouse-core/test/results/sample_v2.json && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing",
+    "update:sample-json": "yarn i18n:collect-strings && node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --throttling-method=devtools --output=json --output-path=./lighthouse-core/test/results/sample_v2.json && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing && yarn compile-proto && yarn build-proto-roundtrip",
```
patrickhulce: that's quite the line these days 😆 

just thinking out loud below 💭 

I guess the contributor thinking here is that if they already had to run `update:sample-json` then they would've failed the build without `compile-proto` so forcing protobuf in the path of this command isn't really a barrier. If they just needed string collection that would still run before all this junk failed anyhow. 👍

paulirish: ^ yes.  any meaningful changes require running both. 

--------

separately, can we bikeshed the script name?? :) how about `update:goldens`?

PR #6410, opened 2018-10-26
core: if verbose, add marks for each protocol command

### LINKED ISSUE #6408: core: if verbose, add marks for each protocol command
would be super

Hoten:
image

https://gist.github.com/Hoten/b5a96f138df2649239499d315b99bf41

Fixes #6408

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -291,11 +291,25 @@ class Driver {
         err.message += ` Method: ${method}`;
         reject(err);
       }), timeout);
+
+      const status = {
+        msg: `Devtools Protocol Command ${method}`,
+        id: `lh:driver:sendCommand:${method}`,
+      };
+      if (log.isVerbose()) {
+        log.time(status, 'verbose');
+      }
       try {
         const result = await this._connection.sendCommand(method, ...params);
+        if (log.isVerbose()) {
+          log.timeEnd(status);
+        }
         clearTimeout(asyncTimeout);
         resolve(result);
       } catch (err) {
+        if (log.isVerbose()) {
+          log.timeEnd(status);
+        }
         clearTimeout(asyncTimeout);
         reject(err);
       }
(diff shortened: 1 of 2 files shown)


Hoten:
`Two measures overlap!` warnings, I guess we can just ignore those.

lighthouse-core/gather/driver.js
```diff
@@ -291,11 +291,25 @@ class Driver {
...
+      const status = {
+        msg: `Devtools Protocol Command ${method}`,
+        id: `lh:driver:sendCommand:${method}`,
+      };
+      if (log.isVerbose()) {
+        log.time(status, 'verbose');
+      }
       try {
         const result = await this._connection.sendCommand(method, ...params);
+        if (log.isVerbose()) {
```
patrickhulce: can we stick this in a `finally` instead?

Hoten: It wouldn't be the same functionality - in `finally`, the resolve/reject cb would happen first.

patrickhulce: Oh I'm sensing a learning opportunity for me! :D 

What situations raise an error or impact the timings in any meaningful way? Until today I thought the callbacks would just be run in a microtask afterwards and resolve/reject is basically a super quick bitflip with a bit of state saved

Hoten: Ok, clearly this just registers a value/promise to resolve to, doesn't actually execute it or anything. Slight confusion :)

Hoten:
Leveraging #6425 would be interesting. Would just need to allow for dynamically setting the mark id, given the args. Would be important for distinguishing between each command.

Then all the marking code would be outside the method itself, and could look more like
```js
if (log.isVerbose()) {
    log.autoMark(Driver, {
        sendCommand: {
            msg: (method) => `Devtools Protocol Command ${method}`,
            id: (method) => `lh:driver:sendCommand:${method}`,
        }
    })
}
```

(can iterate on that, but the important part is that all the tracing code is separated from the implementation)

Hoten:
@brendankenny thoughts on above? If it sounds good I will abandon this PR and implement this in #6425 .

brendankenny:
> If it sounds good I will abandon this PR and implement this in #6425 .

Let's close in favor of that.

PR #6411, opened 2018-10-26
core(jsdom): upgrade jsdom to support Element.prototype.closest

wardpeet:
**Summary**

Upgrade jsdom

**Related Issues/PRs**


yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0-beta.49"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.49.tgz#becd805482734440c9d137e46d77340e64d7f51b"
-  integrity sha1-vs2AVIJzREDJ0TfkbXc0DmTX9Rs=
```
brendankenny: gotta update yarn so these aren't deleted :)

lighthouse-viewer/test/test-helpers.js
```diff
@@ -14,8 +14,9 @@ const jsdom = require('jsdom');
 const PAGE = fs.readFileSync(path.join(__dirname, '../app/index.html'), 'utf8');
 
 function setupJsDomGlobals() {
-  global.document = jsdom.jsdom(PAGE);
-  global.window = global.document.defaultView;
+  const {window} = new jsdom.JSDOM(PAGE);
```
brendankenny: this is throwing an error in the test because viewer has its own version in its `package.json`. I'd vote for removing it there so it can just use the root directory's version

yarn.lock
wardpeet: fixed

wardpeet:
ah i'm on yarn 1.7.0

wardpeet:
@brendankenny this one should be ok now! 👌

PR #6413, opened 2018-10-27
Issue 6407 increase page navigate timeout

Hoten:
i think 60s will be better


lighthouse-core/gather/driver.js
```diff
@@ -840,7 +840,7 @@ class Driver {
     // happen _after_ onload: https://crbug.com/768961
     this.sendCommand('Page.enable');
     this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
-    this.setNextProtocolTimeout(30 * 1000);
+    this.setNextProtocolTimeout(60 * 1000); // see https://github.com/GoogleChrome/lighthouse/pull/6413
```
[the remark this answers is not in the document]
paulirish: sg. i turned it all the way to 90s on lightrider.

might as well use that here, too.

paulirish: I'd give a lil more context in the comment.. maybe

> // We don't need a strict timeout for Page.navigate. See #6413

Hoten:
OK, I've disabled the timeout completely.

lighthouse-core/gather/driver.js
```diff
@@ -286,18 +286,19 @@ class Driver {
       }
     }
     return new Promise(async (resolve, reject) => {
-      const asyncTimeout = setTimeout((_ => {
+      const asyncTimeout = timeout > 0 ? setTimeout((_ => {
```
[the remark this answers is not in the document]
Hoten: I would have gone that route first, but it would need to be `let`, and my personal preferences led me to choose the ternary before reaching for `let`. wdyt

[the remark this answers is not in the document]
brendankenny: whoops, yeah, `let`. And what's with that explicit `undefined`. Going to edit my comment because my brain was apparently off while writing that :)

The main issue is that you have to look down four lines to see the other half of the conditional, and the syntax makes it harder to parse than normal control flow mechanisms (like an if block). Generally long ternaries just aren't worth it, but at the very least, the null branch should be first

lighthouse-core/gather/driver.js
```diff
@@ -286,18 +287,23 @@ class Driver {
       }
     }
     return new Promise(async (resolve, reject) => {
-      const asyncTimeout = setTimeout((_ => {
-        const err = new LHError(LHError.errors.PROTOCOL_TIMEOUT);
-        err.message += ` Method: ${method}`;
-        reject(err);
-      }), timeout);
+      let asyncTimeout;
+      if (timeout > 0) {
```
paulirish: Taking another look at this.
I'm a bit hesitant to absorb this complexity.

What about adding a driver.innerSendCommand() which does the essence of sendCommand
And for the Page.Navigate command we use that.

Hoten: Done

paulirish:
@brendankenny @patrickhulce @exterkamp 	lgty?

lighthouse-core/gather/driver.js
```diff
@@ -278,30 +279,41 @@ class Driver {
...
   }
 
+  /**
+   * Call protocol methods.
+   * @template {keyof LH.CrdpCommands} C
+   * @param {C} method
+   * @param {LH.CrdpCommands[C]['paramsType']} params
+   * @return {Promise<LH.CrdpCommands[C]['returnType']>}
+   */
+  innerSendCommand(method, ...params) {
```
brendankenny: Maybe add a leading `_` and mark it with `@private` (someday tsc will support it) to make it super clear it's intended for internal use

lighthouse-core/test/gather/driver-test.js
```diff
@@ -240,6 +240,24 @@ describe('Browser Driver', () => {
...
+
+    driver.setNextProtocolTimeout(25);
+    await driver.sendCommand('Page.enable');
+
+    driver.setNextProtocolTimeout(5);
+    try {
+      await driver.sendCommand('Page.disable');
+      assert.fail('expected driver.sendCommand to timeout');
+    } catch (err) {
+      // :)
```
patrickhulce: well here we want to assert that we didn't fail with the `assert.fail` you added :)

let's make sure that the message matches roughly what we want it to `.includes('PROTOCOL_TIMEOUT')` maybe?

Hoten: looks correct to me - if this command times out, the next statement in the `try` block should never be reached.

added check for err.code

lighthouse-core/test/gather/driver-test.js
```diff
@@ -240,6 +240,24 @@ describe('Browser Driver', () => {
     });
   });
 
+  it('.sendCommand timesout when commands take too long', async () => {
+    const driver = new Driver(connection);
```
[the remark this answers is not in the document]
Hoten: nice, done

lighthouse-core/test/gather/driver-test.js
patrickhulce: > looks correct to me - if this command times out, the next statement in the try block should never be reached.

If the command times out, it would've thrown. If it doesn't time out as expected, it would've also thrown on the following line with the fail. Both paths led to an empty catch block. New error code assertion looks perfect though!! 👌

Hoten: Ah, of course. I convinced myself that assert.fail didn't throw an error.. I was applying a pattern that works for other testing frameworks. But totally doesn't work here :p

Issue #6202, opened 2018-10-09
☂️ 👣 Breaking changes for next major


paulirish:
A la #4333, the items we are thinking of changing in the next major rls

 - [x] PWA audit reorg #6395
 - [x] displayValue should always be string #6200 
 - [x] proto changes `not_applicable` #6201
 - [ ] benchmarkIndex => CPU throttling #6162
 - [x] Lr_desktop is too generous #6836
 - [x] Nexus5x viewport is slightly different #6615
 - [x] remove children from crc #6211 
 - [x] Rename `typings/` to `types/` #6617
 - [x] Rename `core/gather/computed/` to `core/computed/` #6618

Punt?
 - [x] remove `rawValue` #6199

brendankenny:
this is old now. See #7752 :)

PR #6424, opened 2018-10-29
misc(proto): reduce proto configSettings to minimum needed

exterkamp:
**Summary**
Remove many of the configSettings from the proto definition.

**Related Issues/PRs**
moves #6202 along
fixes: #6309 (by deleting it from the proto)
closes: #6403

### DIFF
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -24,7 +24,7 @@ function processForProto(result) {
 
   // Clean up the configSettings
   // Note: This is not strictly required for conversion if protobuf parsing is set to
-  //       'ignore unknown fields' in the language of conversion.
+  // 'ignore unknown fields' in the language of conversion.
   if (reportJson.configSettings) {
     // The settings that are in both proto and LHR
     const {emulatedFormFactor, locale, onlyCategories} = reportJson.configSettings;
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -23,13 +23,22 @@ function processForProto(result) {
   const reportJson = JSON.parse(result);
 
   // Clean up the configSettings
+  // Note: This is not strictly required for conversion if protobuf parsing is set to
+  //       'ignore unknown fields' in the language of conversion.
   if (reportJson.configSettings) {
-    // make sure the 'output' field is an array
-    if (reportJson.configSettings.output) {
-      if (!Array.isArray(reportJson.configSettings.output)) {
-        reportJson.configSettings.output = [reportJson.configSettings.output];
-      }
-    }
+    // Filter out unwanted fields
+    const valid = ['emulatedFormFactor', 'locale', 'onlyCategories'];
+
+    const minifiedSettings = Object.keys(reportJson.configSettings)
+      .filter(key => valid.includes(key))
+      .reduce((obj, key) => {
+        // @ts-ignore
+        obj[key] = reportJson.configSettings[key];
+        return obj;
+      }, {});
+
+    // @ts-ignore
+    reportJson.configSettings = minifiedSettings;
   }
 
   // Remove runtimeError if it is NO_ERROR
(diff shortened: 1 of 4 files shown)


lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -23,13 +23,22 @@ function processForProto(result) {
...
-        reportJson.configSettings.output = [reportJson.configSettings.output];
-      }
-    }
+    // Filter out unwanted fields
+    const valid = ['emulatedFormFactor', 'locale', 'onlyCategories'];
+
+    const minifiedSettings = Object.keys(reportJson.configSettings)
+      .filter(key => valid.includes(key))
+      .reduce((obj, key) => {
+        // @ts-ignore
```
exterkamp: @brendankenny is there a better way to do this so that ts doesn't mind?

proto/lighthouse-result.proto
```diff
@@ -185,36 +127,14 @@ message LighthouseResult {
     }
 
     // The form factor used in the audit
-    EmulatedFormFactor emulated_form_factor = 8;
+    EmulatedFormFactor emulated_form_factor = 1;
```
exterkamp: @paulirish should we renumber at 1, or reserve 1-16 and start at 17?  IMO we can go ahead and renumber since no one is using this yet.

lighthouse-core/lib/proto-preprocessor.js
Hoten: trying marking `valid` as `@typeof {keyof LH.Config}`

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -23,13 +23,22 @@ function processForProto(result) {
...
+  // Note: This is not strictly required for conversion if protobuf parsing is set to
+  //       'ignore unknown fields' in the language of conversion.
   if (reportJson.configSettings) {
-    // make sure the 'output' field is an array
-    if (reportJson.configSettings.output) {
-      if (!Array.isArray(reportJson.configSettings.output)) {
-        reportJson.configSettings.output = [reportJson.configSettings.output];
-      }
-    }
+    // Filter out unwanted fields
```
patrickhulce: ~~wait I thought these were the only ones we were keeping?~~ oh I see that is what's happening :)

maybe rename the variable so the comment isn't necessary (i.e. `onlyFieldsToKeep` or something) or flesh it out a bit so it's clear `// We'll discard all fields except these because LR doesn't need them` or w/e?

lighthouse-core/lib/proto-preprocessor.js
patrickhulce: If I channeled my inner @paulirish, the better way would be to eliminate all non-numeric `reduce`s from the world.

I'm normally a reduce fan, but this one did confuse me for a half second and the below is actually fewer lines with no ts-ignore, for your consideration :)

```js
/** @type {Partial<LH.Config.Settings> */
const minifiedSettings = {}
for (const [key, value] of Object.entries(reportJson.configSettings) {
  if (keysToKeep.includes(key)) minifiedSettings[key] = value;
}

Comment (exterkamp) [reply lighthouse-core/lib/proto-preprocessor.js]: Consulted @brendankenny for a solution that minimizes the ts ignores and reduces.

I :heart: reduce though.

Comment (brendankenny) [reply lighthouse-core/lib/proto-preprocessor.js]: yeah, the issue is that currently `Object.entries` won't be well typed unless the object passed in has all the same types as its values. Since `configSettings` has a bunch of different types for its values, `[key, value]` ends up as `[string, any]`, and since `minifiedSettings` doesn't have an index, it ends up needing a `ts-ignore` for a different reason.

This solution is pretty good as an alternative, it just won't scale that well if we need more settings passed along :)

Comment (patrickhulce) [reply lighthouse-core/lib/proto-preprocessor.js]: ah my b

yeah if we don't think this list will grow. solution you ended up with is obvs best :D

File: proto/lighthouse-result.proto (hunk 2 above)
Comment (paulirish) [reply proto/lighthouse-result.proto]: We'll need to apply two annotations as described in go/proto-best-practices-checkers but I think this is the right move for now.

LG on the renumber.

File: lighthouse-core/lib/proto-preprocessor.js [hunk 4]
```diff
@@ -23,13 +23,14 @@ function processForProto(result) {
   const reportJson = JSON.parse(result);
 
   // Clean up the configSettings
+  // Note: This is not strictly required for conversion if protobuf parsing is set to
+  //       'ignore unknown fields' in the language of conversion.
```
paulirish: extra spaces?

Issue #6428, opened 2018-10-30
move our windows CI testing to travis


paulirish:
(leaving appveyor)

brendankenny:
Based on discussion in (and closing of) #6429, going to close this as well. We'll move on to trying out the next Windows CI flavor of the month soon, I'm sure :)

PR #6429, opened 2018-10-30
tests: update travis to support windows

wardpeet:
**Summary**
Moving to travis

**Related Issues/PRs**
#6428


paulirish:
@evenstensberg from what i'm seeing here, webpack isn't yet using windows on travis.

you probably would have discovered that secure variables aren't supported and break the build. :)

wardpeet:
@evenstensberg I was going to make a diff project to get chromium running on windows before I move on, on this PR thanks for the references!

wardpeet:
> Thought secure variables are supported, we have used them with GH and NPM permissions on webpack-cli 2->3.0.1 ( https://docs.travis-ci.com/user/encryption-keys/), but maybe that's different on windows too..

Windows on Travis just crashes when using encrypted vars. Lots of bugs actually.

evenstensberg:
I see, could help out but I don't think I'll get such access 😆 Good luck Pete!

wardpeet:
Yes, I'm making a separate repo to test this

wardpeet:
@paulirish @brendankenny So there are a few problems that I encountered
https://travis-ci.org/wardpeet/lighthouse/builds/462406549

- [ ] [Stable (chrome 70) is currently bugged on windows server core 2016](https://bugs.chromium.org/p/chromium/issues/detail?id=873097)
- [ ] [Travis hangs when using encrypted secrets](https://travis-ci.community/t/windows-instances-hanging-before-install/250/14)
- [ ] Chrome needs to run in headless because windows server core has no GUI support. This gives some troubles on our smoketests

brendankenny:
@wardpeet you seemed pretty down on this ever working out on the Travis side, so let's just close for now and start investigating other options for Windows.

PR #6440, opened 2018-11-01
misc(timing): move timing-trace to async events

paulirish:
Motivated by #6410

I was avoiding async events for .. mysterious reasons before. But really these will be overlapping, so it makes more sense to be async.

Overlapping happens in promise.all()s often:
image

Here's how a few timing traces look now:
image
image


lighthouse-core/lib/timing-trace-saver.js
```diff
@@ -5,70 +5,55 @@
...
-
+  entries.forEach(entry => {
     /** @type {LH.TraceEvent} */
-    const traceEvent = {
-      name: entry.name,
-      cat: entry.entryType,
+    const startEvt = {
+      // 1) Remove 'lh:' for readability
+      // 2) Colons in user_timing names get special handling in traceviewer we don't want. https://goo.gl/m23Vz7
+      //    Replace with a 'Modifier letter colon' ;)
```
patrickhulce: 😆

paulirish:
thoughts?


-----------

I've added support for DevTools Perf panel (via adding a TracingStartedInBrowser evt):
image



Though note: -G/-A traces are shown in two diff tracks in chrome:tracing but that's not supported in DevTools:
image

lighthouse-core/test/lib/timing-trace-saver-test.js
```diff
@@ -37,92 +36,11 @@ const mockEntries = [{
...
-      entryType: 'measure',
-    },
-    ];
-
-    generateTraceEvents(overlappingEntries);
-    expect(consoleError).toHaveBeenCalled();
-    expect(consoleError.mock.calls[0][0]).toContain('measures overlap');
+  it('generates a pair of trace events', () => {
+    const events = generateTraceEvents([mockEntries[0]]);
+    expect(events.slice(0, 1)).toMatchSnapshot();
```
patrickhulce: seems a bit odd that we call the test `generates a pair` but then only look at 1 event, just do 'em both?

lighthouse-core/test/lib/timing-trace-saver-test.js
```diff
@@ -37,92 +36,11 @@ const mockEntries = [{
...
   });
 });
```
exterkamp: Add a test to hit a zero duration event?

```javascript
  it('Handles zero duration events', () => {
    const overlappingEntries = [
      {
        startTime: 10,
        name: 'zeroLength',
        duration: 0,
        entryType: 'measure',
      },
    ];
    const expected = {
      name: 'zeroLength',
      cat: 'measure',
      ts: 10000,
      dur: 0,
      args: {},
      pid: 0,
      tid: 50,
      ph: 'n',
      s: 't',
      id: '0x0',
    };
    const event = generateTraceEvents(overlappingEntries);
    assert.deepStrictEqual(event[0], expected);
  });
```

lighthouse-core/test/lib/timing-trace-saver-test.js
paulirish: lol thx

lighthouse-core/test/lib/timing-trace-saver-test.js
paulirish: good call. actually we dont need the `ph:'n'` now that we're on async events. 
problem solved. :)

lighthouse-core/lib/timing-trace-saver.js
```diff
@@ -5,70 +5,48 @@
...
-    if (isOverlap) {
-      // eslint-disable-next-line no-console
-      console.error(`Two measures overlap! ${prevEntry.name} & ${entry.name}`);
-    }
-  }
-}
-
 /**
  * Generates a chromium trace file from user timing measures
  * Adapted from https://github.com/tdresser/performance-observer-tracing
```
brendankenny: maybe add a note about how `threadId` is used?

lighthouse-core/lib/timing-trace-saver.js
```diff
@@ -5,70 +5,48 @@
...
-
     /** @type {LH.TraceEvent} */
-    const traceEvent = {
-      name: entry.name,
-      cat: entry.entryType,
+    const startEvt = {
+      // 1) Remove 'lh:' for readability
+      // 2) Colons in user_timing names get special handling in traceviewer we don't want. https://goo.gl/m23Vz7
+      //    Replace with a 'Modifier letter colon' ;)
+      name: entry.name.replace('lh:', '').replace(/:/g, '꞉'),
```
brendankenny: 😆 

maybe do an explicit `.replace(/:/g, '\ua789'),` just to make it doubly clear that the character is different?

lighthouse-core/lib/timing-trace-saver.js
```diff
@@ -5,70 +5,48 @@
...
+      cat: 'blink.user_timing',
```
brendankenny: it feels like doing this should be possible without faking out a browser trace but w/e traceviewer :)

lighthouse-core/scripts/generate-timing-trace.js
```diff
@@ -50,7 +50,7 @@ function saveTraceFromCLI() {
   // eslint-disable-next-line no-console
   console.log(`
   > Timing trace file saved to: ${traceFilePath}
-  > Open this file in chrome://tracing
+  > Open this file in DevTools perf panel (or if it's a -G/A run, use chrome://tracing)
```
brendankenny: does `-G/A run` refer to a "`-G` or an `-A` run", or a `-GA` run or a regular run (that did both gathering and auditing)?

brendankenny:
(with nits :)

lighthouse-core/lib/timing-trace-saver.js
paulirish: traceviewer wont group these but give each measure its own track otherwise
and devtools needs it to view them.

Issue #6442, opened 2018-11-01
☂️ OYB Next Steps


patrickhulce:
❔ ❔ ❔  _[Curious what OYB is?](https://github.com/GoogleChrome/lighthouse/issues/2146)_ ❔ ❔ ❔ 

**Summary**
I did a quick repurposing of DZL to analyze where our time is going in aggregate. This is the top 30 average timings across ~400 runs today

```
timing-gather-loadblank                 1103.33
timing-gather-afterpass-accessibility    827.09
timing-gather-afterpass-optimizedimages  825.39
timing-gather-afterpass-fontsize         797.62
timing-init-config                       779.76
timing-gather-gettrace                   697.89
timing-audit-speed-index                 578.59
timing-gather-getbenchmarkindex          510.86
timing-audit-screenshot-thumbnails       500.83
timing-config-requireaudits              472.83
timing-gather-afterpass-starturl         248.76
timing-audit-unminified-javascript       243.60
timing-audit-metrics                     241.27
timing-gather-loadpage-offlinepass       234.72
timing-gather-loadpage-redirectpass      230.89
timing-config-requiregatherers           214.00
timing-gather-afterpass-scripts          170.01
timing-gather-afterpass-cssusage         165.33
timing-gather-afterpass-metarobots       146.49
timing-gather-afterpass-offline          122.49
timing-gather-afterpass-crawlablelinks   116.67
timing-audit-estimated-input-latency     116.23
timing-audit-first-contentful-paint      112.46
timing-audit-load-fast-enough-for-pwa    101.72
timing-audit-is-on-https                  90.27
timing-audit-unminified-css               86.65
timing-audit-offscreen-images             81.61
```

Looking closer at the top 10...

1. **Load blank.** #6369 helped, but it's still a bigger impact to our runtime on average than any of our audits 😱, we might want to experiment with shorter blank times until we can revive data URIs (#4310)
1. **axe.** Not much we can do here immediately, I've already filed a couple issues on some of the worst offenders. Might indicate to us it's worth investing a bit of resources to help improve their perf story?
2. **OptimizedImages.** We've already done the biggest thing here by moving it to Chrome side. Only obvious work I see here would reduce our functionality or move more into estimation/heuristics.
3. **FontSize.** @Hoten is already on the case!!
4. **Init config.** This is surprising. I'm on it!
5. **Get trace.** Already aware of it and on it #3596
[...]

PR #6445, opened 2018-11-01
core(config): faster category validation

patrickhulce:
**Summary**
On my machine 90% of the non-requireing time spent in config:init was just validating the categories. Simple conversion of this `.find` to a map eliminated ~60ms.

**Related Issues/PRs**
#6442

### DIFF
--- a/lighthouse-core/config/config.js
+++ b/lighthouse-core/config/config.js
@@ -64,13 +64,18 @@ function validateCategories(categories, audits, groups) {
     return;
   }
 
+  const auditsKeyedById = new Map((audits || []).map(a =>
+    /** @type {[string, LH.Config.AuditDefn]} */
+    ([a.implementation.meta.id, a]))
+  );
+
   Object.keys(categories).forEach(categoryId => {
     categories[categoryId].auditRefs.forEach((auditRef, index) => {
       if (!auditRef.id) {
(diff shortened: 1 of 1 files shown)


lighthouse-core/config/config.js
```diff
@@ -64,13 +64,18 @@ function validateCategories(categories, audits, groups) {
     return;
   }
 
+  const auditsKeyedById = new Map((audits || []).map(audit =>
+    /** @type {[string, LH.Config.AuditDefn]} */
+    ([audit.implementation.meta.id, audit]))
+  );
```
Hoten: the parens here are confusing - it led me to think `Map` had a `.map` function.

is there an alternative way to balance these parentheses?

patrickhulce: Oh haha, whoops yeah I'll move the second one down :)

paulirish:
60ms! Hot damn.

Nice one.

patrickhulce:
> 60ms! Hot damn.

I can't tell if that's sarcasm or not 😆

paulirish:
>> 60ms! Hot damn.

> I can't tell if that's sarcasm or not 😆

It's not. I think this points to the wins of profiling-guided development. Pretty awesome.

PR #6446, opened 2018-11-01
core(driver): wait for Page.frameNavigated for about:blank

patrickhulce:
**Summary**
I'll want to put this through the DZL wringer and maybe someone sanity check this with the DevTools folks, but it seems like we should be able to just wait for the frameNavigated event instead of 300ms all the time.

On my machine it drives down about:blank times from...

300ms -> 200ms
300ms -> 16ms
300ms -> 14ms

If this gets cursory 👍 I'll remove references to blankDuration elsewhere, or we can keep it but just update the minimum to be 0

**Related Issues/PRs**
#6442

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -497,6 +497,16 @@ class Driver {
   }
 
   /**
+   * Returns a promise that resolve when a frame has been navigated.
+   * Used for detecting that our about:blank reset has been completed.
+   */
+  _waitForFrameNavigated() {
+    return new Promise(resolve => {
+      this.once('Page.frameNavigated', resolve);
+    });
+  }
+
+  /**
    * Returns a promise that resolves when the network has been idle (after DCL) for
    * `networkQuietThresholdMs` ms and a method to cancel internal network listeners/timeout.
    * @param {number} networkQuietThresholdMs
@@ -824,10 +834,11 @@ class Driver {
    * possible workaround.
    * Resolves on the url of the loaded page, taking into account any redirects.
    * @param {string} url
-   * @param {{waitForLoad?: boolean, passContext?: LH.Gatherer.PassContext}} options
+   * @param {{waitForLoad?: boolean, waitForNavigated?: boolean, passContext?: LH.Gatherer.PassContext}} options
    * @return {Promise<string>}
    */
   async gotoURL(url, options = {}) {
+    const waitForNavigated = options.waitForNavigated || false;
     const waitForLoad = options.waitForLoad || false;
     const passContext = /** @type {Partial<LH.Gatherer.PassContext>} */ (options.passContext || {});
     const disableJS = passContext.disableJavaScript || false;
@@ -842,6 +853,10 @@ class Driver {
     this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
     this.setNextProtocolTimeout(30 * 1000);
     this.sendCommand('Page.navigate', {url});
+
+    if (waitForNavigated) {
+      await this._waitForFrameNavigated();
+    }
 
     if (waitForLoad) {
       const passConfig = /** @type {Partial<LH.Config.Pass>} */ (passContext.passConfig || {});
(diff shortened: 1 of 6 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -68,18 +68,15 @@ class GatherRunner {
...
    * @param {Driver} driver
    * @param {string=} url
-   * @param {number=} duration
    * @return {Promise<void>}
    */
   static async loadBlank(
       driver,
-      url = constants.defaultPassConfig.blankPage,
-      duration = constants.defaultPassConfig.blankDuration
+      url = constants.defaultPassConfig.blankPage
```
brendankenny: can this go back to being one line now :)

lighthouse-core/gather/driver.js
```diff
@@ -843,6 +854,10 @@ class Driver {
     this.setNextProtocolTimeout(30 * 1000);
     this.sendCommand('Page.navigate', {url});
 
+    if (waitForNavigated) {
+      await this._waitForFrameNavigated();
```
brendankenny: any concerns about racing here? i.e. should we get the promise before `Page.navigate` and then `await` it here?

brendankenny: also, any concerns about delaying `_waitForFullyLoaded` until after `Page.frameNavigated` fires?

lighthouse-core/gather/gather-runner.js
```diff
@@ -68,18 +68,15 @@ class GatherRunner {
    * never fired on it.
    * @param {Driver} driver
    * @param {string=} url
-   * @param {number=} duration
    * @return {Promise<void>}
    */
   static async loadBlank(
       driver,
-      url = constants.defaultPassConfig.blankPage,
-      duration = constants.defaultPassConfig.blankDuration
```
brendankenny: I forget when we've actually used a different `blankDuration`. Was it just for tests?

lighthouse-core/gather/gather-runner.js
patrickhulce: yep! 🎉

lighthouse-core/gather/gather-runner.js
patrickhulce: tests and once upon a time WPT used it for video recording, but for past year or so he's just run LH straight outta the box

https://github.com/WPO-Foundation/wptagent/blob/14d8c49e3139625fd31fa12efd6a3ab7632dd323/internal/devtools_browser.py#L466-L476

lighthouse-core/gather/driver.js
patrickhulce: dgozman convinced me that so long as the listener is attached synchronously (which it is inside `this._waitForFrameNavigated`), it is impossible for the event to be received too early. 

Longer explanation was that the earliest possible resolution of the promise would be the next microtask and because the listener is installed before that microtask is executed there's no way to miss the event

brendankenny: > dgozman convinced me that so long as the listener is attached synchronously (which it is inside this._waitForFrameNavigated), it is impossible for the event to be received too early.

yeahhhh, good point. As long as no one sticks an `await` in front of that `sendCommand`, I guess?

lighthouse-core/gather/driver.js
```diff
@@ -843,6 +843,10 @@ class Driver {
     const passContext = /** @type {Partial<LH.Gatherer.PassContext>} */ (options.passContext || {});
     const disableJS = passContext.disableJavaScript || false;
 
+    if (waitForNavigated && waitForLoad) {
+      throw new Error('Cannot use both waitForNavigated and waitForLoad, pick just one');
```
brendankenny: 😆

lighthouse-core/gather/driver.js
```diff
@@ -496,6 +496,16 @@ class Driver {
     });
   }
 
+  /**
+   * Returns a promise that resolve when a frame has been navigated.
+   * Used for detecting that our about:blank reset has been completed.
+   */
+  _waitForFrameNavigated() {
+    return new Promise(resolve => {
```
brendankenny: at some point we should add a driver method that wraps `once` with a promise. Seems like a lot of our uses of it end up like this

PR #6449, opened 2018-11-01
misc: fix viewer deploy

brendankenny:
fix bug introduced by #6426

Our version of the `gh-pages` npm package was really old, but I brought over the exact same version as before because it was working so why risk it. However, #6426 also tried to use `dest` from the docs (which wasn't supported in the old version) and it tried to promisify the whole thing. That should have worked (error-first callback as the last argument), but it looks like there was some incompatibility.

This PR
- updates the `gh-pages` package
- wraps `ghPages.publish` in an explicit promise
- adds a nicer `gh-pages` branch commit message than `Updates`, with the lighthouse version number in it

It took [two test commits and a revert](https://github.com/GoogleChrome/lighthouse/commits/gh-pages), but verified that it is working and deploying now :)

### DIFF
--- a/build/build-viewer.js
+++ b/build/build-viewer.js
@@ -13,7 +13,7 @@ const writeFileAsync = promisify(fs.writeFile);
 
 const browserify = require('browserify');
 const cpy = require('cpy');
-const ghPages = promisify(require('gh-pages').publish);
+const ghPages = require('gh-pages');
 const glob = promisify(require('glob'));
 const lighthousePackage = require('../package.json');
 const makeDir = require('make-dir');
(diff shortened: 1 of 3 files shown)


build/build-viewer.js
```diff
@@ -147,6 +147,23 @@ async function compileJs() {
...
+/**
+ * Publish viewer to gh-pages branch.
+ * @return {Promise<void>}
+ */
+async function deploy() {
+  return new Promise((resolve, reject) => {
+    ghPages.publish(distDir, {
+      add: true, // keep existing files
+      dest: 'viewer',
+      message: `Update viewer to lighthouse@${lighthousePackage.version}`,
```
patrickhulce: 🎉

Issue #6444, opened 2018-11-01
Reports not rendering in Edge


codepo8:
#### Provide the steps to reproduce
1. Run LH on christianheilmann.com (or whatever)

#### What is the current behavior?

The generated report doesn't render on Edge as there is a JavaScript error
![lighthouse-bug](https://user-images.githubusercontent.com/13508/47838519-ec672c80-ddaf-11e8-80ea-5b6afe890ba7.PNG)
![renderer](https://user-images.githubusercontent.com/13508/47838518-ebce9600-ddaf-11e8-9603-db22fe8825cd.png)

#### What is the expected behavior?

You can read the report

#### Environment Information
* Affected Channels: CLI
* Lighthouse version: 3.2.1
* Node.js version: 10.9.0
* Operating System: Windows 10

**Related issues**

paulirish:
@codepo8 thanks for reporting. We're fixing this along with a few more Edge compat issues.

FWIW, This particular item stems from a spec violation on exceptions thrown from the `URL()` constructor. I've filed here: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/19498561/

codepo8:
Excellent, thank you!

PR #6456, opened 2018-11-02
fix(report): edge Invalid argument error

wardpeet:
**Summary**
Seo font-size audit has a 'Legible Text' value in an url column. Inside the report we run `new URL('Legible Text');` which throws an error in all browsers. We catch it by checking e.name is typeError. Edge just throws an Error with Invalid Argument as a message. We now support both.

**Related Issues/PRs**
#6444

### DIFF
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -127,7 +127,8 @@ class DetailsRenderer {
       displayedHost = parsed.file === '/' ? '' : `(${parsed.hostname})`;
       title = url;
     } catch (/** @type {!Error} */ e) {
-      if (!e.name.startsWith('TypeError')) {
+      // All browsers throw a TypeError except Edge they throw a regular Error with Invalid argument as a message
+      if (!e.name.startsWith('TypeError') && !e.message.startsWith('Invalid argument')) {
         throw e;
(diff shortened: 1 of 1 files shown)


wardpeet:
Issue:
edge error

Demo:
edge load

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -127,7 +127,8 @@ class DetailsRenderer {
       displayedHost = parsed.file === '/' ? '' : `(${parsed.hostname})`;
       title = url;
     } catch (/** @type {!Error} */ e) {
-      if (!e.name.startsWith('TypeError')) {
+      // All browsers throw a TypeError except Edge they throw a regular Error with Invalid argument as a message
+      if (!e.name.startsWith('TypeError') && !e.message.startsWith('Invalid argument')) {
```
paulirish: i think we could afford to remove this `if (...) { throw e; }` block entirely.

Here's the history:

1. https://github.com/GoogleChrome/lighthouse/pull/2312#discussion_r118729232
1. https://github.com/GoogleChrome/lighthouse/pull/5386#discussion_r192170145

From what I can tell, we might as well just discard any exception here. Throwing at this point seems.. bad. :)

@patrickhulce sgty?

Issue #6441, opened 2018-11-01
Localize the error messages


paulirish:
Specifically the text that ends up in `lhr.runtimeError.message` should be localized.

PR #6457, opened 2018-11-02
i18n: add i18n for LighthouseError messages

exterkamp:
**Summary**
Adds i18n strings for LighthouseError messages. Adds support for substituting in values into an error message through i18n UIString.

**Related Issues/PRs**
fixes: #6441


lighthouse-core/lib/lh-error.js
```diff
@@ -5,12 +5,41 @@
...
+  badTraceRecording: 'Something went wrong with recording the trace over your page load. Please run Lighthouse again.',
+  /** Error message explaining that the page loaded too slowly to perform a Lighthouse run.  */
+  pageLoadTookTooLong: 'Your page took too long to load. Please follow the opportunities in the report to reduce your page load time, and then try re-running Lighthouse.',
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailed: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.',
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailedWithStatusCode: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.  Status Code: {substitute}',
```
exterkamp: Proper ICU format for a generic substitution?

lighthouse-core/lib/lh-error.js
```diff
@@ -5,12 +5,41 @@
...
+  protocolTimeout: 'Waiting for DevTools protocol response has exceeded the allotted time. Method: {substitute}',
+};
+
+const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
 
 /**
  * @typedef LighthouseErrorDefinition
  * @property {string} code
  * @property {string} message
+ * @property {string} [substitute]
```
exterkamp: @brendankenny better name for this variable?

Q: Support multiple substitutions?

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -646,7 +647,8 @@ describe('GatherRunner', function() {
       const error = GatherRunner.getPageLoadError(url, [mainRecord]);
       assert.equal(error.message, 'FAILED_DOCUMENT_REQUEST');
       assert.equal(error.code, 'FAILED_DOCUMENT_REQUEST');
-      assert.ok(/^Lighthouse was unable to reliably load/.test(error.friendlyMessage));
+      i18n.replaceIcuMessageInstanceIds(error, 'en-US');
```
exterkamp: Should we assert off the translated text, or on the pre-translation text?

lighthouse-core/lib/lh-error.js
```diff
@@ -5,12 +5,41 @@
...
+/* eslint-disable max-len */
+const UIStrings = {
+  /** Error message explaining that the Lighthouse run was not able to collect screenshots through Chrome.*/
+  didntCollectScreenshots: `Chrome didn't collect any screenshots during the page load. Please make sure there is content visible on the page, and then try re-running Lighthouse.`,
+  /** Error message explaining that the network trace was not able to be recorded for the Lighthouse run. */
+  badTraceRecording: 'Something went wrong with recording the trace over your page load. Please run Lighthouse again.',
+  /** Error message explaining that the page loaded too slowly to perform a Lighthouse run.  */
+  pageLoadTookTooLong: 'Your page took too long to load. Please follow the opportunities in the report to reduce your page load time, and then try re-running Lighthouse.',
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailed: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.',
```
exterkamp: Any ICU way of de-duping `pageLoadFailed`, `pageLoadFailedWithStatusCode`, and `pageLoadFailedWithDetails`?

lighthouse-core/test/gather/gather-runner-test.js
patrickhulce: we have a helper method for this, `toBeDisplayString` https://github.com/GoogleChrome/lighthouse/blob/541e0f029436ce30d4a524dd6c9ff60f44ca7eae/lighthouse-core/test/test-utils.js#L13-L29

usage: `expect(error.friendlyMessage).toBeDisplayString(/Lighthouse was.../)`

lighthouse-core/test/gather/driver-test.js
```diff
@@ -165,7 +166,9 @@ describe('Browser Driver', () => {
     return driverStub.getRequestContent('', MAX_WAIT_FOR_PROTOCOL).then(_ => {
       assert.ok(false, 'long-running getRequestContent supposed to reject');
     }, e => {
+      i18n.replaceIcuMessageInstanceIds(e, 'en-US');
       assert.equal(e.code, 'PROTOCOL_TIMEOUT');
+      assert.ok(/^Waiting for DevTools.*Method: .+/.test(e.friendlyMessage));
```
patrickhulce: same toBeDisplayString comment here :)

lighthouse-core/lib/lh-error.js
patrickhulce: you generally want to make this be the thing that's being substituted, in this case `{statusCode}`

lighthouse-core/lib/lh-error.js
patrickhulce: ICU is very much into repetition for translation clarity and not treating it like a variable template, I think the repetition here is OK

lighthouse-core/test/gather/gather-runner-test.js
exterkamp: :+1: :+1: :+1: TIL

lighthouse-core/lib/lh-error.js
```diff
@@ -24,7 +53,9 @@ class LighthouseError extends Error {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
-    this.friendlyMessage = errorDefinition.message;
+    this.friendlyMessage = (errorDefinition.substitute) ?
```
patrickhulce: WDYT about having good names for each substitution and just pass the whole errorDefinition in as second arg?

lighthouse-core/test/gather/driver-test.js
```diff
@@ -166,6 +166,7 @@ describe('Browser Driver', () => {
       assert.ok(false, 'long-running getRequestContent supposed to reject');
     }, e => {
       assert.equal(e.code, 'PROTOCOL_TIMEOUT');
+      expect(e.friendlyMessage).toBeDisplayString(/^Waiting for DevTools.*Method: .+/);
```
brendankenny: it is possible to get a test method in here so we can make sure the substitution is working?

lighthouse-core/lib/lh-error.js
brendankenny: > better name for this variable?

errrrr, dunno. I believe these are called arguments in ICU speak, so maybe just `icuArgument`?

> Q: Support multiple substitutions?

I'd say hold off until we need them

lighthouse-core/test/gather/driver-test.js
exterkamp: The .+ is 1 or more characters, so it will fail if the output is "...Method: ".  I wasn't sure if it would be consistently calling: `Method: Network.getResponseBody`.  I suspected that was 100% called first time everytime, but was just playing it safe.

brendankenny: right, I think it's probably worth asserting. It's a bit more roundabout than the other tests that are getting touched in here, but the connection override intentionally makes `Network.getResponseBody` take longer than the `MAX_WAIT_FOR_PROTOCOL`, so it should always throw on that method.

Since we're already breaking into the function implementation for the test (overriding a specific driver method to make sure it will timeout), it seems ok to add to that so we can check the argument substitution as well.

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -1059,7 +1062,8 @@ describe('GatherRunner', function() {
         config: new Config({}),
       }).then(artifacts => {
         assert.equal(artifacts.LighthouseRunWarnings.length, 1);
-        assert.ok(/unable.*load the page/.test(artifacts.LighthouseRunWarnings[0]));
+        expect(artifacts.LighthouseRunWarnings[0])
```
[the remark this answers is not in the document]
exterkamp: I think it was already doing that? I might not get what you're implying.  I made the check more accurate since the regex was a little plain.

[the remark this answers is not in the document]
patrickhulce: well I mean that `toBeDisplayString` does the whole `getFormatted` dance for you, so I'm thinkin' we want at least one test that checks if our code is using `getFormatted` at all the right places.

but maybe I'm misunderstanding or we haven't decided to have the strings in artifacts be pre-resolved yet and the test you added in `runner-test` is good enough for now?

hope that makes more sense

lighthouse-core/test/runner-test.js
```diff
@@ -609,6 +609,27 @@ describe('Runner', () => {
     assert.ok(lhr.runtimeError.message.includes(NO_FCP.message));
   });
 
+  it('internationalized thrown errors from driver', async () => {
```
brendankenny: nit: 'localized errors thrown from driver'

lighthouse-core/test/runner-test.js
```diff
@@ -609,6 +609,27 @@ describe('Runner', () => {
...
+    const erroringDriver = {...driverMock, connect:
```
brendankenny: the formatting on here is a bit confusing. `async` could also help. What about something like
```js
const erroringDriver = {
  ...driverMock,
  async connect() {
    const err = new LHError({
      ...LHError.errors.PROTOCOL_TIMEOUT,
      protocolMethod: 'Method.Failure',
    });
    throw err;
  },
};
```

lighthouse-core/lib/lh-error.js
```diff
@@ -5,14 +5,49 @@
...
+const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
+
 
 /**
  * @typedef LighthouseErrorDefinition
  * @property {string} code
  * @property {string} message
  * @property {RegExp} [pattern]
  * @property {boolean} [lhrRuntimeError] True if it should appear in the top-level LHR.runtimeError property.
+ * ICU argument replacement properties
```
brendankenny: s/properties/values

lighthouse-core/lib/lh-error.js
```diff
@@ -24,7 +59,7 @@ class LighthouseError extends Error {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
-    this.friendlyMessage = errorDefinition.message;
+    this.friendlyMessage = str_(errorDefinition.message, errorDefinition);
```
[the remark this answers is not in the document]
brendankenny: What if the second argument to the `LHError` constructor becomes the replacement values? I feel like it would be conceptually easier to understand without having all of it on the first argument and

```js
const err = new LHError(
  LHError.errors.PROTOCOL_TIMEOUT,
  protocolMethod: `${method}`,
);
reject(err);
```

is a lot easier to understand at a glance than

```js
const err = new LHError({
  ...LHError.errors.PROTOCOL_TIMEOUT,
  protocolMethod: `${method}`,
});
reject(err);
```

I was going to say ideally we'd do this but the existing `properties` second arg makes that a problem, but I think it's actually fine. The only place we ever use that argument is in this file for `fromProtocolMessage()`, and we use it for `protocolMethod` and `protocolError`, one of which is already an icu argument name :)

What if `properties` just did both? It puts its properties on the error object (as it does now), and it gets passed into `str_` (as `errorDefinition` is doing here).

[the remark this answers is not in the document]
brendankenny: I think later on we should do stricter type checking...if you use a string that needs a replacement, you should have to provide the replacement

lighthouse-core/lib/lh-error.js
```diff
@@ -70,116 +105,120 @@ const ERRORS = {
...
+    message: UIStrings.pageLoadFailed,
     lhrRuntimeError: true,
   },
   /* Used when DevTools reports loading failed. Usually an internal (Chrome) issue. */
   FAILED_DOCUMENT_REQUEST: {
     code: 'FAILED_DOCUMENT_REQUEST',
-    message: strings.pageLoadFailed,
+    message: UIStrings.pageLoadFailedWithDetails,
     lhrRuntimeError: true,
+    errorDetails: 'No Failure Description Loaded.',
```
brendankenny: do we want defaults for these? What happens if you don't provide them? Throwing if you don't provide one seems ideal

lighthouse-core/lib/lh-error.js
brendankenny: > s/properties/values

oh, I see what this was saying (properties that are for argument replacement). I take it back :)

lighthouse-core/test/gather/gather-runner-test.js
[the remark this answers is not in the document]
patrickhulce: Yeah exactly. I was somewhat wondering outloud if the `artifacts` is one new place we *should* do the translation.

lighthouse-core/lib/lh-error.js
exterkamp: Yeah the ICU message translator throws.  With a 'value that wasn't replaced', good point, we probably do want to throw in order to avoid a silent lack of real replacement.

exterkamp:
Not ready to review yet 😄 should land #6579 first and then merge those changes in and update to i18n that new error!

lighthouse-core/gather/gather-runner.js
```diff
@@ -147,18 +147,23 @@ class GatherRunner {
...
+      props = {
+        errorDetails: `${mainRecord.localizedFailDescription}.`,
+      };
     } else if (mainRecord.hasErrorStatusCode()) {
-      errorDef = {...LHError.errors.ERRORED_DOCUMENT_REQUEST};
-      errorDef.message += ` Status code: ${mainRecord.statusCode}.`;
+      errorDef = LHError.errors.ERRORED_DOCUMENT_REQUEST;
+      props = {
+        statusCode: `${mainRecord.statusCode}.`,
+      };
```
brendankenny: this is getting complex. Can we start doing early returns on these? Should be easier to read with those.

exterkamp:
Should be g2g, ptal

lighthouse-core/gather/gather-runner.js
```diff
@@ -158,18 +157,18 @@ class GatherRunner {
...
         netErr.startsWith('net::ERR_DNS_')
       ) {
-        errorDef = LHError.errors.DNS_FAILURE;
+        return new LHError(LHError.errors.DNS_FAILURE);
       } else {
-        errorDef = {...LHError.errors.FAILED_DOCUMENT_REQUEST};
-        errorDef.message += ` ${netErr}.`;
+        return new LHError(
+          LHError.errors.FAILED_DOCUMENT_REQUEST,
+          {errorDetails: `${netErr}`}
```
patrickhulce: do these still need to be in string templates?

lighthouse-core/lib/lh-error.js
```diff
@@ -5,7 +5,40 @@
...
 'use strict';
 
-const strings = require('./strings');
+const i18n = require('./i18n/i18n.js');
+
+/* eslint-disable max-len */
+const UIStrings = {
+  /** Error message explaining that the Lighthouse run was not able to collect screenshots through Chrome.*/
+  didntCollectScreenshots: `Chrome didn't collect any screenshots during the page load. Please make sure there is content visible on the page, and then try re-running Lighthouse.`,
+  /** Error message explaining that the network trace was not able to be recorded for the Lighthouse run. */
```
patrickhulce: maybe for some of these descriptions we should include for the translators the distinction of whether it was something weird that retrying should fix, or something that the user needs to fix and try again? I'm thinking a stock explanation.

`This is an error message that was because of certain user conditions they must fix before retrying.`
vs.
`This is an error message that was because of irregularities with Lighthouse that should be fixed by trying again.`

just spitballin'

lighthouse-core/lib/lh-error.js
```diff
@@ -70,128 +103,136 @@ const ERRORS = {
...
+   */
   ERRORED_DOCUMENT_REQUEST: {
     code: 'ERRORED_DOCUMENT_REQUEST',
-    message: strings.pageLoadFailed,
+    message: UIStrings.pageLoadFailedWithStatusCode,
     lhrRuntimeError: true,
   },
-  /* Used when security error prevents page load. */
+  /* Used when security error prevents page load.
+   * Requires an additional `securityMessages` field for translation.
```
patrickhulce: I feel like we're going to forget to update these if they ever change, but I'm not sure what to do about it

lighthouse-core/test/runner-test.js
```diff
@@ -609,6 +609,27 @@ describe('Runner', () => {
...
+      },
+    };
+
+    try {
+      await Runner.run(null, {url: 'https://example.com/', driverMock: erroringDriver, config: new Config()});
+      assert.fail('should have thrown');
+    } catch (err) {
+      assert.equal(err.code, LHError.errors.PROTOCOL_TIMEOUT.code);
+      assert.ok(/^Waiting for DevTools protocol.*Method: Method.Failure/.test(err.friendlyMessage),
+        'should have prevented run');
```
patrickhulce: `should have a localized error message` :)

lighthouse-core/lib/lh-error.js
exterkamp: Well the current plan is that if we change them, then there will be a ton of errors thrown from str_ because it is missing the required substitution fields.  But I'm not sure that's a long term strategy.

exterkamp:
@patrickhulce I think I know why that test is failing: https://github.com/GoogleChrome/lighthouse/blob/master/clients/test/lightrider-entry-test.js#L27

lighthouse-core/gather/gather-runner.js
```diff
@@ -180,12 +179,13 @@ class GatherRunner {
...
     if (securityState === 'insecure') {
-      const errorDef = {...LHError.errors.INSECURE_DOCUMENT_REQUEST};
       const insecureDescriptions = explanations
         .filter(exp => exp.securityState === 'insecure')
         .map(exp => exp.description);
-      errorDef.message += ` ${insecureDescriptions.join(' ')}`;
-      throw new LHError(errorDef);
+      throw new LHError(
+        LHError.errors.INSECURE_DOCUMENT_REQUEST,
+        {securityMessages: `${insecureDescriptions.join(' ')}`}
```
Hoten: can remove string template

lighthouse-core/test/runner-test.js
```diff
@@ -625,8 +625,8 @@ describe('Runner', () => {
       assert.fail('should have thrown');
     } catch (err) {
       assert.equal(err.code, LHError.errors.PROTOCOL_TIMEOUT.code);
-      assert.ok(/^Waiting for DevTools protocol.*Method: Method.Failure/.test(err.friendlyMessage),
-        'should have prevented run');
+      /* eslint-disable-next-line max-len */
+      expect(err.friendlyMessage).toBeDisplayString('Waiting for DevTools protocol response has exceeded the allotted time. Method: Method.Failure.');
```
patrickhulce: wait no we really needed this to be the test that makes sure we're localizing without calling `toBeDisplayString` :D

exterkamp: LOL I was wondering why it worked before!  This makes sense now.

lighthouse-core/gather/driver.js
```diff
@@ -280,9 +280,11 @@ class Driver {
     const timeout = this._nextProtocolTimeout;
     this._nextProtocolTimeout = DEFAULT_PROTOCOL_TIMEOUT;
     return new Promise(async (resolve, reject) => {
-      const asyncTimeout = setTimeout((() => {
-        const err = new LHError(LHError.errors.PROTOCOL_TIMEOUT);
-        err.message += ` Method: ${method}`;
+      const asyncTimeout = setTimeout((_ => {
+        const err = new LHError(
+          LHError.errors.PROTOCOL_TIMEOUT,
+          {protocolMethod: `${method}`}
```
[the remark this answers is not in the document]
exterkamp: Personally I like splitting complex constructors into lines, and with it all on one line it skirts the line of es-lint yelling that the lines are too long.  I feel like 4 lines is more readable than crammed onto 1.

clients/test/lightrider-entry-test.js
```diff
@@ -14,7 +14,10 @@ const LHError = require('../../lighthouse-core/lib/lh-error.js');
 describe('lightrider-entry', () => {
   describe('#runLighthouseInLR', () => {
     it('returns a runtimeError LHR when lighthouse throws a runtimeError', async () => {
-      const connectionError = new LHError(LHError.errors.FAILED_DOCUMENT_REQUEST);
+      const connectionError = new LHError(
+        LHError.errors.FAILED_DOCUMENT_REQUEST,
+        {errorDetails: 'Bad connection req'}
+      );
```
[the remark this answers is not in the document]
exterkamp: I think that is what the includes checks, that the runtime message in the end is the same as the orig friendly msg post a round of translation since it is outside of the Runner.run which translates errors before it exits.  It uses the includes instead of equal b/c LR appends the [msg to the friendly msg](https://github.com/GoogleChrome/lighthouse/blob/3b562abb0525ad0605d7cfc754e5f0472fb39d19/clients/lightrider-entry.js#L69) which may not be desired?  It's the first I'm really seeing it.

lighthouse-core/lib/lh-error.js
```diff
@@ -5,7 +5,40 @@
...
+  badTraceRecording: 'Something went wrong with recording the trace over your page load. Please run Lighthouse again.',
+  /** Error message explaining that the page loaded too slowly to perform a Lighthouse run.  */
+  pageLoadTookTooLong: 'Your page took too long to load. Please follow the opportunities in the report to reduce your page load time, and then try re-running Lighthouse.',
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailed: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.',
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailedWithStatusCode: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests.  Status Code: {statusCode}.',
```
paulirish:
```suggestion
  pageLoadFailedWithStatusCode: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Status code: {statusCode})',
```

lighthouse-core/lib/lh-error.js
```diff
@@ -5,7 +5,40 @@
...
+  /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
+  pageLoadFailedWithDetails: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. Detailed Error: {errorDetails}.',
```
paulirish:
```suggestion
  pageLoadFailedWithDetails: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Details: {errorDetails})',
```

lighthouse-core/lib/lh-error.js
```diff
@@ -5,7 +5,40 @@
...
+  /** Error message explaining that the credentials included in the Lighthouse run were invalid, so the URL cannot be accessed. */
+  pageLoadFailedInsecure: 'The URL you have provided does not have valid security credentials. {securityMessages}.',
```
paulirish:
```suggestion
  pageLoadFailedInsecure: 'The URL you have provided does not have valid security credentials. ({securityMessages})',
```

lighthouse-core/lib/lh-error.js
```diff
@@ -5,7 +5,40 @@
...
+  /** Error message explaining that Chrome has encountered an error during the Lighthouse run, and that Chrome should be restarted. */
+  internalChromeError: 'An internal Chrome error occurred. Please restart Chrome and try re-running Lighthouse.',
+  /** Error message explaining that fetching the resources of the webpage has taken longer than the maximum time. */
+  requestContentTimeout: 'Fetching resource content has exceeded the allotted time',
+  /** Error message explaining that the provided URL Lighthouse points to is not valid, and cannot be loaded. */
+  urlInvalid: 'The URL you have provided appears to be invalid.',
+  /** Error message explaining that the Chrome Devtools protocol has exceeded the maximum timeout allowed. */
+  protocolTimeout: 'Waiting for DevTools protocol response has exceeded the allotted time. Method: {protocolMethod}.',
```
paulirish:
```suggestion
  protocolTimeout: 'Waiting for DevTools protocol response has exceeded the allotted time. (Method: {protocolMethod})',
```

clients/test/lightrider-entry-test.js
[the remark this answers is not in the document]
brendankenny: I had noticed that `runLighthouseInLR` doesn't run i18n on its generated output, hence the question, but the new `err.friendlyMessage = i18n.getFormatted(err.friendlyMessage, settings.locale);` line in `runner.js` translates the error `friendlyMessage` before it gets back to `runLighthouseInLR`, so localization is good to go!

PR #6459, opened 2018-11-02
report(edge): fix support for edge

wardpeet:
**Summary**

Edge has some bugs related to css variables & calc function in the transform rule. Edge doesn't support `fill="hsl()"` colors inside svg
- fixed chevron svg css
- fixed lighthouse logo location
- convert hsl colors to hex in svg fill attributes
- add summary & details collapse polyfill

demo:
http://wardpeet-filestorage.surge.sh/lh-store/pr-6459/report.html

**Related Issues/PRs**

#6444


lighthouse-core/report/html/report-styles.css
```diff
@@ -52,8 +52,9 @@
   --report-border-color: #ccc;
   --report-secondary-border-color: #ebebeb;
   --metric-timeline-rule-color: #b3b3b3;
-  --display-value-gray: hsl(216, 5%, 39%);
+  --display-value-gray: #5E6268;
```
paulirish: this can probably remain hsl, right?

lighthouse-core/report/html/report-styles.css
```diff
@@ -77,19 +78,19 @@
...
   --chevron-size: 12px;
 
-  --pass-icon-url: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><title>check</title><path fill="hsl(139, 70%, 30%)" d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z"/></svg>');
-  --average-icon-url: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><title>info</title><path fill="hsl(31, 100%, 45%)" d="M24 4C12.95 4 4 12.95 4 24s8.95 20 20 20 20-8.95 20-20S35.05 4 24 4zm2 30h-4V22h4v12zm0-16h-4v-4h4v4z"/></svg>');
-  --fail-icon-url: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><title>warn</title><path fill="hsl(1, 73%, 45%)" d="M2 42h44L24 4 2 42zm24-6h-4v-4h4v4zm0-8h-4v-8h4v8z"/></svg>');
+  --pass-icon-url: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><title>check</title><path fill="#178239" d="M24 4C12.95 4 4 12.95 4 24c0 11.04 8.95 20 20 20 11.04 0 20-8.96 20-20 0-11.05-8.96-20-20-20zm-4 30L10 24l2.83-2.83L20 28.34l15.17-15.17L38 16 20 34z"/></svg>');
```
paulirish: This is a compat issue with Firefox. (we ran into it before) And apparently Chrome is going to deprecate this too.

https://chromium-review.googlesource.com/c/chromium/src/+/889718/

`# => %23`

lighthouse-core/report/html/report-styles.css
```diff
@@ -966,6 +967,8 @@
  */
 .lh-chevron {
   --chevron-angle: 42deg;
+  /** calc is not working in edge when using rotate. so we define it here */
```
paulirish: > Edge doesn't support `transform: rotate(calc(...))`, so we define it here

lighthouse-core/report/html/report-styles.css
```diff
@@ -52,8 +52,9 @@
   --report-border-color: #ccc;
   --report-secondary-border-color: #ebebeb;
   --metric-timeline-rule-color: #b3b3b3;
-  --display-value-gray: hsl(216, 5%, 39%);
+  --display-value-gray: #5E6268;
   --report-width: calc(60 * var(--body-font-size));
+  --report-width-half: calc(30 * var(--body-font-size));
```
paulirish: if calc is fine here seems like you can do `calc(var(--report-width) / 2)`

lighthouse-core/report/html/report-template.html
```diff
@@ -36,17 +36,30 @@
   </script>
   <script>window.__LIGHTHOUSE_JSON__ = %%LIGHTHOUSE_JSON%%;</script>
   <script>
+    let pollyfillsLoaded = Promise.resolve();
+    if (!("open" in document.createElement("details"))) {
+      pollyfillsLoaded = new Promise((resolve, reject) => {
+        const script = document.createElement('script');
+        script.src = 'https://unpkg.com/details-element-polyfill@2.0.4/dist/details-element-polyfill.js';
+        script.onload = resolve;
+        script.onerror = reject;
```
paulirish: in this case, i think it'd be an uncaught rejection. Perhaps we set onerror to `resolve` as well? Better to try to render something than give up completely in this case.

wardpeet: wasn't 100% sure about that but i'm ok with that!

lighthouse-core/report/html/report-styles.css
wardpeet: yes, it's ok for translate but not for rotate somehow ¯\_(ツ)_/¯

lighthouse-core/report/html/report-styles.css
wardpeet: sweet didn't know

lighthouse-core/report/html/report-styles.css
wardpeet: did a replace all, sorry.

lighthouse-core/report/html/report-styles.css
wardpeet: nested calcs are not working. It seems fixed https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/?page=1&q=nested%20calc but probably in a next release

wardpeet:
there is still an issue with the tooltips which disappear to fast. Which might be better todo in another PR.
lh-edge-popup

paulirish:
doing the tooltips in a sep PR sgtm

lighthouse-core/report/html/report-styles.css
```diff
@@ -123,6 +125,11 @@
   100% { opacity: 0.6;}
 }
 
+/* Edge doesn't recognize these, so we preemptively set to display:block */
+details, summary {
+  display: block;
```
evenstensberg: That's pretty weird, will it be fixed?

PR #6465, opened 2018-11-04
report: add details-element polyfill for Edge

paulirish:
Split this out of #6459 as it's more complex then the other fixes. Hope that's cool, @wardpeet 	!

Edge doesn't support `<details>` but it's [high priority for them](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/detailssummary/?q=details).

### Our clients:

DevTools Audits panel: Only Chrome support
Chrome extension: Only Chrome support
CLI -> HTML: Should support all browsers
Viewer: Should support all browsers
PSI: Should support all browsers

Where "All browsers" == current chrome, saf, ff, edge, but not IE11.

### Polyfillin
This JS is essentially concatenated into the compiled-reportrenderer via the html-report-assets inclusion. It runs immediately and exits early if a feature test succeeds: https://github.com/GoogleChrome/lighthouse/blob/65f9d6b47cf03b56025658e01f8b3165d90ddf70/lighthouse-core/report/html/third_party/details-element-polyfill.js#L13-L39

Shipping this polyfill this way fixes the CLI and Viewer. PSI is configured separately, but has a similar list of files for inclusion. DevTools is also unaffected, so that's cool. Note, this will add the polyfill superfluously to the extension bundle, but... eh.

_____

@wardpeet 	note i removed these lines: https://github.com/GoogleChrome/lighthouse/blob/71501746d1dca08ff9602ae6883419ddaae2dc1d/lighthouse-core/report/html/report-styles.css#L128-L131  I think that's fine as the polyfill will add these styles, and I expect that to happen before we ever render a `<details>` but lmk wyt :)


patrickhulce:
travis is still unhappy about it being require'd from node during collecting strings, maybe we should have an early exit if it's in node?

lighthouse-core/report/html/third_party/details-element-polyfill.js
```diff
@@ -0,0 +1,297 @@
...
+    polyfillToggle();
+    polyfillFocusAndARIA();
+  }
+
+  if (support.element && !support.toggleEvent) {
+    polyfillToggleEvent();
+  }
+
+}).call(this);
+(function() {
```
patrickhulce: what was this one? :)

paulirish:
sg to all. updated.

lighthouse-core/report/html/html-report-assets.js
```diff
@@ -11,6 +11,8 @@ const REPORT_TEMPLATE = fs.readFileSync(__dirname + '/report-template.html', 'ut
 const REPORT_JAVASCRIPT = [
   fs.readFileSync(__dirname + '/renderer/util.js', 'utf8'),
   fs.readFileSync(__dirname + '/renderer/dom.js', 'utf8'),
+  // COMPAT: Remove when Microsoft Edge supports <details>/<summary>
```
brendankenny: can you link the edge bug?

lighthouse-core/report/html/third_party/details-element-polyfill.js
```diff
@@ -0,0 +1,295 @@
+/*
```
brendankenny: I know there's a thing with copying the report renderer files around, but is it possible to support this in the top-level `third-party/`? It should really be there (it makes it very clear its relationship to lighthouse)

brendankenny: also need a license file in parallel to it, and ideally a readme that says the source commit and lists the modifications made

paulirish: I'll do these as a followup. It adds too much complexity to the roll right now.

PR #6482, opened 2018-11-05
report: rename clump classes; give classes to all audit groups

brendankenny:
working on #6395

The PWA category renderer needs to add custom icons to its groups (like the performance renderer does), but to do so without creating each group itself or grossly reaching into things, this PR sets it up by adding a class name to each group based on the group's id in the config (`.lh-audit-group--${groupId}`, modelled after the existing `lh-audit-group--metrics`, `lh-audit-group--opportunities`, etc).

To not collide with anything, the clump names need to move off a mishmash of class names (`lh-failed-audits`, `lh-audit-group--manual`, etc). So we'll just double down on "clump" for those :)

This PR also removes `lh-audit-group--unadorned` since "unadorned" is not a group name and it's somewhat ambiguous what should fall into that class, but that leaves a bit of a mess with `::before` content and margins. I believe I have that all straightened out now. Previously the "unadorned" case took things away from the styling of clumps and important groups, now it's more the default style and clumps and important groups add to it.

### DIFF
--- a/lighthouse-core/test/report/html/renderer/category-renderer-test.js
+++ b/lighthouse-core/test/report/html/renderer/category-renderer-test.js
@@ -142,13 +142,13 @@ describe('CategoryRenderer', () => {
   it('renders manual audits if the category contains them', () => {
     const pwaCategory = sampleResults.reportCategories.find(cat => cat.id === 'pwa');
     const categoryDOM = renderer.render(pwaCategory, sampleResults.categoryGroups);
-    assert.ok(categoryDOM.querySelector('.lh-audit-group--manual .lh-audit-group__summary'));
+    assert.ok(categoryDOM.querySelector('.lh-clump--manual .lh-audit-group__summary'));
     assert.equal(categoryDOM.querySelectorAll('.lh-audit--manual').length, 3,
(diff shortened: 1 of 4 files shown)


lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -294,13 +308,11 @@ describe('CategoryRenderer', () => {
...
       const elem = renderer.render(category, sampleResults.categoryGroups);
-      const passedAudits = elem.querySelectorAll('.lh-passed-audits .lh-audit');
-      const failedAudits = elem.querySelectorAll('.lh-failed-audits .lh-audit');
+      const passedAudits = elem.querySelectorAll('.lh-clump--passed .lh-audit');
+      const failedAudits = elem.querySelectorAll('.lh-clump--failed .lh-audit');
 
       assert.equal(passedAudits.length, 0);
       assert.equal(failedAudits.length, 12);
-
-      assert.equal(elem.querySelector('.lh-passed-audits-summary'), null);
```
brendankenny: `lh-passed-audits-summary` isn't a thing, so of course it was null :)

PR #6483, opened 2018-11-06
core(seo): consider links with no text non-descriptive

### LINKED ISSUE #6481: SEO Audit Failure: link.href.toLowerCase is not a function
1. Run LH on https://www.browardcriminalteam.com/

The SEO portion of the audit is not passing, nor is it providing explanation.

Under the SEO portion of the audit results I see "Error!" in the top right corner.
When I hover over the text an error message is revealed: "Audit Error: link.href.toLowerCase is not a function"
Here is a snapshot for context: https://www.screencast.com/t/XESTDL0iW

#### Environment Information
* Affected Channels: DevTools
* Lighthouse version: 3.0.3
* Operating System: Windows 10

patrickhulce:
**Summary**
Links inside SVG here throwing the error here and links without any text content were passing the audit too, which seemed wrong, @rviscomi does that check out to you too? I couldn't find any discussion in the original PR (#3490) about it.

**Related Issues/PRs**
fixes #6481

### DIFF
--- a/lighthouse-core/audits/seo/link-text.js
+++ b/lighthouse-core/audits/seo/link-text.js
@@ -8,6 +8,7 @@
 const Audit = require('../audit');
 const URL = require('../../lib/url-shim');
 const BLOCKLIST = new Set([
+  '', // no description is also bad
   'click here',
   'click this',
   'go',
@@ -49,6 +50,12 @@ class LinkText extends Audit {
         }
 
         return BLOCKLIST.has(link.text.trim().toLowerCase());
+      })
+      .map(link => {
+        return {
+          href: link.href,
+          text: link.text.trim().length ? link.text : '<NO TEXT>',
+        };
       });
 
     const headings = [
--- a/lighthouse-core/gather/gatherers/seo/crawlable-links.js
+++ b/lighthouse-core/gather/gatherers/seo/crawlable-links.js
@@ -13,19 +13,28 @@ class CrawlableLinks extends Gatherer {
    * @param {LH.Gatherer.PassContext} passContext
    * @return {Promise<LH.Artifacts['CrawlableLinks']>}
    */
-  afterPass(passContext) {
+  async afterPass(passContext) {
     const expression = `(function() {
       ${pageFunctions.getElementsInDocumentString}; // define function on page
+      const resolveURLOrNull = url => {
+        try { return new URL(url, window.location.href).href; }
+        catch (_) { return null; }
+      };
+
       const selector = 'a[href]:not([rel~="nofollow"])';
       const elements = getElementsInDocument(selector);
       return elements
         .map(node => ({
-          href: node.href,
-          text: node.innerText
+          href: node.href instanceof SVGAnimatedString ?
+            resolveURLOrNull(node.href.baseVal) :
+            node.href,
+          text: node.innerText || '',
         }));
     })()`;
 
-    return passContext.driver.evaluateAsync(expression);
+    /** @type {LH.Artifacts['CrawlableLinks']} */
+    const links = await passContext.driver.evaluateAsync(expression);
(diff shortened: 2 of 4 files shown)


lighthouse-core/gather/gatherers/seo/crawlable-links.js
```diff
@@ -13,19 +13,30 @@ class CrawlableLinks extends Gatherer {
    * @param {LH.Gatherer.PassContext} passContext
    * @return {Promise<LH.Artifacts['CrawlableLinks']>}
    */
-  afterPass(passContext) {
+  async afterPass(passContext) {
     const expression = `(function() {
       ${pageFunctions.getElementsInDocumentString}; // define function on page
+      const resolveURLOrNull = url => {
+        try { return new URL(url, window.location.href).href; }
```
paulirish: __nativeURL

lighthouse-core/gather/gatherers/seo/crawlable-links.js
```diff
@@ -13,19 +13,30 @@ class CrawlableLinks extends Gatherer {
...
+        catch (_) { return null; }
+      };
+
       const selector = 'a[href]:not([rel~="nofollow"])';
       const elements = getElementsInDocument(selector);
```
paulirish: how about

```js
getElementsInDocument(selector).filter(e => __ElementMatches.call(e, 'svg a'))
```

patrickhulce: not 100% sure what you're requesting, like we separate them in two different arrays? or just that we use `.matches` to replace the `instanceof` check?

paulirish: hmmm that was the idea but the fact that resolving these baseVal is so annoying makes it kinda not worth it.


actually how about

```js
const elements = getElementsInDocument(selector).filter(e => __ElementMatches.call(e, 'svg a'))
return elements.map(node => ({
  href: new __URL(node.getAttribute('href'), window.location.href).toString(),  // but with try/catch
  text: // ...
        // oh but then links in SVG don't have innerText
```
okay yeah this is all terrible.

lighthouse-core/gather/gatherers/seo/crawlable-links.js
patrickhulce: I'll do ya one better, I'll enable `useIsolation` :)

lighthouse-core/gather/gatherers/seo/crawlable-links.js
patrickhulce: yeah I agree it's all terrible, I started with `node.innerText || node.textContent` but then got worried about potential accidental fallbacks with script/style tag contents when it's empty and whatnot, wanted to keep it as narrow as possible.

lighthouse-core/audits/seo/link-text.js
```diff
@@ -8,6 +8,7 @@
 const Audit = require('../audit');
 const URL = require('../../lib/url-shim');
 const BLOCKLIST = new Set([
+  '', // no description is also bad
```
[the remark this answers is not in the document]
patrickhulce: Alright, I'll revert 👍

cmoss3:
I see that some changes have been merged here but the issue I brought up in my original ticket (#6481) is still occurring. I'm wondering how long it will be before this fix is available for the lighthouse software that ships with Chrome. Do I need to wait until the next release? If so, can I just download the master branch and use it as-is with NPM?

patrickhulce:
You'll have to wait for the next release with this commit merged, it will likely hit Chrome Canary in a few weeks and Chrome stable in a few months. 

If you're feeling especially brave you could try installing LH HEAD directly from GitHub, but proceed at your own risk :) `npm install GoogleChrome/lighthouse#master`

PR #6505, opened 2018-11-08
core: increase default protocol timeout to 10s

### LINKED ISSUE #6475: Proposal: Increase generic protocol timeout to 10s
We're getting a trickle of bug reports for timeouts on random commands (#6450 #6470 #6472). 5s seems somewhat arbitrary and it might be that 10s increases chances of completion? Perhaps there's something else going on though and these are just surfacing other connection problems. If we think that's the case, we should probably dig a little and try to surface a different error message?

WDYT @Hoten ?

Hoten:
Fixes #6475.


Hoten:
@brendankenny @patrickhulce I modified the wrong value here ...

Hoten:
#6531 got it right :)

patrickhulce:
ha, woah can't believe we all missed that 😆

brendankenny:
😳

PR #6508, opened 2018-11-08
core(locales): streamline hover texts

midzer:
Hi :)

**Summary**

Rework minor unbalance on main performance items for hover text

More or less polishing

Let's sync the way hover texts introduce information

More languages need to be checked...

**Related Issues/PRs**

I've tested latest alpha and stumbled upon this issue


lighthouse-core/audits/metrics/estimated-input-latency.js
```diff
@@ -12,7 +12,7 @@ const UIStrings = {
   /** The name of the metric that marks the estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. Shown to users as the label for the numeric metric value. Ideally fits within a ~40 character limit. */
   title: 'Estimated Input Latency',
   /** Description of the Estimated Input Latency metric that estimates the amount of time, in milliseconds, that the app takes to respond to user input. This description is displayed within a tooltip when the user hovers on the metric name to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
-  description: 'The score above is an estimate of how long your app takes to respond to user ' +
```
patrickhulce: Ah yes good catch we should definitely change from `The score above` to `Estimated Input Latency` 👍 

I'm not sure I like `a rough guess` rather than `an estimate`, that feels like it doesn't give us *quite* enough credit for how it's computed :)

lighthouse-core/audits/metrics/interactive.js
```diff
@@ -12,7 +12,7 @@ const UIStrings = {
   /** The name of the metric that marks the time at which the page is fully loaded and is able to quickly respond to user input (clicks, taps, and keypresses feel responsive). Shown to users as the label for the numeric metric value. Ideally fits within a ~40 character limit. */
   title: 'Time to Interactive',
   /** Description of the Time to Interactive (TTI) metric, which evaluates when a page has completed its primary network activity and main thread work. This is displayed within a tooltip when the user hovers on the metric name to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
-  description: 'Interactive marks the time at which the page is fully interactive. ' +
```
patrickhulce: We went with this because the actual moment is called `interactive` and the metric is the timespan until that moment *Time To* Interactive. Not opposed to rewording so that the name of the metric is the subject but we'll need to change up the rest of the sentence then :)

lighthouse-core/lib/locales/ar-XB.json
```diff
@@ -162,7 +162,7 @@
     "message": "‏‮Minimizes‬‏ ‏‮main‬‏ ‏‮thread‬‏ ‏‮work‬‏"
   },
   "lighthouse-core/audits/metrics/estimated-input-latency.js | description": {
-    "message": "‏‮The‬‏ ‏‮score‬‏ ‏‮above‬‏ ‏‮is‬‏ ‏‮an‬‏ ‏‮estimate‬‏ ‏‮of‬‏ ‏‮how‬‏ ‏‮long‬‏ ‏‮your‬‏ ‏‮app‬‏ ‏‮takes‬‏ ‏‮to‬‏ ‏‮respond‬‏ ‏‮to‬‏ ‏‮user‬‏ ‏‮input‬‏, ‏‮in‬‏ ‏‮milliseconds‬‏, ‏‮during‬‏ ‏‮the‬‏ ‏‮busiest‬‏ 5‏‮s‬‏ ‏‮window‬‏ ‏‮of‬‏ ‏‮page‬‏ ‏‮load‬‏. ‏‮If‬‏ ‏‮your‬‏ ‏‮latency‬‏ ‏‮is‬‏ ‏‮higher‬‏ ‏‮than‬‏ 50 ‏‮ms‬‏, ‏‮users‬‏ ‏‮may‬‏ ‏‮perceive‬‏ ‏‮your‬‏ ‏‮app‬‏ ‏‮as‬‏ ‏‮laggy‬‏. [‏‮Learn‬‏ ‏‮more‬‏](‏‮https‬‏://‏‮developers‬‏.‏‮google‬‏.‏‮com‬‏/‏‮web‬‏/‏‮tools‬‏/‏‮lighthouse‬‏/‏‮audits‬‏/‏‮estimated‬‏-‏‮input‬‏-‏‮latency‬‏)."
```
patrickhulce: Thanks for helping with the translations! I'm afraid these files don't actually persist since they get automatically wiped out when new messages come in from the translators.

@exterkamp do you know if we have a way of passing along suggestions to translators? :)

lighthouse-core/audits/metrics/estimated-input-latency.js
midzer: Yeah, simply `estimate` is probably more meaningful. I want to avoid the word twice in a sentence. Maybe there's an alternative word around.

lighthouse-core/audits/metrics/interactive.js
midzer: Yes, this was my intention to bring the title to first words in the description as everywhere else already.

lighthouse-core/audits/metrics/estimated-input-latency.js
exterkamp: Yeah I'm not sure about "rough guess" maybe "an approximation"?

lighthouse-core/lib/locales/ar-XB.json
exterkamp: To my knowledge (other than the above line comments), no? At the risk of this becoming a daisy chain of escalation @paulirish?

lighthouse-core/audits/metrics/interactive.js
exterkamp: Would it be better as "Time to interactive is the amount of time it takes for the page to become fully interactive?"  Since yea, there is a unit disconnect between a span of time (tti) and a point in time (interactive).

nit: What is the capitalization: `Time to Interactive` or `Time to interactive`?

lighthouse-core/audits/metrics/estimated-input-latency.js
brendankenny: +1 on keeping "estimate"

PR #6515, opened 2018-11-10
misc(logger): declaratively mark functions for timings

### LINKED ISSUE #6425: misc(logger): declaratively mark functions for timings
Proposal to replace this:
```js
class Driver {
    async getBenchmarkIndex() {
        const status = {msg: 'Benchmarking machine', id: 'lh:gather:getBenchmarkIndex'};
        log.time(status);
        const indexVal = await this.evaluateAsync(`(${pageFunctions.ultradumbBenchmarkString})()`);
        log.timeEnd(status);
        return indexVal;
    }
}
```

with this:

```js
class Driver {
    getBenchmarkIndex() {
        return this.evaluateAsync(`(${pageFunctions.ultradumbBenchmarkString})()`);
    }
}
log.autoMark({
    'getBenchmarkIndex': {msg: 'Benchmarking machine', id: 'lh:gather:getBenchmarkIndex'},
});
```

Not as good as decorators, but would allow for removing most usages of `log.time`.

Some POC code:

```js
const log = {
    time: (status) => console.time(status),
    timeEnd: (status) => console.timeEnd(status),
    isVerbose: () => true,
}

class SomeClass {
    doWork() {
        console.log('doWork');
        return 200;
    }

    doWorkLogVerbose() {
        console.log('doWorkLogVerbose');
        return 200;
    }

    async doAsyncWork() {
        console.log('doAsyncWork start');
        await new Promise((resolve) => setTimeout(resolve, 2000));
        console.log('doAsyncWork end');
        return 200;
    }

    dontLogMe() {
        console.log('dontLogMe');
        return 200;
    }
}

function autoMark(klass, methods) {
    for (const method of methods) {
        const original = klass.prototype[method];
        klass.prototype[method] = () => {
            const status = `${klass.name}:${method}`;
            log.time(status);
            const result = original();
            if (result && typeof result.then === 'function') {
                return result.then(value => {
                    log.timeEnd(status);
                    return value;
                });
            } else {
                log.timeEnd(status);
                return result;
            }
        }
    }
}

autoMark(SomeClass, ['doWork', 'doAsyncWork']);
if (log.isVerbose()) {
    autoMark(SomeClass, ['doWorkLogVerbose']);
}

(async () => {
    const obj = new SomeClass();
    console.log('obj.doWork()', obj.doWork());
    console.log('obj.doWorkLogVerbose()', obj.doWorkLogVerbose());
    console.log('obj.doAsyncWork()', await obj.doAsyncWork());
    console.log('obj.dontLogMe()', obj.dontLogMe());
})();

/* output

doWork
SomeClass:doWork: 1.433ms
obj.doWork() 200
doWorkLogVerbose
SomeClass:doWorkLogVerbose: 0.054ms
obj.doWorkLogVerbose() 200
doAsyncWork start
doAsyncWork end
SomeClass:doAsyncWork: 2006.274ms
obj.doAsyncWork() 200
dontLogMe
obj.dontLogMe() 200
*/

```

Thoughts?

Hoten:
Fixes #6425.

Whew, this was a journey into TypeScript for me.

I also converted most usages of `time`/`timeEnd`.  The ones remaining either 1) are around just part of a function, and I did not want to decompose it; 2) in a constructor, which cannot be decorated, nor decomposed without making TS very unhappy; 3) behind a conditional, so converting wouldn't be semantically equivalent.

I noticed that VSCode's TS language server can sometimes fail to identify the types. Requires restarting the language server, and also first viewing the type (you can hover over it) - otherwise it will identify as `any`. For example, the `auditDefn` parameter in `runner.js`.

After, follow up on #6410.


lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
     marky.stop(id);
   }
 
+  /* eslint-disable no-invalid-this */
+  static timeDecorate(originalFn, opts) {
+    return function timeDecoratedFn(...args) {
+      const timeStartLogLevel = opts.timeStartLogLevel || 'log';
```
Hoten: Attempting to maintain the same feature set as before, with how messages are only logged w/ the proper verbosity level.

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+        err.stack = err.stack.replace(/.* at timeDecoratedFn .*\n/g, '');
+        throw err;
+      }
+
+      if (result && typeof result.then === 'function') {
+        return result.then(value => {
+          Log.timeEnd(status, timeEndLogLevel);
+          return value;
+        }).catch(err => {
+          err.stack = err.stack.replace(/.* at timeDecoratedFn .*\n/, '');
```
Hoten: wouldn't work if minified. Is just the viewer minified? is it mangled? I need to check.

Hoten: Even if yes ... I don't think it matters.

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
     marky.stop(id);
   }
 
+  /* eslint-disable no-invalid-this */
+  static timeDecorate(originalFn, opts) {
+    return function timeDecoratedFn(...args) {
+      const timeStartLogLevel = opts.timeStartLogLevel || 'log';
+      const timeEndLogLevel = opts.timeStartLogLevel || 'verbose';
```
patrickhulce: seems like this should be a different opt?

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
     marky.stop(id);
   }
 
+  /* eslint-disable no-invalid-this */
+  static timeDecorate(originalFn, opts) {
```
patrickhulce: typedefs here? :)

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+    return function timeDecoratedFn(...args) {
+      const timeStartLogLevel = opts.timeStartLogLevel || 'log';
+      const timeEndLogLevel = opts.timeStartLogLevel || 'verbose';
+
+      let msg;
+      if (typeof opts.msg === 'string') {
+        msg = opts.msg;
+      } else if (typeof opts.msg === 'function') {
+        msg = opts.msg.bind(this)(...args);
```
patrickhulce:
```suggestion
        msg = opts.msg.apply(this, args);
```

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+      }
+      if (!msg) {
+        throw new Error('expected msg');
+      }
+
+      let id;
+      if (typeof opts.id === 'string') {
+        id = opts.id;
+      } else if (typeof opts.id === 'function') {
+        id = opts.id.bind(this)(...args);
```
patrickhulce: same

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+      if (!id) {
+        throw new Error('expected id');
+      }
+
+      const status = {msg, id};
+      Log.time(status, timeStartLogLevel);
+
+      let result;
+      try {
+        result = originalFn.bind(this)(...args);
```
patrickhulce: same

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+      } catch (err) {
+        err.stack = err.stack.replace(/.* at timeDecoratedFn .*\n/g, '');
+        throw err;
+      }
+
+      if (result && typeof result.then === 'function') {
```
patrickhulce: why not just add a oneline `if (typeof result.then === 'function') result = await result` in the try block above?

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+        return result.then(value => {
+          Log.timeEnd(status, timeEndLogLevel);
```
patrickhulce: should we `finally` the timeEnds? doesn't matter?

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+        });
+      } else {
+        Log.timeEnd(status, timeEndLogLevel);
+        return result;
+      }
+    };
+  }
+  /* eslint-enable no-invalid-this */
+
+  static timeDecorateClass(klass, methods) {
```
patrickhulce: typedefs :)

lighthouse-logger/index.js
```diff
@@ -128,6 +128,72 @@ class Log {
...
+    for (const [method, opts] of Object.entries(methods)) {
+      if (!opts.id) {
+        const className = (typeof klass === 'function' ? klass : klass.constructor).name;
```
patrickhulce: maybe a quick note on why we need to handle both? seems like if this is always called on classes we shouldn't need the situation where it's not a function

lighthouse-logger/test/index-test.js
```diff
@@ -0,0 +1,152 @@
+/**
+ * @license Copyright 2016 Google Inc. All Rights Reserved.
```
patrickhulce: 2018 🎉

lighthouse-logger/test/index-test.js
```diff
@@ -0,0 +1,152 @@
...
+    const fn = object.getValue;
+    const decoratedFn = log.timeDecorate(fn, {
+      msg: 'msg',
+      id: 'id',
+    });
+    assert.equal(decoratedFn(), 'works');
+  });
+
+  it('accepts function template for msg and id', () => {
+    // eslint-disable-next-line no-unused-vars
```
patrickhulce: maybe assert the values are what you expect them to be or just get rid of them if they're not important to the test?

typings/lighthouse-logger/index.d.ts
```diff
@@ -16,10 +16,36 @@ declare module 'lighthouse-logger' {
   export function warn(title: string, ...args: any[]): void;
   export function error(title: string, ...args: any[]): void;
   export function verbose(title: string, ...args: any[]): void;
+  export function isVerbose(): boolean;
   export function time(status: Status, level?: string): void;
   export function timeEnd(status: Status, level?: string): void;
+
+  type ArgumentTypes<F extends Function> = F extends (...args: infer A) => any ? A : never;
```
patrickhulce: oh here are the typedefs! very nice :)

lighthouse-logger/index.js
Hoten: It's called on classes and prototypes - the former are functions but the latter are objects. To decorate instance methods, the first arg should be the prototype. For static methods, it should be the class itself.

It's clearer w/ the comments in the `.d.ts` file. Should I just move those to the implementation file?

lighthouse-logger/index.js
patrickhulce: oh right, silly me :) bummer!

Hoten:
Tests will fail until new version of lighthouse-logger is released.

clients/devtools-entry.js
```diff
@@ -7,7 +7,7 @@
 
 const lighthouse = require('../lighthouse-core/index.js');
 const RawProtocol = require('../lighthouse-core/gather/connections/raw.js');
-const log = require('lighthouse-logger');
```
patrickhulce: if we're doing this it seems like we might as well just directly require in from `../lighthouse-logger` and expose the logger object for setting level and whatnot at `require('lighthouse')`, then we wouldn't need to worry about this publish business.

honestly, the only reason it was separate package at all was so that ChromeLauncher could share the logging functions/levels but I'm thinking there are much better ways to go about sharing a log object with chrome launcher than to continue to jump through all these hoops. I realize this is a bit more than you probably wanted to chew on for this change 😆  but if we're rewriting all the logger require statements to make this work, it's worth discussing.

lighthouse-logger/index.js
Hoten: what about packages that bundle lighthouse (and likely minify) themselves? AFAICT, minifiers like `uglify-js` do not support "do not mangle me" comment markers, which would be the only way to ensure `timeDecoratedFn` is always named as such.

Anyway, it's not a big deal. Even if this fails, that just means an extra stack in an already unreadable (minified) stack trace. Unless source maps are applied, then it would show up. Should only address it if anyone ever complains.

lighthouse-logger/index.js
```diff
@@ -118,33 +174,145 @@ class Log {
...
+   */
+  static timeDecorate(originalFn, opts) {
+    /**
+     * @type {(this: T, ...args: Args) => R}
+     */
+    const fn = function timeDecoratedFn(...args) {
+      const timeStartLogLevel = opts.timeStartLogLevel || 'log';
+      const timeEndLogLevel = opts.timeEndLogLevel || 'verbose';
+
+      let msg;
```
patrickhulce: Not a big deal, but it seems like `msg` and `id` could be turned into functions outside of the `timeDecoratedFn` so we could simplify what's going on inside each invocation a bit, i.e. line 212-234 becomes just `const status = {msg: computeMsg(this, args), id: computeId(this, args)}`

lighthouse-logger/index.js
patrickhulce: do we want to `timeEnd` in this `.catch` block too?

lighthouse-logger/index.js
```diff
@@ -118,33 +174,145 @@ class Log {
...
+   * Decorates (like timeDecorate) specified methods of a class.
+   * If decorating instance methods, use the class's prototype.
+   * If decorating static methods, use the class directly.
+   * @template {Object|Function} Class
+   * @template {keyof Class} Prop
+   * @param {Class} klass
+   * @param {{[key in Prop]: TimeDecorateOpts<Class, ArgumentTypes<IsFunction<Class[key]>>>}} methods
+   */
+  static timeDecorateClass(klass, methods) {
+    for (const [method, opts] of Object.entries(methods)) {
```
patrickhulce: ~~I'm surprised this works with class prototypes, last time I dealt with this I feel like they weren't enumerable and I had to use `getOwnPropertyNames` for some reason. If you happened to run into some issues that you could share why I struggled last time I'd love to hear your insight :)~~

Ohhhhhh you're iterating over the methods that are manually specified, not the class members! 👍

lighthouse-logger/index.js
patrickhulce: Ah gotcha!

lighthouse-logger/index.js
```diff
@@ -118,33 +174,145 @@ class Log {
...
+      if (!opts.id) {
+        const className = (typeof klass === 'function' ? klass : klass.constructor).name;
+        opts.id = `lh:${className}:${method}`;
+      }
+      /** @type {IsFunction<Class[typeof method]>} */
+      const original = klass[method];
```
patrickhulce: maybe just do a quick spelling check here? `if (!original) throw new Error('Cannot decorate non-existant method ${method}')`

lighthouse-logger/index.js
```diff
@@ -118,33 +174,145 @@ class Log {
...
+  }
+
+  /**
+   * @param {string} title
+   * @param {ArrayAtLeastOne} args
+   */
   static log(title, ...args) {
     Log.events.issueStatus(title, args);
-    return Log._logToStdErr(title, args);
+    Log._logToStdErr(title, args);
```
patrickhulce: does this really change anything or was it just for ts?

clients/devtools-entry.js
Hoten: What's the better way to share this code with ChromeLauncher? I would very much like to pull this package back into core.

lighthouse-logger/index.js
Hoten: yeah, changes nothing. The return value was void.

lighthouse-logger/index.js
Hoten: You mean type check, yeah? :P

lighthouse-logger/index.js
Hoten: yeah you got it. these are explicitly defined via an object literal when calling the function

lighthouse-logger/index.js
Hoten: yup!

lighthouse-logger/index.js
Hoten: done

clients/devtools-entry.js
patrickhulce: I was thinking ChromeLauncher could just accept a `logger` parameter as an option that conforms to the API we have and we can pass `lighthouse-logger` into, but we could use regular ol' `debug` by default

lighthouse-logger/index.js
patrickhulce: > You mean existence check, yeah? :P

hey hey let's give ms. anonymous future coder the benefit of the doubt that the method does exist and it was just a typo 😆 😛

Hoten: ahhh i see i see. misunderstood.

speaking of spelling mistakes; existent* haha

Hoten:
@patrickhulce how about this;

1) split logger changes in a separate PR, publish new version
2) update usages of it in another PR
3) open issue to discuss pulling the logger package into core

patrickhulce:
Love it @Hoten, let's make it happen!

PR #6526, opened 2018-11-12
report: add PWA category badge gauge

brendankenny:
Part of #6395

~A bit hacked together, but it works. Work in earlier PRs keeps the hackiness fairly well contained, though, and it should be easy to revise what's in this initial draft.~

Currently
- [x] needs tests
- [x] the svg needs optimizing (a naive optimization eliminated the gradients, so skipped for now)
- [x] there should be a cleaner way to reveal score badges or portions of the score badges
- [ ] (maybe eventually) needs a rethink of css classes (e.g. should the "gauge" be anything in that spot, or just the gauge-like gauge. If so, what should that spot be called)

screen shot 2018-11-11 at 19 24 20

screen shot 2018-11-11 at 20 29 50

### DIFF
--- a/lighthouse-core/report/html/renderer/pwa-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/pwa-category-renderer.js
@@ -47,6 +47,42 @@ class PwaCategoryRenderer extends CategoryRenderer {
   }
 
   /**
+   * @param {LH.ReportResult.Category} category
+   * @return {DocumentFragment}
+   */
+  renderScoreGauge(category) {
+    // Defer to parent-gauge style if category error.
+    if (category.score === null) {
+      return super.renderScoreGauge(category);
+    }
+
+    const tmpl = this.dom.cloneTemplate('#tmpl-lh-gauge--pwa', this.templateContext);
+    const wrapper = /** @type {HTMLAnchorElement} */ (this.dom.find('.lh-gauge--pwa__wrapper',
+      tmpl));
+    wrapper.href = `#${category.id}`;
+
+    const passingGroupIds = this._getPassingGroupIds(category.auditRefs);
+    let className = 'lh-badged-na';
+    if (passingGroupIds.has('pwa-fast-reliable')) {
+      if (passingGroupIds.has('pwa-installable')) {
+        if (passingGroupIds.has('pwa-optimized')) {
+          className = 'lh-badged-all';
+        } else {
+          className = 'lh-badged-fast-reliable-installable';
+        }
+      } else {
+        className = 'lh-badged-fast-reliable';
+      }
+    } else if (passingGroupIds.has('pwa-installable')) {
+      className = 'lh-badged-installable';
+    }
+    wrapper.classList.add(className);
+
+    this.dom.find('.lh-gauge__label', tmpl).textContent = category.title;
+    return tmpl;
+  }
+
+  /**
    * Returns the group IDs whose audits are all considered passing.
    * @param {Array<LH.ReportResult.AuditRef>} auditRefs
    * @return {Set<string>}
--- a/lighthouse-core/report/html/renderer/pwa-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/pwa-category-renderer.js
@@ -62,21 +62,9 @@ class PwaCategoryRenderer extends CategoryRenderer {
     wrapper.href = `#${category.id}`;
 
     const passingGroupIds = this._getPassingGroupIds(category.auditRefs);
-    let className = 'lh-badged-na';
-    if (passingGroupIds.has('pwa-fast-reliable')) {
-      if (passingGroupIds.has('pwa-installable')) {
-        if (passingGroupIds.has('pwa-optimized')) {
-          className = 'lh-badged-all';
-        } else {
-          className = 'lh-badged-fast-reliable-installable';
-        }
-      } else {
-        className = 'lh-badged-fast-reliable';
-      }
-    } else if (passingGroupIds.has('pwa-installable')) {
-      className = 'lh-badged-installable';
+    for (const passingGroupId of passingGroupIds) {
+      wrapper.classList.add(`lh-badged--${passingGroupId}`);
     }
-    wrapper.classList.add(className);
 
     this.dom.find('.lh-gauge__label', tmpl).textContent = category.title;
     return tmpl;
--- a/lighthouse-core/report/html/renderer/pwa-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/pwa-category-renderer.js
@@ -61,13 +61,29 @@ class PwaCategoryRenderer extends CategoryRenderer {
       tmpl));
     wrapper.href = `#${category.id}`;
 
+    const allGroups = this._getGroupIds(category.auditRefs);
     const passingGroupIds = this._getPassingGroupIds(category.auditRefs);
-    for (const passingGroupId of passingGroupIds) {
-      wrapper.classList.add(`lh-badged--${passingGroupId}`);
+
+    if (passingGroupIds.size === allGroups.size) {
+      wrapper.classList.add('lh-badged--all');
+    } else {
+      for (const passingGroupId of passingGroupIds) {
+        wrapper.classList.add(`lh-badged--${passingGroupId}`);
+      }
     }
 
     this.dom.find('.lh-gauge__label', tmpl).textContent = category.title;
     return tmpl;
+  }
+
+  /**
+   * Returns the group IDs found in auditRefs.
+   * @param {Array<LH.ReportResult.AuditRef>} auditRefs
+   * @return {Set<string>}
+   */
+  _getGroupIds(auditRefs) {
+    const groupIds = auditRefs.map(ref => ref.group).filter(/** @return {g is string} */ g => !!g);
+    return new Set(groupIds);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


lighthouse-core/report/html/report-styles.css
```diff
@@ -785,10 +785,15 @@ details, summary {
   border: 0;
 }
 
-.lh-scores-header .lh-gauge__wrapper {
+.lh-scores-header .lh-gauge__wrapper,
+.lh-scores-header .lh-gauge--pwa__wrapper, {
   margin: 0 4px;
 }
 
+.lh-scores-header .lh-gauge--pwa__wrapper {
```
brendankenny: it would be nice if this could be `.lh-gauge__wrapper:last-of-type`, but it doesn't work on classes. Open to other ways to do this that don't rely on the pwa wrapper being where it is

lighthouse-core/report/html/renderer/pwa-category-renderer.js
```diff
@@ -46,6 +46,42 @@ class PwaCategoryRenderer extends CategoryRenderer {
     return categoryElem;
   }
 
+  /**
+   * @param {LH.ReportResult.Category} category
+   * @return {DocumentFragment}
+   */
+  renderScoreGauge(category) {
+    // Defer to parent-gauge style if category error.
+    if (category.score === null) {
```
patrickhulce: I can't decide if I like the explicit documentation that `score === null` when it fails or would prefer to play it safe with `Number.isFinite`, any insight you used to share with me?

lighthouse-core/report/html/renderer/pwa-category-renderer.js
```diff
@@ -46,6 +46,42 @@ class PwaCategoryRenderer extends CategoryRenderer {
...
+    }
+
+    const tmpl = this.dom.cloneTemplate('#tmpl-lh-gauge--pwa', this.templateContext);
+    const wrapper = /** @type {HTMLAnchorElement} */ (this.dom.find('.lh-gauge--pwa__wrapper',
+      tmpl));
+    wrapper.href = `#${category.id}`;
+
+    const passingGroupIds = this._getPassingGroupIds(category.auditRefs);
+    let className = 'lh-badged-na';
+    if (passingGroupIds.has('pwa-fast-reliable')) {
```
patrickhulce: 😆

lighthouse-core/report/html/renderer/pwa-category-renderer.js
brendankenny: > any insight you used to share with me?

I used it just because that's what `categoryRenderer.renderScoreGauge` uses :)

Looking more into it now, `null` is what `ReportScoring.arithmeticMean` returns in case of error, and combined with the fact that we haven't seen any issues with it for plain old `categoryRenderer`...seems fine?

lighthouse-core/report/html/renderer/pwa-category-renderer.js
patrickhulce: to clarify the laughing was on this whole nesting business we need to do, not just this line :)

FWIW, I'd probably find something like 

```js
const isReliable = passingGroupIds.has()
const isInstallable = ...
const isOptimized = ...

if (isReliable && isInstallable && isOptimized) className = 'lh-badged-all'
else if (isReliable && isInstallable) className = 'lh-badged-reliable-installable'
...
```

a bit easier to read, but no biggie

lighthouse-core/report/html/templates.html
```diff
@@ -574,6 +574,218 @@
   </a>
 </template>
 
+<!-- Lighthouse PWA badge gauge -->
+<template id="tmpl-lh-gauge--pwa">
+  <style>
+    .lh-gauge--pwa {
+      width: calc(3 * var(--header-font-size));
```
patrickhulce: is there significance to this `3` or just feels about right?

lighthouse-core/report/html/templates.html
```diff
@@ -574,6 +574,218 @@
...
+      height: calc(3 * var(--header-font-size));
+      display: none;
+    }
+
+    .lh-badged-na .lh-gauge--pwa--na {
```
patrickhulce: these are all just lining up the `lh-badged-*` with `lh-gauge--pwa--*` right? since the intention is all identical and we would ideally replace something simpler maybe it does make sense to combine them?

lighthouse-core/report/html/renderer/pwa-category-renderer.js
brendankenny: ha, that is significantly easier to read. I was very tired when I wrote all that is my only excuse :)

While you were writing this I was just working on a more css-centric version. What do you think of this instead?

The looooong class names are the only downside. Maybe we can short circuit the BEMness to pick short names

lighthouse-core/report/html/templates.html
brendankenny: > is there significance to this 3 or just feels about right?

I copied this over from `.lh-gauge` since `lh-gauge` (and its modifiers) had a bunch of svg-score-gauge-specific properties.  Basically it just means `60px` :)

This is one of the properties that make it seem we need a more general "this is a round-ish thing representing the overall category score" name that includes both the gauge we've had and the badge container (though we could just stick with `lh-gauge` and move the current gauge stuff down inside).

lighthouse-core/report/html/report-styles.css
wardpeet: should we really care if it's an .lh-gauge?
couldn't we just do `.lh-scores-header > :last-child` ?

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -201,15 +201,26 @@ class ReportRenderer {
...
 
     if (scoreHeader) {
+      const defaultGauges = [];
+      const customGauges = [];
+      for (const category of report.reportCategories) {
+        const renderer = specificCategoryRenderers[category.id] || categoryRenderer;
+        const categoryGauge = renderer.renderScoreGauge(category);
+
+        // Group gauges that aren't default at the end of the header
+        if (renderer.renderScoreGauge === categoryRenderer.renderScoreGauge) {
```
paulirish: hmm! how does equality on a function work? seems fine though i'm curious.

lighthouse-core/report/html/templates.html
paulirish: lh-gauge defines `--circle-size: calc(3 * var(--header-font-size));`  in its wrapper. 

seems good to either do that here too or to move the `circle-size` into the `.lh-vars` block.

lighthouse-core/report/html/renderer/pwa-category-renderer.js
```diff
@@ -46,14 +46,53 @@ class PwaCategoryRenderer extends CategoryRenderer {
...
+  renderScoreGauge(category) {
+    // Defer to parent-gauge style if category error.
+    if (category.score === null) {
+      return super.renderScoreGauge(category);
+    }
+
+    const tmpl = this.dom.cloneTemplate('#tmpl-lh-gauge--pwa', this.templateContext);
+    const wrapper = /** @type {HTMLAnchorElement} */ (this.dom.find('.lh-gauge--pwa__wrapper',
+      tmpl));
+    wrapper.href = `#${category.id}`;
```
paulirish: I'm wondering if we should also clarify the state in a tooltip.

We could do with a full lh-tooltip and bulleted list, but here's a minimal `[title]` example: 
image

Regardless, I don't think it should hold up this PR. (which is indeed v clean)

lighthouse-core/report/html/renderer/report-renderer.js
brendankenny: > how does equality on a function work?

it's (shared) on the root prototype, so if a renderer descends from `CategoryRenderer` but doesn't override `renderScoreGauge` it'll just be the same function object.

I was originally going to do something more complicated with verifying *which* prototype the function that will be called was defined on, but it doesn't buy anything extra, and considering this is the only special gauge for now, anything much more complicated than a specialized check directly for the pwa renderer would be compounding the silliness.

lighthouse-core/report/html/renderer/pwa-category-renderer.js
brendankenny: yeah, it feels like it should have a tooltip. I'll add it to the list

hwikyounglee: Alternative idea: 
Fast and reliable: 3/3, Installable: 3/3, PWA Optimized: 4/6

lighthouse-core/report/html/templates.html
brendankenny: > seems good to either do that here too or to move the `circle-size` into the `.lh-vars` block.

after discussion, moved into `.lh-vars` block. It makes sense that the gauges should stay the same size and it turns out we were using the same number in a few other places that could benefit from a single definition

lighthouse-core/report/html/report-styles.css
brendankenny: > couldn't we just do .lh-scores-header > :last-child ?

we can't because when you don't run the PWA category you don't want a vertical line before the *next* `last-child` in line once the PWA gauge is gone :) 

Maybe we should make an explicit divider line class, but this seems fine for now (if/when we ever get another special score gauge we can reevaluate)

wardpeet: Of course :) forgot about running without pwa.

Issue #6510, opened 2018-11-08
Cross browser report issues (edge & safari)


wardpeet:
**Safari** 
1) issue with the dropdown of our action button. It's behind the scores box.
![image](https://user-images.githubusercontent.com/1120926/48222474-f3380580-e394-11e8-8b7d-fd51293e04bb.png)

**Edge**
Edge has an issue with opening the viewer. It opens the viewer but no report.

wardpeet:
> Edge
Edge has an issue with opening the viewer. It opens the viewer but no report.

Sadly there is a bug with edge which makes window.opener undefined when changing origins:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11324352/

should we do an edge workaround where we do not do the postmessage dance for chrome and use a setTimeout.

exterkamp:
fixed by #6546 

@paulirish we need to roll this into the hosted viewer I think, and roll to LR.

exterkamp:
Rolled into hosted viewer and LR! :tada:

PR #6546, opened 2018-11-13
report(safari): fix dropdown overlap translateZ

wardpeet:
**Summary**
fixes the dropdown overlap between score container in safari.

demo:
http://wardpeet-filestorage.surge.sh/lh-store/pr-6546/report.html

**Related Issues/PRs**
#6510

### DIFF
--- a/lighthouse-core/report/html/templates.html
+++ b/lighthouse-core/report/html/templates.html
@@ -247,7 +247,7 @@ limitations under the License.
     .lh-export {
       position: absolute;
       right: var(--section-indent);
-      transform: translateY(0);
+      transform: translateY(0) translateZ(1px);
       top: calc(var(--section-padding) / 2);
       will-change: transform;
       z-index: 2;
--- a/lighthouse-core/report/html/templates.html
+++ b/lighthouse-core/report/html/templates.html
@@ -197,7 +197,6 @@ limitations under the License.
     }
     .lh-scores-wrapper {
       margin-top: -30px;
-      transform: translateZ(1px);
     }
     .lh-scores-wrapper__shadow {
       opacity: 0;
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/templates.html
```diff
@@ -247,7 +247,7 @@
     .lh-export {
       position: absolute;
       right: var(--section-indent);
-      transform: translateY(0);
```
patrickhulce: mind a brief explanation of why we're doing this instead of increased z-index? :)

wardpeet: thanks will do.

lighthouse-core/report/html/templates.html
```diff
@@ -247,7 +247,7 @@
...
+      transform: translateY(0) translateZ(1px);
```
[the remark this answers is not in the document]
wardpeet: sure but i'll add a comment.

wardpeet:
Nope, that's the reason

brendankenny:
so should we `will-change` that instead of translating the dropdown or is that not good enough?

wardpeet:
Good question, I'll have a look

wardpeet:
@brendankenny I could even remove them both as it was unnecessary on the scores_wrapper. I updated the demo on http://wardpeet-filestorage.surge.sh/lh-store/pr-6546/report.html but i've tested it on Safari, chrome, firefox and edge.

brendankenny:
yeah, makes sense, I just wasn't sure about the perf work that went into putting it there originally. Rendering looks correct without them, we just don't want to jank on scroll :)

@paulirish for his knowledge

wardpeet:
@brendankenny it shouldn't trigger jank as we're not animating that layer, we are animating it's children so we should be good but lets wait for the perf god 😄

PR #6559, opened 2018-11-14
new_audit: add last painted hero metric prototype

patrickhulce:
**Summary**
Prototype of Last Painted Hero, gathering part is heavily based on the [implementation in WPT](https://github.com/WPO-Foundation/wptagent/blob/bee2afd18a2e4184cef63ca7341ce0dd275f9adf/internal/js/hero_elements.js). Computation is a bit different since we have access through speedline.

Rough two step process:

1. Find all the largest h1, h2, img, CSS background image BCRs
2. Take a look at the screenshots and find the timestamp of the last change that occurred within these BCRs

I haven't done much investigation but I can already identify a couple problems here,

**Changes that occur within the BCR, but unrelated to the hero element, mess with the timings.**

On my site for example, I have a video playing behind the H1 and so Last Painted Hero time is literally the last screenshot that it recorded because every frame is different.

**BCR is measured at the end, not as it was painted**

One might argue this is a feature, not a bug, and it's mostly a similar problem to the speed index issue of late major layout shifts, but it still has a pretty big impact nonetheless. As an example, CNN's header and image BCRs are as measured at the end of page load, after the ad has popped in and shifted everything down. This means we're not really measuring when these elements showed up, but the time at which they reached their final resting location.

**Not many ideas on Lantern implementation**

This isn't necessarily a deal-breaker, especially at this stage, but because this is so tied to individual DOM elements, the implementation in Lantern will be very difficult. I have some basic ideas, but 
I can't image our accuracy here would be very good at all.


brendankenny:
first thing: what's a BCR? :)

patrickhulce:
bump :)

patrickhulce:
> Ideally, I'd like to rely on the instrumentation there, but it's not shipping yet. :/

That sounds great, should eliminate a lot of the problems I identified anyhow.

> Should we schedule a call with maxlg to explore a plan?

Let's. There hasn't seemed to be any urgency from folks about getting this into LH, so just making sure we have a path to relying on the trace events sounds good :)

brendankenny:
From talking to @paulirish, it sounds like we want to delegate to LCP instead so we're not having to replicate/maintain the logic they're developing there. Unfortunately not exposed in traces yet, but work is progressing: https://crbug.com/869924 (somewhere in the middle there the `LCP++` name emerges :)

PR #6578, opened 2018-11-16
core(driver): assert securityIssues before getPageLoadError

paulirish:
I _think_ we want to throw on security issues before we consider pageLoadError situations. wdyt?

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -300,6 +300,8 @@ class GatherRunner {
     const networkRecords = NetworkRecorder.recordsFromLogs(devtoolsLog);
     log.timeEnd(status);
 
+    this.assertNoSecurityIssues(driver.getSecurityState());
+
     let pageLoadError = GatherRunner.getPageLoadError(passContext.url, networkRecords);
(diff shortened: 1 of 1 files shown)


Hoten:
What's the difference in behavior? Seems that either way, on a security issue an error is thrown that is uncaught, going all the way to `runner.js` (And beyond).

image

brendankenny:
> now I see that log.error does more than just log, and it actually does throw an error

I don't think that's right? [`Log.error`](https://github.com/GoogleChrome/lighthouse/blob/9910b9bd5d967514e0f6492ef4892fb7240701ea/lighthouse-logger/index.js#L141-L143)

> The current code pushed the page load error onto the current passContext, but if there is a security error, that will throw an error, exiting the flow of the entire runner.

ah, right. So if there was a security error, the `pageLoadError` gets ignored whether the security error check is before or after it.

From the perspective of reading the code it makes sense to have it first (and maybe even earlier), but I don't think this is going to change the error type distribution :)

Hoten:
Ah, I was looking at the render logger. OK, so yeah, no change really in this PR except for avoiding logging to output a page load error when really it was a security issue. LGTM

PR #6585, opened 2018-11-16
 core(driver): assert page not hung after each gatherer

patrickhulce:
**Summary**
per @paulirish 's request this checks if the page is hung after each gatherer's afterPass. The smoketest takes a full minute since our timeout for evaluate is a full 60s :/ If other's are cool with it, I'll add an option for controlling this? Could be a `process.env.SMOKETEST_PROTOCOL_TIMEOUT` or a real option like `maxWaitForLoad`

**Related Issues/PRs**
#6497

### DIFF
--- a/lighthouse-cli/test/fixtures/infinite-loop.html
+++ b/lighthouse-cli/test/fixtures/infinite-loop.html
@@ -14,8 +14,17 @@
   This is the function that never ends.
 
   <script>
-    while (true) {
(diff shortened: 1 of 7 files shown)


paulirish:
Right now it appears this is no longer necessary. We can resurrect it if we're seeing a problem though.

Thanks @patrickhulce for drafting this up!

PR #6579, opened 2018-11-16
core(gather): define new DNS failure LH Error

paulirish:
Seeing a lot of these errors. I think we want to expose them to the user individually.

PR #6592, opened 2018-11-17
misc(lint): break lines after operators

brendankenny:
inspired by https://github.com/GoogleChrome/lighthouse/pull/6579#discussion_r234372410

Surprised this wasn't on either.
eslint-config-google has the very helpful lines
```
// TODO(philipwalton): add a rule to enforce the operator appearing
// at the end of the line.
```

:P Were we all just following this unconsciously?

I learned from reading about the `operator-linebreak` rule that some people really do prefer to have the ternary (ConditionalExpression) operators at the start of a new line, so we could maybe discuss that for these since there's some preexisting ones.

However, in our code base the `?` and `:` at the end of lines outnumbered them at the beginning of lines by like 5 to 1, so if we go with one of these choices, the changes in this PR seem like the better option.

(`operator-linebreak` is also supported by `--fix`, so should be fine for those using prettier and then fixing)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -68,6 +68,7 @@ module.exports = {
     'require-jsdoc': 0,
     'valid-jsdoc': 0,
     'arrow-parens': 0,
+    'operator-linebreak': [2, 'after'],
   },
   parserOptions: {
     ecmaVersion: 2018,
--- a/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
+++ b/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
@@ -118,9 +118,9 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
       const firstRuleStart = preview.indexOf('{');
       const firstRuleEnd = preview.indexOf('}');
 
-      if (firstRuleStart === -1 || firstRuleEnd === -1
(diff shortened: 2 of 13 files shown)


paulirish:
I think I like 'before' better. 😀

brendankenny:
> I think I like 'before' better. 😀

which one? All of them? One issue is they don't match our style elsewhere

`?` end of line: 33 results
`?` beginning of line: 8 results

`:` end of line: 21 results
`:` beginning of line: 8 results

`+` end of line: 303 results
`+` beginning of line: 2 results

`||` end of line: 59 results
`||` beginning of line: 4 results

`&&` end of line: 67 results
`&&` beginning of line: 1 result

and all other operators have been put at the end of lines 😛

patrickhulce:
I like prettier, and as a majority of the changes in this PR are on my files that were formatted with prettier on save, I'm a bit bummed that dream would die 😞 

Personally though, I do prefer `after` to `before` even if that habit originated from foolish ASI paranoia.

wardpeet:
we might consider [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) if we like the prettier approach so we don't conflict with eslint rules.

brendankenny:
:grinning: :stuck_out_tongue:

paulirish:
image

PR #6593, opened 2018-11-17
report: use details-element-polyfill from node_modules

brendankenny:
followup to #6465

Since LR now takes care of its own copy of `details-element-polyfill`, we can remove it from `report/html/third_party/` and just rely on the one in `node_modules`, copying it into the report from there.


lighthouse-core/report/html/report-styles.css
```diff
@@ -160,11 +160,6 @@
   display: none !important;
 }
 
-/* Edge doesn't recognize these, and the polyfill doesn't set block. https://github.com/javan/details-element-polyfill/issues/13 */
-details, summary {
-  display: block;
-}
```
brendankenny: now [included in the polyfill](https://github.com/javan/details-element-polyfill/blob/5984dc6986720a739a89c484574f9a3e58537a58/dist/details-element-polyfill.js#L22)

package.json
```diff
@@ -140,6 +140,7 @@
     "chrome-launcher": "^0.10.5",
     "configstore": "^3.1.1",
     "cssstyle": "1.1.1",
+    "details-element-polyfill": "2.1.1",
```
brendankenny: locked to the same version we use for LR

paulirish:
budgets

image

PR #6596, opened 2018-11-17
report: expand groups within collapsed clumps

brendankenny:
This is still under discussion, but ready to go if we decide to go this way.

When groups are nested under a closed clump (e.g. the a11y groups under the `Passed audits` section of the a11y category) the groups are also collapsed themselves. This isn't great, as when you open the collapsed section, the whole point is that you're indicating that you want to see the audits contained within.

This PR adds another audit group option when the group is set to be `expandable` that indicates whether it should be expanded or collapsed by default. I went with `open` for this option, but `expanded` might be better (it just seemed too similar to the other option and `open` matches the `<details>` attribute).

Humorously, all groups that are expandable are also set to be `open` *except* the abuse of using `renderAuditGroup` to render clumps, which are expandable but sometimes not open (e.g. `Passed audits`). If we fix the clump TODO on `renderAuditGroup` then we can just drop the `open` option and have them always open.

expanded-groups

### DIFF
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -182,7 +182,7 @@ class CategoryRenderer {
    * Renders the group container for a group of audits. Individual audit elements can be added
    * directly to the returned element.
    * @param {LH.Result.ReportGroup} group
-   * @param {{expandable: boolean, itemCount?: number}} opts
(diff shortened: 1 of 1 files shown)


patrickhulce:
I don't have particularly strong feelings here, but it seems like the a11y category would be the one I'd be most hesitant about. Dumping 100 audits on you at once when you just expanded a single level seems like it might be a bit much. The smaller the category, the more I agree with making this change I suppose.

PR #6607, opened 2018-11-19
report: import <detail> polyfill as string, not Buffer

brendankenny:
Fixes viewer bundle size inflation introduced by #6593. The `details-element-polyfill` was accidentally being inlined as a base64-encoded Buffer and then having `toString()` called on it.

The string version is smaller, and we don't also need `Buffer` browserified in :)


brendankenny:
> ```
> $ bundlesize
>  PASS  ./dist/extension/scripts/lighthouse-ext-bundle.js: 361.57kB < maxSize 520kB gzip 
>  PASS  ./dist/viewer/src/viewer.js: 58.21kB < maxSize 65kB gzip 
>```

back where they should be.

paulirish:
budget success!!!

PR #6608, opened 2018-11-19
core: INSECURE_DOCUMENT_REQUEST errors still return lhr

### LINKED ISSUE #6595: INSECURE error not working as expected
`lighthouse https://zach.diemer.codes/`

This should return an LHR with a runtimeError set to the INSECURE_DO CUMENT_REQUEST

Instead:

1. we first wait for Page.navigate to timeout. (kind of a bummer..)
2. the timeout of Page.navigate triggers a rejection which is unhandled
3. it bails completely instead of making an LHR

Hoten:
Fixes #6595.


Hoten:
Bad bot. why would you close a PR

patrickhulce:
Could we get a bit of explanation for what was happening before and why we want this instead? :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -256,8 +239,13 @@ class GatherRunner {
     if (recordTrace) await driver.beginTrace(settings);
 
     // Navigate.
-    await GatherRunner.loadPage(driver, passContext);
-    log.timeEnd(status);
+    try {
+      await GatherRunner.loadPage(driver, passContext);
```
patrickhulce: we're injecting a lot of new if/try/catch blocks into some already beefy methods, do you think there's a way we might be able to push some of these complications down a level so the busier method is still easy to parse?

Hoten:
When there is a security issue, `Page.navigate` timesout, and the entire LH run is aborted fatally without returning a LHR. If ignored, more commands hang forever (`Network.emulateNetworkConditions`) - so the gather runner aborts as soon as it detects a security issue.

Hoten:
It would be much nicer if we could just throw the security error when discovered (like the PR has it in `driver.goToURL`), and don't attempt to handle it at all in the gather runner, letting it go up to the catch in `Runner.run`, and just returning a bare LHR with the runtimeError set. But that seems like a bigger change.

lighthouse-core/gather/gather-runner.js
Hoten: yes, if we can refactor Runner.run to easily return a (errored) LHR in it's catch. Then all the code but what's in driver.js could be removed.

Hoten:
OK, I removed all that complexity. Test w/ `node lighthouse-cli https://expired.badssl.com/ --view` to see that the run bails immediately on a security issue, instead of hanging.

The original issue requested returning an LHR, but I believe we don't care about that anymore, since LR now correctly forms a LHR when an error is thrown. @paulirish is this correct?

brendankenny:
> The original issue requested returning an LHR, but I believe we don't care about that anymore, since LR now correctly forms a LHR when an error is thrown. 

a million 👍s from me :)

lighthouse-core/lib/lh-error.js
```diff
@@ -22,7 +22,7 @@ const UIStrings = {
   /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
   pageLoadFailedWithDetails: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Details: {errorDetails})',
   /** Error message explaining that the credentials included in the Lighthouse run were invalid, so the URL cannot be accessed. */
-  pageLoadFailedInsecure: 'The URL you have provided does not have valid security credentials. ({securityMessages})',
+  pageLoadFailedInsecure: 'The URL you have provided does not have valid security credentials. {securityMessages}',
```
exterkamp: Any reason for removing these?  I feel like a list that might make no sense as a sentence like this belongs in parens. e.g. The URL...credentials. (Reason 1 Reason 2)

lighthouse-core/lib/file-namer.js
```diff
@@ -16,11 +16,11 @@
  * Generate a filenamePrefix of hostname_YYYY-MM-DD_HH-MM-SS
  * Date/time uses the local timezone, however Node has unreliable ICU
  * support, so we must construct a YYYY-MM-DD date format manually. :/
- * @param {{finalUrl: string, fetchTime: string}} lhr
+ * @param {{requestedUrl: string, finalUrl: string, fetchTime: string}} lhr
```
exterkamp: What is this requestedUrl business about?  Does it relate to INSECURE DOC requests?

lighthouse-core/lib/lh-error.js
exterkamp: A realistic error message reads like this:
>The URL you have provided does not have valid security credentials. This site is missing a valid, trusted certificate (net::ERR_CERT_DATE_INVALID).

Which makes it being a sentence w/o parens more valid, but makes appending multiple errors together with a space make less sense.

lighthouse-core/gather/driver.js
```diff
@@ -512,6 +505,35 @@ class Driver {
...
+      const insecureDescriptions = explanations
+        .filter(exp => exp.securityState === 'insecure')
+        .map(exp => exp.description);
+      return new LHError(LHError.errors.INSECURE_DOCUMENT_REQUEST, {
+        securityMessages: insecureDescriptions.join(' '),
+      });
+    }
+  }
+
+  waitForSecurityIssuesCheck() {
```
brendankenny: can we just combine the two of these? Seems unnecessary to split (`waitForSecurityIssuesCheck` doesn't really do anything in isolation and they aren't called separately)

lighthouse-core/gather/driver.js
```diff
@@ -895,6 +917,13 @@ class Driver {
     // No timeout needed for Page.navigate. See #6413.
     const waitforPageNavigateCmd = this._innerSendCommand('Page.navigate', {url});
 
+    try {
+      await this.sendCommand('Security.enable');
+      await this.waitForSecurityIssuesCheck();
+    } finally {
+      this.sendCommand('Security.disable');
+    }
```
brendankenny: To do this right, I think @paulirish is right and this should move into `waitForFullyLoaded`. We shouldn't be mucking around looking at network traffic before our traffic monitor is allowed to start monitoring :)

Take a look at `this._waitForLoadEvent` or `this._waitForNetworkIdle` for the `promise`/`cancel` cancellable promise model used, and `_waitForFullyLoaded` itself for where to put it in parallel with the timeout. It should be a pretty simple translation from `waitForSecurityIssuesCheck` already in this PR.

(and forgive the state of the code in those methods...they're super old and have pretty much only grown through accretion, not really refactored, so they're a bit creaky and brittle :)

lighthouse-core/gather/gather-runner.js
brendankenny: We're dropping the LHR requirement for now, so we can also drop these changes (there won't be a timing object to add this to if an error was thrown)

lighthouse-core/lib/file-namer.js
Hoten: Oof you're right, forgot to check if that was still necessary.

lighthouse-core/gather/driver.js
Hoten: what about the `_waitForFrameNavigated` case?

brendankenny: > what about the `_waitForFrameNavigated` case?

that's only used for loading `about:blank`, so we're ok there. We should probably make that clearer, though really if you're saying `waitForNavigated: true`, you're basically forgoing all our checks already.

lighthouse-core/gather/driver.js
paulirish: we talked earlier about the order of the listener vs calling enable().

......maybe the order right now is OK. if you listened before calling security.enable then perhaps you would get a statechanged event for the about:blank state.

I'm not totally sure about this.

but typically we do like setting up the event listener _before_ enable is called, because enable _usually_ flushes all events before it results.

lighthouse-core/gather/driver.js
```diff
@@ -512,6 +505,53 @@ class Driver {
...
+    let cancel = () => {
+      throw new Error('_waitForSecurityCheck.cancel() called before it was defined');
+    };
+
+    const promise = new Promise(async (resolve, reject) => {
+      /**
+       * @param {LH.Crdp.Security.SecurityStateChangedEvent} event
+       */
+      const securityStateChangedListener = ({securityState, explanations}) => {
+        this.sendCommand('Security.disable');
```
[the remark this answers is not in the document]
Hoten: Do you mean adding an `await` here?

lighthouse-core/gather/driver.js
```diff
@@ -752,6 +795,7 @@ class Driver {
     // Wait for both load promises. Resolves on cleanup function the clears load
     // timeout timer.
     const loadPromise = Promise.all([
+      waitForSecurityCheck.promise,
```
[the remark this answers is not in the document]
Hoten: I think it'd only work if the `Security` domain is disabled before `waitForPageLoad` runs. If a gatherer were to enable `Security` and forget to disable it, this may break?

Also, isn't this method always called from "about:" -> "target url"?

lighthouse-core/gather/driver.js
```diff
@@ -512,6 +505,53 @@ class Driver {
...
+          });
+          reject(err);
+        } else {
+          resolve();
+        }
+      };
+
+      // wait for Security.enable to resolve, so we skip whatever state change
+      // events are in the pipeline
+      await this.sendCommand('Security.enable');
```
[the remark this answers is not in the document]
Hoten: The test would basically be 1) go to https://expired.badssl.com/ 2) assert LH does not hang / exercise the full page load time out (so, within 5s??)

would that be a good candidate for a smoke test?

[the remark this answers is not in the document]
Hoten: OK, it looks like it starts with the "neutral" security state (about: has no meaningful value for security) and then either goes to "secure" or "insecure", depending on the target page. So I can tweak the logic to listen for state changes until "not neutral" comes up

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
patrickhulce: > Also, isn't this method always called from "about:" -> "target url"?

That's kinda what I'm talking about, just explaining why this is always going to be OK :)

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
patrickhulce: or an empty `.catch` if we don't care?

lighthouse-core/gather/driver.js
```diff
@@ -512,6 +505,52 @@ class Driver {
...
+          });
+          reject(err);
+        } else if (securityState !== 'neutral') {
+          cancel();
+          resolve();
+        }
+      };
+
+      this.on('Security.securityStateChanged', securityStateChangedListener);
+      this.sendCommand('Security.enable');
```
exterkamp: Confirmed this will crash with `Unhandled Rejection. Reason: LHError: PROTOCOL_TIMEOUT`

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
patrickhulce: > would that be a good candidate for a smoke test?

sounds great to me! ~~who maintains badssl? do we expect it to be roughly up most of the time?~~ oh we do, awesome yeah let's do it :D https://github.com/chromium/badssl.com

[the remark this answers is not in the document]
brendankenny: Just `https://localhost:10200` could also work

Hoten:
I went with badssl.com for the smoke test. Also added a special exit code for INSECURE_DOCUMENT_REQUEST exits. Also did #6851.

lighthouse-core/gather/driver.js
Hoten: it's k now

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
Hoten: Ah, OK I misunderstood. How's the comment I added?

lighthouse-core/gather/driver.js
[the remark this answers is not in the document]
Hoten: If in offline mode, the security state is always "neutral" ...

This is me running the pwa smoke test. First two security states  are the online pass (neutral -> secure), last three(?) are the offline pass (neutral -> neutral... but with explanations)
image

But in each case, the first state we care about is the first one with non-empty explanations. So maybe use that?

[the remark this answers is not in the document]
Hoten: Which makes sense... if there's no explanation for a security state, why value it?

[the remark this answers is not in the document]
Hoten: image

For some reason, this url gives three empty, neutral states in the offline pass. Perhaps better to just disable this check for the offline pass.

lighthouse-core/gather/driver.js
```diff
@@ -512,6 +505,57 @@ class Driver {
...
+          resolve();
+        }
+      };
+
+      this.on('Security.securityStateChanged', securityStateChangedListener);
+      this.sendCommand('Security.enable').catch(() => {});
+
+      cancel = () => {
+        this.off('Security.securityStateChanged', securityStateChangedListener);
+        this.sendCommand('Security.disable').catch(() => {});
```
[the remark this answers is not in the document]
Hoten: Sounds reasonable to reject it. Unclear if it would be considered an unhandled rejection.. I think not, because of the `Promise.race`

Hoten:
@brendankenny what about skipping the security check if `http`?

lighthouse-core/gather/driver.js
```diff
@@ -741,6 +790,12 @@ class Driver {
     /** @type {NodeJS.Timer|undefined} */
     let maxTimeoutHandle;
 
+    // Noop if offline or localhost
+    const isLocalhost = new URL(this._monitoredUrl || '').hostname !== 'localhost';
```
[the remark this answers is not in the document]
Hoten: Also, this variable is named incorrectly. :/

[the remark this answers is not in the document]
paulirish: If we need to check the hostname lets use reuse this bit:
https://github.com/GoogleChrome/lighthouse/blob/bb3fcdddd98edbd4c9dd8fb1981acb168f3b7d4c/lighthouse-core/audits/is-on-https.js#L14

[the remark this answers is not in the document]
Hoten: done. what's a good place to move this (if we don't want to duplicate these constants)?

lighthouse-core/gather/driver.js
```diff
@@ -741,6 +792,14 @@ class Driver {
     /** @type {NodeJS.Timer|undefined} */
     let maxTimeoutHandle;
 
+    // Noop if offline or not https
+    // https: => https
+    const protocol = new URL(this._monitoredUrl || '').protocol.replace(/:$/, '');
```
brendankenny: It's possible we can make this approach work, but I don't think it will work like this (correct me if I'm missing something).

`_monitoredUrl` is updated as the page loads to account for redirects, so we would have to recheck every time it updates whether or not we should be checking for security state changes. So I believe a site that redirects http -> https won't get a check since the value will initially be `!isSecureProtocol`?

lighthouse-core/gather/driver.js
```diff
@@ -32,6 +32,8 @@ const DEFAULT_CPU_QUIET_THRESHOLD = 0;
 // Controls how long to wait for a response after sending a DevTools protocol command.
 const DEFAULT_PROTOCOL_TIMEOUT = 30000;
 
+const SECURE_SCHEMES = ['data', 'https', 'wss', 'blob', 'chrome', 'chrome-extension', 'about'];
```
paulirish: remains to be seen if we need this here, but if we do, i'd much rather this array live in a common place than be duplicated here and in the is-on-https audit.
url-shim.js has worked well in the past. move there and import to here and the audit?

brendankenny:
Ha, paul and I were just discussing something very similar and I think you're right. We really only care if an `insecure` event fires at all--not which other states fire, or when they fire, or transitions between security states--so I think that makes sense.

Basically make it another version of the `maxTimeoutPromise` promise, cancelling all the `waitForX`s and then and racing with `loadPromise` and `maxTimeoutPromise` at the end of `waitForFullyLoaded`?

lighthouse-core/gather/driver.js
paulirish: Good call, Brendan.

Example URL that redirects to insecure (but isn't caught currently): http://i.paul.irish  
_(as soon as DNS propagates, http://expiredredirect.paul.irish will do the same)_

lighthouse-core/gather/driver.js
```diff
@@ -832,9 +857,11 @@ class Driver {
 
     // Wait for load or timeout and run the cleanup function the winner returns.
```
patrickhulce: edit this comment to note that `securityCheckPromise` will only ever reject?

patrickhulce: OR crazy idea, the promise resolves with a `cleanupFn` that throws an error :D then the three cases are a little more parallel

lighthouse-core/gather/driver.js
```diff
@@ -792,6 +786,40 @@ class Driver {
...
+    const cancelAll = () => {
+      waitForFCP.cancel();
+      waitForLoadEvent.cancel();
+      waitForNetworkIdle.cancel();
+      waitForCPUIdle.cancel();
+    };
+
+    // Promise that only rejects when an insecure security state is encountered
+    let securityCheckCleanup = () => {};
+    const securityCheckPromise = new Promise((_, reject) => {
```
patrickhulce: there's a lot to unpack in this function that needs longstanding cleanup I'll take a whack at as a followup to #6944, for now WDYT about sticking this in a `_waitForSecurityError` or something following the `{promise, cancel}` paradigm we have currently?

lighthouse-core/gather/driver.js
Hoten: to your second comment, so `securityCheckPromise` would be changed to `resolve` instead of `reject`?

lighthouse-core/gather/driver.js
Hoten: The `cancelAll` and clearing the `maxTimeoutHandle` complicates that a lot. It's not clear to me how to extract into a function without using multiple parameter callbacks (or is that the cleanup you had in mind?)

lighthouse-core/gather/driver.js
patrickhulce: > to your second comment, so securityCheckPromise would be changed to resolve instead of reject?

right it would go hand-in-hand with https://github.com/GoogleChrome/lighthouse/pull/6608#discussion_r246261703 to enable to refactor

lighthouse-core/gather/driver.js
patrickhulce: yeah this goes hand-in-hand with the `resolve` idea. I'm not sure we could easily pull it out with it still rejecting immediately

Hoten: OK, I think I see how that'd work. I'll give it a shot

lighthouse-core/gather/driver.js
```diff
@@ -816,10 +868,8 @@ class Driver {
     }).then(_ => {
       return async () => {
         log.warn('Driver', 'Timed out waiting for page load. Checking if page is hung...');
-        waitForFCP.cancel();
-        waitForLoadEvent.cancel();
-        waitForNetworkIdle.cancel();
-        waitForCPUIdle.cancel();
+        waitForSecurityCheck.cancel();
```
Hoten: I think it'd be easier to reason about all this canceling if each cancelable promise stores state about if it's been canceled yet and avoids doing it more than once. Then we could judiciously use `cancelAll` in one place, after the Promise race.

patrickhulce: Yeah this is ripe for cleanup, but this new parallel style will make it so much easier now, thank you!!

patrickhulce: FWIW, I don't think anything would be harmed today if you threw this into `cancelAll` right?

Hoten: would also have to remove the self-cancel that `_waitForSecurityCheck` does, otherwise it would cancel itself twice.

I'll try the cancel state thing first

patrickhulce: hang on before you do that I was saying I don't think it would matter if we called `cancel` twice, we're just extra disabling the security domain and removing a listener that might've been removed already.

Hoten: OK, LMK if that is better or worse

lighthouse-core/gather/driver.js
```diff
@@ -698,7 +697,6 @@ class Driver {
     const promise = new Promise((resolve, reject) => {
       checkForQuiet(this, resolve).catch(reject);
       cancel = () => {
-        cancelled = true;
```
patrickhulce: this feels like a bug, we want to just add `if (cancelled) return` right?

lighthouse-core/gather/driver.js
```diff
@@ -793,6 +840,13 @@ class Driver {
     // CPU listener. Resolves when the CPU has been idle for cpuQuietThresholdMs after network idle.
     let waitForCPUIdle = this._waitForNothing();
 
+    const waitForSecurityCheck = this._waitForSecurityCheck();
+    const securityCheckPromise = waitForSecurityCheck.promise.then((err) => {
```
patrickhulce: super nit: but I might have the promise resolve with the explanations and this fn create the error.

that feels a little more mainstream to resolve with data and then the type of the error is clear from within this fn

lighthouse-core/test/gather/driver-test.js
```diff
@@ -22,7 +22,10 @@ const redirectDevtoolsLog = require('../fixtures/wikipedia-redirect.devtoolslog.
 const MAX_WAIT_FOR_PROTOCOL = 20;
 
 function createOnceStub(events) {
-  return (eventName, cb) => {
+  return async (eventName, cb) => {
+    // wait a tick b/c real events never fire immediately
```
patrickhulce: ❤️

lighthouse-core/gather/driver.js
Hoten: oops, didn't meant to delete that.

can skip the guard here since reject/clearTimeout should have no effect after the first call.

lighthouse-core/gather/driver.js
Hoten: done

lighthouse-core/gather/driver.js
Hoten: of course, leaving out the guard _looks_ like a mistake, so perhaps I should just add it anyways?

lighthouse-core/gather/driver.js
```diff
@@ -666,7 +665,7 @@ class Driver {
 
     /** @type {NodeJS.Timer|undefined} */
     let lastTimeout;
-    let cancelled = false;
+    let canceled = false;
```
brendankenny: boooo to one L :P

lighthouse-core/gather/driver.js
brendankenny: > > can skip the guard here since reject/clearTimeout should have no effect after the first call.

> of course, leaving out the guard looks like a mistake, so perhaps I should just add it anyways?

I mean, it's also unnecessary everywhere else it's be added, isn't it? All removing listeners and clearing timeouts, which have no problem being run multiple times.

I'm OK with having them since that's how we would write a cancellable promise utility instead of doing each of these separately (in which case all of the methods should have it), but I'm also OK with dropping them since they aren't necessary (except for the `Security.disable` command in `_waitForSecurityCheck`)

lighthouse-core/gather/driver.js
```diff
@@ -698,9 +697,9 @@ class Driver {
     const promise = new Promise((resolve, reject) => {
       checkForQuiet(this, resolve).catch(reject);
       cancel = () => {
-        cancelled = true;
+        canceled = true;
         if (lastTimeout) clearTimeout(lastTimeout);
-        reject(new Error('Wait for CPU idle cancelled'));
+        reject(new Error('Wait for CPU idle canceled'));
```
brendankenny: interesting that we reject here but not in the others

lighthouse-core/gather/driver.js
```diff
@@ -743,6 +745,45 @@ class Driver {
...
+          cancel();
+          resolve(explanations);
+        }
+      };
+      let canceled = false;
+      cancel = () => {
+        if (canceled) return;
+        canceled = true;
+        this.off('Security.securityStateChanged', securityStateChangedListener);
+        this.sendCommand('Security.disable').catch(() => {});
```
brendankenny: maybe add a TODO here for whatever @patrickhulce is planning with the refactor? `cancel()` should really be a promise itself to handle things like this

lighthouse-core/gather/driver.js
```diff
@@ -743,6 +745,45 @@ class Driver {
     };
   }
 
+  /**
+   * Return a promise that resolves when an insecure security state is encountered
+   * and a method to cancel internal listeners.
+   * @return {{promise: Promise<LH.Crdp.Security.SecurityStateExplanation[]>, cancel: function(): void}}
+   * @private
+   */
+  _waitForSecurityCheck() {
```
brendankenny: rename since it resolves only in the case of *in*security (and the check isn't singular, it just goes until it isn't needed)? `_waitForInsecureState`? `_monitorForInsecureState`? Something else?

lighthouse-core/gather/driver.js
```diff
@@ -816,11 +870,6 @@ class Driver {
     }).then(_ => {
       return async () => {
         log.warn('Driver', 'Timed out waiting for page load. Checking if page is hung...');
-        waitForFCP.cancel();
-        waitForLoadEvent.cancel();
-        waitForNetworkIdle.cancel();
-        waitForCPUIdle.cancel();
-
         if (await this.isPageHung()) {
```
brendankenny: FWIW for future refactoring, @paulirish mentioned he doesn't think this check is necessary anymore (errors are successfully being detected elsewhere, I guess)

lighthouse-core/gather/driver.js
```diff
@@ -793,6 +835,19 @@ class Driver {
...
+      return function() {
+        const insecureDescriptions = explanations
+          .filter(exp => exp.securityState === 'insecure')
+          .map(exp => exp.description);
+        const err = new LHError(LHError.errors.INSECURE_DOCUMENT_REQUEST, {
+          securityMessages: insecureDescriptions.join(' '),
+        });
+        throw err;
+      };
+    });
```
brendankenny: ha, I think I disagree with @patrickhulce. The internals of `SecurityStateExplanation` seem like a distraction from the actual purpose of `_waitForFullyLoaded`, since either way we're using `waitForSecurityCheck` to alert us of an error, the question is if we should do more work in there or out here.

If that's not convincing, maybe we could split the difference? Filter and map within `_waitForSecurityCheck` and then it only needs to be 
```js
const securityCheckPromise = waitForSecurityCheck.promise.then(securityMessages => {
  return function() {
    throw new new LHError(LHError.errors.INSECURE_DOCUMENT_REQUEST, {securityMessages});
  }
});
```

(using the cleanupFn to reject the promise is a little weird but w/e :P)

lighthouse-core/gather/driver.js
```diff
@@ -830,11 +879,20 @@ class Driver {
       };
     });
 
-    // Wait for load or timeout and run the cleanup function the winner returns.
+    // Wait for security issue, load or timeout and run the cleanup function the winner returns.
     const cleanupFn = await Promise.race([
```
brendankenny: having this `cleanupFn` setup now is weird if they all can (and should) be cancelled regardless, but that can wait for the refactor too

lighthouse-core/lib/lh-error.js
brendankenny: I think we need to add something about `securityMessages` to the description for the translators. "securityMessages will be replaced with one or more strings from the browser explaining what was insecure about the page load." or whatever

lighthouse-core/test/gather/driver-test.js
```diff
@@ -22,7 +22,10 @@ const redirectDevtoolsLog = require('../fixtures/wikipedia-redirect.devtoolslog.
 const MAX_WAIT_FOR_PROTOCOL = 20;
 
 function createOnceStub(events) {
-  return (eventName, cb) => {
+  return async (eventName, cb) => {
```
brendankenny: this is a weird use of async/await (and depending on `driver.once` calls you'll end up with an unhandled promise rejection for the throw below).

Can we just
```js
// wait a tick b/c real events never fire immediately
setTimeout(_ => cb(events[eventName]), 0);
return;
```
like normal people? :P

lighthouse-core/gather/driver.js
Hoten: I prefer Mr. Webster's simplified spellings, but I could care more.

lighthouse-core/test/gather/driver-test.js
```diff
@@ -535,4 +543,75 @@ describe('Multiple tab check', () => {
...
     });
   });
+
+  describe('Security check', () => {
+    it('does not reject when page is secure', async () => {
+      const secureSecurityState = {
+        explanations: [],
+        securityState: 'secure',
+      };
+      driverStub.on = driverStub.once = createOnceStub({
```
brendankenny: this poor `driverStub` was never meant to be used repeatedly :)

Can you also break the cycle and make a `new Driver(connection)` within each of these tests? We can deal with having a fresh `connection` later

patrickhulce: FWIW I'm also rather fond of keeping all the mocking close to the test like I did over in manifest PR with `jest.spyOn`. Given the way the past few reviews have gone I'm guessing @brendankenny feels the opposite though so take it with a grain of salt 😂

brendankenny: > Given the way the past few reviews have gone I'm guessing @brendankenny feels the opposite

Ha, I'm not entirely sure how to interpret that. I also like maximizing locality so you can see what's actually been done to the mock you have to use. But I am generally against mocks if at all possible, if that's what that means :)

patrickhulce: > I also like maximizing locality

phew! :) 

> Ha, I'm not entirely sure how to interpret that.

only a joke no worries, just I feel like I've lead @Hoten chasing too many wild geese lately 😆

brendankenny: > only a joke no worries

lol I was trying to read too deeply :) :)

brendankenny:
@Hoten you'll need to rerun `yarn diff:sample-json` to pick up the new description

lighthouse-core/test/gather/driver-test.js
brendankenny: I meant drop the async/await in favor of the setTimeout :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -95,23 +96,35 @@ connection.sendCommand = function(command, params) {
...
+}
 
 /* eslint-env jest */
 
+let driverStub;
+beforeEach(() => {
+  const connection = new Connection();
+  connection.sendCommand = sendCommandStub;
+  driverStub = new Driver(connection);
+});
```
brendankenny: ooh nice

brendankenny:
itshappening

Hoten:
image

Hoten:
tumblr_o64l4xtank1tbh1dho1_400

PR #6618, opened 2018-11-20
misc: rename core/gather/computed/ to core/computed/

brendankenny:
part of #6202

~Also fixes a couple things:~
- ~log timing was added at the same time that #6204 updated computed artifacts, so timing info hasn't actually been collected for computed artifacts (timing was still accounted for in the audits that called them, though). This PR adds timing to the new version.~
- ~#6204 messed up and was using strict equality testing for the computed artifact cache, not deep equality. Computed artifacts that take a bare artifact (e.g. `trace-of-tab`) had no issue with this and cached correctly. Any computed artifact that take a more complicated input (e.g. all the metrics) always found keys to be unequal, and so reran their computations every time they were called. This was especially bad in the lantern metrics, which tend to call each other a lot and benefit a good deal from caching.~

  ~This PR defaults `ArbitraryEqualityMap` to using deep equality now (since that's the only way we use it), and the kind of goofball test updated in `computed-artifact-test.js` proved it's worth and will prevent this from regressing again.~

~Good news: runs should speed up :)~

Everything else is path updates.


brendankenny:
ok, we're going to split the PR, functional changes to follow

PR #6619, opened 2018-11-20
core(computed-artifacts): fix cache and add perf timing

brendankenny:
split off from #6618

- log timing was added at the same time that #6204 updated computed artifacts, so timing info hasn't actually been collected for computed artifacts (timing was still accounted for in the audits that called them, though). This PR adds timing to the new version.
- #6204 messed up and was using strict equality testing for the computed artifact cache, not deep equality. Computed artifacts that take a bare artifact (e.g. `trace-of-tab`) had no issue with this and cached correctly. Any computed artifact that take a more complicated input (e.g. all the metrics) always found keys to be unequal, and so reran their computations every time they were called. This was especially bad in the lantern metrics, which tend to call each other a lot and benefit a good deal from caching.

  This PR defaults `ArbitraryEqualityMap` to using deep equality now (since that's the only way we use it), and the kind of goofball test updated in `computed-artifact-test.js` proved it's worth and will prevent this from regressing again.

Good news: runs should speed up :)

### DIFF
--- a/lighthouse-core/computed/computed-artifact.js
+++ b/lighthouse-core/computed/computed-artifact.js
@@ -6,6 +6,7 @@
 'use strict';
 
 const ArbitraryEqualityMap = require('../lib/arbitrary-equality-map.js');
+const log = require('lighthouse-logger');
 
 /**
  * Decorate computableArtifact with a caching `request()` method which will
@@ -25,17 +26,24 @@ function makeComputedArtifact(computableArtifact) {
    */
   const request = (artifacts, context) => {
     const computedCache = context.computedCache;
-    const cache = computedCache.get(computableArtifact.name) || new ArbitraryEqualityMap();
-    computedCache.set(computableArtifact.name, cache);
+    const computedName = computableArtifact.name;
+
+    const cache = computedCache.get(computedName) || new ArbitraryEqualityMap();
+    computedCache.set(computedName, cache);
 
     const computed = /** @type {ReturnType<C['compute_']>|undefined} */ (cache.get(artifacts));
     if (computed) {
       return computed;
     }
 
+    const status = {msg: `Computing artifact: ${computedName}`, id: `lh:computed:${computedName}`};
+    log.time(status, 'verbose');
+
     const artifactPromise = /** @type {ReturnType<C['compute_']>} */
         (computableArtifact.compute_(artifacts, context));
     cache.set(artifacts, artifactPromise);
+
+    artifactPromise.then(() => log.timeEnd(status)).catch(() => log.timeEnd(status));
 
     return artifactPromise;
   };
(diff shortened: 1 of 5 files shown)


lighthouse-core/test/computed/computed-artifact-test.js
```diff
@@ -12,7 +12,7 @@ const assert = require('assert');
 const makeComputedArtifact = require('../../computed/computed-artifact.js');
 
 describe('ComputedArtifact base class', () => {
-  it.skip('caches computed artifacts by strict equality', async () => {
+  it('caches computed artifacts by strict equality', async () => {
     let computeCounter = 0;
 
     const TestComputedArtifact = makeComputedArtifact(class TestComputedArtifact {
```
paulirish: class as expression hurts my head, especially since it has the same name as something different
can you extract?

lighthouse-core/computed/computed-artifact.js
```diff
@@ -25,18 +26,25 @@ function makeComputedArtifact(computableArtifact) {
...
     }
 
+    const status = {msg: `Computing artifact: ${computedName}`, id: `lh:computed:${computedName}`};
+    log.time(status, 'verbose');
+
     const artifactPromise = /** @type {ReturnType<C['compute_']>} */
         (computableArtifact.compute_(artifacts, context));
     cache.set(artifacts, artifactPromise);
 
+    artifactPromise.then(() => log.timeEnd(status)).catch(() => log.timeEnd(status));
```
brendankenny: this is unfortunate until we have `Promise.finally` :)

Otherwise if there's a error thrown in the computed artifact this branch is considered an unhandled rejection and we try to kill Lighthouse early.

brendankenny:
> Good news: runs should speed up :)

effect isn't huge huge, but it helps.

An `-A` run on some saved verge artifacts drops about 11%, from 2.65s to 2.35s on my machine.

lighthouse-core/test/computed/computed-artifact-test.js
```diff
@@ -12,19 +12,19 @@ const assert = require('assert');
 const makeComputedArtifact = require('../../computed/computed-artifact.js');
 
 describe('ComputedArtifact base class', () => {
-  it.skip('caches computed artifacts by strict equality', async () => {
```
patrickhulce: oh I see now :)

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3474,6 +3474,12 @@
         "duration": 100,
         "entryType": "measure"
       },
+      {
+        "startTime": 0,
+        "name": "lh:computed:NetworkRecords",
```
patrickhulce: ❤️

PR #6622, opened 2018-11-21
refactor: Only require html report templates when generating html report

Munter:
**Summary**

Moved importing of html report templates to happen lazily at the time when a html report is being generated.

Reading html template files from disk when not generating html-reports is wasteful. It's also a lot easier to configure a bundler for a serverless deployment if the runtime isn't reading files on disk that are not in the module dendendency graph.

Case in point: I'm bundling up my lighthouse with webpack to minimize my lambda deployment package, but am running into errors because these template files are not included in the dependency graph. I'm only generating a json report, so I really don't need them

### DIFF
--- a/lighthouse-core/report/report-generator.js
+++ b/lighthouse-core/report/report-generator.js
@@ -5,8 +5,6 @@
  */
 'use strict';
 
-const htmlReportAssets = require('./html/html-report-assets');
-
 class ReportGenerator {
   /**
    * Replaces all the specified strings in source without serial replacements.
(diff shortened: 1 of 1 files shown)


patrickhulce:
Thanks for opening @Munter! I'm not sure we're going to be in the business of optimizing our codebase for other bundlers though, just FYI.

Seems like either way you'd want to stub out this HTML generating logic from your bundle anyhow? Does moving it to a synchronous require farther down really prevent the code from being bundled? It seems like that would be impossible to satisfy.

Munter:
@patrickhulce the code gets bundled, but not executed, so the calls to read files never happen.

I do have to stub this out at the moment, but it's not a good solution, nor is it very stable

brendankenny:
I think @patrickhulce is getting at that however it gets pulled in by `require()`, it'll have to be stubbed out. If it's just an empty stub either way, its impact on the bundle size is minimal and which way it's used will just affect where the stub ends up in the output.

Is the build set up something you can share? Maybe there are specifics we're missing.

brendankenny:
Going to close for now, but we can reopen or discuss in an issue if this is something you still want to pursue, @Munter

Issue #6598, opened 2018-11-17
Make image links clickable


jadjoubran:
**Feature request summary**
In audits relating to images, it would be nice to make the image links **clickable** which automatically takes to the image that's failing the audit. 
![screen shot 2018-11-17 at 13 24 28](https://user-images.githubusercontent.com/2265232/48661042-45031e80-ea6c-11e8-8f4b-fce91b29e6a6.png)
Another alternative would be to have a "link opens in new tab" (top-right arrow) icon that will open the image

I would love to work on this new feature

**What is the motivation or use case for changing this?**
When debugging sites with large and/or dynamic images, this makes it much easier to find (and get a link) of the image that's failing the audit.
Most of the times, image links are quite long so they get trimmed and you can't copy them and then you have to find them in the Network tab

**How is this beneficial to Lighthouse?**
Improved Developer Experience 😄

patrickhulce:
Should be easy enough thanks for filing and volunteering @jadjoubran !

The code you're looking for is 
https://github.com/GoogleChrome/lighthouse/blob/7f55977b2a8aa13633c95a1d97b546b0bd74c8be/lighthouse-core/report/html/renderer/details-renderer.js#L193-L206

You'll probably want to wrap that in an `a` and set the href/target to the URL/_blank.

jadjoubran:
maybe then it makes sense to link the **trimmed link** (screenshot below) to the image (opens in a new tab)
<img width="177" alt="screen shot 2018-11-19 at 12 57 26" src="https://user-images.githubusercontent.com/2265232/48705864-b670d780-ebfa-11e8-9dc1-e9bdfff76226.png">

and then a future PR would link the **image preview** (screenshot below) to the DOM element
<img width="42" alt="screen shot 2018-11-19 at 12 57 29" src="https://user-images.githubusercontent.com/2265232/48705890-c8eb1100-ebfa-11e8-9bbf-7667bc1c6e33.png">

What do you think?

patrickhulce:
We've been thinking about making the trimmed link link out for a while now across the board, so that'd be cool!

Linking to the DOM element has always been somewhat trickier but very useful. Definitely start with the trimmed link, and we can handle DOM element next :)

jadjoubran:
Sounds good, just to double check and avoid confusion (because you linked me to the thumbnail)
trimmed link => should link to the image in a new tab
correct?

Hoten:
[...]

Issue #5443, opened 2018-06-06
Linkify paths to resources flagged in the report


addyosmani:
## Feature request

**Summary**

Audits currently list paths to resources that may have an issue. Unfortunately, there is no way to 1) click the text for a path to display the resource in the Network/Sources panel or 2) click to view the full resource in a new tab (if 1. is not possible)

![image](https://user-images.githubusercontent.com/110953/41070306-e547b2d0-69a6-11e8-8ccc-343eeacf25c4.png)

![image](https://user-images.githubusercontent.com/110953/41070392-88c5fcb4-69a7-11e8-80da-6c5b6a5ff2f1.png)

This makes debugging a little trickier as you need to switch to a different panel in DevTools and manually filter down to find the resource you are after. It would be nice if this was something we made a little lower friction :)

**What is the motivation or use case for changing this?**

Time-saving. When LH flags a list of resources, it's currently cumbersome navigating back and forth to filter down different resources to dig in deeper to a problem. 

**How is this beneficial to Ligthhouse?**

Deeper integration with the rest of DevTools provides a smoother debugging experience for developers wanting to more closely analyze why a resource is problematic.

patrickhulce:
Totally agree linking over to network panel in DT would be awesome 👍 the fact that we're looking at a different page load by the time the report is given to you though really complicates things.

Your 2nd suggestion (turning them into links to open the resource in a new tab) has come up a few times and could be done really quickly/easily, but I've always wondered the provided value. I guess the primary argument is that you can open the DevTools network tab on that resource to take a look at the same stats more easily even if looking at the JS itself isn't helpful? If so, I can do that today :)

Alternative approach would be something like https://github.com/GoogleChrome/lighthouse/issues/1975#issuecomment-394091169 where we show the full URL in a tooltip on hover for easy copying? Not sure which is better.

addyosmani:
> Your 2nd suggestion (turning them into links to open the resource in a new tab) has come up a few times and could be done really quickly/easily, but I've always wondered the provided value. I guess the primary argument is that you can open the DevTools network tab on that resource to take a look at the same stats more easily even if looking at the JS itself isn't helpful? If so, I can do that today :)

I personally think there's a lot of value in this direction with the network panel (vs showing the full URL in a tooltip). With the tooltip approach, it would give you more of the path but there's then an exercise to be done navigating to the Network panel and still trying to filter down to it yourself. I think the idea above is a wee bit nicer :)

[...]

PR #6625, opened 2018-11-21
misc: make image links clickable

jadjoubran:
**Summary**

Feature: This PR makes images clickable (only next-gen images audit for now).
I'm opening the PR now to get feedback before I enable it for other image related audits

I'd like to get feedback on my code, style of the URL, should I shorten the URL and only show a portion of it?

**Related Issues/PRs**
Related to #6598

**Screenshot**
screen shot 2018-11-21 at 17 47 50

### DIFF
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -166,7 +166,6 @@ class DetailsRenderer {
     a.target = '_blank';
     a.textContent = details.text;
     a.href = url.href;
-
     return a;
   }
 
@@ -311,6 +310,10 @@ class DetailsRenderer {
               const codeValue = /** @type {(number|string|undefined)} */ (value);
               itemElement = this._renderCode({value: codeValue});
             }
+            break;
+          }
+          case 'link': {
+            itemElement = this._renderLink(/** @type {LinkDetailsJSON} */ {url: value, text: value})
             break;
           }
           case 'timespanMs': {
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -166,6 +166,7 @@ class DetailsRenderer {
     a.target = '_blank';
     a.textContent = details.text;
     a.href = url.href;
+
     return a;
   }
 
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -314,7 +314,7 @@ class DetailsRenderer {
             break;
           }
           case 'link': {
-            itemElement = this._renderLink(/** @type {LinkDetailsJSON} */ {url: value, text: value})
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -313,6 +313,10 @@ class DetailsRenderer {
             }
             break;
           }
+          case 'link': {
+            itemElement = this._renderLink({url: value, text: value});
```
exterkamp: Need to add some typescript typings here for stringifying this value.
```javascript
const strValue = /** @type {string} */ (value);
itemElement = this._renderLink({type: 'link', url: strValue, text: strValue});
```

patrickhulce:
> trim data URI's to a nice length and those kinds of url's aren't linkable anyway, so they need to probably stay as a url type, or be handled in _renderLink as a special case? Maybe we need a test around this?

`_renderLink` should already not link URLs that don't start with `http(s)?`, so I *think* we're OK on that front.

I might intuitively want to click on the image to open it up full screen, but linking the text is inline with a vision to turn all URLs into links (#5443), so I'm fine either way.

lighthouse-core/report/html/renderer/details-renderer.js
agomez314: Linter recommended me to add a type in the following interface in `lhr-lite.d.ts`:
```
 export interface ColumnHeading {
          /** The property key name within DetailsItem being described. */
          key: string;
          /** Readable text label of the field. */
          label: string;
          // TODO(bckenny): should be just string and let lhr be more specific?
          valueType: 'url' | 'timespanMs' | 'bytes' | 'thumbnail' | 'link';
        }
```
And it no longer gave me an error. Perhaps this could work?

paulirish:
One alternative I'm thinking of.... All `type: 'url'` items get a little blue external-link icon placed to the right of them. And the user can click if they want.

Does that satisfy the original UX usecases, @jadjoubran ?

jadjoubran:
Yes that satisfies the original UX usecase 👍
can I get some updated points for the implementation? so I can update the PR
Thanks!

Hoten:
+100 for @paulirish's alternative. for every single url, it can be expected that the user will want to click through to the resource. example: me, just now, with this audit;

image

@jadjoubran you will want to modify the element created [here](https://github.com/GoogleChrome/lighthouse/blob/6b9b9a6971935fe50ccff226e6cecc17f81c884c/lighthouse-core/report/html/renderer/details-renderer.js#L133). I'm not aware of an "external link" icon already in use or in one of our fonts - although I'm not sure why we can't just make the `...some/url.png` the link?

If we go this route, you'll want to remove all of your current changes.

jadjoubran:
Alright sounds good from my side, I'll wait for further feedback before implementing 👍

brendankenny:
> I'll wait for further feedback before implementing 👍

there has been even more rethinking about how to linkify URLs in the report, so let's close for now and reopen when we nail down exactly what we want to do.

PR #6630, opened 2018-11-22
core: rename webapp-install-banner audit to installable-manifest

brendankenny:
Part of #6395 

Updates `webapp-install-banner` to be the new `installable-manifest` audit. SW detection was already deleted from here, so this mostly just updates the `id` and `title` and tightens some tests/removes some useless stuff.

### DIFF
--- a/proto/sample_v2_round_trip.json
+++ b/proto/sample_v2_round_trip.json
@@ -933,6 +933,25 @@
             "scoreDisplayMode": "not_applicable", 
             "title": "`<input type=\"image\">` elements have `[alt]` text"
         }, 
+        "installable-manifest": {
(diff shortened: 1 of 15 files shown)


proto/sample_v2_round_trip.json
```diff
@@ -933,6 +933,25 @@
             "scoreDisplayMode": "not_applicable", 
             "title": "`<input type=\"image\">` elements have `[alt]` text"
         }, 
+        "installable-manifest": {
```
patrickhulce: interesting why did only these two swap order?

brendankenny: looks like they're alphabetized. No idea why, but I guess it doesn't matter :) @exterkamp

PR #6637, opened 2018-11-23
report(tsc): infer createElement type from tag name

brendankenny:
This is a little silly, but there's a small quality of life improvement, it closes an old TODO, and it eliminates a decent number of type assertions.

Bases the created element type on the tag name passed in, so e.g. `DOM.createElement('a')` returns an `HTMLAnchorElement`, not just a generic `Element`. If the tag is unknown, falls back to just `HTMLElement`, which is close enough to correct for us today.

Uses the [same tag-name-to-element-type map that tsc uses for `document.createElement`](https://github.com/Microsoft/TypeScript/blob/3e4c5c95abd515eb9713b881d27ab3a93cc00461/lib/lib.dom.d.ts#L4263-L4270), just with a union for the `HTMLElement` fallback instead of the function overload they use (and since `DOM.createElement` calls `document.createElement` internally, this type is checked by their type).

### DIFF
--- a/lighthouse-core/report/html/renderer/dom.js
+++ b/lighthouse-core/report/html/renderer/dom.js
@@ -18,6 +18,8 @@
 
 /* globals URL self */
 
+/** @typedef {HTMLElementTagNameMap & {[id: string]: HTMLElement}} HTMLElmentByTagName */
+
 class DOM {
   /**
    * @param {Document} document
@@ -34,7 +36,7 @@ class DOM {
    * @param {Object<string, (string|undefined)>=} attrs Attribute key/val pairs.
    *     Note: if an attribute key has an undefined value, this method does not
    *     set the attribute on the node.
-   * @return {(HTMLElementTagNameMap & {[id: string]: HTMLElement})[T]}
+   * @return {HTMLElmentByTagName[T]}
    */
   createElement(name, className, attrs = {}) {
     const element = this._document.createElement(name);
@@ -65,7 +67,7 @@ class DOM {
    * @param {Object<string, (string|undefined)>=} attrs Attribute key/val pairs.
    *     Note: if an attribute key has an undefined value, this method does not
    *     set the attribute on the node.
-   * @return {(HTMLElementTagNameMap & {[id: string]: HTMLElement})[T]}
+   * @return {HTMLElmentByTagName[T]}
    */
   createChildOf(parentElem, elementName, className, attrs) {
     const element = this.createElement(elementName, className, attrs);
--- a/lighthouse-core/report/html/renderer/dom.js
+++ b/lighthouse-core/report/html/renderer/dom.js
@@ -27,14 +27,14 @@ class DOM {
     this._document = document;
   }
 
-  // TODO(bckenny): can pass along `createElement`'s inferred type
   /**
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -364,8 +364,7 @@ class ReportUIFeatures {
     // load event, however it is cross-domain and won't fire. Instead, listen
     // for a message from the target app saying "I'm open".
     const json = reportJson;
-    window.addEventListener('message', function msgHandler(/** @type {Event} */ e) {
-      const messageEvent = /** @type {MessageEvent} */ (e);
```
brendankenny: no idea what was going on here. Maybe fixed on the compiler/built-in defs side.

lighthouse-core/report/html/renderer/dom.js
```diff
@@ -18,6 +18,8 @@
 
 /* globals URL self */
 
+/** @typedef {HTMLElementTagNameMap & {[id: string]: HTMLElement}} HTMLElmentByTagName */
```
patrickhulce: `HTMLElementTagNameMap` is the built-in you're talking about?

lighthouse-core/report/html/renderer/dom.js
```diff
@@ -27,14 +29,14 @@ class DOM {
     this._document = document;
   }
 
-  // TODO(bckenny): can pass along `createElement`'s inferred type
   /**
-   * @param {string} name
+   * @template {string} T
+   * @param {T} name
```
patrickhulce: can we type this as the keys of the `HTMLElementTagNameMap`?

lighthouse-core/report/html/renderer/dom.js
brendankenny: > `HTMLElementTagNameMap` is the built-in you're talking about?

yes

lighthouse-core/report/html/renderer/dom.js
brendankenny: > can we type this as the keys of the `HTMLElementTagNameMap`?

We could, though it would eliminate the fallback to the generic `HTMLElement` when a tag name isn't recognized. Maybe we're ok with that in exchange for better completion/stricter checking?

patrickhulce: would we still get helpful completion if it were strict options | string? :)

brendankenny: > strict options | string

sadly no, it's smart about it and coalesces to just `string`.

brendankenny: ok, I tried to use just `HTMLElementTagNameMap`, and we have one place that's unrecognized: `'summary'` in

https://github.com/GoogleChrome/lighthouse/blob/a203a36026c722895d3c885e282adaf87225f19e/lighthouse-core/report/html/renderer/category-renderer.js#L191

When Edge gets `<details>` support I imagine that they'll also support `<summary>` and their WebIDL will get updated and so the tsc types will get updated (I'm surprised it already has `'details'` support, honestly), but until then we probably just want to go with this as this PR is now.

Issue #5527, opened 2018-06-20
LH3:  does not register a service worker on first report.


kscm:
#### Provide the steps to reproduce
1. Run LH on https://kinoseed.com/

#### What is the current behavior?
LH does not register site's service worker **on the first run** ("generate report").
Running "Generate report" after the first run, performs as expected.

#### What is the expected behavior?
LH should register site's service worker on the first run.

#### Environment Information
* Affected Channels: 
* Lighthouse version: Lighthouse 3.0.0-beta.0
* Node.js version:
* Operating System:

Runtime settings
URL: https://kinoseed.com/
Fetch time: Jun 20, 2018, 11:54 AM GMT+3
Device: Emulated Nexus 5X
Network throttling: 150 ms TCP RTT, 1,638.4 Kbps throughput (Simulated)
CPU throttling: 4x slowdown (Simulated)
User agent: Mozilla/5.0 (Windows NT 6.2; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3359.139 Safari/537.36
Generated by Lighthouse 3.0.0-beta.0 

**Related issues**
Since LH does not register the service worker, PWA checklist fails. Extract from the "report":

Progressive Web App
These checks validate the aspects of a Progressive Web App, as specified by the baseline PWA Checklist.
1
Does not register a service worker
2
Does not respond with a 200 when offline
3
User will not be prompted to Install the Web App
Failures: Site does not register a service worker.

patrickhulce:
I'm not able to reproduce this behavior. On the first run, LH found a service worker.
![image](https://user-images.githubusercontent.com/2301202/41731707-f1ab8df6-7533-11e8-9163-8179fef373ed.png)

If you wait a while to register a service worker, then it could be flaky based on if your registration happened yet before we moved on.

kscm:
@patrickhulce , in your case the "manifest" is missing?

I'm not sure what is going on, but this is what I get:
![lherror](https://user-images.githubusercontent.com/28153166/41734464-b29e57e0-758f-11e8-86a7-34824866297f.jpg)

prash-mi:
+1 for this issue. Finding the same with https://m.jungleerummy.com/ . Lighthouse is reporting that the site doesn't register service worker, but in-fact it does on the landing page load itself. Attached the lighthouse report and SW registration registration screenshots.

[...]

Issue #2924, opened 2017-08-10
Document that service worker needs to be registered early


zoumiaojiang:
URL: https://zoumiaojiang.win
Chrome Version: 60.0.3112.90
Lighthouse Version:  2.3.0

![a3e11ce6773167aed1c120b42211a230](https://user-images.githubusercontent.com/3365978/29154497-8efc9d1c-7dc7-11e7-8766-2bcf4423f344.png)

PS: Service Worker registration is asynchronous

patrickhulce:
Hm, this looks like another case of #599 resurfacing. The registration takes quite some time after network quiet. Though I'm not sure what we can reasonably do about these cases. 5s is already a pretty long time to give room for the service worker to register, and we don't want to wait needlessly for a service worker that will never be registered either. Ideas welcome :)

canarslan12:
I think title of error can be changed to "Does not register a Service Worker or its registration is too slow" and add to information that "registration might be faster" etc

patrickhulce:
That's a good call @canarslan12, let's put that information in the Learn More link. 👍

paulirish:
We identified theres two possible cases here:

1. You called register() very late in the page load cycle. (after Lighthouse detected the page is totally loaded (or after 30 seconds))
1. You called register() early but then the browser was really busy loading the page and never got around to starting the registration/activation process.

patrickhulce:
FWIW, several URLs in #5527 appear to add a third possible case given our new lower 500ms network quiet threshold with Lantern,

3. The service worker was in the process of registering but didn't get to finish before we moved on.

brendankenny:
> You called register() early but then the browser was really busy loading the page and never got around to starting the registration/activation process.

I think this is ruled out, too. The test added in #6648 doesn't strictly keep the page busy, but in my testing, as soon as `navigator.serviceWorker.register()` is called the SW is set to `installing` and it's picked up with a `ServiceWorker.workerVersionUpdated` event, so I believe main thread activity is irrelevant to knowing that registration has started.

At least as of current Chrome (m72), SW installation continues to happen at that point, even if you navigate away from the page before the first script bytes have downloaded.

I had forgotten how this worked. [A long time ago](https://github.com/GoogleChrome/lighthouse/pull/1864), @wardpeet updated the SW version fetcher to wait if the page had no activated service workers but there were SWs in any sort of installation phase:

[...]

PR #6648, opened 2018-11-26
tests: add smoketest for slow service worker

patrickhulce:
**Summary**
Curious what types of late service worker we don't catch. I saw @paulirish's comment that it's just calling `.register` but I was skeptical based on the other reports. I should not have been skeptical it lives up, and this just adds a smoketest to prove it. We wait for delays in loading the service worker script itself and delays in the install event too.

**Related Issues/PRs**
https://github.com/GoogleChrome/lighthouse/issues/2924#issuecomment-332989460
#5527

### DIFF
--- a/lighthouse-cli/test/fixtures/offline-ready.html
+++ b/lighthouse-cli/test/fixtures/offline-ready.html
@@ -24,10 +24,6 @@
 </h4>
 
 <script>
-  function sleep(ms) {
-    return new Promise(resolve => setTimeout(resolve, ms))
-  }
-
   async function registerServiceWorker(qs = '') {
     if (!('serviceWorker' in navigator)) return
 
@@ -45,7 +41,7 @@
   }
 
   if (window.location.search.includes('slow')) {
-    setTimeout(() => registerServiceWorker('?delay=5000&slow'), 500);
+    window.addEventListener('load', () => registerServiceWorker('?delay=5000&slow'));
   } else {
     registerServiceWorker();
   }
--- a/lighthouse-cli/test/fixtures/offline-ready.html
+++ b/lighthouse-cli/test/fixtures/offline-ready.html
@@ -24,18 +24,32 @@
 </h4>
 
 <script>
-  if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.register('/offline-ready-sw.js').then(function(registration) {
+  function sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms))
+  }
+
+  async function registerServiceWorker(qs = '') {
+    if (!('serviceWorker' in navigator)) return
+
+    try {
+      const registration = await navigator.serviceWorker.register(`/offline-ready-sw.js${qs}`)
       console.log('service worker registration complete');
 
       registration.addEventListener('statechange',  e => {
         console.log('sw registration is now', e.target.state);
       });
-    }).catch(function(e) {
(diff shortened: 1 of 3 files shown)


lighthouse-cli/test/fixtures/offline-ready.html
```diff
@@ -24,18 +24,32 @@ <h4>
...
-    }).catch(function(e) {
+    } catch (e) {
       console.error('service worker is not so cool.', e);
       throw e;
-    });
+    }
   }
+
+  if (window.location.search.includes('slow')) {
+    setTimeout(() => registerServiceWorker('?delay=5000&slow'), 500);
```
Hoten: would it make any difference to do this on "load", where the load event is deferred by a significant amount of time?

patrickhulce: This `setTimeout` is mostly useless I can actually remove it since it's complicating the understanding of what's going on. Switching to fire on the load event is a good idea 👍 

the load event is currently being delayed by an image farther down that takes 2s, the only way LH doesn't catch service worker AFAICT is if the page doesn't call `.register` quickly enough, and this test was meant to prove that fact.

Hoten: AFAIK this (registering SW on "load") is basically what pintrest.com (which LH says has no SW) does. not sure if the key difference is how long until "load" fires - or something else entirely.

patrickhulce: If that's all they're doing, then I'm not sure what's going on there. We always wait until load so unless their load event handlers take a very long time such that we don't get to `.register` in time, then it must be something else.

patrickhulce: I can take a look closer at pinterest if there's not already too many cooks in the kitchen :)

Hoten: There's at most 1 :) IDK if ward ever started on it.

patrickhulce: Haha, alright I'll poke around.

lighthouse-cli/test/fixtures/offline-ready.html
```diff
@@ -24,18 +24,28 @@ <h4>
...
+      const registration = await navigator.serviceWorker.register(`/offline-ready-sw.js${qs}`)
       console.log('service worker registration complete');
 
       registration.addEventListener('statechange',  e => {
         console.log('sw registration is now', e.target.state);
       });
-    }).catch(function(e) {
+    } catch (e) {
       console.error('service worker is not so cool.', e);
       throw e;
```
brendankenny: do we check for this anywhere? Or is the hope that a smoke test will fail and this would be used for debugging.

Maybe we should be pulling in `errors-in-console` into all our smoke tests :)

patrickhulce:
> so just to be clear, the intention is this is a slow case that we do indeed handle and this will ensure we don't regress?

yeppers

lighthouse-cli/test/fixtures/offline-ready.html
patrickhulce: > do we check for this anywhere?

nope

> Or is the hope that a smoke test will fail and this would be used for debugging.

Bingo!

> Maybe we should be pulling in errors-in-console into all our smoke tests :)

Not a bad idea, followup! :D

PR #6671, opened 2018-11-28
core(gather-runner): gracefully recover from fatal errors

exterkamp:
**Summary**
Currently a fatal error like PROTOCOL_ERROR will surface as a top level fatal error with no LHR, which is problematic for LR to detect when it is happening, since nothing comes back.  This makes it so that any fatal errors are thrown and caught and triage'd with some basic artifacts and sent back as an errored LHR instead of throwing as a top level exception. 

**~~Related~~ Issues ~~/PRs~~**
In it's current implementation, for simplicity, pageLoadErrors also present in this fashion.  But the catching & throwing out of collected artifacts that might have been collected prior to a fatal error leads to a loss of info (e.g. http-status-code being null instead of detecting it was a 403) there are ways around this (e.g. scanning the err.friendlyMessage for the status code like web.dev) but that isn't exactly clean.  We also could reuse the artifacts that were gathered prior to the fatal error being thrown and just _add_ the error stuff to it so we don't lose any info, but this doesn't work for when the generating of baseArtifacts is _what_ caused the error (e.g. protocol timeout on getting browser version).

### DIFF
--- a/lighthouse-cli/test/smokehouse/seo/expectations.js
+++ b/lighthouse-cli/test/smokehouse/seo/expectations.js
@@ -142,7 +142,8 @@ module.exports = [
     finalUrl: BASE_URL + 'seo-failure-cases.html?status_code=403',
     audits: {
       'http-status-code': {
-        score: null,
+        score: 0,
+        displayValue: '403',
       },
       'viewport': {
         score: null,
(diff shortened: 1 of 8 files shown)


patrickhulce:
+1 to @brendankenny 's suggestion.

Also, I'm a bit confused at this point what the consensus and terminology are around our error strategy, (i.e. #6336). I probably missed a discussion at some point :)

There is "fatal error", "runtime error", and "page load error", along with regular audit and gatherer errors. IIUC, in those terms, this PR...
1. Converts "page load error" into "fatal error"
2. Converts all "fatal error"s into "runtime error"s

Is that right?

Page load errors used to be fatal and we converted them to gatherer errors because of https://github.com/GoogleChrome/lighthouse/issues/1978, so if we're going back we'll need a mechanism to still handle those failures.

exterkamp:
I like the idea of this PR, but since #6739 handles what this was meant to handle I'm going to close it for now.

PR #6678, opened 2018-11-29
core(service-worker): check that start_url is within SW's scope

brendankenny:
part of #6395

Adds a check to the `service-worker` audit to verify that the `start_url` is within the SW's scope.

This joins the #6609 check that the page itself is in scope. As with that check, this is mostly useful when first setting up the SW or someone accidently alters the scope somehow. The explanation for this audit will help with debugging other failures (e.g. offline).

### DIFF
--- a/lighthouse-core/audits/service-worker.js
+++ b/lighthouse-core/audits/service-worker.js
@@ -15,13 +15,71 @@ class ServiceWorker extends Audit {
   static get meta() {
     return {
       id: 'service-worker',
-      title: 'Registers a service worker',
-      failureTitle: 'Does not register a service worker',
+      title: 'Registers a service worker that controls page and start_url',
+      failureTitle: 'Does not register a service worker that controls page and start_url',
       description: 'The service worker is the technology that enables your app to use many ' +
          'Progressive Web App features, such as offline, add to homescreen, and push ' +
          'notifications. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/registered-service-worker).',
-      requiredArtifacts: ['URL', 'ServiceWorker'],
+      requiredArtifacts: ['URL', 'ServiceWorker', 'Manifest'],
     };
+  }
+
+  /**
+   * Find active service workers for this origin.
+   * @param {Array<LH.Crdp.ServiceWorker.ServiceWorkerVersion>} versions
+   * @param {URL} pageUrl
+   * @return {Array<LH.Crdp.ServiceWorker.ServiceWorkerVersion>}
+   */
+  static getVersionsForOrigin(versions, pageUrl) {
+    return versions
+      .filter(v => v.status === 'activated')
+      .filter(v => new URL(v.scriptURL).origin === pageUrl.origin);
+  }
+
+  /**
+   * From the set of active service workers for this origin, find the controlling SW (if any)
+   * and return its scope URL.
+   * @param {Array<LH.Crdp.ServiceWorker.ServiceWorkerVersion>} matchingSWVersions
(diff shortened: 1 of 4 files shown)


lighthouse-core/audits/service-worker.js
```diff
@@ -15,44 +15,109 @@ class ServiceWorker extends Audit {
...
+    const controllingScopeUrl = ServiceWorker.getControllingScopeUrl(versionsForOrigin,
+        registrations, pageUrl);
+    if (!controllingScopeUrl) {
+      return {
+        rawValue: false,
+        explanation: `This origin has one or more service workers, however the page ("${pageUrl.href}") is not in scope.`, // eslint-disable-line max-len
+      };
+    }
+
+    const startUrlFailure = ServiceWorker.checkStartUrl(artifacts.Manifest, controllingScopeUrl);
```
patrickhulce: wouldn't we just want to check that *a* service worker controls the start URL? i.e. if the one controlling this page is more specific but the start URL is the homepage and so it's controlled by a root service worker for example.

lighthouse-core/test/audits/service-worker-test.js
```diff
@@ -75,45 +98,91 @@ describe('Offline: service worker audit', () => {
...
     const swOpts = [{
       status: 'activated',
       scriptURL: 'https://example.com/serviceworker/sw.js',
     }];
+    const manifest = {start_url: finalUrl};
 
-    const output = ServiceWorker.audit(createArtifacts(swOpts, finalUrl));
-    assert.equal(output.rawValue, false);
+    const output = ServiceWorker.audit(createArtifacts(swOpts, finalUrl, manifest));
+    expect(output).toMatchObject({
```
patrickhulce: 🎉

lighthouse-core/test/audits/service-worker-test.js
```diff
@@ -75,45 +98,91 @@ describe('Offline: service worker audit', () => {
...
+      rawValue: false,
+      explanation: expect.stringMatching(new RegExp(`${finalUrl}.*not in scope`)),
```
patrickhulce: well TIL! `stringMatching` and `arrayMatching` are dope

lighthouse-core/test/audits/service-worker-test.js
```diff
@@ -165,8 +236,66 @@ describe('Offline: service worker audit', () => {
...
+    const manifest = {start_url: finalUrl};
+
+    const output = ServiceWorker.audit(createArtifacts(swOpts, finalUrl, manifest));
+    expect(output).toMatchObject({
+      rawValue: false,
+      explanation: expect.stringMatching(new RegExp(`${finalUrl}.*not in scope`)),
+    });
+  });
+
+  it('fails when SW that controls start_url is different than SW that controls page', () => {
```
patrickhulce: oh it seems like it was intentional then, why is that?

lighthouse-core/audits/service-worker.js
brendankenny: > wouldn't we just want to check that *a* service worker controls the start URL? 

Yeah, we may need to think about this a bit. I originally had that check, and I don't think there's any technical problem with it, but then was thinking of the issues if you're testing a page and the declared web app manifest creating the PWA containing that page had a `start_url` that was controlled by a different SW than the page...so I guess the `start_url` is part of a different PWA than the PWA the test page is part of?

So then that seems to violate testing the page as part of *a* PWA. But also I don't know. But also this should only be possible when running with `--disable-storage-reset` anyways, so it's fine to be tougher or not tougher on them :)

patrickhulce: Ah, OK I buy that! SGTM

PR #6684, opened 2018-11-29
deps: update to tsc 3.2.2

brendankenny:
typescript 3.2 is out!

https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#typescript-32

Not a whole lot new for us. Main thing is that `Function.prototype.bind()` (and `call()` and `apply()`) are now strictly typed. Fortunately all our uses of `bind()` (mostly event listeners) seem to have been correct, so...good job us :)

There are a few other things that we'll benefit passively from going forward (see release notes). I'm excited that [`null` and `undefined` are now valid union discriminators](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-unit-types-as-union-discriminants). I'm fairly certain we have a few places in the code where we awkwardly work around this not working in the past, but I can't remember where they are so we'll just update when we find them :)

Changes in this PR are just fixing a few places the compiler got stricter.

### DIFF
--- a/build/build-viewer.js
+++ b/build/build-viewer.js
@@ -113,8 +113,15 @@ async function compileJs() {
   const generatorFilename = `${sourceDir}/../lighthouse-core/report/report-generator.js`;
   const generatorBrowserify = browserify(generatorFilename, {standalone: 'ReportGenerator'})
     .transform('brfs');
-  const generatorBundle = promisify(generatorBrowserify.bundle.bind(generatorBrowserify));
-  const generatorJs = (await generatorBundle()).toString();
+
+  /** @type {Promise<string>} */
+  const generatorJsPromise = new Promise(resolve => {
+    generatorBrowserify.bundle((err, src) => {
+      if (err) throw err;
+      resolve(src.toString());
+    });
+  });
+  const generatorJs = await generatorJsPromise;
 
   // Report renderer scripts.
   const rendererJs = htmlReportAssets.REPORT_JAVASCRIPT;
--- a/build/build-viewer.js
+++ b/build/build-viewer.js
@@ -115,9 +115,9 @@ async function compileJs() {
     .transform('brfs');
 
   /** @type {Promise<string>} */
-  const generatorJsPromise = new Promise(resolve => {
+  const generatorJsPromise = new Promise((resolve, reject) => {
     generatorBrowserify.bundle((err, src) => {
-      if (err) throw err;
+      if (err) return reject(err);
       resolve(src.toString());
     });
   });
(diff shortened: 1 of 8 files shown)


build/build-viewer.js
```diff
@@ -113,8 +113,15 @@ async function compileJs() {
...
-  const generatorJs = (await generatorBundle()).toString();
+
+  /** @type {Promise<string>} */
+  const generatorJsPromise = new Promise(resolve => {
+    generatorBrowserify.bundle((err, src) => {
+      if (err) throw err;
+      resolve(src.toString());
+    });
+  });
+  const generatorJs = await generatorJsPromise;
```
brendankenny: try as I might, I can't convince the compiler that the `promisify` call was working, so just made this an explicit promise

clients/extension/scripts/popup.js
```diff
@@ -234,12 +234,14 @@ async function initPopup() {
     find('header h2').textContent = siteURL.origin;
   });
 
+  const backgroundPagePromise = new Promise(resolve => chrome.runtime.getBackgroundPage(resolve));
```
brendankenny: this was always a lie...the callback to `chrome.runtime.getBackgroundPage` is called with `window`, not just functions from `extension-entry.js`. We didn't tell tsc that `extension-entry.js` actually exposes its functions on `window`...but it doesn't seem worth it to do that now either. This is just a new way to lie :)

Hoten:
Awesome! I was just about to bump this myself. Can use the well-typed `.apply` in the logger changes.

build/build-viewer.js
```diff
@@ -113,8 +113,15 @@ async function compileJs() {
   const generatorFilename = `${sourceDir}/../lighthouse-core/report/report-generator.js`;
   const generatorBrowserify = browserify(generatorFilename, {standalone: 'ReportGenerator'})
     .transform('brfs');
-  const generatorBundle = promisify(generatorBrowserify.bundle.bind(generatorBrowserify));
-  const generatorJs = (await generatorBundle()).toString();
+
+  /** @type {Promise<string>} */
+  const generatorJsPromise = new Promise(resolve => {
+    generatorBrowserify.bundle((err, src) => {
+      if (err) throw err;
```
patrickhulce: not reject?

build/build-viewer.js
Hoten: i hacked my way around it

image

seems TS doesn't like callback style functions having no errors ... which is very weird. see `promisify` in the standard node ts lib. None of them have a nullable Error

build/build-viewer.js
brendankenny: whoops, yes, reject

PR #6688, opened 2018-11-29
report: larger vertical margins for audit-group summaries

brendankenny:
part of #6395

Widens the vertical margins around group header summaries to match the specced design (see discussion starting with https://github.com/GoogleChrome/lighthouse/issues/6395#issuecomment-439259634).

Before:

screen shot 2018-11-29 at 15 19 05

After:

screen shot 2018-11-29 at 15 21 25

Things are a little spread out for e.g. the a11y groups:

screen shot 2018-11-29 at 15 22 46

but it's not egregious

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -652,6 +652,7 @@
   border-bottom: none;
 }
 .lh-audit-group--metrics .lh-audit-group__summary {
+  margin-top: 0;
   margin-bottom: 0;
 }
 
@@ -659,7 +660,8 @@
   display: flex;
   justify-content: space-between;
   padding-right: var(--text-indent);
-  margin-bottom: var(--lh-audit-vpadding);
+  margin-top: calc(var(--section-padding) * 1.5);
+  margin-bottom: var(--section-padding);
 }
 
 .lh-audit-group__itemcount {


patrickhulce:
Does this go for DevTools too? I guess we're already different enough from the rest of the aesthetic what's a little more spacious layout 😆

lighthouse-core/report/html/report-styles.css
```diff
@@ -652,14 +652,16 @@
...
   margin-bottom: 0;
 }
 
 .lh-audit-group__summary {
   display: flex;
   justify-content: space-between;
   padding-right: var(--text-indent);
-  margin-bottom: var(--lh-audit-vpadding);
+  margin-top: calc(var(--section-padding) * 1.5);
+  margin-bottom: var(--section-padding);
```
brendankenny: `--section-padding` is admittedly somewhat a pick out of a hat, but it mostly matches what's being done here, and it gets the slightly more crowded style for devtools (24/16px by default, 18/12px in devtools)

(plus @paulirish suggested using it for this :)

brendankenny:
> Does this go for DevTools too?

ha, ninjaed :)

It's possible it's still too spread out for devtools, but maybe we can tweak? It was `margin-bottom` 4px before, now it's 12px.

PR #6689, opened 2018-11-30
report(pwa): add tooltip to badge gauge

brendankenny:
last item in #6395!

Adds a tooltip as discussed in https://github.com/GoogleChrome/lighthouse/issues/6395#issuecomment-439562286 and comments below that.

screen shot 2018-11-29 at 18 57 01

Was a little more work than expected because we needed `categoryGroups` piped through to get their titles, but not too bad since they were mostly already nearby.

### DIFF
--- a/lighthouse-core/report/html/renderer/pwa-category-renderer.js
+++ b/lighthouse-core/report/html/renderer/pwa-category-renderer.js
@@ -27,7 +27,7 @@ class PwaCategoryRenderer extends CategoryRenderer {
   render(category, groupDefinitions = {}) {
     const categoryElem = this.dom.createElement('div', 'lh-category');
     this.createPermalinkSpan(categoryElem, category.id);
-    categoryElem.appendChild(this.renderCategoryHeader(category));
+    categoryElem.appendChild(this.renderCategoryHeader(category, groupDefinitions));
 
     const auditRefs = category.auditRefs;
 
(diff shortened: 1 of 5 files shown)


lighthouse-core/report/html/renderer/pwa-category-renderer.js
```diff
@@ -104,6 +106,28 @@ class PwaCategoryRenderer extends CategoryRenderer {
...
+   * @param {Array<LH.ReportResult.AuditRef>} auditRefs
+   * @param {Record<string, LH.Result.ReportGroup>} groupDefinitions
+   * @return {string}
+   */
+  _getGaugeTooltip(auditRefs, groupDefinitions) {
+    const groupIds = this._getGroupIds(auditRefs);
+
+    const tips = [];
+    for (const groupId of groupIds) {
+      const groupAudits = auditRefs.filter(ref => ref.group === groupId);
```
patrickhulce: nit: `groupAuditRefs` or `.map(ref => ref.result)`

lighthouse-core/test/report/html/renderer/pwa-category-renderer-test.js
```diff
@@ -152,7 +165,16 @@ describe('PwaCategoryRenderer', () => {
...
       const categoryElem = pwaRenderer.render(category, sampleResults.categoryGroups);
       assert.strictEqual(categoryElem.querySelectorAll('.lh-badged').length, groupIds.length);
-      assert.ok(categoryElem.querySelector('.lh-gauge--pwa__wrapper.lh-badged--all'));
+
+      // Score gauge.
+      const gaugeElem = categoryElem.querySelector('.lh-gauge--pwa__wrapper');
+      assert.ok(gaugeElem.classList.contains('lh-badged--all'));
+
+      // All tooltips should have x/x audits passed.
+      const tips = gaugeElem.title.split(', ');
```
patrickhulce: should we assert any length here?

paulirish:
lg

lighthouse-core/report/html/renderer/pwa-category-renderer.js
brendankenny: > `groupAuditRefs`

done

lighthouse-core/test/report/html/renderer/pwa-category-renderer-test.js
brendankenny: > should we assert any length here?

good idea, done

Issue #6436, opened 2018-10-31
report(font-size): linkify selectors in devtools


connorjclark:
Set `lh-node` attribute for each node shown in the font-size report. See a11y.

Hoten:
This actually isn't useful :)

connorjclark:
> This actually isn't useful :)

That's because I wasn't attempting to link to the range specific to the CSS rule we are highlighting. That would be very useful!

The `CSSStyle.range` property has startLine/startColumn relative to the start of the stylesheet (0 for network stylesheets, relative to `<style>` location for inline).

For inline styles, @paulirish and I just confirmed that we can use `LH.Crdp.CSS.CSSStyleSheetHeader.{isLine,startLine,startColumn}` to link to the sources panel for the HTML at the position where the style rule begins. Usually that'd show you some ugly looking code until you hit Prettify. But luckily there's some movement towards shipping the experimental DT feature "always prettify".

So the idea would be to create something similar to `lh-node`, but have it contain the URL / line / col (`UILocation`) to open in the sources panel. `lh-ui-location`?

connorjclark:
although, the location of a `style` element in the HTML will very likely change across every page load, so we'll see a lot of the same issues from node revealing (#2289)

PR #6702, opened 2018-12-02
core: page function updates relating to upcoming tap targets audit

mattzeunert:
**Summary**

Splitting up [this PR](https://github.com/GoogleChrome/lighthouse/pull/5846) into more digestible bits. 

This PR just contains some page function tweaks, and a new function to get an element selector that we can show to the user.

**Related Issues/PRs**

Tap targets ticket: https://github.com/GoogleChrome/lighthouse/issues/4358

### DIFF
--- a/lighthouse-core/gather/gatherers/accessibility.js
+++ b/lighthouse-core/gather/gatherers/accessibility.js
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-/* global window, document, Node, getOuterHTMLSnippet */
+/* global window, document, getOuterHTMLSnippet, getNodePath */
 
 const Gatherer = require('./gatherer');
 const fs = require('fs');
@@ -45,6 +45,7 @@ function runA11yChecks() {
     // Augment the node objects with outerHTML snippet & custom path string
     // @ts-ignore
     axeResult.violations.forEach(v => v.nodes.forEach(node => {
+      // @ts-ignore - getNodePath put into scope via stringification
       node.path = getNodePath(node.element);
       // @ts-ignore - getOuterHTMLSnippet put into scope via stringification
       node.snippet = getOuterHTMLSnippet(node.element);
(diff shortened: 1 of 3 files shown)


lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -44,5 +44,10 @@ describe('DetailsRenderer', () => {
         ['style-missing', 'aria-label-missing']
       ), '<div id="1" style="style" aria-label="label">');
     });
+
+    it('works if attribute values contain line breaks', () => {
```
patrickhulce: ❤️

lighthouse-core/lib/page-functions.js
```diff
@@ -153,6 +153,71 @@ function ultradumbBenchmark() {
...
+  path.reverse();
+  return path.join(',');
+}
+
+/**
+ * @param {Element} node
+ * @returns {string}
+ */
+/* istanbul ignore next */
+function getNodeSelector(node) {
```
patrickhulce: think we could get a quick test for this function? don't need crazy coverage just something to make sure we don't break it

lighthouse-core/gather/gatherers/accessibility.js
```diff
@@ -45,6 +45,7 @@ function runA11yChecks() {
     // Augment the node objects with outerHTML snippet & custom path string
     // @ts-ignore
     axeResult.violations.forEach(v => v.nodes.forEach(node => {
+      // @ts-ignore - getNodePath put into scope via stringification
       node.path = getNodePath(node.element);
```
Hoten: you could keep this the same, if you also export `getNodePath` in `page-functions.js`.

Will need that module to export `getNodePath` for linking nodes in the report for the font-size audit (#6436) (and for other non-accessibility audits).

patrickhulce: I'm not sure what you mean @Hoten, it's run in the context of the page and he's exported the stringified version which is what he'll need, no?

Hoten: Didn't realize all of that was running in context of the page. Nevermind!

Issue #6675, opened 2018-11-28
Preload audit on lazy iframe


kirill-chirkov-at-clouty:
#### Provide the steps to reproduce
Run LH on https://www.clouty.ru/shop-item/9cedcb77-e064-40a5-9765-1e79152d83bc/jackets-zhaket-emporio-armani

#### What is the current behavior?
Audit suggests me to `Preload key requests` on iframe with src `https://www.clouty.ru/toshop?id=9cedcb77e06440a597651e79152d83bc&type=iframe`.
Iframe is injected lazily via js, some time after `onload` browser event and serves pure technical needs.

#### What is the expected behavior?
Preloading this specified iframe would ruin the lazy behavior, but preload audit lowers score significantly

#### Environment Information
* Affected Channels:
* Lighthouse version: 4.0.0-alpha.2-3.2.1, 4.0.0-alpha.1, 3.2.1 and 3.0.3 (Chrome Lighthouse audit)
* Node.js version: 10.13.0
* Operating System: Windows 10 Pro 64-bit

**Related issues**
https://github.com/GoogleChrome/lighthouse/issues/4425#issuecomment-442416040

wardpeet:
@patrickhulce I did some research on this and when looking into devtools we get a resourceType of undefined back because it's a 302 so my suggestion is to also hide undefined from critical resources.

patrickhulce:
Yeah I think that's a good call @wardpeet! 👍

patrickhulce:
Oh I just realized that might be because of OOPIFs (#6337), still makes sense to do your filter I think though until we have the protocol support we're waiting on.

brendankenny:
fixed in #6704 and verified that it no longer shows up in the report for this site. Additional preload improvements will continue to happen in #4425

PR #6704, opened 2018-12-02
fix(audit): iframe with redirect shouldn't be marked as critical

wardpeet:
**Summary**

Fixes 302 iframes inside a page marked as critical requests

**Related Issues/PRs**
#6675

### DIFF
--- a/lighthouse-core/computed/critical-request-chains.js
+++ b/lighthouse-core/computed/critical-request-chains.js
@@ -31,6 +31,7 @@ class CriticalRequestChains {
     // Iframes are considered High Priority but they are not render blocking
     const isIframe = request.resourceType === NetworkRequest.TYPES.Document
       && request.frameId !== mainResource.frameId;
+    const hasStatusCode200 = request.statusCode === 200;
     // XHRs are fetched at High priority, but we exclude them, as they are unlikely to be critical
     // Images are also non-critical.
     // Treat any missed images, primarily favicons, as non-critical resources
@@ -42,6 +43,8 @@ class CriticalRequestChains {
       NetworkRequest.TYPES.EventSource,
     ];
     if (nonCriticalResourceTypes.includes(request.resourceType || 'Other') ||
+        // if an iframe as a 302 it resourceType is undefined #6675
+        (!request.resourceType && !hasStatusCode200) ||
         isIframe ||
         request.mimeType && request.mimeType.startsWith('image/')) {
       return false;
--- a/lighthouse-core/computed/critical-request-chains.js
+++ b/lighthouse-core/computed/critical-request-chains.js
@@ -28,10 +28,16 @@ class CriticalRequestChains {
       return false;
     }
 
+    let resourceType = request.resourceType;
+    // if resourcetype is undefined we check if it's not a 200 and look for a redirect
+    // when we find a redirect we use the redirectDestination's resourcetype #6675
+    if (!resourceType && request.statusCode !== 200 && request.redirectDestination) {
+      resourceType = request.redirectDestination.resourceType;
+    }
+
     // Iframes are considered High Priority but they are not render blocking
-    const isIframe = request.resourceType === NetworkRequest.TYPES.Document
+    const isIframe = resourceType === NetworkRequest.TYPES.Document
       && request.frameId !== mainResource.frameId;
-    const hasStatusCode200 = request.statusCode === 200;
     // XHRs are fetched at High priority, but we exclude them, as they are unlikely to be critical
     // Images are also non-critical.
     // Treat any missed images, primarily favicons, as non-critical resources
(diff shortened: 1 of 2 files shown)


lighthouse-core/computed/critical-request-chains.js
```diff
@@ -42,6 +43,8 @@ class CriticalRequestChains {
       NetworkRequest.TYPES.EventSource,
     ];
     if (nonCriticalResourceTypes.includes(request.resourceType || 'Other') ||
+        // if an iframe as a 302 it resourceType is undefined #6675
+        (!request.resourceType && !hasStatusCode200) ||
```
wardpeet: I could omit the request.resourceType check this bugs only occured on iframes with a redirect so I like to be really explicit about it.

patrickhulce: Generally, restricting based on status code and missing resource type actually sounds pretty good to me (maybe just no 4xx/5xx in generic terms).

The only sticking point I'm coming up with this is if the root document request was a redirect, what's going to happen? I just realized this bug might be OOPIF related so if we're fairly confident this only happens to iframes we're probably OK.

Aside, I'm somewhat surprised we didn't include `'Other'` in the `nonCriticalResourceTypes` already which would stop this too, but a lot of things can wind up being `Other` I guess :/

wardpeet: it's actually undefined :p

wardpeet: @patrickhulce so resourceType is actually undefined when it's a 302. I'm now checking if redirectDestination exists and use it's resourceType. Maybe I should make it recursive? Now i only check one level

lighthouse-core/computed/critical-request-chains.js
```diff
@@ -28,8 +28,15 @@ class CriticalRequestChains {
       return false;
     }
 
+    let resourceType = request.resourceType;
+    // if resourcetype is undefined we check if it's not a 200 and look for a redirect
+    // when we find a redirect we use the redirectDestination's resourcetype #6675
+    if (!resourceType && request.statusCode !== 200 && request.redirectDestination) {
+      resourceType = request.redirectDestination.resourceType;
```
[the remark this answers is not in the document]
patrickhulce: interesting is it just resourceType that's undefined on redirects? 

might actually make sense to do the whole thing on the final request if other pieces can be missing

maybe we just throw
```js
while (request.redirectDestination) request = request.redirectDestination
```

at the top?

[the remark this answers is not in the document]
wardpeet: @patrickhulce when using the while loop we get a different priority so this means 99% of the time we don't tell people to preload this request. I'm not sure what the perf perspective is on this. (@paulirish)

So for example for the bug https://www.clouty.ru/shop-item/9cedcb77-e064-40a5-9765-1e79152d83bc/jackets-zhaket-emporio-armani it means our html doc gets a low priority which means we won't tell people to preload.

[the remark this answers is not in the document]
patrickhulce: hm that actually makes a lot of sense to me, if the browser later figured out it wasn't very important, it probably wasn't important enough to preload

kirill-chirkov-at-clouty:
@paulirish Yeah, regarding #6675, I don't think it has something to do with 302's...

kirill-chirkov-at-clouty:
@paulirish I might guess that 302 status on our current iframe (which described in #6675) is pure coincidental, point of issue was that iframe was loaded not with the document but later, JS-initiated. Fixing 302 as non-critical request would help in our case, but I might guess that won't help where programmaticaly enabled iframe would not give 302 but 200.

wardpeet:
i'll check it out if it's still a problem ,if not we can just close this one.

paulirish:
#7352 fixed initiators and redirects. It makes a lot more sense now.

We specifically say that iframes aren't critical. 

And we only want critical requests for the user to preload. So if we ever flag a redirect (that eventually is an iframe) in those audits, we have something to fix.

I couldn't get a redirect to show up in preload details. But maybe i wasn't constructing the right situation. 

If they'll never show up in preload, then we're done here. :)

wardpeet:
@paulirish it's not showing up anymore in preload so that's good but the redirect is still marked as a critical request because the resourceType is marked as undefined for a redirect.

I made a repro on glitch
https://erratic-lion.glitch.me

I'm rebasing this branch

lighthouse-core/computed/critical-request-chains.js
```diff
@@ -28,6 +28,12 @@ class CriticalRequestChains {
       return false;
     }
 
+    // Whenever a request is a redirect not all props are filled in correctly (resourceType is undefined)
+    // so we loop until we find the final request so we use the correct data #6675
```
patrickhulce: nit: `https://github.com/GoogleChrome/lighthouse/issues/6675` so it's clickable :)

lighthouse-core/test/computed/critical-request-chains-test.js
```diff
@@ -326,6 +342,15 @@ describe('CriticalRequestChain gatherer: extractChain function', () => {
     networkRecords[2].mimeType = 'text/html';
     networkRecords[2].resourceType = NetworkRequest.TYPES.Document;
     networkRecords[2].frameId = '3';
+    // 4rd record is an iframe in the page with a redirect #6675
```
patrickhulce: nit: `https://github.com/GoogleChrome/lighthouse/issues/6675` :)

wardpeet:
Fixed comments 😄

wardpeet:
thanks @paulirish !

Issue #6660, opened 2018-11-27
request with `intent:` protocol was treated  as network request


echopi:
**Summary**

Shalt we preload the `intent:` request? 

![image](https://user-images.githubusercontent.com/16034964/49057191-85675700-f239-11e8-93e1-b7414e4028ec.png)

patrickhulce:
Nope, this shouldn't be flagged as a key request. Thanks for filing this @echopi!

midzer:
I have time to look at this one, if noone else wants to contribute as a newcomer.

PR #6711, opened 2018-12-03
core(url-shim): add intent protocol to NON_NETWORK_PROTOCOLS

midzer:
**Summary**
What kind of change does this PR introduce?
Don't regard an `intent:` request as key request

Is this a bugfix, feature, refactoring, build related change, etc?
Issue been tagged as bug. So bugfix :)

Link any documentation or information that would help understand this change
I'm unsure whether my change/addition goes in the right direction, as I cannot find more detailed information about `intent` protocol/requests

**Related Issues/PRs**
closes https://github.com/GoogleChrome/lighthouse/issues/6660

Cheers
midzer

### DIFF
--- a/lighthouse-core/test/audits/uses-rel-preload-test.js
+++ b/lighthouse-core/test/audits/uses-rel-preload-test.js
@@ -209,6 +209,18 @@ describe('Performance: uses-rel-preload audit', () => {
       assert.equal(output.details.items.length, 0);
     });
   });
+  
+  it(`shouldn't suggest preload for protocol scheme`, () => {
+    const networkRecords = getMockNetworkRecords();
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/audits/uses-rel-preload-test.js
```diff
@@ -209,6 +209,18 @@ describe('Performance: uses-rel-preload audit', () => {
       assert.equal(output.details.items.length, 0);
     });
   });
+
```
exterkamp:
```suggestion

```

exterkamp: Travis is mad about this trailing space.  I think this suggestion is accurate to just remove the space but keep the /n

Issue #3862, opened 2017-11-20
audit: bundle analysis


addyosmani:
This is a placeholder for a JS bundle analysis feature we will explore in Q3/Q4 2018: deep analysis and webperf recommendations based on your Source Maps. Something that iterates on webpack-bundle-analyzer perhaps.

Most developers I speak to are painfully unaware that they probably need to use a bundle analysis tool to discover issues in their JS and modern tools don't seem to be yelling at them loudly enough when they run into problems.

It wouldn't be easy...so..opening to discuss :)

When working with partner teams, we often find the following to be true:

* Folks are usually not using a tool like [source-map-explorer](https://www.npmjs.com/package/source-map-explorer) or Webpack Bundle Analyzer until they realize they have performance issues with JS bundles. Opportunity for us to inform them of issues before they learn about these tools.
* When folks do discover these tools, it's not trivial to read the dependency graph visualizations and get the "so...what are my next steps here?" guidance out. We could codify some of the patterns we commonly see teams failing on and try to offer up audits around this.
* **Potential audits**:
  - **Library-specific**: 
    * We noticed you're using all of Lodash in your bundle, but you're only actually calling 3 methods. Maybe use lodash-es or lodash-webpack-plugin to reduce the size of your bundle
    * We noticed you're using Moment.js, but including all of the default locales, bloating your bundle by a few hundred KB. Switch to using specific locales you wish to support to automate this using Webpack's contextual replacement plugin to trim down your bundle.
    * Debug version of React. Switch to production version.
  - **Polyfill-specific**: We noticed you're shipping a whole lot of Babel polyfills / core-js. Most modern browsers already support features like ES promises so...maybe use babel-preset-env and don't ship all these polyfills in your bundles.
  - **Duplicate-entries**: We noticed you're including a bunch of duplicate entries (e.g imports, vendor libraries) in your bundle. Maybe you should avoid this. Check out tools like Bundle Buddy and see if they can help.

etc. This could be really neat but does come with some concerns outside of whether it's technically feasible or not:

* How can we keep up with the times when it comes to these toolchain specific recommendations? Do we just bite the bullet and say we'll survey for what is still valid every year?
* Would it make more sense to scope this down to framework specific audits?

Webpack Bundle Analyzer for anyone unfamiliar with it:

[...]

PR #6730, opened 2018-12-05
new_audit(legacy-javascript): detect unnecessary polyfills and transforms

connorjclark:
**EDIT 2**: [Design doc](https://docs.google.com/document/d/1ItjJwAd6e0Ts6yMbvh8TN3BBh_sAd58rYE1whnpuxaA/edit)

**EDIT**: jump to [here](https://github.com/GoogleChrome/lighthouse/pull/6730#issuecomment-533383983) for updates.

I imagine this will go through a few iterations.

image

@paulirish had the idea to surface what polyfills are in use, and suggest removing them. It sounded like fun so I took pass at it.

First, let's talk detecting polyfills. I thought about a few ways to do it, but just two were viable. We could parse out an AST to extract where native functions were being redefined. But, that's likely to be slow, and would add quite a large dependency. Instead, I opted for somewhat-simple regex parsing. The grammar for setting a property (either `=` or `Object.defineProperty`) is fairly simple to grok for, and ought to be much faster. So I went with regex.

Second, let's talk what suggestions to make. I'm not sure we can concretely suggest removing _any_ polyfills. Developers often must support an audience that uses old browsers, so suggesting removing polyfills for features present in the newest browsers would not be found agreeable. Even if a feature is found in browsers going back to IE8, it can still be the wrong call for some sites. So I suggest not attaching any score to this audit, and keeping the language non-prescriptive. Instead, we can just surface what polyfills exist and suggest that they consider researching what they actually need.

Some things worth talking about:

1. Detecting modules that re-implement native JS features, but don't actually polyfill anything. For example, bundling `lodash/startsWith` for a web target, and using that instead of `String.prototype.startsWith`. Bundlers tend to strip module names in production, so such detection likely requires source maps. Otherwise, we're left with analyzing what these modules look like minified. This could certainly be explored further in a separate feature.

2. If a polyfill appears multiple times from multiple scripts, I display a counter in the report. I imagine this would only be actionable if the polyfills are coming from the same origin - if two third party scripts are polyfilling the same thing, there's nothing really to be done. Should we suggest action only if multiple polyfills are coming from the same origin? If so, what's a good way to present that suggestion?

3. The location in the script that the polyfill occurs is provided in the report. Usually, this script is minified, so really it's just an index into an unreadable buffer... Is this useful at all? Should this be removed and only added back with source map support?

4. I grabbed a list of features that can be polyfilled from [here](https://www.npmjs.com/package/mdn-polyfills#supported-polyfills). Certainly not exhaustive. From where should we source this list?

Related: #3862


patrickhulce:
DZL, do a barrel roll!

patrickhulce:
DZL is done! Go check it out http://lh-dzl-6730.surge.sh

Hoten:
CNN
image

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,207 @@
...
+    }).join('|');
+
+    const polysSeen = new Set();
+    const polyMatches = [];
+    const re = new RegExp(`(^\r\n|\r|\n)|${pattern}`, 'g');
+    /** @type {RegExpExecArray | null} */
+    let result;
+    let row = 0;
+    let rowBeginsAtIndex = 0;
+    while ((result = re.exec(code)) !== null) {
```
patrickhulce: I'm gonna at least need a comment here :)

the `...matches` below made me think the results were already somehow all inside one regex match, but the `.exec` suggests we need to iterate through them all

the `..matches` is just the subgroups of each match?

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,207 @@
...
+     * @return {PolyIssue[]}
+     */
+  static detectPolys(polys, code) {
+    const qt = (/** @type {string} */ token) =>
+      `['"]${token}['"]`; // don't worry about matching string delims
+    const pattern = polys.map(({object, property}) => {
+      let subpattern = '';
+
+      // String.prototype.startsWith =
+      subpattern += `${object || ''}.${property}\\s*=`;
```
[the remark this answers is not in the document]
Hoten: yup

lighthouse-core/audits/polyfills.js
Hoten: yeah, and only one subgroup is ever defined in each `matches`. I agree this is arcane bordering on magic. I'll comment accordingly.

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,16 +20,20 @@ class Scripts extends Gatherer {
...
     const scriptRecords = loadData.networkRecords
       .filter(record => record.resourceType === NetworkRequest.TYPES.Script);
 
     for (const record of scriptRecords) {
       try {
         const content = await driver.getRequestContent(record.requestId);
+        const url = record.url;
         if (content) {
-          scriptContentMap[record.requestId] = content;
+          scriptContentMap[record.requestId] = {
```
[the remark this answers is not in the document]
Hoten: Would I necessarily need to create a gather for this then? That's what I wanted to avoid.

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /**
+   * @param {LH.Artifacts} artifacts
+   * @return {Promise<LH.Audit.Product>}
+   */
+  static async audit(artifacts) {
+    // TODO
+    // how do we determine which polys are not necessary?
+    // ex: only reason to polyfill Array.prototype.filter is if IE <9,
```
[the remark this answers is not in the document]
Hoten: That makes a lot of sense, I did not consider that. Very cool that many ES2015+ features can be feature-detected with `<script>` modules.

patrickhulce:
DZL is done! Go check it out http://lh-dzl-6730.surge.sh

lighthouse-core/gather/gatherers/scripts.js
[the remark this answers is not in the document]
patrickhulce: Nope, shouldn't need any gatherer changes at all, just use Scripts artifact as-is and get the URL from `const url = networkRecords.find(record => record.requestId === requestId).url`

will have to compute network records from devtools log, but not the end of the world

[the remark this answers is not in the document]
Hoten: Sweet - I reverted those changes.

patrickhulce:
DZL is done! Go check it out http://lh-dzl-6730.surge.sh

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,225 @@
...
+
+/** @typedef {{object: string | undefined, property: string}} Poly */
+/** @typedef {{row: number, col: number, poly: Poly}} PolyIssue */
+
+const Audit = require('./audit');
+const NetworkRecords = require('../computed/network-records.js');
+const i18n = require('../lib/i18n/i18n.js');
+
+const UIStrings = {
+  /** Imperative title of a Lighthouse audit that tells the user about all JavaScript polyfills loaded on the page. This is displayed in a list of audit titles that Lighthouse generates. */
```
patrickhulce: this one's not imperative :)

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,225 @@
...
+  title: 'Polyfills',
+  /** TODO: write this */
+  // eslint-disable-next-line max-len
+  description: 'Polyfills enable older browsers to use new JavaScript language features. However, they aren\'t always necessary. Research what browsers you must support and consider removing polyfils for features that are well supported by them.',
```
patrickhulce: Maybe our advice here is more like `Consider conditionally serving polyfills based on feature availability.`?

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,225 @@
...
+    const devtoolsLog = artifacts.devtoolsLogs[Polyfills.DEFAULT_PASS];
+    const networkRecords = await NetworkRecords.request(devtoolsLog, context);
+
+    // TODO
+    // how do we determine which polys are not necessary?
+    // ex: only reason to polyfill Array.prototype.filter is if IE <9,
+    //     only reason to polyfill String.prototype.startsWith is if IE, ...
+    // Is there a standard way to declare "we support these browsers"? A meta tag?
+    /** @type {Poly[]} */
+    const polys = [
```
patrickhulce: maybe we could pull some of this code out of the main `audit` function and break it down a bit?

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,225 @@
...
+          // TODO use sourcemaps
+          location: `r: ${row}, c: ${col}`,
+        });
+      }
+    });
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
+      {key: 'description', itemType: 'code', text: 'Description'},
+      {key: 'location', itemType: 'code', text: 'Location'},
+      // TODO include estimate for size based on https://www.npmjs.com/package/mdn-polyfills#supported-polyfills ?
```
patrickhulce: this is interesting, we could then make it an opportunity instead of a diagnostic too!

lighthouse-core/config/default-config.js
```diff
@@ -457,6 +458,8 @@ const defaultConfig = {
         {id: 'password-inputs-can-be-pasted-into', weight: 1},
         {id: 'errors-in-console', weight: 1},
         {id: 'image-aspect-ratio', weight: 1},
+        // Manual audits
+        {id: 'polyfills', weight: 0},
```
patrickhulce: IMO this belongs in performance as a diagnostic and can be upgraded to an opportunity once we have the size data :)

lighthouse-core/audits/polyfills.js
Hoten: I think it might be hilariously small of an impact, though. The average in that table is 300 bytes. Say there's a dozen - about 4KB. is that significant?

patrickhulce: We'd probably have to get a little more aggressive then. Some of the babel polyfills are no joke, maybe async/await IIRC, but the impact was more like ~80KB for some bundles.

patrickhulce: Yeah in phil's article it was ~95K minified plus some extra parse/eval time savings

image


https://philipwalton.com/articles/deploying-es2015-code-in-production-today/

Hoten: But do we know how much of that is generator cruft (async -> es5 code), string templates (aka language level polyfills), vs. web api polyfills?

so far I'm just considering web api polyfills, which is much more easily identifiable than language polyfills. It'd be very powerful to capture those too though. any ideas?

patrickhulce: Yeah these are great points!

I wasn't trying to suggest we do it right away these other categories are what I was trying to get at with be more aggressive :) I'm afraid I don't really have any obvious initial ideas we can pursue. Maybe we'll have to take a look at some babelized bundles and look for distinctive markers of cruft, but we can cross that bridge another day for sure 👍

patrickhulce:
DZL is done! Go check it out http://lh-dzl-6730.surge.sh

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,236 @@
...
+        const polyStatement = `${poly.object ? poly.object + '.' : ''}${poly.property}`;
+        const numOfThisPoly = polyCounter.get(poly) || 0;
+        const isMoreThanOne = numOfThisPoly > 1;
+        const polyIssueOccurrence = polyIssueCounter.get(polyIssue) || 0;
+        const countText = isMoreThanOne ? ` (${polyIssueOccurrence + 1} / ${numOfThisPoly})` : '';
+        tableRows.push({
+          url,
+          description: `${polyStatement}${countText}`,
+          // TODO use sourcemaps
+          location: `r: ${row}, c: ${col}`,
```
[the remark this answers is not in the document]
Hoten: will do

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,233 @@
...
+    return polyMatches;
+  }
+
+  /**
+   * @return {Poly[]}
+   */
+  static getPolyDefinitions() {
+    // if latest Chrome supports a feature natively, we should
+    // complain about the existence of a polyfill
+    return [
```
Hoten: note to self: try replacing this with some logic. look for any `<something?>.prototype.<something>`, if pattern is found ensure it is a native built in (does it exist?) and not some custom thing.

patrickhulce:
Side-note I recently came across a URL in #7027 that had this
image

Not sure how that's possible, could be related to #7005 but all aboard to polyfill removal train!

Hoten:
> Side-note I recently came across a URL in #7027 that had this

wow that's quite the polyfill file.

> Not sure how that's possible, could be related to #7005 but all aboard to polyfill removal train!

could it be another mismatch between expectation & reality re: attribution? similar to #7005

patrickhulce:
> could it be another mismatch between expectation & reality re: attribution?

yeah it seems like it's getting credit for all XHR callbacks, probably polyfilling fetch or something

patrickhulce:
@brendankenny I think you meant to leave that comment on @wardpeet's issue not this PR?

brendankenny:
> I think you meant to leave that comment on @wardpeet's issue not this PR?

Ha, I didn't, but that's what I get when I'm mostly thinking and not reading the code :) I was thinking this was being done in a gatherer, not an audit. Disregard!

Hoten:
@patrickhulce Before i refactor this audit, just wanna run this by you first:

I want to remove the hardcoded list of polyfills to check for. Figure I can just look for a pattern of setting a property on a prototype `SomeClass.prototype.someFn = ...` or setting a property directly on a class `SomeClass.someFn = ...`. Would also check if someFn is actually defined natively in Chrome, which requires doing some stuff in an isolated context in a gatherer, which means I would need to move most of the code to a gatherer.

does that make sense to you?

patrickhulce:
> does that make sense to you?

I personally don't think all of that is worth it. I think this is a great start and keeping it all static makes it very clear what's a violation and is very unlikely to start accidentally including false positives.

Hoten:
BTW, that generated regex looks like [this](https://gist.github.com/Hoten/ddd773fa4de4b7e6f32a4256fb08714b) :)

brendankenny:
not sure what to do with this. A lot has changed with the `stacks` gatherer in place now and more discussion about source maps.

Should we close and move discussion to an issue? Or leave open and keep discussing here?

brendankenny:
Closing in favor of discussing in #8120 and JS payload analysis discussions.

[ext-polys](https://github.com/GoogleChrome/lighthouse/tree/ext-polys) branch will hang out, looking forward to being reopened.

Hoten:
What it looks like now:

image

Hoten:
Maybe I should include a smoke test that uses `babel-polyfill`?

lighthouse-core/audits/polyfills.js
```diff
@@ -0,0 +1,298 @@
...
+      'Array.prototype.findIndex',
+      'Array.prototype.forEach',
+      'Array.prototype.includes',
+      'Array.prototype.indexOf',
+      'Array.prototype.join',
+      'Array.prototype.lastIndexOf',
+      'Array.prototype.slice',
+      'Array.prototype.some',
+      'Array.prototype.sort',
+      'CustomEvent',
```
connorjclark: Should verify that the polyfill parsing logic detected overriding on window:

```
window.CustomEvent = ...
```

connorjclark:
Responding to the questions I posed in the first post:

> Detecting modules that reimplement native JS features, but don't actually polyfill anything. For example, bundling lodash/startsWith for a web target, and using that instead of String.prototype.startsWith. Bundlers tend to strip module names in production, so such detection likely requires source maps. Otherwise, we're left with analyzing what these modules look like minified. This could certainly be explored further in a separate feature.

This can be added to #9097 - once we have bundle analysis we can begin picking on specific packages.

> If a polyfill appears multiple times from multiple scripts, I display a counter in the report. I imagine this would only be actionable if the polyfills are coming from the same origin - if two third party scripts are polyfilling the same thing, there's nothing really to be done. Should we suggest action only if multiple polyfills are coming from the same origin? If so, what's a good way to present that suggestion?

The main concern here is ensuring the items in the table are actionable. Polyfills in third party dependencies are incredibly un-actionable, save for removing the dependency (which is not an action I'd expect is correct in every situation).

Perhaps "show 3rd party resources" should be unchecked by default?

> The location in the script that the polyfill occurs is provided in the report. Usually, this script is minified, so really it's just an index into an unreadable buffer... Is this useful at all? Should this be removed and only added back with source map support?

This offers a great use case for utilizing source maps (#9101) in code snippets. We don't have any details that show code from a ScriptElement, but this could be the first!

> I grabbed a list of features that can be polyfilled from here. Certainly not exhaustive. From where should we source this list?

I guess the current list is good enough for a starting point. If anyone sees some that are missing, please suggest.

housseindjirdeh:
This is super super cool 😍.  I love this!

My two cents: seeing a long list of polyfills in use is great but I wonder how actionable it would be for developers, especially if they're not entirely aware of how to handle them. Linking to a [module/nomodule guide](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/) is a great way to let folks know how they can learn to mitigate polyfill bloat for some browsers, but maybe it would be even nicer if we can show a list of browsers supported by each feature/necessary polyfill as a separate column?

This might provide more visibility so developers are aware of which polyfills are actually unnecessary for their specific use case? Probably easier said then done, so it's just a thought!

connorjclark:
thanks @housseindjirdeh!

> would be even nicer if we can show a list of browsers supported by each feature/necessary polyfill as a separate column?

Evaluated if http://caniuse.com dataset could be used. For super old (EcmaScript 3) features, they just show this:

```
These methods are supported in effectively all browsers (since IE6+, Firefox 2+, Chrome 1+ etc)
```

But for newer features we get more detail:

image

I searched for `Array.of` and it just forwards me to [`compat-table`](http://kangax.github.io/compat-table/es6/). Which looks cool, maybe that'd be a good one to use.

___

So for scripts crazy enough to polyfill super old ES3 features, we could display "all browsers" in the column. Otherwise, we show latest support for IE/Edge/FF/Chrome/Safari/Opera. Certainly useful to gauge how egregious it is to polyfill this particular feature - maybe we could even leverage it to make audit count towards score a bit (if you polyfill an "every browser" feature, -1. Otherwise let it slide).

Would be interested in exploring this further in a followup PR. Since this audit will be non-scoring to start, I don't think it's critical.

lighthouse-core/audits/polyfills.js
connorjclark: added a test for this

wardpeet:
Great work @connorclark! I really love this audit and it will make the web a better place! :heart: Sadly, polyfilling is broken on the web and this audit will show this behavior. I've written down some of my findings on the [web.dev repo](https://github.com/GoogleChrome/web.dev/issues/965) as many projects use `@babel/preset-env` & core-js to polyfill their bundles automatically.

When you typically run a modern vs legacy workflow. You would think that a modern build (setting esmodules to true) wouldn't transpile async-await or polyfill promises. By default they do so you'll need to put in your own target values like ios 11, instead of 10.3, ...

I don't think it's bad that we surface these things, I think we'll get a ton of issues about people not knowing how to fix this.

connorjclark:
Spoke to @paulirish and @philipwalton about how to best give advice on polyfills.

First - although the audit today is only concerning polyfills, language syntax transforms should also be detected. So, we should rename this audit. `unnecessary-javascript` doesn't work, because it conflates with `unused-javascript`. I suggest `legacy-javascript`.

We want to ensure that the advice we give is actionable. The most reasonable way to deploy less code to modern browsers today is the `module`/`nomodule` strategy via `babel-preset-env`'s `esmodule` target.

Using `esmodule` will produce the following browserslist, which denotes the minimum browser versions that support esmodules:

https://github.com/babel/babel/blob/9febf6388233c3d23d4fa9deca5ca8f682567573/packages/babel-preset-env/data/built-in-modules.json
```json
{
  "es6.module": {
    "edge": "16",
    "firefox": "60",
    "chrome": "61",
    "safari": "10.1",
    "opera": "48",
    "ios_saf": "10.3",
    "and_chr": "71",
    "and_ff": "64"
  }
}
```

Any polyfill/transform we surface should be natively and fully supported by all the browsers matching the above browserslist. Example- `transform-async-to-generator` currently must be applied for the `esmodules` target because iso/saf didn't get async/await until version 11, so we should _not_ detect the async generator transform, as the tooling doesn't support it yet. We can begin checking for this transform if [this](https://github.com/babel/babel/pull/10225) is merged.

Further, we want to hand-pick polyfills/transforms that are especially harmful, either in large byte cost or performance cost.

When we have bundle info at hand, we can do some progressive-enhancement-type detection here. Instead of using a regex to find polyfill/transforms, the source maps within a bundle will point out any polyfills or transforms in use. For scripts without bundle info, we fall back to the current regex strategy.

Action items for me:

1. Determine the polyfills and transforms that we should surface.
2. Re-word the audit description to put less emphasis on "don't ship this extra javascript" and more on "use modern best-practices deployment to serve less code to modern browsers".

Just considered this: arguably, only scripts that are themselves bundles (and thus the output of a build process that is already modern and using webpack/browserify/parcel/etc) could follow this advice. So maybe the polyfill checks should be gated on bundles? Would be weird to suggest `babel-preset-env` because an ad tracking script is polyfilling `Array.reverse`.

wardpeet:
I like the update! One small nit about:
>  Example- transform-async-to-generator currently must be applied for the esmodules target because iso/saf didn't get async/await until version 11, so we should not detect the async generator transform, as the tooling doesn't support it yet

It does support it, it just has a bug when using arrow functions inside a class https://bugs.webkit.org/show_bug.cgi?id=166879 that's why compat-table marks it as safari 11.

patrickhulce:
> I suggest legacy-javascript.

WFM 👍 

> only scripts that are themselves bundles (and thus the output of a build process that is already modern and using webpack/browserify/parcel/etc) could follow this advice So maybe the polyfill checks should be gated on bundles?

While I agreeish with the first bit, I wouldn't want to limit the audit to only sites that expose their source maps to us if that's what you were suggesting here.

> Re-word the audit description to put less emphasis on "don't ship this extra javascript" and more on "use modern best-practices deployment to serve less code to modern browsers"

This screams opportunity to me now then :) maybe if you're already doing the analysis on cost of different transforms/polyfills it should just ship as an opportunity?

connorjclark:
> It does support it, it just has a bug when using arrow functions inside a class [bugs.webkit.org/show_bug.cgi?id=166879](https://bugs.webkit.org/show_bug.cgi?id=166879) that's why compat-table marks it as safari 11.

Right, I meant to say it doesn't fully support it, so the polyfill is needed for Safari < 11. With the change I linked to, `esmodules: true` will change the min. saf version to 11, which will remove that transform.

> While I agreeish with the first bit, I wouldn't want to limit the audit to only sites that expose their source maps to us if that's what you were suggesting here.

agreed, just something to consider later I guess. But - @paulirish and I are looking into a a way to identify modules in a bundle without source map ;)

> This screams opportunity to me now then :) maybe if you're already doing the analysis on cost of different transforms/polyfills it should just ship as an opportunity?

Maybe. I'm concerned that the work to detect a transform's application / a JS API being polyfilled is much less than the work needed to accurately attribute the byte cost for the transform/polyfill. The analysis I'm doing is more to be able to confidently say "in 2019, you realllly shouldn't use this polyfill/transform". It then serves as a strong signal that the `module`/`nomdoule` deployment strategy is not being done. Applying that deployment strategy should make this audit pass AND reduce byte size / perf overhead, but it's hard to say by how much.

patrickhulce:
> the work needed to accurately attribute the byte cost for the transform/polyfill. The analysis I'm doing is more to be able to confidently say "in 2019, you realllly shouldn't use this polyfill/transform".

Gotcha, I misunderstood "we want to hand-pick polyfills/transforms that are especially harmful, either in large byte cost or performance cost." then I thought you were already doing the byte cost analysis so we'd be getting it for free. Agreed it is a significant increase in scope to automatically compute this and if we don't already have to do it, let's punt 👍

connorjclark:
Here be the list of transforms / polyfills that `babel-preset-env` _will not_ apply for `esmodules: true`:

```
transform-literals
transform-arrow-functions
transform-block-scoped-functions
transform-classes
transform-object-super
transform-shorthand-properties
transform-duplicate-keys
transform-computed-properties
transform-for-of
transform-sticky-regex
transform-spread
transform-destructuring
transform-typeof-symbol
transform-new-target
transform-regenerator
transform-exponentiation-operator
transform-member-expression-literals
transform-property-literals
transform-reserved-words
es6.array.copy-within
es6.array.every
es6.array.fill
es6.array.filter
es6.array.find
es6.array.find-index
es6.array.for-each
es6.array.from
es7.array.includes
es6.array.index-of
es6.array.is-array
es6.array.iterator
es6.array.last-index-of
es6.array.map
es6.array.of
es6.array.reduce
es6.array.reduce-right
es6.array.some
es6.array.species
es6.date.now
es6.date.to-iso-string
es6.date.to-json
es6.date.to-primitive
es6.date.to-string
es6.function.bind
es6.function.has-instance
es6.function.name
es6.map
es6.math.acosh
es6.math.asinh
es6.math.atanh
es6.math.cbrt
es6.math.clz32
es6.math.cosh
es6.math.expm1
es6.math.fround
es6.math.hypot
es6.math.imul
es6.math.log1p
es6.math.log10
es6.math.log2
es6.math.sign
es6.math.sinh
es6.math.tanh
es6.math.trunc
es6.number.constructor
es6.number.epsilon
es6.number.is-integer
es6.number.is-safe-integer
es6.number.max-safe-integer
es6.number.min-safe-integer
es6.number.parse-float
es6.number.parse-int
es6.object.assign
es6.object.create
es6.object.define-property
es6.object.define-properties
es7.object.entries
es6.object.freeze
es6.object.get-own-property-descriptor
es7.object.get-own-property-descriptors
es6.object.get-own-property-names
es6.object.get-prototype-of
es6.object.prevent-extensions
es6.object.to-string
es6.object.is-frozen
es6.object.is-sealed
es6.object.is-extensible
es6.object.keys
es6.object.seal
es6.object.set-prototype-of
es7.object.values
es6.promise
es6.reflect.apply
es6.reflect.construct
es6.reflect.define-property
es6.reflect.delete-property
es6.reflect.get
es6.reflect.get-own-property-descriptor
es6.reflect.get-prototype-of
es6.reflect.has
es6.reflect.is-extensible
es6.reflect.own-keys
es6.reflect.prevent-extensions
es6.reflect.set
es6.reflect.set-prototype-of
es6.set
es6.string.code-point-at
es6.string.ends-with
es6.string.from-code-point
es6.string.includes
es6.string.iterator
es7.string.pad-start
es7.string.pad-end
es6.string.raw
es6.string.repeat
es6.string.starts-with
es6.string.trim
es6.typed.array-buffer
es6.typed.data-view
es6.typed.int8-array
es6.typed.uint8-array
es6.typed.uint8-clamped-array
es6.typed.int16-array
es6.typed.uint16-array
es6.typed.int32-array
es6.typed.uint32-array
es6.typed.float32-array
es6.typed.float64-array
es6.weak-map
es6.weak-set
```

generated via https://gist.github.com/connorjclark/91373e45e6a9a90ca726c6623a0a7700

connorjclark:
I refactored the pattern matching code to work with detecting transforms too.

___

Some timings. Haven't reduced the # of polyfill audits yet, so these times will go down.

https://www.paulirish.com/
```
CodePatternMatcher ctor: 1.081ms
match https://www.google-analytics.com/analytics.js (44287): 9.118ms
match https://www.paulirish.com/javascripts/firebase-performance.js (63389): 7.584ms
match https://www.paulirish.com/javascripts/firebase-app.js (38293): 4.590ms
match https://www.paulirish.com/ (743): 0.250ms
match https://paulirish.disqus.com/count.js (1517): 0.217ms
match https://www.paulirish.com/ (1): 0.006ms
match https://www.paulirish.com/ (31): 0.028ms
match https://www.paulirish.com/javascripts/modernizr-2.0.js (9736): 1.163ms
match https://www.paulirish.com/javascripts/ender.js (56107): 7.008ms
match https://www.paulirish.com/javascripts/octopress.js (7820): 1.138ms
match https://www.paulirish.com/ (168): 0.065ms
match https://www.paulirish.com/ (395): 0.138ms
match https://www.paulirish.com/ (424): 0.161ms
audit: 39.681ms
```

https://www.nyt.com/
```
CodePatternMatcher ctor: 1.023ms
match https://platform.iteratehq.com/sdk-prod-f5756fbc6a6363b60247.js (821623): 99.109ms
match https://www.googletagservices.com/activeview/js/current/osd.js?cb=%2Fr20100101 (76535): 11.827ms
match https://securepubads.g.doubleclick.net/pagead/js/rum.js (48226): 5.728ms
match https://cdn.optimizely.com/public/3013110282/data/actions/313e3bca30f455ed3ffecf8a39e6df8613fe34d6ad0d12d1f6524daa460407a6.js (9738): 1.636ms
match https://cdn.optimizely.com/public/3013110282/data/actions/f1cecbabdb981796aa512e6315d6ad3dce6d007f0bcc46945f9fb3b28915257c.js (321): 0.044ms
match https://securepubads.g.doubleclick.net/gpt/pubads_impl_rendering_2019070801.js (68707): 8.122ms
match https://cdn3.optimizely.com/js/geo2.js (295): 0.147ms
match https://platform.iteratehq.com/loader.js (1261): 0.157ms
match https://static.chartbeat.com/js/chartbeat.js (36151): 5.541ms
match https://www.google-analytics.com/analytics.js (44287): 9.867ms
match https://static01.nyt.com/ads/google/adsbygoogle.js (28): 0.010ms
match https://www.googletagservices.com/tag/js/gpt.js (36781): 4.356ms
match https://www.nytimes.com/ (19914): 2.432ms
match https://news.google.com/swg/js/v1/swg.js (116057): 14.179ms
match https://www.nytimes.com/ (199): 0.042ms
match https://www.nytimes.com/ (306): 0.052ms
match https://securepubads.g.doubleclick.net/tag/js/gpt.js (36776): 4.348ms
match https://www.nytimes.com/ (285): 0.053ms
match https://www.nytimes.com/ (1848): 0.225ms
match https://contextual.media.net/bidexchange.js?cid=8CULO58R6&dn=www.nytimes.com&https=1 (288628): 34.089ms
match https://c.amazon-adsystem.com/aax2/apstag.js (70033): 8.318ms
match https://adservice.google.com/adsid/integrator.js?domain=www.nytimes.com (109): 0.018ms
match https://securepubads.g.doubleclick.net/gpt/pubads_impl_2019070801.js (153883): 18.165ms
match https://www.nytimes.com/vi-assets/static-assets/vendors~audio~home~paidpost~story~trending~video-3c86fff1a1bfa26e3810.js (20707): 2.482ms
match https://www.nytimes.com/vi-assets/static-assets/vendors~audio~home~paidpost~story-4981bb5b517480c3da4c.js (29302): 3.436ms
match https://www.nytimes.com/vi-assets/static-assets/vendors~home-19879e152701f77239ed.js (18042): 2.131ms
match https://www.nytimes.com/ (5816): 0.714ms
match https://a1.nyt.com/analytics/show-ads.js (45): 0.012ms
match https://www.nytimes.com/ (824227): 95.144ms
match https://www.nytimes.com/ (1539): 0.208ms
match https://www.nytimes.com/vi-assets/static-assets/adslot-c166ae2f982bbe19644d.js (9023): 1.076ms
match https://www.nytimes.com/ (4534): 0.544ms
match https://www.nytimes.com/vi-assets/static-assets/vendor-8c8e3ecfb65a01258107.js (1016079): 96.349ms
match https://www.nytimes.com/vi-assets/static-assets/home-b7b65917a84457be5fd9.js (357572): 33.563ms
match https://www.nytimes.com/vi-assets/static-assets/main-1a4d49bc9dbedb3d5310.js (394744): 36.985ms
match https://www.nytimes.com/ (735): 0.113ms
match https://www.googletagmanager.com/gtm.js?id=GTM-P528B3&gtm_auth=tfAzqo1rYDLgYhmTnSjPqw&gtm_preview=env-130&gtm_cookies_win=x (260204): 28.909ms
match https://www.nytimes.com/ (1532): 0.215ms
match https://contextual.media.net/hmmaster.js?cid=8CULO58R6&p=hb&v=3.1.3&https=1&adt=1 (19145): 1.867ms
match https://cdn.optimizely.com/public/3013110282/s/vi_article.js (680565): 67.796ms
match https://www.nytimes.com/ (1844): 0.196ms
match https://www.nytimes.com/ (738): 0.077ms
match https://www.nytimes.com/ (1220): 0.123ms
match https://www.nytimes.com/ (706): 0.072ms
match https://www.nytimes.com/ (7406): 0.730ms
match https://www.nytimes.com/ (1220): 0.122ms
match https://www.nytimes.com/ (199): 0.022ms
match https://www.nytimes.com/ (199): 0.022ms
match https://www.nytimes.com/ (19587): 1.925ms
match https://et.nytimes.com/?subject=page&url=https%3A%2F%2Fwww.nytimes.com%2F&assetUrl=https%3A%2F%2Fwww.nytimes.com%2F&referrer=&clientTimeStamp=1563409486951&totalTime=0&evtUUID=ca5ba944-a74f-4282-bc37-3a88fc45db9e&gtm=GTM-P528B3-322-Production&page_view_id=hs1X1zjmbt-hN6jks6gDk-BC&performance=%7B%22first-paint%22%3A249%2C%22first-contentful-paint%22%3A249%7D&presentation=%7B%22pageType%22%3A%22Homepage%22%2C%22program%22%3A%7B%22breakpoint%22%3A%7B%22initial%22%3A%22smartphone%22%7D%7D%7D&et2_pageview_id=hs1X1zjmbt-hN6jks6gDk-BC&sourceApp=nyt-vi&assetData=%7B%22url%22%3A%22https%3A%2F%2Fwww.nytimes.com%2F%22%2C%22updatedDate%22%3A1525181301000%2C%22wordCount%22%3A0%2C%22id%22%3A100000005877499%2C%22publishedDate%22%3A1525150800000%2C%22section%22%3A%22Homepage%22%2C%22derivedDesk%22%3A%22%22%2C%22type%22%3A%22sectionfront%22%2C%22jkiddSrc%22%3A%22fe%22%7D&skipFilter=%5B%22content%22%2C%22jkiddata%22%5D&callback=window.EventTracker.cb0&pageMetaData=%7B%22PT%22%3A%22Homepage%22%2C%22CG%22%3A%22Homepage%22%2C%22SCG%22%3A%22%22%2C%22byl%22%3A%22%22%2C%22tom%22%3A%22%22%2C%22hdl%22%3A%22%22%2C%22ptime%22%3A%22%22%2C%22cre%22%3A%22%22%2C%22articleid%22%3A100000005877499%2C%22channels%22%3A%22%22%2C%22CN%22%3A%22%22%2C%22CT%22%3A%22%22%2C%22backfilled%22%3A%22articleid%22%7D&additionalClientData=%7B%22ul%22%3A%22en-US%22%2C%22js%22%3A%22Yes%22%2C%22jv%22%3A%221.5%22%2C%22fi%22%3A%22No%22%2C%22tz%22%3A-7%2C%22bh%22%3A17%2C%22cd%22%3A24%2C%22sr%22%3A%22412x660%22%2C%22bs%22%3A%22412x660%22%2C%22jo%22%3A%22No%22%2C%22or%22%3A%22port%22%7D (105): 0.032ms
match https://www.nytimes.com/ (372): 0.039ms
match https://www.nytimes.com/ (4530): 0.444ms
match https://www.nytimes.com/ (2803): 0.276ms
match https://tags.bluekai.com/site/50550?ret=js&limit=1 (41): 0.013ms
match https://www.nytimes.com/ (1958): 0.195ms
match https://tags.bkrtx.com/js/bk-coretag.js (31380): 3.017ms
match https://www.nytimes.com/ (418): 0.044ms
match https://a1.nyt.com/analytics/comscore-streaming.js (105675): 10.108ms
match https://www.nytimes.com/ (504): 0.051ms
match https://www.nytimes.com/ (1236): 0.149ms
match https://www.nytimes.com/ (514): 0.052ms
match https://www.nytimes.com/ (661): 0.069ms
match https://www.nytimes.com/ (316): 0.034ms
match https://www.nytimes.com/ (7106): 0.693ms
match https://cigsvc.nytimes.com/r1/jp/ip_seg_v2.rep?&callback=callback (170): 0.020ms
audit: 682.113ms
```

connorjclark:
Moved this to a diagnostic, and changed the description to emphasize that the `module`/`nomodule` suggestion only applies to JS that is in a bundle. I'd like this to be considered for merging as a diagnostic, and improve upon the idea at a later date.

Future work:

1) Make `legacy-javascript` an opportunity

For the polyfills, this requires either bundle analysis to attribute the size of the polyfills, or a better parser than regex (there's `acorn`, a "small, fast" parser weighing in at [27.7KB gzipped](https://bundlephobia.com/result?p=acorn@6.2.0)!)

For transforms, this involves estimating how many bytes applying the transform costs. This would be different for each transform. I imagine for some transforms, the cost is exactly calculable with just regex (finding each fingerprint the transform leaves) and some basic arithmetic (factoring in how the transform generates code, and the constant size of the utility functions it requires). For others, I imagine calculating the changes to be less than an exact science, so perhaps an estimate could be made after compiling some real world applications and deriving a formula.

2) Detect with bundle analysis (#9097)

As a method of progressive enhancement, we can detect polyfills for files that provide source-maps based on well-known polyfill packages. For files that do not have maps, the existing RegExp will be the backup.

3) Use `source-location` (#9354)

If that PR merges first, I'll update this PR to utilize it. Otherwise, I'll do it after.

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+
+const Audit = require('./audit.js');
+const NetworkRecords = require('../computed/network-records.js');
+const i18n = require('../lib/i18n/i18n.js');
+
+const UIStrings = {
+  /** Title of a Lighthouse audit that tells the user about legacy polyfills and transforms used on the page. This is displayed in a list of audit titles that Lighthouse generates. */
+  title: 'Legacy JavaScript',
+  // eslint-disable-next-line max-len
+  description: 'Polyfills and transforms enable legacy browsers to use new JavaScript features. However, many aren\'t necessary for modern browsers. For your bundled JavaScript, adopt a modern script deployment strategy using module/nomodule feature detection to reduce the amount of code shipped to modern browsers, while retaining support for legacy browsers. [Learn More](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/)',
```
exterkamp: Split this into ''+'' for line length, lets keep it below 100. Is there a web.dev link for this maybe?  @egsweeny

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+/** @typedef {{name: string, line: number, column: number}} PatternMatchResult */
+
+const Audit = require('./audit.js');
+const NetworkRecords = require('../computed/network-records.js');
+const i18n = require('../lib/i18n/i18n.js');
+
+const UIStrings = {
+  /** Title of a Lighthouse audit that tells the user about legacy polyfills and transforms used on the page. This is displayed in a list of audit titles that Lighthouse generates. */
+  title: 'Legacy JavaScript',
+  // eslint-disable-next-line max-len
```
exterkamp: Add a description string.

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
+      {key: 'description', itemType: 'code', text: 'Description'},
+      {key: 'location', itemType: 'code', text: 'Location'},
+    ];
+    const details = Audit.makeTableDetails(headings, tableRows);
+
+    return {
+      score: Number(patternMatchCounter.size === 0),
+      extendedInfo: {
```
exterkamp:
```javascript
/** Deprecated and does not make its way into the Lighthouse report. */
extendedInfo?: {[p: string]: any};
```

Deprecated?  DisplayValue, or numericValue perhaps?

lighthouse-core/test/audits/legacy-javascript-test.js
```diff
@@ -0,0 +1,166 @@
...
+      acc[String(index)] = {
+        content: code,
+        requestId: String(index),
+      };
+      return acc;
+    }, {}),
+  };
+};
+
+/* eslint-env jest */
```
exterkamp: Move to just below `use strict;`/`const`'s

lighthouse-core/test/audits/legacy-javascript-test.js
```diff
@@ -0,0 +1,166 @@
...
+describe('LegacyJavaScript audit', () => {
```
exterkamp: Use readable name

```suggestion
describe('Legacy Javascript audit', () => {
```

lighthouse-core/test/audits/legacy-javascript-test.js
```diff
@@ -0,0 +1,166 @@
...
+  it('code with no polyfills passes', async () => {
```
exterkamp: test names should be a complete sentence with the 'it'

```suggestion
  it('passes code with no polyfills', async () => {
```

lighthouse-core/test/audits/legacy-javascript-test.js
```diff
@@ -0,0 +1,166 @@
...
+        code: 'Object.defineProperty(SomeClass.prototype, "someFn", function() {})',
+        url: 'https://www.example.com/a.js',
+      },
+    ]);
+    const result = await LegacyJavascript.audit(artifacts, {computedCache: new Map()});
+    assert.equal(result.score, 1);
+    assert.equal(result.extendedInfo.value, 0);
+  });
+
+  it('code with a polyfill fails', async () => {
```
exterkamp: I don't know if I like "fails" being in here.  Since the audit doesn't "fail".  maybe `it('identifies code with a polyfill..`

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+        patternMatchCounter.set(match, val);
+        patternCounter.set(match.name, val + 1);
+      }
+    }
+
+    return {patternCounter, patternMatchCounter, urlToMatchResults};
+  }
+
+  /**
+   * @param {LH.Artifacts} artifacts
```
exterkamp: Definite need a comment here, this `audit` is very heavy and needs some explanation, maybe even an _ascii diagram_!

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+          url,
+          description,
+          location: `Ln: ${line}, Col: ${column}`,
+        });
+      }
+    });
+
+    /** @type {LH.Audit.Details.Table['headings']} */
+    const headings = [
+      {key: 'url', itemType: 'url', text: 'URL'},
```
exterkamp:
```suggestion
      {key: 'url', itemType: 'url', text: str_(i18n.UIStrings.columnURL)},
```

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+      {key: 'description', itemType: 'code', text: 'Description'},
+      {key: 'location', itemType: 'code', text: 'Location'},
```
exterkamp:
```suggestion
      {key: 'location', itemType: 'code', text: str_(i18n.UIStrings.columnLocation)},
```

lighthouse-core/audits/legacy-javascript.js
connorjclark: but why. word wrap is a thing

lighthouse-core/test/audits/legacy-javascript-test.js
connorjclark: other tests use the class name (see unused js test)

lighthouse-core/audits/legacy-javascript.js
connorjclark: oh ya

connorjclark: wait what. there is a `description` key here

lighthouse-core/audits/legacy-javascript.js
connorjclark: also +1 to the web.dev thing. there is @housseindjirdeh guide on web.dev which I also want to link here but I forgot how to find it

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,334 @@
...
+
+    return expression;
+  }
+
+  /**
+   * @return {Pattern[]}
+   */
+  static getPolyfillPatterns() {
+    return [
+      'Array.prototype.filter',
```
connorjclark: this is everything from that spreadsheet, minus ones that don't fit the model (the `*[Symbol.*]` ones) and the ones I figured were too much (`Math.*` mostly)

lighthouse-core/test/audits/legacy-javascript-test.js
connorjclark: this doesn't match other test files i see.

lighthouse-core/audits/legacy-javascript.js
patrickhulce: I think he means the `/**` description for i18n :)

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+    const seen = new Set();
+    /** @type {PatternMatchResult[]} */
+    const matches = [];
+    /** @type {RegExpExecArray | null} */
+    let result;
+    let line = 0;
+    let lineBeginsAtIndex = 0;
+    // Each pattern maps to one subgroup in the generated regex. For each iteration of RegExp.exec,
+    // only one subgroup will be defined. Exec until no more matches.
+    this.re.lastIndex = 0;
```
patrickhulce: is this resetting for the loop if `.match` is called more than once? I was a bit confused by the comment immediately above it, could move a line out of the way perhaps? :)

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+    while ((result = this.re.exec(code)) !== null) {
+      // Discard first - it's the whole matching pattern.
```
patrickhulce:
```suggestion
      // Index 0 - the entire match, discard.
```

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+      // Index 1 is truthy if matching a newline, and is used to track the line number.
```
patrickhulce:
```suggestion
      // Index 1 - truthy if matching a newline, used to track the line number.
```

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+      // `patternExpressionMatches` maps to each possible pattern.
```
patrickhulce:
```suggestion
      // Indexes 2+ - map to the potential match for each pattern in `this.patterns`.
```

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+      // Only one of [isNewline, ...patternExpressionMatches] is ever defined.
```
patrickhulce:
```suggestion
      // Only one element with index >=1 is ever defined.
```

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+      'Object.getOwnPropertyDescriptors',
+      'Object.values',
+      'String.prototype.padEnd',
+      'String.prototype.padStart',
+    ].map(str => {
+      const parts = str.split('.');
+      const object = parts.length > 1 ? parts.slice(0, parts.length - 1).join('.') : null;
+      const property = parts[parts.length - 1];
+      return {
+        name: `${object ? object + '.' : ''}${property}`,
```
patrickhulce: isn't this just piecing `str` back together again or am I missing something?

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+      'Uint8ClampedArray',
+      'WeakMap',
+      'WeakSet',
+      'Array.prototype.includes',
+      'Object.entries',
+      'Object.getOwnPropertyDescriptors',
+      'Object.values',
+      'String.prototype.padEnd',
+      'String.prototype.padStart',
+    ].map(str => {
```
patrickhulce: s/str/polyfillName/

maybe?

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+        const isMoreThanOne = patternOccurrences > 1;
+        const matchOrdinal = patternMatchCounter.get(match) || 0;
+        // Only show ordinal if there is more than one occurrence across all scripts.
+        const description = isMoreThanOne ?
+          `${name} (${matchOrdinal + 1} / ${patternOccurrences})` :
+          name;
+        tableRows.push({
+          url,
+          description,
+          location: `Ln: ${line}, Col: ${column}`,
```
patrickhulce: `source-location` candidate? :)

lighthouse-core/test/audits/legacy-javascript-test.js
```diff
@@ -0,0 +1,166 @@
...
+        code: 'String.prototype.repeat = function() {}; String.prototype.includes = function() {}',
+        url: 'https://www.example.com/a.js',
+      },
+    ]);
+    const result = await LegacyJavascript.audit(artifacts, {computedCache: new Map()});
+    assert.equal(result.score, 0);
+    assert.equal(result.numericValue, 2);
+  });
+
+  it('counts multiple of the same polyfill from the same script only once', async () => {
```
patrickhulce: what's the motivation here? shouldn't it be flagged if they're shipping duplicate polyfills in the same bundle?

lighthouse-core/audits/legacy-javascript.js
```diff
@@ -0,0 +1,344 @@
...
+
+    // Object.defineProperty(String.prototype, 'startsWith'
+    expression += `|defineProperty\\(${object || 'window'},\\s*${qt(property)}`;
+
+    if (object) {
+      const objectWithoutPrototype = object.replace('.prototype', '');
+      // e(e.S,"Object",{values
+      // minified pattern found in babel-polyfill
+      // see https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.2.5/polyfill.min.js
+      expression += `|;e\\([^,]+,${qt(objectWithoutPrototype)},{${property}`;
```
patrickhulce: this one seems like the most brittle, is it fairly certain it will be minified as `e` every time?

lighthouse-core/audits/legacy-javascript.js
connorjclark: yea

lighthouse-core/audits/legacy-javascript.js
connorjclark: yeah.... um..... lol

lighthouse-core/audits/legacy-javascript.js
connorjclark: that's the dream

lighthouse-core/test/audits/legacy-javascript-test.js
connorjclark: this bit:

```
// Don't report more than one instance of a pattern for this code.
      // Would result in multiple matches for the same pattern, ex: if both '='
      // and 'Object.defineProperty' are used conditionally based on feature detection.
      // Would also result in many matches for transform patterns.
```

Can't distinguish between multiple polyfills (somehow) making their way in and just the one polyfill doing their polyfilling in multiple ways. At least, not without a more complicated matching process.

lighthouse-core/audits/legacy-javascript.js
connorjclark: Oh boy :) _cracks knuckles_ tl;dr no, also I overlooked so much, thanks for pointing this out.

The minified JS I linked to is every polyfill that `@babel/present-env` would provide. The recommended distribution of these polyfills is not the module I linked (@babel/polyfills), but instead via `@babel/preset-env`. Regardless, they both use the polyfills as defined by `core-js`. You can see that the polyfills module has a standard bundling and minifying script: https://github.com/babel/babel/blob/fced5cea430cc00e916876b663a8d2a84a5dad1f/packages/babel-polyfill/scripts/build-dist.sh, which is what I linked to before. The question is: does the pattern I'm looking for hold for these `core-js` polyfills regardless of the minifier used? Additionally: where does this pattern come from? I'll focus on that first.

To further complicate, there's `core-js@2` and `core-js@3`, and both seem to be used (or at least, supported in `@babel/preset-env`).

Let's start with `core-js@2`.

## core-js@2

There's a few different mechanisms for defining polyfills.

### _export

Here's a typical polyfill: https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/es6.array.fill.js
```js
// 22.1.3.6 Array.prototype.fill(value, start = 0, end = this.length)
var $export = require('./_export');

$export($export.P, 'Array', { fill: require('./_array-fill') });

require('./_add-to-unscopables')('fill');
```

Here's another: https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/es6.array.find-index.js
```js
'use strict';
// 22.1.3.9 Array.prototype.findIndex(predicate, thisArg = undefined)
var $export = require('./_export');
var $find = require('./_array-methods')(6);
var KEY = 'findIndex';
var forced = true;
// Shouldn't skip holes
if (KEY in []) Array(1)[KEY](function () { forced = false; });
$export($export.P + $export.F * forced, 'Array', {
  findIndex: function findIndex(callbackfn /* , that = undefined */) {
    return $find(this, callbackfn, arguments.length > 1 ? arguments[1] : undefined);
  }
});
require('./_add-to-unscopables')(KEY);
```

[`$export`](https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/_export.js) is a helper function to define an extension to the provided class. The first parameter is a bitmap.
```js
// type bitmap
$export.F = 1;   // forced
$export.G = 2;   // global
$export.S = 4;   // static
$export.P = 8;   // proto
$export.B = 16;  // bind
$export.W = 32;  // wrap
$export.U = 64;  // safe
$export.R = 128; // real proto method for `library`
```

So any number of these flags, but at least one, are passed in as the first parameter. The second parameter the name of the class. The third is an object whose keys are properties/methods that will be added.

### _redefine

There's also a `redefine` helper function, that looks pretty simple: https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/es6.regexp.to-string.js#L10
```js
require('./_redefine')(RegExp.prototype, TO_STRING, fn, true);
```

That constant `TO_STRING` is interesting - I'm seeing that some polyfills use constants for the method name, others don't. Should be the same code post-minification (it'd get inlined), but without minifying the difference is maybe something to consider.

### _collection

Polyfills that create a new collection class uses `_collection`.

https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/es6.set.js
```js
'use strict';
var strong = require('./_collection-strong');
var validate = require('./_validate-collection');
var SET = 'Set';

// 23.2 Set Objects
module.exports = require('./_collection')(SET, function (get) {
  return function Set() { return get(this, arguments.length > 0 ? arguments[0] : undefined); };
}, {
  // 23.2.3.1 Set.prototype.add(value)
  add: function add(value) {
    return strong.def(validate(this, SET), value = value === 0 ? 0 : value, value);
  }
}, strong);
```

### _typed-array

https://github.com/zloirock/core-js/blob/6a3fe85136aaae0e3b099c96a05a5ceb1f515a50/modules/es6.typed.float32-array.js
```js
require('./_typed-array')('Float32', 4, function (init) {
  return function Float32Array(data, byteOffset, length) {
    return init(this, data, byteOffset, length);
  };
});
```

For these, we could combine each typed-array variant into one check by looking for code you'd find in `require('./_typed-array')`.

## core-js@3

Different patterns here.

### export

https://github.com/zloirock/core-js/blob/5adfcd91edfce629ced1eab76a303f65a97e50cb/packages/core-js/modules/es.array.fill.js
```js
var $ = require('../internals/export');
var fill = require('../internals/array-fill');
var addToUnscopables = require('../internals/add-to-unscopables');

// `Array.prototype.fill` method
// https://tc39.github.io/ecma262/#sec-array.prototype.fill
$({ target: 'Array', proto: true }, {
  fill: fill
});

// https://tc39.github.io/ecma262/#sec-array.prototype-@@unscopables
addToUnscopables('fill');
```

### collection, redefine, typed-array.

No change really.

### quicky ditty on minifying

This command: `npx browserify node_modules/core-js/modules/es6.array.find.js | npx uglify-js --mangle keep_fnames`

will transform this:
```js
$export($export.P + $export.F * forced, 'Array', {
  find: function find(
```

to this:
```js
e(e.P+e.F*a,"Array",{find:function find(
```

Drop the mangling, and you get:
```js
$export($export.P+$export.F*forced,"Array",{find:function find(
```

I'm really curious how uglifyjs picks the short name it can use, and why it seems to commonly be `e`. The answer lies somewhere in here: https://github.com/mishoo/UglifyJS2/blob/e8a2c0b5bf18659a3e1285b6038ea755a290220d/lib/scope.js

## so...

With this extra domain knowledge, I'd like to create the regex that'd match each of these few different patterns. And the pattern should survive typical minification. I'd want to verify these patterns too - could bundle each polyfill individually, minify, and apply the pattern.

patrickhulce: *cracks knuckles* is right! 😆 

that all sounds great but like a signficant increase to scope, followup?

connorjclark: > that all sounds great but like a signficant increase to scope, followup?

yeah. luckily it's a lot of the same work that'd be needed for making this an opportunity.

lighthouse-cli/test/smokehouse/legacy-javascript-expectations.js
```diff
@@ -0,0 +1,30 @@
...
+      requestedUrl: 'http://localhost:10200/legacy-javascript.html',
+      finalUrl: 'http://localhost:10200/legacy-javascript.html',
+      audits: {
+        'legacy-javascript': {
+          details: {
+            items: {
+              // 42 are from babel-polyfill
+              // 1 from fixtures/legacy-javascript.js
+              // 1 inline from fixtures/legacy-javascript.html
+              length: 44,
```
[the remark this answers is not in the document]
connorjclark: done

connorjclark:
Because GitHub's way of eliding comments is so awful, here is the last significant comment I made summarizing things: https://github.com/GoogleChrome/lighthouse/pull/6730#issuecomment-513423305

connorjclark:
Too many comments here, so I am closing this PR. Will reopen after making the updates discussed offline.

PR #6739, opened 2018-12-06
core(driver): handle promise rejections from sendCommand

exterkamp:
**Summary**
As of #6347 driver's `sendCommand` can reject its Promise if the underlying protocol command times out.  This new rejection path isn't handled in some invocations of sendCommand leading to unhandled promise rejections.  This would present as us "missing" a PROTOCOL_TIMEOUT in LR, leading to the graceful resolve fix push; however, upon investigation, the cause of missing those timeouts was the timeouts causing an unhandled promise rejection which freaks out LR!  
This should catch all unhandled promise rejections as of right now.  I did a lot of testing to try to track down unhandled promise rejection, but I might have missed some, so if you find any lmk.

**Related Issues/PRs**
- makes #6671 irrelevant
- #6336 doesn't fix this, but it does help LR's story and removes the need for more error handling, because the errors are handled now!

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -628,14 +628,13 @@ class Driver {
     /**
      * @param {Driver} driver
      * @param {() => void} resolve
-     * @param {(error:Error) => void} reject
      */
-    function checkForQuiet(driver, resolve, reject) {
-      if (cancelled) return;
+    function checkForQuiet(driver, resolve) {
+      if (cancelled) resolve();
 
       return driver.evaluateAsync(checkForQuietExpression)
         .then(timeSinceLongTask => {
-          if (cancelled) return;
+          if (cancelled) resolve();
 
           if (typeof timeSinceLongTask === 'number') {
             if (timeSinceLongTask >= waitForCPUQuiet) {
@@ -644,10 +643,10 @@ class Driver {
             } else {
               log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
               const timeToWait = waitForCPUQuiet - timeSinceLongTask;
-              lastTimeout = setTimeout(() => checkForQuiet(driver, resolve, reject), timeToWait);
+              lastTimeout = setTimeout(() => checkForQuiet(driver, resolve), timeToWait);
             }
           }
-        }).catch(e => reject(e));
+        });
     }
 
     /** @type {(() => void)} */
@@ -655,7 +654,7 @@ class Driver {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
     const promise = new Promise((resolve, reject) => {
-      checkForQuiet(this, resolve, reject);
+      checkForQuiet(this, resolve).catch(reject);
       cancel = () => {
         cancelled = true;
         if (lastTimeout) clearTimeout(lastTimeout);
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -628,8 +628,9 @@ class Driver {
     /**
      * @param {Driver} driver
      * @param {() => void} resolve
+     * @param {(error:Error) => void} reject
      */
-    function checkForQuiet(driver, resolve) {
+    function checkForQuiet(driver, resolve, reject) {
       if (cancelled) return;
 
       return driver.evaluateAsync(checkForQuietExpression)
@@ -643,10 +644,10 @@ class Driver {
             } else {
               log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
               const timeToWait = waitForCPUQuiet - timeSinceLongTask;
-              lastTimeout = setTimeout(() => checkForQuiet(driver, resolve), timeToWait);
+              lastTimeout = setTimeout(() => checkForQuiet(driver, resolve, reject), timeToWait);
             }
           }
-        });
+        }).catch(e => reject(e));
     }
 
     /** @type {(() => void)} */
@@ -654,7 +655,7 @@ class Driver {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
     const promise = new Promise((resolve, reject) => {
-      checkForQuiet(this, resolve);
+      checkForQuiet(this, resolve, reject);
       cancel = () => {
         cancelled = true;
         if (lastTimeout) clearTimeout(lastTimeout);
@@ -893,8 +894,8 @@ class Driver {
     // These can 'race' and that's OK.
     // We don't want to wait for Page.navigate's resolution, as it can now
     // happen _after_ onload: https://crbug.com/768961
-    this.sendCommand('Page.enable');
-    this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
+    await this.sendCommand('Page.enable');
+    await this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
     // No timeout needed for Page.navigate. See #6413.
     this._innerSendCommand('Page.navigate', {url});
 
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -630,11 +630,11 @@ class Driver {
      * @param {() => void} resolve
      */
     function checkForQuiet(driver, resolve) {
-      if (cancelled) resolve();
+      if (cancelled) return Promise.resolve();
 
       return driver.evaluateAsync(checkForQuietExpression)
         .then(timeSinceLongTask => {
-          if (cancelled) resolve();
+          if (cancelled) return Promise.resolve();
 
           if (typeof timeSinceLongTask === 'number') {
             if (timeSinceLongTask >= waitForCPUQuiet) {
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -651,7 +651,7 @@ class Driver {
     let cancel = () => {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
-    const promise = new Promise(async (resolve, reject) => {
+    const promise = new Promise((resolve, reject) => {
       checkForQuiet(this, resolve).catch(reject);
       cancel = () => {
         cancelled = true;
@@ -911,7 +911,7 @@ class Driver {
           maxWaitMs);
     }
 
-    // Awaiting so that errors can be properly processed.
+    // Bring `Page.navigate` errors back into the promise chain. See #6739.
     await waitforPageNavigateCmd;
 
     return this._endNetworkStatusMonitoring();
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -628,33 +628,35 @@ class Driver {
     /**
      * @param {Driver} driver
      * @param {() => void} resolve
+     * @return {Promise<void>}
      */
-    function checkForQuiet(driver, resolve) {
-      if (cancelled) return Promise.resolve();
+    async function checkForQuiet(driver, resolve) {
+      if (cancelled) return;
+      const timeSinceLongTask = await driver.evaluateAsync(checkForQuietExpression);
+      if (cancelled) return;
 
-      return driver.evaluateAsync(checkForQuietExpression)
-        .then(timeSinceLongTask => {
-          if (cancelled) return Promise.resolve();
-
-          if (typeof timeSinceLongTask === 'number') {
-            if (timeSinceLongTask >= waitForCPUQuiet) {
-              log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
-              resolve();
-            } else {
-              log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
-              const timeToWait = waitForCPUQuiet - timeSinceLongTask;
-              lastTimeout = setTimeout(() => checkForQuiet(driver, resolve), timeToWait);
-            }
-          }
-        });
+      if (typeof timeSinceLongTask === 'number') {
+        if (timeSinceLongTask >= waitForCPUQuiet) {
+          log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
+          resolve();
+        } else {
+          log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
+          const timeToWait = waitForCPUQuiet - timeSinceLongTask;
+          lastTimeout = setTimeout(() => checkForQuiet(driver, resolve), timeToWait);
+        }
+      }
     }
 
     /** @type {(() => void)} */
     let cancel = () => {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
-    const promise = new Promise((resolve, reject) => {
-      checkForQuiet(this, resolve).catch(reject);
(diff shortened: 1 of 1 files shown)


lighthouse-core/gather/driver.js
```diff
@@ -893,8 +894,8 @@ class Driver {
     // These can 'race' and that's OK.
     // We don't want to wait for Page.navigate's resolution, as it can now
     // happen _after_ onload: https://crbug.com/768961
-    this.sendCommand('Page.enable');
-    this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
+    await this.sendCommand('Page.enable');
```
exterkamp: Is it okay to await these?  The comments above them talk about racing like this might be purposefully not-awaited.

Hoten: Bug is marked as fixed now. If it truly is, can you also remove the comments?

lighthouse-core/gather/driver.js
```diff
@@ -643,18 +644,18 @@ class Driver {
...
+        }).catch(e => reject(e));
     }
 
     /** @type {(() => void)} */
     let cancel = () => {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
     const promise = new Promise((resolve, reject) => {
-      checkForQuiet(this, resolve);
+      checkForQuiet(this, resolve, reject);
```
Hoten: what happens if reject is called twice? is that possible here?

patrickhulce: IIRC, once a promise has been resolved/rejected all further calls will be noops

lighthouse-core/gather/driver.js
```diff
@@ -628,8 +628,9 @@ class Driver {
     /**
      * @param {Driver} driver
      * @param {() => void} resolve
+     * @param {(error:Error) => void} reject
      */
-    function checkForQuiet(driver, resolve) {
+    function checkForQuiet(driver, resolve, reject) {
```
patrickhulce: WDYT about just catching the function call instead of adding another parameter here?

lighthouse-core/gather/driver.js
patrickhulce:
```suggestion
      checkForQuiet(this, resolve).catch(reject);
```

lighthouse-core/gather/driver.js
patrickhulce: the big bug was definitely about not awaiting `Page.navigate`, but why are we needing to `await` these? is it just about unhandled rejections?

lighthouse-core/gather/driver.js
patrickhulce: oh you'd need to change the `return;` to `return Promise.resolve();` too, but I'm not sure I follow what impact the resolve `() => void` is? Maybe I'm skipping a step somewhere :)

lighthouse-core/gather/driver.js
patrickhulce: right I'm onboard with that, but I'm asking because there are then other things we might try instead of serialling awaiting like `this._sendAndSilentlyIgnoreErrors` or something :)

It looks like we used to be just fine waiting on these in sequence though, and the `Page.navigate` change just moved all of 'em to be together, so we don't need to go crazy avoiding the `await` https://github.com/GoogleChrome/lighthouse/pull/3413

I'm a tad spooked by this given that it was a real nasty bug to hunt down, but I think this makes sense if the bug really is fixed. +1 to removing the comments if we're making this change

lighthouse-core/gather/driver.js
exterkamp: I was misinterpreting the linting hint as being important, it was lying to me!

lighthouse-core/gather/driver.js
exterkamp: I like it!  Couldn't get it to work before b/c I was stuck up on the `return;` never thought to _resolve_ them oops!

lighthouse-core/gather/driver.js
exterkamp: Hmmm, I think I might have explained this wrong?  Or mis-titled this PR.  I want these functions to loudly throw their underlying error (PROTOCOL_TIMEOUT) upwards when they reject their promises.  So I don't want to silently ignore errors I just want to handle the unhandled rejections that `sendCommand` can throw now instead of crashing LH for not handling a rejected-promise. 

Is await the best way to do that?  Or are you suggesting some kind of promise catching alternative?

patrickhulce: Oh, so we *do* want these two errors to be fatal? Yeah `await`'ing serially like we were before SGTM, sounds like we also want an empty catch on the `Page.navigate` then as well?

exterkamp: AFAIK since `Page.navigate` uses `_innerSendCommand` directly instead of the `driver.sendCommand` wrapper which has the timeout configured around it I don't think it can reject?  I was only force rejecting promises from `driver.sendCommand` though in my testing, so...?

interrogative: Do we need protocol timeouts for `_innerSendCommand`? What if those timeout, what happens?  Or is that specifically only called directly when we don't want a timeout e.g. `Page.navigate`?

lighthouse-core/gather/driver.js
```diff
@@ -630,11 +630,11 @@ class Driver {
      * @param {() => void} resolve
      */
     function checkForQuiet(driver, resolve) {
-      if (cancelled) return;
+      if (cancelled) resolve();
```
patrickhulce: hang on might need to walk me through this, it was slightly different than what I was thinking :)

if it's been cancelled, we want to stop continuing and our calls to resolve don't really matter since they'll be ignored anyhow.

don't we want
```js
if (cancelled) return Promise.resolve()
```

seems like the current code will still kick off the `evaluateAsync` loop

lighthouse-core/gather/driver.js
Hoten: `_innerSendCommand` is specifically for when you don't want to timeout.

I'm not sure if `_innerSendCommand` can ever reject. You could try setting a breakpoint just before, then kill Chrome, then hit play, and see what happens ...

patrickhulce: Or I guess `sendCommand` is never the real problem it's always the PROTOCOL_TIMEOUT that's the real problem?

lighthouse-core/gather/driver.js
exterkamp: Oh yeah, good catch, I was confusing my `resolves` since it takes in the resolve, but in the cancelled case we want to return an empty resolved promise & return afaik, right?

patrickhulce: > return an empty resolved promise & return afaik

Yep, that checks out to me!

lighthouse-core/gather/driver.js
brendankenny: maybe I'm not following the conversation correctly here, but `_innerSendCommand` can reject via `connection`:

https://github.com/GoogleChrome/lighthouse/blob/9ebf0692de37bd772c59280dd63071bb9357989e/lighthouse-core/gather/connections/connection.js#L120-L124

(throwing inside `Promise.resolve().then(_ => {/* throw */})` rejects it, then `callback.resolve()` with a rejected promise rejects as well. We could probably make that a lot more understandable with an explicit `callback.resolve` *or* `callback.reject`, as well as using an `async` function for the inner part)

lighthouse-core/gather/driver.js
```diff
@@ -630,11 +630,11 @@ class Driver {
      * @param {() => void} resolve
      */
     function checkForQuiet(driver, resolve) {
```
brendankenny: `_waitForCPUIdle` is definitely confusing :)

It would be nice to simplify things a little bit while in here. There are a few ways to do that, but one simple way (accomplishing the same thing you're doing here), just switching to `async` and flattening the `evaluateAsync` call:
```js
/**
 * @param {Driver} driver
 * @param {() => void} resolve
 * @return {Promise<void>}
 */
async function checkForQuiet(driver, resolve) {
  if (cancelled) return;
  const timeSinceLongTask = await driver.evaluateAsync(checkForQuietExpression);
  if (cancelled) return;

  if (typeof timeSinceLongTask === 'number') {
    if (timeSinceLongTask >= waitForCPUQuiet) {
      log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
      resolve();
    } else {
      log.verbose('Driver', `CPU has been idle for ${timeSinceLongTask} ms`);
      const timeToWait = waitForCPUQuiet - timeSinceLongTask;
      lastTimeout = setTimeout(() => checkForQuiet(driver, resolve), timeToWait);
    }
  }
}
```

lighthouse-core/gather/driver.js
brendankenny: so do we want to add an empty `catch` to the `this._innerSendCommand('Page.navigate', {url});` line? Or are we assuming it's so rare that it's ok that a `unhandledRejection` listener can catch it?

lighthouse-core/gather/driver.js
```diff
@@ -893,8 +893,8 @@ class Driver {
     // These can 'race' and that's OK.
     // We don't want to wait for Page.navigate's resolution, as it can now
     // happen _after_ onload: https://crbug.com/768961
```
brendankenny: as others noted, should remove these comments now

lighthouse-core/gather/driver.js
exterkamp: Since it seems like this _never_ happens, should we just let sleeping dragons lie?

lighthouse-core/gather/driver.js
```diff
@@ -890,11 +888,8 @@ class Driver {
...
 
-    // These can 'race' and that's OK.
-    // We don't want to wait for Page.navigate's resolution, as it can now
-    // happen _after_ onload: https://crbug.com/768961
-    this.sendCommand('Page.enable');
-    this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
+    await this.sendCommand('Page.enable');
+    await this.sendCommand('Emulation.setScriptExecutionDisabled', {value: disableJS});
     // No timeout needed for Page.navigate. See #6413.
     this._innerSendCommand('Page.navigate', {url});
```
brendankenny: ok, after talking this through for a little while, we think we do want to incorporate the `Page.navigate` errors into the promise chain. It's a little ugly, so if anyone has any other structure ideas, please chime in :)

Basically it would be something like

```js
const waitforPageNavigateCmd = this._innerSendCommand('Page.navigate', {url});
// ...
let pageWaiter = Promise.resolve();
if (waitForNavigated) {
  pageWaiter = this._waitForFrameNavigated();
} else if (waitForLoad) {
  // ...
  pageWaiter = this._waitForFullyLoaded(/* ... */);
}

await Promise.all([
  waitforPageNavigateCmd,
  pageWaiter,
]);
```
- If `Page.navigate` has no problem, as before it doesn't block or interfere with observing the page load
- If `Page.navigate` throws, it wins the `Promise.all` with a rejection, which leads to lighthouse exit.

The main issue I see (besides reduced readability) is that we don't clean up the `_waitForFullyLoaded` machinery. This is a fatal error so it doesn't matter in most clients, but are we going to have to actually clean up to support when LH is run as a module?

brendankenny: also during error testing we discovered that `_waitForFrameNavigated` should probably also have a timeout instead of blindly waiting on the event.

patrickhulce: > we discovered that _waitForFrameNavigated should probably also have a timeout

WDYT about having an option to `this.once` that does the timeout bit?

patrickhulce: > is that we don't clean up the _waitForFullyLoaded machinery

why don't we just await `waitForPageNavigateCmd` as the ultimate very last thing? the only thing we want to happen is get the errors into the chain, right? if we do it as the very last thing, then it can't prevent other cleanup

brendankenny: > WDYT about having an option to this.once that does the timeout bit?

yeah, good idea. We have at least one or two places we will happily wait forever on an event.

brendankenny: though the service worker one I'm thinking of only waits forever on `about:blank`, which adding an `_waitForFrameNavigated ` timeout will fix :)

brendankenny: > why don't we just await `waitForPageNavigateCmd` as the ultimate very last thing? the only thing we want to happen is get the errors into the chain, right? if we do it as the very last thing, then it can't prevent other cleanup

This would definitely be easier to follow.

The ordering will be a little wonky (I'd imagine a `Page.navigate` failure would end up usually failing on another command within `_waitForFullyLoaded` or by hitting the 45 second timeout), but ordering is already an issue with `Promise.all()` (and without cancelling `_waitForFullyLoaded`), and we've already all agreed that `Page.navigate` rarely if ever fails :)

lighthouse-core/gather/driver.js
```diff
@@ -628,33 +628,31 @@ class Driver {
...
+      }
     }
 
     /** @type {(() => void)} */
     let cancel = () => {
       throw new Error('_waitForCPUIdle.cancel() called before it was defined');
     };
-    const promise = new Promise((resolve, reject) => {
-      checkForQuiet(this, resolve);
+    const promise = new Promise(async (resolve, reject) => {
```
brendankenny: no `async` needed

lighthouse-core/gather/driver.js
```diff
@@ -890,19 +888,14 @@ class Driver {
...
     // No timeout needed for Page.navigate. See #6413.
-    this._innerSendCommand('Page.navigate', {url});
+    const waitforPageNavigateCmd = this._innerSendCommand('Page.navigate', {url});
 
     if (waitForNavigated) {
       await this._waitForFrameNavigated();
-    }
-
-    if (waitForLoad) {
+    } else if (waitForLoad) {
```
brendankenny: 👍

lighthouse-core/gather/driver.js
```diff
@@ -918,6 +911,9 @@ class Driver {
           maxWaitMs);
     }
 
+    // Awaiting so that errors can be properly processed.
```
patrickhulce: nit: while this comment makes total sense in the context of this PR, I can definitely see myself coming back in a year and being like "what?" maybe a link to this PR or a tad more explanation?

lighthouse-core/gather/driver.js
```diff
@@ -918,6 +911,9 @@ class Driver {
...
+    await waitforPageNavigateCmd;
```
patrickhulce: there's nothing in _endNetworkStatusMonitoring we need, right?

lighthouse-core/gather/driver.js
brendankenny: good call. Maybe even just something like `// Bring any 'Page.navigate' errors back into the promise chain.` and a link back here.

The comment above the `Page.navigate` command isn't terribly helpful either, but I don't have a great suggestion :)

lighthouse-core/gather/driver.js
brendankenny: > there's nothing in _endNetworkStatusMonitoring we need, right?

it returns the `finalUrl` after redirects, which is used for updating `passContext` (and eventually the artifact).

A little too convoluted but somewhat difficult to fix since `networkRecords` don't entirely exist by that point.

PR #6745, opened 2018-12-07
core(minification): properly handle regex character classes

### LINKED ISSUE #6744: Minification estimation erroneously reports very large savings
While `minification-estimator.js` currently correctly [ignores license comments](https://github.com/GoogleChrome/lighthouse/blob/54c2b5df7fdcbc289c2ac5942e2d17efa5a8194c/lighthouse-core/lib/minification-estimator.js#L108), it seems like JavaScript and Cascading Style Sheets source mapping comments like `//# sourceMappingURL=foo.min.js.map` or `/*# sourceMappingURL=foo.css.map */` are still counted, which seems incorrect.

patrickhulce:
**Summary**
The minification estimator heuristics weren't accounting for regex character classes properly which meant that it could leave a regex prematurely.

**Related Issues/PRs**
fixes #6744

### DIFF
--- a/lighthouse-core/lib/minification-estimator.js
+++ b/lighthouse-core/lib/minification-estimator.js
@@ -47,6 +47,7 @@ function computeTokenLength(content, features) {
   let isInLicenseComment = false;
   let isInString = false;
   let isInRegex = false;
+  let isInRegexCharacterClass = false;
   let stringOpenChar = null;
 
   for (let i = 0; i < content.length; i++) {
@@ -94,7 +95,13 @@ function computeTokenLength(content, features) {
         // Skip over any escaped characters
         totalTokenLength++;
         i++;
-      } else if (char === '/') {
+      } else if (char === '[') {
+        // Register that we're entering a character class so we don't leave the regex prematurely
+        isInRegexCharacterClass = true;
+      } else if (char === ']') {
+        // Register that we're exiting the character class
+        isInRegexCharacterClass = false;
+      } else if (char === '/' && !isInRegexCharacterClass) {
         // End the string when we hit the regex close character
         isInRegex = false;
         // console.log(i, 'leaving regex', char)
--- a/lighthouse-core/test/lib/minification-estimator-test.js
+++ b/lighthouse-core/test/lib/minification-estimator-test.js
@@ -180,6 +180,23 @@ describe('minification estimator', () => {
       assert.equal(computeJSTokenLength(js), 9);
     });
 
+    it('should handle regular expression character classes', () => {
+      const js = `
+        /regex [^/]\\//.test('this should be in string not comment 123456789')
+      `;
+
+      assert.equal(computeJSTokenLength(js), 69);
+      assert.equal(computeJSTokenLength(js), js.trim().length);
+    });
+
+    it('should handle escaped regular expression characters', () => {
+      const js = `
+        /regex \\[/ // this should be in comment not string 123456789
+      `;
+
+      assert.equal(computeJSTokenLength(js), 10);
+    });
+
     it('should distinguish regex from divide', () => {
       const js = `
         return 1 / 2 // hello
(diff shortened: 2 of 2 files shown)


lighthouse-core/lib/minification-estimator.js
```diff
@@ -94,7 +95,13 @@ function computeTokenLength(content, features) {
         // Skip over any escaped characters
         totalTokenLength++;
         i++;
-      } else if (char === '/') {
+      } else if (char === '[') {
+        // Register that we're entering a character class so we don't leave the regex prematurely
+        isInRegexCharacterClass = true;
+      } else if (char === ']') {
```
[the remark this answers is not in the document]
patrickhulce: shouldn't affect the execution but it's definitely a lot more clear :) 👍 done

Hoten:
Two tangential things:

1) I noticed strings don't get parsed correctly so I made #6746
2) Chrome autocorrect thinks minification should be "magnification"! Almost got me in the making of the above ticket, and I got Patrick in the original post :P

image

patrickhulce:
> I noticed strings don't get parsed correctly so I made #6746

sigh 😞, I don't think we'll ever nail all the edge cases. I'm thinking we might want to do a second gut check of, "Hey if it thinks there's a crazy amount of savings when there's like no whitespace in this file, we should just ignore it's result"

> Chrome autocorrect thinks minification should be "magnification"! Almost got me in the making of the above ticket, and I got Patrick in the original post :P

😆 hahaha it totally did

lighthouse-core/test/lib/minification-estimator-test.js
```diff
@@ -180,6 +180,23 @@ describe('minification estimator', () => {
       assert.equal(computeJSTokenLength(js), 9);
     });
 
+    it('should handle regular expression character classes', () => {
+      const js = `
+        /regex [^/]\\//.test('this should be in string not comment 123456789')
```
exterkamp: Is this valid?
```javascript
// Uncaught SyntaxError: Unexpected token .
/regex [^/]\\//.test('this should be in string not comment 123456789')
```
Should it be:
```javascript
/regex [^/]\\/.test('this should be in string not comment 123456789')
```
Or if you're trying to include a double / then:
```javascript
/regex [^/]\\\//.test('this should be in string not comment 123456789')
```

lighthouse-core/test/lib/minification-estimator-test.js
```diff
@@ -180,6 +180,23 @@ describe('minification estimator', () => {
...
+      `;
+
+      assert.equal(computeJSTokenLength(js), 69);
+      assert.equal(computeJSTokenLength(js), js.trim().length);
+    });
+
+    it('should handle escaped regular expression characters', () => {
+      const js = `
+        /regex \\[/ // this should be in comment not string 123456789
```
exterkamp: I'm also confused here?  What is this testing?  The comment is still part of the regex
```javascript
/regex \\[/ // this should be in comment not string 123456789
```
If it was supposed to be a string then it needs to escape the `[`
```javascript
/regex \\\[/ // this should be in comment not string 123456789
```

lighthouse-core/test/lib/minification-estimator-test.js
patrickhulce: Yeah I'm open to ideas to make this clearer. Basically it's the JS equivalent of

```js
/regex [^/]\//.test('this should be in string not comment 123456789')
```

Because we're sticking the `\` inside a string we need an extra `\` to escape it, but we don't need to escape anything else

lighthouse-core/test/lib/minification-estimator-test.js
patrickhulce: same deal here it's the js equivalent of
```js
/regex \[/ // this should be in comment not string 123456789

lighthouse-core/test/lib/minification-estimator-test.js
patrickhulce: FWIW, I think it's much clearer in vscode which does the highlighting to show what's being escaped, so maybe just add my js example to the comments?

Issue #6200, opened 2018-10-09
Audit.Result.displayValue should always be a string


exterkamp:
#### Provide the steps to reproduce
`displayValue` is sometimes a string when it has been translated (like in the perf category) but it is sometimes still in its untranslated form of an array of elements. e.g.
```json
displayValue: [
    "%dms of time",
    100
]
```

PR #6767, opened 2018-12-10
core(displayvalue): change displayValue to be string only

exterkamp:
**Summary**
Convert displayValue to be a `string` value only instead of `[string, <string | number> ]` for consistency.  Also converted the only 2 audits that still use the old displayValue array format.
- font-size
- load-fast-enough-for-pwa

**Related Issues/PRs**
fixes: #6200

### DIFF
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -15,13 +15,29 @@ const isDeepEqual = require('lodash.isequal');
 const Audit = require('./audit');
 const mobileThrottling = require('../config/constants').throttling.mobileSlow4G;
 const Interactive = require('../computed/metrics/interactive.js');
-
-const displayValueText = `Interactive at %d\xa0s`;
-const displayValueTextWithOverride = `Interactive on simulated mobile network at %d\xa0s`;
+const i18n = require('../lib/i18n/i18n.js');
 
 // Maximum TTI to be considered "fast" for PWA baseline checklist
 //   https://developers.google.com/web/progressive-web-apps/checklist
 const MAXIMUM_TTI = 10 * 1000;
+
+const UIStrings = {
+  /** */
+  title: 'Page load is fast enough on mobile networks',
+  /** */
+  failureTitle: 'Page load is not fast enough on mobile networks',
+  /** */
+  description:
+    'A fast page load over a cellular network ensures a good mobile user experience. ' +
+    '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/fast-3g).',
+  /** */
+  displayValueText: 'Interactive at {timing, number, seconds}\xa0s',
+  /** */
+  displayValueTextWithOverride: 'Interactive on simulated mobile network at ' +
+  '{timing, number, seconds}\xa0s',
+};
+
+const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
 
 class LoadFastEnough4Pwa extends Audit {
   /**
@@ -30,11 +46,9 @@ class LoadFastEnough4Pwa extends Audit {
   static get meta() {
     return {
       id: 'load-fast-enough-for-pwa',
-      title: 'Page load is fast enough on mobile networks',
-      failureTitle: 'Page load is not fast enough on mobile networks',
-      description:
-        'A fast page load over a cellular network ensures a good mobile user experience. ' +
-        '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/fast-3g).',
+      title: str_(UIStrings.title),
+      failureTitle: str_(UIStrings.failureTitle),
+      description: str_(UIStrings.description),
       requiredArtifacts: ['traces', 'devtoolsLogs'],
     };
   }
@@ -57,7 +71,8 @@ class LoadFastEnough4Pwa extends Audit {
     const override = context.settings.throttlingMethod === 'provided' ||
       !isDeepEqual(context.settings.throttling, mobileThrottling);
 
-    const displayValueFinal = override ? displayValueTextWithOverride : displayValueText;
+    const displayValueFinal = override ?
+      UIStrings.displayValueTextWithOverride : UIStrings.displayValueText;
 
     const settings = override ? Object.assign({}, context.settings, settingOverrides) :
       context.settings;
@@ -72,7 +87,7 @@ class LoadFastEnough4Pwa extends Audit {
     /** @type {string|undefined} */
     let explanation;
     if (!score) {
-      displayValue = [displayValueFinal, tti.timing / 1000];
+      displayValue = str_(displayValueFinal, {timing: tti.timing / 1000});
       explanation = 'Your page loads too slowly and is not interactive within 10 seconds. ' +
         'Look at the opportunities and diagnostics in the "Performance" section to learn how to ' +
         'improve.';
@@ -88,3 +103,4 @@ class LoadFastEnough4Pwa extends Audit {
 }
 
 module.exports = LoadFastEnough4Pwa;
+module.exports.UIStrings = UIStrings;
--- a/lighthouse-core/audits/load-fast-enough-for-pwa.js
+++ b/lighthouse-core/audits/load-fast-enough-for-pwa.js
@@ -27,14 +27,12 @@ const UIStrings = {
   /** Imperative title of a Lighthouse audit that tells the user that their page has loaded fast enough to be considered a Progressive Web App. This imperative title is shown to users when the web page has loaded too slowly to be considered a Progressive Web App. */
   failureTitle: 'Page load is not fast enough on mobile networks',
   /** Description of a Lighthouse audit that tells the user *why* they need to load fast enough on mobile networks. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
-  description:
-    'A fast page load over a cellular network ensures a good mobile user experience. ' +
-    '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/fast-3g).',
+  description: 'A fast page load over a cellular network ensures a good mobile user experience. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/fast-3g).',
   /** [ICU Syntax] Label for the audit identifying the time it took for the page to become interactive. */
-  displayValueText: 'Interactive at {timing, number, seconds}\xa0s',
+  displayValueText: 'Interactive at {timeInMs, number, seconds}\xa0s',
   /** [ICU Syntax] Label for the audit identifying the time it took for the page to become interactive on a mobile network. */
   displayValueTextWithOverride: 'Interactive on simulated mobile network at ' +
-  '{timing, number, seconds}\xa0s',
+  '{timeInMs, number, seconds}\xa0s',
 };
 
 const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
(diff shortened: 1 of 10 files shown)


lighthouse-core/audits/seo/font-size.js
```diff
@@ -8,10 +8,26 @@
...
+  title: 'Document uses legible font sizes',
+  /** Imperative title of a Lighthouse audit that tells the user that they should use font sizes that are easily read by the user. This imperative title is shown to users when there is a font that is too small to be read by the user. */
+  failureTitle: 'Document doesn\'t use legible font sizes',
+  /** Description of a Lighthouse audit that tells the user *why* they need to use a larger font size. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
+  description: 'Font sizes less than 12px are too small to be legible and require mobile ' +
+      'visitors to “pinch to zoom” in order to read. Strive to have >60% of page text ≥12px. ' +
+      '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/font-sizes).',
+  /** [ICU Syntax] Label for the audit identifying font sizes that are too small. */
+  displayValue: '{percentFormatted, number, percent4SigFig} legible text',
```
patrickhulce: is there a reason we're upping it to 4 instead of leaving at 1 decimal point?

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -71,6 +71,11 @@ const formats = {
       minimumFractionDigits: 1,
       maximumFractionDigits: 1,
     },
+    percent4SigFig: {
+      // Force allow up to two digits after decimal place in percentages. (Intl.NumberFormat options)
+      maximumFractionDigits: 4,
```
patrickhulce: wouldn't this be 2 then? maximumSignificantDigits is its own thing

still would want to keep this at one though :)

lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
```diff
@@ -51,7 +51,7 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
     return FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()}).then(result => {
       assert.equal(result.score, false, 'not failing a long TTI value');
       assert.equal(result.rawValue, 15000);
-      assert.deepEqual(result.displayValue, ['Interactive at %d\xa0s', 15]);
+      expect(result.displayValue).toBeDisplayString('Interactive at 15.0\xa0s');
```
patrickhulce: let's add an assertion to `font-size-test` while we're at it so we'd catch the 4 digits thing

lighthouse-core/audits/seo/font-size.js
exterkamp: The 4 sigfig percent will make it form XX.XX%, so it matches to two trailing decimal places, I upped it from 1 place to match the failing legible text percent which is formatted to 2 decimal places
`percentageOfFailingText = parseFloat((100 - percentageOfPassingText).toFixed(2));`
It seemed like they should match.

lighthouse-core/lib/i18n/i18n.js
exterkamp: Since it takes the percent decimal in the form `0.XXXX` -> `XX.XX%` it has to be 4 decimal places to prune it to 2 after the decimal place.  I think maximumSignificantDigits would also effectively do the same thing?

patrickhulce: Oh I was very confused then

how does this get turned back into `67.X%`? `style: 'percent'` in ICU automatically multiplies by 100?

seems like we should either use `maximumSignificantDigits` and keep the name or change the name and keep the `maximumFractionDigits` with extra-emphasis that NumberFormat is bonkers and applies the maximumFractionDigits to the pre-multiplied string

I apologize for my confusion, I think I'd prefer sigFig because that's just nuts 😆

lighthouse-core/audits/seo/font-size.js
patrickhulce: oh, not it in diff but elsewhere in the file gotcha! my b 👍

To explain my multi-comment review just about the single decimal place since at this point it seems like it needs explaining not as random hate on your PR (😄): 

In general, I'm a fan of less precision in Lighthouse UI. Especially cases where we can be somewhat hand-wavy to begin with in the impl, or there's really no action to be taken based on the additional precision.

patrickhulce: sorrrrrry to keep going on this.....but can we actually rename the variable then too? `percentFormatted` makes me think we are passing it a `percentFormatted` string when that's what's going to come out on the other side, or am I misunderstanding the point of the ICU syntax names?

lighthouse-core/audits/seo/font-size.js
```diff
@@ -265,8 +279,9 @@ class FontSize extends Audit {
       });
     }
 
+    const percentFormatted = (percentageOfPassingText / 100).toFixed(4);
```
patrickhulce: wait hang on, triple sorry for this I seem to be all over the place with this PR 😭 

isn't this already going to be a string with 4 sigFigs before it ever gets to the i18n bit so all of what I've said is going to be moot anyhow?

exterkamp: So I make it guaranteed to be 4 digits only, but not necessarily 4 sigfigs e.g. 0.0318 -> 3.18% which is 2 decimal places after the . but is only 3 sig figs.  So I guess maybe I should rename `percent4sigfig` to `percent2trailing` because it will clip off any fraction with more than 2 decimal places.  Technically I don't think I need to format this to 4 digits here since Intl will do that with the 4 max digits anyway.

lighthouse-core/lib/i18n/i18n.js
exterkamp: Yes `style: 'percent'` will multiply by 100 and add a `%` to the number automatically and round off so there are no digits after the decimal place.  I can try to use sigfigs, the only worry i have is that setting 4 sigfigs might lead to `8.546%` which would be accurately 4 sigfigs, but not really what we want, I'd imagine we want `8.55%` to be consistent.

lighthouse-core/audits/seo/font-size.js
patrickhulce: OK how's about this.

The `(percentageOfPassingText / 100).toFixed(4)` is super clear to me. Will i18n lib avoid a custom formatting option entirely if we can just pass that directly as just `'percent'` styled?

exterkamp:
Okay I think the naming conventions are in a good spot.  `decimalProportion` for the ICU message var replacement name and `extendedPercent` for the two digit percent override.

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -72,7 +87,7 @@ class LoadFastEnough4Pwa extends Audit {
     /** @type {string|undefined} */
     let explanation;
     if (!score) {
-      displayValue = [displayValueFinal, tti.timing / 1000];
+      displayValue = str_(displayValueFinal, {timing: tti.timing / 1000});
```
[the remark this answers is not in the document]
patrickhulce: Yeah sg we can defer to the i18n lib since the goal was consistency there 👍

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -15,26 +15,38 @@ const isDeepEqual = require('lodash.isequal');
...
+  /** Imperative title of a Lighthouse audit that tells the user that their page has loaded fast enough to be considered a Progressive Web App. This imperative title is shown to users when the web page has loaded too slowly to be considered a Progressive Web App. */
+  failureTitle: 'Page load is not fast enough on mobile networks',
+  /** Description of a Lighthouse audit that tells the user *why* they need to load fast enough on mobile networks. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
+  description: 'A fast page load over a cellular network ensures a good mobile user experience. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/fast-3g).',
+  /** [ICU Syntax] Label for the audit identifying the time it took for the page to become interactive. */
+  displayValueText: 'Interactive at {timeInMs, number, seconds}\xa0s',
+  /** [ICU Syntax] Label for the audit identifying the time it took for the page to become interactive on a mobile network. */
+  displayValueTextWithOverride: 'Interactive on simulated mobile network at ' +
```
exterkamp: Should this be 1 line @brendankenny?

brendankenny: > Should this be 1 line

looks like it won't fit

lighthouse-core/audits/load-fast-enough-for-pwa.js
```diff
@@ -57,7 +69,8 @@ class LoadFastEnough4Pwa extends Audit {
     const override = context.settings.throttlingMethod === 'provided' ||
       !isDeepEqual(context.settings.throttling, mobileThrottling);
 
-    const displayValueFinal = override ? displayValueTextWithOverride : displayValueText;
+    const displayValueFinal = override ?
```
patrickhulce: nit: a bit weird that `displayValueFinal` is not the final display value :)

wdyt about `displayValueTemplate`?

lighthouse-core/audits/seo/font-size.js
```diff
@@ -265,8 +277,9 @@ class FontSize extends Audit {
       });
     }
 
+    const decimalProportion = (percentageOfPassingText / 100).toFixed(4);
```
patrickhulce: still a bit weird that we are passing in a string to the i18n, I would expect this to be a number, is it still necessary to call `.toFixed` now that we've sorted out the decimal piece in our i18n lib?

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -71,6 +71,11 @@ const formats = {
       minimumFractionDigits: 1,
       maximumFractionDigits: 1,
     },
+    extendedPercent: {
+      // Force allow up to two digits after decimal place in percentages. (Intl.NumberFormat options)
+      maximumFractionDigits: 2,
```
patrickhulce: this makes way more sense :)

lighthouse-core/test/lib/i18n/i18n-test.js
```diff
@@ -76,4 +76,47 @@ describe('i18n', () => {
       expect(i18n.lookupLocale('jk-Latn-DE-1996-a-ext-x-phonebk-i-klingon')).toEqual('en');
     });
   });
+
+  describe('#_formatIcuMessage', () => {
```
patrickhulce: nice! great tests!

lighthouse-core/audits/seo/font-size.js
brendankenny: > I would expect this to be a number, is it still necessary to call `.toFixed` now that we've sorted out the decimal piece in our i18n lib?

whoops, missed this, good call!

lighthouse-core/test/lib/i18n/i18n-test.js
brendankenny: oh yeah, maybe rename the `describe`. `describe('Message argument formatting', () => {` or something?

lighthouse-core/audits/seo/font-size.js
exterkamp: Yup!  Good catch!

PR #6774, opened 2018-12-11
core(lantern): drop node id from error message

paulirish:
In LR we collect and report all our errors. This one causes our metrics to get a little funky since it changes every time.
Does this work for you, @patrickhulce.

Also i imagine you want a few URLs that can reproduce this?

b/120845363

### DIFF
--- a/lighthouse-core/lib/dependency-graph/base-node.js
+++ b/lighthouse-core/lib/dependency-graph/base-node.js
@@ -209,7 +209,7 @@ class BaseNode {
       }
     });
 
-    if (!idToNodeMap.has(this.id)) throw new Error(`Cloned graph missing node ${this.id}`);
+    if (!idToNodeMap.has(this.id)) throw new Error(`Cloned graph missing node`);
     return idToNodeMap.get(this.id);
   }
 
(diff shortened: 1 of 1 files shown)


patrickhulce:
> is the id ever helpful when debugging? If not, seems fine to drop it.

It isn't as an isolated aggregate error message. 

It's really useful when you're logging other parts of the graph too like I was when working on lantern and can look at the nodes to see which it is, but at that point, could just add back in a log statement. Might be nice to have some extra use of `debug` where useful log statements can be added that aren't run by default and can be turned back on when re-entering a specific domain. I fear any lantern stuff at this point would be too chatty for global verbose.

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -209,7 +209,7 @@ class BaseNode {
       }
     });
 
-    if (!idToNodeMap.has(this.id)) throw new Error(`Cloned graph missing node ${this.id}`);
+    if (!idToNodeMap.has(this.id)) throw new Error(`Cloned graph missing node`);
```
Hoten: smallest nit in the world of nits....`BACKTICK` -> `COMMA TO THE TOP` since there's no more interpolation

Hoten: Oops. I figured that edit would show up as a suggestion. I just changed it for you :)

PR #6780, opened 2018-12-12
core: handle invalid network timing data

### LINKED ISSUE #6378: Invalid dependency graph created, cycle detected
Got a repro on this bug, finally.  It's running http://www.matsucrossfit.com/ on LR, though I don't see the bug locally.

Here's the dtlog and trace: [lroutput.zip](https://github.com/GoogleChrome/lighthouse/files/2509387/lroutput.zip)

 Also i get "No timing information available" in this run as well, though I don't know if that's related.

@patrickhulce could you take a look?

patrickhulce:
**Summary**
In LR, the `timing` object can be defined but have no valid data (`requestTime: 0, receiveHeadersEnd: -1`). This was unexpected and was causing problems later on down the chain for lantern. 

Masking the problem was the fact that the `rootNode` of the lantern graph was trying to pick the earliest network request but was breaking ties by choosing the *last* network record in the array. When the requestTimes were `0`, it just ended up picking the last one and losing the real root node. This PR also adds some sanity checks around that to make sure we'd fail earlier about those unexpected cases.

**Related Issues/PRs**
fixes #6378

### DIFF
--- a/lighthouse-core/computed/page-dependency-graph.js
+++ b/lighthouse-core/computed/page-dependency-graph.js
@@ -269,14 +269,23 @@ class PageDependencyGraph {
     const networkNodeOutput = PageDependencyGraph.getNetworkNodeOutput(networkRecords);
     const cpuNodes = PageDependencyGraph.getCPUNodes(traceOfTab);
 
-    const rootRequest = networkRecords.reduce((min, r) => (min.startTime < r.startTime ? min : r));
+    // The root request is the earliest network request, using position in networkRecords array to break ties.
+    const rootRequest = networkRecords.reduce((min, r) => (r.startTime < min.startTime ? r : min));
     const rootNode = networkNodeOutput.idToNodeMap.get(rootRequest.requestId);
+    // The main document request is the earliest network request *of type document*.
+    // This will be different from the root request when there are server redirects.
     const mainDocumentRequest = NetworkAnalyzer.findMainDocument(networkRecords);
     const mainDocumentNode = networkNodeOutput.idToNodeMap.get(mainDocumentRequest.requestId);
 
     if (!rootNode || !mainDocumentNode) {
       // Should always be found.
       throw new Error(`${rootNode ? 'mainDocument' : 'root'}Node not found.`);
+    }
+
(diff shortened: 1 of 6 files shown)


lighthouse-core/lib/network-request.js
```diff
@@ -272,6 +272,9 @@ module.exports = class NetworkRequest {
    * @param {LH.Crdp.Network.ResourceTiming} timing
    */
   _recomputeTimesWithResourceTiming(timing) {
+    // Don't recompute times if the data is invalid. RequestTime should always be a thread timestamp.
+    // If we don't have receiveHeadersEnd, we really don't have more accurate data.
+    if (timing.requestTime === 0 || timing.receiveHeadersEnd === -1) return;
```
brendankenny: ha, I believe this would have saved me a lot of time figuring out how to generate the right network request events in #6171

lighthouse-core/computed/page-dependency-graph.js
```diff
@@ -269,8 +269,11 @@ class PageDependencyGraph {
     const networkNodeOutput = PageDependencyGraph.getNetworkNodeOutput(networkRecords);
     const cpuNodes = PageDependencyGraph.getCPUNodes(traceOfTab);
 
-    const rootRequest = networkRecords.reduce((min, r) => (min.startTime < r.startTime ? min : r));
+    // The root request is the earliest network request, using position in networkRecords array to break ties.
```
brendankenny: best comment

lighthouse-core/computed/page-dependency-graph.js
```diff
@@ -269,8 +269,11 @@ class PageDependencyGraph {
...
+    const rootRequest = networkRecords.reduce((min, r) => (r.startTime < min.startTime ? r : min));
```
brendankenny: might be overkill, but should we pull this out into another method on `NetworkAnalyzer` like `findMainDocument`? I can see it usable by others (if they remember it's there).

patrickhulce: I'd want to shift consumers to be using the graph as the source of truth, so *hopefully* no one else needs to compute this too? :)

Issue #6201, opened 2018-10-09
scoreDisplayMode incompatable with proto enum: not-applicable


exterkamp:
#### What is the current behavior?
An audit's `scoreDisplayMode` can be `not-applicable` which is fine in json, but invalid for a proto enum name because of the hyphen.

exterkamp:
@brendankenny brought up a good point to sidestep this issue.  We can rename `not-applicable` to something that can be consistent in both proto and LHR e.g.
- `notapplicable`
- `notApplicable`
- `auditplicable`

something like that.  Thoughts?

patrickhulce:
> auditplicable
😆 we obviously gotta go with this one now, right? 

I like `notApplicable`, personally.

paulirish:
notApplicable wfm

PR #6783, opened 2018-12-12
core: change 'not-applicable' to 'notApplicable'

exterkamp:
**Summary**
Change not-applicable to notApplicable.  This will allow the LHR and protoLHR to use to same value instead of converting it in preprocessing.

Unfortunately since we can't ignore old LHRs that use 'not-applicable' or 'not_applicable' the processors, proto, and renderer need to accommodate those values.

**Related Issues/PRs**
fixes: #6201

### DIFF
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -45,9 +45,12 @@ function processForProto(result) {
 
       // Rewrite the 'not-applicable' scoreDisplayMode to 'not_applicable'. #6201
       if (audit.scoreDisplayMode) {
+        // @ts-ignore ts properly flags this as invalid as it should not happen,
+        // but remains in preprocessor to protect from proto translation errors from
+        // old LHRs.
         if (audit.scoreDisplayMode === 'not-applicable') {
           // @ts-ignore Breaking the LH.Result type
-          audit.scoreDisplayMode = 'not_applicable';
+          audit.scoreDisplayMode = 'notApplicable';
         }
       }
       // Drop raw values. #6199
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -48,7 +48,8 @@ function processForProto(result) {
         // @ts-ignore ts properly flags this as invalid as it should not happen,
         // but remains in preprocessor to protect from proto translation errors from
         // old LHRs.
-        if (audit.scoreDisplayMode === 'not-applicable') {
+        // eslint-disable-next-line max-len
+        if (audit.scoreDisplayMode === 'not-applicable' || audit.scoreDisplayMode === 'not_applicable') {
           // @ts-ignore Breaking the LH.Result type
           audit.scoreDisplayMode = 'notApplicable';
         }
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -43,14 +43,13 @@ function processForProto(result) {
     Object.keys(reportJson.audits).forEach(auditName => {
       const audit = reportJson.audits[auditName];
 
-      // Rewrite the 'not-applicable' scoreDisplayMode to 'not_applicable'. #6201
+      // Rewrite 'not-applicable' and 'not_applicable' scoreDisplayMode to 'notApplicable'. #6201, #6783.
       if (audit.scoreDisplayMode) {
         // @ts-ignore ts properly flags this as invalid as it should not happen,
         // but remains in preprocessor to protect from proto translation errors from
         // old LHRs.
         // eslint-disable-next-line max-len
         if (audit.scoreDisplayMode === 'not-applicable' || audit.scoreDisplayMode === 'not_applicable') {
-          // @ts-ignore Breaking the LH.Result type
           audit.scoreDisplayMode = 'notApplicable';
         }
       }
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -45,18 +45,18 @@ class CategoryRenderer {
    */
   get _clumpDisplayInfo() {
     return {
-      'failed': {
+      failed: {
         className: 'lh-clump--failed',
       },
-      'manual': {
+      manual: {
         title: Util.UIStrings.manualAuditsGroupTitle,
         className: 'lh-clump--manual',
       },
-      'passed': {
+      passed: {
         title: Util.UIStrings.passedAuditsGroupTitle,
         className: 'lh-clump--passed',
       },
-      'notApplicable': {
+      notApplicable: {
         title: Util.UIStrings.notApplicableAuditsGroupTitle,
         className: 'lh-clump--not-applicable',
       },
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -153,7 +153,10 @@ class CategoryRenderer {
    */
   _setRatingClass(element, score, scoreDisplayMode) {
     const rating = Util.calculateRating(score, scoreDisplayMode);
-    element.classList.add(`lh-audit--${rating}`, `lh-audit--${scoreDisplayMode.toLowerCase()}`);
+    element.classList.add(`lh-audit--${rating}`);
+    if (typeof scoreDisplayMode === 'string') {
+      element.classList.add(`lh-audit--${scoreDisplayMode.toLowerCase()}`);
+    }
     return element;
   }
 
--- a/lighthouse-core/report/html/renderer/category-renderer.js
+++ b/lighthouse-core/report/html/renderer/category-renderer.js
@@ -22,7 +22,7 @@
 /** @typedef {import('./report-renderer.js')} ReportRenderer */
 /** @typedef {import('./details-renderer.js')} DetailsRenderer */
 /** @typedef {import('./util.js')} Util */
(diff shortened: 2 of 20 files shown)


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -22,7 +22,7 @@
 /** @typedef {import('./report-renderer.js')} ReportRenderer */
 /** @typedef {import('./details-renderer.js')} DetailsRenderer */
 /** @typedef {import('./util.js')} Util */
-/** @typedef {'failed'|'manual'|'passed'|'not-applicable'} TopLevelClumpId */
+/** @typedef {'failed'|'manual'|'passed'|'notApplicable'} TopLevelClumpId */
```
exterkamp: This doesn't have to change, but since it was a pass-through for the `scoreDisplayMode` values I thought it would make sense to update it.

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -45,9 +45,13 @@ function processForProto(result) {
 
       // Rewrite the 'not-applicable' scoreDisplayMode to 'not_applicable'. #6201
       if (audit.scoreDisplayMode) {
-        if (audit.scoreDisplayMode === 'not-applicable') {
+        // @ts-ignore ts properly flags this as invalid as it should not happen,
+        // but remains in preprocessor to protect from proto translation errors from
+        // old LHRs.
+        // eslint-disable-next-line max-len
```
brendankenny: just break the line :P

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -45,9 +45,13 @@ function processForProto(result) {
...
+        if (audit.scoreDisplayMode === 'not-applicable' || audit.scoreDisplayMode === 'not_applicable') {
           // @ts-ignore Breaking the LH.Result type
```
brendankenny: this `@ts-ignore` is no longer necessary

lighthouse-core/report/html/renderer/category-renderer.js
brendankenny: > This doesn't have to change

I think it does? Because the audits are clumped by their `scoreDisplayMode` (in `_getClumpIdForAuditRef`), so there wouldn't be a matching slot for the `notApplicable` ones if this wasn't updated

lighthouse-core/report/html/renderer/util.js
```diff
@@ -65,9 +65,11 @@ class Util {
     // The proto process turns 'not-applicable' into 'not_applicable'. Correct this to support both.
     // TODO: remove when underscore/hyphen proto issue is resolved. See #6371, #6201.
     for (const audit of Object.values(clone.audits)) {
+      // TODO(exterkamp): remove this when proto enum removes 'not_applicable'
```
brendankenny: remove just the `not_applicable` one? Can we add this to the proto compat umbrella bug instead of as a TODO?

lighthouse-core/report/html/renderer/util.js
```diff
@@ -65,9 +65,11 @@ class Util {
...
       // @ts-ignore tsc rightly flags that this value shouldn't occur.
-      if (audit.scoreDisplayMode === 'not_applicable') {
-        audit.scoreDisplayMode = 'not-applicable';
+      // eslint-disable-next-line max-len
```
brendankenny: line break instead, pls

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -45,9 +45,13 @@ function processForProto(result) {
 
       // Rewrite the 'not-applicable' scoreDisplayMode to 'not_applicable'. #6201
```
brendankenny: update comment

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -56,7 +56,7 @@ class CategoryRenderer {
         title: Util.UIStrings.passedAuditsGroupTitle,
         className: 'lh-clump--passed',
       },
-      'not-applicable': {
+      'notApplicable': {
         title: Util.UIStrings.notApplicableAuditsGroupTitle,
         className: 'lh-clump--not-applicable',
```
brendankenny: need to update

lighthouse-core/report/html/renderer/util.js
```diff
@@ -65,9 +65,11 @@ class Util {
     // The proto process turns 'not-applicable' into 'not_applicable'. Correct this to support both.
```
brendankenny: should update this comment both for the new value and to mention supporting old reports

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -161,7 +161,7 @@ describe('CategoryRenderer', () => {
         '.lh-clump--not-applicable .lh-audit-group__summary'));
 
     const notApplicableCount = a11yCategory.auditRefs.reduce((sum, audit) =>
-        sum += audit.result.scoreDisplayMode === 'not-applicable' ? 1 : 0, 0);
+        sum += audit.result.scoreDisplayMode === 'notApplicable' ? 1 : 0, 0);
```
paulirish: we need the renderer to handle both. so these will have to check both casings. 
Maybe make a util method?

patrickhulce: aren't we fixing all of them before they make it this far though?

lighthouse-core/report/html/renderer/category-renderer.js
patrickhulce: oh we should get a puppeteer screenshot test in for the report to catch these css mismatches

lighthouse-core/test/report/html/renderer/category-renderer-test.js
brendankenny: > aren't we fixing all of them before they make it this far though?

yeah, @paulirish it gets updated in `Util.prepareReportResult` so it's always `'notApplicable'` before it gets passed in to the renderer.

lighthouse-core/report/html/renderer/category-renderer.js
brendankenny: > oh we should get a puppeteer screenshot test in for the report to catch these css mismatches

good idea. Don't we also have a jsdom test or two looking for the not applicable results? Concerning they weren't found...

brendankenny: > Concerning they weren't found...

oh, because the test selectors weren't updated either :)

But now I'm questioning if it should be updated :) We don't currently create the clump names dynamically, so it can be whatever we want. Commonly `notApplicable` in js goes to `not-applicable` in css, but do we care? Or is `notapplicable` fine? Keeping camelCase seems gross, though.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
paulirish: ah sry. yes good call.

lighthouse-core/report/html/renderer/category-renderer.js
exterkamp: I guess I knew that subconsciously and left it :wink:

lighthouse-core/report/html/renderer/util.js
exterkamp: Made a tracking bug for proto convergence #6801

lighthouse-core/report/html/renderer/category-renderer.js
wardpeet: I would go for not-applicable makes it easier to read in my opinion

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -153,7 +153,10 @@ class CategoryRenderer {
    */
   _setRatingClass(element, score, scoreDisplayMode) {
     const rating = Util.calculateRating(score, scoreDisplayMode);
-    element.classList.add(`lh-audit--${rating}`, `lh-audit--${scoreDisplayMode}`);
+    element.classList.add(`lh-audit--${rating}`);
+    if (typeof scoreDisplayMode === 'string') {
```
brendankenny: wait...how would it be possible for it not to be? That seems bad if it happens. Just in tests maybe (hopefully)?

brendankenny: from https://travis-ci.org/GoogleChrome/lighthouse/builds/470305380#L816, it looks like just the unit tests. We should fix those (is that what fdf9755 did?) and then not change the code here, I think. Better to trust our type system and fail fast if we defined/enforced things poorly.

exterkamp: Yeah fdf9755ad113997bdfdb0d476a0827066b921427 fixed the unit tests and added guard rails for if the scoreDisplayMode was undefined.  Good point tho, I removed the guardrails and made all the tests actually represent reality!

lighthouse-core/test/lib/proto-preprocessor-test.js
```diff
@@ -78,7 +78,7 @@ describe('processing for proto', () => {
     const input = {
       'audits': {
         'critical-request-chains': {
-          'scoreDisplayMode': 'not-applicable',
+          'scoreDisplayMode': 'notApplicable',
```
brendankenny: ha, sorry, this one we want

exterkamp: >delete not-applicable 
>don't delete not-applicable

I just can't impress you brendan :sob:

brendankenny:
💥

Issue #6804, opened 2018-12-14
Audit errorMessage not i18n'd


exterkamp:
Per Audit `errorMessage` are not i18n'd.

Who thought they could catch all i18n errors in 1 PR #6457 :cry: 

```
      // Errors become error audit result.
      const errorMessage = err.friendlyMessage ?
        `${err.friendlyMessage} (${err.message})` :
        `Audit error: ${err.message}`;
      auditResult = Audit.generateErrorAuditResult(audit, errorMessage);
    }
```

PR #6812, opened 2018-12-14
i18n: localize audit error strings

exterkamp:
**Summary**
Audit errors were adding the message code to the end of the errorMessage with parens which was destroying the ICU replacement.  This adds `defaultProperties` to easier facilitate mutliple LHErrors using the same UIString to replace and use their own defaults that are constant per error.

**Related Issues/PRs**
fixes: #6804

### DIFF
--- a/lighthouse-core/lib/lh-error.js
+++ b/lighthouse-core/lib/lh-error.js
@@ -12,7 +12,7 @@ const UIStrings = {
   /** Error message explaining that the Lighthouse run was not able to collect screenshots through Chrome.*/
   didntCollectScreenshots: `Chrome didn't collect any screenshots during the page load. Please make sure there is content visible on the page, and then try re-running Lighthouse.`,
   /** Error message explaining that the network trace was not able to be recorded for the Lighthouse run. */
-  badTraceRecording: 'Something went wrong with recording the trace over your page load. Please run Lighthouse again.',
+  badTraceRecording: 'Something went wrong with recording the trace over your page load. Please run Lighthouse again. ({traceCode})',
   /** Error message explaining that the page loaded too slowly to perform a Lighthouse run.  */
   pageLoadTookTooLong: 'Your page took too long to load. Please follow the opportunities in the report to reduce your page load time, and then try re-running Lighthouse.',
   /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
@@ -46,6 +46,7 @@ const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
  * @property {string} message
  * @property {RegExp} [pattern]
  * @property {boolean} [lhrRuntimeError] True if it should appear in the top-level LHR.runtimeError property.
+ * @property {object} [defaultProperties] Optional default properties to use for i18n.
  */
 
 class LighthouseError extends Error {
@@ -57,21 +58,12 @@ class LighthouseError extends Error {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
-    this.friendlyMessage = str_(errorDefinition.message, properties);
+    const mergedProperties = {...errorDefinition.defaultProperties, ...properties};
+    this.friendlyMessage = str_(errorDefinition.message, mergedProperties);
     this.lhrRuntimeError = !!errorDefinition.lhrRuntimeError;
     if (properties) Object.assign(this, properties);
 
     Error.captureStackTrace(this, LighthouseError);
-  }
-
-  /**
-   * @param {LighthouseError} err
-   * @return {LighthouseError}
-   */
-  static fromLighthouseError(err) {
-    const {code, friendlyMessage: message, ...rest} = err;
-    // Note: {...rest} convinces tsc 3.1 that it's assignable to a Record.
-    return new LighthouseError({code, message}, {...rest});
   }
 
   /**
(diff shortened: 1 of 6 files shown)


lighthouse-core/lib/lh-error.js
```diff
@@ -57,23 +58,14 @@ class LighthouseError extends Error {
...
     if (properties) Object.assign(this, properties);
 
     Error.captureStackTrace(this, LighthouseError);
   }
 
-  /**
-   * @param {LighthouseError} err
-   * @return {LighthouseError}
-   */
-  static fromLighthouseError(err) {
```
patrickhulce: was this just never used before? I'm not sure why it would be needed so seems good haha :)

lighthouse-core/runner.js
```diff
@@ -321,9 +321,7 @@ class Runner {
 
       Sentry.captureException(err, {tags: {audit: audit.meta.id}, level: 'error'});
       // Errors become error audit result.
-      const errorMessage = err.friendlyMessage ?
-        `${err.friendlyMessage} (${err.message})` :
-        `Audit error: ${err.message}`;
+      const errorMessage = err.friendlyMessage ? err.friendlyMessage : err.message;
```
patrickhulce: so what errors do we start losing for the code for after this, just the speedline ones? or the page load ones too?

I agree there's not a strong need for our users to know the code, but having it somewhere in the LHR would be nice.

lighthouse-core/lib/lh-error.js
```diff
@@ -57,23 +58,14 @@ class LighthouseError extends Error {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
-    this.friendlyMessage = str_(errorDefinition.message, properties);
+    const mergedProperties = {...errorDefinition.defaultProperties, ...properties};
```
patrickhulce: I'm not a huge fan of `defaultProperties` basically copying the `.code` over, but maybe there's a future world in which the generic is helpful. Did you have any particular use case in mind?

lighthouse-core/lib/lh-error.js
exterkamp: It was, but then the only caller was deleted and this was kept!

lighthouse-core/lib/lh-error.js
brendankenny: yeah, this was kind of my idea, trying to tame some complexity of just trying to get the error codes through in the localized string. I'm kind of leaning in the other direction now, though, and thinking we should just have a unique UIString for each of these so we don't need the error code appended on to them?

Or is there another way we can accomplish this?

lighthouse-core/lib/lh-error.js
```diff
@@ -127,25 +119,40 @@ const ERRORS = {
...
     lhrRuntimeError: true,
+    defaultProperties: {
+      traceCode: 'NO_NAVSTART',
+    },
   },
   NO_FCP: {
     code: 'NO_FCP',
     message: UIStrings.badTraceRecording,
     lhrRuntimeError: true,
+    defaultProperties: {
```
paulirish: icuArgumentValues

lighthouse-core/lib/lh-error.js
```diff
@@ -46,6 +46,7 @@ const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
  * @property {string} message
  * @property {RegExp} [pattern]
  * @property {boolean} [lhrRuntimeError] True if it should appear in the top-level LHR.runtimeError property.
+ * @property {object} [defaultProperties] Optional default properties to use for i18n.
```
paulirish: Argument values to use for i18n message formatting

lighthouse-core/lib/lh-error.js
paulirish: mergedArgValues i guess?

lighthouse-core/runner.js
exterkamp: I think we should follow the same pattern of splitting these out into unique values.

exterkamp:
Okay, flip flopped again.  I was about to split out all the speedline UIStrings, and then noticed all the pageLoadTookTooLong UIStrings which would also need to be split up afaik.  So this seemed like it was getting ridiculous, so @brendankenny and I came to the compromise of providing the `code` of the Error to the ICU replacement by default, this seems to make sense since the code is always enumerated, and is all the replacements needed anyway at this point.

Thoughts?

lighthouse-core/test/runner-test.js
```diff
@@ -606,7 +606,8 @@ describe('Runner', () => {
     assert.ok(lhr.audits['test-audit'].errorMessage.includes(NO_FCP.code));
     // And it bubbled up to the runtimeError.
     assert.strictEqual(lhr.runtimeError.code, NO_FCP.code);
-    assert.ok(lhr.runtimeError.message.includes(NO_FCP.message));
+    expect(lhr.runtimeError.message)
+      .toBeDisplayString(/Something .*(NO_FCP)/);
```
brendankenny: escape `(` and `)`?

lighthouse-core/lib/lh-error.js
```diff
@@ -57,23 +57,13 @@ class LighthouseError extends Error {
     super(errorDefinition.code);
     this.name = 'LHError';
     this.code = errorDefinition.code;
-    this.friendlyMessage = str_(errorDefinition.message, properties);
+    this.friendlyMessage = str_(errorDefinition.message, {errorCode: this.code, ...properties});
```
brendankenny: maybe add a comment mentioning all errors getting `errorCode` as an ICU message argument by default?

lighthouse-core/test/runner-test.js
exterkamp: Nice catch!

PR #6835, opened 2018-12-18
core(uses-long-cache-ttl): ignore private, must-validate, no-cache

### LINKED ISSUE #6140: Third-party Google Assets don't follow their own audit rules, causing report errors
When I run lighthouse reports, my only errors and warnings come from assets delivered by the creator of lighthouse: Google.

Specifically, this warning/error shows up.

## Uses inefficient cache policy on static assets
*All* the items in this report come from 
  * `securepubads.g.doubleclick.net` (google property).
  * `googletagmanager.com` (google property)
  * `pagead2.googlesyndication.com` (google property)
  * `www.google-analytics.com` (google property)
  * `tpc.googlesyndication.com` (google property)

Why does Google serve assets that fail to follow their own recommendations as based on lighthouse?  If they recommend static assets be sent with proper caching headers, why doesn't google follow through with that? Some of their assets have a ttl of *0* or *15m*!

If they aren't going to follow their own recommendations, then it seems that lighthouse should exclude their own domains from this analysis, since it creates a bunch of false positives which I can do nothing about.

Hoten:
Fixes #6140.


lighthouse-core/test/audits/byte-efficiency/uses-long-cache-ttl-test.js
```diff
@@ -103,11 +103,10 @@ describe('Cache headers audit', () => {
 
     const networkRecords = [
       networkRecord({headers: {
-        'cache-control': 'must-revalidate,max-age=3600',
```
patrickhulce: can we come up with a different directive to stick in here to make sure we parse properly :)

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -202,6 +202,12 @@ class CacheHeaders extends Audit {
           cacheControl
         );
 
+        // Ignore assets where policy implies they should not be cached long periods
+        if (cacheControl &&
+          (cacheControl['must-validate'] || cacheControl['no-cache'] || cacheControl['private'])) {
```
patrickhulce: Can we figure out how to consolidate some of this with the logic in `computeCacheLifetimeInSeconds` in L115? right now these type of checks are in there to return a lifetime of `0`

Hoten: maybe? these cache controls could still have ttls, so it'd make `computeCacheLifetimeInSeconds` a bit of a liar. I get that if we have this method return 0 for all cases we want to ignore, things work. which is cool, I'm just arguing semantics I suppose.

lighthouse-core/test/audits/byte-efficiency/uses-long-cache-ttl-test.js
Hoten: ya

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
patrickhulce: it doesn't have to be moving this there, we can extract the L115 part too (and probably should it's not the most obvious right now). I'm just advocating for collecting the special cases we handle so it's clear what's going on.

Hoten: Done. How's that?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -218,7 +238,9 @@ class CacheHeaders extends Audit {
 
         results.push({
           url,
-          cacheControl,
+          // Include cacheControl in results, but cast as any so table types
+          // are happy. cacheControl is not shown in the table so this is OK.
+          cacheControl: /** @type {any} */ (cacheControl),
```
Hoten: thoughts?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -169,6 +160,36 @@ class CacheHeaders extends Audit {
...
+        cacheControl['must-validate'] ||
+        cacheControl['no-cache'] ||
+        cacheControl['no-store'] ||
+        cacheControl['private'])) {
+      return false;
+    }
+
+    if (cacheLifetimeInSeconds !== null && cacheLifetimeInSeconds <= 0) return false;
+
+    if (cacheHitProbability > IGNORE_THRESHOLD_IN_PERCENT) return false;
```
patrickhulce: I'm wondering if we should just move the `cacheHitProbability` calculation to after this `shouldProcessRecord` and switch the ignore threshold to the equivalent time.

That might make it easier to figure out what's going on....

Hoten: > switch the ignore threshold to the equivalent time

I don't understand what you mean by this.

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
Hoten: summoning @brendankenny

lighthouse-core/test/audits/byte-efficiency/uses-long-cache-ttl-test.js
```diff
@@ -74,6 +74,22 @@ describe('Cache headers audit', () => {
     });
   });
 
+  it('ignores nonpositive and nonfinite max-age headers', () => {
+    const infinityMaxAgeStringValue = '1'.repeat(400);
+    assert.equal(Number.parseInt(infinityMaxAgeStringValue), Infinity);
+    const networkRecords = [
```
Hoten: so, we wish to ignore all of these records right?

FYI `max-age=blahblah` would not be ignored.

Shouldn't something silly like `max-age=-anything` be surfaced? or is that out of scope for this audit

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
patrickhulce: cacheLifetimeInSeconds gets converted into a cacheHitProbability

instead of ignoring records with a `cacheHitProbability` above 92.5% what if we ignored records with `cacheLifetimeInSeconds` above 2 weeks (should be the equivalent of 92.5%)

lighthouse-core/test/audits/byte-efficiency/uses-long-cache-ttl-test.js
patrickhulce: what did the spec say for cacheControl? for `expires` it was anything invalid is treated as no cache lifetime, so if cache-control is the same we should probably surface them as having no caching?

brendankenny:
Would really love a PR summary on things like this where parsing the conversation in the original issue takes way too long :P

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
brendankenny: > cacheControl is not shown in the table so this is OK.

It's not ok, why use a table if you don't want it to be a table, but it's not your fault that you're inheriting this :)

Will you add a `TODO(bckenny): fix DetailsItem` to the comment?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
patrickhulce: > I could be making an opposite suggestion to what @patrickhulce has suggested before, but to me there are two different sets of checks in here:

I didn't do a great job of clearly saying what I was lookin for before (sorry @Hoten !!) so it certainly could have been interpreted as opposite to that :)

My primary concern is not sprinkling around the different cases where things get ignored for the same category of reason, before this we checked `private` somewhere different from `no-cache` which are both ignored because of the signal in the `cache-control` directive.

I think it's relatively difficult to segment `shouldProcessRecord` and `computeCacheLifetime` as cleanly as you've done in that example @brendankenny  because there are some interdependencies, but if everyone's happy with some duplicated logic and think it's easy to read, I'm onboard.

IMO the fact that we're dealing with some mixture of `pragma`, `expires`, and `cache-control` all at the same time also complicates this, but I guess those two other cases are pretty simple.

Hoten: understood. I ran a bit too much with the idea of only using one `continue` ;)

Hoten:
anything more?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -169,6 +160,30 @@ class CacheHeaders extends Audit {
     );
   }
 
+  /**
+   * @param {Map<string, string>} headers
```
brendankenny: can you add a jsdoc description to this? Don't want to be redundant with the comments below, just a general category for what would disqualify a record

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -169,6 +160,30 @@ class CacheHeaders extends Audit {
...
+   * @param {ReturnType<typeof parseCacheControl>} cacheControl
+   * @returns {boolean}
+   */
+  static shouldProcessRecord(headers, cacheControl) {
```
brendankenny: should we rename `shouldSkipRecord` or something since it's only used negated?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -197,15 +212,20 @@ class CacheHeaders extends Audit {
...
-          headers,
-          cacheControl
-        );
+        if (!this.shouldProcessRecord(headers, cacheControl)) {
+          continue;
+        }
 
-        // Ignore assets with an explicit no-cache policy
-        if (cacheLifetimeInSeconds === 0) continue;
+        // Ignore if cacheLifetimeInSeconds is defined and a nonpositive number.
```
brendankenny: the `is defined` throws me for a loop when reading this. `Ignore if cacheLifetimeInSeconds is a nonpositive number` seems sufficient and is still a true description :)

Issue #6831, opened 2018-12-18
Bot: LH Issue Runner


Hoten:
Code is here: https://github.com/Hoten/lh-issue-runner

Example run here: https://github.com/GoogleChrome/lighthouse/issues/6830

Currently, the bot runs

1. if an issue has the label `needs-lh-runnner`
2. if a comment requests it, via `LH Runner Go!`

Open questions:
1. Could this be extended to post a comment when a URL is missing? What about issues that don't have a particular URL (like this one! but are there many?) - should there be a "don't bug me about a url" substring that is in the default issue template (kinda gross)?
2. Could this be done automatically, without requiring a label?
3. How could the DevTools entry point be tested? Is it possible to use puppeteer to run LH via the Audits panel?
4. ditto for the extension
5. any suggestions for the comment template?

Once these questions are resolved we can set it up on the same machines running DZL :)

brendankenny:
> 3. How could the DevTools entry point be tested? Is it possible to use puppeteer to run LH via the Audits panel?
> 4. ditto for the extension

- maybe for devtools?
- yes for extension. We do that in [`extension-test.js`](https://github.com/GoogleChrome/lighthouse/blob/51d204762ea6865e8de28ce4b54e36e058833a5d/clients/test/extension/extension-test.js)

Hoten:
Paul nearly got it working for devtools, but kept finding gremlins.

Found a way to pull down the latest LH extension and test that entrypoint.

PR #6839, opened 2018-12-18
extension: return runnerResult

Hoten:
No change in behavior.

I noticed that this function was typed to return a runner result, but it never does. It would be useful for #6831 if it actually did.

### DIFF
--- a/clients/extension/scripts/extension-entry.js
+++ b/clients/extension/scripts/extension-entry.js
@@ -104,7 +104,7 @@ async function runLighthouseInExtension(flags, categoryIDs) {
   const reportHtml = /** @type {string} */ (runnerResult.report);
   const blobURL = createReportPageAsBlob(reportHtml);
   await new Promise(resolve => chrome.windows.create({url: blobURL}, resolve));
-  
+
   return runnerResult;
 }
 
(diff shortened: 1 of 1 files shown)


Hoten:
Also, it may be masking errors in the extension test: https://github.com/GoogleChrome/lighthouse/blob/master/clients/test/extension/extension-test.js#L94

clients/extension/scripts/extension-entry.js
```diff
@@ -104,6 +104,8 @@ async function runLighthouseInExtension(flags, categoryIDs) {
   const reportHtml = /** @type {string} */ (runnerResult.report);
   const blobURL = createReportPageAsBlob(reportHtml);
   await new Promise(resolve => chrome.windows.create({url: blobURL}, resolve));
+
+  return runnerResult;
```
brendankenny: can you update the return type, too? Since the `undefined` branch throws an error now, `void` isn't necessary any more

brendankenny:
> in extension-test.js, let's assert something about lighthouseResult.. like it has fetchedAt on it?

yeah, we shouldn't have had a lack-of-something test by itself in JS :)

brendankenny:
discussed with @Hoten, I'll take the test fix cause I see a few other things it would be nice to clean up in there.

Issue #6837, opened 2018-12-18
Put PWA category last


paulirish:
To match score gauge.

PR #6848, opened 2018-12-19
core(config): reorder categories to match new gauge order

exterkamp:
**Summary**
Move PWA section to the end of the report to match its order on the top level score gauges.

**Related Issues/PRs**

half fix for #6837


patrickhulce:
our tests are harder to please though :)

PR #6851, opened 2018-12-20
tests(smoke): assert errorCode in expectations

Hoten:
`errorCode` wasn't being asserted for smoke tests. with this code, it will be. fin.

only effects the `errors` smoke tests

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -199,8 +199,8 @@ function findDifference(path, actual, expected) {
 
 /**
  * Collate results into comparisons of actual and expected scores on each audit.
- * @param {{finalUrl: string, audits: !Array}} actual
- * @param {{finalUrl: string, audits: !Array}} expected
+ * @param {{finalUrl: string, audits: !Array, errorCode: string}} actual
+ * @param {{finalUrl: string, audits: !Array, errorCode: string}} expected
  * @return {{finalUrl: !Object, audits: !Array<!Object>}}
  */
 function collateResults(actual, expected) {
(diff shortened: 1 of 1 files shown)


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -274,11 +280,12 @@ function reportAssertion(assertion) {
 /**
  * Log all the comparisons between actual and expected test results, then print
  * summary. Returns count of passed and failed tests.
- * @param {{finalUrl: !Object, audits: !Array<!Object>}} results
+ * @param {{finalUrl: !Object, audits: !Array<!Object>, errorCode: !Object}} results
```
patrickhulce: FWIW, we don't need all these `!` anymore, so it'd be cool to be cleaning them up as we go if we can, but I don't think any of the comments are real atm

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -199,8 +199,8 @@ function findDifference(path, actual, expected) {
 
 /**
  * Collate results into comparisons of actual and expected scores on each audit.
- * @param {{finalUrl: string, audits: !Array}} actual
- * @param {{finalUrl: string, audits: !Array}} expected
+ * @param {{finalUrl: string, audits: !Array, errorCode: string}} actual
```
patrickhulce:
```suggestion
 * @param {{finalUrl: string, audits: Array, errorCode?: string}} actual
```

patrickhulce: I don't think these files are actually type checked though, audits isn't an `Array`...

Hoten: Hah, it certainly isn't. Opened #6852.

brendankenny:
we should also be asserting the LHR form rather than the audit results and a few properties moved to the top level, but also for another day :)

PR #6857, opened 2018-12-21
report: support narrow screens

### LINKED ISSUE #6645: Lighthouse report shows horizontal scrollbar on mobile
1. Run LH on mobile for https://www.google.com via https://developers.google.com/web/tools/lighthouse/run
Report is generated, however, there is a horizontal scroll bar when viewed on mobile and rendering isn't optimal: https://builder-dot-lighthouse-ci.appspot.com/report.1543118339965.html
No horizontal scrollbar

#### Environment Information
* Affected Channels: 
* Lighthouse version: 3.2.1
* Node.js version:
* Operating System:

**Related issues**

exterkamp:
Transition from full width to >500px
mobile_view
minimum width without horizontal scroll (400px):
screen shot 2018-12-20 at 5 53 04 pm
minimum width without horizontal scroll (1 category):
screen shot 2018-12-20 at 5 53 55 pm

**Related Issues/PRs**
fixes #6645

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -105,7 +105,7 @@ class ReportUIFeatures {
   }
 
   _setupMediaQueryListeners() {
-    const mediaQuery = self.matchMedia('(max-width: 600px)');
+    const mediaQuery = self.matchMedia('(max-width: 500px)');
     mediaQuery.addListener(this.onMediaQueryChange);
     // Ensure the handler is called on init
     this.onMediaQueryChange(mediaQuery);
@@ -118,6 +118,8 @@ class ReportUIFeatures {
   onMediaQueryChange(mql) {
     const root = this._dom.find('.lh-root', this._document);
     root.classList.toggle('lh-narrow', mql.matches);
+    // Reset animations for the narrow view header height change.
+    this._setupHeaderAnimation();
   }
 
   _setupExportButton() {
@@ -226,8 +228,8 @@ class ReportUIFeatures {
     this.headerSticky.style.transform = `translateY(${heightDiff * scrollPct * -1}px)`;
     this.headerBackground.style.transform = `translateY(${scrollPct * this.headerOverlap}px)`;
     this.lighthouseIcon.style.transform =
-      `translate3d(var(--report-width-half),` +
-      ` calc(-100% - ${scrollPct * this.headerOverlap * -1}px), 0) scale(${1 - scrollPct})`;
+      `translate3d(0,` +
+      ` calc(-${scrollPct * this.headerOverlap * -1}px), 0) scale(${1 - scrollPct})`;
     this.lighthouseIcon.style.opacity = (1 - scrollPct).toString();
 
     // Switch up the score background & shadows
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -45,6 +45,8 @@ class ReportUIFeatures {
     /** @type {HTMLElement} */
     this.headerBackground; // eslint-disable-line no-unused-expressions
     /** @type {HTMLElement} */
+    this.headerMetadata; // eslint-disable-line no-unused-expressions
+    /** @type {HTMLElement} */
     this.lighthouseIcon; // eslint-disable-line no-unused-expressions
     /** @type {!HTMLElement} */
     this.scoresWrapperBg; // eslint-disable-line no-unused-expressions
@@ -118,8 +120,6 @@ class ReportUIFeatures {
   onMediaQueryChange(mql) {
     const root = this._dom.find('.lh-root', this._document);
     root.classList.toggle('lh-narrow', mql.matches);
-    // Reset animations for the narrow view header height change.
-    this._setupHeaderAnimation();
   }
 
   _setupExportButton() {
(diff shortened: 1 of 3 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -118,6 +118,8 @@ class ReportUIFeatures {
   onMediaQueryChange(mql) {
     const root = this._dom.find('.lh-root', this._document);
     root.classList.toggle('lh-narrow', mql.matches);
+    // Reset animations for the narrow view header height change.
```
patrickhulce: 👍

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -226,8 +228,8 @@ class ReportUIFeatures {
     this.headerSticky.style.transform = `translateY(${heightDiff * scrollPct * -1}px)`;
     this.headerBackground.style.transform = `translateY(${scrollPct * this.headerOverlap}px)`;
     this.lighthouseIcon.style.transform =
-      `translate3d(var(--report-width-half),` +
-      ` calc(-100% - ${scrollPct * this.headerOverlap * -1}px), 0) scale(${1 - scrollPct})`;
+      `translate3d(0,` +
```
patrickhulce: you might have to explain this one to me :)

lighthouse-core/report/html/report-styles.css
```diff
@@ -728,6 +731,7 @@
 
 .lh-report {
   background-color: #fff;
+  min-width: var(--report-min-width);
```
patrickhulce: this used to be only `@media screen` do we still want to keep that or doesn't matter?

lighthouse-core/report/html/templates.html
```diff
@@ -164,11 +164,13 @@
...
     }
+    .lh-narrow .lh-header-bg {
+      height: var(--report-header-height-narrow);
+    }
     .lh-lighthouse {
       position: absolute;
-      top: var(--report-header-height);
-      right: 50%;
-      transform: translate3d(var(--report-width-half), -100%, 0);
+      bottom: -4px;
```
patrickhulce: bottom -4?

lighthouse-core/report/html/templates.html
```diff
@@ -164,11 +164,13 @@
...
       will-change: transform;
     }
+    .lh-narrow .lh-header-bg {
+      height: var(--report-header-height-narrow);
+    }
     .lh-lighthouse {
       position: absolute;
-      top: var(--report-header-height);
-      right: 50%;
-      transform: translate3d(var(--report-width-half), -100%, 0);
```
patrickhulce: seems like these all changed because `lh-lighthouse` moved inside the header. what does that get us and do we know if there's a reason it was outside it to begin with? :)

lighthouse-core/report/html/renderer/report-ui-features.js
exterkamp: Since the `lh-lighthouse` element is inside the header container now it doesn't need to be translated on the x axis since it is positioned absolutely to the bottom right corner.

lighthouse-core/report/html/templates.html
exterkamp: The lighthouse SVG element is 4px above the bottom of the svg's wrapper div (which in this screenshot is set to 0px bottom) so I moved it 4px down to match the header height exactly.  I tried to fix the SVG but hand editing an SVG proved a bit difficult.  So I moved the whole thing 4 px down.
image

lighthouse-core/report/html/templates.html
exterkamp: Should have included a screenshot of the current UI:
image
Since the lighthouse isn't tied to the bottom right of the header container it doesn't scale correctly on small screens and clips off the edge like so!

lighthouse-core/report/html/report-styles.css
exterkamp: Hm, I don't think it really matters. I moved it here because I felt like it was always true for every screen not just `max-width: 964px` but I'm not sure it matters?

lighthouse-core/report/html/renderer/report-ui-features.js
patrickhulce: and the `100%` is gone because of the `right: 0` we're apply all the time now?

exterkamp: Yup, it had -100% on in the `.lh-lighthouse` class to position it before, so the new anchoring doesn't need it.

lighthouse-core/report/html/templates.html
```diff
@@ -164,11 +164,13 @@
       width: 100%;
       will-change: transform;
     }
+    .lh-narrow .lh-header-bg {
+      height: var(--report-header-height-narrow);
```
paulirish: eh. i dont think we need this. the jump is a bit awkward and putting the LH behind the text isn't terrible:
image

but if you wanted to avoid that.. we could do `  max-width: 60%;` on `.lh-config`
image

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -226,9 +229,10 @@ class ReportUIFeatures {
...
+      ` calc(-${scrollPct * this.headerOverlap * -1}px), 0) scale(${1 - scrollPct})`;
```
brendankenny: calc not needed anymore?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -226,9 +229,10 @@ class ReportUIFeatures {
...
     this.lighthouseIcon.style.opacity = (1 - scrollPct).toString();
+    this.headerMetadata.style.opacity = (1 - scrollPct).toString();
```
brendankenny: since `.lh-header` only contains the icon and the metadata, maybe save `this.header` instead and set opacity once on the whole thing?

exterkamp: Good call!

PR #6882, opened 2018-12-28
core(starturl): re-use already fetched manifest

### LINKED ISSUE #6881: Lighthouse unable to fetch manifest.json, favicons in Canary 73.0.3654.0
Problem
=====
Favicons and Manifest.json are not accessible to Lighthouse in Canary 73, but they are accessible to Lighthouse in Chrome 71. Failure to retrieve resources by clicking on corresponding hrefs in source view suggests the problem may be Canary itself, not Lighthouse.

OS, Browsers, URL
============
OS: Windows 8.1
Browsers: Chrome 71.0.3578.98 & Canary 73.0.3654.0
URL tested: https://ives-fourth-symphony.com

Lighthouse setup:
=================
Incognito Mode, all browser extensions off
Device: Mobile
All Audits Selected
Throttling: Simulated Fast 3G
Clear Storage: checked

Chrome 71.0.3578.98
===================
No PWA errors detected

Canary 73.0.3654.0
==================
Multiple PWA errors, all from inability to fetch manifest.json:

3: "No usable web app manifest found on page"
6: Failures: No manifest was fetched
8: Failures: No manifest was fetched
9: Failures: No manifest was fetched

The problem may be the browser itself, not Lighthouse.

View source of page in each browser
========================
Go to: view-source:https://ives-fourth-symphony.com/
Search for "href=manifest" and "href=favicon" in the view-source to select the manifest.json and the two favicons
Try clicking on the three links in the two browsers

Chrome 71.0.3578.98
===================
Success: Each link opens in a new tab

Canary 73.0.3654.0
==================
Failure: Neither resource is fetched by clicking on the links:

Theory: Something is wrong with Canary Version 73.0.3654.0

Please advise

patrickhulce:
**Summary**
We've talked about being able to reuse prior artifacts in other future gatherers but were always afraid of the precedence it sets for some horrifying graph of dependencies. This really is great use case though, so I wanted to play with it and see how it feels.

Therefore, I named the method `dangerouslyUsePreviousArtifact` in the hopes it assuages some concerns about this starting to be used everywhere 😆 

Waiting to update start-url test depending on what kind of reaction this gets :)

**Related Issues/PRs**
fixes #6881

### DIFF
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -472,7 +472,7 @@ class GatherRunner {
           /** @param {keyof LH.GathererArtifacts} name */
           async dangerouslyUsePreviousArtifact(name) {
             const gathererPromises = gathererResults[name];
-            if (!gathererPromises) throw new Error(`${name} not yet computed`);
+            if (!gathererPromises) throw new Error(`${name} Gatherer has not yet run`);
 
             const artifacts = {};
             await GatherRunner.convertGathererResultToArtifact(name, gathererPromises, artifacts);
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -365,6 +365,32 @@ class GatherRunner {
   }
 
   /**
+   * @param {keyof GathererResults} gathererName
+   * @param {GathererResults[keyof GathererResults]} phaseResultPromises
+   * @param {Partial<LH.GathererArtifacts>} gathererArtifacts
+   */
+  static async convertGathererResultToArtifact(gathererName, phaseResultPromises, gathererArtifacts) {
+    // If the artifact has already been computed, don't do anything, leave it as-is.
+    if (gathererArtifacts[gathererName] !== undefined) return;
+
+    try {
+      const phaseResults = await Promise.all(phaseResultPromises);
+      // Take last defined pass result as artifact.
+      const definedResults = phaseResults.filter(element => element !== undefined);
+      const artifact = definedResults[definedResults.length - 1];
+      // Typecast pretends artifact always provided here, but checked below for top-level `throw`.
+      gathererArtifacts[gathererName] = /** @type {NonVoid<PhaseResult>} */ (artifact);
+    } catch (err) {
+      // Return error to runner to handle turning it into an error audit.
+      gathererArtifacts[gathererName] = err;
+    }
+
+    if (gathererArtifacts[gathererName] === undefined) {
+      throw new Error(`${gathererName} failed to provide an artifact.`);
+    }
+  }
+
+  /**
    * Takes the results of each gatherer phase for each gatherer and uses the
    * last produced value (that's not undefined) as the artifact for that
    * gatherer. If an error was rejected from a gatherer phase,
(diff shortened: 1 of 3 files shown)


lighthouse-core/gather/gather-runner.js
```diff
@@ -458,6 +469,16 @@ class GatherRunner {
           passConfig,
           // *pass() functions and gatherers can push to this warnings array.
           LighthouseRunWarnings: baseArtifacts.LighthouseRunWarnings,
+          /** @param {keyof LH.GathererArtifacts} name */
+          async dangerouslyUsePreviousArtifact(name) {
```
paulirish: feels like this fn should be defined outside of this for loop.

patrickhulce: what would be your preferred flavor?

`dangerouslyUsePreviousArtifact: GatherRunner.dangerouslyUsePreviousArtifact.bind(gathererResults)`
`dangerouslyUsePreviousArtifact: GatherRunner.createPreviousArtifactFn(gathererResults)`
`dangerouslyUsePreviousArtifact(name) { return GatherRunner.dangerouslyUsePreviousArtifact(gathererResults, name); }`

paulirish: `dangerouslyUsePreviousArtifact: GatherRunner.dangerouslyUsePreviousArtifact.bind(gathererResults)`

patrickhulce:
> We've had use cases before but never a super necessary one, but maybe this is it.

I do think #6881 presents the super necessary use case :D

brendankenny:
>> We've had use cases before but never a super necessary one, but maybe this is it.

> I do think #6881 presents the super necessary use case :D

ha, sorry, I meant to say super necessary to share artifacts between gatherers instead of finding another solution. We definitely definitely should fix #6881 :) :)

paulirish:
DZL, do a barrel roll!

patrickhulce:
@brendankenny what concretely does 

>  I'm not super into this :) but I think it's definitely worth thinking about.

mean :)

Is it any of the following....?

- I won't approve a PR that does this, but it's a starting point for a discussion on what to do. 
- I might approve a PR that does this, but I need 24 hours to think about it.
- I would approve a PR that does this after I review it.

brendankenny:
> Is it any of the following....?

My hesitation is that we already have a lot of implicit interactions in the gathering stage (online/offline, affects perf while a trace is being collected, etc), but at least the `RequiredArtifacts` system works well for figuring out the dependency graph when filtering audits and etc. Adding this method breaks that and it can only stay kind of broken or get worse from here :)

A couple of other options:
- fix upstream (we didn't really look into what change caused #6881. Something in Chrome's handling of manifests? Or a protocol change in relation to `offline`?)
- work around (if the change was for a good reason, maybe there's a better way to do it)
- treat manifests as special like Chrome does. Chrome fetches the manifest on the first visit to a site and has it available from a cache on future visits, even offline. We could do the same thing in LH,  adding it to `baseArtifacts` when it's fetched, like `finalUrl` is updated.

I recognize the underlying issue that it's still vaguely reasonable to want to access artifacts between gatherers, but I think I favor the third option, since it lets us avoid that complexity and it matches how Chrome treats manifests as part of the meta information available for a site. I think we also have some help here, as I recently read that Chrome always fetches the manifest on first visit if a site has one, so we actually don't even need the `driver.getAppManifest()` call to get it out of the network traffic.

patrickhulce:
> We could do the same thing in LH, adding it to baseArtifacts when it's fetched, like finalUrl is updated.

I can get behind this 👍

patrickhulce:
btw @brendankenny I skipped the first two because the bug has already been fixed a bit ago in Chrome :) this is really just about making our starturl gatherer not refetch/reparse (and clumsily so it's currently susceptible to BOM issues for example)

I'll update the issue with a comment.

brendankenny:
> btw @brendankenny I skipped the first two because the bug has already been fixed a bit ago in Chrome :) 

oh, haha, whoops. I had no idea :)

patrickhulce:
@brendankenny scratch that last thing I said, the author fixed the service worker and I attributed that to a 1-day Canary bug instead 😆 they will be sharing source code though so we can also get to bottom of it

brendankenny:
closing in favor of #6957

Issue #6893, opened 2018-12-30
DevTools Error: The asynchronous expression exceeded the allotted time of 60


wertpokler:
**Initial URL**: https://www.myfreecams.com/
**Chrome Version**: 70.0.3538.110
**Error Message**: The asynchronous expression exceeded the allotted time of 60s
**Stack Trace**:
```
Error: The asynchronous expression exceeded the allotted time of 60s
    at _ (chrome-devtools://devtools/bundled/audits2_worker/audits2_worker_module.js:1000:120)
```

patrickhulce:
Thanks for filing! This specific error is no longer possible in latest version of Lighthouse, and I'm no longer able to reproduce. Check out the upcoming 4.0 release in Chrome Canary for the fix.

PR #6900, opened 2018-12-31
core(lantern): remove unnecessary sort calls

patrickhulce:
**Summary**
while investigating #6893 I noticed there was a crazy amount of time spent in graph simulation due to the 1200+ network requests they fire. Turns out most of this was just spent on just 2 `.sort` calls. One of which we don't even need sort, it's just a `max` and the other can be easily solved by maintaining an extra copy of the nodes that's sorted.

It shouldn't impact accuracy either.

**This reduced total auditing time from 17.2 s to 7.8 s**

**Related Issues/PRs**
 #6893

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
@@ -97,8 +97,9 @@ module.exports = class ConnectionPool {
    * @param {Array<TcpConnection>} connections
    */
   _findConnectionWithLargestCongestionWindow(connections) {
-    /** @type {TcpConnection|null} */
-    let maxConnection = connections[0] || null;
+    if (!connections.length) return null;
+
+    let maxConnection = connections[0];
     for (let i = 1; i < connections.length; i++) {
       const connection = connections[i];
       const currentMax = maxConnection.congestionWindow;
--- a/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
@@ -94,6 +94,21 @@ module.exports = class ConnectionPool {
   }
 
   /**
+   * @param {Array<TcpConnection>} connections
+   */
+  _findConnectionWithLargestCongestionWindow(connections) {
+    /** @type {TcpConnection|null} */
+    let maxConnection = null;
+    for (let i = 0; i < connections.length; i++) {
+      const connection = connections[i];
+      const currentMax = (maxConnection && maxConnection.congestionWindow) || -Infinity;
+      if (connection.congestionWindow > currentMax) maxConnection = connection;
+    }
+
+    return maxConnection;
+  }
+
+  /**
    * This method finds an available connection to the origin specified by the network record or null
    * if no connection was available. If returned, connection will not be available for other network
    * records until release is called.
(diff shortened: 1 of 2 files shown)


patrickhulce:
I'm sorting out the fact that it did change behavior hang on :)

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -93,6 +93,21 @@ module.exports = class ConnectionPool {
...
 
+  /**
+   * @param {Array<TcpConnection>} connections
+   */
+  _findConnectionWithLargestCongestionWindow(connections) {
+    /** @type {TcpConnection|null} */
+    let maxConnection = null;
+    for (let i = 0; i < connections.length; i++) {
+      const connection = connections[i];
+      const currentMax = (maxConnection && maxConnection.congestionWindow) || -Infinity;
```
Hoten: could remove this branching for `currentMax` by starting `maxConnection` at the first entry in `connections`, and starting the loop at `i = 1`. 

I assume you wouldn't need to bother with a `.length > 0` check because the previous code didn't bother.

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -114,21 +129,20 @@ module.exports = class ConnectionPool {
...
-      connectionToUse = availableConnections.find(
-        connection => connection.isWarm() === observedConnectionWasReused
-      );
-    }
-
+    // Find the set of available connections based on the acquire options
+    const availableConnections = connections
+      .filter(connection => {
+        // Connections that are in use are never available
+        if (this._connectionsInUse.has(connection)) return false;
```
Hoten: maybe move this check to the end, since it's more work than the others. would be nice to defer.

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
patrickhulce: I think no matter what I'll need a `ts-ignore` or a check on `maxConnection`. I don't think ts is smart enough to know that starting the loop at `1` means that `maxConnection` is necessarily defined

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
patrickhulce: done

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
Hoten: the explicit `return null` was a good call, very nice

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -134,6 +137,12 @@ class Simulator {
    * @param {number} queuedTime
    */
   _markNodeAsReadyToStart(node, queuedTime) {
+    const firstNodeIndexWithGreaterStartTime = this._cachedNodeListByStartTime
+      .findIndex(candidate => candidate.startTime > node.startTime);
+    const insertionIndex = firstNodeIndexWithGreaterStartTime === -1 ?
+      this._cachedNodeListByStartTime.length : firstNodeIndexWithGreaterStartTime;
+    this._cachedNodeListByStartTime.splice(insertionIndex, 0, node);
```
brendankenny: a) How do we only end up with 2x test coverage of this? Or is codecov messed up? Something seems wrong with that.
b) `_nodes[NodeState.ReadyToStart]` (and the local `nodesReadyToStart`) appears unused now and I don't believe we use the uniqueness of `Set` for anything. Should we get rid of it? (I regret all the `_nodes[NodeState]` business now, incidentally :)
c) We only traverse in order of start time, should we switch out with something like a heap? Iterating over them in order goes up to n log n, but insertion/deletion drops to log n

brendankenny: > this LGTM, but if you want to do a little more perf golfing... :)

well, actually we really should get rid of `_nodes[NodeState.ReadyToStart]` or at least leave a note to refactor since it's not used anymore.

patrickhulce: > Or is codecov messed up? 

codecov is definitely messed up, this will get hit for every node on every single simulation call which we have lots of coverage for

> _nodes[NodeState.ReadyToStart] (and the local nodesReadyToStart) appears unused now and I don't believe we use the uniqueness of Set for anything. Should we get rid of it?

seeing as I didn't really think it was necessary in the first place, I'm open to it 😉 

I will say the indexed by state has grown on me though, the clarity at least that they're moving through the system in these sets is appreciated when returning to the code after a long absence. I added a comment explaining this for now.

brendankenny: > I will say the indexed by state has grown on me though, the clarity at least that they're moving through the system in these sets is appreciated when returning to the code after a long absence.

that's a fair point, but maybe there's a better way to handle enumerated state like that. As it is, it's difficult to find all the places we insert nodes :)

patrickhulce: > should we switch out with something like a heap?

I'm somewhat iffy of the ROI on that. Math-wise I'm with ya, but the fact that we just moved from a situation that Math-wise was `O(n log n)` iteration and `O(1)` insertion makes me skeptical it'll be worth the extra complexity since our current tradeoff favored lowering iteration time.

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -114,21 +130,23 @@ module.exports = class ConnectionPool {
...
-      );
-    }
-
+    // Find the set of available connections based on the acquire options
+    const availableConnections = connections
+      .filter(connection => {
+        // Normally, we want to make sure the connection warmth matches the state of the record
+        // we're acquiring for. Do this check first since it's the common case and cheaper than our
+        // "in use" check below.
+        // Use the _warmed property instead of the getter because this is a surprisingly hot code path.
```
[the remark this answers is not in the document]
patrickhulce: yup, sg!

brendankenny:
> codecov is definitely messed up, this will get hit for every node on every single simulation call which we have lots of coverage for

yeah, something is pretty wrong. https://coveralls.io/builds/20889041/source?filename=lighthouse-core/lib/dependency-graph/simulator/simulator.js#L136 shows 10K coverage for these lines

brendankenny:
ha, apparently that number isn't statement hit count. It's "Builds"...whatever that is.

PR #6930, opened 2019-01-04
report: expand groups within Passed Audits

paulirish:
This was the UX conclusion when PSI noticed the groupings in our new renderer.


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -252,7 +251,10 @@ class CategoryRenderer {
 
       // Push grouped audits as a group.
       const groupDef = groupDefinitions[groupId];
-      const auditGroupElem = this.renderAuditGroup(groupDef, opts);
+      // Expanded by default!
+      // 1. The 'failed' clump has all groups expanded.
+      // 2. If a clump is collapsed (passed, n/a), we want the groups within to already be expanded
+      const auditGroupElem = this.renderAuditGroup(groupDef, {expandable: false});
```
patrickhulce: this is a little confusing, it's probably because I don't remember what our opts actually do at this point 😆 so `expandable` means `isExpansionToggleable` and it's expanded by default?

paulirish:
> I'd prefer this, but I believe the conclusion was that there would be no groups under passed.

Hmm I dont remember that.  but Ok. :)


---------

Here's a GIF showing (1) current master - two collapsed groups under Passed. then (2) this PR - two expanded groups under Passed. The gif doesn't show (C) current PSI - groups are intermingled under Passed.

kapture 2019-01-04 at 12 40 32

Brendan thought the conclusion was option C, but I forget, tbh.
 @kusler wanna make a call?

lighthouse-core/report/html/renderer/category-renderer.js
paulirish: agree it's a funny name. `expandable` means `collapsedAndCanBeExpanded`

brendankenny:
yes, I remember it clearly because I was sad to lose the headings under Passed (I believe the argument was that you don't really need them at that point, since they're just a list of passed audits), and I had just finished my sweet ascii art and didn't want to change it :)

Hoten:
should https://github.com/GoogleChrome/lighthouse/pull/6279 be closed?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -272,7 +274,7 @@ class CategoryRenderer {
    */
   renderUnexpandableClump(auditRefs, groupDefinitions) {
```
brendankenny: weren't you going to rename?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -291,7 +293,8 @@ class CategoryRenderer {
    *      ├── audit 5
    *      └── audit 6
    * clump (e.g. 'manual')
-   *   ├── …
+   *   ├── audit 1
+   *   ├── audit 2
    *   ⋮
```
brendankenny: maybe
```
/**
 * Renders a clump (a top level report section), under a status of failed, manual,
 * passed, or notApplicable. The result ends up something like:
 *
 * failed clump
 *   ├── audit 1 (w/o group)
 *   ├── audit 2 (w/o group)
 *   ├── audit group
 *   |  ├── audit 3
 *   |  └── audit 4
 *   └── audit group
 *      ├── audit 5
 *      └── audit 6
 * other clump (e.g. 'manual')
 *   ├── audit 1
 *   ├── audit 2
 *   ├── …
 *   ⋮
```

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -305,17 +308,15 @@ class CategoryRenderer {
       return failedElem;
     }
 
-    const expandable = true;
-    const elements = this._renderGroupedAudits(auditRefs, groupDefinitions, {expandable});
-
     const clumpInfo = this._clumpDisplayInfo[clumpId];
     // TODO: renderAuditGroup shouldn't be used to render a clump (since it *contains* audit groups).
```
brendankenny: this TODO makes no sense now. Maybe `// TODO: renderAuditGroup shouldn't be used to render a clump since it's a clump, not a group.`

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -305,17 +308,15 @@ class CategoryRenderer {
...
     const groupDef = {title: clumpInfo.title, description};
-    const opts = {expandable, itemCount: auditRefs.length};
-    const clumpElem = this.renderAuditGroup(groupDef, opts);
+    const groupOpts = {expandable: true, itemCount: auditRefs.length};
+    const clumpElem = this.renderAuditGroup(groupDef, groupOpts);
     clumpElem.classList.add('lh-clump', clumpInfo.className);
 
-    elements.forEach(elem => clumpElem.appendChild(elem));
+    // For all non-failed clumps, we don't group
```
brendankenny: This might not make much sense outside the context of this PR. Just `// Add all audit results to the clump.` (or add an intermediate `const auditElems = auditRefs.map(...)` and can probably avoid the comment altogether)

lighthouse-core/report/html/renderer/category-renderer.js
brendankenny: > weren't you going to rename?

oh, maybe because `pwa-category-renderer.js` also uses it?

brendankenny:
proposal on top of this in #6982

Issue #6615, opened 2018-11-20
Change mobile viewport size to match Nexus 5X


mattzeunert:
Lighthouse uses a resolution of 412x732 for the mobile setting, but we're trying to match the Nexus 5X viewport size which is only 412x660 (because of the phone chrome around it).

@patrickhulce said we might be able to change it. One upside would be that we could tell users to use Chrome's Nexus 5X emulation settings to test viewport-specific audit results, e.g. tap target sizes.

PR #6932, opened 2019-01-05
misc(emulation): Use Nexus 5X emulation real screen size

exterkamp:
**Summary**
I'll bite and try to adjust this.  I just changed the emulation heights to 660.  I wasn't sure if `height` should be 732 and then `screenHeight` should be 660 or the other way around?  [Docs](https://chromedevtools.github.io/devtools-protocol/tot/Emulation#method-setDeviceMetricsOverride) didn't explain which was which, so I just made them both 660 for now.

When changing the emulated device in the frontend it only sends a screenWidth and screenHeight with width & height set to 0:
image
Note the 660 is derived on the frontend from: 732px - 24px (top bar) - 48px (bottom bar) = 660px

**Related Issues/PRs**
fixes: #6615

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -35,7 +35,7 @@ function getFlags(manualArgv) {
           'lighthouse <url> --disable-device-emulation --throttling-method=provided',
           'Disable device emulation and all throttling')
       .example(
-          'lighthouse <url> --chrome-flags="--window-size=412,732"',
+          'lighthouse <url> --chrome-flags="--window-size=412,660"',
           'Launch Chrome with a specific window size')
       .example(
           'lighthouse <url> --quiet --chrome-flags="--headless"',
--- a/lighthouse-core/lib/emulation.js
+++ b/lighthouse-core/lib/emulation.js
@@ -14,9 +14,9 @@
 const NEXUS5X_EMULATION_METRICS = {
   mobile: true,
   screenWidth: 412,
-  screenHeight: 732,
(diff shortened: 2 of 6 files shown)


exterkamp:
There is a lot of `height: 732` `visibleHeight: 732` etc. in the test/fixtures/traces files, don't know if we need to change those too :man_shrugging:

patrickhulce:
> I wasn't sure if height should be 732 and then screenHeight should be 660 or the other way around? Docs didn't explain which was which, so I just made them both 660 for now.

https://cs.chromium.org/chromium/src/third_party/blink/public/web/web_device_emulation_params.h?sq=package:chromium&g=0&l=23-33 suggests that width/height is the viewport which can be scrolled maybe using view_position? Not sure either 🤷‍♂️ Either way, copying what frontend does SGTM :)


This is probably going to have some unexpected ramifications down the line. I still think it needs to be done obviously since we're claiming Nexus 5X and the right aspect ratio should be used for it 😃, just trying to think of what could happen.

- Speed Index and anything using ViewportDimensions height will likely change (just offscreen images I think that needs height?)
- Screenshots will be different dimensions (probably doesn't matter to us so much but maybe consumers? @benschwarz @mattzeunert both of yours don't rely on default emulation and our screen size do they?)
- Something else??

mattzeunert:
Yeah, I'm not relying on the default emulation settings.

Can't think of anything else that would change (as long as the page doesn't render differently based on the changed window height).

brendankenny:
DZL, do a barrel roll!

PR #6944, opened 2019-01-07
core(driver): waitForFCP when tracing

### LINKED ISSUE #6941: Page load doesn't wait for FCP before network quiet
For a simple HTML that generates content through JS, both PSI and the audit done through the developer tools states an "error!" instead of metric value - [screenshot](https://screenshot.googleplex.com/D35kVNsPQh0). 

URL: http://hkahkjaswke1.appspot.com/SlowServe?page=jsContent.html

patrickhulce:
**Summary**
We don't actually wait for an FCP, and it seems like we should. There could be a lot more to this story of making our `waitForFullyLoaded` match TTI conditions as close as possible, but this is a good start.

**Fully Loaded Conditions After This PR**
```
all [load, FCP, network quiet] --> CPU quiet --> done
```

**Potential Easy Alternative**
```
all [FCP, load] --> network quiet --> CPU quiet --> done
```

**Real TTI Conditions**
```
FCP --> simultaneous network + CPU quiet --> done
```

**Related Issues/PRs**
closes #6941

### DIFF
--- a/lighthouse-cli/test/fixtures/delayed-fcp.html
+++ b/lighthouse-cli/test/fixtures/delayed-fcp.html
@@ -1,6 +1,11 @@
 <html>
 <meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
 <body>
+  <!--
+    This page intentionally contains no content for the first 6 seconds.
+    After 6 seconds, text is finally added with a setTimeout below.
+    We are testing Lighthouse's ability to wait for First Contentful Paint, not just page load.
+  -->
   <script>
     setTimeout(() => {
       const textEl = document.createElement('span');
--- a/lighthouse-cli/test/fixtures/delayed-fcp.html
+++ b/lighthouse-cli/test/fixtures/delayed-fcp.html
@@ -1,0 +1,12 @@
+<html>
+<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
+<body>
+  <script>
+    setTimeout(() => {
+      const textEl = document.createElement('span');
+      textEl.textContent = 'Hello, World!';
+      document.body.appendChild(textEl);
+    }, 6000);
+  </script>
+</body>
+</html>
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -540,25 +540,21 @@ class Driver {
       throw new Error('_waitForFCP.cancel() called before it was defined');
     };
 
-    const promise = (async () => {
-      await this.sendCommand('Page.setLifecycleEventsEnabled', {enabled: true});
-
-      return new Promise(resolve => {
-        /** @param {LH.Crdp.Page.LifecycleEventEvent} e */
-        const lifecycleListener = e => {
-          if (e.name === 'firstContentfulPaint') {
-            resolve();
-            this.off('Page.lifecycleEvent', lifecycleListener);
-          }
-        };
-
-        this.on('Page.lifecycleEvent', lifecycleListener);
-
-        cancel = () => {
+    const promise = new Promise(resolve => {
+      /** @param {LH.Crdp.Page.LifecycleEventEvent} e */
+      const lifecycleListener = e => {
+        if (e.name === 'firstContentfulPaint') {
+          resolve();
           this.off('Page.lifecycleEvent', lifecycleListener);
-        };
-      });
-    })();
+        }
+      };
+
+      this.on('Page.lifecycleEvent', lifecycleListener);
+
+      cancel = () => {
+        this.off('Page.lifecycleEvent', lifecycleListener);
+      };
+    });
 
     return {
       promise,
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -516,6 +516,7 @@ class Driver {
    * Returns a promise that resolves immediately.
    * Used for placeholder conditions that we don't want to start waiting for just yet, but still want
    * to satisfy the same interface.
+   * @return {{promise: Promise<void>, cancel: function(): void}}
    */
   _waitForNothing() {
     return {promise: Promise.resolve(), cancel() {}};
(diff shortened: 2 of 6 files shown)


paulirish:
```
node lighthouse-cli/index.js http://localhost:10200/infinite-loop.html --config-path=/home/travis/build/GoogleChrome/lighthouse/lighthouse-cli/test/smokehouse/error-config.js --output-path=smokehouse-19845.report.json --output=json --quiet --port=0 
Runtime error encountered: _waitForLoadEvent.cancel() called before it was defined
Error: _waitForLoadEvent.cancel() called before it was defined
    at Object.cancel (/home/travis/build/GoogleChrome/lighthouse/lighthouse-core/gather/driver.js:182:163)
    at /home/travis/build/GoogleChrome/lighthouse/lighthouse-core/gather/driver.js:236:424
    at Driver._waitForFullyLoaded (/home/travis/build/GoogleChrome/lighthouse/lighthouse-core/gather/driver.js:237:125)
    at <anonymous>
```

lighthouse-cli/test/fixtures/delayed-fcp.html
```diff
@@ -0,0 +1,12 @@
+<html>
+<meta name="viewport" content="width=device-width, initial-scale=1, minimum-scale=1">
+<body>
+  <script>
+    setTimeout(() => {
```
brendankenny: maybe add a comment as to the goal of this page?

lighthouse-core/gather/driver.js
```diff
@@ -522,6 +531,37 @@ class Driver {
     });
   }
 
+  /**
+   * Returns a promise that resolve when a frame has a FCP.
+   */
+  _waitForFCP() {
```
brendankenny: I'm personally a fan of the explicit return types both for safety (tsc will happily start unioning things in cases of mismatched early returns, etc) and for documentation purposes (intellisense hover shouldn't be required :).

We should probably get into that at some point because I've definitely noticed you leaving them off more recently :)

lighthouse-core/gather/driver.js
```diff
@@ -874,23 +918,25 @@ class Driver {
    * possible workaround.
    * Resolves on the url of the loaded page, taking into account any redirects.
    * @param {string} url
-   * @param {{waitForLoad?: boolean, waitForNavigated?: boolean, passContext?: LH.Gatherer.PassContext}} options
+   * @param {{waitForFCP?: boolean, waitForLoad?: boolean, waitForNavigated?: boolean, passContext?: LH.Gatherer.PassContext}} options
```
brendankenny: is this pulled out as another option for performance reasons, or does fcp not fire in certain passes and we need the option so we don't always wait for the full timeout? It's unfortunate mental overhead since none of the `wait*` options are independent. Is there a way we can reorganize?

patrickhulce: Correcto. On the passes without javascript and offline this can take unnecessarily long when we don't really care about FCP in those passes. In fact, every page that fails the `without javascript` audit would take 45s+ because all we're doing is checking for the existence of text.

> It's unfortunate mental overhead since none of the wait* options are independent. Is there a way we can reorganize?

Agreed. There are really only 3 scenarios despite the 3 total booleans. We could switch to an enum? Opposed to doing it in followup?

lighthouse-core/gather/driver.js
patrickhulce: Yeah I've been leaving it off wherever I see no ambiguous branching going on :)

This one isn't bad though I'll add it.

lighthouse-cli/test/fixtures/delayed-fcp.html
patrickhulce: done

lighthouse-core/gather/driver.js
brendankenny: > There are really only 4 scenarios despite the 3 total booleans. We could switch to an enum? Opposed to doing it in followup?

No, that sounds good. Maybe pair it with a switch to "Potential Easy Alternative" (should mostly match what behavior today) or "Real TTI Conditions" (if we're feeling really brave)

patrickhulce:
> Should we do a DZL?

Fo sho

DZL, do a barrel roll!

brendankenny:
ha, well, I don't know how to interpret the DZL. Two sites slowed significantly in gather time, but the others didn't :) I think that's not unexpected?

One other thing (possibly the cause for the above): for LH runs that end up with the `NO_FCP` error, we'll now hit the pass timeout while we were probably stopping passes earlier before. That's not a huge deal on the CLI, I don't think (since the results are all audit errors regardless), but might be more of a big deal for LR @paulirish

patrickhulce:
> Two sites slowed significantly in gather time, but the others didn't :) I think that's not unexpected?

This is my eternal struggle to get around with DZL for the past month. The comparison hash in that URL was for several days off so the sites were just different `1/4/2019, 8:14:04 PM` vs `1/7/2019, 4:48:23 PM`. If you take a look at the [exact hash from the same day as this was run](http://dzl.patrickhulce.com/dashboard-comparison?label=official-continuous&comparison=branch-6944), you'll find completely different runtime results. I've been doing lots of experiments to figure out how we can get a solid subset, but for now, most runtime results in DZL are pretty much unactionable :(

patrickhulce:
Yes indeed if the page truly never paints content then we will wait until our timeout instead load/network quiet.

> but might be more of a big deal for LR

Interesting, just from a capacity point of view? What would the alternative be?

brendankenny:
> just from a capacity point of view?

Yeah. I think they were having a bunch of NO_FCP at some point, so the timeout rate might significantly increase in the logs? I have no idea if it would matter, though.

brendankenny:
@exterkamp says it doesn't matter and he'll be happy if he can watch the graphs change a bit :)

@paulirish just concurred.

Ready for merge?

patrickhulce:
baller ⛹️‍♀️

PR #6972, opened 2019-01-09
report(css): use padding for PWA vertical border space

exterkamp:
**Summary**
The vertical separator looks squished when looking at small width screens.  On large width screens this makes a not very noticable 1px difference :man_shrugging: 

Don't know if this is WAI, but I wanted to mention it since I thought it looked a bit off.

Current view on min-width screens:
margin
With padding instead of margin:
padding


lighthouse-core/report/html/report-styles.css
```diff
@@ -786,12 +786,13 @@
...
 
-.lh-scores-header .lh-gauge__wrapper,
-.lh-scores-header .lh-gauge--pwa__wrapper {
+.lh-scores-header .lh-gauge__wrapper {
   margin: 0 4px;
 }
 
 .lh-scores-header .lh-gauge--pwa__wrapper {
+  margin-right: 4px;
+  padding-left: 3px; /* 4px - 1px border */
```
patrickhulce: why do we need to subtract the border? didn't we have 9px of space before with margin-left (4) + margin-right (4) + border (1)

exterkamp: Yeah it is currently a 9px space.  I removed the 1px just to make it 4x4 even because the symmetry is pleasing.  It's fair to make it 4x4 + 1px border 🤷‍♂️ I'm just brining up the concern.

patrickhulce: maybe my CSS math is just off :) but isn't having only 3px of padding making the border spacing asymmetrical?  or are you referring to consistent spacing between the category scores?

exterkamp: It makes the overall spacing symmetrical with the rest of the gauges.  Or at least that was my intention.

lighthouse-core/report/html/report-styles.css
```diff
@@ -786,12 +786,13 @@
   border: 0;
 }
 
-.lh-scores-header .lh-gauge__wrapper,
-.lh-scores-header .lh-gauge--pwa__wrapper {
+.lh-scores-header .lh-gauge__wrapper {
   margin: 0 4px;
```
patrickhulce: sorry one last thing, is there a reason we picked margin at all? why not just go `padding: 0 4px;` :)

exterkamp: I assumed there was a reason, but I have no idea.  My CSS gut says this _is margin_ not padding, but I have no strong opinion at all.  Might be better to simplify.

patrickhulce: looks like it was added [here](https://github.com/GoogleChrome/lighthouse/pull/5706/files#diff-f8838d243bb796c2800f4eceab80b7ddR724) @paulirish any memory of why we might not want to just go with padding here now?

paulirish: padding wfm!

brendankenny: there was no border before, so it might have just not mattered at the time

brendankenny: Not a huge deal, but do you want to pull this into a variable since the `pwa__wrapper` padding is based on this one? Someday if we change it, it'll be nice if the connection is explicit

brendankenny: otherwise could make the comment in `.lh-gauge--pwa__wrapper` more explicit about the 4px matching this one

exterkamp: changed to padding :smile:

exterkamp: Since this is just a move to padding now, don't think we need a var anymore.  But good advice!

brendankenny: 💥

PR #6973, opened 2019-01-09
core(font-size): precise text length calculation

castilloandres:
**Summary**
**What kind of change does this PR introduce?**
Count JavaScript string symbols not Unicode code points.

**Is this a bugfix, feature, refactoring, build related change, etc?**
Refactoring.

To calculate text lengths on the font size gatherer with more precision.

**Link any documentation or information that would help understand this change**
```
> '𝐀'.length
2
> Array.from('𝐀').length
1
> '💩'.length
2
> Array.from('💩').length
1
> '再'.length
2
> Array.from('再').length
1
```

### DIFF
--- a/lighthouse-core/gather/gatherers/seo/font-size.js
+++ b/lighthouse-core/gather/gatherers/seo/font-size.js
@@ -182,7 +182,7 @@ function getEffectiveFontRule({inlineStyle, matchedCSSRules, inherited}) {
  * @returns {number}
  */
 function getNodeTextLength(node) {
-  return !node.nodeValue ? 0 : node.nodeValue.trim().length;
+  return !node.nodeValue ? 0 : Array.from(node.nodeValue.trim()).length;
 }
 
 /**
--- a/lighthouse-core/test/gather/gatherers/seo/font-size-test.js
+++ b/lighthouse-core/test/gather/gatherers/seo/font-size-test.js
@@ -10,9 +10,9 @@
 const FontSizeGather = require('../../../../gather/gatherers/seo/font-size');
 let fontSizeGather;
 
-const smallText = ' body small text ';
-const bigText = 'body big text';
-const failingText = 'failing text';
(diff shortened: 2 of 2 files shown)


lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -182,7 +182,7 @@ function getEffectiveFontRule({inlineStyle, matchedCSSRules, inherited}) {
  * @returns {number}
  */
 function getNodeTextLength(node) {
-  return !node.nodeValue ? 0 : node.nodeValue.trim().length;
+  return !node.nodeValue ? 0 : Array.from(node.nodeValue.trim()).length;
```
exterkamp: Can we get a comment mentioning this PR or explaining why this is here. e.g.
```
// Use Array.from in order to more accurately count unicode characters. See: #6973
```
Maybe even add a `TODO` for @patrickhulce's comment on future rabbit hole exploring.

castilloandres:
> I wonder how far down the rabbit hole we should go.

@patrickhulce @exterkamp I think we should go as deep as readable text goes. Apart for emojis this should count more or less foreign character sets correctly 👍

lighthouse-core/gather/gatherers/seo/font-size.js
```diff
@@ -182,7 +182,8 @@ function getEffectiveFontRule({inlineStyle, matchedCSSRules, inherited}) {
  * @returns {number}
  */
 function getNodeTextLength(node) {
-  return !node.nodeValue ? 0 : node.nodeValue.trim().length;
+  // Array.from to count JS symbols not unicode code points. See: #6973
```
Hoten: JS Symbols => character ?

patrickhulce:
Thanks very much for the explanation @Mickael-van-der-Beek! That's the example I pulled from the article I linked which came to similar conclusions as yours :)

PR #6982, opened 2019-01-11
report: clean up more clump/group/expandable crossover noise

brendankenny:
works on top of #6930

@paulirish this is how far I think the simplification can go.

Gets rid of all `expandable` true/false business, groups and clumps that are `<divs>` and others that are `<details>`, etc. Clumps still use the group class names, but no longer use `renderAuditGroup` to make themselves.


brendankenny:
> travis wants pwa-category-renderer updated FYI

I blame Paul's cherry picking skills :P

I'll fix.

brendankenny:
ok, jk, that was my fault :)

Issue #6901, opened 2018-12-31
Code snippet details renderer type


mattzeunert:
We're building this primarily for [JSON-LD validation results](https://github.com/GoogleChrome/lighthouse/issues/4359), but are hoping to re-use it elsewhere eventually.

The type looks like this: 

```
interface DetailsRendererCodeSnippetItem {
      type: "code-snippet",
      code: string,
      title: string,
      highlights: {
          line: number | null,
          message: string
      }[]
} 
```

Two rendered items:

![screenshot 2018-12-31 at 21 38 53](https://user-images.githubusercontent.com/1303660/50568368-9c253000-0d48-11e9-9990-bfa6ed337d38.png)

--- 

Putting the `code-snippet` items into a single-column table looks bad. I'm planning to create another item type `code-snippet-list` to render them. Does that sound reasonable?

I guess it could also be called `lh-list` since it's really just a list with padding between items.

Not sure if we'd ever want it to be part of a table – [here's how that would look like right now](https://user-images.githubusercontent.com/1303660/50568467-2cb04000-0d4a-11e9-990a-3f75a8d8320d.png).

What do you think about how to structure the item types here?

exterkamp:
I like this look, I definitely agree that a 1 column table would look odd.  Definitely like the look of a full width code-snippet list.

I feel like this could be an `lh-list` or maybe `lh-full-width-list`?

Does each snippet contain the full JSON-LD file for expansion?  Should it be a fixed # of lines, I don't see why we would need to include an expansion on each snippet as long as we would preserve the overall global line numberings.  Just like the +/- 2 lines from the error.

@hwikyounglee thoughts on the look and feel?  Stylistically I might want the gray outlines to go away to match the full width tables' style.

mattzeunert:
@rviscomi brought up linking the JSON-LD item to the node in the DevTools elements panel.

The linking works by [looking for `.lh-node` elements](https://github.com/ChromeDevTools/devtools-frontend/blob/da0e4c607c6bbab206b32d406259daa196869fa7/front_end/audits2/Audits2ReportRenderer.js#L39), so we could render one of those and slightly restyle it. Or we can tweak the DevTools code to pick up `data-path` attributes.

Do we want the UI to look something like this? 

![screenshot 2019-01-06 at 09 42 56](https://user-images.githubusercontent.com/1303660/50734465-81fbaf80-1197-11e9-84ed-e175aa0105f3.png)

Hoten:
[...]

PR #6999, opened 2019-01-13
core(details-renderer): snippet details renderer type

mattzeunert:
**Summary**

Details renderer type that shows a code snippet with error highlights.

[Example LH report using code snippets for structured data](https://jsonld-display.glitch.me/#seo)

screenshot 2019-01-13 at 15 54 45

**Status**

~~Not 100% finished but would be good to get some initial review.~~

Functionality is complete now.

Reverted JSON-LD commit is just here because I use it for testing.

**Related Issues/PRs**

Closes: https://github.com/GoogleChrome/lighthouse/issues/6901


mattzeunert:
Getting this on local, is there anything else I need to do for i18n?

> google.protobuf.json_format.ParseError: Failed to parse i18n field: Failed to parse rendererFormattedStrings field: Message type "I18n.RendererFormattedStrings" has no field named "codeSnippetCollapse".

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    return header;
+  }
+
+  /**
+   * @param {DOM} dom
+   * @param {DocumentFragment} tmpl
+   * @param {{content: string, number: number | string,truncated?: boolean}} line
+   * @param {LineOptions} lineOptions
```
patrickhulce: this boolean set of options is a little hard to follow. I'm trying to think if it's possible to have an enum of line states or if the combinations really are sufficiently complicated. I'm convinced we can't reduce it down to 1 but maybe a handful?

Is this right?

1. A line is either `code`, `highlightMessage`, or `empty`
2. A line can be `highlighted` or `notHighlighted`
3. A line is either `alwaysVisible`, `visibleWhenCollapsed`, or `visibleWhenExpanded`

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  /**
+   * @param {DOM} dom
+   * @param {DocumentFragment} tmpl
+   * @param {LH.Audit.DetailsRendererCodeSnippetHighlight} highlight
+   * @return {Element}
+   */
+  static renderHighlightMessage(dom, tmpl, highlight) {
+    return CodeSnippetRenderer.renderLine(dom, tmpl, {
```
patrickhulce: this is when it clicked for me that `renderLine` doesn't mean `render a line of code` but just `render a line of content in our DOM`. not sure how to make that clearer earlier, maybe just jsdoc? or if every call had the `contentType: 'code'|'highlightMessage'|'empty'` enum suggestion I mentioned :D

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+    const nonLineSpecificHighlights = highlights.filter(h => typeof h.lineNumber !== 'number');
+    nonLineSpecificHighlights.forEach(h =>
+      snippet.append(CodeSnippetRenderer.renderHighlightMessage(dom, tmpl, h))
+    );
+
+    snippet.append(CodeSnippetRenderer.renderSnippetLines(dom, tmpl, details));
+
+    // If expanded view still doesn't include all lines then indicate that that
+    const firstLineIsVisible = lines[0].number === 1;
+    const lastLineIsVisible = lines.slice(-1)[0].number === lineCount;
```
patrickhulce: clever! I'm not 100% convinced it's that much easier to grok than `lines[lines.length - 1].number` though ;)

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+  static renderSnippetLines(dom, tmpl, details) {
+    const {highlights, lineCount, lines} = details;
+    const collapsedLines = Util.filterRelevantLines(lines, highlights, 2);
+    const hasLineSpecificHighlights = highlights.some(h => typeof h.lineNumber === 'number');
+    const hasOnlyNonLineSpecficHighlights = highlights.length > 0 && !hasLineSpecificHighlights;
+
+    const lineContainer = dom.createFragment();
+
+    // Keep track if a highlighted line has been seen, because in collapsed mode
+    // we don't show the omitted lines indication before the first visible set of lines
```
patrickhulce: is this because it's handled by the parent already? would it simplify things to let this piece handle it instead?

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+          lineContainer.append(
+            CodeSnippetRenderer.renderOmittedLines(dom, tmpl, {expandedOnly: true}));
+        }
+        if (collapsedLine && !collapsedPreviousLine) {
+          lineContainer.append(
+            CodeSnippetRenderer.renderOmittedLines(dom, tmpl, {collapsedOnly: true}));
+        }
+      }
+
+      if (!line) {
```
patrickhulce: is this when we iterate through the omitted lines?

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -264,6 +272,22 @@ class DetailsRenderer {
...
+  /**
+   * @param {TableDetailsJSON} details
+   * @return {Element}
+   */
+  _renderList(details) {
+    const listContainer = this._dom.createElement('div', 'lh-list');
+
+    details.items.forEach(item => {
+      const itemEl = this._dom.createElement('div');
+      itemEl.append(this.render(item));
```
patrickhulce: I feel like we had to avoid forward looking nested details types because of protobuf/LR concerns, but I don't remember exactly where that landed.

was this a real limitation @exterkamp or am I making that up?

if we do, maybe we just keep it a `code-snippet-list` for now

lighthouse-core/test/audits/audit-test.js
```diff
@@ -99,4 +99,77 @@ describe('Audit', () => {
     assert.equal(result.score, null);
     assert.equal(result.scoreDisplayMode, 'error');
   });
+
+  // is audit the right place to put this logic?
```
patrickhulce: not really, but it's where we have it right now so I think you're fine :)

might make sense to split off into some `details-helpers` file or something in the future

lighthouse-core/test/report/html/renderer/code-snippet-renderer-test.js
```diff
@@ -0,0 +1,92 @@
...
+    const {document} = new jsdom.JSDOM(TEMPLATE_FILE).window;
+    dom = new DOM(document);
+  });
+
+  afterAll(() => {
+    global.Util = undefined;
+  });
+
+
+  it('Renders code snippet', () => {
```
patrickhulce: there's *a lot* of complexity in the combinations of cases I'd love to test a few more once we settle on how it should actually look :)

patrickhulce:
> Getting this on local, is there anything else I need to do for i18n?

yeah you'll have to update the `I18n.RendererFormattedStrings` message in `proto/lighthouse-result.proto` with the new property

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+ */
+'use strict';
+
+/* eslint-env jest */
+
+/* globals self, Util */
+
+/** @typedef {import('./details-renderer')} DetailsRenderer */
+
+const SHOW_IF_EXPANDED_CLASS = 'lh-code-snippet__show-if-expanded';
```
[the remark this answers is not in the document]
mattzeunert: > just the individual line visibility and the expand/collapse button text?

Yeah. We could have specific classes for each one like you suggested.

I went through 3 versions for the renderer logic:

1) Re-render on expand click
2) Pre-render two separate listings upfront with one hidden, then show/hide one on expand click
3) Pre-render a single listing for both states and toggle with the `show-if-expanded/collapsed` classes

The current solution (3) makes toggling easy, but the rendering logic is more complex because we need to track state for both the collapsed and expanded version. For (1) and (2) the code is simpler because we can focus on rendering one listing at a time.

Do you think maybe we should use (2) to simplify the code a bit? Only downside would be having more elements in the DOM.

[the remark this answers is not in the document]
patrickhulce: I actually really like that you've done 3! I personally think it's very workable with a few other tweaks mentioned below, so I'd hold off on any major changes for now

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+
+    const {codeSnippetCollapse, codeSnippetExpand} = Util.UIStrings;
+    dom.find('.' + SHOW_IF_EXPANDED_CLASS, header).textContent = codeSnippetCollapse;
+    dom.find('.' + SHOW_IF_COLLAPSED_CLASS, header).textContent = codeSnippetExpand;
+
+    const toggleShowAllButton = dom.find('.lh-code-snippet__toggle-show-all', header);
+    if (showAll) {
+      toggleShowAllButton.remove();
+    } else {
+      toggleShowAllButton.addEventListener('click', () => toggleExpandedFn());
```
brendankenny: What's the js restriction for the report in devtools again? (We should really write these down :)

I believe we currently don't run any after report generation that's not in devtools itself. Is that still a requirement? If so, we'll either need a decent way to display this statically or land this interaction code in devtools as well.


@paulirish

paulirish: > What's the js restriction for the report in devtools again?

There is none, these days. :)  (But there used to be.)

And yes, as of today, only interaction JS is defined in the DevTools files, but there's nothing preventing us from registering event handlers in our renderers. Seems cool.

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+const SHOW_IF_EXPANDED_CLASS = 'lh-code-snippet__show-if-expanded';
+const SHOW_IF_COLLAPSED_CLASS = 'lh-code-snippet__show-if-collapsed';
+
+/**
+ * @param {LH.Audit.DetailsRendererCodeSnippetLine[]} lines
+ * @param {number} lineNumber
+ */
+function getLineAndPreviousLine(lines, lineNumber) {
+  return {
+    line: lines.find(l => l.number === lineNumber),
```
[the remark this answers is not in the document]
mattzeunert: Yeah!

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,260 @@
...
+    const showAll = lineCount <= 4;
+
+    const header = dom.cloneTemplate('#tmpl-lh-code-snippet__header', tmpl);
+    dom.find('.lh-code-snippet__title', header).textContent = title;
+
+    const {codeSnippetCollapse, codeSnippetExpand} = Util.UIStrings;
+    dom.find('.' + SHOW_IF_EXPANDED_CLASS, header).textContent = codeSnippetCollapse;
+    dom.find('.' + SHOW_IF_COLLAPSED_CLASS, header).textContent = codeSnippetExpand;
+
+    const toggleShowAllButton = dom.find('.lh-code-snippet__toggle-show-all', header);
```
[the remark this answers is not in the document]
mattzeunert: Renamed to `toggleExpandButton` 👍

lighthouse-core/report/html/renderer/code-snippet-renderer.js
mattzeunert: I really like this suggestion! So much cleaner 🙂

lighthouse-core/report/html/renderer/code-snippet-renderer.js
mattzeunert: Hmm, that "because" comment doesn't make a lot of sense. Updated it like this:

    // Keep track of whether a highlighted line has been seen, because we don't want this code
    // to render an omitted lines indication before the first line

Rendering the first and last omitted line indicators would fit here too in terms of code structure, although I don't think it would make the code simpler.

Wondering if there's a good way to clean up `renderSnippetLines` more.

lighthouse-core/report/html/renderer/code-snippet-renderer.js
mattzeunert: Yeah, we skip the lines where we don't know the content.

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: I don't know what a forward looking nested details type is, but would this also affect the `DetailsRendererNodeDetailsJSON` in `DetailsRendererCodeSnippet.node`?

Is the list any different from table, or does table have some special treatment?

mattzeunert:
We'd like the error message text to always wrap within the visible area,  so you can read it without scrolling. Like this:

screenshot 2019-01-23 at 20 31 20

But right now it doesn't work if the lines themselves are scrollable:

screenshot 2019-01-23 at 20 31 28

@rviscomi and I are ok with it, since having error messages that long is pretty rare. But let me know if you know some magic CSS trick that would fix this 🙂

[This is the live example.](https://jsonld-display.glitch.me/#seo)

lighthouse-core/report/html/renderer/details-renderer.js
exterkamp: I'm not 100% what @patrickhulce is implying with this?  Nor can I see exactly what the format is?  Is it a details element that basically contains more details elements? like:
```javascript
{
   details: {
      items: [
         details,
         details,
         details...
         ],
}
```
?

mattzeunert: Yeah exactly, we have a `list` type that's very similar to `table`. In this case the `list` contains a `code-snippet` which contains a `node`. Not sure why this might cause problems.

exterkamp: This isn't a blocker from a proto perspective, it will run fine.  Buuuuuut this level of nesting is _uncomfortable_ since our protos + the struct syntax are already highly nested.  How deep will this go?  Is there an upper limit?  Does it nest more than once?  

This nesting in the details (here and in tables) from a proto perspective is very bad and should be represented more flatly and predictably, but it is not a blocker and since we already do it for tables I think this is _okay_ as long as the nesting is not overly deep.

exterkamp: Although if the list contains `code-snippet` objects we should call that out in the types instead of just throwing in more details objects as a proxy.

mattzeunert:
@patrickhulce What do you think of the PR now?

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,283 @@
...
+    const hasOnlyNonLineSpecficHighlights = highlights.length > 0 && !hasLineSpecificHighlights;
+
+    const lineContainer = dom.createFragment();
+
+    // Keep track of whether a highlighted line has been seen, because we don't want this code
+    // to render an omitted lines indication before the first line
+    let hasSeenHighlight = false;
+    for (let lineNumber = 1; lineNumber <= lineCount; lineNumber++) {
+      const {line, previousLine} = getLineAndPreviousLine(lines, lineNumber);
+      const {line: collapsedLine, previousLine: collapsedPreviousLine} =
```
patrickhulce: WDYT about some comments or more descriptive names here?

is this right?

- `!previousLine` is just telling us if this current line is the start of a visible block and we need to render the omitted lines placeholder for the stuff that came before?
- `collapsedLine` is just telling us if this current line should be shown even when the snippet is collapsed?
- `!collapsedPreviousLine` is just telling us if this current line is the start of a visible block and we need to render the omitted lines placeholder for the stuff that came before?

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,283 @@
...
+        getLineAndPreviousLine(collapsedLines, lineNumber);
+
+      if (hasSeenHighlight) {
+        // Show if some lines were omitted
+        if (line && !previousLine) {
```
patrickhulce: is it true that `collapsedLine` can never be truthy if `line` is not truthy? 
is it the same for `previousLine` and `collapsedPreviousLine`?

what I'm getting at here is that it feels like we have 2 cases of omitted line placeholders not 3, either this line is always shown as omitted or it's only shown as omitted when collapsed.

lighthouse-core/report/html/renderer/details-renderer.js
patrickhulce: Yeah that's sensible @mattzeunert!

I'm not entirely convinced we need a full-blown generic list details that people accidentally start using when we don't really want them to :)

I think I'd be a bit more comfortable with a code-snippet-list details type that doesn't do the nesting, but I don't feel *too* strongly.

lighthouse-core/test/audits/audit-test.js
```diff
@@ -99,4 +99,75 @@ describe('Audit', () => {
...
+    });
+
+    it('Limits the number of lines around highlights', () => {
+      const code = makeLines(99) + 'A\n' + makeLines(99) + '\nB';
+      const allLines = code.split('\n');
+      const details = Audit.makeCodeSnippetDetails({
+        code,
+        title: 'Title',
+        highlights: [{
+          lineNumber: allLines.findIndex(l => l === 'A') + 1,
```
patrickhulce: seems like we could probably just put `100` here :)

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,283 @@
...
+    });
+  }
+
+  /**
+   * @param {DOM} dom
+   * @param {DocumentFragment} tmpl
+   * @param {LH.Audit.DetailsRendererCodeSnippet} details
+   * @return {DocumentFragment}
+   */
+  static renderSnippet(dom, tmpl, details) {
```
patrickhulce: I think some explanation of the overall structure would help a bit here.

something like the below if it's accurate? please feel free to adjust to what's happening :)

- First we add the highlights that relate to the whole snippet and don't reference particular lines
- Then we'll render the lines with each of their line-specific highlights and omitted segments
- Finally we'll make sure we render the omitted lines placeholder for the beginning and end

lighthouse-core/report/html/renderer/code-snippet-renderer.js
```diff
@@ -0,0 +1,283 @@
...
+    const snippet = dom.find('.lh-code-snippet__snippet-inner', template);
+
+    const nonLineSpecificHighlights = highlights.filter(h => typeof h.lineNumber !== 'number');
+    nonLineSpecificHighlights.forEach(h =>
+      snippet.append(CodeSnippetRenderer.renderHighlightMessage(dom, tmpl, h))
+    );
+
+    snippet.append(CodeSnippetRenderer.renderSnippetLines(dom, tmpl, details));
+
+    // If expanded view still doesn't include all lines then indicate that that
```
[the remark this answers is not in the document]
patrickhulce: also what about the collapsed view?

I'm very shaky on all the omitted lines placeholder logic, why wouldn't the other chunk of logic handle this case already?

if `lines[0].number !== 1` then wouldn't `line && !previousLine` be true in `renderSnippetLines` for `lines[0]`?

lighthouse-core/audits/audit.js
```diff
@@ -125,6 +125,70 @@ class Audit {
...
+   * node?: LH.Audit.DetailsRendererNodeDetailsJSON;
+   * maxLineLength?: number;
+   * maxLinesAroundHighlight?: number;
+   * }} CodeSnippetInfo */
+  /**
+   * @param {CodeSnippetInfo} opts
+   * @return {LH.Audit.DetailsRendererCodeSnippet}
+   */
+  static makeCodeSnippetDetails({
+    code,
```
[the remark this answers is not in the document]
Hoten: content makes sense, but only if we change it to "content-snippet". Which might be better. can you call JSON code?

lighthouse-core/test/audits/audit-test.js
Hoten: I think this is more clear about what's being tested. `allLines.findIndex(...` -> "the line with A on it" is more immediately understandable than "100" -> "the line with A on it".

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: I think the main reason for the nesting is that we might want to put a code snippet inside a table at some point.

But we don't have any concrete plans and I'm not sure how we'd make that look good anyway, so I think having a `code-snippet-list`/`content-snippet-list`/... is fine.

lighthouse-core/report/html/renderer/util.js
```diff
@@ -449,6 +449,50 @@ class Util {
...
+   * @param {LH.Audit.DetailsRendererCodeSnippetLine[]} lines
+   * @param {LH.Audit.DetailsRendererCodeSnippetHighlight[]} highlights
+   * @param {number} surroundingLineCount Number of lines to include before and after
+   * the highlight. If this is e.g. 2 this function might return 5 lines.
+   */
+  static filterRelevantLines(lines, highlights, surroundingLineCount) {
+    const lineSpecificHighlights = highlights.filter(h => typeof h.lineNumber === 'number');
+    if (lineSpecificHighlights.length === 0) {
+      // no specific line highlighted, just return the first bunch of lines
+      return lines.slice(0, surroundingLineCount * 2 + 1);
```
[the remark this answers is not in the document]
mattzeunert: We don't know which lines are relevant to the message ,if there is one, but a few arguments for using the first few lines:

- one goal of the snippet is to help users identify what JSON-LD element they're looking at, and having more consecutive lines makes that easier
- the last visible line will often just be `}`, so it provides less information than extra lines at the top
- we avoid visual clutter by not needing to render the "..."

lighthouse-core/report/html/renderer/code-snippet-renderer.js
mattzeunert: Here's an example, (X) means only in expanded lines array, `-` means the line is not in either array

```
 A 
(B) => should insert "..." before this, but only if collapsed because
            there are no missing lines before C
 C
(D) => hmm, actually this is a bit trickier... when we want to show "..."
            for both we want to show it after the last line that's visible in
            the expanded state, so we need to look forward when the 
            line is missing in linesVisibleWhenCollapsed, and backward
            if the line is missing in all available lines
 -
(K)
 L
(M)
```

mattzeunert: I guess it can all just be handled by setting flags, so it's less complex than I thought.

lighthouse-core/report/html/renderer/util.js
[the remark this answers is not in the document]
Hoten: Those points make sense for JSON content, but that won't be the only content to use these snippets. I'm wondering if the first + last few lines would be preferable for content like CSS or JS or HTML. But we can consider that more when we add uses of those contents, and possible add a new mode controlling default behavior (when no specific highlights are present).

lighthouse-core/report/html/renderer/code-snippet-renderer.js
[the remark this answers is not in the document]
mattzeunert: > also what about the collapsed view?

We don't show it there, but maybe we should. The reason against showing it is that it makes the UI look more cluttered. Usually it's clear from the code that it's cut off and you have the "Expand snippet" button as well.

screenshot 2019-02-02 at 07 56 26

lighthouse-core/report/html/renderer/util.js
```diff
@@ -449,6 +449,50 @@ class Util {
...
+        // make sure we still show (surroundingLineCount * 2 + 1) lines in total
+        firstSurroundingLineNumber++;
+        lastSurroundingLineNumber++;
+      }
+      for (let i = firstSurroundingLineNumber; i <= lastSurroundingLineNumber; i++) {
+        const surroundingLineNumber = i;
+        lineNumbersToKeep.add(surroundingLineNumber);
+      }
+    });
+
```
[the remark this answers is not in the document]
mattzeunert: Done for x=3 now.

lighthouse-core/report/html/renderer/code-snippet-renderer.js
mattzeunert: I've cleaned up the rendering logic a bit more, though it's still not exactly nice code 🤔

mattzeunert:
> for the amount of complexity that's here I'm hoping for a bit broader test coverage if possible. Especially around some of the cases I'm talking about in the code where I'm confused :)

I think the cases where you're confused about line rendering all already have test cases 😂 Maybe you meant something else, or I need make the test names more descriptive?

mattzeunert:
> FAIL  ./dist/viewer/src/viewer.js: 65.11kB > maxSize 65kB gzip 

So sad 😞

lighthouse-core/report/html/renderer/snippet-list-renderer.js
```diff
@@ -0,0 +1,307 @@
...
+ * @param {number} lineNumber
+ */
+function getLineAndPreviousLine(lines, lineNumber) {
+  return {
+    line: lines.find(l => l.number === lineNumber),
+    previousLine: lines.find(l => l.number === lineNumber - 1),
+  };
+}
+
+class SnippetListRenderer {
```
brendankenny: I feel like we need an overview for this type/renderer somewhere, though I'm not sure exactly where. Basically the only way to understand how it works right now is to read this file :)

In this file? On the details type in `audit.d.ts`?

lighthouse-core/test/report/html/renderer/snippet-list-renderer-test.js
```diff
@@ -0,0 +1,163 @@
...
+    dom = new DOM(document);
+  });
+
+  afterAll(() => {
+    global.Util = undefined;
+  });
+
+  function renderSnippet(highlights, lineRanges = undefined) {
+    const details = makeDetails(highlights, lineRanges);
+    const el = SnippetListRenderer.render(dom, dom.document(), details, {});
```
brendankenny: no fourth argument?

lighthouse-core/test/report/html/renderer/snippet-list-renderer-test.js
```diff
@@ -0,0 +1,163 @@
...
+    const {highlightLines} = renderSnippet([{
+      message: 'General error',
+    }, {
+      lineNumber: 5,
+      message: 'Error on line',
+    }]);
+
+    assert.equal(highlightLines.length, 2);
+  });
+});
```
brendankenny: seems like a few more tests are need. The list as a whole (I think), header, toggling collapsed state (I think you can `.click()` in jsdom), `truncated`...any other states?

Can wait on some of these until the rest of the PR is settled so you don't have to keep changing them

types/audit.d.ts
```diff
@@ -76,24 +76,49 @@ declare global {
...
+    export interface DetailsRendererSnippetHighlight {
+      /** Line number, starting from 1 */
+      lineNumber: number | null,
+      message: string
+    }
+
+    export interface DetailsRendererSnippetLine {
+      content: string
+      /** Line number, starting from 1 */
+      number: number
```
brendankenny: number vs lineNumber?

lighthouse-core/audits/audit.js
```diff
@@ -125,6 +125,62 @@ class Audit {
...
+
+  /**
+   * @param {string} content
+   * @param {number} maxLineLength
+   * @returns {LH.Audit.DetailsRendererSnippetLine[]}
+   */
+  static _makeSnippetLinesArray(content, maxLineLength) {
+    return content.split('\n').map((line, lineIndex) => {
+      const lineNumber = lineIndex + 1;
+      /** @type LH.Audit.DetailsRendererSnippetLine */
```
brendankenny: seems like this isn't needed?

brendankenny:
> `FAIL ./dist/viewer/src/viewer.js: 65.11kB > maxSize 65kB gzip`

This might be unavoidable because right now the viewer is like 60K after gzip and it includes all report renderer code twice (once for rendering, once for self replication), so a 11K+ new file would be expected to run right into that. There may be no choice but to bump the limit.

lighthouse-core/test/report/html/renderer/snippet-list-renderer-test.js
mattzeunert: Fixed, now takes four args.

lighthouse-core/audits/audit.js
mattzeunert: Without this TS complains about adding the `truncated` property.

lighthouse-core/report/html/renderer/util.js
```diff
@@ -449,6 +449,51 @@ class Util {
...
+        return;
+      }
+      let firstSurroundingLineNumber = lineNumber - surroundingLineCount;
+      let lastSurroundingLineNumber = lineNumber + surroundingLineCount;
+      while (firstSurroundingLineNumber < 1) {
+        // make sure we still show (surroundingLineCount * 2 + 1) lines in total
+        firstSurroundingLineNumber++;
+        lastSurroundingLineNumber++;
+      }
+      // If we only a few lines would be omitted normally then we prefer to include
```
Hoten: "If we only" -> "If only"

lighthouse-core/report/html/renderer/snippet-list-renderer.js
```diff
@@ -0,0 +1,307 @@
...
+    // If we leave the button in though, wire up the click listener to toggle visibility!
+    if (showAll) {
+      toggleExpandButton.remove();
+    } else {
+      toggleExpandButton.addEventListener('click', () => toggleExpandedFn());
+    }
+
+    const nodeContainer = dom.find('.lh-snippet__node', header);
+    if (details.node && dom.isDevTools()) {
+      nodeContainer.appendChild(detailsRenderer.renderNode(details.node));
```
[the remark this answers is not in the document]
mattzeunert: Hmm, I don't actually... can I try it without building Chrome locally?

The screenshot at the top of the PR actually includes the element, but in DevTools they'll be converted to clickable links to the Elements panel.

lighthouse-core/report/html/renderer/snippet-list-renderer.js
```diff
@@ -0,0 +1,307 @@
...
+   * @param {LH.Audit.DetailsRendererSnippet} details
+   * @return {DocumentFragment}
+   */
+  static renderSnippetContent(dom, tmpl, details) {
+    const {highlights} = details;
+
+    const template = dom.cloneTemplate('#tmpl-lh-snippet__content', tmpl);
+    const snippet = dom.find('.lh-snippet__snippet-inner', template);
+
+    const nonLineSpecificHighlights = highlights.filter(h => typeof h.lineNumber !== 'number');
```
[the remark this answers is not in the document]
mattzeunert: Doesn't seem like moving the highlights onto lines is a great improvement:
- `highlights` are what we get from the audit, so we'd have to add them to the lines
- to check if we have highlights we now need to iterate through all lines
- Assuming we still want to allow multiple highlight messages per line and might want to add properties like `type: "error"` or `type: "warning"` in the future we'd need a nested type like `lines[0].highlights[0].text`, rather than just a `higlightMessage` string property

What do you think about splitting highlights into `lineHighlights` and `generalHighlights`?

Or maybe `lineMessages` and `generalMessages` would be better?

brendankenny:
having spent some time back in `details` land looking at basically all of our existing audits, I'd like to propose splitting this (back?) into two details types.

I've noticed that we already have several audits using single-column tables to get an ugly list, and switching them to a dedicated `list` type (sometime in the future) would be great, so let's leave that path open to them. So:

```ts
export interface DetailsRendererList {
  type: 'list',
  items: DetailsRendererSnippet[]
} 
```

aka `items` still has to be an array of only snippets, but that will allow us to easily expand what's allowed in `items` in the future. THen

- move `snippet-list-renderer.js` to `snippet-renderer.js`, and move `SnippetListRenderer.render()` into `details-renderer.js` as `DetailsRenderer._renderList()`
- `SnippetListRenderer.renderSnippet()` moves to `SnippetRenderer.render()`, and the new `DetailsRenderer._renderList()` would then call that (fine to leave `_renderList()` specialized as it is because we only support rendering snippets in a list for now.
- add `type: 'snippet'` to the `DetailsRendererSnippet` interface

I don't think this will be much of a code change, but it will leave us with plenty of flexibility.

WDYT?

lighthouse-core/report/html/renderer/snippet-list-renderer.js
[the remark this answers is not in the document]
brendankenny: > Doesn't seem like moving the highlights onto lines is a great improvement

yeah, all your points are good ones

> What do you think about splitting highlights into `lineHighlights` and `generalHighlights`?

sounds good to me, if that seems like a win (I'm fine with either naming scheme)

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: Split back into `snippet` and `list` since @brendankenny found a few of other places where we'd want to use a list.

lighthouse-core/report/html/renderer/snippet-list-renderer.js
```diff
@@ -0,0 +1,307 @@
...
+  };
+}
+
+class SnippetListRenderer {
+  /**
+   * @param {DOM} dom
+   * @param {DocumentFragment} tmpl
+   * @param {LH.Audit.DetailsRendererSnippet} details
+   * @param {DetailsRenderer} detailsRenderer
+   * @param {function} toggleExpandedFn
```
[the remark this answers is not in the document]
mattzeunert: We could also add the event listener to the `lh-snippet` container and then check `e.target`, but that seems worse than what we have now.

lighthouse-core/test/report/html/renderer/snippet-list-renderer-test.js
```diff
@@ -0,0 +1,163 @@
...
+    require('../../../../report/html/renderer/snippet-list-renderer.js');
+
+const TEMPLATE_FILE = fs.readFileSync(__dirname +
+    '/../../../../report/html/templates.html', 'utf8');
+
+/**
+ * @param {*} highlights
+ * @returns {LH.Audit.DetailsRendererSnippet}
+ */
+function makeDetails(highlights, lineRanges = [{from: 1, to: 6}]) {
```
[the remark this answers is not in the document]
mattzeunert: Have tried to make it more clear what each test case is testing 👍

mattzeunert:
Have separated `snippet` and `list` into separate details renderer types.

Also, `highlights` is now `lineMessages` and `generalMessages`.

```
export interface List {
  type: "list";
  items: Snippet[];
}

export interface Snippet {
  type: "snippet";
  lines: {
    content: string;
    lineNumber: number;
    truncated?: boolean;
  }[];
  title: string;
  lineMessages: {
    lineNumber: number;
    message: string;
  }[];
  generalMessages: {
    message: string;
  }[];
  lineCount: number;
  node?: NodeValue;
}
```

lighthouse-core/report/html/renderer/snippet-list-renderer.js
mattzeunert: Added some comments in `audit-details.d.ts`.

lighthouse-core/report/html/renderer/snippet-list-renderer.js
```diff
@@ -0,0 +1,307 @@
...
+  static renderHeader(dom, tmpl, details, detailsRenderer, toggleExpandedFn) {
+    const {lineCount, title} = details;
+    const showAll = lineCount <= 4;
+
+    const header = dom.cloneTemplate('#tmpl-lh-snippet__header', tmpl);
+    dom.find('.lh-snippet__title', header).textContent = title;
+
+    const {snippetCollapseBtnLabel, snippetExpandBtnLabel} = Util.UIStrings;
+    dom.find('.lh-snippet__btn-label-collapse', header).textContent = snippetCollapseBtnLabel;
+    dom.find('.lh-snippet__btn-label-expand', header).textContent = snippetExpandBtnLabel;
```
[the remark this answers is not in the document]
mattzeunert: Maybe `data-ui-string="snippetCollapseBtnLabel"`?

lighthouse-core/report/html/renderer/details-renderer.js
patrickhulce: :( I'm sorry @mattzeunert! mind linking to the convo I missed?

mattzeunert: https://github.com/GoogleChrome/lighthouse/pull/6999#issuecomment-461657358

danbri:
How does it get its knowledge of what terms and structures are in Schema.org? e.g. the "unexpected property" in the screenshot at the top of this PR. /cc @vholland @rvguha

paulirish:
Dan, see https://github.com/GoogleChrome/lighthouse/pull/6750

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -68,6 +68,15 @@ class DetailsRenderer {
       case 'table':
         // @ts-ignore - TODO(bckenny): Fix type hierarchy
         return this._renderTable(/** @type {TableDetailsJSON} */ (details));
+      case 'list':
+        return this._renderList(
+          // @ts-ignore
```
patrickhulce: is this the same `TODO:` reason as the above?

lighthouse-core/report/html/renderer/snippet-renderer.js
```diff
@@ -0,0 +1,364 @@
...
+
+/** @typedef {{
+    content: string;
+    lineNumber: string | number;
+    contentType: LineContentType;
+    truncated?: boolean;
+    visibility?: LineVisibility;
+}} LineDetails */
+
+const classNameByContentType = {
```
patrickhulce: nit: `classNamesByContentType` or something to signify it's an array when being used?

lighthouse-core/report/html/renderer/snippet-renderer.js
```diff
@@ -0,0 +1,364 @@
...
+    if (details.node && dom.isDevTools()) {
+      const nodeContainer = dom.find('.lh-snippet__node', header);
+      nodeContainer.appendChild(detailsRenderer.renderNode(details.node));
+    }
+
+    return header;
+  }
+
+  /**
+   * Renders a line of DOM content (text content, message, or empty line)
```
patrickhulce: `DOM content` confuses me a bit since we're using `content` to refer to a particular type of line (which I love!), maybe `Renders a line (text content, message, or empty line) as a DOM element.`?

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: Yes, added that. I think @brendankenny [has fixed these todos in this PR](https://github.com/GoogleChrome/lighthouse/pull/7192/files).

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -68,6 +68,15 @@ class DetailsRenderer {
       case 'table':
         // @ts-ignore - TODO(bckenny): Fix type hierarchy
         return this._renderTable(/** @type {TableDetailsJSON} */ (details));
+      case 'list':
+        return this._renderList(
+          // @ts-ignore - TODO(bckenny): Fix type hierarchy
+          /** @type {LH.Audit.Details.List} */ (details)
+        );
+      case 'snippet':
```
brendankenny: let's not add `snippet` here (it's just going to be immediately removed anyways)

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -210,6 +219,21 @@ class DetailsRenderer {
...
+  /**
+   * @param {LH.Audit.Details.List} details
+   * @returns {Element}
+   */
+  _renderList(details) {
+    const listContainer = this._dom.createElement('div', 'lh-list');
+
+    details.items.forEach(item => {
+      // @ts-ignore TODO(bckenny): this can never be null
+      listContainer.appendChild(this.render(item));
```
brendankenny: rather than call `this.render(item)`, just call `SnippetRenderer.render()` directly (then also don't have to deal with `null`)

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -455,6 +479,12 @@ if (typeof module !== 'undefined' && module.exports) {
   }} TableDetailsJSON
  */
 
+/** @typedef {{
+      type: string,
+      items: Array<DetailsJSON>
+  }} ListDetailsJSON
+ */
```
brendankenny: I think this can be deleted?

types/audit-details.d.ts
```diff
@@ -164,6 +169,38 @@ declare global {
         type: 'url';
         value: string;
       }
+
+      /**
+       * Snippet of text with line numbers and annotations.
+       */
+      export interface Snippet {
+        type: "snippet",
```
brendankenny:
```suggestion
        type: 'snippet',
```

types/audit-details.d.ts
```diff
@@ -164,6 +169,38 @@ declare global {
         type: 'url';
         value: string;
       }
+
+      /**
+       * Snippet of text with line numbers and annotations.
+       */
+      export interface Snippet {
```
brendankenny: `SnippetValue` to keep in line with the others, but otherwise this looks great!

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: Yup!

Issue #4093, opened 2017-12-19
Enable categories other than a11y to have audit groups


rviscomi:
Right now only the Accessibility audit category can have groups of audits. Make this feature reusable so other categories like SEO can have groups.

Context: https://github.com/GoogleChrome/lighthouse/pull/4057#issuecomment-351805856

paulirish:
this'll probably mean a decent refactor within report-renderer. 
done generally, it probably means that we enable the `notApplicable` group for all other categories too.

seems good!

kdzwinel:
@paulirish I'm merging a11y category renderer with the default one. a11y renderer doesn't have the 'X Failed Audits' header. Should we:

1. have it in all audit categories,
1. remove it from all audit categories,
1. have it only for grouped audit categories (as it is now)

?

IMO it's redundant and we could skip it everywhere for consistency and to save some space. E.g. PWA section with an w/o the header:

<img width="1231" alt="screen shot 2018-01-18 at 00 59 21" src="https://user-images.githubusercontent.com/985504/35073901-78fa0da4-fbeb-11e7-8d9f-d0347e00927b.png">

paulirish:
Yeah I like nuking the heading (and indentation). 

@vinamratasingal does that work for you?

vinamratasingal:
I actually like keeping the failed audits section- it creates consistency between the other sections. Paul/Konrad- any reasons why you like nuking the header?

cc: @hwikyounglee in case she wants to add her perspective :)

hwikyounglee:
Header gives a visual help to group/separate things. I suggest we keep it.

kdzwinel:
IMO failed audits are clearly marked (by the description, ❌ and, in many cases, red debugString) and don't require any more introduction. Removing 'X Failed Audits' header saves us some horizontal space that, especially inside of DevTools, is very precious.

If we decide to keep them, we should probably still deal with the inconsistency between sections that use groups ('Accessibility', 'SEO') and sections that don't use them ('PWA', 'Best Practices'). Should we add headers to all of them?

vinamratasingal:
I understand that the failed audits have the X, but I still think for consistency/visual help perspective, we should keep it in there. I suggest we push this paradigm across all our sections. 

cc @paulirish @hwikyounglee

kdzwinel:
OK, I see your point👍 I made a quick change to show how the accessibility section will look like with this header:

[...]

PR #7006, opened 2019-01-14
report: restore old, disabled failed grouping test

brendankenny:
This was an old test we made the test runner skip because of changing architecture at one point, but it turns out we don't really have another test of the grouping of failing audits, which is kind of important. This restores the test :)

TMI:
I mentioned it in https://github.com/GoogleChrome/lighthouse/issues/5327#issuecomment-429466664, linked to making sure audits with warnings are shown; which is kind of weird given the test contents. I *think* the connection (in my head?) was that we previously marked audits with warnings as failures to make sure they showed up by default, but that all changed when we added groups for all categories in #4278. 

However [the conversation at the time](https://github.com/GoogleChrome/lighthouse/issues/4093#issuecomment-358491941) was about the "X failed audits" header, which might be the "header" mentioned in the TODO. But that would have required a very simple adjustment of the selector to fix.

Lessons:
- don't leave `it.skip()` tests for very long because you're going to forget why you skipped
- make sure to document the results of "discussing in the next meeting" in the PR/issue because otherwise all reasoning will be lost in the mists of time ⏳💨

### DIFF
--- a/lighthouse-core/test/report/html/renderer/category-renderer-test.js
+++ b/lighthouse-core/test/report/html/renderer/category-renderer-test.js
@@ -215,16 +215,36 @@ describe('CategoryRenderer', () => {
       assert.ok(description.querySelector('a'), 'description contains converted markdown links');
     });
 
-    // TODO waiting for decision regarding this header
-    it.skip('renders the failed audits grouped by group', () => {
-      const categoryDOM = renderer.render(category, sampleResults.categoryGroups);
-      const failedAudits = category.auditRefs.filter(audit => {
-        return audit.result.score !== 1 && !audit.result.scoreDisplayMode === 'notApplicable';
+    it('renders the failed audits grouped by group', () => {
+      // Fail all the audits.
+      const categoryClone = JSON.parse(JSON.stringify(category));
+      const auditRefs = categoryClone.auditRefs;
+      auditRefs.forEach(ref => {
+        ref.result.score = 0;
+        ref.result.scoreDisplayMode = 'binary';
       });
-      const failedAuditTags = new Set(failedAudits.map(audit => audit.group));
+      const categoryDOM = renderer.render(categoryClone, sampleResults.categoryGroups);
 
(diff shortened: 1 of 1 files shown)


lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -215,16 +215,36 @@ describe('CategoryRenderer', () => {
...
+      auditRefs.forEach(ref => {
+        ref.result.score = 0;
+        ref.result.scoreDisplayMode = 'binary';
       });
-      const failedAuditTags = new Set(failedAudits.map(audit => audit.group));
+      const categoryDOM = renderer.render(categoryClone, sampleResults.categoryGroups);
+
+      // All the group names in the config.
+      const groupNames = Array.from(new Set(auditRefs.map(ref => ref.group))).filter(Boolean);
+      assert.ok(groupNames.length > 5); // Make sure there are groups to test.
```
patrickhulce: we could turn the comment into an assertion message just to make things easier when it fails. though jest comes to our aid here pretty nicely already :)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -215,16 +215,36 @@ describe('CategoryRenderer', () => {
...
-      assert.equal(failedAuditGroups.length, failedAuditTags.size);
+      // All the group roots in the DOM.
+      const failedGroupElems = Array.from(
+          categoryDOM.querySelectorAll('.lh-clump--failed > .lh-audit-group'));
+
+      assert.strictEqual(failedGroupElems.length, groupNames.length);
+
+      for (const groupName of groupNames) {
+        const groupAuditRefs = auditRefs.filter(ref => ref.group === groupName);
+        assert.ok(groupAuditRefs.length > 0); // Make sure there are audits to find.
```
patrickhulce: same

brendankenny:
good call

PR #7008, opened 2019-01-14
i18n: import translated accessibility and lh-error messages

exterkamp:
**Summary**
Importing new i18n messages for a11y and lh-errors!


paulirish:
Let's land after 4.0 is cut.

exterkamp:
Yes, imported them :smile:

PR #7024, opened 2019-01-16
4.0.0

brendankenny:


lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js
```diff
@@ -17,7 +17,7 @@ module.exports = [
         score: 0,
         details: {
           items: {
-            length: 6,
+            length: 7,
```
paulirish: was ist das?

brendankenny: > was ist das?

 whoops, was testing :)

Issue #6747, opened 2018-12-07
Gatherer/artifact cleanup


paulirish:
In a plugin world, the gatherers become part of our public API. We know about a few things we're interested in changing. Let's capture them here.

* [x] add a new `MetaElements` gatherer which gets name/content of all of them. This would nullify the following: `ThemeColor`, `Viewport`, `MetaRobots`, `MetaDescription`, (#7025)
* [x] add a new `LinkElements` gatherer (..actually introduced in #6694 as `ExternalResourceLinks`). This could nullify `Canonical` and `Hreflang` (#7080)
* [x] add a new `AnchorElements` gatherer which gets rel/href/target/text of all `a` elements, nullifies `AnchorsWithNoRelNoopener` and `CrawlableLinks` (#7101)
* [x] convert `ImageUsage` into `ImageElements` gatherer (#7030)
* [x] convert `Scripts` into `ScriptElements` gatherer (#7920)
* [ ] ...

~~@patrickhulce you had a few in mind, too, right?~~

patrickhulce:
`meta` and `links` were the two big ones 👍 I also wanted to cleanup

- `Images` into `ImageElements`
- `AnchorElements`

patrickhulce:
I like the standardization on `*Elements` I'll update #6694, my only concern with `LinkElements` was that it would be easily confused with `a` elements :) I think it makes sense though when they'll be side-by-side

kusler:
Is this issue comprehensive with what we're wanting to accomplish with artifact cleanup, or are there more that need to be tended to? 

If it's not the full list, then can Connor take on fleshing it out?

patrickhulce:
I think the primary goal is getting our artifacts in a state ready and useful for plugins. If someone wants to do some brainstorming what audits might need to know about a page and how we should standardize it, that'd be super useful :)

patrickhulce:
OK decision time and input needed! Over in https://github.com/GoogleChrome/lighthouse/pull/8133#discussion_r274014854 we were discussing exposing the DevTools-style child-index path on the `ScriptElements` artifact. This brings up a host of discussions about our public API for artifacts in v5.


My proposal is **we only expose `*Elements` artifacts and the base artifacts to plugins by default**. This will give plugins access to network records and the primary elements on the page which should take you pretty far. Everything else we make opt-in via `dangerouslyUseExperimentalArtifacts` or an `Experimental` prefix in the artifact name or something.

[...]

PR #7025, opened 2019-01-16
core(meta): refactor all meta audits to single artifact

patrickhulce:
**Summary**
Replace all our scattered meta gatherers with a single `MetaElements` artifact.

**Related Issues/PRs**
#6747


lighthouse-core/audits/seo/font-size.js
```diff
@@ -199,7 +199,7 @@ class FontSize extends Audit {
       title: str_(UIStrings.title),
       failureTitle: str_(UIStrings.failureTitle),
       description: str_(UIStrings.description),
-      requiredArtifacts: ['FontSize', 'URL', 'Viewport'],
+      requiredArtifacts: ['FontSize', 'URL', 'MetaElements'],
```
patrickhulce: no changes to functionality needed here because it just calls out to the viewport audit

types/artifacts.d.ts
```diff
@@ -89,10 +89,8 @@ declare global {
       JsUsage: Crdp.Profiler.ScriptCoverage[];
       /** Parsed version of the page's Web App Manifest, or null if none found. */
       Manifest: Artifacts.Manifest | null;
-      /** The value of the <meta name="description">'s content attribute, or null. */
-      MetaDescription: string|null;
-      /** The value of the <meta name="robots">'s content attribute, or null. */
-      MetaRobots: string|null;
+      /** The value of the <meta>elements in the head. */
```
brendankenny: values

lighthouse-cli/test/fixtures/seo/seo-failure-cases.html
```diff
@@ -12,7 +12,7 @@
   <!-- FAIL(font-size): missing viewport -->
   <meta name="viewport" content="invalid-content=should_have_looked_it_up">
   <!-- no <meta name="description" content=""> -->
-  <meta name="robots" content="nofollow, NOINDEX, all">
+  <meta name="RoBots" content="nofollow, NOINDEX, all">
```
brendankenny: document the reason for this?

lighthouse-core/audits/themed-omnibox.js
```diff
@@ -43,13 +43,13 @@ class ThemedOmnibox extends MultiCheckAudit {
   }
 
   /**
-   * @param {LH.Artifacts['ThemeColor']} themeColorMeta
+   * @param {LH.Artifacts['MetaElements'][0]|undefined} themeColorMeta
```
brendankenny: should we make an alias of `LH.Artifacts.MetaElement` since we use this so often?

lighthouse-core/test/audits/seo/font-size-test.js
```diff
@@ -14,10 +14,12 @@ const validViewport = 'width=device-width';
 /* eslint-env jest */
 
 describe('SEO: Font size audit', () => {
+  const makeMetaElements = viewport => viewport ? [{name: 'viewport', content: viewport}] : [];
+
   it('fails when viewport is not set', () => {
     const artifacts = {
       URL,
-      Viewport: null,
+      MetaElements: makeMetaElements(null),
```
brendankenny: Is the conditional in `makeMetaElements` just for this case? `null` could just be `undefined` or this one could just be `[]` an no conditional needed :)

lighthouse-core/test/audits/seo/is-crawlable-test.js
```diff
@@ -78,7 +80,7 @@ describe('SEO: Is page crawlable audit', () => {
     const artifacts = {
       devtoolsLogs: {[IsCrawlableAudit.DEFAULT_PASS]: devtoolsLog},
       URL: {finalUrl},
-      MetaRobots: null,
+      MetaElements: makeMetaElements(null),
```
brendankenny: switch to no argument or just use `[]` and no function call?

lighthouse-core/test/config/config-test.js
```diff
@@ -94,7 +94,7 @@ describe('Config', () => {
     const configJson = {
       passes: [{
         passName: unlikelyPassName,
-        gatherers: ['viewport'],
+        gatherers: ['meta-elements'],
```
brendankenny: I apologize for the number of times I've used `viewport` as a dummy gatherer name :)

patrickhulce:
> maybe it's unavoidable, but are we setting ourselves up for ballooning the size of the artifacts? People put crazy things in meta tags sometimes.

FWIW, I'm not really worried about this. Of all the artifact size concerns, we know these artifacts have to be constrained to roughly the size of the page. trace/devtoolslog/screenshots are always going to dwarf this.

> We could have a list of allowed meta names and add the ones we need as we go. On the other hand, that would prevent plugins from being able to just use the meta tags they're interested in without getting changes landed in core.

This limitation would kinda defeat the purpose of having generic gatherers IMO, would it not? Seems like if we just limited to what we needed then it's no different from the position we were in before functionality-wise.

lighthouse-cli/test/fixtures/seo/seo-failure-cases.html
patrickhulce: done

lighthouse-core/audits/themed-omnibox.js
patrickhulce: this is the only usage of it, but sure :)

patrickhulce:
@brendankenny LGTY?

lighthouse-core/audits/themed-omnibox.js
brendankenny: > this is the only usage of it, but sure

ha, whoops

PR #7031, opened 2019-01-16
viewer: allow loading RunnerResult json

paulirish:
If you want to drop the full RunnerResult json into the viewer, it should work.

cc @benschwarz


Hoten:
typescript error:
```
lighthouse-viewer/app/src/lighthouse-report-viewer.js(141,19): error TS2339: Property 'lhr' does not exist on type 'ReportResult'.
```

PR #7032, opened 2019-01-16
misc: fix extension package path and viewer comments

brendankenny:
These came up in the process of pushing 4.0.

For the extension, using `archiver.glob` with an absolute path nests the extension directory within a recreation of the absolute path of directories leading up to it within the zip file. Kind of amazing that that still works as an extension, but weird to ship it that way. Switching to `archiver.directory()` allows `dist/extension/` to be the root of the zip file.

The viewer comment changes are just to assure the next release that both 3x and 4x results are handled by the current viewer.

I also updated the changelog with the "Breaking changes" section I added to the 4.0 release notes.

### DIFF
--- a/lighthouse-viewer/app/src/lighthouse-report-viewer.js
+++ b/lighthouse-viewer/app/src/lighthouse-report-viewer.js
@@ -209,8 +209,8 @@ class LighthouseReportViewer {
     const warnMsg = `Version mismatch between viewer and JSON. Opening compatible viewer...`;
     logger.log(warnMsg, false);
 
+    // TODO: Handle 4x reports if we break viewer compat moving to v5.
     // Place report in IDB, then navigate current tab to the legacy viewer
-    // TODO: viewer4x :)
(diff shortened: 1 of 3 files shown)


lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -211,6 +210,7 @@ class LighthouseReportViewer {
     logger.log(warnMsg, false);
 
     // Place report in IDB, then navigate current tab to the legacy viewer
+    // TODO: viewer4x :)
```
paulirish: i know this one is probably my fault but can you expand the comment?

is it.. "TODO: Handle 4x reports if we break viewer compat moving to v5"  ??

brendankenny: done

brendankenny:
I also updated the changelog with the "Breaking changes" section I added to the [4.0 release notes](https://github.com/GoogleChrome/lighthouse/releases/tag/v4.0.0).

PR #7049, opened 2019-01-17
misc: replace TODOs with github issue links

patrickhulce:
**Summary**
Cleans up my TODOs with github issues so they're more visible and easier to prioritize.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -1241,7 +1241,6 @@ class Driver {
    * @return {Promise<void>}
    */
   async beginEmulation(settings) {
-    // TODO: https://github.com/GoogleChrome/lighthouse/issues/7044
     if (!settings.disableDeviceEmulation) {
       if (settings.emulatedFormFactor === 'mobile') {
         await emulation.enableNexus5X(this);
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -1241,7 +1241,7 @@ class Driver {
    * @return {Promise<void>}
    */
   async beginEmulation(settings) {
-    // TODO(phulce): remove this flag on next breaking change
+    // TODO: https://github.com/GoogleChrome/lighthouse/issues/7044
     if (!settings.disableDeviceEmulation) {
       if (settings.emulatedFormFactor === 'mobile') {
         await emulation.enableNexus5X(this);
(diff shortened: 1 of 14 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -129,8 +129,6 @@ class OptimizedImages extends Gatherer {
    * @return {Promise<?{fromProtocol: boolean, originalSize: number, jpegSize: number, webpSize: number}>}
    */
   calculateImageStats(driver, networkRecord) {
-    // TODO(phulce): remove this dance of trying _getEncodedResponse with a fallback when Audits
```
patrickhulce: at this point, I'm not sure it makes sense. It's a solid fallback option for when the protocol fails.

lighthouse-core/lib/network-request.js
```diff
@@ -51,7 +51,6 @@ const RESOURCE_TYPES = {
 module.exports = class NetworkRequest {
   constructor() {
     this.requestId = '';
-    // TODO(phulce): remove default DevTools connectionId
```
patrickhulce: I'm not convinced this is necessary anymore

lighthouse-core/scripts/assert-golden-lhr-unchanged.sh
```diff
@@ -28,7 +28,6 @@ trap teardown EXIT
 
 colorText "Generating a fresh LHR..." "$purple"
 set -x
-# TODO(phulce): add a lantern LHR-differ
```
patrickhulce: I'm not convinced this is necessary anymore

lighthouse-core/lib/network-recorder.js
```diff
@@ -231,7 +231,6 @@ class NetworkRecorder extends EventEmitter {
     const modifiedData = {
       ...data,
       // Copy over the initiator as well to match DevTools behavior
-      // TODO(phulce): abandon this DT hack and update Lantern graph to handle it
```
patrickhulce: I'm not convinced this is necessary anymore, it's probably easier to just stay in sync with wht devtools does

lighthouse-core/gather/driver.js
```diff
@@ -1241,7 +1241,7 @@ class Driver {
    * @return {Promise<void>}
    */
   async beginEmulation(settings) {
-    // TODO(phulce): remove this flag on next breaking change
+    // TODO: https://github.com/GoogleChrome/lighthouse/issues/7044
```
brendankenny: I'm not sure if it's worth keeping the TODO for things like this that aren't so targeted at a single point (or that single point is easy to find again since the goal is to remove something altogether)

patrickhulce: removed

patrickhulce:
sooooooo it sounds like you'd prefer just removing all of them?

I'm down for that.

brendankenny:
> sooooooo it sounds like you'd prefer just removing all of them?

ha, probably. I'm also fine with this PR for now. What do you want to do?

patrickhulce:
I'll remove most of them for now. There are a few that I like having the easy exact line handle for finding when we actually address the issue, but most are stand alone.

PR #7051, opened 2019-01-17
misc(viewer): fix types to reference LH.Result

paulirish:
We're dealing with the real LHR here, not the ReportResult. Not sure how this happened but it's funny.

(EDIT: [it was me](https://github.com/GoogleChrome/lighthouse/commit/ed5b38ecb40869dda0e817b0d268ee65bd5ad109#diff-fe42ad1345a50fa5e61842d4e35473b3) :)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -80,7 +80,7 @@ class ReportUIFeatures {
   /**
    * Adds export button, print, and other functionality to the report. The method
    * should be called whenever the report needs to be re-rendered.
-   * @param {LH.ReportResult} report
+   * @param {LH.Result} report
    */
   initFeatures(report) {
```
brendankenny: should also drop the `Util.prepareReportResult` in `report-ui-features-test.js`

(that is a seriously anemic test file for the amount of jsdom setup it does :)

lighthouse-viewer/app/src/lighthouse-report-viewer.js
```diff
@@ -132,7 +132,7 @@ class LighthouseReportViewer {
   }
 
   /**
-   * @param {LH.ReportResult} json
```
brendankenny: Really all the callers to this are either doing it with `any` or have casted `any`, so this is kind of a lie.

`json` should really have type `unknown` and we could have `_validateReportJson` verify that it's an `LH.Result` (as it's superficially doing already), but that can be a project for another day.

Should we add a TODO or file an issue?

PR #7056, opened 2019-01-18
core(lhr): include resourceSize in network-requests audit

paulirish:
along with the idea of #7052.... i noticed resource size isn't exposed in the lhr anywhere. seems reasonable to tack on here. wdyt?


lighthouse-core/audits/network-requests.js
```diff
@@ -63,6 +64,13 @@ class NetworkRequests extends Audit {
           granularity: 1,
           text: 'Transfer Size',
         },
+        {
+          key: 'resourceSize',
+          itemType: 'bytes',
+          displayUnit: 'kb',
```
Hoten: isn't this bytes?

image

brendankenny: yeah, it's interpreted by the `itemType` but is put into the LHR as the `displayUnit`

lighthouse-core/test/results/sample_v2.json
```diff
@@ -870,6 +880,7 @@
             "startTime": 636.6400000115391,
             "endTime": 1213.2910000218544,
             "transferSize": 139,
+            "resourceSize": 0,
```
brendankenny: why do these 0 `resourceSize` items have `transferSize` set?

lighthouse-core/audits/network-requests.js
brendankenny: > but is put into the LHR as the displayUnit

er, that's not right. It is if it's in an i18ned string, otherwise it's not until `details-renderer.js` that the `displayUnit` is used...which is pointless here because these aren't displayed in the html report

lighthouse-core/test/results/sample_v2.json
brendankenny: > why do these 0 resourceSize items have transferSize set?

and this makes a lot more sense now when 139 isn't the `transferSize` in KB, it's in bytes :)

Issue #7005, opened 2019-01-14
Performance Audit attribution from third-party wrapping


nicjansma:
**Feature request summary**

[We](https://akamai.com) are a [third-party analytics](https://www.akamai.com/us/en/products/web-performance/mpulse-real-user-monitoring.jsp) provider that provides Real User Monitoring (RUM) and other features such as JavaScript Error Tracking through a JavaScript library our customers include on their pages called [Boomerang](https://github.com/akamai/boomerang).

Lately, we've had several customers raise support issues over their concerns that our JavaScript library is causing excessive JavaScript CPU usage, according to Lighthouse/PSI reports.

Here's an example:

![image](https://user-images.githubusercontent.com/1004649/51137113-1a370b80-180c-11e9-8a87-988ec91827ba.png)

Upon investigation, we found that our [Error Tracking plugin](https://akamai.github.io/boomerang/BOOMR.plugins.Errors.html) was causing Lighthouse to be confused about work being done in our customer's app.  The plugin wraps several high-level functions, and the callbacks of, `setTimeout()`, `addEventListener()` and others.  It does this to ensure that the customer's application can get detailed error messages (instead of the generic `Script error.` message) when errors are generated from scripts in other origins.  More details are [here](https://akamai.github.io/boomerang/BOOMR.plugins.Errors.html#toc11__anchor), and other [error tracking](https://docs.trackjs.com/browser-agent/sdk-reference/agent-config/#callbackenabled) libraries do this wrapping as well.

Because Boomerang is wrapping the callbacks in `setTimeout()`/etc, _any_ function in _any_ file that calls `setTimeout()` will be attributed to Boomerang instead, because Boomerang's wrapped callback is at the "bottom" of the stack.  As you can imagine, this accounts for a lot of the JavaScript CPU time on many sites.  The actual work done in the wrapped function is [minimal (but non-zero)](https://github.com/akamai/boomerang/blob/master/plugins/errors.js#L1319), but the time of the _entire_ callback is attributed to Boomerang instead of the "real" work being done.  Our understanding is the "URL" of the _JavaScript boot-up time_ /etc tables are based on the function/URL at the bottom of the stack.

If we were to disable Error Tracking and re-load the same page, we can see the real attribution moves a lot to the Page and other scripts (as well as Boomerang), but the "cost" of Boomerang has decreased from 770ms to 127ms:

[...]

PR #7059, opened 2019-01-18
core(bootup-time): better attribution

patrickhulce:
**Summary**
There are a couple things going on here all sparked by #7005. This bootup time audit is somewhat inconsistent because it's pulling double duty as two different things.

1. It flags the amount of time spent loading due to script execution.
2. It flags the CPU time hogged by the responsible script URL.

These are different because a script can force a layout. It's not fair to say the layout time was "script execution" necessarily because it was the browser doing the work, but it's also not fair to ignore the cost of layouts from third party scripts when saying "these are your problem scripts". Any ideas to solve this problem are welcome :)

Now on to the PR changes...

1. Script executions that are the result of a setTimeout are now assigned to the script that installed the timer rather than the code that executed during the timer. This solves part of #7005.
1. If we cannot find a URL to attribute the execution cost to, we put it into an "Other" bucket. While trying to get a solution to the problem presented in #7005 for XHR/fetch callbacks it uncovered the fact that because async stacks are off, frequently a script execution with a clear URL loses its stack trace and becomes URL-less. Previously all such work was disappearing from our totals.
1. The display value of the audit is now the sum of all script execution + script parse time. With the creation of the "Other" bucket, summing all execution time makes the total exactly the same as "Main thread breakdown". 

Future things to address:
- Turning on async stacks should enable us to better attribute the 'Other' category.
- We aren't utilizing all of the URLs we could be for attribution. Leveraging `frame` data and `scriptId` could help here.
- How do we handle legit inline script executions? Right now we prefer to find the first script file that executed because nearly all executions have the main page as the first "attributable URL" (obviously everything started because the page asked it to one way or another). This also means though that for scripts that legitimately have the inline script of the page to blame we still blame the first script file that executed. i.e. this PR does not actually solve the specific repro example provided in #7005

**Related Issues/PRs**
#7005

### DIFF
--- a/lighthouse-core/computed/main-thread-tasks.js
+++ b/lighthouse-core/computed/main-thread-tasks.js
@@ -41,6 +41,8 @@ const TraceOfTab = require('./trace-of-tab.js');
  * @prop {TaskGroup} group
  */
 
+/** @typedef {{timers: Map<string, TaskNode>}} PriorTaskData */
+
 class MainThreadTasks {
   /**
    * @param {LH.TraceEvent} event
@@ -71,15 +73,24 @@ class MainThreadTasks {
 
   /**
    * @param {LH.TraceEvent[]} mainThreadEvents
+   * @param {PriorTaskData} priorTaskData
    * @return {TaskNode[]}
    */
-  static _createTasksFromEvents(mainThreadEvents) {
+  static _createTasksFromEvents(mainThreadEvents, priorTaskData) {
     /** @type {TaskNode[]} */
     const tasks = [];
     /** @type {TaskNode|undefined} */
     let currentTask;
 
     for (const event of mainThreadEvents) {
+      // Save the timer data, TimerInstall events are instant events `ph === 'I'` so process them first.
+      if (event.name === 'TimerInstall' && currentTask) {
+        /** @type {string} */
+        // @ts-ignore - timerId exists when name is TimerInstall
+        const timerId = event.args.data.timerId;
+        priorTaskData.timers.set(timerId, currentTask);
+      }
+
       // Only look at X (Complete), B (Begin), and E (End) events as they have most data
       if (event.ph !== 'X' && event.ph !== 'B' && event.ph !== 'E') continue;
 
@@ -141,11 +152,13 @@ class MainThreadTasks {
   /**
    * @param {TaskNode} task
    * @param {string[]} parentURLs
+   * @param {{timers: Map<string, TaskNode>}} priorTaskData
    */
-  static _computeRecursiveAttributableURLs(task, parentURLs) {
+  static _computeRecursiveAttributableURLs(task, parentURLs, priorTaskData) {
     const argsData = task.event.args.data || {};
     const stackFrameURLs = (argsData.stackTrace || []).map(entry => entry.url);
 
+    /** @type {Array<string|undefined>} */
     let taskURLs = [];
     switch (task.event.name) {
       /**
@@ -161,6 +174,15 @@ class MainThreadTasks {
       case 'v8.compileModule':
         taskURLs = [task.event.args.fileName].concat(stackFrameURLs);
         break;
+      case 'TimerFire': {
+        /** @type {string} */
+        // @ts-ignore - timerId exists when name is TimerFire
+        const timerId = task.event.args.data.timerId;
+        const timerInstallerTaskNode = priorTaskData.timers.get(timerId);
+        if (!timerInstallerTaskNode) break;
+        taskURLs = timerInstallerTaskNode.attributableURLs.concat(stackFrameURLs);
+        break;
+      }
       default:
         taskURLs = stackFrameURLs;
         break;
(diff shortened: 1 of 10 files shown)


Hoten:
> Turning on async stacks should enable us to better attribute the 'Other' category.

Could you talk about why this is currently not enabled for LH? Were there performance concerns?

[Next v8 release](https://v8.dev/blog/v8-release-73#async-stack-traces) will do that by default. They claim zero-cost, so perhaps performance issues have been resolved?

lighthouse-core/computed/main-thread-tasks.js
```diff
@@ -71,15 +73,24 @@ class MainThreadTasks {
...
    * @param {LH.TraceEvent[]} mainThreadEvents
+   * @param {PriorTaskData} priorTaskData
    * @return {TaskNode[]}
    */
-  static _createTasksFromEvents(mainThreadEvents) {
+  static _createTasksFromEvents(mainThreadEvents, priorTaskData) {
     /** @type {TaskNode[]} */
     const tasks = [];
     /** @type {TaskNode|undefined} */
     let currentTask;
```
Hoten: `currentTask` is never set, what's going on?

patrickhulce:
> Could you talk about why this is currently not enabled for LH? Were there performance concerns?

There were performance boogeyman concerns from enabling the debugger category. I never found statistically significant changes though.

> [Next v8 release](https://v8.dev/blog/v8-release-73#async-stack-traces) will do that by default. They claim zero-cost, so perhaps performance issues have been resolved?

This is awesome news!!! Enables more accurate lantern estimates, better attribution, huzzah!!!! :D 🎉 🎉 🎉 🎉

paulirish:
In order to satisfy #7005 (and to generally improve `bootup-time`), I think we need to handle attribution a little differently. ⛵️  First, changes 2 and 3 above SGTM.. I only have a concern with the new timer attribution. 

Example scenario: 

```js
jQuery(document).ready(function(){ 
	myapp.initalize();
});
```

`jquery.js` owns the event handler so it will get all the self-time attribution, even though the cost of this function is over in `myapp.js`. :/  

Basically anytime the flame chart has [a different color at the top vs the bottom](https://user-images.githubusercontent.com/39191/52501628-9fa4a600-2b95-11e9-9c9d-17e3a29c83e5.png), our current method leads to inaccurate results.


We have avoided turning on the JS sampling profiler 'till now, but I think that's the correct move here.

-------

Here's Nic's [test page](https://output.jsbin.com/kadohut/4/quiet) with JS sampling off.. see that all bottomup time is associated to the start of the JS stack.
image

And again but with JS sampling on. Correct selftime association now:
image

patrickhulce:
Hm, I'm not sure we're on the same page with that example @paulirish. `myapp.js` should still receive attribution in that case because while the jquery file may have been the line that installed the handler, the *attributable URL* that installed the handler was still `myapp.js`.

If in the `jquery.js` script evaluation it installed it's own `jQuery(document).ready` listeners *then* `jquery.js` would receive attribution. The only edge case that remains after this is when the `jQuery(document).ready` occurs inline of the page in which case jQuery gets mistakenly blamed.

patrickhulce:
I'm definitely not against enabling the sampling profiler though if we can use DZL to prove it's OK :D better fidelity is better fidelity and I won't fight that!

paulirish:
Okay slightly diff example:

```
+------------------------------------------+
|       jQuery DCL handler (100ms)         |
+-------------------+----------------------+
|    jquery fns     |     myapp fns        |
+-------------------+----------------------+
```

I'd expect to see jquery.js get 50ms and myapp to get 50ms.  Right?

patrickhulce:
Ah, I see what you're saying now. 

Yeah, we don't split tasks at all right now. It's a bit messy to try to half do self-time though, don't you think?


In 
```
+------------------------------------------+
|               Myapp (100ms)              |
+-------------------+----------------------+
|    myapp fns      |     jquery fns       |
+-------------------+----------------------+
```

I want myapp to get full blame for all of this work.  From #7005 these were my thoughts:

Two Main Options I see:
* We attribute time to each script according to the aggregate selfTime of all its executions. i.e. if we have jQuery on the page, then jQuery would receive the attribution for all the users' calls to `$.whatever`. This is typically what a developer wants to see when working on identifying bottlenecks during a performance deep dive. **However, this is probably not what we want to surface in Lighthouse**. This somewhat presumes that the calls to functions are fixed and we're trying to have the largest overall impact improving things, but we're trying to surface what scripts _instigated_ the most work. Let's take a look at number 2.
* We attribute time to each script according to the aggregate totalTime of all tasks it initiated. i.e. if we have jQuery on the page, then jQuery would receive attribution for its initial installation and any handlers, setTimeouts, or additional work it defers, but _not_ for any calls to jQuery functions from other scripts. This is what we're trying to achieve, but there's currently a hole where we are attributing to the first URL that executed in the timeout instead of the URL that scheduled the timeout.

patrickhulce:
Basically, I'm saying it's really difficult to understand that a library installed it's own handler while it was being initialized and then is also letting other scripts piggyback on the same exact handler. Instead of registering new ones once it's invoked.

patrickhulce:
@paulirish we settled on this being a good step forward, right? LGTY?

paulirish:
> we settled on this being a good step forward, right? LGTY?

yes

PR #7065, opened 2019-01-19
core: add inline scripts to Scripts artifact

### LINKED ISSUE #7060: Include inline JS scripts in Scripts artifact
Audits that analyze js currently do not take into account inline js from `script` elements.

Hoten:
Fixes #7060.

I think it would be better to change the `Scripts` artifact to this shape:

```
{
  code: string
  requestId: string
}[]
```

I went with the lesser-diff change for the initial PR, but if the above seems like a good idea, I'll change it to that. thoughts?


lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -30,7 +30,7 @@ module.exports = [
           overallSavingsBytes: '>45000',
           overallSavingsMs: '>500',
           items: {
-            length: 1,
+            length: 3,
```
Hoten: now takes into account [these two inline scripts](  // FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES -- FILLER DATA JUST TO OCCUPY BYTES)

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,7 +22,7 @@ class Scripts extends Gatherer {
   async afterPass(passContext, loadData) {
     const driver = passContext.driver;
 
-    /** @type {Object<string, string>} */
+    /** @type {LH.Artifacts['Scripts']} */
```
Hoten: 😶

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -78,20 +78,24 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
     const items = [];
     const warnings = [];
     for (const requestId of Object.keys(artifacts.Scripts)) {
-      const scriptContent = artifacts.Scripts[requestId];
+      const scriptContentEntry = artifacts.Scripts[requestId];
```
Hoten: PSA: typescript doesn't support type guarding for expressions like

```
const scriptContent = Array.isArray(artifacts.Scripts[requestId]) ?
        artifacts.Scripts[requestId] : [artifacts.Scripts[requestId]];
```

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -78,20 +78,24 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
...
-        // If the total number of bytes to be saved is quite small, it's also safe to ignore.
-        if (result.wastedPercent < IGNORE_THRESHOLD_IN_PERCENT ||
-          result.wastedBytes < IGNORE_THRESHOLD_IN_BYTES ||
-          !Number.isFinite(result.wastedBytes)) continue;
-        items.push(result);
-      } catch (err) {
-        warnings.push(`Unable to process ${networkRecord.url}: ${err.message}`);
+      for (const scriptContent of scriptContents) {
+        try {
+          const result = UnminifiedJavaScript.computeWaste(scriptContent, networkRecord);
```
patrickhulce: I don't think this is what we want, right? IIUC, the network record is going to be the main document for all inline scripts, so we'll calculate the percent waste of a single inline script and report X% savings on the main document. Then we'll hit the next inline script and do it again?

Hoten: I'll get some reports uploaded and we'll take a look.

FYI this is the relevant code: https://github.com/GoogleChrome/lighthouse/blob/1098f952d5c75b22668aa572532415f81d2817af/lighthouse-core/audits/byte-efficiency/byte-efficiency-audit.js#L82

It should be reporting each un-minified script, and reporting some percentage relative to the main doc that can be saved.

lighthouse-core/gather/gatherers/scripts.js
patrickhulce: I think we should do `Array<LH.Artifacts['Scripts'][0]>` 😆

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,21 +22,46 @@ class Scripts extends Gatherer {
...
+      return getElementsInDocument('script')
+        .filter(meta => !meta.src && meta.text.trim())
+        .map(meta => meta.text);
+    })()`, {useIsolation: true});
+
+    if (inlineScripts.length) {
+      const mainResource = loadData.networkRecords.find(
+        request => URL.equalWithExcludedFragments(request.url, passContext.url));
+      if (!mainResource) {
+        throw new Error('could not locate mainResource');
```
patrickhulce: I'm a teensy bit worried that this is going to start throwing errors into our perf category on edge cases we previously didn't really care about. We don't actually *need* a request ID for the whole thing to work. WDYT about making `requestId` optional?

lighthouse-core/test/results/sample_v2.json
```diff
@@ -1998,41 +1998,10 @@
...
-            "wastedPercent": 42.52388078488413
-          }
-        ],
-        "overallSavingsMs": 150,
-        "overallSavingsBytes": 30470
-      }
+      "score": null,
+      "scoreDisplayMode": "error",
+      "rawValue": null,
+      "errorMessage": "artifacts.Scripts is not iterable"
```
patrickhulce: update the artifacts.json?

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,21 +22,46 @@ class Scripts extends Gatherer {
...
-    const scriptContentMap = {};
+    /** @type {LH.Artifacts['Scripts']} */
+    const scripts = [];
+
+    /** @type {string[]} */
+    const inlineScripts = await driver.evaluateAsync(`(() => {
+      ${getElementsInDocumentString};
+
+      return getElementsInDocument('script')
+        .filter(meta => !meta.src && meta.text.trim())
```
patrickhulce: seems like this one isn't a `meta` ;)

lighthouse-core/gather/gatherers/scripts.js
Hoten: 😱 

counter arg: you know it's an array in fewer characters when you see `[]`

lighthouse-core/gather/gatherers/scripts.js
patrickhulce: If I knew them I would fix them :)

I just meant that we have seen this error popup in peoples reports before and it's mostly 🤷‍♂️and move on since it wasn't that big a deal. Now it'll be a bit more user-facing. 

I actually had to tackle this for canonical though and think it'll be solid.

https://github.com/GoogleChrome/lighthouse/pull/7080/files#diff-2113a8215d43931339dbb50287d89dcbR450

Hoten: done. thoughts on handling reporting of unminified scripts in the JS audit? I just put `?` as a placeholder url :)

Hoten: also, one of us should extract that function to somewhere else (whoever merges last)

patrickhulce: extract it off of `network-analyzer`? that actually seems like a good place for it :)

though if you mean we should finally move `network-analyzer` out of the `depdency-graph/simulator` folder then I agree with you :)

patrickhulce: > done. thoughts on handling reporting of unminified scripts in the JS audit? I just put ? as a placeholder url 

Ideally we'd try to provide a snippet of the code like we do for CSS, but I'm OK with this for now until we see how common it is.

patrickhulce: maybe when the fancy code snippet preview (#6901) lands we can do something cool here :D

Hoten: yeah having it in `network-analyzer` is perfect

💯 to snippets

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -77,21 +77,24 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
     /** @type {Array<LH.Audit.ByteEfficiencyItem>} */
     const items = [];
     const warnings = [];
-    for (const requestId of Object.keys(artifacts.Scripts)) {
-      const scriptContent = artifacts.Scripts[requestId];
+    for (const {requestId, code} of artifacts.Scripts) {
+      if (!code) continue;
```
brendankenny: should always be code, shouldn't there? Just excluding the zero-length case?

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -77,21 +77,24 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
...
+        const result = UnminifiedJavaScript.computeWaste(code, networkRecord);
         // If the ratio is minimal, the file is likely already minified, so ignore it.
         // If the total number of bytes to be saved is quite small, it's also safe to ignore.
         if (result.wastedPercent < IGNORE_THRESHOLD_IN_PERCENT ||
           result.wastedBytes < IGNORE_THRESHOLD_IN_BYTES ||
           !Number.isFinite(result.wastedBytes)) continue;
         items.push(result);
       } catch (err) {
-        warnings.push(`Unable to process ${networkRecord.url}: ${err.message}`);
+        if (networkRecord) {
```
brendankenny: can we have a fallback like for `url` in `computeWaste()` since we know it's only scripts in the main page missing the network record?

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -61,7 +61,7 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
     const wastedBytes = Math.round(totalBytes * wastedRatio);
 
     return {
-      url: networkRecord.url,
+      url: networkRecord ? networkRecord.url : '?',
```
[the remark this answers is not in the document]
brendankenny: > seems like we always know the undefined case is the main page

admittedly we don't know that will always be the case, so maybe it's better to make that certain in the gatherer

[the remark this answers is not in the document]
patrickhulce: > seems like we always know the undefined case is the main page

I think the problem here is inline scripts/evals added via other scripts. If it works like CSS those tend to be "inline" but shouldn't be considered part of the main page. I think the preference should be for code snippets showing the code itself, but I'd be more comfortable conveying it's in the "Other" / "Inline" / "Eval" bucket than the main page.

lighthouse-core/test/results/artifacts/artifacts.json
```diff
@@ -10,11 +10,40 @@
...
...n            // we can patch the real socket\n            proxySocket = socket;\n        }\n        patchOnProperties(proxySocket, ['close', 'error', 'message', 'open']);\n        return proxySocket;\n    };\n    for (var prop in WS) {\n        _global.WebSocket[prop] = WS[prop];\n    }\n}\n\n/**\n * @license\n * Copyright Google Inc. All Rights Reserved.\n *\n * Use of this source code is governed by an MIT-style license that can be\n * found in the LICENSE file at https://angular.io/license\n */\nvar eventNames = 'copy cut paste abort blur focus canplay canplaythrough change click contextmenu dblclick drag dragend dragenter dragleave dragover dragstart drop durationchange emptied ended input invalid keydown keypress keyup load loadeddata loadedmetadata loadstart message mousedown mouseenter mouseleave mousemove mouseout mouseover mouseup pause play playing progress ratechange reset scroll seeked seeking select show stalled submit suspend timeupdate volumechange waiting mozfullscreenchange mozfullscreenerror mozpointerlockchange mozpointerlockerror error webglcontextrestored webglcontextlost webglcontextcreationerror'\n    .split(' ');\nfunction propertyDescriptorPatch(_global) {\
```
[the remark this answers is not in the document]
Hoten: seems that the initial network record gets a different sort of ID than the records it later requests.

keep in mind that before, the inline scripts weren't in this Scripts object. So that's why you don't see the weird `44FDCEE5FC54A761B82815FAC6719910` request id previously.

Hoten: I shouldn't have changed these request id values. reverted.

lighthouse-core/test/results/sample_v2.json
```diff
@@ -1998,10 +1998,10 @@
       "id": "unminified-javascript",
       "title": "Minify JavaScript",
       "description": "Minifying JavaScript files can reduce payload sizes and script parse time. [Learn more](https://developers.google.com/speed/docs/insights/MinifyResources).",
-      "score": 0.88,
+      "score": 1,
       "scoreDisplayMode": "numeric",
-      "rawValue": 150,
-      "displayValue": "Potential savings of 30 KB",
+      "rawValue": 0,
+      "displayValue": "Potential savings of 15 KB",
```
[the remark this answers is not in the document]
Hoten: nope- due to mistakenly changing the requestId values in Scripts w/o also changing the values in the devtools logs. 

I assume I shouldn't just regenerate the whole thing, and instead minimize the diff by changing just what I needed to. Is that how we typically update sample_v2.json/artifacts.json?

lighthouse-core/test/results/sample_v2.json
```diff
@@ -2024,14 +2024,14 @@
         ],
         "items": [
           {
-            "url": "http://localhost:10200/zone.js",
-            "totalBytes": 71654,
-            "wastedBytes": 30470,
+            "url": "?",
```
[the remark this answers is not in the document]
Hoten: see above

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
Hoten: The new main doc scripts I added already have a zero-length check, so all good there. But the previous code was doing a zero-length check, and I didn't want to modify that behavior.

we could add a zero length check to the other scripts too, and then remove this check.

Hoten:
(fyi, did a first pass, will do second pass changes after a meeting)

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
Hoten: done. the data was stable too so I include it.

lighthouse-core/test/results/sample_v2.json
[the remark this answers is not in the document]
patrickhulce: I usually edit it by hand, unless there's something really massive that we need to generate from a real run

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,21 +23,46 @@ class Scripts extends Gatherer {
...
+      ${getElementsInDocumentString};
+
+      return getElementsInDocument('script')
+        .filter(script => !script.src && script.text.trim())
+        .map(script => script.text);
+    })()`, {useIsolation: true});
+
+    if (inlineScripts.length) {
+      const mainResource = loadData.networkRecords.find(
+        request => URL.equalWithExcludedFragments(request.url, passContext.url));
```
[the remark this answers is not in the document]
Hoten: done. after the two merge PRs merge maybe we should just move that check into `URL.equalWithExcludedFragments` as an early return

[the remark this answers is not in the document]
patrickhulce: > maybe we should just move that check into URL.equalWithExcludedFragments as an early return

yeah it's a *bit* different to do generically because in these cases we know for a fact there are no fragments on the network records but agreed we should not have to do these things, the URL.* methods should just be fast by default :)

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
Hoten: ACTUALLY, it reports the percentage saved relative to the script length, not the document length. fyi.

patrickhulce: relative to the javascript string length times the compression ratio of the main document, right?

Hoten: yup! it's very clever like that

Hoten:
Strangely, `yarn smoke byte` fails locally but not in appveyor

```
yarn smoke byte                                                                                            ✔  7201  13:46:19
yarn run v1.12.3
$ node lighthouse-cli/test/smokehouse/run-smoke.js byte
Running ONLY smoketests for: byte

Smoketest batch: perf-opportunity
byte smoketest starting…

byte smoketest results:
Command failed: node lighthouse-cli/test/smokehouse/smokehouse.js --config-path=lighthouse-cli/test/smokehouse/byte-config.js --expectations-path=byte-efficiency/expectations.js

Doing a run of 'http://localhost:10200/byte-efficiency/tester.html'...
$ node lighthouse-cli/index.js http://localhost:10200/byte-efficiency/tester.html --config-path=/Users/cjamcl/src/lighthouse/lighthouse-cli/test/smokehouse/byte-config.js --output-path=smokehouse-10774.report.json --output=json --quiet --port=0
Asserting expected results match those found. (http://localhost:10200/byte-efficiency/tester.html)
  ✓ final url: http://localhost:10200/byte-efficiency/tester.html
  ✓ error code: undefined

  ✘ difference at unused-javascript.details.overallSavingsBytes
              expected: ">=25000"
                 found: 21018

          found result:
      {
        "id": "unused-javascript",
        "title": "Remove unused JavaScript",
        "description": "Remove unused JavaScript to reduce bytes consumed by network activity.",
        "score": 0.75,
        "rawValue": 300,
        "displayValue": "Potential savings of 21 KB",
        "details": {
          "type": "opportunity",
          "headings": [
            {
              "key": "url",
              "valueType": "url",
              "label": "URL"
            },
            {
              "key": "totalBytes",
              "valueType": "bytes",
              "label": "Size (KB)"
            },
            {
              "key": "wastedBytes",
              "valueType": "bytes",
              "label": "Potential Savings (KB)"
            }
          ],
          "items": [
            {
              "url": "http://localhost:10200/byte-efficiency/script.js",
              "totalBytes": 53181,
              "wastedBytes": 14415,
              "wastedPercent": 27.105307847614018
            },
            {
              "url": "http://localhost:10200/byte-efficiency/tester.html",
              "totalBytes": 15073,
              "wastedBytes": 6603,
              "wastedPercent": 43.80430423740142
            }
          ],
          "overallSavingsMs": 300,
          "overallSavingsBytes": 21018
        }
      }


  ✘ difference at offscreen-images.details.items[3].url
              expected: "/large.svg$/"
                 found: "http://localhost:10200/byte-efficiency/lighthouse-480x320.webp?lazilyLoaded=true"

          found result:
      {
        "id": "offscreen-images",
        "title": "Defer offscreen images",
        "description": "Consider lazy-loading offscreen and hidden images after all critical resources have finished loading to lower time to interactive. [Learn more](https://developers.google.com/web/tools/lighthouse/audits/offscreen-images).",
        "score": 1,
        "rawValue": 0,
        "displayValue": "Potential savings of 88 KB",
        "warnings": [],
        "details": {
          "type": "opportunity",
          "headings": [
            {
              "key": "url",
              "valueType": "thumbnail",
              "label": ""
            },
            {
              "key": "url",
              "valueType": "url",
              "label": "URL"
            },
            {
              "key": "totalBytes",
              "valueType": "bytes",
              "label": "Size (KB)"
            },
            {
              "key": "wastedBytes",
              "valueType": "bytes",
              "label": "Potential Savings (KB)"
            }
          ],
          "items": [
            {
              "url": "http://localhost:10200/byte-efficiency/lighthouse-unoptimized.jpg",
              "requestStartTime": 98037.281547,
              "totalBytes": 51941,
              "wastedBytes": 51941,
              "wastedPercent": 100
            },
            {
              "url": "http://localhost:10200/byte-efficiency/lighthouse-480x320.webp",
              "requestStartTime": 98039.276147,
              "totalBytes": 11460,
              "wastedBytes": 11460,
              "wastedPercent": 100
            },
            {
              "url": "http://localhost:10200/byte-efficiency/lighthouse-480x320.webp?invisible",
              "requestStartTime": 98039.286467,
              "totalBytes": 11460,
              "wastedBytes": 11460,
              "wastedPercent": 100
            },
            {
              "url": "http://localhost:10200/byte-efficiency/lighthouse-480x320.webp?lazilyLoaded=true",
              "requestStartTime": 98045.118763,
              "totalBytes": 11460,
              "wastedBytes": 11460,
              "wastedPercent": 100
            },
            {
              "url": "http://localhost:10200/byte-efficiency/large.svg",
              "requestStartTime": 98039.743435,
              "totalBytes": 3281,
              "wastedBytes": 3281,
              "wastedPercent": 100
            }
          ],
          "overallSavingsMs": 0,
          "overallSavingsBytes": 89602
        }
      }

  Correctly passed 7 assertions

7 passing
2 failing
smoketest-byte: 32674.869ms
byte smoketest complete.

We have 1 failing smoketests: byte
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Hoten:
The above only happens in canary, not stable. ¯\_(ツ)_/¯

patrickhulce:
> The above only happens in canary

geez, what's goin on with canary these days!? 😆

Hoten:
for reference, the only difference for `unused-javascript` is `wastedBytes` for `http://localhost:10200/byte-efficiency/script.js`. Stable has `22317` but canary gives `14415`.

wastedData (`UnusedJavaScript.computeWaste`) for `script.js`...

stable:
```
[ { unusedLength: 22240, contentLength: 52997 } ]
```

canary
```
[ { unusedLength: 14365, contentLength: 52997 } ]
```

Great, so the coverage data is what's different.

Side by side, this shows the only difference:
image
Left is canary, right is stable.

Hmm.

Hoten:
as a sanity check I tried setting breakpoints in the "used" unused function. I suppose breakpoints on code that v8 determines is dead code is not possible?

image

impossible to set a breakpoint there. (also, in the `if (false)` block). I have seen this behavior before but never connected it to dead code, just thought devtools was buggy :)

Hoten:
https://bugs.chromium.org/p/chromium/issues/detail?id=927464

brendankenny:
> https://bugs.chromium.org/p/chromium/issues/detail?id=927464

we're not blocked on this, though, right? Since it affects things regardless of the shape of the artifact?

Hoten:
correct, that test fails locally on canary regardless of any changes.

Hoten:
Reference image:

image

We could also link to the `script` tag, but that seems like an enhancement that can be punted. Perhaps would require adding a `nodePath` to the Scripts artifact?

Hoten:
Do we consider changing an artifact to be a breaking change?

patrickhulce:
> Do we consider changing an artifact to be a breaking change?

I think only after 5.0 when we make plugins an official public thing

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -26,9 +26,26 @@ module.exports = [
...
           overallSavingsMs: '>500',
-          items: {
-            length: 1,
-          },
+          items: [
+            {
+              url: 'http://localhost:10200/byte-efficiency/script.js',
+              totalBytes: 53181,
+              wastedBytes: 46481,
+              wastedPercent: 87.40117365133875,
```
patrickhulce: I would feel slightly better about a looser assertion here

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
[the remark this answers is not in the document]
patrickhulce: Oh I gotcha, yeah agreed, we should just nix the finding of the main resource in the gatherer 👍 

I think just putting `inline` without giving any idea to the content is pretty rough UX. Maybe we can at least show the first 40 characters or something?

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
Hoten: wb a new syntax like `x % 5` to indicate within +/- 5% of x?

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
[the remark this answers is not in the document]
Hoten: I don't like omitting the `requestId` if it is readily available for inline content. How about not using an inline script's `requestId` in the audit, but keeping it in the artifact so that something else might use it if needed w/o grabbing the main record manually?

new pic: 
image

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
patrickhulce: Oooh I love that, how about `x +/- y` though ;)

brendankenny:
> `+/-`

can you move to a separate PR :)

Hoten:
> > `+/-`
> 
> can you move to a separate PR :)

sure. stashing this here for now:
```
items: [
    {
      url: 'http://localhost:10200/byte-efficiency/script.js',
      wastedBytes: '46481 +/- 100',
      wastedPercent: '87 +/- 5',
    },
    {
      url: 'inline: \n  function unusedFunction() {\n    // Un...',
      wastedBytes: '6581 +/- 100',
      wastedPercent: '99.6 +/- 0.1',
    },
    {
      url: 'inline: \n  // Used block #1\n  // FILLER DATA JUS...',
      wastedBytes: '6559 +/- 100',
      wastedPercent: 100,
    },
  ],
```

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,21 +23,53 @@ class Scripts extends Gatherer {
...
+    const inlineScripts = await driver.evaluateAsync(`(() => {
+      ${getElementsInDocumentString};
+
+      return getElementsInDocument('script')
+        .filter(script => !script.src && script.text.trim())
+        .map(script => script.text);
+    })()`, {useIsolation: true});
+
+    if (inlineScripts.length) {
+      const mainResource = loadData.networkRecords.find(request =>
```
brendankenny: how many places do we figure this out now? :)

lighthouse-core/gather/gatherers/scripts.js
```diff
@@ -20,21 +23,53 @@ class Scripts extends Gatherer {
...
+          URL.equalWithExcludedFragments(request.url, passContext.url));
+      if (!mainResource) {
+        log.warn('Scripts', 'could not locate mainResource');
+      }
+      scripts.push(
+        ...inlineScripts.map(content => {
+          return {
+            content,
+            inline: true,
+            requestId: mainResource ? mainResource.requestId : undefined,
```
brendankenny: move this out of the `map` to a named variable to keep it cleaner in here.

It definitely sucks that we have to add this for the case that should never happen but turns out does. Maybe add a comment too about `mainResource` is expected to be found, but on the very small chance something weird happens and it isn't found, we'll roll without it for now?

brendankenny: oh, actually, forgot about the earlier conversation. We should switch to `NetworkAnalyzer.findMainDocument()` once #7080 lands so we can make `requestId` not optional

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
patrickhulce: +/- landed right we can update these now?

lighthouse-core/test/audits/byte-efficiency/unminified-javascript-test.js
```diff
@@ -58,30 +67,67 @@ describe('Page uses optimized responses', () => {
     }));
 
     expect(results).toMatchObject([
-      {url: 'foo.js', wastedPercent: 57, wastedKB: 11},
-      {url: 'other.js', wastedPercent: 53, wastedKB: 27},
+      {url: 'foo.js', wastedPercent: 56, wastedKB: 11},
```
patrickhulce: why did this change?

lighthouse-core/test/audits/byte-efficiency/unminified-javascript-test.js
```diff
@@ -58,30 +67,67 @@ describe('Page uses optimized responses', () => {
...
+      {url: 'other.js', wastedPercent: 53, wastedKB: 26},
```
patrickhulce: same here

lighthouse-core/audits/byte-efficiency/unminified-javascript.js
```diff
@@ -61,7 +62,9 @@ class UnminifiedJavaScript extends ByteEfficiencyAudit {
     const wastedBytes = Math.round(totalBytes * wastedRatio);
 
     return {
-      url: networkRecord.url,
+      url: networkRecord && !inline ?
```
patrickhulce: super nit: it feels like the inline is the special case we need to check for so it reads a bit easier in my head to say

an inline thing is something that's explicitly marked as such or doesn't have a network record
inline things get a prefix and content while everything else has a URL like normal

types/artifacts.d.ts
```diff
@@ -105,8 +105,8 @@ declare global {
       RobotsTxt: {status: number|null, content: string|null};
       /** Set of exceptions thrown during page load. */
       RuntimeExceptions: Crdp.Runtime.ExceptionThrownEvent[];
-      /** The content of all scripts loaded by the page, keyed by networkRecord requestId. */
-      Scripts: Record<string, string>;
+      /** The content of all scripts loaded by the page, and the networkRecord requestId that loaded them. Note, HTML documents will have one entry per script tag, all with the same requestId. */
```
patrickhulce: I would say "the networkRecord requestId that contained their content" since for it very well could have been a different script that "loaded" that other script.

lighthouse-core/test/audits/byte-efficiency/unminified-javascript-test.js
Hoten: whitespace changes in the provided network record contents. the `content: ` eats up some of the ws :)

PR #7090, opened 2019-01-24
core(renderer): do not attempt fireworks for devtools

Hoten:
This was a hotfix applied to the 4.0.0 devtools roll. Now adding here.

### DIFF
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -207,14 +207,12 @@ class ReportRenderer {
     }
 
     // Fireworks
-    if (!this._dom.isDevTools()) {
-      const scoresAll100 = report.reportCategories.every(cat => cat.score === 1);
-      if (scoresAll100) {
-        headerContainer.classList.add('score100');
-        this._dom.find('.lh-header', headerContainer).addEventListener('click', _ => {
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -207,12 +207,14 @@ class ReportRenderer {
     }
 
     // Fireworks
-    const scoresAll100 = report.reportCategories.every(cat => cat.score === 1);
-    if (scoresAll100) {
-      headerContainer.classList.add('score100');
-      this._dom.find('.lh-header', headerContainer).addEventListener('click', _ => {
-        headerContainer.classList.toggle('fireworks-paused');
-      });
+    if (!this._dom.isDevTools()) {
```
patrickhulce: is this a pavel-reaction undoing? 😆 

or do we just not want to deal with the differences right now getting it to work in a smaller header?

exterkamp: I like this more simply as 
```javascript
if (!this._dom.isDevTools() && scoresAll100) {
```
to be more flat, but it wfm either way.

Hoten: we are brushing up on the release deadline for next chrome, so quick is good. but also we don't even have a design for fireworks in DT. @exterkamp should totes look into that

PR #7103, opened 2019-01-29
core(fcp-3g): remove unused i18n for LR compatibility

exterkamp:
**Summary**
Drop the `require.resolve`to find the i18n strings in exchange for an inlined `UIStrings` so that browserify can find them and the audit can be run in Lightrider.

### DIFF
--- a/lighthouse-core/audits/metrics/first-contentful-paint-3g.js
+++ b/lighthouse-core/audits/metrics/first-contentful-paint-3g.js
@@ -56,7 +56,7 @@ class FirstContentfulPaint3G extends Audit {
         context.options.scoreMedian
       ),
       rawValue: metricResult.timing,
-      displayValue: `${metricResult.timing}ms`,
+      displayValue: `${metricResult.timing}\xa0ms`,
     };
   }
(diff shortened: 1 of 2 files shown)


paulirish:
We don't need to i18n them. This will only be pulled programmatically.

Similar to metrics and screenshot-thumbnails which don't get i18n'd. So we can just leave them unadorned as well.

lighthouse-core/audits/metrics/first-contentful-paint-3g.js
```diff
@@ -60,7 +56,7 @@ class FirstContentfulPaint3G extends Audit {
         context.options.scoreMedian
       ),
       rawValue: metricResult.timing,
-      displayValue: str_(i18n.UIStrings.seconds, {timeInMs: metricResult.timing}),
+      displayValue: `${metricResult.timing}s`,
```
exterkamp: Is there a way to statically fill in i18n formatted to use timeInMs with formatting?  I figured this might be enough since we aren't really i18n'ing it.

PR #7105, opened 2019-01-29
misc(tap-targets): remove lines with trailing whitespace

exterkamp:
**Summary**
2 lines in tap-targets were being browserified with trailing white space which fails on the LR bundler.  So just remove those lines.

### DIFF
--- a/lighthouse-core/gather/gatherers/seo/tap-targets.js
+++ b/lighthouse-core/gather/gatherers/seo/tap-targets.js
@@ -324,6 +324,7 @@ class TapTargets extends Gatherer {
       ${pageFunctions.getNodePathString};
       ${pageFunctions.getNodeSelectorString};
       ${gatherTapTargets.toString()};
+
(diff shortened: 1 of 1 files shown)


patrickhulce:
Side-note should these failures you're finding be reflected in our own test suite somewhere @exterkamp? Is that easy?

Feel bad we keep breakin' ya :(

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -324,9 +324,7 @@ class TapTargets extends Gatherer {
       ${pageFunctions.getNodePathString};
       ${pageFunctions.getNodeSelectorString};
       ${gatherTapTargets.toString()};
-
```
brendankenny: can you keep the line break here but delete the trailing whitespace? It definitely helps with readability (the other line can definitely go)

Issue #7010, opened 2019-01-14
A11y i18n


exterkamp:
#### Provide the steps to reproduce
As of #7008 there are some gaps in a11y translation:
- title, and category description
![image](https://user-images.githubusercontent.com/6392995/51146336-5f0e7280-180b-11e9-9716-c3971a1408c0.png)
- Manual Audits
![image](https://user-images.githubusercontent.com/6392995/51146361-70f01580-180b-11e9-89cb-10464325b855.png)

paulirish:
> As of #7008 there are some gaps in a11y translation:

We could bring back https://github.com/GoogleChrome/lighthouse/pull/5726/files#diff-4bcfedf0736e1cd08b26fe659537ad90 to help avoid this.

patrickhulce:
> We could bring back https://github.com/GoogleChrome/lighthouse/pull/5726/files#diff-4bcfedf0736e1cd08b26fe659537ad90

Huzzah!! Yes please :D

brendankenny:
> We could bring back https://github.com/GoogleChrome/lighthouse/pull/5726/files#diff-4bcfedf0736e1cd08b26fe659537ad90 to help avoid this.

wait, is that to make it easier to spot when we've missed some strings when manually testing? Is it that different than just loading in russian and looking for english?

patrickhulce:
@brendankenny you have to wait for the translations to come back though, that bit was us programmatically generating it so you could notice as you're going.

exterkamp:
Well we can't load new russian until after its translated.  We need the script to load in the en-XA to see that it's bad before we send it to be translated afaik.

paulirish:
I also want to comment in this thread right now.

brendankenny:
> We need the script to load in the en-XA to see that it's bad before we send it to be translated afaik.

Cool. But I do predict the exact same pace of people actually loading another locale to check if we've missed something :P

exterkamp:
Well now I have to leave a comment.

PR #6229, opened 2018-10-12
i18n: convert accessibility audits

patrickhulce:
**Summary**
converts the non-manual accessibility audits to use i18n strings

PR #7113, opened 2019-01-30
i18n: a11y category title and description

exterkamp:
**Summary**
This i18n's the a11y category title, description, and manual description in the default config.

**Related Issues/PRs**
Follow up to Patricks #6229 
helps, but doesnt close: #7010


lighthouse-core/config/default-config.js
```diff
@@ -31,6 +31,12 @@ const UIStrings = {
   diagnosticsGroupTitle: 'Diagnostics',
   /** Description of the diagnostics section of the Performance category. Within this section are audits with non-imperative titles that provide more detail on the page's page load performance characteristics. Whereas the 'Opportunities' suggest an action along with expected time savings, diagnostics do not. Within this section, the user may read the details and deduce additional actions they could take. */
   diagnosticsGroupDescription: 'More information about the performance of your application.',
+  /** Title of the Accessibility category of audits. Accessibility contains audits focused on making web content accessible to users with disabilities. Also used as a label of a score gauge; try to limit to 20 characters. */
```
brendankenny: "This section contains audits focused..." instead of starting the sentence with "Accessibility"

lighthouse-core/config/default-config.js
```diff
@@ -31,6 +31,12 @@ const UIStrings = {
...
+  a11yCategoryTitle: 'Accessibility',
+  /** Description of the Accessibility category. This is displayed immediately under the Accessibility title element. No character length limits. 'improve the accessibility of your web app' becomes link text to additional documentation. */
```
brendankenny: I think you want to repeat the "This section contains audits focused on making..." description line for each of these, since (we're told) the translators don't generally cross reference the strings, so need the context for each.

The "under the Accessibility title element" may also be too specific, and may want to instead mention it's "displayed at the top of a list of audits focused on making..."

lighthouse-core/config/default-config.js
```diff
@@ -31,6 +31,12 @@ const UIStrings = {
...
+  a11yCategoryDescription: 'These checks highlight opportunities to [improve the accessibility of your web app](https://developers.google.com/web/fundamentals/accessibility). Only a subset of accessibility issues can be automatically detected so manual testing is also encouraged.',
+  /** Description of the Accessibility manual checks category. This is displayed above the manual checks section of the Accessibility category. No character length limits. 'conducting an accessibility review' becomes link text to additional documentation. */
```
brendankenny: the second sentence doesn't clarify much. Maybe something like "This description is displayed above a list of accessibility audits that currently have no automated test and so must be verified manually by the user. No character length limits...".

exterkamp:
Great feedback thanks! I'll take another pass at the SEO category title & description, since they are very similar, and could be improved in similar ways.

PR #7125, opened 2019-01-31
i18n: seo audits; add it in Polish

verseopl:
**Summary**

i18n

Ability to create reports in other languages

none
**Related Issues/PRs**

I think it isn't necessary


verseopl:
Ok, no problem;) I'm glad to hear that. What about other transltations e.g. dobetterweb audits?

PR #7020, opened 2019-01-15
deps: update axe-core to latest

brendankenny:
Taking over from #5565.

The PR is so big  mostly because all the `Learn more` links now point to the 3.1 docs instead of the 2.2 ones and I updated the `Accessibility` sample artifact.

fixes #5940

PR #7129, opened 2019-01-31
report: reorganize a11y audit groups

### LINKED ISSUE #7019: Proposed better organization of a11y audits
I was wondering how LH folks feel about me putting together a PR to better categorize the accessibility audits? The current categories are kind of vague, e.g. "uses attributes correctly" for example.

Here's a proposed reorg:

### Best practices

```
- [id] attributes on the page are unique
- [user-scalable="no"] is not used in the <meta name="viewport"> element and the [maximum-scale] attribute is not less than 5.
- The document does not use <meta http-equiv="refresh">
```

### Navigation

```
- [accesskey] values are unique
- No element has a [tabindex] value greater than 0
- The page contains a heading, skip link, or landmark region
```

### Names and labels

```
- Document has a <title> element
- Form elements have associated labels
- Buttons have an accessible name
- Links have a discernible name
- <frame> or <iframe> elements have a title
- <object> elements have [alt] text
- Image elements have [alt] attributes
- <input type="image"> elements have [alt] text
```

### ARIA

```
- [role]s have all required [aria-*] attributes
- Elements with [role] that require specific children [role]s, are present
- [role]s are contained by their required parent element
- [role] values are valid
- [aria-*] attributes match their roles
- [aria-*] attributes have valid values
- [aria-*] attributes are valid and not misspelled
```

### Audio and video

```
- <audio> elements contain a <track> element with [kind="captions"]
- <video> elements contain a <track> element with [kind="captions"]
- <video> elements contain a <track> element with [kind="description"]
```

### Tables and lists

```
- Cells in a <table> element that use the [headers] attribute only refer to other cells of that same table.
- <th> elements and elements with [role="columnheader"/"rowheader"] have data cells they describe.
- Presentational <table> elements avoid using <th>, <caption> or the [summary] attribute.
- <dl>'s contain only properly-ordered <dt> and <dd> groups, <script> or <template> elements.
- Definition list items are wrapped in <dl> elements
- Lists contain only <li> elements and script supporting elements (<script> and <template>).
- List items (<li>) are contained within <ul> or <ol> parent elements
```

### Contrast

```
- Background and foreground colors have a sufficient contrast ratio
```

### Internationalization and localization

```
- <html> element has a [lang] attribute
- <html> element has a valid value for its [lang] attribute
- [lang] attributes have a valid value
```

### Manual audits

```
# Note, these are unchanged

- Interactive elements indicate their purpose and state
- Visual order on the page follows DOM order
- The page has a logical tab order
- Interactive controls are keyboard focusable
- The user's focus is directed to new content added to the page
- User focus is not accidentally trapped in a region
- Offscreen content is hidden from assistive technology
- Headings don't skip levels
- HTML5 landmark elements are used to improve navigation
- Custom controls have associated labels
- Custom controls have ARIA roles
```

robdodson:
**Summary**

Improve the categorization of accessibility audits.

- Note: Also enables the `accesskeys` audit which was accidentally turned off in [a previous PR](https://github.com/GoogleChrome/lighthouse/pull/7020/files#diff-427cb962a42ed8d3463f1a773c2f4cc8L9).

**Related Issues/PRs**
fixes #7019

### DIFF
--- a/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
+++ b/lighthouse-cli/test/fixtures/a11y/a11y_tester.html
@@ -11,6 +11,11 @@
     <meta name="viewport" content="user-scalable=no, maximum-scale=1.0">
   </head>
   <body>
+    <p>accesskeys</p>
+    <section>
+      <button id="accesskeys1" accesskey="s">Foo</button>
+      <button id="accesskeys2" accesskey="s">Bar</button>
+    </section>
     <p>aria-allowed-attr</p>
     <section>
       <div
--- a/lighthouse-cli/test/smokehouse/a11y/expectations.js
+++ b/lighthouse-cli/test/smokehouse/a11y/expectations.js
@@ -230,8 +230,12 @@ module.exports = [
         },
       },
       'accesskeys': {
-        score: null,
-        scoreDisplayMode: 'manual',
+        score: 0,
+        details: {
+          items: {
+            length: 1,
+          },
+        },
       },
     },
   },
--- a/lighthouse-core/config/default-config.js
+++ b/lighthouse-core/config/default-config.js
@@ -40,7 +40,7 @@ const UIStrings = {
   /** Title of the best practices section of the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
   a11yBestPracticesGroupTitle: 'Best practices',
   /* Description of the best practices section within the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
-  a11yBestPracticesGroupDescription: 'These are opportunities to improve the interpretation of your content by users in different locales.',
+  a11yBestPracticesGroupDescription: 'These items highlight common accessibility best practices.',
   /* Title of the color contrast section within the Accessibility category. Within this section are audits with descriptive titles that highlight the color and vision aspects of the page's accessibility that are passing or failing. */
   a11yColorContrastGroupTitle: 'Contrast',
   /* Description of the color contrast section within the Accessibility category. Within this section are audits with descriptive titles that highlight the color and vision aspects of the page's accessibility that are passing or failing. */
--- a/lighthouse-core/config/default-config.js
+++ b/lighthouse-core/config/default-config.js
@@ -31,44 +31,44 @@ const UIStrings = {
   diagnosticsGroupTitle: 'Diagnostics',
   /** Description of the diagnostics section of the Performance category. Within this section are audits with non-imperative titles that provide more detail on the page's page load performance characteristics. Whereas the 'Opportunities' suggest an action along with expected time savings, diagnostics do not. Within this section, the user may read the details and deduce additional actions they could take. */
   diagnosticsGroupDescription: 'More information about the performance of your application.',
-  /** Title of the Accessibility category of audits. This section contains audits focused on making web content accessible to users with disabilities. Also used as a label of a score gauge; try to limit to 20 characters. */
+  /** Title of the Accessibility category of audits. This section contains audits focused on making web content accessible to all users. Also used as a label of a score gauge; try to limit to 20 characters. */
   a11yCategoryTitle: 'Accessibility',
-  /** Description of the Accessibility category. This is displayed at the top of a list of audits focused on making web content accessible to users with disabilities. No character length limits. 'improve the accessibility of your web app' becomes link text to additional documentation. */
+  /** Description of the Accessibility category. This is displayed at the top of a list of audits focused on making web content accessible to all users. No character length limits. 'improve the accessibility of your web app' becomes link text to additional documentation. */
   a11yCategoryDescription: 'These checks highlight opportunities to [improve the accessibility of your web app](https://developers.google.com/web/fundamentals/accessibility). Only a subset of accessibility issues can be automatically detected so manual testing is also encouraged.',
   /** Description of the Accessibility manual checks category. This description is displayed above a list of accessibility audits that currently have no automated test and so must be verified manually by the user. No character length limits. 'conducting an accessibility review' becomes link text to additional documentation. */
   a11yCategoryManualDescription: 'These items address areas which an automated testing tool cannot cover. Learn more in our guide on [conducting an accessibility review](https://developers.google.com/web/fundamentals/accessibility/how-to-review).',
+  /** Title of the best practices section of the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
+  a11yBestPracticesGroupTitle: 'Best practices',
+  /* Description of the best practices section within the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
+  a11yBestPracticesGroupDescription: 'These are opportunities to improve the interpretation of your content by users in different locales.',
   /* Title of the color contrast section within the Accessibility category. Within this section are audits with descriptive titles that highlight the color and vision aspects of the page's accessibility that are passing or failing. */
-  a11yColorContrastGroupTitle: 'Color Contrast Is Satisfactory',
+  a11yColorContrastGroupTitle: 'Contrast',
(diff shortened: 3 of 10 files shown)


lighthouse-core/config/default-config.js
```diff
@@ -369,42 +369,42 @@ const defaultConfig = {
...
+        {id: 'meta-refresh', weight: 1, group: 'a11y-best-practices'},
+        {id: 'meta-viewport', weight: 3, group: 'a11y-best-practices'},
+        {id: 'object-alt', weight: 4, group: 'a11y-names-labels'},
+        {id: 'tabindex', weight: 4, group: 'a11y-navigation'},
+        {id: 'td-headers-attr', weight: 1, group: 'a11y-tables-lists'},
+        {id: 'th-has-data-cells', weight: 1, group: 'a11y-tables-lists'},
         {id: 'valid-lang', weight: 1, group: 'a11y-language'},
-        {id: 'video-caption', weight: 4, group: 'a11y-describe-contents'},
-        {id: 'video-description', weight: 3, group: 'a11y-describe-contents'},
+        {id: 'audio-caption', weight: 4, group: 'a11y-audio-video'},
```
patrickhulce: is the new sorting scheme alphabetical unless it's audio-caption? :)

lighthouse-core/config/default-config.js
```diff
@@ -369,42 +369,42 @@ const defaultConfig = {
       description: str_(UIStrings.a11yCategoryDescription),
       manualDescription: str_(UIStrings.a11yCategoryManualDescription),
       auditRefs: [
+        {id: 'accesskeys', weight: 3, group: 'a11y-navigation'},
```
patrickhulce: @paulirish did we officially decide what our criteria for breaking changes was for v4 going forward? do we need to introduce this as `weight: 0` to start or are we doing that starting v5?

lighthouse-core/config/default-config.js
```diff
@@ -369,42 +369,42 @@ const defaultConfig = {
...
         {id: 'aria-required-parent', weight: 2, group: 'a11y-aria'},
         {id: 'aria-roles', weight: 3, group: 'a11y-aria'},
         {id: 'aria-valid-attr-value', weight: 2, group: 'a11y-aria'},
         {id: 'aria-valid-attr', weight: 5, group: 'a11y-aria'},
-        {id: 'audio-caption', weight: 4, group: 'a11y-correct-attributes'},
-        {id: 'button-name', weight: 10, group: 'a11y-element-names'},
-        {id: 'bypass', weight: 10, group: 'a11y-describe-contents'},
+        {id: 'button-name', weight: 10, group: 'a11y-names-labels'},
+        {id: 'bypass', weight: 10, group: 'a11y-navigation'},
         {id: 'color-contrast', weight: 6, group: 'a11y-color-contrast'},
```
patrickhulce: heh, is this really the only audit in the group?

robdodson: It is! Sorry, it's an odd duck. But oh so important.

lighthouse-core/config/default-config.js
robdodson: I think I did that one first while I was still figuring out if the audits were sorted by id or group. Fixed

paulirish:
Before / After:

image

(via `lighthouse http://localhost:10200/a11y/a11y_tester.html`)

paulirish:
@meggynw as fyi

robdodson:
I think all of the weights should be the same with the exception of `accesskeys` which was formerly listed as a manual audit. I could move it back to 0 weight and consider giving it a weight at a later, more breakage friendly time?

lighthouse-core/config/default-config.js
```diff
@@ -31,44 +31,44 @@ const UIStrings = {
   diagnosticsGroupTitle: 'Diagnostics',
   /** Description of the diagnostics section of the Performance category. Within this section are audits with non-imperative titles that provide more detail on the page's page load performance characteristics. Whereas the 'Opportunities' suggest an action along with expected time savings, diagnostics do not. Within this section, the user may read the details and deduce additional actions they could take. */
   diagnosticsGroupDescription: 'More information about the performance of your application.',
-  /** Title of the Accessibility category of audits. This section contains audits focused on making web content accessible to users with disabilities. Also used as a label of a score gauge; try to limit to 20 characters. */
+  /** Title of the Accessibility category of audits. This section contains audits focused on making web content accessible to all users. Also used as a label of a score gauge; try to limit to 20 characters. */
```
brendankenny: 👍 👍

lighthouse-core/config/default-config.js
brendankenny: > do we need to introduce this as `weight: 0` to start or are we doing that starting v5?

I think since this is not in the performance section it's less of a big deal (people might appreciate it, actually).

It was also a misunderstanding that disabled this audit, so we can treat the temporary `weight: 0` as a bug that we're now fixing? :D

robdodson:
Updated PTAL

brendankenny:
there's a test snapshot that needs to be updated (it's in the travis logs, but you can run `yarn run unit-cli:ci -u` to automatically fix) and the smoke test expectations will need to be switched back to not a manual result (https://github.com/GoogleChrome/lighthouse/blob/5c0483ca831af82b833cb27742e3bf0937813194/lighthouse-cli/test/smokehouse/a11y/expectations.js#L232-L235)

robdodson:
Updated PTAL

lighthouse-cli/test/smokehouse/a11y/expectations.js
```diff
@@ -230,8 +230,12 @@ module.exports = [
         },
       },
       'accesskeys': {
-        score: null,
-        scoreDisplayMode: 'manual',
+        score: 0,
```
brendankenny: to fix the CI error, it looks like this might be
```js
score: null,
scoreDisplayMode: 'notApplicable',
```
?

But did something change with the audit since removal in #7020?

lighthouse-core/config/default-config.js
```diff
@@ -31,44 +31,44 @@ const UIStrings = {
...
   /** Description of the Accessibility manual checks category. This description is displayed above a list of accessibility audits that currently have no automated test and so must be verified manually by the user. No character length limits. 'conducting an accessibility review' becomes link text to additional documentation. */
   a11yCategoryManualDescription: 'These items address areas which an automated testing tool cannot cover. Learn more in our guide on [conducting an accessibility review](https://developers.google.com/web/fundamentals/accessibility/how-to-review).',
+  /** Title of the best practices section of the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
+  a11yBestPracticesGroupTitle: 'Best practices',
+  /* Description of the best practices section within the Accessibility category. Within this section are audits with descriptive titles that highlight common accessibility best practices. */
+  a11yBestPracticesGroupDescription: 'These are opportunities to improve the interpretation of your content by users in different locales.',
```
brendankenny: not clear what different locales have to do with `duplicate-id`/`meta-refresh`/`meta-viewport`?

brendankenny: also seems like for almost all of these we could remove the beginning "These are", but maybe we want to wait for the efforts looking at all descriptions in the report

brendankenny: > different locales

ah, looks like this may be a c/p holdover from the "Internationalization and localization" group

lighthouse-cli/test/fixtures/a11y/a11y_tester.html
```diff
@@ -11,6 +11,11 @@
     <meta name="viewport" content="user-scalable=no, maximum-scale=1.0">
   </head>
   <body>
+    <p>accesskeys</p>
+    <section>
+      <button id="accesskeys1" accesskey="s">Foo</button>
+      <button id="accesskeys2" accesskey="s">Bar</button>
+    </section>
```
brendankenny: 👍 👍 thanks for adding this!

PR #7132, opened 2019-02-01
core(metrics): do not fail if TTI fails

patrickhulce:
This one's for you @benschwarz


benschwarz:
Big ❤️, thank you @patrickhulce

PR #7154, opened 2019-02-05
core(tsc): begin clean up of audit details types

brendankenny:
wading through #6999 makes it clear that our `details` type situation is approaching untenable :)

This starts us off, unifying several top-level details types under `LH.Audit.Details`. There are no functional changes, just types, so this is just describing what we're already doing in a slightly different way (hopefully better).

This is the cleanest set of things I could find to do before starting to need code changes, so we'll build from here :)

Happy to bikeshed on names

### DIFF
--- a/lighthouse-core/test/report/html/renderer/crc-details-renderer-test.js
+++ b/lighthouse-core/test/report/html/renderer/crc-details-renderer-test.js
@@ -23,7 +23,6 @@ const superLongURL =
     'https://example.com/thisIsASuperLongURLThatWillTriggerFilenameTruncationWhichWeWantToTest.js';
 const DETAILS = {
   type: 'criticalrequestchain',
-  header: {type: 'text', text: 'CRC Header'},
   chains: {
     0: {
       request: {
--- a/types/audit-details.d.ts
+++ b/types/audit-details.d.ts
@@ -1,0 +1,71 @@
+/**
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+
+declare global {
+  module LH.Audit.Details {
+    export interface Filmstrip {
+      type: 'filmstrip';
+      scale: number;
+      items: {
+        timing: number;
+        timestamp: number;
+        data: string;
+      }[];
+    }
+
+    export interface Screenshot {
+      type: 'screenshot';
+      timestamp: number;
+      data: string;
+    }
+
+    export interface Opportunity {
+      type: 'opportunity';
+      overallSavingsMs: number;
+      overallSavingsBytes?: number;
+      headings: TableColumnHeading[];
+      items: OpportunityItem[];
+    }
+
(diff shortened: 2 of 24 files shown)


lighthouse-core/test/report/html/renderer/crc-details-renderer-test.js
```diff
@@ -23,7 +23,6 @@ const superLongURL =
     'https://example.com/thisIsASuperLongURLThatWillTriggerFilenameTruncationWhichWeWantToTest.js';
 const DETAILS = {
   type: 'criticalrequestchain',
-  header: {type: 'text', text: 'CRC Header'},
```
brendankenny: only executable code change (though in a test). We haven't done anything with a `header` field in I don't know how long, though, so might as well delete it today :)

types/lhr-lite.d.ts
```diff
@@ -70,49 +70,10 @@ declare global {
         /** An explanation of audit-related issues encountered on the test page. */
         explanation?: string;
         /** Extra information provided by some types of audits. */
-        details?: Audit.MetricDetails | Audit.OpportunityDetails;
+        details?: never;
```
brendankenny: I'm going to start deleting `lhr-lite-d.ts` since its use didn't pan out, so starting by snipping off everything starting here

types/audit-details.d.ts
```diff
@@ -0,0 +1,71 @@
...
+      };
+      chains: Audit.SimpleCriticalRequestNode;
+    }
+
+    // TODO(bckenny)
+    // export interface MultiCheck {}
+    // export interface Table {}
+    // export interface SomeKindOfHiddenThing
+
+    // Contents of details below here
```
brendankenny: maybe we want a different namespace here, not sure. These are things that are only inside details (table headings, table items).

types/audit-details.d.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+declare global {
+  module LH.Audit.Details {
+    export interface Filmstrip {
+      type: 'filmstrip';
+      scale: number;
+      items: {
+        timing: number;
+        timestamp: number;
```
patrickhulce: wanna throw a comment on these bad boys for filmstrip? we were inconsistent when we added the full page screenshot so documenting what they represent seems like a decent idea :)

types/audit-details.d.ts
```diff
@@ -0,0 +1,71 @@
...
+        length: number;
+        transferSize: number;
+      };
+      chains: Audit.SimpleCriticalRequestNode;
+    }
+
+    // TODO(bckenny)
+    // export interface MultiCheck {}
+    // export interface Table {}
+    // export interface SomeKindOfHiddenThing
```
patrickhulce: these intentionally left in here? 

these would be like a foreign language to me if I were ever tasked with following up maybe we could flesh out the comment a tad? :D

types/audit-details.d.ts
patrickhulce: `LH.Audit.Details.Details` ;)

I'd personally be in favor of naming convention over a different namespace. they categorically they definitely belong under "details". maybe we return to `*Details` suffix if these get out of hand?

brendankenny: > I'd personally be in favor of naming convention over a different namespace. they categorically they definitely belong under "details". maybe we return to `*Details` suffix if these get out of hand?

yeah, I'm ok with living with this for now, just trying to think of ways to shrink the amount of thinking one has to do. Part of this was motivated by reviewing #6999 and seeing some of our existing types like `DetailsRendererDetailsJSON` and `DetailsRendererNodeDetailsJSON` and going "wat" :)

types/audit-details.d.ts
brendankenny: > wanna throw a comment on these bad boys for filmstrip

done

PR #7177, opened 2019-02-07
core(tsc): fix audit details type hierarchy

brendankenny:
Finally maps out types for the `details` objects!

This PR was already getting long enough, so this (mostly) only touches the production of `details` object in audits. I'll follow up with a change to make `details-renderer.js` and etc use these types. I already have that working in a branch, so I promise it will work on that side as well :)

For the major thrust of things see `audit-details.d.ts` (and what was deleted from `audit.d.ts`). Everything else is mostly just switching over to use the types, not changes to the audits themselves. 

The new `details` structure just describes what we currently do: there are a few details object shapes that we use directly (`criticalrequestchain`, `filmstrip`, `opportunity`, `screenshot`, `table`, and now `diagnostic`), then a bunch of things that are only ever used inside those details (`text`, `url`, 'code', etc). If and when we need some of those `code` or whatever as top-level details, it's easy to add them to the union of `details` output by audits, just like it's easy to add additional properties.

There are two relatively small changes to audit output:
- no more `null` in `details`. This was a minor thing, but we only used it in two non-visible audits, 
- as discussed in chat, a new `'diagnostic'` type. There are plenty of places you can still stick values that aren't visible in the final report in `details` (like unreferenced properties on table `items`), but this is the place to stick full-on objects or entire `details` objects like `metrics.js` or `diagnostics.js`. It's an explicit signal that a property won't be in the html report, and it's a way of keeping the `details` types relatively narrow so that type checking is useful (you can still put anything in them, but they just have to be labeled with `type: 'diagnostic'`).

  In practice the effect of both are relatively minor (at most you get some `diagnostic` data one level deeper in the a11y audits). Check out the changes to `sample_v2.json` for all the LHR changes that occur.

### DIFF
--- a/lighthouse-core/audits/accessibility/axe-audit.js
+++ b/lighthouse-core/audits/accessibility/axe-audit.js
@@ -44,11 +44,11 @@ class AxeAudit extends Audit {
     const impact = rule && rule.impact;
     const tags = rule && rule.tags;
 
-    /** @type {Array<{node: LH.Audit.DetailsRendererNodeDetailsJSON}>} */
+    /** @type {LH.Audit.Details.Table['items']}>} */
     let items = [];
     if (rule && rule.nodes) {
       items = rule.nodes.map(node => ({
-        node: /** @type {LH.Audit.DetailsRendererNodeDetailsJSON} */ ({
+        node: /** @type {LH.Audit.Details.NodeValue} */ ({
           type: 'node',
           selector: Array.isArray(node.target) ? node.target.join(' ') : '',
           path: node.path,
@@ -58,16 +58,27 @@ class AxeAudit extends Audit {
       }));
     }
 
+    /** @type {LH.Audit.Details.Table['headings']} */
     const headings = [
       {key: 'node', itemType: 'node', text: str_(UIStrings.failingElementsHeader)},
     ];
+
+    /** @type {LH.Audit.Details.Diagnostic|undefined} */
+    let diagnostic;
+    if (impact || tags) {
+      diagnostic = {
+        type: 'diagnostic',
+        impact,
+        tags,
+      };
+    }
 
     return {
       rawValue: typeof rule === 'undefined',
       extendedInfo: {
         value: rule,
       },
-      details: {...Audit.makeTableDetails(headings, items), impact, tags},
+      details: {...Audit.makeTableDetails(headings, items), diagnostic},
     };
   }
 }
--- a/lighthouse-core/audits/bootup-time.js
+++ b/lighthouse-core/audits/bootup-time.js
@@ -159,6 +159,7 @@ class BootupTime extends Audit {
 
     const summary = {wastedMs: totalBootupTime};
 
+    /** @type {LH.Audit.Details.Table['headings']} */
     const headings = [
       {key: 'url', itemType: 'url', text: str_(i18n.UIStrings.columnURL)},
       {key: 'total', granularity: 1, itemType: 'ms', text: str_(UIStrings.columnTotal)},
--- a/lighthouse-core/audits/diagnostics.js
+++ b/lighthouse-core/audits/diagnostics.js
@@ -67,7 +67,11 @@ class Diagnostics extends Audit {
     return {
       score: 1,
       rawValue: 1,
-      details: {items: [diagnostics]},
+      details: {
+        type: 'diagnostic',
+        // TODO: Consider not nesting diagnostics under `items`.
+        items: [diagnostics],
+      },
     };
   }
 }
--- a/lighthouse-core/audits/multi-check-audit.js
+++ b/lighthouse-core/audits/multi-check-audit.js
@@ -11,6 +11,15 @@
 
 const Audit = require('./audit');
 
+/** @typedef {
+  Partial<Record<LH.Artifacts.ManifestValueCheckID, boolean>> &
+  Partial<LH.Artifacts.ManifestValues> & {
+    failures: Array<string>;
+    manifestValues?: undefined;
+    allChecks?: undefined;
+  }} MultiCheckDiagnosticDetails
+ */
+
 class MultiCheckAudit extends Audit {
   /**
    * @param {LH.Artifacts} artifacts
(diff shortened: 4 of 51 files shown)


types/audit-details.d.ts
```diff
@@ -5,63 +5,166 @@
...
-    }
+      export interface Table {
+        type: 'table';
+        headings: TableColumnHeading[];
+        items: TableItem[];
+        summary?: {
+          wastedMs?: number;
+          wastedBytes?: number;
+        };
+        diagnostic?: Diagnostic;
```
brendankenny: turns out I couldn't do `[p: string]: Diagnostic` here because that means *any* string accessor of `Table` should return a value assignable to `Diagnostic`, which we don't want. So this is a middle ground, and seems flexible enough (@exterkamp rightly pointed out that if a `Diagnostic` object can contain anything, why would you ever need two of them as properties on the same object?)

brendankenny: being able to define a set of properties with certain types and a different type for every other property not explicitly defined will be possible if/when negated types land in tsc (https://github.com/Microsoft/TypeScript/pull/29317)

types/audit-details.d.ts
```diff
@@ -5,63 +5,166 @@
...
-        length: number;
-        transferSize: number;
-      };
-      chains: Audit.SimpleCriticalRequestNode;
-    }
+      /** Possible types of values found within table items. */
+      type ItemValueTypes = 'bytes' | 'code' | 'link' | 'ms' | 'node' | 'numeric' | 'text' | 'thumbnail' | 'timespanMs' | 'url';
 
-    // Contents of details below here
+      // TODO(bckenny): unify Table/Opportunity headings and items on next breaking change.
```
brendankenny: `TableColumnHeading` and `OpportunityColumnHeading` are very close but not close enough. I think we just didn't get around to making them the same :S

Something for v5 :)

lighthouse-core/audits/bootup-time.js
```diff
@@ -159,6 +159,7 @@ class BootupTime extends Audit {
 
     const summary = {wastedMs: totalBootupTime};
 
+    /** @type {LH.Audit.Details.Table['headings']} */
```
brendankenny: had to add a ton of these explicit `headings` type lines because the heading `itemType` now has a more stringent type (union of literals instead of just `string`) which won't work if it's been widened. There's an upcoming tsc 3.4 feature (`as const`) that will make this a bit easier all over the codebase, but this doesn't seem too bad for now

lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -58,16 +58,27 @@ class AxeAudit extends Audit {
...
     }
 
+    /** @type {LH.Audit.Details.Table['headings']} */
     const headings = [
       {key: 'node', itemType: 'node', text: str_(UIStrings.failingElementsHeader)},
     ];
 
+    /** @type {LH.Audit.Details.Diagnostic|undefined} */
+    let diagnostic;
+    if (impact || tags) {
```
brendankenny: left undefined if these weren't defined to match current LHR behavior

lighthouse-core/audits/diagnostics.js
```diff
@@ -67,7 +67,11 @@ class Diagnostics extends Audit {
     return {
       score: 1,
       rawValue: 1,
-      details: {items: [diagnostics]},
+      details: {
+        type: 'diagnostic',
+        // TODO: Consider not nesting diagnostics under `items`.
+        items: [diagnostics],
```
patrickhulce: awww this is just to remain non-breaking? :)

the idea would be `details: {type: 'diagnostic', ...diagnostics}` if we were writing it today right?

brendankenny: > awww this is just to remain non-breaking?

Ha, well, since we just added this and we consider it internal only, I'm happy to change, I just didn't want to assume and/or cause churn in people's lives, especially because i mostly don't use these :)

We could change these newer ones and leave the older ones like `metrics` until v5 since we know of at least a few people who rely on them.

patrickhulce: I meant "awwww" as in how thoughtful :D I'm probably the only person that relies on this and that would indeed be a pain, lol

I like leaving them until v5 :)

brendankenny: haha, got it. SGTM :)

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -160,8 +160,9 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
     const thumbnailResult = thumbnailAudit && thumbnailAudit.result;
     if (thumbnailResult && thumbnailResult.details) {
       timelineEl.id = thumbnailResult.id;
+      // @ts-ignore TODO(bckenny): fix detailsRenderer.render input type
```
patrickhulce: input type? output type looks busted too based on the next line change 😆

lighthouse-core/audits/multi-check-audit.js
```diff
@@ -11,6 +11,15 @@
 
 const Audit = require('./audit');
 
+/** @typedef {
+  Partial<Record<LH.Artifacts.ManifestValueCheckID, boolean>> &
```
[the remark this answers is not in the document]
brendankenny: > Is there a long-term plan here with the multi-check audits?
> I'm not sure the js typedefs are a readability win in this step.

ah, now that you say that, there's no reason it can't keep living in `audit.d.ts`. I was deleting all the `details` types from there, but there's no reason to consider it a `details` type now, it's just a handy typedef. I'll put it back.

(personally I hate multi-check-audit and hope it dies as a base class someday, but that's more the long-long-term plan :)

lighthouse-core/report/html/renderer/performance-category-renderer.js
brendankenny: > input type? output type looks busted too based on the next line change

haha, maybe both then

PR #7184, opened 2019-02-07
core: suppress protocol timeout for app manifest bug in LR

### LINKED ISSUE #7147: PROTOCOL_TIMEOUT (Method: Page.getAppManifest)
1. Run LH on `https://seo4.net` or `https://krakowskakancelaria.pl/`

Error: Lighthouse returned error: PROTOCOL_TIMEOUT. Waiting for DevTools protocol response has exceeded the allotted time. (Method: Page.getAppManifest) (PROTOCOL_TIMEOUT)

No errors.

#### Environment Information
* Affected Channels: Pagespeed API returns 500, web.dev and pagespeedinsights return issue.
* Lighthouse version: ?
* Node.js version: ?
* Operating System: ?

#### Extra info
Issues started to occur since thursday, no action were performed on website, we tried modifying appmanifest but it doesn't seem to be the issue as it is valid and working.
~100 other websites we are tracking regularly do have manifest and they work fine.

Hoten:
Fixes #7147

See b/124008171.

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -407,7 +407,7 @@ class Driver {
       if (err.code === 'PROTOCOL_TIMEOUT') {
         // LR will timeout fetching the app manifest in some cases, move on without one.
         // https://github.com/GoogleChrome/lighthouse/issues/7147#issuecomment-461210921
-        log.error('Driver', 'Failed fetching...', err);
+        log.error('Driver', 'Failed fetching manifest', err);
         return null;
       }
 
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -400,7 +400,16 @@ class Driver {
    */
   async getAppManifest() {
     this.setNextProtocolTimeout(3000);
-    const response = await this.sendCommand('Page.getAppManifest');
+    let response;
+    try {
+      response = await this.sendCommand('Page.getAppManifest');
+    } catch (err) {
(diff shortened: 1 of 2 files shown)


brendankenny:
> WDYT about just catching this in driver's `getAppManifest` instead?

what should it return, though? Ideally you'd get a "timed out waiting for the manifest" explanation in the report, and not just "you didn't have a manifest".

brendankenny:
> Yeah that would be better, but I think we want to deploy something soon...Are we okay with this hotfix for the LR fix, and following up with a more thorough workaround later?

sorry, yes, I definitely agree with that. I guess just anticipating the follow up discussion :)

lighthouse-core/gather/driver.js
```diff
@@ -400,7 +400,20 @@ class Driver {
...
     this.setNextProtocolTimeout(3000);
-    const response = await this.sendCommand('Page.getAppManifest');
+    let response;
+    try {
+      response = await this.sendCommand('Page.getAppManifest');
+    } catch (err) {
+      if (err.code === 'PROTOCOL_TIMEOUT') {
+        // LR will timeout fetching the app manifest in some cases, move on without one.
+        // https://github.com/GoogleChrome/lighthouse/issues/7147#issuecomment-461210921
+        log.error('Driver', 'Failed fetching...', err);
```
brendankenny: ha, sorry, I meant ... for the rest of the message you had before, I just didn't want to retype it :)

PR #7192, opened 2019-02-09
report(details-renderer): use new details types

brendankenny:
Uses the new audit details types from #7177 in `DetailsRenderer`, finally deleting all those super old (and confusing) typedefs. Internal functionality is changed, but the generated DOM should be exactly the same.

Majority of the PR is tests because `details-renderer-test.js` was really really light before this :)

Main things were
- finally completing an old TODO to merge `opportunity` and `table` rendering (since they're both really tables). This required a `_getCanonicalizedTableHeadings` method because they have different `heading` formats (which we can hopefully fix next breaking version).
- deleting a bunch of cases from the `DetailsRenderer.render()` switch statement reflecting the [`audit-details.d.ts`](https://github.com/GoogleChrome/lighthouse/blob/c79e3579cbcf214f29680a463be7162e7bdc89fe/types/audit-details.d.ts) type structure that has a [few root `details` types](https://github.com/GoogleChrome/lighthouse/blob/c79e3579cbcf214f29680a463be7162e7bdc89fe/types/audit-details.d.ts#L9-L15), with all the common types (`text`, `url`, etc) within details `items`. These are now rendered in `_renderTableValue()`.

As mentioned in #7177, there's nothing special about which types fall into `render` or `_renderTableValue`, this organization just keeps things narrowed to how we actually use these types. If at some point we need `code` to be directly used, it's as simple as adding `LH.Audit.Details.Code` to the `LH.Audit.Details` type union and adding `_renderCode()` back to the `DetailsRenderer.render()` switch.

### DIFF
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -19,9 +19,7 @@
 /* globals self CriticalRequestChainRenderer SnippetRenderer Util URL */
 
 /** @typedef {import('./dom.js')} DOM */
-/** @typedef {LH.Audit.Details.Opportunity} OpportunityDetails */
 
-/** @type {Array<string>} */
 const URL_PREFIXES = ['http://', 'https://', 'data:'];
 
 class DetailsRenderer {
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -43,52 +41,29 @@ class DetailsRenderer {
   }
 
   /**
-   * @param {DetailsJSON|OpportunityDetails} details
+   * @param {LH.Audit.Details} details
```
exterkamp: Add a comment for what this render function does.  i.e. render the top level categories of details.

types/audit.d.ts
```diff
@@ -100,7 +100,6 @@ declare global {
       id: string;
       /** A more detailed description that describes why the audit is important and links to Lighthouse documentation on the audit; markdown links supported. */
       description: string;
-      // TODO(bckenny): define details
```
exterkamp: 🎉

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -43,52 +41,32 @@ class DetailsRenderer {
...
-        return this._renderNumeric(/** @type {StringDetailsJSON} */ (details));
+        return this._renderTable(details);
 
       // Internal-only details, not for rendering.
       case 'screenshot':
       case 'diagnostic':
         return null;
+      // Fallback for old LHRs, where no type meant don't render.
+      case undefined:
+        return null;
```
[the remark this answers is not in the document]
brendankenny: > should we console.error or something?

I don't think it matters. This is for loading < 4.2 or whatever LHRs in the viewer. Things like `diagnostics.js` and `metrics.js` didn't have a `type` before, now they do (`type: 'diagnostic'`), so just need an explicit handling of the no-`type` type.

Issue #7158, opened 2019-02-05
Shows "Does not use HTTP/2 for all of its resources" tip when files are loaded from service worker via http/1.1 protocol


asistapl:
#### Provide the steps to reproduce
1. Run LH on https://www.anthonydesigner.com/

#### What is the current behavior?
All files are served with H2 initially but it still shows "Does not use HTTP/2 for all of its resources" tip because files are loaded from service worker via http/1.1 protocol.

#### What is the expected behavior?
I think it should show that tip only if files aren't served with H2 initially, not when service worker is used.

#### Environment Information
* Affected Channels: DevTools
* Lighthouse version: 3.2.0
* Operating System: Mac OSX

wardpeet:
related issue https://github.com/gatsbyjs/gatsby/issues/11123

I'm putting up a PR for it.

wardpeet:
@asistapl PR is up might giving this a spin yourself?

asistapl:
@wardpeet I launched cli from "bug/sw-h2" branch and it works for me! Thank you 🎉

paulyabsley:
@patrickhulce Do you know when this bug fix will be released?

PR #7197, opened 2019-02-09
fix(uses-http2): ignore requests from service-worker

wardpeet:
**Summary**
Requests from service workers are returned with a http/1.1 protocol so if a service worker is loaded on time and calls the claim function we might route some functions through the SW before we stop listening to the page. This can happen with for example `<link rel="fetch" ../>`

An example that fails regularly is https://remarkable-airboat.glitch.me
image

**Related Issues/PRs**
fixes: #7158


lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
```diff
@@ -61,4 +61,21 @@ describe('Resources are fetched over http/2', () => {
       assert.ok(auditResult.displayValue === '');
     });
   });
+
+  it('passes when all resources were requested via service worker', () => {
```
exterkamp: Should we also get a test for failing with mixed records. i.e. only convert 1 or 2 1.x requests to serviceWorkers so that it isn't all serviceWorker requests (so it fails), so that it's more realistic.

wardpeet: sure I just looked at the test of h2 and copied it, I can scope to 1 or 2 :)

lighthouse-core/audits/dobetterweb/uses-http2.js
```diff
@@ -44,6 +44,9 @@ class UsesHTTP2Audit extends Audit {
       const seenURLs = new Set();
       // Filter requests that are on the same host as the page and not over h2.
       const resources = networkRecords.filter(record => {
+        // check if record is not served through the service worker, servicer worker uses http/1.1 as a protocol
+        // these can generate false positives (bug: #7158)
```
patrickhulce: mind replacing `#7158` with the full link for easy clicking in text editors? :)

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
patrickhulce: to @exterkamp's point, the tests are currently passing yet our test case is mutating `fromServiceWorker` and our code is checking `fetchedViaServiceWorker`, so something isn't quite right :)

wardpeet: `fromServiceWorker` is what devtools has in it's network trace but we mutate it to fetchedViaServiceWorker. 

see https://github.com/GoogleChrome/lighthouse/blob/1098f952d5c75b22668aa572532415f81d2817af/lighthouse-core/lib/network-request.js#L260

it's also from devtools:
https://github.com/ChromeDevTools/devtools-frontend/blob/ecf32a28e6156b8a2c9dc025faa9315ec8652e6a/front_end/sdk/NetworkManager.js#L337-L338

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
```diff
@@ -61,4 +61,22 @@ describe('Resources are fetched over http/2', () => {
       assert.ok(auditResult.displayValue === '');
     });
   });
+
+  it('results are correct when some requests are handled by service worker', () => {
+    const clonedNetworkRecords = JSON.parse(JSON.stringify(networkRecords));
+    clonedNetworkRecords.forEach(record => {
+      // convert http 1.x to service worker requests
+      if (record.protocol === 'http/1.1') {
+        record.fromServiceWorker = true;
```
patrickhulce: @wardpeet think you could fix the bug over in network-records-to-devtools-log? it's reading the wrong property name when doing the `networkRecordsToDevtoolsLog` round-trip :)

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
```diff
@@ -61,4 +61,22 @@ describe('Resources are fetched over http/2', () => {
...
+      }
+    });
+
+    return UsesHTTP2Audit.audit(
+      getArtifacts(clonedNetworkRecords, URL), {computedCache: new Map()}
+    ).then(auditResult => {
+      assert.equal(auditResult.rawValue, false);
+      assert.ok(auditResult.displayValue.match('1 request not'));
+      expect(auditResult.details).toMatchSnapshot();
```
patrickhulce: I do ❤️ me some `toMatchSnapshot` :D 

Since this test doesn't really have any impact on anything but the items though, I'd want to limit it to that.

Aside, up to this point, all uses of snapshot testing expect for 1 are uses where there's arguably not a permanent correct answer and/or we update the values frequently. When it's essentially used in place of `.toEqual`, I think it should be kept to `InlineSnapshot`. I believe @brendankenny feels differently about these sorts of uses though 😉

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
patrickhulce: ahhhhhhh that makes more sense it's doing the devtoolslog roundtrip!

@wardpeet think you could fix the bug over in network-records-to-devtools-log? it's reading the wrong property name when doing the `networkRecordsToDevtoolsLog` round-trip :)

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
wardpeet: I remove the snapshot as it wasn't a big deal here :)

wardpeet:
@patrickhulce @exterkamp  I rebooted travis, PR should be good to go :)

lighthouse-core/test/audits/dobetterweb/uses-http2-test.js
```diff
@@ -61,4 +61,22 @@ describe('Resources are fetched over http/2', () => {
...
+        record.fetchedViaServiceWorker = true;
+      }
+    });
+
+    return UsesHTTP2Audit.audit(
+      getArtifacts(clonedNetworkRecords, URL), {computedCache: new Map()}
+    ).then(auditResult => {
+      assert.equal(auditResult.rawValue, false);
+      assert.ok(auditResult.displayValue.match('1 request not'));
+      assert.equal(auditResult.details.items[0].url, 'https://webtide.com/wp-content/themes/clean-retina-pro/library/js/tinynav.js?ver=4.5.4');
```
patrickhulce: why doesn't this one go away too? is it not `http/1.1`?

patrickhulce: friendly ping :)

wardpeet: will have a look today :)

wardpeet:
```js
{
  protocol: 'http/1.0',
  url: 'https://webtide.com/wp-content/themes/clean-retina-pro/library/js/tinynav.js?ver=4.5.4'
}
```
Protocol is http/1.0 not http/1.1 so should be flagged.

patrickhulce: oh gotcha, can we leave a comment so it's explained or just convert the test to flip all to sw or something? :)

wardpeet: i'll add a comment, not adding all requests to sw was added because of this comment:
https://github.com/GoogleChrome/lighthouse/pull/7197#discussion_r255352918

patrickhulce: Oh my b, yeah comment is ace 👌

svewag:
Hey there 👋 

this is now fixed in v4.3.0, right?

Hoten:
[correct](https://github.com/GoogleChrome/lighthouse/blob/master/changelog.md#430-2019-04-08). are you seeing otherwise?

svewag:
No, I was just wondering because the issue is still not closed 😉

paulirish:
@svewag good catch. i just closed #7383 after verifying that it was fixed by this PR.

PR #7212, opened 2019-02-11
tests: re-organize driver tests by method

patrickhulce:
**tl;dr - for EASIER REVIEWING just look at https://github.com/GoogleChrome/lighthouse/pull/7212/commits/877264306c5351120c3a79c69f4e2609a6a62d80**

**Summary**
It's no secret that driver tests have been somewhat of a mess for a long-time. Driver is one of the most critical pieces of Lighthouse logic and yet adding tests has always been intimidating and somewhat unconvincing that things are working. This PR is a first step in an attempt to change that.

**The Changes**

- Organizes all the tests by the method. Already much easier to see where we have coverage and where we're lacking.
- Minor renaming of driverStub -> driver. The driver here is a real driver, its the connection and sendCommand functions that are stubs.
- Moves to a more readable format where each method stubs the `sendCommand` methods that it actually uses. This should hopefully make it much easier and less magical for contributors to understand what each method on driver requires, how to add tests that use new protocol methods, etc. A nice side effect is decent documentation about what each protocol method will actually do.

image

**Related Issues/PRs**
#7174,  didn't want to proceed with more major gotoURL without the promised cleanup and testing first

### DIFF
--- a/lighthouse-core/test/gather/driver-test.js
+++ b/lighthouse-core/test/gather/driver-test.js
@@ -122,37 +122,37 @@ function sendCommandStub(command, params) {
 
 /* eslint-env jest */
 
-let driverStub;
+let driver;
 let connectionStub;
 beforeEach(() => {
   connectionStub = new Connection();
   connectionStub.sendCommand = sendCommandStub;
-  driverStub = new Driver(connectionStub);
+  driver = new Driver(connectionStub);
   sendCommandParams = [];
   sendCommandMockResponses.clear();
 });
 
 describe('.querySelector(All)', () => {
   it('returns null when DOM.querySelector finds no node', () => {
-    return driverStub.querySelector('invalid').then(value => {
+    return driver.querySelector('invalid').then(value => {
       assert.equal(value, null);
     });
   });
 
   it('returns element when DOM.querySelector finds node', () => {
-    return driverStub.querySelector('meta head').then(value => {
+    return driver.querySelector('meta head').then(value => {
       assert.equal(value instanceof Element, true);
     });
   });
 
   it('returns [] when DOM.querySelectorAll finds no node', () => {
-    return driverStub.querySelectorAll('invalid').then(value => {
+    return driver.querySelectorAll('invalid').then(value => {
       assert.deepEqual(value, []);
     });
   });
 
   it('returns element when DOM.querySelectorAll finds node', () => {
-    return driverStub.querySelectorAll('a').then(value => {
+    return driver.querySelectorAll('a').then(value => {
       assert.equal(value.length, 1);
       assert.equal(value[0] instanceof Element, true);
     });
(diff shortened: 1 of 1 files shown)


lighthouse-core/test/gather/driver-test.js
```diff
@@ -30,37 +27,30 @@ function createOnceStub(events) {
...
+  return connectionStub.sendCommand.mock.calls.find(call => call[0] === command)[1];
 }
 
-function createActiveWorker(id, url, controlledClients, status = 'activated') {
-  return {
-    registrationId: id,
-    scriptURL: url,
-    controlledClients,
-    status,
+function createSendCommandMockFn() {
```
brendankenny: docs would be good here too (e.g. the returned function has a `mockSendCommandResponse` on it, use it to push responses, can be used multiple times, where each call will get the next one, etc

lighthouse-core/test/gather/driver-test.js
```diff
@@ -30,37 +27,30 @@ function createOnceStub(events) {
...
-function sendCommandStub(command, params) {
-  sendCommandParams.push({command, params});
-
-  if (sendCommandMockResponses.has(command)) {
-    const mockResponse = sendCommandMockResponses.get(command);
-    sendCommandMockResponses.delete(command);
-    return Promise.resolve(mockResponse);
-  }
-
+function sendCommandOldStub(command, params) {
```
brendankenny: keeping this because it'd be too much churn to remove it and break up across tests?

lighthouse-core/test/gather/driver-test.js
```diff
@@ -78,14 +68,20 @@ function sendCommandStub(command, params) {
...
-        result: params.objectId === 'invalid' ? [] : [{
-          name: 'test',
-          value: {
-            value: '123',
-          },
-        }, {
-          name: 'novalue',
-        }],
+        result:
+          params.objectId === 'invalid'
```
brendankenny: prefer the old formatting :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -30,37 +27,30 @@ function createOnceStub(events) {
...
 }
 
-function createSWRegistration(id, url, isDeleted) {
-  return {
-    isDeleted: !!isDeleted,
-    registrationId: id,
-    scopeURL: url,
-  };
+function findSendCommandInvocation(command) {
+  expect(connectionStub.sendCommand).toHaveBeenCalledWith(command, expect.anything());
```
brendankenny: it's a little weird to me that it's using the assumed global `connectionStub` here (while `createSendCommandMockFn` is all encapsulated), could we put this as another method on the return value of `createSendCommandMockFn()`? Or have the particular `sendCommand` method passed in

lighthouse-core/test/gather/driver-test.js
patrickhulce: yeah I had to draw the line somewhere for the first pass. half the tests still need `sendCommandOldStub` but absolutely the idea is to migrate the rest of them too

lighthouse-core/test/gather/driver-test.js
patrickhulce: reverted

lighthouse-core/test/gather/driver-test.js
patrickhulce: yeah sounds good to add to the mock function! 👍 done

lighthouse-core/test/gather/driver-test.js
```diff
@@ -30,37 +27,30 @@ function createOnceStub(events) {
   };
 }
 
-function createSWRegistration(id, url, isDeleted) {
-  return {
-    isDeleted: !!isDeleted,
-    registrationId: id,
-    scopeURL: url,
-  };
+function findSendCommandInvocation(command) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/gather/driver-test.js
patrickhulce: done

patrickhulce:
> Would be great to get rid of it and the long list of commands it's not clear which one goes to which test :)

You cool with following up with this and the rest of the driver cleanup in separate PR?

patrickhulce:
^^ @brendankenny that one was for you :)

PR #7217, opened 2019-02-12
tests: covert only non-volatile snapshot usage

patrickhulce:
**Summary**
In #7197 I realized I had introduced the only non-volatile snapshot usage into the codebase. @brendankenny I felt bad, so here we go :)


brendankenny:
> but I have no idea what "non-volatile" means in this context

oh, ha, I see the https://github.com/GoogleChrome/lighthouse/pull/7197#discussion_r255706473 now, got it :)

PR #6860, opened 2018-12-21
i18n: seo

exterkamp:
**Summary**
:us: :gb: :de: :india: :canary_islands: :mexico: 
i18n for SEO category audits.

If you have opinions on the i18n description text please comment, I am not the best at explaining things in those also spelling is hard.

PR #7223, opened 2019-02-12
core(mobile-friendly): Remove manual mobile friendly audit

mattzeunert:
**Summary**

Now that tap targets is merged we don't need need the manual mobile friendly check audit any more.


brendankenny:
since @exterkamp is ok with fixing #6860 again, let's land :)

PR #7227, opened 2019-02-12
tests(smokehouse): fail on finalUrl/errorCode mismatches

### LINKED ISSUE #7225: Smokehouse tests do not fail on finalUrl/errorCode mismatches
```
diff --git a/lighthouse-cli/test/smokehouse/pwa-expectations.js b/lighthouse-cli/test/smokehouse/pwa-expectations.js
index 9db56360..3ebd2c37 100644
--- a/lighthouse-cli/test/smokehouse/pwa-expectations.js
+++ b/lighthouse-cli/test/smokehouse/pwa-expectations.js
@@ -24,7 +24,7 @@ const pwaDetailsExpectations = {
 module.exports = [
   {
     requestedUrl: 'https://airhorner.com',
-    finalUrl: 'https://airhorner.com/',
+    finalUrl: 'https://airhorner.com/asdasdasdas',
     audits: {
       'is-on-https': {
         score: 1,

```

image

Note that the status code is 0.

This is due to the way errors are counted in `smokehouse.js report()`.

Hoten:
Fixes #7225.

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -311,11 +311,23 @@ function reportAssertion(assertion) {
  * @return {{passed: number, failed: number}}
  */
 function report(results) {
-  reportAssertion(results.finalUrl);
-  reportAssertion(results.errorCode);
-
   let correctCount = 0;
   let failedCount = 0;
+
+  reportAssertion(results.finalUrl);
+  if (results.finalUrl.equal) {
+    correctCount++;
+  } else {
+    failedCount++;
+  }
+
+  reportAssertion(results.errorCode);
+  if (results.errorCode.equal) {
+    correctCount++;
+  } else {
+    failedCount++;
+  }
+
   results.audits.forEach(auditAssertion => {
     if (auditAssertion.equal) {
       correctCount++;
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -266,7 +266,7 @@ function collateResults(actual, expected) {
 }
 
 /**
- * @param {*} obj
+ * @param {unknown} obj
  */
 function isPlainObject(obj) {
   return Object.prototype.toString.call(obj) === '[object Object]';
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -266,6 +266,13 @@ function collateResults(actual, expected) {
 }
 
 /**
+ * @param {*} obj
+ */
+function isPlainObject(obj) {
+  return Object.prototype.toString.call(obj) === '[object Object]';
+}
+
+/**
  * Log the result of an assertion of actual and expected results.
  * @param {Comparison} assertion
  */
@@ -277,7 +284,7 @@ function reportAssertion(assertion) {
   RegExp.prototype.toJSON = RegExp.prototype.toString;
 
   if (assertion.equal) {
-    if (assertion.actual !== null && typeof assertion.actual === 'object') {
+    if (isPlainObject(assertion.actual)) {
       console.log(`  ${log.greenify(log.tick)} ${assertion.category}`);
     } else {
       console.log(`  ${log.greenify(log.tick)} ${assertion.category}: ` +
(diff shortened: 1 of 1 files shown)


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -311,11 +311,23 @@ function reportAssertion(assertion) {
...
   reportAssertion(results.errorCode);
+  if (results.errorCode.equal) {
+    correctCount++;
+  } else {
+    failedCount++;
+  }
 
-  let correctCount = 0;
-  let failedCount = 0;
   results.audits.forEach(auditAssertion => {
```
paulirish:
```suggestion
  [results.finalUrl, results.errorCode, ...results.audits].forEach(auditAssertion => {
```

paulirish: could do this instead and drop the two reportAssertions above.

Hoten: nice

do we want to continue showing green ticks for passing? this is what it looks like  with those changes + moving `reportAssertion` out of the if statement:

image


Keeping `reportAssertion` where it is-

image

Hoten: i like this best

image

patrickhulce:
I kinda like the verbose version. My most common use case when actually running smoketests is picking a particular smoke test and running it frequently so I kinda like the ability to see what's passing. I have no strong feelings on the length of passing CI logs since those are collapsed for me anyhow :)

Hoten:
`--verbose` it is!

Hoten:
Went with an env flag since wiring an option would be more work (see how run-smokehouse parses argv).

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -35,6 +35,7 @@ const PAGE_HUNG_EXIT_CODE = 68;
 const INSECURE_DOCUMENT_REQUEST_EXIT_CODE = 69;
 const RETRIES = 3;
 const NUMERICAL_EXPECTATION_REGEXP = /^(<=?|>=?)((\d|\.)+)$/;
+const VERBOSE = process.env.LH_SMOKE_VERBOSE === '1';
```
patrickhulce: WDYT about just `Boolean(process.env.LH_SMOKE_VERBOSE)`?

patrickhulce: later down we use a env var `SMOKEHOUSE_DEBUG` maybe we should unify these two 🤔 

maybe we rename that one to `LH_SMOKE_DEBUG`?

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -276,8 +277,12 @@ function reportAssertion(assertion) {
   RegExp.prototype.toJSON = RegExp.prototype.toString;
 
   if (assertion.equal) {
-    console.log(`  ${log.greenify(log.tick)} ${assertion.category}: ` +
-        log.greenify(assertion.actual));
+    if (assertion.actual !== null && typeof assertion.actual === 'object') {
```
patrickhulce: super nit but mind flipping the order of these to `is an object && is not null`? took me a bit to realize this was just an object check

paulirish: can we just use a helper method like so?  

```js
const isPlainObject = function (obj) {
	return Object.prototype.toString.call(obj) === '[object Object]';
};
```

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -62,7 +63,7 @@ function resolveLocalOrCwd(payloadPath) {
  * @return {ExpectedLHR}
  */
 function runLighthouse(url, configPath, isDebug) {
-  isDebug = isDebug || Boolean(process.env.SMOKEHOUSE_DEBUG);
+  isDebug = isDebug || Boolean(process.env.LH_SMOKE_DEBUG);
```
brendankenny: doesn't debug also save the LHR? Up to y'all if you want that :)

lighthouse-cli/test/smokehouse/smokehouse.js
brendankenny: > `Object.prototype.toString.call(obj) === '[object Object]';`

what year is it?

brendankenny: :P

lighthouse-cli/test/smokehouse/smokehouse.js
Hoten: nah. just writing to disk is good enough for me. verbose shouldn't be _too_ verbose

PR #7235, opened 2019-02-13
core(unused-css): change title Defer -> Remove

### LINKED ISSUE #6588: Rename Performance Opportunity "Defer Unused CSS"
I find that this opportunity needs to be renamed as it's unclear in the way it's formatted. Defer Unused CSS sounds like you're suggesting for the developer/user to remove render blocking CSS files using loadCSS or another method. Which is why I was driving myself crazy as to why this rule kept coming up even though it was implemented. 

However, me not paying attention I went ahead and clicked it to see the following 
"Remove unused rules from stylesheets to reduce unnecessary bytes consumed by network activity).  Learn More.

That's an entirely different suggestion from the language used to describe the opportunity. Deferring the CSS would be as I suggested, loading it asynchronously to not interrupt the rendering of the webpage. This rule should be renamed to something more obvious like. 
"Remove Unused CSS Rules"
"Remove Unused CSS"

Using the word Defer though, in this case, doesn't make sense in my opinion thoughts?

patrickhulce:
**Summary**
It's a bit confusing that the unused CSS is probably dead CSS that's completely unused. We'll save that distinction for the help text and make the title a little more aggressive.

**Related Issues/PRs**
closes #6588

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
+++ b/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
@@ -10,10 +10,10 @@ const i18n = require('../../lib/i18n/i18n.js');
 
 const UIStrings = {
   /** Imperative title of a Lighthouse audit that tells the user to remove content from their CSS that isn’t needed immediately and instead load that content at a later time. This is displayed in a list of audit titles that Lighthouse generates. */
-  title: 'Defer unused CSS',
+  title: 'Remove unused CSS',
   /** Description of a Lighthouse audit that tells the user *why* they should defer loading any content in CSS that isn’t needed at page load. This is displayed after a user expands the section to see more. No word length limits. 'Learn More' becomes link text to additional documentation. */
-  description: 'Remove unused rules from stylesheets to reduce unnecessary ' +
-    'bytes consumed by network activity. ' +
+  description: 'Remove dead rules from stylesheets and defer the loading of CSS not used for ' +
+    'above-the-fold content to reduce unnecessary bytes consumed by network activity. ' +
     '[Learn more](https://developers.google.com/web/tools/lighthouse/audits/unused-css).',
 };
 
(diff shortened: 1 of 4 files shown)


cagross:
This is good news--thank you.  This should help clear up the the defer vs. remove issue.

But what about the remaining ambiguity on the [Learn More page](https://developers.google.com/web/tools/lighthouse/audits/unused-css)?  Specifically, the title refers to 'unused CSS.'  But the Learn More page describes 'critical CSS' and 'uncritical CSS.'  I was under the impression all three were separate:

Critical CSS:  CSS rules used on the page that affect above-the-fold content, which, if removed, could result in noticeable differences in layout/style.

Uncritical CSS:  CSS rules used on the page that affect below-the-fold content, or content not viewable on immediate page load (e.g. in a modal window that the user might open on the page)/

Unused CSS:  CSS rules that have no chance at all of being used on the page.

If we use these three designations, it sounds like the optimum way to deal with all would be to inline critical CSS, defer uncritical CSS, and remove unused CSS.  Is this indeed how PageSpeed Insights suggests we do things?  If so, I don't think the "Learn More" page makes that very clear.  Or am I misunderstanding the way in which PageSpeed Insights suggests we load the three types of CSS?

patrickhulce:
You are right @cagross that is the breakdown of the 3 categories of CSS we're recommending. The documentation is currently managed over in [another repo](https://github.com/google/WebFundamentals) which moves a little slower than core. That's where a fix would be done for that issue.

cagross:
@patrickhulce OK thanks for confirming that there are three distinct CSS 'types,' and what I described corresponds to Google's suggested treatment for each.  But where would I suggest that the verbiage of the [Learn More page](https://developers.google.com/web/tools/lighthouse/audits/unused-css) be modified to reflect this?  Are you saying I should bring that up as in issue in the WebFundamentals repo?  Or are you perhaps saying that the WebFundamentals repo needs to first approve this as best practice, then the Lighthouse repo needs to modify the text on the Learn More page accordingly?  Or something else?  Or is this already in the works, and I need not concern myself too much with it?

patrickhulce:
The change itself will take place in that other repo. It's in the backlog to update, but I'll go ahead and re-open the issue this closed to track its progress.

No further action from you necessary :)

cagross:
@patrickhulce OK got it, sounds good.  I see that you re-opened the other ticket.  I'll monitor progress there.  Thanks very much for confirming everything and giving helpful info.

Issue #7236, opened 2019-02-13
Lighthouse fails to flag a 24MP image


patrickhulce:
#### Provide the steps to reproduce
1. Run LH on https://www.portuguesetrails.com/en
2. Observe the horrifying 33 MB of images it loads
2. Look through all the image opportunity audits for their whopping 24 megapixel image https://www.portuguesetrails.com/sites/default/files/gallery/iStock-621859064BicicletasPraia_AG_PT_moedas1.jpg whose dimensions are 6000x4000

#### What is the current behavior?
Lighthouse does not flag this image as being massive.

#### What is the expected behavior?
Lighthouse flags this image.

I'm not really sure what to specifically do about this one though. It falls through the cracks because...

- It passes optimized images because for a 24MP image, it is fairly well compressed at only 2 MB.
- It passes webp images for the same reason.
- It passes "use responsive images" because it is offscreen and has client dimensions of 0.
- It passes "use offscreen" because it is requested after TTI.

I don't really want to change any of those constraints in our existing audits, so it feels like there's just a hole in our coverage for "this image is just way too big".

**Related issues**
#7173

patrickhulce:
We decided to count images that are offscreen in the "use-responsive-images" audit with an assumed display size of the entire screen.

PR #7237, opened 2019-02-13
core(optimized-images): cap execution to 5 seconds

### LINKED ISSUE #7173: OptimizedImages gatherer can take >20s
Locally, I see it take 30s when running LH on https://www.portuguesetrails.com/en
On LR, that's enough to hit our 60s render timeout and fail the entire run. (Which it does)

I think 15s is a decent time budget for this gatherer. (And probably all gatherers).

1. We can look into speeding up this gatherer.
2. We can enforce a deadline for all gatherers.

Thoughts? @patrickhulce

patrickhulce:
**Summary**
We cap execution of our OptimizedImage gatherer at 5s and estimate the savings on all the images we didn't get to.

I know it got big, but it's really just the capping + all the consequences of capping + tests.

**Related Issues/PRs**
closes #7173

### DIFF
--- a/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
+++ b/lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
@@ -16,6 +16,12 @@ const NetworkRequest = require('../../../lib/network-request');
 const Sentry = require('../../../lib/sentry');
 const Driver = require('../../driver.js'); // eslint-disable-line no-unused-vars
 
+// Image encoding can be slow and we don't want to spend forever on it.
+// Cap our encoding to 5 seconds, anything after that will be estimated.
+const MAX_TIME_TO_SPEND_ENCODING = 5000;
+// Cap our image file size at 1MB, anything bigger than that will be estimated.
+const MAX_RESOURCE_SIZE_TO_ENCODE = 1000 * 1024;
+
 const JPEG_QUALITY = 0.92;
 const WEBP_QUALITY = 0.85;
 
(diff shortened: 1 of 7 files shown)


lighthouse-core/gather/gatherers/dobetterweb/optimized-images.js
```diff
@@ -126,48 +84,24 @@ class OptimizedImages extends Gatherer {
...
-          fromProtocol: false,
-          originalSize: isBase64DataUri ? base64Length : networkRecord.resourceSize,
-          jpegSize: isBase64DataUri ? stats.jpeg.base64 : stats.jpeg.binary,
-          webpSize: isBase64DataUri ? stats.webp.base64 : stats.webp.binary,
-        };
-      });
-    });
+  async calculateImageStats(driver, networkRecord) {
+    const originalSize = networkRecord.resourceSize;
+    // We'll use our heuristic on the audit side for images that are too large and once we've hit our cap.
```
paulirish: reword comment so it fits the order of the conditional? :)

patrickhulce:
> Yes?

Yes!

> To me the only tricky thing is that we don't attempt images >2mb but prioritize all results by size. But I understand the tradeoff here between precision of our estimate and runtime.

Yeah, this limit to me is more that we should be flagging this image via other means. If you're serving an image over 2MB, it's not just an "encode with the right JPEG settings issue" it's a "this image is way too effin' big issue", so let's not spend time focusing on it (related #7236). I'm open to other constants for this though. Maybe 3MB? 5MB?

Assuming it falls into the optimization bucket though, it does make sense to prioritize the biggest byte savings first.

paulirish:
> Yeah, this limit to me is more that we should be flagging this image via other means. If you're serving an image over 2MB, it's not just an "encode with the right JPEG settings issue" it's a "this image is way too effin' big issue", so let's not spend time focusing on it 

Yah I like that approach.

Works for me.

Issue #6775, opened 2018-12-11
DZL Tracking Issue


patrickhulce:
See #6152 for historical discussion and DZL's purpose. Comment or edit this issue to add feature requests.

**Features**
- [x] On-demand runs of specific hashes/URLs
  - [x] support for arbitrary # of runs
  - [ ] support for PSI runs instead of hashes
- [x] Tooltip on hover of hash with commit info (author, subject, date?)
- [x] Expose page vital stats **in progress**
  - [x] Number of requests
  - [x] Number of scripts
  - [x] Number of tasks
  - [x] Number of tasks >10ms
  - [x] Total task execution time
  - [x] connection RTT
  - [x] RTT by origin
  - [x] server latency by origin
  

**Bugs**
?

patrickhulce:
Alright folks, if you're like me, you've found DZL results not very useful. They're interesting, but not really actionable. Every time it fails something I assume it wasn't really the PR and when it passes it doesn't really give me confidence that the PR is safe. Not a resounding success :/

Instead of adding features, I've spent the past (slow holiday) month or so when working on DZL playing around with different URLs and intentionally regressive LH versions to suss out what's going on. Here's a general dump of what I've learned and what I think that means for LH.

### tl;dr - we need to isolate and surface every aspect of variance and their sources

## We must separate "good", "bad", and "unnecessary" variance
There's "good", "bad", and "unnecessary" variance. "Good variance" is the variance due to changes in the user experience that were within the developer's control, i.e. resources changed and payloads got heavier, scripts got less efficient and took more time executing, etc. "Bad variance" is the variance due to changes in the user experience that were *mostly* outside the developer's control, i.e. connection speed changes, random spikes in server traffic that slow responses, etc. "Unnecessary variance" is a change in our metrics that does not track any change in the user experience. I'm not currently aware of any such variance, but it's worth calling out and eliminating if we find it :)

## Metric graphs without page vitals are meaningless
Every time I see a graph of a FCP that's different, I want to know why it's different. Were there different resources loaded? Were the same resources just larger? Did our network analysis estimate different RTTs/server latency to the origins? Did the CPU tasks take longer? Why? Without this answer, I have no reason to believe the implementation changed or even had anything to do with the difference.

To me, a page vital is anything that lantern uses in its estimates.

1. Number of requests
2. Total byte weight
3. Total CPU tasks
4. Total CPU time
5. Estimated Connection RTT
5. Estimated RTT by origin
[...]

PR #7239, opened 2019-02-13
core(lantern): add configuration for precomputed network analysis

patrickhulce:
**Summary**
As discussed in the variance meeting earlier, this adds the ability to read in precomputed network analysis values and write them to a file. Nothing happens automatically for now it's up to the caller to do all the booking, etc to limit the complexity and difficult questions until we can validate how effective this is.

**Related Issues/PRs**
#6775


patrickhulce:
> I was thinking, should the data be additive?

Definitely! Great idea!

IMO, this could probably wait until we prove out how much of an impact it has on the metrics, and even long-term fancy median/avg/merging logic could potentially be the caller's responsibility. I see potentially very different cases for CI and LR for example.

lighthouse-core/computed/load-simulator.js
```diff
@@ -26,6 +26,13 @@ class LoadSimulator {
       serverResponseTimeByOrigin: networkAnalysis.serverResponseTimeByOrigin,
     };
 
+    if (precomputedLanternData) {
```
brendankenny: add a comment about overwriting the above?

lighthouse-cli/cli-flags.js
```diff
@@ -97,6 +97,8 @@ function getFlags(manualArgv) {
         'max-wait-for-load':
             'The timeout (in milliseconds) to wait before the page is considered done loading and the run should continue. WARNING: Very high values can lead to large traces and instability',
         'extra-headers': 'Set extra HTTP Headers to pass with request',
+        'precomputed-lantern-data-path': 'Path to the file where lantern simulation data should be read from.',
+        'lantern-data-output-path': 'Path to the file where lantern simulation data should be written to.',
```
brendankenny: needs more explanation? Unclear from this what "simulation data" is or what it'll be used for

lighthouse-core/lib/asset-saver.js
```diff
@@ -272,11 +273,34 @@ async function logAssets(artifacts, audits) {
...
+
+  /** @type {LH.PrecomputedLanternData} */
+  const lanternData = {additionalRttByOrigin: {}, serverResponseTimeByOrigin: {}};
+  for (const [origin, value] of networkAnalysis.additionalRttByOrigin.entries()) {
+    if (origin.startsWith('http')) lanternData.additionalRttByOrigin[origin] = value;
+  }
+
+  for (const [origin, value] of networkAnalysis.serverResponseTimeByOrigin.entries()) {
+    if (origin.startsWith('http')) lanternData.serverResponseTimeByOrigin[origin] = value;
+  }
```
brendankenny: it feels like this shouldn't be in `asset-saver.js`, like it should be somewhere that can just be requested as `whatever.getSaveableLanternData()` or whatever, but I don't have a great suggestion. `Simulator`, since it's also in charge of understanding the format upon load?

lighthouse-core/test/computed/load-simulator-test.js
```diff
@@ -45,14 +53,52 @@ describe('Simulator artifact', () => {
...
+    expect(additionalRttByOrigin).toEqual(new Map([
+      ['https://pwa.rocks', 1000],
+      ['https://www.googletagmanager.com', 500],
+      ['https://www.google-analytics.com', 1000],
+    ]));
+    expect(serverResponseTimeByOrigin).toEqual(new Map([
+      ['https://pwa.rocks', 150],
+      ['https://www.googletagmanager.com', 200],
+      ['https://www.google-analytics.com', 400],
+    ]));
```
brendankenny: since these tests are all very specific to the internal implementation (have to call `simulate()`, options are on the connectionPool, etc), maybe a test somewhere that the precomputed numbers are used when simulated vs what would have resulted from the original graph?

lighthouse-cli/bin.js
```diff
@@ -105,6 +105,17 @@ if (cliFlags.extraHeaders) {
...
+if (cliFlags.precomputedLanternDataPath) {
+  const lanternDataStr = fs.readFileSync(cliFlags.precomputedLanternDataPath, 'utf8');
+  /** @type {LH.PrecomputedLanternData} */
+  const data = JSON.parse(lanternDataStr);
+  if (!data.additionalRttByOrigin || !data.serverResponseTimeByOrigin) {
+    throw new Error('Invalid precomputed lantern data file');
+  }
+
+  cliFlags.precomputedLanternData = data;
+}
```
brendankenny: add test(s)?

lighthouse-core/lib/asset-saver.js
patrickhulce: good point 👍 moved

lighthouse-core/test/computed/load-simulator-test.js
patrickhulce: good call, done 👍

Issue #6806, opened 2018-12-14
Head childs doesn't have to be evaluated in Maximum Child Elements


marcmrf:
#### Provide the steps to reproduce
1. Run LH on https://andro4all.com/
2. Check "Avoid an excessive DOM size" point on Diagnostics.

#### What is the current behavior?
Right now is failing due to Maximum Child Elements, 79, which is bigger than the 60 recommended. The parent element is the head.

#### What is the expected behavior?
Most of the elements placed in the head are metas, pre-connects, fetch's... As the number of the elements in the head doesn't affect to rendering performance, they have to be omitted in this check.

#### Environment Information
* Affected Channels: Devtools
* Lighthouse version: Lighthouse 4.0.0-beta
* Operating System: macOS

patrickhulce:
A fair point thanks for filing @marcmrf! 

@ebidel any reason you can remember that you included the `head` elements long, long ago? :)

midzer:
So it seems we're only looking for render relevant nodes under `<body>`. Might be an idea to filter them by using `nodeInBody(node)` function located right now in `font-size.js`

ebidel:
The original goal of the audit was to raise awareness to apps that create excessive DOM trees which can put a burden on layout/style recalcs and the general advise we got from the Chrome team was < 60 nodes per parent. I'm not sure there was any reason in particular we included `<head>` elements, other than for completeness. Possibly call out usage of old school meta/link tags which are no longer needed.

ebidel:
The other thing might be to avoid loads of markup in head. Means more markup to parse on a slow mobile connections, possible blocking inline scripts and other resources, and a slower first render. But those things could be captured in other perf audits.

paulirish:
Yeah lets only include elements inside the body.

patrickhulce:
Sounds good @midzer thanks!

brendankenny:
Fixed by #7241

PR #7241, opened 2019-02-13
core(dom-size): collect only body nodes for DOM stats

midzer:
Hi :)

**Summary**

Respect only nodes within `<body>` for dom-size calculation audit

This has been tagged as bug, but it appears it's more like an adjustment how calculation of DOM size is done

More accurate results whether there are too many DOM elements which might result in bad performance for layouting and rendering

**Related Issues/PRs**

https://github.com/GoogleChrome/lighthouse/issues/6806

Have a nice day
midzer

### DIFF
--- a/lighthouse-core/audits/dobetterweb/dom-size.js
+++ b/lighthouse-core/audits/dobetterweb/dom-size.js
@@ -6,8 +6,8 @@
 
 /**
  * @fileoverview Audits a page to see how the size of DOM it creates. Stats like
- * tree depth, # children, and total nodes are returned. The score is calculated
- * based solely on the total number of nodes found on the page.
+ * tree depth, # children, and total elements are returned. The score is calculated
+ * based solely on the total number of elements found on the page.
  */
 
 'use strict';
@@ -25,9 +25,9 @@ const UIStrings = {
   title: 'Avoids an excessive DOM size',
   /** Title of a diagnostic audit that provides detail on the size of the web page's DOM. The size of a DOM is characterized by the total number of DOM elements and greatest DOM depth. This imperative title is shown to users when there is a significant amount of execution time that could be reduced. */
   failureTitle: 'Avoid an excessive DOM size',
-  /** Description of a Lighthouse audit that tells the user *why* they should reduce the size of the web page's DOM. The size of a DOM is characterized by the total number of DOM nodes and greatest DOM depth. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
+  /** Description of a Lighthouse audit that tells the user *why* they should reduce the size of the web page's DOM. The size of a DOM is characterized by the total number of DOM elements and greatest DOM depth. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
   description: 'Browser engineers recommend pages contain fewer than ' +
-    `~${MAX_DOM_ELEMENTS.toLocaleString()} DOM nodes. The sweet spot is a tree ` +
+    `~${MAX_DOM_ELEMENTS.toLocaleString()} DOM elements. The sweet spot is a tree ` +
     `depth < ${MAX_DOM_TREE_DEPTH} elements and fewer than ${MAX_DOM_TREE_WIDTH} ` +
     'children/parent element. A large DOM can increase memory usage, cause longer ' +
     '[style calculations](https://developers.google.com/web/fundamentals/performance/rendering/reduce-the-scope-and-complexity-of-style-calculations), ' +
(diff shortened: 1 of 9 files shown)


lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -148,8 +148,8 @@ class DOMStats extends Gatherer {
     })()`;
     return passContext.driver.sendCommand('DOM.enable')
       .then(() => passContext.driver.evaluateAsync(expression, {useIsolation: true}))
-      .then(results => passContext.driver.getElementsInDocument().then(allNodes => {
-        results.totalDOMNodes = allNodes.length;
+      .then(results => passContext.driver.getElementsInDocument('body').then(bodyNodes => {
```
patrickhulce: I don't think this will do what we want it to do. The implementation of `getElementsInDocument` looks like it doesn't accept a querySelector

https://github.com/GoogleChrome/lighthouse/blob/e6dbefa15c07903733a9b1f90dbb9ae393b35cb6/lighthouse-core/gather/driver.js#L1141-L1153

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
```diff
@@ -148,8 +148,8 @@ class DOMStats extends Gatherer {
     })()`;
     return passContext.driver.sendCommand('DOM.enable')
       .then(() => passContext.driver.evaluateAsync(expression, {useIsolation: true}))
```
patrickhulce: we'll probably want to piggyback on our `getDOMStats` function to compute the total body nodes

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
midzer: Yup, I misleadingly thought I can pick `getElementsInDocument()` from those page-functions...

lighthouse-core/gather/gatherers/dobetterweb/domstats.js
midzer: Yeah, sounds reasonable to do calculation right there :)

patrickhulce:
Just checking in here @midzer do you think you'd be able to make the smoketest change we're talking about? You'll basically add an assertion in the [expectations file](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-cli/test/smokehouse/dobetterweb/dbw-expectations.js#L15) and run `yarn smoke dbw` to make sure it passes with your fix and fails without it :)

midzer:
Right now I'm kinda stuck with lighthouse due https://github.com/GoogleChrome/lighthouse/issues/7246
Yeah, I can try to finish those tests anyway @patrickhulce

patrickhulce:
Oh bummer :( I wonder if there's another solution. Could downloading Chromium straight from https://download-chromium.appspot.com/ make any difference?

midzer:
Thanks for your support @patrickhulce 
Meanwhile I've installed regular Google Chrome and can run LH properly via `node lighthouse-cli http://example.com --view --chrome-flags="--no-sandbox --headless --disable-gpu"`

Running `yarn smoke dbw` fails though:

> Unable to connect to Chrome

even if I add those chrome-flags in `runSmokehouse()` in `run-smoke.js`

brendankenny:
ok, since this is a breaking change in the artifacts, I wanted to get this in for 5.0 :) I added an entry in the expectations file for `dom-size` as requested.

I also realized we're really counting elements, not nodes, so I changed all names to reflect that fact.

@exterkamp, @patrickhulce are you still good with this with the extra smoke test?

lighthouse-core/audits/dobetterweb/dom-size.js
```diff
@@ -16,35 +16,35 @@ const Audit = require('../audit');
...
+  /** Title of a diagnostic audit that provides detail on the size of the web page's DOM. The size of a DOM is characterized by the total number of DOM elements and greatest DOM depth. This imperative title is shown to users when there is a significant amount of execution time that could be reduced. */
   failureTitle: 'Avoid an excessive DOM size',
   /** Description of a Lighthouse audit that tells the user *why* they should reduce the size of the web page's DOM. The size of a DOM is characterized by the total number of DOM nodes and greatest DOM depth. This is displayed after a user expands the section to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
   description: 'Browser engineers recommend pages contain fewer than ' +
-    `~${MAX_DOM_NODES.toLocaleString()} DOM nodes. The sweet spot is a tree ` +
+    `~${MAX_DOM_ELEMENTS.toLocaleString()} DOM nodes. The sweet spot is a tree ` +
```
patrickhulce: shall we change the text here then to to read `DOM elements`?

paulirish:
```suggestion
    `~${MAX_DOM_ELEMENTS.toLocaleString()} DOM elements. The sweet spot is a tree ` +
```

brendankenny: > shall we change the text here then to to read `DOM elements`?

definitely, thanks for catching that

brendankenny: weird, I can push manually to this branch but not like this

Issue #7084, opened 2019-01-24
Move viewport audit logic to computed artifact


mattzeunert:
Tap targets and font size audit currently call the viewport audit, and we don't like audits calling each other.

https://github.com/GoogleChrome/lighthouse/pull/5846#discussion_r250334020

brendankenny:
We should maybe expose some sort of `notApplicable` in the result from that computed artifact if the test was done with a desktop form factor. Since this will be at least three audits using it, we may not want to replicate it in each one, but it might also be overkill.

See #6687 and https://github.com/GoogleChrome/lighthouse/pull/5846#discussion_r250335779

@patrickhulce were you just talking about form factor information on the audit `context`? Or was that fixes for the report `Runtime settings` I'm mixing up with this?

patrickhulce:
> @patrickhulce were you just talking about form factor information on the audit context? Or was that fixes for the report Runtime settings I'm mixing up with this?

Yep I just filed #7043 that's very similar to this concern. Separately, I also want to fix runtime settings that I'm fairly vocal about in chat but I haven't collected my thoughts enough into an issue yet haha :)

chrisandflora:
> Tap targets and font size audit currently call the viewport audit, and we don't like audits calling each other.
> 
> [#5846 (comment)](https://github.com/GoogleChrome/lighthouse/pull/5846#discussion_r250334020)

This is really Awesome..

mattzeunert:
Should we create a `viewportMeta` computed artifact for this, which basically does the logic of the viewport audit and exposes the results?

@patrickhulce https://github.com/GoogleChrome/lighthouse/issues/7043 is just similar but wouldn't actually share code, right?

patrickhulce:
Yeah @mattzeunert I don't think they would end up sharing code, just related 👍

Issue #7043, opened 2019-01-17
Add `isMobile` boolean to audit context / create `Environment` artifact


patrickhulce:
**Summary**
In content width, we branch our logic based on if the audit was on a mobile page or not. I could see plugins wanting to do something similar. Moving this to an artifact or context boolean could be helpful.

*Note:* Created from a TODO in code so it can be more visible and properly prioritized

mattzeunert:
I want to do this so we can use it in https://github.com/GoogleChrome/lighthouse/issues/6687.

Should we use the `Environment` artifact approach? Right now the `isMobile` logic uses `artifacts.HostUserAgent`, so it we don't know the `isMobile` value upfront.

patrickhulce:
I like the idea of an `Environment` artifact 👍It could even be a base artifact so other gatherers could rely on it if needed.

mattzeunert:
Cool, I didn't know about base artifacts 🙂. Seems like they're already very similar to an `Environment` artifact, so we could just add `isMobile` there? Or refactor it and move stuff like `HostUserAgent` and `BenchmarkIndex` into `Environment`.

Right now we use the `HostUserAgent` rather than the `NetworkUserAgent` to detect `isMobile`. Is that correct when Lighthouse is run on a real mobile device?

Do we even need `HostUserAgent`?

---

Found this relevant PR: https://github.com/GoogleChrome/lighthouse/pull/5871

patrickhulce:
> Seems like they're already very similar to an `Environment` artifact, so we could just add `isMobile` there?

I think we want something similar to what content-width does now. We want to know both `isMobileHost` and `isMobile`/`isMobileFromSitesPerspective` where `isMobileFromSitesPerspective === (isMobileHost && !isEmulatingNonMobile) || isEmulatingMobile`.

> Right now we use the `HostUserAgent` rather than the `NetworkUserAgent` to detect `isMobile`. Is that correct when Lighthouse is run on a real mobile device?

We use the host user agent combined with our emulation settings to compute the above. When `NetworkUserAgent` is the same as JS-side user agent `isMobileFromSitesPerspective === (isMobileHost && !isEmulatingNonMobile) || isEmulatingMobile` is probably the same as just using the `NetworkUserAgent` yes 👍 

> Do we even need `HostUserAgent`? I guess we want it in the LHR so we know what Chrome version was used to generate the report, but should it be an artifact?

[...]

Issue #6687, opened 2018-11-29
Add ability to disable audits depending on config


rviscomi:
Related to #1769

**Feature request summary**

When running Lighthouse in desktop mode, there may be certain audits that are mobile-specific and would be irrelevant and misleading.

Applicable audits should be marked as mobile-only and disabled when run in a desktop configuration. This may also apply to desktop-only audits for mobile configs.

**What is the motivation or use case for changing this?**

If you run Lighthouse in a desktop configuration, mobile-specific audits (for example the Mobile Friendly group of SEO audits) would also run and affect the SEO score. In that case, the results would not be applicable to the page.

**How is this beneficial to Lighthouse?**

Showing only relevant audits improves the accuracy/applicability of the results.

cc @mattzeunert

paulirish:
SG. If we're not emulating as mobile, then we should mark the audit as notApplicable.
We can use the settings provided in the audit context?

brendankenny:
> We can use the settings provided in the audit context?

Yeah, since we don't have that many audits that will need to be marked `notApplicable`, I think we should do this audit-by-audit rather than add another audit `meta` entry.

In `audit(artifacts, context) {}`, adding a `if (context.settings.emulatedFormFactor === 'desktop') return {rawValue: false, notApplicable: true};` should be sufficient.

(it isn't possible to get `context.settings.emulatedFormFactor === 'none'` in devtools/extension/PSI, but if we want to support more of the CLI we would need some runtime detection. Not sure if that's worth it?)

mattzeunert:
I think @rviscomi and I decided we only need to do this for tap targets.

The font size audit is applicable on desktop, but should not fail because of a missing viewport meta tag.

rviscomi:
In addition to tap targets, I could see the font size audit getting disabled for desktop as well. The motivation for that audit is to catch mobile unfriendliness and while small font sizes also affect desktop usability there is no direct connection to SEO.

PR #7264, opened 2019-02-17
core(viewport): create ViewportMeta computed artifact

mattzeunert:
**Summary**

Get the viewport info from a ViewportMeta computed artifact, instead of the font size and tap target audits calling the viewport audit.

Involves making the three audits involved async because computed artifacts are async.

@brendankenny also mentioned including a `notApplicable` property in the artifact result, for when Lighthouse is run in desktop mode. I've not done that for now, but will consider it when [adding the `isMobile` artifact](https://github.com/GoogleChrome/lighthouse/issues/7043) or [adding `notApplicable` logic to tap targets](https://github.com/GoogleChrome/lighthouse/issues/6687).

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/7084

### DIFF
--- a/lighthouse-core/test/audits/seo/tap-targets-test.js
+++ b/lighthouse-core/test/audits/seo/tap-targets-test.js
@@ -12,7 +12,7 @@ const assert = require('assert');
 
 const fakeContext = {computedCache: new Map()};
 
-async function auditTapTargets(tapTargets, metaElements = [{
+function auditTapTargets(tapTargets, metaElements = [{
   name: 'viewport',
   content: 'width=device-width',
 }]) {
--- a/lighthouse-core/test/audits/seo/tap-targets-test.js
+++ b/lighthouse-core/test/audits/seo/tap-targets-test.js
@@ -10,7 +10,9 @@
 const TapTargetsAudit = require('../../../audits/seo/tap-targets.js');
 const assert = require('assert');
 
-function auditTapTargets(tapTargets, metaElements = [{
+const fakeContext = {computedCache: new Map()};
+
+async function auditTapTargets(tapTargets, metaElements = [{
   name: 'viewport',
   content: 'width=device-width',
 }]) {
@@ -19,7 +21,7 @@ function auditTapTargets(tapTargets, metaElements = [{
     MetaElements: metaElements,
   };
 
-  return TapTargetsAudit.audit(artifacts);
+  return TapTargetsAudit.audit(artifacts, fakeContext);
 }
 
 const tapTargetSize = 10;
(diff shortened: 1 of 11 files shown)


lighthouse-core/test/audits/seo/tap-targets-test.js
```diff
@@ -10,7 +10,9 @@
 const TapTargetsAudit = require('../../../audits/seo/tap-targets.js');
 const assert = require('assert');
 
-function auditTapTargets(tapTargets, metaElements = [{
+const fakeContext = {computedCache: new Map()};
```
patrickhulce: doesn't seem like this one needed to become async ;) but I like it 👍

lighthouse-core/test/audits/seo/tap-targets-test.js
```diff
@@ -152,43 +154,44 @@ describe('SEO: Tap targets audit', () => {
...
+    const auditResult = await auditTapTargets(
       getBorderlineTapTargets({
         overlapSecondClientRect: true,
       })
     );
     assert.equal(auditResult.rawValue, false);
   });
 
-  it('reports 2 items if the main target is overlapped both vertically and horizontally', () => {
+  it('reports 2 items if the main target is overlapped both vertically and horizontally',
```
patrickhulce: super 🚲 🏠 but I'd avoid the line and indentation split by getting rid of "the main target is" since none of the other tests really need to specify that piece


😆

types/artifacts.d.ts
```diff
@@ -348,6 +348,15 @@ declare global {
         }[];
       }
 
+      export type ViewportMeta = {
+        /** Whether the page has any viewport tag  */
+        hasViewportTag: boolean;
+        /** Whether the viewport tag is optimized for mobile screens */
+        hasMobileViewport: boolean;
```
patrickhulce: based on the description what about `isMobileOptimized` then?

lighthouse-core/test/audits/seo/tap-targets-test.js
mattzeunert: Computed artifacts return a promise, so I think tap targets needs to be async to use one?

patrickhulce: yeah but you're not using `await` in here right?

mattzeunert: Oh yeah that makes sense.

mattzeunert:
Thanks for the feedback @patrickhulce and @brendankenny!

> The computed artifact needs tests, though. I'd imagine most of the `computed/viewport-meta-test.js` tests will be what's in `viewport-test.js` now (plus any other coverage you think we need), so hopefully it'll be painless.

Move most of the tests over to `viewport-meta-test`, what do you think?

patrickhulce:
> Move most of the tests over to `viewport-meta-test`, what do you think?

heh, I specifically didn't make the test comment because in refactoring PRs I like to see that the original case and flowthrough works exactly as it used it :)

buuuuuut, since @brendankenny pointed it out, I say yes 👍

mattzeunert:
> Move most of the tests over to `viewport-meta-test`, what do you think?

Missed a `d` here, have already made the change.

mattzeunert:
@brendankenny What do you think of the PR now?

PR #7270, opened 2019-02-18
tests: refactor driver, more gotoURL tests

patrickhulce:
**Summary**
Final part of driver test reorg. Adds more coverage for gotoURL, the original point of this. Also convinced me that all the logic associated with `gotoURL` should be its own `NavigationManager` or something since this was crazy amount of work for still lots of coverage to go 😆

### DIFF
--- a/lighthouse-core/test/gather/driver-test.js
+++ b/lighthouse-core/test/gather/driver-test.js
@@ -60,8 +60,8 @@ function createMockSendCommandFn() {
  */
 function createMockOnceFn() {
   const mockEvents = [];
-  const mockFn = jest.fn().mockImplementation((event, listener) => {
-    const indexOfResponse = mockEvents.findIndex(entry => entry.event === event);
+  const mockFn = jest.fn().mockImplementation((eventName, listener) => {
+    const indexOfResponse = mockEvents.findIndex(entry => entry.event === eventName);
     if (indexOfResponse === -1) return;
     const {response} = mockEvents[indexOfResponse];
     mockEvents.splice(indexOfResponse, 1);
@@ -79,10 +79,6 @@ function createMockOnceFn() {
     return mockFn.mock.calls.find(call => call[0] === event)[1];
   };
 
-  mockFn.findListeners = event => {
-    return mockFn.mock.calls.filter(call => call[0] === event).map(call => call[1]);
-  };
-
   return mockFn;
 }
 
@@ -93,7 +89,7 @@ function createMockOnceFn() {
  * @param {Promise<T>} promise
  * @return {Promise<T> & {isDone: () => boolean, isResolved: () => boolean, isRejected: () => boolean}}
  */
-function createInspectablePromise(promise) {
+function makePromiseInspectable(promise) {
   let isResolved = false;
   let isRejected = false;
   let resolvedValue = undefined;
@@ -117,6 +113,13 @@ function createInspectablePromise(promise) {
 }
 
 expect.extend({
+  /**
+   * Asserts that an inspectable promise created by makePromiseInspectable is currently resolved or rejected.
+   * This is useful for situations where we want to test that we are actually waiting for a particular event.
+   *
+   * @param {ReturnType<makePromiseInspectable>} received
+   * @param {string} failureMessage
+   */
   toBeDone(received, failureMessage) {
     const pass = received.isDone();
 
@@ -430,7 +433,7 @@ describe('.goOffline', () => {
 });
 
 describe('.gotoURL', () => {
-  function createMockListenerFn() {
+  function createMockWaitForFn() {
     let resolve;
     let reject;
     const promise = new Promise((r1, r2) => {
@@ -511,16 +514,23 @@ describe('.gotoURL', () => {
 
   describe('when waitForNavigated', () => {
     it('waits for Page.frameNavigated', async () => {
-      driver.on = driver.once = createMockOnceFn()
-        .mockEvent('Page.frameNavigated');
+      driver.on = driver.once = createMockOnceFn();
 
       const url = 'https://www.example.com';
       const loadOptions = {
         waitForNavigated: true,
       };
 
-      const loadPromise = driver.gotoURL(url, loadOptions);
+      const loadPromise = makePromiseInspectable(driver.gotoURL(url, loadOptions));
       await flushAllTimersAndMicrotasks();
+      expect(loadPromise).not.toBeDone('Did not wait for frameNavigated');
+
+      // Use `findListener` instead of `mockEvent` so we can control exactly when the promise resolves
+      const listener = driver.on.findListener('Page.frameNavigated');
+      listener();
+      await flushAllTimersAndMicrotasks();
+      expect(loadPromise).toBeDone('Did not resolve after frameNavigated');
+
       await loadPromise;
     });
   });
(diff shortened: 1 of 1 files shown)


lighthouse-core/test/gather/driver-test.js
```diff
@@ -49,6 +49,88 @@ function createMockSendCommandFn() {
...
+    return mockFn;
+  };
+
+  mockFn.findListener = event => {
+    expect(mockFn).toHaveBeenCalledWith(event, expect.anything());
+    return mockFn.mock.calls.find(call => call[0] === event)[1];
+  };
+
+  mockFn.findListeners = event => {
+    return mockFn.mock.calls.filter(call => call[0] === event).map(call => call[1]);
```
Hoten: this is some jiu jestsu 😮

lighthouse-core/test/gather/driver-test.js
```diff
@@ -49,6 +49,88 @@ function createMockSendCommandFn() {
...
+}
+
+/**
+ * Transparently augments the promise with inspectable functions to query its state.
+ *
+ * @template T
+ * @param {Promise<T>} promise
+ * @return {Promise<T> & {isDone: () => boolean, isResolved: () => boolean, isRejected: () => boolean}}
+ */
+function createInspectablePromise(promise) {
```
brendankenny: maybe we need another name for this? It's less creating and more making a promise inspectable

lighthouse-core/test/gather/driver-test.js
```diff
@@ -421,27 +509,133 @@ describe('.gotoURL', () => {
...
+        driver._waitForNetworkIdle = createMockListenerFn();
+        driver._waitForCPUIdle = createMockListenerFn();
+
+        const listener = driver[`_waitFor${name}`];
+        const otherListeners = [
+          driver._waitForFCP,
+          driver._waitForLoadEvent,
+          driver._waitForNetworkIdle,
+          driver._waitForCPUIdle,
+        ].filter(l => l !== listener);
```
brendankenny: just learned that computed property names work with destructuring

```js
const waitFns = {
  _waitForFCP: createMockListenerFn(),
  _waitForLoadEvent: createMockListenerFn(),
  _waitForNetworkIdle: createMockListenerFn(),
  _waitForCPUIdle: createMockListenerFn(),
};
const {[`_waitFor${name}`]: listener, ...otherListeners} = waitFns;

Object.assign(driver, waitFns);
```

But I can't decide how evil that is :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -485,14 +678,19 @@ describe('.gotoURL', () => {
         },
       };
 
-      expect.assertions(2);
+      // 2 assertions in the catch block and the 1 implicit in `findListener`
+      expect.assertions(3);
```
brendankenny: 👍

lighthouse-core/test/gather/driver-test.js
```diff
@@ -485,14 +678,19 @@ describe('.gotoURL', () => {
...
 
       try {
         const loadPromise = driver.gotoURL(startUrl, loadOptions);
         await flushAllTimersAndMicrotasks();
+
+        const listener = driver.on.findListener('Security.securityStateChanged');
+        listener(insecureSecurityState);
```
brendankenny: hmm, why do it this way rather than with `mockEvent()`? (which would have to be modified, obvs)

Or leave a comment giving a heads up on doing it this way to have it fire later

lighthouse-core/test/gather/driver-test.js
```diff
@@ -485,14 +678,19 @@ describe('.gotoURL', () => {
...
+        await flushAllTimersAndMicrotasks();
         await loadPromise;
       } catch (err) {
-        expect(err.code).toEqual('INSECURE_DOCUMENT_REQUEST');
+        expect(err).toMatchObject({code: 'INSECURE_DOCUMENT_REQUEST'});
```
brendankenny: old way seems slightly clearer to me (since it's just a single property)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -49,6 +49,88 @@ function createMockSendCommandFn() {
...
+  });
+
+  mockFn.mockEvent = (event, response) => {
+    mockEvents.push({event, response});
+    return mockFn;
+  };
+
+  mockFn.findListener = event => {
+    expect(mockFn).toHaveBeenCalledWith(event, expect.anything());
+    return mockFn.mock.calls.find(call => call[0] === event)[1];
```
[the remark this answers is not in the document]
patrickhulce: as in, you won't be able to find the second listener installed for the same event?

that's what `findListeners` was for but turns out we don't actually need it for these tests as you already noticed :)

lighthouse-core/test/gather/driver-test.js
```diff
@@ -49,6 +49,88 @@ function createMockSendCommandFn() {
...
+ * Closely mirrors `createMockSendCommandFn`.
+ *
+ * It is decorated with two methods:
+ *    - `mockEvent` which pushes protocol event payload for consumption
+ *    - `findListener` which asserts that `on` was invoked with the given event name and
+ *      returns the listener .
+ */
+function createMockOnceFn() {
+  const mockEvents = [];
+  const mockFn = jest.fn().mockImplementation((event, listener) => {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/gather/driver-test.js
```diff
@@ -49,6 +49,88 @@ function createMockSendCommandFn() {
...
+  inspectablePromise.isDone = () => isResolved || isRejected;
+  inspectablePromise.isResolved = () => isResolved;
+  inspectablePromise.isRejected = () => isRejected;
+  inspectablePromise.getDebugValues = () => ({resolvedValue, rejectionError});
+
+  return inspectablePromise;
+}
+
+expect.extend({
+  toBeDone(received, failureMessage) {
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/gather/driver-test.js
```diff
@@ -360,6 +430,24 @@ describe('.goOffline', () => {
 });
 
 describe('.gotoURL', () => {
+  function createMockListenerFn() {
```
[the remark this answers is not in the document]
patrickhulce: good point, renamed to "createMockWaitForFn"

lighthouse-core/test/gather/driver-test.js
```diff
@@ -421,27 +509,133 @@ describe('.gotoURL', () => {
...
+        .mockEvent('Page.frameNavigated');
+
+      const url = 'https://www.example.com';
+      const loadOptions = {
+        waitForNavigated: true,
+      };
+
+      const loadPromise = driver.gotoURL(url, loadOptions);
+      await flushAllTimersAndMicrotasks();
+      await loadPromise;
```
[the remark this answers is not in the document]
patrickhulce: done

lighthouse-core/test/gather/driver-test.js
patrickhulce: 😮

lighthouse-core/test/gather/driver-test.js
patrickhulce: mockEvent would respond immediately, we want to control exact flow so we can assert, added a comment

lighthouse-core/test/gather/driver-test.js
patrickhulce: maybe, but it's a major PITA to debug when something goes wrong. this way will actually print the full error out so you see what was throwing.

compromise on `toHaveProperty`?

lighthouse-core/test/gather/driver-test.js
patrickhulce: sounds good 👍 

create, make, it's all the same to me 🤷‍♂️

patrickhulce:
> I'm not super wild about about being _so_ prescriptive about the internals of `gotoURL` in order to test it (any substantial refactor is going to have to change all of these tests)

I'd argue that if it were broken down and refactored these would be deps that would be mocked anyhow, so yes, the tests would change, but the meat should all be pretty much the same :)

IMO, this is also much less dependent on the specific impl details of these functions too. A refactor is very likely to change how we wait for network or CPU idle but not that we must wait for idle

Issue #7273, opened 2019-02-19
Stuck in ImageElements gatherer


paulirish:
```sh
lighthouse http://www.hallobest.com/ --only-categories=performance
```

Took over 60 seconds in the ImageElements gatherer

PR #7274, opened 2019-02-19
core(image-elements): cap natural size fetch time

patrickhulce:
**Summary**
Caps our time spent fetching natural size information in `ImageElements` gatherer.

Aside, this got me thinking that we really need artifact smoketests. The current logic is covered by existing smoketests, but it'd be a lot more comfortable to make changes to gatherers with lots of `evaluateAsync` if we could easily add gatherer smoketests.

**Related Issues/PRs**
closes https://github.com/GoogleChrome/lighthouse/issues/7273

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -353,6 +353,10 @@ class Driver {
    * @return {Promise<*>}
    */
   async _evaluateInContext(expression, contextId) {
+    // Use a higher than default timeout if the user hasn't specified a specific timeout.
+    // Otherwise, use whatever was requested.
+    const timeout = this._nextProtocolTimeout === DEFAULT_PROTOCOL_TIMEOUT ?
+      60000 : this._nextProtocolTimeout;
     const evaluationParams = {
       // We need to explicitly wrap the raw expression for several purposes:
       // 1. Ensure that the expression will be a native Promise and not a polyfill/non-Promise.
@@ -372,11 +376,11 @@ class Driver {
       includeCommandLineAPI: true,
       awaitPromise: true,
       returnByValue: true,
-      timeout: 60000,
+      timeout,
       contextId,
     };
 
-    this.setNextProtocolTimeout(60000);
+    this.setNextProtocolTimeout(timeout);
     const response = await this.sendCommand('Runtime.evaluate', evaluationParams);
     if (response.exceptionDetails) {
       // An error occurred before we could even create a Promise, should be *very* rare
--- a/lighthouse-core/gather/gatherers/image-elements.js
+++ b/lighthouse-core/gather/gatherers/image-elements.js
@@ -130,6 +130,8 @@ class ImageElements extends Gatherer {
   async fetchElementWithSizeInformation(driver, element) {
     const url = JSON.stringify(element.src);
     try {
+      // We don't want this to take forever, 250ms should be enough for images that are cached
+      driver.setNextProtocolTimeout(250);
       /** @type {{naturalWidth: number, naturalHeight: number}} */
       const size = await driver.evaluateAsync(`(${determineNaturalSize.toString()})(${url})`);
       return Object.assign(element, size);
@@ -147,7 +149,9 @@ class ImageElements extends Gatherer {
   async afterPass(passContext, loadData) {
     const driver = passContext.driver;
     const indexedNetworkRecords = loadData.networkRecords.reduce((map, record) => {
-      if (/^image/.test(record.mimeType) && record.finished) {
+      // The network record is only valid for size information if it finished with a successful status
+      // code that indicates a complete resource response.
+      if (/^image/.test(record.mimeType) && record.finished && record.statusCode === 200) {
         map[record.url] = record;
       }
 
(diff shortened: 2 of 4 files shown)


lighthouse-core/gather/gatherers/image-elements.js
```diff
@@ -147,7 +149,9 @@ class ImageElements extends Gatherer {
   async afterPass(passContext, loadData) {
     const driver = passContext.driver;
     const indexedNetworkRecords = loadData.networkRecords.reduce((map, record) => {
-      if (/^image/.test(record.mimeType) && record.finished) {
+      // The network record is only valid for size information if it finished with a successful status
+      // code that indicates a complete resource response.
+      if (/^image/.test(record.mimeType) && record.finished && record.statusCode === 200) {
```
patrickhulce: this is technically the only fix explicitly required for #7273, but seemed like a good idea to address other potential situations too

lighthouse-core/gather/gatherers/image-elements.js
```diff
@@ -163,6 +167,9 @@ class ImageElements extends Gatherer {
     const elements = await driver.evaluateAsync(expression);
 
     const imageUsage = [];
+    const top50Images = Object.values(indexedNetworkRecords)
+      .sort((a, b) => b.resourceSize - a.resourceSize)
```
brendankenny: should this be using the `transferSize` fallback like below?

patrickhulce: IMO, no. That fallback is for making sure we don't miss the super rare case of an image being GZIPped in our final savings estimate but has negligible effect on whether it counts in the top 50.

lighthouse-core/gather/gatherers/image-elements.js
```diff
@@ -177,8 +184,13 @@ class ImageElements extends Gatherer {
...
       // CSS images have no natural size information at all. Try to get the actual size if we can.
-      // Additional fetch is expensive; don't bother if we don't have a networkRecord for the image.
-      if ((element.isPicture || element.isCss) && networkRecord) {
+      // Additional fetch is expensive; don't bother if we don't have a networkRecord for the image,
+      // or it's not in the top 50 largest images.
+      if (
+        (element.isPicture || element.isCss) &&
+        networkRecord &&
+        top50Images.includes(networkRecord)
+      ) {
```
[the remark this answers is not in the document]
patrickhulce: I'll just convert the default to be these since the goal is to set every image to `0` or the real value anyhow.

lighthouse-core/gather/gatherers/image-elements.js
```diff
@@ -130,6 +130,8 @@ class ImageElements extends Gatherer {
   async fetchElementWithSizeInformation(driver, element) {
     const url = JSON.stringify(element.src);
     try {
+      // We don't want this to take forever, 250ms should be enough for images that are cached
+      driver.setNextProtocolTimeout(250);
```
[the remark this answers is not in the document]
patrickhulce: Correct loading the image to get the size info was taking a long time because it was reissuing the request.

lighthouse-core/gather/gatherers/image-elements.js
```diff
@@ -177,8 +184,13 @@ class ImageElements extends Gatherer {
 
       // Images within `picture` behave strangely and natural size information isn't accurate,
       // CSS images have no natural size information at all. Try to get the actual size if we can.
-      // Additional fetch is expensive; don't bother if we don't have a networkRecord for the image.
-      if ((element.isPicture || element.isCss) && networkRecord) {
+      // Additional fetch is expensive; don't bother if we don't have a networkRecord for the image,
+      // or it's not in the top 50 largest images.
```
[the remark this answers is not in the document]
patrickhulce: sure

Hoten:
Instead of doing `fetchElementWithSizeInformation` one element at a time, could we do it all at once? I'm wondering how much overhead there is in the protocol back and forth.

patrickhulce:
> ATM this gatherer could still timeout the run (in LR), if it has enough images.

How so? Or is this assuming the worst case imagineable where all 50 images need fetching and the gatherer takes 12.5 seconds.

> Instead of doing `fetchElementWithSizeInformation` one element at a time, could we do it all at once? I'm wondering how much overhead there is in the protocol back and forth.

We certainly could, just somewhat complicates things a bit and the idea of trying to load 50 images all at once if things go poorly means scares me a bit. I'd vote to avoid doing it until we have a compelling example case for it.

lighthouse-core/gather/driver.js
```diff
@@ -353,6 +353,10 @@ class Driver {
    * @return {Promise<*>}
    */
   async _evaluateInContext(expression, contextId) {
+    // Use a higher than default timeout if the user hasn't specified a specific timeout.
+    // Otherwise, use whatever was requested.
+    const timeout = this._nextProtocolTimeout === DEFAULT_PROTOCOL_TIMEOUT ?
+      60000 : this._nextProtocolTimeout;
```
brendankenny: since you just made `driver-test.js` shiny and new, could you add a quick test for this? I could see us easily breaking it at some point and the issue being difficult to notice

patrickhulce: good call! 👍

PR #7280, opened 2019-02-20
core(base-artifacts): add FormFactorIsMobile base artifact

mattzeunert:
**Summary**

Extract the logic for this from the `content-width` audit so we can use it for other audits.

@patrickhulce I know we were talking about an `Environment` artifact to wrap these two properties, but since there are already other environment-related properties directly on `baseArtifacts` I just added them there.

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/7043

### DIFF
--- a/lighthouse-core/test/audits/content-width-test.js
+++ b/lighthouse-core/test/audits/content-width-test.js
@@ -13,12 +13,12 @@ const assert = require('assert');
 describe('Mobile-friendly: content-width audit', () => {
   it('fails when scroll width differs from viewport width', () => {
     const result = Audit.audit({
-      HostUserAgent: 'Desktop',
+      IsMobile: true,
       ViewportDimensions: {
         innerWidth: 100,
         outerWidth: 300,
       },
-    }, {settings: {emulatedFormFactor: 'mobile'}});
+    });
 
     assert.equal(result.rawValue, false);
     assert.ok(result.explanation);
@@ -26,12 +26,12 @@ describe('Mobile-friendly: content-width audit', () => {
 
   it('fails when host user agent is a phone', () => {
     const result = Audit.audit({
-      HostUserAgent: 'Mobile Android',
+      IsMobile: true,
       ViewportDimensions: {
         innerWidth: 100,
         outerWidth: 300,
       },
-    }, {settings: {emulatedFormFactor: 'none'}});
+    });
 
     assert.equal(result.rawValue, false);
     assert.ok(result.explanation);
@@ -47,13 +47,13 @@ describe('Mobile-friendly: content-width audit', () => {
     }, {settings: {emulatedFormFactor: 'mobile'}}).rawValue, true);
   });
 
-  it('not applicable when device emulation is turned off', () => {
+  it('not applicable when run on desktop', () => {
     return assert.equal(Audit.audit({
-      HostUserAgent: 'Mobile Android Chrome',
+      IsMobile: false,
       ViewportDimensions: {
         innerWidth: 300,
         outerWidth: 450,
       },
-    }, {settings: {emulatedFormFactor: 'desktop'}}).notApplicable, true);
+    }).notApplicable, true);
   });
 });
--- a/lighthouse-core/test/audits/content-width-test.js
+++ b/lighthouse-core/test/audits/content-width-test.js
@@ -24,19 +24,6 @@ describe('Mobile-friendly: content-width audit', () => {
     assert.ok(result.explanation);
   });
 
-  it('fails when host user agent is a phone', () => {
-    const result = Audit.audit({
-      IsMobile: true,
-      ViewportDimensions: {
-        innerWidth: 100,
-        outerWidth: 300,
-      },
-    });
-
-    assert.equal(result.rawValue, false);
-    assert.ok(result.explanation);
-  });
-
   it('passes when widths match', () => {
     return assert.equal(Audit.audit({
       HostUserAgent: '',
--- a/lighthouse-core/test/audits/content-width-test.js
+++ b/lighthouse-core/test/audits/content-width-test.js
@@ -13,7 +13,7 @@ const assert = require('assert');
 describe('Mobile-friendly: content-width audit', () => {
   it('fails when scroll width differs from viewport width', () => {
     const result = Audit.audit({
-      FormFactorIsMobile: true,
+      TestedAsMobileDevice: true,
       ViewportDimensions: {
         innerWidth: 100,
         outerWidth: 300,
@@ -36,7 +36,7 @@ describe('Mobile-friendly: content-width audit', () => {
 
   it('not applicable when run on desktop', () => {
     return assert.equal(Audit.audit({
-      FormFactorIsMobile: false,
+      TestedAsMobileDevice: false,
       ViewportDimensions: {
         innerWidth: 300,
         outerWidth: 450,
--- a/lighthouse-core/test/audits/content-width-test.js
+++ b/lighthouse-core/test/audits/content-width-test.js
@@ -13,7 +13,7 @@ const assert = require('assert');
 describe('Mobile-friendly: content-width audit', () => {
   it('fails when scroll width differs from viewport width', () => {
     const result = Audit.audit({
-      IsMobileFormFactor: true,
+      FormFactorIsMobile: true,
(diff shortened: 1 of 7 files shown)


lighthouse-core/test/audits/content-width-test.js
```diff
@@ -13,25 +13,25 @@ const assert = require('assert');
...
         outerWidth: 300,
       },
-    }, {settings: {emulatedFormFactor: 'mobile'}});
+    });
 
     assert.equal(result.rawValue, false);
     assert.ok(result.explanation);
   });
 
   it('fails when host user agent is a phone', () => {
```
patrickhulce: FWIW I think this is a duplicate test now

lighthouse-core/test/gather/fake-driver.js
```diff
@@ -5,6 +5,79 @@
  */
 'use strict';
 
+function makeFakeDriver({protocolGetVersionResponse}) {
```
patrickhulce: 👏 👏

types/artifacts.d.ts
```diff
@@ -23,6 +23,10 @@ declare global {
       fetchTime: string;
       /** A set of warnings about unexpected things encountered while loading and testing the page. */
       LighthouseRunWarnings: string[];
+      /** Whether the page was loaded on either a real or emulated mobile device */
```
brendankenny:
```suggestion
      /** Whether the page was loaded on either a real or emulated mobile device. */
```

types/artifacts.d.ts
```diff
@@ -23,6 +23,10 @@ declare global {
...
+      IsMobile: boolean;
+      /** Whether Lighthouse was run on a mobile device (i.e. not on a desktop machine) */
```
brendankenny:
```suggestion
      /** Whether Lighthouse was run on a mobile device (i.e. not on a desktop machine). */
```

types/artifacts.d.ts
```diff
@@ -23,6 +23,10 @@ declare global {
       fetchTime: string;
       /** A set of warnings about unexpected things encountered while loading and testing the page. */
       LighthouseRunWarnings: string[];
+      /** Whether the page was loaded on either a real or emulated mobile device */
+      IsMobile: boolean;
```
brendankenny: we might need to bikeshed on these. When I see `artifacts.isMobile`, my first thought is "is what mobile?" :)

It'll make a little more sense when nested in an environment artifact, but do we have other ideas?

(I also like reusing `host` for the UA string artifacts, but (super nitpicky) dislike the `host` prefix vs suffix :)

lighthouse-core/test/gather/gather-runner-test.js
```diff
@@ -156,6 +157,50 @@ describe('GatherRunner', function() {
...
+      const driver = fakeDriver;
+      const config = new Config({passes: [{}]});
+      const settings = {};
+      const options = {url, driver, config, settings};
+
+      const results = await GatherRunner.run(config.passes, options);
+      expect(results.IsMobile).toBe(false);
+      expect(results.IsMobileHost).toBe(false);
+    });
+    it('works when running on desktop device with mobile emulation', async () => {
```
brendankenny: new lines between these

types/artifacts.d.ts
mattzeunert: `IsMobileFormFactor`?

types/artifacts.d.ts
```diff
@@ -23,6 +23,10 @@ declare global {
       fetchTime: string;
       /** A set of warnings about unexpected things encountered while loading and testing the page. */
       LighthouseRunWarnings: string[];
+      /** Whether the page was loaded on either a real or emulated mobile device. */
+      IsMobile: boolean;
+      /** Whether Lighthouse was run on a mobile device (i.e. not on a desktop machine). */
+      IsMobileHost: boolean;
```
paulirish: apparently, this is `false` if it was real mobile hardware but desktop was emulated. 
which isn't what I expected given the name.

paulirish: how about slightly different booleans?

`isMobileHardware`
`isMobileEmulated`

They will combine differently, but I think they communicate all the same things?

brendankenny: > how about slightly different booleans?
> `isMobileHardware`
> `isMobileEmulated`

That kind of just punts the question back to the audits again, though. Really we just want to be able to easily answer "were my artifacts gathered in a mobile environment, whether real or emulated?".

The value should be true if run on desktop with mobile emulation, run on mobile with mobile emulation, or run on mobile with no emulation, otherwise false.

Maybe we should just drop `IsMobileHost` for now (I don't think anything needs it at this point?) and just pick a name we all like for `isMobile`/`IsMobileFormFactor`/`isMobileEmulated`?

patrickhulce: I like `isMobileHardware` to replace `isMobileHost` 👍 

when discussing @paulirish 's comment we just used the phrase `isMobileConditions` which I kinda like over `isMobile`/`isMobileFormFactor`. I'd want to avoid `Emulated` since that's not really what we are getting at

paulirish: `isMobileHardware`
`wasMobileConditions` 
???

the second is past tense because it's sometimes true and sometimes not. 
crazy but semantically reasonable?

brendankenny: yeah, the `is` part is kind of weird since you're asking about how the data was collected, not the current state of things. OTOH we use `is` all over the place for exactly this kind of meaning and probably still should.

Could adapt to something like `isFromMobileConditions` to clear that up, though that's getting long winded :)

mattzeunert: I like `isMobileHardware`, but if we use that should `HostUserAgent` be `HardwareUserAgent`?

Maybe the user agent isn't even the right thing to look at. These are the special cases we have in mind right now:

- tap targets: is this device operated with imprecise inputs (i.e. touch)?
- font size: is this a small screen ([not 100% why that makes a difference](https://developers.google.com/web/tools/lighthouse/audits/font-sizes))
- viewport width: is this a small screen? I assume the reasoning for disabling it on desktop is that on desktop horizontal scrolling is more likely to be intentional, plus it's an SEO audit, not an accessibility one

A purely descriptive term for `isMobile` would be `isMobileNetworkUserAgent`, even though right now we use `HostUserAgent` and `emulatedFormFactor` to determine it.

mattzeunert: > apparently, this is `false` if it was real mobile hardware but desktop was emulated.
which isn't what I expected given the name.

Are you talking about `isMobileHost`? Right now I'd expect it to be true on a mobile device regardless of emulation.

patrickhulce: @mattzeunert Paul was just confused, he's got it now 👍 

I like `HardwareUserAgent` personally but that's a report-breaking change

mattzeunert: Renamed to `FormFactorIsMobile` and removed `IsMobileHost` from the base artifacts for now.

brendankenny: at first read, I'd probably have to look up whether `FormFactorIsMobile` meant either tested on actual mobile hardware or the definition we're going for.

What about something around `TestedAsMobileDevice`?

patrickhulce: I'm starting to circle back around to @mattzeunert 's suggestion that this represents whether the device presented itself as mobile to the site, i.e network user agent was mobile :)

at this point though, I think we've all identified it's hard to name and we should just document well in the comment above it what it really represents

brendankenny: user agent is independent of form factor, though, isn't it? That's why I was suggesting we go with whatever awkward string reasonably encapsulates `RanLighthouseWithADeviceThatEitherWasOrPretendedToBeAMobileOne`, which would include the UA string and screen characteristics

mattzeunert: `TestedAsMobileDevice` work for me, updated the PR.

mattzeunert:
We're using `TestedAsMobileDevice` now. @brendankenny  What do you think?

brendankenny:
everyone else ok with `TestedAsMobileDevice`? :)

patrickhulce:
> everyone else ok with TestedAsMobileDevice?

WFM

paulirish:
lgtm on naming :)

PR #7285, opened 2019-02-20
core(tsc): polish switch to new audit-details types

brendankenny:
Removes a few leftover `TODO`s and `@ts-ignore`s and removed the middle man on some of the details types being wrapped in an object and then immediately unwrapped again.

With this PR the `LH.Audit.Details` fix process should be finished.

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
+++ b/lighthouse-core/audits/byte-efficiency/unused-css-rules.js
@@ -151,8 +151,7 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
     }
 
     const usage = UnusedCSSRules.computeUsage(stylesheetInfo);
-    // @ts-ignore TODO(bckenny): fix index signature on ByteEfficiencyItem.
-    return Object.assign({url}, usage);
+    return {url, ...usage};
   }
 
   /**
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -80,7 +80,7 @@ class DetailsRenderer {
   _renderBytes(details) {
     // TODO: handle displayUnit once we have something other than 'kb'
     const value = Util.formatBytesToKB(details.value, details.granularity);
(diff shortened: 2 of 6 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -447,7 +447,6 @@ class ReportUIFeatures {
    */
   getReportHtml() {
     this._resetUIState();
-    // @ts-ignore - technically documentElement can be null, but that's dumb - https://dom.spec.whatwg.org/#document-element
```
brendankenny: this was unrelated, but noticed it when I was searching for `@ts-ignore`. Fixed in some tsc release since the comment was added.

lighthouse-core/audits/byte-efficiency/unused-css-rules.js
```diff
@@ -151,8 +151,7 @@ class UnusedCSSRules extends ByteEfficiencyAudit {
     }
 
     const usage = UnusedCSSRules.computeUsage(stylesheetInfo);
-    // @ts-ignore TODO(bckenny): fix index signature on ByteEfficiencyItem.
-    return Object.assign({url}, usage);
+    return {url, ...usage};
```
patrickhulce: 😆

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -207,7 +201,7 @@ class DetailsRenderer {
       // The value's type overrides the heading's for this column.
       switch (value.type) {
         case 'code': {
-          return this._renderCode(value);
+          return this._renderCode(value.value);
```
patrickhulce: 🤔 not that it's new, but seems like we need a new local name for `value`

brendankenny: > but seems like we need a new local name for `value`

haha, yeah. I actually like the local `value` and would rather change the property (e.g. `value.code`), but that's probably way too much for now :)

PR #7286, opened 2019-02-21
tests(smokehouse): gzip test to assert transfer and resource sizes

Hoten:
Assert that gzip savings are correctly being provided by the browser for LH to use.

Regression test for when a version of headless Chrome was giving weird `dataReceivedEvent` values in the face of gzipped resources (see [this doc](https://docs.google.com/document/d/1FcvF1kKtWuKh2OLXXkTsBuh7qq9f6PxmA-JWcre1wFo/edit) (private to Googlers) for more).

Related: #7284

### DIFF
--- a/lighthouse-cli/test/fixtures/static-server.js
+++ b/lighthouse-cli/test/fixtures/static-server.js
@@ -9,6 +9,7 @@
 /* eslint-disable no-console */
 
 const http = require('http');
+const zlib = require('zlib');
 const path = require('path');
 const fs = require('fs');
 const parseQueryString = require('querystring').parse;
@@ -71,6 +72,7 @@ function requestHandler(request, response) {
     }
 
     let delay = 0;
+    let gzip = false;
     if (queryString) {
       const params = new URLSearchParams(queryString);
       // set document status-code
@@ -93,10 +95,18 @@ function requestHandler(request, response) {
         }
       }
 
+      if (params.has('gzip')) {
+        gzip = Boolean(params.get('gzip'));
+      }
+
       // redirect url to new url if present
       if (params.has('redirect')) {
         return setTimeout(sendRedirect, delay, params.get('redirect'));
       }
+    }
+
+    if (gzip) {
+      headers['Content-Encoding'] = 'gzip';
     }
 
     response.writeHead(statusCode, headers);
@@ -106,7 +116,7 @@ function requestHandler(request, response) {
       return setTimeout(finishResponse, delay, data);
     }
 
-    finishResponse(data);
+    finishResponse(data, gzip);
   }
 
   function sendRedirect(url) {
@@ -117,8 +127,12 @@ function requestHandler(request, response) {
     response.end();
   }
 
-  function finishResponse(data) {
-    response.write(data, 'binary');
+  function finishResponse(data, gzip) {
+    if (gzip) {
+      response.write(zlib.gzipSync(data), 'binary');
+    } else {
+      response.write(data, 'binary');
+    }
     response.end();
   }
 }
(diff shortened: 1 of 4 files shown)


lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -103,4 +103,29 @@ module.exports = [
...
+    audits: {
+      'network-requests': {
+        details: {
+          items: [
+            {
+              url: 'http://localhost:10200/byte-efficiency/gzip.html',
+            },
+            {
+              url: 'http://localhost:10200/byte-efficiency/script.js?gzip=1',
+              transferSize: 1136,
```
patrickhulce: are these stable across platforms, like the native zlib mac/linux is the same byte-for-byte as js and transfer size is always the same and all that?

seems like we could loosen the assertion a bit

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
```diff
@@ -103,4 +103,29 @@ module.exports = [
...
+              resourceSize: 52997,
+            },
+            {
+              url: 'http://localhost:10200/byte-efficiency/script.js',
```
patrickhulce: is this is just for exact comparison with the gzip?

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
Hoten: Good question, no idea. For the purposes of the regression a ballpark is good enough (less than 1500 i guess).

Looks like travis / appveyor got the same thing I did.

This [SO answer](https://stackoverflow.com/questions/26516369/zlib-gzip-produces-different-results-for-same-input-on-different-oses) perhaps suggests that we can rely on zlib being the same? Interestingly there's a byte that represents the OS used, but that wouldn't change the length. You've got me curious now so I'm gonna dig in way deeper than necessary :P

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
Hoten: Just the ?gzip=1 one is important for the regression test. This is just for comparing. And maybe it'd catch some other future issue? I'm open to removing.

lighthouse-cli/test/smokehouse/byte-efficiency/expectations.js
patrickhulce: Any findings from the 🐰 🕳 ?

Hoten: Just that it may change when Node upgrades zlib, but they haven't update it in years. Seems fine to punt.

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -71,6 +72,7 @@ function requestHandler(request, response) {
     }
 
     let delay = 0;
+    let gzip = false;
```
paulirish: `useGzip`

lighthouse-cli/test/fixtures/static-server.js
```diff
@@ -93,20 +95,28 @@ function requestHandler(request, response) {
...
+
     response.writeHead(statusCode, headers);
 
     // Delay the response
     if (delay > 0) {
       return setTimeout(finishResponse, delay, data);
     }
 
-    finishResponse(data);
+    finishResponse(data, gzip);
```
brendankenny: if passing it this way, would need to pass into the `delay` path above as well (to get`delay` and `gzip` to work together).

Could also doing the gzip encoding when doing the header? e.g.
```js
if (gzip) {
  headers['Content-Encoding'] = 'gzip';
  data = zlib.gzipSync(data);
}

response.writeHead(statusCode, headers);

// ...
```

Hoten: Done

Issue #6070, opened 2018-09-19
Implement support for LH plugins


paulirish:

brendankenny:
let's call it done :)

Further plugin issues to be filed as needed

PR #7304, opened 2019-02-22
core: add group support for plugins

jburger424:
**Summary**
Allow for plugin to add custom groups and assign custom to plugin auditRefs. 
Modify tests to verify groups added and assigned properly.

**Related Issues/PRs**
#6959
#6070

### DIFF
--- a/lighthouse-core/config/config-plugin.js
+++ b/lighthouse-core/config/config-plugin.js
@@ -84,7 +84,7 @@ class ConfigPlugin {
     }
 
     return auditRefsJson.map(auditRefJson => {
-      const {id, weight, ...invalidRest} = auditRefJson;
+      const {id, weight, group: group, ...invalidRest} = auditRefJson;
       assertNoExcessProperties(invalidRest, pluginName, 'auditRef');
 
       if (typeof id !== 'string') {
@@ -93,10 +93,14 @@ class ConfigPlugin {
       if (typeof weight !== 'number') {
         throw new Error(`${pluginName} has an invalid auditRef weight.`);
       }
+      if (typeof group !== 'string' && typeof group !== 'undefined') {
+        throw new Error(`${pluginName} has an invalid auditRef group.`);
+      }
 
       return {
         id,
         weight,
+        ...group && {group},
       };
     });
   }
@@ -141,6 +145,56 @@ class ConfigPlugin {
     };
   }
 
+
+  /**
+   * Extract and validate groups JSON added by the plugin.
+   * @param {unknown} groupsJson
+   * @param {string} pluginName
+   * @return {Record<string, LH.Config.GroupJson>|undefined}
+   */
+  static _parseGroups(groupsJson, pluginName) {
+    if (groupsJson === undefined) {
+      return undefined;
+    }
+
+    if (!isObjectOfUnknownProperties(groupsJson)) {
+      throw new Error(`${pluginName} group json is not valid.`);
+    }
+
+    if (typeof groupsJson !== 'object') {
+      throw new Error(`${pluginName} has an invalid group type.`);
+    }
+    const groupIds = Object.keys(groupsJson);
+    if (!groupIds.length) {
+      throw new Error(`${pluginName} has an an empty groups object.`);
+    }
+    /** @type {Record<string, LH.Config.GroupJson>} **/
+    const parsedGroupsJson = {};
+    groupIds.forEach(groupId => {
+      if (typeof groupId !== 'string') {
+        throw new Error(`${pluginName} has an invalid group ID.`);
+      }
+      const groupJson = groupsJson[groupId];
+      if (!isObjectOfUnknownProperties(groupJson)) {
+        throw new Error(`${pluginName} has an invalid group.`);
+      }
+      const {title, description, ...invalidRest} = groupJson;
+      assertNoExcessProperties(invalidRest, pluginName, 'group');
+
+      if (typeof title !== 'string') {
+        throw new Error(`${pluginName} has an invalid group name.`);
+      }
+      if (typeof description !== 'string' && typeof description !== 'undefined') {
+        throw new Error(`${pluginName} has an invalid group description.`);
+      }
+      parsedGroupsJson[groupId] = {
+        title,
+        ...description && {description},
+      };
+    });
+    return parsedGroupsJson;
+  }
+
   /**
    * Extracts and validates a ConfigJson from the provided plugin input, throwing
    * if it deviates from the expected object shape.
@@ -156,6 +210,7 @@ class ConfigPlugin {
     }
 
     const {
+      groups: pluginGroupsJson,
       audits: pluginAuditsJson,
       category: pluginCategoryJson,
       ...invalidRest
@@ -164,6 +219,7 @@ class ConfigPlugin {
     assertNoExcessProperties(invalidRest, pluginName);
 
     return {
+      ...pluginGroupsJson && {groups: ConfigPlugin._parseGroups(pluginGroupsJson, pluginName)},
       audits: ConfigPlugin._parseAuditsList(pluginAuditsJson, pluginName),
       categories: {
         [pluginName]: ConfigPlugin._parseCategory(pluginCategoryJson, pluginName),
--- a/lighthouse-core/config/config-plugin.js
+++ b/lighthouse-core/config/config-plugin.js
@@ -159,18 +159,16 @@ class ConfigPlugin {
     }
 
     if (!isObjectOfUnknownProperties(groupsJson)) {
-      throw new Error(`${pluginName} groups json is not valid.`);
+      throw new Error(`${pluginName} groups json is not defined as an object.`);
     }
 
     const groups = Object.entries(groupsJson);
-    if (!groups.length) {
-      throw new Error(`${pluginName} has an empty groups object.`);
-    }
-    /** @type {Record<string, LH.Config.GroupJson>} **/
+
+    /** @type {Record<string, LH.Config.GroupJson>} */
     const parsedGroupsJson = {};
     groups.forEach(([groupId, groupJson]) => {
       if (!isObjectOfUnknownProperties(groupJson)) {
-        throw new Error(`${pluginName} has an invalid group.`);
+        throw new Error(`${pluginName} has a group not defined as an object.`);
       }
       const {title, description, ...invalidRest} = groupJson;
       assertNoExcessProperties(invalidRest, pluginName, 'group');
@@ -204,20 +202,20 @@ class ConfigPlugin {
     }
 
     const {
-      groups: pluginGroupsJson,
       audits: pluginAuditsJson,
       category: pluginCategoryJson,
+      groups: pluginGroupsJson,
       ...invalidRest
     } = pluginJson;
 
     assertNoExcessProperties(invalidRest, pluginName);
 
     return {
-      groups: ConfigPlugin._parseGroups(pluginGroupsJson, pluginName),
       audits: ConfigPlugin._parseAuditsList(pluginAuditsJson, pluginName),
       categories: {
         [pluginName]: ConfigPlugin._parseCategory(pluginCategoryJson, pluginName),
       },
+      groups: ConfigPlugin._parseGroups(pluginGroupsJson, pluginName),
     };
   }
 }
--- a/lighthouse-core/config/config-plugin.js
+++ b/lighthouse-core/config/config-plugin.js
@@ -84,7 +84,7 @@ class ConfigPlugin {
     }
 
     return auditRefsJson.map(auditRefJson => {
-      const {id, weight, group: group, ...invalidRest} = auditRefJson;
+      const {id, weight, group, ...invalidRest} = auditRefJson;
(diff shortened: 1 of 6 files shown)


lighthouse-core/config/config-plugin.js
```diff
@@ -93,10 +93,14 @@ class ConfigPlugin {
...
         throw new Error(`${pluginName} has an invalid auditRef weight.`);
       }
+      if (typeof group !== 'string' && typeof group !== 'undefined') {
+        throw new Error(`${pluginName} has an invalid auditRef group.`);
+      }
 
       return {
         id,
         weight,
+        ...group && {group},
```
patrickhulce: I don't think I've seen this pattern before, definitely gave me pause.

This is all to avoid having `.group` be in the object but undefined, correct? Is there a case that this messes with to warrant the extra layer of mental parsing?

lighthouse-core/config/config-plugin.js
```diff
@@ -84,7 +84,7 @@ class ConfigPlugin {
     }
 
     return auditRefsJson.map(auditRefJson => {
-      const {id, weight, ...invalidRest} = auditRefJson;
+      const {id, weight, group: group, ...invalidRest} = auditRefJson;
```
patrickhulce: nit: just `group,`

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +145,56 @@ class ConfigPlugin {
...
+      const {title, description, ...invalidRest} = groupJson;
+      assertNoExcessProperties(invalidRest, pluginName, 'group');
+
+      if (typeof title !== 'string') {
+        throw new Error(`${pluginName} has an invalid group name.`);
+      }
+      if (typeof description !== 'string' && typeof description !== 'undefined') {
+        throw new Error(`${pluginName} has an invalid group description.`);
+      }
+      parsedGroupsJson[groupId] = {
```
patrickhulce: WDYT about prepending the plugin name to the groupId to avoid collisions?

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +145,56 @@ class ConfigPlugin {
...
+      }
+      const groupJson = groupsJson[groupId];
+      if (!isObjectOfUnknownProperties(groupJson)) {
+        throw new Error(`${pluginName} has an invalid group.`);
+      }
+      const {title, description, ...invalidRest} = groupJson;
+      assertNoExcessProperties(invalidRest, pluginName, 'group');
+
+      if (typeof title !== 'string') {
+        throw new Error(`${pluginName} has an invalid group name.`);
```
patrickhulce: Should we just call it 'title' since that's the property name anyway?

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +145,56 @@ class ConfigPlugin {
...
+    }
+    const groupIds = Object.keys(groupsJson);
+    if (!groupIds.length) {
+      throw new Error(`${pluginName} has an an empty groups object.`);
+    }
+    /** @type {Record<string, LH.Config.GroupJson>} **/
+    const parsedGroupsJson = {};
+    groupIds.forEach(groupId => {
+      if (typeof groupId !== 'string') {
+        throw new Error(`${pluginName} has an invalid group ID.`);
```
patrickhulce: is this possible? isn't it coming from `Object.keys`?

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +145,56 @@ class ConfigPlugin {
...
+    if (typeof groupsJson !== 'object') {
+      throw new Error(`${pluginName} has an invalid group type.`);
+    }
+    const groupIds = Object.keys(groupsJson);
+    if (!groupIds.length) {
+      throw new Error(`${pluginName} has an an empty groups object.`);
+    }
+    /** @type {Record<string, LH.Config.GroupJson>} **/
+    const parsedGroupsJson = {};
+    groupIds.forEach(groupId => {
```
patrickhulce: super nit: could just use `Object.entries` since the first thing you do is grab the value

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +145,56 @@ class ConfigPlugin {
...
   }
 
+
+  /**
+   * Extract and validate groups JSON added by the plugin.
+   * @param {unknown} groupsJson
+   * @param {string} pluginName
+   * @return {Record<string, LH.Config.GroupJson>|undefined}
+   */
+  static _parseGroups(groupsJson, pluginName) {
```
patrickhulce: I wonder if we should be enforcing any extra constraints here, like if there are groups, all audits belong to a group or something

lighthouse-core/config/config-plugin.js
jburger424: It is not, removed.

lighthouse-core/config/config-plugin.js
jburger424: My one concern with this would be the case of including manual audits. This doesn't apply to our use case at the moment, but I imagine it may apply to others.

lighthouse-core/config/config-plugin.js
jburger424: You're right, I thought this was necessary but it isn't needed and is certainly unreadable. Removed.

jburger424:
> Thanks for the contribution @jburger424!! mind posting a screenshot with the motivating example? :)
> 
> Apologies if you may have already communicated this internally, and I'm just out of the loop.

The motivating example for this is our plugin [Ad Speed Insights](https://github.com/googleads/ad-speed-insights). 
ad-speed-insights-screenshot

lighthouse-core/config/config-plugin.js
patrickhulce: That's true, I'm fine with holding as-is to start 👍

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +146,52 @@ class ConfigPlugin {
...
+    if (groupsJson === undefined) {
+      return undefined;
+    }
+
+    if (!isObjectOfUnknownProperties(groupsJson)) {
+      throw new Error(`${pluginName} group json is not valid.`);
+    }
+
+    if (typeof groupsJson !== 'object') {
+      throw new Error(`${pluginName} has an invalid group type.`);
```
patrickhulce: `isObjectOfUnknownProperties` should already cover this case

lighthouse-core/config/config-plugin.js
```diff
@@ -164,6 +216,7 @@ class ConfigPlugin {
     assertNoExcessProperties(invalidRest, pluginName);
 
     return {
+      ...pluginGroupsJson && {groups: ConfigPlugin._parseGroups(pluginGroupsJson, pluginName)},
```
patrickhulce: is it possible to just do away with the `...pluginGroupsJson && {` construct here too?

lighthouse-core/config/config-plugin.js
```diff
@@ -93,10 +93,15 @@ class ConfigPlugin {
       if (typeof weight !== 'number') {
         throw new Error(`${pluginName} has an invalid auditRef weight.`);
       }
+      if (typeof group !== 'string' && typeof group !== 'undefined') {
+        throw new Error(`${pluginName} has an invalid auditRef group.`);
+      }
 
+      const prependedGroup = group ? `${pluginName}-${group}` : group;
```
brendankenny: 👍

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +146,49 @@ class ConfigPlugin {
...
+   * @param {string} pluginName
+   * @return {Record<string, LH.Config.GroupJson>|undefined}
+   */
+  static _parseGroups(groupsJson, pluginName) {
+    if (groupsJson === undefined) {
+      return undefined;
+    }
+
+    if (!isObjectOfUnknownProperties(groupsJson)) {
+      throw new Error(`${pluginName} groups json is not valid.`);
```
brendankenny: "not defined as an object"? Or something else more specific than "not valid"

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +146,49 @@ class ConfigPlugin {
...
+    }
+
+    const groups = Object.entries(groupsJson);
+    if (!groups.length) {
+      throw new Error(`${pluginName} has an empty groups object.`);
```
brendankenny: do we care if it's empty? Seems ok to support it for e.g. putting in a placeholder `{}` while creating the plugin

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +146,49 @@ class ConfigPlugin {
...
+    }
+    /** @type {Record<string, LH.Config.GroupJson>} **/
```
brendankenny:
```suggestion
    /** @type {Record<string, LH.Config.GroupJson>} */
```

lighthouse-core/config/config-plugin.js
```diff
@@ -141,6 +146,49 @@ class ConfigPlugin {
...
+    const parsedGroupsJson = {};
+    groups.forEach(([groupId, groupJson]) => {
+      if (!isObjectOfUnknownProperties(groupJson)) {
+        throw new Error(`${pluginName} has an invalid group.`);
```
brendankenny: "not an object" or whatever

lighthouse-core/config/config-plugin.js
```diff
@@ -156,6 +204,7 @@ class ConfigPlugin {
     }
 
     const {
+      groups: pluginGroupsJson,
```
brendankenny: total nit, but put `groups` after `category` (and `categories` in the `return` below) to match the order of the type declarations

lighthouse-core/test/config/config-test.js
```diff
@@ -721,6 +721,19 @@ describe('Config', () => {
       assert.strictEqual(config.categories['lighthouse-plugin-simple'].title, 'Simple');
     });
 
+    it('should append a group', () => {
```
brendankenny: should also have one of the audits in the fixture using the group? (maybe better to check up in 'should append audits'?)

lighthouse-core/test/config/config-plugin-test.js
```diff
@@ -84,6 +93,7 @@ describe('ConfigPlugin', () => {
       categories: {
         'lighthouse-plugin-evil': evilCategory,
       },
+      groups: undefined,
```
brendankenny: do these two additions add anything to the test? My brain may be just foggy today and I'm missing it :)

lighthouse-core/test/config/config-plugin-test.js
```diff
@@ -253,4 +265,55 @@ describe('ConfigPlugin', () => {
       });
     });
   });
+
+  describe('`groups`', () => {
+    it('throws if auditRef has an invalid group id', () => {
```
brendankenny: should move this test into the ``describe('`category.auditRefs`')`` test group

lighthouse-core/test/config/config-plugin-test.js
```diff
@@ -253,4 +265,55 @@ describe('ConfigPlugin', () => {
...
+      assert.throws(() => ConfigPlugin.parsePlugin(pluginClone, nicePluginName),
+        /^Error: lighthouse-plugin-nice-plugin has an invalid group title/);
+    });
+
+    it('throws if group description is invalid', () => {
+      const pluginClone = deepClone(nicePlugin);
+      pluginClone.groups['group-a'].description = 55;
+      assert.throws(() => ConfigPlugin.parsePlugin(pluginClone, nicePluginName),
+        /^Error: lighthouse-plugin-nice-plugin has an invalid group description/);
+    });
```
brendankenny: can you also add
- "throws if excess property on a group" test and
- a passing test similar to the `'correctly passes through the contained auditRefs'` one just to check that the groups are coming through with prefixed names and expected shape?

lighthouse-core/config/config-plugin.js
jburger424: Agreed, removed.

lighthouse-core/test/config/config-plugin-test.js
jburger424: Tests were failing without this because the undefined group object is not yet stripped at this point.

lighthouse-core/test/config/config-test.js
jburger424: Moved to 'should append audits' and added auditRef check.

lighthouse-core/test/fixtures/config-plugins/lighthouse-plugin-simple/plugin-simple.js
```diff
@@ -7,12 +7,19 @@
...
+  },
   audits: [
     {path: 'redirects'},
     {path: 'user-timings'},
   ],
   category: {
     title: 'Simple',
-    auditRefs: [],
+    auditRefs: [
+      {id: 'redirects', weight: 1, group: 'new-group'},
```
brendankenny: oh, ha, there weren't any audit refs to add a group to before. Sorry I missed that :)

Issue #5950, opened 2018-09-03
Audits comparison


nicoabie:
**Feature request summary**
Audits comparison would be a nice feature in order to try different building configurations and see easily the benefits or drawbacks being introduced.

With some help and guidance I'm willing to work on it.

**What is the motivation or use case for changing this?**
It would greatly help the auditor and developer team to ship better and more performant code. 

**How is this beneficial to Ligthhouse?**
I think audits comparison is inline with the end goals of Lighthouse
BTW, I've using the [viewer](https://googlechrome.github.io/lighthouse/viewer/) which is great.
It would just be great to be able to upload more than just one file and see the advantages and disadvantages

Thanks for the great work

paulirish:
Yeah!  We've called this "report diffing" but I think we're talking about the same thing.  We definitely have wanted to do this as well.

We'd love to have you work on this. And I like your idea of adding to the viewer app first. We can then figure out where else we can expose the functionality.

We actually have two older mocks that may be useful:

![image](https://user-images.githubusercontent.com/39191/45000003-39787f00-af76-11e8-9461-2b593ab1beb7.png)
![image](https://user-images.githubusercontent.com/39191/45000009-3e3d3300-af76-11e8-98c8-3f3c2f9bb4f3.png)

Does that UI look mostly like what you had in mind?

If so, then we can work out how we want to model the data.

nicoabie:
Yeah @paulirish, those mockups look very good to me!

paulirish:
Great. I think first we'll need a data structure to represent the two reports and their differences. It's going to be a little while before I have the time to make a proposal here. (Quite busy with other work, unfortch)... But if you want to take a stab, I think that'd be great.

At a high level, within each category, I think we want to know which items 1) regressed 2) improved 3) are still failing 4) are still passing. And we should keep in mind that we want to compare `details`, eg. a certain URL is no longer listed in the image-compression audit, because it was fixed.

Maybe these groups are used instead of any existing category groups? (perf cat will get special handling, though.)

nicoabie:
@paulirish I can begin with that, sure! I've been looking into the code and readme(s) and couldn't find the definition of the json report that it is being currently used by the [report-renderer](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/report/html/renderer/report-renderer.js). For what I read that is where all begins when the html of the report is being generated.
Do you have any hint, so I can start from that json and extend it?

I think the viewer should be able to:

[...]

PR #7306, opened 2019-02-22
WIP: report diff

connorjclark:
image

I wanted to see how the existing report generating code could be extended to a report diff. Would be great to enable diffs through some minor changes in this code, rather than build something completely new.

This is a WIP, so don't focus too much on the implementation details. I'd mostly like feedback on if this is a good direction.

For the higher-level rendering functions, I found it simplest to split functionality into a 1 LHR vs n LHRs (normal vs diff view - example, `render` / `renderDiff` in category render). But for the lower level ones (`populateAuditValues`), I converged them. I think the others could be converged as well (where n = 1 LHR would be the normal view), but I won't bother doing that until getting feedback. Also, it may be a bit simpler to grok the changes this way.

See this for a mock made some time ago: https://github.com/GoogleChrome/lighthouse/issues/5950#issuecomment-418183019

Diff between 2 LHRs:
http://bashful-week.surge.sh/diff-2.html

4 LHRs:
http://bashful-week.surge.sh/diff-4.html

As of now:

1. I have punted the bar visualization of the perf metrics. It's good, just delaying implementation.
2. For the perf opportunities / diagnostics, I simply show an LHR's opps/diags all together. Would be better to show each opp/diag just once with a column for each LHR's value. Another punt.
3. The audits in accessibility/seo/best practices group the audits like the mock does. Thoughts?
4. I like the visual distinction given by the long colored nodes denoting the LHR letter.  Good to keep? Should each long color node also show the URL?
5. Not a lot of real estate to show the category score for each LHR. I currently just show the first LHR. Thoughts on a workaround?
6. Need to populate the header/footer with data from all the LHRs, not just the first.

*5:
image

FYI: If you want to check out this branch and experiment, my current hack to get a diff is this:

```
MORE_LHR="tmp-test-lhrs/exterkamp.json,tmp-test-lhrs/hoten.json,tmp-test-lhrs/paulirish.json" node lighthouse-cli https://www.cnn.com --view
```


Hoten:
> I think it is okay to not show the bars here, this is already A LOT of info, so I think that might just be more noise, I don't think we lose much, and maybe frees us up to do # 2

I'll put all the metrics in one table (not two, like the normal view) with a column for each report. See how that looks.

> (# 2) Continued from first point, I think if we elide the bars, then we might have enough room to combine the results into 1 table like the rest of the categories (column for each report's results), maybe order the rows in the order from the first report and just fill in each other reports estimated savings based on that order. Or make the table sortable on each reports column.

I'm not really following the ordering part, could you explain further?

> Love the color coding. I think just the color + Letter code is fine. Do we need to trim the URLs on top though? Urls can get really long, don't know if that is already part of the logic, or we need to define a layout for long URLs.
> Ultra nit: can we use a more [material palette](https://standards.google/guidelines/google-material/color/palettes.html#brand-palette) for the colors?

Probably shouldn't trim. If you're comparing two different long URLs, you really need to see the whole thing. To prevent it breaking the layout, could have a color coded legend in the header (instead of doing it on the category scores). Or simply make it take up the whole width so they don't steal width from the gauges.

Yeah, I can use those colors. Link is dead tho...

> I'd almost say remove it. I find it confusing to just have the first one's score there. I'd say either:
> 
> * Make the shown score the average.  _meh_
> * Show all score gauges.
> * No gauges in each section.

Average would be meaningless, especially if comparing LHRs from different sites. Would like to find a way to display all of them. Maybe instead of having the gauge in the top right corner, could show all of them on their own row just before or after the description.

> Yeah. Not sure if its best to do a color coded set of multiple runtime settings footers, or just combine them all into a big table of settings? The header could be modified to be something a little more generic like, "Diff of multiple Lighthouse Reports" instead of a URL, and then the date range of all reports instead of just a date. Some ideas:
> screen shot 2019-02-22 at 11 21 58 am
> screen shot 2019-02-22 at 11 22 38 am

For the runtime settings, wb one table, and highlight the rows that differ?

I'm happy with the title being general and the dates being a the range of all runs, because all that data would be in the runtime settings table for each LHR. Same for the emulation summary - could that part could be omitted for diffing or is it important to show here in the header?

> Q's:
> 
> * Do we need to do diffing to say basically "these reports were run with different runtime settings, they shouldn't be compared directly and may not correlate" if the reports were made with diff throttling settings or diff benchmark idx's?

I think highlighting the rows in the runtime settings table is sufficient.

> * Audit errors aren't colored red?

I'll look into it.

> * Instead of "Changed" when there are mixed results, is "Mixed" or "Differing" or "Different" or "Assorted" better?  I don't know if I like "Changed".

Varied?

patrickhulce:
Woohoo this is an awesome start @Hoten!! Absolutely ❤️ that this is a first-class citizen.

image


@exterkamp way to go!! ;)


High-level meta comment: I really like the idea of this being a first-class citizen of the report code. Agreed it will be somewhat tricky/complicated to get it integrated and we should probably spend a decent amount of time discussing the options to make sure everyone's happy but I think it's the right way forward 👍 


Now on to the details, you covered a few of these things already but just wanted to throw my perspective in too

**What I Love**:
- This is a first-class citizen of the report!! 🎉 
- The regular audit display
image
This is simple, easy to understand, and easy to scan!
- The screenshot display
image
Super easy to see what changed!


**What I Slightly Less Love**: 😉 
- The performance comparisons. I know you punted opps/diagnostics but metrics also aren't super easy to compare right now. I think we'll have to change things around a bit more here to make it work.
- +1 to @exterkamp's callout of "Changed", I like "Mixed" personally.
- Runtime settings. I know this was punted too, but while the current strategy works great for other audits I think we'll have to do something a little more creative here too.

**Other Things To Consider**:
- Timescale of screenshot display. Right now each one will be relative to the its own TTI which isn't as helpful when comparing the same site. Ideally we'd get these all on the same timescale.
- A view where you can filter to only things that changed more than X%

That's the flush of what I've got so far. Look

Hoten:
> * Timescale of screenshot display. Right now each one will be relative to the its own TTI which isn't as helpful when comparing the same site. Ideally we'd get these all on the same timescale.

It seems like it'd be really tricky to line them up. What if we just put the `timing` value in the bottom right corner of each frame? In fact, I'd like to see that in the normal view too.

> * A view where you can filter to only things that changed more than X%

Neat. could define default values for this at report generation time, but allow it to be changed in the report too.

Hoten:
Various changes from some (but not all) of the feedback so far.

http://bashful-week.surge.sh/2/diff-2.html
http://bashful-week.surge.sh/2/diff-4.html

exterkamp:
Love the header & layout of URLs on the topline!

Styling idea for the performance category:
image

For the Opportunites and Diagnostics section:
 1. The 4 seperate tables should be merged into 1 table just like the `Metrics` table.  IMO.  With a column for each report's estimated savings.  
 1. Since the tables are normally organized by most estimated savings -> least savings this presents a problem, since each site would have different "estimated savings" ordering.  I propose we sort by the first reports' estimated savings by default.  This would result in the next X number of reports being "out of order" since they might not be sorted the same way.  But then the table could be resorted to any column's highest to lowest estimated savings' order.

Or perhaps make the Opportunities and Diagnostics collapsed by default?  Then click the colored header to expand each table?  That would reduce the amount of noise on the page.  Since the point is really diffing, not individual analysis?

@patrickhulce +1 I like "Mixed"

Hoten:
> Styling idea for the performance category:
> image
> 
> For the Opportunites and Diagnostics section:
> 
> 1. The 4 seperate tables should be merged into 1 table just like the `Metrics` table.  IMO.  With a column for each report's estimated savings.
> 2. Since the tables are normally organized by most estimated savings -> least savings this presents a problem, since each site would have different "estimated savings" ordering.  I propose we sort by the first reports' estimated savings by default.  This would result in the next X number of reports being "out of order" since they might not be sorted the same way.  But then the table could be resorted to any column's highest to lowest estimated savings' order.

Done.

http://bashful-week.surge.sh/3/diff-2.html
http://bashful-week.surge.sh/3/diff-4.html

brendankenny:
Do we want to support diffing more than two LHRs? It might be a lot easier to reason about layout if only dealing with a max of two LHRs, and many of them at once seems like it's venturing more into a graph view (albeit with symbols) than a diff anyways.

Hoten:
> Do we want to support diffing more than two LHRs? It might be a lot easier to reason about layout if only dealing with a max of two LHRs,

The only layout issues I'm seeing are with the category score gauges (not the ones in the header) - and even then, for just 2 a different layout is needed. So far, everything else seems straightforward, and no more complex to support 2 vs 2+.

> and many of them at once seems like it's venturing more into a graph view (albeit with symbols) than a diff anyways.

While the 2-diffs would be less common than n-diffs, I believe it's useful to support an arbitrary (but reasonably small) number of reports. There are pages I've worked on in the past with more than 2 variations (same page structure, slightly different content and features enabled on the page), and being able to compare them all at a glance would have been very informative.

AFAICT the code to support 2-diff is much like the code to support n-diff - just 2 would either be the exact same (just with arrays no more than length 2), or unrolled with variables like `reportA` and `reportB` flying around (which I think would be not be ideal). If it's reasonable that supporting n-diff is the same amount of complexity (or just a tad more) than supporting 2-diff, I think we should go with the former.

mattzeunert:
Not sure how feasible/in scope this is, but in general it would be nice to be able to only show changes or highlight them somehow.

E.g. if my page failed with 12 tap target failures before, and now I have 14, my attention wouldn't be drawn to that audit if there are just two error triangles side by side.

Hoten:
> Not sure how feasible/in scope this is, but in general it would be nice to be able to only show changes or highlight them somehow.
> 
> E.g. if my page failed with 12 tap target failures before, and now I have 14, my attention wouldn't be drawn to that audit if there are just two error triangles side by side.

I'm imagining a green up arrow / red down arrow layered over the Green check / Red triangle. That'd indicate the state of the audit for that LHR + the score relative to the base LHR. Is that a sensible way to convey all that information?

mattzeunert:
> I'm imagining a green up arrow / red down arrow layered over the Green check / Red triangle. That'd indicate the state of the audit for that LHR + the score relative to the base LHR. Is that a sensible way to convey all that information?

Yeah that's a good solution. You'd still have to manually compare the details tables to see what exactly changed, but that's lower priority.

> Also, what about thresholds? Is a score of 0.79 vs 0.78 worth drawing attention to?

For audits that are generally stable I'd say yes. We may already have 22 failing tap targets on the page, but if my FE change adds another one then I'd like to know that.

Hoten:
@jonkeller No telling when this will land, we gotta have some design discussions first. In the meantime, you can always check out this branch locally. Just gotta use my environment variable hacks here: https://github.com/GoogleChrome/lighthouse/pull/7306/files#diff-6a4779f17aa030f3223e4de92085eda9R51

Hoten:
Merged master back in, and _mostly_ put things back to how it was. Def. regressed though - will only bother to continue after some design talks.

3Q5278:
*http://vieons.com/?fromuid=1160494 <http://vieons.com/?fromuid=1160494>*

cjamcl <notifications@github.com> 於 2019年5月24日 週五 上午7:04寫道：

connorjclark:
@patrickhulce close this?

patrickhulce:
@connorjclark yeah, let's close but hang onto the branch for sure :)

mostly awaiting UX decisions before moving forward.

Issue #4663, opened 2018-03-01
When linking to LH docs, include utm parameters for analytics


rviscomi:
Right now when we link to the [LH docs](https://developers.google.com/web/tools/lighthouse/) we don't include any tracking information. For example:

https://developers.google.com/web/tools/lighthouse/audits/has-viewport-meta-tag

We could include utm parameters to better understand how visitors arrived at the docs and why. For example:

https://developers.google.com/web/tools/lighthouse/audits/has-viewport-meta-tag?utm_source=lighthouse&utm_medium=extension&utm_content=failure

Ideally we'd use `utm_medium` to differentiate the Chrome extension from the Audits panel in Dev Tools. Not sure if the code is self-aware enough to make this distinction.

Other params: `utm_source` always `lighthouse` to indicate that we sent users there from this tool, and `utm_content` to identify why users were sent to the docs (the audit failed, passed, maybe specific details about the failure reason).

cc @paulirish @patrickhulce @kaycebasques

kdzwinel:
@paulirish regarding `utm_source` - I haven't found any existing code that recognizes the environment (I thought that maybe sentry code is doing that, but sentry only works in cli). How does this look in your opinion:

```js
function getEnvironment() {
  if (typeof window !== 'undefined') {
    if (typeof InspectorMain !== 'undefined') {
      return 'devtools';
    }
    return 'extension';
  }
  return 'cli';
}
```

rviscomi:
^ small correction: `utm_source` should always be `lighthouse`, but we should customize the `utm_medium` param to match the environment

kdzwinel:
As discussed on the eng meeting - we are going to include environment info in the build time (instead of detecting it in the runtime as I originally proposed).

brendankenny:
#4972 was well suited to things before we changed the whole build system.

The changes broke the PR pretty badly, but hopefully made it simpler to accomplish the same task. There's a separate entry point for the extension/DevTools/LR/CLI now as well as a separate build entry point for each, which means we can easily add something to each platform's call out to the report generator (or add a `process.env` variable if it really comes down to it).

PR #7312, opened 2019-02-24
core(config): specify lighthouse channel

mattzeunert:
**Summary**

Determine whether Lighthouse was run through the CLI, npm module, Chrome extension, Chrome DevTools, or Lightrider.

Is settings the right place to put this? Feels like it's similar to `locale`.

**Related Issues/PRs**

Part of https://github.com/GoogleChrome/lighthouse/issues/4663


types/externs.d.ts
```diff
@@ -116,6 +116,8 @@ declare global {
       skipAudits?: string[] | null;
       // List of extra HTTP Headers to include
       extraHeaders?: Crdp.Network.Headers | null; // See extraHeaders TODO in bin.js
+      // How Lighthouse was run, e.g. from the Chrome extension or from the npm module */
```
patrickhulce: dangling `*/` seems unnecessary ;)

lighthouse-core/test/results/sample_v2.json
```diff
@@ -5,7 +5,7 @@
     "hostUserAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/66.0.3358.0 Safari/537.36",
     "benchmarkIndex": 1000
   },
-  "lighthouseVersion": "4.1.0",
+  "lighthouseVersion": "4.1.0+cli",
```
exterkamp: I don't know if I like an addition to the version itself, since it really _isn't_ part of the version.  I like it better as it's own item in `channel` only.

lighthouse-core/test/results/sample_v2.json
```diff
@@ -3172,6 +3172,7 @@
     "disableStorageReset": false,
     "disableDeviceEmulation": false,
     "emulatedFormFactor": "mobile",
+    "channel": "cli",
```
exterkamp: Do we want this info in LR by adding it to the proto definition?

lighthouse-core/test/results/sample_v2.json
patrickhulce: Enough of the channels have had their own cherry-picks though that I think one *could* make the argument that it has effectively become a subversion. Not sure if I feel strongly enough to make that argument myself, but one could :)

lighthouse-core/test/results/sample_v2.json
patrickhulce: Related: If a proto property exists, but it's always the default, does anyone notice it?

😆

lighthouse-core/test/results/sample_v2.json
mattzeunert: I don't know why we'd need it to be part of the version. The [old PR](https://github.com/GoogleChrome/lighthouse/pull/4972) mentioned there was a decision to include it, and @paulirish [made a comment](https://github.com/GoogleChrome/lighthouse/pull/4972#issuecomment-384365102) that seems like a summary of some kind of decision.

mattzeunert: Back then there was no `channel` value in the settings, so maybe all we wanted is to include the channel in the LHR somehow?

brendankenny: yeah, i would definitely prefer a separate `channel` entry. I believe @paulirish may have changed his thinking on this as well.

I don't remember what the earlier discussion was about, and there isn't any mention in old meeting notes that I can find, so let's call that moot :)

lighthouse-core/test/results/sample_v2.json
mattzeunert: @exterkamp Would I just add it as as field number 4 here? https://github.com/GoogleChrome/lighthouse/blob/master/proto/lighthouse-result.proto#L141

proto/lighthouse-result.proto
```diff
@@ -139,6 +139,9 @@ message LighthouseResult {
     // List of the categories that were run, empty if all were run
     // nullable list of strings
     google.protobuf.Value only_categories = 3;
+
+    // How Lighthouse was run, e.g. from the Chrome extension or from the npm module
+    string channel = 4;
```
exterkamp: If we are adding it to the proto, I would do it as an enum, i.e.
```
    enum Channel {
      UNKNOWN_CHANNEL = 0;
      ext = 1;
      cdt = 2;
      cli = 3;
      nm = 4;
      lr = 5;
    }
    Channel channel = 4;
```

But I would confirm with @paulirish we would want to add it.  We have elided much of the configSettings from the proto.  Esp. since the channel would always be `lr` for proto LHRs I'm not 100% it would be useful 🤷‍♂️

mattzeunert: Ok, if we don't include it then when rendering the report we don't know it was generated by Lightrider? The channel value would just be `undefined`?

@rviscomi Do we care much about being able to identify Lightrider-generated reports from the UTM links?

exterkamp: Oh yeah.  If it is an LR generated LHR and the proto didn't include this new field the `channel` field would be totally elided.  If we need this info from a rendering side, we would need to add this new field.  Which to me is a valid justification to include it in the proto imo.  

If we use the enum we need to include `UNKNOWN_CHANNEL` in the js so that can propagate from protos as a valid value that people would need to expect.

rviscomi: It'd be nice to have complete coverage of all channels but not critical.

lighthouse-core/index.js
```diff
@@ -48,6 +48,11 @@ async function lighthouse(url, flags = {}, configJSON, connection) {
   flags.logLevel = flags.logLevel || 'error';
   log.setLevel(flags.logLevel);
 
+  const channelOverride = /** @type {any} */(process.env.LH_CHANNEL) || flags.channel;
```
brendankenny: do we have an immediate need to have a `process.env` override? If not, let's wait to add it

lighthouse-core/test/results/sample_v2.json
brendankenny: should it also be in the top level of the LHR? Maybe inside `environment`, next to `benchmarkIndex`, etc?

clients/devtools-entry.js
```diff
@@ -36,6 +36,7 @@ function getDefaultConfigForCategories(categoryIDs) {
 function runLighthouseInWorker(port, url, flags, categoryIDs) {
   // Default to 'info' logging level.
   flags.logLevel = flags.logLevel || 'info';
+  flags.channel = 'cdt';
```
brendankenny: do we have a good place to test any/all of these? It's going to be difficult to notice if something is accidentally falling back to the node default, but I don't know if we have a good place to check

lighthouse-core/index.js
Hoten: I'm guessing Matt does.

types/externs.d.ts
```diff
@@ -116,6 +116,8 @@ declare global {
       skipAudits?: string[] | null;
       // List of extra HTTP Headers to include
       extraHeaders?: Crdp.Network.Headers | null; // See extraHeaders TODO in bin.js
+      // How Lighthouse was run, e.g. from the Chrome extension or from the npm module
+      channel?: 'ext' | 'cdt' | 'cli' | 'nm' | 'lr';
```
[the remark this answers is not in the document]
Hoten: yeah, especially "nm" doesn't mean anything to me without context.

lighthouse-core/index.js
patrickhulce: I believe I originally made this request in the old PR for WebPageTest

types/externs.d.ts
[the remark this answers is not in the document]
patrickhulce: sgtm 👍

patrickhulce:
> Is it still useful information to know the original source, or should it be `utm_medium=viewer` at that point? Especially if it's shared with multiple other people...none of them are touching the original source in that case.

I think the original source is far more important for environment info than where they are currently looking at it. Maybe not the correct thing from utm medium perspective, but I think viewer is the only real case where there will be a mismatch and maybe viewer just overrides that with UI features or something?

lighthouse-core/index.js
mattzeunert: Yeah just added it because the old PR mentioned wpt.

brendankenny: WPT makes sense. If we want to let folks override it, should it just be a regular cli/node flag (overriding only if not set)?

The type will also need to widen to just `string`, since we don't want anyone relying on it being a small set of strings when it could be anything.

mattzeunert: Have removed the `LH_CHANNEL` env logic, and added a channel CLI flag with a default value of `cli`.

lighthouse-core/test/results/sample_v2.json
mattzeunert: Is there a benefit over using `lhr.configSettings.channel`?

clients/devtools-entry.js
mattzeunert: Added a test case for the Chrome extension.

proto/lighthouse-result.proto
mattzeunert: @exterkamp Could PSI and web.dev pass in their own channel values and override the Lightrider default?

exterkamp: PSI could definitely pass it's own channel.  Web.dev, we'll have to make sure.  @kusler?

mattzeunert:
Any more feedback on this?

mattzeunert:
> I don't really think I'm satisfied with the field just being freeform string, I'd prefer a enumerated type on ts and on the proto. I like the longer names, but personally I want them enumerated.

@exterkamp We allow CLI and node module consumers to pass in their own `channel` values. Does it still make sense to enumerate the possible values, since I think we need `string` anyway?

I feel it's similar on the Lightrider/proto side, but maybe we can put `psi` and `webdotdev` in the enum?

exterkamp:
Hm, I might have missed this discussion (I don't see it in this PR?), but why do they get to pass their own channel?  Wouldn't their channel be "cli" or "node-module"?  Since that is the channel they were on?  What's the use case for sending a custom run channel that we don't know about beforehand and have enumerated?

brendankenny:
I think it comes down to what the purpose of the field is.

As helpful debug info after the fact, it makes sense to stick to just our "clients" (cli/node/devtools/extension/lr). That's the programmatic entry point, great.

However, #4663 was about getting a sense of what lead users to our docs. In that case it makes some sense to have it open so we can distinguish between e.g. the cli and PSI and WPT, which are all very different user experience entry points into our docs.

exterkamp:
I've been convinced irl.  sgtm!

mattzeunert:
@exterkamp Should be good to merge now 🙂

clients/test/extension/extension-test.js
```diff
@@ -176,4 +179,8 @@ describe('Lighthouse chrome extension', function() {
     // this audit has regressed in the extension twice, so make sure it passes
     assert.ok(await extensionPage.$('#is-crawlable.lh-audit--pass'), 'did not pass is-crawlable');
   });
+
+  it('should specify the channel as extension', async () => {
+    assert.equal(lhr.configSettings.channel, 'extension');
```
brendankenny: 👍

clients/lightrider-entry.js
```diff
@@ -34,6 +34,7 @@ async function runLighthouseInLR(connection, url, flags, {lrDevice, categoryIDs,
   // disableStorageReset because it causes render server hang
   flags.disableStorageReset = true;
   flags.logLevel = flags.logLevel || 'info';
+  flags.channel = 'lr';
```
brendankenny: can assert this in `lightrider-entry-test.js`

lighthouse-cli/cli-flags.js
```diff
@@ -146,6 +147,7 @@ function getFlags(manualArgv) {
       .default('port', 0)
       .default('hostname', 'localhost')
       .default('enable-error-reporting', undefined) // Undefined so prompted by default
+      .default('channel', 'cli')
```
brendankenny: can assert this in `'runLighthouse completes a LH round trip'` in `run-test.js`

lighthouse-core/config/constants.js
```diff
@@ -51,6 +51,7 @@ const defaultSettings = {
   disableStorageReset: false,
   disableDeviceEmulation: false,
   emulatedFormFactor: 'mobile',
+  channel: 'node',
```
brendankenny: and can assert this in `lighthouse-core/test/index-test.js` (basically just copy `'should return formatted LHR when given no categories'` and delete most of it :)

lighthouse-core/test/results/sample_v2.json
brendankenny: > Is there a benefit over using lhr.configSettings.channel?

because it's not really a setting, but now it kind of is, so that's fine

mattzeunert:
@brendankenny Thanks for those suggestions, added them now!

clients/test/lightrider-entry-test.js
```diff
@@ -56,5 +58,18 @@ describe('lightrider-entry', () => {
       assert.strictEqual(parsedResult.runtimeError.code, LHError.UNKNOWN_ERROR);
       assert.ok(parsedResult.runtimeError.message.includes(errorMsg));
     });
+
+    it('specifies the channel as lr', async () => {
+      const runStub = sinon.stub(Runner, 'run');
```
brendankenny: can you use jest mocks for this? I'm hoping we can drop `sinon` since we currently only use it in one place

lighthouse-core/test/index-test.js
```diff
@@ -132,6 +132,31 @@ describe('Module Tests', function() {
     });
   });
 
+  it('should specify the channel as node by default', function() {
```
brendankenny: not a big deal, but if making one last commit, could you make these async/await? no need to stick to the style of the file for that :)

clients/test/lightrider-entry-test.js
patrickhulce: oh +1 to using jest please :)


```
const runStub = Runner.run = jest.fn()
...
const config = runStub.mock.calls[0][1].config
```

I think?

mattzeunert: Good point, somehow I felt I needed to use sinon in mocha tests 🙂

brendankenny:
thanks!

rviscomi:
After this feature is released we'll also need to update the PSI and web.dev clients to specify their channel configs. @paulirish is there somewhere we can file an issue to do this?

paulirish:
@rviscomi interesting... :)  right now PSI webapp, PSI API clients, and web.dev will be tagged as 'lr'. 

We don't have a straightforward way to distinguish between different PSI API consumers easily. 

How do you want to make use of this channel data?

rviscomi:
The web apps should include their channel in a UTM param appended to their learn more links so we can see where documentation visitors are coming from.

brendankenny:
The web front ends could change the setting in the lhr before generating the report. Pretty gross, though :/

web.dev and the PSI do have their own analytics (I assume), which would allow teasing out those sources, at least.

rviscomi:
Yeah I assume we do have analytics on the tools themselves, this is just for the documentation on d.g.co to easily identify the LH channel that brought the user to the docs.

PR #7318, opened 2019-02-25
misc: export UA strings from emulation.js

paulirish:
potential improvement for cl/235073833

> Just some basic stuff to help improve keeping the UA that LR uses in sync. right now its hardcoded in LR.. but with this we can at least use an import script to keep them in sync.


brendankenny:
> potential improvement for b/235073833

actual description? :P

(that's not even a real bug)

paulirish:
>> potential improvement for b/235073833

>(that's not even a real bug)

oops. it's http://cl/235073833

>actual description? :P

Just some basic stuff to help improve keeping the UA that LR uses in sync. right now its hardcoded in LR.. but with this we can at least use an import script to keep them in sync.

PR #7319, opened 2019-02-25
deps: updates to weed out transitive vulnerabilities

brendankenny:
alerted by the github UI about a few vulnerable transitive dependencies. Almost all are descended from devDependencies (and the rest don't end up running, AFAICT), so they aren't a huge deal, but better to move off them.

This involved mostly removing and then re-adding the same dependency to get yarn to update transitive ones, but a few had to be manually updated in the `package.json`.

~~Nine~~ Ten have been eliminated, but a ~~couple remain~~ one remains. We need to get finally get rid of mocha (or update to latest, but I didn't want to change that without testing) to get rid of one ~~and there's one last one I haven't figured out yet.~~


package.json
```diff
@@ -101,10 +101,10 @@
     "browserify": "^16.2.3",
     "bundlesize": "^0.14.4",
     "chalk": "^2.4.1",
-    "codecov": "^2.2.0",
+    "codecov": "^3.2.0",
     "commitizen": "^2.9.6",
     "conventional-changelog-cli": "^1.3.4",
-    "coveralls": "^2.11.9",
+    "coveralls": "^3.0.3",
```
brendankenny: codecov and coveralls had to go up a major version to get the latest `request`

package.json
```diff
@@ -113,9 +113,6 @@
     "eslint-config-google": "^0.9.1",
     "gh-pages": "^2.0.1",
     "glob": "^7.1.3",
-    "gulp": "^3.9.1",
-    "gulp-replace": "^0.5.4",
-    "gulp-util": "^3.0.7",
```
brendankenny: I was so excited about deleting these before, I don't know how they remained

patrickhulce:
Travis isn't happy though. Maybe jsdom doesn't like the new changes 🤷‍♂️

brendankenny:
> Travis isn't happy though. Maybe jsdom doesn't like the new changes 🤷‍♂️

ok, not sure what happened, but made the changes in a slightly different order and should be good now :)

PR #7339, opened 2019-02-27
cli: write report in runLighthouse before quitting Chrome

brendankenny:
while we wait for the fix for https://crbug.com/936272 and/or in chrome-launcher, this change at least writes the report to disk before waiting for Chrome to close.

Also removes the `unhandledRejection` listener when done (fixing the `MaxListenersExceededWarning` for @Hoten) and converts`runLighthouse` to use async/await, as the ancient promise business in there was inscrutable. I think there's still more to improve here, both bikeshedding and file organization (e.g. `handleError` should maybe be in `bin.js`, not in here); happy for more ideas or leaving all that for yet another day.

### DIFF
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -62,53 +62,57 @@ function getDebuggableChrome(flags) {
   });
 }
 
+/** @return {never} */
 function showConnectionError() {
   console.error('Unable to connect to Chrome');
-  process.exit(_RUNTIME_ERROR_CODE);
+  return process.exit(_RUNTIME_ERROR_CODE);
 }
 
+/** @return {never} */
 function showProtocolTimeoutError() {
   console.error('Debugger protocol timed out while connecting to Chrome.');
-  process.exit(_PROTOCOL_TIMEOUT_EXIT_CODE);
+  return process.exit(_PROTOCOL_TIMEOUT_EXIT_CODE);
 }
 
-/** @param {LH.LighthouseError} err */
+/** @param {LH.LighthouseError} err @return {never} */
 function showPageHungError(err) {
   console.error('Page hung:', err.friendlyMessage);
-  process.exit(_PAGE_HUNG_EXIT_CODE);
+  return process.exit(_PAGE_HUNG_EXIT_CODE);
 }
 
-/** @param {LH.LighthouseError} err */
+/** @param {LH.LighthouseError} err @return {never} */
 function showInsecureDocumentRequestError(err) {
   console.error('Insecure document request:', err.friendlyMessage);
-  process.exit(_INSECURE_DOCUMENT_REQUEST_EXIT_CODE);
+  return process.exit(_INSECURE_DOCUMENT_REQUEST_EXIT_CODE);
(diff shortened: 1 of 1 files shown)


lighthouse-cli/run.js
```diff
@@ -62,53 +62,57 @@ function getDebuggableChrome(flags) {
...
 /**
  * @param {LH.LighthouseError} err
+ * @return {never}
  */
 function showRuntimeError(err) {
   console.error('Runtime error encountered:', err.friendlyMessage || err.message);
   if (err.stack) {
     console.error(err.stack);
   }
-  process.exit(_RUNTIME_ERROR_CODE);
```
patrickhulce: what's the benefit of these `return process.exit` changes?

lighthouse-cli/run.js
```diff
@@ -159,66 +163,62 @@ async function saveResults(runnerResult, flags) {
...
+/**
+ * Attempt to kill the launched Chrome, if defined.
+ * @param {ChromeLauncher.LaunchedChrome=} launchedChrome
+ * @return {Promise<void>}
+ */
+async function potentiallyKillChrome(launchedChrome) {
+  if (!launchedChrome) return;
+
+  try {
+    await launchedChrome.kill();
```
patrickhulce: should we time this out?

lighthouse-cli/run.js
```diff
@@ -159,66 +163,62 @@ async function saveResults(runnerResult, flags) {
...
-    if (launchedChrome !== undefined) {
-      return launchedChrome.kill();
+    // If in gatherMode only, there will be no runnerResult.
+    if (runnerResult) {
+      await saveResults(runnerResult, flags);
     }
-    return Promise.resolve({});
+
+    await potentiallyKillChrome(launchedChrome);
+    process.removeListener('unhandledRejection', handleTheUnhandled);
```
patrickhulce: should this be removed in `catch` too?

lighthouse-cli/run.js
brendankenny: > what's the benefit of these `return process.exit` changes?

purely tsc, you have to return it to take advantage of the `@return {never}` chain

lighthouse-cli/run.js
brendankenny: > should this be removed in `catch` too?

`handleError` always ends with a `process.exit()`, so it doesn't matter in the `catch` case. If we moved `handleError` to `bin.js`, though, we could put both `potentiallyKillChrome` and removing the listener to a `finally` block, which would be kind of nice.

patrickhulce: oh right 👍

lighthouse-cli/run.js
brendankenny: > should we time this out?

probably, but the Chrome child process still keeps the main process going, so it just prints a `Couldn't quit Chrome process. Timed out.` error message but still sits there spinning. Is that better?

We could also `process.exit(1)` in that case, but assuming we want to always exit in that case seems like it might be excessive?

PR #7340, opened 2019-02-27
misc: run NetworkService in-process to avoid hanging process kill

paulirish:
temporary workaround for the hang on `ChromeLauncher Killing Chrome instance`

### DIFF
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -57,7 +57,7 @@ function parseChromeFlags(flags = '') {
 function getDebuggableChrome(flags) {
   return ChromeLauncher.launch({
     port: flags.port,
-    chromeFlags: parseChromeFlags(flags.chromeFlags),
+    chromeFlags: parseChromeFlags(flags.chromeFlags + ' --enable-features=NetworkServiceInProcess'),
     logLevel: flags.logLevel,
   });
 }


lighthouse-cli/run.js
```diff
@@ -57,7 +57,7 @@ function parseChromeFlags(flags = '') {
 function getDebuggableChrome(flags) {
   return ChromeLauncher.launch({
     port: flags.port,
-    chromeFlags: parseChromeFlags(flags.chromeFlags),
+    chromeFlags: parseChromeFlags(flags.chromeFlags + ' --enable-features=NetworkServiceInProcess'),
```
patrickhulce: nit: should we do this on the post parse-side?

`.concat('--enable-features=NetworkServiceInProcess')` feels a bit safer somehow

brendankenny: can you also add a TODO or whatever to remove again with a crbug link

Hoten: does this override any other `--enable-features`, or are they combined

paulirish:
Fixed in Chrome as of rev 637480


```
Comment #12 on issue 936272 by bugdroid: Network Service prevents Chrome from being killed gracefully.
https://bugs.chromium.org/p/chromium/issues/detail?id=936272#c12
  https://chromium.googlesource.com/chromium/src.git/+/1ae9dbb37fc72e7682bedfb32c86608ed687cd13

commit 1ae9dbb37fc72e7682bedfb32c86608ed687cd13
Author: John Abd-El-Malek
Date: Mon Mar 04 23:40:47 2019

Don't try to recreate network service at shutdown repeatedly

When the network process is killed by the system and the browser knows it's in shutdown mode, it
won't connect to the network service. However this could lead to subsystems repeatedly trying to
connect to the network service. Fix this by creating a mojo pipe to the service that isn't
bound but which doesn't have a connection error.

Bug: 936272,931436

Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/1500776
Reviewed-by: Robbie McElrath 
Commit-Queue: John Abd-El-Malek
Cr-Commit-Position: refs/heads/master@{#637480}
```

closing this PR

PR #7352, opened 2019-02-28
core(network-recorder): redirects started by script are set as initiators

paulirish:
Given this test page:

```html
<!DOCTYPE html>
<script>
setTimeout(_ => {
  // add an iframe to the page via script
  // the iframe will open :10200/redirects-script.html
  // which redirects to :10503/redirects-script.html
  // which redirects to airhorner.com
  const elem = document.createElement('iframe');
  elem.src = 'http://localhost:10200/redirects-script.html?redirect=http%3A%2F%2Flocalhost%3A10503%2Fredirects-script.html%3Fredirect%3Dhttps%253A%252F%252Fairhorner.com%252F';
  document.body.append(elem);
}, 400);
</script>
```

We'd expect to see a request chain like so:
image

But we currently see this instead:
image

This PR resolves this. 

Our technique of setting initiator was mostly considering network redirects of the main document, but not when the initiator was Script.  A quick logic fix later and we're good.

-----------

I ran into this while looking at #6704 and #6675.


lighthouse-core/lib/network-recorder.js
```diff
@@ -368,7 +368,7 @@ class NetworkRecorder extends EventEmitter {
     for (const record of records) {
       const stackFrames = (record.initiator.stack && record.initiator.stack.callFrames) || [];
       const initiatorURL = record.initiator.url || (stackFrames[0] && stackFrames[0].url);
-      const initiator = recordsByURL.get(initiatorURL) || record.redirectSource;
```
patrickhulce: presumably there was some reason I fellback this way but I have no idea what is was at this point and this seems to make more sense, so..... cool!

maybe add a comment so we don't forget again?

lighthouse-core/lib/network-recorder.js
```diff
@@ -368,7 +368,9 @@ class NetworkRecorder extends EventEmitter {
...
+      // If we were redirected to this request, our initiator is that redirect, otherwise, it's the
+      // initiator provided by the protocol. See #7352
```
brendankenny:
```suggestion
      // initiator provided by the protocol. See https://github.com/GoogleChrome/lighthouse/pull/7352/
```

paulirish:
lantern tests results:
image

PR #7353, opened 2019-02-28
i18n: fix max potential FID description

paulirish:
this description was accidentally referring to EIL.


patrickhulce:
`update:sample-json`

Issue #7344, opened 2019-02-27
Documentation around config presets needs to be clearer


gidztech:
**Background:**
You can extend the default Lighthouse config like so:

```
{
        extends: 'lighthouse:default',
        settings: {
          emulatedFormFactor: 'desktop',
          ...
        },
}
```
This can be passed as an object to the Lighthouse node API. 

I found a number of presets (`default`, `full`, `lr-desktop`, `lr-mobile`, `mixed-content`, `perf`) in [lighthouse-core/config](https://github.com/GoogleChrome/lighthouse/tree/master/lighthouse-core/config). **It's not clear from the documentation how these are meant to be referenced.** 

I had to dive into the source code to see how they were loaded ([lighthouse-core/config/config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/config.js#L353)). It's now clear to me that only `lighthouse:default` and `lighthouse:full` are loaded in this way. All the other configs in that directory may just be there as examples.

Therefore, my understanding is that you either have to extend one of the two presets and add your own settings, audits etc. to it, or you create the full config from scratch. 

If you are using the CLI, you can import config from a file with the `--config-path` flag (as described [here](https://github.com/GoogleChrome/lighthouse/blob/master/docs/recipes/custom-audit/readme.md), which incidentially took me a while to find). You can reference those presets inside the Lighthouse repository directly or copy them into the your own repository. 

However, the documentation for the CLI also has a `preset` flag with a different set of options (`full`, `perf`, `mixed-content`). So now it's getting more confusing because I don't know what would happen if you used both `config-path` and `preset` flags together.

**Conclusion**
I think some improvements are needed regarding the documentation of the presets for both the Node API and the CLI. I'm happy to help, but I'd need some clarification first.

patrickhulce:
Thanks for filing @gidztech! We would love some help clarifying the docs around this :)

> All the other configs in that directory may just be there as examples.

Yes, they are there because they are used in various environments and tests. They can also be used via `presets` but aren't meant to be extended because they are rather specific to their respective use cases.

> Therefore, my understanding is that you either have to extend one of the two presets and add your own settings, audits etc. to it, or you create the full config from scratch.

Correct!

> However, the documentation for the CLI also has a `preset` flag with a different set of options (`full`, `perf`, `mixed-content`) So now it's getting more confusing because I don't know what would happen if you used both `config-path` and `preset` flags together.

[...]

PR #7354, opened 2019-03-01
docs: improve docs around preset configuration and using the Node module

gidztech:
**Summary**
Added improvements to the docs around configuring Lighthouse. I recently found some aspects of the existing docs confusing or lack. See https://github.com/GoogleChrome/lighthouse/issues/7344 for discussion.

This PR makes the following docs changes:
- ~~Adds `lr-desktop` and `lr-mobile` to the list of choices available with the `--preset` CLI flag~~
- Adds warning about the `--preset` flag being ignored if the `--config-path` flag is provided
- Adds an example config file reference for the `--config-path`  CLI flag so user knows config format
- Adds a configuration section with example in a more prominent place in the additional docs readme
- Adds note about the limitations of the `extends` property and how to use presets programmatically
- Adds links to newer presets available in the configuration readme

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -196,7 +196,11 @@ The groups property controls how to visually group audits within a category. For
 
 ## Config Extension
 
-The stock Lighthouse configurations can be extended if you only need to make small tweaks, such as adding an audit or skipping an audit, but wish to still run most of what Lighthouse offers. When adding the `extends: 'lighthouse:default'` property to your config, the default passes, audits, groups, and categories will be automatically included, allowing you modify settings or add additional audits to a pass. See [more examples below](#more-examples) to see different types of extension in action.
+The stock Lighthouse configurations can be extended if you only need to make small tweaks, such as adding an audit or skipping an audit, but wish to still run most of what Lighthouse offers. When adding the `extends: 'lighthouse:default'` property to your config, the default passes, audits, groups, and categories will be automatically included, allowing you modify settings or add additional audits to a pass.
+
+Please note that you can only extend from `lighthouse:default` or `lighthouse:full` using the `extends` property. Other internal configs found in the [lighthouse-core/config](https://github.com/GoogleChrome/lighthouse/tree/master/lighthouse-core/config) directory can be used by importing the config object from file reference, or by using the [`--preset`](https://github.com/GoogleChrome/lighthouse#cli-options) CLI flag.
+
+See [more examples below](#more-examples) to view different types of extensions in action.
 
 **Config extension is the recommended way to run custom Lighthouse**. If there's a use case that extension doesn't currently solve, we'd love to [hear from you](https://github.com/GoogleChrome/lighthouse/issues/new)!
 
@@ -205,7 +209,11 @@ The stock Lighthouse configurations can be extended if you only need to make sma
 The best examples are the ones Lighthouse uses itself! There are several reference configuration files that are maintained as part of Lighthouse.
 
 * [lighthouse-core/config/default-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/default-config.js)
+* [lighthouse-core/config/full-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/full-config.js)
+* [lighthouse-core/config/lr-desktop-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/lr-desktop-config.js)
+* [lighthouse-core/config/lr-mobile-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/lr-mobile-config.js)
 * [lighthouse-core/config/perf-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/perf-config.js)
+* [lighthouse-core/config/mixed-content-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/config/mixed-content-config.js)
 * [docs/recipes/custom-audit/custom-config.js](https://github.com/GoogleChrome/lighthouse/blob/master/docs/recipes/custom-audit/custom-config.js)
 * [pwmetrics](https://github.com/paulirish/pwmetrics/blob/v4.1.1/lib/perf-config.ts)
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -112,8 +112,8 @@ When installed globally via `npm i -g lighthouse` or `yarn global add lighthouse
 instance with an open debugging port.
 
 1. Run `chrome-debug`. This will log the debugging port of your Chrome instance
-2. Navigate to your site and log in.
-3. In a separate terminal tab, run `lighthouse http://mysite.com --port port-number` using the port number from chrome-debug.
+1. Navigate to your site and log in.
+1. In a separate terminal tab, run `lighthouse http://mysite.com --port port-number` using the port number from chrome-debug.
 
 ## Testing on a mobile device
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -82,6 +82,29 @@ log.setLevel(flags.logLevel);
 launchChromeAndRunLighthouse('https://example.com', flags).then(...);
 ```
 
+## Configuration
+In order to configure Lighthouse programmatically, you need to pass the config object as the 3rd argument.
+
+**Example:**
+```js
+{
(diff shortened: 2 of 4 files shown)


docs/configuration.md
```diff
@@ -196,7 +196,11 @@ The groups property controls how to visually group audits within a category. For
...
+The stock Lighthouse configurations can be extended if you only need to make small tweaks, such as adding an audit or skipping an audit, but wish to still run most of what Lighthouse offers. When adding the `extends: 'lighthouse:default'` property to your config, the default passes, audits, groups, and categories will be automatically included, allowing you modify settings or add additional audits to a pass.
+
+Please note that you can only extend from `lighthouse:default` or `lighthouse:full` using the `extends` property. Other internal configs found in the [lighthouse-core/config](https://github.com/GoogleChrome/lighthouse/tree/master/lighthouse-core/config) directory can be used by importing the config object from file reference, or by using the [`--preset`](https://github.com/GoogleChrome/lighthouse#cli-options) CLI flag.
```
brendankenny: we have more or less always intended to open `extends` to any other config, but no one has ever asked for it yet, so we've waited :)

Maybe we should open an issue for that and link it here?

docs/readme.md
```diff
@@ -82,15 +82,38 @@ log.setLevel(flags.logLevel);
...
+  settings: {
+    onlyAudits: [
+      'first-meaningful-paint',
+      'speed-index-metric',
+      'estimated-input-latency',
+      'first-interactive',
+      'consistently-interactive',
+    ],
+  },
+}
```
brendankenny: I can see both the benefits and the downsides to having this repeated here and in `configuration.md`.  Seems ok in exchange for the explanation in place? Just more places to update when updates are needed

docs/readme.md
```diff
@@ -82,15 +82,38 @@ log.setLevel(flags.logLevel);
 launchChromeAndRunLighthouse('https://example.com', flags).then(...);
 ```
 
+## Configuration
+In order to configure Lighthouse programmatically, you need to pass the config object as the 3rd argument.
```
brendankenny: > you need to pass the config object as the 3rd argument

The config parameter isn't required (if not provided, it defaults to the default config), so maybe this needs more emphasis on it being an option if needing a more custom LH run?

docs/readme.md
```diff
@@ -82,15 +82,38 @@ log.setLevel(flags.logLevel);
...
 
 When installed globally via `npm i -g lighthouse` or `yarn global add lighthouse`,
 `chrome-debug` is added to your `PATH`. This binary launches a standalone Chrome
 instance with an open debugging port.
 
 1. Run `chrome-debug`. This will log the debugging port of your Chrome instance
-1. Navigate to your site and log in.
-1. In a separate terminal tab, run `lighthouse http://mysite.com --port port-number` using the port number from chrome-debug.
+2. Navigate to your site and log in.
+3. In a separate terminal tab, run `lighthouse http://mysite.com --port port-number` using the port number from chrome-debug.
```
brendankenny: these line numberings were intentional (markdown takes care of the actual numbers making changes to the list easier)

readme.md
```diff
@@ -56,6 +56,7 @@ Configuration:
   --print-config                 Print the normalized config for the given config and options, then exit.                                  [boolean]
   --additional-trace-categories  Additional categories to capture with the trace (comma-delimited).
   --config-path                  The path to the config JSON.
+                                 An example config file: lighthouse-core/config/lr-desktop-config.js
```
brendankenny: this section mirrors the printout from `--help`, so will need to make these changes in `lighthouse-cli/cli-flags.js` to have them there as well

readme.md
```diff
@@ -64,7 +65,8 @@ Configuration:
                                  Chromium version 66.0 or later. By default, any detected Chrome Canary or Chrome (stable) will be launched.
                                                                                                                                        [default: ""]
   --port                         The port to use for the debugging protocol. Use 0 for a random port                                    [default: 0]
-  --preset                       Use a built-in configuration.                                            [choices: "full", "perf", "mixed-content"]
+  --preset                       Use a built-in configuration.                                            [choices: "full", "perf", "mixed-content", "lr-desktop", "lr-mobile"]
```
brendankenny: `lr-desktop` and `lr-mobile` are really only meant for configuring runs for PSI and are subject to change at any time, so I don't think we want to expose them as an option here

readme.md
```diff
@@ -164,6 +166,12 @@ lighthouse -GA=./gmailartifacts https://gmail.com
 
 The first time you run the CLI you will be prompted with a message asking you if Lighthouse can anonymously report runtime exceptions. The Lighthouse team uses this information to detect new bugs and avoid regressions. Opting out will not affect your ability to use Lighthouse in any way. [Learn more](https://github.com/GoogleChrome/lighthouse/blob/master/docs/error-reporting.md).
 
+## Using the Node module
+You can also use Lighthouse programmatically with the Node module.
+
+Read [Using Lighthouse programmatically](./docs/readme.md#using-programmatically) for help getting started.\
+Read [Lighthouse Configuration](./docs/configuration.md) to learn more about the configuration options available.
```
brendankenny: maybe this second link isn't necessary?

docs/configuration.md
gidztech: @brendankenny 

Perhaps what would be useful is creating a `lighthouse-presets` package (or similar) that has a load of useful preset configs that you can import into your own project. `lighthouse-core` could also bring it in interally as a dependency so it's all managed in one place. 

At that point, if a user wants to extend any of those presets, they can import one and do an object merge with their own config rather than relying on the `extends` property. The problem with `extends` is that you have to look up what the available options are, and name it correctly on file name convention.

Let me know what you think of that idea. I'm happy to help with it, but what should we do in the meantime? Do we start with this updated documentation first and then update again later if and when changes are made to how config is done?

brendankenny: > Do we start with this updated documentation first and then update again later if and when changes are made to how config is done?

I think this, definitely. Any progress we can make immediately is a good thing.

Maybe open an issue to discuss the other idea? I guess my only hesitation is I haven't seen a ton of interesting variations on configs out there, so I'm not sure if there are enough to justify a standalone collection, but I could be very wrong :)

docs/readme.md
gidztech: @brendankenny 

> Seems ok in exchange for the explanation in place? 

I'm not too sure what you meant here?

docs/readme.md
gidztech: I think there was an extra space before the first number, which got resolved when I commited.

docs/readme.md
brendankenny: > I'm not too sure what you meant here?

I meant, it is duplicating content, which can get lost and fall behind as we update docs in the future and don't hit all the places that need to be updated.

But on the other hand, it means someone reading this doesn't have to click through to another page just to get a simple example of the config, which is kind of nice

docs/readme.md
brendankenny: > I think there was an extra space before the first number, which got resolved when I commited.

can you revert these lines? They're meant to be just 1., 1., 1., etc

readme.md
gidztech: It would require two clicks to get to the configuration options from the main README. Having both here means I can get there quicker, especially if I keep coming back to the repository to remind myself.

docs/readme.md
gidztech: It's showing as 1. 2. 3. on the master version. See:

https://github.com/GoogleChrome/lighthouse/blob/master/docs/readme.md#testing-on-a-site-with-authentication

Or am I missing something?

docs/readme.md
gidztech: @brendankenny Yep, I understand the pros and cons, but what is our preference in this case?

brendankenny: > what is our preference in this case?

that's why I said

> Seems ok

:)

docs/readme.md
brendankenny: check out the raw version: https://raw.githubusercontent.com/GoogleChrome/lighthouse/master/docs/readme.md

it's the markdown viewer that's makes it rendered as 1. 2. 3.

readme.md
brendankenny: makes sense, sounds good to me

readme.md
gidztech: Ok reverted that.

docs/readme.md
gidztech: I've clarified the fact if omitted, a default config will be used.

PR #7358, opened 2019-03-02
core: remove NO_ERROR from default responses

exterkamp:
**Summary**
Removes NO_ERROR codes from default responses that contain no errors.
From some initial testing I did, this should have no effect on LR.

**Related Issues/PRs**
part of: #6336

### DIFF
--- a/lighthouse-cli/index.js
+++ b/lighthouse-cli/index.js
@@ -6,12 +6,7 @@
  */
 'use strict';
 
-require('./bin.js').begin().then((lhr) => {
-  // Exit with non-zero exit code if LHR contains runtime error.
-  if (lhr && lhr.lhr.runtimeError) {
-    process.exit(1);
-  }
-}).catch(err => {
+require('./bin.js').begin().catch(err => {
   process.stderr.write(err.stack);
   process.exit(1);
 });
--- a/lighthouse-cli/index.js
+++ b/lighthouse-cli/index.js
@@ -6,7 +6,12 @@
  */
 'use strict';
 
-require('./bin.js').begin().catch(err => {
+require('./bin.js').begin().then((lhr) => {
+  // Exit with non-zero exit code if LHR contains runtime error.
+  if (lhr && lhr.lhr.runtimeError) {
+    process.exit(1);
+  }
+}).catch(err => {
   process.stderr.write(err.stack);
   process.exit(1);
 });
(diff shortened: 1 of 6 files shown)


lighthouse-core/runner.js
```diff
@@ -345,10 +345,7 @@ class Runner {
       }
     }
 
-    return {
-      code: LHError.NO_ERROR,
-      message: '',
-    };
+    return undefined;
```
brendankenny: can you add a test for this in `runner-test.js` to go with the existing "includes a top-level runtimeError when a gatherer throws one" test

Something like
```js
it('does not include a top-level runtimeError when gatherers were successful', async () => {
  const config = new Config({
    settings: {
      auditMode: __dirname + '/fixtures/artifacts/perflog/',
    },
    audits: [
      'content-width',
    ],
  });

  const {lhr} = await Runner.run(undefined, {config});
  assert.strictEqual(lhr.runtimeError, undefined);
});
```
(fails on current master and should pass with this PR?)

exterkamp:
Added `exit code 1` when there is a runtime error present.  I was trying to add a `exit code 1` when there was _no LHR_ but that got complicated since there isn't an LHR when it is `print-config` or in `gather-only` mode.  Which are valid cases to not see an LHR.  So we can refactor more to detect those/return blank empty LHRs, but I wanted to get feedback on those ideas/if it was worth it @paulirish.

brendankenny:
sorry, I wasn't trying to tell you you needed to add that to this PR, just that we might not be done with that issue yet :)

lighthouse-cli/index.js
```diff
@@ -6,7 +6,12 @@
  */
 'use strict';
 
-require('./bin.js').begin().catch(err => {
+require('./bin.js').begin().then((lhr) => {
+  // Exit with non-zero exit code if LHR contains runtime error.
+  if (lhr && lhr.lhr.runtimeError) {
```
brendankenny: ideally we can handle this in `run.js` with the other runtime error/exiting code

lighthouse-core/runner.js
exterkamp: Done.

exterkamp:
Rolled back exit code changes for follow up PR.  Added test.

Ready for re-review :smile:

PR #7359, opened 2019-03-02
misc: use jsdoc-style comments on Flags properties

brendankenny:
converts them to `/** */` style so they appear on hover in vscode, etc

### DIFF
--- a/types/externs.d.ts
+++ b/types/externs.d.ts
@@ -59,19 +59,19 @@ declare global {
 
     /** Simulation settings that control the amount of network & cpu throttling in the run. */
     interface ThrottlingSettings {
-      // The round trip time in milliseconds
+      /** The round trip time in milliseconds. */
       rttMs?: number;
-      // The network throughput in kilobytes per second
+      /** The network throughput in kilobytes per second. */
       throughputKbps?: number;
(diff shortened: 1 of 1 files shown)


types/externs.d.ts
```diff
@@ -143,31 +147,35 @@ declare global {
...
+      /** A flag to print the normalized config for the given config and options, then exit. */
       printConfig: boolean;
-      // Path to the file where precomputed lantern data should be read from.
+      /** Path to the file where precomputed lantern data should be read from. */
       precomputedLanternDataPath?: string;
-      // Path to the file where precomputed lantern data should be written to.
+      /** Path to the file where precomputed lantern data should be written to. */
       lanternDataOutputPath?: string;
+
+      // The following are given defaults in cli-flags, so are not optional like in Flags or SharedFlagsSettings.
```
exterkamp: Should this also be `/** */`?

brendankenny: > Or is it just // b/c it is a comment instead of an annotation for a single field?

exactly. It's an explanation for why the properties below that point are different (those three properties also have doc strings from their base definitions in `SharedFlagsSettings` and `Flags`)

Issue #7338, opened 2019-02-27
SEO score of 99 without further details


rentalh0sting:
We really like using LH (https://web.dev/measure) in helping us fix various issues on our websites and we consistently can get an SEO score of 100 for all of our websites, except for https://chiceventrentals.com/ where LH shows a score of 99 and without any further details on why it's not 100.   Can you please let me know why LH is not showing what needs to be corrected?

![image](https://user-images.githubusercontent.com/10382057/53513667-b8102e00-3a7a-11e9-8903-8bbc5a7ce752.png)

patrickhulce:
Thanks for filing @rentalh0sting. The failure is with tap targets.

![image](https://user-images.githubusercontent.com/2301202/53513995-8a77b480-3a7b-11e9-85b1-25831dcfbbd5.png)

cc @rviscomi @mattzeunert it is a bit weird that we'll sweep it into "Passing" if the score is high enough. Maybe we should keep it below 90 if there are any?

mattzeunert:
Agreed that that the failures should be more prominent. Thanks @rentalh0sting for letting us know!

The other SEO/Accessibility audits don't have this problem because they return a binary pass/fail result rather than returning a score. E.g the font audit size has a 40% illegible text threshold before failing. But having a score for tap targets makes sense, since it rewards fixing individual problems.

> Maybe we should keep it below 90 if there are any?

Yeah, multiplying the score by 0.89 if it's below 100 sounds good. An alternative would be to keep the score at 100 if more than 90% of tap targets pass, but that feels like sweeping problems under the rug.

mattzeunert:
I talked to @rviscomi and he'd prefer if there was a way to show the results without affecting the score. Maybe we can vary the pass threshold per audit, or do we have ways to highlight "passing but score < 90" in the audit somehow?

If not, multiplying by 0.89 is ok too.

patrickhulce:
> I talked to @rviscomi and he'd prefer if there was a way to show the results without affecting the score

@rviscomi as in you want the audit to be passing, but shown along with the other failing results?

Our position is that if it's passing it means it is not worth looking at, i.e. it should be collapsed by default and shown with the other hidden by default passing audits. If there's something that needs to be looked at, it should be failing. I'd say go with `* .9`/`Math.min(score, .89)`/`targets.length ? 0 : 1` if each and every failure should be looked at.

rviscomi:
[...]

PR #7373, opened 2019-03-04
core(tap targets): make tap targets failures more visible

mattzeunert:
**Summary**

Right now if the audit score is above 90 the tap targets audit doesn't fail even if there are overlap failures. That way it's hard to identify what's dragging the SEO score down. So if there are failures we should cap the score at 89.

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/7338

### DIFF
--- a/lighthouse-core/audits/seo/tap-targets.js
+++ b/lighthouse-core/audits/seo/tap-targets.js
@@ -302,8 +302,15 @@ class TapTargets extends Audit {
     const failingTapTargetCount = new Set(overlapFailures.map(f => f.tapTarget)).size;
     const passingTapTargetCount = tapTargetCount - failingTapTargetCount;
 
-    const score = tapTargetCount > 0 ? passingTapTargetCount / tapTargetCount : 1;
-    const displayValue = str_(UIStrings.displayValue, {decimalProportion: score});
+    let score = 1;
+    let passingTapTargetRatio = 1;
+    if (failingTapTargetCount > 0) {
+      passingTapTargetRatio = (passingTapTargetCount / tapTargetCount);
+      // If there are any failures then we don't want the audit to pass,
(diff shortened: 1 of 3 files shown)


lighthouse-core/audits/seo/tap-targets.js
```diff
@@ -302,8 +302,15 @@ class TapTargets extends Audit {
...
 
-    const score = tapTargetCount > 0 ? passingTapTargetCount / tapTargetCount : 1;
-    const displayValue = str_(UIStrings.displayValue, {decimalProportion: score});
+    let score = 1;
+    let passingTapTargetRatio = 1;
+    if (failingTapTargetCount > 0) {
+      passingTapTargetRatio = (passingTapTargetCount / tapTargetCount);
+      // If there are any failures then we don't want the audit to pass,
+      // so keep the score below 90.
+      score = Math.min(passingTapTargetRatio * 0.9, 0.89);
```
patrickhulce: do we need to do both? isn't just doing 1 of these sufficient?

mattzeunert: Yeah just `* 0.89` is enough.

I was trying to interpret [this](https://github.com/GoogleChrome/lighthouse/issues/7338#issuecomment-468723187) and thought maybe you meant that multiplying by 0.9 gives a smoother score or something.

> I'd say go with `* .9/Math.min(score, .89)/targets.length ? 0 : 1` if each and every failure should be looked at.

patrickhulce: oh yeah was just trying to suggest using one of those not in combination :)

mattzeunert: Ahh, it makes sense now 🙂

PR #7374, opened 2019-03-04
docs: fix throttling rate comment and tweak comcast bandwidth

mattzeunert:
**Summary**

Throttling rate is in kilobits not kilobytes.

LH uses a bandwidth of 1638kbps/1.6mpbs by default, so we should recommend using that with comcast too.

### DIFF
--- a/docs/throttling.md
+++ b/docs/throttling.md
@@ -48,7 +48,7 @@ Currently, `comcast` will also throttle the websocket port that Lighthouse uses 
 
 ```sh
 # Enable system traffic throttling
-comcast --latency=150 --target-bw=1600
+comcast --latency=150 --target-bw=1638
 
 # Run Lighthouse with its own throttling disabled
 lighthouse --throttling.requestLatencyMs=0 --throttling.downloadThroughputKbps=0 --throttling.uploadThroughputKbps=0 # ...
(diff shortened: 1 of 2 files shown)


paulirish:
the throttling.md change works for me.

But I'm not sure about the kilobytes => kilobits change. 

But looking at tcp-connection i'm getting even more confused..

https://github.com/GoogleChrome/lighthouse/blob/5865c873776b975a2c25632bc6a17c07b3c0050b/lighthouse-core/lib/dependency-graph/simulator/tcp-connection.js#L47

^ @patrickhulce why divide by 8? should this be called `bitsPerSecond` instead?

paulirish:
Took another look and Lantern is good. Also confirmed that DevTools's `emulateNetworkConditions` speaks in **bytes** per second. 

That leads me to wonder why we never convert our configuration value... My best guess is it's somehow wrapped up in the `DEVTOOLS_THROUGHPUT_ADJUSTMENT_FACTOR` but that seems very odd...  @patrickhulce plz tell me i'm being really dumb here. :)

patrickhulce:
We do convert our numbers before sending to DevTools

https://github.com/GoogleChrome/lighthouse/blob/5865c873776b975a2c25632bc6a17c07b3c0050b/lighthouse-core/lib/emulation.js#L107-L110

Nice catch @mattzeunert yeah the lowercase `b` in `Kbps` was meant to be the signal for `kilobits per second`, not sure how we messed that doc string up 😆

PR #7389, opened 2019-03-05
Testing travis build time [do not commit]

paulirish:
Back in Nov 2018 our travis builds were very consistently taking 9-10 min. Lately they have been taking much longer.. 14-15min. 

image

This branch is off a commit in november, so we'd expect it to take 9-10 minutes for travis to build it...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 sudo: required
 language: node_js
-branches:
-  only:
-  - master
 matrix:
   include:
     - node_js: "8"


brendankenny:
compared to the original travis run for that commit: https://travis-ci.org/GoogleChrome/lighthouse/builds/454674401

everything's just a liiiiiittle bit slower (27 vs 34s, 20 vs 28s, 190 vs 228s), except the smoke tests which went from 179s to 385s.

(and to be fair, `yarn test-lantern` ran for this PR, which added 119s that didn't happen in the original PR)

paulirish:
Oops this is a PR build, where it takes the commits and merges them to current master.. and builds that. 

I just allowed travis to build this branch solo. So this should be much more comparable... :)

image

paulirish:
This time the differences were more subtle. Total runtime was 11:29 (compared to 9:55). Worse, but not a LOT worse.

Smoke tests had the biggest difference again: 178s => 208s. That's ~17% slower.

Our new smoketests certainly had an impact. And travis generally just runs them slower than it used to. Perhaps something going on with xvfb.
```diff
- yarn smoke ally pwa pwa2 pwa3 dbw redirects seo offline byte perf tti
+ yarn smoke a11y pwa pwa2 pwa3 dbw redirects seo offline byte perf errors oopif lantern metrics
```

PR #7393, opened 2019-03-06
core(tap-targets): disable font size and tap targets audit on desktop

mattzeunert:
**Summary**

These two audits aren't important to desktop SEO, so we should disable them there.

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/6687

### DIFF
--- a/lighthouse-core/audits/seo/font-size.js
+++ b/lighthouse-core/audits/seo/font-size.js
@@ -207,7 +207,7 @@ class FontSize extends Audit {
       title: str_(UIStrings.title),
       failureTitle: str_(UIStrings.failureTitle),
       description: str_(UIStrings.description),
-      requiredArtifacts: ['FontSize', 'URL', 'MetaElements'],
+      requiredArtifacts: ['FontSize', 'URL', 'MetaElements', 'TestedAsMobileDevice'],
     };
   }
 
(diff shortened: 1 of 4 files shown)


paulirish:
> Is the alternative to this to have a separate desktop-config

Doing a desktop run is already something the config allows users to do (`disableDeviceEmulation`). In this PR's approach we take on the complexity and dynamically change the runtime based on the passed in config. 

The alternative is we pass the complexity to the user and have them choose which audits are appropriate given their selected configuration. I think we'd agree that is less desirable.

jabcreations:
This kind of screw up is why I won't show Lighthouse to clients inquiring. If I have to explain your bugs for you and then show not one though two developers who think it's perfectly acceptable to run a mobile test for desktop then the entire test is invalid. That applies to all my clients for my web platform business.

paulirish:
@jabcreations hmmmm..  you don't?

The change here (from 5 years ago) made it so you don't need to worry about this.

font-size and tap-target audits are only evaluated if it's a mobile run. if it's a desktop run then these two audits will be passing. 

From an SEO POV these checks are moot in desktop.

jabcreations:
A desktop device test with Lighthouse in Chrome 122.0:
seo

So yeah, either the change didn't work or someone has since then broken it.

paulirish:
for eng team:

reproable on https://www.jabcreations.com/features/

image

and here's it in CDP: [`CSS.createStyleSheet`](https://chromedevtools.github.io/devtools-protocol/tot/CSS/#method-createStyleSheet)

we're gonna drop this audit in exchange for axe, so this will get resolved via that. ✅

PR #7407, opened 2019-03-07
 cli: add --plugins flag to load from the command line

brendankenny:
Should make e.g. the [recipe](https://github.com/GoogleChrome/lighthouse/pull/6965/) a lot clearer since you don't need a config file.

Old:
```js
{
  plugins: ['lighthouse-plugin-preload-as'],
}
```
`lighthouse --config-path=my-config.json ...`

New:
`lighthouse --plugins=lighthouse-plugin-preload-as ...`

<hr>

Two things:
- `--plugins` or `--plugin`? `--plugin` seems like it might be more natural (`--plugin=lighthouse-plugin-preload-as`) and it's [what e.g. `eslint` does](https://eslint.org/docs/user-guide/command-line-interface#options), but technically you can have multiple values (`--plugin=plugin1 plugin2`), `plugins` is the property name in the config, and we use the plural for other possibly single things, e.g. `--only-categories`.
- ~~Since `plugins` isn't a setting (it's a [top level field](https://github.com/GoogleChrome/lighthouse/blob/35b285b87d681905777ed2545e48292312910485/types/config.d.ts#L23) in the config), it can't go through our normal flags/settings pipeline and instead has to be yet another optional argument to `run.runLighthouse()` and the `lighthouse()` module interface. This is ugly enough that I almost don't want to do it, but we've already talked about switching to an options object for v5, so maybe it's fine for now?~~ (see below)

### DIFF
--- a/lighthouse-cli/test/cli/run-test.js
+++ b/lighthouse-cli/test/cli/run-test.js
@@ -19,42 +19,71 @@ const fastConfig = {
   },
 };
 
+// Map plugin name to fixture since not actually installed in node_modules/.
+jest.mock('lighthouse-plugin-simple', () => {
+  return require('../../../lighthouse-core/test/fixtures/config-plugins/lighthouse-plugin-simple');
+}, {virtual: true});
+
 const getFlags = require('../../cli-flags').getFlags;
 
 describe('CLI run', function() {
-  it('runLighthouse completes a LH round trip', () => {
-    const url = 'chrome://version';
+  describe('LH round trip', () => {
+    /** @type {LH.RunnerResult} */
+    let passedResults;
     const filename = path.join(process.cwd(), 'run.ts.results.json');
-    const timeoutFlag = `--max-wait-for-load=${9000}`;
-    const flags = getFlags(`--output=json --output-path=${filename} ${timeoutFlag} ${url}`);
-    return run.runLighthouse(url, flags, fastConfig).then(passedResults => {
-      if (!passedResults) {
-        assert.fail('no results');
-        return;
+    /** @type {LH.Result} */
+    let fileResults;
+
+    beforeAll(async () => {
+      const url = 'chrome://version';
+      const timeoutFlag = `--max-wait-for-load=${9000}`;
+      const flags = getFlags(`--output=json --output-path=${filename} ${timeoutFlag} ${url}`);
+      const plugins = ['lighthouse-plugin-simple'];
+
+      const rawResult = await run.runLighthouse(url, flags, fastConfig, plugins);
+
+      if (!rawResult) {
+        return assert.fail('no results');
       }
+      passedResults = rawResult;
 
-      const {lhr} = passedResults;
       assert.ok(fs.existsSync(filename));
-      /** @type {LH.Result} */
-      const results = JSON.parse(fs.readFileSync(filename, 'utf-8'));
-      assert.equal(results.audits.viewport.rawValue, false);
+      fileResults = JSON.parse(fs.readFileSync(filename, 'utf-8'));
+    }, 20 * 1000);
 
-      // passed results match saved results
-      assert.strictEqual(results.fetchTime, lhr.fetchTime);
-      assert.strictEqual(results.requestedUrl, lhr.requestedUrl);
-      assert.strictEqual(results.finalUrl, lhr.finalUrl);
-      assert.strictEqual(results.audits.viewport.rawValue, lhr.audits.viewport.rawValue);
-      assert.strictEqual(
-          Object.keys(results.audits).length,
-          Object.keys(lhr.audits).length);
-      assert.deepStrictEqual(results.timing, lhr.timing);
-      assert.ok(results.timing.total !== 0);
-
-      assert.equal(results.configSettings.channel, 'cli');
-
+    afterAll(() => {
       fs.unlinkSync(filename);
     });
-  }, 20 * 1000);
+
+    it('returns results that match the saved results', () => {
+      const {lhr} = passedResults;
+      assert.equal(fileResults.audits.viewport.rawValue, false);
+
+      // passed results match saved results
+      assert.strictEqual(fileResults.fetchTime, lhr.fetchTime);
+      assert.strictEqual(fileResults.requestedUrl, lhr.requestedUrl);
+      assert.strictEqual(fileResults.finalUrl, lhr.finalUrl);
+      assert.strictEqual(fileResults.audits.viewport.rawValue, lhr.audits.viewport.rawValue);
+      assert.strictEqual(
+          Object.keys(fileResults.audits).length,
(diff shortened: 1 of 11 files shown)


lighthouse-cli/test/cli/run-test.js
```diff
@@ -19,40 +19,67 @@ const fastConfig = {
...
+jest.mock('lighthouse-plugin-simple', () => {
+  return require('../../../lighthouse-core/test/fixtures/config-plugins/lighthouse-plugin-simple');
+}, {virtual: true});
+
 const getFlags = require('../../cli-flags').getFlags;
 
 describe('CLI run', function() {
-  it('runLighthouse completes a LH round trip', () => {
-    const url = 'chrome://version';
+  describe('LH round trip', () => {
```
brendankenny: split this up as it seems like there are a few different things being tested (and `channel` is about to be another check added to this)

lighthouse-cli/test/cli/run-test.js
```diff
@@ -19,40 +19,67 @@ const fastConfig = {
...
+
+    afterAll(() => {
+      fs.unlinkSync(filename);
+    });
+
+    it('returns results that match the saved results', () => {
+      const {lhr} = passedResults;
+      assert.equal(fileResults.audits.viewport.rawValue, false);
 
       // passed results match saved results
```
brendankenny: no idea why this is being checked this way, but going with it :)

lighthouse-core/test/config/config-test.js
```diff
@@ -705,11 +706,24 @@ describe('Config', () => {
...
       };
       const config = new Config(configJson, {configPath: configFixturePath});
-      const groupIds = Object.keys(config.groups);
       assert.deepStrictEqual(config.audits.map(a => a.path),
         ['installable-manifest', 'metrics', 'redirects', 'user-timings']);
-      assert.ok(groupIds.length === 1);
-      assert.strictEqual(groupIds[groupIds.length - 1], 'lighthouse-plugin-simple-new-group');
+    });
+
+    it('should append and use plugin-prefixed groups', () => {
```
brendankenny: @jburger424 I split the `should append audits and a group` in two because I was having trouble following it all in one :) Should be exactly the same functionality

jburger424: :+1: this is cleaner

brendankenny:
> What if instead, CLI could call into Config.mergePlugins if the plugins CLI flag was set? It's already doing the work to figure out configJSON based on a set of flags that affect the config, so adding one more conditional to check for plugins seems like a small price to pay for nuking all the other complexity to our entrypoints that would be introduced otherwise.

We could do this now (`generateConfig` is already exposed for `--print-config`, and might be a nice spot for this?), though the other entry points are going to have to do something like this too (joining `categoryIds` as an option passed in) to support plugins, so is it better to just centralize now?

brendankenny:
I think we'll want node module users to be able to use this too, but that just made me realize I made this way too complicated (I think).

We don't want it in `settings`, but it can still be on `flags`, which is passed into the config constructor. I think this can be dealt with entirely within `config.js`.

patrickhulce:
> We don't want it in settings, but it can still be on flags, which is passed into the config constructor. I think this can be dealt with entirely within config.js.

Perfect 👍

brendankenny:
ah, so much better

lighthouse-cli/test/cli/run-test.js
```diff
@@ -19,42 +19,73 @@ const fastConfig = {
...
+      assert.ok(passedResults.lhr.timing.total !== 0);
+    });
 
-      assert.equal(results.configSettings.channel, 'cli');
+    it('correctly sets the channel', () => {
+      assert.equal(passedResults.lhr.configSettings.channel, 'cli');
+    });
 
-      fs.unlinkSync(filename);
+    it('merged the plugin into the config', () => {
```
brendankenny: this is well beyond just stretching the "unit" part of "unit tests", but this check has trivial overhead and until we have a smoketest for plugins I'd like to leave this in

lighthouse-cli/test/cli/run-test.js
```diff
@@ -19,42 +19,73 @@ const fastConfig = {
   },
 };
 
+// Map plugin name to fixture since not actually installed in node_modules/.
+jest.mock('lighthouse-plugin-simple', () => {
+  return require('../../../lighthouse-core/test/fixtures/config-plugins/lighthouse-plugin-simple');
```
patrickhulce: `.js` of the real file?

lighthouse-core/test/config/config-test.js
```diff
@@ -727,6 +741,38 @@ describe('Config', () => {
...
+      const baseConfigJson = {
+        audits: ['installable-manifest'],
+        categories: {
+          myManifest: {
+            auditRefs: [{id: 'installable-manifest', weight: 9000}],
+          },
+        },
+      };
+      const baseFlags = {configPath: configFixturePath};
+      const simplePlugin = 'lighthouse-plugin-simple';
```
patrickhulce: nit: could we go with `PluginName` here? just easier to remember down below it's a string

lighthouse-core/config/config.js
```diff
@@ -459,22 +459,23 @@ class Config {
 
   /**
    * @param {LH.Config.Json} configJSON
+   * @param {LH.Flags=} flags
    * @param {string=} configDir
    * @return {LH.Config.Json}
    */
-  static mergePlugins(configJSON, configDir) {
-    const pluginNames = configJSON.plugins;
+  static mergePlugins(configJSON, flags, configDir) {
```
patrickhulce: these are like the only meaningful changes, right?

so clean ✨ 😃

lighthouse-cli/test/cli/run-test.js
brendankenny: > or we are making it work like a package so it's ok?

I was making it like a package (I added a `/` to make it clear it's a directory), but can also make it the js file if you'd rather

lighthouse-core/test/config/config-test.js
brendankenny: good call

lighthouse-core/config/config.js
brendankenny: > these are like the only meaningful changes, right?

yep, that's it! soooo much better

brendankenny: haha, I also just realized this was the original plan all along, so at least we got back to it :)

> so it seems ok to add this to the handful of special-cased flag/config mergings in config.js when we add support for the CLI flag.

https://github.com/GoogleChrome/lighthouse/pull/6959#issuecomment-452540140

lighthouse-cli/test/cli/run-test.js
patrickhulce: nah, it's cool

PR #7421, opened 2019-03-08
devtools: import html generator code

Hoten:
Bundle the HTML generation code. No longer import templates/stylesheets, as they are already in the new bundled fine.

[Chromium changes](https://chromium-review.googlesource.com/c/chromium/src/+/1510732)


lighthouse-core/scripts/roll-to-devtools.sh
```diff
@@ -37,9 +37,22 @@ lh_bg_js="dist/lighthouse-dt-bundle.js"
...
 echo -e "\033[96m ✓\033[39m (Potentially stale) lighthouse-dt-bundle copied."
+
+# bundle html generator
+node -e "
+  const htmlReportAssets = require('./lighthouse-core/report/html/html-report-assets.js');
+  const reportGenerator = require('./lighthouse-core/report/report-generator.js');
+  console.log('const htmlReportAssets =', JSON.stringify(htmlReportAssets, null, 2), ';');
+  console.log('class ReportGenerator {');
+  console.log('  static Assets = htmlReportAssets;');
+  console.log('  static', reportGenerator.generateReportHtml.toString());
```
patrickhulce: this feels....fragile to say the least :)

Hoten: What about bundling w/ browserify + this plugin? https://github.com/browserify/brfs

The file system access is what made it tricky to consume in DT.

Hoten: Not doing above, but I did a refactor. Changed the report generator module to load from Runtime.cachedAssets if the env is Devtools, and replaced the node bundling code with just copying those assets.

lighthouse-core/scripts/roll-to-devtools.sh
```diff
@@ -30,14 +30,23 @@ else
...
 fe_lh_dir="$frontend_dir/audits2/lighthouse"
 
 lh_bg_js="dist/lighthouse-dt-bundle.js"
 lh_worker_dir="$frontend_dir/audits2_worker/lighthouse"
 
 # copy report files
-cp -pPR $report_dir/{report-styles.css,templates.html,renderer} "$fe_lh_dir"
+cp -pPR $report_dir/report-generator.js "$fe_lh_dir"
+cp -pPR $report_dir/html/renderer "$fe_lh_dir"/renderer
+# import report assets
```
Hoten: say "import report" 10 times quickly :)

lighthouse-core/scripts/roll-to-devtools.sh
Hoten: Bonus - fewer bytes wasted from string escaping :)

brendankenny:
Seems like this has a large overlap with what viewer has to do to save the HTML report. Is it possible we could share that between the two at build time rather than check at runtime?

Hoten:
> Seems like this has a large overlap with what viewer has to do to save the HTML report. Is it possible we could share that between the two at build time rather than check at runtime?

good idea, I'll see what could be done.

Hoten:
updated chromium cl. for some reason there's this at the bottom of the downloaded report :/
image

Hoten:
> > Seems like this has a large overlap with what viewer has to do to save the HTML report. Is it possible we could share that between the two at build time rather than check at runtime?
> 
> good idea, I'll see what could be done.

At first glance, it seems like it'd be a large refactor and wouldn't be as simple to understand as the current code in `roll-to-devtools.sh`. Maybe wait for a future cleanup? At the least, tackle in a follow up PR.

brendankenny:
we crafted report generation to work with devtools before, going through a few solutions until finding a nice one where we could just cut off the dependency tree at `html-report-assets`. It would be great to make this work within that same solution rather than having the devtools worker be build-time and the panel be run-time.

Some ideas:

1) the original commit was pretty much just `report-generator.js` browserified but dropping the csv and json output lines. Those methods are pretty small, could we just include them unused and browserify with `brfs` as you mentioned? `build-viewer.js` does this with just
```js
const generatorFilename = `./lighthouse-core/report/report-generator.js`;
const generatorBrowserify = browserify(generatorFilename, {standalone: 'ReportGenerator'})
  .transform('brfs');
```
(which puts `ReportGenerator` on `window`/`global`/`self`)

2) if we really want to reuse assets loaded in the devtools bundle, we could write a custom `html-report-assets.js` just for devtools, which neatly keeps that division point. Something like
```js
// clients/devtools-report-assets.js
'use strict';

module.exports = {
  REPORT_TEMPLATE: Runtime.cachedResources['audits2/lighthouse/report-template.html'],
  REPORT_TEMPLATES: Runtime.cachedResources['audits2/lighthouse/report-templates.html'],
  REPORT_JAVASCRIPT: Runtime.cachedResources['audits2/lighthouse/report.js'],
  REPORT_CSS: Runtime.cachedResources['audits2/lighthouse/report.css'],
};
```
and then have browserify insert the replacement by adding
```js
const pathToReportAssets = require.resolve('./clients/devtools-report-assets.js');
generatorBrowserify.require(pathToURLShim, {expose: './html/html-report-assets'});
```

Hoten:
> 1. the original commit was pretty much just report-generator.js browserified but dropping the csv and json output lines. Those methods are pretty small, could we just include them unused and browserify with brfs as you mentioned? build-viewer.js does this with just

Did this. 2) would be nice in terms of reducing bytes (would remove tons of escaping), but would require the build script to copy the [assets](https://github.com/GoogleChrome/lighthouse/pull/7421/commits/af95b3a01d69e3ac63167acd148b7b993037ee31#diff-cc71e0547b282357dfff898cfa3f3211R43). but should we bother with the extra complexity?

lighthouse-core/report/report-generator.js
```diff
@@ -8,6 +8,10 @@
 const htmlReportAssets = require('./html/html-report-assets');
 
 class ReportGenerator {
+  static get Assets() {
```
brendankenny:
```suggestion
  static getAssets() {
```
any specific reason to have it as a getter?

lighthouse-core/scripts/roll-to-devtools.sh
```diff
@@ -37,9 +37,12 @@ lh_bg_js="dist/lighthouse-dt-bundle.js"
...
-cp -pPR $report_dir/{report-styles.css,templates.html,renderer} "$fe_lh_dir"
+cp -pPR $report_dir/renderer "$fe_lh_dir"
 echo -e "\033[32m ✓\033[39m Report renderer files copied."
 
 # copy lighthouse-dt-bundle (potentially stale)
 cp -pPR "$lh_bg_js" "$lh_worker_dir/lighthouse-dt-bundle.js"
 echo -e "\033[96m ✓\033[39m (Potentially stale) lighthouse-dt-bundle copied."
+
+# bundle report generator
+yarn browserify lighthouse-core/report/report-generator.js -o $fe_lh_dir/report-generator.js -s ReportGenerator -t brfs
```
brendankenny: I think it could be helpful to move this into a script in `build/` and output to `dist/dt-report-generator-bundle.js` (or whatever) and have `roll-to-devtools.sh` only responsible for copying things over. That way all our client building is all in one place and the requirements (e.g. browserify compatible) are plain to see.

Buuuut, this is also so succinct, if you don't want to move on that today I'm ok with that too :)

lighthouse-core/report/report-generator.js
Hoten: i thought it looked better when used: https://chromium-review.googlesource.com/c/chromium/src/+/1510732/5/third_party/blink/renderer/devtools/front_end/audits2/Audits2Panel.js

brendankenny: I think they're usually just lying as an API--maybe justified if you're trying to match an expected interface--but we don't really have an established style on this front so to each their own :)

Should be lowercase `assets`, though

lighthouse-core/scripts/roll-to-devtools.sh
Hoten: moved to `build/`

build/dt-report-generator-bundle.js
```diff
@@ -0,0 +1,12 @@
+const browserify = require('browserify');
```
brendankenny: Needs a license

build/dt-report-generator-bundle.js
```diff
@@ -0,0 +1,19 @@
...
+
+const distDir = __dirname + '/../dist';
+const outFile = `${distDir}/report-generator.js`;
+const generatorFilename = `./lighthouse-core/report/report-generator.js`;
+browserify(generatorFilename, {standalone: 'ReportGenerator'})
+  .transform('brfs')
+  .bundle((err, src) => {
+    if (err) throw err;
+    fs.writeFileSync(outFile, src.toString());
+  });
```
brendankenny: sweet, I think this might help @exterkamp as well. At least I think he was looking for a standalone report generator?

Hoten: Cool! He can just remove the `dt-` bit then, that'd be a great PR @exterkamp  :)

PR #7429, opened 2019-03-09
tests: finish removing mocha, fix CI errors

brendankenny:
another possible approach to fixing #7418/#7413, assuming something wonky is going on trying to do coverage over jest and mocha at once.

If it doesn't work...we've been meaning to finish getting rid of mocha anyways :)


lighthouse-core/test/config/config-test.js
```diff
@@ -1046,7 +1046,7 @@ describe('Config', () => {
 
     it('loads a gatherer from node_modules/', () => {
       // Use a lighthouse dep as a stand in for a module.
-      assert.throws(_ => loadGatherer('mocha'), function(err) {
+      assert.throws(_ => loadGatherer('lighthouse-logger'), function(err) {
```
brendankenny: these are testing that `Config.resolveModule` will look in `node_modules/` for audits or gatherers, so it just needs to be *something* in there.

package.json
```diff
@@ -50,7 +50,6 @@
     "fast": "yarn start --disable-device-emulation --throttlingMethod=provided",
     "deploy-viewer": "yarn build-viewer --deploy",
     "bundlesize": "bundlesize",
-    "plots-smoke": "bash plots/test/smoke.sh",
```
brendankenny: blast from the past

package.json
```diff
@@ -138,12 +137,10 @@
     "configstore": "^3.1.1",
     "cssstyle": "1.2.1",
     "details-element-polyfill": "2.2.0",
-    "esprima": "^4.0.1",
```
brendankenny: still used by `collect-strings.js`, but no longer in `unminified-javascript`, so just a devDependency

brendankenny:
> TypeError: Cannot read property 'decl' of undefined

no dice

brendankenny:
"fix CI errors" -> lies

Issue #6945, opened 2019-01-07
Unescaped HTML in translation text


exterkamp:
These translated strings contain unescaped HTML which makes the translation linter unhappy.

- `[user-scalable="no"]` is not used in the `<meta name="viewport">` element and the `[maximum-scale]` attribute is not less than 5.
- The document does not use `<meta http-equiv="refresh">`
- `[user-scalable="no"]` is used in the `<meta name="viewport">` element or the `[maximum-scale]` attribute is less than 5.
- `<input type="image">` elements have `[alt]` text
- The document uses `<meta http-equiv="refresh">`
- `<input type="image">` elements do not have `[alt]` text

brendankenny:
This is already a problem in some of our strings. The translators usually kept the tags unchanged, but sometimes treated things like attribute values as you would the `alt` tag on an image and translated it. e.g. `<link rel=preload>` became [`<link rel=предварительную загрузку>` in Russian](https://github.com/GoogleChrome/lighthouse/blob/809da7b6caa92626741a54cff538453e4609b081/lighthouse-core/lib/i18n/locales/ru.json#L255).

brendankenny:
Apparently the "proper" way to do this is to have the html inserted as an ICU text replacement. This makes it guaranteed that the translators won't translate them. It's also nice to give an example to the translators of what will be in the replacement...in this case, since it's just a literal string, the example is exactly what will be put in the replacement.

We didn't see another easy way to be able to mark within a string "don't translate this section", at least with the format we've chosen (the XML form has a little more flexibility, but come on).

The replacement approach has pretty rotten UX in our current i18n setup, however, since the substrings are just literals, not used as markup, and they aren't dynamic. The string and the html substring end up separated and it becomes difficult to read.

We could:
- embrace this and just try to remember to do string replacement if we need a substring untranslated
- always remember to put something in the description to leave part of the message untranslated (I believe we've done that in a few descriptions already)
- add support for some kind of demarcator to `i18n.js` that automatically extracts substrings and adds them to the replacements that will be done when the formatted string is retrieved (kind of `Function.p.bind` for our strings). We could automatically add a mention of the extracted substring to the description in that case?

patrickhulce:
> add support for some kind of demarcator to i18n.js that automatically extracts substrings and adds them to the replacements that will be done when the formatted string is retrieved

This was my thought as well.

> always remember to put something in the description to leave part of the message untranslated (I believe we've done that in a few descriptions already)

[...]

PR #7447, opened 2019-03-11
Wrap HTMl in markdown code

StefanoChiodino:
**Summary**

I noticed the link tag was being rendered as actual HTML in my application, needs to be escaped!

**Related Issues/PRs**

n/a


StefanoChiodino:
@patrickhulce thanks, that's good to know. I've unknowingly been converting descriptions to markdown, and that was the only problem I've found. If this is not the way to go, I'll just encode HTML markup or something :)

brendankenny:
looks like we have six uses of backticks in audits' `description`, all in accessibility (probably written in a flow of writing descriptions, or assuming we support backticks in there the same way we do for titles).

I vote for deferring to a solution for #6945 before committing more to backticks for this. And we probably also need to be more consistent on what subset of markdown features we use since folks are using our strings from the LHR  (web.dev is another example) and it's hard to know that sometimes we use \` to make preformatted code but that we're also comfortable using \` to mark cases that *would* be formatted if we were formatting, but we aren't :)

@StefanoChiodino in the meantime, except for markdown links (in some strings) and backticks (in other strings), all our strings are intended to be plain text, so you should be ok just doing a normal html escape on them to get them to display properly.

PR #7473, opened 2019-03-12
misc(coverage): drop negated coverageReporters from config

paulirish:
We have collectCoverage as `false` by default but flip it on with `--coverage` in CI.  But then `coverageReporters` is `none` which doesn't make sense when coverage is on. 

Anyway, I guess jest changed something recently. 

This seems fine and I think resolves our coverage collecting issues. (Or maybe just one more of them)

https://jestjs.io/docs/en/configuration.html#coveragereporters-array-string
https://jestjs.io/docs/en/cli#coverage


patrickhulce:
> Is none even a valid value, though?

Yes, [it is.](https://github.com/gotwarlost/istanbul/blob/master/lib/report/none.js), [full list](https://github.com/gotwarlost/istanbul/tree/master/lib/report)

paulirish:
> but the default value is ["json", "lcov", "text", "clover"], which I don't think we want.

Since this has caused us so many problems I feel better sticking to the defaults than trying to be selective. At most the extra formatters are costing us like 5s, which aint too bad IMO.

merged master to rekick travis.

brendankenny:
> At most the extra formatters are costing us like 5s, which aint too bad IMO.

they both write to `coverage/`, so I'm not sure what's going to happen when nyc tries to report from there

paulirish:
okay. the results seem odd. its gone.

PR #7478, opened 2019-03-12
clients(lightrider): always get transferSize from X-TotalFetchedSize header

Hoten:
After inspecting the devtools log locally with the NS on and off, I'm mostly convinced that Chrome is right, and LH needs to be tweaked

The pattern dataLength=x,encodedDataLength=0 + dataLength=0,encodedDataLength=y only happens for resources that load in one chunk. And x = y only when the content is not compressed. It seems that if all the data for a resource comes in one chunk, the first "dataReceived" event always has 0 for encodedLength, but the second event will have the entire encoded length. If the resource is loaded in multiple chunks, the first dataReceived event has some of the total encoded length, the next won't have much or any, and the remaining unreported encoded length is on the last event. Inspecting the logic in the code linked in `#3` seems to support this.

I think the behavior before could be considered incorrect, and the "bugged" version is what should be expected. An encodedDataLength of zero doesn't really make sense, but we were depending on it :) In other words, no issue with Chrome, let's fix LH.

Fixes internal bug b/123996856. Googlers, see internal changes cl/238109521.


brendankenny:
even the comment in the PR itself is a better PR description than this one :P :P

Hoten:
Didn't want to repeat myself ;)

patrickhulce:
> It seems that if all the data for a resource comes in one chunk, the first "dataReceived" event always has 0 for encodedLength, but the second event will have the entire encoded length. If the resource is loaded in multiple chunks, the first dataReceived event has some of the total encoded length, the next won't have much or any, and the remaining unreported encoded length is on the last event

I had thought that in WRS loaders there was no incremental loading on the Chrome side and it always appeared to showup in one big chunk over the protocol, or at least that's what early devtools logs appeared to show, so if that wasn't true, it certainly explains the troubles we see now.

Hoten:
> > It seems that if all the data for a resource comes in one chunk, the first "dataReceived" event always has 0 for encodedLength, but the second event will have the entire encoded length. If the resource is loaded in multiple chunks, the first dataReceived event has some of the total encoded length, the next won't have much or any, and the remaining unreported encoded length is on the last event
> 
> I had thought that in WRS loaders there was no incremental loading on the Chrome side and it always appeared to showup in one big chunk over the protocol, or at least that's what early devtools logs appeared to show, so if that wasn't true, it certainly explains the troubles we see now.

That is true. Orthogonal to how many dataReceived events are seen is - whether or not encodedDataLength is set on them. The weird pattern of `dataLength=x,encodedDataLength=0 + dataLength=0,encodedDataLength=y` stuck out because of the guaranteed single-chunk loading in WRS, and looked like a weird bug at first, until realizing that a request with 0 encodedDataLength doesn't make much sense.

paulirish:
Copying from the internal thread...

Here's how I understand things.. 

#### "Correct" pattern (we see in desktop chrome regardless of NS) when data is received in one chunk:
1. Network.responseReceived: ...
1. Network.dataReceived: {dataLength:`<allbytes>`, encodedDataLength:0},
1. Network.dataReceived: {dataLength:0,            encodedDataLength:`<gzippedbytes>`},

#### Old WRS:
1. Network.responseReceived {headers: {X-TotalFetchedSize: `<gzippedbytes>`}}
1. Network.dataReceived: {dataLength:`<allbytes>`, encodedDataLength:0},  _(encoded=0 IS A LIE)_

#### ToT WRS:
1. Network.responseReceived {headers: {X-TotalFetchedSize: `<gzippedbytes>`}}
1. Network.dataReceived: {dataLength:`<allbytes>`, encodedDataLength:0},
1. Network.dataReceived: {dataLength:0,            encodedDataLength:`<allbytes>`}, _(encoded=allbytes IS A LIE)_


This seems correct to everyone?

if so...

Due to reasons, we can't have encodedDataLength populated correctly in WRS, so we have to accept a lie. So we **have to** handle the special situation here on the LH side.

`ResponseReceived` is received first and already has the full encodedDataLength. 

**Proposal**: If this is Lightrider and we receive that header, then we use that as our `transferSize` and ignore all the `encodedDataLength`'s. 
That's roughly in line with your PR here but I'm wondering if we should set transferSize (via `_updateTransferSizeForLightRiderIfNecessary`) from inside of `onResponseReceived()` instead. wdyt?

patrickhulce:
> "Correct" pattern (we see in desktop chrome regardless of NS) when data is received in one chunk:

In practice I see, `Network.loadingFinished` being the actual source of all transfer size information the normal case. We may get those `encodedDataLength` events, but they always get overridden by `loadingFinished`. Otherwise that seems correct to me!

> That's roughly in line with your PR here but I'm wondering if we should set transferSize (via _updateTransferSizeForLightRiderIfNecessary) from inside of onResponseReceived() instead. wdyt?

seems fine to me 👍

paulirish:
> In practice I see, Network.loadingFinished being the actual source of all transfer size information the normal case.

Ah I see. loadingFinished's value should override the sum of dataReceived's. I guess we'd use the dataReceived ones if it was loadingFailed or some other weird situation. mkay.

Okay in that case, let's continue to set LR transferSize here in loadingFinished, but let's add a shitton of comments that explain this shit so we don't have to research it so hard next time. 

YES? YES.

Hoten:
> Here's how I understand things..
> This seems correct to everyone?

Yupper

>  I'm wondering if we should set transferSize (via `_updateTransferSizeForLightRiderIfNecessary`) from inside of `onResponseReceived()` instead. wdyt?

`_updateTransferSizeForLightRiderIfNecessary` is currently called for `onResponseReceived` and `onLoadingFinished`. The order goes:

```
responseReceived
dataReceived
dataReceived
loadingFinished / loadingFailed
```

It seems we could remove the call to `_updateTransferSizeForLightRiderIfNecessary` in `responseReceived` and just have it in `loadingFinished` (and maybe `loadingFailed` too?). We could just have it in just `responseReceived`, but we'd also need to not accumulate the `encodedDataLength` as `dataReceived` events are processed in LR.

options as I see it;
1. call `_updateTransferSizeForLightRiderIfNecessary` in `loadingFinished` / `loadingFailed`
2. call `_updateTransferSizeForLightRiderIfNecessary` in `responseReceived` and never accumulate `encodedDataLength` on `dataReceived` in LR.

brendankenny:
Can we keep the changes for lightrider limited to a single place? What happened to

> and just have it in `loadingFinished` (and maybe `loadingFailed` too?).

Seems like the new behavior is the same as in 8406bc9 (disregard normal way of setting `this.transferSize`, use the header instead) if we just move the `_updateTransferSizeForLightRiderIfNecessary` call to finished/failed?

We should also get some tests in for this.

patrickhulce:
I'd strongly prefer to just have `_updateTransferSizeForLightRiderIfNecessary` be called as frequently as necessary (loadingFinished/loadingFailed) then have extra logic for not accumulating on `dataReceived` :)

Hoten:
doneski :)

lighthouse-core/lib/network-request.js
```diff
@@ -246,6 +247,18 @@ module.exports = class NetworkRequest {
 
     this.responseReceivedTime = timestamp;
 
+    this.responseHeadersText = response.headersText || '';
```
brendankenny: move these lines back?

lighthouse-core/lib/network-request.js
```diff
@@ -246,6 +247,18 @@ module.exports = class NetworkRequest {
...
+
+    // The total length of the encoded data is spread out among multiple events. The sum of the
+    // values in onResponseReceived and all the onDataReceived events will equal the value
+    // seen on the onLoadingFinished event. As we process onResonseReceived and onDataReceived
+    // we accumulate the total encodedDataLength. When we process onLoadingFinished, we override
+    // the accumulated total. We do this so that if the request is aborted or fails, we still get
+    // a value via the accumulation.
+    // In Lightrider, we do not have true value for encodedDataLength, and we get the actual size
+    // of the encoded data via a special response header. Because the values are totally bogus,
+    // we do no accumulation.
```
brendankenny: move comment down to `_updateTransferSizeForLightrider`?

brendankenny:
> Or we can trim down to just the requests we care about?

like ...

```js
json.filter(entry => {
  return entry.method.startsWith('Network') &&
    ['B38A026891F997E88601C20D85DC4616', '1000170085.13'].includes(entry.params.requestId);
});
```

?

:)

lighthouse-core/lib/network-request.js
```diff
@@ -246,6 +247,18 @@ module.exports = class NetworkRequest {
 
     this.responseReceivedTime = timestamp;
 
+    this.responseHeadersText = response.headersText || '';
+    this.responseHeaders = NetworkRequest._headersDictToHeadersArray(response.headers);
+
+    // The total length of the encoded data is spread out among multiple events. The sum of the
+    // values in onResponseReceived and all the onDataReceived events will equal the value
+    // seen on the onLoadingFinished event. As we process onResonseReceived and onDataReceived
```
paulirish: > The sum of the
>    // values in onResponseReceived and all the onDataReceived events will equal the value
>    // seen on the onLoadingFinished event. 

I don't know exactly what leads to a situation when the loadingFinished total is different than the sum of the parts, but apparently it does happen.

I just tried against 6 saved devtoolslogs i have. (Disclaimer: the artifacts were collected months ago definitely without networkservice, though I wouldn't expect it to matter)... There's, 320 network requests in total and _nearly all_ of them have a sum that matches (or the sum was 0).. But I found 3 requests where this wasn't the case.

| encodedSum | encodedTotalInLoadingFinished | url |
|-|-|-|
| 108 | 12640 |  http://localhost:10200/dobetterweb/dbw_tester.html |
| 139 | 144 |  http://localhost:10200/dobetterweb/empty_module.js?delay=500 |
| 650 | 714 |  https://s.amazon-adsystem.com/x/da2e6c890e6e3636 |

Again.. don't know what leads to this situation, but it happens. I'm happy with admitting that in the comment or we can ask caseq/network people about the circumstances..

lighthouse-core/lib/network-request.js
```diff
@@ -246,6 +247,18 @@ module.exports = class NetworkRequest {
...
+    // we accumulate the total encodedDataLength. When we process onLoadingFinished, we override
+    // the accumulated total. We do this so that if the request is aborted or fails, we still get
+    // a value via the accumulation.
+    // In Lightrider, we do not have true value for encodedDataLength, and we get the actual size
```
paulirish: > we do no accumulation.

Looks like we do, as of now.. We just override the transferSize in loadingFinished

Thank you VERY MUCH for writing all this shit out. These comments are so Good. 

here's a very small rewrite of this second para for clarity:

> In Lightrider, due to instrumentation limitations, our values for encodedDataLength are bogus and not valid. However the resource's true `encodedDataLength`/`transferSize` is shared via a  special response header, `X-TotalFetchedSize`. In this situation, we read this value from responseReceived, use it for the `transferSize` and ignore the encodedDataLength values in both dataReceived and loadingFinished.

Hoten: oops thanks for the revisions :heart:

lighthouse-core/lib/network-request.js
```diff
@@ -297,12 +296,24 @@ module.exports = class NetworkRequest {
 
   /**
    * LR loses transfer size information, but passes it in the 'X-TotalFetchedSize' header.
+   * 'X-TotalFetchedSize' is the canonical transfer size in LR. Nothing should supersede it.
+   *
+   * The total length of the encoded data is spread out among multiple events. The sum of the
+   * values in onResponseReceived and all the onDataReceived events will equal the value
```
paulirish: will equal => typically equals

and add 

> In <1% of cases we see the values differ.

lighthouse-core/lib/network-request.js
```diff
@@ -297,12 +296,24 @@ module.exports = class NetworkRequest {
...
+   * In Lightrider, due to instrumentation limitations, our values for encodedDataLength are bogus
+   * and not valid. However the resource's true encodedDataLength/transferSize is shared via a
+   * special response header, X-TotalFetchedSize. In this situation, we read this value from
+   * responseReceived, use it for the transferSize and ignore the encodedDataLength values in
+   * both dataReceived and loadingFinished.
    */
-  _updateTransferSizeForLightRiderIfNecessary() {
-    // Bail if we're not in LightRider, this only applies there.
-    if (!global.isLightRider) return;
+  _updateTransferSizeForLightrider() {
```
paulirish: can you also leave a comment on L70 that says go read the _updateTransferSizeForLightrider comment?

lighthouse-core/lib/network-request.js
```diff
@@ -297,12 +297,24 @@ module.exports = class NetworkRequest {
 
   /**
    * LR loses transfer size information, but passes it in the 'X-TotalFetchedSize' header.
+   * 'X-TotalFetchedSize' is the canonical transfer size in LR. Nothing should supersede it.
+   *
+   * The total length of the encoded data is spread out among multiple events. The sum of the
+   * values in onResponseReceived and all the onDataReceived events typically equals the value
+   * seen on the onLoadingFinished event. In <1% of cases we see the values differ. As we process
+   * onResonseReceived and onDataReceived we accumulate the total encodedDataLength. When we
```
patrickhulce: typo `onResponseReceived`

lighthouse-core/lib/network-request.js
```diff
@@ -297,12 +297,24 @@ module.exports = class NetworkRequest {
 
   /**
    * LR loses transfer size information, but passes it in the 'X-TotalFetchedSize' header.
+   * 'X-TotalFetchedSize' is the canonical transfer size in LR. Nothing should supersede it.
+   *
+   * The total length of the encoded data is spread out among multiple events. The sum of the
+   * values in onResponseReceived and all the onDataReceived events typically equals the value
+   * seen on the onLoadingFinished event. In <1% of cases we see the values differ. As we process
```
patrickhulce: fwiw in ~70% of cases in my local devtoolslog of a run I just did the final `encodedDataLength` differs from the sum of the events, not sure if I'm broken or the something else has changed :)

lighthouse-core/lib/network-request.js
```diff
@@ -297,12 +297,24 @@ module.exports = class NetworkRequest {
...
+   * onResonseReceived and onDataReceived we accumulate the total encodedDataLength. When we
+   * process onLoadingFinished, we override the accumulated total. We do this so that if the
+   * request is aborted or fails, we still get a value via the accumulation.
```
patrickhulce: > or fails

did we decide not to override for `onLoadingFailed`?

Hoten: `_updateTransferSizeForLightrider` is being called on failed now, or did you mean  something else?

lighthouse-core/lib/network-request.js
Hoten: sent a pr

lighthouse-core/lib/network-request.js
Hoten: interesting. could you or paul share your test scripts / LH code diff / devtool logs? We should control for what logs we are using + NS  + Chrome version so we know what's actually happening.

lighthouse-core/lib/network-request.js
patrickhulce: oh I just missed it and comment didn't mention it so I got confused, we're all good :)

lighthouse-core/lib/network-request.js
patrickhulce: oh I just looked at the last latest-run/defaultPass.devtoolslog.json I had, so it's probably fairly old

Issue #7041, opened 2019-01-17
Beacon "impossible" cases to Sentry


patrickhulce:
**Summary**
There are several "impossible" branches of code that should never happen but could be happening in the wild. It'd be nice to beacon these to Sentry to see what's going on.

*Note:* Created from a TODO in code so it can be more visible and properly prioritized

PR #7513, opened 2019-03-14
core(lantern): FCP improvements

patrickhulce:
**Summary**

While doing some investigation into resilient FCP, I noticed a few cases popping up where we could be more accurate.

Specifically:

- Previously, if a network request was for a script, it finished before FCP, and it had a render blocking priority, we would include it in the FCP graph *even if the EvaluateScript task for that script happened after FCP*. This doesn't make sense so we now exclude such scripts. This in particular was a source of variance as sometimes FCP would happen to be before the script finished downloading and sometimes it wouldn't. Now such scripts will always be excluded.
- Previously, we were only including EvaluateScript tasks for blocking scripts, no other CPU nodes were included. That frequently misses some pretty big/important tasks like parsing the HTML, the first layout, etc. We now include the first layout/paint/parse HTML task if they happened before FCP.

**Impact**

Huge boosts to our FCP accuracy!! 🎉 FMP takes a slight hit to be consistent here, but it doesn't move any sites to worse buckets and FCP has a much higher priority/visibility anyway.

This also shrunk the observed variance for the Hulu case from +/- 2000ms to +/-200ms, which is far more tolerable.

image


lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+
+    if (cpuFilter) cpuNodes.filter(cpuFilter).forEach(node => blockingCpuNodeIds.add(node.id));
+
+    return {possiblyRenderBlockingScriptUrls, actuallyRenderBlockingScriptUrls, blockingCpuNodeIds};
   }
 
   /**
    * @param {Node} dependencyGraph
-   * @param {LH.Artifacts.TraceOfTab} traceOfTab
+   * @param {number} fcpTs
```
brendankenny: some docs would be good. Not at clear from the outside why this is called `fcpTs` but `lantern-first-meaningful-paint.js` is passing in `fmp`, for instance :)

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
   /**
-   * @param {Node} dependencyGraph
-   * @param {LH.Artifacts.TraceOfTab} traceOfTab
-   * @return {Node}
+   *
+   * @param {Node} graph
+   * @param {number} fcpTs
+   * @param {function(NetworkNode):boolean} networkFilter
+   * @param {function(CPUNode):boolean=} cpuFilter
+   * @return {{possiblyRenderBlockingScriptUrls: Set<string>, actuallyRenderBlockingScriptUrls: Set<string>, blockingCpuNodeIds: Set<string>}}
```
brendankenny: `actually` reads like `possibly` blocking nodes really weren't. `definitelyRenderBlockingScriptUrls`? :)

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
 
   /**
-   * @param {Node} dependencyGraph
-   * @param {LH.Artifacts.TraceOfTab} traceOfTab
-   * @return {Node}
+   *
+   * @param {Node} graph
+   * @param {number} fcpTs
+   * @param {function(NetworkNode):boolean} networkFilter
+   * @param {function(CPUNode):boolean=} cpuFilter
```
brendankenny: does `{(function(CPUNode):boolean)=}` work? `=` isn't supposed to be applicable to return types, but with tsc you never know, so it ends up hard to read

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+      for (const cpuNode of cpuNodes) {
+        if (cpuNode.isEvaluateScriptFor(new Set([url]))) {
+          actuallyRenderBlockingScriptUrls.add(url);
+          blockingCpuNodeIds.add(cpuNode);
+          break;
+        }
+      }
+    }
+
+    const firstLayout = cpuNodes.find(node => node.didPerformLayout());
```
brendankenny: comment over this block would be good too

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+
+    return {possiblyRenderBlockingScriptUrls, actuallyRenderBlockingScriptUrls, blockingCpuNodeIds};
   }
 
   /**
    * @param {Node} dependencyGraph
-   * @param {LH.Artifacts.TraceOfTab} traceOfTab
+   * @param {number} fcpTs
+   * @param {function(NetworkNode):boolean} blockingScriptFilter
+   * @param {function(CPUNode):boolean=} cpuNodeFilter
```
brendankenny: `cpuNodeFilter` does the opposite of what I expected...maybe incorporate `block`/`blocking` into its name as well?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,109 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+      if (node.type === BaseNode.TYPES.NETWORK) {
+        // Exclude all nodes that ended after FCP (except for the main document which we always consider necessary)
+        if (node.endTime > fcpTs && !node.isMainDocument()) return false;
 
-      // Include non-script-initiated network requests with a render-blocking priority
-      return node.hasRenderBlockingPriority();
+        const url = node.record.url;
+        if (possiblyRenderBlockingScriptUrls.has(url))
+          return actuallyRenderBlockingScriptUrls.has(url);
+        return node.hasRenderBlockingPriority();
```
brendankenny: why override the check in `hasRenderBlockingPriority` if the url is in `possiblyRenderBlockingScriptUrls` but not in `actuallyRenderBlockingScriptUrls`?

paulirish:
> Previously, if a network request was for a script, it finished before FCP, and it had a render blocking priority, we would include it in the FCP graph even if the EvaluateScript task for that script happened after FCP. This doesn't make sense so we now exclude such scripts

Yeah this doesn't make sense at all. :) Like.... the script is *render blocking* so why would chrome do a real contentful paint before the evaluateScript??? That's not really supposed to happen. Can we look closer at what's going on here?


> We now include the first layout/paint/parse HTML task if they happened before FCP.

that sounds awesome.

patrickhulce:
> Like.... the script is render blocking so why would chrome do a real contentful paint before the evaluateScript??? That's not really supposed to happen. Can we look closer at what's going on here?

Oh the thing is it *wasn't* really render-blocking. It just had a render-blocking priority. Scripts added to the head by other scripts for example aren't render-blocking but will be requested by Chrome with the same priority as render-blocking scripts. We exclude these types of scripts from the optimistic estimate but still include them in the pessimistic one since they're usually *important* and *could* have been the script to actually render the content, but aren't *necessarily* required.

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: haha good point sounds good 👍

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: done 👍

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: done

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: yeah this was a bad name, renamed both of them to `extraBlockingCpuNodesToIncludeFilter`

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: well, on further reflection `definitely` feels a tad too strong :)

we don't not with 100% certainty, but it's the delta between the two is `definitelyNotRenderBlocking`, maybe I just return that directly

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: done

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: added more comments and deciding to restructure this a bit

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+  static getBlockingCpuData(
+      graph,
+      filterTimestamp,
+      blockingScriptFilter,
+      extraBlockingCpuNodesToIncludeFilter
+  ) {
+    /** @type {Array<CPUNode>} */
+    const cpuNodes = [];
+    graph.traverse(node => {
+      if (node.type === BaseNode.TYPES.CPU && node.startTime <= filterTimestamp) {
```
brendankenny: maybe a similar comment as below that the node is considered if *started* before the timestamp? So it's clear that it's not a typo of comparing CPUNode `startTime` against the timestamp while NetworkNode `endTime` below

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
-      if (node.endTime > fcp && !node.isMainDocument()) return false;
-      // Include EvaluateScript tasks for blocking scripts
-      if (node.type === BaseNode.TYPES.CPU) {
-        return node.isEvaluateScriptFor(blockingScriptUrls);
-      }
+    cpuNodes.sort((a, b) => a.startTime - b.startTime);
 
-      // Include non-script-initiated network requests with a render-blocking priority
-      return node.hasRenderBlockingPriority() && node.initiatorType !== 'script';
+    // A script is *possibly* render blocking if it finished loading before FCP
```
brendankenny: `s/FCP/filterTimestamp` ?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+    const possiblyRenderBlockingScriptUrls = LanternMetric.getScriptUrls(graph, node => {
+      return node.endTime <= filterTimestamp && blockingScriptFilter(node);
     });
+
+    // A script is *definitely not* render blocking if its EvaluateScript task finished after FCP.
```
brendankenny: seems like this comment doesn't belong here anymore? (the comment over the `if (!hadEvaluateScript)` seems to clarify it there, but timing doesn't seem to be checked in here anymore anyways)

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+    }
+
+    // The first layout, first paint, and first ParseHTML are almost always necessary for first paint,
+    // so we always include those CPU nodes.
+    const firstLayout = cpuNodes.find(node => node.didPerformLayout());
+    if (firstLayout) blockingCpuNodeIds.add(firstLayout.id);
+    const firstPaint = cpuNodes.find(node => node.childEvents.some(e => e.name === 'Paint'));
+    if (firstPaint) blockingCpuNodeIds.add(firstPaint.id);
+    const firstParse = cpuNodes.find(node => node.childEvents.some(e => e.name === 'ParseHTML'));
+    if (firstParse) blockingCpuNodeIds.add(firstParse.id);
```
brendankenny: are the FCP/FMP results weird (and not worth keeping) if any of these are missing?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
   }
 
   /**
+   * This function computes the graph required for the first paint of interest.
+   *
    * @param {Node} dependencyGraph
-   * @param {LH.Artifacts.TraceOfTab} traceOfTab
+   * @param {number} paintTs
+   * @param {function(NetworkNode):boolean} blockingScriptFilter
+   * @param {function(CPUNode):boolean=} extraBlockingCpuNodesToIncludeFilter
```
brendankenny: `{(function(CPUNode):boolean)=}` here too

brendankenny: also, copy the parameter descriptions from above to here as well, since this is the public-ish API called for FMP as well?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
-      }
+      if (node.type === BaseNode.TYPES.NETWORK) {
+        // Exclude all nodes that ended after FCP (except for the main document which we always consider necessary)
+        if (node.endTime > paintTs && !node.isMainDocument()) return false;
 
-      // Include non-script-initiated network requests with a render-blocking priority
-      return node.hasRenderBlockingPriority();
+        const url = node.record.url;
+        // If the URL definitely wasn't render-blocking then we filter it out.
+        if (definitelyNotRenderBlockingScriptUrls.has(url)) {
```
brendankenny: `definitelyNotRenderBlockingScriptUrls` is a little confusing inside of `getBlockingCpuData`, but really nice here!

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+  /**
+   * @param {Node} dependencyGraph
+   * @param {LH.Artifacts.TraceOfTab} traceOfTab
+   * @return {Node}
+   */
+  static getOptimisticGraph(dependencyGraph, traceOfTab) {
+    return this.getFirstPaintBasedGraph(
+      dependencyGraph,
+      traceOfTab.timestamps.firstContentfulPaint,
+      node => node.hasRenderBlockingPriority() && node.initiatorType !== 'script'
```
brendankenny: would a comment (here and in FMP) on the addition of `&& node.initiatorType !== 'script'` for the optimistic graph be good? I don't want to get redundant with other source of information, but not immediately clear while reading this why there'd be this difference between this and the pessimistic graph

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,152 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+        if (node.endTime > paintTs && !node.isMainDocument()) return false;
 
-      // Include non-script-initiated network requests with a render-blocking priority
-      return node.hasRenderBlockingPriority();
+        const url = node.record.url;
+        // If the URL definitely wasn't render-blocking then we filter it out.
+        if (definitelyNotRenderBlockingScriptUrls.has(url)) {
+          return false;
+        }
+        return node.hasRenderBlockingPriority();
```
brendankenny: this line seems to get into a double negative situation, though maybe I'm just reading wrong here.

The optimistic graph includes a filter for `node => node.hasRenderBlockingPriority() && node.initiatorType !== 'script'`, which means a node with render blocking priority but isn't `initiatorType` `script` is not considered possibly blocking.

That means it can't end up in `definitelyNotRenderBlockingScriptUrls`, which means the return value falls to this line, and since the node has render blocking priority it'll return true and be included as blocking, even though the filter was saying it shouldn't be considered blocking.

Is that not intended (and this line should maybe be `return blockingScriptFilter(node)`?) or it is intended and we should do some renaming or something to make it clearer?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: theoretically yes, but in practice I've seen some of these be shifted slightly so I wouldn't want to throw it all away

maybe another candidate for #7041 ?

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: done

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: hm, no it still is by virtue of the fact that we're checking the set of CPU nodes.

added two sentences to clarify this

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: done

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: sure sg 👍

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
patrickhulce: yes great catch!! and now we improve the metrics even more, huzzah 🎉 :D

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,163 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
-      // Include non-script-initiated network requests with a render-blocking priority
-      return node.hasRenderBlockingPriority() && node.initiatorType !== 'script';
+    // A script is *possibly* render blocking if it finished loading before filterTimestamp.
+    const possiblyRenderBlockingScriptUrls = LanternMetric.getScriptUrls(graph, node => {
+      return node.endTime <= filterTimestamp && blockingScriptFilter(node);
     });
+
+    // A script is *definitely not* render blocking if its EvaluateScript task started after filterTimestamp.
+    // We'll check `cpuNodes` to see if we can find its EvaluateScript.
+    // If we can't, that means it started after filterTimestamp, and we add it to our set.
```
brendankenny: ah, this is great

brendankenny:
@paulirish not sure if you want to take another look before we land

brendankenny:
looks like need a new golden lhr and `render-blocking-resources` expectations need adjusting

patrickhulce:
sorry y'all the fix for rendering blocking smoke is slightly more involved but I think it makes the `definitelyNotRenderBlocking` computation a little clearer too in the process.

just *one* more quick look? :D
 
the good news is we keep goin' up accuracy-wise! 😃 

this is the incremental progress on what we already had 💣 💥 🥇 

image

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,174 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+      return node.endTime <= filterTimestamp && blockingScriptFilter(node);
     });
+
+    // A script is *definitely not* render blocking if its EvaluateScript task started after filterTimestamp.
+    /** @type {Set<string>} */
+    const definitelyNotRenderBlockingScriptUrls = new Set();
+    /** @type {Set<string>} */
+    const blockingCpuNodeIds = new Set();
+    for (const url of possiblyRenderBlockingScriptUrls) {
+      // Lookup the CPU node that had the EvaluateScript for this URL.
```
brendankenny:
```suggestion
      // Lookup the CPU node that had the earliest EvaluateScript for this URL.
```

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,174 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+      const cpuNodeForUrl = scriptUrlToNodeMap.get(url);
+
+      // If we can't find it at all, we'll assume it was render-blocking.
+      if (!cpuNodeForUrl) continue;
+
+      // If we found it and it was in our `cpuNodes` set that means it finished before filterTimestamp, we're good.
```
brendankenny: `we're good` -> "so it really is blocking" or something like that

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
brendankenny: > maybe another candidate for #7041 ?

yeah, probably the best thing before we start throwing on it

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,174 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+    );
 
     return dependencyGraph.cloneWithRelationships(node => {
-      if (node.endTime > fcp && !node.isMainDocument()) return false;
-      // Include EvaluateScript tasks for blocking scripts
-      if (node.type === BaseNode.TYPES.CPU) {
-        return node.isEvaluateScriptFor(blockingScriptUrls);
-      }
+      if (node.type === BaseNode.TYPES.NETWORK) {
+        // Exclude all nodes that ended after FCP (except for the main document which we always consider necessary)
```
brendankenny: s/FCP/paintTs

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -45,7 +45,7 @@
 <link rel="stylesheet" href="./dbw_tester.css?delay=3000&async=true" disabled onload="this.disabled = false">
 
 <!-- FAIL: block rendering -->
-<script src="./dbw_tester.js"></script>
+<script src="./dbw_tester.js?delay=2000"></script>
```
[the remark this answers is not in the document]
patrickhulce: oh it's not anymore that was just for testing, removed

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,174 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+    /** @type {Array<CPUNode>} */
+    const cpuNodes = [];
+    graph.traverse(node => {
       if (node.type === BaseNode.TYPES.CPU) {
-        return node.isEvaluateScriptFor(blockingScriptUrls);
+        // A task is *possibly* render blocking if it *started* before filterTimestamp.
+        // We use startTime here because the paint event can be *inside* the task that was render blocking.
+        if (node.startTime <= filterTimestamp) cpuNodes.push(node);
+
+        // Make sure we still build our script URL to node map.
```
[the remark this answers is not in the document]
patrickhulce: haha that's fair removed :)

lighthouse-core/computed/metrics/lantern-first-contentful-paint.js
```diff
@@ -24,52 +26,174 @@ class LanternFirstContentfulPaint extends LanternMetric {
...
+    // A script is *definitely not* render blocking if its EvaluateScript task started after filterTimestamp.
+    /** @type {Set<string>} */
+    const definitelyNotRenderBlockingScriptUrls = new Set();
+    /** @type {Set<string>} */
+    const blockingCpuNodeIds = new Set();
+    for (const url of possiblyRenderBlockingScriptUrls) {
+      // Lookup the CPU node that had the EvaluateScript for this URL.
+      const cpuNodeForUrl = scriptUrlToNodeMap.get(url);
+
+      // If we can't find it at all, we'll assume it was render-blocking.
```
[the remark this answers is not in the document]
patrickhulce: yeah plenty of scripts don't actually get an evaluatescript task, it makes more sense to explain that we can't conclude it's "definitely not render blocking" just because we couldn't find its task, so i've switched to taht :)

paulirish:
We discussed shipping this in v5 (#7752), so we may wait to land it.

brendankenny:
I think we're good to go here now, right?

patrickhulce:
We are! let's do it!

PR #7517, opened 2019-03-14
core(oopif): autoattach to all descendants

patrickhulce:
**Summary**
While trying to revert #6078 discovered a few iframes that don't get autoattached for some reason. ~~It's unclear to me why `setAutoAttach` doesn't find them. At first I thought it was a grandchild iframe problem, but that didn't really explain the pattern and recursively sending `setAutoAttach` didn't solve it either.~~ It's cross-origin grandchild frames that get put into new processes distinct from all other previous frames. AFAICT, in many cases the grandchild issue wasn't a problem before because the site also had a direct child that got grouped into that same process (i.e. the verge has a doubleclick iframe so when a different ad iframe adds a doubleclick grandchild frame we were still good, or at least that's why I believe my test cases were all my passing). The reproducible case I've got now is our localhost iframing paulirish.com which iframes youtube, 3 distinct origins that always get separated.

~~Switching over to autodiscover solved it though, so I've switched to that here. I feel a tad out of my depth here maybe someone from devtools would want to comment on what my issue is. Perhaps `flatten: true` solves this and isn't just about the `sessionId` bookkeeping change?~~ Per Pavel we should avoid autodiscover and go with autoattach. This means we must start keeping track of session ids on our connection, so here comes the plumbing! 😭 

~~I'm also not 100% how to add a test for this since I don't fully understand why devtools doesn't attach to all targets and I have trouble reproducing it anyhow, but suggestions welcome :)~~

**Related Issues/PRs**
#6337  #6078

### DIFF
--- a/lighthouse-cli/test/smokehouse/oopif-expectations.js
+++ b/lighthouse-cli/test/smokehouse/oopif-expectations.js
@@ -16,10 +16,11 @@ module.exports = [
       'network-requests': {
         details: {
           items: {
-            // The page itself only makes a few requests.
-            // We want to make sure we are finding the iframe's requests (airhorner) while being flexible enough
-            // to allow changes to the live site.
-            length: '>10',
+            // We want to make sure we are finding the iframe's requests (paulirish.com) *AND*
+            // the iframe's iframe's iframe's requests (youtube.com/doubleclick/etc).
+            // - paulirish.com ~50-60 requests
+            // - paulirish.com + all descendant iframes ~80-90 requests
+            length: '>70',
           },
         },
       },
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -69,17 +69,17 @@ class Driver {
      */
     this._monitoredUrl = null;
 
-    let targetProxyMessageId = 0;
     this.on('Target.attachedToTarget', event => {
-      targetProxyMessageId++;
       // We're only interested in network requests from iframes for now as those are "part of the page".
       if (event.targetInfo.type !== 'iframe') return;
 
       // We want to receive information about network requests from iframes, so enable the Network domain.
       // Network events from subtargets will be stringified and sent back on `Target.receivedMessageFromTarget`.
-      this.sendCommand('Target.sendMessageToTarget', {
-        message: JSON.stringify({id: targetProxyMessageId, method: 'Network.enable'}),
-        sessionId: event.sessionId,
+      this.sendCommand({method: 'Network.enable', sessionId: event.sessionId});
+      this.sendCommand({method: 'Target.setAutoAttach', sessionId: event.sessionId}, {
+        autoAttach: true,
+        waitForDebuggerOnStart: false,
+        flatten: true,
       });
     });
 
@@ -279,11 +279,15 @@ class Driver {
    * Call protocol methods, with a timeout.
    * To configure the timeout for the next call, use 'setNextProtocolTimeout'.
    * @template {keyof LH.CrdpCommands} C
-   * @param {C} method
+   * @param {C|{method: C, sessionId: string}} methodObj
    * @param {LH.CrdpCommands[C]['paramsType']} params
    * @return {Promise<LH.CrdpCommands[C]['returnType']>}
    */
-  sendCommand(method, ...params) {
+  sendCommand(methodObj, ...params) {
+    /** @type {C} */
+    const method = typeof methodObj === 'string' ? methodObj : methodObj.method;
+    const sessionId = typeof methodObj === 'string' ? undefined : methodObj.sessionId;
+
     const timeout = this._nextProtocolTimeout;
     this._nextProtocolTimeout = DEFAULT_PROTOCOL_TIMEOUT;
     return new Promise(async (resolve, reject) => {
@@ -295,7 +299,7 @@ class Driver {
         reject(err);
       }), timeout);
       try {
-        const result = await this._innerSendCommand(method, ...params);
+        const result = await this._innerSendCommand(method, sessionId, ...params);
         resolve(result);
       } catch (err) {
         reject(err);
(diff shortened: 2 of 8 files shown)


paulirish:
@pavelfeldman @aslushnikov can you take a look and lend some insight?

paulirish:
Pavel says setDiscoverTargets will report all targets in the browser.. (so.. iframes in other tabs would make it through). the targetInfo does report a parent.. so it's possible to use this and filter to just your tree, but funky.

autoAttach is preferred and _should_ work for us. And Pavel's interested in a test case if recursively calling setAutoattach isn't reliably finding iframes.

patrickhulce:
> And Pavel's interested in a test case if recursively calling setAutoattach isn't reliably finding iframes.

Finally found the narrow case where it's happening and I see where I went wrong 👍 


Take 3 is up, if we're going with setAutoAttach there's not really a way to avoid threading `sessionId` down to the connection level which is what I was hoping to avoid. If we wanted to start reporting opportunities on iframe resources though this was somewhat inevitable. 

I did it in the least impactful way I could think of right now. WDYT?

paulirish:
DZL, do a barrel roll!

lighthouse-cli/test/smokehouse/oopif-expectations.js
```diff
@@ -16,10 +16,11 @@ module.exports = [
...
           items: {
-            // The page itself only makes a few requests.
-            // We want to make sure we are finding the iframe's requests (airhorner) while being flexible enough
-            // to allow changes to the live site.
-            length: '>10',
+            // We want to make sure we are finding the iframe's requests (paulirish.com) *AND*
+            // the iframe's iframe's iframe's requests (youtube.com/doubleclick/etc).
+            // - paulirish.com ~50-60 requests
+            // - paulirish.com + all descendant iframes ~80-90 requests
+            length: '>70',
```
paulirish: I have a feeling this is going to have a big effect on TTI as network idle is likely going to be pushed out in a bunch of cases.

lighthouse-core/gather/driver.js
```diff
@@ -279,11 +279,15 @@ class Driver {
    * Call protocol methods, with a timeout.
    * To configure the timeout for the next call, use 'setNextProtocolTimeout'.
    * @template {keyof LH.CrdpCommands} C
-   * @param {C} method
+   * @param {C|{method: C, sessionId: string}} methodObj
```
paulirish: Yah I dig this solution.

A followup PR can move `this._nextProtocolTimeout` into this methodObj. cc @Hoten

lighthouse-core/gather/driver.js
```diff
@@ -69,17 +69,17 @@ class Driver {
...
     this._monitoredUrl = null;
 
-    let targetProxyMessageId = 0;
     this.on('Target.attachedToTarget', event => {
-      targetProxyMessageId++;
       // We're only interested in network requests from iframes for now as those are "part of the page".
       if (event.targetInfo.type !== 'iframe') return;
 
       // We want to receive information about network requests from iframes, so enable the Network domain.
       // Network events from subtargets will be stringified and sent back on `Target.receivedMessageFromTarget`.
```
paulirish: We still get back these messages over the target domain?

In https://github.com/GoogleChrome/puppeteer/pull/3827/files#diff-0c1fe9dd78cc7f29823c09b251c8aae0R109 I saw the sessionId being received and handled in the `handleRawMessage` method (basically).

lighthouse-core/gather/driver.js
brendankenny: I feel the same way about this as I did for #6347 :) We're trying to generalize a simple thing by making it more complicated, just for the two calls that need it.

Since they're both inside `driver`, would it be possible for now to make those two calls be calls of `_innerSendCommand` instead, leave `sendCommand` the same, and we setup a time to figure out what `driver` should look like in a multi target world (which we're going to need for SW work anyways)? I think long term that can also help with complexity when we start doing more opportunities with iframe resources, as ideally we can minimize the manual sessionId management that has to be done in gatherers.

lighthouse-core/gather/driver.js
```diff
@@ -310,18 +314,19 @@ class Driver {
...
    * @param {C} method
+   * @param {string|undefined} sessionId
    * @param {LH.CrdpCommands[C]['paramsType']} params
    * @return {Promise<LH.CrdpCommands[C]['returnType']>}
    */
-  _innerSendCommand(method, ...params) {
+  _innerSendCommand(method, sessionId, ...params) {
     const domainCommand = /^(\w+)\.(enable|disable)$/.exec(method);
-    if (domainCommand) {
+    if (domainCommand && !sessionId) {
```
brendankenny: if we're going to expand to other targets, we should probably track enabled domains there as well...

lighthouse-core/test/lib/network-recorder-test.js
```diff
@@ -249,9 +249,7 @@ describe('network recorder', function() {
       ];
 
       const periods = NetworkRecorder.findNetworkQuietPeriods(records, 0);
-      assert.deepStrictEqual(periods, [
-        {start: 1200, end: Infinity},
-      ]);
+      assert.deepStrictEqual(periods, []);
```
brendankenny: what changed with the quiet period?

lighthouse-core/gather/driver.js
patrickhulce: > Since they're both inside driver, would it be possible for now to make those two calls be calls of _innerSendCommand instead, leave sendCommand the same, and we set up a time to figure out what driver should look like in a multi target world (which is also going to help us do better SW work)?

sure

lighthouse-core/gather/driver.js
patrickhulce: whoops no, I need to update the comment 👍

lighthouse-core/gather/driver.js
patrickhulce: I'll use your point back atcha for now ;) 

It's only the two messages that are never disabled, so we can cross that bridge when we get to it IMO

lighthouse-core/test/lib/network-recorder-test.js
patrickhulce: this was the intentional removal of our hack for frame roots, now if an iframe never fires `loadingFinished` it shouldn't be seen as finished

lighthouse-core/gather/driver.js
brendankenny: > It's only the two messages that are never disabled, so we can cross that bridge when we get to it IMO

ha, I am 👌👌 with crossing that bridge later, but maybe leave a comment about not bothering to check other targets?

lighthouse-core/test/lib/network-recorder-test.js
brendankenny: ah, got it, I wasn't thinking about the `network-recorder.js` changes

patrickhulce:
Soooooooo, I totally forgot the extension existed again (as I tend to do) and the story is a lot more complicated there. I'll put up what I have in a separate PR so it's easier to see, but basically my conclusion is that we cannot use `setAutoAttach` there.

patrickhulce:
This has changed enough from the beginning that I'm just opening a new PR

PR #5322, opened 2018-05-23
core(seo): support japanese in link-text audit

rdlabo:
**Summary**
lint link-text is great!! I think that it will more wonderful if it supports multiple languages.

I added Japanese easily. But I think this should be separate file, and that various people should be involved in editing.
Please let the opinion. 

thanks.

PR #7547, opened 2019-03-17
core(seo): support spanish in link-text audit

MarkelFe:
I have added a couple of translations for the [Links Do Not Have Descriptive Text](https://developers.google.com/web/tools/lighthouse/audits/descriptive-link-text) audit in Spanish.

Note that in Spanish there is this symbol: `´`. People forget to write it so we may take into account these errors, let me know your opinion.

**Related Issues/PRs**
#5322 but in Spanish

### DIFF
--- a/lighthouse-core/audits/seo/link-text.js
+++ b/lighthouse-core/audits/seo/link-text.js
@@ -43,7 +43,7 @@ const BLOCKLIST = new Set([
   'este',
   'enlace',
   'este enlace',
-  'empezar', 
+  'empezar',
 ]);
 const i18n = require('../../lib/i18n/i18n.js');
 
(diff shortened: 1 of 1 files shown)


lighthouse-core/audits/seo/link-text.js
```diff
@@ -25,6 +25,25 @@ const BLOCKLIST = new Set([
   '続きを読む',
   '続く',
   '全文表示',
+  // Spanish
+  'click aquí',
```
patrickhulce: There's a fair amount of duplication with the permutations, I wonder if we could do some replacements to keep the signal in here high?


could be as simple as an easy map for now

```js
const CHARACTER_REPLACEMENTS = new Map([
  [/í/g, 'i'],
  [/á/g, 'a'],
  [/ó/g, 'o'],
])

function normalizeText(text) {
  let normalized = text.trim().toLowerCase();
  for (const [regex, replacement] of CHARACTER_REPLACEMENTS.entries()) {
     normalized = normalized.replace(regex, replacement)
  }

  return normalized;
}
...

return BLOCKLIST.has(normalizeText(link.text));
```

DISCLAIMER: untested

PR #7562, opened 2019-03-18
core(tsc): update to tsc 3.3.333

Hoten:
no changes needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "prettier": "^1.14.3",
     "pretty-json-stringify": "^0.0.2",
     "puppeteer": "^1.10.0",
-    "typescript": "3.2.2",
+    "typescript": "3.3.3333",
     "uglify-es": "3.0.15",
     "url-search-params": "0.6.1",
     "whatwg-fetch": "2.0.1",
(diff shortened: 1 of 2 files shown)


Hoten:
I didn't exhaustively look to see how the new features could be used. I really wanted 3.4 but that's still in beta.

patrickhulce:
> I didn't exhaustively look to see how the new features could be used. I really wanted 3.4 but that's still in beta.

Oh, well I mean what prompted the update then I guess is what I'm asking. Why do we care about 3.3.3? :)

Hoten:
> Why do we care about 3.3.3? :)

I don't, I just had 5 minutes to kill on the bus. Do you suggest we not update?

patrickhulce:
Oh haha 😆 then yeah I'd vote for not updating if there's no reason too. Just seems like noise otherwise.

Hoten:
I was thinking there might be necessary changes, which would need to be made for the 3.4 update (which I do care about). But there were none :yay:

fine with me

brendankenny:
> Why do we care about 3.3.3?

I haven't seen anything we'd immediately want. [3.4 will have some nice things](https://github.com/Microsoft/TypeScript/wiki/Roadmap#34-march-2019) around better generic inference and better readonly support, but I don't believe it should affect anything we currently use either, only open up options for future use.

([`as const`](https://github.com/Microsoft/TypeScript/pull/29510) would be super helpful for us, but it doesn't work in JS (yet))

PR #7566, opened 2019-03-18
OOPIF full support in extension

patrickhulce:
broken out from #7517 for easier diffing

**tl;dr** I don't see an immediate way to make extensions work with `setAutoAttach`

**Summary**
The problem: [extension debugger API](https://developers.chrome.com/extensions/debugger#method-sendCommand) does not have a parameter for `sessionId`. It's method of talking to targets uses `targetId` and a different `sendCommand` just to them. This PR implements that approach, *however*, we run into a problem.

1. You cannot `chrome.debugger.sendCommand` to a target you have not yet attached to.
2. You cannot `chrome.debugger.attach` to a target that has already been attached to by another debugger (which `Target.setAutoAttach` seems to do).

The options as I see it:

1. Get `chrome.debugger.sendCommand` to support `sessionId` as a `Debuggee` target (maybe there's hidden support I haven't found?? I tried sending it as part of params too with no luck)
2. Switch back to `discover` all the targets and filter the frame tree instead of `setAutoAttach`
3. Something I'm missing?

**Related Issues/PRs**
#7517

### DIFF
--- a/clients/test/extension/extension-test.js
+++ b/clients/test/extension/extension-test.js
@@ -70,7 +70,7 @@ describe('Lighthouse chrome extension', function() {
     });
 
     const page = await browser.newPage();
-    await page.goto('https://www.paulirish.com', {waitUntil: 'networkidle2'});
+    await page.goto('https://www.paulirish.com/page/2', {waitUntil: 'networkidle2'});
     const targets = await browser.targets();
     const extensionTarget = targets.find(({_targetInfo}) => {
       return _targetInfo.title === 'Lighthouse' && _targetInfo.type === 'background_page';
@@ -179,5 +179,9 @@ describe('Lighthouse chrome extension', function() {
 
   it('should specify the channel as extension', async () => {
     assert.equal(lhr.configSettings.channel, 'extension');
+  });
+
+  it('should find OOPIF network requests', async () => {
+    expect(lhr.audits['network-requests'].details.items.length).toBeGreaterThan(60);
   });
 });
--- a/lighthouse-core/gather/connections/connection.js
+++ b/lighthouse-core/gather/connections/connection.js
@@ -52,14 +52,14 @@ class Connection {
    * Call protocol methods
    * @template {keyof LH.CrdpCommands} C
    * @param {C} method
-   * @param {string|undefined} sessionId
+   * @param {{sessionId: string, targetId: string}|undefined} target
    * @param {LH.CrdpCommands[C]['paramsType']} paramArgs,
    * @return {Promise<LH.CrdpCommands[C]['returnType']>}
    */
-  sendCommand(method, sessionId, ...paramArgs) {
+  sendCommand(method, target, ...paramArgs) {
     // Reify params since we need it as a property so can't just spread again.
     const params = paramArgs.length ? paramArgs[0] : undefined;
-
+    const sessionId = target && target.sessionId;
     log.formatProtocol('method => browser', {method, params, sessionId}, 'verbose');
     const id = ++this._lastCommandId;
     const message = JSON.stringify({id, method, params, sessionId});
--- a/lighthouse-core/gather/connections/extension.js
+++ b/lighthouse-core/gather/connections/extension.js
@@ -16,6 +16,8 @@ class ExtensionConnection extends Connection {
   constructor() {
     super();
     this._tabId = null;
+    /** @type {Set<string>} */
+    this._attachedTargetIds = new Set();
 
     this._onEvent = this._onEvent.bind(this);
     this._onUnexpectedDetach = this._onUnexpectedDetach.bind(this);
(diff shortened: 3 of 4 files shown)


paulirish:
talked with pavel and dmitry and they think setAutoAttach can work. (without it, we have a race)

### Introducing @pavelfeldman's Foolproof Guide to Making `setAutoAttach` Work with the Extension API

1. Target.setAutoAttach({setAutoAttach:true, waitForDebuggerOnStart: true})
1. targetCreated event arrives. 
1. attachedToTarget event arrives.
   - attachedToTarget has sessionID, targetinfo.targetID
1. newdebuggee = chrome.debugger.attachtoTarget(targetID)
1. newdebugeee: set up all agents, overrides
1. main connection: Target.detachFromTarget(sessionID)


@patrickhulce since you've been deep in this, does it sound reasonable to you?

brendankenny:
is it possible there will be a significant impact from `waitForDebuggerOnStart: true`, or are we connecting quickly enough that it's ok?

paulirish:
>is it possible there will be a significant impact from waitForDebuggerOnStart: true, or are we connecting quickly enough that it's ok? 

without using it we can very likely miss out on network activity that happened while we weren't attached. (probably for a datauri iframe).
with pausing yes we wait for our JS thread to respond and attach, but I'd expect this to happen in <5ms. So it does add some latency, but probably not a lot. Our thread _shouldnt_ be busy at this time.

patrickhulce:
@paulirish does Pavel think the `waitForDebuggerOnStart` will fix the fact that the chrome extension throws when you try to attach to a target that was discovered via `setAutoAttach`? 

That was the core problem I was trying to communicate that we're in a catch 22. The extension API wants you to connect to a target before sending a command yet when you try to attach it says the target is already attached (I'm assuming because it was attached via `setAutoAttach` but possibly it's another reason).

patrickhulce:
Aside from `waitForDebuggerOnStart` I've implemented exactly as he has described, I'll give it a shot.

patrickhulce:
The plot thickens, you can't use `Target.detachFromTarget` at all from extension 🤔 

> Protocol error (Target.detachFromTarget): Not allowed

This is the core of the problem AFAICT: extension API just can't interact with sessionId-based attached targets.

- It can't sendCommand based on a sessionId (this is based on the docs and everywhere I've tried to stick `sessionId` to no effect or error)
- It throws when attaching to a targetId that has been auto-attached (via `chrome.debugger.attach`)
- It throws when attaching to a targetId via `Target.attachToTarget` (pretty sure I tried this earlier and it threw you have to go through `chrome.debugger.attach`)
- It throws when detaching from a sessionId that has been auto-attached (via `Target.detachFromTarget`, see above)
- It can't detach via `targetId` (via `chrome.debugger.detach`) because it doesn't think it's attached in the first place.

paulirish:
+ @dgozman @pavelfeldman

pavelfeldman:
It looks like our `Target` domain handler for `chrome.debugger` endpoints is [constrained](https://cs.chromium.org/chromium/src/content/browser/devtools/render_frame_devtools_agent_host.cc?sq=package:chromium&g=0&l=300) due to security considerations. It will auto-attach, but it won't let you attach or detach :/

Which means that magic that I was suggesting does not work, sorry.

patrickhulce:
So it seems like our options are....

- Use `setAutoAttach` and don't support OOPIFs in the extension
- Use `discoverTargets` and live with potential races
- Implement both methods and choose which one we use based on some environment flag
- Other??

pavelfeldman:
> Use discoverTargets and live with potential races

Can't do even that. ([link](https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/target_handler.cc?sq=package:chromium&g=0&l=554))

pavelfeldman:
Remind me why we don't want to use autoAttach / sendMessageToTarget?

patrickhulce:
> Use discoverTargets and live with potential races
> Can't do even that. (link)

@pavelfeldman yes, it would never be both. I know y'all have been recommending against `discoverTargets` from the beginning but it is technically *an* option :)

> Remind me why we don't want to use autoAttach / sendMessageToTarget?

Extension API throws when you try to use `sendMessageToTarget` and tells you to use `chrome.debugger.sendCommand` which has the aforementioned attachment problems

pavelfeldman:
>  it is technically an option

not in chrome.debugger land, right?

pavelfeldman:
> Extension API throws when you try to use `sendMessageToTarget`

I'll look into this one.

patrickhulce:
> not in chrome.debugger land, right?

I haven't implemented it but it seems like it should work from there, no or am I missing something? I can still see and attach to targets when they aren't auto attached so I should be able to attach to *them*, right?

patrickhulce:
> I'll look into this one.

To clarify, our current method when it's not `flat` works for `sendMessageToTarget` it only started failing when I started using the new method

pavelfeldman:
> haven't implemented it but it seems like it should work from there

I sent you a chrome link above that suggests there is no way it does

patrickhulce:
> Error: Protocol error (Target.sendMessageToTarget): When using flat protocol, messages are routed to the target via the sessionId attribute.

That's what I get once I enable `flatten`

pavelfeldman:
> To clarify, our current method when it's not flat works for `sendMessageToTarget`

`sendMessageToTarget` is not used in flat mode at all. So it is fine for it to throw when misused with flat. But if non-flat works, why touching it?

patrickhulce:
> I sent you a chrome link above that suggests there is no way it does

Oh I misunderstood I thought only the subtargets were kAutoAttachOnly but now I see.

patrickhulce:
> But if non-flat works, why touching it?

Because we want to do this for all descendants not just children. I guess you're saying we can continue to use the old way if we just properly nest the `Target.sendMessageToTarget` commands enough?

pavelfeldman:
> I guess you're saying we can continue to use the old way if we just properly nest the `Target.sendMessageToTarget` commands enough

Exactly this

patrickhulce:
Alrighty we've found our solution even if it means we can't use any of the flattening niceness

Thanks @pavelfeldman !

PR #7574, opened 2019-03-19
core(navigation): Add an option to ignore https errors during navigation

Janpot:
**Summary**

Adds an option, similar like puppeteer, to ignore https errors during page navigation.
I'm still planning to add tests, but opening a PR for discussion.

IMO, If the intent is to encourage people to sort out their security, then I think this should be turned into an audit, rather than a hard error.

I'd like to be able to run lighthouse on pages that didn't sort out their security yet. And be able to extend lighthouse to include an audit that reflects the chrome security state.

**Related Issues/PRs**


lighthouse-core/gather/driver.js
```diff
@@ -1454,6 +1458,14 @@ class Driver {
 
     await this.sendCommand('Page.enable');
   }
+
+  /**
+   * Set chrome security settings to ignore certificate errors.
+   * @return {Promise<void>}
+   */
+  async setIgnoreHttpsErrors() {
+    await this.sendCommand('Security.setIgnoreCertificateErrors', {ignore: true});
```
patrickhulce: I believe puppeteer uses 

```js
await client.send('Security.setOverrideCertificateErrors', {override: true});
client.on('Security.certificateError', event => this._onCertificateError(event));

_onCertificateError(event) {
    this._client.send('Security.handleCertificateError', {
      eventId: event.eventId,
      action: 'continue'
    }).catch(debugError);
}
```

but in classic Google fashion those two methods are `deprecated` and `setIgnoreCertificateErrors ` is `experimental`, soooo we're stuck between a 💎and a hard place

lighthouse-core/gather/driver.js
```diff
@@ -885,7 +886,9 @@ class Driver {
     const monitorForInsecureState = this._monitorForInsecureState();
     const securityCheckPromise = monitorForInsecureState.promise.then(securityMessages => {
       return function() {
-        throw new LHError(LHError.errors.INSECURE_DOCUMENT_REQUEST, {securityMessages});
+        if (!ignoreHttpsErrors) {
```
patrickhulce: I'm not sure this is entirely what we want since we'll have already fallen into the wrong `Promise.race` branch below. I think we want a `_waitForever` version of `_waitForNothing`


```js
const securityCheckPromise = ignoreHttpsErrors ? this._waitForever() : <promise as it is now>

File: lighthouse-core/gather/driver.js (hunk 1 above)
Comment (Janpot) [reply lighthouse-core/gather/driver.js]: Seems like lighthouse already uses some experimental methods already, like e.g. [Emulation.setCPUThrottlingRate](
https://github.com/GoogleChrome/lighthouse/blob/d76b2f5e02df7b04d861c25f723bc69514d2e863/lighthouse-core/lib/emulation.js#L136).
That being said, it looks like puppeteer is actually [waiting for it to stabilize](https://github.com/GoogleChrome/puppeteer/issues/1634). I'm fine with either, but I'd do `setIgnoreCertificateErrors`.

File: lighthouse-core/gather/driver.js (hunk 2 above)
Comment (Janpot) [reply lighthouse-core/gather/driver.js]: makes sense, I'll refactor

**Comment by Janpot** [comment]
@patrickhulce could it be that there is a broken/flaky test (`(http://localhost:10200/tricky-main-thread.html?setTimeout)`)? Can't seem to reproduce locally, and I'm not sure how it's related.

**Comment by patrickhulce** [comment]
> could it be that there is a broken/flaky test ((http://localhost:10200/tricky-main-thread.html?setTimeout))?

Yeah it's only failing on appveyor we've yet to look into it. Debugging appveyor-only failures isn't the easiest/most fun thing to do :)

File: lighthouse-core/gather/driver.js (hunk 1 above)
Comment (patrickhulce) [reply lighthouse-core/gather/driver.js]: Oh yeah we use plenty of experimental that's not an issue on its own, it's just rarer that there is an alternative in the "stable" protocol set we can use

Comment (Janpot) [reply lighthouse-core/gather/driver.js]: I changed to the puppeteer method, for some strange reason, this conflicts with the `redirectPass` (and only with that pass).
Screenshot 2019-03-19 at 16 59 04
Seems like internally chrome executes the `Security.handleCertificateError` just fine though.

File: lighthouse-core/gather/driver.js [hunk 3]
```diff
@@ -550,6 +550,17 @@ class Driver {
...
   }
 
+  /**
+   * Returns a promise that never resolves.
+   * Used for placeholder conditions that we don't want race with, but still want to satisfy the same
+   * interface.
+   * @return {{promise: Promise<void>, cancel: function(): void}}
+   */
+  _waitForever() {
+    return {promise: new Promise(() => null), cancel() {}};
```
patrickhulce: we probably still want to be able to cancel it though to avoid hanging async junk, just meant that the promise will "wait forever" until cancelled :)

sorry for the confusion!

Janpot: no problem, updated the PR

Janpot: Not sure yet why, but for some reason it seems like this also fixed some `NO_FCP` errors I was seeing in some pass (that I was still looking into)

Janpot:
@brendankenny @paulirish Any input on how to proceed here? I don't need this merged exactly tomorrow but it would be helpful for me to at least know what's the line of thinking on this of the team.

IMO lighthouse should treat security errors in a similar way as the `http-status-code` audit. Basically, accept insecure urls and use the `Security.securityStateChanged` to craft an audit that fails on those. I don't mind working on a PR for this.

Hoten:
I think this is great. Asking people to muck with certificates has proven to be too big an ask. I'll go ahead and update this branch w/ master.

lbineau:
+1 for merging this feature, I cannot audit local website with invalidate HTTPS certificate.

zzo:
+1 please merge :)

vitalyzhakov:
I have interesting to merge this pull, because I want to make our website better.
Can I help something to do this big thing?

paulirish:
closing in favor of #8865

Issue #7534, opened 2019-03-15
Running with headless Chrome crashes when provided with non-existent url.


jburger424:
#### Provide the steps to reproduce
1. `lighthouse https://not-a-url.com --chrome-flags=--headless`

#### What is the current behavior?
```
events.js:167
      throw er; // Unhandled 'error' event
      ^

Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
Emitted 'error' event at:
    at Socket.socketErrorListener (_http_client.js:391:9)
    at Socket.emit (events.js:182:13)
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

No report is saved.

#### What is the expected behavior?
`Runner:warn ... encountered an error: DNS_FAILURE`

Report should save with following error:
```
There were issues affecting this run of Lighthouse:
  DNS servers could not resolve the provided domain.
```

#### Environment Information
* Affected Channels: CLI
* Lighthouse version: 4.2.0
* Node.js version: 6.9.0
* Operating System: Debian GNU/Linux

**Related issues**

patrickhulce:
FWIW this seems linux specific as I cannot reproduce on Mac. Resident linux folks can take a look? :)

If Chrome is crashing, then it's probably worth filing a crbug instead of just LH bug.

brendankenny:
I'm able to get the same thing on my Mac.

We should be catching this, so I assume (based on that and the error) it's happening in an event and so outside our normal execution flow.

jburger424:
It looks like this is a new issue in 4.2.0, unable to reproduce on 4.1.0.

brendankenny:
hmm, at first glance it appears to be from running this line with an already killed Chrome instance, which is supposed to not do this :)

https://github.com/GoogleChrome/lighthouse/blob/9efe177e47fd0b8a4762c5b770b0fd4ef7f1f63f/lighthouse-core/gather/driver.js#L821-L822

brendankenny:
sorry, I take that back. That line happily runs and returns, it just happens to occur while other things are going wrong.

[...]

PR #7576, opened 2019-03-19
core(logging): Don't log when closing already closed page

Janpot:
**Summary**
Screenshot 2019-03-19 at 15 29 54

Happens when running lighthouse on insecure pages:
```
lighthouse https://wrong.host.badssl.com/
```

**Related Issues/PRs**

I also sometimes get this error (like one in 10 times or so).

Screenshot 2019-03-19 at 15 56 40


patrickhulce:
> I also sometimes get this error (like one in 10 times or so).

I believe that might be https://github.com/GoogleChrome/lighthouse/issues/7534

PR #7578, opened 2019-03-19
cli(logging): Use lighthouse logger to display runtime errors

Janpot:
**Summary**

Turns these logs:
Screenshot 2019-03-19 at 16 09 38
into these logs:
Screenshot 2019-03-19 at 16 08 38

Just to clean up the output and make it more uniform.

**Related Issues/PRs**

### DIFF
--- a/lighthouse-cli/run.js
+++ b/lighthouse-cli/run.js
@@ -64,25 +64,25 @@ function getDebuggableChrome(flags) {
 
 /** @return {never} */
 function showConnectionError() {
-  log.error('CLI runtime', 'Unable to connect to Chrome');
+  log.error('CLI', 'Unable to connect to Chrome');
   return process.exit(_RUNTIME_ERROR_CODE);
 }
 
 /** @return {never} */
 function showProtocolTimeoutError() {
-  log.error('CLI runtime', 'Debugger protocol timed out while connecting to Chrome.');
+  log.error('CLI', 'Debugger protocol timed out while connecting to Chrome.');
   return process.exit(_PROTOCOL_TIMEOUT_EXIT_CODE);
 }
 
 /** @param {LH.LighthouseError} err @return {never} */
 function showPageHungError(err) {
-  log.error('CLI runtime', 'Page hung:', err.friendlyMessage);
+  log.error('CLI', 'Page hung:', err.friendlyMessage);
   return process.exit(_PAGE_HUNG_EXIT_CODE);
 }
 
 /** @param {LH.LighthouseError} err @return {never} */
 function showInsecureDocumentRequestError(err) {
-  log.error('CLI runtime', 'Insecure document request:', err.friendlyMessage);
+  log.error('CLI', 'Insecure document request:', err.friendlyMessage);
   return process.exit(_INSECURE_DOCUMENT_REQUEST_EXIT_CODE);
 }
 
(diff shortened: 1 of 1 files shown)


lighthouse-cli/run.js
```diff
@@ -64,25 +64,25 @@ function getDebuggableChrome(flags) {
 
 /** @return {never} */
 function showConnectionError() {
-  console.error('Unable to connect to Chrome');
+  log.error('CLI runtime', 'Unable to connect to Chrome');
```
patrickhulce: I think we can just have these be `'CLI'`, that matches our two other log statements from the CLI 👍

Janpot: Done

patrickhulce:
> do we want --quiet to prevent these errors from being printed (to stderr)? We've made --quiet purposefully silence other log.error instances, but this is also only for the CLI

IMO we should move quiet from the CLI to be the equivalent of `logLevel: 'error'`, auditing our own usage of `.error` most of them feel like they're really just warnings and the class of error should be reserved for real errors.

> it's much more likely that there could be an error thrown from log.error itself than console.error

sure but has this *ever* happened? if we legitimately thought this was a concern we probably shouldn't be using a non `console.*` logger at all :)

brendankenny:
> IMO we should move quiet from the CLI to be the equivalent of logLevel: 'error', auditing our own usage of .error most of them feel like they're really just warnings and the class of error should be reserved for real errors.

I mean, we can, but it would have to be done before this PR. I think the current `console.error` is how we get the `NO_FCP` logging out of smokehouse (which uses `--quiet`), for instance.

Not sure how I feel about getting rid of `log.error` elsewhere, though. Can be helpful to have proximate logging, but it's possible the existing messages could be moved to `warning` with no issue.

Or we could make yet another logging level. `log.thisIsTheEndError()` :)

Hoten:
> Or we could make yet another logging level. log.thisIsTheEndError() :)

[yes please](https://www.youtube.com/watch?v=JSUIQgEVDM4)

patrickhulce:
> it would have to be done before this PR. I think the current console.error is how we get the NO_FCP logging out of smokehouse (which uses --quiet), for instance.

Yikes how is it passing our smoketests then that's not good 🤔 

> Or we could make yet another logging level. log.thisIsTheEndError()

I like that actually `log.fatal`?

Janpot:
> I think the current console.error is how we get the NO_FCP logging out of smokehouse (which uses --quiet), for instance.

Been looking through the code, but can't find any reference to this in smokehouse. It seems to use exit codes to detect certain erroneous output. I'm totally new to this code so I might be completely off here.

Hoten:
Your changes are ok as-is afaict. The travis CI tests still pass (ignore appveyor, it is flaky).

I'm not seeing any assertions made of NO_FCP errors in our smoke tests. The couple smoke tests we have for fatal error conditions are in `error-expectations.js`, and those are fine b/c we signify with an error code after catching the error in the run.

brendankenny:
>> it would have to be done before this PR. I think the current console.error is how we get the NO_FCP logging out of smokehouse (which uses --quiet), for instance.

> Yikes how is it passing our smoketests then that's not good 🤔

> I'm not seeing any assertions made of NO_FCP errors in our smoke tests.

I meant how we see it printed in case there is a runtime error during a smoke run. We inherit `stderr` so we see the error printed to console when smokehouse runs lighthouse in a child process (it doesn't affect the smoke test results because those are read from disk).

brendankenny:
short of us implementing a `log.fatal()` or moving errors to warnings, I don't think we should move on this so `--quiet` doesn't silence our error logging. Sorry for leaving this hanging for so long!

PR #7603, opened 2019-03-20
core(tap-targets): exclude sticky elements from tap targets audit

mattzeunert:
**Summary**

Whether sticky elements overlap depends on the current scroll position, so we should exclude them like we do with position absolute and fixed.

**Related Issues/PRs**

Pointed out here: https://github.com/GoogleChrome/lighthouse/issues/7365#issuecomment-474095920

### DIFF
--- a/lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
+++ b/lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
@@ -119,7 +119,7 @@
       This is <a>a link in a text block.</a>
     </p>
 
-    <a style="background: #afa; top: 0; left: 0; display: block; width: 100vw; height: 800px;">
+    <a style="background: #afa; display: block; width: 100%; height: 800px;">
       This link is intentionally placed at the bottom of the page, so that when we've scrolled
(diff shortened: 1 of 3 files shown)


lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
```diff
@@ -108,13 +111,19 @@ <h1>SEO Tap targets</h1>
...
-    <p style="width: 100px">
+    <p style="width: 100px;margin-bottom: 30px;">
       This is <a>a link in a text block.</a>
       This is <a>a link in a text block.</a>
       This is <a>a link in a text block.</a>
       This is <a>a link in a text block.</a>
       This is <a>a link in a text block.</a>
     </p>
+
+    <a style="background: #afa; top: 0; left: 0; display: block; width: 100vw; height: 800px;">
```
patrickhulce: what are `top` and `left` doing here? isn't this element `position: static` so they will have no effect?

mattzeunert: Good catch, they're not relevant.

PR #7613, opened 2019-03-21
clients(lr): allow for custom config

Hoten:
**Summary**

Main motivator: Smokerider needs to pass a custom config to Lightrider, otherwise the results do not correspond to the expectations.

But also, hey, a way to configure plugins in Lightrider! Just provide a config that extends the LR preset + adds your plugin coolness.

I can open a WIP internal CL for the LR branch that uses these changes, if necessary to review.

**Related Issues/PRs**

Just the proto changes: #7614
[Internal Googler bug](b/128444698)


clients/lightrider-entry.js
```diff
@@ -24,11 +24,11 @@ const LR_PRESETS = {
  * @param {Connection} connection
  * @param {string} url
  * @param {LH.Flags} flags Lighthouse flags, including `output`
- * @param {{lrDevice?: 'desktop'|'mobile', categoryIDs?: Array<string>, logAssets: boolean, keepRawValues: boolean}} lrOpts Options coming from Lightrider
+ * @param {{lrDevice?: 'desktop'|'mobile', categoryIDs?: Array<string>, logAssets: boolean, keepRawValues: boolean, configOverride?: LH.Config.Json}} lrOpts Options coming from Lightrider
```
patrickhulce: Should we note somewhere that lrDevice and `categoryIDs` get ignored with `configOverride`?

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -25,17 +25,6 @@ function processForProto(result, opts = {}) {
   /** @type {LH.Result} */
   const reportJson = JSON.parse(result);
 
-  // Clean up the configSettings
-  // Note: This is not strictly required for conversion if protobuf parsing is set to
-  // 'ignore unknown fields' in the language of conversion.
-  if (reportJson.configSettings) {
```
patrickhulce: why is this being removed?

Hoten: It was pruning the JSON sent back to LR, which would then place it in a Config proto that only had `{emulatedFormFactor, locale, onlyCategories}`. Since the rest weren't saved it made sense to cut the extras here. With the proto changes, most of the settings (i left out things like `output`) can be saved, so I'm undoing the pruning. The fields that aren't in the proto are safely ignored when serialized.

clients/lightrider-entry.js
Hoten: done

lighthouse-core/lib/proto-preprocessor.js
patrickhulce: Gotcha, yeah basically I'm wondering why we don't need to prune anymore even though we still don't support 100% of the settings :)

Are the remaining extra properties any less of a concern than it was before? Based on the comments here, it seems like it was still technically unnecessary then but we decided to do it anyway 🤷‍♂️

Hoten:
Forgot about the round trip test :)

lighthouse-core/lib/proto-preprocessor.js
Hoten: it's just a couple string values now. before it was like, 95% of it. pruning is slightly more unnecessary now :)

proto/scripts/json_roundtrip_via_proto.py
```diff
@@ -33,7 +33,7 @@ def clean():
 proto_lhr = lhr_pb2.LighthouseResult()
 
 # fill proto lhr with data from JSON
-Parse(json.dumps(data), proto_lhr)
```
patrickhulce: I guess to finally figure out my real concern, is this something that anyone else will now have to change and match or are we the only people ever doing this proto-json conversion :)

apologies for being *very* slow on the uptake here 😆

Hoten: pretty sure the proto stuff has one user: us, with LR.

patrickhulce: awesome sauce

Hoten:
I added back the config setting pruning, since we are punting on the proto changes.

I think we could keep the ` lighthouse-core/test/report/proto-test.js` changes even without the proto changes. Good to keep in this PR?

Hoten:
Any more review on this?

patrickhulce:
Ah sorry connor I didn't flush my comments! I wish GH had an outbox of some sort that showed you all the PRs you had pending comments on :D

lighthouse-core/test/report/proto-test.js
```diff
@@ -55,21 +55,28 @@ describe('round trip JSON comparison subsets', () => {
   });
 
   it('has the same config values', () => {
-    expect(roundTripJson.configSettings).toMatchObject(sampleJson.configSettings);
+    // Config settings from proto round trip should be a subset of the actual settings.
+    expect(sampleJson.configSettings).toMatchObject(roundTripJson.configSettings);
```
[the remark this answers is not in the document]
Hoten: These objects are actually 100% equal, so either way would be fine. It was only apparent that the order is wrong when I was allowing more properties to be in the proto result.

so the test change isn't needed, but it is more correct. whatever settings are in the round trip json (from proto) should match the values as found in the originating settings json

lighthouse-core/test/report/proto-test.js
```diff
@@ -55,21 +55,28 @@ describe('round trip JSON comparison subsets', () => {
...
-// Note: In a failing diff, if you see details.summary going from {} to null, it's OK.
-// Jest considers this not a failure, and neither do we, here in the python roundtrip
-// Meanwhile, The PSI roundtrip maintains {} to {}.
 describe('round trip JSON comparison to everything', () => {
   let sampleJson;
 
   beforeEach(() => {
     sampleJson = JSON.parse(preprocessor.processForProto(sample));
+
+    // Proto conversion turns null summaries into an empty object. This is OK,
```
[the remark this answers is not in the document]
Hoten: The result is fine, but the jest matcher doesn't like `{}` when it expects `null`. The distinction is meaningless in proto world but important just for jest.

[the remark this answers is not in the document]
patrickhulce: I get that, but proto doesn't actually seem to turn `null` summaries into `{}` based on the previous comments and the fact that you are turning `{}` into `null` in the sample json, which means that proto seems to be turning `{}` into `null`, not the other way around...right?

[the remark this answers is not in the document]
Hoten: ahhhh yes, I got it all mixed up here. updated with what I intended to do

PR #7614, opened 2019-03-21
misc(proto): LighthouseConfig

Hoten:
**Summary**

Just the proto changes. See #7613 for more.

**Related Issues/PRs**
#7613

### DIFF
--- a/proto/lighthouse-result.proto
+++ b/proto/lighthouse-result.proto
@@ -130,6 +130,17 @@ message LighthouseResult {
       none = 3;
     }
 
+    enum ThrottlingMethod {
+      // Unknown. This should not be used.
+      THROTTLING_METHOD_UNSPECIFIED = 0;
+
+      devtools = 1;
+
+      simulate = 2;
+
+      provided = 3;
+    }
+
     // The form factor used in the audit
     EmulatedFormFactor emulated_form_factor = 1;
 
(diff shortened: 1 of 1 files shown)


proto/lighthouse-result.proto
```diff
@@ -138,10 +149,49 @@ message LighthouseResult {
...
+
+    // If present, the run should skip this list of audits.
+    // nullable list of strings
+    google.protobuf.ListValue skip_audits = 14;
+
+    // List of extra HTTP Headers to include.
+    map<string, string> extra_headers = 15;
+
+    // Precomputed lantern estimates to use instead of observed analysis.
+    PrecomputedLanternData precomputed_lantern_data = 16;
```
patrickhulce: I'm not super excited about exposing this in its current form to the world, is everything here going to be available over the PSI API automatically?

proto/lighthouse-result.proto
```diff
@@ -138,10 +149,49 @@ message LighthouseResult {
 
     // List of the categories that were run, empty if all were run
     // nullable list of strings
+    // Note: this should have been google.protobuf.ListValue, but too late now.
```
patrickhulce: What impact does this have on us? (same with onlyAudits I assume?

proto/lighthouse-result.proto
Hoten: All these changes will only be exposed in LR atm. But perhaps in the future we'd expand PSI to allow a config. We could disallow whatever settings we want from PSI.

Do you have a refactor of this setting in mind? If so, we should do that before encoding in proto. If this is just a setting that's only useful for testing I don't think the shape matters much.

proto/lighthouse-result.proto
Hoten: Little. The python / go proto serializers handle both just fine, given a list of strings. Technically LR users can pass anything to `only_categories` (like a string or a struct), and it would pass the serialization with flying colors, but ultimately fail the request (or at least, the setting would get ignored. i have not tried it). But the same is true for ListValue (pass a list of int32s, no bueno). so it's already an existing shortcoming.

proto/lighthouse-result.proto
patrickhulce: > Do you have a refactor of this setting in mind?

Well in my variance doc I proposed removing it completely depending on what direction we want to head, so I wouldn't want a brand new setting to popup for PSI users only to take it away a few weeks later :)

patrickhulce:
> I'd argue against keeping it private - why would we want to limit LR users?

It's mostly `precomputed_lantern_data` I'm concerned about, though now that I'm thinking it through more it might actually be a decent option for people trying to further mitigate their variance when using PSI for monitoring so...... 🤔

proto/lighthouse-result.proto
Hoten: We could drop it completely + disable the smokeTest that uses in Smokerider indefinitely.

Hoten:
@patrickhulce we can cut the lantern stuff, you ok with losing w/e coverage the lantern-expectations would get us in LR?

patrickhulce:
> we can cut the lantern stuff, you ok with losing w/e coverage the lantern-expectations would get us in LR?

I'm fine either way :)

You've reconvinced me that it would be worthwhile to have this property in LR contexts, and waiting to expose it also wouldn't hurt anything, so I'm happy whichever you decide 👍

Hoten:
We decided to spend more time and thought on the config before going all-in on the proto. I can do a hacky thing for Smokerider's needs for now.

Parking things in the branch `lr-config-proto`. Will continue work in #7613.

PR #7624, opened 2019-03-21
misc: fix unit-viewer command for windows

patrickhulce:
I don't tend to look at the appveyor tests too closely now that they are so flakey, but apparently our viewer filter has been ignored and so it defaults to re-running *all* our tests. As if it weren't slow enough already! 😭 

Turns out it wasn't quoted like the others, so this *should* fix it.

image

### DIFF
--- a/lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
+++ b/lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
@@ -40,7 +40,8 @@ module.exports = [
         details: {
           items: {
             0: {
-              url: /main-thread-consumer/,
+              // FIXME: Appveyor finds this particular assertion very flaky for some reason :(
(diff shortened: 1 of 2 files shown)


brendankenny:
I had removed the quotes because I get `Invalid testPattern lighthouse-viewer/**/*-test.js supplied. Running all tests instead.`

brendankenny:
oh, I think if it's in quotes it has to be a regex (that jest uses to filter all inputs) vs just globbing a list of input files (like this is doing, that would be broken if the shell doesn't support that)

patrickhulce:
> oh, I think if it's in quotes it has to be a regex (that jest uses to filter all inputs) vs just globbing a list of input files (like this is doing, that would be broken if the shell doesn't support that)

Ha I just found this too, great minds :)

patrickhulce:
I also disabled the one assertion that's particularly flaky on appveyor for now

PR #7627, opened 2019-03-21
misc(travis): yarn build-devtools

Hoten:
let's see what happens


brendankenny:
should we close? Without a lot of work those TS annotations aren't going to work inside closure :)

PR #7672, opened 2019-03-24
report: add basic messageFormat() for '7 audits'

paulirish:
We've put off this one for a while. But this allows us to translate the last english string that's visible in current PSI.

image

And opens up our ability to translate other strings that include placeholder values such as:
image

### DIFF
--- a/lighthouse-core/report/html/renderer/util.js
+++ b/lighthouse-core/report/html/renderer/util.js
@@ -95,6 +95,22 @@ class Util {
   }
 
   /**
+   * Poor man's implementation of i18n._formatIcuMessage()
+   * @param {{message: string, values: Object|undefined}} audit
+   * @return {string}
+   */
+  static basicMsgFormat({message, values}) {
+    let resolvedMsg = message;
+    for (const [valueName, value] of Object.entries(values)) {
+      // Look for `{valueName}` or `{valueName, number}`
+      // eslint-disable-next-line no-useless-escape
+      const re = new RegExp('\{\\s*?' + valueName + '\\b.*?\}');
+      resolvedMsg = resolvedMsg.replace(re, value);
+    }
+    return resolvedMsg;
+  }
+
+  /**
    * Used to determine if the "passed" for the purposes of showing up in the "failed" or "passed"
    * sections of the report.
    *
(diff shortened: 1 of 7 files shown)


lighthouse-core/report/html/renderer/util.js
```diff
@@ -94,6 +94,22 @@ class Util {
...
+   * Poor man's implementation of i18n._formatIcuMessage()
+   * @param {{message: string, values: Object|undefined}} audit
+   * @return {string}
+   */
+  static basicMsgFormat({message, values}) {
+    let resolvedMsg = message;
+    for (const [valueName, value] of Object.entries(values)) {
+      // Look for `{valueName}` or `{valueName, number}`
+      // eslint-disable-next-line no-useless-escape
+      const re = new RegExp('\{\\s*?' + valueName + '\\b.*?\}');
```
patrickhulce: should this be a little narrower given we don't support everything? i.e. maybe `(,\s*number)?}`

lighthouse-core/report/html/renderer/util.js
```diff
@@ -503,6 +519,9 @@ Util.UIStrings = {
   /** Label for button that only shows a few lines of the snippet when clicked */
   snippetCollapseButtonLabel: 'Collapse snippet',
 
+  /** Label shown above a list of collapsed to describe how many are in the group. The `{itemCount}` placeholder will be replaced with the number of audits, typically an integer between 1 and 30 */
```
patrickhulce: should we note that we don't handle plurality and they should pick the version that will just work the most often?

patrickhulce: didn't even realize we don't support plurality for english 😆

lighthouse-core/report/html/renderer/util.js
```diff
@@ -94,6 +94,22 @@ class Util {
...
+  /**
+   * Poor man's implementation of i18n._formatIcuMessage()
+   * @param {{message: string, values: Object|undefined}} audit
+   * @return {string}
+   */
+  static basicMsgFormat({message, values}) {
+    let resolvedMsg = message;
+    for (const [valueName, value] of Object.entries(values)) {
+      // Look for `{valueName}` or `{valueName, number}`
+      // eslint-disable-next-line no-useless-escape
```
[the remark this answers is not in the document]
paulirish: oh man i was so convinced it was flagging the double escapes, but i think you're right.. :)

lighthouse-core/report/html/renderer/util.js
paulirish: wfm

lighthouse-core/report/html/renderer/util.js
paulirish: Oh plurals! Yes that's why nobody had started this one yet. 🤦‍♂️

exterkamp: +1 ICU Plurals for this

brendankenny:
ideas:
- more custom formatter. One confusing issue is that these strings are special, ending up in the LHR *without* having their message replaced
- ship `intl-messageformat` in our report. How big does it get? (and same issue as above)
- bite the bullet and have a certain set of strings we put multiple times in `rendererFormattedStrings`, once for each category (won't help with per-audit strings like "Show 3rd party resources", unless we also have another set of strings we put in the LHR many more times)
- (like above, but dynamic) actually run the html report generator in `runner`, collecting any strings it translates along the way, then put them in `rendererFormattedStrings` in the LHR when done
- Figure out a nice way to have a translated string and a number without message replacement. Similar to how concatenation is usually fine for translated strings, maybe there's a nice way to do the "Show 3rd party resources" + "(1)" that makes sense across languages. ("Audits (7)") is less nice than "7 audits", but maybe we can live with that.

paulirish:
Consensus is do #5 above. In the future we could improve. We should also fix metric names to not be translated, as that's a more pressing issue.

brendankenny:
report's a-changin, this'll be out of sync soon, so...

brendankenny:
(but the good news is we basically are going with option 5)

Issue #7700, opened 2019-03-26
main-thread-tasks audit does not use toplevel events


alekseykulikov:
#### Provide the steps to reproduce
1. Run LH on any URL
2. Compare `main-thread-tasks` and `diagnostics`. It's not possible to reproduce diagnostics using data from `main-thread-tasks`.

#### What is the current behavior?

`main-thread-tasks` audit exclude toplevel events:
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/main-thread-tasks.js#L35

`diagnostics` uses only toplevel events:
https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/diagnostics.js#L40

#### What is the expected behavior?

It looks like a typo in `main-thread-tasks` and the filter expression should be: 

```js
const results = tasks.filter(task => task.duration > 5 && !task.parent)
```
I can provide a PR if the proposed change is correct.
Or do I have a wrong assumption that `main-thread-tasks` audit should not accompany `diagnostics`?

#### Environment Information
* Lighthouse version: 4.2.0

patrickhulce:
thanks for filing @alekseykulikov this is indeed a typo, I guess this is the problem with adding audits that aren't used for anything :)

the `main-thread-tasks` audit should be `!task.parent`, diagnostics audit is correct

alekseykulikov:
@patrickhulce thank you for fixing it quickly!
The audit is extremely useful, as well as other diagnostics added in v4.2.0.

patrickhulce:
Fixed by #7711

PR #7711, opened 2019-03-26
core(main-thread-tasks): use toplevel tasks in main thread audit

patrickhulce:
**Summary**
Typo in the `main-thread-tasks` audit, I didn't get around to surfacing this yet, so it shouldn't have hurt anything. I think I was originally going for first inner child of toplevel task so we'd actually have a type, but then again I wasn't using the results yet, so 🤷‍♂️ 

**Related Issues/PRs**
#7700

### DIFF
--- a/lighthouse-core/audits/main-thread-tasks.js
+++ b/lighthouse-core/audits/main-thread-tasks.js
@@ -35,7 +35,6 @@ class MainThreadTasks extends Audit {
       .filter(task => task.duration > 5 && !task.parent)
       .map(task => {
         return {
-          type: task.group.id,
           duration: task.duration,
           startTime: task.startTime,
         };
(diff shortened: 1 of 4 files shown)


patrickhulce:
cc @alekseykulikov

lighthouse-core/audits/main-thread-tasks.js
```diff
@@ -32,18 +32,16 @@ class MainThreadTasks extends Audit {
     const tasks = await MainThreadTasksComputed.request(trace, context);
 
     const results = tasks
-      .filter(task => task.duration > 5 && task.parent)
+      .filter(task => task.duration > 5 && !task.parent)
```
brendankenny: maybe add a comment that top level === no parent here? Easy to miss the point of the filter though it's obvious once you realize :)

patrickhulce: yeah good call, I mean I missed it the first time around 😆

alekseykulikov:
LGTM!

PR #7723, opened 2019-03-26
misc: drive-by code cleanup in BaseNode

brendankenny:
While starting work on robust TTI, there were some things that were a little hard to follow as I was stepping through graphs. Tried to make them a bit clearer in code and comments while keeping everything working the same.

@patrickhulce let me know if this feels like too much change for change's sake and I can scale it back :)

### DIFF
--- a/lighthouse-core/lib/dependency-graph/base-node.js
+++ b/lighthouse-core/lib/dependency-graph/base-node.js
@@ -168,96 +168,79 @@ class BaseNode {
 
   /**
    * Clones the entire graph connected to this node filtered by the optional predicate. If a node is
-   * included by the predicate, all nodes along the paths between the two will be included. If the
-   * node that was called clone is not included in the resulting filtered graph, the method will throw.
+   * included by the predicate, all nodes along the paths between the node and the root will be included. If the
+   * node this was called on is not included in the resulting filtered graph, the method will throw.
    * @param {function(Node):boolean} [predicate]
    * @return {Node}
    */
   cloneWithRelationships(predicate) {
     const rootNode = this.getRootNode();
 
-    /** @type {function(Node): boolean} */
-    let shouldIncludeNode = () => true;
-    if (predicate) {
-      const idsToInclude = new Set();
-      rootNode.traverse(node => {
-        if (predicate(node)) {
-          node.traverse(
-            node => idsToInclude.add(node.id),
-            node => node._dependencies.filter(parent => !idsToInclude.has(parent))
-          );
-        }
-      });
+    /** @type {Map<string, Node>} */
+    const idsToIncludedClones = new Map();
 
-      shouldIncludeNode = node => idsToInclude.has(node.id);
-    }
+    // Walk down dependents.
+    rootNode.traverse(node => {
+      // Find new and includable nodes.
+      if (idsToIncludedClones.has(node.id)) return;
+      if (predicate && !predicate(node)) return;
 
-    const idToNodeMap = new Map();
-    rootNode.traverse(originalNode => {
-      if (!shouldIncludeNode(originalNode)) return;
-      const clonedNode = originalNode.cloneWithoutRelationships();
-      idToNodeMap.set(clonedNode.id, clonedNode);
+      // Walk back up dependencies, cloning all nodes in the path to this one.
+      node.traverse(
+        node => idsToIncludedClones.set(node.id, node.cloneWithoutRelationships()),
+        node => node._dependencies.filter(parent => !idsToIncludedClones.has(parent.id))
+      );
     });
 
+    // Copy dependencies between nodes.
     rootNode.traverse(originalNode => {
-      if (!shouldIncludeNode(originalNode)) return;
-      const clonedNode = idToNodeMap.get(originalNode.id);
+      const clonedNode = idsToIncludedClones.get(originalNode.id);
+      if (!clonedNode) return;
 
       for (const dependency of originalNode._dependencies) {
-        const clonedDependency = idToNodeMap.get(dependency.id);
+        const clonedDependency = idsToIncludedClones.get(dependency.id);
+        if (!clonedDependency) throw new Error('Dependency somehow not cloned');
         clonedNode.addDependency(clonedDependency);
       }
     });
 
-    if (!idToNodeMap.has(this.id)) throw new Error('Cloned graph missing node');
-    return idToNodeMap.get(this.id);
+    const clonedThisNode = idsToIncludedClones.get(this.id);
+    if (!clonedThisNode) throw new Error('Cloned graph missing node');
+    return clonedThisNode;
   }
 
   /**
-   * Traverses all paths in the graph, calling iterator on each node visited. Decides which nodes to
-   * visit with the getNext function.
-   * @param {function(Node, Node[]): void} iterator
-   * @param {function(Node): Node[]} getNext
+   * Traverses all connected nodes in BFS order, calling `callback` exactly once
+   * on each. `traversalPath` is the shortest (though not necessarily unique)
+   * path from `node` to the root of the iteration.
+   *
+   * The `getNextNodes` function takes a visited node and returns which nodes to
+   * visit next. It defaults to returning the node's dependents.
+   * @param {(node: Node, traversalPath: Node[]) => void} callback
+   * @param {function(Node): Node[]} [getNextNodes]
    */
-  _traversePaths(iterator, getNext) {
-    const stack = [[/** @type {Node} */(/** @type {BaseNode} */(this))]];
-    while (stack.length) {
-      /** @type {Node[]} */
-      // @ts-ignore - stack has length so it's guaranteed to have an item
-      const path = stack.shift();
-      const node = path[0];
-      iterator(node, path);
+  traverse(callback, getNextNodes) {
+    if (!getNextNodes) {
+      getNextNodes = node => node.getDependents();
+    }
 
-      const nodesToAdd = getNext(node);
-      for (const nextNode of nodesToAdd) {
-        stack.push([nextNode].concat(path));
+    const visited = new Set([this.id]);
+    const queue = [[/** @type {Node} */(/** @type {BaseNode} */(this))]];
+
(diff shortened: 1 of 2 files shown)


brendankenny:
slightly better diff with [?w=1](https://github.com/GoogleChrome/lighthouse/pull/7723/files?w=1)

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
 
-    /** @type {function(Node): boolean} */
-    let shouldIncludeNode = () => true;
-    if (predicate) {
-      const idsToInclude = new Set();
-      rootNode.traverse(node => {
-        if (predicate(node)) {
-          node.traverse(
-            node => idsToInclude.add(node.id),
-            node => node._dependencies.filter(parent => !idsToInclude.has(parent))
```
brendankenny: dangers of `Set<any>`...we were checking the `parent` node object against a set of string IDs, so weren't saving any time here

patrickhulce: nice fix!!

patrickhulce: I guess this one is difficult to test since it's just a marginal time savings huh

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
+    // Walk down dependents.
+    rootNode.traverse(node => {
+      // Find new and includable nodes.
+      if (idsToIncludedClones.has(node.id)) return;
+      if (predicate && !predicate(node)) return;
+
+      // Walk back up dependencies, cloning all nodes in the path to this one.
+      node.traverse(
+        node => idsToIncludedClones.set(node.id, node.cloneWithoutRelationships()),
+        node => node._dependencies.filter(parent => !idsToIncludedClones.has(parent.id))
```
patrickhulce: can we get a comment on this one that explains the filter while we're at it? :) it even took me a sec to remember why this was happening at all

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
-      if (!shouldIncludeNode(originalNode)) return;
-      const clonedNode = originalNode.cloneWithoutRelationships();
-      idToNodeMap.set(clonedNode.id, clonedNode);
+    // Walk down dependents.
+    rootNode.traverse(node => {
+      // Find new and includable nodes.
+      if (idsToIncludedClones.has(node.id)) return;
+      if (predicate && !predicate(node)) return;
+
+      // Walk back up dependencies, cloning all nodes in the path to this one.
```
patrickhulce: can we only do the reverse traversal when there's a predicate? 

that's one nice feature of the previous impl I wouldn't want to lose

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
-   * visit with the getNext function.
-   * @param {function(Node, Node[]): void} iterator
-   * @param {function(Node): Node[]} [getNext] Defaults to returning the dependents.
-   */
-  traverse(iterator, getNext) {
-    if (!getNext) {
-      getNext = node => node.getDependents();
-    }
+    const visited = new Set([this.id]);
+    const queue = [[/** @type {Node} */(/** @type {BaseNode} */(this))]];
```
patrickhulce: do we still need this? I'm surprised we need to cast `BaseNode` for `this` hopefully this is something that was fixed :)

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
+   * on each. `traversalPath` is the shortest (though not necessarily unique)
+   * path from `node` to the root of the iteration.
+   *
+   * The `getNextNodes` function takes a visited node and returns which nodes to
+   * visit next. It defaults to returning the node's dependents.
+   * @param {(node: Node, traversalPath: Node[]) => void} callback
+   * @param {function(Node): Node[]} [getNextNodes]
    */
-  _traversePaths(iterator, getNext) {
-    const stack = [[/** @type {Node} */(/** @type {BaseNode} */(this))]];
```
patrickhulce: lol I guess this is a queue :)

lighthouse-core/lib/dependency-graph/base-node.js
```diff
@@ -168,96 +168,79 @@ class BaseNode {
...
 
-    getNext = node => {
-      visited.add(node.id);
-      const allNodesToVisit = originalGetNext(node);
-      const nodesToVisit = allNodesToVisit.filter(nextNode => !visited.has(nextNode.id));
-      nodesToVisit.forEach(nextNode => visited.add(nextNode.id));
-      return nodesToVisit;
-    };
+      for (const nextNode of getNextNodes(node)) {
+        if (visited.has(nextNode.id)) continue;
```
patrickhulce: having trouble finding the substantial changes here, it's just inlining the `getNext` rewrite from before and eliminating `traverse` right?

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > I guess this one is difficult to test since it's just a marginal time savings huh

yeah, and even at its worst it's really not that bad compared to other things we do with the graph

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > can we get a comment on this one that explains the filter while we're at it? 

good call

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > do we still need this? 

Annoyingly, yes. `this` *extends* `BaseNode`, but the lie is that it's a `Node`. Humans know the only types of `BaseNode` are `CPUNode | NetworkNode` so the cast to `Node` is ok, but the compiler accounts for the fact that we could be calling on a different class derived from `BaseNode`, so it's not necessarily compatible with `Node`.

So to cast while following the type assertion rules, we have to upcast to `BaseNode` and then downcast to `Node`. The other options is an assertion and a `@ts-ignore`...maybe that's more honest (albeit also ugly).

brendankenny: > Humans know the only types of `BaseNode` are `CPUNode | NetworkNode` so

I mean, even that's not entirely true. In tests we call `traverse()` with a graph of `BaseNode` nodes, but in real code it's much more convenient to only get `CPUNode | NetworkNode` back, so we should continue to pretend it's true :)

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > lol I guess this is a queue :)

your comment in `hasCycle` (which does DFS) underlined the humor: 

> // Get the last node in the stack (DFS uses stack, not queue)

I was thinking the approach changed at some point during the writing of `traverse` :)

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > it's just inlining the `getNext` rewrite from before and eliminating `_traversePaths` right?

yeah, main thing was the `getNext` rewriting and call to `_traversePaths` was breaking my brain when I was first (re)reading it, and it seemed like `traverse` has proven its design, so it seemed ok to inline at this point.

This was probably the bigger one for me saying it's ok to push back on changes if you don't think it reads better :)

*technically* we save n-1 `visited.add(node.id)` calls, but that's not worth much and could have been done without all the other things.

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: > can we only do the reverse traversal when there's a predicate?

do you mean something like

```js
// Walk down dependents.
rootNode.traverse(node => {
  if (idsToIncludedClones.has(node.id)) return;

  if (!predicate) {
    // Clone every node.
    idsToIncludedClones.set(node.id, node.cloneWithoutRelationships());
  } else if (predicate(node)) {
    // Walk back up dependencies, cloning all nodes in the path to this one.
    node.traverse(
      node => idsToIncludedClones.set(node.id, node.cloneWithoutRelationships()),
      // Dependencies already cloned have already cloned ancestors, so no need to visit again.
      node => node._dependencies.filter(parent => !idsToIncludedClones.has(parent.id))
    );
  }
});
```

Or other ideas for formatting to make it less dense looking?

Another option is to make a default always true predicate, so it's a little clearer.

Agreed it is a bit weird to do the reverse traversal in the no predicate case. My thinking before was that we only have a single call of `cloneWithRelationships` without a predicate, so it didn't seem to be a super important case to call out (and since the filter line makes there be no next nodes for every node, it doesn't do any significant extra work)

lighthouse-core/lib/dependency-graph/base-node.js
patrickhulce: haha yeah that was my guess as well when I was looking at this, I'll take your benefit of the doubt on this one lol 🤞

lighthouse-core/lib/dependency-graph/base-node.js
patrickhulce: that's exactly what I meant :)

we could do an early return in the `!predicate` case to avoid the nesting? but IMO this is fine

lighthouse-core/lib/dependency-graph/base-node.js
patrickhulce: nah, I like this. originally I had used `_traversePaths` in another method but that need hasn't materialized in the 1.5 years since so I think we're pretty safe :)

lighthouse-core/lib/dependency-graph/base-node.js
brendankenny: done

Issue #7691, opened 2019-03-25
Investigate increase in "bootup-time" errors


connorjclark:
@patrickhulce 

awful repro:
```

N=30

export CHROME_PATH="/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary"

mkdir -p timing-runs

for run in {1..$N}
do
	echo "run NS $run"
	output_path="timing-runs/ns-$run.json"
	node lighthouse-cli https://www.elbatal-egypt.com/nader/ --output json --output-path="$output_path" --chrome-flags="--enable-features=NetworkService" --save-assets 2> /dev/null
	grep -q "Invalid task timing data" "$output_path"
	if [ $? -eq 0 ]; then
		echo broke
	fi
done

for run in {1..$N}
do
	echo "run no NS $run"
	output_path="timing-runs/no-ns-$run.json"
	node lighthouse-cli https://www.elbatal-egypt.com/nader/ --output json --output-path="$output_path" --chrome-flags="--disable-features=NetworkService" --save-assets 2> /dev/null
	grep -q "Invalid task timing data" "$output_path"
	if [ $? -eq 0 ]; then
		echo broke
	fi
done

for trace in timing-runs/*.trace.json; do
	echo $trace
	grep '"ph":"B"' $trace | wc -l ; grep '"ph":"E"' $trace | wc -l ;
done
```

Runs 12 and 15 for NS on failed. NS off never failed. Using Canary 75.0.3745.0.

Just NS 12 and 15:

[ns-12-and-15.zip](https://github.com/GoogleChrome/lighthouse/files/3005753/ns-12-and-15.zip)

```
timing-runs/no-ns-1-0.trace.json
   11057
   11052
timing-runs/no-ns-10-0.trace.json
   13173
   13168
timing-runs/no-ns-11-0.trace.json
   11194
   11189
timing-runs/no-ns-12-0.trace.json
   10710
   10705
timing-runs/no-ns-13-0.trace.json
   10992
   10986
timing-runs/no-ns-14-0.trace.json
   11276
   11270
timing-runs/no-ns-17-0.trace.json
   10534
   10529
timing-runs/no-ns-18-0.trace.json
   10845
   10839
timing-runs/no-ns-19-0.trace.json
   11046
   11041
timing-runs/no-ns-2-0.trace.json
    9754
    9750
timing-runs/no-ns-20-0.trace.json
   11345
   11339
timing-runs/no-ns-21-0.trace.json
[...]

PR #7728, opened 2019-03-27
core(main-thread-tasks): handle last task not finishing

patrickhulce:
**Summary**
Treats `'B'` events without a corresponding `'E'` as if they ended at the time of trace end.

**Related Issues/PRs**
#7691

### DIFF
--- a/lighthouse-core/computed/main-thread-tasks.js
+++ b/lighthouse-core/computed/main-thread-tasks.js
@@ -74,20 +74,16 @@ class MainThreadTasks {
   /**
    * @param {LH.TraceEvent[]} mainThreadEvents
    * @param {PriorTaskData} priorTaskData
+   * @param {number} traceEndTs
    * @return {TaskNode[]}
    */
-  static _createTasksFromEvents(mainThreadEvents, priorTaskData) {
+  static _createTasksFromEvents(mainThreadEvents, priorTaskData, traceEndTs) {
     /** @type {TaskNode[]} */
     const tasks = [];
     /** @type {TaskNode|undefined} */
     let currentTask;
-    let lastTraceTs = 0;
 
     for (const event of mainThreadEvents) {
-      // Keep track of the last trace event timestamp
-      if (event.ts > lastTraceTs) lastTraceTs = event.ts;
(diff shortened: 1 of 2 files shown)


Hoten:
Looks good. I'll verify in Canary now.

lighthouse-core/computed/main-thread-tasks.js
```diff
@@ -81,8 +81,13 @@ class MainThreadTasks {
     const tasks = [];
     /** @type {TaskNode|undefined} */
     let currentTask;
+    let lastTraceTs = 0;
 
     for (const event of mainThreadEvents) {
+      // Keep track of the last trace event timestamp
+      if (event.ts > lastTraceTs) lastTraceTs = event.ts;
+      if (currentTask && currentTask.endTime > lastTraceTs) lastTraceTs = currentTask.endTime;
```
brendankenny: rather than doing this (which is a hard to follow with the `currentTask` check and the (I assume) `NaN` comparison), could we either pass in `TraceOfTab.timestamps.traceEnd` or (probably less ideal) replicate the logic from there since the events are sorted by ts?

lighthouse-core/computed/main-thread-tasks.js
```diff
@@ -133,6 +138,12 @@ class MainThreadTasks {
       }
     }
 
+    while (currentTask && !Number.isFinite(currentTask.endTime)) {
+      // The last event didn't finish before tracing stopped, just end it when tracing ended.
+      currentTask.endTime = lastTraceTs;
+      currentTask = currentTask.parent;
+    }
+
```
paulirish: Lets add a comment here with our expectation like

    // At this point we expect all tasks to have a finite startTime and endTime.

lighthouse-core/computed/main-thread-tasks.js
```diff
@@ -133,6 +138,12 @@ class MainThreadTasks {
       }
     }
 
+    while (currentTask && !Number.isFinite(currentTask.endTime)) {
```
paulirish: add this sorta comment in here? took me a second to determine what currentTask was at this point.

    // Starting from the last and bottom-most task,

Issue #7160, opened 2019-02-05
Ideas for trimming the LHR


brendankenny:
The LHR is getting big, often 1MB or more. Some unnecessary things we could trim:
- [ ] a11y per-node `explanation` are on every details item, often repeated verbatim on multiple nodes, and we don't use them in the report
- [ ] `tap-targets` includes all possible bad targets. We should limit the number.

Feel free to add to the list.

brendankenny:
whoops, some unnecessary things we could *trim*, feel free to contribute to the list :)

Hoten:
We could remove entire source contents if the file is not referenced in any audit.

Hoten:
Timings are ~9KB

![image](https://user-images.githubusercontent.com/4071474/52352041-b9e95300-29e0-11e9-9101-4de88e4d4b22.png)

We could fix the numbers so we don't have 100000 digits pass the decimal :)
"measure" -> "m" (actually, is that property even necessary?)

Hoten:
Could we embed the `i18n` data to the viewer, or would that get complicated real quick?

patrickhulce:
> * a11y per-node `explanation` are on every details item, often repeated verbatim on multiple nodes, and we don't use them in the report

This was an explicit bug report when we removed it the first time around and had to add it back because for some nodes the information is specific to that node (color contrast for example). https://github.com/GoogleChrome/lighthouse/issues/5402, we should be careful about removing it.

> Could we embed the `i18n` data in the report renderer, or would that get complicated real quick?

There's enough blubber elsewhere in the LHR to remove that I'd want to do this last. No need to hamper the beautiful simplicity and flexibility of i18n yet IMO :D

Also worth clarifying here what we're worried about. Is it API transfer sizes? Is it storage size on disk? In a database? JSON parse times? Biggest gzipped wins will probably be different from biggest uncompressed wins, and some strategies might reduce LHR size but not really help certain use cases.

Example: I can think of many ways in which we could greatly shrink total API bytes by splitting the LHR into its dynamic and static components just for transport.

brendankenny:
> This was an explicit bug report when we removed it the first time around and had to add it back because for some nodes the information is specific to that node (color contrast for example).

arg, I forgot about that. I wonder if there's some deduping we could do then...some of the strings are quite long and occur multiple times.

> Also worth clarifying here what we're worried about. Is it API transfer sizes? Is it storage size on disk? In a database? JSON parse times?

[...]

PR #7748, opened 2019-03-27
core: truncate measure timings to hundredths

Hoten:
Reduces LHR size by ~4KB. That was ~1% of the LHR I tested it with.

**Related Issues/PRs**

#7160

### DIFF
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -179,13 +179,16 @@ class Runner {
       ...timingEntriesFromRunner,
       // As entries can share a name, dedupe based on the startTime timestamp
     ].map(entry => /** @type {[number, PerformanceEntry]} */ ([entry.startTime, entry]));
-    const timingEntries = Array.from(new Map(timingEntriesKeyValues).values());
-    for (const timing of timingEntries) {
-      // @ts-ignore - ignore readonly
-      timing.startTime = parseFloat(timing.startTime.toFixed(2));
-      // @ts-ignore - ignore readonly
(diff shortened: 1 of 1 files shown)


Hoten:
Wait, wrong timings :)

lighthouse-core/runner.js
```diff
@@ -180,6 +180,12 @@ class Runner {
       // As entries can share a name, dedupe based on the startTime timestamp
     ].map(entry => /** @type {[number, PerformanceEntry]} */ ([entry.startTime, entry]));
     const timingEntries = Array.from(new Map(timingEntriesKeyValues).values());
+    for (const timing of timingEntries) {
```
paulirish: add a comment above this loop to explain why this is being done.

lighthouse-core/runner.js
```diff
@@ -180,6 +180,12 @@ class Runner {
...
+      // @ts-ignore - ignore readonly
+      timing.startTime = parseFloat(timing.startTime.toFixed(2));
```
brendankenny: It looks like you may have to make a copy of these in order to do this. Real `PerformanceEntry`'s properties really do appear to be read only

paulirish: yeaup. 
image

Hoten: done

paulirish:
shout out to @wardpeet 	who made extension-test a thing!  that failure was so rad. :D

patrickhulce:
hip hip hurray!! @wardpeet is like, *the* best example of open source awesomeness around 😃

PR #7778, opened 2019-03-28
core(tap-targets): don't exclude visible position absolute elements from audit

mattzeunert:
**Summary**

- include tap targets that are position absolute or have a position absolute ancestor
- check that tap targets are visible using elementFromPoint

Screenshot 2019-03-28 at 16 29 48

Also catching this now:

Screenshot 2019-04-25 at 13 58 45

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/7365


lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -287,17 +346,66 @@ function gatherTapTargets() {
...
+    // todo: do soething to prevent infinite loop here
+    while (!pointIsInViewport(largestRectCenterPoint)) {
+      const finalScrollPos = document.documentElement.scrollHeight;
+      if (document.documentElement.scrollTop >= finalScrollPos) {
+        throw Error('scrolled all the way but not found');
+      }
+      document.documentElement.scrollTop += window.innerHeight;
+    }
+
+    const isTop = elementCenterIsAtZAxisTop(tapTargetElement, largestRectCenterPoint);
```
patrickhulce: so this is really beyond just working with absolute elements this is about making sure we're only looking at what's on top for *all* elements, right?

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -287,17 +346,66 @@ function gatherTapTargets() {
...
+  enhancedTapTargets.sort(
+    (a, b) => {
+      return a.largestRectCenterPoint.y - b.largestRectCenterPoint.y;
+    }
+  );
+
+  // Disable point events so that tap targets below them don't get
+  // detected as non-tappable (they are tappable, just not while the viewport
+  // is at the current scroll position)
+  const reenableFixedAndStickyElementPointerEvents = disableFixedAndStickyElementPointerEvents();
```
patrickhulce: so elementFromPoint uses pointer events to decide what's on top? interesting, so it really is from the mouse-perspective then, not visual. TIL :)

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: It is right now, but just because I forgot to just do the check for position absolute elements. I think we should filter it.

mattzeunert:
> high level looks pretty good! did you have the stats on sites that start failing more now in the issue or was that something else I was remembering?

I did a few of those recently, so you're probably thinking of something else. But I do plan to get some stats for this one too.

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Just realized that we need to handle full-page app install/newsletter/GDPR overlays...

We could check if all elements are top layer, not just position absolute ones. Then if you have a full-page overlay you get an easy 100% score.

Or we could only check position absolute elements, but then we can get false failures with the content behind them.

I'm leaning towards the first option... if your site has a full-page overlay it kinda makes sense that the tap targets audit isn't very useful?

mattzeunert:
DZL, do a barrel roll!

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: +1 to the first approach. We should keep the signal as high as possible in this audit, and I think false positives from absolute elements would hurt that if we went with option 2.

If they have a overlay, it makes sense that we can only test what we see :)

mattzeunert:
Here are [some examples of how scores are impacted](https://gist.github.com/mattzeunert/7e1d8f3bfd2ce7aa29ab1b4e6790966c). Generally it looks good, not just in terms of including position absolute elements, we're also detecting much better whether elements are visible. Do you think we need to do anything about the `overflow: hidden` issue on Godaddy?

For 100 sites I tested there's a change of ±5.7 point on average. Positive and negative changes even out overall and there's a drop of just over 1 point in the average score.

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -287,17 +352,66 @@ function gatherTapTargets() {
...
+
+  let item;
+  while (item = enhancedTapTargets.shift()) {
+    const {tapTargetElement, largestRectCenterPoint, visibleClientRects} = item;
+
+    while (!pointIsInViewport(largestRectCenterPoint)) {
+      const previousScrollY = window.scrollY;
+      window.scrollTo(0, window.scrollY + window.innerHeight);
+
+      if (window.scrollY === previousScrollY) {
```
mattzeunert: I wonder how we should handle this case... it's not really supposed to happen so throwing might just be ok. On the other hand if we can't find it in the doc that means it's not visible, so we could also just set `isTop` to false

One thing we need to handle is scrollable divs that can contain elements outside the document  dimensions. Right now those are filtered out by `filterClientRectsWithinAncestorsVisibleScrollArea`, but it would be nice if could just get rid of all that logic.

mattzeunert:
@patrickhulce Should be good for some more review now 🙂

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -94,6 +99,7 @@ function getVisibleClientRects(element) {
   // - tap targets audit doesn't consider different containers/layers
   // - having most content in an explicit scroll container is rare
   // - treating them as hidden only generates false passes, which is better than false failures
+  // TODO: is this still needed?
```
mattzeunert: Right now only need this to filter out rects that are outside the visible scroll area, and which end up being the largest rect. Without `filterClientRectsWithinAncestorsVisibleScrollArea` we'll fail to detect those tap targets.

Maybe that's fine? If we didn't filter that rect out, and that rect did overlap with another tap target, then the failure would be invalid.

Maybe we shouldn't check the largest rect of each tap target with `elementFromPoint`, but rather filter the list of client rects calling `elementFromPoint` on each one?

lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -185,8 +185,8 @@ module.exports = [
     audits: {
       'tap-targets': {
         score: (() => {
-          const PASSING_TAP_TARGETS = 11;
-          const TOTAL_TAP_TARGETS = 12;
+          const PASSING_TAP_TARGETS = 12;
+          const TOTAL_TAP_TARGETS = 13;
```
mattzeunert: I'd would be nice to assert each tap target from the artifact here, maybe we can modify the expectations logic to allow that?

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -115,7 +121,7 @@ function filterClientRectsWithinAncestorsVisibleScrollArea(element, clientRects)
     const parentBCR = parent.getBoundingClientRect();
     clientRects = clientRects.filter(cr => rectContains(parentBCR, cr));
   }
-  if (parent.parentElement && parent.parentElement.tagName !== 'BODY') {
+  if (parent.parentElement && parent.parentElement !== document.documentElement) {
```
patrickhulce: just curious, what specific dom structure did this fix :)

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -249,6 +255,42 @@ function truncate(str, maxLength) {
...
+
+/* istanbul ignore next */
+function disableFixedAndStickyElementPointerEvents() {
+  const className = 'lighthouse-disable-pointer-events';
+  const styleTag = document.createElement('style');
+  styleTag.innerHTML = `.${className} { pointer-events: none !important }`;
+  document.body.appendChild(styleTag);
+
+  Array.from(document.querySelectorAll('*')).forEach(el => {
+    const position = /** @type string */(getComputedStyle(el).position);
```
patrickhulce: hm, why do we need the type assertion?

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -257,10 +299,20 @@ function gatherTapTargets() {
...
   // @ts-ignore - getElementsInDocument put into scope via stringification
   const tapTargetElements = getElementsInDocument(tapTargetsSelector);
 
+
+  /** @type {{
+    tapTargetElement: Element,
+    largestRectCenterPoint: {x: number, y: number},
+    visibleClientRects: ClientRect[]
+  }[]} */
+  const enhancedTapTargets = [];
```
patrickhulce: i don't love `enhanced` since I'm not sure what that means but I don't have a ton of other suggestions


`tapTargetsWithClientRects` or something maybe?

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: Sorry, what's the final state of this now then?

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: > Maybe we shouldn't check the largest rect of each tap target with elementFromPoint, but rather filter the list of client rects calling elementFromPoint on each one?

I'm not sure I understand the proposal. What's the distinction between that and what we do now? what would start failing/passing?

lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
```diff
@@ -93,13 +93,15 @@ <h1>SEO Tap targets</h1>
...
     <!-- Only target that's being overlapped should fail, the overlapping one shouldn't -->
-    <div>
-      <a style="display: block; width: 100px; height: 30px;background: #ddd;">
-        too small target
-      </a>
-      <a style="display: block; width: 100px; height: 100px;background: #aaa;">
-        big enough target
-      </a>
+    <div style="position: relative; height: 200px;">
+      <div style="position: absolute;">
```
patrickhulce: maybe add a note that this is just testing our `absolute` handling ability?

I feel like we should also throw in false positive case that we handle correctly. Maybe one of the ones you were thinking about in https://github.com/GoogleChrome/lighthouse/pull/7778/files#r271846819?

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: - we go with option 1, so if a lot of content is covered by a position absolute element then we ignore those tap targets
- we ignore position `fixed`/`sticky` elements when calling `getElementFromPoint`, so tap targets underneath them are detected – we were doing that before but I got confused because reddit had a fixed overlay with `pointer-events: all`. My CSS rule is now `!important`, so that's fixed

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Current approach in this PR:

- get all client rects that aren't 0x0px or outside the scroll ancestor
- get the largest of those rects
- the element is visible if `getElementFromPoint` at the center of the largest rect matches the element (but the client rect list also contains rects where the center is invisible)

Proposal:

- get all client rects that aren't 0x0px (or outside the scroll ancestor)
- each rect is visible if `getElementFromPoint` at its center matches the element
- the element is visible if it has a visible rect

With the current approach we fail to detect the tap targets in the [GoDaddy example](https://gist.github.com/mattzeunert/7e1d8f3bfd2ce7aa29ab1b4e6790966c). That's because the center of the largest rect is invisible, so we decide that the whole element must be invisible. By moving to the new approach we fix the GoDaddy problem.

I was also thinking we might be able to get rid of `filterClientRectsWithinAncestorsVisibleScrollArea`, but if we just check center of each react we still risk false failures caused by partially invisible rects inside `overflow: hidden` containers. So maybe we should test all four corners with `getElementFromPoint`?

Let me know if you have any thoughts. I think I need to do some more experimentation, and also see what the performance cost of `elementFromPoint` and `scrollTo` is.

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: We don't 🙂

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: I've changed it to do the z index test on each client rect now.

I don't think we need `filterClientRectsWithinAncestorsVisibleScrollArea` any more, because the z index check covers most of those cases. The only difference is that z index only checks the center of the tap target whereas before we were checking that the whole rect is visible.

Stats for this [wiki page](https://en.m.wikipedia.org/wiki/List_of_2017_albums):

|    | master | this branch |
| - | ------- | ----------- |
| Gathered tap targets | 2018 |  2091 |
| Run time | 405ms | 497ms |

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Can't find it anymore, but that code's gone now anyway.

lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
mattzeunert: Added two more test cases in there.

lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
```diff
@@ -85,29 +95,81 @@ <h1>SEO Tap targets</h1>
 
     <br/><br/>
 
-    <div style="position: relative">
-      <a style="display: block; position: absolute; top:0; height: 40px; width: 40px;">inner</a>
-      <a style="display: block; height: 100px; width: 100px; background: yellowgreen">outer</a>
+    <!-- The left tap target has a large client rect that would normally overlap with the right tap target,
```
patrickhulce: 😍 LOVIN' these tests :D

lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -190,46 +201,56 @@ module.exports = [
       audits: {
         'tap-targets': {
           score: (() => {
-            const PASSING_TAP_TARGETS = 11;
-            const TOTAL_TAP_TARGETS = 12;
+            const totalTapTargets = expectedGatheredTapTargets.length;
+            const passingTapTargets = totalTapTargets - expectedFailingTapTargetsMatch.length;
             const SCORE_FACTOR = 0.89;
-            return Math.floor(PASSING_TAP_TARGETS / TOTAL_TAP_TARGETS * SCORE_FACTOR * 100) / 100;
+            return Math.round(passingTapTargets / totalTapTargets * SCORE_FACTOR * 100) / 100;
```
patrickhulce: a tad worried by how much logic ends up living in our expectations, but I sympathize with the tedium of updating these and would rather have it be more precise so 👍

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -257,10 +230,18 @@ function gatherTapTargets() {
   /** @type {LH.Artifacts.TapTarget[]} */
   const targets = [];
 
+  // Capture element positions relative to the top of the page
+  window.scrollTo(0, 0);
```
patrickhulce: aside, I wonder if we should do this after every gatherer. I recently got burned by it not starting at 0,0 in the images one

lighthouse-cli/test/smokehouse/seo/expectations.js
brendankenny: > a tad worried by how much logic ends up living in our expectations, but I sympathize with the tedium of updating these and would rather have it be more precise so 👍

I think this might be going too far in this direction :/ It's one thing when it's broken down but still defined by constants, but since this is pulling live from the test file, it would be difficult to notice e.g. if we broke one of the overlapping tap target examples in the test page. So long as it also no longer matched the regex it would still pass and we wouldn't notice we were losing coverage.

It's also nice to have a place we can point to with "here's what we expect a failing/real value for this audit to look like", and with this approach the only way to do that is to put in a breakpoint and open up a debugger.

lighthouse-core/lib/rect-helpers.js
```diff
@@ -68,6 +63,7 @@ function filterOutRectsContainedByOthers(rects) {
 /**
  * @param {LH.Artifacts.Rect} rect
  */
+/* istanbul ignore next */
```
brendankenny: doesn't this get coverage info from use in `tappable-rects.js`, so don't need to ignore?

lighthouse-core/lib/rect-helpers.js
```diff
@@ -208,13 +204,15 @@ function getRectAtCenter(rect, centerRectSize) {
...
```
brendankenny: same with this and `getLargestRect`, which are called from the `tap-targets.js` audit?

lighthouse-core/gather/gatherers/seo/tap-targets.js
brendankenny: one issue is what if the page purposely scrolled down? Seems similar to the axe gathering, e.g. if we took a screenshot (for some reason) in a gatherer after this one, as a user I'd expect it to be the targeted final look of the page, not scrolled back to the top. Does this gatherer need to be last (or next to last) if it wants to change  scroll position?

(of course there are lots of issues with actually supporting this that we don't account for in e.g. offscreen-images)

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -249,6 +180,48 @@ function truncate(str, maxLength) {
...
+  const topEl = document.elementFromPoint(
+    elCenterPoint.x,
+    elCenterPoint.y - window.scrollY
+  );
+
+  return topEl === el || el.contains(topEl);
+}
+
+/* istanbul ignore next */
+function disableFixedAndStickyElementPointerEvents() {
```
brendankenny: need types and explanation jsdocs on this

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -249,6 +180,48 @@ function truncate(str, maxLength) {
...
+  const className = 'lighthouse-disable-pointer-events';
+  const styleTag = document.createElement('style');
+  styleTag.innerHTML = `.${className} { pointer-events: none !important }`;
```
brendankenny:
```suggestion
  styleTag.textContent = `.${className} { pointer-events: none !important }`;
```

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -249,6 +180,48 @@ function truncate(str, maxLength) {
...
+  document.body.appendChild(styleTag);
+
+  Array.from(document.querySelectorAll('*')).forEach(el => {
```
brendankenny: can just do
```suggestion
  document.querySelectorAll('*').forEach(el => {
```
since this code only ever runs in recent Chrome

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -273,30 +254,66 @@ function gatherTapTargets() {
...
+  const tapTargetsWithVisibleClientRects = [];
+  // We use separate loop here to get visible client rects because that involves
+  // scrolling around the page for elementCenterIsAtZAxisTop, which would affect the
+  // client rect positions.
+  tapTargetsWithClientRects.forEach(({tapTargetElement, clientRects}) => {
+    // Filter out empty client rects
+    let visibleClientRects = clientRects.filter(cr => cr.width !== 0 && cr.height !== 0);
+
+    // Filter out client rects that are invisible, e.g because they are in a position absolute element
+    // with a lower z-index than the main contnet
```
brendankenny:
```suggestion
    // with a lower z-index than the main content.
```

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: hm I wonder why it's not stacking our convos...

anyhoo :)

I disagree we need to be worried about always returning the scroll position to the exact same spot as before. I think it's a far more reasonable contract that every gatherer should scroll to where it needs to be and not assume anything about scroll position and we reset scroll position to 0,0 between every gatherer. It doesn't preclude anyone from taking screenshots of a particular scrolled area with the protocol unless I'm misunderstanding your objection?

brendankenny: > It doesn't preclude anyone from taking screenshots of a particular scrolled area with the protocol unless I'm misunderstanding your objection?

FWIW, I meant if the page scrolled itself (e.g. it's an SPA with a router that scrolls to a particular part of the page based on whatever), but I'm not sure if that changes much about your point

lighthouse-core/gather/gatherers/seo/tap-targets.js
mattzeunert: Oh, TIL. 👍

lighthouse-core/lib/rect-helpers.js
mattzeunert: I think nyc renames the functions when doing the code coverage test. So we disable it on all functions that are injected into the gatherer with `.toString()`.

mattzeunert:
> to sidestep the need for switching on/off pointer-events, did you consider document.elementsFromPoint()? Since the returned array is in paint order, seems like it would be possible to shift off any starting elements that are fixed or sticky?

No I didn't consider it, that's a good idea!

I tried it now and `elementsFP` is a lot slower than `elementFP`. (200x slower in a simple loop for the same position over and over.) The gatherer as a whole is 2.5x slower on the List of 2017 albums.

lighthouse-core/gather/gatherers/seo/tap-targets.js
patrickhulce: > it's an SPA with a router that scrolls to a particular part of the page based on whatever

Ah, gotcha. Well I think my point remains unchanged but replace "we reset scroll position to 0,0 between every gatherer" with "we reset scroll position to endOfPassX,endOfPassY between every gatherer" :)

lighthouse-core/lib/rect-helpers.js
brendankenny: > I think nyc renames the functions when doing the code coverage test

ah. That's really annoying :)

we really should switch to native node coverage if we can...

paulirish:
@brendankenny want another pass or is this g2g?

PR #7785, opened 2019-03-28
misc(gather): update comment re: ClientRect copying

paulirish:
the comment is out of date.

getBoundingClientRect  used to return a ClientRect which stringified to {}
but [as of july 2017](https://bugs.chromium.org/p/chromium/issues/detail?id=719246#c20), it returns a DOMRect, which stringifies fine.

Still, we dont want all the props, so we'll keep the basic functionality here.

this change is so minor. sorry for the noise.
credit to @brendankenny who brought this up


brendankenny:
> credit to @brendankenny who brought this up

actually credit to @thebengeu who found the JSONify comment and noticed it didn't actually act like that (anymore): https://github.com/googleads/ad-speed-insights/pull/14#discussion_r270190264 :)

PR #7894, opened 2019-04-02
misc: simplifications in simulator/connection-pool

brendankenny:
another one I accumulated while stepping through the simulator. Feel free to reject anything @patrickhulce :)

1) remove `records` from `LH.Artifacts.NetworkAnalysis`. The caller of the computed artifact has to already have them to get the `NetworkAnalysis`, and we never use them on any instance of the type anyways.
2) early return in `PageDependencyGraph.linkNetworkNodes()` for the common case where there were no redirects of a particular record
3) split `ConnectionPool.acquire()` into the case where a connection is needed to fetch a record and the case where a connection is already fetching a record.

   This seems ok as a place where the generality hasn't proven necessary...we only call `acquire()` in three places, and the two done that return assigned connections already explicitly call out with a comment and a type cast that it knows a connection already exists for this record.
4) removed `NodeTimingIntermediate.queuedTime` as it isn't used for anything. I can add back if this was useful for debugging or something.

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
@@ -139,14 +139,10 @@ module.exports = class ConnectionPool {
    * @return {?TcpConnection}
    */
   acquire(record, options = {}) {
-    if (this._connectionsByRecord.has(record)) {
-      // @ts-ignore
-      return this._connectionsByRecord.get(record);
-    }
+    if (this._connectionsByRecord.has(record)) throw new Error('Record already has a connection');
 
-    const origin = String(record.parsedURL.securityOrigin);
+    const origin = record.parsedURL.securityOrigin;
     const observedConnectionWasReused = !!this._connectionReusedByRequestId.get(record.requestId);
-    /** @type {TcpConnection[]} */
     const connections = this._connectionsByOrigin.get(origin) || [];
     const connectionToUse = this._findAvailableConnectionWithLargestCongestionWindow(connections, {
       ignoreConnectionReused: options.ignoreConnectionReused,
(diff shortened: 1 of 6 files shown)


lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
```diff
@@ -139,14 +139,10 @@ module.exports = class ConnectionPool {
...
-    if (this._connectionsByRecord.has(record)) {
-      // @ts-ignore
-      return this._connectionsByRecord.get(record);
-    }
+    if (this._connectionsByRecord.has(record)) throw new Error('Record already has a connection');
 
-    const origin = String(record.parsedURL.securityOrigin);
+    const origin = record.parsedURL.securityOrigin;
     const observedConnectionWasReused = !!this._connectionReusedByRequestId.get(record.requestId);
-    /** @type {TcpConnection[]} */
```
patrickhulce: this isn't necessary anymore?

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -141,9 +141,8 @@ class Simulator {
 
   /**
    * @param {Node} node
-   * @param {number} queuedTime
    */
-  _markNodeAsReadyToStart(node, queuedTime) {
```
patrickhulce: I'd prefer to keep this, it's pretty helpful when stepping through the debugger

lighthouse-core/lib/dependency-graph/simulator/connection-pool.js
brendankenny: > this isn't necessary anymore?

no, its inferred from the first part of the init and the `_findAvailableConnectionWithLargestCongestionWindow` callsite

lighthouse-core/lib/dependency-graph/simulator/simulator.js
brendankenny: > I'd prefer to keep this, it's pretty helpful when stepping through the debugger

gotcha :)

lighthouse-core/lib/dependency-graph/simulator/simulator.js
```diff
@@ -503,7 +501,7 @@ module.exports = Simulator;
  * @typedef NodeTimingIntermediate
  * @property {number} [startTime]
  * @property {number} [endTime]
- * @property {number} [queuedTime]
+ * @property {number} [queuedTime] Helpful for debugging.
```
patrickhulce: 👍

Issue #7882, opened 2019-04-01
Feature request: support gatherers in plugin


paullewis:
**Feature request summary**
The plugin supports custom audits, but AFAICT not custom gatherers. It would be great if we were able to append gatherers in the same way as audits.

**What is the motivation or use case for changing this?**
I have a plugin that requires artifacts not currently captured by the existing gatherers. Specifically pulling the contents of scripts that have non-standard types, e.g. `<script type="foo" src="bar.something">` since these won't get requested or parsed and I need their contents.

**How is this beneficial to Lighthouse?**
Moar powerful for plugin makers 👍

patrickhulce:
Heya Paul :)

It was an intentional decision to keep plugins from getting involved in gathering, but I'm struggling to find the document with the discussion since #6070 is *quite* sparse. IIRC, it essentially boiled down to security/privacy concerns, along with the fact that gathering can be significantly more difficult to do correctly than auditing and we'd probably want to be involved if there's an important fact about the page we don't have right now.

For your use case, it seems reasonable for us to collect the URLs and types of all the script elements  on the page even if we don't request the content of ones that weren't loaded. The plugin audit could still fetch the ones that are missing even if it violates the purity of a gather and audit step :)

paullewis:
Yeah, that could work! Since it's a non-core plugin I can theoretically fetch the file in the audit. A little gross, but not the worst thing either. I can totally see why you'd be cautious of adding gatherers to plugins more generally.

Okay, so let's say the Scripts artifact. It currently only returns inline scripts or those that have triggered a network request, and my particular one won't do either, as you rightly noted. If we reduce this request to including non-fetched scripts as well, does that seem reasonable to you?

brendankenny:
Going to close this issue. The tension between limited/simple plugins vs capable/harder-to-support plugins is still there, but we can focus on specific requests and don't need an umbrella tracking issue. For now we still recommend not using plugins if a custom gatherer is needed and to either request the gatherer in core or use a custom config where you can add new gatherers all day.

Maintainably extending Lighthouse is still a wide-open area, so anyone coming across this old issue, please feel free to open new requests.

PR #7920, opened 2019-04-02
core(scripts): move to ScriptElements artifact

patrickhulce:
**Summary**
Moves `Scripts` artifact to `ScriptElements` and includes information about *all* script elements on the page.

**Related Issues/PRs**
#6747 #7882


brendankenny:
I *think* #7882 is going to be satisfied by structured data parsing. I can see the argument in going general like the other `*Elements`, but are we making this harder to use for not much benefit?

e.g. doesn't `unminified-javascript.js` need to check the `type` now? Otherwise people doing inline shaders or whatever are going to get that showing up in that audit's results. That's going to be difficult to remember to always in audits, especially since on *most* websites js scripts are all you'll usually find.

Basically the more general we get the more widely useful these get (like if you want to write a `unminified-glsl.js` audit), but we end up passing more corner cases and trickiness onto users of artifacts instead of the gatherer taking the hit.

brendankenny:
Maybe it's inevitable, like if you want to give warnings on invalid `*Elements`, you have to have those in the artifact.

We could make help structurally in the artifact shape, like dividing the top level `ScriptElements` into sub objects of js scripts and all-the-others scripts (or something like that)

patrickhulce:
> doesn't unminified-javascript.js need to check the type now?

Not AFAICT. We never checked the type before, just if it had content and didn't have a src, which is the same as what's happening now. Also, the type is almost always `null` most scripts in my experience do not specify a type nor do they need to.

This change shouldn't meaningfully impact the breadth of what we collect. It primarily impacts the properties and amount of data we provide about each script. If there are scripts with a src that weren't fetched for some reason, then those are new, but they won't have content so there's not much I can imagine what might try to do with them.

patrickhulce:
> We could make help structurally in the artifact shape, like dividing the top level ScriptElements into sub objects of js scripts and all-the-others scripts (or something like that)

I definitely like the idea of helping consumers, but IMO the benefit of consistentcy of `*Element*` artifacts isn't worth splitting the structure. I'd prefer to address this concern through exposed helper methods or additional flags.

brendankenny:
> We never checked the type before, just if it had content and didn't have a src, which is the same as what's happening now.

Oh, that's a good point. Inline glsl output definitely ends up in `unminified-javascript` output right now

> Also, the type is almost always null most scripts in my experience do not specify a type nor do they need to.

but *we* need to know what the type was :)

> I definitely like the idea of helping consumers, but IMO the benefit of consistency of *Element* artifacts isn't worth splitting the structure. I'd prefer to address this concern through exposed helper methods or additional flags.

Doing something like making the structure of the data into documentation of the possible pitfalls means you can't forget it (like a `type` bug left unnoticed for 1.25 years :). I don't know if anyone will go looking for a helper method if what we have here is "obviously" just an array of all the javascript elements in the page

patrickhulce:
> like a type bug left unnoticed for 1.25 years

if you're referring to the inline scripts here, you'll be glad to learn this has only existed for 28 days, so it's not that old :)

I agree that it will be easy to overlook and worth some of our attention to document/highlight/etc, but seeing as there's only a single usage of this artifact currently and who knows how it will be consumed, I'm just saying I'm not sure it's worth crafting bespoke artifact shapes for each `*Elements` artifact because we're concerned people will forget to filter for what they really were looking for.

patrickhulce:
@Hoten any thoughts here since you were just touching `Scripts`?

Hoten:
> Oh, that's a good point. Inline glsl output definitely ends up in unminified-javascript output right now

this was surprising to me. In order to fix that, we would need the type... good thing this PR adds it to the artifact! let's follow up with an explicit check for JS scripts in that audit.

I do see most use cases of `ScriptElements` wanting to only touch javascripts. plugins that use that artifact would easily overlook that (who thinks to test inline glsl?!). can we provide a function that just returns JS scripts elements in ScriptElements? where would that go? maybe via a computed artifact `JavscriptElements`?

patrickhulce:
> can we provide a function that just returns JS scripts elements in ScriptElements? where would that go? maybe via a computed artifact JavscriptElements?

IMO, we should just normalize the type in the artifact in the followup that fixes `unminified-javascript` so that consuming the type is much easier and we/others can check `script.normalizedType === 'javascript'` or whatever :)

I think doing anything beyond that puts the cart before the horse a bit. Are people writing plugins that consume script element contents mostly going to target javascript? I'm not convinced the answer is yes :) The most obvious first plugins to me are actually targeting non-standard script types like json-ld or framework-specific template content. I think this highlights more that we need robust guides for consuming artifacts to help folks think through what they intend to flag, so they don't write the same bugs we do 😆 


I'm going to go ahead and merge this as a step forward, and we can always discuss the ideal way to guide plugin author usage of our artifacts in the future. @Hoten did you say that to mean you wanted to take the followup or shall I?

Hoten:
sounds good

>  @Hoten did you say that to mean you wanted to take the followup or shall I?

go for it!

PR #7962, opened 2019-04-03
misc: specify types for some dependency-graph options objects

brendankenny:
No issues, just for documentation and ease of cross referencing.

Where the only caller(s) passed in definitely defined arguments, also made the properties required (I think this only happened for the `ConnectionPool` and `DNSCache` constructors).

### DIFF
--- a/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
+++ b/lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
@@ -260,14 +260,14 @@ class NetworkAnalyzer {
    * available in the records themselves appears untrustworthy.
    *
    * @param {LH.Artifacts.NetworkRequest[]} records
-   * @param {object} [options]
+   * @param {{forceCoarseEstimates: boolean}} [options]
    * @return {Map<string, boolean>}
    */
   static estimateIfConnectionWasReused(records, options) {
-    options = Object.assign({forceCoarseEstimates: false}, options);
+    const {forceCoarseEstimates = false} = options || {};
 
     // Check if we can trust the connection information coming from the protocol
-    if (!options.forceCoarseEstimates && NetworkAnalyzer.canTrustConnectionInformation(records)) {
+    if (!forceCoarseEstimates && NetworkAnalyzer.canTrustConnectionInformation(records)) {
       // @ts-ignore
       return new Map(records.map(record => [record.requestId, !!record.connectionReused]));
     }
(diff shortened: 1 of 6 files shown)


lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
```diff
@@ -474,3 +462,12 @@ module.exports = NetworkAnalyzer;
  * @property {number} avg
  * @property {number} median
  */
+
+/**
+ * @typedef NetworkAnalyzer.RTTEstimateOptions
+ * @property {boolean} [forceCoarseEstimates] TCP connection handshake information will be used when available, but for testing it's useful to see how the coarse estimates compare with higher fidelity data.
```
brendankenny: I don't believe the emphasis on "for testing it's useful..." is entirely true anymore (`ConnectionPool` calls `NetworkAnalyzer.estimateIfConnectionWasReused` with `forceCoarseEstimates: true`), but I don't know enough to update this :)

lighthouse-core/test/lib/dependency-graph/simulator/dns-cache-test.js
```diff
@@ -61,4 +61,12 @@ describe('DependencyGraph/Simulator/DNSCache', () => {
       expect(dns.getTimeUntilResolution(request, {requestedAt: 2000})).toBe(0);
     });
   });
+
+  describe('.setResolvedAt', () => {
```
brendankenny: coverage low hanging fruit :)

lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
```diff
@@ -303,52 +303,44 @@ class NetworkAnalyzer {
...
-        // useful for testing to isolate the different methods of estimation
-        useDownloadEstimates: true,
-        useSendStartEstimates: true,
-        useHeadersEndEstimates: true,
-      },
-      options
-    );
+    const {
+      forceCoarseEstimates = false,
+      coarseEstimateMultiplier = 0.3,
```
patrickhulce: FWIW, I think we lose a little value by moving the explanation away from the default on this one

lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
patrickhulce: Oh yeah we needed to turn it on over there because the connection id information just lied so often in certain enviornments so it was just easier to always ignore it.

Probably worth rewording to something like

`TCP connection handshake information will be used when available, but in some circumstances this data can be unreliable. This flag exposes an option to ignore the handshake data and use the coarse download/TTFB timing data.`

?

lighthouse-core/lib/dependency-graph/simulator/network-analyzer.js
brendankenny: > FWIW, I think we lose a little value by moving the explanation away from the default on this one

Yeah, sounds good. We should maybe move it to a top-level constant (with the comment), but just moved the comment back into place for now

PR #7991, opened 2019-04-05
misc: remove lhr-lite.d.ts

### LINKED ISSUE #4867: Resolve LHR differences between master/LHR-Lite/external service
I think we have 3 API shapes in play atm. Perhaps best way to work through the differences is a 3-column doc working through the LHR schema?  I believe we now have all 3 defined in .d.ts. 

We'll then need to: 
* Define the differences
* decide on their resolution
* get review from dgozman/caseq
* submit a CL to external service to discuss changes with API owners

brendankenny:
fixes #4867

all our output has been unified around the full [LHR](https://github.com/GoogleChrome/lighthouse/blob/7a431253ff19397d4b8db9ccdf599e93f7fe40c6/types/lhr.d.ts) (and/or a subset(ish) of it via the proto), so no need for `lhr-lite.d.ts` anymore.

This wasn't a comprehensive adding of strings to all the properties currently missing them, just copying over anything in `lhr-lite.d.ts` that wasn't currently in any of the type files we do use.


types/audit.d.ts
```diff
@@ -64,23 +70,31 @@ declare global {
     export interface Product {
       rawValue: boolean | number | null;
       displayValue?: string;
+      /** An explanation of audit-related issues encountered on the test page. */
       explanation?: string;
+      /** Error message from any exception thrown while running this audit. */
       errorMessage?: string;
       warnings?: string[];
+      /** The scored value determined by the audit, in the range `0-1`, or null if `scoreDisplayMode` indicates not scored. */
```
paulirish:
```suggestion
      /** The scored value of the audit, provided in the range `0-1`. */
```

types/audit.d.ts
```diff
@@ -64,23 +70,31 @@ declare global {
...
+    /* Audit result returned in Lighthouse report. All audits offer a description and score of 0-1. */
     export interface Result {
       rawValue: boolean | number | null;
       displayValue?: string;
+      /** An explanation of audit-related issues encountered on the test page. */
       explanation?: string;
+      /** Error message from any exception thrown while running this audit. */
       errorMessage?: string;
       warnings?: string[];
+      /** The scored value determined by the audit, in the range `0-1`, or null if `scoreDisplayMode` indicates not scored. */
```
paulirish:
```suggestion
      /** The scored value of the audit, provided in the range `0-1`. */
```

types/audit.d.ts
```diff
@@ -64,23 +70,31 @@ declare global {
...
+      /** Extra information provided by some types of audits. */
       details?: Audit.Details;
     }
 
-    /* Audit result returned in Lighthouse report. All audits offer a description and score of 0-1 */
+    /* Audit result returned in Lighthouse report. All audits offer a description and score of 0-1. */
     export interface Result {
       rawValue: boolean | number | null;
       displayValue?: string;
+      /** An explanation of audit-related issues encountered on the test page. */
```
patrickhulce:
```suggestion
      /** An explanation of why the audit failed on the test page. */
```

types/audit.d.ts
```diff
@@ -64,23 +70,31 @@ declare global {
...
+      /** Error message from any exception thrown while running this audit. */
       errorMessage?: string;
       warnings?: string[];
+      /** The scored value determined by the audit, in the range `0-1`, or null if `scoreDisplayMode` indicates not scored. */
       score?: number;
+      /** Deprecated and does not make its way into the Lighthouse report. */
       extendedInfo?: {[p: string]: any};
       /** Overrides scoreDisplayMode with notApplicable if set to true */
       notApplicable?: boolean;
+      /** Extra information provided by some types of audits. */
```
patrickhulce: seems about as descriptive as `details` ;)

do we want to talk about how this is structured data for the renderer?

types/audit.d.ts
```diff
@@ -64,23 +70,31 @@ declare global {
     export interface Product {
```
patrickhulce: somewhat unrelated, but instead of duplicating should this `Pick<>` + extend from `Result`?

brendankenny: > somewhat unrelated, but instead of duplicating should this `Pick<>` + extend from `Result`?

yeah, I was thinking about the same thing while doing this, I think it's worth trying to make it work.

types/audit.d.ts
brendankenny: > seems about as descriptive as `details` ;)

Fair :)

> do we want to talk about how this is structured data for the renderer?

 I took a two second stab at it :)

PR #8044, opened 2019-04-07
tests: support artifact assertions in smoke tests

mattzeunert:
**Summary**

We'd [like to](https://github.com/GoogleChrome/lighthouse/pull/7778#discussion_r271847241) make assertions against the artifacts in the smoke tests.

- Wrap existing expectations in `lhr` top level object
- Add support for having `artifacts` expectations in addition to `lhr`

### DIFF
--- a/lighthouse-cli/test/smokehouse/smokehouse-report.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse-report.js
@@ -103,12 +103,12 @@ function findDifference(path, actual, expected) {
 }
 
 /**
- * @param {string} name
+ * @param {string} name – description of what's being asserted on (e.g. the result of a certain audit)
  * @param {any} actualResult
  * @param {any} expectedResult
- * @returns {Smokehouse.Comparison}
+ * @return {Smokehouse.Comparison}
  */
-function makeAssertion(name, actualResult, expectedResult) {
+function makeComparison(name, actualResult, expectedResult) {
   const diff = findDifference(name, actualResult, expectedResult);
 
   return {
@@ -121,7 +121,7 @@ function makeAssertion(name, actualResult, expectedResult) {
 }
 
 /**
- * Collate results into comparisons of actual and expected scores on each audit.
+ * Collate results into comparisons of actual and expected scores on each audit/artifact.
  * @param {Smokehouse.ExpectedRunResult} actual
  * @param {Smokehouse.ExpectedRunResult} expected
  * @return {Smokehouse.Comparison[]}
@@ -138,23 +138,21 @@ function collateResults(actual, expected) {
       }
 
       const expectedResult = (expected.artifacts || {})[artifactName];
-      return makeAssertion(artifactName + ' artifact', actualResult, expectedResult);
+      return makeComparison(artifactName + ' artifact', actualResult, expectedResult);
     });
   }
 
   /** @type {Smokehouse.Comparison[]} */
   let auditAssertions = [];
-  if (expected.lhr.audits) {
-    auditAssertions = Object.keys(expected.lhr.audits).map(auditName => {
-      const actualResult = actual.lhr.audits[auditName];
-      if (!actualResult) {
-        throw new Error(`Config did not trigger run of expected audit ${auditName}`);
-      }
+  auditAssertions = Object.keys(expected.lhr.audits).map(auditName => {
+    const actualResult = actual.lhr.audits[auditName];
+    if (!actualResult) {
+      throw new Error(`Config did not trigger run of expected audit ${auditName}`);
+    }
 
-      const expectedResult = expected.lhr.audits[auditName];
-      return makeAssertion(auditName + ' audit', actualResult, expectedResult);
-    });
-  }
+    const expectedResult = expected.lhr.audits[auditName];
+    return makeComparison(auditName + ' audit', actualResult, expectedResult);
+  });
 
   return [
     {
--- a/lighthouse-cli/test/smokehouse/smokehouse-report.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse-report.js
@@ -130,14 +130,15 @@ function collateResults(actual, expected) {
   /** @type {Smokehouse.Comparison[]} */
   let artifactAssertions = [];
   if (expected.artifacts) {
-    const artifactNames = /** @type {(keyof LH.Artifacts)[]} */ (Object.keys(expected.artifacts));
+    const expectedArtifacts = expected.artifacts;
+    const artifactNames = /** @type {(keyof LH.Artifacts)[]} */ (Object.keys(expectedArtifacts));
     artifactAssertions = artifactNames.map(artifactName => {
       const actualResult = (actual.artifacts || {})[artifactName];
       if (!actualResult) {
         throw new Error(`Config run did not generate artifact ${artifactName}`);
       }
 
-      const expectedResult = (expected.artifacts || {})[artifactName];
+      const expectedResult = expectedArtifacts[artifactName];
       return makeComparison(artifactName + ' artifact', actualResult, expectedResult);
     });
   }
--- a/lighthouse-cli/test/smokehouse/smokehouse-report.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse-report.js
@@ -103,7 +103,7 @@ function findDifference(path, actual, expected) {
 }
 
 /**
- * @param {string} name – description of what's being asserted on (e.g. the result of a certain audit)
+ * @param {string} name – name of the value being asserted on (e.g. the result of a certain audit)
  * @param {any} actualResult
  * @param {any} expectedResult
  * @return {Smokehouse.Comparison}
@@ -112,7 +112,7 @@ function makeComparison(name, actualResult, expectedResult) {
   const diff = findDifference(name, actualResult, expectedResult);
 
   return {
-    category: name,
+    name,
     actual: actualResult,
     expected: expectedResult,
     equal: !diff,
@@ -156,13 +156,13 @@ function collateResults(actual, expected) {
 
   return [
     {
-      category: 'error code',
+      name: 'error code',
       actual: actual.errorCode,
       expected: expected.errorCode,
       equal: actual.errorCode === expected.errorCode,
     },
     {
-      category: 'final url',
+      name: 'final url',
       actual: actual.lhr.finalUrl,
       expected: expected.lhr.finalUrl,
       equal: actual.lhr.finalUrl === expected.lhr.finalUrl,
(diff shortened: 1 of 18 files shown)


lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,45 +102,74 @@ function findDifference(path, actual, expected) {
...
-  const collatedAudits = auditNames.map(auditName => {
-    const actualResult = actual.audits[auditName];
-    if (!actualResult) {
-      throw new Error(`Config did not trigger run of expected audit ${auditName}`);
-    }
+  /** @type {Smokehouse.Assertion[]} */
+  const artifactAssertions = [];
+  if (expected.artifacts) {
+    Object.keys(expected.artifacts).map(artifactName => {
+      const actualResult = /** @type {any} */ (actual.artifacts)[artifactName];
```
patrickhulce: can we type these as partial artifacts? I guess it doesn't help much since we don't type check our assertions :)

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,45 +102,74 @@ function findDifference(path, actual, expected) {
...
 
   return {
-    audits: collatedAudits,
+    assertions,
     errorCode: {
       category: 'error code',
-      actual: actual.errorCode,
-      expected: expected.errorCode,
-      equal: actual.errorCode === expected.errorCode,
+      actual: actual.lhr.errorCode,
```
patrickhulce: now that we are fixing things, this isn't really in the LHR, so makes sense to move out at top level

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -60,9 +60,9 @@ function runLighthouse(url, configPath, isDebug) {
     '--port=0',
   ];
 
-  if (isDebug) {
-    args.push('-GA');
-  }
+  // Save artifacts
+  args.push('-GA');
```
patrickhulce: we should specify a tmp path if we're *always* doing this otherwise every smoke run will overwrite the users saved artifacts

lighthouse-cli/test/smokehouse/smokehouse-report.js
mattzeunert: Do artifacts have a type other than the union type of all artifact values? Is that what you were thinking of?

mattzeunert: Oh I get it now, we can type the artifacts key instead of it being string and then the artifact value will be the union type.

patrickhulce: Awesome yeah LGTM :)

patrickhulce:
For other reviewers [?w=1](https://github.com/GoogleChrome/lighthouse/pull/8044/files?w=1) was a lifesaver here, it's really only ~100 or so line change without the whitespace diffs :)

types/smokehouse.d.ts
```diff
@@ -19,14 +19,28 @@
...
+    artifacts?: Partial<LH.Artifacts>
+  }
+
+  export interface RunComparison {
+    assertions: Comparison[];
     errorCode: Comparison;
     finalUrl: Comparison;
   }
 
+  export interface Assertion{
```
patrickhulce: nit: space after `Assertion`

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,34 +102,64 @@ function findDifference(path, actual, expected) {
...
+      const expectedResult = expected.lhr.audits[auditName];
+      return makeAssertion(auditName + ' audit', actualResult, expectedResult);
+    });
+  }
+
+  const assertions = [...artifactAssertions, ...auditAssertions];
 
   return {
-    audits: collatedAudits,
+    assertions,
```
patrickhulce: a bit strange that we separate out the others but mix these into `assertions` IMO, I might prefer a separate artifacts property or just collapsing everything into a comparison array

but not a huge deal

types/smokehouse.d.ts
mattzeunert: Whoops, realized this is the same as `Smokehouse.Comparison`.

lighthouse-cli/test/smokehouse/smokehouse-report.js
mattzeunert: Merged to just return a Comparison array.

mattzeunert:
@brendankenny What do you think about this?

brendankenny:
I only got to take a cursory read yesterday with the 4.3 stuff going on, but the concept and implementation were looking great to me. I'll take a closer look later today.

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,47 +102,76 @@ function findDifference(path, actual, expected) {
   return null;
 }
 
+/**
+ * @param {string} name
+ * @param {any} actualResult
+ * @param {any} expectedResult
+ * @returns {Smokehouse.Comparison}
+ */
+function makeAssertion(name, actualResult, expectedResult) {
```
brendankenny: maybe call this `makeComparison` now?

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,47 +102,76 @@ function findDifference(path, actual, expected) {
...
+    category: name,
+    actual: actualResult,
+    expected: expectedResult,
+    equal: !diff,
+    diff,
+  };
+}
+
 /**
  * Collate results into comparisons of actual and expected scores on each audit.
```
brendankenny: should update the "on each audit" part

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,47 +102,76 @@ function findDifference(path, actual, expected) {
...
-      category: auditName,
-      actual: actualResult,
-      expected: expectedResult,
-      equal: !diff,
-      diff,
-    };
-  });
+  /** @type {Smokehouse.Comparison[]} */
+  let auditAssertions = [];
+  if (expected.lhr.audits) {
```
brendankenny: I think we require `audits` to be defined, fwiw

brendankenny:
I'd like to go all the way and use the real types for the `LH.RunnerResult` (using a recursive `Partial<>` except for the few things which are in the `Pick<>` type today), which after this PR would mostly just need testing `lhr.runtimeError` instead of our constructed `errorCode` (though we could still test that too). We could do that in another PR if others are ok with that.

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -101,22 +101,29 @@ function runLighthouse(url, configPath, isDebug) {
...
   }
 
-  const lhr = fs.readFileSync(outputPath, 'utf8');
-  if (isDebug) {
-    console.log('LHR output available at: ', outputPath);
-  } else if (fs.existsSync(outputPath)) {
-    fs.unlinkSync(outputPath);
-  }
+  const artifacts = JSON.parse(
+    fs.readFileSync(`${artifactsDirectory}/artifacts.json`).toString());
```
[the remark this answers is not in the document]
mattzeunert: Wow, I did not know you could do that!

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -102,47 +102,76 @@ function findDifference(path, actual, expected) {
   return null;
 }
 
+/**
+ * @param {string} name
```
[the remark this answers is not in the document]
mattzeunert: I've changed `category` to `name` now. We also use it for `difference at TapTargets artifact.length`, so it's more of an object name than a custom error message you'd have for an assertion. If we compare the whole runner result we could just use the key path here maybe.

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -103,46 +103,73 @@ function findDifference(path, actual, expected) {
...
+  let artifactAssertions = [];
+  if (expected.artifacts) {
+    const artifactNames = /** @type {(keyof LH.Artifacts)[]} */ (Object.keys(expected.artifacts));
+    artifactAssertions = artifactNames.map(artifactName => {
+      const actualResult = (actual.artifacts || {})[artifactName];
+      if (!actualResult) {
+        throw new Error(`Config run did not generate artifact ${artifactName}`);
+      }
+
+      const expectedResult = (expected.artifacts || {})[artifactName];
```
patrickhulce: should be able to nuke this `||` given the if on line 132

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -199,22 +226,22 @@ function reportAssertion(assertion) {
...
  */
-function report(results) {
+function report(comparisons) {
   let correctCount = 0;
   let failedCount = 0;
 
-  [results.finalUrl, results.errorCode, ...results.audits].forEach(auditAssertion => {
-    if (auditAssertion.equal) {
+  comparisons.forEach(assertion => {
+    if (assertion.equal) {
```
patrickhulce: feels like we've still got quite a bit of assertion/comparison mixing going on throughout but that was preexisting :)

mattzeunert: It feels like it should always be comparison? Assertion seems more narrow in scope to me than what we have here.

lighthouse-cli/test/smokehouse/smokehouse-report.js
mattzeunert: Hmm, TS fails without the `|| {}` if the artifacts are accessed in the `map` callback. In general constraining the type seems to work in nested functions, but not here somehow...

lighthouse-cli/test/smokehouse/smokehouse-report.js
patrickhulce: agreed :) cleanup for another time

lighthouse-cli/test/smokehouse/smokehouse-report.js
patrickhulce: ah right the TS callback problem, my preferred way around this recently has been to assign `expected.artifacts` to a variable right after we've asserted its definedness, then when you access it in the closure TS already knows it's safe

brendankenny:
I really don't understand the lint issue here. They aren't inconsistently quoted, and I can't get the failure in a completely fresh download and install of the lighthouse repo

brendankenny:
welllll, I'm going to override travis again, assuming that when this is merged back into master everything will be happy. We'll see how this goes for the second time this week :)

PR #8074, opened 2019-04-08
tests: generate devtoolsLogs for network-request-test.js tests

brendankenny:
@exterkamp (when merged it'll become part of #7888) 

tests are mostly of the form:
- use `networkRecordsToDevtoolsLog` to generate devtools logs from network records to be tested
- turn on (or don't) `global.isLightrider`
- use `NetworkRecord.recordsFromLogs` to get back network records just as LH normally parses them (with LR changes from headers if applicable)
- use `expect()` for assertions (even though `expect(x).toBe(y)` is really ugly :) so we can use `expect(record).toMatchObject(originalReq)` when no changes from headers should have been made

### DIFF
--- a/lighthouse-core/test/lib/network-request-test.js
+++ b/lighthouse-core/test/lib/network-request-test.js
@@ -6,189 +6,170 @@
 'use strict';
 
 const NetworkRequest = require('../../lib/network-request.js');
-const assert = require('assert');
+const NetworkRecorder = require('../../lib/network-recorder.js');
+const networkRecordsToDevtoolsLog = require('../network-records-to-devtools-log.js');
 
 /* eslint-env jest */
-describe('network request', function() {
+describe('NetworkRequest', () => {
   afterEach(() => {
     global.isLightrider = undefined;
   });
 
-  function getRequest() {
-    const req = new NetworkRequest();
-    req.transferSize = 100;
-    req.responseHeaders = [{name: NetworkRequest.HEADER_FETCHED_SIZE, value: 10}];
-    return req;
-  }
+  describe('update transfer size for Lightrider', () => {
+    function getRequest() {
+      return {
+        transferSize: 100,
+        responseHeaders: [{name: NetworkRequest.HEADER_FETCHED_SIZE, value: '10'}],
+      };
+    }
 
-  describe('update transfer size for lightrider', function() {
-    it('does nothing if not Lightrider', function() {
+    it('does nothing if not Lightrider', () => {
       const req = getRequest();
-      global.isLightrider = false;
+      expect(req.transferSize).toBe(100);
 
-      assert.equal(req.transferSize, 100);
-      req._updateTransferSizeForLightrider();
-      assert.equal(req.transferSize, 100);
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+
+      expect(record.transferSize).toBe(100);
     });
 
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/lib/network-request-test.js
```diff
@@ -6,189 +6,170 @@
...
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
       global.isLightrider = true;
-      req.responseHeaders = [{name: NetworkRequest.HEADER_FETCHED_SIZE, value: 'ten'}];
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
 
-      assert.equal(req.transferSize, 100);
-      req._updateTransferSizeForLightrider();
-      assert.equal(req.transferSize, 100);
-      assert.notEqual(req.transferSize, NaN);
+      expect(record).toMatchObject(req);
```
patrickhulce: nit: I think it'd be clearer if it was just like the others `expect(req.transferSize).toBe(100)`

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -6,189 +6,170 @@
...
+      expect(req.transferSize).toBe(100);
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      global.isLightrider = true;
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
 
-      assert.equal(req.transferSize, 100);
-      req._updateTransferSizeForLightrider();
-      assert.equal(req.transferSize, 100);
+      expect(record).toMatchObject(req);
```
patrickhulce: same

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -6,189 +6,170 @@
...
-      assert.equal(req.startTime, 0);
-      assert.equal(req.endTime, 2);
-      assert.equal(req.responseReceivedTime, 1);
+      req.responseHeaders = [];
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      global.isLightrider = false;
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+
+      expect(record).toMatchObject(req);
```
patrickhulce: these tests feel a bit hollow when combined with `toMatchObject` since the only affect the code will ever have right now is mutating `lrStatistics` and we'll never asset anything about that property

should we assert something on `lrStatistics`? or are these tests mostly just for saving time when we do the followup with the real manipulation code?

patrickhulce:
also @brendankenny what are your thoughts on `toEqual` then? should be the same for primitives like this :)

lighthouse-core/test/lib/network-request-test.js
brendankenny: > these tests feel a bit hollow when combined with `toMatchObject` since the only effect the code will ever have right now is mutating `lrStatistics` and we'll never asset anything about that property

Yeah, I think these were intended as counterpoints to tests where the other timing numbers were changing. Those tests were removed when that was dropped from the PR, but these remained. Seems ok to leave in since it's a little hard to tease out a dividing line and the timing changes should come very soon, but yeah, these are mostly just a test of `networkRecordsToDevtoolsLog` right now :)

I can put an explicit check for `lrStatistics` just so it's clear it's expected to not be defined.

brendankenny:
> also @brendankenny what are your thoughts on `toEqual` then? should be the same for primitives like this :)

My thoughts are that I wish they had fewer variations of these :) Seems better, though; switched.

PR #8078, opened 2019-04-08
tests(smokehouse): set Content-Type for json served by static-server.js

Hoten:


Hoten:
[That's it](https://www.youtube.com/watch?v=8bbTtPL1jRs)!

PR #8109, opened 2019-04-09
clients(lr): apply Lightrider timings to NetworkRequests

exterkamp:
**Summary**
Follow up to #7888 which now actually applies those LR timings.

Call out if I forgot to re-add anything.

### DIFF
--- a/lighthouse-core/lib/network-request.js
+++ b/lighthouse-core/lib/network-request.js
@@ -428,8 +428,41 @@ class NetworkRequest {
       return;
     }
 
+    // Init blank timing if needed.
+    if (this.timing === undefined) {
+      this.timing = {
+        requestTime: this.startTime,
+        proxyStart: -1,
+        proxyEnd: -1,
+        dnsStart: -1,
+        dnsEnd: -1,
+        connectStart: -1,
+        connectEnd: -1,
+        sslStart: -1,
+        sslEnd: -1,
+        workerStart: -1,
+        workerReady: -1,
+        sendStart: -1,
+        sendEnd: -1,
+        pushStart: -1,
+        pushEnd: -1,
+        receiveHeadersEnd: -1,
+      };
+    }
+
+    const origEnd = this.endTime;
+    // EndTime and responseReceivedTime are in seconds, so conversion is necessary.
+    this.endTime = this.startTime + (totalMs / 1000);
+    this.responseReceivedTime = this.startTime + ((TCPMs + requestMs) / 1000);
+    this.timing.connectStart = 0;
+    this.timing.connectEnd = TCPMs;
+    this.timing.sslStart = TCPMs - SSLMs;
+    this.timing.sslEnd = TCPMs;
+    this.timing.sendStart = TCPMs;
+    this.timing.sendEnd = TCPMs;
+    this.timing.receiveHeadersEnd = TCPMs + requestMs;
     this.lrStatistics = {
-      endTimeDeltaMs: (this.endTime - (this.startTime + (totalMs / 1000))) * 1000,
+      endTimeDeltaMs: (origEnd - this.endTime) * 1000,
       TCPMs: TCPMs,
       requestMs: requestMs,
       responseMs: responseMs,
--- a/lighthouse-core/lib/network-request.js
+++ b/lighthouse-core/lib/network-request.js
@@ -451,9 +451,10 @@ class NetworkRequest {
     }
 
     const origEnd = this.endTime;
-    // EndTime and responseReceivedTime are in seconds, so conversion is necessary.
+    // `this.endTime` and `this.responseReceivedTime` are in seconds, so convert from milliseconds.
     this.endTime = this.startTime + (totalMs / 1000);
     this.responseReceivedTime = this.startTime + ((TCPMs + requestMs) / 1000);
+
     this.timing.connectStart = 0;
     this.timing.connectEnd = TCPMs;
     this.timing.sslStart = TCPMs - SSLMs;
@@ -461,6 +462,7 @@ class NetworkRequest {
     this.timing.sendStart = TCPMs;
     this.timing.sendEnd = TCPMs;
     this.timing.receiveHeadersEnd = TCPMs + requestMs;
+
     this.lrStatistics = {
       endTimeDeltaMs: (origEnd - this.endTime) * 1000,
       TCPMs: TCPMs,
--- a/lighthouse-core/lib/network-request.js
+++ b/lighthouse-core/lib/network-request.js
@@ -428,27 +428,25 @@ class NetworkRequest {
       return;
     }
 
-    // Init blank timing if needed.
-    if (this.timing === undefined) {
-      this.timing = {
-        requestTime: this.startTime,
-        proxyStart: -1,
-        proxyEnd: -1,
-        dnsStart: -1,
-        dnsEnd: -1,
-        connectStart: -1,
-        connectEnd: -1,
-        sslStart: -1,
-        sslEnd: -1,
-        workerStart: -1,
-        workerReady: -1,
-        sendStart: -1,
-        sendEnd: -1,
-        pushStart: -1,
-        pushEnd: -1,
-        receiveHeadersEnd: -1,
-      };
-    }
+    // Init timing.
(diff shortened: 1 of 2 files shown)


lighthouse-core/lib/network-request.js
```diff
@@ -428,8 +428,41 @@ class NetworkRequest {
...
+        sendStart: -1,
+        sendEnd: -1,
+        pushStart: -1,
+        pushEnd: -1,
+        receiveHeadersEnd: -1,
+      };
+    }
+
+    const origEnd = this.endTime;
+    // EndTime and responseReceivedTime are in seconds, so conversion is necessary.
```
brendankenny:
```suggestion
    // `this.endTime` and `this.responseReceivedTime` are in seconds, so convert from milliseconds.
```

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -90,16 +90,16 @@ describe('NetworkRequest', () => {
       };
     }
 
-    it('updates lrStatistics if in Lightrider', () => {
+    it('updates fetch stats and lrStatistics if in Lightrider', () => {
```
brendankenny:
```suggestion
    it('updates fetch timing numbers if in Lightrider', () => {
```
really this is about those, `lrStatistics` is just the logging instrument, so makes sense to focus on the timing numbers

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +176,64 @@ describe('NetworkRequest', () => {
       expect(record).toMatchObject(req);
       expect(record.lrStatistics).toStrictEqual(undefined);
     });
+
+    it('Handles negative timing data', function() {
```
brendankenny:
```suggestion
    it('treats negative timing numbers as 0', () => {
```

lighthouse-core/test/lib/network-request-test.js
brendankenny: should maybe update the `'does not update lrStatistics when'` below to the same "does nothing if" used in the other tests, too

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +176,64 @@ describe('NetworkRequest', () => {
...
+      expect(record.responseReceivedTime).toStrictEqual(0);
+      expect(record.lrStatistics).toStrictEqual({
+        endTimeDeltaMs: -8000,
+        TCPMs: 0,
+        requestMs: 0,
+        responseMs: 10000,
+      });
+    });
+
+    it('does not updates fetch stats if SLLMs > TCPMs', function() {
```
brendankenny: there's a `does nothing if SSL time exceeds TCP time` test above, so can probably delete this one

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +176,64 @@ describe('NetworkRequest', () => {
...
+
+    it('initialized a blank timing', function() {
+      const req = getRequest();
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+      global.isLightrider = true;
+
+      expect(record.timing).toStrictEqual(undefined);
+      record._updateTimingsForLightrider();
```
brendankenny: weird merge or copy/paste? Doesn't seem necessary to call `_updateTimingsForLightrider`

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +176,64 @@ describe('NetworkRequest', () => {
...
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      global.isLightrider = true;
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+
+      expect(record).toMatchObject(req);
+      expect(record.lrStatistics).toStrictEqual(undefined);
+    });
+
+    it('initialized a blank timing', function() {
```
brendankenny:
```suggestion
    it('created a new timing property if one did not exist', () => {
```
or something like that

lighthouse-core/test/lib/network-request-test.js
brendankenny: maybe add another similar test that shows that it works if a header is missing? (since it'll default to 0 in that case, as long as the numbers still add up it should work)

lighthouse-core/test/lib/network-request-test.js
brendankenny: I think we want the opposite of this, too, that a network request with an existing timing property has the new numbers merged into it

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +176,64 @@ describe('NetworkRequest', () => {
...
+
+      expect(record.startTime).toStrictEqual(0);
+      expect(record.endTime).toStrictEqual(10);
+      expect(record.responseReceivedTime).toStrictEqual(7.5);
+      expect(record.lrStatistics).toStrictEqual({
+        endTimeDeltaMs: -8000,
+        TCPMs: 5000,
+        requestMs: 2500,
+        responseMs: 2500,
+      });
```
brendankenny: for all the tests where the numbers are updated, seems like we also want to check that the `record.timing` objects are also correct?

lighthouse-core/test/lib/network-request-test.js
exterkamp: I thought it looked familiar.

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +183,125 @@ describe('NetworkRequest', () => {
...
+        {name: NetworkRequest.HEADER_RES, value: '10000'}];
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      global.isLightrider = true;
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+
+      expect(record.startTime).toStrictEqual(0);
+      expect(record.endTime).toStrictEqual(10);
+      expect(record.responseReceivedTime).toStrictEqual(0);
+      expect(record.timing.connectStart).toStrictEqual(0);
```
patrickhulce: you've already done all these so no need to change or anything but this type of situation for `.timing` is exactly what `toMatchObject` is made for :)

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -176,5 +183,125 @@ describe('NetworkRequest', () => {
...
+      expect(lrRecord.timing.receiveHeadersEnd).toStrictEqual(7500);
+      expect(lrRecord.lrStatistics).toStrictEqual({
+        endTimeDeltaMs: -8000,
+        TCPMs: 5000,
+        requestMs: 2500,
+        responseMs: 2500,
+      });
+    });
+
+    it('merges with existing timing properties', function() {
```
patrickhulce: is there any situation where we'll create inconsistent timings here? seems like we probably *don't* want to merge the timings right?

patrickhulce: then again I imagine this won't be a real-world scenario right?

exterkamp: Hm, interesting thought.  @paulirish thoughts?  I think this is moot since we expect it to never happen, but we may want to just override everything.

lighthouse-core/test/lib/network-request-test.js
exterkamp: *cries in jest* whoops, yea this is a lot cleaner.  Done.

lighthouse-core/test/lib/network-request-test.js
paulirish: yup lets just override. seems like we can just drop the `if` statement here

https://github.com/GoogleChrome/lighthouse/blob/a0e5fd24f95a166c3495a13f0e2ff5a5f9e907ef/lighthouse-core/lib/network-request.js#L432


cool

exterkamp: Done.

Issue #7752, opened 2019-03-27
☂️ 👣 Breaking changes for v5


paulirish:
A la #6202 and #4333, the items we are thinking of changing in the next major rls

 - [x] Kill `rawValue` #6199 (PRs #8343, #8385, #????)
 - [x] Remove `disableDeviceEmulation` flag before v5 #7044 (PR #8289)
 - [x] (minor) Inconsistent format of screenshot-thumbnails and final-screenshot #7092 (PR #8299)
 - [x] Switch to a different name for [`LH.Audit.Details.Diagnostic`](https://github.com/GoogleChrome/lighthouse/blob/7a431253ff19397d4b8db9ccdf599e93f7fe40c6/types/audit-details.d.ts#L78). "Diagnostic" is too [overloaded](https://github.com/GoogleChrome/lighthouse/blob/7a431253ff19397d4b8db9ccdf599e93f7fe40c6/lighthouse-core/config/default-config.js#L31) in LH. (PR #8298)
 - [x] Move to minimum Node 10 #7215 (PR #8117)
 - [x] Turn on advanced LR network timings (PR #8109)
 - [x] Lantern FCP improvements (PR #7513)
 - [x] Collect only elements in body for `dom-size` (PR #7241)

**Maybe:**

**Punted: 🏈**
 - benchmarkIndex => CPU throttling #6162
 - Consider always filtering configs #7787

**No:**
 - ~Expose page functions to custom gatherers #7317~
 - ~drop null from `SharedFlagsSettings` properties #7788~

Please add any additional above.

brendankenny:
one bullet point might be to decide on our v5 public interface, at least the start of one. This could also involve finally publishing the our public types :)

I bring this up because

> Expose page functions to custom gatherers #7317

Right now we're reluctant to call even artifacts a public interface since it will make some (even minor) changes a lot harder. We should be judicious in what we expose because we can't tweak or take them back later.

Page functions are short enough (and usually self contained enough) that they could be a good candidate for a list of "useful functions you might want to copy/paste" if we didn't want to commit to their current form, for instance.

benschwarz:
Thanks for posting. We'll be tracking against this.

exterkamp:
Are we going to touch rawValues at all for v5?

patrickhulce:
oh yeah let's make the `.details.timeInMs` or whatever a thing :)

exterkamp:
> benchmarkIndex => CPU throttling #6162

[...]

PR #8117, opened 2019-04-09
misc: drop support for Node before 10.13 (LTS)

### LINKED ISSUE #7215: Move to Node 10
It's [that](https://github.com/GoogleChrome/lighthouse/issues/449) time [again](https://github.com/GoogleChrome/lighthouse/issues/3742).

Node 10.13 was moved to LTS at the end of October 2018. Historically we've waited until some time in the new year to move over to the new version. [Last time we did it](https://github.com/GoogleChrome/lighthouse/pull/4703) with the split to 3.0.

There's not a ton of new stuff (for-await (including for readable streams), `Promise.finally`, BigInt, `fs/promises`, global `URL`, `--experimental-worker`), but they offer some pretty decent QoL improvements.

Regardless of timeline, we'll need to update our testing configs to start testing node 11 as well.

brendankenny:
fixes #7215 and part of #7752

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,11 +9,11 @@ branches:
 environment:
   fast_finish: true
   matrix:
-    - nodejs_version: "8"
+    - nodejs_version: "10"
       platform: x86
     # other Node versions are skipped, as appveyor only allows 1 concurrent job
     # and we want appveyor finishing ASAP. see #2382
-    # - nodejs_version: "9"
+    # - nodejs_version: "11"
     #   platform: x86
 
 build: off
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,10 +11,6 @@ environment:
   matrix:
     - nodejs_version: "10"
       platform: x86
-    # other Node versions are skipped, as appveyor only allows 1 concurrent job
-    # and we want appveyor finishing ASAP. see #2382
-    # - nodejs_version: "11"
-    #   platform: x86
 
 build: off
 
--- a/lighthouse-core/lib/asset-saver.js
+++ b/lighthouse-core/lib/asset-saver.js
@@ -139,9 +139,8 @@ async function prepareAssets(artifacts, audits) {
 }
 
 /**
- * Generates a JSON representation of traceData line-by-line to avoid OOM due to very large traces.
- * COMPAT: As of Node 9, JSON.parse/stringify can handle 256MB+ strings. Once we drop support for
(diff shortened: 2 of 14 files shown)


lighthouse-core/lib/asset-saver.js
```diff
@@ -139,9 +139,8 @@ async function prepareAssets(artifacts, audits) {
 }
 
 /**
- * Generates a JSON representation of traceData line-by-line to avoid OOM due to very large traces.
- * COMPAT: As of Node 9, JSON.parse/stringify can handle 256MB+ strings. Once we drop support for
- * Node 8, we can 'revert' PR #2593. See https://stackoverflow.com/a/47781288/89484
```
brendankenny: even though we *can* do this now, I think we like the one trace event per line format, so it made sense to drop the old comment.

.appveyor.yml
```diff
@@ -9,11 +9,11 @@ branches:
 environment:
   fast_finish: true
   matrix:
-    - nodejs_version: "8"
+    - nodejs_version: "10"
       platform: x86
     # other Node versions are skipped, as appveyor only allows 1 concurrent job
     # and we want appveyor finishing ASAP. see #2382
-    # - nodejs_version: "9"
+    # - nodejs_version: "11"
```
patrickhulce: should we even bother to keep this in here?

patrickhulce:
Hang on though, tsc isn't happy with new node types

.appveyor.yml
brendankenny: > should we even bother to keep this in here?

lol, no :)

brendankenny:
there was a conflict with the new node types assuming that there would always be a `Error.code` (as there usually is in node). Fortunately it's easy to drop our `Error` augmentation since we've mostly switched over to `LHError`.

The last holdout fortunately fixes a long-time TODO (that we probably would have missed again for the major version change :), but on the downside is more annoying to review :)

lighthouse-core/lib/asset-saver.js
Hoten: yes, we love it

types/externs.d.ts
```diff
@@ -202,10 +196,6 @@ declare global {
       group: string;
     }
 
-    export interface LighthouseError extends Error {
```
patrickhulce: might have been nice to just export our class as this type so we could easily reference it as `LH.LighthouseError` still but I don't feel strongly

brendankenny: yeah, I'd like to add this to our public types (when those actually exist :)

PR #8121, opened 2019-04-10
report(redesign): new gauges, score scale, category headers, variables

Hoten:
Design reference: https://yuinchien.github.io/sandbox/lighthouse/

A possibly exhaustive list of what I did:

* Gauges have a light background.
* Separator removed between normal categories and the PWA.
* Gauges are bigger.
* Line separator between sections.
* Score scale is centered, given a border, and order reversed.
* Previous sticky heady behavior disabled for now - new sticky header will be done in a follow up PR.
* Category description is centered and larger.
* Various new variables defining font sizes, colors, spacings, etc.

See: https://lh-pr-report-redesign.surge.sh/1

#8185


exterkamp:
Love this 😍 

I have nits but I realize that this is only step 1.  From a _gauges only_ perspective: LGTM.

Can we move the header stuff inside the margin?
image

This is the view I get from scrolling once.  I don't know if it is like that for all screens, but on mine it cuts off (see blue circle).  Also the categories float away from the divider (red arrow) it feels a little weird when I scroll.
image

This is the height that the categories jump to, it'd be more natural if the gauge was at the top if possible so that there isnt any info above.
image

And the dividers between categories should be full width or at least wider than they are, the categories seem to run together to me rn.

patrickhulce:
Woohoo awesome stuff already!! Excellent!!! 😍 

A few nit differences I noticed from the design:

- Same as Shane, "the categories float away from the divider (red arrow) it feels a little weird when I scroll."
- Score legend isn't rounded
- Score legend doesn't have enough margin with the lower divider
image

- Section score gauge is 120px in design but 106px in report (the container is 120 in the report, but the circle ends up being quite a bit smaller for some reason)
- Header score gauges are 90px in design but 85px in report
- Section header text is 28px in design but 20px in report
- Section header text for "Performance" is missing margin between the header and the metrics
- Section header description (for a11y and SEO) is missing margin between header and the description

lighthouse-core/report/html/templates.html
```diff
@@ -134,30 +132,15 @@
...
-      height: 100%;
-      background: white;
-      border-radius: 8px;
-    }
-    .lh-scores-wrapper__shadow {
-      border-radius: 0;
-      box-shadow: rgba(0, 0, 0, 0.2) 0px 3px 0px -2px
-    }
     .lh-scores-container {
+      display: grid;
```
patrickhulce: 😍

patrickhulce: do we actually use any grid properties though? what role does this play?

lighthouse-core/report/html/report-styles.css
```diff
@@ -82,6 +82,33 @@
   --chevron-size: 12px;
   --circle-size: calc(3 * var(--header-font-size));
 
+  --color-black-100: #F5F5F5;
```
patrickhulce: doesn't seem like we use most of these, just copied and prepping for future?

Hoten:
Thanks for the detailed bullets. I went through each, and if it took longer than 10 minutes to make progress I skipped for now.

Shane:

> Can we move the header stuff inside the margin?

done

> This is the view I get from scrolling once. I don't know if it is like that for all screens, but on mine it cuts off (see blue circle). Also the categories float away from the divider (red arrow) it feels a little weird when I scroll.

I can't figure out how to remove that, and I don't know enough to know if removing it would be going against the grain in terms of the follow up sticky work.

> And the dividers between categories should be full width or at least wider than they are, the categories seem to run together to me rn.

done

Patrick:

> * Same as Shane, "the categories float away from the divider (red arrow) it feels a little weird when I scroll."



> * Score legend isn't rounded

If you mess with the viewport of the design reference, it goes from rounded -> square (and then gets stuck even after restoring viewport? hmm). @yuinchien is the score scale supposed to always be rounded, or rounded in desktop & squared in mobile?

> * Score legend doesn't have enough margin with the lower divider

done

> * Section score gauge is 120px in design but 106px in report (the container is 120 in the report, but the circle ends up being quite a bit smaller for some reason)

Looks the same to me.

image

> * Header score gauges are 90px in design but 85px in report

this is hard for me. idk why. maybe because the previous styles have the svg positioned normally & the percentage/ absolute, but the design reference does it the other way, which really messes with my tried-and-true "copy and paste and hammer" css skills.

> * Section header text is 28px in design but 20px in report

done

> * Section header text for "Performance" is missing margin between the header and the metrics

done

> * Section header description (for a11y and SEO) is missing margin between header and the description

done

See: http://lh-pr-report-redesign.surge.sh/2

yuinchien:
Screen Shot 2019-04-10 at 10 45 53 AM
@Hoten the score scale should have 100% rounded corner consistently across screen sizes.

patrickhulce:
> Looks the same to me.

Hm it's definitely rendering the arcs smaller on my machine. The stroke width is also 1px larger in the report than the design which probably contributes to why it visually feels smaller even if the difference is only 6px.


image

image

patrickhulce:
lookin' real good my friend! who even needs 2 weeks psh 😎

Hoten:
ah, I needed to change the radius.

See: https://lh-pr-report-redesign.surge.sh/3

exterkamp:
Just a nit for the future and possibly with that _NOW_ stuff @paulirish: can we make these examples with a site that doesn't do so well on every category and audit like example.com?  Maybe something more like amazon or theverge or apple?  So there is some more variety in the perf section and looks a little more realistic?

Hoten:
See: https://lh-pr-report-redesign.surge.sh/4

paulirish:
>  can we make these examples with a site that doesn't do so well on every category and audit like example.com

SG. I have 5 sites worth of artifacts locally that I use for this purpose. i can incorporate that into #8159

Hoten:
I know @paulirish is working on the sticky header so I suggest leaving the weird sticky behavior. push this to master? or nah?

patrickhulce:
@Hoten I know my comments on the code itself weren't big or anything, but an explanation to satisfy my curiosity would be nice before a merge :)

Hoten:
doh! I missed all the comments.

lighthouse-core/report/html/report-styles.css
Hoten: yeah. there's gonna be a last step for cleaning up stuff. I found myself doing a lot of back and forth copying and pasting these values as I came across them.

lighthouse-core/report/html/templates.html
Hoten: It centered the thingy :P

patrickhulce: oh haha, there's probably a more obvious way to express centering :) 

we could do `display: flex` with `justify-content: center`?

lighthouse-core/report/html/report-styles.css
patrickhulce: > there's gonna be a last step for cleaning up stuff.

in general that sentence scares me a bit with CSS 😆 

I wish in addition to unused CSS tools there were "useless CSS" tools. i.e. this property you have set isn't actually doing anything, but I dream :)

lighthouse-core/report/html/templates.html

lighthouse-core/report/html/report-styles.css
Hoten: > I wish in addition to unused CSS tools there were "useless CSS" tools. i.e. this property you have set isn't actually doing anything, but I dream :)

i can dig it

curate many different reports, render in Chrome w/ a couple different viewports, attach a DT session to each and for each node, record which rules(*) are active. The rules that were never marked active are always useless and can be remove. bonus points if you toggle spacing rules (margin, padding, max-*/min-* etc) to see if the node changes spacing at all. if not, delete that rule.

* i assume this must be possible since the DT inspector shows rules as grayed out when they are overridden by a more specific rule.

Hoten:
nah i'd say the closest is 🌡

lighthouse-core/report/html/report-styles.css
patrickhulce: That's a nice approach! I'm also thinking the rules that are technically active they just have no effect on the final display (maybe that's just your bonus points 😃)

Hoten: take it a step further and make screenshots after each change :P

I want this tool now

lighthouse-core/report/html/templates.html
```diff
@@ -681,19 +663,6 @@
 <!-- Lighthouse PWA badge gauge -->
 <template id="tmpl-lh-gauge--pwa">
   <style>
-    .lh-gauge--pwa__wrapper {
```
paulirish: you managed to have it share the styles?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -336,13 +334,13 @@ class CategoryRenderer {
     // Cast `null` to 0
     const numericScore = Number(category.score);
     const gauge = this.dom.find('.lh-gauge', tmpl);
-    // 329 is ~= 2 * Math.PI * gauge radius (53)
+    // 352 is ~= 2 * Math.PI * gauge radius (56)
     // https://codepen.io/xgad/post/svg-radial-progress-meters
-    // score of 50: `stroke-dasharray: 164.5 329`;
+    // score of 50: `stroke-dasharray: 176 352`;
```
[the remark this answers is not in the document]
Hoten: fixed

lighthouse-core/report/html/report-styles.css
```diff
@@ -106,11 +135,57 @@
...
+    --score-size: 72px;
+    --score-title-font-size-big: 22px;
+    --score-title-font-size: 14px;
+    --score-title-line-height-big: 26px;
+    --score-title-line-height: 20px;
+    --section-padding: 24px;
+    --topbar-height: 32px;
+  }
+
+  .two-col.audits .audit {
```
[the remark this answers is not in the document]
Hoten: removed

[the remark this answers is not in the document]

lighthouse-core/report/html/report-styles.css
```diff
@@ -807,47 +872,48 @@
...
-  text-align: right;
-  transform-origin: bottom right;
-  will-change: opacity; /* opacity is changed on scroll */
+  padding: 8px 8px;
 }
 
 .lh-scorescale-range {
-  margin-left: 10px;
+  margin: 0 12px;
+  font-family: 'Roboto Mono', monospace;
```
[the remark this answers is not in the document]

lighthouse-core/report/html/templates.html
```diff
@@ -581,45 +554,43 @@
...
-    .lh-gauge__wrapper--fail,
-    .lh-gauge__wrapper--fail .lh-gauge {
-      --circle-color: var(--fail-color);
-      color: var(--circle-color);
+    .lh-gauge {
+      stroke-linecap: round;
+      width: var(--score-size);
+      height: var(--score-size);
+      font-size: var(--score-number-fontsize);
+      font-family: 'Roboto Mono', monospace;
```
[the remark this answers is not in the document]

lighthouse-core/report/html/templates.html
```diff
@@ -581,45 +554,43 @@
...
+      stroke: var(--color-red);
     }
 
-    .lh-gauge__wrapper--fail,
-    .lh-gauge__wrapper--fail .lh-gauge {
-      --circle-color: var(--fail-color);
-      color: var(--circle-color);
+    .lh-gauge {
+      stroke-linecap: round;
+      width: var(--score-size);
```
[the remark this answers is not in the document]

lighthouse-core/report/html/templates.html
```diff
@@ -633,44 +604,55 @@
...
   <a href="#" class="lh-gauge__wrapper">
-    <svg viewBox="0 0 120 120" class="lh-gauge" fill="none" stroke-width="2">
-      <circle class="lh-gauge-base" r="53" cx="60" cy="60"></circle>
-      <circle class="lh-gauge-arc" transform="rotate(-90 60 60)" stroke-dasharray="0 329" stroke-dashoffset="0" r="53" cx="60" cy="60"></circle>
+    <svg viewBox="0 0 120 120" class="lh-gauge">
+      <circle class="lh-gauge-base" r="56" cx="60" cy="60"></circle>
+      <circle class="lh-gauge-arc" transform="rotate(-90 60 60)" r="56" cx="60" cy="60"></circle>
     </svg>
     <div class="lh-gauge__percentage"></div>
     <div class="lh-gauge__label"></div>
```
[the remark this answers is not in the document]

lighthouse-core/report/html/templates.html
Hoten: yup

lighthouse-core/report/html/report-styles.css
```diff
@@ -106,11 +135,57 @@
...
+    display: block;
+  } */
+}
+
 .lh-vars.lh-devtools {
   --text-font-family: '.SFNSDisplay-Regular', 'Helvetica Neue', 'Lucida Grande', sans-serif;
   --monospace-font-family: 'Menlo', 'dejavu sans mono', 'Consolas', 'Lucida Console', monospace;
   --body-font-size: 12px;
-  --body-line-height: 16px;
+  --body-line-height: 20px;
```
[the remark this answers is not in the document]
Hoten: reverted - didn't mean to change this value, but the value for mobile.

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -204,7 +204,8 @@ class ReportRenderer {
 
     for (const category of report.reportCategories) {
       const renderer = specificCategoryRenderers[category.id] || categoryRenderer;
-      categories.appendChild(renderer.render(category, report.categoryGroups));
+      const wrapper = renderer.dom.createChildOf(categories, 'div', 'lh-category-wrapper');
```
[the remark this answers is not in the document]
paulirish: ^

lighthouse-core/report/html/report-styles.css
paulirish: bump on 


    /* TODO(cjamcl) clean up unused variables. */

lighthouse-core/report/html/renderer/report-renderer.js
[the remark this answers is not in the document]
Hoten: oops, i hit "batch" instead of "commit" then forgot to push

Hoten:
@paulirish gotcha

Issue #2289, opened 2017-05-17
DevTools: node revealing is flaky


paulirish:
First a little primer on how this system works:

#### On the Lighthouse side:

* we collect a selector path, an outerHTML snippet, we generate the proprietary node path. 
* we set all that on data attributes on the element for safekeeping.

#### [On the devtools side](https://github.com/ChromeDevTools/devtools-frontend/blob/8dd291f42b57c4d74fd637dbaed00e544a85606c/front_end/audits2/Audits2Panel.js#L675-L710):

1. we let Lighthouse make that renderNode element
2. lighthouse is currently reloading the page to clean things up so we need to wait for that to finish.
3. we wait for `SDK.ResourceTreeModel.Events.MainFrameNavigated` to fire once before proceeding
4. once that's done we use the `domModel.pushNodeByPathToFrontend` method to push our fancy node path to the backend and have it return a current nodeID.
1. sometimes it doesnt return any results. sometimes it does return a node, but the `nodeForId` lookup fails because the domModel isn't populated yet. one of these can happen if the node hasn't been inserted into the DOM yet. (something rather lazy). 
1. if we have a working nodeID we give it to `linkifyNodeReference` which hooks up the blue highlight and click

#### notes

1. I've noticed `linkifyNodeReference` being passed nodeIds that appear to be slightly "deprecated" nodeIds for a given element. As in.. the click-to-reveal will work, but the hover will not give a blue highlight. And if you get the current nodeid of the element, it's soemthing like `243` rather than `78`. 
1. there seems to be some other reasons the transition from nodePath to nodeId fail. A) race conditions with reload B) race conditions with the DOM updating and dom model being build/rebuilt. C) something about the DOM structure trips up this resolving phase.

@robdodson wrestled with this a bit tonight which has unearthed some things..

On that last point (2/C), rob noticed that on his test page, having "Any script at the end of body" will make it fail.  If there are no scripts, the whole thing works.
Also if the script is in the head and async, things work.

Why a script at the end could affect this nodePath doesn't really add up. 

But that's where things are at the moment!

kaycebasques:
Can repro on `theverge.com`

<img width="695" alt="screen shot 2017-06-28 at 2 45 31 pm" src="https://user-images.githubusercontent.com/4713486/27662895-b01cb98e-5c14-11e7-869a-52941743928d.png">

carmacleod:
Same problem here: http://www.carbondesignsystem.com/
[...]

PR #8133, opened 2019-04-10
core: add path property to script elements artifact

mattzeunert:
**Summary**

When displaying JSON-LD script elements in DevTools we want to link them to the Elements panel, so we need to collect the node `path` to do that.

I saw [the ScriptElements PR](https://github.com/GoogleChrome/lighthouse/pull/7920) also updated `lighthouse-core/test/results/artifacts/artifacts.json`, but I'm not sure how that's updated...


lighthouse-core/gather/gatherers/script-elements.js
```diff
@@ -9,6 +9,9 @@ const Gatherer = require('./gatherer.js');
 const NetworkAnalyzer = require('../../lib/dependency-graph/simulator/network-analyzer.js');
 const NetworkRequest = require('../../lib/network-request.js');
 const getElementsInDocumentString = require('../../lib/page-functions.js').getElementsInDocumentString; // eslint-disable-line max-len
+const pageFunctions = require('../../lib/page-functions.js');
+
+/* global getNodePath */
 
 /**
  * @return {LH.Artifacts['ScriptElements']}
```
patrickhulce: we need to update the types too

lighthouse-core/gather/gatherers/script-elements.js
```diff
@@ -26,6 +29,8 @@ function collectAllScriptElements() {
       async: script.async,
       defer: script.defer,
       source: /** @type {'head'|'body'} */ (script.closest('head') ? 'head' : 'body'),
+      // @ts-ignore - getNodePath put into scope via stringification
+      path: getNodePath(script),
```
patrickhulce: `path` might be a tad too generic on the `*Elements` artifacts for my taste, given it's basically a devtools-specific crazy node child index array

WDYT about...

`devtoolsNodePath`
`devtoolsChildIndexNodePath`
`nodePathForDevToolsRevealing`

any other suggestions welcome :)

mattzeunert: `nodeReferencePath`? I'd like to avoid tying it to DevTools, since other consumers could theoretically use it too.

So for now we'd use the new name here and keep `path` everywhere else?

patrickhulce: > since other consumers could theoretically use it too

😱 I hope not!

😆 If you can't tell, I'm not fond of this particular path approach, so I would prefer to limit its scope and push a more flexible strategy that can be resilient to reloads.

> So for now we'd use the new name here and keep path everywhere else?

yeah just thinking out loud here. I've proposed to only expose the `*Elements` artifacts as the public API for v5 so I'm thinking of them quite separately from the others at this point, which might not align with the rest of LH team yet :)

Hoten: >  I would prefer to limit its scope and push a more flexible strategy that can be resilient to reloads.

https://github.com/GoogleChrome/lighthouse/issues/2289#issuecomment-446606864 fyi

patrickhulce: Thanks @Hoten I'm glad we've already made some awesome improvements here :D 

To clarify, my main goal if we're exposing this to plugin consumers in `*Elements` would be to provide a path that has some value even when the node cannot be found, i.e. based on some sort of identifiers/classes/etc on the page and not just child indexes that don't provide much value to a human, some selector/path hybrid I guess.

mattzeunert: > 😱 I hope not!

Ok, `__legacyDevToolsPathDontUseEver` it is then 😂

> some selector/path hybrid I guess

Were you thinking of the `selector` property we show currently when the user hovers over a node in the report?

Axe already [aims to make that a unique selector](https://github.com/dequelabs/axe-core/blob/develop/lib/core/utils/get-selector.js#L336). [`getNodeSelector`](https://github.com/GoogleChrome/lighthouse/blob/b374ea409fc56fa1af7af8e9f3a5c6c1f96ee30d/lighthouse-core/lib/page-functions.js#L193) generates selectors based on the id/class name if available, but doesn't attempt to make that unique.

Seems like we don't need `path` at all if we can make `selector` unique.

patrickhulce: `__legacyDevToolsPathDontUseEver` 😍 love it! 😆 

I don't want to hold things up on this, but I'm just afraid of what we won't be able to take back once 5.0 hits :)

Anyone else core have thoughts on our public artifacts API? I think I might stir up this convo over on the issue...

brendankenny: `devtoolsNodePath` sounds great to me, personally. Sounds uninvitingly opaque if you aren't devtools, you still know exactly what it is, and even if you do figure out how to use it, it sounds like the gears of devtools turning could change the format at any time.

Also fine with a leading underscore (but I vote we stay away from two :) and a strongly worded comment, but I think that's more than enough at that point.

brendankenny: @paulirish if he has an opinion

paulirish: fine with `devtoolsNodePath`. seems cool.

Here's where the protocol uses it: https://vanilla.aslushnikov.com/?DOM.pushNodeByPathToFrontend

patrickhulce:
travis is harder to please though :) I think you'll have to provide an empty one for the network-discovered scripts

Issue #4359, opened 2018-01-25
[SEO Audits] Structured data is valid


rviscomi:
**Status:** This is in the exploratory phase, not ready for development.

The idea is to have an audit that comprehensively lints/validates structured data of any flavor:
- JSON-LD
- Microdata
- RDFa 1.1

To do so we would need to keep up with changes to the schemas and ensure that they are well-defined. These are not currently feasible so this issue is on the back burner (or chillin :snowflake:  depending on how you look at it).

cc @kdzwinel

rviscomi:
@kdzwinel let's start by setting up a new repo on your personal GH account and getting the plumbing to work so that it is registered on npm and installed by LH. You can send me any code reviews outside of the LH repo.

rviscomi:
TODO: Is there anything we can reuse from Chromium's "copyless paste" feature, which does JSON-LD parsing? https://github.com/chromium/chromium/search?utf8=%E2%9C%93&q=json-ld&type=

paulirish:
What's in scope for the first pass here? The three checkboxes above?

Also, do we know how much of the heavy lifting can be done by existing libraries?

paulirish:
I do see quite a few json-ld projects already, so I was mostly curious on that side.
Seems like JSONlint would be good to provide better diagnostic data than what `JSON.parse` does; seems good if we're finding a lot of invalid JSON.

The three checkboxes above are what you're scoping this MVP to?

rviscomi:
Oh I actually linked to the wrong project. jsonlint does look interesting and may help. I meant to link to https://github.com/digitalbazaar/jsonld.js, which is JSON-LD specific.

As for the MVP, the first two (valid JSON, standard props) are required and the third is a stretch goal.

Paul, on the subject of an external repo, I was curious to hear your thoughts on internal vs external. My hope/expectation is for the structured data validation code to be reusable in other contexts and also maintained by the web community. I think an external repo would be more conducive to those things, eg how the axe a11y logic is external to LH. WDYT?

kdzwinel:
> Paul, on the subject of an external repo, I was curious to hear your thoughts on internal vs external.

We have been talking a lot on the engineering meeting about solutions such as lerna, yarn workspaces and doing what we are doing with lh-logger (same repo, different package), but I haven't asked why we don't want a separate repo in the first place?

paulirish:
breaking out the seperate repo discussion into #4955

kdzwinel:
[...]

PR #8137, opened 2019-04-10
core(jsonld): Structured data validation updates

mattzeunert:
**Summary**

Updates to the SD validation logic:

- add line numbers to error messages
- store invalid entity type for "unexpected property" errors

We use the entity type to include them in the error messages and link to their schema.org URL so the user can see what properties are valid:

Screenshot 2019-04-10 at 13 32 28

From the audit's perspective the validator would ideally generate the full error message including the markdown link, but that logic is in the audit to make the sd validator more re-usable.

**Related Issues/PRs**

Part of https://github.com/GoogleChrome/lighthouse/issues/4359


lighthouse-core/lib/sd-validation/jsonld-keyword-validator.js
```diff
@@ -41,7 +41,7 @@ module.exports = function validateJsonLD(json) {
     if (name.startsWith('@') && !VALID_KEYWORDS.has(name)) {
       errors.push({
         path: path.join('/'),
-        message: 'Unknown keyword',
+        message: 'Unknown keyword ' + name,
```
patrickhulce: should we put this in quotes or after a colon or something? this also makes we wonder what our i18n story is here 😬

lighthouse-core/lib/sd-validation/json-linter.js
```diff
@@ -9,21 +9,21 @@ const jsonlint = require('jsonlint-mod');
...
- * @returns {{message: string, lineNumber: string|null}|null}
+ * @returns {{message: string, lineNumber: number|null}|null}
  */
 module.exports = function parseJSON(input) {
   try {
     jsonlint.parse(input);
   } catch (error) {
-    /** @type {string|null} */
+    /** @type {number|null} */
     let line = error.at;
```
patrickhulce: oh is `error.at` actually a number then?

lighthouse-core/test/lib/sd-validation-test.js
```diff
@@ -17,7 +17,7 @@ describe('JSON validation', () => {
     `);
 
     assert.equal(errors.length, 1);
-    assert.equal(errors[0].path, 2);
```
patrickhulce: we should probably use strict equal here then, these would have been failing before I guess

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,73 @@ module.exports = async function validate(textInput) {
...
     });
   }
 
   return [];
 };
+
+/**
+ * @param {*} obj
+ * @param {string} path
+ * @returns null | number - line number of the path value in the prettified JSON
```
patrickhulce: whoa are the `{}` optional? 😮

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,73 @@ module.exports = async function validate(textInput) {
...
+ */
+function getLineNumberFromJsonLDPath(obj, path) {
+  // To avoid having an extra dependency on a JSON parser we set a unique key in the
```
patrickhulce: has that shipped sailed with `jsonlint-mod` is there any way to leverage that?

patrickhulce: also deja vu, I feel like I've reviewed this before so sorry if you've answered this in the previous PR

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,73 @@ module.exports = async function validate(textInput) {
...
+  // object and then use that to identify the correct line
+  const searchKey = Math.random().toString();
+  obj = JSON.parse(JSON.stringify(obj));
+
+  setValueAtJsonLDPath(obj, path, searchKey);
+  const jsonLines = JSON.stringify(obj, null, 2).split('\n');
```
patrickhulce: if we're reformatting the JSON then this line number is pretty much unrelated to the line number in the source right?

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,73 @@ module.exports = async function validate(textInput) {
...
+    }
+
+    const isLastPart = pathParts.length - 1 === i;
+    let keyFound = false;
+    for (const key of Object.keys(currentObj)) {
+      // The actual key in JSON might be an absolute IRI like "http://schema.org/author"
+      // but key provided by validator is "author"
+      const keyParts = key.split('/');
+      const relativeKey = keyParts[keyParts.length - 1];
+      if (relativeKey === pathPart && currentObj[key] !== undefined) {
```
patrickhulce: seems like if we have `currentObj[key]` as `undefined` we could just return before this whole loop

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,73 @@ module.exports = async function validate(textInput) {
...
+        // If we've arrived at the end of the provided path set the value, otherwise
+        // continue iterating with the object at the key location
+        if (isLastPart) {
+          currentObj[key] = value;
+        } else {
+          currentObj = currentObj[key];
+        }
+        keyFound = true;
```
patrickhulce: seems like `keyFound` is unnecessary, we could just return here and throw if we ever make it through the loop without returning

lighthouse-core/test/lib/sd-validation-test.js
```diff
@@ -169,7 +171,9 @@ describe('schema.org validation', () => {
     }`);
 
     assert.equal(errors.length, 1);
+    assert.equal(errors[0].invalidTypes[0], 'http://schema.org/Article');
```
patrickhulce: this name is a bit confusing to me for how it's used. the type isn't invalid, right? it's just the type of the entity that has some *other* invalid stuff going on?

lighthouse-core/lib/sd-validation/schema-validator.js
```diff
@@ -114,20 +123,22 @@ module.exports = function validateSchemaOrg(expandedObj) {
...
-      keyErrorMessages.forEach(message =>
+      keyErrors.forEach(error => {
         errors.push({
+          invalidTypes: error.invalidTypes,
+          message: error.message,
           // get rid of the first chunk (/@type) as it's the same for all errors
           path:
             '/' +
             path
               .slice(0, -1)
```
patrickhulce: wait a second, the path is backwards? I think we're missing some tests that have paths of length >2 then 🤔

lighthouse-core/lib/sd-validation/json-linter.js
mattzeunert: Yeah https://github.com/circlecell/jsonlint-mod/blob/master/lib/doug-json-parse.js#L67

lighthouse-core/lib/sd-validation/sd-validation.js
mattzeunert: > has that shipped sailed with jsonlint-mod is there any way to leverage that?

I don't see a direct way to leverage jsonlint-mod here. It just uses Douglas Crockford's JSON parser.

I guess we could parse it ourselves or find a small parser that we can include?

> also deja vu, I feel like I've reviewed this before so sorry if you've answered this in the previous PR

Yup, I sent you a [pre-PR](https://github.com/mattzeunert/lighthouse/pull/1) to get some initial feedback.

lighthouse-core/lib/sd-validation/sd-validation.js
mattzeunert: Seems like it. Will fix 😂

lighthouse-core/lib/sd-validation/sd-validation.js
mattzeunert: Yeah. I tried to express that with `line number of the path value in the prettified JSON`, but do you have any suggestions to make it more explicit?

lighthouse-core/lib/sd-validation/schema-validator.js
mattzeunert: I can look into this more tomorrow, but what makes you think the path is backwards? Here's an example of the path:

```[ 'http://schema.org/author', '0', 'http://schema.org/colleague', '0', '@type' ]```

For this: 

```
{
  "@context": "https://schema.org",
  "@type": "Article",
  "author": {
    "@type": "Person",
    "displayNaame": "Sally",
    "colleague": {
      "@type": "Person",
      "displayNaame": "Sally"
    }
  }
}
```

File: lighthouse-core/test/lib/sd-validation-test.js (hunk 9 above)
Comment (mattzeunert) [reply lighthouse-core/test/lib/sd-validation-test.js]: The entity is not a valid `type` instance because it has a property that doesn't exist on `type`.

I originally just called that array `types`, but then it wasn't clear that having that array means invalid. (And we use that array to generate the `Invalid Event: unexpected property asdf` message.)

Maybe we can call the array `invalidForTypes`?

File: lighthouse-core/lib/sd-validation/sd-validation.js (hunk 6 above)
Comment (patrickhulce) [reply lighthouse-core/lib/sd-validation/sd-validation.js]: Oh yeah yeah I think it's already clear in the code, I meant more from the users perspective!

File: lighthouse-core/lib/sd-validation/schema-validator.js (hunk 10 above)
Comment (patrickhulce) [reply lighthouse-core/lib/sd-validation/schema-validator.js]: Oh the comment confused me. It says it's getting rid of the "first chunk" but it's removing the last chunk. That made me think it is backwards and all paths started with `/@type`, but I see now :)

maybe update the comment? a test for longer paths still couldn't hurt though sometime :)

File: lighthouse-core/test/lib/sd-validation-test.js (hunk 9 above)
Comment (patrickhulce) [reply lighthouse-core/test/lib/sd-validation-test.js]: oh gotcha, I actually like the direction you were headed with `.types` then. I think it's clear *something* is invalid by virtue of the fact that we are giving an error. Maybe `typeOfInvalidEntity` or something similar if we're going the more explicit route?

File: lighthouse-core/lib/sd-validation/sd-validation.js (hunk 6 above)
Comment (mattzeunert) [reply lighthouse-core/lib/sd-validation/sd-validation.js]: Yeah, that's a good point. Is there a strong need to address that though? The line numbers probably won't match anyway if the snippet is embedded in an HTML template.

File: lighthouse-core/lib/sd-validation/sd-validation.js (hunk 7 above)
Comment (mattzeunert) [reply lighthouse-core/lib/sd-validation/sd-validation.js]: True, I don't see why we'd need it at all actually...

File: lighthouse-core/lib/sd-validation/schema-validator.js (hunk 10 above)
Comment (mattzeunert) [reply lighthouse-core/lib/sd-validation/schema-validator.js]: Oooh, yeah good catch, that's a very confusing comment! Added a test case for deeper nesting too.

**Comment by mattzeunert** [comment]
I don't think this can happen if the input is a JSON-LD object and the path is from the expanded version of that object. Should we just take it out?

Screenshot 2019-04-11 at 11 36 17

File: lighthouse-core/lib/sd-validation/sd-validation.js (hunk 6 above)
Comment (patrickhulce) [reply lighthouse-core/lib/sd-validation/sd-validation.js]: This is true, I'm just thinking there's not a ton of value in exposing the line number to the user then, but knowing where to highlight is massively awesome, so sorry for the distracting thread :)

File: types/structured-data.d.ts [hunk 11]
```diff
@@ -7,6 +7,21 @@
 declare global {
   module LH {
     module StructuredData {
+
+      export type ValidatorType =
+        | "json"
+        | "json-ld"
+        | "json-ld-expand"
+        | "schema-org";
```
brendankenny: ok, where is this leading `|` coming from :P Not even prettier does this, I don't think

lighthouse-core/lib/sd-validation/schema-validator.js
```diff
@@ -49,7 +49,7 @@ function findType(type) {
  *
  * @param {string|Array<string>} typeOrTypes
  * @param {Array<string>} keys
- * @returns {Array<string>}
+ * @returns {Array<{message: string, key?: string, typesOfInvalidEntity?: Array<string>, path?: string}>}
```
brendankenny: maybe pull into a typedef so these properties can have docs?

lighthouse-core/lib/sd-validation/sd-validation.js
```diff
@@ -64,12 +62,66 @@ module.exports = async function validate(textInput) {
...
 
   return [];
 };
+
+/**
+ * @param {*} obj
+ * @param {string} path
+ * @returns {null | number} - line number of the path value in the prettified JSON
+ */
+function getLineNumberFromJsonLDPath(obj, path) {
```
brendankenny: these two are utility-y enough, maybe they're worth exposing and having explicit tests for them?

They're getting test coverage right now but I doubt it's much, and it would have the benefit of documenting them a bit more. e.g. these aren't called with literals anywhere, so it's not possible to know what the `path` string generally looks like without also looking at the code that generates paths.

types/structured-data.d.ts
patrickhulce: my prettier definitely does this in other projects, not sure about here :)

lighthouse-core/lib/sd-validation/sd-validation.js
mattzeunert: Yeah, adding some tests is a good idea 👍

lighthouse-core/test/lib/sd-validation-test.js
mattzeunert: I'm at `validTypes` right now, since we only include the types in the validation error if they are valid. (And we don't want to show invalid types in the UI, because in that case they'll be part of the error message.)

types/structured-data.d.ts
brendankenny: > my prettier definitely does this in other projects, not sure about here :)

maybe it's a typescript prettier thing? I [couldn't get it to repro in their playground ](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEA3AhgJwAQHcAW68qcOAvNgIzYA+2ATLdgMxMAsTArEwGxMDsTABxMAnE0oAGCdTqVGc1nI5zucvnMFyRc8XXqSA3CAA0ICAAcYAS2gBnZKCyYIuAApYEDlOlQRrACamIABGmOhgANZwMADKFhHWUADmyDCYAK5wZvgwALYANgDq+NbwdglgcLFeZdaoZQCeyOB2DmZJdqQwbuHJeejIAGboBV1mAFZ2AB4AQuFRMbHoeXAAMklww6PjIFPTsUnJBXAAihkQ8Ntj2SAJmF2YLSHoIY0F0MEWmEkwRYEwfDIISSMzfCBdIrhCwtb5wR4kYIARwu8F6lm8IHQdgAtFA4HAAoTgpg4CjrKTeuh+oMkCMbmYunlrGlMrc7EcTudLls6TtbjBXv8AoDkPQzOl0NYCkcAMIQPIDFpQaBbMwZLoAFVe3npXQAvvqgA)

mattzeunert:
@patrickhulce LGTY?

PR #8163, opened 2019-04-10
tests: assert no unused css in report styles

Hoten:
This is what a failure would look like:

image

I noticed it doesn't take into account variables. Didn't evaluate any other options yet, but will try @patrickhulce's nukecss if he gets it working again.

### DIFF
--- a/lighthouse-core/test/report/html/renderer/styles-test.js
+++ b/lighthouse-core/test/report/html/renderer/styles-test.js
@@ -1,0 +1,63 @@
+/**
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/* eslint-env jest */
+
+const fs = require('fs');
+const ReportGenerator = require('../../../../report/report-generator.js');
+const sampleResults = require('../../../../../lighthouse-core/test/results/sample_v2.json');
+const purify = require('purify-css');
+const jsdom = require('jsdom');
+
+/* eslint-disable no-console */
+function captureConsoleLog(fn) {
+  const outputData = [];
+  const originalLog = console.log;
+  console.log = (...inputs) => (outputData.push(...inputs));
+  try {
+    fn();
+  } finally {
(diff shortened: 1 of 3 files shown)


brendankenny:
still not sure I understand how this can work without a lot of work to make sure all permutations are tested. How to make sure we're rendering PWA-badge-specific style rules, or firework styles, etc, so they're counted as used? Is there a typical way of handling this?

Hoten:
my understanding is that purifycss just splits the given HTML by spaces, and uses those words as potential ids/classes. All the pwa / firework stuff would be found in the various html templates, so would't need to be be in the actual rendered dom.

patrickhulce:
> How to make sure we're rendering PWA-badge-specific style rules, or firework styles, etc, so they're counted as used? Is there a typical way of handling this?

I can explain how nukecss works to avoid these problems at least :) 

It parses all HTML and JS and finds all id/class attributes and string values and object property names to see what candidates are for ids and classes. Then it has two modes of operation, strict and non-strict.

In strict mode, there must be an exact match for each component of a selector.

In non-strict mode, there must be some permutation of source tokens that matches each component of a selector.

To @Hoten 's point, purify and nuke both look at source so the idea is exactly that you find the source class in the templates and source code without needing to see all rendered versions of the page.

brendankenny:
*searches for `` `.*\$\{.*` `` in report renderer*

...good luck ;)

patrickhulce:
if we really want to do this, I'd be happy to dust off nuke and make it return the removed selectors so it's easily assertable :)

Hoten:
> if we really want to do this, I'd be happy to dust off nuke and make it return the removed selectors so it's easily assertable :)

I'd say, only bother if you have more functionality than purify-css. Maybe you could add support for vars?

This is interesting: this rule doesn't trigger an error
```
.lh-generated-orange-class-name {
  color: black;
}
```

this does:
```
.lh-generated-grape-class-name {
  color: black;
}
```

gotta be the token permutation stuff.

Hoten:
> gotta be the token permutation stuff.

It is.

I looked at the code for purify-css. it works like how @patrickhulce says nukecss works - it groks tokens from the HTML input ("lh-my-class" -> lh, my, class). that makes sense.

Then, for each css rule it verifies that all tokens were seen in the first step. but it does it in a weird way that isn't obviously correct - https://github.com/purifycss/purifycss/blob/be52c1c6b1b6e287b6989428c57f05a79aa135dc/src/utils/ExtractWordsUtil.js#L40 . Also, there's the whole false positive issues as shown by `.lh-generated-grape-class-name`.

would be cooler to use the DT protocol to identify the styles that are being applied to each node, and do an actual accurate analysis, rather than this wishy-washy subtoken thing.

Hoten:
The token permutation strategy these tools use isn't ideal. Closing b/c to very useful. Could do something cool w/ approach described here: https://github.com/GoogleChrome/lighthouse/pull/8121#discussion_r274985730

PR #8169, opened 2019-04-11
core: adjustments to stack-pack comments and types

brendankenny:
changes:
- clean up of some of the jsdocs for stack pack types and properties in various `d.ts` files
- made `artifacts.Stacks` always defined (just init with a `[]`, get to remove the null checks). makes sense cause it's an array (so no results can just be an empty array) and it's a base artifact, so really should always be defined
- moved `lib/gatherer-stacks.js` to `lib/stack-collector.js` after discussing names with @paulirish. Happy to bikeshed more tho :)
- added types to `js-library-detector` output (in `stack-collector.js`), and settled on normalizing its various combinations of `undefined` and `null` for missing output to just using `undefined` once it becomes an artifact.
- while fixing `js-library-detector` noticed our incorrect async `Object.entries().forEach()`, which gives us our first `for await...of` in honor of moving to node 10 :)

TODO (tomorrow):
- stack pack tests :|

### DIFF
--- a/build/build-bundle.js
+++ b/build/build-bundle.js
@@ -53,7 +53,7 @@ async function browserifyFile(entryPath, distPath) {
 
   bundle
     // Transform the fs.readFile etc into inline strings.
-    .transform('brfs', {global: true, parserOpts: {ecmaVersion: 9}})
+    .transform('brfs', {global: true, parserOpts: {ecmaVersion: 10}})
     // Strip everything out of package.json includes except for the version.
     .transform('package-json-versionify');
 
@@ -121,6 +121,7 @@ function minifyScript(filePath) {
     shouldPrintComment: () => false, // Don't include @license or @preserve comments either
     plugins: [
       'syntax-object-rest-spread',
+      'syntax-async-generators',
     ],
     // sourceMaps: 'both'
   };
--- a/lighthouse-core/gather/gather-runner.js
+++ b/lighthouse-core/gather/gather-runner.js
@@ -7,7 +7,7 @@
 
 const log = require('lighthouse-logger');
 const manifestParser = require('../lib/manifest-parser.js');
-const stacksGatherer = require('../lib/gatherer-stacks.js');
+const stacksGatherer = require('../lib/stack-collector.js');
 const LHError = require('../lib/lh-error.js');
 const URL = require('../lib/url-shim.js');
 const NetworkRecorder = require('../lib/network-recorder.js');
@@ -412,7 +412,7 @@ class GatherRunner {
       NetworkUserAgent: '', // updated later
       BenchmarkIndex: 0, // updated later
       WebAppManifest: null, // updated later
-      Stacks: null, // updated later
+      Stacks: [], // updated later
       traces: {},
       devtoolsLogs: {},
       settings: options.settings,
@@ -480,11 +480,17 @@ class GatherRunner {
         }
         await GatherRunner.beforePass(passContext, gathererResults);
         await GatherRunner.pass(passContext, gathererResults);
+
         if (isFirstPass) {
+          // Fetch the manifest, if it exists. Currently must be fetched before gatherers' `afterPass`.
           baseArtifacts.WebAppManifest = await GatherRunner.getWebAppManifest(passContext);
+        }
+
+        const passData = await GatherRunner.afterPass(passContext, gathererResults);
+
+        if (isFirstPass) {
           baseArtifacts.Stacks = await stacksGatherer(passContext);
         }
-        const passData = await GatherRunner.afterPass(passContext, gathererResults);
 
         // Save devtoolsLog, but networkRecords are discarded and not added onto artifacts.
         baseArtifacts.devtoolsLogs[passConfig.passName] = passData.devtoolsLog;
--- a/lighthouse-core/lib/stack-collector.js
+++ b/lighthouse-core/lib/stack-collector.js
@@ -15,45 +15,51 @@
 
 const fs = require('fs');
 const libDetectorSource = fs.readFileSync(
-  require.resolve('js-library-detector/library/libraries.js'),
-  'utf8'
-);
+  require.resolve('js-library-detector/library/libraries.js'), 'utf8');
+
+/**
+ * @typedef JSLibraryDetectorTest
+ * @property {string} icon Essentially an id, useful if no npm name is detected.
+ * @property {string} url
+ * @property {string|null} npm npm module name, if applicable to library.
+ * @property {function(Window): false|{version: string|null} | Promise<false|{version: string|null}>} test Returns false if library is not present, otherwise returns an object that contains the library version (set to null if the version is not detected).
+ */
 
 /**
  * @typedef JSLibrary
  * @property {string} name
- * @property {string} version
- * @property {string} npm
- * @property {string} iconName
+ * @property {string} icon
+ * @property {string|null} version
+ * @property {string|null} npm
  */
 
 /**
  * Obtains a list of detected JS libraries and their versions.
  */
 /* istanbul ignore next */
-function detectLibraries() {
+async function detectLibraries() {
   /** @type {JSLibrary[]} */
   const libraries = [];
 
   // d41d8cd98f00b204e9800998ecf8427e_ is a consistent prefix used by the detect libraries
   // see https://github.com/HTTPArchive/httparchive/issues/77#issuecomment-291320900
+  /** @type {Record<string, JSLibraryDetectorTest>} */
   // @ts-ignore - injected libDetectorSource var
-  Object.entries(d41d8cd98f00b204e9800998ecf8427e_LibraryDetectorTests).forEach(
-    async ([name, lib]) => {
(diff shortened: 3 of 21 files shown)


build/build-bundle.js
```diff
@@ -121,6 +121,7 @@ function minifyScript(filePath) {
     shouldPrintComment: () => false, // Don't include @license or @preserve comments either
     plugins: [
       'syntax-object-rest-spread',
+      'syntax-async-generators',
```
brendankenny: this change was needed for supporting `for await...of` in our client minifying pipeline. It *feels* like we shouldn't need the plugin now that it's stage 4 and shipping everywhere, but I don't care enough to look into babel more :)

note that this isn't transpiling anything (we only minify), it's just so the babel parser understands the syntax and doesn't throw when seeing this.

(no idea if the `brfs` change is needed, it's just that 10 is the latest version, so might as well)

lighthouse-core/lib/stack-packs.js
```diff
@@ -6,42 +6,48 @@
...
-        continue;
-      }
-
-      packs.push({
-        id: matchedPack.id,
-        title: matchedPack.title,
-        iconDataURL: matchedPack.iconDataURL,
-        descriptions: matchedPack.descriptions,
-      });
+  for (const pageStack of pageStacks) {
```
brendankenny: this is all just whitespace changes due to removing the `if (artifacts.Stacks) {` check

lighthouse-core/lib/stack-packs.js
```diff
@@ -6,42 +6,48 @@
...
+      stackPackToIncl.requiredStacks.includes(`${pageStack.detector}:${pageStack.id}`));
+    if (!stackPackToIncl) {
+      continue;
     }
+
+    // Grab the full pack definition
+    const matchedPack = stackPacks.find(pack => pack.id === stackPackToIncl.packId);
+    if (!matchedPack) {
+      log.warn('StackPacks',
+        `'${stackPackToIncl.packId}' stack pack was matched but is not found in stack-packs lib`);
```
brendankenny: except for warning here :)

lighthouse-core/lib/stack-collector.js
```diff
@@ -15,45 +15,51 @@
...
-  'utf8'
-);
+  require.resolve('js-library-detector/library/libraries.js'), 'utf8');
+
+/**
+ * @typedef JSLibraryDetectorTest
+ * @property {string} icon Essentially an id, useful if no npm name is detected.
+ * @property {string} url
+ * @property {string|null} npm npm module name, if applicable to library.
+ * @property {function(Window): false|{version: string|null} | Promise<false|{version: string|null}>} test Returns false if library is not present, otherwise returns an object that contains the library version (set to null if the version is not detected).
```
patrickhulce: this is kinda a confusing type to read :) could we do something like

```js
/** @typedef {false|{version:string|null}} JSLibraryDetectorTestResult */
/** ... 
  @property {function(Window): (JSLibraryDetectorTestResult|Promise<JSLibraryDetectorTestResult>)}
```

lighthouse-core/gather/gather-runner.js
```diff
@@ -480,12 +480,18 @@ class GatherRunner {
...
         if (isFirstPass) {
+          // Fetch the manifest, if it exists. Currently must be fetched before gatherers' `afterPass`.
           baseArtifacts.WebAppManifest = await GatherRunner.getWebAppManifest(passContext);
-          baseArtifacts.Stacks = await stacksGatherer(passContext);
         }
+
         const passData = await GatherRunner.afterPass(passContext, gathererResults);
 
+        if (isFirstPass) {
+          baseArtifacts.Stacks = await stacksGatherer(passContext);
```
patrickhulce: any reason for moving this after after pass?

lighthouse-core/lib/stack-collector.js
```diff
@@ -15,45 +15,51 @@
...
-            npm: lib.npm,
-            iconName: lib.icon,
-          });
-        }
-      } catch (e) {}
-    }
-  );
+  const libraryDetectorTests = d41d8cd98f00b204e9800998ecf8427e_LibraryDetectorTests; // eslint-disable-line 
+
+  for await (const [name, lib] of Object.entries(libraryDetectorTests)) {
```
patrickhulce: do we get anything out of this? I assumed nearly all of them are synchronous, if they make network requests or something that wouldn't be good 😨

lighthouse-core/lib/stack-collector.js
```diff
@@ -15,45 +15,51 @@
...
+          name: name,
+          icon: lib.icon,
+          version: result.version,
+          npm: lib.npm,
+        });
+      }
+    } catch (e) {}
+  }
 
   return libraries;
```
patrickhulce: does this mean we always returned an empty array before?

seems important enough to warrant a smoketest and separate fix if so

types/lhr.d.ts
```diff
@@ -103,16 +103,18 @@ declare global {
...
+        /** The unique string ID for this stack pack. */
         id: string;
-        /** The name of the stack pack which we display on our report */
+        /** The title of the stack pack, to be displayed in the report. */
         title: string;
-        /** The base64 url of the icon  */
+        /** A base64 data url to be used as the stack pack's icon. */
         iconDataURL: string;
-        /** An object containing the descriptions of the audits, keyed by the audits' `id` identifier. */
+        /** A set of descriptions for some of Lighthouse's audits, keyed by the audits' `id`. */
```
Hoten:
```suggestion
        /** A set of descriptions for some of Lighthouse's audits, keyed by the audit `id`. */
```

lighthouse-core/lib/stack-collector.js
brendankenny: yah

lighthouse-core/gather/gather-runner.js
brendankenny: > any reason for moving this after after pass?

so we aren't tracing while it runs. The web manifest one can move out of during the trace too, but it has to be run before gatherer's `afterPass`, so we'd need a refactor on `GatherRunner.afterPass`

lighthouse-core/lib/stack-collector.js
brendankenny: > do we get anything out of this

well, first my change makes no sense...`Object.entries` doesn't return an async iterator :) We really only need an `async` on `detectLibraries` and to ditch the inner function calls.

We do get something out of being async. I believe only the `workbox` test is async, but since the function isn't currently async and waiting for the test to return, we don't detect workbox on test pages (though that fact is currently only available in the hidden `js-libraries` audit).

> if they make network requests or something that wouldn't be good

they don't *today*, and hopefully they won't in the future. The workbox one is async because it's [going through the SW APIs to get the SW source file](https://github.com/johnmichel/Library-Detector-for-Chrome/blob/68a527631db0a256932b001e17e8e173e591b55a/library/libraries.js#L1400).

lighthouse-core/lib/stack-collector.js
brendankenny: > does this mean we always returned an empty array before?

when it was `JSLibraries? Yes.

> seems important enough to warrant a smoketest and separate fix if so

an artifact smoketest would be excellent (we have an audit smoke test with the `js-libraries` audit in dbw picking up jquery), but not sure what you mean about fix. This still returns an empty array when nothing is found?

types/lhr.d.ts
brendankenny: > audit

done

lighthouse-core/lib/stack-collector.js
patrickhulce: ah ok gotcha 👍

lighthouse-core/gather/gather-runner.js
patrickhulce: maybe add a comment to this effect? it's hard to remember all the steps that happen inside those when glancing at this section :)

lighthouse-core/lib/stack-collector.js
patrickhulce: I'm confused given your two answers as they seem mutually exclusive :)

Given that we're not `Promise.all` and `.map`'ing I'm surprised that the previous code was working at all, so I asked "does this mean we always returned an empty array before?" to which you say "Yes"

Yet our smoketest for this was still passing and you believe there's no fix necessary, so it seems like it wasn't always returning an empty array.

I guess the answer is that because our evaluateAsync Promise wrapping logic in driver waits for all microtasks to complete, it's sort of coincidence that we were still getting all of the synchronously computed ones by  `.push`'ing to the array that was returned after the fact...

Crazy times 😆

brendankenny: ohhh, I understand what you were asking. I thought you meant in the case where there were no libraries detected did we always return an array, even if empty :)

Yes, it was buggy but accidentally working *except* for the `workbox` test promise, which resolved enough microtasks later (+ whatever else SW infrastructure was doing) that `evaluateAsync` had already returned the array.

PR #8190, opened 2019-04-11
tests: add artifact snapshot testing

patrickhulce:
**Summary**
In addition to @mattzeunert's PR on smoketesting artifacts (#8044) I wanted to offer another proposal for artifact testing that would hopefully lower the barrier to entry and provide us with more robust assertion capabilities as we begin to expose our artifacts as a public API.

This PR adds a way to snapshot test a subset of our artifacts from within jest. Because it's jest-based we can also make any additional assertions we might want.

The primary limitation here is that we are only loading dobetterweb and not all the different nuances of the smoketest pages, so I think each still has their place. This strikes a different balance with the level of detail in our assertions.

**Related Issues/PRs**
#8044


patrickhulce:
> would these tests be useful to run in LR via smokerider?

Somewhat, but I'm not terribly concerned about the differences here. The primary value of these IMO is flagging when we accidentally break the primary gatherers without realizing we're making a breaking change :)

lighthouse-cli/test/cli/artifacts-test.js
```diff
@@ -0,0 +1,87 @@
...
+  'ScriptElements',
+];
+
+/** @type {Array<[(path: string[], value: any) => boolean, (value: any) => any]>} */
+const cleaningFunctions = [
+  [path => path[0] === 'requestId', () => '<requestId>'],
+  [path => path[0] === 'content', value => value && value.slice(0, 100)],
+  [(_, value) => /localhost:\d+/.test(value), value => value.replace(/localhost:\d+/, 'localhost')],
+  [(_, value) => /^blob:/.test(value), () => '<blob url>'],
+];
```
brendankenny: can you add a comment (or comments) for each of these

lighthouse-cli/test/cli/artifacts-test.js
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  if (Array.isArray(value)) {
+    value = value.map((childValue, index) => {
+      const childKey = index.toString();
+      return cleanArtifactsForSnapshot(childValue, [childKey, ...pathToValue]);
+    });
+  } else if (typeof value === 'object' && value !== null) {
+    Object.entries(value).forEach(([childKey, childValue]) => {
+      value[childKey] = cleanArtifactsForSnapshot(childValue, [childKey, ...pathToValue]);
```
brendankenny: applying this recursively definitely seems like it could get fiddly and hard to maintain over time (at least there are snapshots to make sure you aren't deleting someone else's test). e.g. how do I remove `score` over here but not for everyone else or whatever.

patrickhulce: > how do I remove score over here but not for everyone else or whatever.

for posterity, the answer to this question is that you can check the complete path of the value :)

lighthouse-cli/test/cli/artifacts-test.js
patrickhulce: sure thing if we want to proceed!

patrickhulce:
> In that vein, I'm not sure how this realllllllly differs from #8044

If we move all our smoke tests over to jest assertions then there is no difference :)

I thought of this as a first step toward that overall goal, but if we are ready to make the jump now that works for me! 😍

brendankenny:
@patrickhulce, close in favor of the approach in #8962?

patrickhulce:
yeah if  #8962 is happening 😍

brendankenny:
> yeah if #8962 is happening 😍

well, *I* like where it's going, and everyone else seemed to like it or not say anything, so... :)

PR #8202, opened 2019-04-12
cli: print ansii lighthouse on --version

Hoten:
:)

image

I made the ansii with [img2txt.py](https://github.com/hit9/img2txt):
```
img2txt.py ~/Desktop/lighthouse-logo.png --ansi --maxLen=40 --targetAspect=.5
```


Hoten:
or...

dithered
image

a different img:

image

brendankenny:
my vote is no, FWIW :P

I'd find it really annoying if just trying to get the version :)

brendankenny:
(if people realllllllly want it, maybe we could do it with a secret `--i-can-be-your-lighthouse` or whatever flag :)

Hoten:
fair enough :)

I put it in the --version because it's one of the few (maybe the only, i dont recall) fields in `yargs` that does not preprocess the text.

Issue #8185, opened 2019-04-11
Report redesign for v5


connorjclark:
Design reference: https://yuinchien.github.io/sandbox/lighthouse/

- [x] gauges
- - [x] light fill color (#8121)
- - [x] bigger (#8121)
- - [x] separator removed between normal categories and the PWA (#8121)
- - [x] plugin icon (#8307)
- - [x] gauge order: default, pwa, plugins (#8529)
- [x] score scale (#8121, #8590)
- - [x] center
- - [x] round border
- - [x] reverse order
- [x] opportunities
- - [x] hide savings estimate bar in mobile (#8562)
- [x] audit results
- - [x] sections have a line separator (#8121)
- - [x] category description is centered and larger (#8121)
- - [x] bold audit group title with inline description (#8222 #8592)
- - [x] remove audit result numbering (#8222)
- - [x] change audit count to be grouped with title (#8222)
- [x] move score icons to the left (in metrics, opportunities, audit results) (#8222)
- [x] colors (#8222)
- [x] icons
- - [x] use simple shapes for score icons (#8222)
- - [x] green PWA icons (#8532)
- - [x] remove icons from audit group headers (#8329)
- - [ ] doublecheck gray circle for resource summary audit (all informative)
- [ ] fonts
- [x] metric description toggle
- [x] whitespace (#8531, #8591)
- [x] remove old lighthouse header (#8300)
- [ ] sticky topbar
- - [x] LH icon in top left (#8300)
- - [x] followed by url (#8300)
- - [x] score gauge sticky header
- - - [x] appears when scrolling past score scale (#8524)
- - - [x] current category in view is underlined (#8524)
- - - [x] clicking category scrolls to it (#8720)
- [ ] dark mode 
- - [x] styles (#8425)
- - [ ] persist
- [ ] clients
- - [ ] Devtools
- - [x] PSI
- [x] runtime settings (#8229)
- - [x] alternating background colors for rows
- - [x] style like a table
- [x] browser compatibility (#8653)
- - [x] stable Firefox
- - [x] stable Safari
- - [x] Edge 17/18
- [ ] cleanup

Likely punting till after I/O

- [ ] logo
- [ ] run config
- [ ] error messaging
- [x] fireworks (#8660)
- [ ] full bleed thumbnail images

Feel free to edit with anything else.

Hoten:
> Likely punting till after I/O
> > logo

@paulirish hmmmm i did kill it already. ok?

[...]

PR #8222, opened 2019-04-12
report(redesign): score icons, audit group headers, colors

Hoten:
Design reference: https://yuinchien.github.io/sandbox/lighthouse/

Some questions that came up:

* PWA badge icons (the ones in the audit group headers) - do we still want the grayed version when incomplete, and the new green versions when completed? Also, am missing the green icon for PWA Optimized
* want to confirm that we are removing the clump icons (check mark/mangnifying glass/dash-in-circle)

Will fix tests after comments on design.

#8185


Hoten:
(just noticed I need to add back the description for clumps when expanded, ex: "Run these additional validators on your site to check additional SEO best practices.")

patrickhulce:
looks good! haven't taken an deep look yet but the text color jumped out, is the split between icon and text colors for color contrast purposes in an upcoming PR?

Hoten:
ah, did not realize those were different colors. I did a lot of this in low lighting haha

brendankenny:
I thought we were going to try to split this work up (mostly) aligned with #8185? And except for "colors" (all of them? some of them?), none of the things listed match up to anything in that original list :) It could use some sub-bullet points, probably.

For a PR like this, it seems like it could work as three smaller changes (even if one or two are dependent on the other). At least for me, with CSS especially, it's a *lot* easier to see why something is being changed or moved if I don't have to figure out which of several things it's being changed to fix.

Hoten:
I'm having difficulty working like that. Is the planned clean up phase after all this enough of a repentance? Do you prefer me to start over?

brendankenny:
I mean, it doesn't have to be divided up as in #8185. That's just the list @paulirish put together as a conceptual breakdown of what needs to be addressed; if it makes more sense to divide the work up differently based on actual implementation constraints, we should absolutely do it that way.

And I'm not trying to suggest some heavyweight process here. I just mean listing something like, ok clumps need:
- remove icons
- restyle title
- change audit count to be grouped with title
- remove audit numbering
- margin/padding changes
- and whatever else

Once it's a list, paste it into #8185 and then make the changes in 2 or 3 or 5 PRs, depending on how related and involved the changes are, and refer back to the list in the PR descriptions.

Ideally it's just a reflection of the mental work you have to do anyway to translate from the design to what needs to be changed.

> Do you prefer me to start over?

No, I think we should definitely keep moving forward. We should work on this for the changes that will build on top of the existing PRs.

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -54,22 +54,20 @@ class CategoryRenderer {
...
   }
 
   /**
    * Populate an DOM tree with audit details. Used by renderAudit and renderOpportunity
    * @param {LH.ReportResult.AuditRef} audit
-   * @param {number} index
    * @param {DocumentFragment} tmpl
    * @return {Element}
    */
-  populateAuditValues(audit, index, tmpl) {
```
exterkamp: Yay no more numbering!

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -293,16 +285,12 @@ class CategoryRenderer {
     const headerEl = this.dom.find('.lh-audit-group__header', clumpElement);
     const title = this._clumpTitles[clumpId];
     headerEl.textContent = title;
-    if (description) {
-      const markdownDescriptionEl = this.dom.convertMarkdownLinkSnippets(description);
-      const auditGroupDescription = this.dom.createElement('div', 'lh-audit-group__description');
-      auditGroupDescription.appendChild(markdownDescriptionEl);
-      clumpElement.appendChild(auditGroupDescription);
-    }
+    // if (description) {
```
exterkamp: remove?

Hoten:
@brendankenny I'll update the list now, and try to bite off smaller pieces for the rest.

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: idk, do we want "Run these additional validators on your site to check additional SEO best practices"??

Hoten:
I'll do the PWA icons in a different PR.

> PWA badge icons (the ones in the audit group headers) - do we still want the grayed version when incomplete, and the new green versions when completed? Also, am missing the green icon for PWA Optimized

@yuinchien

lighthouse-core/report/html/renderer/category-renderer.js
exterkamp: I think that is extraneous.  I vote remove this entirely.

Hoten: Keeping for now, it may or may not survive design review.

yuinchien:
> I'll do the PWA icons in a different PR.
> 
> > PWA badge icons (the ones in the audit group headers) - do we still want the grayed version when incomplete, and the new green versions when completed? Also, am missing the green icon for PWA Optimized
> 
> @yuinchien

@Hoten Please see icons in the zip file. Let me know if you need any other assets!
[Archive.zip](https://github.com/GoogleChrome/lighthouse/files/3086119/Archive.zip)

Hoten:
thanks @yuinchien!

@paulirish @brendankenny I've fixed the failing tests and lints for this PR.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -165,7 +165,9 @@ describe('CategoryRenderer', () => {
 
     assert.ok(pwaCategory.manualDescription);
     const description = categoryDOM
-      .querySelector('.lh-clump--manual .lh-audit-group__description').textContent;
+      .querySelector('.lh-clump--manual .lh-audit-group__description').textContent
+      // remove " — "
+      .substring(3);
```
patrickhulce: nit: this feels a tad brittle, maybe a regex match or something at least?

lighthouse-core/report/html/report-styles.css
```diff
@@ -99,17 +96,19 @@
   --color-red: #FF4E42;
   --color-white: #FFFFFF;
   --header-padding: 20px 0 20px 0;
+  --polygon-square: polygon(6% 6%, 94% 6%, 94% 94%, 6% 94%);
```
patrickhulce: 😮 TIL

lighthouse-core/report/html/report-styles.css
```diff
@@ -811,7 +812,7 @@
 
 .lh-warnings {
   --item-margin: calc(var(--body-line-height) / 6);
-  border: 1px solid var(--average-color);
+  border: 1px solid var(--color-orange);
```
patrickhulce: what's everyone's opinions on color variable names?

their logical usage like `average-color` or what they are like `black-400`?

Hoten: I just copied these from yuinchen's design, but I'll share what I think about them:

* starting with `color-` is great for autocomplete
* `color-xxx` is neat, IDK what the scale represents tho.
* using `orange` over `average` is good for quick-understanding, but bad for theming (do we care?)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: done

lighthouse-core/report/html/report-styles.css
yuinchien: @Hoten FYI, the color scale is from material color system: https://material.io/design/color/the-color-system.html#tools-for-picking-colors. 500 is the primary, above is darker, below is lighter.

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -193,16 +190,17 @@ class CategoryRenderer {
...
 
+    const auditGroupHeader = this.dom.createElement('div', 'lh-audit-group__header');
+
+    this.dom.createChildOf(auditGroupHeader, 'span', 'lh-audit-group__title')
+      .textContent = group.title;
     if (group.description) {
-      const auditGroupDescription = this.dom.createElement('div', 'lh-audit-group__description');
-      auditGroupDescription.appendChild(this.dom.convertMarkdownLinkSnippets(group.description));
-      groupEl.appendChild(auditGroupDescription);
+      const descriptionEl = this.dom.convertMarkdownLinkSnippets(` — ${group.description}`);
```
brendankenny: what's the i18n call on these?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -292,17 +288,15 @@ class CategoryRenderer {
...
-      clumpElement.appendChild(auditGroupDescription);
+      const descriptionEl = this.dom.convertMarkdownLinkSnippets(` — ${description}`);
+      descriptionEl.classList.add('lh-audit-group__description');
+      headerEl.appendChild(descriptionEl);
     }
 
     const itemCountEl = this.dom.find('.lh-audit-group__itemcount', clumpElement);
-    // TODO(i18n): support multiple locales here
-    itemCountEl.textContent = `${auditRefs.length} audits`;
+    itemCountEl.textContent = `(${auditRefs.length})`;
```
brendankenny: i18n here too

lighthouse-core/report/html/report-styles.css
```diff
@@ -478,47 +472,45 @@
...
 .lh-metric--average .lh-metric__value {
-  color: var(--average-color);
-  padding-left: 16px;
+  color: var(--color-orange-700);
 }
-.lh-metric--average .lh-metric__value::after {
-  background: var(--average-icon-url) no-repeat 50% 50%;
+.lh-metric--average .lh-metric__innerwrap::before {
+  background: var(--color-orange);
+  clip-path: var(--polygon-square);
```
brendankenny: is this going to be a browser compat issue?

lighthouse-core/report/html/report-styles.css
brendankenny: I'm happy to defer to others on this, but to me `--average-color` tells me how to use it, while `--color-orange` is something I'll have to look up when I'm trying to figure out what to use for the average color.

There's limits to the expressiveness of variable names (who knows what `--report-secondary-border-color` really refers to :), but it does seem useful to learn more towards how it's used rather than what it contains.

lighthouse-core/report/html/renderer/category-renderer.js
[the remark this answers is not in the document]
Hoten: done

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -193,16 +190,17 @@ class CategoryRenderer {
    */
   renderAuditGroup(group) {
     const groupEl = this.dom.createElement('div', 'lh-audit-group');
-    const summaryEl = this.dom.createChildOf(groupEl, 'div');
-    const summaryInnerEl = this.dom.createChildOf(summaryEl, 'div', 'lh-audit-group__summary');
-    const headerEl = this.dom.createChildOf(summaryInnerEl, 'div', 'lh-audit-group__header');
 
+    const auditGroupHeader = this.dom.createElement('div', 'lh-audit-group__header');
```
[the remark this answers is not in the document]
Hoten: > In this case I think i'd love it if we just leave a comment like

done

i may try combining these two into the sample template later, in the clean up phase.

lighthouse-core/report/html/report-styles.css
```diff
@@ -732,27 +716,44 @@
   margin-bottom: var(--section-padding);
 }
 
+.lh-audit-group__caption {
```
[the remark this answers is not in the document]
Hoten: trash

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: @exterkamp help

lighthouse-core/report/html/renderer/category-renderer.js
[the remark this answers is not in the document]
brendankenny: > i may try combining these two into the sample template later, in the clean up phase.

I was serious about the original TODO :)

>  TODO: group classes shouldn't be reused for clumps.

They don't need to be the same thing, and we don't win anything by creating a grand unified clump-group theorem except a new headache for every point where they're almost but not quite the same thing :P

Seriously, we'll save like two total lines of code and gain some extra conditionals and a bunch of head scratching later about why groups are sometimes what we call groups everywhere else and why they're sometimes something completely different.

Rather than make a comment here, the clean up phase should just `s/group/clump` where that original comment is and be done with it 👍 👍

lighthouse-core/report/html/report-styles.css
Hoten: havent tested in Edge, but apparently Edge support `clip-path` but only for svg urls. could transition to that.

Hoten: err.. but apparently safari has no support.

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: idk. @exterkamp

lighthouse-core/report/html/renderer/category-renderer.js
exterkamp: This one looks A-Okay because the final output is all numeric, correct?  
"17 Audits" becomes "(17)" <- this should be fine.

The concern here is non latin numerics, like ar:
"(17)" -ar-> "(١٧)"

But afaik we already don't support that yet in LH or PSI consistently so I'd leave in the i18n TODO, but LGTM for now.

lighthouse-core/report/html/renderer/category-renderer.js
exterkamp: This seems fine for LTR.  But seems like a landmine for RTL.  But it actually seems to work itself out: 
image

So LGTM?  I think that we need to keep this in mind for an eventual RTL rework, but looks fine for now.  Put in an i18n TODO here though to be safe I'd say.

brendankenny: what if we avoid at least the string concatenation and let the DOM handle it by also having the dash in a `<span>` in addition to the title and description spans?

(that should also help with the tests a bit)

Hoten: done

exterkamp:
Any idea why this is happening?
image
Seems to take the last character and put it first?

Hoten:
> Any idea why this is happening?

no :(

lighthouse-core/report/html/report-styles.css
paulirish: yup safari can't do the triangle. Let's just fall back to normal css triangles? 
squares and circles are easy enough to do on their own.. i guess we miss out on using polygon() this time. :(

lighthouse-core/report/html/report-styles.css
Hoten: ya i'm with ya on this. done.

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -193,16 +190,18 @@ class CategoryRenderer {
    */
   renderAuditGroup(group) {
     const groupEl = this.dom.createElement('div', 'lh-audit-group');
-    const summaryEl = this.dom.createChildOf(groupEl, 'div');
-    const summaryInnerEl = this.dom.createChildOf(summaryEl, 'div', 'lh-audit-group__summary');
-    const headerEl = this.dom.createChildOf(summaryInnerEl, 'div', 'lh-audit-group__header');
 
+    // Group header DOM roughly matches the clump header DOM, and that's OK.
```
brendankenny: should remove this comment. See reasoning in earlier thread, but, more importantly, this wouldn't be the place to put the comment anyways :)

This group header is using `group` class names, nothing weird about that or worth commenting on; it's the *clump* header using `group` class names that's the aberration.

lighthouse-core/report/html/report-styles.css
```diff
@@ -113,11 +110,21 @@
   --score-title-line-height-big: 36px;
   --score-title-line-height: 26px;
   --section-padding: 40px;
+  --shape-margin: 7px 12px 0 4px;
+  --shape-size: 12px;
```
brendankenny: this would probably be a little more helpful with an additional qualifier in the variable name. Which shape? (same with `circle` and `square`?)

Hoten: it's used for both metric icons and audit score icons so I went with `--icon-shape-size`...

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: done

lighthouse-core/test/report/html/renderer/psi-test.js
```diff
@@ -118,15 +118,16 @@ describe('DOM', () => {
         const {perfCategoryEl} = prepareLabData(sampleResultsStr, document);
         const metricsGroupEl = perfCategoryEl.querySelector('.lh-audit-group--metrics');
 
+        const expectedDescription = Util.UIStrings.lsPerformanceCategoryDescription
+          // Replacing markdown because ".textContent" will be post-markdown.
+          .replace('[Lighthouse](https://developers.google.com/web/tools/lighthouse/)', 'Lighthouse');
+
         // Assume using default locale.
```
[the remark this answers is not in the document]
Hoten: done

Hoten:
Let's defer the RTL to a follow up PR. Everything else look swell?

lighthouse-core/report/html/report-styles.css
brendankenny: > it's used for both metric icons and audit score icons so I went with `--icon-shape-size`...

"icon" and "shape" mean basically the same thing :)

what about `--score-shape-size`, etc?

lighthouse-core/report/html/report-styles.css
```diff
@@ -152,6 +160,7 @@
     --score-title-line-height-big: 26px;
     --score-title-line-height: 20px;
     --section-padding: 24px;
+    --shape-margin: 5px 12px 0 2px;
```
brendankenny: `--score-shape-margin` here too?

PR #8228, opened 2019-04-12
report: extract init() function from report-template

paulirish:
Providing a hook in case one uses the report-generator from a webapp (after DCL has fired).


lighthouse-core/report/html/report-template.html
```diff
@@ -36,7 +36,7 @@
   </script>
   <script>window.__LIGHTHOUSE_JSON__ = %%LIGHTHOUSE_JSON%%;</script>
   <script>
-    window.addEventListener('DOMContentLoaded', _ => {
+    function __initLighthouseReport__(){
```
brendankenny:
```suggestion
    function __initLighthouseReport__() {
```
were you raised in a barn

PR #8192, opened 2019-04-11
Report ui runtime settings

Hoten:
Branched off of #8121. Will rebase off master when that is merged.

See: https://lh-pr-report-redesign.surge.sh/5

#8185

PR #8229, opened 2019-04-13
report(redesign): runtime settings footer

Hoten:
messed up #8192


Hoten:
this was a silly time. Should be good now.

lighthouse-core/report/html/report-styles.css
```diff
@@ -98,13 +101,14 @@
   --color-red-700: #EB0F00;
   --color-red: #FF4E42;
   --color-white: #FFFFFF;
+  --env-name-min-width: 220px;
   --header-padding: 20px 0 20px 0;
+  --item-bg: var(--color-black-100);
+  --item-padding: 10px 0px;
```
paulirish: i presume `item` isn't specific to the env stuff?

Hoten: just checked, it's only uses for the env table. renamed.

Hoten:
brb deleting `git-rebase` from my machine

lighthouse-core/report/html/report-styles.css
```diff
@@ -83,6 +83,9 @@
   --circle-size: calc(3 * var(--header-font-size));
 
   /* TODO(cjamcl) clean up unused variables. */
+  --audits-margin-bottom: 40px;
+  --circle-size-big: 120px;
+  --circle-size: 96px;
```
patrickhulce: ~~what are these doing here/should we remove the `circle-size` above it?~~

oh looks like they're just alphabetized now :) still remove the one duplicate above it?

Hoten:
linking to #8185

Hoten:
@paulirish look good?

paulirish:
Yup

PR #8243, opened 2019-04-13
misc: update sample_v2 artifacts

brendankenny:
I started writing a test that needed the `settings` on the artifacts that generate `sample_v2.json` and I noticed they weren't there. Did a little housecleaning on the easy to update artifacts (deleting `EventListeners` and `WebSQL`, finally adding `RobotsTxt` and `TapTargets`), though at some point (maybe for v5) we should still do a whole-hog update of the artifacts, trace, and devtools log.

At that point, surprisingly `sample_v2.json` still wasn't a valid `LH.Result`, which isn't super important but is kind of annoying. To fix we just had to
- not delete `lhr.configSettings.auditMode` in the cleanup script (just hardcoded to `true` to sidestep the travis issue mentioned in the comment in that file) and
- `Artifacts.MeasureEntry` (which makes up the `Timing` artifact) needed to not inherit from `PerformanceEntry`. `PerformanceEntry` has a [`toJSON` method](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry#Methods), but when we save artifacts as JSON methods aren't preserved, so the resulting object is no longer a `PerformanceEntry`. I just copied the properties over to the type definition rather than inherit now.

### DIFF
--- a/lighthouse-core/scripts/cleanup-LHR-for-diff.js
+++ b/lighthouse-core/scripts/cleanup-LHR-for-diff.js
@@ -26,10 +26,10 @@ function cleanAndFormatLHR(lhrString) {
   /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
 
-  // TODO: Resolve the below so we don't need to delete it:
+  // TODO: Resolve the below so we don't need to force it to a boolean value:
   // 1) The string|boolean story for proto
   // 2) Travis gets a absolute path during yarn diff:sample-json
-  delete lhr.configSettings.auditMode;
+  lhr.configSettings.auditMode = true;
 
   // Set timing values, which change from run to run, to predictable values
   lhr.timing.total = 12345.6789;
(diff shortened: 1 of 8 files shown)


lighthouse-core/scripts/cleanup-LHR-for-diff.js
```diff
@@ -26,10 +26,10 @@ function cleanAndFormatLHR(lhrString) {
   /** @type {LH.Result} */
   const lhr = JSON.parse(lhrString);
 
-  // TODO: Resolve the below so we don't need to delete it:
+  // TODO: Resolve the below so we don't need to force it to a boolean value:
   // 1) The string|boolean story for proto
   // 2) Travis gets a absolute path during yarn diff:sample-json
-  delete lhr.configSettings.auditMode;
+  lhr.configSettings.auditMode = true;
```
brendankenny: this is valid, it's just as if we were running audit mode from the default location instead of from the saved artifacts

lighthouse-viewer/test/viewer-test-pptr.js
```diff
@@ -133,10 +133,6 @@ describe('Lighthouse Viewer', function() {
     const errorSelectors = '.lh-audit-explanation, .tooltip--error';
     const auditErrors = await viewerPage.$$eval(errorSelectors, getErrors, selectors);
     const errors = auditErrors.filter(item => item.explanation.includes('Audit error:'));
-    const unexpectedErrrors = errors.filter(item => {
-      return !item.explanation.includes('Required RobotsTxt gatherer did not run') &&
-        !item.explanation.includes('Required TapTargets gatherer did not run');
-    });
-    assert.deepStrictEqual(unexpectedErrrors, [], 'Audit errors found within the report');
```
brendankenny: no longer necessary :)

PR #8294, opened 2019-04-15
extension: remove github link in favor of copy to clipboard

wardpeet:
**Summary**
Remove the github issue creation link and move to a clipboard button instead.

lighthouse-copy

<details>
<summary>The following text is copied into the clipboard:</summary>

**Lighthouse Version**: 4.3.0
**Lighthouse Commit**: f46bcd78779ca5ce6052d6655a15a22d5f264478
**Chrome Version**: 73.0.3683.103
**Initial URL**: https://www.paulirish.com/
**Error Message**: Attempted to add event listener after connection disposed.
**Stack Trace**:
```
Error: Attempted to add event listener after connection disposed.
    at ExtensionConnection.on (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:20743:7)
    at new Driver (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:21189:12)
    at Function._gatherArtifactsFromBrowser (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:32187:37)
    at Function.run (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:32062:24)
    at lighthouse (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:23321:15)
    at runLighthouseInExtension (chrome-extension://kolnikflpalakcdkahhcfkiahbhfjfod/scripts/lighthouse-ext-bundle.js:14595:20)
``` 
</details>


clients/extension/scripts/popup.js
```diff
@@ -176,13 +177,13 @@ async function onGenerateReportButtonClick(background, settings) {
     // Close popup once report is opened in a new tab
     window.close();
   } catch (err) {
-    let message = err.message;
+    let message = err.friendlyMessage || err.message;
```
patrickhulce: nice this is what we get the protocol timeout method now right? :)

clients/extension/scripts/popup.js
```diff
@@ -176,13 +177,13 @@ async function onGenerateReportButtonClick(background, settings) {
...
     let includeReportLink = true;
 
     // Check for errors in how the user ran Lighthouse and replace with a more
     // helpful message (and remove 'Report Error' link).
     for (const [test, replacement] of Object.entries(NON_BUG_ERROR_MESSAGES)) {
-      if (message.includes(test)) {
+      if (err.message.includes(test)) {
```
patrickhulce: I think some of our tests actually test our friendly message, maybe we should fail on either?

wardpeet: Oh, it's to replace messages to something more descriptive and remove the button as it's more or less a user issue or a chrome issue.

```
const NON_BUG_ERROR_MESSAGES = {
  'Another debugger': 'You probably have DevTools open. Close DevTools to use Lighthouse',
  'multiple tabs': 'You probably have multiple tabs open to the same origin. ' +
      'Close the other tabs to use Lighthouse.',
  // The extension debugger API is forbidden from attaching to the web store.
  // @see https://chromium.googlesource.com/chromium/src/+/5d1f214db0f7996f3c17cd87093d439ce4c7f8f1/chrome/common/extensions/chrome_extensions_client.cc#232
  'The extensions gallery cannot be scripted': 'The Lighthouse extension cannot audit the ' +
      'Chrome Web Store. If necessary, use the Lighthouse CLI to do so.',
  'Cannot access a chrome': 'The Lighthouse extension cannot audit ' +
      'Chrome-specific urls. If necessary, use the Lighthouse CLI to do so.',
  // The user tries to review an error page or has network issues
  'Unable to load the page': 'Unable to load the page. Please verify the url you ' +
      'are trying to review.',
  'Cannot access contents of the page': 'Lighthouse can only audit URLs that start' +
      ' with http:// or https://.',
  'INSECURE_DOCUMENT_REQUEST': 'The URL you have provided does not have valid' +
      ' security credentials.',
  'INVALID_URL': 'Lighthouse can only audit URLs that start' +
      ' with http:// or https://.',
};
```

I can also check for for friendly but don't think we need it at this point but doesn't harm

clients/extension/scripts/popup.js
wardpeet: Yes, now we get these nice messages:
Waiting for DevTools protocol response has exceeded the allotted time. (Method: Browser.getVersion)

clients/extension/scripts/popup.js
patrickhulce: Oh yeah I meant here that some of our string includes I think are looking for strings that are only in the friendly message, so we might want to do something like `err.message.includes(test) || err.friendlyMessage.includes(test)`. 

I just checked all of them though, and I don't think this is actually the case. Most of them are direct from Chrome/extension-specific, some look old and no longer applicable.

Even though we are never showing the file bug button anymore I still think some of the instructions here are worth keeping 👍 

Proposal:

- Replace 'Unable to load the page' test with 'ERRORED_DOCUMENT_REQUEST' and 'FAILED_DOCUMENT_REQUEST' and 'DNS_FAILURE'
- Reorder so 'multiple tabs' and our 'ERROR_CODE*' tests are first and add a comment that the rest are issued by chrome extension API.
- Reword 'INSECURE_DOCUMENT_REQUEST' to say 'valid SSL certificate'?

I know these are only partially related to the change but I'm glad we uncovered them! :D

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
+ * @return {HTMLButtonElement}
  */
 function buildReportErrorLink(err) {
   const issueBody = `
 **Lighthouse Version**: ${getLighthouseVersion()}
 **Lighthouse Commit**: ${getLighthouseCommitHash()}
 **Chrome Version**: ${getChromeVersion()}
 **Initial URL**: ${siteURL}
-**Error Message**: ${err.message}
+**Error Message**: ${err.friendlyMessage || err.message}
```
wardpeet: the original error (err.message) is still shown in err.stack.

clients/extension/scripts/popup.js
wardpeet: - [x] "Replace 'Unable to load the page' test with 'ERRORED_DOCUMENT_REQUEST' and 'FAILED_DOCUMENT_REQUEST' and 'DNS_FAILURE'"
- [x] Reorder so 'multiple tabs' and our 'ERROR_CODE*' tests are first and add a comment that the rest are issued by chrome extension API.
  * I did it the way around, put the regular errors at the top and chrome stuff at the bottom
- [x] Reword 'INSECURE_DOCUMENT_REQUEST' to say 'valid SSL certificate'?
  * Did it on both places inside lh-errors and extension.

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
-  url.searchParams.append('title', `Extension Error: ${errorTitle}`);
-  url.searchParams.append('body', issueBody.trim());
-
-  const reportErrorEl = document.createElement('a');
+  const reportErrorEl = document.createElement('button');
   reportErrorEl.className = 'button button--report-error';
-  reportErrorEl.href = url.href;
-  reportErrorEl.textContent = 'Report Error';
-  reportErrorEl.target = '_blank';
+  reportErrorEl.textContent = 'Copy to clipboard';
```
Hoten: at the first glance, the current copy suggests it would only copy the provided error message to the clipboard. Not obvious there are more details if only you copy to the clipboard.

"Copy details to clipboard"?

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
 
 /**
  * @param {Error} err
- * @return {HTMLAnchorElement}
+ * @return {HTMLButtonElement}
  */
 function buildReportErrorLink(err) {
```
brendankenny:
```suggestion
function buildErrorCopyButton(err) {
```
maybe?

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
     `;
 
-  const url = new URL('https://github.com/GoogleChrome/lighthouse/issues/new');
-
-  const errorTitle = err.message.substring(0, MAX_ISSUE_ERROR_LENGTH);
-  url.searchParams.append('title', `Extension Error: ${errorTitle}`);
-  url.searchParams.append('body', issueBody.trim());
-
-  const reportErrorEl = document.createElement('a');
+  const reportErrorEl = document.createElement('button');
```
brendankenny:
```suggestion
  const errorButtonEl = document.createElement('button');
```
maybe?

clients/extension/scripts/popup.js
brendankenny: I like `Copy details to clipboard` (or something like it) as well (do we also want a copy to cliboard icon?)

lighthouse-core/lib/lh-error.js
```diff
@@ -22,7 +22,7 @@ const UIStrings = {
   /** Error message explaining that Lighthouse could not load the requested URL and the steps that might be taken to fix the unreliability. */
   pageLoadFailedWithDetails: 'Lighthouse was unable to reliably load the page you requested. Make sure you are testing the correct URL and that the server is properly responding to all requests. (Details: {errorDetails})',
   /** Error message explaining that the credentials included in the Lighthouse run were invalid, so the URL cannot be accessed. securityMessages will be replaced with one or more strings from the browser explaining what was insecure about the page load. */
-  pageLoadFailedInsecure: 'The URL you have provided does not have valid security credentials. {securityMessages}',
```
brendankenny: revert?

clients/extension/scripts/popup.js
brendankenny: this will have to be ts-ignored or cast (we're saying it's an `Error` while `friendlyMessage` is only on `LighthouseError`)

Or we could pass in `message` as a string parameter since the caller has already figured that out.

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
   reportErrorEl.className = 'button button--report-error';
-  reportErrorEl.href = url.href;
-  reportErrorEl.textContent = 'Report Error';
-  reportErrorEl.target = '_blank';
+  reportErrorEl.textContent = 'Copy to clipboard';
+
+  reportErrorEl.addEventListener('click', async () => {
+    try {
+      // @ts-ignore
```
brendankenny:
```suggestion
      // @ts-ignore - tsc doesn't include `clipboard` on `navigator`.
```

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
+      await navigator.clipboard.writeText(issueBody);
+      reportErrorEl.textContent = 'Copied to clipboard';
+      setTimeout(() => {
```
brendankenny: I wasn't sure at first what this was for, so maybe add a comment saying something like `// Return button to inviting state after timeout.`

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
+        reportErrorEl.textContent = 'Copy to clipboard';
+      }, 1000);
+    } catch (err) {
+    }
```
brendankenny: do we need a `try/catch` since this is only in the extension UI, so will always be Chrome only?

If we do need it, we should have a `catch` fallback action, e.g. set the button text to `'Cannot copy to clipboard'` and disable the button?

clients/extension/scripts/popup.js
```diff
@@ -80,32 +82,36 @@ function find(query, context = document) {
...
-  reportErrorEl.target = '_blank';
+  reportErrorEl.textContent = 'Copy to clipboard';
+
+  reportErrorEl.addEventListener('click', async () => {
+    try {
+      // @ts-ignore
+      await navigator.clipboard.writeText(issueBody);
+      reportErrorEl.textContent = 'Copied to clipboard';
+      setTimeout(() => {
+        reportErrorEl.textContent = 'Copy to clipboard';
```
brendankenny: should pull this string out to a variable in a higher scope so it can be shared between this and the original setting on line 102

lighthouse-core/lib/lh-error.js
wardpeet: Oh @patrickhulce was suggesting this in an obve comment
> Reword 'INSECURE_DOCUMENT_REQUEST' to say 'valid SSL certificate'?
https://github.com/GoogleChrome/lighthouse/pull/8294#discussion_r275837518

patrickhulce: it seemed like we never really touch these strings and this was the closest PR we'd naturally get :)

patrickhulce: The other changes we're making here that touched the triggering of these is what originally brought this up, but I'm fine to punt if others feel it's sufficiently unrelated.

wardpeet:
It now looks like
image

clients/extension/scripts/popup.js
wardpeet: Fixed in https://github.com/GoogleChrome/lighthouse/pull/8294/commits/84bac693c0ce0d722bd3bf2fa95838afcab258b2

clients/extension/scripts/popup.js
wardpeet: Fixed in https://github.com/GoogleChrome/lighthouse/pull/8294/commits/84bac693c0ce0d722bd3bf2fa95838afcab258b2

clients/extension/scripts/popup.js
wardpeet: I still needed `err.stack` so I'm expecting a LighthouseError object.
Fixed in https://github.com/GoogleChrome/lighthouse/pull/8294/commits/84bac693c0ce0d722bd3bf2fa95838afcab258b2

clients/extension/scripts/popup.js
wardpeet: Fixed in https://github.com/GoogleChrome/lighthouse/pull/8294/commits/84bac693c0ce0d722bd3bf2fa95838afcab258b2

clients/extension/scripts/popup.js
wardpeet: Fixed in https://github.com/GoogleChrome/lighthouse/pull/8294/commits/84bac693c0ce0d722bd3bf2fa95838afcab258b2

clients/extension/scripts/popup.js
wardpeet: I removed the try catch, we'll add it back if we get error reports of this. As I'm unsure when chrome can't write to clipboard

lighthouse-core/lib/lh-error.js
brendankenny: I'd say we probably shouldn't have it in this PR, but we could probably get everybody who needs to weigh in if we really want to :)

It's a translated string, is used as a return value for the PSI API if it runs into an SSL error, etc, so we'd definitely want to get eyes on it before landing it

patrickhulce: aye, sorry @wardpeet I'll open a different PR for this then :)

patrickhulce: wait hang on, I never actually wanted this string to be changed in lh-error in this PR, lol

I was just talking about the one in the detection popup :)

I'll still file a different PR, but let's not revert the extension-specific string?

clients/extension/scripts/popup.js
```diff
@@ -193,7 +201,7 @@ async function onGenerateReportButtonClick(background, settings) {
 
     if (includeReportLink) {
       feedbackEl.className = 'feedback feedback-error';
-      feedbackEl.appendChild(buildReportErrorLink(err));
+      feedbackEl.appendChild(buildErrorCopyButton(/** @type {LighthouseError} */ (err)));
```
brendankenny: Ideally we wouldn't do this cast because we don't actually know this...e.g. even if we someday catch all exceptions inside Lighthouse and only throw LighthouseErrors, if an extension API throws, it'll still be a plain old `Error`.

I just meant something like `function buildErrorCopyButton(friendlyMessage, err) {` so we don't have to do the `friendlyMessage || message` dance again inside `buildErrorCopyButton` when we've already figured it out here.

lighthouse-core/lib/lh-error.js
wardpeet: fixed, no problem :)

brendankenny:
@patrickhulce if you want to look again as well

patrickhulce:
Gah, get out of here googlebot we both have CLAs! 😛

PR #8298, opened 2019-04-15
core: rename diagnostics to debugData

patrickhulce:
**Summary**
All the data here is essentially used for debugging, so `debugData` seems fair.

`metrics` audit is probably the one exception where I'd expect folks to be explicitly relying on this for presentation, so I'm not sure what to do there. OTOH I also don't feel particularly strongly it needs to be renamed away from `diagnostics` since neither of them describe how it's being used in `metrics` :)

**Related Issues/PRs**
No issue was filed specifically, so it can't be closed yet! ;)
#7752

### DIFF
--- a/lighthouse-core/audits/accessibility/axe-audit.js
+++ b/lighthouse-core/audits/accessibility/axe-audit.js
@@ -63,11 +63,11 @@ class AxeAudit extends Audit {
       {key: 'node', itemType: 'node', text: str_(UIStrings.failingElementsHeader)},
     ];
 
-    /** @type {LH.Audit.Details.Diagnostic|undefined} */
+    /** @type {LH.Audit.Details.DebugData|undefined} */
     let diagnostic;
     if (impact || tags) {
       diagnostic = {
-        type: 'diagnostic',
+        type: 'debug',
         impact,
         tags,
       };
@@ -78,7 +78,7 @@ class AxeAudit extends Audit {
       extendedInfo: {
         value: rule,
       },
-      details: {...Audit.makeTableDetails(headings, items), diagnostic},
+      details: {...Audit.makeTableDetails(headings, items), debug: diagnostic},
     };
   }
 }
--- a/lighthouse-core/audits/accessibility/axe-audit.js
+++ b/lighthouse-core/audits/accessibility/axe-audit.js
@@ -64,10 +64,10 @@ class AxeAudit extends Audit {
     ];
 
     /** @type {LH.Audit.Details.DebugData|undefined} */
-    let diagnostic;
+    let debugData;
     if (impact || tags) {
-      diagnostic = {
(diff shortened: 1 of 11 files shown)


lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -63,11 +63,11 @@ class AxeAudit extends Audit {
       {key: 'node', itemType: 'node', text: str_(UIStrings.failingElementsHeader)},
     ];
 
-    /** @type {LH.Audit.Details.Diagnostic|undefined} */
+    /** @type {LH.Audit.Details.DebugData|undefined} */
     let diagnostic;
```
brendankenny: rename variable?

types/audit-details.d.ts
```diff
@@ -51,7 +51,7 @@ declare global {
         overallSavingsBytes?: number;
         headings: OpportunityColumnHeading[];
         items: OpportunityItem[];
-        diagnostic?: Diagnostic;
+        diagnostic?: DebugData;
```
brendankenny: I think all these `diagnostic` need to be renamed?

lighthouse-core/audits/byte-efficiency/uses-long-cache-ttl.js
```diff
@@ -238,18 +238,18 @@ class CacheHeaders extends Audit {
         if (url.includes('?')) queryStringCount++;
 
         // Include cacheControl info (if it exists) per url as a diagnostic.
-        /** @type {LH.Audit.Details.Diagnostic|undefined} */
-        let diagnostic;
+        /** @type {LH.Audit.Details.DebugData|undefined} */
+        let debugData;
```
brendankenny: just `debug`?

types/audit-details.d.ts
```diff
@@ -68,15 +68,15 @@ declare global {
...
 
       /**
        * A details type that is not rendered in the final report; usually used
-       * for including diagnostic information in the LHR. Can contain anything.
+       * for including debug information in the LHR. Can contain anything.
        */
-      export interface Diagnostic {
-        type: 'diagnostic';
+      export interface DebugData {
+        type: 'debug';
```
brendankenny: if we go with type `DebugData` maybe this string should be `debugdata` and the property name `debugData`?

It's a little confusing having both `debug` and `debugData` as reasonable local variable names

patrickhulce: agreed, `debugdata` SGTM :)

types/audit-details.d.ts
patrickhulce: yeah fo sho, didn't expect the first name go around to be accepted :D

paulirish:
name wfm. :+1:

PR #8299, opened 2019-04-15
core(screenshots): match format of final-screenshot

patrickhulce:
**Summary**
Reworks our `screenshot-thumbnails` to match `final-screenshot`. I personally prefer the format used in `screenshot-thumbnails` but I see the value in an immediately consumable data URI without misinterpretation of what its contents are. I played around with just a mime-type property but then that seemed like overkill so I went back to just matching the `data:image/jpeg` format. 

Potentially we might consider a rename of `.data` to `.dataURL` or something at the same time?

**Related Issues/PRs**
#7752


lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -128,8 +129,8 @@ class ScreenshotThumbnails extends Audit {
       }
       thumbnails.push({
         timing: Math.round(targetTimestamp - speedline.beginning),
-        timestamp: targetTimestamp * 1000,
-        data: base64Data,
+        timestamp: targetTimestamp,
```
brendankenny: for traceOfTab we use microseconds for `timestamp`, I think we should keep matching it here (so `final-screenshot.js` will have to move to * 1000

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -128,8 +129,8 @@ class ScreenshotThumbnails extends Audit {
       }
       thumbnails.push({
         timing: Math.round(targetTimestamp - speedline.beginning),
```
brendankenny: should we also add `timing` to `final-screenshot.js` just to completely match?

lighthouse-core/audits/screenshot-thumbnails.js
```diff
@@ -128,8 +129,8 @@ class ScreenshotThumbnails extends Audit {
...
-        timestamp: targetTimestamp * 1000,
-        data: base64Data,
+        timestamp: targetTimestamp,
+        data: `data:image/jpeg;base64,${base64Data}`,
```
brendankenny: yeah, having the prefix makes sense so no one has to guess what type of image it is, even if it's annoying to add it here

lighthouse-core/test/audits/screenshot-thumbnails-test.js
```diff
@@ -30,14 +30,14 @@ describe('Screenshot thumbnails', () => {
         const framePath = path.join(__dirname,
             `../fixtures/traces/screenshots/progressive-app-frame-${index}.jpg`);
         const expectedData = fs.readFileSync(framePath, 'base64');
-        assert.equal(expectedData.length, result.data.length);
+        expect(result.data.length - 'data:image/jpeg;base64,'.length).toEqual(expectedData.length);
```
brendankenny: weird test :)

lighthouse-core/audits/screenshot-thumbnails.js
patrickhulce: yeah good call done

lighthouse-core/test/audits/screenshot-thumbnails-test.js
patrickhulce: heh, I can slice it instead to make it a little more explicit :)

brendankenny: > > weird test :)
>
> heh, I can slice it instead to make it a little more explicit :)

ha, I meant the original test is a weird one, but I like the change :)

lighthouse-core/audits/final-screenshot.js
```diff
@@ -30,18 +31,23 @@ class FinalScreenshot extends Audit {
...
+    const traceOfTab = await TraceOfTab.request(trace, context);
     const screenshots = await Screenshots.request(trace, context);
     const finalScreenshot = screenshots[screenshots.length - 1];
 
     if (!finalScreenshot) {
       throw new LHError(LHError.errors.NO_SCREENSHOTS);
     }
 
+    // trace-of-tab uses microseconds for the timestamp, so we'll convert for consistency
+    const finalScreenshotTs = finalScreenshot.timestamp * 1000;
```
brendankenny: `computed/screenshots.js` is [dividing the timestamp by 1000](https://github.com/GoogleChrome/lighthouse/blob/81a8a5cb3d5e785494fe8549cbe747de2325c17c/lighthouse-core/computed/screenshots.js#L22), and this is the only audit using that computed artifact (did we use it for something else at some point?), so, want to just not divide it there?

patrickhulce: yeah sg, I wanted to kill it at some point for being mostly useless, but that'll have to wait for another day :)

brendankenny:
lol how does that computed artifact have no test for the timestamp :)

it should definitely get culled at some point.

Issue #8159, opened 2019-04-10
Deploy the updated report to the web after every push


paulirish:
We have a bunch of report UI changes upcoming.. the first of which from @hoten updates the gauges (#8121).

[Zeit/now.sh has an integration with GitHub](https://zeit.co/docs/v2/integrations/now-for-github/) nowadays that makes this very straightforward. (Last year, I had a real hacky solution for this, which is no longer necessary yay!)

Please check it out in action here: https://github.com/paulirish/test-lh-for-now/pull/3

* It'll create new deployment items and create a commit status entry (or two) to communicate status of the deployment:
![image](https://user-images.githubusercontent.com/39191/55911874-650bb980-5b96-11e9-9a7f-ff1f330425ee.png)
* Click on `View Deployment` and you'll be brought to something like https://test-lh-for-now-8viv2go5c.now.sh/

These live forever and are free. The report deployment is usually up in ~30s after the commit push. 

Considerations:
1. We can't block the deployment from starting. So we can't really say "only do this on report-affecting PRs". The best we could do is fail a deployment because we don't want it to finish.
1. The GoogleChrome org needs to approve the Now github application.

I think this would be super helpful for review of upcoming report redesign PRs. We could also incorporate stackpacks or plugins if we wanted to always see those too. 

What do folks think? Good stuff or annoying?

patrickhulce:
it's dope and super necessary IMO as we work on total report redesign!

paulirish:
@addyosmani can you help us enable the Now github app for this org? 
I _think_ this is the url: https://github.com/organizations/GoogleChrome/settings/installations

paulirish:
Feature requests:

* PSI UI
* Other reports with different states.. 
  * cnn, lots-of-errors.html

connorjclark:
now is now now

I think we can close this. improving what we put on now will be an ongoing endeavor, hopefully. (like #10151)

PR #8307, opened 2019-04-16
report: plugin badge for plugin score gauges

Hoten:
#8185

image


lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -331,6 +331,10 @@ class CategoryRenderer {
     wrapper.href = `#${category.id}`;
     wrapper.classList.add(`lh-gauge__wrapper--${Util.calculateRating(category.score)}`);
 
+    if (!['performance', 'accessibility', 'best-practices', 'seo', 'pwa'].includes(category.id)) {
```
exterkamp: @brendankenny is this the best way to detect plugins?  Seems _okay_ to me, but the code smell is not great.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,16 @@ describe('CategoryRenderer', () => {
 
     const audits = categoryDOM.querySelectorAll('.lh-audit');
     assert.equal(audits.length, category.auditRefs.length, 'renders correct number of audits');
+
+    assert.ok(categoryDOM.querySelector('.lh-gauge__wrapper--plugin') === null);
+  });
+
+  it('plugin category has plugin badge', () => {
```
exterkamp: Testing 😍

lighthouse-core/report/html/renderer/category-renderer.js
patrickhulce: let's at least make it a util function :)

patrickhulce: also plugin category ids should have their name match the module that was passed in `--plugins` so we could check that instead?

I suppose it's a question of if we want to flag all plugins with this icon or flag all non-core groups with this icon.

Hoten: > let's at least make it a util function :)

sure

> also plugin category ids should have their name match the module that was passed in --plugins so we could check that instead?

how would I access that from this function?

Hoten:
@exterkamp made the badges more responsive

lighthouse-core/report/html/report-styles.css
```diff
@@ -143,6 +149,9 @@
     --circle-size-big: 96px;
     --circle-size: 72px;
     --header-padding: 16px 0 16px 0;
+    --plugin-badge-size: 28px;
+    --plugin-badge-size-big: 32px;
+    --plugin-icon-size: 75%;
```
paulirish: this value seems like it can be inlined in the style rule

lighthouse-core/report/html/templates.html
```diff
@@ -598,6 +598,31 @@
         animation-delay: 250ms;
     }
 
+    .lh-gauge__svg-wrapper {
+      position: relative;
+    }
+    .lh-gauge__wrapper--plugin .lh-gauge__svg-wrapper:before {
```
paulirish: add comment above

    /* The plugin badge overlay */

lighthouse-core/report/html/report-styles.css
```diff
@@ -99,19 +101,23 @@
   --color-red: #FF4E42;
   --color-white: #FFFFFF;
   --header-padding: 20px 0 20px 0;
+  --plugin-badge-bg: var(--color-white);
+  --plugin-badge-size: 36px;
```
paulirish: let's define plugin-badge-size based on circle-size via calc()

I have a plan to dynamically set circle-size to slowly shrink the badges as they scroll, so having these all using the same base numbers would help a lot.

lighthouse-core/report/html/report-styles.css
```diff
@@ -143,6 +149,9 @@
     --circle-size-big: 96px;
     --circle-size: 72px;
     --header-padding: 16px 0 16px 0;
+    --plugin-badge-size: 28px;
```
paulirish: also once plugin-badge-size is sized off circle-size we dont need this secondary definition. ;)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,16 @@ describe('CategoryRenderer', () => {
...
+    const category = sampleResults.reportCategories.find(c => c.id === 'seo');
+    // Any category not of the core
+    category.id = 'someplugin';
```
paulirish: this taints the sampleResults in all remaining tests. you should clone them with the JSON.stringify/parse dance first.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,16 @@ describe('CategoryRenderer', () => {
 
     const audits = categoryDOM.querySelectorAll('.lh-audit');
     assert.equal(audits.length, category.auditRefs.length, 'renders correct number of audits');
+
+    assert.ok(categoryDOM.querySelector('.lh-gauge__wrapper--plugin') === null);
```
paulirish: assert.equal(thing, null)

lighthouse-core/report/html/renderer/util.js
```diff
@@ -465,6 +465,13 @@ class Util {
 
     return lines.filter(line => lineNumbersToKeep.has(line.lineNumber));
   }
+
+  /**
+   * @param {string} categoryId
+   */
+  static isPluginCategory(categoryId) {
```
paulirish: adding onto [previous discussion](https://github.com/GoogleChrome/lighthouse/pull/8307#discussion_r275633709)....  yeahhhhhh this is feelin bad.  :/

my first inclination: `LH.Result.Category` should have an optional `plugin` property that uses the full plugin id.

@brendankenny we'll need your call on this.

brendankenny: I guess it depends on what we mean by a "plugin category", if we want to badge basically anything that's not under the lighthouse repo's control (basically what's in the default or full configs, I guess), or just plugin plugins.

> my first inclination: LH.Result.Category should have an optional plugin property that uses the full plugin id.

Got you covered :)

If we want it to be strictly things passed in with `--plugins` (or `plugins` in the config), then (as @patrickhulce was saying in https://github.com/GoogleChrome/lighthouse/pull/8307#discussion_r275854898), the `categoryId` from any Lighthouse plugin has the same name as the plugin itself, and plugin names [have to start with `lighthouse-plugin-`](https://github.com/GoogleChrome/lighthouse/blob/53cea6d12883799627cbdbfe7ad8b30485e8bc13/lighthouse-core/config/config.js#L173), so checking that will be sufficient.

lighthouse-core/report/html/report-styles.css
Hoten: but it varies over mobile/desktop

lighthouse-core/report/html/templates.html

lighthouse-core/report/html/report-styles.css

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: done

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: done

lighthouse-core/report/html/report-styles.css

lighthouse-core/report/html/renderer/util.js
Hoten: > and plugin names have to start with lighthouse-plugin-, so checking that will be sufficient.

done

paulirish: > and plugin names have to start with lighthouse-plugin-

we are so smrt

lighthouse-core/scripts/build-report-for-autodeployment.js
```diff
@@ -14,6 +14,13 @@ const path = require('path');
 const ReportGenerator = require('../../lighthouse-core/report/report-generator.js');
 const lhr = /** @type {LH.Result} */ (require('../../lighthouse-core/test/results/sample_v2.json'));
 
+lhr.categories.test = {
+  id: 'plugin',
```
exterkamp:
```suggestion
  id: 'lighthouse-plugin-someplugin',
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -104,6 +104,10 @@
   --env-name-min-width: 220px;
   --env-tem-padding: 10px 0px;
   --header-padding: 20px 0 20px 0;
+  --plugin-badge-bg: var(--color-white);
+  --plugin-badge-size-big: calc(var(--circle-size-big) / 2.7);
```
paulirish: looks like you can drop the -big variant now :)

lighthouse-core/report/html/report-styles.css
paulirish: ok sg.

lighthouse-core/report/html/templates.html
```diff
@@ -612,6 +612,32 @@
...
+      content: "";
+      position: absolute;
+      right: -6px;
+      bottom: -2px;
+      display: block;
+      z-index: 100;
+      box-shadow: 0 0 4px rgba(0,0,0,.2);
+      border-radius: 25%;
+    }
+    .lh-category .lh-gauge__wrapper--plugin .lh-gauge__svg-wrapper:before {
```
paulirish: as mentioned you can remove this now that we're calc() from circle-size

lighthouse-core/report/html/templates.html
```diff
@@ -664,10 +690,13 @@
 
   </style>
   <a href="#" class="lh-gauge__wrapper">
-    <svg viewBox="0 0 120 120" class="lh-gauge">
-      <circle class="lh-gauge-base" r="56" cx="60" cy="60"></circle>
-      <circle class="lh-gauge-arc" transform="rotate(-90 60 60)" r="56" cx="60" cy="60"></circle>
-    </svg>
+    <!-- Wrapper exists for the ::before plugin icon. Cannot create pseudo-elements on svgs. -->
```
paulirish: HTML comments in templates end up in the final report DOM so...

it'd be slightly better to have them on the outide of the <template> (where they are hardly any use.)  sucks as I love having these, but they're awkward additions to our shipped DOM.

paulirish: fuck it lets keep these and remove comment nodes from DOM.cloneTemplate() in a separate PR

lighthouse-core/report/html/report-styles.css
Hoten: hmm, can you take another look, b/c this seems necessary to me. the plugin badge is larger for the gauge in the section, so the `--plugin-badge-size-big` var is derived from the `--circle-size-big` var

lighthouse-core/report/html/templates.html
Hoten: fudge it #8390

lighthouse-core/scripts/build-report-for-autodeployment.js
```diff
@@ -14,6 +14,13 @@ const path = require('path');
 const ReportGenerator = require('../../lighthouse-core/report/report-generator.js');
 const lhr = /** @type {LH.Result} */ (require('../../lighthouse-core/test/results/sample_v2.json'));
 
+lhr.categories.test = {
```
brendankenny: key on `categories` should be the same as `id`. Not sure we actually depend on that anywhere in the report renderer, but just in case...

can you also add a comment (`// Add a plugin to demo plugin rendering` or whatever) and drop a note in #8159 for the feature request?

lighthouse-core/report/html/templates.html
```diff
@@ -497,6 +497,32 @@
         animation-delay: 250ms;
     }
 
+    /* The plugin badge overlay */
```
brendankenny: this comment might make more sense below this, above `.lh-gauge__wrapper--plugin`, since *all* category gauges will have an `svg-wrapper`, they just won't have the `:before` content

lighthouse-core/report/html/templates.html
```diff
@@ -549,10 +575,13 @@
 
   </style>
   <a href="#" class="lh-gauge__wrapper">
-    <svg viewBox="0 0 120 120" class="lh-gauge">
-      <circle class="lh-gauge-base" r="56" cx="60" cy="60"></circle>
-      <circle class="lh-gauge-arc" transform="rotate(-90 60 60)" r="56" cx="60" cy="60"></circle>
-    </svg>
+    <!-- Wrapper exists for the ::before plugin icon. Cannot create pseudo-elements on svgs. -->
```
brendankenny: should stick to either `:before` or `::before` in comments and in the stylesheet :)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,17 @@ describe('CategoryRenderer', () => {
 
     const audits = categoryDOM.querySelectorAll('.lh-audit');
     assert.equal(audits.length, category.auditRefs.length, 'renders correct number of audits');
+
+    assert.equal(categoryDOM.querySelector('.lh-gauge__wrapper--plugin'), null);
```
brendankenny: a string explanation like the above asserts would be nice (or a comment, but when in Rome, etc). No plugin categories in the sample results.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,17 @@ describe('CategoryRenderer', () => {
...
+  });
+
+  it('plugin category has plugin badge', () => {
+    const category = JSON.parse(
+      JSON.stringify(sampleResults.reportCategories.find(c => c.id === 'seo')));
+    // Any category not of the core
```
brendankenny: not sure what this means :)

```
Be bloody, bold, and resolute; laugh to scorn
The power of core, for none of plugin born
Shall harm the test.
```

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,17 @@ describe('CategoryRenderer', () => {
...
+    category.id = 'lighthouse-plugin-someplugin';
```
brendankenny: also needs the key and `id` to match, so maybe just append a fake category? Just need a `id`/`title`/`score`/empty `auditRefs`

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -144,6 +144,17 @@ describe('CategoryRenderer', () => {
...
+    const categoryDOM = renderer.render(category, sampleResults.categoryGroups);
+    assert.ok(categoryDOM.querySelector('.lh-gauge__wrapper--plugin'));
```
brendankenny: maybe assert slightly more...e.g. there's only one plugin wrapper in the `categoryDOM`, and `pluginWrapper.querySelector('.lh-gauge__label').textContent` is the category's `title`

lighthouse-core/report/html/report-styles.css
brendankenny: @paulirish ^

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: i must have left for lunch mid-comment :)

just gonna delete this.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: There's no key here.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: There's only one category gauge being rendered here, so there's could only be the one wrapper to assert on. I think your advice makes sense only if this test was rendering the whole report?

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: done

lighthouse-core/report/html/templates.html

lighthouse-core/report/html/templates.html

lighthouse-core/scripts/build-report-for-autodeployment.js
Hoten: done

lighthouse-core/test/report/html/renderer/category-renderer-test.js
brendankenny: > There's no key here.

er, yeah, whoops :)

lighthouse-core/report/html/report-styles.css
paulirish: ah true. 

what i was probably thinking was changing uses of circle-size-big to assign it to circle size:

```diff
     .lh-category .lh-gauge {
-      width: var(--circle-size-big);
-      height: var(--circle-size-big);
+      --circle-size: var(--circle-size-big);
     }
```

and then all your base stuff is derived off the same value. :D

lighthouse-core/test/report/html/renderer/category-renderer-test.js
brendankenny: > There's only one category gauge being rendered here, so there's could only be the one wrapper to assert on. I think your advice makes sense only if this test was rendering the whole report?

oh yeah :)

- the `.lh-gauge__label` check still makes sense, though, and that reminds me
- we should have another test similar to this but out in `report-renderer-test.js`, since report-renderer is in charge of rendering the `lh-scores-header` with all the gauges in it.

lighthouse-core/report/html/report-styles.css
Hoten: ok but i accidentally did it for #8222

Hoten: btw this specific example didnt work, i think there's a problem with the data dependency of the variables / scope.

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: done

brendankenny:
🧩 📛

PR #8328, opened 2019-04-16
new_audit: json-ld audit

mattzeunert:
**Summary**

Use the JSON-LD validator to audit JSON-LD snippets with Lighthouse.

[Sample Lighthouse report](https://invalid-json-ld-report.glitch.me/#seo)

**Related Issues/PRs**

Closes https://github.com/GoogleChrome/lighthouse/issues/4359


lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,173 @@
...
+  /** @type {LH.Audit.Details.SnippetValue['lineMessages']} */
+  const lineMessages = [];
+  /** @type {LH.Audit.Details.SnippetValue['generalMessages']} */
+  const generalMessages = [];
+  errors.forEach(({
+    message, lineNumber, validTypes, validator,
+  }) => {
+    if (validTypes && validator === 'schema-org') {
+      const typeStrings = validTypes.map(type => {
+        const typeNameMatch = type.match(/[\w]+$/);
```
patrickhulce: this feels unfortunate we have to do this regex business here

we control the library is there another way to feed this more directly, or you think it still makes sense to do this parsing?

mattzeunert: Let's do the regex business in sd-validator so that `validTypes` is an object with `uri` and `name` properties? I don't think we can avoid the regex altogether.

patrickhulce: oh yeah sg! that's what I meant, just that having to parse out with a regex *post* sd library feels rough and I'm assuming if others ever wanted to use this they could want something similar

mattzeunert:
```
 yarn build-extension && yarn build-devtools && yarn build-lr && yarn build-viewer
$ node ./build/build-extension.js
events.js:174
      throw er; // Unhandled 'error' event
      ^
Error: Cannot find module './doug-json-parse' from 'C:\projects\lighthouse\node_modules\jsonlint-mod\web'
```

@patrickhulce Do you have any thoughts on this failure? I know we used to have that problem, and I think it disappeared when we worked on the JSON-LD logic because that wasn't included in the bundle yet.

Browserify is trying to resolve this require:

```
var dougJSONParse = typeof ___dougJSONParse === 'undefined' ? require('./doug-json-parse') : ___dougJSONParse;
```

I can look into this more tomorrow, just wanted to check if you already have thoughts on what the solution is.

patrickhulce:
sorry @mattzeunert not sure off the top of my head, that file is indeed missing from that directory for me, so we might need to shim something or install another dep?

mattzeunert:
It's inlined in that file already, so there's no need to require it. Will look into shimming 👍

build/build-bundle.js
```diff
@@ -64,7 +64,14 @@ async function browserifyFile(entryPath, distPath) {
...
     .ignore('rimraf')
-    .ignore('pako/lib/zlib/inflate.js');
+    .ignore('pako/lib/zlib/inflate.js')
+    .ignore('file') // required by jsonlint-mod
+    .ignore('system'); // required by jsonlint-mod
+
+  // There is no way to add './doug-json-parse' to ignored packages via public API
+  // w/o browserify resolving the path into an absolute path
+  // @ts-ignore
+  bundle._ignore.push('./doug-json-parse');
```
patrickhulce: this feels like a recipe for struggles on next browserify update 😬 

maybe it's time to investigate moving to something better maintained/more flexible post 5.0

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,173 @@
...
+    if (jsonLDElements.length === 0) {
+      return {
+        notApplicable: true,
+        rawValue: true,
+      };
+    }
+
+    const validatedSnippets = await Promise.all(jsonLDElements.map(async (element) => {
+      // We don't want to show empty lines around the snippet
+      const content = (element.content || '').trim();
```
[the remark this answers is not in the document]
patrickhulce: bump :)

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,163 @@
...
+        content,
+        errors: await validateJsonLD(content),
+      };
+    }));
+    // Show invalid snippets at the top
+    validatedSnippets.sort((a, b) => {
+      return b.errors.length - a.errors.length;
+    });
+
+    const renderedSnippets = await Promise.all(
```
patrickhulce: this doesn't seem to be async?

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,163 @@
...
+ * @param {{content: string, devtoolsNodePath: string, errors: LH.StructuredData.ValidationError[]}} validatedSnippet
+ */
+function renderValidatedSnippet(validatedSnippet) {
+  const {content, devtoolsNodePath, errors} = validatedSnippet;
+
+  let parsedContent;
+  let topLevelType;
+  let topLevelName;
+  try {
+    parsedContent = JSON.parse(content);
```
patrickhulce: can we reuse a type for this somewhere from sd-validation types?

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,163 @@
...
+
+  let title = '';
+  if (topLevelName && topLevelType) {
+    title = `${topLevelType}: ${topLevelName}`;
+  } else if (topLevelType) {
+    title = `@type ${topLevelType}`;
+  } else {
+    title = 'Invalid JSON-LD element';
+  }
+  title += ` (${errors.length} Error${errors.length !== 1 ? 's' : ''})`;
```
patrickhulce: can we i18n this sucker?

lighthouse-core/lib/sd-validation/schema-validator.js
```diff
@@ -96,11 +96,18 @@ function validateObjectKeys(typeOrTypes, keys) {
...
-      message: `Unexpected property "${key}"`,
-      key,
-      validTypes: types,
-    }));
+    .map(key => {
+      return ({
+        message: `Unexpected property "${key}"`,
+        key,
+        validTypes: types.map(typeUri => {
+          const typeNameMatch = typeUri.match(/[\w]+$/);
```
patrickhulce: 👍

lighthouse-core/lib/sd-validation/schema-validator.js
```diff
@@ -96,11 +96,18 @@ function validateObjectKeys(typeOrTypes, keys) {
...
+          const name = typeNameMatch ? typeNameMatch[0] : typeUri;
+          const uri = typeUri.replace('http://', 'https://');
```
patrickhulce: what effect does this end up having, just curious :)

mattzeunert: We don't end up linking to non-https pages from the report.

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: Yeah maybe. I'm not quite sure how to do it, is there some way to put the `snippetName ? snippetName: 'Invalid JSON-LD element'` in the i18n string? It seems like we can't concatenate individual i18n strings because the final conversion is done on the full strings.

I wonder if we even need to, given that the actual error messages are in English and if the snippet can be parsed it'll use the English schema.org name.

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: It's kinda like `ExpandedSchemaRepresentationItem` – is there actually anything special about that type being `Expanded`?

lighthouse-core/audits/seo/structured-data-automatic.js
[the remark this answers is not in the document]
mattzeunert: It's `notApplicable` now because we do this:

``` javascript
const jsonLDElements = artifacts.ScriptElements.filter(
      script => script.type === 'application/ld+json' && !!script.content);
```

Have changed the `|| ''` to a `/** @type ` cast.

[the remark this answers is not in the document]
patrickhulce: we don't seem to be trimming that filter about though was what I was trying to say

lighthouse-core/audits/seo/structured-data-automatic.js
patrickhulce: ah yeah the mingling with lib error messages messes with everything huh, nevermind

the manual pluralization might stick out to someone in the future, mind throwing a quick comment? :)

lighthouse-core/lib/sd-validation/schema-validator.js
patrickhulce: oh, no impact at all from the structured data side of things the https/http are always interchangable?

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: Added one 🙂

lighthouse-core/lib/sd-validation/schema-validator.js
mattzeunert: [I think so](https://schema.org/docs/faq.html#19).
> Q: Should we write 'https://schema.org' or 'http://schema.org' in our markup?
> 
> There is a general trend towards using 'https' more widely, and you can already write 'https://schema.org' in your structured data. Over time we will migrate the schema.org site itself towards using https: as the default version of the site and our preferred form in examples. However 'http://schema.org' -based URLs in structured data markup will remain widely understood for the forseeable future and there should be no urgency about migrating existing data. This is a lengthy way of saying that both 'https://schema.org' and 'http://schema.org' are fine.
> 
>

mattzeunert:
@patrickhulce Any more feedback on this?

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,165 @@
...
+const i18n = require('../../lib/i18n/i18n.js');
+
+const UIStrings = {
+  /** Title of a Lighthouse audit that provides detail on whether JSON-LD structured data snippets are valid. This descriptive title is shown when no invalid JSON-LD snippets were found. */
+  title: 'JSON-LD structured data syntax is valid',
+  /** Title of a Lighthouse audit that provides detail on whether JSON-LD structured data snippets are valid. This descriptive title is shown when JSON-LD snippets with invalid content were found. */
+  failureTitle: 'JSON-LD structured data syntax is invalid',
+  /** Description of a Lighthouse audit that tells the user whether JSON-LD snippets on the page are invalid. This is displayed after a user expands the section to see more. No character length limits. */
+  /* eslint-disable-next-line max-len */
+  description: 'Structured data contains rich metadata about a web page. The data is used in search results and social sharing. Invalid metadata will affect how the page appears in these contexts. This audit currently validates a subset of JSON-LD rules. See also the manual audit below to learn how to validate other types of structured data.',
```
patrickhulce: I wonder if we should just include the link to the tool here rather than force them to go to the other audit?

lighthouse-core/audits/seo/structured-data-automatic.js
patrickhulce: I don't totally remember :) buuuuuuut I think it's just that there are more compact forms/shortcuts/references that get resolved to their actual values?

https://json-ld.org/spec/latest/json-ld-api/#expansion

You've been much deeper in this more recently than me though, so I'll leave it to your judgement. We are trying to access the `@type` and `name` properties though so have *some* type would be nice

lighthouse-core/test/results/sample_v2.json
```diff
@@ -2934,6 +2934,13 @@
       "score": null,
       "scoreDisplayMode": "notApplicable"
     },
+    "structured-data-automatic": {
+      "id": "structured-data-automatic",
+      "title": "JSON-LD structured data syntax is valid",
+      "description": "Structured data contains rich metadata about a web page. The data is used in search results and social sharing. Invalid metadata will affect how the page appears in these contexts. This audit currently validates a subset of JSON-LD rules. See also the manual audit below to learn how to validate other types of structured data.",
+      "score": null,
+      "scoreDisplayMode": "notApplicable"
```
patrickhulce: would be nice to update our artifacts sometime with all the new goodies you've worked on, then we'd get the report previews for free :) 

seems fine for a separate PR though 👍

lighthouse-core/audits/seo/structured-data-automatic.js
```diff
@@ -0,0 +1,165 @@
...
+  /** Explanatory message stating what percentage of JSON-LD structured data snippets are invalid */
+  displayValue: '{validSnippetProportion, number, percent} valid snippets',
```
patrickhulce: I'm not positive that percentage is what we care most about here, maybe we just report # invalid?

build/build-bundle.js
patrickhulce: still a bit concerned about this but I think whatever we end up doing will be hacky in some way 🤷‍♂️

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: This is the message for the manual structured data audit:

> Run the [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/) and the [Structured Data Linter](http://linter.structured-data.org/) to validate structured data. [Learn more](https://developers.google.com/search/docs/guides/mark-up-content?utm_source=lighthouse&utm_medium=cli).

The Structured Data Linter doesn't look very welcoming, so maybe we can remove it and just link the SDTT from both audits.

Do we even need the manual audit? If we removed it we'd need to change the title of the automatic audit to not only mention JSON-LD. But maybe that's misleading, if we say your structured data is fine but we only checked JSON-LD?

@avgp any thoughts on this?

(Related: should we call the audit `structured-data-automatic` internally or just `json-ld`?)

patrickhulce: I'll defer to @AVGP for removal of the manual one. `json-ld` seems like a reasonable audit id to me. Do we foresee this growing in scope to include other types of data?

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: I've added a type for it now. Turns out we weren't actually returning a valid `ExpandedSchemaRepresentation` from `expand`, so fixed that type too.

> I don't totally remember :) buuuuuuut I think it's just that there are more compact forms/shortcuts/references that get resolved to their actual values?

That's a good point, ideally we should use the expanded representation and look for "http://schema.org/name" on there. What do you think of having the the validator return an object with `errors` and `expandedSchema` properties, instead of just the errors array?

Fwiw, I've never actually seen anyone do anything fancy with JSON-LD in the wild.

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: Changed to "N invalid snippets"

lighthouse-core/test/results/sample_v2.json
mattzeunert: Why is updating the artifacts such a pain? Couldn't we just have script that reverts everything that's expected to vary between runs?

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: Renamed the audit to `json-ld`.

Will bring up description/manual audit with @avgp again in a week or two, but shouldn't be merge blocking.

lighthouse-core/test/results/sample_v2.json
patrickhulce: > everything that's expected to vary between runs

The problem is that list is enormous and not always separable from what you wanted to change :) 

I think to your point though, there's room for a more targeted script that says "only accept these particular artifact changes" and throw away everything else. 👍

lighthouse-core/audits/seo/structured-data-automatic.js
patrickhulce: Ah nice catch!! Yeah that sounds reasonable to me, though maybe if it's not particularly common we punt for now. Just file an issue?

mattzeunert: Filed: https://github.com/GoogleChrome/lighthouse/issues/8668

mattzeunert:
@brendankenny Unstuffed the merge now, should be good for review.

lighthouse-core/audits/seo/structured-data-automatic.js
mattzeunert: [Updates](https://github.com/GoogleChrome/lighthouse/pull/8328/commits/991bf3c0e30e201224d3956fba4e65a1cb0b159a) after talking to @avgp:

- removed the link to the Structured Data Linter
- link to SDTT from json-ld audit description

AVGP: Yep, I think there's value in having the manual audit as people get confused with "valid JSON-LD" vs. "Valid SD for rich results" and such, plus the SDTT catches most of that, so linking directly to it is a good idea.

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -108,6 +108,15 @@
 <!-- Force FCP to be ~5s out, necessary to make sure our render-blocking audits still work -->
 <!-- even when other forces were responsible for delaying the render. -->
 <script src="./fcp-delayer.js?delay=5000"></script>
+
+<!-- FAIL(json-ld): buttons too close together -->
```
brendankenny: update failure reason :)

build/build-bundle.js
```diff
@@ -64,7 +64,14 @@ async function browserifyFile(entryPath, distPath) {
     .ignore('raven')
     .ignore('mkdirp')
     .ignore('rimraf')
-    .ignore('pako/lib/zlib/inflate.js');
+    .ignore('pako/lib/zlib/inflate.js')
+    .ignore('file') // required by jsonlint-mod
+    .ignore('system'); // required by jsonlint-mod
+
+  // There is no way to add './doug-json-parse' to ignored packages via public API
+  // w/o browserify resolving the path into an absolute path
```
brendankenny: thanks for documenting these. They are pretty annoying to have to manually prune :(

lighthouse-core/audits/seo/json-ld.js
```diff
@@ -0,0 +1,172 @@
...
+
+const UIStrings = {
+  /** Title of a Lighthouse audit that provides detail on whether JSON-LD structured data snippets are valid. This descriptive title is shown when no invalid JSON-LD snippets were found. */
+  title: 'JSON-LD structured data syntax is valid',
+  /** Title of a Lighthouse audit that provides detail on whether JSON-LD structured data snippets are valid. This descriptive title is shown when JSON-LD snippets with invalid content were found. */
+  failureTitle: 'JSON-LD structured data syntax is invalid',
+  /** Description of a Lighthouse audit that tells the user whether JSON-LD snippets on the page are invalid. This is displayed after a user expands the section to see more. No character length limits. */
+  /* eslint-disable-next-line max-len */
+  description: 'Structured data contains rich metadata about a web page. The data is used in search results and social sharing. Invalid metadata will affect how the page appears in these contexts. This audit currently validates a subset of JSON-LD rules. See also the [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool/) to validate other types of structured data.',
```
brendankenny: could this be trimmed down a little. Really we just want enough information for why this is important and have elaboration in the linked content (though the subset disclaimer seems important)

lighthouse-core/audits/seo/json-ld.js
```diff
@@ -0,0 +1,172 @@
...
+    if (jsonLDElements.length === 0) {
+      return {
+        notApplicable: true,
+        score: 1,
+      };
+    }
+
+    const validatedSnippets = await Promise.all(jsonLDElements.map(async (element) => {
+      // We don't want to show empty lines around the snippet
+      const content = /** @type string */ (element.content).trim();
```
brendankenny: can you add a note that `element.content` is known to be not null due to the `filter()` above, so the cast is ok

lighthouse-core/audits/seo/json-ld.js
```diff
@@ -0,0 +1,172 @@
...
+      script => script.type === 'application/ld+json' && !!script.content);
+
+    if (jsonLDElements.length === 0) {
+      return {
+        notApplicable: true,
+        score: 1,
+      };
+    }
+
+    const validatedSnippets = await Promise.all(jsonLDElements.map(async (element) => {
```
brendankenny: `validatedSnippets` is kind of confusing here as a variable name (e.g. invalid validatedSnippets breaks my brain a little)

Just `snippets`, maybe? Doesn't seem to lose out on context below (a snippet in `snippets` can either be valid or invalid)

lighthouse-core/audits/seo/json-ld.js
```diff
@@ -0,0 +1,172 @@
...
+        errors: await validateJsonLD(content),
+      };
+    }));
+    // Show invalid snippets at the top
+    validatedSnippets.sort((a, b) => {
+      return b.errors.length - a.errors.length;
+    });
+
+    const renderedSnippets = validatedSnippets.map(
+      snippetWithErrors => renderValidatedSnippet(snippetWithErrors)
```
brendankenny: does just `const renderedSnippets = validatedSnippets.map(renderValidatedSnippet);` work?

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -108,6 +108,15 @@
...
+<script type="application/ld+json">
+{
+  "@context": "http://schema.org",
+  "@type": "Event",
+  "unknownProp": "Oh no"
+}
```
brendankenny: is there an expectations change for this?

build/build-bundle.js
Hoten: should we add a test that the bundle output does not contain some code in `douge-json-parse`? so we remember to fix this when it break.

brendankenny: > should we add a test that the bundle output does not contain some code in douge-json-parse? so we remember to fix this when it break.

the build throws if we forget this, so it will definitely fail the tests :)

The issue is that `jsonlint-mod`'s bundled js has a reference to this file, which isn't included in the module. Browserify throws when it tries to resolve it.

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
AVGP: I'm not sure I understand the question - could you elaborate a bit?

AVGP:
@paulirish Is there something I gotta do about the CLA or the bundlesize?

When I run `yarn bundlesize` locally, I get different results :/

AVGP:
@paulirish I'm not sure what's up with the bundlesize but I think this PR is ready to move forward?

connorjclark:
(2410564 - 1544755) / 1000 = +865KB to devtools bundles. Can expect slightly less if PR is updated to include recent build improvements.

connorjclark:
this was rather recently superseded by another effort to bring structured data audit to LH: #12324 , so closing for that.

PR #8329, opened 2019-04-16
report(redesign): remove icons from non-pwa group headings

Hoten:
#8185


lighthouse-core/report/html/report-styles.css
```diff
@@ -659,36 +648,6 @@
   vertical-align: middle;
```
paulirish: I think this rule will also be removed as well.

paulirish: oh it's kept for the PWA ones.  ;)

PR #8289, opened 2019-04-15
core: remove disableDeviceEmulation flag

wardpeet:
**Summary**
Remove disableDeviceEmulation flag and fore people on emulatedFormFactor

**Related Issues/PRs**
closes https://github.com/GoogleChrome/lighthouse/issues/7044

PR #8334, opened 2019-04-16
core: add deprecation message for disable-device-emulation flag

wardpeet:
**Summary**
Adding small deprecation message for disableDeviceEmulation flag.

For more info chat with @patrickhulce 😄 

**Related Issues/PRs**
https://github.com/GoogleChrome/lighthouse/pull/8289#issuecomment-483693944

### DIFF
--- a/lighthouse-cli/bin.js
+++ b/lighthouse-cli/bin.js
@@ -94,7 +94,7 @@ async function begin() {
     cliFlags.outputPath = 'stdout';
   }
 
-  // show deprecation message for disableDeviceEmulation
+  // @ts-ignore - deprecation message for removed disableDeviceEmulation; can remove warning in v6.
   if (cliFlags.disableDeviceEmulation) {
     log.warn('config', 'The "--disable-device-emulation" has been removed in v5.' +
         ' Please use "--emulated-form-factor=none" instead.');
--- a/lighthouse-cli/bin.js
+++ b/lighthouse-cli/bin.js
@@ -96,8 +96,8 @@ async function begin() {
 
   // show deprecation message for disableDeviceEmulation
   if (cliFlags.disableDeviceEmulation) {
-    log.warn('config', 'The "--disable-device-emulation" option is deprecated.' +
-      ' Please use "--emulated-form-factor=none" instead.');
+    log.warn('config', 'The "--disable-device-emulation" has been removed in v5.' +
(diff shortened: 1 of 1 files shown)


patrickhulce:
It just occurred to me that we can do this after it's been removed too :)

I propose tweaking the text slightly and cherry-picking this back to v4.3.1 if we ever publish a patch release

lighthouse-cli/bin.js
```diff
@@ -94,6 +94,12 @@ async function begin() {
     cliFlags.outputPath = 'stdout';
   }
 
+  // show deprecation message for disableDeviceEmulation
+  if (cliFlags.disableDeviceEmulation) {
+    log.warn('config', 'The "--disable-device-emulation" option is deprecated.' +
```
patrickhulce: we can make this pull double duty if we say `"--disable-device-emulation" has been removed in v5.`

wardpeet:
```js
log.warn('config', 'The "--disable-device-emulation" option is deprecated.' +
    ' Please use "--emulated-form-factor=none" instead.');
```

wardpeet:
Feel free to keep this PR around until a patch is going to be made or just close if seems unecessary.

lighthouse-cli/bin.js
```diff
@@ -94,6 +94,12 @@ async function begin() {
     cliFlags.outputPath = 'stdout';
   }
 
+  // show deprecation message for disableDeviceEmulation
```
brendankenny: maybe add a note we should remove for v6? (we don't have to remove it then, but then we'll know it's alright to at any time after that :)

lighthouse-cli/bin.js
```diff
@@ -94,6 +94,12 @@ async function begin() {
...
+  if (cliFlags.disableDeviceEmulation) {
```
brendankenny: I think you'll have to `// @ts-ignore disableDeviceEmulation has been removed from cliFlags` this line now that #8289 has removed it

wardpeet:
@brendankenny we could go for something more in your face https://www.npmjs.com/package/boxen

brendankenny:
> @brendankenny we could go for something more in your face https://www.npmjs.com/package/boxen. we could also pause and put some new lines below

sounds terrible, let's not do that :D

brendankenny:
@patrickhulce does this work for you?

PR #8338, opened 2019-04-16
core(driver): only fail security state if scheme is not cryptographic

patrickhulce:
**Summary**
Currently no `http` site seems to be auditable by the extension in Canary. AFAICT, Chrome has started to fire security state `insecure` events, but not over the websocket protocol for some reason, just chrome.debugger... 😕 

It *seems* like we should be able to just do the security check if the scheme is cryptographic and ignore it if it wasn't cryptographic? There are some other properties in the object that look promising (`displayedInsecureContentStyle === 'insecure'` maybe?), but want @Hoten 's take since he's our resident security state expert :)

**Event we get for bad ssl site**
image

**Event we get for http site**
image

### DIFF
--- a/lighthouse-core/gather/driver.js
+++ b/lighthouse-core/gather/driver.js
@@ -912,8 +912,12 @@ class Driver {
       /**
        * @param {LH.Crdp.Security.SecurityStateChangedEvent} event
        */
-      const securityStateChangedListener = ({securityState, explanations}) => {
-        if (securityState === 'insecure') {
+      const securityStateChangedListener = ({
+        securityState,
+        explanations,
+        schemeIsCryptographic,
+      }) => {
+        if (securityState === 'insecure' && schemeIsCryptographic) {
           cancel();
           const insecureDescriptions = explanations
             .filter(exp => exp.securityState === 'insecure')
--- a/lighthouse-core/test/gather/driver-test.js
+++ b/lighthouse-core/test/gather/driver-test.js
@@ -700,6 +700,31 @@ describe('.gotoURL', () => {
       await loadPromise;
     });
 
+    it('does not reject when page is insecure but http', async () => {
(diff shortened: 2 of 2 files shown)


Hoten:
Would we prefer to be more explicit and just do the check if `https`?

https://cs.chromium.org/chromium/src/url/gurl.cc?type=cs&sq=package:chromium&g=0&l=385

that's all Chrome is doing.

patrickhulce:
> Would we prefer to be more explicit and just do the check if https?

I could be on board for that! Just means we have to thread the URL down to here which is kinda pain

Hoten:
> I could be on board for that!

cool

> Just means we have to thread the URL down to here which is kinda pain

ok you've convinced me

lighthouse-core/gather/driver.js
```diff
@@ -912,8 +912,12 @@ class Driver {
...
        * @param {LH.Crdp.Security.SecurityStateChangedEvent} event
        */
-      const securityStateChangedListener = ({securityState, explanations}) => {
-        if (securityState === 'insecure') {
+      const securityStateChangedListener = ({
+        securityState,
+        explanations,
+        schemeIsCryptographic,
+      }) => {
+        if (securityState === 'insecure' && schemeIsCryptographic) {
```
brendankenny: yeah, the nice thing about `schemeIsCryptographic` is that it's always testing the scheme of the URL on which this event was fired†. It would be difficult to get the current URL in here *and* (in the case of redirects) to make sure the URL we're checking is the one associated with the event we're checking.

I guess the question is are there cases where `schemeIsCryptographic` is false but we should still be checking for the `securityState`, otherwise we're testing a Chrome-generated error page?

†uh, at least I assume it is :)

paulirish:
wfm

Issue #6199, opened 2018-10-09
Remove rawValue from Audit.Result


exterkamp:
#### What is the current behavior?
`rawValue` is still in each Audit in the LHR, but the `Metrics` audit contains all `rawValue` information and so `rawValue` is redundant.

paulirish:
The big change here is asking people to pull metric values from `audits.metrics.details` instead of `audits[metricId].rawValue`

I'm cool with that, though.

benschwarz:
I might've misunderstood this issue a little bit, but on a cursory look at the lighthouse JSON, `dom-size`, `time-to-first-byte` and `bootup-time` are not contained within the metrics audit.

paulirish:
* dom-size: yes we dont have a great answer for this. we changed it for formatting recently and now we broke the usecase of grabbing these values
* time-to-first-byte: hmmmm yes we dont expose a value here that'd be useful. `.details.overallSavingsMs` is something else.
* bootup-time: same # available via `.details.summary.wastedMs`.

benschwarz:
I guess for `dom-size` we can add a `.details.summary.value` so that it's similar to `bootup-time`.

exterkamp:
Started to work on this with [branch](https://github.com/GoogleChrome/lighthouse/tree/remove-rawValue).

Currently I have removed rawValue from `auditResult` objects so that the value is never surfaced to the LHR.  But it is still used internally in the `auditProduct` objects.

brendankenny:
> But it is still used internally in the auditProduct objects.

Since (I believe) this isn't blocking the proto work, I think we should do this right and eliminate the `rawValue` from the product as well.

This might be one of those cases we all need to get in a single (virtual) room and make sure every useful `rawValue` has a reasonable `details` that handles it

exterkamp:
@brendankenny :+1: I agree.  The current branch _works_ but obviously isn't clean and will rawValue will definitely rot forever.  I will continue to make the changes to `auditResult`.

Hoten:
Since we delete `rawValue` for LR, that presents problems with smoke tests that assert on it. But I'll rectify that by swapping over to assertions on `score`.

Are we leaning towards deleting `rawValue` entirely?

Hoten:
via LR, the `robots-txt` audit doesn't provide a sane way to check for success (since `rawValue` is dropped, and we provide no score value). Just another example of how our scoring story could be improved.

```
// OK
{
[...]

PR #8343, opened 2019-04-16
core: require score in audit product; make rawValue numeric only

brendankenny:
The majority of the change in https://github.com/GoogleChrome/lighthouse/issues/6199#issuecomment-477720022.

- makes `score` a required part of `LH.Audit.Product`, either a number or null (in case of ERROR, NOT_APPLICABLE, or INFORMATIVE)
- make all boolean uses of `rawValue` into a `score` of 0 or 1, then drop those `rawValue`s

Mostly string replace, except in `audit.js` and `audit-test.js`

after this I think we want to (at least) rename `rawValue` to finish out #6199

### DIFF
--- a/lighthouse-core/test/audits/audit-test.js
+++ b/lighthouse-core/test/audits/audit-test.js
@@ -20,6 +20,30 @@ class B extends Audit {
   static audit() {}
 }
 
+class PassOrFailAudit extends Audit {
+  static get meta() {
+    return {
+      id: 'pass-or-fail',
+      title: 'Passing',
+      failureTitle: 'Failing',
+      description: 'A pass or fail audit',
+      requiredArtifacts: [],
+    };
+  }
+}
+
+class NumericAudit extends Audit {
+  static get meta() {
+    return {
+      id: 'numeric-time',
+      title: 'Numbersssss',
+      description: '01000000001011011111100001010100',
+      requiredArtifacts: [],
+      scoreDisplayMode: Audit.SCORING_MODES.NUMERIC,
+    };
+  }
+}
+
 describe('Audit', () => {
   it('throws if an audit does not override the meta', () => {
     assert.throws(_ => A.meta);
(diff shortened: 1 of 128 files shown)


patrickhulce:
> Mostly string replace, except in audit.js and audit-test.js

Only done a quick scan, but it's fair to say this is the only meat right?

brendankenny:
> Only done a quick scan, but it's fair to say those two files are the only meat right?

yeah, `tap-targets-test.js` needed some rounding due to the scoring used in that audit and there are the limited type/comment changes in `audit.d.ts`, but the rest are all trivial replacements. Even `sample_v2.json` only had `"rawValue": (true|false)` lines removed, and thanks to @Hoten's work in #7780, smoke expectations didn't change at all (all remaining `rawValue`s in those are numeric)

lighthouse-core/test/runner-test.js
```diff
@@ -389,7 +390,7 @@ describe('Runner', () => {
     return Runner.run({}, {config}).then(results => {
       const audits = results.lhr.audits;
       assert.equal(audits['critical-request-chains'].displayValue, '5 chains found');
-      assert.equal(audits['critical-request-chains'].rawValue, false);
+      assert.equal(audits['critical-request-chains'].details.longestChain.transferSize, 2468);
```
brendankenny: oh, I guess not quite string replace in this file because it made no sense to assert the score on these. They're all `null` since the audit is `INFORMATIVE`

lighthouse-core/test/audits/audit-test.js
```diff
@@ -20,6 +20,30 @@ class B extends Audit {
...
+    };
+  }
+}
+
+class NumericAudit extends Audit {
+  static get meta() {
+    return {
+      id: 'numeric-time',
+      title: 'Numbersssss',
+      description: '01000000001011011111100001010100',
```
patrickhulce: `@-�T`?

c'mon gotta have some next level messaging in here :)

lighthouse-core/test/audits/audit-test.js
```diff
@@ -37,64 +61,94 @@ describe('Audit', () => {
...
+      const auditResult = Audit.generateAuditResult(PassOrFailAudit, {score: 0});
+      assert.strictEqual(auditResult.score, 0);
       assert.equal(auditResult.title, 'Failing');
     });
+
+    it('chooses the title if audit is not scored due to scoreDisplayMode', () => {
+      const auditResult = Audit.generateAuditResult(PassOrFailAudit,
+          {score: 0, errorMessage: 'what errors lurk'});
+      assert.strictEqual(auditResult.score, null);
+      assert.equal(auditResult.title, 'Passing');
```
patrickhulce: we use the passing title for errors? funny the things you learn in testing :)

lighthouse-core/test/audits/audit-test.js
Hoten: `01101101011001010110111101110111`

brendankenny: not everything's a character :P

`new Float32Array(new Uint32Array([0b01000000001011011111100001010100]).buffer)[0]`

lighthouse-core/test/audits/audit-test.js
brendankenny: > we use the passing title for errors? funny the things you learn in testing :)

yeah, I was really surprised when I made `score === null` explicitly use the `failureTitle` and a whole bunch of things changed in `sample_v2.json` :)

I'm not entirely convinced it was on purpose, but it makes sense for `NOT_APPLICABLE` and `INFORMATIVE`, and it kind of makes sense for `ERROR`, since we don't know the page was doing the failure thing.

lighthouse-core/test/audits/audit-test.js
patrickhulce: touche :)

PR #8370, opened 2019-04-17
deps: axe-core@3.2.2

paulirish:
has two perf boosts: https://github.com/dequelabs/axe-core/releases/tag/v3.2.0

3.2.0 introduced rules [`aria-hidden-focus`](https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus) and [`form-field-multiple-labels`](https://dequeuniversity.com/rules/axe/3.2/form-field-multiple-labels) which are tagged wcag2a (which means we'll run them now).

The hidden-focus rule scares me a bit, implementation-wise, so I'm hesitant to include that in our results. The multiple-labels rule looks straightfoward.

@robdodson @marcysutton do yall have feelings on this?

fixes #7496
fixes #7355

Issue #7127, opened 2019-01-31
evaluate new axe audits for inclusion


brendankenny:
When updating axe in #7020, I noticed there were some axe audits that are running but don't have audits to show them in the report. We should look at them and see if we should be showing them in the report or if we should disable them to save gathering time. We should also see if there are any other tests we should add from the more comprehensive axe of today.

**The current list of running tests with no LH audits:**

1. `aria-dpub-role-fallback`: Ensures unsupported DPUB roles are only used on elements with implicit fallback roles.
2. `aria-hidden-body`: Ensures aria-hidden='true' is not present on the document body.
3. `duplicate-id-active`: Ensures every id attribute value of active elements is unique.
4. `duplicate-id-aria`: Ensures every id attribute value used in ARIA and in labels is unique.
5. `html-xml-lang-mismatch`: Ensure that HTML elements with both valid lang and xml:lang attributes agree on the base language of the page.
1. [`aria-hidden-focus`](https://dequeuniversity.com/rules/axe/3.2/aria-hidden-focus): ensures aria-hidden elements do not contain focusable elements.
1. [`form-field-multiple-labels`](https://dequeuniversity.com/rules/axe/3.2/form-field-multiple-labels): Ensures form field does not have multiple label elements.
1. [aria-input-field-name](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/aria-input-field-name.json) Ensures every ARIA input field has an accessible name
1. [aria-toggle-field-name](https://github.com/dequelabs/axe-core/blob/develop/lib/rules/aria-toggle-field-name.json) Ensures every ARIA toggle field has an accessible name

On the other side of things, we don't appear to have any audits that don't have results somewhere in the axe results, so that's good :)

----

Updated April 17: merged in shane's descriptions. Paul added 6,7 from #8370. Patrick added 8, 9 from #9343

exterkamp:
A little more color to these since I was kinda lost:

1. `aria-dpub-role-fallback`: Ensures unsupported DPUB roles are only used on elements with implicit fallback roles.
2. `aria-hidden-body`: Ensures aria-hidden='true' is not present on the document body.
3. `duplicate-id-active`: Ensures every id attribute value of active elements is unique.
4. `duplicate-id-aria`: Ensures every id attribute value used in ARIA and in labels is unique.
5. `html-xml-lang-mismatch`: Ensure that HTML elements with both valid lang and xml:lang attributes agree on the base language of the page.

[...]

PR #8373, opened 2019-04-17
core(a11y): disable unused axe rules

paulirish:
Investigated in #7127

And proposed here: https://github.com/GoogleChrome/lighthouse/pull/8370#issuecomment-484186501

We don't show these rules to our LH users, but they run. Their runtime cost is probably short, but we can keep them disabled for now, until it's clear that we want them. (#7127 tracks that evaluation for inclusion)


paulirish:
> does it fail with unrecognized rule or something? :)

it does. which is nice. 

added those two as disabled in #8370

patrickhulce:
> it does. which is nice.

oh sweet! sorry I missed that update :)

PR #8379, opened 2019-04-17
tests: add smoketest to capture microtask regression

patrickhulce:
**Summary**
There was a recent Chrome change where we lost all CPU time that was spent in microtasks. It's being fixed (https://bugs.chromium.org/p/chromium/issues/detail?id=953914), but we should have a test that would've caught this earlier.


brendankenny:
needs lint fixes

lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
```diff
@@ -31,9 +31,9 @@ module.exports = [
       requestedUrl: 'http://localhost:10200/tricky-main-thread.html?setTimeout',
       finalUrl: 'http://localhost:10200/tricky-main-thread.html?setTimeout',
       audits: {
-        interactive: {
+        'interactive': {
```
Hoten: hmmmmmmmmmmmmm do we have to? :neckbeard:

lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
```diff
@@ -26,4 +26,50 @@ module.exports = [
...
+      audits: {
+        'interactive': {
+          // Make sure all of the CPU time is reflected in the perf metrics as well.
+          score: '<.2',
+        },
+        'bootup-time': {
+          details: {
+            items: {
+              0: {
+              // TODO: requires async stacks, https://github.com/GoogleChrome/lighthouse/pull/5504
```
brendankenny: seems like we don't need the TODO anymore? (`// requires async stacks` might be worth keeping, though)

patrickhulce: oh yeah, I'll update this, but turns out async stacks are just one required part for this particular bug 😞 

I'll file a new one to track progress.

patrickhulce:
> what do you want to do here? Is the smoke test failure expected or is something else still in progress?

The smoketest is accurately capturing the failure mode this test would have alerted us to, so that's good news.

The bad news is Travis is running with `Google Chrome 73.0.3683.103` and shouldn't have the commit we bisected to that caused it. Further investigation reveals that it's apparently been broken before. I don't have the exact commit range, but it looks like the timeline was somewhere along the lines of the following


July 2018 and earlier (example: 68.0.3410.0)
Microtask work nested inside toplevel event

Fall 2018 - February 2019 (examples: 70.0.3538.0, 71.0.3551.0, 73.0.3683.103)
Microtask work no longer nested inside toplevel event (I suspect https://bugs.chromium.org/p/chromium/issues/detail?id=871204 but haven't done an actual bisect)

? - March 19 2019 (example: 74.0.3729.0)
Microtask work nested inside toplevel event

March 20 2019 - April 18 2019 (example: 75.0.3766.0)
Microtask work no longer nested inside toplevel event

April 18 2019 - Present (example: 76.0.3773.0)
Microtask work nested inside toplevel event


Good news is that because it was broken and fixed within 75 before branch I think we just need to wait for m74 until this can be merged (next Chrome stable).

patrickhulce:
> Travis is testing m74 and all seems good now? Good to merge?

Yeppers!

Hoten:
Could we define the assertions for `interactive` relative to `numericValue` instead of `score`? This smoke test is failing in LR:

image


A score of 0.94 vs the expected <0.2 doesn't tell me as much as a numericValue of 3250 vs the expected value of Y would be. I know I can calculate Y after determining how the score is calculated, but ew maths :)

patrickhulce:
yeah sure @Hoten , also chatting with paul the version of m75 that's in LR was indeed a broken (just a few hours before the fix IIRC) so it is catching good things

PR #8374, opened 2019-04-17
core(config): add dedicated ID to audit/gatherer defns

patrickhulce:
**Summary**
So blessed artifacts turned out to be quite the rabbit hole. This is part 1 of 3 stages.

1. (This PR) Split apart some config logic into a helper file and add a dedicated ID to audit and gatherer definitions so that they are more easily referenceable. This also has the happy side effect of enabling the use of options in plugin audits in step 2 :D
2. Enforce that every plugin audit id is prefixed with `<plugin-id>`. It will still be able to reference our existing core audit implementations. 
3. Load each audit when loading a plugin to check its requiredArtifacts against a set of blessed artifacts

**Related Issues/PRs**
#6747

PR #8381, opened 2019-04-17
core(plugins): load plugins to validate artifacts

patrickhulce:
**Summary**
Lighter version of #8374 that's all in one-go. The other breaking change we might want to consider before v5 related to plugins is that audit paths are currently relative to the master config instead of relative to the plugin's config, which might make more sense. i.e. in the plugin config you want to say 'use the audit right next to my config file'

Nearly all of these changes would be necessary to make that happen so I think it's pretty worthwhile :)

**Related Issues/PRs**
#8374


patrickhulce:
Would like a decision here on what we consider breaking for v5. If the answer is "this is all puntable", that's fine :)

Two breaking changes discussed/blocked on this PR:

- Throwing on the use of non-blessed artifacts
- Changing what file audit paths in plugins are relative to

*see*
> audit paths are currently relative to the master config instead of relative to the plugin's config, which might make more sense. i.e. in the plugin config you want to say 'use the audit right next to my config file'

patrickhulce:
Friendly ping here. 

By the lack of excitement I'm guessing we don't really care about the audit file path resolution behavior for v5, but explicit confirmation of this fact would be nice :)

patrickhulce:
Closing in favor of #8374, this PR was only relevant as an easier transition path that could land before 5.0.0.

PR #8382, opened 2019-04-17
core(artifacts): create a PublicGathererArtifacts type

patrickhulce:
**Summary**
In conjuction with #8381, this moves us to blessed artifacts. The only non-typing change is to `ChromeConsoleMessages` which just removes the `Chrome` prefix. IMO, that feels more like the public API we might want to expose, but it's a trivial change if we want to revert.

Our friends at adspeed only use `ViewportDimensions`, `IFrameElements`, `devtoolsLogs`, and `traces`. We haven't created `IFrameElements` yet but we can always increase the scope of our blessed artifacts later.

After discussion with @brendankenny and @paulirish, we'll probably rely on documentation to prevent folks from venturing too far on their own with computed artifacts.

**Related Issues/PRs**
#8381


lighthouse-core/gather/gatherers/chrome-console-messages.js
```diff
@@ -12,7 +12,7 @@
 
 const Gatherer = require('./gatherer');
 
-class ChromeConsoleMessages extends Gatherer {
+class ConsoleMessages extends Gatherer {
```
paulirish: thanks :)

brendankenny:
(looks like there are a few more that are simple/stable that we can move over, but we aren't in a rush)

PR #8385, opened 2019-04-18
core: replace rawValue with numericValue in audit result

brendankenny:
part 2 of apparently 3 for #6199

Following #8343, this now renames the public `rawValue` on `LH.Audit.Result` to `numericValue`. We have very few places we rely on this except the smoke tests, so it's an easy change and a better place to bikeshed on `numericValue` and the public interface than the next PR :)

The last PR will be renaming the variable internal to audits/audit tests.

### DIFF
--- a/docs/understanding-results.md
+++ b/docs/understanding-results.md
@@ -55,7 +55,7 @@ An object containing the results of the audits, keyed by their name.
 | explanation | <code>string&#124;undefined</code> | A string indicating the reason for audit failure. |
 | warnings | <code>string[]&#124;undefined</code> | Messages identifying potentially invalid cases |
 | errorMessage | <code>string&#124;undefined</code> | A message set |
-| rawValue | <code>boolean&#124;number</code> | The unscored value determined by the audit. Typically this will match the score if there's no additional information to impart. For performance audits, this value is typically a number indicating the metric value. |
+| numericValue | <code>number&#124;undefined</code> | The unscored value determined by the audit. Typically this will match the score if there's no additional information to impart. For performance audits, this value is typically a number indicating the metric value. |
 | displayValue | `string` | The string to display in the report alongside audit results. If empty, nothing additional is shown. This is typically used to explain additional information such as the number and nature of failing items. |
 | score | <code>number</code> | The scored value determined by the audit as a number `0-1`, representing displayed scores of 0-100. |
 | scoreDisplayMode | <code>"binary"&#124;"numeric"&#124;"error"&#124;"manual"&#124;"notApplicable"&#124;"informative"</code> | A string identifying how the score should be interpreted for display i.e. is the audit pass/fail (score of 1 or 0), did it fail, should it be ignored, or are there shades of gray (scores between 0-1 inclusive). |
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -17,11 +17,8 @@ const fs = require('fs');
 
 /**
   * @param {string} result
-  * @param {{keepRawValues?: boolean}} opts
   */
-function processForProto(result, opts = {}) {
-  const {keepRawValues = false} = opts;
-
+function processForProto(result) {
   /** @type {LH.Result} */
   const reportJson = JSON.parse(result);
 
@@ -56,13 +53,8 @@ function processForProto(result, opts = {}) {
           audit.scoreDisplayMode = 'notApplicable';
         }
       }
-      // Drop numeric values until we decide what to do with the optional type in proto.
-      // https://github.com/GoogleChrome/lighthouse/issues/6199
-      if (!keepRawValues && 'numericValue' in audit) {
-        delete audit.numericValue;
-      }
-      // Normalize displayValue to always be a string, not an array. #6200
 
+      // Normalize displayValue to always be a string, not an array. #6200
       if (Array.isArray(audit.displayValue)) {
         /** @type {Array<any>}*/
         const values = [];
--- a/lighthouse-core/lib/proto-preprocessor.js
+++ b/lighthouse-core/lib/proto-preprocessor.js
@@ -56,9 +56,10 @@ function processForProto(result, opts = {}) {
           audit.scoreDisplayMode = 'notApplicable';
         }
       }
-      // Drop raw values. https://github.com/GoogleChrome/lighthouse/issues/6199
-      if (!keepRawValues && 'rawValue' in audit) {
-        delete audit.rawValue;
+      // Drop numeric values until we decide what to do with the optional type in proto.
+      // https://github.com/GoogleChrome/lighthouse/issues/6199
+      if (!keepRawValues && 'numericValue' in audit) {
+        delete audit.numericValue;
       }
       // Normalize displayValue to always be a string, not an array. #6200
 
--- a/lighthouse-core/test/lib/manifest-parser-test.js
+++ b/lighthouse-core/test/lib/manifest-parser-test.js
@@ -347,7 +347,6 @@ describe('Manifest Parser', function() {
       const display = parsedManifest.value.display;
       assert.ok(!display.warning);
       assert.equal(display.value, 'browser');
-      assert.equal(display.rawValue, undefined);
     });
 
     it('trims whitespace', () => {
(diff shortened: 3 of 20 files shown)


lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -56,9 +56,10 @@ function processForProto(result, opts = {}) {
           audit.scoreDisplayMode = 'notApplicable';
         }
       }
-      // Drop raw values. https://github.com/GoogleChrome/lighthouse/issues/6199
-      if (!keepRawValues && 'rawValue' in audit) {
-        delete audit.rawValue;
+      // Drop numeric values until we decide what to do with the optional type in proto.
```
brendankenny: punting on the question of `DoubleValue`/`Value`/whatever in proto by maintaining the deleting by default behavior of `rawValue`. We can change it to whatever later.

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -56,9 +56,10 @@ function processForProto(result, opts = {}) {
...
+      // https://github.com/GoogleChrome/lighthouse/issues/6199
+      if (!keepRawValues && 'numericValue' in audit) {
```
brendankenny: not sure how flexible we are in changing the `keepRawValues` parameter name in the lr entry point, but happy to rename if it's not a big deal

lighthouse-core/test/lib/manifest-parser-test.js
```diff
@@ -347,7 +347,6 @@ describe('Manifest Parser', function() {
       const display = parsedManifest.value.display;
       assert.ok(!display.warning);
       assert.equal(display.value, 'browser');
-      assert.equal(display.rawValue, undefined);
```
brendankenny: this made no sense, not a thing

patrickhulce:
I'm a little unclear on what the purpose of this PR is, are we planning on exposing an audit-level property called `numericValue`/whatever we decide on that has the numers and not the `details.timingMs` route or whatever?

If we are, then I think I was confused about what killing `rawValue` meant since this doesn't seem fundamentally different :)

If we aren't and this is all temporary, then I have absolutely no preference on the name so merge away 😆

brendankenny:
> are we planning on exposing an audit-level property called `numericValue`/whatever we decide on that has the numbers here and not the details.timingMs route or whatever?

That's what's this PR does but it's not necessarily the plan :)

I don't think we ever actually settled on a plan, so this is bikeshedding on property name and object structure. Fun for everyone!

I'll try to make a better case tomorrow, but basically I think a large portion of our programmatic users just need something like this, a single numeric value that's a companion to the score, and don't want (or need) to dig into details to do it. They're also not all ms numbers, so it's hard to name the value. I also realize that this doesn't really differentiate the property from `rawValue` except for the (already removed) primacy and booleaness of the property in the audit product :)

brendankenny:
(I am curious how many of these remaining rawValues are actually already covered by things in their `details`...)

benschwarz:
> I think a large portion of our programmatic users just need something like this, a single numeric value that's a companion to the score, and don't want (or need) to dig into details to do it. 

👬 Strong upvote from me on that @brendankenny

exterkamp:
>  a single numeric value that's a companion to the score

This is the concept.  While I've been trying to analyze variance this has come up a lot.  RawValue is not suitable since it is non-numeric and has a lot of legacy baggage.  So we wanted a numeric only value that could be used for analysis. e.g. TTFB score: 0.7, numericValue: 1159 (1159ms) | DOM size score: 0.9, numericValue: 14 (14 nodes).

lighthouse-core/lib/proto-preprocessor.js
exterkamp: I think it is going to have to be `Value` but this is something that we can punt 🏈

docs/understanding-results.md
```diff
@@ -55,7 +55,7 @@ An object containing the results of the audits, keyed by their name.
 | explanation | <code>string&#124;undefined</code> | A string indicating the reason for audit failure. |
 | warnings | <code>string[]&#124;undefined</code> | Messages identifying potentially invalid cases |
 | errorMessage | <code>string&#124;undefined</code> | A message set |
-| rawValue | <code>boolean&#124;number</code> | The unscored value determined by the audit. Typically this will match the score if there's no additional information to impart. For performance audits, this value is typically a number indicating the metric value. |
+| numericValue | <code>number&#124;undefined</code> | The unscored value determined by the audit. Typically this will match the score if there's no additional information to impart. For performance audits, this value is typically a number indicating the metric value. |
```
exterkamp: Should this be `number|null` like score?  I'm foreseeing the proto issue of default values on `undefined` going to 0.

patrickhulce:
> I think a large portion of our programmatic users just need something like this, a single numeric value that's a companion to the score, and don't want (or need) to dig into details to do it. 

As one of those programmatic users I also strongly upvote this :) 

SGTM!

lighthouse-core/lib/proto-preprocessor.js
paulirish: we can def change it.

lighthouse-core/lib/proto-preprocessor.js
paulirish: On the other hand.. this seems a great time to not punt on the proto typing for this prop. :) All our PSI API users currently don't get rawValue because we're punting and thus have to parse strings like "3,241ms" in order to get some metric values.

`.numericValue` being typed as the mega-generic proto `Value` seems real bad. 

What are the challenges to typing `.numericValue` as `DoubleValue`?

patrickhulce: Maybe I misunderstood but we are punting on adding the property at all to the proto right, not proposing to start with it as `Value`?

lighthouse-core/test/scoring-test.js
```diff
@@ -35,7 +35,7 @@ describe('ReportScoring', () => {
   describe('#scoreAllCategories', () => {
     it('should score the categories', () => {
       const resultsByAuditId = {
-        'my-audit': {rawValue: 'you passed', score: 0},
+        'my-audit': {score: 0},
         'my-boolean-audit': {score: 1, extendedInfo: {}},
```
patrickhulce: man, never thought I'd see the day where extendedInfo outlives rawValue 😆

lighthouse-core/lib/proto-preprocessor.js
exterkamp: Afaik DoubleValue doesn't support a difference between 'null' or 'undefined' and '0' since it is the default.  I can do some more testing around that, but thats the crux.  BUT if we don't care about undefined = 0, then we can for sure use a more typical and better proto type!

Hoten: Hold up - DoubleValue _is_ nullable. `double` and `float` are the ones that can't distinguish between 0 and not provided.

lighthouse-core/test/scoring-test.js
brendankenny: > man, never thought I'd see the day where extendedInfo outlives rawValue 😆

hahaha, maybe it's time to do the final purge on that

lighthouse-core/lib/proto-preprocessor.js
brendankenny: > we can def change it.

do you want me to change it now and LR will just have to remember to update on the next roll?

Hoten: well, if we are happy with numericValue now shouldn't we just always include it. remove the option entirely.

lighthouse-core/lib/proto-preprocessor.js
brendankenny: > Maybe I misunderstood but we are punting on adding the property at all to the proto right, not proposing to start with it as Value?

yeah, that's what I was proposing, not changing the status quo on proto, but we can do this for real

brendankenny: @exterkamp and I just tested a few possibilities and, yeah, `DoubleValue` will work well for us. `0` comes through, and then any audit with it not set will come back with no `numericValue` in JSON, which works just fine for us.

lighthouse-core/lib/proto-preprocessor.js
brendankenny: > well, if we are happy with numericValue now shouldn't we just always include it. remove the option entirely.

removed :)

PR #8390, opened 2019-04-18
report: remove html comments from templates

Hoten:
see https://github.com/GoogleChrome/lighthouse/pull/8307/files#r276452642

### DIFF
--- a/lighthouse-core/report/html/renderer/dom.js
+++ b/lighthouse-core/report/html/renderer/dom.js
@@ -78,16 +78,20 @@ class DOM {
   }
 
   /**
-   * @param {Node} node
+   * @param {Document} document
    */
-  removeCommentNodes(node) {
-    for (const child of Array.from(node.childNodes)) {
-      // Comment node
-      if (child.nodeType === 8) {
-        child.remove();
-      // Element node
-      } else if (child.nodeType === 1) {
-        this.removeCommentNodes(child);
+  removeCommentNodes(document) {
+    const walker = document.createTreeWalker(document, 128, null);
+
+    const commentNodes = [];
+    let currentNode;
+    while (currentNode = walker.nextNode()) {
+      commentNodes.push(currentNode);
+    }
+
+    for (const node of commentNodes) {
+      if (node.parentNode) {
+        node.parentNode.removeChild(node);
       }
     }
   }
@@ -113,7 +117,6 @@ class DOM {
     }
     template.setAttribute('data-stamped', 'true');
 
-    this.removeCommentNodes(clone);
     return clone;
   }
 
--- a/lighthouse-core/report/html/renderer/dom.js
+++ b/lighthouse-core/report/html/renderer/dom.js
@@ -81,7 +81,9 @@ class DOM {
    * @param {Document} document
    */
   removeCommentNodes(document) {
-    const walker = document.createTreeWalker(document, 128, null);
+    // NodeFilter is undefined in node context.
+    const NodeFilter_SHOW_COMMENT = 128;
+    const walker = document.createTreeWalker(document, NodeFilter_SHOW_COMMENT, null);
 
     const commentNodes = [];
     let currentNode;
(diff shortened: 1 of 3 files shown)


lighthouse-core/report/html/renderer/dom.js
```diff
@@ -77,6 +77,21 @@ class DOM {
     return element;
   }
 
+  /**
+   * @param {Node} node
+   */
+  removeCommentNodes(node) {
+    for (const child of Array.from(node.childNodes)) {
```
paulirish: wdyt about using the TreeWalker API?

https://stackoverflow.com/a/16152333

I just tried it out and this current implementation takes ~40ms for an entire report DOM. And the TW snippet takes 0.1ms. Plus it reads pretty nice.

patrickhulce: TIL :)

Hoten: ok, here's what changed...

TW only works on documents, so I move it up to the entire Doc (instead of each template). This now removes every comment, including the license in the html. Should I fix that?

some craziness with tests (jsdom's document doesn't support innerHtml)

treewalker, just like childNodes, doesn't like it when you mutate during iteration. silly web apis don't error when you do that so that was annoying to figure out.

lighthouse-core/report/html/renderer/dom.js
```diff
@@ -77,6 +77,25 @@ class DOM {
     return element;
   }
 
+  /**
+   * @param {Document} document
+   */
+  removeCommentNodes(document) {
+    const walker = document.createTreeWalker(document, 128, null);
```
paulirish: 128 is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants). let's use `NodeFilter.SHOW_COMMENT`

lighthouse-core/report/html/renderer/dom.js
```diff
@@ -77,6 +77,25 @@ class DOM {
...
+
+    const commentNodes = [];
+    let currentNode;
+    while (currentNode = walker.nextNode()) {
+      commentNodes.push(currentNode);
+    }
+
+    for (const node of commentNodes) {
```
paulirish: for (const node of commentNodes) node.remove();

lighthouse-core/report/html/renderer/dom.js
Hoten: I did, but that fails in Node (during tests).

```
ReferenceError: NodeFilter is not defined
```

compromised w/ a const.

lighthouse-core/report/html/renderer/dom.js
Hoten: ok. had to convince tsc.

brendankenny:
this doesn't appear to be removing any comments except the top-level license, btw, at least in the devtools element panel and in `document.documentElement.outerHTML`.

two questions:
- do we really care? :) everyone has dumb comments in their markup
  (though more contemplative ones like "should likely be an h2" or "don't reuse group classes for clumps" should really be in an issue instead of a comment :)
- where are we anticipating folks seeing these so we want to remove them? Like, when you view it in devtools? Cause obviously if you do an old fashioned view source they'll still be in there.

Hoten:
> this doesn't appear to be removing any comments except the top-level license, btw, at least in the devtools element panel and in `document.documentElement.outerHTML`.

oops. had to move after the render. 

Even then, it doesn't traverse into templates.

> two questions:
> 
> * do we really care? :) everyone has dumb comments in their markup
>   (though more contemplative ones like "should likely be an h2" or "don't reuse group classes for clumps" should really be in an issue instead of a comment :)
> * where are we anticipating folks seeing these so we want to remove them? Like, when you view it in devtools? Cause obviously if you do an old fashioned view source they'll still be in there.

excellent points.

patrickhulce:
I think I agree with @brendankenny on this one. I'm not sure what the value is, but it was cool to learn about TreeWalker amirite 😃

paulirish:
Alright fine. :)

Well we all learned some cool hax to remove comments should we ever need them.

Thx for trying this out anyhow, @hoten

PR #8415, opened 2019-04-18
i18n: import stack pack translations

exterkamp:
**Summary**
Importing new i18n strings as they come in.  In this import: stack packs.
image

SEO groups will be imported in another PR, as it is separate in tc/

### DIFF
--- a/lighthouse-core/test/lib/i18n/i18n-test.js
+++ b/lighthouse-core/test/lib/i18n/i18n-test.js
@@ -59,6 +59,7 @@ describe('i18n', () => {
     it('returns icu messages in the specified locale', () => {
       const strings = i18n.getRendererFormattedStrings('en-XA');
       expect(strings.passedAuditsGroupTitle).toEqual('[Þåššéð åûðîţš one two]');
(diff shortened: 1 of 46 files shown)


lighthouse-core/test/lib/i18n/i18n-test.js
```diff
@@ -59,7 +59,6 @@ describe('i18n', () => {
     it('returns icu messages in the specified locale', () => {
       const strings = i18n.getRendererFormattedStrings('en-XA');
       expect(strings.passedAuditsGroupTitle).toEqual('[Þåššéð åûðîţš one two]');
-      expect(strings.scorescaleLabel).toEqual('[Šçöŕé šçåļé: one two]');
```
brendankenny: can you replace this rather than removing it

exterkamp: Done.

exterkamp:
> > SEO groups will be imported in another PR, as it is separate in tc/
> 
> seems like this isn't true? Are they all in here?

They are all in here in English as only lang for all langs (artifact of tc/ since they are being translated they appear as English in all langs).  They just aren't translated yet.  That's the follow up.

brendankenny:
> They just aren't translated yet

oh, I see

PR #8417, opened 2019-04-18
style: Removing a dead flag

asteinha:
See https://chromium.googlesource.com/chromium/src.git/+/ba49d00fddeb39c0816090dad7d359df0dce9a1a
Code cleanup.


Hoten:
Better move would be to update these fixtures. those trace files are uses in tests w/o running Chrome at all.

brendankenny:
Hi @asteinha,

Thanks for your PR over on `chrome-launcher`!

For these files in `lighthouse`, though, the traces are really just records of what they were at the time they were run, including those command line flags. You'll see a lot of other out of date stuff in there too, like the fact they were run with Chrome 60 :)

They're used as saved test inputs, not used for running future invocations of Lighthouse, so no need to update them in this way

Issue #458, opened 2016-06-22
Resolve audit result values


paulirish:
#### Currently...

* `result.value` - most of the time is actually === a score
* `result.rawValue` - many times its a string of "1232.32ms"  .. not exactly "raw"

#### Proposal 
* `result.score`
* `result.rawValue` - which is either a number or boolean, never a string
*  `result.displayValue` - which is a string

brendankenny:
LGTM

paullewis:
LGTM

wardpeet:
I'm working on this ticket as we speak but what should happen with rawValue when it's a color? should it just say true?

paulirish:
> I'm working on this ticket as we speak but what should happen with rawValue when it's a color? should it just say true?

Theme-color?  TBH it's kinda superfluous that we report it. The rawValue there should be boolean. Can move the color into `extendedInfo`

> What with score false? Should it convert to 0 and have a rawValue of false and displayValue of ''?

shrug...

wardpeet:
I'll just prepare a pr and i'll just wait for feedback :p

wardpeet:
Fixing some bugs, almost done so will have a pr ready today or tomorrow

paulirish:
awesome! thanks for the update man

PR #74, opened 2016-03-24
Adds aggregation

paullewis:
Makes the output more inline with what matters to a user, e.g. 'Will get an Add to Homescreen prompt'

screen shot 2016-03-24 at 1 05 18 pm

:boom: 

PTAL @paulirish - I think this matches what we discussed. Also note that the `will-get-add-to-homescreen-prompt` is a kind-of "composite aggregation", which should aid testing.

Issue #61, opened 2016-03-23
Spec the return value of audits


paulirish:
Brendan is put up a strawman for this.

It will include warning/severity levels and handling atomic results.

paullewis:
Cool, we need to make sure that we know how to bubble up to the top-level category / tag. For example, if you fail a manifest audit, what does that say for its tags? What if the category has 10 audits, and you fail 1?

brendankenny:
I've been looking over some of the equivalents of our architecture within catapult and I'm increasingly not sure here. At this point I don't know if we can plan a declarative tag structure that's sufficient to express all we'll need for our audits, as https://github.com/GoogleChrome/lighthouse/issues/61#issuecomment-200306979 anticipates. There's some evidence of this happening within catapult, with tacked on meta information where it looks (at first impression) like there were a few cases of "oh, we actually need to pipe one more thing through to evaluate this" happening :)

At this point I'd say I want us to avoid ending up writing an audits of audits layer here (and an audits of audits of audits layer...) until (and if) we really really need them, which means it's going to be hard to express categories with tags beyond ones that are scored with a simple combination of boolean states or subscores.

Sooo...barring a magical perf taxonomy handed to us on some stone tablets, maybe it's best to just avoid any architecture astronautism and continue with the more straightforward path we're on now. I could be very wrong on the above and tags will work just fine :) I think this will become increasingly clear one way or the other, and we just need to settle on something simple for now so we can continue with some auditing momentum:

```
Promise.resolve({
  rawValue: {}, // could be a primitive value or object, as long as it can easily be stringified and displayed
  score: 0 // boolean or number
  tags: ['what have you'],
  description: 'whatnot',
  warning: 'some *specific* error value worth display, e.g. feedback on how to fix what's wrong' // although the reporter can handle *general* feedback on how to fix, e.g. links to the docs
});
```

After walking through some of catapult this afternoon, I feel like I have less insight here than I had before, so I'm interested in others' thoughts :)

brendankenny:
Where we currently are:
```
Promise.resolve({
  value: 0, // "score", in our current use. boolean or number
  rawValue: {}, // could be anything, as long as it can easily be stringified and displayed, e.g. 'your score is bad because you wrote ${rawValue}'
  debugString: 'some *specific* error string for helping the user figure out why they failed here' // e.g. 'your manifest 404ed'. The reporter can handle *general* feedback on how to fix, e.g. links to the docs
  name: 'audit-name',
[...]

PR #8421, opened 2019-04-19
core: replace remaining rawValue with numericValue

### LINKED ISSUE #6199: Remove rawValue from Audit.Result

brendankenny:
part 3 of 3 after #8343 and #8385

Replaces `rawValue` with `numericValue` in the internal audit output (`LH.Audit.Product`). Went suspiciously easily :)

fixes #6199

### DIFF
--- a/docs/recipes/lighthouse-plugin-example/audits/preload-as.js
+++ b/docs/recipes/lighthouse-plugin-example/audits/preload-as.js
@@ -37,7 +37,7 @@ class PreloadAsAudit extends Audit {
     const passed = noAsLinks.length === 0;
 
     return {
-      rawValue: passed,
+      score: passed ? 1 : 0,
       displayValue: `Found ${noAsLinks.length} preload requests with missing \`as\` attributes`,
     };
   }
--- a/lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
+++ b/lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
@@ -26,7 +26,7 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
     const settings = {throttlingMethod: 'devtools', throttling: mobileSlow4GThrottling};
     return FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()}).then(result => {
       assert.equal(result.score, true, 'fixture trace is not passing audit');
-      assert.equal(result.rawValue, 1582.189);
+      assert.equal(result.numericValue, 1582.189);
     });
   });
 
@@ -50,7 +50,7 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
     const settings = {throttlingMethod: 'devtools', throttling: mobileSlow4GThrottling};
     return FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()}).then(result => {
       assert.equal(result.score, false, 'not failing a long TTI value');
-      assert.equal(result.rawValue, 15000);
+      assert.equal(result.numericValue, 15000);
       expect(result.displayValue).toBeDisplayString('Interactive at 15.0\xa0s');
       assert.ok(result.explanation);
     });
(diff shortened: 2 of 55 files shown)


lighthouse-core/test/audits/metrics/first-contentful-paint-3g-test.js
```diff
@@ -26,7 +26,7 @@ describe('Performance: first-contentful-paint-3g audit', () => {
 
     const result = await FCP3G.audit(artifacts, {settings: {}, options, computedCache: new Map()});
     // Use InlineSnapshot here so changes to Lantern coefficients can be easily updated en masse
```
brendankenny: `:roll_eyes:` :P

lighthouse-core/test/audits/load-fast-enough-for-pwa-test.js
```diff
@@ -75,8 +75,7 @@ describe('PWA: load-fast-enough-for-pwa audit', () => {
 
     const settings = {throttlingMethod: 'provided', throttling: {rttMs: 40, throughput: 100000}};
     const result = await FastPWAAudit.audit(artifacts, {settings, computedCache: new Map()});
-    expect(result.rawValue).toBeGreaterThan(2000); // If not overridden this would be 1582
-    expect(Math.round(result.rawValue)).toMatchSnapshot();
```
brendankenny: wasn't sure why this was this way, but just a single number, so removed

patrickhulce: @brendankenny it's a lantern computed value it does it this way so that lantern changes can be updated with `-u` in jest

patrickhulce: could you revert with a comment?

lighthouse-core/test/audits/metrics/first-contentful-paint-3g-test.js
patrickhulce: oh puh-lease 😛 there's like 40 freaking values to manually and copy and paste when lantern changes and I've had to do it like 8 times. do you want to volunteer to manually fix every expectation every time this happens from here on out?

brendankenny:
In memoriam:

The earliest reference to `rawValue` I can find is in [#61](https://github.com/GoogleChrome/lighthouse/issues/61#issuecomment-200695806) in a proposed audit output format, where you can see me worrying about tags (what were those going to be?) and the complexity of a possible emerging "audits of audits" architecture. We don't discuss the `rawValue` name, so it may have already been decided upon.

Also we were going to have `auditResult.fault` as the explanation of why the audit screwed up and couldn't handle your site, which is the modern day `LH.Audit.Result['warnings']`. I kind of like `fault` :)

The types landed a day later in the great [Aggregator PR](https://github.com/GoogleChrome/lighthouse/pull/74/files#diff-d01d9d70eddf296180be0003f332eb0aR35) from @paullewis. At that point, audits had a `value` and a `rawValue`.

`rawValue` rapidly became the dumping ground for whatever seemed raw-ish. Finally, a few months 
 later, there's issue #458 from @paulirish suggesting we rename `value` to `score`, and describing `rawValue` as `not exactly "raw"` :) That issue birthed `displayValue` and gave us a refined `rawValue` of `either a number or boolean [and eventually null], never a string` that we kept around for some reason for another three years :)

docs/recipes/lighthouse-plugin-example/audits/preload-as.js
```diff
@@ -37,7 +37,7 @@ class PreloadAsAudit extends Audit {
     const passed = noAsLinks.length === 0;
 
     return {
-      rawValue: passed,
+      score: passed ? 1 : 0,
```
Hoten: `+passed`? idk how hacky you'll allow js to be :)

patrickhulce: I like the explicitness of this for the recipe :)

brendankenny:
@patrickhulce you're still requesting changes

patrickhulce:
Oh my bad sorry thought I came back to this already :)

PR #8450, opened 2019-04-19
feat(lightwallet): Add Details.LightWallet

khempenius:
**Summary**
Adds details for LightWallet audit.

**Related Issues/PRs**
https://github.com/GoogleChrome/lighthouse/pull/8427

### DIFF
--- a/types/audit-details.d.ts
+++ b/types/audit-details.d.ts
@@ -13,7 +13,8 @@ declare global {
       Details.List |
       Details.Opportunity |
       Details.Screenshot |
-      Details.Table;
+      Details.Table |
+      Details.LightWallet;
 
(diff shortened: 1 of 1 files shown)


types/audit-details.d.ts
```diff
@@ -172,6 +173,34 @@ declare global {
         value: string;
       }
 
+      export interface LightWallet {
```
khempenius: My thinking here was to have the Details.LightWallet format mirror the budgets format as closely as possible bc that’s probably not going to change much but UX could.

Keeping the info for the diagnostics section separate from budget results made the audit/renderer logic much cleaner, even if the info is a bit duplicative in some cases.

PR #8514, opened 2019-04-22
report(tables): change table layout to fixed to ensure proper display…

ahmadalfy:
Change table layout to fixed to ensure proper display of all columns

**Summary**
This is a fix for layout bug in the generated report. Currently when the number of characters in the URLs exceed the allowed space it results in a display that looks like this:

image

As you can see, the information for "Potential Savings (KB)" is pushed and cannot be displayed. Fixing that by changing the table layout to `fixed` fix the issue.

image

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -987,6 +987,7 @@
 .lh-table {
   --image-preview-size: 24px;
   border-collapse: collapse;
+  table-layout: fixed;
 }
 
 .lh-table thead {


ahmadalfy:
Thank you, sorry if I merged to master directly should I create a branch and resubmit the PR? I didn't find that in the contribution guide.

brendankenny:
I know @paulirish has looked at the table layouts before...

brendankenny:
we're going to wait on this until after the rest of #8185 lands to make table changes

paulirish:
@ahmadalfy thx very much for writing this. I ended up going a bit deeper and sorting this out in #8820

PR #8524, opened 2019-04-22
report(redesign): add sticky scores header

Hoten:
#8185

sticky header w/ mini gauges
...appears when scrolling past score scale
...current category in view is underlined

### DIFF
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -256,8 +256,9 @@ class ReportRenderer {
     // }
 
     if (scoreHeader) {
-      scoreHeader.append(
-        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
+      const scoreGauges =
+        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
+      scoreHeader.append(...scoreGauges);
       const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
       const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
       scoresContainer.appendChild(scoreHeader);
@@ -277,10 +278,9 @@ class ReportRenderer {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
 
-      // The sticky header is just the score gauges, but styled to be smaller. Just
-      // clone the gauges from the score header.
-      stickyHeader.append(
-        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
+      const scoreGauges =
+        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
+      stickyHeader.append(...scoreGauges);
 
       reportFragment.appendChild(stickyHeader);
     }
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -159,6 +159,39 @@ class ReportRenderer {
 
   /**
    * @param {LH.ReportResult} report
+   * @param {CategoryRenderer} categoryRenderer
+   * @param {Record<string, CategoryRenderer>} specificCategoryRenderers
+   * @return {DocumentFragment[]}
+   */
+  _renderScoreGauges(report, categoryRenderer, specificCategoryRenderers) {
+    // Group gauges in this order: default, pwa, plugins.
+    const defaultGauges = [];
+    const customGauges = []; // PWA.
+    const pluginGauges = [];
+
+    for (const category of report.reportCategories) {
+      const renderer = specificCategoryRenderers[category.id] || categoryRenderer;
+      const categoryGauge = renderer.renderScoreGauge(category, report.categoryGroups || {});
+
+      if (Util.isPluginCategory(category.id)) {
+        pluginGauges.push(categoryGauge);
+      } else if (renderer.renderScoreGauge === categoryRenderer.renderScoreGauge) {
+        // The renderer for default categories is just the default CategoryRenderer.
+        // If the functions are equal, then renderer is an instance of CategoryRenderer.
+        // For example, the PWA category uses PwaCategoryRenderer, which overrides
+        // CategoryRenderer.renderScoreGauge, so it would fail this check and be placed
+        // in the customGauges bucket.
+        defaultGauges.push(categoryGauge);
+      } else {
+        customGauges.push(categoryGauge);
+      }
+    }
+
+    return [...defaultGauges, ...customGauges, ...pluginGauges];
+  }
+
+  /**
+   * @param {LH.ReportResult} report
    * @return {DocumentFragment}
    */
   _renderReport(report) {
@@ -223,29 +256,8 @@ class ReportRenderer {
     // }
 
     if (scoreHeader) {
-      // Group gauges in this order: default, pwa, plugins.
-      const defaultGauges = [];
-      const customGauges = []; // PWA.
-      const pluginGauges = [];
-      for (const category of report.reportCategories) {
-        const renderer = specificCategoryRenderers[category.id] || categoryRenderer;
-        const categoryGauge = renderer.renderScoreGauge(category, report.categoryGroups || {});
-
-        if (Util.isPluginCategory(category.id)) {
-          pluginGauges.push(categoryGauge);
-        } else if (renderer.renderScoreGauge === categoryRenderer.renderScoreGauge) {
-          // The renderer for default categories is just the default CategoryRenderer.
-          // If the functions are equal, then renderer is an instance of CategoryRenderer.
-          // For example, the PWA category uses PwaCategoryRenderer, which overrides
-          // CategoryRenderer.renderScoreGauge, so it would fail this check and be placed
-          // in the customGauges bucket.
-          defaultGauges.push(categoryGauge);
-        } else {
-          customGauges.push(categoryGauge);
-        }
-      }
-      scoreHeader.append(...defaultGauges, ...customGauges, ...pluginGauges);
-
+      scoreHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
       const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
       const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
       scoresContainer.appendChild(scoreHeader);
@@ -267,9 +279,8 @@ class ReportRenderer {
 
       // The sticky header is just the score gauges, but styled to be smaller. Just
       // clone the gauges from the score header.
-      for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
-        stickyHeader.appendChild(gaugeWrapperEl.cloneNode(true));
-      }
+      stickyHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
 
       reportFragment.appendChild(stickyHeader);
     }
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -255,6 +255,7 @@ class ReportRenderer {
 
     if (scoreHeader && !this._dom.isDevTools()) {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
 
       for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
         const gaugeWrapperElClone = gaugeWrapperEl.cloneNode(true);
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -257,9 +257,10 @@ class ReportRenderer {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
 
+      // The sticky header is just the score gauges, but styled to be smaller. Just
+      // clone the gauges from the score header.
       for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
-        const gaugeWrapperElClone = gaugeWrapperEl.cloneNode(true);
-        stickyHeader.appendChild(gaugeWrapperElClone.cloneNode(true));
+        stickyHeader.appendChild(gaugeWrapperEl.cloneNode(true));
       }
 
       reportFragment.appendChild(stickyHeader);
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -247,10 +247,24 @@ class ReportRenderer {
     reportSection.appendChild(this._renderReportFooter(report));
 
     const reportFragment = this._dom.createFragment();
+
     if (!this._dom.isDevTools()) {
       const topbarDocumentFragment = this._renderReportTopbar(report);
       reportFragment.appendChild(topbarDocumentFragment);
     }
+
+    if (scoreHeader && !this._dom.isDevTools()) {
+      const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      stickyHeader.classList.add('stuck');
+
+      for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
+        const gaugeWrapperElClone = gaugeWrapperEl.cloneNode(true);
+        stickyHeader.appendChild(gaugeWrapperElClone.cloneNode(true));
+      }
+
+      reportFragment.appendChild(stickyHeader);
+    }
+
     reportFragment.appendChild(headerContainer);
     reportFragment.appendChild(container);
 
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -255,7 +255,6 @@ class ReportRenderer {
 
     if (scoreHeader && !this._dom.isDevTools()) {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
-      stickyHeader.classList.add('stuck');
 
       for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
         const gaugeWrapperElClone = gaugeWrapperEl.cloneNode(true);
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -406,6 +406,7 @@ class ReportUIFeatures {
     const scoreScaleTop = this.scoreScaleEl.getBoundingClientRect().top;
     const showStickyHeader = topbarBottom >= scoreScaleTop;
     this.stickyHeaderEl.classList.toggle('lh-sticky-header--visible', showStickyHeader);
+    if (!showStickyHeader) return;
 
     // Highlight mini gauge when section is in view.
     // In view = the last category that starts above the middle of the window.
@@ -416,8 +417,6 @@ class ReportUIFeatures {
       categoriesAboveTheMiddle.length > 0 ? categoriesAboveTheMiddle.length - 1 : 0;
 
     // Category order matches gauge order in sticky header.
-    // TODO(hoten): not 100% true yet, need to order gauges like: core, pwa, plugins. Remove
-    // this comment when that is done.
     const gaugeWrapperEls = this.stickyHeaderEl.querySelectorAll('.lh-gauge__wrapper');
     const gaugeToHighlight = gaugeWrapperEls[highlightIndex];
     this.highlightEl.style.left = gaugeToHighlight.getBoundingClientRect().left + 'px';
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -405,8 +405,6 @@ class ReportUIFeatures {
     const topbarBottom = this.topbarEl.getBoundingClientRect().bottom;
     const scoreScaleTop = this.scoreScaleEl.getBoundingClientRect().top;
     const showStickyHeader = topbarBottom >= scoreScaleTop;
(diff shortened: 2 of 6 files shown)


patrickhulce:
Nice!! The only thing that jumps out to me is that the plugin gets underlined before the PWA category as you scroll and the PWA only gets underlined after you scroll past it. Seems like they should be flipped but I guess I'll find out in the code in a second :)

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -247,10 +247,24 @@ class ReportRenderer {
...
     if (!this._dom.isDevTools()) {
       const topbarDocumentFragment = this._renderReportTopbar(report);
       reportFragment.appendChild(topbarDocumentFragment);
     }
+
+    if (scoreHeader && !this._dom.isDevTools()) {
+      const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
+
+      for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
```
patrickhulce: this is creating a copy of all the gauges in the score header and sticking them into the sticky header, right?

It took me halfway writing out a different comment to figure out what this was doing (until I noticed the `, scoreHeader` 😆 ). Maybe a comment explaining it would help?

I like the savings of not having to recompute all the gauge details just took some thought to realize the cleverness :)

lighthouse-core/report/html/templates.html
```diff
@@ -144,10 +144,85 @@
...
+      width: var(--score-container-width);
+      height: 1px;
+      background: var(--color-highlighter-bg);
+      position: absolute;
+      bottom: -1px;
+      left: 0;
+    }
   </style>
+  <script>
+    function handleStickyHeader() {
```
patrickhulce: This is the first `script` we have in the templates to my knowledge. Was there a particular reason it's here instead of in `report-ui-features` or another dedicated js file?

lighthouse-core/report/html/templates.html
```diff
@@ -144,10 +144,85 @@
...
+      const topbarEl = document.querySelector('.lh-topbar');
+      const scoreScaleEl = document.querySelector('.lh-scorescale');
+      const stickyHeaderEl = document.querySelector('.lh-sticky-header');
+      const highlightEl = document.querySelector('.lh-highlighter');
+
+      // Show sticky header when the score scale begins to go underneath the topbar.
+      const showStickyHeader =
+        topbarEl.getBoundingClientRect().bottom - scoreScaleEl.getBoundingClientRect().top >= 0;
+      stickyHeaderEl.classList.toggle('stuck', showStickyHeader);
```
patrickhulce: can we use `lh-sticky-header--stuck`? :)

lighthouse-core/report/html/templates.html
```diff
@@ -144,10 +144,85 @@
...
+
+      // Highlight mini gauge when section is in view.
+      // Use the middle of the viewport as an anchor - the closest category to the middle
+      // is the one "in view".
+      let highlightIndex = 0;
+      let highlightIndexDistance = Number.POSITIVE_INFINITY;
+      for (const [index, categoryEl] of Object.entries(document.querySelectorAll('.lh-category'))) {
```
patrickhulce: I think this index-based approach is what is causing the behavior I was seeing with plugin and PWA flipped. Do we want the gauges to match their appearance order in the report instead?

lighthouse-core/report/html/templates.html
```diff
@@ -144,10 +144,85 @@
...
+        const distanceToMiddle = categoryEl.getBoundingClientRect().top - window.innerHeight / 2;
+        // Closest negative distance to zero wins.
+        if (distanceToMiddle < 0 && highlightIndexDistance > distanceToMiddle) {
+          highlightIndex = index;
+          highlightIndexDistance = -distanceToMiddle;
+        }
+      }
+
+      // Category order matches gauge order in sticky header.
+      // TODO(hoten): not 100% true yet, need to order gauges like: core, pwa, plugins. Remove
```
patrickhulce: oh, I see :) 

historically comments without a direct issue link tend to be forgotten, is that PR like the next one though?

Hoten: There is a task entry in the report-redesign issue for this.

lighthouse-core/report/html/templates.html
Hoten: just confirming you are correct in your next comment - this code will work as expected when the gauge order is as expected.

lighthouse-core/report/html/templates.html
Hoten: moved to report-ui-features

lighthouse-core/report/html/templates.html
patrickhulce: ah of course!

lighthouse-core/report/html/templates.html
```diff
@@ -144,10 +144,49 @@
...
+      padding-top: var(--score-container-padding);
+      padding-bottom: 8px;
+      z-index: 1000;
+      pointer-events: none;
+      opacity: 0;
+    }
+
+    .lh-sticky-header--stuck {
+      pointer-events: auto;
+      opacity: 1;
```
patrickhulce: I guess this is really only messing with visibility, huh? 

I wish I would've thought of `--visible` the first time around :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +386,42 @@ class ReportUIFeatures {
...
+    // Show sticky header when the score scale begins to go underneath the topbar.
+    const showStickyHeader =
+      topbarEl.getBoundingClientRect().bottom - scoreScaleEl.getBoundingClientRect().top >= 0;
+    stickyHeaderEl.classList.toggle('lh-sticky-header--stuck', showStickyHeader);
+
+    // Highlight mini gauge when section is in view.
+    // Use the middle of the viewport as an anchor - the closest category to the middle
+    // is the one "in view".
+    let highlightIndex = 0;
+    let highlightIndexDistance = Number.POSITIVE_INFINITY;
```
patrickhulce: just curious, why not `Infinity`? :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +386,42 @@ class ReportUIFeatures {
...
+    const categoryEls = this._document.querySelectorAll('.lh-category');
+    for (const [index, categoryEl] of Object.entries(categoryEls)) {
+      // Normalize to middle of viewport.
+      const distanceToMiddle = categoryEl.getBoundingClientRect().top - window.innerHeight / 2;
+      // Closest negative distance to zero wins.
+      if (distanceToMiddle < 0 && highlightIndexDistance > distanceToMiddle) {
```
patrickhulce: seems like we could simplify some of this but maybe I'm misunderstanding

- `highlightIndexDistance` will only ever update when `distanceToMiddle` is negative
- `highlightIndexDistance` is `-distanceToMiddle`, so `highlightIndexDistance` is always positive
-  `highlightIndexDistance` will always be greater than `distanceToMiddle` if `distanceToMiddle` is negative, so it's really just `if (distanceToMiddle < 0)`

this seems equivalent to "we highlight the last category that starts above the middle of the window" which at least for me is a lot easier to understand.

maybe something like the below?
```js
const categoriesAboveTheMiddle = categories.filter(el => el.getBoundingClientRect().top - window.innerHeight / 2)
const categoryToHighlight = categoriesAboveTheMiddle[categoriesAboveTheMiddle.length - 1]
const guageToHighlight = gauges[categories.indexOf(categoryToHighlight)]

File: lighthouse-core/report/html/renderer/report-ui-features.js [hunk 9]
```diff
@@ -380,6 +386,42 @@ class ReportUIFeatures {
...
+        highlightIndex = parseInt(index);
+        highlightIndexDistance = -distanceToMiddle;
+      }
+    }
+
+    // Category order matches gauge order in sticky header.
+    // TODO(hoten): not 100% true yet, need to order gauges like: core, pwa, plugins. Remove
+    // this comment when that is done.
+    const gaugeToHighlight = stickyHeaderEl.querySelectorAll('.lh-gauge__wrapper')[highlightIndex];
+    // @ts-ignore
```
patrickhulce: what are we ignoring?

Hoten: `Property 'style' does not exist on type 'Element'` idk why

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: this works:

```js
// Highlight mini gauge when section is in view.
// In view = the last category that starts above the middle of the window.
const categoryEls = Array.from(this._document.querySelectorAll('.lh-category'));
const categoriesAboveTheMiddle =
  categoryEls.filter(el => el.getBoundingClientRect().top - window.innerHeight / 2 < 0);
const highlightIndex =
  categoriesAboveTheMiddle.length > 0 ? categoriesAboveTheMiddle.length - 1 : 0;

// Category order matches gauge order in sticky header.
// TODO(hoten): not 100% true yet, need to order gauges like: core, pwa, plugins. Remove
// this comment when that is done.
const gaugeToHighlight = stickyHeaderEl.querySelectorAll('.lh-gauge__wrapper')[highlightIndex];
// @ts-ignore
highlightEl.style.left = gaugeToHighlight.getBoundingClientRect().left + 'px';
```

thanks for making more declarative, i'll update

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: needs to be HTMLElement

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +386,34 @@ class ReportUIFeatures {
...
+    const categoriesAboveTheMiddle =
+      categoryEls.filter(el => el.getBoundingClientRect().top - window.innerHeight / 2 < 0);
+    const highlightIndex =
+      categoriesAboveTheMiddle.length > 0 ? categoriesAboveTheMiddle.length - 1 : 0;
+
+    // Category order matches gauge order in sticky header.
+    // TODO(hoten): not 100% true yet, need to order gauges like: core, pwa, plugins. Remove
+    // this comment when that is done.
+    const gaugeToHighlight = stickyHeaderEl.querySelectorAll('.lh-gauge__wrapper')[highlightIndex];
+    (/** @type {HTMLElement} */ (highlightEl)).style.left =
```
patrickhulce: Ah yeah I guess `querySelector` can return things that aren't HTMLElements, bummer! Maybe make the assertion at the top where it's declared?

Hoten: just refactored to use _dom + make member vars

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -61,10 +70,16 @@ class ReportUIFeatures {
     this.json = report;
     this._setupMediaQueryListeners();
     this._setupExportButton();
+    this._setupStickyHeaderElements();
     this._setUpCollapseDetailsAfterPrinting();
     this._resetUIState();
     this._document.addEventListener('keydown', this.printShortCutDetect);
     this._document.addEventListener('copy', this.onCopy);
+    this._document.addEventListener('scroll', this._handleStickyHeader);
+    // window.addEventListener is undefined in jest tests.
```
brendankenny: which test file(s) fails? This sounds like jsdom wasn't set up correctly

lighthouse-core/test/report/html/renderer/report-renderer-test.js
```diff
@@ -80,7 +80,7 @@ describe('ReportRenderer', () => {
       assert.ok(output.querySelector('.lh-header-sticky'), 'has a header');
       assert.ok(output.querySelector('.lh-report'), 'has report body');
       assert.equal(output.querySelectorAll('.lh-gauge__wrapper, .lh-gauge--pwa__wrapper').length,
-          sampleResults.reportCategories.length * 2, 'renders category gauges');
+          sampleResults.reportCategories.length * 3, 'renders category gauges');
```
brendankenny: maybe add a comment for where this multiplier comes from

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -247,10 +247,25 @@ class ReportRenderer {
...
       reportFragment.appendChild(topbarDocumentFragment);
     }
+
+    if (scoreHeader && !this._dom.isDevTools()) {
+      const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
+
+      // The sticky header is just the score gauges, but styled to be smaller. Just
+      // clone the gauges from the score header.
+      for (const gaugeWrapperEl of this._dom.findAll('.lh-gauge__wrapper', scoreHeader)) {
```
brendankenny: this feels pretty squicky to me :)

Can we make a `renderScoresHeader()` method and use it in both places?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +402,29 @@ class ReportUIFeatures {
     this._document.body.removeChild(a);
     setTimeout(_ => URL.revokeObjectURL(href), 500);
   }
+
+  _handleStickyHeader() {
```
brendankenny: maybe rename to make it clear this is for adjusting the sticky header on scroll events?

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: yup that was it, fixed.

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: It handles sticky header visibility, and the highlighting. It also fires on scroll and resize.

`_handleStickyHeaderVisibilityAndHighlighter`? :)

brendankenny: haha, I was thinking more like `_updateStickyHeaderOnScroll` or whatever :)

Arguably (it's a stretch :) resize is a special case of scrolling for this element, resizing is uncommon compared to scrolling, and the main thing is the name can communicate something about what it's up to and when

lighthouse-core/report/html/renderer/report-renderer.js
Hoten: done

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: sold, to `_updateStickyHeaderOnScroll`

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -157,6 +157,39 @@ class ReportRenderer {
     return container;
   }
 
+  /**
+   * @param {LH.ReportResult} report
+   * @param {CategoryRenderer} categoryRenderer
+   * @param {Record<string, CategoryRenderer>} specificCategoryRenderers
```
brendankenny: `categoryRenderer` and `specificCategoryRenderers` probably need docstrings (sorry for that second name :)

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -223,29 +256,8 @@ class ReportRenderer {
...
-          // in the customGauges bucket.
-          defaultGauges.push(categoryGauge);
-        } else {
-          customGauges.push(categoryGauge);
-        }
-      }
-      scoreHeader.append(...defaultGauges, ...customGauges, ...pluginGauges);
-
+      scoreHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
```
brendankenny: kind of hard to read the spread after a line break like this. Maybe save to a local var first?

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -255,10 +267,24 @@ class ReportRenderer {
...
       const topbarDocumentFragment = this._renderReportTopbar(report);
       reportFragment.appendChild(topbarDocumentFragment);
     }
+
+    if (scoreHeader && !this._dom.isDevTools()) {
+      const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
+
+      // The sticky header is just the score gauges, but styled to be smaller. Just
+      // clone the gauges from the score header.
```
brendankenny: not cloned anymore

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -255,10 +267,24 @@ class ReportRenderer {
...
+      stickyHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
```
brendankenny: same thing with the line break and spread as above

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +399,29 @@ class ReportUIFeatures {
     this._document.body.removeChild(a);
     setTimeout(_ => URL.revokeObjectURL(href), 500);
   }
+
+  _updateStickyHeaderOnScroll() {
```
brendankenny: does this need to be called once on startup?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +399,29 @@ class ReportUIFeatures {
...
+    // Show sticky header when the score scale begins to go underneath the topbar.
+    const topbarBottom = this.topbarEl.getBoundingClientRect().bottom;
+    const scoreScaleTop = this.scoreScaleEl.getBoundingClientRect().top;
+    const showStickyHeader = topbarBottom >= scoreScaleTop;
+    this.stickyHeaderEl.classList.toggle('lh-sticky-header--visible', showStickyHeader);
+
+    // Highlight mini gauge when section is in view.
```
brendankenny: early return if `showStickyHeader === false`?

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -255,10 +267,24 @@ class ReportRenderer {
...
 
     const reportFragment = this._dom.createFragment();
+
     if (!this._dom.isDevTools()) {
       const topbarDocumentFragment = this._renderReportTopbar(report);
       reportFragment.appendChild(topbarDocumentFragment);
     }
+
+    if (scoreHeader && !this._dom.isDevTools()) {
+      const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
```
brendankenny: we need a test for this, though jsdom (AFAIK) doesn't support scrolling, so it will have to be a limited one. Since we have a nice complete "renders score gauges in this order" test, maybe just a test that `.lh-sticky-header` has child gauges that match the ones in `.lh-scores-header`?

Or something like that

brendankenny:
will also need a PR rename.

@paulirish do you feel comfortable with the overhead of this scroll handler? Seems comparable to the one we had before

lighthouse-core/report/html/renderer/report-ui-features.js
brendankenny: > does this need to be called once on startup?

I guess not since it starts out hidden

Hoten: TIL - https://stackoverflow.com/a/34095654
> The scroll event does not fire on every load, only when refreshing a page that was scrolled, or when navigating to an anchor directly.

So fresh view (top of page), state should be hidden, and the default is just that. Refresh after scrolling a bit, or click on a link w/ an anchor to the report, and a scroll event fires.

Hoten:
Attempted a rebase - it's a lot of conflicts over many commits. Resolving them is not realistic imo.

I ran all the tests locally and they passed.

brendankenny:
> Attempted a rebase - it's a lot of conflicts over many commits. Resolving them is not realistic imo.

just merge? a rebase isn't necessary

> I ran all the tests locally and they passed.

I got burned the last two times I overrode the Travis check, so I'd rather not do that anymore

Hoten:
Ah, failed to mention that I did merge master. Still failing.

brendankenny:
Looks like it was the NO_TRACING_STARTED flakiness. Green now

brendankenny:
@paulirish had some performance feedback

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -380,6 +399,28 @@ class ReportUIFeatures {
     this._document.body.removeChild(a);
     setTimeout(_ => URL.revokeObjectURL(href), 500);
   }
+
+  _updateStickyHeaderOnScroll() {
+    // Show sticky header when the score scale begins to go underneath the topbar.
+    const topbarBottom = this.topbarEl.getBoundingClientRect().bottom;
+    const scoreScaleTop = this.scoreScaleEl.getBoundingClientRect().top;
+    const showStickyHeader = topbarBottom >= scoreScaleTop;
+    this.stickyHeaderEl.classList.toggle('lh-sticky-header--visible', showStickyHeader);
```
paulirish: we'll want to change this at the bottom.

otherwise we force a layout within this function.

image

Hoten: Interesting... The order is the same, so is the issue that code in between these two changes allows for the browser to jump in and do an extra layout? Idk how layouts work

paulirish: https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing

or http://wilsonpage.co.uk/preventing-layout-thrashing/

Hoten: got it, makes sense.

AFAICT from the performance tab w/ this code, there are no extra layouts. One change I saw was "Recalculate Styles" was within the update function, but w/ your changes it ran just after on the main thread. Would like to know if I'm reading the performance timeline wrong.

Even if there technically isn't a layout cost, moving the DOM mutation to the bottom SGTM b/c it makes it obvious there is no layout thrashing.

lighthouse-core/report/html/renderer/report-renderer.js
Hoten: > a test that .lh-sticky-header has child gauges that match the ones in .lh-scores-header?

imo that doesn't seem like a useful test, since these gauges are created with the same function.

> jsdom (AFAIK) doesn't support scrolling

https://github.com/jsdom/jsdom/issues/1422#issuecomment-471038750 this makes me think it should be possible to fake it

Hoten: ok it'd be way too much faking, basically all the `getBoundingClientRect` needs to be mocked

PR #8530, opened 2019-04-22
core: reword insecure error message

patrickhulce:
**Summary**
The error message for INSECURE_DOCUMENT_REQUEST, while correct, definitely takes the browser perspective on these issues rather than what I expect a user might think. When I hear "invalid credentials" as a user, I'm thinking there's some auth issue and it's a 403 or something. Really 90% of the time these are certificate issues and feels like we can rephrase a bit (source: made-up-stats.com).

It's not as big a deal here in `lh-error` where we give them the real message too compared to the extension, so I won't die on this hill if others disagree like I would over there :)

**Related Issues/PRs**
Came up in #8294 while seeing our copy of these messages over in the extension.


patrickhulce:
Yes! @Hoten any insight you'd like to share here? Hoping to find some middle ground between technically accurate from the range of possible errors and suggesting to the user what the problem is :)

Hoten:
much better

PR #8531, opened 2019-04-22
report(redesign): misc whitespace and color changes

Hoten:
#8185

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -830,7 +830,7 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: center;
-  padding-top: var(--section-indent) calc(var(--section-indent) / 2) calc(var(--section-indent) * 2);
+  padding-top: var(--section-indent);
 }
 .lh-scores-header__solo {
   padding: 0;
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -830,7 +830,6 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: center;
-  padding-top: var(--section-indent);
 }
 .lh-scores-header__solo {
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/report-styles.css
```diff
@@ -830,7 +830,7 @@
   display: flex;
   flex-wrap: wrap;
   justify-content: center;
-  padding: var(--section-indent) calc(var(--section-indent) / 2) calc(var(--section-indent) * 2);
+  padding-top: var(--section-indent) calc(var(--section-indent) / 2) calc(var(--section-indent) * 2);
```
paulirish: this is 3 values, but padding-top just takes one? I guess you just want the section indent and that's it?

lighthouse-core/report/html/templates.html
```diff
@@ -562,14 +562,15 @@
     .lh-gauge__label {
       font-size: var(--score-title-font-size);
       line-height: var(--score-title-line-height);
-      margin: calc(0.5 * var(--score-title-line-height));
+      margin-bottom: calc(0.5 * var(--score-title-line-height));
       text-align: center;
       color: black;
     }
 
     .lh-category .lh-gauge__label {
```
paulirish: side note: last year, I probably should have done this more bem-y as `.lh-gauge__label--big` instead of relying on descendent selector, etc.

can you add a todo for that?

lighthouse-core/report/html/report-styles.css

lighthouse-core/report/html/templates.html
brendankenny: > .lh-gauge__label--big

@Hoten did you note this anywhere? (can add to e.g.  #8185)

Hoten: Sorry, forgot to push. I made a TODO(#8185) comment and will follow up as clean up.

PR #8535, opened 2019-04-23
core(font-display): do not use invalid sourceURLs

### LINKED ISSUE #8534: Font display doesn't handle styles with relative/invalid sourceURL comments
1. Run LH on https://weston.ruter.net/

Font display audit fails

Font display audit doesn't fail

I've traced this down to to the inline style which comes through with an invalid/relative `source.header.sourceURL`, this throws because it's not a valid URL and we don't end up whitelisting the font URL.

**Related issues**
https://github.com/GoogleChrome/lighthouse/issues/7708#issuecomment-485598419

patrickhulce:
**Summary**
Apparently `/*# sourceURL=custom-url.css */` sets the `sourceURL` in DevTools to whatever the attribute was, so we should handle that case.

**Related Issues/PRs**
fixes #8534

### DIFF
--- a/lighthouse-core/test/audits/font-display-test.js
+++ b/lighthouse-core/test/audits/font-display-test.js
@@ -231,6 +231,8 @@ describe('Performance: Font Display audit', () => {
   });
 
   it('handles varied font-display declarations', async () => {
+    // Make sure we don't use sourceURL when it's not a valid URL, see https://github.com/GoogleChrome/lighthouse/issues/8534
+    stylesheet.header.sourceURL = 'custom-attribute';
     stylesheet.content = `
       @font-face {
         src: url(font-0.woff);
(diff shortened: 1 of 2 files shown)


westonruter:
> Apparently `<style custom-name>` sets the `sourceURL` in DevTools to whatever the attribute was, so we should handle that case.

Actually, the custom URL was set manually. See https://github.com/GoogleChrome/lighthouse/issues/8534#issuecomment-485617493

patrickhulce:
Thanks @westonruter! Updated accordingly :)

lighthouse-core/test/audits/font-display-test.js
```diff
@@ -231,6 +231,8 @@ describe('Performance: Font Display audit', () => {
   });
 
   it('handles varied font-display declarations', async () => {
+    // Make sure we don't use sourceURL when it's not a valid URL, see https://github.com/GoogleChrome/lighthouse/issues/8534
+    stylesheet.header.sourceURL = 'custom-attribute';
```
brendankenny: any chance you could split this off into its own test? I could see us revising `font-display` at some point and having to reconfigure the tests and losing this very important line :)

patrickhulce: sure

PR #8536, opened 2019-04-23
core(stackpacks): proto uses icon_data_URL and stackPack test

exterkamp:
**Summary**
Main bugfix: icon_data_url -> icon_data_URL. Also added stack packs to artifacts so that sample_v2 has a `stackPacks` element so this kind of thing is moderately prevented.

fixes: #8537


lighthouse-core/test/results/artifacts/artifacts.json
```diff
@@ -1757,6 +1757,12 @@
       "name": "jQuery",
       "version": "2.1.1",
       "npm": "jquery"
+    },
+    {
+      "detector": "js",
```
brendankenny: these artifacts come from running LH against `dbw_tester.html`, and while we don't update this file often, it is pretty brittle to have to manually add this in here.

I'm not sure what's better/worse...adding it in the generation script or [making it look like `dbw_tester.html` is served by wordpress](https://github.com/johnmichel/Library-Detector-for-Chrome/blob/68a527631db0a256932b001e17e8e173e591b55a/library/libraries.js#L1369-L1370).

exterkamp: Yeah either way its not great 😦 we could make an example stack pack that only dbw_tester passes on?  or make a flag to manually run stack packs on the cli?

brendankenny: not sure what others think, but I think my vote would be to make it look like `dbw_tester.html` was a wordpress page.

Seems like matching `script[src*="wp-includes"]` would be the easiest way to pass, just making sure to pick a MIME type that won't do anything (like [`'example'`](https://tools.ietf.org/html/rfc4735)). So could add e.g.

```html
<!-- Add non-functional script to appear to be a Wordpress page for Stack Packs. -->
<script type="example" src="wp-includes"></script>
```

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -152,7 +152,25 @@ describe('util helpers', () => {
 
       // Original audit results should be restored.
       const preparedResult = Util.prepareReportResult(clonedSampleResult);
+
+      // Delete stack packs from preparedResult
```
patrickhulce:
```suggestion
      // Delete stack packs from preparedResult; we test it separately below.
```

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -53,6 +53,9 @@
 <!-- FAIL(errors-in-console): exception thrown -->
 <script type="text/javascript">throw new Error('An error');</script>
 
+<!-- Stack Pack Wordpress detection script -->
+<script src="fake-script-wp-includes"></script>
```
patrickhulce: do we need `type="just-for-library-detector"` in here?

brendankenny: I suggested `type="example"` since it's specced to always be an example MIME type, but I guess it's unlikely there will ever be a `just-for-library-detector` type we'll collide with :)

exterkamp: Took brendan's type="example" for this :smile:

lighthouse-core/test/report/html/renderer/util-test.js
```diff
@@ -152,7 +152,25 @@ describe('util helpers', () => {
 
       // Original audit results should be restored.
       const preparedResult = Util.prepareReportResult(clonedSampleResult);
+
+      // Delete stack packs from preparedResult; we test it separately below.
+      Object.values(preparedResult.audits).forEach(audit => {
+        if (audit.stackPacks) {
+          delete audit.stackPacks;
+        }
+      });
```
brendankenny: I kind of hate we have to do this. See my proposal in #8633 for avoiding it altogether

exterkamp: Removed now that that landed!

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -53,6 +53,9 @@
 <!-- FAIL(errors-in-console): exception thrown -->
 <script type="text/javascript">throw new Error('An error');</script>
 
+<!-- Add non-functional script to appear to be a Wordpress page for Stack Packs. -->
+<script type="example" src="fake-script-wp-includes"></script>
```
brendankenny: since we're adding this here, mind adding it to the smoke test?

Should just need to add something like
```js
artifacts: {
  Stacks: [{
    id: 'jquery',
  }, {
    id: 'wordpress',
  }],
},
```
in parallel to the `lhr` entry in `dbw-expectations.js`

PR #8563, opened 2019-04-24
report(redesign): two rows for filmstrip on mobile

Hoten:
#8185

image

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -604,7 +604,6 @@
 /* Filmstrip */
 
 .lh-filmstrip-container {
-  padding: 0 var(--expandable-indent);
   /* smaller gap between metrics and filmstrip */
   margin: -8px auto 0 auto;
 }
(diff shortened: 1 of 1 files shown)


Hoten:
Should I consume more of that whitespace? (remove most of the space between)

patrickhulce:
@Hoten 

>  // screenshots in content shell are flaky and NO_NAVSTART occurs on bots way more frequently than local
  // ignore the results of the trace-dependent audits, just make sure they ran

https://cs.chromium.org/chromium/src/third_party/blink/web_tests/http/tests/devtools/audits2/audits2-successful-run.js?l=6-7

paulirish:
Can you describe what this is fixing?

This is the before/after i'm seeing:
image

Hoten:
@yuinchien how should the filmstrip images be laid out in mobile?

yuinchien:
@Hoten on mobile if we want to keep all 10 images, can we do 5 per row like in the image below?

Screen Shot 2019-04-25 at 10 58 24 AM

Hoten:
done
image

Hoten:
> Just throwing this out there as a potential option, maybe we could show 1 row of just however many there are space for? All CSS impl might be a bit messy, but it might feel more natural to keep the spacing even and reveal more images as room allows

So if there is room for 8 images, the last two won't be displayed?

patrickhulce:
> So if there is room for 8 images, the last two won't be displayed?

Well I don't think we would want to just hide the last ones, more like the intermittent ones. i.e. if there's space for 8 we hide frame 3 and 7 or something

Hoten:
I don't like the idea of eliding random frames, or the weird CSS I'm sure would come with it. @yuinchien / @hwikyounglee any ideas to deal with the spacing issues here?

yuinchien:
How about for mobile & desktop, we scale the image and keep the gutter size constant? 
Mobile will be 5 images per row, gutter size 10px, 2 rows total.
Desktop 10 images per row, gutter size 20px, 1 row total.

Hoten:
:+1: to landing

brendankenny:
we had LGTMs with "a possible nit" and "looks a little weird when they get too far apart", so let's call them LGTMs with the desire for tweaks in the future if folks want to drive them :)

PR #8583, opened 2019-04-24
misc: handle connection errors even before the timeout

patrickhulce:
**Summary**
Should fix travis and give us better error messages when Chrome crashes unexpectedly.

Previously we only handled connection errors after we reached our timeout when we should be handling them as soon as we try to connect in case Chrome crashes unexpectedly.

**Related Issues/PRs**
#7534

### DIFF
--- a/lighthouse-core/gather/connections/cri.js
+++ b/lighthouse-core/gather/connections/cri.js
@@ -108,22 +108,17 @@ class CriConnection extends Connection {
         });
       });
 
-      let ignoreECONNRESET = false;
-      request.on('error', err => {
-        // @ts-ignore - not in @types yet. See https://nodejs.org/api/errors.html#errors_class_systemerror.
-        if (err.code === 'ECONNRESET' && ignoreECONNRESET) return;
-        reject(err);
-      });
+      // This error handler is critical to ensuring Lighthouse exits cleanly even when Chrome crashes.
+      // See https://github.com/GoogleChrome/lighthouse/pull/8583.
+      request.on('error', reject);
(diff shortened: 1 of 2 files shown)


patrickhulce:
So the errors smoke test has been fixed :)

but the NO_TRACING_STARTED is still very in our face :(

lighthouse-core/gather/connections/cri.js
```diff
@@ -108,16 +108,17 @@ class CriConnection extends Connection {
         });
       });
 
-      request.setTimeout(CONNECT_TIMEOUT, () => {
-        request.abort();
+      let ignoreECONNRESET = false;
```
brendankenny: can you add a comment explaining this

patrickhulce: as in move up the comment that's where we set it or a new one duplicating? :)

patrickhulce:
> so what changed with 74 exactly?

Chrome's shutdown procedure is clearly different though I'm not sure what exactly. If I had to guess I would say it's related to network service since that's dramatically changed how Chrome shuts down (usually takes 4-6s now locally).

lighthouse-core/gather/connections/cri.js
brendankenny: duplicate-ish?

patrickhulce: refactored so it's not necessary instead

brendankenny: love it

PR #8590, opened 2019-04-24
report(redesign): adjust score scale to match new design

Hoten:
#8185

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -118,6 +118,8 @@
   --score-title-font-size: 20px;
   --score-title-line-height-big: 36px;
   --score-title-line-height: 26px;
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/report-styles.css
```diff
@@ -853,18 +853,20 @@
...
   margin: 0 12px;
   font-family: var(--monospace-font-family);
   white-space: nowrap;
 }
 
 .lh-scorescale-range::before {
   content: '';
-  width: var(--body-font-size);
-  height: calc(var(--body-font-size) * .50);
+  width: 18px;
```
patrickhulce: should we var these babies up somewhere?

Hoten:
lgtm b/c it's pretty isolated (and done :))

I'll use #8591 for the rest of the pixel pushing

PR #8591, opened 2019-04-24
report(redesign): various whitespace adjustments (DRAFT)

Hoten:
#8185

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -688,11 +688,6 @@
   background-image: var(--pwa-optimized-color-url);
 }
 
-/* Removing too much whitespace */
-.lh-audit-group--metrics {
-  margin-top: calc(var(--gauge-circle-size)/2 * -1);
-  border-bottom: none;
-}
 .lh-audit-group--metrics .lh-audit-group__summary {
   margin-top: 0;
   margin-bottom: 0;
@@ -924,6 +919,7 @@
   max-width: 400px;
   width: auto;
   margin: 0px auto;
+  padding: var(--score-container-padding);
 }
 
 .lh-category-header .lh-audit__title {
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -648,11 +648,12 @@
   border-bottom: 1px solid var(--report-secondary-border-color);
 }
 
-/*
-* Apply border-top to just the first audit. First child is the header.
-*/
-.lh-audit:nth-child(2) {
+/* Apply border-top to just the first audit. */
+.lh-audit {
   border-top: 1px solid var(--report-secondary-border-color);
+}
+.lh-audit ~ .lh-audit {
+  border-top: none;
 }
 
 
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -829,6 +829,10 @@
   border: 0;
 }
 
+.lh-gauge__wrapper {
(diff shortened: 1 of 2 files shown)


patrickhulce:
we don't get anymore top margin now though between the category header should we bump that a tad?

image

Hoten:
Ok, I tried to adjust that @patrickhulce, and some other things that didn't match re: section spacing with the report, and it looks closer now ?

Super hard to split up changes that affect spacing (padding, margin of various components, font sizes, etc). I want to pull up the design and the impl side by side, and make pixel-perfect adjustments. but if the font size goes in one PR, and spacing for gauges in one, and margins for sections in another ... well, my brain goes kaploot.

patrickhulce:
> I want to pull up the design and the impl side by side, and make pixel-perfect adjustments.

Why don't we just do that then? :) That's always the strategy that makes more sense to me personally, sit down with a designer and make all the adjustments at once.

Hoten:
> Why don't we just do that then? :)

wunderbar

lighthouse-core/report/html/report-styles.css
```diff
@@ -646,14 +648,25 @@
   border-bottom: 1px solid var(--report-secondary-border-color);
 }
 
+/*
+* Apply border-top to just the first audit.
+*/
+.lh-audit {
+  border-top: 1px solid var(--report-secondary-border-color);
+}
+.lh-audit ~ .lh-audit {
```
paulirish: this is a somewhat expensive selector. can we just do border-top on the `.lh-audit:nth-child(1)` ?

lighthouse-core/report/html/templates.html
```diff
@@ -500,6 +500,9 @@
 
     .lh-gauge__svg-wrapper {
       position: relative;
+      /* The wrapper adds some height to the element. display: inline-block doesn't
+          work for some reason, so instead just shave off the extra height. */
+      margin-bottom: -7px;
```
paulirish: hmmm i'm only seeing 5 extra pixels, not 7. 

also i'd much rather determine where the phantom height is coming from and address that rather than go with negative margin hacks.

lighthouse-core/report/html/report-styles.css
Hoten: done. had to be 2.

paulirish: ah. i see. i was looking at best practices where the DOM is a little different.

image

hmm. if we can still keep to something simple like nth-child and support both these cases, i'd like to. but if its a PITA then perhaps we have to go back to the tilde.

lighthouse-core/report/html/templates.html
Hoten: image
image

I tried setting the svg wrapper to inline-block (which should not consume those extra 7 pixels):

image

But it gets ignored.

...After typing this, I realized I needed to set the height: `height: var(--gauge-circle-size)`.

lighthouse-core/report/html/report-styles.css
Hoten: I'd prefer the tilde. this dom changes too much, and this would easily get out of date

lighthouse-core/report/html/report-styles.css
```diff
@@ -646,14 +648,22 @@
   border-bottom: 1px solid var(--report-secondary-border-color);
 }
 
+/*
+* Apply border-top to just the first audit. First child is the header.
+*/
+.lh-audit:nth-child(2) {
```
paulirish: we can go back to the tilde since this doesn't work as consistently.

PR #8592, opened 2019-04-24
report(redesign): inline display text. bold audit header text

Hoten:
#8185

image


lighthouse-core/report/html/report-styles.css
```diff
@@ -387,14 +386,20 @@
...
 }
 
-.lh-audit__title {
+.lh-audit__display-text {
   flex: 1;
 }
 
+/* Prepend display text with separator. But not in Opportunities. */
+.lh-audit-group:not(.lh-audit-group--load-opportunities) .lh-audit__display-text:not(:empty):before {
+  content: '—';
```
paulirish:
```suggestion
  content: '\2014';
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -387,14 +386,20 @@
   margin-right: 0;
 }
 
-.lh-audit__title {
+.lh-audit__display-text {
   flex: 1;
 }
 
+/* Prepend display text with separator. But not in Opportunities. */
+.lh-audit-group:not(.lh-audit-group--load-opportunities) .lh-audit__display-text:not(:empty):before {
```
paulirish: while this selector does look crazy, it turns out to be reasonable..  👍

lighthouse-core/report/html/report-styles.css
```diff
@@ -387,14 +386,20 @@
   margin-right: 0;
 }
 
-.lh-audit__title {
+.lh-audit__display-text {
   flex: 1;
 }
 
+/* Prepend display text with separator. But not in Opportunities. */
```
paulirish:
```suggestion
/* Prepend display text with em dash separator. But not in Opportunities. */
```

lighthouse-core/report/html/report-styles.css
paulirish: per discussion in chat.. let's use whitespace on both sides of the dash.

Assuming we want to use `var(--audit-item-gap)` the we'll need some margin in here too:

```cs
  margin-right: var(--audit-item-gap);
```

brendankenny:
these things realllllly need a screenshot. No idea what "inline display text" means here :)

lighthouse-core/report/html/report-styles.css
brendankenny: can this just go in the HTML? Kind of weird to do it with a pseudo-element

Hoten: why? dt encoding issues are resolved

Hoten: @brendankenny can't just be HTML, need some sort of logic check. makes sense to me to have the display + logic in one place here.

lighthouse-core/report/html/report-styles.css
Hoten: :)

lighthouse-core/report/html/report-styles.css

lighthouse-core/report/html/report-styles.css
```diff
@@ -453,13 +452,20 @@
   margin-right: 0;
 }
 
-.lh-audit__title {
+.lh-audit__display-text {
   flex: 1;
 }
 
+/* Prepend display text with em dash separator. But not in Opportunities. */
+.lh-audit-group:not(.lh-audit-group--load-opportunities) .lh-audit__display-text:not(:empty):before {
```
brendankenny: best practices aren't grouped so aren't getting the `:before`

Hoten: fixed

lighthouse-core/report/html/report-styles.css
brendankenny: > can't just be HTML, need some sort of logic check. makes sense to me to have the display + logic in one place here.

I don't think it needs to be. Opportunities and the rest of the audits share text with a `.lh-audit__display-text` class, but they actually come from different templates, `#tmpl-lh-opportunity` vs `#tmpl-lh-audit`. If the `#tmpl-lh-audit` template gets the em dash inline, then it'll only show up there

Hoten: Good catch on the template thing. But would still need a length check - and can't make a CSS rule for preceding sibling (pseudo code below).

image

```
.lh-audit__separator {
  margin: 0 var(--audit-item-gap);
}
/* not actually a thing */
.lh-audit__display-text:not(:empty) - .lh-audit__separator {
  display: none;
}
```

Issue #8598, opened 2019-04-25
ensure 4.x LHR and 5.0 report renderer compatibility


brendankenny:
both for ease of viewer maintenance and the PSI update dance, we need to make sure 5.0 renders 4.x (and preferably 3.x) LHRs. Should be pretty simple, but there's been at least one thing that's slipped through with 5.0.

Please add to the list as we discover more:
- [ ] `diagnostics` audit details were renamed to `debugdata` (#8298), so `details-renderer.js` will throw when trying to render `diagnostics` now. Could correct in util or have details renderer handle this case [like it does no details `type`](https://github.com/GoogleChrome/lighthouse/blob/90ae4851630eac42b5e901649c3ce40fbf3e48b4/lighthouse-core/report/html/renderer/details-renderer.js#L64-L66)
- [ ] screenshots in the `filmstrip` audit details type are now assumed to include the data URL jpg prefix (#8299), so the screenshots from old LHRs without it won't render

PR #8605, opened 2019-04-25
core: fix sample_v2 generation, proto errors

### LINKED ISSUE #8245: lhr.stackPacks doesn't make the proto round trip
you'll see this if you run `yarn update:sample-json` on master. `stackPacks` will be deleted and then `npx jest proto-test.js` will fail.

Since CI only runs `yarn diff:sample-json` (just regenerating the original json, not the proto round-trip json), as long as `stackPacks` was manually added to `sample_v2_round_trip.json`, everything will be green on Travis :)

@exterkamp were the proto changes in #7243 not sufficient to get this automatically in the round trip result? I haven't been able to get good insight into why `json_roundtrip_via_proto.py` is dropping it.

brendankenny:
recently we (I) have asked a few PRs to update a subset of example artifacts and then `sample_v2.json`. However if you try to run `yarn update:sample-artifacts` and then `yarn update:sample-json`, you'll notice there are a few errors preventing you from actually doing that :)

This PR
- fixes the `throttling-method` mismatch between the `-G` and the `-A` by switching `-G` to `devtools` to match `-A`
- adds the optional `gather` boolean to the proto `Timing` entry interface so that gatherer entries, when reimported with `-A`, will still result in a proto-able LHR at the end.

  I'm not sure how often that will come up in practice for PSI, but it's possible, and with `false`/`undefined` as the default, it's zero overhead when there are no `gather` props (as there aren't in the current `sample_v2.json` because the source artifacts have `Timing: []`)
- makes `lhr.stackPacks` optional, matching the proto's treatment of them as a `repeated` field. This is probably the easiest resolution to #8245 (and solves #8598 for stack packs automatically :), and now folks won't have to keep manually adding that back.

cc @exterkamp 

fixes #8245

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "type-check": "tsc -p . && tsc -p lighthouse-viewer/",
     "i18n:checks": "./lighthouse-core/scripts/i18n/assert-strings-collected.sh",
     "i18n:collect-strings": "node lighthouse-core/scripts/i18n/collect-strings.js",
-    "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
+    "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js",
     "update:sample-json": "yarn i18n:collect-strings && node ./lighthouse-cli -A=./lighthouse-core/test/results/artifacts --throttling-method=devtools --output=json --output-path=./lighthouse-core/test/results/sample_v2.json && node lighthouse-core/scripts/cleanup-LHR-for-diff.js ./lighthouse-core/test/results/sample_v2.json --only-remove-timing && yarn compile-proto && yarn build-proto-roundtrip",
     "diff:sample-json": "yarn i18n:checks && bash lighthouse-core/scripts/assert-golden-lhr-unchanged.sh",
     "ultradumbBenchmark": "./lighthouse-core/scripts/benchmark.js",
--- a/proto/lighthouse-result.proto
+++ b/proto/lighthouse-result.proto
@@ -161,6 +161,9 @@ message LighthouseResult {
       string entry_type = 2;
       google.protobuf.DoubleValue start_time = 3;
       google.protobuf.DoubleValue duration = 4;
+
+      // Whether timing entry was collected during artifact gathering.
+      bool gather = 5;
     }
 
     // The total duration of Lighthouse's run
--- a/proto/sample_v2_round_trip.json
+++ b/proto/sample_v2_round_trip.json
@@ -3653,11 +3653,10 @@
             "warningAuditsGroupTitle": "Passed audits but with warnings", 
             "warningHeader": "Warnings: "
         }
-    },
+    }, 
     "lighthouseVersion": "4.3.0", 
     "requestedUrl": "http://localhost:10200/dobetterweb/dbw_tester.html", 
-    "runWarnings": [],
-    "stackPacks": [],
+    "runWarnings": [], 
     "timing": {
         "entries": [
             {
--- a/types/lhr.d.ts
+++ b/types/lhr.d.ts
@@ -61,7 +61,7 @@ declare global {
       /** The record of all formatted string locations in the LHR and their corresponding source values. */
       i18n: {rendererFormattedStrings: I18NRendererStrings, icuMessagePaths: I18NMessages};
       /** An array containing the result of all stack packs. */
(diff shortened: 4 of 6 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "type-check": "tsc -p . && tsc -p lighthouse-viewer/",
     "i18n:checks": "./lighthouse-core/scripts/i18n/assert-strings-collected.sh",
     "i18n:collect-strings": "node lighthouse-core/scripts/i18n/collect-strings.js",
-    "update:sample-artifacts": "node lighthouse-core/scripts/update-report-fixtures.js -G",
```
brendankenny: doesn't do anything :)

brendankenny:
>  FAIL  lighthouse-core/test/report/proto-test.js
>  ● round trip JSON comparison subsets › has the same top level values
>     -   "stackPacks": Array [],

ok, not being able to even have an empty array might be a bridge too far. Have to think about it :)

proto/lighthouse-result.proto
```diff
@@ -161,6 +161,9 @@ message LighthouseResult {
       string entry_type = 2;
       google.protobuf.DoubleValue start_time = 3;
       google.protobuf.DoubleValue duration = 4;
+
+      // Whether timing entry was collected during artifact gathering.
+      bool gather = 5;
```
patrickhulce: didn't @exterkamp *just* remove this ? maybe I'm just misremembering what I saw haha

types/lhr.d.ts
```diff
@@ -61,7 +61,7 @@ declare global {
       /** The record of all formatted string locations in the LHR and their corresponding source values. */
       i18n: {rendererFormattedStrings: I18NRendererStrings, icuMessagePaths: I18NMessages};
       /** An array containing the result of all stack packs. */
-      stackPacks: Result.StackPack[];
+      stackPacks?: Result.StackPack[];
```
patrickhulce: wait this had no other impact being required the whole time? 😆

proto/lighthouse-result.proto
brendankenny: Ha, unless he had amnesia I don't think so... @exterkamp?

types/lhr.d.ts
brendankenny: > wait this had no other impact being required the whole time? 😆

Yeah, the only code using [was treating it as optional](https://github.com/GoogleChrome/lighthouse/blob/ca650db26339d05a63910bd240e269a683ee850e/lighthouse-core/report/html/renderer/util.js#L82) anyways, so easy change (or so I thought).

proto/lighthouse-result.proto
exterkamp: I removed it from my PR so it could be here where it made more sense 😄

proto/sample_v2_round_trip.json
```diff
@@ -3653,11 +3653,10 @@
             "warningAuditsGroupTitle": "Passed audits but with warnings", 
             "warningHeader": "Warnings: "
         }
-    },
+    }, 
     "lighthouseVersion": "4.3.0", 
     "requestedUrl": "http://localhost:10200/dobetterweb/dbw_tester.html", 
-    "runWarnings": [],
-    "stackPacks": [],
```
exterkamp: This makes the proto round trip test fail b/c empty list is elided.  [] == nothing in proto as repeated.  This comes back to the whole ListValue thing.  See appveyor.

brendankenny: Ha, yes I know. See my comment https://github.com/GoogleChrome/lighthouse/pull/8605#issuecomment-486534646 :)

brendankenny:
OK, 2a22744 is the solution I'll advocate for.

Basically the JS side is helping out the proto side by allowing `lhr.stackPacks` to be optional, but the proto is going to have to meet the JS halfway by acknowledging that to it `[]` and not defined are equivalent. The end result is that both sides agree that `stackPacks: []` and no `stackPacks` at all should be treated as the same thing, so we'll never have a mismatch.

I've enabled that by modifying the proto test expectation so that if `stackPacks` is an empty array, it can pass if the round-trip proto has an omitted `stackPacks`. This has the nice benefit of no additions to `proto-preprocessor`, which we're still hoping to eliminate over time. And to be conservative, it's doing this only for `stackPacks`, not all default values in our proto (though maybe we eventually should...).

PR #8625, opened 2019-04-25
core(gather-runner): always reset scroll position

patrickhulce:
**Summary**
Came up a few times now, but we should give gatherers a consistent view of the page. This PR saves the scroll position of the page after we're done loading and scrolls back to it after every gatherer.

**Related Issues/PRs**
https://github.com/GoogleChrome/lighthouse/pull/7778#discussion_r276864149


lighthouse-core/gather/driver.js
```diff
@@ -86,6 +86,14 @@ class Driver {
       this._handleReceivedMessageFromTarget(event, []).catch(this._handleEventError);
     });
 
+    this.on('Runtime.executionContextDestroyed', event => {
+      if (event.executionContextId === this._isolatedExecutionContextId) {
+        this._clearIsolatedContextId();
+      }
+    });
+
+    this.on('Page.frameNavigated', this._clearIsolatedContextId.bind(this));
```
brendankenny:
```suggestion
    this.on('Page.frameNavigated', () => this._clearIsolatedContextId);
```
?

lighthouse-core/gather/driver.js
```diff
@@ -86,6 +86,14 @@ class Driver {
       this._handleReceivedMessageFromTarget(event, []).catch(this._handleEventError);
     });
 
+    this.on('Runtime.executionContextDestroyed', event => {
```
brendankenny: maybe add a comment above these

lighthouse-core/gather/driver.js
```diff
@@ -464,7 +472,16 @@ class Driver {
    */
   async evaluateAsync(expression, options = {}) {
     const contextId = options.useIsolation ? await this._getOrCreateIsolatedContextId() : undefined;
-    return this._evaluateInContext(expression, contextId);
+    return this._evaluateInContext(expression, contextId).catch(async err => {
```
brendankenny: can you switch to try/await/catch? `.catch(async err => {})` is confusing :)

lighthouse-core/gather/gather-runner.js
```diff
@@ -319,6 +319,10 @@ class GatherRunner {
     await driver.setThrottling(passContext.settings, {useThrottling: false});
     log.time(apStatus, 'verbose');
 
+    // Some gatherers scroll the page which can cause unexpected results for other gatherers.
+    // We reset the scroll position in between each gatherer.
+    const scrollPosition = pageLoadError ? null : await driver.getScrollPosition();
```
brendankenny: `pageLoadError` control flow is seriously the worst :S

lighthouse-core/test/gather/driver-test.js
```diff
@@ -317,6 +317,19 @@ describe('.evaluateAsync', () => {
...
+  it('recovers from isolation failures', async () => {
+    connectionStub.sendCommand = createMockSendCommandFn()
+      .mockResponse('Page.getResourceTree', {frameTree: {frame: {id: 1337}}})
+      .mockResponse('Page.createIsolatedWorld', {executionContextId: 9001})
+      .mockResponse('Runtime.evaluate', Promise.reject(new Error('Cannot find context')))
+      .mockResponse('Page.getResourceTree', {frameTree: {frame: {id: 1337}}})
+      .mockResponse('Page.createIsolatedWorld', {executionContextId: 9002})
+      .mockResponse('Runtime.evaluate', {result: {value: 2}});
+
+    const value = await driver.evaluateAsync('1 + 1', {useIsolation: true});
```
brendankenny: maybe pick a evaluateAsync and response that makes it obvious that real computation isn't being done here?

lighthouse-core/gather/driver.js
patrickhulce: oops that's not right though haha, but point taken :)

lighthouse-core/gather/driver.js
patrickhulce: *can* I? yes, but let me try to convince you why we shouldn't

I personally find it's semantics very subtly deceptive for this situation.

You start off with 

```js
try {
  return this._evaluateInContext(..)
} catch (err) {
  // do stuff
}
```

then you realize that this won't ever catch anything so you change it to

```js
try {
  return await this._evaluateInContext(..)
} catch (err) {
  // do stuff
}
```

but there are linters out there that prevent the `return await <value>` pattern so it's been engrained in folks to think it's unnecessary (even though it is in this case and good linters like eslint won't flag it).

then you might think OK fine I'll assign it to a value first!
```js
try {
  const value = await this._evaluateInContext(..)
  return value
} catch (err) {
  // do stuff
}
```

that sidesteps the `return await` problem but still looks unnecessary to the casual observer

all of this to avoid a `.catch` which, even if it mixes `.then` and `async` styles is very straightforward to understand what's happening :)

brendankenny: Haha, I like this response.

I mean, I don't really care what bad linters say, but setting that aside... :)

My issue is that an async function passed as a callback always raises my alert level, making me double check that the callee can handle an async function. I'd definitely get used to it over time for `.then`/`.catch`, but I can't really foresee this pattern taking the world by storm at any point in the future :)

My vote is to not mix and match, so either
- go all in on async/await here, make readers get used to `await` patterns, even returns
- go all in on promises here, converting the `.catch()` callback to not be `async`

patrickhulce: > I don't really care what bad linters say

To clarify, good linters still flag `return await <some value>` which gets folks in the habit of thinking it's bad. I consider this to be a more serious version of the semicolon problem. 

At least linters can warn when you needed to insert a semicolon, making them unnecessary IMO, but no one can tell if you meant to catch all the errors or just the sync ones!

I will give up here though and go with option 1

paulirish:
let's land this after we ship 5.0.0

patrickhulce:
> let's land this after we ship 5.0.0

It is not post-5.0.0, reviews welcome :)

brendankenny:
yeah, no, bad merge, bad :)

PR #8629, opened 2019-04-25
report(redesign): three dots for export button

Hoten:
#8185

@yuinchien the dots in your design use the material fonts icon font. I believe adding a font for just that is not ideal (@paulirish can confirm). I was hoping the unicode character ⋮ would suffice, but the spacing between dots is off.


yuinchien:
@Hoten i understand your concern, but it's important to have a consistent icon style throughout the report. The source file (svg and png) can be access here: https://icons.googleplex.com/#icon=ic_more_vert&search=more 

Can we do inline svg instead of loading the whole font file?

lighthouse-core/report/html/templates.html
```diff
@@ -224,21 +224,15 @@
       will-change: transform;
     }
     .lh-export__button {
-      background-color: #fff;
-      border: 1px solid #dadada;
-      border-radius: 2px;
+      opacity: 0.54;
```
exterkamp: Matches with #757575 comment.
```suggestion

```

lighthouse-core/report/html/templates.html
```diff
@@ -310,7 +304,12 @@
     <a href="" class="lh-topbar__url" target="_blank" rel="noopener"></a>
     
     <div class="lh-export">
-      <button class="report-icon report-icon--share lh-export__button" title="Export report"></button>
+      <div class="report-icon report-icon--share lh-export__button" title="Export report">
+        <svg width="100%" height="100%" viewBox="0 0 24 24" fill="#000000">
```
exterkamp: How about using grey-600 instead of #000000 and opacity 0.54?
```suggestion
        <svg width="100%" height="100%" viewBox="0 0 24 24" fill="#757575">
```

patrickhulce: or even better, can we set the fill with our CSS variable :)

brendankenny:
proposal: what if we kept the share button on desktop and only used the three dots on a mobile breakpoint?

The share icon communicates more about what will contained in the drop down and it looks less like we took a mobile UI and blew it up to desktop size (and in devtools we won't suffer from a double three-dot menu icon)

Hoten:
> proposal: what if we kept the share button on desktop and only used the three dots on a mobile breakpoint?
> 
> The share icon communicates more about what will contained in the drop down and it looks less like we took a mobile UI and blew it up to desktop size (and in devtools we won't suffer from a double three-dot menu icon)

@yuinchien

yuinchien:
@Hoten thanks for the suggestion. I realized the double menu icon situation after reading it, they are not aligned vertically or having similar sizes that you'd confuse the chrome system icon versus the menu icon in the report. It's also better keep the icon consistent than switching it for diff form factors.

I did start with the share icon in my prototype, but visually they are huge and don't look pretty. Our design update strives for simplicity, I think the menu icon serves its purpose, people are used to access it for more actions. Can we keep the current design and track user's behaviour? If click rate drop significantly we can revisit the design :)

paulirish:
i'm not following the thread here too well... where are we on this?

Hoten:
@brendankenny suggested keeping the share icon for desktop. @yuinchien gave reason not to.

PR #8630, opened 2019-04-25
misc: simplify LH.Config settings types

brendankenny:
We missed this simplification while doing the flags/settings unification stuff.

We needed `extraHeaders` to be a more specific type for configJson settings at some point, but it's now the same type as on the base `SharedFlagsSettings`, so `LH.Config.Json['settings']` can just be `SharedFlagsSettings`, as you'd pretty much expect.

### DIFF
--- a/types/config.d.ts
+++ b/types/config.d.ts
@@ -15,16 +15,12 @@ declare global {
        */
       export interface Json {
         extends?: 'lighthouse:default' | 'lighthouse:full' | string | boolean;
-        settings?: SettingsJson;
+        settings?: SharedFlagsSettings;
         passes?: PassJson[] | null;
         audits?: Config.AuditJson[] | null;
         categories?: Record<string, CategoryJson> | null;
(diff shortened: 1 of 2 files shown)


brendankenny:
this doesn't have any material effect, just noticed when looking at budget types that we had an intermediate type we don't need anymore.

types/config.d.ts
```diff
@@ -15,18 +15,14 @@ declare global {
        */
       export interface Json {
         extends?: 'lighthouse:default' | 'lighthouse:full' | string | boolean;
-        settings?: SettingsJson;
+        settings?: SharedFlagsSettings;
```
patrickhulce: seems like we could just call this settings and CliFlags extends settings then :)

brendankenny: > seems like we could just call this settings and CliFlags extends settings then :)

ha, we could. For whatever reason the name `SharedFlagsSettings` does a poor job to indicate that it's literally "the properties shared by flags and settings" to me :)

PR #8632, opened 2019-04-26
report: create svg data urls from file

Hoten:
This extra step in our report asset pipeline makes editing svgs 100x easier.

I did just one to show proof of concept. Will do the rest if this is blessed.


Hoten:
image

Hoten:
> I dislike that the variables are now somewhat confusing to trace to their source

how about a comment above each one?

```css
/* ./lighthouse-core/report/html/pwa-fast-reliable-gray-url.svg */
--pwa-fast-reliable-gray-url: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg">  <g fill="none" fill-rule="nonzero">    <circle fill="%23DAE0E3" cx="12" cy="12" r="12" />    <path      d="M12.3 4l6.3 2.8V11c0 3.88-2.69 7.52-6.3 8.4C8.69 18.52 6 14.89 6 11V6.8L12.3 4zm-.56 12.88l3.3-5.79.04-.08c.05-.1.01-.29-.26-.29h-1.96l.56-3.92h-.56L9.6 12.52c0 .03.07-.12-.03.07-.11.2-.12.37.2.37h1.97l-.56 3.92h.56z"      fill="%23FFF" />  </g></svg>');
```

patrickhulce:
> how about a comment above each one?

I like that! But my main concern was I'm looking at the CSS file where this variable is used, now I want to see where it comes from. I ctrl-f, find nothing; I grep the entire codebase, nothing.

The comment at least makes it so I can grep in the final HTML source which is nice :)

brendankenny:
> we won't end up touching any of them until next year

yeah, that would be my concern as well. Trading build system complexity to gain development ease...but we change these so rarely it doesn't seem like we'll be taking advantage of it.

I wonder if this would be useful instead as a standalone script I could drop an svg into and get back a data url I could paste somewhere.

brendankenny:
closing for now, let's pick back up when we're figuring out report renderer cleanup and #8884

PR #8637, opened 2019-04-26
clients(devtools): remove ascii encoded requirement

### LINKED ISSUE #8525: Investigate devtools ascii requirement
Devtools currently does not support non-ascii assets. Figure out if that restriction can be dropped, so we can stop converting the LH import to unicode-escaped ascii.

Extracted from #8456.

Hoten:
awaiting https://chromium-review.googlesource.com/c/chromium/src/+/1584569

Fixes #8525


Hoten:
Upstream changes have landed.

PR #8658, opened 2019-04-26
report(redesign): remove units from table columns

Hoten:
#8185


brendankenny:
before/after screenshot?

Hoten:
before
image
after
image

exterkamp:
> LGTM. We need to sync this ASAP if we want it within a week. But the not so bad news is that the other locales will just say "X (ms)" instead of just "X" so it's not crazy critical.

Oh how naive I was.  This breaks tc/ :laughing:

patrickhulce:
*breaks* tc? How!?

PR #8660, opened 2019-04-27
report(redesign): fireworks

Hoten:
#8185

run this in the now deployment:
```
const scoresContainer = document.querySelector('.lh-scores-container');
scoresContainer.classList.add('score100');
document.body.classList.add('dark');
scoresContainer.addEventListener('click', _ => {
  scoresContainer.classList.toggle('fireworks-paused');
});
```


lighthouse-core/report/html/templates.html
```diff
@@ -362,6 +367,7 @@
     */
     .pyro {
       display: none;
+      z-index: 1;
```
paulirish: we need `pointer-events: none` now. it's hard but possible to click on the fireworks div blocking the click on the gauge

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -244,16 +244,16 @@ class ReportRenderer {
...
-    //     const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
-    //     scoresContainer.classList.add('score100');
-    //     scoresContainer.addEventListener('click', _ => {
-    //       scoresContainer.classList.toggle('fireworks-paused');
-    //     });
-    //   }
-    // }
+    const scoresAll100 = report.reportCategories.every(cat => cat.score === 1);
+    if (!this._dom.isDevTools() && scoresAll100) {
+      if (scoresAll100) {
```
brendankenny: no need to double check `scoresAll100`

Hoten:
> needs a test :)

ez

paulirish:
Can we flip on dark mode for just scoresHeader?

>

Hoten:
> Can we flip on dark mode for just scoresHeader?

i think it'd look weird dawg but @exterkamp suggests he'll try it post IO

paulirish:
let's switch the whole report into dark mode if they are getting fireworks.

PR #8661, opened 2019-04-27
report(redesign): add toggle control to show metric descriptions

Hoten:
#8185


lighthouse-core/report/html/templates.html
```diff
@@ -60,13 +60,37 @@
...
         </div>
         <div class=""></div>
       </div>
     </summary>
   </details>
 </template>
 
+<!-- Lighthouse metrics toggle -->
+<template id="tmpl-lh-metrics-toggle">
+  <div class="lh-metrics-toggle">
```
paulirish: i'd like to use a `<input type=checkbox>` here (as the underlying control).  

we shouldn't need JS for this guy (except for ` this.metricAuditGroup.classList.toggle('lh-audit-group--metrics__show-descriptions')`)

this guide seems to be solid on implementation.. also including their resources at the bottom. https://webdesign.tutsplus.com/tutorials/how-to-make-custom-accessible-checkboxes-and-radio-buttons--cms-32074

lighthouse-core/report/html/templates.html
```diff
@@ -60,13 +60,37 @@
...
+    <div class="lh-metrics-toggle-less lh-metrics-toggle--active">
+      <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">
+        <defs>
+          <path id="a" d="M0 0h24v24H0V0z" />
+        </defs>
+        <clipPath id="b">
```
paulirish: afaik this clippath isn't being used. i guess its just an artifact from sketch exports.

lighthouse-core/report/html/templates.html
```diff
@@ -60,13 +60,37 @@
...
+          <use xlink:href="#a" overflow="visible" />
+        </clipPath>
+        <path class="lh-metrics-toggle-lines" d="M4 9h16v2H4zm0 4h10v2H4z" clip-path="url(#b)" />
+      </svg>
+    </div>
+    <div class="lh-metrics-toggle-more">
+      <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+        <path class="lh-metrics-toggle-lines" d="M3 18h12v-2H3v2zM3 6v2h18V6H3zm0 7h18v-2H3v2z" />
+        <path fill="none" d="M0 0h24v24H0V0z" />
```
paulirish: ditto with this unused path

lighthouse-core/report/html/templates.html
Hoten: done, hows that?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -157,6 +166,25 @@ class ReportUIFeatures {
     this._copyAttempt = false;
   }
 
+  _setupMetricDescriptionToggleElements() {
+    this.metricDescriptionToggleEl =
+      /** @type {HTMLInputElement} */ (this._dom.find('.lh-metrics-toggle input', this._document));
+    this.metricAuditGroup = this._dom.find('.lh-audit-group--metrics', this._document);
+  }
+
+  _setupMetricDescriptionToggleListeners() {
```
paulirish: i definitely didnt expect this..

click on the newly expanded descriptions to make them go away?

this UX doesn't make sense to me

paulirish: IMO this would only make sense if the toggle is scoped to just the metric title.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -157,6 +166,25 @@ class ReportUIFeatures {
...
+    const toggle = () => this.metricDescriptionToggleEl.click();
+    this.metricDescriptionToggleEl.addEventListener('input', this._toggleMetricDescription);
+    for (const el of this._dom.findAll('.lh-metric', this._document)) {
```
paulirish: event delegation will be cleaner here.

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -81,6 +88,8 @@ class ReportUIFeatures {
     this._document.addEventListener('copy', this.onCopy);
     this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
     window.addEventListener('resize', this._updateStickyHeaderOnScroll);
+    this._setupMetricDescriptionToggleElements();
+    this._setupMetricDescriptionToggleListeners();
```
paulirish: let's just merge these two

lighthouse-core/report/html/report-styles.css
```diff
@@ -102,6 +103,7 @@
   --env-item-bg: var(--color-black-100);
   --env-name-min-width: 220px;
   --env-tem-padding: 10px 0px;
+  --expandable-padding: 0 0 2px 28px;
```
paulirish: magic number here...  :/

```
    --score-shape-margin: 7px 12px 0 4px;
    --score-shape-size: 12px;
```

afaik the 28 is 12 + 4 + 12.  

I wonder if we can redefine score-shape with some base variable so these remain synchronized?

lighthouse-core/report/html/report-styles.css
```diff
@@ -548,11 +551,66 @@
...
+}
+.lh-metrics-toggle input:not(:checked) + label .lh-metrics-toggle-less,
+.lh-metrics-toggle input:checked + label .lh-metrics-toggle-more {
+  background-color: var(--color-blue);
+}
+.lh-metrics-toggle-lines {
+  fill: #7F7F7F;
+}
+.lh-metrics-toggle input:not(:checked) + label .lh-metrics-toggle-less .lh-metrics-toggle-lines,
+.lh-metrics-toggle input:checked + label .lh-metrics-toggle-more .lh-metrics-toggle-lines {
```
paulirish: instead of this, you could override a css var up in the conditional L595 rule ... 

```css
  --color-lines: var(--color-white);
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -548,11 +551,66 @@
...
+  fill: white;
+}
+.lh-metrics-toggle label div.lh-metrics-toggle-less {
```
paulirish: you could afford to be more BEM-y here

.lh-metrics-toggle__icon
.lh-metrics-toggle__icon--expand

lighthouse-core/report/html/report-styles.css
```diff
@@ -548,11 +551,66 @@
...
+  width: 74px;
+  height: 28px;
+}
+.lh-metrics-toggle label {
+  display: flex;
+  background-color: #eee;
+  border-radius: 20px;
+  overflow: hidden;
+}
+.lh-metrics-toggle label div {
```
paulirish: this should be defined closer to the L605 rules. and BEMy would be `.lh-metrics-toggle__icon`

lighthouse-core/report/html/report-styles.css
```diff
@@ -548,11 +551,66 @@
...
+.lh-metrics-toggle {
+  position: relative;
+}
+.lh-metrics-toggle input {
+  cursor: pointer;
+  left: 0;
+  opacity: 0;
+  position: absolute;
+  top: 0;
+  width: 74px;
```
paulirish: instead of these magic numbers you could just do bottom:0; right: 0; and now its 100% 100%. ;)

also not putting left & top together in a rule? what kind of sadist are you?  😭

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: image
https://docs.google.com/document/d/15L5iaWo4ieouegpSlaqGGWpA13JnTv73xnpxuVhzkBg/edit?ts=5cc77eb3

Hoten: silly that u cant link to doc comments

paulirish: aiight. i'm not feeling this UX at all, so I raised a concern there: https://docs.google.com/a/google.com/document/d/15L5iaWo4ieouegpSlaqGGWpA13JnTv73xnpxuVhzkBg/edit?disco=AAAAC6njMnI


> silly that u cant link to doc comments

image

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -200,6 +200,11 @@ class CategoryRenderer {
       descriptionEl.classList.add('lh-audit-group__description');
       auditGroupHeader.appendChild(descriptionEl);
     }
+    if (group.title === 'Metrics') {
```
brendankenny: should move this over to `performance-category-renderer.js`. It'll have to do an extra step to re-find the header, but it would be good to keep specialized knowledge of the perf category encapsulated in that file if we can

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -157,6 +165,24 @@ class ReportUIFeatures {
     this._copyAttempt = false;
   }
 
+  _setupMetricDescriptionToggleElements() {
+    this.metricDescriptionToggleEl =
+      /** @type {HTMLInputElement} */ (this._dom.find('.lh-metrics-toggle input', this._document));
```
brendankenny: will need to guard against when there's no perf category

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -60,6 +64,9 @@ class ReportUIFeatures {
     this.expandAllDetails = this.expandAllDetails.bind(this);
     this._toggleDarkTheme = this._toggleDarkTheme.bind(this);
     this._updateStickyHeaderOnScroll = this._updateStickyHeaderOnScroll.bind(this);
+    this._toggleMetricDescription = this._toggleMetricDescription.bind(this);
+    this._setupMetricDescriptionToggleElements =
```
brendankenny: doesn't look like it need to be bound?

lighthouse-core/report/html/report-styles.css
Hoten: done-ish. shoot, that icon width varies based on the state ...

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: hm?

Hoten:
@brendankenny added tests. Also tried a refactor (`render fn`) but not sure if better

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
```diff
@@ -26,11 +26,23 @@ const TEMPLATE_FILE_REPORT = fs.readFileSync(__dirname +
...
 describe('ReportUIFeatures', () => {
-  let renderer;
-  let reportUIFeatures;
   let sampleResults;
   let dom;
 
+  /**
+   * @param {LH.JSON} lhr
+   */
+  function render(lhr) {
```
brendankenny: render function looks 👍 

main issue is that it can be pretty slow, so we want to share rendered reports between tests as much as possible, but that's what you do here, so looks good to me

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
```diff
@@ -96,16 +104,11 @@ describe('ReportUIFeatures', () => {
 
   describe('initFeatures', () => {
     it('should init a report', () => {
-      // render a report onto the UIFeature dom
-      const container = reportUIFeatures._dom._document.querySelector('main');
-      renderer.renderReport(sampleResults, container);
-
-      assert.equal(reportUIFeatures.json, undefined);
-      reportUIFeatures.initFeatures(sampleResults);
-      assert.ok(reportUIFeatures.json);
```
brendankenny: testing `json` was being used as a proxy test for the reportUIFeatures being initialized, otherwise this is just testing the report renderer.

We could definitely do something much better if you want to come up with something to test :) or just restore and we'll think about it another day

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
```diff
@@ -161,4 +162,41 @@ describe('ReportUIFeatures', () => {
...
+
+  describe('metric description toggles', () => {
+    let container;
+    let metricsAuditGroup;
+    let toggle;
+    const metricsClass = 'lh-audit-group--metrics';
+    const toggleClass = 'lh-metrics-toggle__input';
+    const showClass = 'lh-audit-group--metrics__show-descriptions';
+
+    describe('works', () => {
```
brendankenny:
```suggestion
    describe('works if there is a performance category', () => {
```
or whatever

lighthouse-core/report/html/report-styles.css
```diff
@@ -795,6 +855,14 @@
   font-weight: bold;
 }
 
+/* Pushes the metric description toggle button to the right. */
```
brendankenny: not sure if it's really better down here or up with the toggle stuff

lighthouse-core/report/html/renderer/report-ui-features.js
brendankenny: we really only need one click listener, then just make sure the target was an `.lh-metric` (though it might as well be on the entire section considering how much of it `lh-metric`s take up...I think I agree with @paulirish about the clicking thing)

paulirish: here's event delegation. and also only using the metric titles as click targets.

```js

    this.metricAuditGroup.addEventListener('click', e => {
      if (e.target.closest('.lh-metric__title')) this.metricDescriptionToggleEl.click()
    });
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -553,11 +560,64 @@
...
+.lh-metrics-toggle__input:not(:checked) + label .lh-metrics-toggle__icon--less,
+.lh-metrics-toggle__input:checked + label .lh-metrics-toggle__icon--more {
+  background-color: var(--color-blue);
+}
+.lh-metrics-toggle__lines {
+  fill: var(--color-metric-toggle-lines);
+}
+.lh-metrics-toggle__input:not(:checked) + label .lh-metrics-toggle__icon--less .lh-metrics-toggle__lines,
+.lh-metrics-toggle__input:checked + label .lh-metrics-toggle__icon--more .lh-metrics-toggle__lines {
+  --color-metric-toggle-lines: var(--color-white);
```
paulirish: now you can move this prop into the L601 rule and delete this L608 rule

lighthouse-core/report/html/report-styles.css
```diff
@@ -553,11 +560,64 @@
...
+}
+.lh-metrics-toggle label .lh-metrics-toggle__icon--less {
```
paulirish: the base `.lh-metrics-toggle__icon` rule could assume the `--less` case  so this rule isnt necessary

regardless you can move width/height into the  base `lh-metrics-toggle__icon` rule

lighthouse-core/report/html/templates.html
```diff
@@ -60,13 +60,36 @@
...
+<template id="tmpl-lh-metrics-toggle">
+  <div class="lh-metrics-toggle">
+    <input class="lh-metrics-toggle__input" type="checkbox" name="toggle-metric-descriptions">
+    <label for="toggle-metric-descriptions">
+      <div class="lh-metrics-toggle__icon lh-metrics-toggle__icon--less" aria-hidden="true">
+        <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">
+          <defs>
+            <path id="a" d="M0 0h24v24H0V0z" />
+          </defs>
+          <path class="lh-metrics-toggle__lines" d="M4 9h16v2H4zm0 4h10v2H4z" clip-path="url(#b)" />
```
paulirish: i think you can just delete clip-path attribute and the `<defs>` above. it seems to be ignored.

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
Hoten: How about just removing it? Doesn't seem useful now that we have all these other tests checking individual behaviors.

Hoten:
not crazy about how the tab control works, thoughts?

lighthouse-core/test/report/html/renderer/report-ui-features-test.js
brendankenny: > How about just removing it?

seems fine now, yeah

brendankenny:
up to @paulirish on if he likes this :)

PR #8708, opened 2019-04-29
report(lightwallet): render performance-budget section

khempenius:
**Summary**

**Related Issues/PRs**
#8675, #8427, #8709, #8708,#8522, #8539

Screen Shot 2019-04-29 at 8 55 04 AM


lighthouse-core/report/html/report-styles.css
```diff
@@ -679,6 +679,19 @@
   vertical-align: middle;
 }
 
+.lh-audit-group--budgets .lh-table tbody tr td:nth-child(4),
+.lh-audit-group--budgets .lh-table tbody tr td:nth-child(5){
```
paulirish: might as well do this in the next PR where you add support for the 'over budget' column?

already thinking you could use a `.lh-table-column--overbudge` class instead of nth-child.

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -122,6 +122,21 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
       element.appendChild(this.renderCategoryHeader(category, groups));
     }
 
+    // Budgets
```
paulirish: I think the Budgets block should follow **after** Metrics.

I know we didn't discuss this much before.. so apologies for that. :/


a few reasons why, though:

1. The top level metrics are just that... 'top level'.. and it's a bit confusing to put something above them.
1. The perf score is determined from the metrics, and their current proximity helps make this connection. Putting budgets in between adds some confusion... ("Does budget failure affect perf score?")
1. Just considering screen real-estate, the budgets block is really tall, and pushes the metrics often below the fold.

@khempenius wdyt?

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -122,6 +122,21 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
+    const budgetAudit = category.auditRefs.find((audit) => audit.id === 'resource-budget');
+    if (budgetAudit && budgetAudit.result && budgetAudit.result.details) {
+      const budgetsGroupEl = this.renderAuditGroup(groups['budgets']);
```
brendankenny:
```suggestion
      const budgetsGroupEl = this.renderAuditGroup(groups.budgets);
```

brendankenny:
```suggestion
      const budgetsGroupEl = this.renderAuditGroup(groups.budgets);
```

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -122,6 +122,21 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
...
+      const tmpl = this.dom.cloneTemplate('#tmpl-lh-opportunity-header', this.templateContext);
+      budgetsGroupEl.appendChild(this.dom.find('.lh-load-opportunity__header', tmpl));
```
brendankenny: can you make a new header template (even if it's just a copy with new budget class names)

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -122,6 +122,21 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
       element.appendChild(this.renderCategoryHeader(category, groups));
     }
 
+    // Budgets
+    const budgetAudit = category.auditRefs.find((audit) => audit.id === 'resource-budget');
+    if (budgetAudit && budgetAudit.result && budgetAudit.result.details) {
```
brendankenny:
```suggestion
    if (budgetAudit && budgetAudit.result.details) {
```
(`budgetAudit.result` is always defined if `budgetAudit` is)

lighthouse-core/report/html/report-styles.css
khempenius: This PR "has" the over budget column (see below for what it looks like when all the branches are together) in the sense that it renders it, but @brendankenny wanted me to split the renderer code from the audit code for PR purposes.

Agreed that this CSS is kinda gross. However, I did it that way because it allows the budgets table to re-use Audit.makeTableDetails() so there's really no custom code - just the code in this branch which all reuses existing methods. If you think this is worth changing, I'd be happy to do that.

With a budget:
// Only lists resources that have budgets associated with them.
Screen Shot 2019-04-23 at 2 35 19 AM

Without a budget:
// Lists all resources
Screen Shot 2019-04-23 at 2 26 29 AM

lighthouse-core/report/html/renderer/performance-category-renderer.js
khempenius: Totally fine with me (& I think that makes more sense) - I was just following the mocks.

paulirish: groovy! lets switcheroo. :)

lighthouse-core/report/html/report-styles.css
paulirish: whats the easiest way for me to generate that first screenshot of yours?

khempenius: The easiest way would be to use the khempenius/lightwallet_pt2 branch.

That was the branch that I used to split this out into separate PRs. It doesn't have some of the minor code cleanup changes that have occurred on the PR branches, but is functionally the same and has all the upstream dependencies necessary to generate this.

lighthouse-core/report/html/renderer/performance-category-renderer.js
paulirish: agree with the others but i dont think its necessary to maek a new header template. seems okay to reuse this one.

brendankenny: > agree with the others but i dont think its necessary to maek a new header template. seems okay to reuse this one.

booooooo 😛 

It's 300 bytes. We can afford them in exchange for avoiding the question "why does the budget section have a `lh-load-opportunity__header` className? :)

Make it a two-column performance header and then we can talk

khempenius:
Updated.

brendankenny:
@paulirish is it possible to redeploy this with new master?

paulirish:
woohoo! now we got budget section in the now deployment! https://lighthouse-gce0hkow5.now.sh/

image

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -151,7 +151,8 @@ describe('PerfCategoryRenderer', () => {
     const passedSection = categoryDOM.querySelector('.lh-category > .lh-clump--passed');
 
     const passedAudits = category.auditRefs.filter(audit =>
-        audit.group && audit.group !== 'metrics' && Util.showAsPassed(audit.result));
+      (audit.group === 'load-opportunities' || audit.group === 'diagnostics')
+      && audit.group && Util.showAsPassed(audit.result));
```
exterkamp: The `&& audit.group` seems redundant here afaict.
```suggestion
(audit.group === 'load-opportunities' || audit.group === 'diagnostics')
      && Util.showAsPassed(audit.result));
```

lighthouse-core/test/report/html/renderer/performance-category-renderer-test.js
```diff
@@ -68,7 +68,7 @@ describe('PerfCategoryRenderer', () => {
   it('renders the sections', () => {
     const categoryDOM = renderer.render(category, sampleResults.categoryGroups);
     const sections = categoryDOM.querySelectorAll('.lh-category > .lh-audit-group');
-    assert.equal(sections.length, 4);
+    assert.equal(sections.length, 5);
```
brendankenny: should also add some tests to ensure rendering is happening, etc. We already have tests for the table details renderer, so it can probably stay relatively light. Starting suggestion:
```js
describe('budgets', () => {
  it('renders a performance budget', () => {
    const categoryDOM = renderer.render(category, sampleResults.categoryGroups);

    const budgetsGroup = categoryDOM.querySelector('.lh-audit-group.lh-audit-group--budgets');
    assert.ok(budgetsGroup);
    const budgetTable = budgetsGroup.querySelector('#performance-budget.lh-table');
    assert.ok(budgetTable);

    const lhrBudgetEntries = sampleResults.audits['performance-budget'].details.items;
    const tableRows = budgetTable.querySelectorAll('tbody > tr');
    assert.strictEqual(tableRows.length, lhrBudgetEntries.length);
  });

  it('does not render a budget table when performance-budget audit is notApplicable', () => {
    const budgetlessCategory = JSON.parse(JSON.stringify(category));
    const budgetRef = budgetlessCategory.auditRefs.find(a => a.id === 'performance-budget');
    budgetRef.result.scoreDisplayMode = 'notApplicable';
    delete budgetRef.result.details;

    const categoryDOM = renderer.render(budgetlessCategory, sampleResults.categoryGroups);
    const budgetsGroup = categoryDOM.querySelector('.lh-audit-group.lh-audit-group--budgets');
    assert.strictEqual(budgetsGroup, null);
  });
});
```

lighthouse-core/config/default-config.js
```diff
@@ -379,6 +379,7 @@ const defaultConfig = {
         {id: 'time-to-first-byte', weight: 0, group: 'load-opportunities'},
         {id: 'redirects', weight: 0, group: 'load-opportunities'},
         {id: 'uses-rel-preload', weight: 0, group: 'load-opportunities'},
+        {id: 'performance-budget', weight: 0, group: 'budgets'},
```
patrickhulce: did I miss something about why this was moved?

kinda weird for it to be in the middle of `load-opportunities`

lighthouse-core/report/html/renderer/performance-category-renderer.js
```diff
@@ -165,6 +165,20 @@ class PerformanceCategoryRenderer extends CategoryRenderer {
       filmstripEl && timelineEl.appendChild(filmstripEl);
     }
 
+    // Budgets
+    const budgetAudit = category.auditRefs.find((audit) => audit.id === 'performance-budget');
```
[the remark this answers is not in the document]
patrickhulce: also not sure if we should do this by audit id or by group and/or if there's any point to having the `budgets` group anymore if we keep it by id?

FWIW I think I'd lean towards group since that's how we're doing everything else, but then again it might be a bit confusing that it's a single audit in the group so 🤷‍♂

lighthouse-core/test/report/html/renderer/report-renderer-test.js
```diff
@@ -255,7 +255,8 @@ describe('ReportRenderer', () => {
 
     let notApplicableCount = 0;
     Object.values(clonedSampleResult.audits).forEach(audit => {
-      if (audit.scoreDisplayMode === 'notApplicable') {
+      // The performance-budget audit is omitted from the DOM if it is not applicable
```
patrickhulce: I missed that we were building the actual expected count since that was down on 271 and I just looked at 265, WDYT

```suggestion
      // The performance-budget audit is omitted from the DOM when it is not applicable, so omit from our expected count too.
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -852,6 +852,19 @@
   vertical-align: middle;
 }
 
+.lh-audit-group--budgets .lh-table tbody tr td:nth-child(4),
```
patrickhulce: could we get a comment explainin what `nth-child(4)` and 5 are in case they change in the future?

```suggestion
/* Style the "over budget" columns red. */
.lh-audit-group--budgets .lh-table tbody tr td:nth-child(4),
```

lighthouse-core/report/html/report-styles.css
```diff
@@ -852,6 +852,19 @@
...
+.lh-audit-group--budgets .lh-table tbody tr td:nth-child(5){
+  color: var(--color-red-700);
+}
+
+.lh-audit-group--budgets .lh-table tbody tr td:nth-child(4){
```
patrickhulce:
```suggestion
/* Align the "over budget request count" text to be close to the "over budget bytes" column.
.lh-audit-group--budgets .lh-table tbody tr td:nth-child(4){
```

patrickhulce: or whatever the reason is we want it aligned right :)

khempenius:
Updated.

All resolved except this comment from @patrickhulce:
"also not sure if we should do this by audit id or by group and/or if there's any point to having the budgets group anymore if we keep it by id?

FWIW I think I'd lean towards group since that's how we're doing everything else, but then again it might be a bit confusing that it's a single audit in the group so 🤷‍♂"

brendankenny:
🎉 🎉

PR #8724, opened 2019-04-29
core: filter out blob urls more often

paulirish:
I noticed this blob url in our caching audit.
Doesn't make much sense. :)

image


lighthouse-core/test/audits/byte-efficiency/uses-long-cache-ttl-test.js
```diff
@@ -24,6 +24,7 @@ function networkRecord(options = {}) {
     statusCode: options.statusCode || 200,
     resourceType: options.resourceType || NetworkRequest.TYPES.Script,
     transferSize: options.transferSize || 10000,
+    protocol: options.protocol || 'http/1.1',
```
patrickhulce: super confusing that `protocol` on devtools isn't `new URL().protocol` 😩

patrickhulce:
(needs sample json update though :))

PR #8726, opened 2019-04-30
misc(chore): Common Typos

exterkamp:
**Summary**
Caught by common-typos again.  And we would've gotten away with it too!

Fixing seperately in g3 for :hourglass: sake.

### DIFF
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -335,7 +335,7 @@ class DetailsRenderer {
 
   /**
    * @param {LH.Audit.Details.List} details
-   * @returns {Element}
+   * @return {Element}
    */
   _renderList(details) {
     const listContainer = this._dom.createElement('div', 'lh-list');
--- a/lighthouse-core/report/html/templates.html
+++ b/lighthouse-core/report/html/templates.html
@@ -617,7 +617,7 @@ limitations under the License.
       color: var(--body-text-color);
     }
 
-    /* TODO(#8185) use more BEM (.lh-gauge__label--big) instead of relying on descendent selector */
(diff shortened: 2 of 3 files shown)


lighthouse-core/report/html/templates.html
```diff
@@ -617,7 +617,7 @@
       color: var(--body-text-color);
     }
 
-    /* TODO(#8185) use more BEM (.lh-gauge__label--big) instead of relying on descendent selector */
+    /* TODO(#8185) use more BEM (.lh-gauge__label--big) instead of relying on descendant selector */
```
patrickhulce: I stand by @Hoten on this one

Oxford says descendant === noun, descendent === adjective
https://en.oxforddictionaries.com/definition/descendant

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -335,7 +335,7 @@ class DetailsRenderer {
 
   /**
    * @param {LH.Audit.Details.List} details
-   * @returns {Element}
+   * @return {Element}
```
patrickhulce: fwiw http://usejsdoc.org/tags-returns.html

brendankenny: > fwiw http://usejsdoc.org/tags-returns.html

I don't feel bad about us standardizing on one of these :)

brendankenny:
with this and the `@return {}` discussion, maybe we should look at jsdoc linting. Last time I looked it was pretty heavy handed, but maybe it's improved. It would be nice to just have it automated

lighthouse-core/report/html/renderer/details-renderer.js
patrickhulce: Oh I don't feel bad about standardizing on one either :) 

but I couldn't stand to left `@returns` be called incorrect when in fact it is the actual "spec" one

PR #8727, opened 2019-04-30
core(lightwallet): stricter verification of budget types

brendankenny:
@khempenius @patrickhulce 

This is my proposal for slightly stricter checking of the budgets file, primarily by assuming nothing about the type when it comes into `budget.js` so that we can get a little more verification by the compiler that we're checking the types well.

There are also a few places where the type checks have been tightened and some more tests have been added.

Even though this now returns a totally new budget object instead of just validating in place, it does have the same issue as in #8675 where *in practice* it's just validating in place because it throws on any input that isn't valid. Renaming to reflect that as suggested in https://github.com/GoogleChrome/lighthouse/pull/8675#discussion_r279401765 makes a lot of sense to me, though not creating the object on the fly may reduce tsc's ability to verify the type. We could always create but not return? Or maybe it doesn't matter.

### DIFF
--- a/lighthouse-core/config/budget.js
+++ b/lighthouse-core/config/budget.js
@@ -5,10 +5,36 @@
  */
 'use strict';
 
+/**
+ * @param {unknown} arr
+ * @return {arr is Array<Record<string, unknown>>}
+ */
+function isArrayOfUnknownObjects(arr) {
+  return Array.isArray(arr) && arr.every(isObjectOfUnknownProperties);
+}
+
+/**
+ * @param {unknown} val
+ * @return {val is Record<string, unknown>}
+ */
+function isObjectOfUnknownProperties(val) {
+  return typeof val === 'object' && val !== null && !Array.isArray(val);
+}
+
+/**
+ * Returns whether `val` is numeric. Will not coerce to a number. `NaN` will
+ * return false, however ±Infinity will return true.
+ * @param {unknown} val
+ * @return {val is number}
+ */
+function isNumber(val) {
+  return typeof val === 'number' && !isNaN(val);
+}
+
 class Budget {
   /**
    * Asserts that obj has no own properties, throwing a nice error message if it does.
-   * Plugin and object name are included for nicer logging.
+   * `objectName` is included for nicer logging.
    * @param {Record<string, unknown>} obj
    * @param {string} objectName
    */
@@ -21,13 +47,14 @@ class Budget {
   }
 
   /**
-   * @param {LH.Budget.ResourceBudget} resourceBudget
+   * @param {Record<string, unknown>} resourceBudget
    * @return {LH.Budget.ResourceBudget}
    */
   static validateResourceBudget(resourceBudget) {
     const {resourceType, budget, ...invalidRest} = resourceBudget;
     Budget.assertNoExcessProperties(invalidRest, 'Resource Budget');
 
+    /** @type {Array<LH.Budget.ResourceType>} */
     const validResourceTypes = [
       'total',
       'document',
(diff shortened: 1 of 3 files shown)


lighthouse-core/config/budget.js
```diff
@@ -87,43 +117,44 @@ class Budget {
   /**
    * More info on the Budget format:
    * https://github.com/GoogleChrome/lighthouse/issues/6053#issuecomment-428385930
-   * @param {Array<LH.Budget>} budgetArr
+   * @param {unknown} budgetJson
```
brendankenny: it seems a little weird to act like `cliFlags.budgets` is an `Array<LH.Budget>` in all of `lighthouse-cli/`, but then treat it distrustfully as soon as it gets into this file, but switching `SharedFlagsSettings['budgets']` to `unknown also seemed a bit silly because by that argument *all* of the CLI flags should be treated as `unknown` until we verify them.

And maybe we should be, but then that degrades the handiness and safety of having a single verification point inside `config.js`, and I'm not sure it's worth verifying in `bin.js` *and* in `config.js`...

lighthouse-core/config/budget.js
```diff
@@ -5,10 +5,36 @@
  */
 'use strict';
 
+/**
+ * @param {unknown} arr
+ * @return {arr is Array<Record<string, unknown>>}
+ */
+function isArrayOfUnknownObjects(arr) {
```
patrickhulce: i'd love to share more of this logic across the config files

lighthouse-core/config/budget.js
```diff
@@ -39,46 +66,49 @@ class Budget {
...
         `Valid timing metrics are: ${validTimingMetrics.join(', ')}`);
     }
-    if (isNaN(timingBudget.budget)) {
-      throw new Error('Invalid budget: ${timingBudget.budget}');
+    if (!isNumber(budget)) {
+      throw new Error(`Invalid budget: ${budget}`);
     }
-    if (timingBudget.tolerance !== undefined && isNaN(timingBudget.tolerance)) {
-      throw new Error('Invalid tolerance: ${timingBudget.tolerance}');
+    if (typeof tolerance !== 'undefined' && !isNumber(tolerance)) {
```
patrickhulce: I thought you were against `typeof x === 'undefined'` checks @brendankenny ? 😛

lighthouse-core/test/config/budget-test.js
```diff
@@ -74,55 +74,103 @@ describe('Budget', () => {
...
+      budget[0].timings[0].metric = 'medianMeaningfulPaint';
+      assert.throws(_ => Budget.initializeBudget(budget),
+        // eslint-disable-next-line max-len
+        /Invalid timing metric: medianMeaningfulPaint. \nValid timing metrics are: first-contentful-paint, /);
     });
 
     it('throws when an invalid budget is supplied', () => {
       budget[0].timings[0].budget = '100KB';
-      assert.throws(_ => Budget.initializeBudget(budget), /Invalid budget/);
+      assert.throws(_ => Budget.initializeBudget(budget), / Invalid budget: 100KB/);
```
patrickhulce: why the space...?

khempenius:
LGTM.

lighthouse-core/config/budget.js
brendankenny: > i'd love to share more of this logic across the config files

agreed!

lighthouse-core/config/budget.js
brendankenny: > I thought you were against `typeof x === 'undefined'` checks @brendankenny ? 😛

haha, I am! For some reason typescript won't treat it as `number | undefined` with `tolerance !== undefined`, and it's really annoying :)

lighthouse-core/test/config/budget-test.js
brendankenny: > why the space...?

bad at copy/paste :)

PR #8729, opened 2019-04-30
core: replace EIL with MP-FID

patrickhulce:
**Summary**
Promotes Max Potential FID to the spot of EIL, moves EIL to JSON only.

> @patrickhulce: oh we're def making that happen for v5? 😀
> @paulirish: ya

It's difficult to analyze our thresholds because the HTTPArchive data we have on max potential FID will all by impacted by the missing microtask data bug (see #8379), but MPFID differentiates **MUCH* better than than EIL, they feel about right, and its still not a scored metric anyway...

**Ventiles**
```sql
SELECT
  APPROX_QUANTILES(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.max-potential-fid.rawValue') as FLOAT64),
    20) as mpfid,
  APPROX_QUANTILES(CAST(JSON_EXTRACT_SCALAR(report, '$.audits.estimated-input-latency.rawValue') as FLOAT64),
    20) as eil
FROM
  httparchive.lighthouse.2019_03_01_mobile
```
image

**Related Issues/PRs**
picks up the mantle of #5842 and raises it to glory.

### DIFF
--- a/lighthouse-core/audits/metrics/max-potential-fid.js
+++ b/lighthouse-core/audits/metrics/max-potential-fid.js
@@ -13,8 +13,8 @@ const UIStrings = {
   /** The name of the metric "Maximum Potential First Input Delay" that marks the maximum estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. Shown to users as the label for the numeric metric value. Ideally fits within a ~40 character limit. */
   title: 'Max Potential FID',
   /** Description of the Maximum Potential First Input Delay metric that marks the maximum estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. This description is displayed within a tooltip when the user hovers on the metric name to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
-  description: 'The potential First Input Delay that your users could experience is the ' +
-      'duration, in milliseconds, of the longest task.',
+  description: 'The maximum potential First Input Delay that your users could experience is the ' +
+      'duration, in milliseconds, of the longest task. [Learn more](https://developers.google.com/web/updates/2018/05/first-input-delay).',
 };
 
 const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
--- a/lighthouse-cli/test/cli/__snapshots__/index-test.js.snap
+++ b/lighthouse-cli/test/cli/__snapshots__/index-test.js.snap
@@ -694,11 +694,11 @@ Object {
         },
         Object {
           "group": "metrics",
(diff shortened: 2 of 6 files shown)


lighthouse-core/audits/metrics/max-potential-fid.js
```diff
@@ -13,8 +13,8 @@ const UIStrings = {
   /** The name of the metric "Maximum Potential First Input Delay" that marks the maximum estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. Shown to users as the label for the numeric metric value. Ideally fits within a ~40 character limit. */
   title: 'Max Potential FID',
   /** Description of the Maximum Potential First Input Delay metric that marks the maximum estimated time between the page receiving input (a user clicking, tapping, or typing) and the page responding. This description is displayed within a tooltip when the user hovers on the metric name to see more. No character length limits. 'Learn More' becomes link text to additional documentation. */
-  description: 'The potential First Input Delay that your users could experience is the ' +
-      'duration, in milliseconds, of the longest task.',
+  description: 'The maximum potential First Input Delay that your users could experience is the ' +
+      'duration, in milliseconds, of the longest task. [Learn more](https://developers.google.com/web/updates/2018/05/first-input-delay).',
```
brendankenny: Not having the "maximum" part of the string in all the translations (until the next round come out) isn't a huge deal, but  how much do we care about getting this link out to everyone? We could manually edit the other languages to add the `Learn More` link.

brendankenny:
> but MPFID differentiates *MUCH better than than EIL, they feel about right

that's because one tries to be robust to outliers and the other is "here's your outlier", but w/e :P

patrickhulce:
> that's because one tries to be robust to outliers and the other is "here's your outlier", but w/e :P

apologies, didn't intend to be hating on EIL's usefulness as a whole. I know it's by design but it's definitively not as suitable for differentiating perf scores on top half of the web that's all

lighthouse-core/audits/metrics/max-potential-fid.js
brendankenny: > We could manually edit the other languages to add the Learn More link.

note: this could also be a really really bad idea :)

patrickhulce: Ha, Given that these are hidden by default, IMO it's probably not the end of the world to update these ASAP, but others could feel differently.

brendankenny: Yeah, SG

PR #8387, opened 2019-04-18
misc: add releasing scripts

patrickhulce:
**Summary**
Step 1 to writing a script to auto test our tarball without the test files bundled is writing a script to auto test our tarball *with* the test files :)

PR #8751, opened 2019-04-30
misc: rename release scripts properly

patrickhulce:
**Summary**
As @brendankenny and @Hoten pointed out the release scripts didn't get renamed. I must've messed something up with git mv because all the references got updated but not the files themselves.

Sorry for the noise!

**Related Issues/PRs**
#8387


Hoten:
image
:)

patrickhulce:
> is it possible to drop the second release, so it's just release/clean-pristine.sh, release/prepare-commit.sh, etc?

done!

PR #8752, opened 2019-04-30
clients(lr): revert Lightrider timings applied to NetworkRequests

exterkamp:
This reverts commit ca1731aba7df31b16f68abc0de23f9fffac38efd, which _may have_, _slightly_, _kinda_, messed up some performance data.  Or whatever, I mean you never know.

### DIFF
--- a/lighthouse-core/test/lib/network-request-test.js
+++ b/lighthouse-core/test/lib/network-request-test.js
@@ -176,5 +176,43 @@ describe('NetworkRequest', () => {
       expect(record).toMatchObject(req);
       expect(record.lrStatistics).toStrictEqual(undefined);
     });
+
+    it('treats negative timings as 0', function() {
+      const req = getRequest();
+      req.responseHeaders = [{name: NetworkRequest.HEADER_TOTAL, value: '10000'},
+        {name: NetworkRequest.HEADER_TCP, value: '-1'},
+        {name: NetworkRequest.HEADER_REQ, value: '-1'},
+        {name: NetworkRequest.HEADER_SSL, value: '-1'},
+        {name: NetworkRequest.HEADER_RES, value: '10000'}];
+
+      const devtoolsLog = networkRecordsToDevtoolsLog([req]);
+      global.isLightrider = true;
+      const record = NetworkRecorder.recordsFromLogs(devtoolsLog)[0];
+
+      expect(record.lrStatistics).toStrictEqual({
+        endTimeDeltaMs: -8000,
+        TCPMs: 0,
+        requestMs: 0,
+        responseMs: 10000,
+      });
+    });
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/lib/network-request-test.js
```diff
@@ -185,133 +176,5 @@ describe('NetworkRequest', () => {
       expect(record).toMatchObject(req);
       expect(record.lrStatistics).toStrictEqual(undefined);
     });
-
-    it('treats negative timings as 0', function() {
```
brendankenny: if we're going to keep the rest of the header code, we should keep these tests as they significantly improved coverage of the code. Would just need to remove (or better, update?) the `startTime`/`endTime`/`responseReceivedTime` numbers

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -185,133 +176,5 @@ describe('NetworkRequest', () => {
...
-      });
-      expect(record.lrStatistics).toStrictEqual({
-        endTimeDeltaMs: -8000,
-        TCPMs: 1000,
-        requestMs: 0,
-        responseMs: 9000,
-      });
-    });
-
-    it('created a new timing property if one did not exist', function() {
```
brendankenny: I guess we won't need this one

lighthouse-core/test/lib/network-request-test.js
```diff
@@ -185,133 +176,5 @@ describe('NetworkRequest', () => {
...
-      });
-      expect(lrRecord.lrStatistics).toStrictEqual({
-        endTimeDeltaMs: -8000,
-        TCPMs: 5000,
-        requestMs: 2500,
-        responseMs: 2500,
-      });
-    });
-
-    it('overrides existing timing properties', function() {
```
brendankenny: but we will still want this one? (but updated, obvs)

lighthouse-core/test/lib/network-request-test.js
exterkamp: We have seperate checks for the timing no-op this is just for negatives, so I removed all the startTime/endTime stuff

lighthouse-core/test/lib/network-request-test.js
exterkamp: I think this is covered by `updates lrStatistics if in Lightrider` which makes sure no record data is different.  I can expand that one to timing data as well, but it will be all `-1`'s i think

exterkamp: actually timing is `null` in that test.  hm.

exterkamp: Readded :smile:

PR #8760, opened 2019-05-01
core(runner): audits can only use requiredArtifacts

patrickhulce:
**Summary**
One more breaking change I want to get in we discussed in the meeting that I misunderstood and thought was already on someone's plate :) 

**an audit can only use its `requiredArtifacts`**

With plugins it feels like folks aren't going to remember to use `requiredArtifacts` and if we ever want to enforce it, seems like we need to do it now.

This also fixes the dozen or so audits that were cheating :)

### DIFF
--- a/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
+++ b/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
@@ -70,10 +70,9 @@ class RenderBlockingResources extends Audit {
       title: str_(UIStrings.title),
       scoreDisplayMode: Audit.SCORING_MODES.NUMERIC,
       description: str_(UIStrings.description),
-      // This audit also looks at CSSUsage but has a graceful fallback if it failed, so do not mark
-      // it as a "requiredArtifact".
-      // TODO: look into adding an `optionalArtifacts` property that captures this
-      requiredArtifacts: ['URL', 'TagsBlockingFirstPaint', 'traces', 'devtoolsLogs'],
+      // TODO: look into adding an `optionalArtifacts` property that captures the non-required nature
+      // of CSSUsage
+      requiredArtifacts: ['URL', 'TagsBlockingFirstPaint', 'traces', 'devtoolsLogs', 'CSSUsage'],
     };
   }
 
--- a/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
+++ b/lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
@@ -73,7 +73,7 @@ class RenderBlockingResources extends Audit {
       // This audit also looks at CSSUsage but has a graceful fallback if it failed, so do not mark
       // it as a "requiredArtifact".
       // TODO: look into adding an `optionalArtifacts` property that captures this
-      requiredArtifacts: ['URL', 'TagsBlockingFirstPaint', 'traces'],
+      requiredArtifacts: ['URL', 'TagsBlockingFirstPaint', 'traces', 'devtoolsLogs'],
     };
   }
 
--- a/lighthouse-core/computed/viewport-meta.js
+++ b/lighthouse-core/computed/viewport-meta.js
@@ -11,10 +11,10 @@ const makeComputedArtifact = require('./computed-artifact.js');
 
 class ViewportMeta {
   /**
-   * @param {{MetaElements: LH.GathererArtifacts['MetaElements']}} artifacts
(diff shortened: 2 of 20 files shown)


lighthouse-core/runner.js
```diff
@@ -325,7 +325,12 @@ class Runner {
         ...sharedAuditContext,
       };
 
-      const product = await audit.audit(artifacts, auditContext);
+      const requiredArtifacts = audit.meta.requiredArtifacts
```
patrickhulce: the magic

lighthouse-core/computed/viewport-meta.js
```diff
@@ -11,10 +11,10 @@ const makeComputedArtifact = require('./computed-artifact.js');
 
 class ViewportMeta {
   /**
-   * @param {{MetaElements: LH.GathererArtifacts['MetaElements']}} artifacts
+   * @param {LH.GathererArtifacts['MetaElements']} MetaElements
    * @return {Promise<ViewportMetaResult>}
   */
-  static async compute_({MetaElements}) {
+  static async compute_(MetaElements) {
```
patrickhulce: a necessary refactor to ensure that they're still cached, also this seems like a bad idea to cache on the entire artifacts object anyway :)

brendankenny: lol whoops

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
```diff
@@ -73,7 +73,7 @@ class RenderBlockingResources extends Audit {
       // This audit also looks at CSSUsage but has a graceful fallback if it failed, so do not mark
```
brendankenny: do you want to change anything about `CSSUsage` for this audit, since after this PR it won't ever see it unless we get around to optionalArtifacts?

lighthouse-core/runner.js
brendankenny: can you add a comment about narrowing down the artifacts to just what is requested by the audit

lighthouse-core/runner.js
```diff
@@ -325,7 +325,12 @@ class Runner {
...
+        .reduce((requiredArtifacts, artifactName) => {
+          requiredArtifacts[artifactName] = artifacts[artifactName];
+          return requiredArtifacts;
+        }, /** @type {LH.Artifacts} */ ({}));
```
brendankenny: this is unfortunate, but fine enough. Maybe add a line about the masquerading type to the above comment as well?

I was hoping at one point that we could type the `artifacts` input of an audit based on its declared `requiredArtifacts`, but it would require not having `Array<string>` for the `requiredArtifacts` type, was finicky because it was nested on the `meta` getter, and didn't gain us all that much, but maybe in some future version of typescript...

lighthouse-core/audits/byte-efficiency/render-blocking-resources.js
patrickhulce: oh yeah nice catch!

lighthouse-core/runner.js
patrickhulce: 👍 done

lighthouse-core/runner.js
patrickhulce: yeah I thought about it for about 5 seconds before thinking there was no way it was going to be usable 😆 

hopefully one day typescript will let you work your magic here :)

lighthouse-core/test/runner-test.js
```diff
@@ -339,6 +339,38 @@ describe('Runner', () => {
...
+      const config = new Config({
+        settings: {
+          auditMode: __dirname + '/fixtures/artifacts/alphabet-artifacts/',
+        },
+        audits: [
+          SimpleAudit,
+        ],
+      });
+
+      const results = await Runner.run({}, {config})
```
brendankenny: lint error here

patrickhulce: pesky semicolons

brendankenny:
ha, fascinating. The `yarn smoke dbw` failure is real (going from 6 console errors to 8). Turns out since `CSSUsage` was optional it wasn't running via the dbw config. Now that it does run, it appears to be causing re-requests of two of the 404ing stylesheets.

Is there an easy check we can add to `css-usage.js` or should we just bump the error count for now?

patrickhulce:
> Is there an easy check we can add to css-usage.js or should we just bump the error count for now?

We can add a check but it's non trivial and I'm not sure it's worth it :) I'll bump the error count for now 👍

Issue #8331, opened 2019-04-16
[meta] 2019 I/O Burndown


paulirish:
Channeling our [ancestors](https://github.com/GoogleChrome/lighthouse/issues/5008), we introduce this year's abbreviated I/O burndown list.

## Report
- [ ] Redesign #8185
- [x] Improved report handling in g3 #8197  [@exterkamp]
- [x] ensure 4.x LHR and 5.0 report renderer compatibility #8598
- [x] third-party table filtering #6351

## Core
- [x] Ship LightWallet / Perf Budgeting [@khempenius]
   - [x] [Learn more] link for lightwallet audits https://github.com/GoogleChrome/lighthouse/pull/8522#discussion_r279479493 (PR #8871)
   - [x] smoke test(s) (PR #8853)
- [x] Ship 4.3.0 [@hoten]
- [x] Ship 4.3.1 [@hoten]
   - [x] merge 4.3.1 in devtools to beta (https://crbug.com/958952)
- [x] Ship 5.0
   - [x] PSI rls notes and announcement [@exterkamp]
   - [x] Changelog highlighting major (interesting) and breaking changes, screenshots, etc
   - [x] update Chrome Webstore extension screenshots #6112
   - [x] ship 5.0 to devtools
- [x] Lighthouse v5 Breaking Changes #7752 [@brendankenny]
   - [x] Kill `rawValue` #6199 (PRs #8343, #8385, #8421)
   - [x] Remove `disableDeviceEmulation` flag before v5 #7044 (PR #8289)
   - [x] (minor) Inconsistent format of screenshot-thumbnails and final-screenshot #7092 (PR #8299)
   - [x] Switch to a different name for [`LH.Audit.Details.Diagnostic`](https://github.com/GoogleChrome/lighthouse/blob/7a431253ff19397d4b8db9ccdf599e93f7fe40c6/types/audit-details.d.ts#L78). "Diagnostic" is too [overloaded](https://github.com/GoogleChrome/lighthouse/blob/7a431253ff19397d4b8db9ccdf599e93f7fe40c6/lighthouse-core/config/default-config.js#L31) in LH. (PR #8298)
   - [x] Move to minimum Node 10 #7215 (PR #8117)
   - [x] Lantern FCP improvements (PR #7513)
   - [x] Collect only elements in body for `dom-size` (PR #7241)
   - [x] New a11y audit weights (PR #8823)
[...]

PR #8781, opened 2019-05-01
deps: chrome-launcher@0.10.7

paulirish:
#8331


package.json
```diff
@@ -132,7 +132,7 @@
   },
   "dependencies": {
     "axe-core": "3.2.2",
-    "chrome-launcher": "^0.10.5",
+    "chrome-launcher": "^0.10.6",
```
brendankenny: will you also delete the `resolutions` entry down on L164-166? yay!

brendankenny:
is it possible the built js files aren't in 0.10.6? Or something else went wrong? Both CIs are still saying `Cannot find module 'chrome-launcher'`

patrickhulce:
hm yeah something looks wrong there, there's no `dist/` folder

image

patrickhulce:
heh @paulirish we were implicitly re-using the rimraf types from chrome-launcher in lighthouse-viewer :D

want to manually add `@types/rimraf`?

who knew a chrome-launcher bump would be so pesky

brendankenny:
:arrows_counterclockwise::arrows_counterclockwise::arrows_counterclockwise:

PR #8783, opened 2019-05-01
core(lightwallet): add budget to sample artifacts config

brendankenny:
#8427 manually added a budget to the sample artifacts. This PR changes `update-report-fixtures.js` to add a budget to the run so that it will always be in the sample artifacts when it's updated in the future.

Originally discussed in https://github.com/GoogleChrome/lighthouse/pull/8427#discussion_r277925455.

I just copied the budget added in #8427. The property order changes in the `artifacts.json` file because this is the order that `budget.js` initializes things.

### DIFF
--- a/lighthouse-core/scripts/update-report-fixtures.js
+++ b/lighthouse-core/scripts/update-report-fixtures.js
@@ -12,6 +12,25 @@ const {server} = require('../../lighthouse-cli/test/fixtures/static-server.js');
 
 /** @typedef {import('net').AddressInfo} AddressInfo */
 
+/** @type {LH.Config.Json} */
+const budgetedConfig = {
+  extends: 'lighthouse:default',
+  settings: {
+    budgets: [{
+      resourceSizes: [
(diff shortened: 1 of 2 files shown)


lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -31,7 +50,7 @@ async function update() {
     url,
   ].join(' ');
   const flags = cliFlags.getFlags(rawFlags);
-  await cli.runLighthouse(url, flags, undefined);
+  await cli.runLighthouse(url, flags, budgetedConfig);
```
brendankenny: since we don't run `bin.js` (which would load from a `--budget-path`), this could either put a budget on `flags.budgets` or make a config as it does here. The config seemed cleaner, but I'm not attached to the approach.

patrickhulce: config sgtm 👍

brendankenny:
> though not sure I'd call it core.... :)

good point :)

Issue #8647, opened 2019-04-26
How should we handle content changing when scrolling down the page?


mattzeunert:
When a gatherer scroll down the page it can change the page layout, e.g. due lazy loading or infinite scroll loading more content. That might matter because:

- page content varies between gatherers, so whether the page has been scrolled can affect the scores of audits that look at the DOM
- [the new tap targets gatherer](https://github.com/GoogleChrome/lighthouse/pull/7778) first collects elements location and then scrolls through the page to call `elementFromPoint`. If an element moves while scrolling down the page the gatherer will think the tap target is invisible
- if we take a full page screenshot it should match what other gatherers see so we correctly identify elements on it

The accessibility gatherer used to run early on and it scrolled to the bottom of the page. [Moving it to run last](https://github.com/GoogleChrome/lighthouse/pull/8216) had a [minor impact on some sites' scores](https://gist.github.com/mattzeunert/79cb7605e6388c76f4090b7124a64482).

Should we do anything to make this more consistent and predictable?
- can we scroll all the way to the bottom of the page once before running any gatherers?
- maybe after scrolling down once we could get the height of the page and if the page grows again we ignore content beyond that page height? That page height could also be the max height of the full page screenshot.

cc @patrickhulce

patrickhulce:
Discussed this the other day with @mattzeunert (Thanks for filing it!) and I think it would be a great followup to #8625 to scroll to the bottom of the current page height before all the afterPass runs to address most of this.

The ignoring all elements that get created seems a little tricker to enforce on gatherers, so I'm more willing to punt that for now/let individual gatherers handle that on a case-by-case basis.

paulirish:
> - [the new tap targets gatherer](https://github.com/GoogleChrome/lighthouse/pull/7778) first collects elements location and then scrolls through the page to call `elementFromPoint`. If an element moves while scrolling down the page the gatherer will think the tap target is invisible
> - if we take a full page screenshot it should match what other gatherers see so we correctly identify elements on it

Looks like @connorjclark resolved these concerns with the #11536 and #11118 :D   


> - page content varies between gatherers, so whether the page has been scrolled can affect the scores of audits that look at the DOM

this part is still unresolved, but i'm gonna close the issue for now as there's no action plan on the horizon.

PR #8797, opened 2019-05-02
WIP: Element screenshots

### LINKED ISSUE #6683: Improve how HTML elements are displayed in the results table
A few things that could be improved:
- UX: hide irrelevant info, make sure important info is shown (like the text content)
- Reduce result size (by hiding irrelevant info)
- Be more consistent across different audits (different data sources might make this difficult, e.g. expression evaluated in page, `driver.getNodesInDocument`, or aXe)

Here are two example results:

Related ticket about the node selectors we show on hover - might also benefit from some standardization: https://github.com/GoogleChrome/lighthouse/issues/4015

cc @rviscomi @patrickhulce

### LINKED ISSUE #7327: Visual audit reporting
**Feature request summary**

Visually convey why applicable audits are failing by highlighting affected UI in the context of the page, either statically overlaid on a screenshot or interactively on the page itself.

The [VisBug](https://github.com/GoogleChromeLabs/ProjectVisBug) project is already well-positioned to handle the visualization of audit failures so it's worth exploring possible integration points between that and LH. For example, perhaps VisBug could run LH audits under the hood and highlight audit failures based on the JSON LHR. cc @argyleink

**What is the motivation or use case for changing this?**

Many of the audits are inherently visual in nature. They are testing that visible parts of the page behave as expected. When Lighthouse identifies elements that fail the audit, we show their innerHTML or selector to convey what is failing and rely on text to explain why it failed.

By showing the results visually, it's much easier to convey both what is failing and why. For example, the tap targets SEO audit checks for overlapping links/buttons. The results are a table of elements and box sizes. In addition to this easily serializable format, it would be helpful for users to be able to see these elements overlapping on the page so it's not only obvious what the issue is but also how to fix it.

Prototype of a visual report of overlapping tap targets:

image

Contrast that with the text-based approach:

image

Here is a list of audits that are candidates for this visual reporting approach:

- Performance
  - properly size images
  - defer offscreen images
  - efficiently encode images
  - serve images in next-gen formats
  - use video formats for animated content
- Accessibility
  - all text remains visible during webfont loads
  - form elements do not have associated labels
  - image elements do not have [alt] attributes
  - some elements have a [tabindex] value greater than 0
  - links do not have a discernible name
  - buttons have an accessible name
  - background and foreground colors have a sufficient contrast ratio
  - [id] attributes on the page are unique
  - [accesskey] attributes on the page are unique
  - `<input type="image">` elements have [alt] text
  - `<object>` elements have [alt] text
  - `<video>` elements contain a `<track>` element with [kind="captions"]
  - `<video>` elements contain a `<track>` element with [kind="description"]
- Best Practices
  - displays images with correct aspect ratio
- SEO
  - document doesn't use legible font sizes
  - links have descriptive text
  - document doesn't have overlapping tap targets

**How is this beneficial to Lighthouse?**

Visual reporting will make it easier for LH users to make sense of the results and fix issues.

mattzeunert:
**Summary**

Very WIP 🏗

- take full-page screenshot (created a [separate PR for this](https://github.com/GoogleChrome/lighthouse/pull/9064))
- capture element bounding boxes
- render element screenshot in report

Todo:
- [ ] figure out final UX
- [ ] we should scroll down the page once before capturing element positions or the screenshot
   - otherwise page content size may change on scroll, e.g. because of lazy loading images/ads
   - example: footer promo slider at the bottom of the BBC homepage
   - Ticket: https://github.com/GoogleChrome/lighthouse/issues/8647
otherwise page content size may change on scroll (e.g. on the BBC homepage) 
- [x] sort out how chrome devtools instantiates detailsrenderer
    - I think this may actually be resolved now, seems like [Chrome no longer has custom code to initialize the details renderer](https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/audits/lighthouse/report.js?l=3853&rcl=df13cdcb4cd91c79eb63bdd33267b8e67f9f43b3)
- [ ] don't use clippath (see Patrick's note on browser support, also there's no reason we need it)
- [ ] do we need to check if the element is not visible and then not capture the bounding box so that we don't show a screenshot that doesn't show the element (e.g. because it's overflowing a scroll container or there's an overlay on top of it)
- [ ] handle elements that are outside the captured screenshot area (e.g. for pages > 16000px)

Known issues:
- if resizing the window changes the layout then the element marker positions will be off (e.g. this happens for position fixed elements, or if the page uses `vh`)
- elements in `overflow: scroll` containers may be hidden
- elements can be obscured by full-screen overlays

**Related Issues/PRs**

fixes #6683 
fixes #7327


lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -355,15 +355,43 @@ class DetailsRenderer {
...
     if (item.path) element.setAttribute('data-path', item.path);
     if (item.selector) element.setAttribute('data-selector', item.selector);
     if (item.snippet) element.setAttribute('data-snippet', item.snippet);
+    element.innerHTML = `<div style="color: black;font-family: var(--text-font-family);margin-bottom: 10px;">${item.textContent}</div>`;
+    const snippetEl = document.createElement('div');
+    snippetEl.style.fontSize = '12px';
+    snippetEl.innerText = item.snippet;
+    element.appendChild(snippetEl);
+
+    const fullpageScreenshotUrl = __LIGHTHOUSE_JSON__.audits['full-page-screenshot'].details.data;
```
mattzeunert: What's a good way to access this in the renderer?

lighthouse-core/gather/gatherers/full-page-screenshot.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const driver = passContext.driver;
+    const metrics = await driver.sendCommand('Page.getLayoutMetrics');
+    const width = Math.ceil(metrics.contentSize.width);
+    let height = Math.ceil(metrics.contentSize.height);
+
+    height = Math.min(maxScreenshotHeight, height);
+
+    await driver.sendCommand('Emulation.setDeviceMetricsOverride', {
+      // todo: figure out what this means and set appropriately
```
patrickhulce: 😆 I wonder how many devtools protocol settings we should add this comment to 🤔

lighthouse-core/report/html/renderer/details-renderer.js
patrickhulce: I would say it should probably be passed into a `DetailsRenderer` as the context of the report, almost like `setTemplateContext`

lighthouse-core/report/html/renderer/element-screenshot-renderer.js
```diff
@@ -0,0 +1,97 @@
...
+   * @param {ParentNode} templateContext
+   * @param {LH.Audit.Details.NodeValue} item
+   * @param {string} fullpageScreenshotUrl
+   * @return {Element}
+   */
+  static render(dom, templateContext, item, fullpageScreenshotUrl) {
+    const tmpl = dom.cloneTemplate('#tmpl-lh-element-screenshot', templateContext);
+    const previewContainer = dom.find('.lh-element-screenshot', tmpl);
+
+    // todo: support desktop screenshots
```
patrickhulce: ideally just passed around along with the `fullPageScreenshot` artifact so there's no support for specific screen sizes :)

types/artifacts.d.ts
```diff
@@ -127,6 +127,8 @@ declare global {
       TagsBlockingFirstPaint: Artifacts.TagBlockingFirstPaint[];
       /** Information about tap targets including their position and size. */
       TapTargets: Artifacts.TapTarget[];
+      /** Screenshot of the entire page (rather than just the above the fold content). */
+      FullPageScreenshot: {data: string}
```
patrickhulce: dimensions here would be 👌

lighthouse-core/report/html/report-styles.css
```diff
@@ -1366,3 +1366,29 @@
...
+
+/* Element screenshot */
+.lh-element-screenshot {
+  position: relative;
+  pointer-events: none;
+}
+.lh-element-screenshot__content {
+  position: absolute;
+  right: 0px;
+  top: -190px;
```
patrickhulce: this is based on the dynamic height we set it js? anyway to fix it or not have the hanging magic value? :)

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -352,15 +352,43 @@ class DetailsRenderer {
...
+
+    if (item.boundingRect) {
+      /** @type {Element} */
+      let elementScreenshot;
+      element.addEventListener('mouseenter', () => {
```
patrickhulce: not sure what our stance is on JS in report now

@paulirish 🤓 🔫 , CSS-only impl? ;)

lighthouse-core/report/html/renderer/element-screenshot-renderer.js
```diff
@@ -0,0 +1,97 @@
...
+    elMarker.style.height = boundingRect.height + 'px';
+    elMarker.style.left = boundingRect.left + 'px';
+    elMarker.style.top = topOffset + 'px';
+
+    const mask = /** @type {HTMLElement} */
+        (previewContainer.querySelector('.lh-element-screenshot__mask'));
+    const clipId = 'clip-' + Math.floor(Math.random() * 100000000);
+    mask.style.width = previewWidth + 'px';
+    mask.style.height = previewHeight + 'px';
+    mask.style.clipPath = 'url(#' + clipId + ')';
```
patrickhulce: clever! I think the chart is scarier than reality given this narrow usecase but still might be troublesome for us? https://caniuse.com/#search=clipPath

would you mind explaining in words what the approach is here? I think I'm getting it but not sure why the mask is necessary

mattzeunert:
> do you think you could update the artifacts so we can play around with it in the now.sh deployment report on the PR

Done now 👍

lighthouse-core/audits/accessibility/axe-audit.js
```diff
@@ -53,6 +53,8 @@ class AxeAudit extends Audit {
           selector: Array.isArray(node.target) ? node.target.join(' ') : '',
           path: node.path,
           snippet: node.html || node.snippet,
+          textContent: node.textContent,
```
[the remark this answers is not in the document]
mattzeunert: I added the text content to the rendered elements to make them easier to read. If we only show the element screenshots on hover/tap then I think it's worth showing. What do you think?

Do we need the property for that? Yes because the HTML may be truncated, so we can't extract the text from it. But we can limit it to 80 characters or so.

Screenshot 2019-05-15 at 09 48 31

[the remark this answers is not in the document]
mattzeunert: Maybe we should call it `text` and we can also count `alt` or `aria-label` attributes if there's no textContent

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: Passing it into the `DetailsRenderer` constructor now.

lighthouse-core/report/html/renderer/element-screenshot-renderer.js
mattzeunert: There isn't anything that can't be done with divs positioned over the image, will change it.

lighthouse-core/report/html/report-styles.css
mattzeunert: It's so that the screenshot doesn't cover the table item you're hovering over and instead appears above it. Not really based on anything though.

Do you think showing the screenshot above the table item on hover is the way to go? And then on mobile we could expand the table item to include the screenshot when the user taps on it.

lighthouse-core/report/html/renderer/element-screenshot-renderer.js
mattzeunert: Added support for desktop screenshots now, i.e. screenshots that are wider than the element screenshot display area.

Right now if there's a big element highlighted on the screenshot it'll just be cut off with `overflow: hidden`. Maybe in the future we could address that and maybe zoom out a bit? Probably not something that needs to be part of v1.

lighthouse-core/report/html/renderer/details-renderer.js
mattzeunert: I can't find the comment, but I think we decided that's ok on the snippet renderer PR.

Not sure how feasible this would be to do without JS, browsers might not like having a hundred 500kb base64 images on one page.

patrickhulce: Oh that sounds right. I was remembering some I/O convo about going back and forth to make things easier on PSI, but we did already break that barrier.

lighthouse-core/audits/accessibility/axe-audit.js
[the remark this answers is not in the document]
patrickhulce: Hm, might be difficult if we're specifically flagging elements that are using `alt`/`aria-label` incorrectly 😆 

how about we stick to textContent for now but limit to 80 characters as you suggest? 👍

lighthouse-core/report/html/renderer/element-screenshot-renderer.js
patrickhulce: Gotcha yeah let's punt on that for v1 :)

patrickhulce:
Nice thanks for deploying it! The experience is A-MAZ-ING 😍 


Few findings:

- Small screen sizes it gets a little weird that it goes full screen and completely covers the node, maybe we scale it down a bit so we can shift it to the side?
image

- This is the sort of thing I was worried about with textContent, maybe we actually want innerText or perhaps the 80 character limit is enough?
image

lighthouse-core/audits/accessibility/axe-audit.js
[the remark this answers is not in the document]
mattzeunert: I think using `alt/aria-label` is quite helpful, e.g. for buttons that just consist of an image.

Created a PR here: https://github.com/GoogleChrome/lighthouse/pull/8961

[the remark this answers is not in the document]
patrickhulce: Oh if you've already done it then more power too ya! 👍 ❤️

mattzeunert:
> This is the sort of thing I was worried about with textContent, maybe we actually want innerText or perhaps the 80 character limit is enough?

Addressed in [this PR](https://github.com/GoogleChrome/lighthouse/pull/8961).

> It wasn't able to get the screenshot for these two links here, not sure whats going on with that

They have a visible size of 0x0px – do you think we should leave some kind of indication if there's no screenshot?

> Sometimes the full-size preview isn't really what we want I think, we might actually prefer context for where it is on the page? Example where we have a ton of duplicate lighthouse images on the page so I still don't know which one this is talking about.

Maybe if the element is taller than 60px we try to show where on the page it is and zoom out by 2x or 4x, and for smaller ones we do something similar to what we have now?

> Small screen sizes it gets a little weird that it goes full screen and completely covers the node, maybe we scale it down a bit so we can shift it to the side?

I changed the CSS so that the hover always appears above the table item. Let's see how the zoomed out stuff from above works and see if it can help with this too.

What do you mean by "shift it to the side"?

> Another small screen size failure case where the important part of the screenshot is outside the window.

Will do this:
- if there's a hover contain it within the window
- if the page is narrower than say 800px users have to tap the line with the element

mattzeunert:
@patrickhulce I just introduced zooming out for large elements, and it's a big improvement!

Played around with always zooming out, but for e.g. small tap targets the immediate environment is more important than the page context.

The screenshot positioning feels a bit awkward to me sometimes, but not sure if there's a better approach. Maybe we could always put them in a corner of the screen?

Also, here are two screenshots from a desktop report:



Screenshot 2019-05-16 at 16 55 37
Screenshot 2019-05-16 at 16 55 41

patrickhulce:
Woah yes zooming out makes a massive difference, great work!!

> do you think we should leave some kind of indication if there's no screenshot?

Yeah I think that'd be a good call 👍 

> What do you mean by "shift it to the side"?

I just meant always make sure we can see a portion of the table item, i.e. let's move the screenshot out of the way a little bit.

> Maybe we could always put them in a corner of the screen?

This is an interesting idea, I like it! A consistent place for the screenshots to appear might be nice.

mattzeunert:
> > Maybe we could always put them in a corner of the screen?
> 
> This is an interesting idea, I like it! A consistent place for the screenshots to appear might be nice.
 
Tried that now, and it's really nice on a big screen (except for the flicker when hovering over the td padding).
 
I was originally thinking we'd put the screenshot inside the td on mobile, but actually putting it outside seems better because we don't have to shrink it down, e.g. in the 3-column tap targets table.
 
On small and medium size screens the screenshot ends up overlapping the main content. Maybe we should show the image at the top of the screen if it overlaps? I thought about trying it out but then realized how much of a pain it is to get the location of the `td` in the screenshot renderer 🙂

---

Do you think we should try to re-use the `final-screenshot` instead of duplicating the top of the page in `full-page-screenshot`?

- pro: we'd save some JSON size
- pro: part of the full-page screenshot would be higher quality
- con: it adds complexity
- con: visible change in jpeg quality in element screenshots

My feeling is "no".

I assume `full-page-screenshot` can't be used as `final-screenshot` since its quality is too low and it's too big.

patrickhulce:
> My feeling is "no".

👍

Maybe in the future we can explore that but separate for now seems fine to me.

mattzeunert:
@patrickhulce Been thinking of creating a PR for the full-page screenshot artifact/audit, since I think we know what we want there. Any reason we'd want to hold off on that, or do you think I should go ahead?

patrickhulce:
Sounds like a great place to start to me @mattzeunert! The only hesitation I can see would be if we don't want the immediate byte impact in the LHR if nothing is using it yet, but maybe we can leave it ungrouped or something?

connorjclark:
hey @mattzeunert we're wanting to land this (mostly for #10517) and willing to invest eng time to get it the rest of the way. mind if we take it from here, or would you be able to pick this back up?

I'll go through the PR tomorrow and leave my thoughts.

mattzeunert:
Hey @connorjclark! Having someone on the team pick it up sounds good 👍

connorjclark:
I forked and started work here: https://github.com/GoogleChrome/lighthouse/compare/elscreens

PR #8802, opened 2019-05-02
tests: partial artifact update

mattzeunert:
**Summary**

Make sample artifacts easier to update by specifying only one artifact that should be updated, like this:

`yarn run update:sample-artifacts ScriptElements`


lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -43,15 +46,32 @@ async function update() {
...
   await new Promise(res => server.close(res));
+
+  if (artifactName) {
+    // Revert everything except the one artifact
+    const newArtifacts = readArtifacts();
+    const finalArtifacts = oldArtifacts;
+    finalArtifacts[artifactName] = newArtifacts[artifactName];
+    fs.writeFileSync(
+      artifactPath + '/artifacts.json',
+      JSON.stringify(finalArtifacts, null, 2) + '\n'
```
brendankenny: would it be crazy overkill to use `assetSaver.saveArtifacts(finalArtifacts, artifactPath)` for this? 

`JSON.stringify(artifacts, null, 2)` is the same exact thing we do there, but it's possible that will change or the default file name will change, etc, and we'd get those changes for free here.

Only awkwardness is this would need to set `finalArtifacts.traces = {};` and `finalArtifacts.devtoolsLogs = {};` to keep `saveArtifacts` happy.

CONTRIBUTING.md
```diff
@@ -73,6 +73,14 @@ Don't:
 If no reference doc exists yet, then you can use the `description` as a stopgap for explaining
 both why the audit is important and how to fix it.
 
+## Updating sample artifacts and LHR JSON
+
+```
+yarn run update:sample-artifacts # update all artifacts
+yarn run update:sample-artifacts ScriptElements # update just one artifact
```
brendankenny: should we note that they'll need to revert the devtoolslog and trace files if they want only this artifact?

lighthouse-core/scripts/update-report-fixtures.js
brendankenny: > would it be crazy overkill to use `assetSaver.saveArtifacts(finalArtifacts, artifactPath)` for this?

oh! We should definitely do this, but we should also do 
`const oldArtifacts = assetSaver.loadArtifacts(artifactPath);`
and
`const newArtifacts = assetSaver.loadArtifacts(artifactPath);`

That would pull in all the devtoolsLogs and traces into `oldArtifacts`, which would then be rewritten to the same files on `assetSaver.saveArtifacts(finalArtifacts, artifactPath)`, so no need to manually revert the traces and whatnot.

(and if, for some unknown reason someone wants to update the sample traces, `	yarn run update:sample-artifacts traces` will do the trick)

All this is untested, but hopefully just works :)

mattzeunert: > All this is untested, but hopefully just works :)

It does! Thanks, this is much better.

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -32,9 +34,10 @@ const budgetedConfig = {
 };
 
 /**
- * Update the report artifacts
+ * Update the report artifacts. If artifactName is set only that artifact will be updated.
+ * @param {keyof LH.Artifacts} artifactName?
```
brendankenny:
```suggestion
 * @param {keyof LH.Artifacts=} artifactName
```
or 
```suggestion
 * @param {keyof LH.Artifacts} [artifactName]
```

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -7,6 +7,8 @@
 
 const cli = require('../../lighthouse-cli/run.js');
 const cliFlags = require('../../lighthouse-cli/cli-flags.js');
+const assetSaver = require('../lib/asset-saver');
```
brendankenny:
```suggestion
const assetSaver = require('../lib/asset-saver.js');
```

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -43,15 +46,25 @@ async function update() {
...
+    // Revert everything except the one artifact
+    const newArtifacts = await assetSaver.loadArtifacts(artifactPath);
+    const finalArtifacts = oldArtifacts;
+    finalArtifacts[artifactName] = newArtifacts[artifactName];
+    await assetSaver.saveArtifacts(finalArtifacts, artifactPath);
+  }
 }
 
-update();
+update(/** @type {keyof LH.Artifacts} */ (process.argv[2]));
```
brendankenny: technically need an `|undefined`

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -43,15 +46,25 @@ async function update() {
...
   ].join(' ');
   const flags = cliFlags.getFlags(rawFlags);
   await cli.runLighthouse(url, flags, budgetedConfig);
   await new Promise(res => server.close(res));
+
+  if (artifactName) {
+    // Revert everything except the one artifact
+    const newArtifacts = await assetSaver.loadArtifacts(artifactPath);
+    const finalArtifacts = oldArtifacts;
+    finalArtifacts[artifactName] = newArtifacts[artifactName];
```
brendankenny: do we want to log a warning if `artifactName` isn't a key on `newArtifacts` or `oldArtifacts` (need to check both in case it's adding a new one or removing an old one?)

At worst an unknown key will just have no effect and we don't want to throw because it'll leave things in a bad state, but it might be good to let the user know they spelled something wrong or whatever

CONTRIBUTING.md
mattzeunert: Added, feel free to reword.

CONTRIBUTING.md
```diff
@@ -73,6 +73,16 @@ Don't:
...
+yarn run update:sample-json # update sample LHR based on sample artifacts
+```
+
+Usually you'll need to revert changes to the `*.devtoolslog.json` and `*.trace.json` files and manually review changes to `artifacts.json` to make sure they are related to your work.
```
brendankenny: maybe it's unnecessary now? Or 
```suggestion
If you do a full artifacts update, usually you'll need to revert changes to the `*.devtoolslog.json` and `*.trace.json` files and manually review changes to `artifacts.json` to make sure they are related to your work.
```
I don't have a strong opinion on keeping/wording, what do you think would have been helpful for you?

lighthouse-core/scripts/update-report-fixtures.js
mattzeunert: Can we have a lint rule for this? And update all the places where we don't use `.js` right now?

lighthouse-core/scripts/update-report-fixtures.js
mattzeunert: Will throw now if it can't find the artifact name.

CONTRIBUTING.md
mattzeunert: You have to revert those files even if you just do just a partial update, because we just revert the artifacts file.

> what do you think would have been helpful for you?

Just having some docs, and knowing that having lots of changes you need to revert is normal.

lighthouse-core/scripts/update-report-fixtures.js
brendankenny: > Can we have a lint rule for this? 

I'd love that. I couldn't find a good one last time I looked, but I didn't try very hard to find one.

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -7,6 +7,8 @@
 
 const cli = require('../../lighthouse-cli/run.js');
 const cliFlags = require('../../lighthouse-cli/cli-flags.js');
+const assetSaver = require('../lib/asset-saver.js');
+const artifactPath = 'lighthouse-core/test/results/artifacts';
```
paulirish:
```suggestion

const artifactPath = 'lighthouse-core/test/results/artifacts';
```

just add some whitespace after our requires

CONTRIBUTING.md
brendankenny: > You have to revert those files even if you just do just a partial update, because we just revert the artifacts file.

that's one nice side effect of using `loadArtifacts`/`saveArtifacts`, they [automatically handle collecting and resaving](https://github.com/GoogleChrome/lighthouse/blob/03e6566fc999befc3e12ca19a3f1bc4b87e8f2d7/lighthouse-core/lib/asset-saver.js#L51-L66) the devtoolsLog and trace files too :)

CONTRIBUTING.md
```diff
@@ -73,6 +73,16 @@ Don't:
...
 
+## Updating sample artifacts and LHR JSON
+
+```
+yarn run update:sample-artifacts # update all artifacts
+yarn run update:sample-artifacts ScriptElements # update just one artifact
+yarn run update:sample-json # update sample LHR based on sample artifacts
+```
+
+When updating artifacts, usually you'll need to revert changes to the `*.devtoolslog.json` and `*.trace.json` files and manually review changes to `artifacts.json` to make sure they are related to your work.
```
brendankenny:
```suggestion
When updating all artifacts, usually you'll need to revert changes to the `*.devtoolslog.json` and `*.trace.json` files and manually review changes to `artifacts.json` to make sure they are related to your work.
```

lighthouse-core/scripts/update-report-fixtures.js
```diff
@@ -43,15 +47,28 @@ async function update() {
...
   ].join(' ');
   const flags = cliFlags.getFlags(rawFlags);
   await cli.runLighthouse(url, flags, budgetedConfig);
   await new Promise(res => server.close(res));
+
+  if (artifactName) {
+    // Revert everything except the one artifact
+    const newArtifacts = await assetSaver.loadArtifacts(artifactPath);
+    if (!(artifactName in newArtifacts) && !(artifactName in oldArtifacts)) {
+      throw Error('Unknown artifact name: ' + artifactName);
```
brendankenny:
```suggestion
      console.warn(`❌ Unknown artifact name: '${artifactName}'. Reverting artifacts...`); // eslint-disable-line no-console
```
throwing ends up as an unhandled promise rejection (since the function is async) and leaves the artifacts as the newly generated ones. With this it logs and continues. Since the `artifactName` is unknown, it's just setting the property to `undefined` which doesn't change the JSON output and so the original artifacts are restored.

PR #8803, opened 2019-05-02
tests: add tap targets to dobetterweb sample page

mattzeunert:
**Summary**

Include tap targets in the sample artifacts, so they are part of the tests and the now.sh preview.


lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
```diff
@@ -182,6 +182,16 @@ <h2>Do better web tester page</h2>
 <!-- FAIL(efficient-animated-content): animated gif found -->
 <img src="lighthouse-rotating.gif" width="811" height="462">
 
+<!-- FAIL(tap-targts): buttons too close together -->
```
brendankenny: this is a little weird because usually the PASS/FAIL markers are indicators for the specific smoke test that it's a pass/fail. `yarn smoke dbw` doesn't run `tap-targets`, so it wouldn't be an audit failure.

Buuuuuut, we've been using these test files in multiple tests for a while now (sometimes multiple smoke configs, and this one for `sample_v2.json`), so maybe it's fine you just have to cross reference the audit id with the config that's running the smoke test?

Hoten:
```suggestion
<!-- FAIL(tap-targets): buttons too close together -->
```

brendankenny:
might need to rebase this one? Pull `sample_v2_round_trip.json` and rerun the update:sample-json? That change was brutal for conflicts for some reason

lighthouse-cli/test/fixtures/dobetterweb/dbw_tester.html
mattzeunert: Oh I see what you mean. Maybe the ultimate solution would be to separate the dbw smoke test from the sample page?

PR #8805, opened 2019-05-02
 tests: convert smoke metric assertions to numericValue

patrickhulce:
**Summary**
Converts some smoke metric assertions to assert on numericValue instead of just score. Some folks find this more readable as to the intention of the smoke test (https://github.com/GoogleChrome/lighthouse/pull/8379#issuecomment-488563344), others feel that there's no harm in including both (https://github.com/GoogleChrome/lighthouse/pull/8805#discussion_r280561086).

### DIFF
--- a/lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
+++ b/lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
@@ -33,7 +33,8 @@ module.exports = [
       audits: {
         'interactive': {
           // Make sure all of the CPU time is reflected in the perf metrics as well.
-          score: '<.2',
+          // The scripts stalls for 3 seconds and lantern has a 4x multiplier so 12s minimum.
+          numericValue: '>12000',
         },
         'bootup-time': {
           details: {
@@ -56,7 +57,8 @@ module.exports = [
       audits: {
         'interactive': {
           // Make sure all of the CPU time is reflected in the perf metrics as well.
-          score: '<.2',
+          // The scripts stalls for 3 seconds and lantern has a 4x multiplier so 12s minimum.
+          numericValue: '>12000',
         },
         'bootup-time': {
           details: {
--- a/lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
+++ b/lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
@@ -15,11 +15,11 @@ module.exports = [
       finalUrl: 'http://localhost:10200/tricky-tti.html',
       audits: {
         'first-cpu-idle': {
-          score: '<75',
+          // stalls for 5 seconds, 5 seconds out, so should be around 10s
           numericValue: '>9000',
         },
         'interactive': {
-          score: '<75',
+          // stalls for 5 seconds, 5 seconds out, so should be around 10s
           numericValue: '>9000',
         },
       },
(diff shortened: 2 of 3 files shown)


lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
```diff
@@ -15,11 +15,11 @@ module.exports = [
       finalUrl: 'http://localhost:10200/tricky-tti.html',
       audits: {
         'first-cpu-idle': {
-          score: '<75',
```
patrickhulce: this was just wrong from a time long past :)

Hoten:
thanks!

lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
```diff
@@ -33,7 +33,8 @@ module.exports = [
       audits: {
         'interactive': {
           // Make sure all of the CPU time is reflected in the perf metrics as well.
-          score: '<.2',
+          // The scripts stalls for 3 seconds and lantern has a 4x multiplier so 12s minimum.
+          numericValue: '>12000',
```
brendankenny: is the value just that these are more informative? Because from that screenshot in the other issue it seems like LR is going to be just as wrong with these numbers as the score numbers :)

lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
```diff
@@ -15,11 +15,11 @@ module.exports = [
...
+          // stalls for 5 seconds, 5 seconds out, so should be around 10s
```
brendankenny: is there any harm in having `score` and `numericValue`? It's kind of nice to have a `first-cpu-idle` score curve assertion, even if it's only a single (very fuzzy) data point

lighthouse-cli/test/smokehouse/perf/lantern-expectations.js
Hoten: yes, just more informative.

lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
patrickhulce: sure, fine with me. `<75` is definitely useless though 😆

brendankenny:
> can you add a clearer PR description?

bump :P

lighthouse-cli/test/smokehouse/tricky-metrics/expectations.js
brendankenny: > sure, fine with me. <75 is definitely useless though 😆

lol, didn't notice the lack of a decimal point

...so do you want to do the score as well? I can't tell :)

We could tighten them up if we're this confident about the numericValues:
`'>9000'` for first-cpu-idle means a score of `'<.25'`
`'>9000'` for interactive means a score of `'<.35'`
`'>12000'` for interactive means a score of `'<.17'`

patrickhulce: > ...so do you want to do the score as well?

personally, I don't. I think it creates extra noise when the point of the tests was to assert on a particular metric value and not specifics of our scoring curve. I concede it makes sense to have at least a few score assertions to make sure we're not going crazy there, but IMO the unit tests are good enough there and we don't need to copy it everywhere.

I don't feel strongly enough to fight it on what was supposed to be a quick feedback followup PR though so 😆  🤷‍♂

brendankenny: haha, fair enough

PR #8818, opened 2019-05-02
misc(proto): remove trailing whitespace; add contributing messages

brendankenny:
kind of a follow up to #8815

This
- adds a note to our readme that if you're going to be touching audits you'll likely need protobuf installed (so that `yarn update:sample-json` can run successfully)
- adds a more explicit error message if `yarn update:sample-json` fails on the proto step to install protobuf
- trims trailing whitespace from `proto/sample_v2_round_trip.json`. We've had a decent amount of churn from PRs that have needed to manually edit the file and have had whitespace automatically (or manually) trimmed. Hopefully this reduces future noise

### DIFF
--- a/proto/scripts/json_roundtrip_via_proto.py
+++ b/proto/scripts/json_roundtrip_via_proto.py
@@ -40,4 +40,4 @@ round_trip_lhr = json.loads(MessageToJson(proto_lhr, including_default_value_fie
 
 # write the json to disk
 with open(path_round_trip, 'w') as f:
-    json.dump(round_trip_lhr, f, indent=4, sort_keys=True)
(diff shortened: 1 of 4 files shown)


proto/scripts/json_roundtrip_via_proto.py
```diff
@@ -40,4 +40,4 @@ def clean():
 
 # write the json to disk
 with open(path_round_trip, 'w') as f:
-    json.dump(round_trip_lhr, f, indent=4, sort_keys=True)
+    json.dump(round_trip_lhr, f, indent=4, sort_keys=True, separators=(',', ': '))
```
patrickhulce: 🎉

PR #8819, opened 2019-05-02
client(devtools): update version in tests on lh roll

Hoten:
b/c I keep forgetting to do it myself.


brendankenny:
@Hoten what do you want to do with the feedback here?

PR #6833, opened 2018-12-18
report: fix audits DOM style

tgfjt:
Hi! 😃 

## Summary
This PR fixes: some texts protruded from the box of runtime-settings.

screenshot: 

image

## Change

Add `word-break: break-all` into `.lh-env__description` .

screenshot: 
image

example URL: `https://www.amazon.co.jp/%E3%83%AF%E3%83%B3%E3%82%B9%E3%83%BB%E3%82%A2%E3%83%9D%E3%83%B3%E3%83%BB%E3%82%A2%E3%83%B3%E3%83%BB%E3%82%A2%E3%83%AB%E3%82%B4%E3%83%AA%E3%82%BA%E3%83%A0-%E7%89%A9%E8%AA%9E%E3%81%A7%E8%AA%AD%E3%81%BF%E8%A7%A3%E3%81%8F%E8%A8%88%E7%AE%97-Martin-Erwig/dp/4320124413/ref=sr_1_6?ie=UTF8&qid=1546476091&sr=8-6&keywords=martin+erwig`

PR #8820, opened 2019-05-02
report: improve text wrapping

paulirish:
avoiding stuff like this:
image

now with this PR: 
image

ref #6833 #5138 ##8192


lighthouse-core/report/html/report-styles.css
```diff
@@ -1190,18 +1206,13 @@
   width: 12%;
 }
 
-.lh-text__url {
-  overflow: hidden;
-  text-overflow: ellipsis;
```
Hoten: yayyyyyyy

paulirish:
@yuinchien does this work for you?

PR #8822, opened 2019-05-03
report: make v5 renderer compatible with v3 and v4 LHRs

### LINKED ISSUE #8598: ensure 4.x LHR and 5.0 report renderer compatibility

brendankenny:
fixes #8598

In addition to the `diagnostic` -> `debug` update that needed to be done, #8299 also meant we needed to update the <5.0 filmstrip screenshots to be able to render in the report.

Tested LHRs back to 3.1, everything is 👍 .

*However*: turns out even the 3.1 viewer isn't compatible with 3.0 LHRs because of the i18n switch. Not sure what we want to do about that, though at least 5.0 isn't a regression on this front :)

metrics screenshot

https://googlechrome.github.io/lighthouse/viewer/?gist=86d005fa4297bdefacd01b7228ab47fa

### DIFF
--- a/lighthouse-core/report/html/renderer/details-renderer.js
+++ b/lighthouse-core/report/html/renderer/details-renderer.js
@@ -61,9 +61,6 @@ class DetailsRenderer {
       case 'screenshot':
       case 'debugdata':
         return null;
-      // Fallback for old LHRs, where no type meant don't render.
-      case undefined:
-        return null;
 
       default: {
         // @ts-ignore tsc thinks this unreachable, but ts-ignore for error message just in case.
--- a/lighthouse-core/report/html/renderer/util.js
+++ b/lighthouse-core/report/html/renderer/util.js
@@ -21,6 +21,7 @@
 const ELLIPSIS = '\u2026';
 const NBSP = '\xa0';
 const PASS_THRESHOLD = 0.9;
+const SCREENSHOT_PREFIX = 'data:image/jpeg;base64,';
 
 const RATINGS = {
   PASS: {label: 'pass', minScore: PASS_THRESHOLD},
@@ -48,7 +49,8 @@ class Util {
 
   /**
    * Returns a new LHR that's reshaped for slightly better ergonomics within the report rendereer.
-   * Also, sets up the localized UI strings used within renderer and number/date formatting
+   * Also, sets up the localized UI strings used within renderer and makes changes to old LHRs to be
+   * compatible with current renderer.
    * The LHR passed in is not mutated.
    * TODO(team): we all agree the LHR shape change is technical debt we should fix
    * @param {LH.Result} result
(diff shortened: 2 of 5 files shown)


lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -61,9 +61,6 @@ class DetailsRenderer {
       case 'screenshot':
       case 'debugdata':
         return null;
-      // Fallback for old LHRs, where no type meant don't render.
-      case undefined:
-        return null;
```
brendankenny: moved into `util.prepareReportResult()` so all the backwards compat stuff is in one place

lighthouse-core/report/html/renderer/util.js
```diff
@@ -62,24 +64,45 @@ class Util {
     if (!clone.configSettings.locale) {
       clone.configSettings.locale = 'en';
     }
-    Util.setNumberDateLocale(clone.configSettings.locale);
```
brendankenny: these are just rearranged so backward compatibility stuff, report locale setting, and smooshing aren't all interleaved

paulirish:
> because of the i18n switch

i've been seeing these but wasn't sure the source issue. what switch you talkin about?

brendankenny:
> i've been seeing these but wasn't sure the source issue. what switch you talkin about?

switching from our old displayValue string replacement to the i18n ICU message version, I believe

brendankenny:
> > i've been seeing these but wasn't sure the source issue. what switch you talkin about?

> switching from our old displayValue string replacement to the i18n ICU message version, I believe

actually, I guess this is at least a limited issue for LHRs up until 4.0 when #6767 landed and finished off the old `displayValue` format (`font-size` and `load-fast-enough-for-pwa` were the stragglers).

It's just 3.1 where the perf section moved to the modern i18n system and so the above the fold content is correct from then on.

We could add the old `displayValue` code to `util.prepareReportResult` and have it convert any `displayValue` of the old format it finds, but I vote not worth it until (maybe) later since this won't be a new problem coming from 5.0 :)

Issue #3444, opened 2017-09-29
Rethink accessibility scoring


robdodson:
Currently the accessibility tests are all equally weighted. This means even if a test is not applicable it still counts as a pass and artificially inflates the score. As a result, pages in the [WCAG's "bad" section](https://www.w3.org/WAI/demos/bad/) still get a score of >89%.

One suggestion is to score inapplicable tests at a weight of 0, however based on [this issue](https://github.com/dequelabs/axe-core/issues/473) I'm not sure if aXe actually returns inapplicable tests.

Another thing we should definitely do in the near term is re-weight the tests. aXe itself has criticality ratings, and we can look at httparchive stats to figure out which tests fail most often and maybe boost their weight even more.

Ultimately though, an automated tool like aXe can only ever test a subset of accessibility issues, so giving someone a score of 100% can be misleading. It's entirely possible to make something that gets a 100% accessibility score but is still not very useable. For this reason we might consider ditching the accessibility score altogether and replacing it with something else. Maybe just an indicator that there are (minor|major|critical) errors? Open to suggestion here :)

@marcysutton @WilcoFiers

robdodson:
Forgot to mention, I think a tool like Tenon.io also weighs the score based on page complexity. So if there are 1000 DOM nodes and 1 failing test, it's considered less sever than if there are 100 DOM nodes and a failing test.

jnurthen:
I think having a % based score is a bad idea - 100% implies that the job is done. Perhaps consider reversing it so that the goal is to get to 0 errors. In my mind having 0 errors doesn't have the same implications as reaching 100%

rpkoller:
@jnurthen maybe in a vein like HTML_Sniffer displays results. There you get errors, warnings and notices for the page. You might spice things up with some sort of color coding. Red background color for too many errors on a page with few DOM nodes or a certain amount of errors for an average amount of DOM nodes and so forth like @robdodson suggested . Yellow background for better results and green background for desired results.

paulirish:
>  an automated tool like aXe can only ever test a subset of accessibility issues

True, but this is basically the case for anything.  For performance or best practices or PWA or anything, we'll always be looking at a subset of all worthwhile issues. So I think that just places the onus on the tool developers to be as comprehensive as possible. 

>  re-weight the tests. 

SGTM. I'm totally on board with looking at weighting things differently, even if it includes aspects like DOM node count.

marcysutton:
[...]

PR #8823, opened 2019-05-03
core(a11y): update scoring weights based on severity

robdodson:
**Summary**

Re-weight a11y audits using the same severity scale that axe uses for their docs.

Items which really should be critical failures were weighted too low.

https://dequeuniversity.com/rules/axe/3.2/

**Related Issues/PRs**

#3444


paulirish:
can you leave a comment somewhere that indicates where we can pull these severity weights from? i feel like it is in the axe response somewhere..

robdodson:
Yeah so I'm not 100% happy with this scoring, though it is based on how axe
ranks items from minor, serious, critical.

Downranking something because it is uncommon in hindsight feels wrong
because it means getting an Aria role wrong--which is critically bad--is
considered low impact in current lighthouse.

Another option would be to boost the Aria scores. That's actually what got
me started on this idea.

*From: *Patrick Hulce <notifications@github.com>
*Date: *Thu, May 2, 2019, 6:48 PM
*To: *GoogleChrome/lighthouse
*Cc: *Rob Dodson, Author

*@patrickhulce* commented on this pull request.

robdodson:
OK I staggered things a bit more so minor is 1, serious is 3, and critical is 10. In testing this seems to produce a result that feels about right.

This change is tricky because I think a lot of sites will see their scores go up but that's because we were penalizing them for issues that were common but not necessarily critical failures. On the flip side, there were critical failures that we were severely undercounting because they were uncommon.

For example, having a button without a label—if it's the only failure on the page—is currently worth ~21 points. But misspelling the `button` role on a `div` is only worth about 8 points. Without a proper role the div is not announced as a button to the user, so it feels like it's just as bad of a failure, but it's undercounted.

In the new scheme they are both worth something like 10 or 11 points.

Things which are common, but not necessarily show stoppers, like color contrast are downranked from critical to serious. A _lot_ of sites will benefit from this, but it really didn't make sense to say it was "more severe" than having uncaptioned video or audio. Those are legitimate barriers to access, even if they are less common.

One thing that does make me feel better is the [W3C bad accessibility example pages](https://www.w3.org/WAI/demos/bad/) all score lower in the new scheme than in current Lighthouse. That's because they have more legitimate barriers to access so they get penalized more.

> can you leave a comment somewhere that indicates where we can pull these severity weights from?

@paulirish in the json or here on github?
The ranking comes from deque's scoring system. If you click on an audit listed here: https://dequeuniversity.com/rules/axe/3.2/ there is a column on the right hand side that shows severity. Everything falls into either minor, serious, or critical.

brendankenny:
> proto/sample_v2_round_trip.json

sorry, I changed some stuff. You may need to `git checkout master -- proto/sample_v2_round_trip.json` and then run `yarn update:sample-json` again

robdodson:
@brendankenny I have done the thing.

paulirish:
> @paulirish in the json or here on github?

in the default-config.js right at the top of the auditRefs block.

robdodson:
Added

lighthouse-core/config/default-config.js
```diff
@@ -79,7 +79,8 @@ const UIStrings = {
   seoCategoryManualDescription: 'Run these additional validators on your site to check additional SEO best practices.',
   /* Title of the navigation section within the Search Engine Optimization (SEO) category. Within this section are audits with descriptive titles that highlight opportunities to make a page more usable on mobile devices. */
   seoMobileGroupTitle: 'Mobile Friendly',
-  /* Description of the navigation section within the Search Engine Optimization (SEO) category. Within this section are audits with descriptive titles that highlight opportunities to make a page more usable on mobile devices. */
+  /* Description of the navigation section within the Search Engine Optimization (SEO) category. Within this section are
```
brendankenny: errant hit of the return key? :) this is breaking the strings test (I think maybe we aren't set up to extract multi-line comments into message descriptions) and linting (no trailing spaces)

brendankenny:
thanks for getting this in so fast, @robdodson!

robdodson:
sure thing! Thanks for being so accommodating y'all

PR #8827, opened 2019-05-03
clients(psi): pass score scale to psi

exterkamp:
**Summary**
The new PSI report needs a score scale passed to it! :smile:

### DIFF
--- a/lighthouse-core/report/html/renderer/psi.js
+++ b/lighthouse-core/report/html/renderer/psi.js
@@ -64,7 +64,9 @@ function prepareLabData(LHResult, document) {
   scoreGaugeWrapperEl.removeAttribute('href');
 
   const finalScreenshotDataUri = _getFinalScreenshot(perfCategory);
-  const scoreScaleEl = dom.cloneTemplate('#tmpl-lh-scorescale', dom.document()).firstElementChild;
+
+  const clonedScoreTemplate = dom.cloneTemplate('#tmpl-lh-scorescale', dom.document());
+  const scoreScaleEl = dom.find('.lh-scorescale', clonedScoreTemplate);
 
   return {scoreGaugeEl, perfCategoryEl, finalScreenshotDataUri, scoreScaleEl};
 }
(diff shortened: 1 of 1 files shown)


lighthouse-core/report/html/renderer/psi.js
```diff
@@ -64,7 +64,9 @@ function prepareLabData(LHResult, document) {
   scoreGaugeWrapperEl.removeAttribute('href');
 
   const finalScreenshotDataUri = _getFinalScreenshot(perfCategory);
-  return {scoreGaugeEl, perfCategoryEl, finalScreenshotDataUri};
+  const scoreScaleEl = dom.cloneTemplate('#tmpl-lh-scorescale', dom.document()).firstElementChild;
```
brendankenny: may not really be a thing, but maybe better to do a `const scoreScaleEl = dom.find('.lh-scorescale', clonedScoreTemplate);`?

More resistant to the unlikely event that an element is added before the `div.lh-scorescale` in the template, but more importantly will throw if the template changes significantly and that class name changes since `dom.find` will throw while `firstElementChild` will happily return whatever is there instead.

exterkamp: Yeah good idea.  FirstElementChild was a hack, this seems much safer!

lighthouse-core/report/html/renderer/psi.js
```diff
@@ -31,7 +31,7 @@
  *
  * @param {LH.Result | string} LHResult The stringified version of {LH.Result}
  * @param {Document} document The host page's window.document
- * @return {{scoreGaugeEl: Element, perfCategoryEl: Element, finalScreenshotDataUri: string|null}}
+ * @return {{scoreGaugeEl: Element, perfCategoryEl: Element, finalScreenshotDataUri: string|null, scoreScaleEl: Element|null}}
```
brendankenny:
```suggestion
 * @return {{scoreGaugeEl: Element, perfCategoryEl: Element, finalScreenshotDataUri: string|null, scoreScaleEl: Element}}
```
cause `dom.find` will always find it

PR #8840, opened 2019-05-03
report(redesign): tooltip for category description

connorjclark:
#8185

@yuinchien


lighthouse-core/report/html/templates.html
```diff
@@ -45,8 +45,13 @@
 <!-- Lighthouse category header -->
 <template id="tmpl-lh-category-header">
   <div class="lh-category-header">
-    <div class="lh-score__gauge"></div>
-    <div class="lh-category-header__description"></div>
+    <div class="lh-score__gauge">
+      <!-- Will be added to .lh-gauge__label via renderCategoryHeader -->
+      <!-- <span class="tooltip-boundary">
```
Hoten: should i make this a template

paulirish: yup, i think so.  include in this one and .remove() if there's no description?

lighthouse-core/report/html/report-styles.css
```diff
@@ -61,8 +61,9 @@
   --navitem-hpadding: var(--body-font-size);
   --navitem-vpadding: calc(var(--navitem-line-height) / 2);
   --lh-score-highlight-bg: hsla(0, 0%, 90%, 0.2);
-  --lh-score-icon-background-size: 24px;
   --lh-group-icon-background-size: var(--lh-score-icon-background-size);
+  --lh-info-icon-background-size: 20px;
+  --lh-score-icon-background-size: 24px;
```
paulirish: just -lh-score-icon-size ?

Hoten: btw ur commenting on something i just reordered

i added `--lh-info-icon-background-size`

could rename stuff after all this tho

paulirish:
lighthouse-core/test/report/html/renderer/category-renderer-test.js is cranky.

Hoten:
done. had to introduce a wrapper el so that the label el makes more sense :)

lighthouse-core/test/report/html/renderer/category-renderer-test.js
```diff
@@ -185,7 +185,7 @@ describe('CategoryRenderer', () => {
     const categoryDOM = renderer.render(category, sampleResults.categoryGroups);
     assert.ok(categoryDOM.querySelector('.lh-gauge__wrapper--plugin'));
     const label = categoryDOM.querySelector('.lh-gauge__label').textContent;
-    assert.equal(category.title, label);
+    assert.equal(label, category.title);
```
Hoten: was backwards

paulirish:
can you resolve the conflict?

lighthouse-core/test/report/html/renderer/category-renderer-test.js
brendankenny: > was backwards

just for code correctness purposes, not a functional change in the test, I hope?

Hoten:
Not functional. Error message was wrong when the test was failing.

lighthouse-core/report/html/templates.html
```diff
@@ -656,11 +666,35 @@
...
+      height: var(--lh-info-icon-background-size);
+    }
+
+    .lh-gauge__label-wrapper .tooltip {
+      font-size: var(--caption-font-size);
+      line-height: var(--caption-line-height);
+    }
   </style>
-  <a href="#" class="lh-gauge__wrapper">
+  <a class="lh-gauge__wrapper">
```
brendankenny: should these not be `<a>` now?

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -178,8 +178,18 @@ class CategoryRenderer {
...
       const descEl = this.dom.convertMarkdownLinkSnippets(category.description);
-      this.dom.find('.lh-category-header__description', tmpl).appendChild(descEl);
+      this.dom.find('.tooltip', tooltipEl).appendChild(descEl);
+
+      // Wrap the label so the tooltip icon can be positioned relative to it.
+      const gaugeWrapper = this.dom.find('.lh-gauge__wrapper', gaugeContainerEl);
+      const labelWrapper = this.dom.createChildOf(gaugeWrapper, 'div', 'lh-gauge__label-wrapper');
+      const gaugeLabelEl = this.dom.find('.lh-gauge__label', gaugeContainerEl);
+      labelWrapper.appendChild(gaugeLabelEl);
+      labelWrapper.appendChild(tooltipEl);
```
brendankenny: for some reason I can't for the life of me follow what this is doing :) Taking stuff in the template, appending it to the wrapper, than putting the wrapper into what was in the template? Can the `labelWrapper` just live in the template? Seems like that it might simplify this dance

lighthouse-core/report/html/renderer/category-renderer.js
```diff
@@ -324,8 +334,7 @@ class CategoryRenderer {
    */
   renderScoreGauge(category, groupDefinitions) { // eslint-disable-line no-unused-vars
     const tmpl = this.dom.cloneTemplate('#tmpl-lh-gauge', this.templateContext);
-    const wrapper = /** @type {HTMLAnchorElement} */ (this.dom.find('.lh-gauge__wrapper', tmpl));
-    wrapper.href = `#${category.id}`;
+    const wrapper = this.dom.find('.lh-gauge__wrapper', tmpl);
```
brendankenny: what's the point of moving this? (this same code is also in `pwa-category-renderer`)

Hoten: It was suppressing the click event for links in the description. Only the header should be a clickable link, so the href= was moved to where the header gauges are made.

Should have removed the PWA one too.

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: The issue is that the tooltip must be placed relative to the label, and the label is in `tmpl-lh-gauge`, which is shared by the category header gauges and the main header gauge.

Perhaps `renderScoreGauge` could also take a description, the tooltip stuff moved to the `tmpl-lh-gauge` template, and hide the tooltip if the description is empty.

lighthouse-core/report/html/templates.html
Hoten: well, they are still links but only in the main gauge header

lighthouse-core/test/report/html/renderer/category-renderer-test.js
Hoten: yes

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: hows that

lighthouse-core/report/html/renderer/pwa-category-renderer.js
```diff
@@ -54,7 +54,7 @@ class PwaCategoryRenderer extends CategoryRenderer {
   renderScoreGauge(category, groupDefinitions) {
     // Defer to parent-gauge style if category error.
     if (category.score === null) {
-      return super.renderScoreGauge(category, groupDefinitions);
```
patrickhulce: I think we want false here or we need a fix somewhere else?

image

Hoten: oops yea, there are two gauge templates and I forgot to handle the pwa one.

paulirish:
sorry @Hoten 

image

Hoten:
merged master

next steps on this?

lighthouse-core/report/html/renderer/category-renderer.js
Hoten: that didn't work (just linked to top of page instead). Added a new 'disable' class as a quick fix. but maybe a better (but bigger) change could be done?

Hoten:
image

uh....bump it up?

brendankenny:
- > bump it up? if we truly want to keep it under 70, we can open an issue to research how to do that

   I think that's fine. We've been hovering above 69KB for a while now, and adding things to the report renderer was going to do it sooner or later. 75KB?

- The tooltips don't appear to be working right now, at least in the deployed example.

- so, there are two states here: gauge+label for the scores header, gauge+label+optional-description for categories? I'm not entirely following what's needed with `<a>` vs not `<a>`, though.

   Rather than passing the option into `renderScoreGauge`, can it go back to how it was before, adding a tooltip if there's a description in `renderCategoryHeader()` (ensuring the score header gauges don't get one).

  If there's more needed to sort out the anchor business, maybe it needs to be split more. The different category vs scores-header gauges aren't really the same thing *except* the score gauge...we could have `renderScoreGauge` return only what's really shared between the two uses and move more functionality back to the call sites.

Hoten:
incorporated @brendankenny's feedback

patrickhulce:
Sorry @Hoten seems like this regressed again at some point in the feedback cycle.

image

Hoten:
oops, fixed that

lighthouse-core/report/html/report-styles.css
```diff
@@ -1246,7 +1281,6 @@
 }
 
 .lh-category-header .lh-score__gauge {
-  max-width: 400px;
```
patrickhulce: just unnecessary or do we want the gauges to get bigger than this?

lighthouse-core/report/html/report-styles.css
```diff
@@ -1470,7 +1504,6 @@
   display: block;
   animation: fadeInTooltip 250ms;
   animation-fill-mode: forwards;
-  animation-delay: 850ms;
```
patrickhulce: this removes this everywhere, we want that right? I would guess so because a lot of folks don't seem to know that they can hover over the error to get the tooltip 👍

lighthouse-core/report/html/report-styles.css
```diff
@@ -1139,6 +1142,16 @@
...
 }
 
+/* Category header gauges shouldn't be links. */
+.lh-category-wrapper .lh-gauge__wrapper {
+  pointer-events: none;
+  cursor: default;
+}
+.lh-gauge__label-info-icon {
+  pointer-events: fill;
+  cursor: help;
```
patrickhulce: feels a tad odd to me to flip the cursor since there's no actual click action here, is there a broader UX strategy here? I would love to learn more about these kinds of things :)

lighthouse-core/test/report/html/renderer/pwa-category-renderer-test.js
```diff
@@ -242,7 +242,7 @@ describe('PwaCategoryRenderer', () => {
   describe('#renderScoreGauge', () => {
     it('renders an error score gauge in case of category error', () => {
       category.score = null;
-      const badgeGauge = pwaRenderer.renderScoreGauge(category);
+      const badgeGauge = pwaRenderer.renderScoreGauge(category, null);
```
patrickhulce: what's the point of this one? typedef suggests both are unexpected but looking through the code it looks like we never used groups for the score gauge no matter what.... 🤷‍♂

lighthouse-core/report/html/report-styles.css
Hoten: honestly, I had the bring the cursor back with CSS so I just picked one. Since there's no click action to prompt perhaps it should just remain a normal cursor. IDK.

@yuinchen what do you think of this?
https://lighthouse-33lan68g9.now.sh/#accessibility (hover over `i`)

lighthouse-core/test/report/html/renderer/pwa-category-renderer-test.js
Hoten: don't recall, removed.

lighthouse-core/report/html/report-styles.css
Hoten: yeah, I figured it doesn't make sense to inject 850ms of latency for this.

lighthouse-core/report/html/report-styles.css
Hoten: unnecessary

lighthouse-core/report/html/report-styles.css
yuinchien: @Hoten I suggest shifting the position 4px downward and to the right.

I applied `transform: translate3d(4px,4px,0)` to the info icon and it looks like this.

Screen Shot 2019-05-30 at 10 22 10 AM

brendankenny:
https://www.matuzo.at/blog/building-the-most-inaccessible-site-possible-with-a-perfect-lighthouse-score/

is an argument against moving the accessibility description to a tooltip (see especially the conclusion at the end). Maybe it would be better just to shorten + only exist if there's something really important to say

connorjclark:
what are we doing with this?

brendankenny:
> what are we doing with this?

I don't think we want to do this. Implementation complexity, not a huge win in terms of simplification, and an [arguable net loss of explainable utility](https://github.com/GoogleChrome/lighthouse/pull/8840#issuecomment-499651562)

PR #8845, opened 2019-05-04
report(redesign): make dark mode details less intense

exterkamp:
**Summary**
Dark mode seemed a bit high in contrast to me, and made me not really like it.  These are a few nits that make it more _subtle_ and put together to me.

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -85,6 +85,7 @@
   --color-black-600: #757575;
   --color-black-800: #424242;
   --color-black-900: #212121;
+  --off-black: #1b1b1b;
   --color-black: #000000;
   --color-blue: #2962FF;
   --color-green-700: #018642;
@@ -178,10 +179,10 @@
   --color-black-400: var(--color-black-600);
   --color-black-600: var(--color-black-500);
 
-  --topbar-bg: var(--color-black);
+  --topbar-bg: var(--off-black);
   --plugin-badge-bg: var(--color-black-800);
   --header-bg-color: var(--color-black-900);
-  --env-item-bg: rgba(0, 0, 0, 0.5);
+  --env-item-bg: var(--off-black);
   --report-secondary-border-color: var(--color-black-200);
 
   --body-background-color: var(--color-black-900);
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -179,10 +179,10 @@
   --color-black-400: var(--color-black-600);
   --color-black-600: var(--color-black-500);
 
-  --topbar-bg: var(--off-black);
+  --topbar-bg: var(--color-black);
   --plugin-badge-bg: var(--color-black-800);
   --header-bg-color: var(--color-black-900);
-  --env-item-bg: var(--off-black);
+  --env-item-bg: var(--color-black);
   --report-secondary-border-color: var(--color-black-200);
 
   --body-background-color: var(--color-black-900);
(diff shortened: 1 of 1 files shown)


paulirish:
@yuinchien lgty?

yuinchien:
I think the clean definition of the black topbar, it helps accentuate the color scores on the page. I also think it needs to have enough contrast to differentiate from the sticky bar. The new change loose the definition, I recommend the original black design.
Screen Shot 2019-05-04 at 1 31 29 PM

I prefer the original borders looking slightly more clean cut and defined. The new change feels a bit blurry to me. I recommend keeping the original design.
Screen Shot 2019-05-04 at 1 31 56 PM

What I think needs work are the following:

- Removing the gray background in the heading 
Screen Shot 2019-05-04 at 1 40 37 PM

- Align toggle to the right edge
Screen Shot 2019-05-04 at 1 42 06 PM

- Wrap text after em dash properly
Screen Shot 2019-05-04 at 1 44 23 PM

exterkamp:
@yuinchien 
This is the current side-by-side I see, left old, right this branch:
image
These borders are extremely high contrast, much more than the light theme, and don't match the full width horizontal borders.  I feel like they draw too much attention away from the text.

I'm not gonna bikeshed this _too_ hard.  But I just really don't like the pure black topbar.  I think it needs to be toned down a bit.  Is this better?  It's #111111.
image


I think these are valid, but I wasn't touching them in this PR:
* The toggle alignment
* Wrap text

paulirish:
let's go with our designer's recommendations :)

yuinchien:
@exterkamp Seeing your screenshot, yes the border color is too bright in current implementation. In my prototype (the source of truth for design) https://yuinchien.github.io/sandbox/lighthouse/ the border color is much more subtle. In my css, the black-200 is replaced with black-800 (#424242).   

body.dark 
    --color-black-200: var(--color-black-800);

I would recommend keeping the colors consistent with the existing color system, and not introduce arbitrary hex that's only used in one place. We already have 8 shades of black in the report used in our components, it's better to have the consistency. I have to say i like the sharper black in the topbar, it makes the LH symbol pop more too. 

LH shades of black:
  --color-black: #000000;
  --color-black-900: #212121;
  --color-black-800: #424242;
  --color-black-600: #757575;
  --color-black-500: #9E9E9E;
  --color-black-400: #BDBDBD;
  --color-black-200: #E0E0E0;
  --color-black-100: #F5F5F5;
  --color-white: #FFFFFF;

exterkamp:
Okay, sgtm.  Yeah this branch brings the borders in line with the design. So, the borders are ✅ 

The toggle pill in dark mode in your design doesn't have a dark mode, so wdyt of mine?
image

I'm not gonna fight over the topbar color.  If black works for you, then it works for me.

yuinchien:
Screen Shot 2019-05-04 at 5 24 27 PM

For the toggle pill, l think the following background should work well (screenshot above):
Light theme: --color-black-200
Dark theme: --color-black-800

Thanks @exterkamp for your dedication, in my experience, products are most successful because everyone is deeply passionate and cares enough to communicate what they think is best!

exterkamp:
You beat me to a response!  But awesome, thanks for the feedback!

Summary of all changes side-by-side -> left (old) | right (new)
Metrics borders:
image

Toggle Pill:
image

Just a note on the runtime settings & topbar: the runtime settings in your darkmode mock use rgba(0,0,0,0.5) which resolves to #101010.  Which I like a lot better than a pure black.  Which is what I was trying to emulate with the `off-black` color scheme by putting it on the runtime-settings and the topbar. That was my line of thinking.
image

PR #8853, opened 2019-05-04
tests(lightwallet): add perf-budget smoke test

brendankenny:
added a smoke test for the `resource-summary` (#8522) and `performance-budget` (#8539) audits.

This just rips off some things from the two other perf testing pages so we could get a decent mix of resources loaded. Feel free to make suggestions for improvements :)

Three other things done in the process:
- moved the `perf` smoke test to its own config file (off of `--preset=perf`) so that it could include a budget
- made `undefined` in smoke expectations match with a property not found on the `actual` result. Previously the matching `actual` property had to explicitly be set to `undefined` to match, which isn't even possible because smoke test results are saved/loaded from JSON, where `undefined` properties get dropped.
- added `'±'` as an alias for `'+/-'` as one of the allowed operators, cause it's cooler :)

part of #8331


brendankenny:
cc @khempenius

lighthouse-cli/test/smokehouse/perf/expectations.js
```diff
@@ -65,6 +65,83 @@ module.exports = [
...
+    lhr: {
+      requestedUrl: 'http://localhost:10200/perf/perf-budgets/load-things.html',
+      finalUrl: 'http://localhost:10200/perf/perf-budgets/load-things.html',
+      audits: {
+        'resource-summary': {
+          score: null,
+          displayValue: '11 requests • 164 KB',
+          details: {
+            items: [
+              {resourceType: 'total', requestCount: 11, size: '168000±1000'},
```
brendankenny: used ± to leave some leeway on some of the assets because most of them are shared with other smoke tests. This will allow the test to absorb changes to them of a few hundred bytes or so, but we can always make it exact.

lighthouse-cli/test/smokehouse/perf/expectations.js
```diff
@@ -65,6 +65,83 @@ module.exports = [
...
+              },
+              {
+                resourceType: 'stylesheet',
+                countOverBudget: undefined,
+                sizeOverBudget: '450±100',
+              },
+              {
+                resourceType: 'image',
+                countOverBudget: '1 request',
+                sizeOverBudget: undefined,
```
brendankenny: Letting `undefined` values in the expectations match *not defined* properties in `actual` allows testing for things like `sizeOverBudget` when not over budget, [when its value is set to `undefined`](https://github.com/GoogleChrome/lighthouse/blob/ec08e7b9375d3cbe38938b95a0dadfd8287d532e/lighthouse-core/audits/performance-budget.js#L76-L84) and so it lost in the JSON shuffle.

If we just left this `sizeOverBudget` off the expectations it would indeed match an `actual` `undefined` value, which is good...but it would also match if `actual` was erroneously returning a numeric `sizeOverBudget`, so it would be useless.

patrickhulce: should we add a comment somewhere to this effect?

lighthouse-cli/test/smokehouse/perf/perf-config.js
```diff
@@ -0,0 +1,50 @@
...
+'use strict';
+
+/** @type {LH.Config.Json} */
+const perfConfig = {
+  extends: 'lighthouse:default',
+  settings: {
+    throttlingMethod: 'devtools',
+    onlyCategories: ['performance'],
+
+    // A mixture of under, over, and meeting budget to exercise all paths.
```
patrickhulce: maybe include the path to the file where these budgets are being tested so we can look them up for correctness later? it's a bit weird that our "expectations" are basically here in the config that's shared with several other smokes

lighthouse-cli/test/smokehouse/perf/perf-config.js
```diff
@@ -0,0 +1,50 @@
...
+        {resourceType: 'stylesheet', budget: 0},
+        {resourceType: 'image', budget: 30}, // meets budget
+        {resourceType: 'media', budget: 0},
+        {resourceType: 'font', budget: 75},
+        {resourceType: 'script', budget: 30},
+        {resourceType: 'document', budget: 1},
+        {resourceType: 'other', budget: 2}, // meets budget
+        {resourceType: 'third-party', budget: 0},
+      ],
+      timings: [
```
patrickhulce: these aren't being used right now, correct?

brendankenny: > these aren't being used right now, correct?

right, they're parsed by budget.js, but the audit doesn't check them yet

lighthouse-cli/test/smokehouse/perf/perf-config.js
brendankenny: > it's a bit weird that our "expectations" are basically here in the config
I mean, that's kind of what all configs are :)

> that's shared with several other smokes

I could split it into a new smoke `budget` category? Or is it fine to leave as one big perf category

lighthouse-cli/test/smokehouse/perf/expectations.js
brendankenny: > should we add a comment somewhere to this effect?

added a comment to make it clear to anyone reading the expectations file

PR #8863, opened 2019-05-05
misc(proto): assert libprotoc version

Hoten:
Prevents churn from unintended changes caused by version differences. See https://github.com/GoogleChrome/lighthouse/pull/8857#discussion_r281004034


brendankenny:
I'm like 90% sure this is wrong, btw :)

Issue #7292, opened 2019-02-21
Test lighthouse on website with invalid certificate fails


abarre:
This bug has been found via Webpagetest. I also did some tests with the command line.

#### Provide the steps to reproduce

The following possibilities instruct Chrome to ignore certificate errors and resolve www.lemonde.fr to IP 62.210.77.17. Note that the server 62.210.77.17 doesn't have a valid certificate for www.lemonde.fr.

Three possibilities :
1. Run lighthouse with chrome flags like so :
```bash
lighthouse "https://www.lemonde.fr/?frz-demo=true" --chrome-flags="--ignore-certificate-errors --host-resolver-rules='MAP www.lemonde.fr 62.210.77.17'"
```

2. or Run Chrome and lighthouse separately :
```bash
chrome --host-resolver-rules="MAP www.lemonde.fr 62.210.77.17" --remote-debugging-port=9251 --ignore-certificate-errors about:blank
lighthouse "https://www.lemonde.fr/?frz-demo=true" --port=9251
```

3. or Via Webpagetest :
- Enable lighthouse report
- Enable "ignore certificate errors"
- Use the script :
```
setDns  www.lemonde.fr  62.210.77.17
navigate  https://www.lemonde.fr/?frz-demo=true
```

Ref : http://wpt.fasterize.com/result/190221_X6_R/

#### What is the current behavior?

The report is not launched whereas the page is correctly displayed on Chrome

```
Insecure document request: The URL you have provided does not have valid security credentials. This site is missing a valid, trusted certificate (net::ERR_CERT_COMMON_NAME_INVALID).
```

Full log : 

```
status:verbose Create config +0ms
  status:verbose Requiring gatherers +3ms
  statusEnd:verbose Requiring gatherers +13ms
  status:verbose Requiring audits +1ms
  statusEnd:verbose Requiring audits +100ms
  statusEnd:verbose Create config +9ms
  status:verbose Runner setup +2ms
  status Connecting to browser +1ms
  statusEnd:verbose Connecting to browser +61ms
  status:verbose Getting browser version +1ms
  method => browser:verbose Browser.getVersion  +3ms
  method <= browser OK:verbose Browser.getVersion {"protocolVersion":"1.3","product":"Chrome/72.0.3626.109","revision":"@fae8db7ab9280fa6704a59980263c804f809ebd5","userAgent":"Mozilla/5.0 (Ma +9ms
  statusEnd:verbose Getting browser version +1ms
[...]

Issue #7011, opened 2019-01-14
Consistent style for lh-warnings


exterkamp:
Mentioned in #6989.  I think the lh-warnings style looks out of place with the report style e.g.
![image](https://user-images.githubusercontent.com/6392995/51146678-608c6a80-180c-11e9-9d2d-386318470f8e.png)

My suggestion is to replace the 1px border with a light orange background (#fff5e3):
![image](https://user-images.githubusercontent.com/6392995/51146759-9d586180-180c-11e9-9206-1bcdc18b37ab.png)

Would love some thoughts. @hwikyounglee

paulirish:
@yuinchien is looking at this.

hwikyounglee:
Su^per late - 

@exterkamp 's suggestion on using the bg color instead of border looks good to me while we're currently looking at a longer term approach in a separate effort.

exterkamp:
I'm gonna give this to @Hoten as part of the report-redesign :smile:

PR #8865, opened 2019-05-05
core(driver): security errors are no longer a fatal or pageload error

### LINKED ISSUE #6655: Don't fail localhost pages with INSECURE
Testing a local site using a self-signed certificate seems pretty reasonable. 
Wdyt about us allowing `localhost` in our INSECURE checks?

paulirish:
fix #6655, #7292
closes #7574

Following the investigation and proposal outlined here: https://github.com/GoogleChrome/lighthouse/issues/6655#issuecomment-484964210

I've tested with every badssl page and we never hang indefinitely on any of them. I did see a few PROTOCOL_TIMEOUT on the offlinePass though.

Also every badssl case where a security interstitial is shown hits our FAILED _DOC_REQ case and the `localizedFailDescription` that we show in the warning captures the same error code we got via securityState.insecureDescriptions. \o/

see also #8340


lighthouse-core/gather/driver.js
```diff
@@ -903,53 +903,6 @@ class Driver {
     };
   }
 
-  /**
-   * Return a promise that resolves when an insecure security state is encountered
-   * and a method to cancel internal listeners.
-   * @return {{promise: Promise<string>, cancel: function(): void}}
-   * @private
-   */
-  _monitorForInsecureState() {
```
Hoten: rip

lighthouse-cli/test/smokehouse/error-expectations.js
```diff
@@ -18,10 +18,10 @@ module.exports = [
     },
   },
   {
-    errorCode: 'INSECURE_DOCUMENT_REQUEST',
+    errorCode: undefined,
```
brendankenny: this should have other expectations then? And maybe a comment that this is ensuring there *isn't* an error (since it's in the `error-expectations` file :)

patrickhulce: This should be fixed by followup to #8340, the comments I have there address the fact that none of these are currently fatal errors but IMO should be non-zero exit codes.

patrickhulce:
What's our opinion on this one are we nuking it entirely like this PR does?

Or are we just going to move it to be non-fatal with a warning like followup to #8340 would?

brendankenny:
(I can help come up with actual solutions--not just a list of problems--tomorrow, I just had to leave for the day :)

patrickhulce:
My proposal for path forward:

1. Land this as-is.
2. Follow-up PR to make all `runtimeError`s exit with a non-zero exit code as discussed here and in #8340. This should not impact any environment other than CLI. Other consumers should already be deciding whether to consume the LHR or not based on the existence of `runtimeError`.
4. Land a modified version of #7574 to actually make HTTPS pages with bad certs loadable. This PR as-is doesn't fix the #7574 concern because the pages still fail to load.
5. Land a rebased version of #8866.
6. Land a version of #8866 but for audits that fails them out with a specifc `runtimeError` error and skips the unnecessary logging.

Specific responses and rationale to @brendankenny's great points below:

> but is it strange to be showing all this content when none of it is meaningful?

IMO, yes it's super strange. But that's the paradigm we've settled on elsewhere and I can understand the rationale for certain types of integrations that want some of the information we collect.

> a whole bunch of logging from work that didn't need to be done

Fixed by step 6.

> also weird that we still take a trace and a devtools log for a page that isn't the page being targeted 

As I understand it this is actually the primary desire and reason for providing this mostly useless LHR and returning the `runtimeError` instead of throwing. So, feature not bug ;)

> as a result of how we currently put together runtimeError from the artifacts, NO_NAVSTART doesn't bubble up as a runtime error (even though it's marked as one), so e.g. the current badssl error smoke test doesn't get a runtimeError or an error exit code. That definitely shouldn't depend on what audits are being run.

I think this is fixed by the combination of steps 2, 5, and 6.

brendankenny:
> Follow-up PR to make all `runtimeError`s exit with a non-zero exit code as discussed here and in #8340. This should not impact any environment other than CLI. Other consumers should already be deciding whether to consume the LHR or not based on the existence of `runtimeError`.

I think with #8340 this is already done? (for all instances of `runtimeError` that actually become an `lhr.runtimeError`)

> Land a modified version of #7574 to actually make HTTPS pages with bad certs loadable. This PR as-is doesn't fix the #7574 concern because the pages still fail to load.

with this PR landed, isn't `--chrome-flags="--ignore-certificate-errors"` sufficient? (see discussion from https://github.com/GoogleChrome/lighthouse/issues/6655#issuecomment-465410790 onwards)

steps 4 and 5 sound good

>> but is it strange to be showing all this content when none of it is meaningful?

> IMO, yes it's super strange. But that's the paradigm we've settled on elsewhere and I can understand the rationale for certain types of integrations that want some of the information we collect.

well the strangeness feels bad :) https://googlechrome.github.io/lighthouse/viewer/?gist=b8c85fa12128c68f017e5074c4e50e2d doesn't seem like something we should be returning. `lhr.runtimeError` and the exit code are good to go in that case, but everything else is at best useless, at worst confusing.

Maybe we need the Lighthouse version of a server error page, a design reminiscent of the main report but not trying to render a bunch of stuff that makes no sense.

>> also weird that we still take a trace and a devtools log for a page that isn't the page being targeted

> As I understand it this is actually the primary desire and reason for providing this mostly useless LHR and returning the runtimeError instead of throwing. So, feature not bug ;)

I agree in the case that `--chrome-flags="--ignore-certificate-errors"` is passed in, they get a `net::ERR_CERT_DATE_INVALID` error but since we can still load the page and test it we should. But without the flag, a trace and devtoolsLog for a page (the security interstitial) that's different than the page they were trying to test seems like a bug not a feature :)

It seems like @paulirish's "we shouldn't throw if we can technically finish loading the page" shouldn't apply if we're loading something but it isn't "the page".

So, suggested addenda to suggested plans:
- step 5 += a Chrome interstitial detector that leads to a runtime error
- removing some base artifacts in the face of an interstitial (e.g. trace and devtoolsLog of the interstitial)
- a new error-state report. Maybe it keeps the ? ? ? ? ? header and #7011-ified warning but doesn't render categories or something

patrickhulce:
> I think with #8340 this is already done? 

*Almost* but we're not quite there yet. I called it out in the PR description in #8340 but PAGE_HUNG and INSECURE_DOCUMENT_REQUEST both don't abide by this because of https://github.com/GoogleChrome/lighthouse/pull/8340/files#diff-42278a7ee772120215ea8cfe1b0cb1b1R74. INSECURE_DOCUMENT_REQUEST will take care of itself by landing this

> with this PR landed, isn't --chrome-flags="--ignore-certificate-errors" sufficient? 

Well depends on what you mean by sufficient, it'll still be subject to all the other problems you've called out and seems like a LH flag to do both for you sounds like a good idea :)

> everything else is at best useless

The point as I understand it is that an integration can still take a look at the artifacts we managed to collect to get screenshots/request specifics and display a better debug screen. Maybe we should be building such a screen based on artifacts too, is part of your argument here?

> step 5 += a Chrome interstitial detector that leads to a runtime error

👍 

> removing some base artifacts in the face of an interstitial (e.g. trace and devtoolsLog of the interstitial)

As I understand it, this seems like it destroys the only value we have in not throwing. Why don't we want to make this information available for debugging?

> a new error-state report. Maybe it keeps the ? ? ? ? ? header and #7011-ified warning but doesn't render categories or something

👍

brendankenny:
>> with this PR landed, isn't --chrome-flags="--ignore-certificate-errors" sufficient?

with this PR

`lighthouse https://expired.badssl.com --view --chrome-flags="--ignore-certificate-errors"`

really does work fine since Chrome can proceed to load the underlying page:

[expired-badssl Lighthouse report](https://googlechrome.github.io/lighthouse/viewer/?gist=6797a0f98ab9c1a9f3f86abad2bea41a)

https://googlechrome.github.io/lighthouse/viewer/?gist=6797a0f98ab9c1a9f3f86abad2bea41a

I personally think we should have an audit that calls out the bad certificate even while everything else is green, but everyone else seems to think everything should be all green in this case since the user explicitly opted in to it, so  the `--ignore-certificate-errors` case may be good to go after this lands.

patrickhulce:
Wait I thought you were just saying it was the other screenshot and that was the main problem! Oh yes yes, still seems like a flag might be useful at some point in the future but not necessary anymore.

Janpot:
> I personally think we should have an audit that calls out the bad certificate even while everything else is green, but everyone else seems to think everything should be all green in this case since the user explicitly opted in to it

IMO It makes sense to make this opt in. And when I do, lighthouse should behave very similar when a website has no certificate vs. a bad certificate. I opt in, not because I don't care about security, but because it shouldn't block me from auditing every other aspect of my website's performance. If there's an audit that calls me out for having no https, then it should also call me out for having broken https.

patrickhulce:
So where do we stand with this then. @brendankenny do you have specific requests for this PR or are all of your concerns with points other than number 1 in the plan?

cjolif:
any idea on when a release will that fix will be made?

connorjclark:
#9442 release is imminent

cjolif:
Thanks!!

PR #8866, opened 2019-05-05
core: bail on gathering if we have a failure in the first pass

paulirish:
While working on #8865 I noticed the offline pass is troublesome for security interstitial cases. This would help us fail the whole run quickly if we have a pageLoadError when loading the original URL.

_In essence: if we can't load the provided URL, we shouldn't try to load it offline and we shouldn't try to see if the http redirect works._

There's a few ways of doing this, but I went for the dumb and straightforward way. :)


paulirish:
@patrickhulce i wrote this before reading #8340 ... seeing that we want both... what do you think makes sense?

patrickhulce:
Doing both this and #8340 makes sense to me 👍 

The only part about this I don't love is that the audits in those other passes will now fail with "required artifact Blah missing" instead of the load error, right?

PR #8867, opened 2019-05-05
misc(proto): require protobuf 3.7.1, add stricter audit details test

brendankenny:
follow up to #8863 and https://github.com/GoogleChrome/lighthouse/pull/8857#discussion_r281034929

Something was fixed between protobuf 3.6.1 and 3.7.1 that handled the `details.summary` type when the value was just an empty `{}`. 3.6.1 round tripped it as `null`, 3.7.1 correctly round trips it as `{}`.

The round trip test wasn't catching the issue because `{}` was the `summary` expectation and `{}` matches basically everything with `toMatchObject`. The round trip is good enough for audit results now that we can just do the deep, strict `toEqual` instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "ultradumbBenchmark": "./lighthouse-core/scripts/benchmark.js",
     "mixed-content": "./lighthouse-cli/index.js --chrome-flags='--headless' --preset=mixed-content",
     "minify-latest-run": "./lighthouse-core/scripts/lantern/minify-trace.js ./latest-run/defaultPass.trace.json ./latest-run/defaultPass.trace.min.json && ./lighthouse-core/scripts/lantern/minify-devtoolslog.js ./latest-run/defaultPass.devtoolslog.json ./latest-run/defaultPass.devtoolslog.min.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
     "ultradumbBenchmark": "./lighthouse-core/scripts/benchmark.js",
     "mixed-content": "./lighthouse-cli/index.js --chrome-flags='--headless' --preset=mixed-content",
     "minify-latest-run": "./lighthouse-core/scripts/lantern/minify-trace.js ./latest-run/defaultPass.trace.json ./latest-run/defaultPass.trace.min.json && ./lighthouse-core/scripts/lantern/minify-devtoolslog.js ./latest-run/defaultPass.devtoolslog.json ./latest-run/defaultPass.devtoolslog.min.json",
-    "compile-proto": "[ \"$(protoc --version)\" == 'libprotoc 3.6.1' ] && protoc --python_out=./ ./proto/lighthouse-result.proto && mv ./proto/*_pb2.py ./proto/scripts || (echo \"❌ Install protobuf to compile the proto file.\" && false)",
+    "compile-proto": "[ \"$(protoc --version)\" == 'libprotoc 3.7.1' ] && protoc --python_out=./ ./proto/lighthouse-result.proto && mv ./proto/*_pb2.py ./proto/scripts || (echo \"❌ Install protobuf to compile the proto file.\" && false)",
```
brendankenny: @paulirish sounds like we don't want the version test anymore, though? Probably good cause future versions wouldn't match and I'm not sure if the version check will work cross platform (we don't run `update:sample-json` on appveyor...)

PR #8870, opened 2019-05-06
tests(lightwallet): add budget to golden LHR

brendankenny:
#8427 added a budget to the sample artifacts, but because the `-A` step in `yarn update:sample-json` was run without the same config, the budget didn't make its way into the golden `sample_v2.json`.

This PR *actually* solves the original problem and gets a budget all the way into the final JSON :)

This will make things like #8708 easier to review because it'll actually be in the deployed report.

I just picked budget numbers. Happy to haggle.


lighthouse-core/test/results/sample-config.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+/** @type {LH.Config.Json} */
+const budgetedConfig = {
+  extends: 'lighthouse:default',
+  settings: {
+    throttlingMethod: 'devtools',
+    budgets: [{
+      resourceCounts: [
+        {resourceType: 'total', budget: 10},
```
patrickhulce: wish we could reuse this for the smokes to keep it in sync :)

PR #8891, opened 2019-05-06
report(redesign): metrics padding, font weights

exterkamp:
**Summary**
* Remove metrics horizontal padding
* Set group headers explicitly to font-weight 700 (bold -> 700 is a no-op)
image

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -894,7 +894,7 @@
 }
 
 .lh-audit-group__header span.lh-audit-group__title {
-  font-weight: 700;
+  font-weight: bold;
 }
 
 .lh-audit-group__header span.lh-audit-group__itemcount {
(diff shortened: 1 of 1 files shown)


yuinchien:
@exterkamp "Values are estimated and may vary." should be left aligned.

lighthouse-core/report/html/report-styles.css
```diff
@@ -894,7 +894,7 @@
 }
 
 .lh-audit-group__header span.lh-audit-group__title {
-  font-weight: bold;
+  font-weight: 700;
```
paulirish: i'd rather keep this as `bold`, unless you want to change all `bold` to `700`.

exterkamp: @yuinchien?

yuinchien: I think you guys should decide this one, as long as the text rendered the correct weight i'm happy.

exterkamp: using *bold*

PR #8893, opened 2019-05-07
report(redesign): move gauge styles to main stylesheet

Hoten:
#8185 

Fixes: 
image

The PWA gauge depends on many style declarations defined in the template for regular gauges. When there are no regular gauges, that `style` element never gets add. I moved all those styles to the main stylesheet to cope.


paulirish:
i can tell you love these tests by now :)

image

lighthouse-core/test/report/html/renderer/psi-test.js
```diff
@@ -82,7 +82,6 @@ describe('DOM', () => {
         const result = prepareLabData(sampleResultsStr, document);
         assert.ok(result.scoreGaugeEl instanceof document.defaultView.Element);
         assert.equal(result.scoreGaugeEl.querySelector('.lh-gauge__wrapper').href, '');
-        assert.ok(result.scoreGaugeEl.outerHTML.includes('<style>'), 'score gauge comes with CSS');
```
brendankenny: I assume this is ok? The PSI renderer already includes `report-styles.css`?

Hoten: pretty sure it does. @exterkamp  to confirm

exterkamp: Yup, PSI imports report-styles.  Should be g2g!

PR #8899, opened 2019-05-07
report(redesign): fix word breaking in audit display text

Hoten:
Issue:

image

now

image

image

Better but the em dash does get its own line on some viewports. may need to rethink where we put the warnings/errors text.

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -470,11 +470,8 @@
   margin: 0 var(--audit-item-gap);
 }
 .lh-chevron-container {
+  margin-left: auto;
   margin-right: 0;
-}
-
-.lh-audit__display-text {
-  flex: 1;
 }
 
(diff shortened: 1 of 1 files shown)


patrickhulce:
Am I the only one that feels like our whitespace issues are more fragile than a house of cards? 😆 

I read Paul's beautiful comment and thought I understood what was going on and yet I still have no clue 😢

paulirish:
i tried moving the display-text span inside of the audit title span and it looked better. all part of one flow instead of two columns side-by-side.



image
vs
image


I was just moving in devtools. didnt try it in the actual renderer. 
can we try this, though?

brendankenny:
> i tried moving the display-text span inside of the audit title span and it looked better. all part of one flow instead of two columns side-by-side.

two columns definitely makes it weird when we sometimes have a ton of extra text to add. Besides just flow with the `displayValue`, should we be putting the audit warning somewhere else than just sticking it in with the title?

paulirish:
addressed by #9030

PR #8933, opened 2019-05-11
cli: accept csv for array values

Hoten:
```
node lighthouse-cli https://www.example.com/ --only-categories=performance,seo --only-categories=pwa
```


lighthouse-cli/cli-flags.js
```diff
@@ -12,12 +12,21 @@ const pkg = require('../package.json');
 const printer = require('./printer');
 
 /**
- * @param {string=} manualArgv
+ * @param {(string | string[])[]} arr
+ * @return {string[]}
+ */
+function flatten(arr) {
```
Hoten: fyi a generic "flatten" is impossible in ts :)

patrickhulce: is it really? how come :)

lighthouse-cli/cli-flags.js
```diff
@@ -152,6 +161,17 @@ function getFlags(manualArgv) {
       .default('enable-error-reporting', undefined) // Undefined so prompted by default
       .default('channel', 'cli')
       .check(/** @param {LH.CliFlags} argv */ (argv) => {
+        // ".middleware" does not exist in this version of yargs, so do some preprocessing here.
+        for (const _key of Object.keys(argv)) {
```
patrickhulce: I don't love that this is here and I'm not sure we can actually do this across the board.

(for example `--chrome-flags="--window-size 800,600"`)

WDYT about just doing this to our specific `.array` properties minus `blockedUrlPatterns`?

lighthouse-cli/test/cli/cli-flags-test.js
```diff
@@ -28,4 +28,14 @@ describe('CLI bin', function() {
       assert.ok(optionsWithDescriptions.includes(opt), `cli option '${opt}' has no description`);
     });
   });
+
+  it('array values support csv', () => {
+    const flags = getFlags(
```
patrickhulce: once we figure out other feedback would love the `--window-size` and `blockedUrlPattern` counter examples here too :)

lighthouse-cli/cli-flags.js
Hoten: https://www.reddit.com/r/typescript/comments/apfkxh/how_should_i_correctly_type_an_array_flatten/eg82tcu/

I tried to do the multiple variations approach, via jsdoc's `@variation`, but no luck. Unsure if that's fully supported in TS's jsdoc interop.

lighthouse-cli/cli-flags.js
Hoten: ah ok. sounds like either way we need a hard coded list (either a blacklist of a whitelist). Whitelist is easier to reason about so let's do that :)

lighthouse-cli/test/cli/cli-flags-test.js
Hoten: `blockedUrlPattern` just b/c comma could be in a pattern? what if we could escape it?

Hoten: nvm lets not do that :)

lighthouse-cli/cli-flags.js
patrickhulce: oh like an infinite depth flatten, yeah gotcha :)

lighthouse-cli/cli-flags.js
```diff
@@ -152,6 +161,22 @@ function getFlags(manualArgv) {
       .default('enable-error-reporting', undefined) // Undefined so prompted by default
       .default('channel', 'cli')
       .check(/** @param {LH.CliFlags} argv */ (argv) => {
+        // ".middleware" does not exist in this version of yargs, so do some preprocessing here.
```
patrickhulce: still not 100% sure that I love mutating the arguments in a `.check` method 

WDYT of just saving the `.argv` to a variable and mutating *that* before we return in `getFlags`?

I think that might help make it more explicit that we are doing some non-standard yargs mutations

Hoten: done

lighthouse-cli/cli-flags.js
```diff
@@ -171,6 +181,24 @@ function getFlags(manualArgv) {
           'For more information on Lighthouse, see https://developers.google.com/web/tools/lighthouse/.')
       .wrap(yargs.terminalWidth())
       .argv;
+
+  // ".middleware" does not exist in this version of yargs, so do some post-processing here.
+  /** @type {(keyof LH.CliFlags)[]} */
```
patrickhulce: I think we're settling on `Array<keyof LH.CliFlags>` for complex types that aren't just `string[]` or w/e, though if you ask @brendankenny we're *always* settling on `Array<*>` ;)

brendankenny: > though if you ask @brendankenny we're always settling on Array<*> ;)

well, yes, because it's an array, not a string, so why would you put that information last? :P. They could have gone `[string]`, but they used that up on tuples

lighthouse-cli/cli-flags.js
```diff
@@ -12,19 +12,32 @@ const pkg = require('../package.json');
...
+ * @param {string[]} manualArgv
  * @return {LH.CliFlags}
  */
-function getFlags(manualArgv) {
-  // @ts-ignore yargs() is incorrectly typed as not returning itself
-  const y = manualArgv ? yargs(manualArgv) : yargs;
-  return y.help('help')
+function getFlags(...manualArgv) {
+  const y = manualArgv.length ? yargs(manualArgv) : yargs;
+  /** @type {LH.CliFlags} */
```
brendankenny: this doesn't buy us much except having to coerce with `/** @type {string[]} */` below

lighthouse-cli/cli-flags.js
```diff
@@ -12,19 +12,32 @@ const pkg = require('../package.json');
 const printer = require('./printer');
 
 /**
- * @param {string=} manualArgv
+ * @param {(string | string[])[]} arr
+ * @return {string[]}
```
brendankenny: isn't this simpler, just
```js
@param {Array<Array<string>>}
@return {Array<string>}
```
(`str.split(',')` always returns an array, even if the substring isn't found)

lighthouse-cli/cli-flags.js
```diff
@@ -12,19 +12,32 @@ const pkg = require('../package.json');
...
+/**
+ * @param {string[]} manualArgv
  * @return {LH.CliFlags}
  */
-function getFlags(manualArgv) {
-  // @ts-ignore yargs() is incorrectly typed as not returning itself
-  const y = manualArgv ? yargs(manualArgv) : yargs;
-  return y.help('help')
+function getFlags(...manualArgv) {
+  const y = manualArgv.length ? yargs(manualArgv) : yargs;
```
brendankenny: this ends up somewhat confusing from the caller's perspective since it's no longer explicitly optional (even though we usually call it with no args) and when we do pass an argument it's still a single string (it's annoying that the `yargs` types require this because [`yargs` definitely supports parsing just a string](https://github.com/yargs/yargs-parser/blob/981e1512d15a0af757a74a9cd75c448259aa8efa/index.js#L9-L11)).

What about keeping it `string=` and doing `const y = manualArgv ? yargs([manualArgv]) : yargs;` instead?

lighthouse-cli/cli-flags.js
```diff
@@ -12,19 +12,32 @@ const pkg = require('../package.json');
...
+  /** @type {LH.CliFlags} */
+  const argv = y.help('help')
       .version(() => pkg.version)
       .showHelpOnFail(false, 'Specify --help for available options')
 
       .usage('lighthouse <url> <options>')
       .example(
           'lighthouse <url> --view', 'Opens the HTML report in a browser after the run completes')
+      .example(
+          'lighthouse <url> --only-categories=performance,pwa',
```
brendankenny: kick somewhere below? I'd call `--only-categories` less important than `--config-path`, `--chrome-flags`, etc

lighthouse-cli/cli-flags.js
```diff
@@ -171,6 +184,24 @@ function getFlags(manualArgv) {
...
+    'onlyAudits',
+    'onlyCategories',
+    'output',
+    'plugins',
+    'skipAudits',
+  ];
+  arrayKeysThatSupportCsv.forEach(key => {
+    const input = /** @type {string[]} */ (argv[key]);
+    if (input) {
+      argv[key] = flatten(input.map(value => value.split(',')));
```
brendankenny: it's quite concerning this (`Array<string>`) is assignable back to `argv[key]`, which should be something like `number`|`boolean`|`string`|`Array<string>` not to mention our enums, etc...

The values are all there, it's just not flagging it...

brendankenny: also, `[].flatMap`, coming in Node 11 :)

lighthouse-cli/cli-flags.js
brendankenny: ha, actually, that test failure may be because our version of yargs doesn't take an array at all?

lighthouse-cli/cli-flags.js
brendankenny: > it's quite concerning this (Array<string>) is assignable back to argv[key], which should be something like number|boolean|string|Array<string> not to mention our enums, etc...

apparently this is a long-standing issue that will supposedly be fixed in https://github.com/microsoft/TypeScript/pull/30769

...but we can't run the current nightlies because they OOM on our code base, so I don't know for sure :)

We implicitly rely on this behavior in a few places in our codebase, so it'll be interesting to see how much breaks vs how much is still fine (last time I looked at that PR, our technically-unsound protocol typing was still going to work without error)

lighthouse-cli/cli-flags.js
Hoten: I'm confused on how it's confusing to the caller?

if reverting the change, would you want the code using multiple args (admittedly it's just test code) to look like this?

```
getFlags([
	'http://www.example.com',
	'--only-categories=performance,seo',
	'--skipAudits=unused-javascript,redirects',
	'--skipAudits=bootup-time',
].join(' '));
```

or this?
```
getFlags('http://www.example.com --only-categories=performance,seo --skipAudits=unused-javascript,redirects --skipAudits=bootup-time');
```

Hoten: It passes locally.

lighthouse-cli/cli-flags.js
Hoten: would like to keep `/** @type {string[]} */` otherwise more errors come up. I added a comment saying why we can assume it to be a string[].

`LH.CliFlags` one could be removed if you want.

lighthouse-cli/cli-flags.js
Hoten: ok the issue is that if you pass in an array, you cannot have each item have multiple args. test just needs to change

Hoten: also I guess they expect some preprocessing, b/c `=` is required too. https://github.com/GoogleChrome/lighthouse/pull/8933/commits/6f58b9007f2b9be45e7a399187b7d07b3d46554c

lighthouse-cli/cli-flags.js
brendankenny: > would like to keep /** @type {string[]} */ otherwise more errors come up.

but it's wrong because *all* of this is wrong :)

The ts-ignore was suppressing a different error than it said (I assume/hope it changed at some point), but the comment was still correct ('`yargs()` is incorrectly typed as not returning itself'): all the calls on yargs after `help()` are of type `any` and return `any`. In addition:
- the type of `LH.CliFlags[keyof LH.CliFlags]` is definitely *not* a string array, really it should be `LH.CliFlags['onlyAudits'|'onlyCategories'|'output'|'plugins'|'skipAudits']`
- if we pretend the type is `string[]` because yargs is returning that for the subset of keys we care about, assigning back as `string[]` is still not correct even for that subset because e.g. `output` is `LH.OutputMode[]`, not `string[]` (and that's ignoring trying to reassign back to `LH.CliFlags[keyof LH.CliFlags]`). This will be an error in tsc 3.5+.

I think the most honest thing is to just leave as `any` internal to `getFlags` with a note about why and leave it up to tests to verify that `getFlags` really is returning  `LH.CliFlags`.

lighthouse-cli/cli-flags.js
brendankenny: > if reverting the change, would you want the code using multiple args (admittedly it's just test code) to look like this?

yes from me. Maybe it's just that I haven't been completely indoctrinated to rest parameters yet, but to me an optional `manualArgv?: string` parameter is a lot clearer than `...manaulArgv: string[]` at informing that the argument is optional. It's not a huge deal since it's not a commonly used interface, but it's only test code that uses this, so it makes sense to me to make the real use case clear in the signature.

brendankenny: > also I guess they expect some preprocessing, b/c = is required too.

You can only omit the `=` when passing in as a single string? This seems like the real downside...we should be able to write tests using both styles since the real deal accepts both

Hoten:
@brendankenny updated

lighthouse-cli/cli-flags.js
```diff
@@ -11,14 +11,25 @@ const yargs = require('yargs');
...
+
 /**
  * @param {string=} manualArgv
  * @return {LH.CliFlags}
  */
 function getFlags(manualArgv) {
   // @ts-ignore yargs() is incorrectly typed as not returning itself
   const y = manualArgv ? yargs(manualArgv) : yargs;
-  return y.help('help')
+  const argv = y.help('help')
```
brendankenny: maybe add a comment about ``// Intentionally left as type `any` because @types/yargs doesn't chain correctly.`` or something just so it doesn't look like a glaring error later

lighthouse-cli/cli-flags.js
```diff
@@ -11,14 +11,25 @@ const yargs = require('yargs');
...
+  const result = [];
+  return result.concat(...arr);
+}
+
 /**
  * @param {string=} manualArgv
  * @return {LH.CliFlags}
  */
 function getFlags(manualArgv) {
   // @ts-ignore yargs() is incorrectly typed as not returning itself
```
brendankenny: Maybe update this one as well? `// @ts-ignore yargs() is incorrectly typed as not accepting a single string.`

brendankenny: honestly why are we even using `@types/yargs` :)

lighthouse-cli/cli-flags.js
```diff
@@ -171,6 +185,29 @@ function getFlags(manualArgv) {
...
+    'plugins',
+    'skipAudits',
+  ];
+  arrayKeysThatSupportCsv.forEach(key => {
+    // If a key is defined as an array in yargs, the value (if provided)
+    // will always be a string array. However, we keep argv and input as any,
+    // since assigning back to argv as string[] would be unsound for enums,
+    // for example: output is LH.OutputMode[].
+    const input = argv[key];
+    // Existence check + convinces TS that this is an array.
```
brendankenny: "Existence check + "...something else? Or no '+' needed

lighthouse-cli/cli-flags.js
```diff
@@ -171,6 +185,29 @@ function getFlags(manualArgv) {
           'For more information on Lighthouse, see https://developers.google.com/web/tools/lighthouse/.')
       .wrap(yargs.terminalWidth())
       .argv;
+
+  // ".middleware" does not exist in this version of yargs, so do some post-processing here.
```
brendankenny:
```suggestion
  // Support comma-separated values for some array flags by splitting on any ',' found.
```
(fine to be specific and we can generalize it if we add more post-processing in the future)

PR #8956, opened 2019-05-14
core(tracehouse): move files to core/lib/tracehouse/

patrickhulce:
**Summary**
https://github.com/aslushnikov/tracium is basically a port of our files with a few things stripped out that we're encouraging external folks to use when trying to make sense of a trace. Keeping them in sync is going to be a nightmare unless we can publish from the source of truth. This is step 1 towards that goal.

1. (This PR) Move the required files to a `lib/tracium` directory 
2. (Followup) Eliminate any dependencies from `lib/tracium` to the rest of LH
3. (Followup) Port back the patches in https://github.com/aslushnikov/tracium.
4. Retire https://github.com/aslushnikov/tracium / decide our monorepo strategy / do something to make external consumption *and* internal maintenance of tracium easier
5. Continue to make ongoing generic improvements as necessary

cc @aslushnikov

**IMPORTANT FOR MERGE** Before merging we want to keep at least two commits separate to preserve history, all the file moves are in `core(tracium): move files to lib/tracium/` commit which should allow us not to lose blame history, etc


aslushnikov:
> Retire https://github.com/aslushnikov/tracium / decide our monorepo strategy / do something to make external consumption and internal maintenance of tracium easier

@patrickhulce Yes! If tracium lives in Lighthouse repo (which is optimal imo) and publishes to tracium's npm, than we can just deprecate and archive the aslushnikov/tracium.

Very happy this is happening! 🎉

paulirish:
offline we discussed naming this tracehouse instead.

lighthouse-core/audits/bootup-time.js
```diff
@@ -7,7 +7,7 @@
 
 const Audit = require('./audit');
 const NetworkRequest = require('../lib/network-request');
-const {taskGroups} = require('../lib/task-groups');
+const {taskGroups} = require('../lib/tracehouse/task-groups');
```
brendankenny: .js

lighthouse-core/audits/mainthread-work-breakdown.js
```diff
@@ -11,7 +11,7 @@
 'use strict';
 
 const Audit = require('./audit');
-const {taskGroups} = require('../lib/task-groups');
+const {taskGroups} = require('../lib/tracehouse/task-groups');
```
brendankenny: .js :)

lighthouse-core/computed/trace-of-tab.js
```diff
@@ -1,191 +1,22 @@
 /**
- * @license Copyright 2017 Google Inc. All Rights Reserved.
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
```
brendankenny: no reason to update these?

brendankenny: oh, I guess the moved file keeps the year and this file is "new". Blame seems to work on the new files, but I'm curious what it'll do here given that the github PR UI doesn't see it that way

patrickhulce: yeah as long as we don't squash I think we're OK, when you look at a long history it won't do the smart diff, but blame and individual steps should be 👌

patrickhulce:
> Any ideas?

Monorepo where we actually depend on our own `@lighthouzzzzzzz/core` packages :)

brendankenny:
> Monorepo where we actually depend on our own @lighthouzzzzzzz/core packages :)

the files would still be named the same without changes to either the lib/ or computed artifact files themselves, though?

patrickhulce:
> the files would still be named the same without changes to either the lib/ or computed artifact files themselves, though?

Oh, I see what you were saying now. Why don't we start naming our computed `*-computed`? this has been a problem in metrics for a long time and it's definitely annoying :)

patrickhulce:
What's the final naming verdict?

Twitter poll time? 😈

paulirish:
brendan voiced a few more things:

* not great that we end up with a computed/trace-of-tab that's nearly empty. 
* we got 3 main-thread-tasks.js now. 
* between main-thread-tasks, tracing-processor and trace-of-tab, surely we can kill off one of those.

paulirish:
> What's the final naming verdict?

IMO tracehouse naming seems fine.

> Why don't we start naming our computed *-computed? 

agree. i'd love this. (separate PR, natch)

> not great that we end up with a computed/trace-of-tab that's nearly empty.

aye. feels like a codesmell. dunno what we can do about it without adding too much magic.


> we got 3 main-thread-tasks.js now.

audits could be called 'long-tasks' instead, but perhaps too late for that? the -computed fix would then disambiguate between the others..

> between main-thread-tasks, tracing-processor and trace-of-tab, surely we can kill off one of those.

ayup. i think patrick was waiting to do the move first, but that seems rather tempting.

patrickhulce:
> not great that we end up with a computed/trace-of-tab that's nearly empty.

I'm not quite as concerned about this thing. One's the meat and the other is the adaptation of our meat to the ultra-LH-specific "computed artifact" concept which as long as we have generic library type things is relatively unavoidable. If `tracehouse` were it's own package on NPM we require'd in I don't think we'd be that worried about having a lightweight `-computed` wrapper.

> audits could be called 'long-tasks' instead, but perhaps too late for that? the -computed fix would then disambiguate between the others..

The obvious step to me here is to parallel `-computed` and rename all our audits `*-audit` without changing their ids :) 

Maybe I'm misunderstanding the primary concern here, but it sounds like these are all problems because the files have the same name?

> i think patrick was waiting to do the move first, but that seems rather tempting.

Indeed I was, my plan for killing one of them is going to be controversial enough to warrant its own PR for discussion 😈

patrickhulce:
@brendankenny what do you need to see from me to get this merged? :)

(if all looks good I will go ahead and re-rebase to have just the two commits)

patrickhulce:
> It's a definite architecture smell.

Ooooooooooh if we're just making commentary on the fact that it's not ideal that our repo is organized this way, then you know I wholeheartedly agree. But you already know my thoughts on this so no need for me to get the boxing gloves out 😉

aslushnikov:
yay!! 🎉

What are the next steps?

patrickhulce:
@aslushnikov here's the roadmap, probably 2 PRs away from matching current tracium :)

1. Eliminate any dependencies from `lib/tracium` to the rest of LH
3. Port back the patches in https://github.com/aslushnikov/tracium.
4. Retire https://github.com/aslushnikov/tracium / decide our monorepo strategy / do something to make external consumption *and* internal maintenance of tracium easier
5. Continue to make ongoing generic improvements as necessary

PR #8961, opened 2019-05-15
core(report): show nodeLabel for DOM nodes in addition to snippet

mattzeunert:
**Summary**

Collect a human-readable title for each node and display it above the snippet.

Screenshot 2019-05-16 at 14 41 28

**Related Issues/PRs**

https://github.com/GoogleChrome/lighthouse/issues/6683

### DIFF
--- a/lighthouse-cli/test/smokehouse/seo/expectations.js
+++ b/lighthouse-cli/test/smokehouse/seo/expectations.js
@@ -205,6 +205,7 @@ module.exports = [
                  '\n        too small target\n      </a>',
                   'path': '2,HTML,1,BODY,3,DIV,21,DIV,0,A',
                   'selector': 'body > div > div > a',
+                  'title': 'too small target',
                 },
                 'overlappingTarget': {
                   'type': 'node',
@@ -213,6 +214,7 @@ module.exports = [
                   '\n        big enough target\n      </a>',
                   'path': '2,HTML,1,BODY,3,DIV,21,DIV,1,A',
                   'selector': 'body > div > div > a',
+                  'title': 'big enough target',
                 },
                 'size': '100x30',
                 'width': 100,
--- a/lighthouse-cli/test/smokehouse/seo/expectations.js
+++ b/lighthouse-cli/test/smokehouse/seo/expectations.js
@@ -205,7 +205,7 @@ module.exports = [
                  '\n        too small target\n      </a>',
                   'path': '2,HTML,1,BODY,3,DIV,21,DIV,0,A',
                   'selector': 'body > div > div > a',
-                  'title': 'too small target',
+                  'nodeLabel': 'too small target',
                 },
                 'overlappingTarget': {
                   'type': 'node',
@@ -214,7 +214,7 @@ module.exports = [
                   '\n        big enough target\n      </a>',
                   'path': '2,HTML,1,BODY,3,DIV,21,DIV,1,A',
                   'selector': 'body > div > div > a',
-                  'title': 'big enough target',
+                  'nodeLabel': 'big enough target',
                 },
                 'size': '100x30',
                 'width': 100,
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -228,13 +228,13 @@ function getNodeTitle(node) {
     // too broad, contains all page content
     return null;
   }
-  let title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
+  const title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
   if (title) {
     return truncate(title, 80);
   } else {
     const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
     if (nodeToUseForTitle) {
-      title = getNodeTitle(nodeToUseForTitle);
+      return getNodeTitle(nodeToUseForTitle);
     }
   }
 }
@@ -264,6 +264,7 @@ module.exports = {
   getNodePathString: getNodePath.toString(),
   getNodeSelectorString: getNodeSelector.toString(),
   getNodeSelector: getNodeSelector,
+  getNodeTitle: getNodeTitle,
   getNodeTitleString: getNodeTitle.toString(),
   truncateString: truncate.toString(),
 };
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -219,8 +219,8 @@ function getNodeSelector(node) {
 }
 
 /**
- * @param {Element} node
- * @returns {string}
+ * @param {HTMLElement} node
+ * @returns {string|null}
  */
 /* istanbul ignore next */
 function getNodeTitle(node) {
@@ -234,9 +234,10 @@ function getNodeTitle(node) {
   } else {
     const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
     if (nodeToUseForTitle) {
-      return getNodeTitle(nodeToUseForTitle);
+      return getNodeTitle(/** @type {HTMLElement} */ (nodeToUseForTitle));
     }
   }
+  return null;
 }
 
 /**
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -218,6 +218,38 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
+ * @param {Element} node
+ * @returns {string}
+ */
+/* istanbul ignore next */
+function getNodeTitle(node) {
+  let title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
+  if (!title) {
+    const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
+    if (nodeToUseForTitle) {
+      title = nodeToUseForTitle.getAttribute('alt') || nodeToUseForTitle.getAttribute('aria-label');
+    }
+  }
+  if (title) {
+    title = truncate(title, 80);
+  }
+  return title;
+}
+
+/**
+ * @param {string} str
+ * @param {number} maxLength
+ * @returns {string}
+ */
+/* istanbul ignore next */
+function truncate(str, maxLength) {
+  if (str.length <= maxLength) {
+    return str;
+  }
+  return str.slice(0, maxLength - 1) + '…';
+}
+
 module.exports = {
   wrapRuntimeEvalErrorInBrowserString: wrapRuntimeEvalErrorInBrowser.toString(),
   registerPerformanceObserverInPageString: registerPerformanceObserverInPage.toString(),
@@ -230,4 +262,6 @@ module.exports = {
   getNodePathString: getNodePath.toString(),
   getNodeSelectorString: getNodeSelector.toString(),
   getNodeSelector: getNodeSelector,
+  getNodeTitleString: getNodeTitle.toString(),
+  truncateString: truncate.toString(),
 };
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -227,6 +227,18 @@ function getNodeSelector(node) {
  */
 /* istanbul ignore next */
 function getNodeLabel(node) {
+  /**
+   * @param {string} str
+   * @param {number} maxLength
+   * @return {string}
+   */
+  function truncate(str, maxLength) {
+    if (str.length <= maxLength) {
+      return str;
+    }
+    return str.slice(0, maxLength - 1) + '…';
+  }
+
   const tagName = node.tagName.toLowerCase();
   // html and body content is too broad to be useful, since they contain all page content
   if (tagName !== 'html' && tagName !== 'body') {
@@ -245,19 +257,6 @@ function getNodeLabel(node) {
   return tagName;
 }
 
-/**
- * @param {string} str
- * @param {number} maxLength
- * @return {string}
- */
-/* istanbul ignore next */
-function truncate(str, maxLength) {
-  if (str.length <= maxLength) {
-    return str;
-  }
-  return str.slice(0, maxLength - 1) + '…';
-}
-
 module.exports = {
   wrapRuntimeEvalErrorInBrowserString: wrapRuntimeEvalErrorInBrowser.toString(),
   registerPerformanceObserverInPageString: registerPerformanceObserverInPage.toString(),
@@ -272,5 +271,4 @@ module.exports = {
   getNodeSelector: getNodeSelector,
   getNodeLabel: getNodeLabel,
   getNodeLabelString: getNodeLabel.toString(),
-  truncateString: truncate.toString(),
 };
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -224,20 +224,19 @@ function getNodeSelector(node) {
  */
 /* istanbul ignore next */
 function getNodeTitle(node) {
-  if (node.tagName === 'HTML' || node.tagName === 'BODY') {
-    // too broad, contains all page content
-    return null;
-  }
-  const title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
-  if (title) {
-    return truncate(title, 80);
-  } else {
-    const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
-    if (nodeToUseForTitle) {
-      return getNodeTitle(/** @type {HTMLElement} */ (nodeToUseForTitle));
+  const tagName = node.tagName.toLowerCase();
+  if (tagName !== 'html' && tagName !== 'body') {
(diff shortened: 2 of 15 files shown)


lighthouse-core/report/html/report-styles.css
```diff
@@ -388,10 +388,15 @@
...
 /* Node */
-.lh-node {
-  display: block;
+.lh-node__snippet {
   font-family: var(--monospace-font-family);
   color: hsl(174, 100%, 27%);
+  font-size: 12px;
+}
+.lh-node__title {
+  color: black;
```
Hoten: click the icon in the top left to toggle dark mode. use `--body-text-color` instead

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
+ * @param {string} str
+ * @param {number} maxLength
+ * @returns {string}
+ */
+/* istanbul ignore next */
+function truncate(str, maxLength) {
+  if (str.length <= maxLength) {
+    return str;
+  }
+  return str.slice(0, maxLength - 1) + '…';
```
Hoten: could you add some tests for this edge case behavior?

`truncate('123', 3) === '123'`
`truncate('1234', 3) === '123…'`

or w/e way you expect truncate to work.

lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -213,6 +214,7 @@ module.exports = [
                   '\n        big enough target\n      </a>',
                   'path': '2,HTML,1,BODY,3,DIV,21,DIV,1,A',
                   'selector': 'body > div > div > a',
```
brendankenny: can you add to the `a11y` expectations as well?

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
+ * @param {Element} node
+ * @returns {string}
```
brendankenny: `@return` here (and elsewhere) (sorry)

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
+ */
+/* istanbul ignore next */
+function getNodeTitle(node) {
+  if (node.tagName === 'HTML' || node.tagName === 'BODY') {
+    // too broad, contains all page content
+    return null;
```
brendankenny: supposedly returns `string` but we have type checking turned off for this file :)

You could turn it on, fix any errors in the new code, then turn it back off.

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
+  }
+  const title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
+  if (title) {
+    return truncate(title, 80);
+  } else {
+    const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
+    if (nodeToUseForTitle) {
+      return getNodeTitle(nodeToUseForTitle);
+    }
+  }
```
brendankenny: intentionally `undefined` if this all falls through?

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
+ * @param {Element} node
+ * @returns {string}
+ */
+/* istanbul ignore next */
+function getNodeTitle(node) {
+  if (node.tagName === 'HTML' || node.tagName === 'BODY') {
```
brendankenny: it seems like there's something nice we could print for these...could we just always have a lowercase `tagName` as the fallback (here and if nothing is found below)? Or does that end up looking bad because it's right above `<body>` anyways

types/artifacts.d.ts
```diff
@@ -145,6 +145,7 @@ declare global {
             snippet: string;
             target: string[];
             failureSummary?: string;
+            title?: string;
```
brendankenny: is there a different word we could use? `title` seems very particular to html rendering rather than describing what the value is...but I don't have a good suggestion. `snippet` is already taken :) `description`? `summary`? ...

patrickhulce: +1 to using a different name, I kept thinking this was the `title` attribute until I opened the PR 😆

lighthouse-core/lib/page-functions.js
mattzeunert: Works for me, what do you think?

Screenshot 2019-05-16 at 14 25 49

lighthouse-core/lib/page-functions.js
mattzeunert: Yeah, explicitly returning null now.

types/artifacts.d.ts
mattzeunert: `snippetTitle`?

lighthouse-cli/test/smokehouse/seo/expectations.js
mattzeunert: Right now all we assert there is `{ length: 1 }`, but I'll change it when I do the final artifact update for this PR.

mattzeunert:
> I wonder if there's more we could do to tie the new text and the snippet together visually, though

Removed some unnecessary margin between the two. Not sure what else to do exactly, @hwikyounglee any suggestions?

types/artifacts.d.ts
patrickhulce: That could work. Just brainstorming some other ideas

`nodeText`
`nodeLabel`
`textLabel`
`textContentOrAriaLabel` :)

mattzeunert: I quite like `nodeLabel`, if we're fine with `node.nodeLabel`.

patrickhulce: I'm down with that. A prefix seems almost necessary to signify that it isn't the literal `label` the node but some derived thing that's label-like, so 👍 

@brendankenny any thoughts on those?

hwikyounglee:
> > I wonder if there's more we could do to tie the new text and the snippet together visually, though
> 
> Removed some unnecessary margin between the two. Not sure what else to do exactly, @hwikyounglee any suggestions?

Thanks Matt. Only thing I would double check is the contrast of the green text on gray and on white. We need to aim for 4.5 at minimum. 

@yuinchien for any further guidance

mattzeunert:
> Only thing I would double check is the contrast of the green text on gray and on white. We need to aim for 4.5 at minimum.

Good point. We weren't using a var for it before, using `--color-green-700` for now:

Screenshot 2019-05-16 at 17 11 54

lighthouse-cli/test/smokehouse/seo/expectations.js
mattzeunert: Added `details.items` assertions.

types/artifacts.d.ts
brendankenny: > A prefix seems almost necessary to signify that it isn't the literal label the node but some derived thing that's label-like, so 👍

this makes sense to me too, `nodeLabel` sounds good

lighthouse-cli/test/smokehouse/seo/expectations.js
brendankenny: > Right now all we assert there is { length: 1 }, but I'll change it when I do the final artifact update for this PR.

haha, uh oh, I guess I didn't know how much I was asking for :) I guess we can just cross our fingers that these are stable and all worth testing. I do really like the extra coverage this gets us

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
 
+/**
+ * @param {HTMLElement} node
+ * @return {string|null}
+ */
+/* istanbul ignore next */
+function getNodeLabel(node) {
+  const tagName = node.tagName.toLowerCase();
+  if (tagName !== 'html' && tagName !== 'body') {
+    const title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
```
brendankenny: not a big deal, but I guess s/title/label

brendankenny: also should use the standard `node.textContent` (unless `innerText` was intentional?)

lighthouse-core/report/html/renderer/details-renderer.js
```diff
@@ -352,15 +352,24 @@ class DetailsRenderer {
    */
   renderNode(item) {
     const element = this._dom.createElement('span', 'lh-node');
+    if (item.nodeLabel) {
+      const titleEl = this._dom.createElement('div');
```
brendankenny: `labelEl`

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -59,4 +59,31 @@ describe('Page Functions', () => {
...
     });
   });
+
+  describe('getNodeLabel', () => {
+    it('Returns innerText if element has visible text', () => {
+      const el = dom.createElement('div');
+      el.innerText = 'Hello';
+      assert.equal(pageFunctions.getNodeLabel(el), 'Hello');
+    });
+    it('Falls back to children and alt/aria-label if a title can\'t be determined', () => {
```
brendankenny: nit: newline between the `it()` blocks

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
```
brendankenny: can you add a comment for this function? Basically trying to provide a human recognizable label for the given element. Falls back to the tagName if no useful label is found.

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
+ * @param {HTMLElement} node
+ * @return {string|null}
+ */
+/* istanbul ignore next */
+function getNodeLabel(node) {
+  const tagName = node.tagName.toLowerCase();
+  if (tagName !== 'html' && tagName !== 'body') {
```
brendankenny: can you add back that comment about why we skip html and body

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,40 @@ function getNodeSelector(node) {
...
+    const title = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
+    if (title) {
+      return truncate(title, 80);
+    } else {
+      const nodeToUseForTitle = node.querySelector('[alt], [aria-label]');
```
brendankenny: maybe add comment about why it's useful to recursively search (what cases does this help with?)

types/audit-details.d.ts
```diff
@@ -161,6 +161,7 @@ declare global {
         path?: string;
         selector?: string;
         snippet?: string;
+        nodeLabel?: string;
```
brendankenny: also add a comment here?

`/** A text descriptor for the node intended to help quickly recognize which node is being referenced. */`

or something

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -59,4 +59,31 @@ describe('Page Functions', () => {
...
+    });
+    it('Uses tag name for html tags', () => {
+      const el = dom.createElement('html');
+      assert.equal(pageFunctions.getNodeLabel(el), 'html');
+    });
+    it('Uses tag name if there is no better label', () => {
+      const el = dom.createElement('div');
+      assert.equal(pageFunctions.getNodeLabel(el), 'div');
+    });
+  });
```
brendankenny: also add a test for the recursive case?

lighthouse-core/lib/page-functions.js
patrickhulce: I specifically requested innerText because `textContent` gets us all that inline script instead of actual text

lighthouse-core/test/lib/page-functions-test.js
mattzeunert: We already have the `Falls back to children` test case above, do you mean checking it falls back to the parent tag name if neither the parent nor any children are useful? I've modified the test case to check that.

lighthouse-core/report/html/report-styles.css
```diff
@@ -100,6 +100,8 @@
   --color-white: #FFFFFF;
   --color-blue-200: #90CAF9;
   --color-blue-900: #0D47A1;
+  --color-teal-300: #00d6c1;
+  --color-teal-700: #008577;
```
mattzeunert: Re-introduced teal since it was a bit nicer than the green. Do the numbers at the end have a specific meaning, other than broadly identifying darkness?

patrickhulce: It's the shade in the material design color palette https://material.io/design/color/#

lighthouse-core/test/lib/page-functions-test.js
brendankenny: > We already have the `Falls back to children` test case above, do you mean checking it falls back to the parent tag name if neither the parent nor any children are useful? I've modified the test case to check that.

whoops, no, that's me not reading carefully enough. I like the new change too, though!

lighthouse-core/report/html/report-styles.css
brendankenny: is this the right number, then?

mattzeunert: Changed it to use material design colors.

Also cut the snippet line height a bit.

Screenshot 2019-05-20 at 20 31 26
Screenshot 2019-05-20 at 20 31 33

lighthouse-cli/test/smokehouse/seo/expectations.js
patrickhulce: When @mattzeunert does something, he does it right, and you get the works! 😎

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,46 @@ function getNodeSelector(node) {
...
+ * @return {string|null}
+ */
+/* istanbul ignore next */
+function getNodeLabel(node) {
+  const tagName = node.tagName.toLowerCase();
+  // html and body content is too broad to be useful, since they contain all page content
+  if (tagName !== 'html' && tagName !== 'body') {
+    const nodeLabel = node.innerText || node.getAttribute('alt') || node.getAttribute('aria-label');
+    if (nodeLabel) {
+      return truncate(nodeLabel, 80);
```
patrickhulce: I was making this comment over in #8979 , but it's unfortunate when page-functions in a lib have dependencies on each other that you have to know about ahead of time. 

since it's basically a single ternary line WDYT about copying the function definition in the two files instead?

lighthouse-core/report/html/report-styles.css
```diff
@@ -175,6 +178,7 @@
 .lh-vars.dark {
   --color-red-700: var(--color-red);
   --color-green-700: var(--color-green);
+  --color-teal-600: var(--color-cyan-500);
```
patrickhulce: dark mode is weird o.O

brendankenny:
> I don't love the truncate situation, but I won't block an otherwise approved merge because of it either.

> I was making this comment over in #8979 , but it's unfortunate when page-functions in a lib have dependencies on each other that you have to know about ahead of time.

with this, tap-targets, iframeelements, and...other stuff(?) making changes to `page-functions` in current PRs, my vote is definitely to make things simple rather than have to come back later and try to untangle (which will be really hard because debugging `evaluateAsync` functions in gatherers is really hard)

lighthouse-core/lib/page-functions.js
mattzeunert: Are you thinking have it in `tap-targets.js` and also inlined inside `getNodeLabel`?

mattzeunert:
Maybe `fnString` shouldn't just be `fn.toString()` but also all its dependencies?

lighthouse-core/lib/page-functions.js
patrickhulce: Yeah that was my thought

patrickhulce:
> Maybe the exported fnString shouldn't just be fn.toString() but also all its dependencies?

Yeah that's a good idea, though perhaps dupes might be an issue then? Since the function is so simple here I think inline is decent option, but we'll have to figure this out long-term for sure either.

mattzeunert:
> > Maybe the exported fnString shouldn't just be fn.toString() but also all its dependencies?
> 
> Yeah that's a good idea, though perhaps dupes might be an issue then? Since the function is so simple here I think inline is decent option, but we'll have to figure this out long-term for sure either.

Is there an issue with dupes other than aesthetics? There'd be a perf impact but that seems very minor.

patrickhulce:
> Is there an issue with dupes other than aesthetics?

I was thinking of variable declarations not function declarations, but I suppose all our dupes would be of function declarations, so it's a non-issue. 👍

Issue #8955, opened 2019-05-14
"Show 3rd-party resources" checkbox should be hidden when all resources are 3rd party. 


jburger424:
**Feature request summary**
Similar to how the checkbox is not shown when nothing is 3rd-party, it shouldn't be shown when everything is 3rd-party.

This is specifically applicable to our plugin as we have several audits that will only display 3rd party resources. In that situation it seems to make no sense to provide the user a way to clear the table.

I'd be happy to implement this myself.

brendankenny:
Makes perfect sense to me

> I'd be happy to implement this myself.

👍 sounds great!

Hoten:
sgtm

> This is specifically applicable to our plugin as we have several audits that will only display 3rd party resources. In that situation it seems to make no sense to provide the user a way to clear the table.

We probably shouldn't have [hardcoded the exceptions](https://github.com/GoogleChrome/lighthouse/blob/ffb1debfb6ed909c2d44b9c3cd578a5b6d17da8d/lighthouse-core/report/html/renderer/report-ui-features.js#L159) :)

jburger424:
> We probably shouldn't have [hardcoded the exceptions](https://github.com/GoogleChrome/lighthouse/blob/ffb1debfb6ed909c2d44b9c3cd578a5b6d17da8d/lighthouse-core/report/html/renderer/report-ui-features.js#L159) :)

Haha missed that. What if the exceptions were part of the config file?

brendankenny:
as you noted, there's [already a check for not adding it if there's no third-party URLs](https://github.com/GoogleChrome/lighthouse/blob/ffb1debfb6ed909c2d44b9c3cd578a5b6d17da8d/lighthouse-core/report/html/renderer/report-ui-features.js#L177-L178), it seems completely reasonable  and nicely simple to add a check right next to there for when a table is all third-party URLs.

(and if `uses-rel-preconnect` just turns out to be a specific case of this as the comment above it implies, we could remove the `thirdPartyFilterAuditExclusions`, but I don't think that's a pressing concern)

jburger424:
Agreed, will just stick with this for now.

PR #8966, opened 2019-05-15
report: hide 3rd party filter checkbox if all rows are 3rd party

jburger424:
**Related Issues/PRs**
Resolves https://github.com/GoogleChrome/lighthouse/issues/8955

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -173,9 +173,10 @@ class ReportUIFeatures {
       });
 
     tablesWithUrls.forEach((tableEl, index) => {
+      const numRows = this._getUrlItems(tableEl).length;
       const thirdPartyRows = this._getThirdPartyRows(tableEl, this.json.finalUrl);
-      // No 3rd parties, no checkbox!
-      if (!thirdPartyRows.size) return;
+      // If all or none of the rows are 3rd party, no checkbox!
+      if (thirdPartyRows.size === numRows || !thirdPartyRows.size) return;
 
       // create input box
       const filterTemplate = this._dom.cloneTemplate('#tmpl-lh-3p-filter', this._document);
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -173,9 +173,10 @@ class ReportUIFeatures {
       });
 
     tablesWithUrls.forEach((tableEl, index) => {
+      const numRows = this._getUrlItems(tableEl).length;
       const thirdPartyRows = this._getThirdPartyRows(tableEl, this.json.finalUrl);
-      // No 3rd parties, no checkbox!
-      if (!thirdPartyRows.size) return;
+      // If all or none of the rows are 3rd party, no checkbox!
+      if (thirdPartyRows.size === numRows || !thirdPartyRows.size) return;
```
patrickhulce: I think we should be able to nuke `thirdPartyFilterAuditExclusions` with just this :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -219,7 +220,7 @@ class ReportUIFeatures {
    * @return {Map<number, HTMLTableRowElement>}
    */
   _getThirdPartyRows(el, finalUrl) {
```
patrickhulce: maybe we should just pass the url items in then since it's happening the line before?

lighthouse-core/report/html/renderer/report-ui-features.js
jburger424: I've found a few cases where this doesn't work, where resources are being loaded from a different subdomain on the same domain.

patrickhulce: Ah yeah if they're intermixed it would be weird to show it good call 👍

brendankenny:
and actually, that "no filter" test is just for `thirdPartyFilterAuditExclusions`, not for if there are no third party URLs, so if it's easy would you mind adding that as well? 🥇 🥇

PR #8979, opened 2019-05-17
core(iframe-elements): Include new IFrameElements gatherer

jburger424:
**Summary**
This moves https://github.com/googleads/pub-ads-lighthouse-plugin/blob/master/lighthouse-plugin-ad-speed-insights/gatherers/iframe-elements.js to core Lighthouse repo.

**Related Issues/PRs**
This resolves https://github.com/googleads/pub-ads-lighthouse-plugin/issues/58.


jburger424:
TODO: unit tests

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+'use strict';
+
+const Gatherer = require('./gatherer.js');
+const pageFunctions = require('../../lib/page-functions.js');
+
+/* eslint-env browser, node */
+
+/**
+ * @return {LH.Artifacts['IFrameElements']}
+ */
```
patrickhulce: you'll need `/* istanbul ignore next */` here, our tests instrument these functions for coverage if we don't

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,63 @@ function getNodeSelector(node) {
...
+  }
+  return false;
+}
+
+/**
+ * @param {?HTMLElement} element
+ * @return {Boolean}
+ */
+/* istanbul ignore next */
+function isFixed(element) {
```
patrickhulce: WDYT about `isPositionFixed`?

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+    // @ts-ignore
+    const clientRect = getClientRect(node).toJSON();
+    // Marking 1x1 as non-visible to ignore tracking pixels.
+    const isVisible = (clientRect.width > 1 && clientRect.height > 1);
+    return {
+      id: node.id,
+      src: node.src,
+      clientRect,
+      isVisible,
+      // @ts-ignore
```
patrickhulce:
```suggestion
      // @ts-ignore - put into scope via stringification
```

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+      isFixed: isVisible && isFixed(node),
+    };
+  });
+}
+
+/** @inheritdoc */
```
patrickhulce: TIL 😃 

what effects does this end up having for us? at least in vscode I seem to get the same docstring without it, but curious what else takes it into account

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+   * @return {Promise<LH.Artifacts['IFrameElements']>}
+   * @override
+   */
+  async afterPass(passContext) {
+    const driver = passContext.driver;
+
+    const {frameTree} = await driver.sendCommand('Page.getFrameTree');
+    const framesByDomId = new Map();
+    if (frameTree.childFrames) {
+      for (const {frame} of frameTree.childFrames) {
```
patrickhulce: maybe make a comment about why we're not recursing?

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,63 @@ function getNodeSelector(node) {
...
+  }
+  return false;
+}
+
+/**
+ * @param {HTMLIFrameElement} element
+ * @return {DOMRect | ClientRect}
+ */
+/* istanbul ignore next */
+function getClientRect(element) {
```
patrickhulce: this seems a bit unnecessary IMO 

it's exactly `getBoundingClientRect` and adding more in page functions is a pain as you've no doubt discovered for yourself :)

lighthouse-core/lib/page-functions.js
patrickhulce: also not great that the caller has to know to bring along all of these other functions, we tried to keep that isolated to just the tap-targets audit which has so much complexity in it already.

others may disagree but since we're not reusing these anywhere else right now, WDYT about defining these helper functions inside `isFixed` so we don't need to think about those pieces? :)

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * @return {LH.Artifacts['IFrameElements']}
+ */
+function collectIFrameElements() {
+  // @ts-ignore - put into scope via stringification
+  const iFrameElements = getElementsInDocument('iframe'); // eslint-disable-line no-undef
+  return iFrameElements.map(/** @param {HTMLIFrameElement} node */ (node) => {
+    // @ts-ignore
+    const clientRect = getClientRect(node).toJSON();
```
patrickhulce:
```suggestion
    const clientRect = node.getBoundingClientRect().toJSON();
```

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,78 @@
...
+      ${pageFunctions.hasScrollableAncestorString};
+      ${pageFunctions.isFixedString};
+      return (${collectIFrameElements})();
+    })()`;
+
+    /** @type {LH.Artifacts['IFrameElements']} */
+    const elements =
+        await driver.evaluateAsync(expression, {useIsolation: true});
+    for (const el of elements) {
+      el.frame = framesByDomId.get(el.id);
```
patrickhulce: should we throw if we can't find it? leave as optional on the typedef?

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
...
+        /** The `id` attribute of the IFrame. */
+        id: string,
+        /** The `src` attribute of the IFrame. */
+        src: string,
+        /** The IFrame's ClientRect. @see https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect */
+        clientRect: ClientRect,
+        /** The visibility of the IFrame. */
+        isVisible: boolean,
+        /** If the IFrame or an ancestor of the IFrame is fixed in position. */
+        isFixed: boolean,
```
patrickhulce: same `isPositionFixed` comment

lighthouse-core/lib/page-functions.js
jburger424: Yeah, completely unnecessary, removed.

lighthouse-core/gather/gatherers/iframe-elements.js
warrengm: After reading your comment it should recuse. I didn't realize that `getElementsInDoc` looks at nested frames by default: https://paulirish.github.io/lighthouse/docs/api/lighthouse/2.5.1/Driver.html#getElementsInDocument

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,63 @@ function getNodeSelector(node) {
...
+      }
+    }
+    currentEl = currentEl.parentElement;
+  }
+  return false;
+}
+
+/**
+ * @param {?HTMLElement} element
+ * @return {Boolean}
```
warrengm: Any reason not to use primitive types (`string` and `boolean` vs `String` and `Boolean`)? here and elsewhere

lighthouse-core/gather/gatherers/iframe-elements.js
jburger424: TBH this was part of the gatherer when I started on the project and it was just never evaluated. Removed for consistency.

lighthouse-core/gather/gatherers/iframe-elements.js
jburger424: @warrengm will leave this for you to decide

brendankenny:
for the travis failure, will need to update the `index-test.js.snap` snapshot with `yarn run unit-cli -u` to include the new gatherer.

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
         params: {name: string; value: string}[];
       }
 
+      export interface IFrameElement {
+        /** The `id` attribute of the IFrame. */
+        id: string,
+        /** The `src` attribute of the IFrame. */
+        src: string,
+        /** The IFrame's ClientRect. @see https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect */
+        clientRect: ClientRect,
```
brendankenny: we've had issues with putting some of the real DOM types in artifacts because of the extra constraints sometimes put on them, e.g. for `ClientRect`, `width` and `height` are marked `readonly`, which can pose an annoyance e.g. in the type-checked tests, scripts that overwrite properties on artifacts, etc.

Since it's not really a `ClientRect` anyways (we pass it through JSON via the protocol), can you switch this to a manually specified object like `ImageElement.clientRect` does (though with `width` and `height`, since I vaguely recall that you do use those)?

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
...
+        /** The visibility of the IFrame. */
+        isVisible: boolean,
+        /** If the IFrame or an ancestor of the IFrame is fixed in position. */
+        isPositionFixed: boolean,
+        /** The Frame of the IFrame. @see https://chromedevtools.github.io/devtools-protocol/tot/Page#type-Frame */
+        frame: Crdp.Page.Frame,
```
brendankenny: what's this used for? It seems like a lot of it isn't going to be usable after the browser connection is closed.

types/artifacts.d.ts
```diff
@@ -59,6 +59,8 @@ declare global {
     export interface PublicGathererArtifacts {
       /** Console deprecation and intervention warnings logged by Chrome during page load. */
       ConsoleMessages: Crdp.Log.EntryAddedEvent[];
+      /** All the IFrame elements in the page.*/
```
brendankenny:
```suggestion
      /** All the iframe elements in the page. */
```

lowercase? I always think `IFrame` looks awkward but maybe that's just me :)

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,52 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
+ * @param {?HTMLElement} element
```
brendankenny: why nullable? I don't believe it needs to be (maybe originally called recursively and no longer necessary?)

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,52 @@ function getNodeSelector(node) {
   return parts.join(' > ');
 }
 
+/**
```
brendankenny: can you add an explanatory comment for the function? From just the name I would assume it's just checking if the element itself is `position: fixed`, so a sentence explaining why you might need to check for more would be 👍

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * @param {LH.Gatherer.PassContext} passContext
+   * @return {Promise<LH.Artifacts['IFrameElements']>}
+   * @override
+   */
+  async afterPass(passContext) {
+    const driver = passContext.driver;
+
+    const {frameTree} = await driver.sendCommand('Page.getFrameTree');
+    const framesByDomId = new Map();
```
brendankenny: add a type here, otherwise it's `Map<any, any>`. Annoyingly this doesn't get flagged with `--no-implicit-any`

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+    if (frameTree.childFrames) {
+      for (const {frame} of frameTree.childFrames) {
+        if (framesByDomId.has(frame.name)) {
+          // DOM ID collision, mark it as null.
```
brendankenny: can you add the reason to mark it null?

lighthouse-core/gather/gatherers/iframe-elements.js
brendankenny: yeah, when the map isn't a `Map<any, any>`, this choice will have to be made explicit in the `IFrameElement` artifact type

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const expression = `(() => {
+      ${pageFunctions.getOuterHTMLSnippetString};
+      ${pageFunctions.getElementsInDocumentString};
+      ${pageFunctions.isPositionFixedString};
+      return (${collectIFrameElements})();
+    })()`;
+
+    /** @type {LH.Artifacts['IFrameElements']} */
+    const elements =
```
brendankenny: `const iframeElements` or whatever

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+        await driver.evaluateAsync(expression, {useIsolation: true});
```
brendankenny: total nit: we use 100 character line lengths, so this doesn't have to be on a new line

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @return {LH.Artifacts['IFrameElements']}
+ */
+/* istanbul ignore next */
+function collectIFrameElements() {
+  // @ts-ignore - put into scope via stringification
+  const iFrameElements = getElementsInDocument('iframe'); // eslint-disable-line no-undef
+  return iFrameElements.map(/** @param {HTMLIFrameElement} node */ (node) => {
+    // @ts-ignore - put into scope via stringification
+    const clientRect = node.getBoundingClientRect().toJSON(); // eslint-disable-line no-undef
```
brendankenny: this is about to go back over the protocol via JSON, so are the explicit `.toJSON()` and `// @ts-ignore` necessary?

brendankenny: (and if it is necessary, need to remove the `eslint-disable-line` and update the `ts-ignore` comment. It's really that `getBoundingClientRect` will always return a `DOMRect` since this code only ever runs in Chrome, so `toJSON` will always be defined)

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
         params: {name: string; value: string}[];
       }
 
+      export interface IFrameElement {
+        /** The `id` attribute of the IFrame. */
```
brendankenny: I feel like people usually always go lowercase, so my vote is for `iframe` for all these comments (though variable/type names are fine)

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
...
+        id: string,
+        /** The `src` attribute of the IFrame. */
+        src: string,
+        /** The IFrame's ClientRect. @see https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect */
+        clientRect: ClientRect,
+        /** The visibility of the IFrame. */
```
brendankenny: can you update this to be more specific (e.g. mention the size requirement)

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,75 @@
...
+    // Marking 1x1 as non-visible to ignore tracking pixels.
+    const isVisible = (clientRect.width > 1 && clientRect.height > 1);
+    return {
+      id: node.id,
+      src: node.src,
+      clientRect,
+      isVisible,
```
brendankenny: should this be renamed? This would catch `display: none` (maybe should mention this in the comment over `isVisible` declaration), but not `opacity` or if it's occluded by other elements, etc. What's the use of the property, exactly? (maybe "is this a frame we might care about"?)

lighthouse-core/gather/gatherers/iframe-elements.js
patrickhulce: I think those were leftover from when it was a helper function :)

the `toJSON` definitely used to be necessary because protocol didn't serialize the same way, but IIRC someone was saying it might be unnecessary these days? worth checking either way :)

types/artifacts.d.ts
```diff
@@ -179,6 +181,21 @@ declare global {
...
+        isVisible: boolean,
```
paulirish: with `visibility` i think about visible within viewport or css visibility... this is neither.. what about a name like `hasSize` or... multiply width x height and call the result `pixelArea` and filter out the 0s later..

thoughts?

lighthouse-core/lib/page-functions.js
```diff
@@ -218,6 +218,52 @@ function getNodeSelector(node) {
...
+ * @param {?HTMLElement} element
+ * @return {boolean}
+ */
+/* istanbul ignore next */
+function isPositionFixed(element) {
```
paulirish: do you care about `position:sticky` ?

types/artifacts.d.ts
paulirish: aye. i'm not seeing this used in the audits right now.

lighthouse-core/gather/gatherers/iframe-elements.js
jburger424: Following Paul's pixelArea idea, so no longer applicable.

types/artifacts.d.ts
jburger424: Following Paul's pixelArea idea, so no longer applicable.

types/artifacts.d.ts
jburger424: Like the idea of pixelArea

lighthouse-core/gather/gatherers/iframe-elements.js
warrengm: I think it's quite common not to find it, e.g. at least a couple iframes on the page with no DOM IDs. Leaving it optional in the typedef SGTM

types/artifacts.d.ts
warrengm: It's used in the paint timing audit, to match the paint trace events with corresponding iframes

lighthouse-core/gather/gatherers/iframe-elements.js
warrengm: Good to know! I assumed the compiler would infer the type parameters

types/artifacts.d.ts
```diff
@@ -182,6 +184,28 @@ declare global {
...
+          right: number;
+          width: number;
+          height: number;
+        },
+        /** The pixel area iframe. */
+        pixelArea: number,
+        /** If the iframe or an ancestor of the iframe is fixed in position. */
+        isPositionFixed: boolean,
+        /** The Frame of the iframe. @see https://chromedevtools.github.io/devtools-protocol/tot/Page#type-Frame */
+        frame?: Crdp.Page.Frame | null,
```
connorjclark: could this be just `Crdp.Page.Frame | null` or `Crdp.Page.Frame | undefined`? current is `Crdp.Page.Frame | null | undefined`

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @return {LH.Artifacts['IFrameElements']}
+ */
+/* istanbul ignore next */
+function collectIFrameElements() {
+  // @ts-ignore - put into scope via stringification
+  const iFrameElements = getElementsInDocument('iframe'); // eslint-disable-line no-undef
+  return iFrameElements.map(/** @param {HTMLIFrameElement} node */ (node) => {
+    const clientRect = node.getBoundingClientRect();
+    // Marking 1x1 as non-visible to ignore tracking pixels.
```
warrengm: Comment is out of date?

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,83 @@
...
+
+    const {top, bottom, left, right, width, height} = clientRect;
+    return {
+      id: node.id,
+      src: node.src,
+      clientRect: {top, bottom, left, right, width, height},
```
warrengm: Should this just be `clientRect`? Or is it necessary to copy the object?

jburger424: This change was due to https://github.com/GoogleChrome/lighthouse/pull/8979#discussion_r285728948

types/artifacts.d.ts
jburger424: SGTM, went with `frame?: Crdp.Page.Frame` as that seems to follow the precedent on that file.

lighthouse-core/gather/gatherers/iframe-elements.js
warrengm: Got it. Thanks!

lighthouse-core/lib/page-functions.js
```diff
@@ -225,6 +225,55 @@ function getNodeSelector(node) {
...
+  /**
+   * @param {HTMLElement} element
+   * @return {boolean}
+   */
+  function hasScrollableAncestor(element) {
+    let currentEl = element.parentElement;
+    while (currentEl) {
+      if (currentEl.scrollHeight > currentEl.clientHeight) {
+        const yScroll = getStyleAttrValue(currentEl, 'overflowY');
```
patrickhulce: what's the circumstance that `yScroll` is falsy? isn't it always going to be set to one of the valid overflow values?

maybe I just need some help understanding the scrollable ancestor logic :) 

is it basically "the scroll height is greater than client height and the overflow isn't hidden"?

types/artifacts.d.ts
```diff
@@ -182,6 +184,28 @@ declare global {
...
+        /** The iframe's ClientRect. @see https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect */
+        clientRect: {
+          top: number;
+          bottom: number;
+          left: number;
+          right: number;
+          width: number;
+          height: number;
+        },
+        /** The pixel area iframe. */
```
patrickhulce: maybe this is the place for some notes on future plans other than `width * height`?

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,81 @@
...
+  // @ts-ignore - put into scope via stringification
+  const iFrameElements = getElementsInDocument('iframe'); // eslint-disable-line no-undef
+  return iFrameElements.map(/** @param {HTMLIFrameElement} node */ (node) => {
+    const clientRect = node.getBoundingClientRect();
+    const {top, bottom, left, right, width, height} = clientRect;
+    return {
+      id: node.id,
+      src: node.src,
+      clientRect: {top, bottom, left, right, width, height},
+      pixelArea: width * height,
```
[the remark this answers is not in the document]
jburger424: Now that you mention it, this is pretty useless. Removed.

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,81 @@
...
+    const toVisit = [frameTree];
+    /** @type {Map<string | undefined, LH.Crdp.Page.Frame | undefined>} */
+    const framesByDomId = new Map();
+
+    while (toVisit.length) {
+      const tempFrameTree = toVisit.shift();
+      if (tempFrameTree) {
+        for (const childFrameTree of tempFrameTree.childFrames || []) {
+          if (childFrameTree.childFrames) {
+            toVisit.concat(childFrameTree.childFrames);
```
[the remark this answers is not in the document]
jburger424: good catch, fixed. Will add a test in next commit.

lighthouse-core/lib/page-functions.js
jburger424: Wow, this was a complete oversight. I've modified it to behave as expected.

> is it basically "the scroll height is greater than client height and the overflow isn't hidden"?

Yeah that's what we're looking for. Essentially saying that an element isn't truly fixed unless it behaves as fixed,ie . stays in place when the user scrolls.

types/artifacts.d.ts
jburger424: No longer applicable.

lighthouse-core/lib/page-functions.js
```diff
@@ -225,6 +225,59 @@ function getNodeSelector(node) {
...
+        // In case where we reach outermost element (`<html>`), default computed value of visible
+        // will be scrollable.
+        if ((yScroll === 'scroll' || yScroll === 'auto') ||
+        (yScroll === 'visible' && currentEl.parentElement === null)
```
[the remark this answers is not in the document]
patrickhulce: now that I'm thinking about this more, I can't really think of a situation other than the document isn't at least the window height where this function will ever return false... is that expected it's fairly rare?

maybe we could add some comments to the jsdoc of `hasScrollableAncestor` describing the concrete examples we need this to watchout for?

lighthouse-core/lib/page-functions.js
```diff
@@ -225,6 +225,59 @@ function getNodeSelector(node) {
...
+  function getStyleAttrValue(element, attr) {
+    // Check style before computedStyle as computedStyle is expensive.
+    return element.style[attr] || window.getComputedStyle(element)[attr];
+  }
+
+  let currentEl = element;
+  while (currentEl) {
+    const position = getStyleAttrValue(currentEl, 'position');
+    // Only truly fixed if an ancestor is scrollable.
+    if ((position === 'fixed' || position === 'sticky') && hasScrollableAncestor(currentEl)) {
```
patrickhulce: we should be able to return `false` early if `hasScrollableAncestor` is falsy, right? could probably do that outside the loop entirely

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,80 @@
...
+    const driver = passContext.driver;
+
+    const {frameTree} = await driver.sendCommand('Page.getFrameTree');
+    let toVisit = [frameTree];
+    /** @type {Map<string | undefined, LH.Crdp.Page.Frame | undefined>} */
+    const framesByDomId = new Map();
+
+    while (toVisit.length) {
+      const tempFrameTree = toVisit.shift();
+      if (tempFrameTree) {
```
patrickhulce: is this just to satisfy `ts`? `tempFrameTree` should never be undefined if `toVisit.length` is truthy, right?

maybe we can just flip this to avoid the nesting at least?

```js
if (!tempFrameTree) continue;

File: lighthouse-core/gather/gatherers/iframe-elements.js [hunk 36]
```diff
@@ -0,0 +1,80 @@
...
+        for (const childFrameTree of tempFrameTree.childFrames || []) {
+          if (childFrameTree.childFrames) {
+            toVisit = toVisit.concat(childFrameTree.childFrames);
+          }
+          if (framesByDomId.has(childFrameTree.frame.name)) {
```
patrickhulce: nit: feels a bit confusing to do all this grandchildren work instead of just letting the while loop take care of the 2nd level.

WDYT about

```js
while (toVisit.length) {
  const frameTree = toVisit.shift()
  // Shouldn't be necessary but make tsc happy
  if (!frameTree) continue;
  // Set the dom ID
  if (framesByDomId.has(frameTree.frame.name)) {
    framesByDomId.set(frameTree.frame.name, undefined);
  } else {
    framesByDomId.set(frameTree.frame.name, frameTree.frame);
  }

  // Add all the children to our queue
  if (frameTree.childFrames) {
    toVisit = toVisit.concat(frameTree.childFrames);
  }
}

File: lighthouse-core/gather/gatherers/iframe-elements.js [hunk 37]
```diff
@@ -0,0 +1,80 @@
...
+  async afterPass(passContext) {
+    const driver = passContext.driver;
+
+    const {frameTree} = await driver.sendCommand('Page.getFrameTree');
+    let toVisit = [frameTree];
+    /** @type {Map<string | undefined, LH.Crdp.Page.Frame | undefined>} */
+    const framesByDomId = new Map();
+
+    while (toVisit.length) {
+      const tempFrameTree = toVisit.shift();
```
patrickhulce: nit: `tempFrameTree` is a bit unusual to me for some reason. it's only temporary in the sense that it's the current one in the loop, correct?

do you like any of these?

- currentFrameTree
- frameTree (and rename frameTree to rootFrameTree)

lighthouse-core/lib/page-functions.js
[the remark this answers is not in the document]
jburger424: You're completely right, that really is the only case where this will fail so doing lots of unneeded work. Will modify this to only look for that condition.

lighthouse-core/gather/gatherers/iframe-elements.js
jburger424: Went with above suggestion so no longer applicable.

jburger424:
@patrickhulce I'm ensuring that frame traversal is working properly in the smoke test, by ensuring that the nested iframe has a `frame` attr. Do you still think it would be good to test this in a unit test?

lighthouse-cli/test/smokehouse/oopif-expectations.js
```diff
@@ -27,5 +27,48 @@ module.exports = [
...
+    artifacts: {
+      IFrameElements: [
+        {
+          id: 'oopif',
+          src: 'https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/',
+          clientRect: {
+            width: '>0',
+            height: '>0',
+          },
+          isPositionFixed: false,
```
patrickhulce: can we assert anything about the frame here or do we not get this info about cross origin frames?

patrickhulce: also worth noting that we picked this page because it has a nested youtube embed iframe in it, should that be expected to appear in this list?

or because it's an OOPIF we don't expect our artifact to capture it at all? 

i.e. what do we do about frames that are in the devtools protocl frame tree but we can't access from the dom due to origin concerns

lighthouse-core/lib/page-functions.js
```diff
@@ -225,6 +232,44 @@ function getNodeSelector(node) {
...
+  const htmlEl = document.querySelector('html');
+  if (htmlEl.scrollHeight <= htmlEl.clientHeight ||
+      !['scroll', 'auto', 'visible'].includes(getStyleAttrValue(htmlEl, 'overflowY'))) {
+    return false;
+  }
+
+  let currentEl = element;
+  while (currentEl) {
+    const position = getStyleAttrValue(currentEl, 'position');
+    // Only truly fixed if an ancestor is scrollable.
```
patrickhulce: this comment feels out of place now :)

lighthouse-core/lib/page-functions.js
```diff
@@ -98,6 +98,13 @@ function getElementsInDocument(selector) {
       if (el.shadowRoot) {
         _findAllElements(el.shadowRoot.querySelectorAll('*'));
       }
+      // If the element has a contentWindow (IFrame), dig deeper.
+      // Try/Catch needed in case of cross-origin frame.
+      try {
+        if (el.contentWindow) {
+          _findAllElements(el.contentWindow.document.querySelectorAll('*'));
```
patrickhulce: is there something *slightly* more specific we could catch here? like testing the content window if its accesible? right now it'll silently drop all errors in the entire recursive call

lighthouse-core/lib/page-functions.js
```diff
@@ -225,6 +232,44 @@ function getNodeSelector(node) {
...
+   * @param {HTMLElement} element
+   * @param {string} attr
+   * @return {string}
+   */
+  function getStyleAttrValue(element, attr) {
+    // Check style before computedStyle as computedStyle is expensive.
+    return element.style[attr] || window.getComputedStyle(element)[attr];
+  }
+
+  // Position fixed/sticky has no effect in case when document does not scroll.
```
patrickhulce: FWIW, I didn't mean to discourage the entire pattern since there are legitimate cases where it will have an effect and the document isn't scrollable. 

My primary goal was to see what we cared about for the `isPositionFixed` in `adspeed` since I expected those cases to be rarer in iframes and making sure we documented this function accordingly with what to look out for :)

types/artifacts.d.ts
```diff
@@ -182,6 +184,26 @@ declare global {
...
+          top: number;
+          bottom: number;
+          left: number;
+          right: number;
+          width: number;
+          height: number;
+        },
+        /** If the iframe or an ancestor of the iframe is fixed in position. */
+        isPositionFixed: boolean,
+        /** The Frame of the iframe, undefined if cross-origin. @see https://chromedevtools.github.io/devtools-protocol/tot/Page#type-Frame */
```
patrickhulce: ah perhaps my question above is already answered then :)

this is somewhat curious, I would at least expect the root OOPIF entry to appear even if we don't get the OOPIF-child information. I expect this to be very relevant for ads in particular. Then again, I'm not really sure what additional information `frame` really conveys to the user. How is it used in the adspeed plugin?

lighthouse-cli/test/smokehouse/oopif-expectations.js
jburger424: added an assertion that frame is undefined. Unfortunately it doesn't look like we are able to access frame information about OOPIF, found this old comment from @paulirish when I was looking into the issue: https://github.com/cyrus-and/chrome-remote-interface/issues/276#issuecomment-435553068.

I don't think there are any such frames, I think it's the same frames that we can't access that also don't show up in the frametree.

lighthouse-core/lib/page-functions.js
jburger424: Figured out a better way to do this without the try/catch :), turns out that contentDocument is null, rather than throwing, for OOPIFs

lighthouse-core/lib/page-functions.js
jburger424: Gotcha, after your comment though I realized that this really is the only case we need to worry about.

Our motivation for including this is specifically for our ad-top-of-viewport audit. We're telling publishers to not put an ad at the top of the viewport because it can be scrolled past before loading, however, in the case of a fixed ad this advice is irrelevant.

types/artifacts.d.ts
jburger424: it certainly is relevant, but it doesn't seem like there's any way for us to access it, open to suggestions though if you have any thoughts.

> How is it used in the adspeed plugin?

It's used in the [paint timing audit](https://github.com/googleads/pub-ads-lighthouse-plugin/blob/master/lighthouse-plugin-ad-speed-insights/audits/first-ad-paint.js), to match the paint trace events with corresponding iframes

patrickhulce:
@paulirish @brendankenny do you have any desire to take another look at this one?

lighthouse-core/gather/gatherers/iframe-elements.js
```diff
@@ -0,0 +1,80 @@
...
+
+    while (toVisit.length) {
+      const frameTree = toVisit.shift();
+      // Should never be undefined, but needed for tsc.
+      if (!frameTree) continue;
+      if (framesByDomId.has(frameTree.frame.name)) {
+        // DOM ID collision, mark as undefined.
+        framesByDomId.set(frameTree.frame.name, undefined);
+      } else {
+        framesByDomId.set(frameTree.frame.name, frameTree.frame);
```
paulirish: So I don't think this works.

We're using the `name` of a frame to set it, but later we match by `id`. These are different things and not guaranteed to match: 
image

lighthouse-core/lib/page-functions.js
```diff
@@ -98,6 +98,10 @@ function getElementsInDocument(selector) {
       if (el.shadowRoot) {
         _findAllElements(el.shadowRoot.querySelectorAll('*'));
       }
+      // If the element has a contentDocument (IFrame), dig deeper.
```
paulirish: only some iframes have contentDocuments... I honestly dont understand why some do and some dont (even within google ads iframes...)
image

1) can we make this comment more specific to which iframes are included
2) i think we need this behind a `pierceIframes` option that defaults to off, otherwise our `dom-size` audit is about to go pretty crazy.

types/artifacts.d.ts
```diff
@@ -182,6 +184,26 @@ declare global {
...
+        frame?: Crdp.Page.Frame,
```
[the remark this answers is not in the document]
brendankenny: > that aside, I support the idea of including this type here in the artifact. Seems useful for any audit work.

I'm not super wild about it, but I can live with it :)

But:
- the type should be defined inline, not referencing the protocol so that (theoretical?) users of our types won't need the protocol types as a dependency
- Since it's going into `PublicGathererArtifacts` we should trim it to just the properties we need (with any others which would be ridiculous to not include) since ideally we should only be adding to this type in the future, not trimming it, even on breaking versions

jburger424:
@paulirish Sorry for the late response, have been OOO for past week. Will start digging into these today.

Issue #8223, opened 2019-04-12
Adopt new Page.getInstallabilityErrors method for PWA checks


paulirish:
https://chromium-review.googlesource.com/c/chromium/src/+/1560533
https://chromedevtools.github.io/devtools-protocol/tot/Page#method-getInstallabilityErrors

It tests against the real implementation, so we won't have a gap.
https://cs.chromium.org/chromium/src/chrome/browser/installable/installable_logging.cc?l=15-57

This also means we can probably drop manifest parser, manifest-values, etc. 
Their offline check is a little different though. 

The strings returned are hardcoded english.. but then again so are our current strings. ;)

Hoten:
Won't be able to drop manifest-values completely, as many checks aren't handled via the protocol. Example - background color, theme color, icons >= 512px.

connorjclark:
see: https://docs.google.com/document/d/1vRjvfLRm7IgqTDf_wza01UJ9nTo-tbaSfY5wlVZAV4c/edit#

paulirish:
now a dupe of #10515

PR #8995, opened 2019-05-19
audits(installable-manifest): defer to Page.getInstallabilityErrors when all of our checks pass

connorjclark:
See #8223

This is just meant to cover any possible gaps between our checks and the checks made via the protocol.

I don't see an obvious way to remove our checks completely (see comment block in the PR). Perhaps we could augment the protocol to give us some error codes. That'd allow us to convert errors (or lack thereof) from the protocol to a particular check, which can be consumed by other audits (like `splash-screen`).

### DIFF
--- a/lighthouse-core/gather/gatherers/installability-errors.js
+++ b/lighthouse-core/gather/gatherers/installability-errors.js
@@ -1,0 +1,25 @@
+/**
+ * @license Copyright 2019 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


lighthouse-core/gather/gatherers/installability-errors.js
```diff
@@ -0,0 +1,25 @@
...
+const Gatherer = require('./gatherer');
+
+class InstallabilityErrors extends Gatherer {
+  /**
+    * @param {LH.Gatherer.PassContext} passContext
+    * @return {Promise<LH.Artifacts['InstallabilityErrors']>}
+    */
+  afterPass(passContext) {
+    const driver = passContext.driver;
+    // @ts-ignore - types are missing in @types/chrome.
```
brendankenny: it can be updated in the devtools-protocol repo

connorjclark:
agreed. this need a lot of protocol work. and seems nowhere on our roadmap.

PR #9002, opened 2019-05-20
clients(lr): insert assets in lhr for logging purposes

Hoten:
This enables logging from PSI.

Maybe we should log all the artifacts too?

Tests awaiting feedback.

Note: this PR is building on [this](https://groups.google.com/a/google.com/forum/#!searchin/lightrider-eng/assets%7Csort:date/lightrider-eng/GZneqL21Kas/qtfhkVdDBgAJ) (Googlers only).

### DIFF
--- a/clients/lightrider/lightrider-entry.js
+++ b/clients/lightrider/lightrider-entry.js
@@ -56,9 +56,7 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
     if (!runnerResult) return;
 
     if (logAssets) {
-      const assetsToLog =
-        await assetSaver.prepareAssetsForLogging(runnerResult.artifacts, runnerResult.lhr.audits);
-      assetSaver.logAssets(assetsToLog);
+      await assetSaver.logAssets(runnerResult.artifacts, runnerResult.lhr.audits);
     }
 
     // pre process the LHR for proto
--- a/clients/lightrider/lightrider-entry.js
+++ b/clients/lightrider/lightrider-entry.js
@@ -55,10 +55,6 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
     const runnerResult = await lighthouse(url, flags, config, connection);
     if (!runnerResult) return;
 
-    if (logAssets) {
-      await assetSaver.logAssets(runnerResult.artifacts, runnerResult.lhr.audits);
-    }
-
     // pre process the LHR for proto
     if (flags.output === 'json' && typeof runnerResult.report === 'string') {
       // When LR is called with |internal: {keep_raw_response: true, save_lighthouse_assets: true}|,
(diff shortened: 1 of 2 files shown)


brendankenny:
at some point we were going to delete all of this because the artifacts are returned by the module now. Can lr use that instead of this logging business?

Hoten:
Discussed offline a bit.

We considered changed the LR entry to return the entire RunnerResult, but that would involve some annoying parsing changes in LR.

Realized we can just put all of the `runnerResults.artifacts` on the LR return value, without touching the asset saving module. I can revert all the changes there, unless it's better now?

lighthouse-core/lib/asset-saver.js
```diff
@@ -255,22 +255,39 @@ async function saveAssets(artifacts, audits, pathWithBasename) {
  * Log trace(s) and associated devtoolsLog(s) to console.
  * @param {LH.Artifacts} artifacts
  * @param {LH.Audit.Results} audits
- * @return {Promise<void>}
+ * @return {Promise<Array<{name: string, contents: string}>>}
  */
-async function logAssets(artifacts, audits) {
```
paulirish: once we have the LR-side to process your nested artifacts, I think we can delete this logAssets function.

clients/lightrider/lightrider-entry.js
```diff
@@ -52,19 +52,29 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
...
 
   try {
-    const results = await lighthouse(url, flags, config, connection);
-    if (!results) return;
+    const runnerResult = await lighthouse(url, flags, config, connection);
+    if (!runnerResult) return;
 
     if (logAssets) {
-      await assetSaver.logAssets(results.artifacts, results.lhr.audits);
+      const assetsToLog =
```
paulirish: this refactor doesn't seem to be a win. if anything, logAssets() should call its own private prepare method, but due to other comments, i dont think we'll end up needing this code anyhow.

clients/lightrider/lightrider-entry.js
```diff
@@ -52,19 +52,29 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
...
     }
 
     // pre process the LHR for proto
-    if (flags.output === 'json' && typeof results.report === 'string') {
-      return preprocessor.processForProto(results.report);
+    if (flags.output === 'json' && typeof runnerResult.report === 'string') {
+      // When LR is called with |internal: {keep_raw_response: true, save_lighthouse_assets: true}|,
+      // this code will log artifacts to raw_response.artifacts.
+      if (logAssets) {
+        const reportObj = JSON.parse(runnerResult.report);
```
paulirish: let's use runnerResult.lhr instead of parsing the string back

lighthouse-core/lib/asset-saver.js
Hoten: done

PR #9003, opened 2019-05-20
core(config): split out config helpers

patrickhulce:
**Summary**
Smallest portion I can see to split out from #8374. I made separate copy and then modify commits to make it easier to review, but it didn't really help much since config is so massive. Sorry! Part 1 of 4.

1. (This PR) Split apart some config logic into a helper file.
2. Add a dedicated ID to audit and gatherer definitions so that they are more easily referenceable. This also has the happy side effect of enabling the use of options in plugin audits in step 3 :D
2. Enforce that every plugin audit id is prefixed with `<plugin-id>`. It will still be able to reference our existing core audit implementations. 
3. Load each audit when loading a plugin to check its requiredArtifacts against a set of blessed artifacts.

Besides blessed artifacts and validating plugin audits at load time, the primary benefit of this effort is that plugins can start to reuse core audits with their own options. Immediate use cases I'd like to offer are a simulated 3G performance plugin and simulated warm load performance plugin. i.e. user gets all three scores with just 1 LH run.

**Related Issues/PRs**
#6747  #8374


lighthouse-core/config/config-helpers.js
```diff
@@ -0,0 +1,208 @@
+/**
+ * @license Copyright 2016 Google Inc. All Rights Reserved.
```
exterkamp: 2019

lighthouse-core/config/config-helpers.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    existingItem.options = Object.assign({}, existingItem.options, item.options);
+  }
+
+  return mergedItems;
+};
+
+/**
+ * @param {LH.Config.AuditDefn} auditDefinition
```
exterkamp: Add a function comment?

Hoten:
image

====

image

PR #9005, opened 2019-05-20
misc: remove compat for headless <m59

paulirish:


lighthouse-core/gather/connections/cri.js
```diff
@@ -36,20 +36,6 @@ class CriConnection extends Connection {
   connect() {
     return this._runJsonCommand('new')
       .then(response => this._connectToSocket(/** @type {LH.DevToolsJsonTarget} */(response)))
-      .catch(_ => {
-        // Compat: headless didn't support `/json/new` before m59. (#970, crbug.com/699392)
-        // If no support, we fallback and reuse an existing open tab
-        log.warn('CriConnection', 'Cannot create new tab; reusing open tab.');
```
patrickhulce: I want to say I've actually gotten this log message before in headful somehow, but I can't remember the circumstances.

brendankenny:
Should we really close this? This seems like a prime case for deletion.

I guess we could still catch but only throw a clear error so we'd be able to spot this case later if it still comes up

patrickhulce:
Yeah sorry wasn't trying to say we shouldn't do this just offering a data point :)

Seems like a good candidate for removal just brainstorming the other cases we might have run into this since then

PR #9007, opened 2019-05-21
misc(compat): clean up top-level task tracing

paulirish:
This removes support for handling traces from m70 and before, but gets rid of some cruft.


lighthouse-core/gather/driver.js
```diff
@@ -100,9 +100,6 @@ class Driver {
       // Exclude default categories. We'll be selective to minimize trace size
       '-*',
 
-      // Used instead of 'toplevel' in Chrome 71+
-      'disabled-by-default-lighthouse',
```
patrickhulce: Is this not a thing anymore then? That seems short-lived 😆

lighthouse-core/lib/traces/tracing-processor.js
```diff
@@ -264,14 +255,23 @@ class TraceProcessor {
...
+   * and regressed as didn't catch top-level events inside microtasks,
+   * - This was identified & fixed in http://crbug.com/953914. The fix made it to m75 before stable release.
+   *
+   * In summary: as of m75, `cat: 'disabled-by-default-devtools.timeline', name: 'RunTask'` events are top-level tasks.
+   *
    * @param {LH.TraceEvent} evt
    * @return {boolean}
    */
   static isScheduleableTask(evt) {
-    return evt.name === SCHEDULABLE_TASK_TITLE_LH ||
```
patrickhulce: Can we leave this in here? Especially for tracehouse parsing older traces still has value and isn't *that* big of a maintenance burden.

paulirish: aiight.

PR #9010, opened 2019-05-21
tests: update a11y artifacts

mattzeunert:
**Summary**

Update out of date a11y data. Thought it might be related to https://github.com/GoogleChrome/lighthouse/pull/8823, but that PR did update the the artifacts – not sure what changes these are.

Also, when running `npm run update:sample-json` on master you get the `summary: null` sample change, which broke the proto tests. Some of the tests already had a `beforeAll` that fixed that, so moved that beforeAll to be top-level.

### DIFF
--- a/proto/sample_v2_round_trip.json
+++ b/proto/sample_v2_round_trip.json
@@ -1115,7 +1115,7 @@
                         "name": "WordPress"
                     }
                 ],
-                "summary": null,
+                "summary": {},
                 "type": "table"
             },
             "id": "js-libraries",
@@ -1247,7 +1247,7 @@
             "details": {
                 "headings": [],
                 "items": [],
-                "summary": null,
+                "summary": {},
                 "type": "table"
             },
             "id": "link-text",
@@ -1903,7 +1903,7 @@
                         "vulnCount": 2.0
                     }
                 ],
-                "summary": null,
+                "summary": {},
                 "type": "table"
             },
             "displayValue": "2 vulnerabilities detected",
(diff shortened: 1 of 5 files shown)


proto/sample_v2_round_trip.json
```diff
@@ -1115,7 +1115,7 @@
                         "name": "WordPress"
                     }
                 ],
-                "summary": {},
```
patrickhulce: @exterkamp this is expected now, right? I've lost track of all the summary fixes and what's real and what's just tests 😆

exterkamp: No.  @brendankenny can get his proto to output `{}` instead of `null`, but I can't get mine to, and I guess @mattzeunert can't either.  `{}` should be considered canonical imo, and we need to source out this difference.

Hoten: make sure you're using the latest protobuf compiler.

exterkamp: I'm on protoc v=3.7.1

I just recompiled from source.  I'm going to try to uninstall and reinstall.

mattzeunert: Ran `brew upgrade protobuf` to upgrade to 3.7.1 and it works fine for me now.

brendankenny: > No. @brendankenny can get his proto to output {} instead of null, but I can't get mine to, and I guess @mattzeunert can't either

I guess @exterkamp now stands alone with his buggy protoc :)

brendankenny:
> Update out-of-date a11y data. Thought it might be related to #8823, but that PR did update the the artifacts – not sure what changes these are.

I think it was mostly from #8373 (with another one from rob that added `accesskeys` after it has been removed for a little while) and we just never updated the artifacts to match.

PR #9016, opened 2019-05-21
proto: update processForProto method signature, string -> LH.Result

paulirish:
Also, lightrider will only request JSON output now.


brendankenny:
does this supersede #9002?

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -16,11 +16,12 @@ const fs = require('fs');
  */
 
 /**
-  * @param {string} result
+  * Transform an LHR into a proto-friendly LHR
```
brendankenny:
```suggestion
  * Transform an LHR into a proto-friendly, mostly-compatible LHR.
```

lighthouse-core/test/lib/proto-preprocessor-test.js
```diff
@@ -5,7 +5,7 @@
  */
 'use strict';
 
-const processForProto = require('../../lib/proto-preprocessor').processForProto;
```
brendankenny: can you add a test that it doesn't modify the passed-in lhr?

brendankenny:
> does this supersede #9002?

oh, jk, only touching `results.lhr`

clients/lightrider/lightrider-entry.js
```diff
@@ -25,9 +25,9 @@ const LR_PRESETS = {
  * If configOverride is provided, lrDevice and categoryIDs are ignored.
  * @param {Connection} connection
  * @param {string} url
- * @param {LH.Flags} flags Lighthouse flags, including `output`
+ * @param {LH.Flags} flags Lighthouse flags
  * @param {{lrDevice?: 'desktop'|'mobile', categoryIDs?: Array<string>, logAssets: boolean, configOverride?: LH.Config.Json}} lrOpts Options coming from Lightrider
- * @return {Promise<string|Array<string>|void>}
+ * @return {Promise<string|void>}
```
brendankenny: is `void` still a return possibility? (I think it can be dropped now)

clients/lightrider/lightrider-entry.js
```diff
@@ -25,9 +25,9 @@ const LR_PRESETS = {
  * If configOverride is provided, lrDevice and categoryIDs are ignored.
  * @param {Connection} connection
  * @param {string} url
- * @param {LH.Flags} flags Lighthouse flags, including `output`
+ * @param {LH.Flags} flags Lighthouse flags
```
brendankenny: make sure you aren't accidentally still passing in `output: 'html'` and doing all that extra work :)

clients/lightrider/lightrider-entry.js
brendankenny: > is `void` still a return possibility?

Oh, yes it is for the `if (!results) return` case. Should it throw to get a runtime error when that happens? Weird to return nothing if Lighthouse was acting poorly (I think it's actually a `// should never happen` case to make tsc happy).

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -16,11 +16,12 @@ const fs = require('fs');
...
+  * @param {LH.Result} lhr
+  * @return {LH.Result}
   */
-function processForProto(result) {
-  /** @type {LH.Result} */
-  const reportJson = JSON.parse(result);
+function processForProto(lhr) {
+  const reportJson = JSON.parse(JSON.stringify(lhr));
```
Hoten: could we skip the cloning? Only the lr entry uses this fn, and the original lhr is not needed there. if some other client comes along with different requirements they can clone the input themselves.

assuming cloning a lhr takes a measurable amount of time.

I clocked 4ms for example.com and for 8ms nyt.com:

```
console.time('clone'); JSON.parse(JSON.stringify(lhr)); console.timeEnd('clone');
```

clients/lightrider/lightrider-entry.js
```diff
@@ -25,9 +25,9 @@ const LR_PRESETS = {
...
  * @param {{lrDevice?: 'desktop'|'mobile', categoryIDs?: Array<string>, logAssets: boolean, configOverride?: LH.Config.Json}} lrOpts Options coming from Lightrider
- * @return {Promise<string|Array<string>|void>}
+ * @return {Promise<string|void>}
  */
 async function runLighthouseInLR(connection, url, flags, lrOpts) {
   const {lrDevice, categoryIDs, logAssets, configOverride} = lrOpts;
```
Hoten: I'm unclear how the default for the output flag is handled. Could we be explicit and add `flags.output = 'json'` here https://github.com/GoogleChrome/lighthouse/pull/9016/files#diff-872604c92afd40cde2ae727a77bae800R41 ?

brendankenny: it can, but that's the default (at some point you have to trust the defaults, because there are a whole lot of them :). I meant my reminder mostly because I know something used to be requesting `html` output

lighthouse-core/lib/proto-preprocessor.js
paulirish: I find it unexpected when methods mutate the passed in object, so this is my preferred pattern. It's also what we do for Util.prepareReportResult, swapLocale, our tests, etc.

We adopted this pattern in our tests, as previously we had some unexpected mutations that caused failures that were tricky to track down.

clients/lightrider/lightrider-entry.js
paulirish: Added a comment.. i dont think we need a stronger check.

lighthouse-core/lib/proto-preprocessor.js
Hoten: yeah I prefer avoiding mutation too. if the few ms is worth the change, I'd suggest renaming the method to `modifyForProto` to properly set expectations.

clients/lightrider/lightrider-entry.js
brendankenny: > Added a comment.. i dont think we need a stronger check.

yeah, I meant just double check existing code, definitely don't need a check *in* the code

clients/lightrider/lightrider-entry.js
```diff
@@ -38,6 +38,7 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
   flags.disableStorageReset = true;
   flags.logLevel = flags.logLevel || 'info';
   flags.channel = 'lr';
+  // flags.output assumed to be unset by Lightrider
```
brendankenny: yeah, probably just drop this line. At worst it's just extra work, and is just as assumed as every other property on `flags` could be (and we don't make a comment for any of those)

clients/lightrider/lightrider-entry.js
```diff
@@ -52,22 +53,17 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
...
-      // this code will log artifacts to raw_response.artifacts.
-      if (logAssets) {
-        // @ts-ignore - Regenerate the report, but tack on the artifacts.
-        runnerResult.lhr.artifacts = runnerResult.artifacts;
-        runnerResult.report = JSON.stringify(runnerResult.lhr);
-      }
-
-      return preprocessor.processForProto(runnerResult.report);
+    // When LR is called with |internal: {keep_raw_response: true, save_lighthouse_assets: true}|,
+    // this code will log artifacts to raw_response.artifacts.
```
brendankenny: this comment is inherited but is pretty confusing here since it appears to refer entirely to stuff outside of LH? e.g. `keep_raw_response` and/or `save_lighthouse_assets` somehow become `lrOpts.logAssets`, and `lhr` becomes `raw_response`?

If it's helpful for users of lr, though, seems ok to keep, though

clients/lightrider/lightrider-entry.js
```diff
@@ -52,22 +53,17 @@ async function runLighthouseInLR(connection, url, flags, lrOpts) {
...
+    if (logAssets) {
+      // @ts-ignore - Regenerate the report, but tack on the artifacts.
```
brendankenny: maybe regenerate was referring to what it used to be doing? what about just `// @ts-ignore - piggyback the artifacts on the LHR.`

lighthouse-core/lib/proto-preprocessor.js
```diff
@@ -16,11 +16,12 @@ const fs = require('fs');
...
-  * @param {string} result
+  * Transform an LHR into a proto-friendly, mostly-compatible LHR.
+  * @param {LH.Result} lhr
+  * @return {LH.Result}
   */
-function processForProto(result) {
-  /** @type {LH.Result} */
-  const reportJson = JSON.parse(result);
+function processForProto(lhr) {
+  const reportJson = JSON.parse(JSON.stringify(lhr));
```
brendankenny: the type error (failure in travis) is because this whole function is operating on an `any` `reportJson` (due to the return type of `JSON.parse`). Luckily there was a `forEach` in there or we wouldn't have noticed the whole thing wasn't being checked :)

Need to keep the `/** @type {LH.Result} */` above this line.

clients/lightrider/lightrider-entry.js
patrickhulce: should it instead throw if output is set to anything other than `undefined` or `json`? seems like it could be frustrating to figure that out on your own

 but you both seem to be on the same page that we "*definitely* don't need a check in the code" so I feel like I'm missing something

brendankenny: > should it instead throw if output is set to anything other than undefined or json?

yeah, that's why I was suggesting removing the comment, because from this comment it *definitely* sounds like it should be an assertion, not a comment :)

AIUI, the `runnerResult.report` is ignored completely on the LR side now, it uses the LHR directly (with its own HTML reportGenerator bundle), so at worst a defined `output` is just extra work.

...but there are several possible things on `flags` that could add extra work and we're just assuming those are correct too, so I really don't think it's worth thinking about. Really it's something for the caller to worry about, either way anyways

Issue #9019, opened 2019-05-22
DevTools report redesign


connorjclark:
I collected some screenshots of what master looks like now in DevTools. Not pretty (yet) :)

This is what 4.2 (in stable) looks like now (yes, DevTools LH had dark mode already):
<img width="733" alt="lh-4 2" src="https://user-images.githubusercontent.com/4071474/58145899-bf258300-7c08-11e9-97b7-0ac5d92550e3.png">

Dark mode, master (I zoomed out a lot here):
<img width="730" alt="lh-5 0-before-dark" src="https://user-images.githubusercontent.com/4071474/58145919-d6647080-7c08-11e9-9aa2-c03fc3843ce8.png">

Light mode, master (same zoom settings as 4.2 img):
<img width="725" alt="lh-5 0-before-light" src="https://user-images.githubusercontent.com/4071474/58145929-debcab80-7c08-11e9-847f-775176e4c52f.png">
<img width="726" alt="lh-5 0-before-light-2" src="https://user-images.githubusercontent.com/4071474/58145936-e419f600-7c08-11e9-9435-966894a42ccb.png">

--

Dark mode is messed up b/c it was implemented (in DT) with a theme - a bunch of custom CSS overwriting the pre 5.0 stuff. Figure we just need to remove most (or all) of what that theme is doing.

Chromium changes: https://chromium-review.googlesource.com/c/chromium/src/+/1623894

Hoten:
Metric error state looks odd w/o an icon:
![image](https://user-images.githubusercontent.com/4071474/58146987-58ef2f00-7c0d-11e9-9020-1f81c0d0f4cd.png)

Hoten:
Mouse over "Error" in metrics doesn't show the error.
![image](https://user-images.githubusercontent.com/4071474/58197458-2e3ebe00-7c81-11e9-8193-1946438567d8.png)

Hoten:
Oh, you'd have to toggle the pill meant for metric descriptions to see the error..

brendankenny:
fixed! (modulo other open report rendering issues)

PR #9023, opened 2019-05-22
devtools: enable sticky header, top bar, and report ui features

Hoten:
LH changes for #9019.

image

### DIFF
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -230,7 +230,7 @@ class ReportRenderer {
     if (scoreHeader) {
       const makeScoreGauges = () =>
         this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
-      
+
       const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
       const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
       scoreHeader.append(...makeScoreGauges());
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -238,12 +238,6 @@ class ReportRenderer {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       stickyHeader.append(
         ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
-      // The highlighter is placed in the first gauge to avoid an layout issue where the first
-      // paint comes before the highlighter's calculated position is applied. This causes
-      // the initial positioning to be rendered for a split second. As a workaround, the
-      // highlighter is placed in the first score gauge so that the initial position is correct.
-      const firstGauge = this._dom.find('.lh-gauge__wrapper', stickyHeader);
-      this._dom.createChildOf(firstGauge, 'div', 'lh-highlighter');
       reportFragment.appendChild(stickyHeader);
     }
 
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -228,18 +228,17 @@ class ReportRenderer {
     reportFragment.appendChild(topbarDocumentFragment);
 
     if (scoreHeader) {
-      const makeScoreGauges = () =>
-        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
-
       const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
       const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
-      scoreHeader.append(...makeScoreGauges());
+      scoreHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
       scoresContainer.appendChild(scoreHeader);
       scoresContainer.appendChild(scoreScale);
 
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
+      stickyHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
       this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
-      stickyHeader.append(...makeScoreGauges());
       reportFragment.appendChild(stickyHeader);
     }
 
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -223,35 +223,29 @@ class ReportRenderer {
       wrapper.appendChild(renderer.render(category, report.categoryGroups));
     }
 
-    if (scoreHeader) {
-      const scoreGauges =
-        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
-      scoreHeader.append(...scoreGauges);
-      const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
-      const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
-      scoresContainer.appendChild(scoreHeader);
-      scoresContainer.appendChild(scoreScale);
-    }
-
-    reportSection.appendChild(this._renderReportFooter(report));
-
     const reportFragment = this._dom.createFragment();
     const topbarDocumentFragment = this._renderReportTopbar(report);
     reportFragment.appendChild(topbarDocumentFragment);
 
     if (scoreHeader) {
+      const makeScoreGauges = () =>
+        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
+      
+      const scoreScale = this._dom.cloneTemplate('#tmpl-lh-scorescale', this._templateContext);
+      const scoresContainer = this._dom.find('.lh-scores-container', headerContainer);
+      scoreHeader.append(...makeScoreGauges());
+      scoresContainer.appendChild(scoreHeader);
+      scoresContainer.appendChild(scoreScale);
+
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
-
-      const scoreGauges =
-        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
-      stickyHeader.append(...scoreGauges);
-
+      stickyHeader.append(...makeScoreGauges());
       reportFragment.appendChild(stickyHeader);
     }
 
     reportFragment.appendChild(headerContainer);
     reportFragment.appendChild(container);
+    reportSection.appendChild(this._renderReportFooter(report));
 
     return reportFragment;
   }
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -101,7 +101,6 @@ class ReportRenderer {
     return shortHeaderContainer;
   }
 
-
   /**
    * @param {LH.ReportResult} report
    * @return {DocumentFragment}
@@ -255,13 +254,10 @@ class ReportRenderer {
     reportSection.appendChild(this._renderReportFooter(report));
 
     const reportFragment = this._dom.createFragment();
+    const topbarDocumentFragment = this._renderReportTopbar(report);
+    reportFragment.appendChild(topbarDocumentFragment);
 
-    if (!this._dom.isDevTools()) {
-      const topbarDocumentFragment = this._renderReportTopbar(report);
-      reportFragment.appendChild(topbarDocumentFragment);
-    }
-
-    if (scoreHeader && !this._dom.isDevTools()) {
+    if (scoreHeader) {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
 
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -92,16 +92,6 @@ class ReportRenderer {
   }
 
   /**
-   * @return {Element}
-   */
-  _renderReportShortHeader() {
-    const shortHeaderContainer = this._dom.createElement('div', 'lh-header-container');
-    const wrapper = this._dom.cloneTemplate('#tmpl-lh-scores-wrapper', this._templateContext);
-    shortHeaderContainer.appendChild(wrapper);
-    return shortHeaderContainer;
-  }
-
-  /**
    * @param {LH.ReportResult} report
    * @return {DocumentFragment}
    */
@@ -194,16 +184,8 @@ class ReportRenderer {
    * @return {DocumentFragment}
    */
   _renderReport(report) {
-    let header;
     const headerContainer = this._dom.createElement('div');
-    if (this._dom.isDevTools()) {
-      headerContainer.classList.add('lh-header-plain');
-      header = this._renderReportShortHeader();
-    } else {
-      headerContainer.classList.add('lh-header-sticky');
-      header = this._renderReportHeader();
-    }
-    headerContainer.appendChild(header);
+    headerContainer.appendChild(this._renderReportHeader());
 
     const container = this._dom.createElement('div', 'lh-container');
     const reportSection = container.appendChild(this._dom.createElement('div', 'lh-report'));
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -242,7 +242,7 @@ class ReportRenderer {
       // paint comes before the highlighter's calculated position is applied. This causes
       // the initial positioning to be rendered for a split second. As a workaround, the
       // highlighter is placed in the first score gauge so that the initial position is correct.
-      const firstGauge = this._dom.find('.lh-gauge__wrapper', stickyHeader)
+      const firstGauge = this._dom.find('.lh-gauge__wrapper', stickyHeader);
       this._dom.createChildOf(firstGauge, 'div', 'lh-highlighter');
       reportFragment.appendChild(stickyHeader);
     }
--- a/lighthouse-core/report/html/renderer/report-renderer.js
+++ b/lighthouse-core/report/html/renderer/report-renderer.js
@@ -238,7 +238,12 @@ class ReportRenderer {
       const stickyHeader = this._dom.createElement('div', 'lh-sticky-header');
       stickyHeader.append(
         ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
-      this._dom.createChildOf(stickyHeader, 'div', 'lh-highlighter');
+      // The highlighter is placed in the first gauge to avoid an layout issue where the first
+      // paint comes before the highlighter's calculated position is applied. This causes
+      // the initial positioning to be rendered for a split second. As a workaround, the
+      // highlighter is placed in the first score gauge so that the initial position is correct.
+      const firstGauge = this._dom.find('.lh-gauge__wrapper', stickyHeader)
+      this._dom.createChildOf(firstGauge, 'div', 'lh-highlighter');
       reportFragment.appendChild(stickyHeader);
     }
 
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -111,7 +111,8 @@ class ReportUIFeatures {
     // There is only a sticky header when at least 2 categories are present.
     if (Object.keys(this.json.categories).length >= 2) {
       this._setupStickyHeaderElements();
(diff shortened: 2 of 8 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -185,7 +185,7 @@ class ReportUIFeatures {
 
       filterInput.id = id;
       filterInput.addEventListener('change', e => {
-        // Remove rows from the dom and keep track of them to readd on uncheck.
+        // Remove rows from the dom and keep track of them to re-add on uncheck.
```
Hoten: image

patrickhulce: *so* close to `tracium` :)

Hoten:
We're bringing report ui features (mostly) to DevTools. Plus, the sticky heady and the topbar.

image

Hoten:
Small issue: when you resize the docked DevTools, the highlighter becomes off:

image

Normally, resizing is handled easily w/ a resize event handler. But how can listen to the DevTools dock resizing?

Hoten:
fireworks:

image

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -195,16 +184,8 @@ class ReportRenderer {
...
     const headerContainer = this._dom.createElement('div');
-    if (this._dom.isDevTools()) {
-      headerContainer.classList.add('lh-header-plain');
-      header = this._renderReportShortHeader();
-    } else {
-      headerContainer.classList.add('lh-header-sticky');
-      header = this._renderReportHeader();
-    }
-    headerContainer.appendChild(header);
+    headerContainer.appendChild(this._renderReportHeader());
```
patrickhulce: this is a beautiful, beautiful line 😃

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -78,33 +80,36 @@ class ReportUIFeatures {
...
   initFeatures(report) {
-    if (this._dom.isDevTools()) return;
-
     this.json = report;
-    this._setupMediaQueryListeners();
-    this._setupExportButton();
+
+    if (this._dom.isDevTools()) {
+      // Not ready for DevTools.
+      // Saving as HTML save the entire AuditsPanel.
```
patrickhulce: hm it's not *super* clear what's being enabled in DT and what's not since the comments sorta line up with what's in the `else`.

not sure how to fix this maybe it's just moving to have a comment for each line explaining why it can't be enabled and/or when it might be moved out of the `else`?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +119,34 @@ class ReportUIFeatures {
     // There is only a sticky header when at least 2 categories are present.
     if (Object.keys(this.json.categories).length >= 2) {
       this._setupStickyHeaderElements();
-      this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
+
+      let elToAddScrollListener = this._document;
```
patrickhulce: just curious, how do folks feel about about ternary `const` expressions compared to the `let`/`if` construct?

lighthouse-core/report/html/report-styles.css
```diff
@@ -260,6 +261,10 @@
   --audit-group-indent: 16px;
   --audit-indent: 16px;
   --expandable-indent: 16px;
+  --gauge-circle-size-big: 96px;
+  --gauge-circle-size: 72px;
+  /* Consider offset of the DevTools UI. */
```
patrickhulce: these feels slightly magical is there anyway we can link to how to compute this number in the future? :)

Hoten: So I actually just increased the offset value until the topbar was 100% visible and there was no gap between the DevTools UI + the topbar.

Think I should look into computing the offset at runtime, and somehow injecting the CSS variable?

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: yeah i kinda leaned on a C++ style here for some reason ..... i can change it.

lighthouse-core/report/html/renderer/report-renderer.js
Hoten: yeah I was stoke I got to do this

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: sure - `...does not work.` is not a good comment :)

lighthouse-core/report/html/report-styles.css
patrickhulce: oh haha, nah maybe just make a note of "this is the height of the devtools UI as eyeballed by a developer in styles panel" or something? 😆

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -255,13 +236,10 @@ class ReportRenderer {
...
+    const topbarDocumentFragment = this._renderReportTopbar(report);
+    reportFragment.appendChild(topbarDocumentFragment);
 
-    if (!this._dom.isDevTools()) {
-      const topbarDocumentFragment = this._renderReportTopbar(report);
-      reportFragment.appendChild(topbarDocumentFragment);
-    }
-
-    if (scoreHeader && !this._dom.isDevTools()) {
+    if (scoreHeader) {
```
brendankenny: can we combine these `if (scoreHeader)` blocks now?

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -565,11 +598,13 @@ class ReportUIFeatures {
...
     const gaugeToHighlight = gaugeWrapperEls[highlightIndex];
-    const offset = gaugeToHighlight.getBoundingClientRect().left + 'px';
+    // This is normally 0 - but in DevTools, it's not, since the entire report could be docked.
+    const origin = this.stickyHeaderEl.getBoundingClientRect().left;
+    const offset = gaugeToHighlight.getBoundingClientRect().left - origin;
 
     // Mutate at end to avoid layout thrashing.
     this.stickyHeaderEl.classList.toggle('lh-sticky-header--visible', showStickyHeader);
-    this.highlightEl.style.left = offset;
+    this.highlightEl.style.left = offset + 'px';
```
Hoten: If you look closely, the highlighter starts off on the far left before quickly rendering to the first gauge. I think this is b/c of the `display: none` changes. I thought doing `this.highlightEl.style.left = ...` first would fix it, but it does not.

lighthouse-core/report/html/renderer/report-renderer.js
Hoten: done

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -545,7 +574,14 @@ class ReportUIFeatures {
    * @param {boolean} [force]
    */
   _toggleDarkTheme(force) {
-    this._document.body.classList.toggle('dark', force);
+    const el = this._dom.find('.lh-vars', this._document);
+    // This seems unnecessary, but in DevTools, passing "undefined" as the second
```
Hoten: this is weird

brendankenny: https://github.com/ChromeDevTools/devtools-frontend/blob/dd6a6d4153647c2a4203c327c595692c5e0a4256/front_end/dom_extension/DOMExtension.js#L809-L819

:/

Hoten: this is amazing

Hoten:
Open problems:

https://github.com/GoogleChrome/lighthouse/pull/9023#discussion_r288315625
https://github.com/GoogleChrome/lighthouse/pull/9023#issuecomment-495831344

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -78,33 +80,36 @@ class ReportUIFeatures {
...
-    this._setUpCollapseDetailsAfterPrinting();
-    this._resetUIState();
-    this._document.addEventListener('keyup', this.onKeyUp);
-    this._document.addEventListener('copy', this.onCopy);
+
+    // Some features in the top right drop down menu don't work in the DevTools
+    // client. They could with some tweaks, but currently they don't. For example:
+    // Saving as HTML/JSON - does not bring up a file dialog, as one would expect in DevTools.
+    //                       also, it saves the AuditsPanel HTML, which is funky.
+    if (this._dom.isDevTools()) {
```
brendankenny: what about just `display:none` on the export button? Then this stuff won't have to be special-cased

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +119,35 @@ class ReportUIFeatures {
     // There is only a sticky header when at least 2 categories are present.
     if (Object.keys(this.json.categories).length >= 2) {
       this._setupStickyHeaderElements();
-      this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
+
+      /** @type {Document | HTMLElement} */
+      let elToAddScrollListener = this._document;
```
brendankenny: this doesn't need to be fixed now, but we shouldn't have to hardcode the different possible scroll containers the report is in. Either it should be something obvious, something we can query, or something we shouldn't be messing with in the first place because who knows how the caller is using the report.

This was the reason we originally had the separate `devtools-ui-features.js` or whatever (and viewer has `viewer-ui-features.js`), because `report-ui-features.js` is written very specifically for `report-template.html`. We got rid of the file because devtools didn't need it at the time, but now that it does, maybe it's worth splitting ui features down a different axis. That way if PSI, web.dev, etc etc start needing things in here we won't have to add a case for each of their scroll containers too :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -565,11 +602,13 @@ class ReportUIFeatures {
     // Category order matches gauge order in sticky header.
     const gaugeWrapperEls = this.stickyHeaderEl.querySelectorAll('.lh-gauge__wrapper');
     const gaugeToHighlight = gaugeWrapperEls[highlightIndex];
-    const offset = gaugeToHighlight.getBoundingClientRect().left + 'px';
+    // This is normally 0 - but in DevTools, it's not, since the entire report could be docked.
```
brendankenny: weird, apparently devtools covers the whole window when docked but has a weird ghost element taking up the other section of the page where the content goes that's not selectable with inspect, etc

lighthouse-core/report/html/templates.html
```diff
@@ -171,25 +171,30 @@
...
       --score-container-width: 60px;
       --score-number-font-size: 13px;
-      position: fixed;
+      position: sticky;
       left: 0;
       right: 0;
       top: var(--topbar-height);
       font-weight: 700;
-      display: flex;
+      display: none;
```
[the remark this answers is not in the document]
Hoten: I continued b/c was incorporating feedback from Paul: https://docs.google.com/document/d/1WfzBo6VsS7hBvDMs1sMm31WpTcgRfguYty5Rv6RuziA/edit?usp=sharing

For this change, after making the sticky header `position: sticky`, even at opacity 0 the height of the element was creating a big margin:

image

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: maybe find the nearest ancestor with scroll overflow?

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: done

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -78,16 +80,16 @@ class ReportUIFeatures {
...
-    if (this._dom.isDevTools()) return;
-
     this.json = report;
+
     this._setupMediaQueryListeners();
     this._setupExportButton();
-    this._setupThirdPartyFilter();
     this._setUpCollapseDetailsAfterPrinting();
     this._resetUIState();
+    this._setupThirdPartyFilter();
```
patrickhulce: super nit: could revert these now that we don't need any changes

lighthouse-core/report/html/renderer/report-ui-features.js
patrickhulce: heh, yeah one of the 🤯moments for me was learning that with devtools the entire page is really devtools and we're just sticking this other renderer inside it that happens to take up most of the space

brendankenny: > 🤯

lol how is that in unicode but a lighthouse isn't

also (at least on a mac), that looks a lot gorier small than it does big and you realize that's a mushroom cloud coming out

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: did that. works well.

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -242,38 +223,29 @@ class ReportRenderer {
       wrapper.appendChild(renderer.render(category, report.categoryGroups));
     }
 
+    const reportFragment = this._dom.createFragment();
+    const topbarDocumentFragment = this._renderReportTopbar(report);
+    reportFragment.appendChild(topbarDocumentFragment);
+
     if (scoreHeader) {
-      const scoreGauges =
+      const makeScoreGauges = () =>
```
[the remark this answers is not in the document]
Hoten: Ideally I'd like no variables for them at all.

```
scoreHeader.append(
	...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
```

?

lighthouse-core/report/html/renderer/report-ui-features.js
brendankenny: > did that. works well

I was thinking more like IntersectionObserver but that works for now too. Hopefully stack overflow did their due diligence there :P

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: can we poke someone about the two pending Lighthouse unicode proposals? :)

lighthouse-core/report/html/renderer/report-renderer.js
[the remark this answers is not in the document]
brendankenny: > Ideally I'd like no variables for them at all.

I don't really see what that gains you over a named variable, but sure, that works too :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +111,50 @@ class ReportUIFeatures {
...
+    if (isScrollable && node.scrollHeight >= node.clientHeight) {
+      return node;
+    }
+
+    if (node.parentNode instanceof HTMLElement) {
+      return this._getScrollParent(node.parentNode);
+    } else {
+      return document;
+    }
+  }
```
brendankenny: the HTMLElement checks are overkill if we're going to fall back to `document` in the end anyways (since it'll be the only non-element node encountered). I believe we can do something like
```js
/**
 * Finds the first scrollable ancestor of `element`. Falls back to the document.
 * @param {HTMLElement} element
 * @return {Node}
 */
_getScrollParent(element) {
  const {overflowY} = window.getComputedStyle(element);
  const isScrollable = overflowY !== 'visible' && overflowY !== 'hidden';

  if (isScrollable && element.scrollHeight >= element.clientHeight) {
    return element;
  }

  if (element.parentElement) {
    return this._getScrollParent(element.parentElement);
  }

  return document;
}
```

Does the `element.scrollHeight >= element.clientHeight` check always work out? What happens in DevTools with a short report (only one category, lots of passing)? Still works somehow? No scrolling so it doesn't matter anyways? Or should this return nothing in that case so we don't add a listener to whatever `document` is in devtools?

brendankenny: ha, and this will help with the test errors :)

brendankenny: > Does the element.scrollHeight >= element.clientHeight check always work out? What happens in DevTools with a short report (only one category, lots of passing)? 

oh, this would be handled by the `=` case. But now I'm wondering is it possible for `element.scrollHeight >= element.clientHeight` to ever be false :)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +111,61 @@ class ReportUIFeatures {
...
       this._setupStickyHeaderElements();
-      this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
-      window.addEventListener('resize', this._updateStickyHeaderOnScroll);
+      const containerEl = this._dom.find('.lh-container', this._document);
+      const elToAddScrollListener = this._getScrollParent(containerEl);
+      elToAddScrollListener.addEventListener('scroll', this._updateStickyHeaderOnScroll);
+      // We can't rely on listening to the window resize event for DevTools, so we first
+      // attempt the new ResizeObserver web platform feature. It has poor cross browser
+      // support, so we should check that it's supported. However, there are some performance
+      // issues with using window.ResizeObserver - it updates much more often than the 'resize'
```
Hoten: When using the ResizeObserver in the deploy app (It's disabled outside devtools but if you revert that bit), and add some logging to the callback:

```
console.log(topbarBottom, scoreScaleTop, showStickyHeader);
```

image


You see that the position of our elements are jumping all over the place!

Hoten: So we should just disable it outside DevTools for now, luckily it works there perfectly.

lighthouse-core/report/html/renderer/report-renderer.js
```diff
@@ -242,38 +223,33 @@ class ReportRenderer {
...
-      const scoreGauges =
-        this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers);
-      stickyHeader.append(...scoreGauges);
-
+      stickyHeader.append(
+        ...this._renderScoreGauges(report, categoryRenderer, specificCategoryRenderers));
+      // The highlighter is placed in the first gauge to avoid an layout issue where the first
+      // paint comes before the highlighter's calculated position is applied. This causes
+      // the initial positioning to be rendered for a split second. As a workaround, the
+      // highlighter is placed in the first score gauge so that the initial position is correct.
```
brendankenny: I think the context of this will be lost over time and not make much sense in isolation (there's no need for a workaround cause the only thing we're doing *is* the workaround). 

Really we just need something like `// Position highlighter at first gauge; will be transformed on scroll.`

(should this div actually be added in `report-ui-features` since it's kind of broken in reports that don't use `report-ui-features` and so don't get the scroll handler?)

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +111,61 @@ class ReportUIFeatures {
...
+      // event fires, and the experience is choppy in LH. For now, limit use to just DevTools,
+      // which doesn't seem affected for some reason.
```
brendankenny: since in theory `ResizeObserver` is more efficient and the future path to monitoring resizes, what about
`// Use ResizeObserver where available. TODO: there is an issue with incorrect position numbers and, as a result, performance issues due to layout thrashing. See https://github.com/GoogleChrome/lighthouse/pull/9023/files#r288822287 for details.`

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -114,11 +111,61 @@ class ReportUIFeatures {
...
+  /**
+   * Finds the first scrollable ancestor of `element`. Falls back to the document.
+   * @param {HTMLElement} element
+   * @return {Node}
+   */
+  _getScrollParent(element) {
+    const {overflowY} = window.getComputedStyle(element);
+    const isScrollable = overflowY !== 'visible' && overflowY !== 'hidden';
+
+    if (isScrollable && element.scrollHeight >= element.clientHeight) {
```
brendankenny: drop the height check since it's always true?

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: are you also suggesting removing the devtools checks?

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: done

lighthouse-core/report/html/renderer/report-renderer.js
Hoten: good call, done

lighthouse-core/report/html/renderer/report-ui-features.js
brendankenny: > are you also suggesting removing the devtools check?

whoops, no, good point. I like your version

brendankenny:
there's a type check error...the type definition isn't working? It should be declared on window...

Hoten:
tsconfig for viewer was wrong. Compiler option `typeRoots` don't resolve to the nearest node_modules as one would expect.

Hoten:
friggen flakes...

Issue #9026, opened 2019-05-22
PSI Mobile view style guffaws


connorjclark:
I snapped these from PSI, but they should be issues in LH too.

![image](https://user-images.githubusercontent.com/4071474/58201148-3e5a9b80-7c89-11e9-9589-24d1db81d48b.png)

![image](https://user-images.githubusercontent.com/4071474/58201161-4581a980-7c89-11e9-9b28-5af4abfa5489.png)

Hoten:
LH looks like this (disclaimer in the wrong spot)

![image](https://user-images.githubusercontent.com/4071474/58215193-85f31e80-7cad-11e9-958d-c462947f1e7b.png)

brendankenny:
is this fixed now? "Values are estimated and may vary" fixed in #9030, filmstrip in #8563. Other issues?

Hoten:
The margin issue on the metrics table is PSI-only. Think that's all that remains.

exterkamp:
Fixed in PSI, waiting on next deploy :smile: surely that will be very quick.

exterkamp:
SURELY this is happening soon :sweat_smile:

exterkamp:
Done w/5.2!
![image](https://user-images.githubusercontent.com/6392995/62984592-a92ef500-bde8-11e9-84e8-2e13f436e50f.png)

connorjclark:
neat

PR #9030, opened 2019-05-22
renderer: fix word break issue for display text, move metric disclaimer

Hoten:
Before:

image

image

image

After:

image

image

image

See #9026

### DIFF
--- a/lighthouse-core/report/html/report-styles.css
+++ b/lighthouse-core/report/html/report-styles.css
@@ -478,7 +478,7 @@
   margin-right: 0;
 }
 
-.lh-audit__display-text {
+.lh-audit__title-and-text {
   flex: 1;
 }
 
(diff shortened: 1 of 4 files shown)


lighthouse-core/report/html/report-styles.css
```diff
@@ -480,6 +480,7 @@
 
 .lh-audit__display-text {
   flex: 1;
+  word-break: normal;
```
patrickhulce: I would love to see a diagram of our web of word-break fixes over the course of LH history 😆

paulirish:
I was attempting a fix for the wrapping myself yesterday. I just pushed it here. Here's the difference between the two:

| | |
|-|-|
| image |  image |

sorry for jumping into the PR

Hoten:
fiiiine. only cuz it's better √√

lgtm

PR #9036, opened 2019-05-23
core: add traceDuration to diagnostics audit

paulirish:
exposing this would help in cases where we have two runs like this:

image


brendankenny:
FWIW, this is in `metrics` as `observedTraceEnd` or `observedTraceEndTs`

patrickhulce:
To @brendankenny 's point, the diagnostics was originally a collection of junk that's otherwise hard to grab from the JSON. Since this is easily grabbed from metrics I'd be inclined to keep out of here, but was there a specific reason you needed it in diagnostics?

paulirish:
All good. I definitely thought it landed.. just didn't go looking in `metrics` for it. :)

Issue #9050, opened 2019-05-24
Improve `--plugins` resolution algorithm


alekseykulikov:
First of all, thank you for adding the way to define plugins. It's a fantastic feature!

We've just published the [lighthouse-plugin-field-performance](https://github.com/treosh/lighthouse-plugin-field-performance)  and discovered a few issues with `--plugins` resolution algorithm.

**The story**:
1. Start local development using the recipe. It's not possible to run the current folder as a plugin since Lighthouse expects *lighthouse-plugin-* prefix. I have added the folder `lighthouse-plugin-field-performance` with all the related code.
2. Release 🎉Check the published module, but `--plugins` algorithm does not recognize `package.json`'s `main` field and can't find the code. So we used the exact structure from [the recipe](https://github.com/GoogleChrome/lighthouse/tree/master/docs/recipes/lighthouse-plugin-example) and `ln -s ../ node_modules/lighthouse-plugin-field-performance` a symlink to use the current directory as a part of node_modules.
3. `plugin.js` is a wrong name since it can only resolve `index.js`.

The whole story explained with [commit messages](https://github.com/treosh/lighthouse-plugin-field-performance/commits/master):
<img width="447" alt="Screenshot 2019-05-24 at 16 22 45" src="https://user-images.githubusercontent.com/158189/58334970-dbac0000-7e40-11e9-9bbe-bd423f1fa739.png">

**Conclusion**
In order to make the development of the plugins more straightforward, it would be great to **improve `--plugins` resolution algorithm** by adding support for `package.json`'s `main` field and checking if the current folder is a lighthouse plugin. (for example, checking the name in `package.json`).

patrickhulce:
Thanks for filing and being such an early adopter! :D

The behavior you'd like is the current intended behavior and the `lighthouse-plugin-example` works locally for me using `"main": "./plugin.js"` so we definitely want to make sure this is the experience for plugin developers as well.

A few notes:

> It's not possible to run the current folder as a plugin since Lighthouse expects lighthouse-plugin- prefix.

Right now I get this to work by doing `yarn add ./path/to/lighthouse-plugin-folder` at which point its put into `node_modules` with the proper structure.

> --plugins algorithm does not recognize package.json's main field and can't find the code

Our algorithm is `require.resolve` so if it can be required exactly as you pass it in to `--plugins` then it should work with us. If that doesn't work we start falling back to some workarounds but if it's in node_modules and `require.resolve` works, it should work here too. 

[...]

PR #9051, opened 2019-05-24
docs(plugins): add link to field-performance plugin

paulirish:
Followup to #9049.. linking here will be useful for prospective plugin authors :)

--------

Also...  

@alekseykulikov @denar90 I was hoping to get some feedback from you guys on writing a plugin.. It looks like there were some rough edges and I think we could improve docs and perhaps some APIs to make it more straightforward for the next plugin :)

What did you notice? What was annoying? What was hard to figure out? What could we document better? :)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -350,5 +350,6 @@ Most artifacts will try to represent as truthfully as possible what was observed
 
 ## Examples
 
+- [Field Performance](https://github.com/treosh/lighthouse-plugin-field-performance) - A plugin to gather and display Chrome UX Report field data
 - [Publisher Ads Audits](https://github.com/googleads/pub-ads-lighthouse-plugin) - a well-written, but complex, plugin
 - [Lighthouse Plugin Recipe](./recipes/lighthouse-plugin-example)


denar90:
Hi, thanks, folks :)

We were so happy to have that API and really enjoyed working on the plugin. Other than #9050 #9052 I can add about html rendering.

The first intention was to make performance score representation like it's for "performance" category, but digging into the code we've noticed it's not so easy without patching LH.

> There are special rendering for PWA and Performance - https://github.com/GoogleChrome/lighthouse/blob/v5.0.0/lighthouse-core/report/html/renderer/report-renderer.js#L227-L230. Will be nice to accept the custom for other categories.

So then we switched to table view which we found perfect for distribution representation. What we also brainstormed about there, were:
- style row or table somehow (to underline slow, average, fast  categories and their values)
- accept values as html. I had this crazy idea to render some graph or other visualization via preact and then do render-to-string to pass the result as value (plain html).

One more, mixing static audit `meta` data with dynamic values. We couldn't add some info about the current page and origin like:

image

brendankenny:
> fantastically written example 👍

ha, it took me a while to realize you weren't sarcastically referring to @paulirish's one line change here :)

> https://github.com/GoogleChrome/lighthouse/pull/9051#issuecomment-495779502

Good feedback, we should capture in some issues in addition to #9053

patrickhulce:
>  it took me a while to realize you weren't sarcastically referring to @paulirish's one line change here 

Haha, yes sorry that was intended as a compliment to @denar90 and @alekseykulikov , no dig on @paulirish :)

alekseykulikov:
Hi all!

In addition to @denar90's reply, I could list a few minor issues:
- We use TypeScript with `allowJs` (similar to your setup), and validation requires `devtools-protocol`, which is a part of `devDependencies` of Lighthouse and not installed. It required us to add `devtools-protocol` to our [package.json](https://github.com/treosh/lighthouse-plugin-field-performance/blob/master/package.json#L44) to bypass the validator.
- More recommendations regarding scoring performance values would be helpful for plugin authors. We used Lighthouse's `Audit.computeLogNormalScore` to score FCP/FID ([code](https://github.com/treosh/lighthouse-plugin-field-performance/blob/master/lighthouse-plugin-field-performance/utils/audit-helpers.js#L56)), and I'm not sure is it part of the stable API or not.
- I would review the assumption, that plugin can't gather custom data from the page (artifacts). `field-performance` plugin fetches the data from PSI API, then cache it internally to use in the rest of audits. ([code](https://github.com/treosh/lighthouse-plugin-field-performance/blob/master/lighthouse-plugin-field-performance/utils/audit-helpers.js#L37)). It's possible to cache internally, but still would be valuable to hook to Lighthouse's gather process. For example, if @patrickhulce would build a third-party plugin, he would need to make another Lighthouse run without all 3rd-party scripts, which is not possible now.

I have some feedback about PSI API/UI, and I will send you @paulirish by email :)

PR #9054, opened 2019-05-24
report: show metric descriptions by default when errors exist

Hoten:
See https://github.com/GoogleChrome/lighthouse/issues/9019#issuecomment-494961806

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -119,7 +119,9 @@ class ReportUIFeatures {
     }
 
     // Show the metric descriptions by default when there is an error.
-    if (report.audits.metrics.errorMessage) {
+    const hasMetricError = report.categories.performance.auditRefs
+      .some(audit => Boolean(audit.group === 'metrics' && report.audits[audit.id].errorMessage));
+    if (hasMetricError) {
       const toggleInputEl = /** @type {HTMLInputElement} */ (
         this._dom.find('.lh-metrics-toggle__input', this._document));
       toggleInputEl.checked = true;
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -119,7 +119,7 @@ class ReportUIFeatures {
     }
 
     // Show the metric descriptions by default when there is an error.
-    const hasMetricError = report.categories.performance.auditRefs
+    const hasMetricError = report.categories.performance && report.categories.performance.auditRefs
       .some(audit => Boolean(audit.group === 'metrics' && report.audits[audit.id].errorMessage));
     if (hasMetricError) {
       const toggleInputEl = /** @type {HTMLInputElement} */ (
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -117,6 +117,13 @@ class ReportUIFeatures {
       this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
       window.addEventListener('resize', this._updateStickyHeaderOnScroll);
     }
+
+    // Show the metric descriptions by default when there is an error.
+    if (report.audits.metrics.errorMessage) {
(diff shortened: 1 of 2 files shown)


lighthouse-core/test/report/html/renderer/report-ui-features-test.js
```diff
@@ -239,4 +239,19 @@ describe('ReportUIFeatures', () => {
       assert.ok(container.querySelector('.score100'), 'has fireworks treatment');
     });
   });
+
+  describe('metric descriptions', () => {
+    it('with no errors, hide by default', () => {
+      const lhr = JSON.parse(JSON.stringify(sampleResults));
+      const container = render(lhr);
+      assert.ok(!container.querySelector('.lh-metrics-toggle__input').checked);
```
Hoten: these assertions seem questionable..

Hoten: as in, I'm very explicitly testing against the if statement. idk seems odd

patrickhulce: yeah it feels like we really just want to assert that the errors are visible, do we do any gCS in these UI tests yet?

Hoten: gCS?

I first tried to do something like "are the description elements visible", but I don't think we can do that b/c jsdom doesn't try to surface stuff like that. as in, there is no rendering/css engine.

patrickhulce: I thought styles directly applied to elements still resolved though? I could be wrong I just thought there was no cascading.

Hoten: Ah, so I was actually thinking of _layout_ (from reading [this](https://github.com/jsdom/jsdom/issues/1048#issuecomment-76524044)). Tried a bit more, but still no dice:

image

Hoten: LOL I forgot we shim getComputedStyles in this test. I thought it was provided by jsdom. anyways, idk how to do this.

Hoten: so if we don't overwrite their `getComputedStyles`, this happens:

image

weird

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -117,6 +117,13 @@ class ReportUIFeatures {
       this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
       window.addEventListener('resize', this._updateStickyHeaderOnScroll);
     }
+
+    // Show the metric descriptions by default when there is an error.
+    if (report.audits.metrics.errorMessage) {
```
patrickhulce: worth noting that TTI, speed index, and FCPUI will just be undefined and not throw an errorMessage onto metrics so this will only handle it for fcp/fmp/eil/traceOfTab errors

but that seems to be the case we want to auto-expand anyhow, when all of them fail basically?

Hoten: My intention was to expand if there are any "Error"s in the metric table. I assumed audits.metrics.errorMessage would be that, but that's probably a bad assumption...

patrickhulce: ah you'll probably want to switch to `report.audits['first-contentful-paint'].errorMessage || report.audits.interactive.errorMessage || ...` then :/

Hoten: figured it out

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -117,6 +117,15 @@ class ReportUIFeatures {
       this._document.addEventListener('scroll', this._updateStickyHeaderOnScroll);
       window.addEventListener('resize', this._updateStickyHeaderOnScroll);
     }
+
+    // Show the metric descriptions by default when there is an error.
+    const hasMetricError = report.categories.performance && report.categories.performance.auditRefs
```
patrickhulce: even better :)

PR #9056, opened 2019-05-25
core(perf): speed up tap-target's isVisible()

paulirish:
thanks jquery!
https://github.com/jquery/jquery/blob/master/src/css/hiddenVisibleSelectors.js

Seems good...


paulirish:
Looking at the failure...

This implementation includes this node from the smoketest, but it should not:

```html
 <a style="display: inline-block; width: 0; overflow-x: hidden;height: 1px">
        width 0 and overflow x hidden
      </a>
```

lighthouse-core/gather/gatherers/seo/tap-targets.js
```diff
@@ -32,38 +32,12 @@ const TARGET_SELECTORS = [
...
-  }
-
-  const parent = element.parentElement;
-  if (parent && parent.tagName !== 'BODY') {
-    // if a parent is invisible then the current element is also invisible
-    return elementIsVisible(parent);
-  }
-
-  return true;
+  return !!(element.offsetWidth || element.offsetHeight || element.getClientRects().length);
```
patrickhulce: to fix the failure can we check if any of these dimensions are 0?

it feels like everything was so carefully written the first time around though I'm not sure I fully understand how these compare...

paulirish: Yeahh... something with a 1px height but 0x width is still invisible.  doesnt matter the height.

Not sure what the test case was supposed to be for. @mattzeunert

mattzeunert: The test failure is fixed after merging master, because the node is filtered out by `elementCenterIsAtZAxisTop`. The [position absolute PR](https://github.com/GoogleChrome/lighthouse/pull/7778/files) also generally improves the smoke test a bunch.

> Not sure what the test case was supposed to be for.

The tap target can't be tapped on because the content is hidden by the `overflow: hidden`. We care about the total number of tap targets in the artifacts because the pass rate is `1 - failureCount/totalTapTargets`, and we don't want invisible elements to bring the score up.

> to fix the failure can we check if any of these dimensions are 0?

Just because the element has a width of 0 doesn't mean the tap target is untappable, because there could be overflowing child content that the user can tap on.

I added a test case for a `width: 0` element with visible child content.
 
> it feels like everything was so carefully written the first time around though I'm not sure I fully understand how these compare...

I think this change is fine, we have reasonable smoke test coverage and I tried it on 50 sites and didn't notice any changes.

I think the new `elementIsVisible` will deem more elements visible than before, but they'll then be filtered out by `elementCenterIsAtZAxisTop`.

lighthouse-cli/test/fixtures/seo/seo-tap-targets.html
```diff
@@ -159,6 +159,24 @@ <h1>SEO Tap targets</h1>
 
     <br/><br/>
 
+    <!-- even if a tap target's size is 0x0px it can still fail if overflowing content is visible -->
+    <div>
+      <a
+        data-gathered-target="zero-width-tap-target-with-overflowing-child-content"
+        style="display: block; width: 0; white-space: nowrap">
+        <!-- TODO: having the span should not be necessary to cause a failure here, but
```
mattzeunert: I think we could do this with https://stackoverflow.com/a/28457044/1290545. Konrad's original code actually used it, but then neither of us could figure out what it was needed for 🙂

mattzeunert:
> Is the test failure expected given your smoketest changes though @mattzeunert? Or is that a bug that needs fixing?

Seems like CI rendered the link 1px less tall than on my local.

PR #9070, opened 2019-05-28
report: increase UI density for devtools

paulirish:


lighthouse-core/report/html/report-styles.css
```diff
@@ -261,11 +261,27 @@
   --audit-group-indent: 16px;
   --audit-indent: 16px;
   --expandable-indent: 16px;
-  --gauge-circle-size-big: 96px;
-  --gauge-circle-size: 72px;
   /* Consider offset of the DevTools UI. */
   --top-of-page-offset: 54px;
 
+  --gauge-circle-size-big: 72px;
+  --gauge-circle-size: 64px;
```
paulirish: these are adjusted

lighthouse-core/report/html/report-styles.css
```diff
@@ -261,11 +261,27 @@
...
+
+  --audits-margin-bottom: 20px;
```
paulirish: this block is copy/pasted from the narrow media query

though `score-container-width` is also reduced to get all 5 icons without wrapping.

Issue #8638, opened 2019-04-26
Total timing is equal to 0 always via API


jmartinezpoq:
When im running lighthouse report via API the result in the JSON  'lighthouseResult' -> 'timing' -> 'total' is always 0

1. Run LH on https://www.google.com

2. https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=https://www.google.com&strategy=desktop&category=accessibility&category=performance&category=pwa&category=best-practices&category=seo

3. Response
![image](https://user-images.githubusercontent.com/47970682/56790159-1ac84080-6804-11e9-8f58-f11a0850da51.png)

#### What is the current behavior?
The response for all the websites is always 0

#### What is the expected behavior?
This is happening since April 17th, before it returned the total time in millis

#### Environment Information
* Lighthouse version: 4.3.0

exterkamp:
So I can confirm this.  I'll take a deeper look into it.

Hoten:
Corresponds with 4.3 in LR.

walterfrancis:
Thank you @jmartinezpoq I had been seeing this too and had not had time to investigate why a test tool of mine using PSI was failing.

jmartinezpoq:
any updates?

jmartinezpoq:
any updates? @exterkamp

paulirish:
@exterkamp this is a valuable signal. Let's get a fix in prod for this soon.

jmartinezpoq:
hi, 
this is still happening.
@exterkamp @paulirish

paulirish:
It's now fixed for PSI.

PR #9072, opened 2019-05-28
core(runner): fix missing timing properties in the browser

exterkamp:
**Summary**
PerformanceEntries in Lightrider (and the extension) had no names and no entryTypes.  This manifested as a 0 `timing.total` in the API as well.  

@brendankenny found that browser PerformanceEntry objects do not spread, but in Node they do b/c of the marky lib.  So, simple fix to explicitly call out the `name` and `entryType` fields when parsing the entries.  This will also populate the missing `lh:runner:run` entry to fill in `timing.total`.

**Related Issues/PRs**
fixes: #8638

### DIFF
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -186,6 +186,8 @@ class Runner {
     // resolution.
     .map(entry => {
       return /** @type {PerformanceEntry} */ ({
+        // Don't spread entry because browser PerformanceEntries can't be spread.
+        // https://github.com/GoogleChrome/lighthouse/issues/8638
         startTime: parseFloat(entry.startTime.toFixed(2)),
         name: entry.name,
         duration: parseFloat(entry.duration.toFixed(2)),
--- a/lighthouse-core/runner.js
+++ b/lighthouse-core/runner.js
@@ -186,7 +186,8 @@ class Runner {
     // resolution.
     .map(entry => {
       return /** @type {PerformanceEntry} */ ({
-        ...entry,
+        name: entry.name,
+        entryType: entry.entryType,
         duration: parseFloat(entry.duration.toFixed(2)),
         startTime: parseFloat(entry.startTime.toFixed(2)),
       });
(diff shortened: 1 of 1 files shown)


lighthouse-core/runner.js
```diff
@@ -186,7 +186,8 @@ class Runner {
     // resolution.
     .map(entry => {
       return /** @type {PerformanceEntry} */ ({
-        ...entry,
+        name: entry.name,
+        entryType: entry.entryType,
         duration: parseFloat(entry.duration.toFixed(2)),
         startTime: parseFloat(entry.startTime.toFixed(2)),
```
brendankenny: can you reorder so it's (`startTime`, `name`, `duration`, `entryType`) just to prevent the churn in `sample_v2.json` :)

exterkamp: We found that at the same time! Done.

brendankenny:
> @brendankenny found that browser PerformanceEntry objects do not spread, but in Node they do b/c of the marky lib

`marky` makes its own `PerformanceEntry` objects, which spread just fine, but the properties on a native `PerformanceEntry` are getters in the prototype chain, so nothing is spread.

Interestingly, running LH in DevTools seems to defeat marky's environment sniffing and so it uses the fallback objects and the entries are ok. The extension is like lr here and only has `{startTime, duration}`

lighthouse-core/runner.js
```diff
@@ -186,9 +186,10 @@ class Runner {
     // resolution.
     .map(entry => {
       return /** @type {PerformanceEntry} */ ({
-        ...entry,
-        duration: parseFloat(entry.duration.toFixed(2)),
         startTime: parseFloat(entry.startTime.toFixed(2)),
+        name: entry.name,
```
patrickhulce: what to add a comment with link to this issue on why we can't spread? :)

Hoten:
can we add a hexa test to LR

PR #9073, opened 2019-05-28
report: show fireworks only if all core categories score 100

### LINKED ISSUE #9058: Fireworks don't work for solo category
```
node lighthouse-cli https://exterkamp.codes --view --only-categories=seo
```

b/c we do all the fireworks magic in the score header. Which has zero height in solo category mode.

Hoten:
Fixes #9058

### DIFF
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -25,9 +25,6 @@
 
 /* globals getFilenamePrefix Util */
 
-/** @typedef {import('./dom.js')} DOM */
-/** @typedef {import('./util.js')} Util */
-
 /**
  * @param {HTMLTableElement} tableEl
  * @return {Array<HTMLTableRowElement>}
--- a/lighthouse-core/report/html/renderer/report-ui-features.js
+++ b/lighthouse-core/report/html/renderer/report-ui-features.js
@@ -26,6 +26,7 @@
 /* globals getFilenamePrefix Util */
 
 /** @typedef {import('./dom.js')} DOM */
+/** @typedef {import('./util.js')} Util */
 
 /**
  * @param {HTMLTableElement} tableEl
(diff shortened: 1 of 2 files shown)


lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -98,7 +99,9 @@ class ReportUIFeatures {
 
     // Fireworks.
     const scoresAll100 = Object.values(report.categories).every(cat => cat.score === 1);
-    if (!this._dom.isDevTools() && scoresAll100) {
+    const hasAllCoreCategories =
```
patrickhulce: hm, I guess this would still get fireworks for custom config categories but if the user is going all that way to get some fireworks they probably deserve 'em 😆

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -98,7 +99,9 @@ class ReportUIFeatures {
...
+      Object.keys(report.categories).filter(id => !Util.isPluginCategory(id)).length >= 5;
+    if (!this._dom.isDevTools() && scoresAll100 && hasAllCoreCategories) {
```
patrickhulce: brave soul creating merge conflicts for yourself 😉

lighthouse-core/report/html/renderer/report-ui-features.js
```diff
@@ -26,6 +26,7 @@
 /* globals getFilenamePrefix Util */
 
 /** @typedef {import('./dom.js')} DOM */
+/** @typedef {import('./util.js')} Util */
```
brendankenny: doesn't the `html-renderer.d.ts` take care of this? (I don't think either typedef is needed)

lighthouse-core/report/html/renderer/report-ui-features.js
brendankenny: I kind of feel like it should be at least perf + any other categories need to be 100. So solo perf 100 would be ok.

But I also don't care that much and this seems fine :)

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: ya, removed

lighthouse-core/report/html/renderer/report-ui-features.js
Hoten: solo + fireworks would require a refactor, since the fireworks were designed to work with the scores container header.

Issue #9048, opened 2019-05-24
Avoid an excess DOM size goes "Error!"


Westbrook:
When auditing https://westbrook.github.io/minesweeper/ the "Avoid an excess DOM size" returns `Error!`.

#### Provide the steps to reproduce
1. Visit https://westbrook.github.io/minesweeper/
2. Run DevTools audit
3. See "Avoid an excess DOM size" report

#### What is the current behavior?
![image](https://user-images.githubusercontent.com/1156657/58328417-07f35c80-7e00-11e9-82c2-bceda3ed4d75.png)

#### What is the expected behavior?
It tells me something useful... right?

#### Environment Information
* Affected Channels: DevTools
* Lighthouse version: Stable & Canary
* Operating System: Mac OS

**Related issues**

patrickhulce:
Thanks for filing @Westbrook!

The full error message can be seen on hover, and this is indeed a bug!

![image](https://user-images.githubusercontent.com/2301202/58336537-830d4080-7e09-11e9-8a52-1cb9922452d7.png)

Westbrook:
Oh, great! Thanks for pointing that out, it will be super helpful if I run into future issues.

I would never have caught that bubble interaction, however. Seems like there might be room for some UI/X love on that (being the visuals tell me to toggle, not hover).

If there is anything else I can do to support clearing up this issue, please let me know.

patrickhulce:
For anyone that might want to take this on, the error is being thrown here

https://github.com/GoogleChrome/lighthouse/blob/df85d681886d1ae163798c715fd29c72cc96c58f/lighthouse-core/lib/page-functions.js#L115-L126

A good fix might be to simply `try/catch` and return `<${node.localName}>` if we can't determine the full outerHTML :)

NickolasBenakis:
I would like to work on this

Hoten:
thanks @NickolasBenakis!

Tried to assign you but I guess only project members can be assigned to issues :)

NickolasBenakis:
seems harder than I thought, I tried the `try/catch` senario but it gives me the same error again...
I am still working on this

NickolasBenakis:
I fixed it!

brendankenny:
Fixed by #9079

PR #2131, opened 2017-05-03
DOMStat Audit: shadow roots don't have .classList

ebidel:
R: all

Fixes https://github.com/GoogleChrome/lighthouse/issues/1986.

PR #9079, opened 2019-05-29
core(page-functions): don't try to clone a ShadowRoot

NickolasBenakis:
**Summary**
This PR fixes the  #9048 issue .
image

It is a bug fix.

### DIFF
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -6,7 +6,7 @@
 // @ts-nocheck
 'use strict';
 
-/* global window document Node */
+/* global window document Node ShadowRoot */
 
 /**
  * Helper functions that are passed by `toString()` by Driver to be evaluated in target page.
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -107,14 +107,15 @@ function getElementsInDocument(selector) {
 
 /**
  * Gets the opening tag text of the given node.
- * @param {Element} element
+ * @param {Element|ShadowRoot} element
  * @param {Array<string>=} ignoreAttrs An optional array of attribute tags to not include in the HTML snippet.
  * @return {string}
  */
 /* istanbul ignore next */
 function getOuterHTMLSnippet(element, ignoreAttrs = []) {
   try {
-    if (element instanceof ShadowRoot && element.host && element.localName !== 'a') {
+    // ShadowRoots are sometimes passed in; use their hosts' outerHTML.
+    if (element instanceof ShadowRoot) {
       element = element.host;
     }
 
@@ -126,6 +127,7 @@ function getOuterHTMLSnippet(element, ignoreAttrs = []) {
     const match = clone.outerHTML.match(reOpeningTag);
     return (match && match[0]) || '';
   } catch (_) {
+    // As a last resort, fall back to localName.
     return `<${element.localName}>`;
   }
 }
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -114,7 +114,7 @@ function getElementsInDocument(selector) {
 /* istanbul ignore next */
 function getOuterHTMLSnippet(element, ignoreAttrs = []) {
   try {
-    if (ShadowRoot.prototype.isPrototypeOf(element) && element.host && element.localName !== 'a') {
+    if (element instanceof ShadowRoot && element.host && element.localName !== 'a') {
       element = element.host;
     }
 
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -113,17 +113,19 @@ function getElementsInDocument(selector) {
  */
 /* istanbul ignore next */
 function getOuterHTMLSnippet(element, ignoreAttrs = []) {
-  const clone = element.cloneNode();
-
-  ignoreAttrs.forEach(attribute =>{
-    clone.removeAttribute(attribute);
-  });
-
-  const reOpeningTag = /^[\s\S]*?>/;
-  const match = clone.outerHTML.match(reOpeningTag);
-
-  return (match && match[0]) || '';
+  try {
+    const clone = element.cloneNode();
+    ignoreAttrs.forEach(attribute =>{
+      clone.removeAttribute(attribute);
+    });
+    const reOpeningTag = /^[\s\S]*?>/;
+    const match = clone.outerHTML.match(reOpeningTag);
+    return (match && match[0]) || '';
+  } catch (_) {
+    return `<${element.localName}>`;
+  }
 }
+
 
 /**
  * Computes a memory/CPU performance benchmark index to determine rough device class.
--- a/lighthouse-core/lib/page-functions.js
+++ b/lighthouse-core/lib/page-functions.js
@@ -114,6 +114,10 @@ function getElementsInDocument(selector) {
 /* istanbul ignore next */
 function getOuterHTMLSnippet(element, ignoreAttrs = []) {
   try {
+    if (ShadowRoot.prototype.isPrototypeOf(element) && element.host && element.localName !== 'a') {
+      element = element.host;
+    }
+
     const clone = element.cloneNode();
     ignoreAttrs.forEach(attribute =>{
       clone.removeAttribute(attribute);
--- a/lighthouse-core/test/lib/page-functions-test.js
+++ b/lighthouse-core/test/lib/page-functions-test.js
@@ -43,7 +43,7 @@ describe('Page Functions', () => {
       ), '<div id="1">');
     });
 
-    it('should handle dom nodes that cannot be inspected', () => {
+    it('should handle dom nodes that cannot be cloned', () => {
       const element = dom.createElement('div');
       element.cloneNode = () => {
         throw new Error('oops!');
--- a/lighthouse-core/test/lib/page-functions-test.js
+++ b/lighthouse-core/test/lib/page-functions-test.js
@@ -43,6 +43,13 @@ describe('Page Functions', () => {
       ), '<div id="1">');
     });
 
+    it('should handle dom nodes that cannot be inspected', () => {
+      const element = dom.createElement('div');
+      element.cloneNode = () => {
+        throw new Error('oops!');
+      };
+      assert.equal(pageFunctions.getOuterHTMLSnippet(element), '<div>');
+    });
     it('ignores when attribute not found', () => {
       assert.equal(pageFunctions.getOuterHTMLSnippet(
         dom.createElement('div', '', {'id': '1', 'style': 'style', 'aria-label': 'label'}),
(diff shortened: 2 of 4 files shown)


NickolasBenakis:
> Thanks for your pull request. It looks like this may be your first contribution to a Google open source project (if not, look below for help). Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 **Please visit https://cla.developers.google.com/ to sign.**
> 
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.
> 
> #### What to do if you already signed the CLA
> ##### Individual signers
> * It's possible we don't have your GitHub username or you're using a different email address on your commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> 
> ##### Corporate signers
> * Your company has a Point of Contact who decides which employees are authorized to participate. Ask your POC to be added to the group of authorized contributors. If you don't know who your Point of Contact is, direct the Google project maintainer to [go/cla#troubleshoot](http://go/cla#troubleshoot) ([Public version](https://opensource.google.com/docs/cla/#troubleshoot)).
> * The email used to register you as an authorized contributor must be the email used for the Git commit. Check [your existing CLA data](https://cla.developers.google.com/clas) and verify that your [email is set on your git commits](https://help.github.com/articles/setting-your-email-in-git/).
> * The email used to register you as an authorized contributor must also be [attached to your GitHub account](https://github.com/settings/emails).
> 
> ℹ️ **Googlers: [Go here](https://goto.google.com/prinfo/https%3A%2F%2Fgithub.com%2FGoogleChrome%2Flighthouse%2Fpull%2F9079) for more info**.

I signed it!

lighthouse-core/lib/page-functions.js
```diff
@@ -113,18 +113,20 @@ function getElementsInDocument(selector) {
...
+  try {
+    const clone = element.cloneNode();
+    ignoreAttrs.forEach(attribute =>{
+      clone.removeAttribute(attribute);
+    });
+    const reOpeningTag = /^[\s\S]*?>/;
+    const match = clone.outerHTML.match(reOpeningTag);
+    return (match && match[0]) || '';
+  } catch (error) {
+    return element.localName;
```
patrickhulce:
```suggestion
    return `<${element.localName}>`;
```

lighthouse-core/lib/page-functions.js
```diff
@@ -113,18 +113,20 @@ function getElementsInDocument(selector) {
...
-  return (match && match[0]) || '';
+  try {
+    const clone = element.cloneNode();
+    ignoreAttrs.forEach(attribute =>{
+      clone.removeAttribute(attribute);
+    });
+    const reOpeningTag = /^[\s\S]*?>/;
+    const match = clone.outerHTML.match(reOpeningTag);
+    return (match && match[0]) || '';
+  } catch (error) {
```
patrickhulce:
```suggestion
  } catch (_) {
```

we name our unused variables `_` :)

lighthouse-core/lib/page-functions.js
patrickhulce: just to match the expected output of this function under normal conditions

NickolasBenakis: Thanks a lot for the feedback @patrickhulce .
You want me to remove completely the `package-lock.json` ?
I am checking the a11y tester case now, I 've never done this before :) 
But I am trying to figure out

NickolasBenakis: image
It seems that it pass,
but I didnt add any error-prone element inside the a11y_tester.html.

patrickhulce: That's expected :)

Steps:

1. Temporarily comment out your fix
1. Add an error-prone element inside a11y_tester.html
2. Run `yarn smoke a11y`, it should fail
3. Uncomment your fix
2. Run `yarn smoke a11y`, it should pass
4. Commit the changes! 🎉

NickolasBenakis: The error-prone element should be printed somewhere in the bug report right?

NickolasBenakis: my problem is that i can't find an error-prone element to add it in the smoke a11y test,
Is this performance report supposed to show me an element?
or is there another report that displays an the error element?

patrickhulce: hm it looks like the problematic element has no `localName` too `<undefined>` is definitely not what we want to be returning here.

patrickhulce: OK I've done a bit of digging and this turned out to be a bit more complicated than it appeared, sorry about that! I'll go ahead and add the test to your PR :)

patrickhulce: OK new plan!

Add a test to https://github.com/GoogleChrome/lighthouse/blob/767116696fdb1fc0903d369f811005a1cdae8fbb/lighthouse-core/test/lib/page-functions-test.js#L48-L52 that tests we can handle shadow root elements and elements that fail when their `cloneNode` is called :)

i.e. `element.cloneNode = () => { throw new Error('Oops!') }` and make sure it recovers

NickolasBenakis: image

NickolasBenakis: you mean something like this?

NickolasBenakis: image

NickolasBenakis: meh either way,
I run the test with `npm run test`
and it crashes... 
well :'(

patrickhulce: I meant something like

```js
it('should handle dom nodes that cannot be inspected', () => {
  const element = dom.createElement('div');
  element.cloneNode = () => { throw new Erorr('oops!') };
  assert.equal(pageFunctions.getOuterHTMLSnippet(element), '<div>');
});

Comment (NickolasBenakis) [reply lighthouse-core/lib/page-functions.js]: @patrickhulce  I commited the test case :)

File: lighthouse-core/lib/page-functions.js [hunk 3]
```diff
@@ -113,18 +113,24 @@ function getElementsInDocument(selector) {
...
-  return (match && match[0]) || '';
+    const clone = element.cloneNode();
+    ignoreAttrs.forEach(attribute =>{
+      clone.removeAttribute(attribute);
+    });
+    const reOpeningTag = /^[\s\S]*?>/;
+    const match = clone.outerHTML.match(reOpeningTag);
+    return (match && match[0]) || '';
+  } catch (_) {
+    return `<${element.localName}>`;
```
Hoten: can this entire range fail? if it's just the first couple statements, could you reduce the try/catch to just that part?

a comment in the catch block would be useful.

lighthouse-core/lib/page-functions.js
```diff
@@ -113,18 +113,24 @@ function getElementsInDocument(selector) {
...
-  const clone = element.cloneNode();
-
-  ignoreAttrs.forEach(attribute =>{
-    clone.removeAttribute(attribute);
-  });
-
-  const reOpeningTag = /^[\s\S]*?>/;
-  const match = clone.outerHTML.match(reOpeningTag);
+  try {
+    if (ShadowRoot.prototype.isPrototypeOf(element) && element.host && element.localName !== 'a') {
```
brendankenny: nit:
```suggestion
    if (element instanceof ShadowRoot && element.host && element.localName !== 'a') {
```
reads a little simpler than the prototype check, but

non-nit: can we also add a comment to this line? Both to explain the shadow root case, but also to explain the `localName` check. I don't really understand it...`ShadowRoot` doesn't have a `localName`, so by that point in the conditional won't the comparison always be true?

brendankenny: (also according to spec ["Shadow roots’s associated `host` is never null."](https://dom.spec.whatwg.org/#interface-shadowroot))

lighthouse-core/lib/page-functions.js
```diff
@@ -113,18 +113,24 @@ function getElementsInDocument(selector) {
...
+      element = element.host;
+    }
 
-  return (match && match[0]) || '';
+    const clone = element.cloneNode();
+    ignoreAttrs.forEach(attribute =>{
+      clone.removeAttribute(attribute);
```
brendankenny: the types on this whole function get kind of broken if we aren't really accepting just `Element`.

`removeAttribute` and `outerHTML` (and more or less `localName`) exist only on `Element`.

Maybe we should type it as `@param {Element|ShadowRoot} element` and (if we don't need that `localName` check) have the conditional be just `if (element instanceof ShadowRoot) element = element.host;` (with a comment somewhere explaining that this function handles the ShadowRoot case).

It's still playing a little fast and loose with types (it doesn't help that the tsc built in types don't recognize `ShadowRoot` as a global value), but it simplifies things a bit.

lighthouse-core/test/lib/page-functions-test.js
```diff
@@ -38,6 +43,13 @@ describe('Page Functions', () => {
       ), '<div id="1">');
     });
 
+    it('should handle dom nodes that cannot be inspected', () => {
```
brendankenny:
```suggestion
    it('should handle dom nodes that cannot be cloned', () => {
```
maybe?

exterkamp:
Just updating the CLA

lighthouse-core/lib/page-functions.js
brendankenny: @patrickhulce since you added this line, what's the deal with the `element.localName !== 'a'` check?

patrickhulce: This entire piece is straight up copied from the dom-size audit that filters this way, so you'd have to ask Eric as to why it was written this way :)

It indeed seemed like overkill and I thought about trying to extract this to a `isShadowRoot` function with a canonical version to do this or something but bringing in all the baggage of `page-functions` with dependencies into this PR seemed unfair.

I don't see why we can't just use `element instanceof ShadowRoot` personally, but I'm assumed Eric thought of something I didn't which is a pretty safe bet when it comes to our comparative web component/shadow dom knowledge 😆

NickolasBenakis: :P I commited the suggested changes, 
is there any other issue that we need to tackle?

brendankenny: > This entire piece is straight up copied from the dom-size audit that filters this way, so you'd have to ask Eric as to why it was written this way :)

OK! After consulting the old texts (#2131), I think this was from adapting earlier code (which identified shadow roots just by looking for an `element.host` but had to rule out anchors because they also have a `host` property). That code was wrapped without modification in the `ShadowRoot.prototype` check in that PR, which I believe was the error (but since it's always true, it doesn't cause any bad behavior).

I'm going to remove based on the spec and my own testing saying it's fine. I'll take responsibility for any failures :)

NickolasBenakis:
> LGTM! Good to go when the build is green.
> 
> Thanks for sticking with this, @NickolasBenakis!

My pleasure...
This is one of my first steps in open source and I had fun doing this.
I will keep my eyes open for any future issue that I could help.
=)

PR #9086, opened 2019-05-30
5.1.0

Hoten:


patrickhulce:
fixtures need some love though

Hoten:
> LGTM!
> 
> I'm curious @Hoten are you using the scripts at all? Are they helpful? How can they be more helpful?

`./lighthouse-core/scripts/release/test.sh`

I ended up commenting out the smoke parts because it kept flaking. Would be good to prompt "Smokes failed, was it a flake? Try again? Continue?" etc.

Maybe skip the first test run if it's just going to run the tests again via the npm package.

```
echo "${TXT_DIM}Press any key to see the git diff, CTRL+C to exit...${TXT_RESET}"
read -n 1 -r unused_variable
git --no-pager diff HEAD^
```

I was confused on this part, because I think you can't page up/down to see the whole diff.

changelog.md
```diff
@@ -1,3 +1,76 @@
...
+* config: split out config helpers ([#9003](https://github.com/googlechrome/lighthouse/pull/9003))
+* gather-runner: treat NO_FCP as a pageLoadError ([#8340](https://github.com/googlechrome/lighthouse/pull/8340))
+* gather-runner: always reset scroll position ([#8625](https://github.com/googlechrome/lighthouse/pull/8625))
+* image-elements: drop spritesheet logic ([#8940](https://github.com/googlechrome/lighthouse/pull/8940))
+* refactor gather-runner control flow ([#8964](https://github.com/googlechrome/lighthouse/pull/8964))
+* logging: change log messages for gathering and trace retrieval ([#9029](https://github.com/googlechrome/lighthouse/pull/9029))
+* perf: speed up getNodePath ([#9055](https://github.com/googlechrome/lighthouse/pull/9055))
+* performance-budget: add MP-FID as budget metric ([#8905](https://github.com/googlechrome/lighthouse/pull/8905))
+* runner: fix missing timing properties in the browser ([#9072](https://github.com/googlechrome/lighthouse/pull/9072))
+* tap-targets: don't exclude visible position absolute elements from audit ([#7778](https://github.com/googlechrome/lighthouse/pull/7778))
```
brendankenny: can you throw
- `gather-runner: treat NO_FCP as a pageLoadError `
- `gather-runner: always reset scroll position`
- `logging: change log messages for gathering and trace retrieval`
- `tap-targets: don't exclude visible position absolute elements from audit`

at the top? They're probably most significant for most users

changelog.md
```diff
@@ -1,3 +1,76 @@
+<a name="5.1.0"></a>
+# 5.1.0 (2019-05-30)
+[Full Changelog](https://github.com/googlechrome/lighthouse/compare/v5.0.0...v5.1.0)
+
+## New contributors!
+
+Thanks to our new contributors 👽🐷🐰🐯🐻! 
+Beng Eu <beng@bengeu.com>
+David Boyne <david.boyne@bglgroup.co.uk>
```
brendankenny: this is a thing we should probably change about the release scripts. Even though it's in the git log, most people probably don't want their email in our release notes. In the past we would have done `@thebengeu and @boyney123` for these. We *could* include the user.name from the commits if folks want to, though.

(the release script might have to hit the github api for those? unless they're adding some commit metadata on squash in there somewhere)

changelog.md
brendankenny: oh, actually, the scroll one was reverted. Maybe don't put `gather-runner: always reset scroll position` up top and add a `(reverted)` or something so it's not confusing?

changelog.md
patrickhulce: Hm yeah fair point. 

I always assumed the emails in my GitHub commits are there for spreading my exposure to recruiter spam! 😆

brendankenny:
all good?

PR #9089, opened 2019-05-30
deps: update bundlesize, add more budgets

brendankenny:
bundlesize has been having a bunch of issues in CI lately, and now there's a moderate security alert in one of its deps, so let's try bumping it to latest.

Unfortunately it brings in a rather heavyweight dep itself in a native brotli encoding library, but development seems to be picking back up again (https://github.com/siddharthkp/bundlesize/issues/306) and that'll soon be optional if the node version has native brotli support.

Also added budgets for the other bundles as they vary quite a bit in size due to things included or not included in them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
     "babel-plugin-syntax-object-rest-spread": "^6.13.0",
     "brfs": "^1.6.1",
     "browserify": "^16.2.3",
-    "bundlesize": "^0.14.4",
+    "bundlesize": "^0.17.2",
     "chalk": "^2.4.1",
     "codecov": "^3.2.0",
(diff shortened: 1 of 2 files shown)


brendankenny:
If this doesn't work, we could also try out bundlewatch, which has a pretty nice details page for the bundle sizes



but doesn't have such a large community using it. https://bundlewatch.io/#/getting-started/using-bundlewatch

brendankenny:
well, the initial api call is still 500ing...I think that's when it's trying to query the bundlesize from the parent commit (hence the `-null` in the status).

We need to update anyways and I think the new status message is improved (even better when it can start querying master so we get some update on size change in that PR), so still worth landing.

PR #9109, opened 2019-06-03
core(cli): accept flags from path

patrickhulce:
**Summary**
In order to easily accept an LH settings object over in Lighthouse CI, we need a way to easily passthrough the object. This can be done with config today but you miss out on some nice features like being able to use `--budgets-path`, set the `--output-path`, etc. 

This is a low-cost way to enable consumers of the LH CLI that run in a child process programmatic access to all our CLI flags without manually (and error-pronely) constructing an argument list.

### DIFF
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -89,6 +89,7 @@ function getFlags(manualArgv) {
         ],
         'Configuration:')
       .describe({
+        'cli-flags-path': 'The path to a JSON file that contains the desired CLI flags to apply. Flags specified at the command line will still override the file-based ones.',
         // We don't allowlist specific locales. Why? So we can support the user who requests 'es-MX' (unsupported) and we'll fall back to 'es' (supported)
         'locale': 'The locale/language the report should be formatted in',
         'enable-error-reporting':
--- a/lighthouse-cli/cli-flags.js
+++ b/lighthouse-cli/cli-flags.js
@@ -62,6 +62,9 @@ function getFlags(manualArgv) {
           'lighthouse <url> --only-categories=performance,pwa',
           'Only run specific categories.')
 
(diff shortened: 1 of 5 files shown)


lighthouse-cli/cli-flags.js
```diff
@@ -62,6 +62,9 @@ function getFlags(manualArgv) {
           'lighthouse <url> --only-categories=performance,pwa',
           'Only run specific categories.')
 
+      // Accept a file for all of these flags.
+      .config('cli-settings-path')
```
brendankenny: can you add a description specifying that other flags will override this one? (e.g. `--throttling-method=provided` on the command line will override `"throttling-method": "devtools"` in the settings file)

brendankenny: also, bikeshed time: should we call it something like `cli-flags-path`?

brendankenny: also, copy over to readme :)

patrickhulce: > can you add a description specifying that other flags will override this one?

Yes, test added too! 👍 

> cli-flags-path

WFM!

> also, copy over to readme 

This feels close to an autopopulate script too if only we made changes more frequently 🤔

paulirish:
wfm! 👍

lighthouse-cli/cli-flags.js
brendankenny: > This feels close to an autopopulate script too if only we made changes more frequently 🤔

definitely, and it's a little silly we have it at all, but I know I've at least found it useful in the past, so....

patrickhulce:
> I don't entirely understand how you're going to use it

Basically I want to use the existing logic for `outputPath` so I can't use the node integration, but I also want to pass a budget directly as JSON, which I can't really do from CLI flags :)

PR #9112, opened 2019-06-03
docs: add instructions for testing with self-signed certificate

brendankenny:
With #8865, `--chrome-flags="--ignore-certificate-errors"` now works for self-signed certificates. Adds a note to the docs about this so we have a place to point to in the future.

see also #6655, #7292, #7574


docs/readme.md
```diff
@@ -115,6 +115,17 @@ instance with an open debugging port.
 1. Navigate to your site and log in.
 1. In a separate terminal tab, run `lighthouse http://mysite.com --port port-number` using the port number from chrome-debug.
 
+## Testing on a site with an untrusted certificate
+
+When testing a site with an untrusted certificate, Chrome will be unable to load the page and so the Lighthouse report will mostly contain errors.
+
+If this certificate **is one you control** and is necessary for development (for instance, `localhost` with a self-signed certificate for local HTTP/2 testing), you can
```
patrickhulce: IMO a tad ambiguous if you need to do one or both of these currently, is that intentional to hope they go the cert route? ;)

PR #9121, opened 2019-06-04
core(gather-runner): convert PAGE_HUNG to non-fatal runtimeError

patrickhulce:
**Summary**
This should be last step in making all our page-load focused runtimeErrors non-fatal and exiting with an error code. This PR also cleans up some loose ends from #8865 where INSECURE_DOCUMENT_REQUEST was lying around.

**Notable Changes**

- PAGE_HUNG is now in the same boat as NO_FCP, it gets special cased as a `pageLoadError` and is no longer fatal.
- PAGE_HUNG does not exit with its own special exit code, it's just 1 like all the other runtimeErrors. Smoketests are now asserted via `{runtimeError: {code: '...'}}`
- If a pass fails with a `pageLoadError`, the rest of the passes are skipped and the rest of baseArtifacts are not populated. Necessary for recovery from PAGE_HUNG and just made sense, see https://github.com/GoogleChrome/lighthouse/pull/8865#issuecomment-497481383 step 3. 
- `clearDataForOrigin` will not throw PROTOCOL_TIMEOUT if it can't complete, it will just warn and move on. This is necessary for trying to recover from PAGE_HUNG in a non-fatal way, and it was one of the flagged non-responsive methods in our PROTOCOL_TIMEOUT umbrella anyhow.

**Related Issues/PRs**
followup according to https://github.com/GoogleChrome/lighthouse/pull/8865#issuecomment-497481383 (step 2 of the process)

### DIFF
--- a/clients/extension/scripts/popup.js
+++ b/clients/extension/scripts/popup.js
@@ -13,14 +13,7 @@
  * Lighthouse itself, mapped to more useful strings to report to the user.
  */
 const NON_BUG_ERROR_MESSAGES = {
-  // The user tries to review an error page or has network issues
-  'ERRORED_DOCUMENT_REQUEST': 'Unable to load the page. Please verify the url you ' +
-      'are trying to review.',
-  'FAILED_DOCUMENT_REQUEST': 'Unable to load the page. Please verify the url you ' +
-      'are trying to review.',
   'DNS_FAILURE': 'DNS servers could not resolve the provided domain.',
-  'INSECURE_DOCUMENT_REQUEST': 'The URL you have provided does not have a valid' +
-      ' SSL certificate.',
   'INVALID_URL': 'Lighthouse can only audit URLs that start' +
       ' with http:// or https://.',
 
--- a/lighthouse-cli/test/smokehouse/smokehouse.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse.js
@@ -16,8 +16,6 @@ const log = require('lighthouse-logger');
 const {collateResults, report} = require('./smokehouse-report.js');
 
 const PROTOCOL_TIMEOUT_EXIT_CODE = 67;
-const PAGE_HUNG_EXIT_CODE = 68;
-const INSECURE_DOCUMENT_REQUEST_EXIT_CODE = 69;
 const RETRIES = 3;
 
 /**
@@ -45,9 +43,6 @@ function resolveLocalOrCwd(payloadPath) {
  */
 function isUnexpectedFatalResult(exitCode, outputPath) {
   return exitCode !== 0
-    // These runtime errors are currently fatal but "expected" runtime errors we are asserting against.
-    && exitCode !== PAGE_HUNG_EXIT_CODE
-    && exitCode !== INSECURE_DOCUMENT_REQUEST_EXIT_CODE
     // On runtime errors we exit with a error status code, but still output a report.
     // If the report exists, it wasn't a fatal LH error we need to abort on, it's one we're asserting :)
     && !fs.existsSync(outputPath);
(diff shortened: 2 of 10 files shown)


lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -116,15 +111,11 @@ function runLighthouse(url, configPath, isDebug) {
 
   let errorCode;
   let lhr = {requestedUrl: url, finalUrl: url, audits: {}};
-  if (runResults.status === PAGE_HUNG_EXIT_CODE) {
-    errorCode = 'PAGE_HUNG';
-  } else if (runResults.status === INSECURE_DOCUMENT_REQUEST_EXIT_CODE) {
-    errorCode = 'INSECURE_DOCUMENT_REQUEST';
-  } else {
+  if (fs.existsSync(outputPath)) {
```
brendankenny: should we turn `isUnexpectedFatalResult` into `exitCode !== 0 || !fs.existsSync(outputPath)` and this will always exist by this point? Seems like no output is an unexpected failure, even if status code is 0

lighthouse-core/gather/gather-runner.js
```diff
@@ -552,6 +552,9 @@ class GatherRunner {
         const passResults = await GatherRunner.runPass(passContext);
         Object.assign(artifacts, passResults.artifacts);
 
+        // If we encountered a pageLoadError, don't try to keep loading the page in future passes.
+        if (passResults.pageLoadError) break;
```
brendankenny: this gets a bit weird because we don't populate but still return them. All the other artifacts are either undefined, an Error, or the artifact, but we haven't until now done half-finished artifacts.

Seems ok for now, though

lighthouse-core/lib/lh-error.js
```diff
@@ -172,14 +172,6 @@ const ERRORS = {
     message: UIStrings.pageLoadFailedWithStatusCode,
     lhrRuntimeError: true,
   },
-  /* Used when security error prevents page load.
```
brendankenny: re: discussion in #8865, I don't think we should get rid of this LighthouseError. I think it should be repurposed as the `lhr.runtimeError` in case of interstitial detected instead of it being `FAILED_DOCUMENT_REQUEST` bubbling up from one of the audits

clients/extension/scripts/popup.js
```diff
@@ -13,14 +13,7 @@
  * Lighthouse itself, mapped to more useful strings to report to the user.
  */
 const NON_BUG_ERROR_MESSAGES = {
-  // The user tries to review an error page or has network issues
-  'ERRORED_DOCUMENT_REQUEST': 'Unable to load the page. Please verify the url you ' +
-      'are trying to review.',
-  'FAILED_DOCUMENT_REQUEST': 'Unable to load the page. Please verify the url you ' +
```
brendankenny: this should still be possible, shouldn't it? e.g. for https://expired.badssl.com/

lighthouse-cli/test/smokehouse/smokehouse.js
patrickhulce: Yeah good call 👍

clients/extension/scripts/popup.js
patrickhulce: Not as a fatal error. The popup only catches fatal errors and replaces the bug prompt with these messages. `FAILED_DOCUMENT_REQUEST` hasn't been fatal for quite some time.

lighthouse-core/lib/lh-error.js
patrickhulce: reverted, followup can bring interstitial detection magic back to make this INSECURE_DOCUMENT_REQUEST again :)

Issue #9126, opened 2019-06-05
i18n, performance-budgets columns not translated


exterkamp:
`audits/performance-budgets.js` is mostly translated, but the `columnHeaders` are not.

exterkamp:
`resource-summary` has same problem 😉

PR #9127, opened 2019-06-05
i18n: perf-budget and res-summary column headers

exterkamp:
**Summary**
Noticed this when snooping around for i18n stuff.

fixes: #9126

### DIFF
--- a/lighthouse-core/audits/performance-budget.js
+++ b/lighthouse-core/audits/performance-budget.js
@@ -20,6 +20,14 @@ const UIStrings = {
     =1 {1 request}
     other {# requests}
    }`,
+  /** Table column header for classifying website resources by type, example: "Scripts", "Third-Party", "Stylesheet". */
+  columnResourceType: 'Resource Type',
+  /** Table column header for the count of requests made per resource type. */
+  columnRequests: 'Requests',
+  /** Table column header for the sizes of all network requests of a specific type. */
+  columnTransferSize: 'Transfer Size',
+  /** Table column header for how displaying how much larger the size of network requests were than a predetermined budget. */
+  columnOverBudget: 'Over Budget',
 };
 
 const str_ = i18n.createMessageInstanceIdFn(__filename, UIStrings);
@@ -130,11 +138,11 @@ class ResourceBudget extends Audit {
 
     /** @type { LH.Audit.Details.Table['headings'] } */
     const headers = [
-      {key: 'label', itemType: 'text', text: 'Resource Type'},
-      {key: 'requestCount', itemType: 'numeric', text: 'Requests'},
-      {key: 'size', itemType: 'bytes', text: 'Transfer Size'},
+      {key: 'label', itemType: 'text', text: str_(UIStrings.columnResourceType)},
+      {key: 'requestCount', itemType: 'numeric', text: str_(UIStrings.columnRequests)},
+      {key: 'size', itemType: 'bytes', text: str_(UIStrings.columnTransferSize)},
       {key: 'countOverBudget', itemType: 'text', text: ''},
-      {key: 'sizeOverBudget', itemType: 'bytes', text: 'Over Budget'},
+      {key: 'sizeOverBudget', itemType: 'bytes', text: str_(UIStrings.columnOverBudget)},
     ];
 
     return {
--- a/lighthouse-core/audits/performance-budget.js
+++ b/lighthouse-core/audits/performance-budget.js
@@ -20,12 +20,6 @@ const UIStrings = {
     =1 {1 request}
     other {# requests}
    }`,
-  /** Table column header for classifying website resources by type, example: "Scripts", "Third-Party", "Stylesheet". */
-  columnResourceType: 'Resource Type',
-  /** Table column header for the count of requests made per resource type. */
-  columnRequests: 'Requests',
-  /** Table column header for the sizes of all network requests of a specific type. */
-  columnTransferSize: 'Transfer Size',
   /** Table column header for how displaying how much larger the size of network requests were than a predetermined budget. */
   columnOverBudget: 'Over Budget',
 };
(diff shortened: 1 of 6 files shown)


lighthouse-core/audits/performance-budget.js
```diff
@@ -20,6 +20,14 @@ const UIStrings = {
     =1 {1 request}
     other {# requests}
    }`,
+  /** Table column header for classifying website resources by type, example: "Scripts", "Third-Party", "Stylesheet". */
+  columnResourceType: 'Resource Type',
+  /** Table column header for the count of requests made per resource type. */
+  columnRequests: 'Requests',
+  /** Table column header for the sizes of all network requests of a specific type. */
+  columnTransferSize: 'Transfer Size',
```
patrickhulce: I faintly recall having a discussion in a PR about whether these should be generic in `i18n.js` or not, but I guess we ended up doing nothing instead 😆 

Anyhoo, transfer size elsewhere is `columnSize` in `i18n.js` maybe we standardize on that and reuse?

exterkamp: Made use of `i18n` columns.  I will defer to @khempenius /@hwikyounglee . to decide if they want to keep "Transfer Size" vs. just "Size".  I kept it just to be no-op here, also `resource-summary` also uses "Transfer Size".

lighthouse-core/audits/performance-budget.js
```diff
@@ -20,6 +20,8 @@ const UIStrings = {
     =1 {1 request}
     other {# requests}
    }`,
+  /** Table column header for how displaying how much larger the size of network requests were than a predetermined budget. */
```
exterkamp: This desc is different than the ones in `i18n`.  Which format do you guys prefer?

lighthouse-core/audits/performance-budget.js
khempenius: I'm in favor of keeping it as "Transfer Size". There's a big difference between "transfer size" and "resource size" and I think this is the only place in the UI where we distinguish which interpretation of "size" this uses.

hwikyounglee: Being specific, i.e. 'Transfer Size', SGTM too

lighthouse-core/audits/performance-budget.js
patrickhulce: This seems a bit better to me to clarify the fact that it's a table column front and center 👍

lighthouse-core/audits/performance-budget.js
patrickhulce: All things being equal I definitely prefer `Transfer Size` too, the mixture of resource/transfer and column width restrictions make sense elsewhere but if we can let's :)

exterkamp: SGTM will keep as is 😄

lighthouse-core/audits/performance-budget.js
exterkamp: I personally find this format _less_ specific in general.  I also don't really see this as a huge problem because a lot fo files use both 🤷‍♂ 

Worth unifying?

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -56,6 +56,12 @@ const UIStrings = {
   columnWastedMs: 'Potential Savings',
   /** Label for the time spent column in data tables, entries will be the number of milliseconds spent during a particular activity */
   columnTimeSpent: 'Time Spent',
+  /** Label for the type of network resource column in data tables, entries will be enumerated kinds of network requests e.g. "Scripts", "Third-Party", "Stylesheet".*/
```
brendankenny: not sure when these descriptions switched formats above, but the ones below (like `totalResourceType`) seem like they'd read a lot clearer to someone with no context (like, what's a "type of network resource column in data tables"?).

e.g.
`Label for a row in a data table; entries will be types of resources loaded over the network, e.g. "Scripts", "Third-Party", "Stylesheet".`
seems better

lighthouse-core/lib/i18n/i18n.js
```diff
@@ -56,6 +56,12 @@ const UIStrings = {
...
+  columnResourceType: 'Resource Type',
+  /** Label for the count of network requests column in data tables, entries will be the number of network requests done by a webpage.*/
+  columnRequests: 'Requests',
+  /** Label for the transfer size of network requests column in data tables, entries will be the total bytes transferred of a group of network requests.*/
```
brendankenny: should be something like "total number of bytes transferred to load a set of files" or something like that?

lighthouse-core/audits/performance-budget.js
brendankenny: wait, @exterkamp, didn't you write this? Are you suggesting it and then shooting it down? :)

khempenius:
```suggestion
  /** Table column header displaying: how much greater the quantity of network requests were than a predetermined budget, how much larger the size of network requests were than a predetermined budget. */
```

lighthouse-core/lib/i18n/i18n.js
exterkamp: The ones below are different.  Those are `rows` in the perf-budgets table, not the columns.  That's why they're different.  But I agree they're more clear.

lighthouse-core/audits/performance-budget.js
exterkamp: I wrote it then had regrets, wondered which format was better #ragrets

exterkamp: Integrated this with the column format from `i18n.js`

exterkamp: I like that we landed on the format in `18n.js`.  Resolving.

lighthouse-core/audits/performance-budget.js
```diff
@@ -20,6 +20,8 @@ const UIStrings = {
     =1 {1 request}
     other {# requests}
    }`,
+  /** Label for a column in a data table; entries will be how much greater the quantity of network requests were than a predetermined budget, or how much larger the size of network requests were than a predetermined budget.*/
```
brendankenny: I think this is probably overkill :) What about combining the last two parts:
```suggestion
  /** Label for a column in a data table; entries will be how much the quantity or size of network requests exceeded a predetermined budget.*/
```

exterkamp:
Please Googlebot.  Please.

PR #9129, opened 2019-06-05
core(tracehouse): merge trace-of-tab and tracing-processor

patrickhulce:
**Summary**
- Merges trace-of-tab and tracing-processor into `trace-processor`, blame for `trace-of-tab` should be preserved by squashing down into 2 separate commits at the end.
- Removes all the dependencies from trace-processor into the rest of LH. Really this is just LHError. Behavior has been preserved by an overrideable method that constructs the errors it should throw.

Individual commits hopefully aid in reviewing as the final diff is a overinflated, final PR will be squashed down to just 2, one for file renames and one with all the changes.

**Related Issues/PRs**
Step 2 from the plan outlined in https://github.com/GoogleChrome/lighthouse/pull/8956#issue-278881171


lighthouse-core/computed/trace-of-tab.js
```diff
@@ -6,7 +6,34 @@
...
+const LHError = require('../lib/lh-error.js');
+const TraceProcessor = require('../lib/tracehouse/trace-processor.js');
+
+// TraceProcessor throws generic errors, but we'd like our special localized and code-specific LHError
+// objects to be thrown instead.
+class LHTraceProcessor extends TraceProcessor {
+  /**
+   * @return {Error}
+   */
+  static createNoNavstartError() {
```
Hoten: this is a problem I've never seen before (translating error codes from one domain to another). What made you go with this method override approach vs try/catch/replacing (or .... drum roll.... catch and release)

patrickhulce: Catch and rethrow would be dependent on the specific string used inside `trace-processor`. I've been burned by such an approach before, so wanted to avoid it.

This method also makes it very clear which important errors can be thrown by the trace processor and gives other consumers of tracehouse the ability to override if they wish too.

lighthouse-core/test/lib/tracehouse/trace-processor-test.js
```diff
@@ -0,0 +1,444 @@
+/**
```
brendankenny: is bisect going to be preserved for any of the things in here?

lighthouse-core/lib/tracehouse/trace-processor.js
```diff
@@ -17,18 +32,75 @@ const SCHEDULABLE_TASK_TITLE_ALT2 = 'ThreadControllerImpl::DoWork';
...
-class TraceProcessor {
   /**
    * There should *always* be at least one top level event, having 0 typically means something is
    * drastically wrong with the trace and we should just give up early and loudly.
    *
    * @param {LH.TraceEvent[]} events
    */
   static assertHasToplevelEvents(events) {
-    const hasToplevelTask = events.some(TraceProcessor.isScheduleableTask);
+    const hasToplevelTask = events.some(this.isScheduleableTask);
```
brendankenny: any reason to change all these? They should all still work, shouldn't they? Typically we've used the ClassName.method() approach to call out static methods

patrickhulce: `this` is the only thorny part of the inheritance approach since we *must* invoke it with `this` if the overrideable errors are to work.

Rather than split between those that need it today and those that don't, I moved them all to `this` for consitency so a future implementer who might copy/paste doesn't accidentally use static :)

Different proposal though that helps solve this?

lighthouse-core/test/lib/tracehouse/trace-processor-test.js
patrickhulce: not really, the file was moved from `trace-of-tab` so that history should be there but since I indented everything the blame will be pretty much wiped out

PR #9130, opened 2019-06-06
tests(smokehouse): always assert lhr.runtimeError

brendankenny:
follow up to #9121 and part of the discussion from #8865. Adds two additional checks to smokehouse:
- if Lighthouse exits with a non-zero exit code the lhr must have a `runtimeError` (or if there's a `runtimeError` it must have had a non-zero exit code). This just double checks the change in https://github.com/GoogleChrome/lighthouse/commit/ff4f486298b5ac1db583a85bfc8a45c9689de5fb#diff-8035764042a5861297234f1e1b922ca7R221 (and should always be 👍)
- if there's a `runtimeError` in the run's LHR there *must* be a `runtimeError` in the expectations (slightly more strict than usual where something left out of the expectations isn't tested). We're asking everyone to keep an eye on `runtimeError` to spot problems, so our tests should be expecting them if one is in there :)

Since we weren't checking `runtimeError` before, it means [`error-expectations.js`](https://github.com/GoogleChrome/lighthouse/blob/9ebae67202a6df281781024ffa911a27643e629d/lighthouse-cli/test/smokehouse/error-expectations.js) needs to be updated because those runs don't actually include a `runtimeError` (because the audit being run is a metric and doesn't need anything but base artifacts -- mentioned in the last bullet point of https://github.com/GoogleChrome/lighthouse/pull/8865#pullrequestreview-243000692).

Also does a drive-by fix to delete the smokehouse test artifact files after use. I was up to several hundred MBs of them in `.tmp/` :)

### DIFF
--- a/lighthouse-cli/test/smokehouse/error-config.js
+++ b/lighthouse-cli/test/smokehouse/error-config.js
@@ -14,6 +14,8 @@ module.exports = {
     maxWaitForLoad: 5000,
     onlyAudits: [
       'first-contentful-paint',
+      // TODO: this audit collects a non-base artifact, allowing the runtimeError to be collected.
+      'content-width',
     ],
   },
 };
--- a/lighthouse-cli/test/smokehouse/seo/expectations.js
+++ b/lighthouse-cli/test/smokehouse/seo/expectations.js
@@ -210,6 +210,10 @@ module.exports = [
       // Note: most scores are null (audit error) because the page 403ed.
       requestedUrl: BASE_URL + 'seo-failure-cases.html?status_code=403',
       finalUrl: BASE_URL + 'seo-failure-cases.html?status_code=403',
+      runtimeError: {
+        code: 'ERRORED_DOCUMENT_REQUEST',
+        message: /Status code: 403/,
+      },
       audits: {
         'http-status-code': {
           score: 0,
--- a/lighthouse-cli/test/smokehouse/smokehouse-report.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse-report.js
@@ -127,6 +127,7 @@ function makeComparison(name, actualResult, expectedResult) {
  */
 function collateResults(actual, expected) {
   // If actual run had a runtimeError, expected *must* have a runtimeError.
+  // Relies on the fact that an `undefined` argument to makeComparison() can only match `undefined`.
   const runtimeErrorAssertion = makeComparison('runtimeError', actual.lhr.runtimeError,
       expected.lhr.runtimeError);
 
--- a/lighthouse-cli/test/smokehouse/smokehouse-report.js
+++ b/lighthouse-cli/test/smokehouse/smokehouse-report.js
@@ -126,6 +126,10 @@ function makeComparison(name, actualResult, expectedResult) {
  * @return {Smokehouse.Comparison[]}
  */
 function collateResults(actual, expected) {
+  // If actual run had a runtimeError, expected *must* have a runtimeError.
+  const runtimeErrorAssertion = makeComparison('runtimeError', actual.lhr.runtimeError,
+      expected.lhr.runtimeError);
+
   /** @type {Smokehouse.Comparison[]} */
   let artifactAssertions = [];
   if (expected.artifacts) {
@@ -161,6 +165,7 @@ function collateResults(actual, expected) {
       expected: expected.lhr.finalUrl,
       equal: actual.lhr.finalUrl === expected.lhr.finalUrl,
     },
+    runtimeErrorAssertion,
     ...artifactAssertions,
     ...auditAssertions,
   ];
@@ -194,7 +199,8 @@ function reportAssertion(assertion) {
   } else {
     if (assertion.diff) {
       const diff = assertion.diff;
-      const fullActual = JSON.stringify(assertion.actual, null, 2).replace(/\n/g, '\n      ');
+      const fullActual = String(JSON.stringify(assertion.actual, null, 2))
+          .replace(/\n/g, '\n      ');
       const msg = `
   ${log.redify(log.cross)} difference at ${log.bold}${diff.path}${log.reset}
               expected: ${JSON.stringify(diff.expected)}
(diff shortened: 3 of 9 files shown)


lighthouse-cli/test/smokehouse/seo/expectations.js
```diff
@@ -210,6 +210,10 @@ module.exports = [
       // Note: most scores are null (audit error) because the page 403ed.
       requestedUrl: BASE_URL + 'seo-failure-cases.html?status_code=403',
       finalUrl: BASE_URL + 'seo-failure-cases.html?status_code=403',
+      runtimeError: {
+        code: 'ERRORED_DOCUMENT_REQUEST',
+        message: /Status code: 403/,
+      },
```
brendankenny: has to be asserted now

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -194,7 +199,8 @@ function reportAssertion(assertion) {
   } else {
     if (assertion.diff) {
       const diff = assertion.diff;
-      const fullActual = JSON.stringify(assertion.actual, null, 2).replace(/\n/g, '\n      ');
+      const fullActual = String(JSON.stringify(assertion.actual, null, 2))
```
brendankenny: fixes the case when `assertion.actual` is `undefined`, which surprisingly I guess we've never encountered before because this throws on `undefined.replace(...)`

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -88,27 +91,34 @@ function runLighthouse(url, configPath, isDebug) {
...
-    console.error(`Lighthouse run failed to produce a report and exited with ${runResults.status}. stderr to follow:`); // eslint-disable-line max-len
+    console.error(`Lighthouse run failed to produce a report and exited with ${exitCode}. stderr to follow:`); // eslint-disable-line max-len
     console.error(runResults.stderr);
-    process.exit(runResults.status);
+    process.exit(exitCode);
   }
 
+  /** @type {LH.Result} */
   const lhr = JSON.parse(fs.readFileSync(outputPath, 'utf8'));
+  const artifacts = assetSaver.loadArtifacts(artifactsDirectory);
```
brendankenny: we weren't loading the trace(s) and devtoolsLogs before, and I guess it's possible someone will want them at some point :) Regardless, switching to `loadArtifacts` is the better solution

lighthouse-core/lib/asset-saver.js
```diff
@@ -33,9 +33,9 @@ const devtoolsLogSuffix = '.devtoolslog.json';
  * Load artifacts object from files located within basePath
  * Also save the traces to their own files
  * @param {string} basePath
- * @return {Promise<LH.Artifacts>}
+ * @return {LH.Artifacts}
  */
-async function loadArtifacts(basePath) {
+function loadArtifacts(basePath) {
```
brendankenny: fix to keep smokehouse's `runLighthouse` sync. Since we dropped the streaming JSON loader in #6099, this doesn't need to be async anyways

brendankenny:
added an audit that needs more than the base artifacts to the errors config so that the gathering `runtimeError`s bubble up to the lhr for now

lighthouse-cli/test/smokehouse/error-config.js
```diff
@@ -14,6 +14,8 @@ module.exports = {
     maxWaitForLoad: 5000,
     onlyAudits: [
       'first-contentful-paint',
+      // TODO: this audit collects a non-base artifact, allowing the runtimeError to be collected.
```
patrickhulce: heh, yikes I forgot about that!

lighthouse-cli/test/smokehouse/smokehouse-report.js
```diff
@@ -126,6 +126,11 @@ function makeComparison(name, actualResult, expectedResult) {
  * @return {Smokehouse.Comparison[]}
  */
 function collateResults(actual, expected) {
+  // If actual run had a runtimeError, expected *must* have a runtimeError.
+  // Relies on the fact that an `undefined` argument to makeComparison() can only match `undefined`.
+  const runtimeErrorAssertion = makeComparison('runtimeError', actual.lhr.runtimeError,
```
patrickhulce: doh! everything is off of `lhr` now we should just iterate through the keys

lighthouse-cli/test/smokehouse/smokehouse.js
```diff
@@ -88,27 +91,34 @@ function runLighthouse(url, configPath, isDebug) {
...
+    rimraf.sync(artifactsDirectory);
   }
 
-  // Artifacts are undefined if they weren't written to disk (e.g. if there was an error).
-  let artifacts;
-  try {
-    artifacts = JSON.parse(fs.readFileSync(`${artifactsDirectory}/artifacts.json`, 'utf8'));
-  } catch (e) {}
+  // There should either be both an error exitCode and a lhr.runtimeError or neither.
+  if (!exitCode !== !lhr.runtimeError) {
```
patrickhulce: nit: there's a lot of `!` to parse :) 

`Boolean(exitCode) !== Boolean(lhr.runtimeError)`?

lighthouse-core/lib/asset-saver.js
patrickhulce: `runLighthouse` needs to become `async` for #8962 (which still needs review btw :P)

brendankenny: > runLighthouse needs to become async for #8962 (which still needs review btw :P)

ha, fair. Nothing is async in it anyways, though, so saves us a microtask or two :)

lighthouse-cli/test/smokehouse/smokehouse-report.js
brendankenny: > everything is off of lhr now we should just iterate through the keys

agreed, though I really want to make sure we check `runtimeError`, even if it's not in the expectations

lighthouse-cli/test/smokehouse/smokehouse.js
brendankenny: > nit: there's a lot of `!` to parse :)
>
> `Boolean(exitCode) !== Boolean(lhr.runtimeError)`?

obviously we should go with `if (!!(!!exitCode !== !!lhr.runtimeError))`

(your suggestion sounds good :)

